PR #4496, opened 2015-12-14
[minor] raise SetupWizardComplete error

anandpdoshi:
- Depends on frappe/frappe#1459


nabinhait:
lets allow to come these errors, will help to fix the underlines errors.

Issue #4497, opened 2015-12-14
Time Log only calculates based on Minutes rather then seconds


partzsch:
Hi all,

due to request of neilLasrado I´m raising the following

https://discuss.erpnext.com/t/time-log-timesheet-only-calculates-to-minutes/8969/5

when filling in time logs the resulting hours only get calculated down to Minutes rather then seconds 

had a look @ https://github.com/frappe/erpnext/blob/develop/erpnext/projects/doctype/time_log/time_log.py

as I understand with my limited knowledge of Python, this should be the responsible snippit at line 112

```
def calculate_total_hours(self):
if self.to_time and self.from_time:
    from frappe.utils import time_diff_in_seconds
    self.hours = flt(time_diff_in_seconds(self.to_time, self.from_time)) / 3600
```

this is accomplished by https://github.com/frappe/frappe/blob/develop/frappe/utils/data.py1

```
def time_diff_in_seconds(string_ed_date, string_st_date):
return time_diff(string_ed_date, string_st_date).total_seconds()
```

it seems to meet our needs so far as it looks for seconds ...

but https://github.com/frappe/erpnext/blob/develop/erpnext/projects/doctype/time_log/time_log.js

```
// set hours if to_time is updated
frappe.ui.form.on("Time Log", "to_time", function(frm) {
if(frm._setting_hours) return;
frm.set_value("hours", moment(cur_frm.doc.to_time).diff(moment(cur_frm.doc.from_time),
"minutes") / 60);
```

Is it right that this java script then brings our seconds down to minutes before inserting the value into the Field hours??

so wouldn be the solution to change https://github.com/frappe/erpnext/blob/develop/erpnext/projects/doctype/time_log/time_log.js

```
// set hours if to_time is updated
frappe.ui.form.on("Time Log", "to_time", function(frm) {
if(frm._setting_hours) return;
frm.set_value("hours", moment(cur_frm.doc.to_time).diff(moment(cur_frm.doc.from_time),
"seconds") / 3600);
```

??

partzsch:
can be closed as solution merged with #4520

PR #4517, opened 2015-12-19
Time Log now down to seconds

spa49:
according to:
https://github.com/frappe/erpnext/issues/4497
please find the following commit


anandpdoshi:
@partzsch too many commits for such a small change. Can you create a fresh branch, make this change and create another pull request?

Issue #4543, opened 2015-12-25
Creating variants with Naming Series: Item Code is mandatory because Item is not automatically numbered


gabtzi:
Hello,

I'm trying to setup some items in ERPNext but I'd like the item code to be created based on Naming Series for better management. 

In doing so I created the naming series called `SUP-` and the item code `SUP-00003` was automatically created just fine when I created the base/parent item. The bug came when I tried to create a variant for that item for one attribute. When I try to save the item it won't let me save it complaining that 
`Item Code is mandatory because Item is not automatically numbered`

![variants problem](https://cloud.githubusercontent.com/assets/11225818/12003145/a002b972-ab1c-11e5-915f-e3ad317561f1.JPG)

I think that in such cases one of the rest should happen:
1. Use parent Item Code and append the attribute identifier code with value (check below)
2. User parent item code and append a numeric counter
3. Simply increment the parent's Naming Series 
4. At last let the user select an Item Code by himself. 

Regarding 1. 
While creating the item attribute I noticed that there is no field like an Item Attribute Code but only the name exists. In my opinion Item Attributes should also have a unique Item Attribute Code value or Abbreviation to separate them because we might want them to be looking the same but refer to different things. 

For example the sizes in shoes and shirts and jackets and trousers are completely different but we might want their name to appear as just Size and not `Shoe Size`  or `Shirt Size` for example. Imagine the complications if an employee were to mess things up when adding the variant's attribute values because he had to browse through 200 different values. 

I appreciate any input regarding this

EDIT:
This happens when in attribute the "Numeric Values" checkbox is selected.

gabtzi:
For anyone who might face the same issue as me 

In my case and for my needs and inspired by Magento I worked around my issue by altering the code in file: 
`erpnext/stock/doctype/item/item.py`

so that the code I added will handle the creation of the new code by appending to the parent product item code the concatenated attribute values with dashes.
```python
elif frappe.db.get_default("item_naming_by")=="Naming Series" and self.variant_of:
            itemCodeSuffix = ""
            for attribute in self.attributes:
                itemCodeSuffix += "-" + str(attribute.attribute_value)
            self.item_code = str(self.variant_of) + itemCodeSuffix
```

So for product `SERIES-00003`  and for `NumericAttribute1=5`,  `Color=RED` the new item code would be `SERIES-00003-5-RED`

So the file now in line 32 looks like:
```python
	def autoname(self):
[...]

PR #4550, opened 2015-12-28
 Fix for Issue #4543: Automatic item_code generation for variants when parent item is named after a Naming Series

gabtzi:
Fix for Issue #4543: Automatic item_code generation for variants when parent item is named after a Naming Series. 

This will handle the creation of the new code by appending to the parent product item code the concatenated attribute values with dashes. So for product `SERIES-00003` and for `NumericAttribute1=5`, `Color=RED` the new item code would be `SERIES-00003-5-RED`

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -30,14 +30,15 @@ class Item(WebsiteGenerator):
 		self.get("__onload").sle_exists = self.check_if_sle_exists()
 
 	def autoname(self):
-		if frappe.db.get_default("item_naming_by")=="Naming Series" and not self.variant_of:
-			from frappe.model.naming import make_autoname
-			self.item_code = make_autoname(self.naming_series+'.#####')
-		elif frappe.db.get_default("item_naming_by")=="Naming Series" and self.variant_of:
-			item_code_suffix = ""
-			for attribute in self.attributes:
-				item_code_suffix += "-" + str(attribute.attribute_value)
-			self.item_code = str(self.variant_of) + item_code_suffix
+		if frappe.db.get_default("item_naming_by")=="Naming Series":
+			if self.variant_of:
+				item_code_suffix = ""
+				for attribute in self.attributes:
+					item_code_suffix += "-" + str(attribute.attribute_value)
+				self.item_code = str(self.variant_of) + item_code_suffix
+			else:
+				from frappe.model.naming import make_autoname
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item.py
```diff
@@ -33,6 +33,11 @@ def autoname(self):
 		if frappe.db.get_default("item_naming_by")=="Naming Series" and not self.variant_of:
 			from frappe.model.naming import make_autoname
 			self.item_code = make_autoname(self.naming_series+'.#####')
+		elif frappe.db.get_default("item_naming_by")=="Naming Series" and self.variant_of:
```
rmehta: maybe add this in the earlier if so that it does not query twice:
```
if frappe.db.get_default("item_naming_by")=="Naming Series":
  if self.variant_of:
    ...
  else:
    ...
```

erpnext/stock/doctype/item/item.py
```diff
@@ -33,6 +33,11 @@ def autoname(self):
...
+			item_code_suffix = ""
+			for attribute in self.attributes:
+				item_code_suffix += "-" + str(attribute.attribute_value)
```
rmehta: Do we still use abbreviations?

```
"-" + str(attribute.abbr or attribute.attribute_value)
```

gabtzi: I was thinking about abbreviations but they don't exist in the Numeric Value attributes so I opted for using just the values because the attribute name could be really long.

erpnext/stock/doctype/item/item.py
```diff
@@ -30,9 +30,16 @@ def onload(self):
 		self.get("__onload").sle_exists = self.check_if_sle_exists()
 
 	def autoname(self):
-		if frappe.db.get_default("item_naming_by")=="Naming Series" and not self.variant_of:
-			from frappe.model.naming import make_autoname
-			self.item_code = make_autoname(self.naming_series+'.#####')
+		if frappe.db.get_default("item_naming_by")=="Naming Series":
+			if self.variant_of:
+				item_code_suffix = ""
```
nabinhait: Item code for variants can be changed manually. Hence, it should set auto_name **only if item_code is blank**. Also, when creating variant using UI, item_code is already set.

nabinhait: Fixed and merged

PR #4573, opened 2016-01-04
[fix] Stock Entry: calculate amount in javascript

anandpdoshi:


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -227,6 +227,56 @@ erpnext.stock.StockEntry = erpnext.stock.StockController.extend({
...
+			this.distribute_additional_costs();
+		}
+
+		// get the new changes
+		item = frappe.model.get_doc(item.doctype, item.name);
+
+		item.amount = flt(item.basic_amount + flt(item.additional_cost),
+			precision("amount", item));
+
+		item.valuation_rate = flt(flt(item.basic_rate)
```
nabinhait: Amount and valuation rate should be calculated for all rows, on trigger of basic_rate and qty.

anandpdoshi:
@nabinhait please review

PR #4601, opened 2016-01-11
Páginas de inicio de la traducción al español.

paulocoronado:
work in progress
Spanish translation


rmehta:
Thanks for starting! Can you add a line that this is "work in progress" until we complete it :)

PR #4618, opened 2016-01-14
Added additional cost to time log

neilLasrado:
- Added additional cost to time log
- Fixed issues in Time log billing amount calculation and billing amount calculation while making time log batch


rmehta:
@neilLasrado patch?

neilLasrado:
@rmehta Not required as i added a new field and the calculation problems were client side.

PR #4635, opened 2016-01-19
Add localization in the on_submit event

MaxMorais:


MaxMorais:
PS: Without the localization, this method fail in all other languages, because the <a> have a bad URL

PR #4652, opened 2016-01-20
Update item_list.js - Filter improvements

RicardoJohann:
Added a default filter and fixed filter when clicked on "activate"

### DIFF
--- a/erpnext/stock/doctype/item/item_list.js
+++ b/erpnext/stock/doctype/item/item_list.js
@@ -1,6 +1,7 @@
 frappe.listview_settings['Item'] = {
 	add_fields: ["item_name", "stock_uom", "item_group", "image", "variant_of",
 		"has_variants", "end_of_life", "disabled", "is_sales_item"],
+	filters: [["disabled", "=", "1"]],
 
 	get_indicator: function(doc) {
 		if (doc.disabled) {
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item_list.js
```diff
@@ -1,6 +1,7 @@
 frappe.listview_settings['Item'] = {
 	add_fields: ["item_name", "stock_uom", "item_group", "image", "variant_of",
 		"has_variants", "end_of_life", "disabled", "is_sales_item"],
+	filters: [["disabled", "=", "1"]],
```
nabinhait: disabled = 0

erpnext/stock/doctype/item/item_list.js
```diff
@@ -12,7 +13,7 @@ frappe.listview_settings['Item'] = {
 		} else if (doc.variant_of) {
 			return [__("Variant"), "green", "variant_of,=," + doc.variant_of];
 		} else {
-			return [__("Active"), "blue", "end_of_life,>=,Today"];
+			return [__("Active"), "blue", "disabled,=,1"];
```
nabinhait: this change not needed

nabinhait:
manually added the change

PR #4653, opened 2016-01-20
Update issue_list.js - Filter improvement

RicardoJohann:
Added also "Replied" in the default filter list


nabinhait:
Don't make this change, "Replied" issues do not need any immediate action, hence not needed in the default list.

PR #4659, opened 2016-01-20
[Renamed] - the filename of images in the articles docs to work with …

syedqadeer:
…windows


nabinhait:
What the changes have you made?

syedqadeer:
i have only renamed the filename of the images, they were have a timeformat with ":" (colons) windows is causing issue for the colons so i have changed it to be "." as your other images.

nabinhait:
Ok. Better would be giving a proper filename based on what it does.

You have to change the associated .md file as well.

syedqadeer:
okay, i am working on it.

nabinhait:
Sorry, actually we have removed this page recently. In the latest release, it is covered in https://frappe.github.io/erpnext/user/manual/en/stock/articles/managing-batch-wise-inventory.html

Closing the PR

syedqadeer:
@nabinhait  :

Kindly, remove the ":" (colon) from those 4 images filenames, since you have said that you removed that page already, but the images for that page still exists, due to the presence of colons ":" in the filename, i am unable to clone on windows. Please look into this matter, if you want i can create a new pull request by renaming the file names ?

Thanks.

nabinhait:
Removed those image files

Issue #4661, opened 2016-01-20
Cannot Undo Journal Entry in Bank Recon


ccfiel:
Hello @anandpdoshi @nabinhait 

I want to submit a PR for this if the logic is acceptable in ERPNext.

The problem now in ERPNext is you cannot Undo a record in bank recon. Even you edit the clearance date.

My propose solution allow to set clearance date to blank again (None) the record will be back to unreconcilled state.

or is there a better way to undo reconciled journal entry

What do you think?

ccfiel:
made a PR https://github.com/frappe/erpnext/pull/4662

any comments? :)

ccfiel:
thanks! its now merge in dev.

nabinhait:
Thanks for the PR

PR #4662, opened 2016-01-20
undo reconciled journal entry record

ccfiel:
see https://github.com/frappe/erpnext/issues/4661

### DIFF
--- a/erpnext/accounts/doctype/bank_reconciliation/bank_reconciliation.py
+++ b/erpnext/accounts/doctype/bank_reconciliation/bank_reconciliation.py
@@ -51,14 +51,10 @@ class BankReconciliation(Document):
 				if d.cheque_date and getdate(d.clearance_date) < getdate(d.cheque_date):
 					frappe.throw(_("Clearance date cannot be before check date in row {0}").format(d.idx))
 
+			if d.clearance_date or self.include_reconciled_entries:
 				frappe.db.set_value("Journal Entry", d.voucher_id, "clearance_date", d.clearance_date)
 				frappe.db.sql("""update `tabJournal Entry` set clearance_date = %s, modified = %s
 					where name=%s""", (d.clearance_date, nowdate(), d.voucher_id))
-				vouchers.append(d.voucher_id)
-			else:
-				frappe.db.set_value("Journal Entry", d.voucher_id, "clearance_date", None)
-				frappe.db.sql("""update `tabJournal Entry` set clearance_date = NULL , modified = %s
-					where name=%s""", (nowdate(), d.voucher_id))
 				vouchers.append(d.voucher_id)
 
 		if vouchers:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/bank_reconciliation/bank_reconciliation.py
```diff
@@ -55,6 +55,11 @@ def update_details(self):
 				frappe.db.sql("""update `tabJournal Entry` set clearance_date = %s, modified = %s
 					where name=%s""", (d.clearance_date, nowdate(), d.voucher_id))
 				vouchers.append(d.voucher_id)
+			else:
+				frappe.db.set_value("Journal Entry", d.voucher_id, "clearance_date", None)
+				frappe.db.sql("""update `tabJournal Entry` set clearance_date = NULL , modified = %s
+					where name=%s""", (nowdate(), d.voucher_id))
+				vouchers.append(d.voucher_id)
```
nabinhait: May be the following will be better:

    if d.clearance_date:
    	if d.cheque_date and getdate(d.clearance_date) < getdate(d.cheque_date):
    		frappe.throw(_("Clearance date cannot be before check date in row {0}").format(d.idx))
    
    if d.clearance_date or self.include_reconciled_entries:
    	frappe.db.set_value("Journal Entry", d.voucher_id, "clearance_date", d.clearance_date)
    	frappe.db.sql("""update `tabJournal Entry` set clearance_date = %s, modified = %s
    		where name=%s""", (d.clearance_date, nowdate(), d.voucher_id))
    	vouchers.append(d.voucher_id)

Whats say?

tindaio:
@nabinhait I check the code I did not see that you reset the clearance date.  With this code how can undo a journal entry that has been already click the process update clearance date?

ccfiel:
@nabinhait I check the code I did not see that you reset the clearance date. With this code how can undo a journal entry that has been already click the process update clearance date?

nabinhait:
If "include_reconciled_entries" is checked, it will fetch all entries. And in that case, I have bypassed `if clearance_date` check, so it will update for all rows, even if clearance date is blank.

nabinhait:
By the way, `set_value` and `update via sql query` is duplicate statement.

ccfiel:
@nabinhait I check and test your suggestion its much cleaner :) I have already change it.

ccfiel:
@nabinhait Hello we still have the bug I have notice in your suggested code when you set the clearance date to blank again it meriadb will set it just set it to 0000-00-00 not null still the report in bank reconciliation still thinks its reconciled because it not null. https://github.com/ccfiel/erpnext/commit/cb66d3a8c94051b1027cb6e3a61f9e639130b02d I think this is still the solution?

.
```
	def update_details(self):
		vouchers = []
		for d in self.get('journal_entries'):
			if d.clearance_date:
				if d.cheque_date and getdate(d.clearance_date) < getdate(d.cheque_date):
					frappe.throw(_("Clearance date cannot be before check date in row {0}").format(d.idx))

			if d.clearance_date or self.include_reconciled_entries:
				frappe.db.set_value("Journal Entry", d.voucher_id, "clearance_date", d.clearance_date)
				frappe.db.sql("""update `tabJournal Entry` set clearance_date = %s, modified = %s
					where name=%s""", (d.clearance_date, nowdate(), d.voucher_id))
				vouchers.append(d.voucher_id)
```

ccfiel:
@nabinhait  I create an alternative solution that will incorporate your idea https://github.com/frappe/erpnext/pull/4744. What do you think?

anandpdoshi:
Pass None instead of blank string of clearance date is missing. This will prevent the bad date.

ccfiel:
@anandpdoshi Yes this the PR I made. https://github.com/frappe/erpnext/pull/4744 . What do you think?

PR #4663, opened 2016-01-21
Compact Print Layout Items

ShashaQin:
Hi,
I added a new check box on stock setting page called 'Compact Print Layout' to set item list print view. if this check box is checked, item list just have four columns in print view which are description, qty, rate and amount, and other columns will be shown in description column.


ShashaQin:
found some problems for this update, close it temporarily.

PR #4677, opened 2016-01-23
[Feature] feature to add margin in Quotation Item and Sales Order Item

mbauskar:
Hi,

User Can apply the margin on Quotation Item and Sales Order Item using following two options.
1)Price Rule: With the help of this method user can apply the margin on Quotation and Sales Order based on condition. We have added section margin on pricing rule where a user has to select the type of margin whether it is Percentage or Amount and rate or amount. The system will apply the margin on quotation item and sales order item if pricing rule is enabled.
2) Apply margin direct on Item: If user wants to apply the margin without pricing rule, they can use this option. In Quotation Item and Sales Order Item, user can select the margin type and rate or amount. The system will calculate the margin and apply it on price list rate to calculate the rate of the product.

Rate will be calculated as follows:
Rate = Price List Rate + Margin Rate

Note: We have applied the margin on price list rate not on the rate. So margin feature will get disabled if price list rate is not defined against the item.

Thanks,
Indictrans Technologies Pvt Ltd


rmehta:
@mbauskar thanks!

@nabinhait did you have a look at this?

erpnext/selling/doctype/quotation_item/quotation_item.json
```diff
@@ -412,6 +428,33 @@
...
+   "search_index": 0, 
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "depends_on": "eval:doc.price_list_rate", 
+   "fieldname": "type",
```
rmehta: should be `margin_type`

erpnext/selling/doctype/quotation_item/quotation_item.json
```diff
@@ -545,6 +594,56 @@
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
+   "depends_on": "eval: doc.type != \"\"", 
+   "fieldname": "rate_or_amount",
```
rmehta: should be `margin_rate_or_amount`

rmehta:
@mbauskar can you rename the fields? 

In general, *naming* of fields is **very very** important.

**Margin Type** must always be `margin_type` and not just `type`. Using ambiguous field names is not maintainable in the long run.

Also adding a couple of screenshots will be great for review.

Please mention the sponsor name too, so we can post it in the popup.

mbauskar:
Hi @rmehta 

Renamed the field to margin_type and margin_rate_or_amount,
also added the adding-margin.md and screenshot in `docs` directory.
please check https://github.com/indictranstech/erpnext/commit/ba4dbeddbbac731a05e072e5a135513c5a34117d

This feature has been developed for Daniel Noel.

Thanks,
Makarand

rmehta:
:+1: will check and merge this tomorrow.

erpnext/stock/get_item_details.py
```diff
@@ -69,8 +69,70 @@ def get_item_details(args):
...
+	if args.document_type in ["Quotation Item", "Sales Order Item"]:
+		# calculate rate by appling discount to total_margin
+		if(args.total_margin):
+			out["rate"] = discount_on_total_margin(args.total_margin, out.discount_percentage);
+		elif not args.type:
+			data = set_margin_fields(out.pricing_rule, out.price_list_rate)
+			out["rate"] = discount_on_total_margin(data.get("total_margin"), out.discount_percentage)
+			out.update(data)
+
 	return out
```
rmehta: @mbauskar since this is related to pricing rule, the correct place to put this code is here:

https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L158

Usually its easy to put everything related to a feature (like pricing rule) in one place so that it will be easy to debug. If this is here, someone needs to remember that margin is in `get_item_details` and discount is in `apply_pricing_rule`

Also since discount and margin are similar, they should be together.

Can you fix. Thanks!

mbauskar:
@rmehta we will put the mentioned code in `pricing_rule.py`. 
need some clarification regarding `Also since discount and margin are similar, they should be together.`
should I move the `margin_type, margin_rate_or_amount, and total_margin` fields in Sales Order Item, Quotation Item to Quantity and Rate section ??

rmehta:
@mbauskar create a new section "Discount and Margin"

Also they need to be in all the sales transactions (quotation, sales order, delivery note, sales invoice) for consistency

rmehta:
@mbauskar update?

mbauskar:
Hi @rmehta,

I have moved the suggested methods to pricing_rule.py, and added fields in Delivery Note and Sales Invoice

Thanks,
Makarand

rmehta:
@rohitwaghchaure let me know the update.

rmehta:
Merging via https://github.com/frappe/erpnext/pull/5044

PR #4685, opened 2016-01-25
Added the credit_days_based_on field to Supplier and Supplier Type doctype

### LINKED ISSUE #3727: Enhancement: Supplier Credit Days Allow Great Flexibility
Currently in Suppliers you can only specify the number of days after the invoice date on which payment is due. Of all my Suppliers with whom I have credit arrangements only one conforms to the fixed days scenario which 30 days in that Suppliers case. All the others are either End of the Next month ie an invoice raised on 1 July 2015 would become due for payment on 31 August 2015. All the rest are End of the Month + 30 days which would 30 August 2015 in the previous example.

In essence the Supplier screen should behave more like the Customers Screen.

So basically a dropdown list where you choose the type of credit terms given ie:
Fixed Days, End of Next Month, End of Current Month Plus Fixed Days

Like the Customer Form if the Fixed Days Options (and in this case here End of Current Month Plus Fixed Days) is chosen then a new box will need to appear to allow input of the number of credit days.

shreyasp:
Added the credit_days_based_on field to Supplier and Supplier Type doctype

Fixed #3727


anandpdoshi:
:100: / :100:

PR #4691, opened 2016-01-26
[cleanup] Removed 'Is Service Item' checkbox and filters

anandpdoshi:
'Is Service Item' checkbox only worked for filtering Items in Maintenance Schedule and Maintenance Visit, and validating that Items in a Maintenance type Order were of type 'Service'. However, it doesn't fit an actual use case where any Sales Item could be given for Maintenance, making the checkbox an unnecessary addition.


anandpdoshi:
@nabinhait please review

PR #4700, opened 2016-01-28
Simplified Timelog

RicardoJohann:
Our staff need to record the hours spent in an easier way on a daily basis. 

They input the date, activity type and how many hours worked on a project, support ticket or more of them in that day. 

As we don't need to record the time and have the overlap validation we added this option in a way it can be merged to your standard version. 

Be aware 'date_worked' field has been added as a custom field in our DB.


anandpdoshi:
@RicardoJohann Thanks for the pull request.

However, this seems very specific to your use case and not generic enough. Also, the custom field should be a standard field if you want to contribute the code that uses it.

PR #4721, opened 2016-02-01
Report Improvement - Profit and Loss, Balance Sheet

ShashaQin:
Hi, 
We have same issue with this link https://discuss.erpnext.com/t/query-on-the-profit-loss-report/2329.
So I changed the code to make Profit and Loss and Balance Sheet report non-cumulative. First, I added a check box named 'accumulated values' on these two reports. The reports keep same with current if it checked, and calculate monthly value and adding a total column at the last if it unchecked.


nabinhait:
@ShashaQin 

On top of your PR, I have made many changes. Please check https://github.com/frappe/erpnext/pull/4730. It will be merged and released tomorrow.

Issue #4710, opened 2016-01-29
[Bug] Pricing Rule price currency is not in Supplier Currency


bcornwellmott:
I noticed that when using pricing rules, the overwritten price is overwritten in the currency of the company, not the currency of the supplier listed in the pricing rule. I believe this is a bug.

To explain my set-up:

I have a number of suppliers that have quoted a part at various quantities. In the example I've taken screen shots of, there are two suppliers with the following pricing:

Item Code: m0209
Supplier Name: Amazon
Price 1: US$144 (Qty 1-9)
Price 2: US$75 (Qty 10+)

Supplier Name: AliExpress
Price 1: US$107.38 (Qty 3+)

The way I store this information in ERPNext is by using Price Lists, Item Prices and Pricing Rules. For each supplier I have a Price List. For the lowest quantity price, I have an Item Price, and for each of the additional prices (at higher quantities) I have a Pricing Rule. 

Price Lists: Amazon, AliExpress
Item Price: ITEM-PRICE-00971 (for the Amazon price of US$144), ITEM-PRICE-00972 (for the AliExpress price of $107.38)
Pricing Rules: Amazon-m0209-10-10000 (for the Amazon price of US$75)

When I go to create a purchase order for the item with Amazon selected as a supplier and a quantity of 1, I get the correct price in the correct currecy:

![po-goodprice](https://cloud.githubusercontent.com/assets/16806041/12690668/019dbdc4-c69a-11e5-9e2a-62642a063b83.png)

When I change the quantity of the item to 10, the price in the pricing rule is taken in the wrong currency (my company currency), instead of the supplier currency.

![po-badprice](https://cloud.githubusercontent.com/assets/16806041/12690569/58c34cf0-c699-11e5-8ef0-930f4c93c625.png)

I've attached the Supplier, Price List and Pricing Rule screenshots below to demonstrate that the Price List and Supplier are both set the the USD currency.

Supplier:
![supplier](https://cloud.githubusercontent.com/assets/16806041/12690577/69f2fb06-c699-11e5-9b95-0196dd125b19.png)

Price List:
![pl](https://cloud.githubusercontent.com/assets/16806041/12690611/97d9b834-c699-11e5-94ac-d58af434fb6d.png)

Pricing Rule:
![pr](https://cloud.githubusercontent.com/assets/16806041/12690586/76929114-c699-11e5-8974-0fb896ca37d7.png)

[...]

PR #4731, opened 2016-02-02
Conversion_rate not applicable for type 'Supplier'

bcornwellmott:
If the type is supplier, then the conversion rate is not applicable because the pricing rule is already in the currency of the supplier.

[Issue #4710 ](https://github.com/frappe/erpnext/issues/4710)


nabinhait:
This is not the correct solution. We should add Currency option in Pricing Rule. And if Pricing Rule currency matches with transaction currency, then only we should apply it in the transaction.

PR #4733, opened 2016-02-03
Update leave_application.py

ShashaQin:
Hi,
Found an issue about leave application: Employee cannot apply half day of Unpaid Leave after lodging half day of Annual leave on the same day. This issue is caused by validating leave overlap. I fixed this issue by checking half day check box and total_leave_days.


rmehta:
@ShashaQin merging this and thanks for contributing, but please include test-cases + documentation update in future contributions.

rmehta:
@ShashaQin you have also copy pasted the query from above, always make a function and not duplicate code.

rmehta:
Please use tabs instead of spaces as per the convention in this app

PR #4744, opened 2016-02-04
undo bank recon.

ccfiel:
set to null if clearance_date is blank because if not it will set it just 0000-00-00


nabinhait:
@ccfiel Why is it needed? System also considers "0000-00-00" as non-reconciled entry. Check https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/bank_reconciliation/bank_reconciliation.py#L18

ccfiel:
@nabinhait in bank reconciliation you are correct but in the report bank reconciliation statement 0000-00-00 is still considered as reconciled. Check https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/report/bank_reconciliation_statement/bank_reconciliation_statement.py

PR #4759, opened 2016-02-07
Corrected typo while translating

Tropicalrambler:
Found a switch of letters:  "wrokstation" instead of "workstation".  Corrected the item appropriately.  Found this while translating my fork of this page to spanish.


fderyckel:
Actually the term operation is also wrongly spelled.

rmehta:
@fderyckel I saw that, but the branch is un-mergeable because of conflicts.

PR #4762, opened 2016-02-08
Advance adjustment in invoice cannot be greater than grand total

nabinhait:


hiousi:
Why ? I am in the following scenario:
advance paiement of 100% of the order. Order splitted into 2 deliveries. Need to make 2 invoices (for customs, it's a legal requirement).
First invoice will fire "Advance amount cannot be greater than..."

rmehta:
@hiousi  can you open a new issue for this?

hiousi:
@rmehta No ;)  just because I discovered you can assign a partial amount of the advance to an invoice. The rest of the total is put on hold for next invoice.
So it works well. Sorry for your time.

PR #4768, opened 2016-02-08
Updated help on manufacturing module.

umairsy:


rmehta:
remove files from `docs/current`

umairsy:
Removed files from docs/current. thanks to Nabin.

PR #4777, opened 2016-02-09
Compact Items Print Layout

ShashaQin:
Hi,
We'd like to compact items print layout like version 4. So I added a new check box on feature setup page called 'Compact Print Layout' to set item list print view. if this check box is checked, item list just have four columns in print view which are description, qty, rate and amount, and other columns will be shown in description column. If this check box is unchecked, everything will be same as current standard.


anandpdoshi:
can you share a screenshot of what do you mean?

erpnext/controllers/print_settings.py
```diff
@@ -1,9 +1,28 @@
...
+import frappe
+from frappe.utils import cint
+
 def print_settings_for_item_table(doc):
+	
 	doc.print_templates = {
 		"description": "templates/print_formats/includes/item_table_description.html",
 		"qty": "templates/print_formats/includes/item_table_qty.html"
 	}
+	customised_print_preview = cint(frappe.db.get_value("Features Setup", None, "customised_print_preview"))
```
rmehta: Please use 'compact_item_view` here. The field names and labels should be clear to the end user.

ShashaQin:
field name and label are changed to 'Compact Item View'. 
Features setup screenshot:
features setup page
Item print view screenshot when checked Compact Item View:
check compact item view
Item print view screenshot when unchecked Compact Item View:
uncheck compact item view

anandpdoshi:
merged manually after cleanup

anandpdoshi:
Thanks @ShashaQin

PR #4788, opened 2016-02-10
Added Make Receipt (Update Stock) and IsCash Options Into Purchase Invoice

jof2jc:
- Make Receipt option is to do purchase receipt directly from within Purchase Invoice.
- IsCash is to make cash purchase directly from Purchase Invoice.

These will simplify purchase cycle especially for small businesses. Just use single purchase invoice entry to book payable, receive stock items and deduct payable with cash payment,


rmehta:
@jof2jc Thanks for the PR!

Can add make a couple of test cases and also update the documentation. It will make it much easier for us to merge.

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -706,6 +784,178 @@
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
+   "fieldname": "sc_wh",
```
rmehta: minor, keep fieldnames explicit `warehouse_section`

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -208,6 +208,31 @@
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
+   "default": "0", 
+   "fieldname": "is_cash",
```
rmehta: Should this be `include_payment` (as payment may not necessarily be "cash", but could be a check?)

jof2jc: How about IsPaid?

jof2jc:
@rmehta how to make test cases? Sorry I'm still new with this.. it would be great if you can give guidances.. Thanks

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
rmehta: Yeah, `is_paid` should also work

rmehta:
@jof2jc check this: https://frappe.github.io/frappe/user/guides/basics/writing-tests

You will have to write another method like `test_paid_purchase_receipt` in this file: https://github.com/frappe/erpnext/blob/develop/erpnext/stock/doctype/purchase_receipt/test_purchase_receipt.py#L13

You can test it by

`bench run-tests --doctype "Purchase Receipt"`

jof2jc:
@rmehta, sorry still get no luck and lack of time to.create test method as i'm still a newbie.developer and still need deep learning on python..

If you don't mind..can you guys take it over from here and get it merged. Promise this time only!! 

I have tested it with real data entries right on the system and it works fine. This feature will be very helpful in case users really need to simplify purchase cycle.

Many thanks

rmehta:
@jof2jc This also has a lot of code repetition from Purchase Receipt. Atleast the code needs to be commonified. Fixing your PR will take a longer time than making it again.

Also Test Cases are the biggest part of this PR since, the code is already there in Purchase Receipt.

Do you want to try it, or should we close it?

The feature itself is good and we would like to accept such requests, but its not a frequently asked features so I am not sure how much time we can devote on this right now.

anandpdoshi:
@jof2jc I have a simpler solution to this:

1. Make a checkbox "Make Receipt and Payment on Submit"
1. On submission of Purchase Invoice, use mapper functions to create and submit Receipt and Payment records.

This will take less code and be consistent.

jof2jc:
@anand, can you give examples how to use mapper functions?

anandpdoshi:
See:
1. https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order/purchase_order.py#L310
1. https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order/purchase_order.py#L276

rmehta:
Merging via https://github.com/frappe/erpnext/pull/5040

jof2jc:
@rmehta really thanks for this...but it's not on master yet?

PR #4791, opened 2016-02-11
[fix] reorder items

makeen:
To test this issue:
Create new Item with 'Maintain Stock' and reorder_level=1 reorder_qty=1
Create and submit a Sales Order with 2 units of previous item.
For two days pass in two minutes: move tasks "erpnext.stock.reorder_item.reorder_item" from "daily" to "all" in hooks.py and DEFAULT_SCHEDULER_INTERVAL from 300 to 60 in celery_app.py

First run task: this will make a 'Material Request' with 3 units of the item, up here all right.

Second run task: this will make a second 'Material Request' with 1 units. I think this is wrong.

### DIFF
--- a/erpnext/stock/reorder_item.py
+++ b/erpnext/stock/reorder_item.py
@@ -48,7 +48,7 @@ def _reorder_item():
 		# projected_qty will be 0 if Bin does not exist
 		projected_qty = flt(item_warehouse_projected_qty.get(item_code, {}).get(warehouse))
 
-		if reorder_level and projected_qty <= reorder_level:
+		if reorder_level and projected_qty < reorder_level:
 			deficiency = reorder_level - projected_qty
 			if deficiency > reorder_qty:
 				reorder_qty = deficiency


anandpdoshi:
@makeen why do you think this is wrong? If the re-order level is set to 5, and if the stock reaches 5 or less, it should re-order.

makeen:
@anandpdoshi I don't think that this is wrong, but why after the first 'material request' with 5 units, the second day other 'material request' is create with 1 units of same item.
`if reorder_level and projected_qty <= reorder_level:` if 'projected_qty' is equal to reorder_level why reorder.

anandpdoshi:
Projected Qty = Actual Qty + Planned Qty + Requested Qty + Ordered Qty - Reserved Qty

This is the quantity that is exists to fulfill current orders. You want to raise a new material request automatically when this quantity reaches or reduces below a certain value which is the re-order level. Hence we use `<=` and not just `<`. Ideally its name should be 'Auto-request Level' and 'Auto-request Quantity'.

makeen:
If you have an Item with reorder_level=5 and reorder_qty=1, is normal that erpnext create two 'material request' one with 5 units and other(the next day) with 1 units?

 ```
if reorder_level and projected_qty <= reorder_level:
			deficiency = reorder_level - projected_qty
			if deficiency > reorder_qty:
				reorder_qty = deficiency
```
first day:
**reorder_qty = 1**
reorder_level = 5
proyected_qty = 0
deficiency = 5
**reorder_qty = 5**, this is because if deficiency(5) > reorder_qty(1) then reorder_qty = deficiency(5)
"Material request with 5 units"

second day:
**reorder_qty = 1**
reorder_level = 5
**proyected_qty = 5**
deficiency = 0, reorder_qty keeps its value because deficiency is less than reorder_qty
**reorder_qty = 1**
"Material request with 1 units" (I think this is wrong)

creador30:
I understand the reorder-level as the security stock you want to keep. So, this is the minimum quantity you want to have in your warehouse. Whether auto reorder is created at or below reorder-level is just a matter of interpretation. The important point is the minimum stock level you have, as this affects costs.

However with the current implementation we are losing some functionality. Take a case where you want to keep just one piece of security stock of an item. With <= you will have at least 2 items at stock and there is no way you can go below that.

Even with < there is no way to keep the auto reorder functionality, if you don't want to keep any security stock. That's why I created the pull request #4807

makeen:
@anandpdoshi With `<=` reorder_items task tries to keep the stock of the items above (not at) the reorder_level making MRs but, Why create two MR for each item? one MR with the quantity for reach reorder_level and other (the next day) with one unit more to exceed the value. Reorder level concept is not equivalent to minimum stock concept due to `<=`.

nabinhait:
@makeen @creador30 @anandpdoshi 
According to the [Wikipedia](https://en.wikipedia.org/wiki/Reorder_point) , it stock falls below reorder_level, then only we should reorder items. Hence, I think "<" is correct one. Accepting the PR #4807

PR #4806, opened 2016-02-14
Product Bundle improvement

creador30:
This change allows addition of multiple instances of the same product inside a product bundle, if each has a different description in the product bundle list (thus no duplicates).

In our case we use product bundles to include scrap material into our stock calculations. Here we need to enter the same product with different quantities into the same bundle. We explicitly specify scrap in the description of each line in product bundle.


erpnext/stock/doctype/packed_item/packed_item.json
```diff
@@ -523,6 +523,32 @@
    "search_index": 0, 
    "set_only_once": 0, 
    "unique": 0
+  },
+  {
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "fieldname": "packing_description",
```
anandpdoshi: Why do you need to add a new field when there is already a field called `description` in Packed Item doctype?

creador30:
The field "description" is the same as the item description. This is a single value for each item.
However while creating a product bundle there is a description field you can use to give some detail about e.g. why this item is added to the bundle. This description is saved within the doctype Product Bundle Item in field 'description'.
I did not want to change the default behaviour of the description field of doctype Packed Item. That is why I added a new field to store the description field value from Product Bundle Item doctype.

Below picture shows the product bundle screen with the same item added twice with different descriptions:
screen shot 2016-02-16 at 22 00 15

This picture shows a line from the packing list of a sales order. You can see that the description field stores the item description, whereas the packing_description field store the description from Product Bundle:
screen shot 2016-02-16 at 21 59 11

nabinhait:
Still we will recommend one "Description" field for the packed item? Can you please change the PR as per that?

creador30:
Thanks for the feedback, the PR is now corrected accordingly.

PR #4807, opened 2016-02-14
Allow reorder_level to be set to 0

creador30:
For items you don’t want to keep in stock, you can set the reorder_level to 0 and just define reorder_qty.

If there is e.g. a sales order for that item, the projected quantity will fall below 0, which will trigger an automatic material request.

Includes #4791

### DIFF
--- a/erpnext/stock/reorder_item.py
+++ b/erpnext/stock/reorder_item.py
@@ -48,7 +48,7 @@ def _reorder_item():
 		# projected_qty will be 0 if Bin does not exist
 		projected_qty = flt(item_warehouse_projected_qty.get(item_code, {}).get(warehouse))
 
-		if reorder_level and projected_qty <= reorder_level:
+		if (reorder_level or reorder_qty) and projected_qty < reorder_level:
 			deficiency = reorder_level - projected_qty
 			if deficiency > reorder_qty:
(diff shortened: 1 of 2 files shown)


erpnext/stock/reorder_item.py
```diff
@@ -48,7 +48,7 @@ def add_to_material_request(item_code, warehouse, reorder_level, reorder_qty, ma
 		# projected_qty will be 0 if Bin does not exist
 		projected_qty = flt(item_warehouse_projected_qty.get(item_code, {}).get(warehouse))
 
-		if reorder_level and projected_qty <= reorder_level:
+		if (reorder_level or reorder_qty) and projected_qty < reorder_level:
```
anandpdoshi: why change `<=` to `<` in level comparison?

nabinhait:
@creador30 Merged manually after fixing the condition, please check https://github.com/frappe/erpnext/commit/38be62c7e3e668cef8b1c633cea2e310368a83d2

nabinhait:
@creador30 After checking the Wikipedia, reverted my changes via https://github.com/frappe/erpnext/commit/e32f9c3a24abbad79ac81fd89ccb2c7bfb60e36b

Issue #4780, opened 2016-02-10
Displaying stock balance while creating quotation (In Quotation Item Line)


Devens:
It is very important that while creating quotation, sales user would know - what is the availability of stock (Displaying stock balance will further empower him by understanding total balance of stock in various warehouses). 

We suggest that in Quotation Item (similar to Actual Qty, etc being shown in Sales order Item), stock balance should be displayed as read only field. By clicking on that stock balance, if user can be directed to stock balance report for respective item, it can be further helpful.

kamtex:
Agreed. That will be a good enhancement.

PR #4825, opened 2016-02-17
[enhance] Displaying stock balance while creating quotation

saurabh6790:
https://github.com/frappe/erpnext/issues/4780


erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -162,3 +163,10 @@ cur_frm.cscript.on_submit = function(doc, cdt, cdn) {
 frappe.ui.form.on("Quotation Item", "items_on_form_rendered", function(frm, cdt, cdn) {
 	// enable tax_amount field if Actual
 })
+
+frappe.ui.form.on("Quotation Item", "stock_balance", function(frm, cdt, cdn) {
+	console.log("here")
```
anandpdoshi: :hankey:

erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -162,3 +163,10 @@ cur_frm.cscript.on_submit = function(doc, cdt, cdn) {
...
+	var d = locals[cdt][cdn];
```
anandpdoshi: Nicer way to write: `var d = frappe.model.get_doc(cdt, cdn);`

erpnext/stock/get_item_details.py
```diff
@@ -165,6 +165,7 @@ def get_basic_details(args, item):
 		"discount_percentage": 0.0,
 		"supplier": item.default_supplier,
 		"delivered_by_supplier": item.delivered_by_supplier,
+		"actual_qty": get_item_balance_qty_for_default_warehouse(item)
```
anandpdoshi: This need not always be default warehouse, since you are allowed to change the warehouse. Also, use the existing function: `erpnext.stock.utils.get_stock_balance`.

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -106,7 +106,8 @@ def set_missing_values(source, target):
 			"Quotation Item": {
 				"doctype": "Sales Order Item",
 				"field_map": {
-					"parent": "prevdoc_docname"
+					"parent": "prevdoc_docname",
+					"warehouse": "warehouse"
```
nabinhait: Warehouse will be auto copied, if "No Copy" is not checked.

erpnext/stock/get_item_details.py
```diff
@@ -165,6 +165,7 @@ def get_basic_details(args, item):
 		"discount_percentage": 0.0,
 		"supplier": item.default_supplier,
 		"delivered_by_supplier": item.delivered_by_supplier,
+		"actual_qty": get_available_qty(item.name, user_default_warehouse or args.warehouse or item.default_warehouse)
```
nabinhait: Not needed, it already maps actual qty via https://github.com/frappe/erpnext/blob/develop/erpnext/stock/get_item_details.py#L46

PR #4832, opened 2016-02-18
Need to use set_balance_in_account_currency

revant:
https://github.com/revant/drop_ship/commit/b9ad5178e4f75a73be9f7b36a1b3d691c8937746
@nabinhait 
needed to make GL Entry

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -239,7 +239,7 @@ class AccountsController(TransactionBase):
 
 		if self.doctype not in ["Journal Entry", "Period Closing Voucher"]:
 			self.validate_account_currency(gl_dict.account, account_currency)
-			self.set_balance_in_account_currency(gl_dict, account_currency)
+			set_balance_in_account_currency(gl_dict, account_currency, self.conversion_rate, self.company_currency)
 
 		return gl_dict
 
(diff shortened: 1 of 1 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -541,3 +524,20 @@ def _on_previous_row_error(row_range):
 			_on_previous_row_error("1 - %d" % (tax.row_id,))
 		elif tax.get("category") == "Valuation":
 			frappe.throw(_("Valuation type charges can not marked as Inclusive"))
+
+def set_balance_in_account_currency(gl_dict, conversion_rate=None, company_currency=None, account_currency=None):
```
nabinhait: Don't change the position of the argument "account_currency", keep it in the second place.

nabinhait: You should also change wherever it is called in erpnext.

revant:
I searched set_balance_in_account_currency in all files, found it used only in accounts_controller.py, that I changed.
Are arguments correct?

nabinhait:
@revant I have fixed an issue on your PR, it gives an error for stock entry. Check https://github.com/frappe/erpnext/commit/67f40775ff588ea997522ebb18028e2a3b09e6cf

PR #4834, opened 2016-02-18
[enhance] add date in currency exchange

saurabh6790:


rmehta:
@saurabh6790 test cases failing

saurabh6790:
@rmehta fixed

erpnext/shopping_cart/doctype/shopping_cart_settings/shopping_cart_settings.py
```diff
@@ -34,20 +33,16 @@ def validate_exchange_rates_exist(self):
...
-
-			missing = list(set(expected_to_exist).difference(exists))
-
-			if missing:
-				msgprint(_("Missing Currency Exchange Rates for {0}").format(comma_and(missing)),
+					where from_currency = %s and to_currency = %s
+					order by date desc limit 1""", (currency, company_currency))
+					
+			if currency == company_currency:
+				exists = [1]
```
anandpdoshi: Why write it this way?

erpnext/setup/utils.py
```diff
@@ -62,19 +62,21 @@ def before_tests():
...
 	
-	exchange = "%s-%s" % (from_currency, to_currency)
-	value = flt(frappe.db.get_value("Currency Exchange", exchange, "exchange_rate"))
-
+	if not posting_date:
+		posting_date = frappe.utils.nowdate()
+		
+	value = flt(frappe.db.get_value("Currency Exchange", 
+		{"from_currency": from_currency, "to_currency": to_currency,
+			"date": posting_date}, "exchange_rate"))
```
anandpdoshi: What about the case when a user wants to use a "fixed" exchange rate? The code should also look for a Currency Exchange record with a blank date. Or did I miss a discussion? @nabinhait what's your opinion?

erpnext/setup/utils.py
```diff
@@ -62,19 +62,21 @@ def before_tests():
...
 	if not value:
 		try:
 			cache = frappe.cache()
-			key = "currency_exchange_rate:{0}:{1}".format(from_currency, to_currency)
+			key = "currency_exchange_rate:{0}:{1}:{2}".format(from_currency, to_currency, posting_date)
 			value = cache.get(key)
-
 			if not value:
 				import requests
 				response = requests.get("http://api.fixer.io/latest", params={
```
anandpdoshi: Shouldn't we pass the date to fixer.io?

erpnext/setup/utils.py
nabinhait: Yes, there should be a option to set fixed exchange rate

anandpdoshi:
To Do:

- [ ] Add "From Date" and "To Date" in Currency Exchange
- [ ] Validate overlap of exchange records
- When looking for an exchange rate:
    - [ ] From Date <= Posting Date <= To Date
    - [ ] From Date <= Posting Date and To Date is null
    - [ ] From Date is null and Posting Date <= To Date
    - [ ] Both From Date and To Date are null
    - [ ] Fixer.io on that date

PR #4836, opened 2016-02-18
Deprecate close button in SO/PO since Stop and Close has same functionality 

patilsangram:
Hello,

We have merge stop and close button.

Thanks,
Sangram
New Indictrans Technologies Pvt Ltd


anandpdoshi:
@patilsangram I think we should deprecate "Stop" and keep "Close" as it makes more sense in most situations

@rmehta @nabinhait what do you think?

rmehta:
Yeah, remove Stop, keep Close.

@patilsangram thanks for the PR!

PR #4842, opened 2016-02-19
Multilingual print format

rohitwaghchaure:
Given option to the user to set language for their customer, supplier. For an example if I have customers from germany, france who want quotation in german, french language respectively will be possible with these feature.


rohitwaghchaure:
error is coming because dependency in frappe is not yet merged

Issue #909, opened 2013-09-26
Recording similar data for Job Applicants, transferrable to Employee


lxnow:
Here's an idea: let job applicants apply through ERPNext portal as, say, a "guest" user that is limited to only applying and not viewing any other records.  This collects all pertinent information about the applicant which can later be simply transferred to an Employee record if he/she gets hired.

It's pretty much like a PO->PR or SO->DN process.

This would be helpful for recruitment agencies type businesses, and I'd certainly find use for this as well. Saves on paper.

vjFaLk:
Fixed

Issue #637, opened 2013-07-24
Job Opening & Jobs page on website


rmehta:
- [x] Create a new DocType Job Opening
- [ ] Create a Genarator and a list page (/jobs) on website
- [ ] Allow user to set an introductory (sales pitch) text on the Jobs list page.

PR #4846, opened 2016-02-19
Added feature to allow people to apply for job from website

KanchanChauhan:
Fix for https://github.com/frappe/erpnext/issues/909 and https://github.com/frappe/erpnext/issues/637


rmehta:
@KanchanChauhan the Job Application web-form is not a part of the commit. Need to fix this tomorrow.

PR #4849, opened 2016-02-22
Create at_austria_chart_Einheitskontenrahmen.json

rasos:
Austrian Chart of Accounts compiled by fairkom.eu from various sources, among them the official "Österreichischer Kontenrahmen" published by ÖPWZ - Fachsenat für Betriebswirtschaft und Organisation der Kammer der Wirtschaftstreuhänder Version May 2014. 
Numbering scheme has been slightly adapted to fit best with ERPnext setup of a company, where the most important accounts are automatically assigned. 
When using this chart of accounts, feel free to reduce / adapt the accounts - it is not legally binding but good practice to be able to compare the performance of companies in Austria.


nabinhait:
Thanks for the PR, we will test it soon.

PR #4855, opened 2016-02-22
better validation example for developer

sbkolate:
throw "past date selected" will freeze the form and user can not correct value of  from_date.
but validated false will just throw the pop-up and user can correct value of from_date and save the form.

Example changed from Event to Task because same is available in Custom Script Help for desk user


nabinhait:
thanks

PR #4859, opened 2016-02-23
Shopping Cart with multi currency and checkout fixes

saurabh6790:


erpnext/shopping_cart/cart.py
```diff
@@ -307,6 +323,32 @@ def get_customer(user=None):
 
 		return customer
 
+def get_debtors_account(cart_settings):
+	payment_gateway_account_currency = \
+		frappe.get_doc("Payment Gateway Account", cart_settings.payment_gateway_account).currency
+	
+	account_name = _("Debtors ({0})".format(payment_gateway_account_currency))
+	
+	debtors_account = get_account("Receivable", "Asset", is_group=0, account_name=account_name)
```
nabinhait: Check existing debtors account based on account_type and account_currency and company, not based on account_name. If not found create it.

erpnext/shopping_cart/cart.py
```diff
@@ -307,6 +323,32 @@ def get_customer(user=None):
...
+	
+	if not debtors_account:
+		debtors_account = frappe.get_doc({
+			"doctype": "Account",
+			"account_type": "Receivable",
+			"root_type": "Asset",
+			"is_group": 0,
+			"parent_account": get_account(root_type="Asset", is_group=1, \
```
nabinhait: Again, get parent account based on account_type, is_group, company, not based on account_name

PR #4869, opened 2016-02-24
opportunities status fix in Sales Funnel report

sbkolate:
Reference https://discuss.erpnext.com/t/sales-funnel-not-showing-updated-opportunities/10702/4

As opportunity is not submittable, docstatus always zero for opportunity.

Thanks,
Sambhaji Kolate
New Indictrans Technologies Pvt Ltd


nabinhait:
thanks :+1:

PR #4871, opened 2016-02-25
Improvements about recurring documents

ShashaQin:
According to https://discuss.erpnext.com/t/suggestions-about-recurring-documents/10466, I've made some improvements about recurring documents.
1. Made next_date calculated by from_date and can be changed by user
2. Added a check box named 'Notify by Email', if checked to generate a submitted doc and send it via Email (current work flow), if unchecked to generate a draft document without sending email.


nabinhait:
@ShashaQin Test case is failing, can you please check?

ShashaQin:
It is because the docstatus of new recurring doc is 0 as new code set it as draft. The test case will repeat 5 time, but my new code will not generate the second recurring doc when run the second time test case because the docstatus of last recurring doc is still 0...
I'm not sure how to modify the test_recurring_document.py in this case.

ShashaQin:
made some changes, now checks passed.

PR #4876, opened 2016-02-25
[enhancement] get valuation rate and gross profit on sales order item

saurabh6790:


erpnext/stock/get_item_details.py
```diff
@@ -68,6 +67,8 @@ def get_item_details(args):
 
 	if args.get("is_subcontracted") == "Yes":
 		out.bom = get_default_bom(args.item_code)
+		
+	get_goss_profit(out)
```
nabinhait: spelling?

erpnext/stock/get_item_details.py
```diff
@@ -463,3 +467,10 @@ def get_default_bom(item_code=None):
 			return bom
 		else:
 			frappe.throw(_("No default BOM exists for Item {0}").format(item_code))
+			
+			
+def get_goss_profit(out):
+	out.update({
+		"gross_profit": ((out.price_list_rate - out.valuation_rate) * out.qty) * (out.conversio_rate or 1)
```
nabinhait: spelling of conversion_rate?

gross_profit = (out.base_rate - out.valuation_rate) * out.qty

It is in company_currency

erpnext/selling/doctype/sales_order_item/sales_order_item.json
```diff
@@ -1222,6 +1222,56 @@
...
+   "search_index": 0, 
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "fieldname": "gross_profit", 
+   "fieldtype": "Currency",
```
nabinhait: set options to show company currency symbol in gross profit value

erpnext/public/js/controllers/transaction.js
```diff
@@ -866,6 +867,10 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				refresh_field('to_date');
 			}
 		}
+	},
+	
+	gross_profit: function(item) {
```
nabinhait: set_gross_profit()

correct the formulae

and it is only applicable to sales order

saurabh6790: @nabinhait 

Not getting value of base_rate fields from js side on rate trigger if transaction is in company currency.

saurabh6790:
@nabinhait 

fixed calculation for non-stock and product bundle items

anandpdoshi:
needs to be fixed as per discussion with @nabinhait

saurabh6790:
@anandpdoshi  Fixed. as per discussions.

PR #4887, opened 2016-02-26
enable uom in BOM Table View

sbkolate:
reference 
https://discuss.erpnext.com/t/uom-in-item-tables/10733/4


nabinhait:
thanks :+1:

Issue #2090, opened 2014-08-20
Add stripe payment service as a 'Pay Now' option on Sales Invoices


jevonearth:
Feature request in User Story: 

As a customer of a company that uses ERPNext, I want to receive invoices via email, and click a 'Pay Now' link on the invoice, that opens a webpage (https://erpnext.example.net/invoice/036zfd34-c3cd-4dce-a808-32xefaa1302b) so that I can pay the invoice by credit-card quickly and easily.

As a billing user of ERPNext, I want to send my customers invoices that have a 'Pay Now' link on the invoice PDF that, when clicked, the customer is taken to a on-line payment screen where the payment is processed via stripe, so that erpnext automatically updates the Sales Invocies as paid.

Examples of how a similar feature is implemented in the Xero accounting product;

_Image showing payment link on a Invoice PDF_
![Image showing payment link on a Invoice PDF](https://cloud.githubusercontent.com/assets/1114943/3980401/b21795fe-2861-11e4-9136-99825ee6166a.png)

_Image showing invoice on-line and prompting user for payment details._
![Image showing invoice on-line and prompting user for payment details.](https://cloud.githubusercontent.com/assets/1114943/3980411/db56ecb2-2861-11e4-9e1a-54a7527919e7.png)

The Pay Now link on the Payment Invoice has a hashed URL, which makes it a 'private link'. When the customer opens the link, they see their invoice, and the option to pay that invoice via credit-card (in this case stripe). 

This feature request is related to #2088

betabank:
It is possible to add also the "Pay now" div in the shopping cart?

system19:
+1

pifiu:
+1 you could also add braintree or some other gateways. If you implement stripe, it would not be much more work to do any others.

yashodhank:
:+1: 

Option to Automate Recurring Payments.

To handle Payment Subscriptions will be very useful addition if not implemented yet.

rmehta:
sorry moving this to 5.1 - we need to close new development in 5.0

abhijeetramgir:
How do we do this?

abhijeetramgir:
how do we do this?

marri2k:
Still waiting for this !!!!

hubdotcom:
Could this been done using current actions / filters without modifying core code?

rmehta:
Not out of the box (yet)

hubdotcom:
Soon :)

[...]

PR #4891, opened 2016-02-27
Add stripe integration #2090

mach-kernel:
Hi,

This is a basic stripe integration for the ERPNext POS system. Future work probably involves reconciliation via webhook, but as it currently stands, the `stripe_checkout_id` is the `token` which Stripe returns as part of the checkout process so manual reconciliation is possible via the Stripe API.

[Here](http://imgur.com/a/hm7of) is the POS checkout flow with Stripe. If the `Customer` used in the checkout process has an e-mail address associated with their record, then it will automatically be filled in via the Stripe checkout process. 

TODO:
- Webhooks
- Add e-mail field to new customer modal from POS (probably as separate issue).

Thanks!


erpnext/public/js/pos/pos.js
```diff
@@ -489,13 +491,35 @@ erpnext.pos.PointOfSale = Class.extend({
 	},
 	set_pay_button: function(dialog) {
 		var me = this;
+
+		var stripe_handler = StripeCheckout.configure({
+			key: 'pk_test_6pRNASCoBOKtIshFeQd4XMUh',
```
anandpdoshi: Hard coded. Can be moved to a separate app called "Payment Gateways"

rmehta:
Needs to be made into a separate app.

Issue #4754, opened 2016-02-07
Disable "In Words" for Grand Total


dottenbr:
The "In Words" like "EUR Three Thousand, Six Hundred only." for "€ 3.599,75" is a cool feature but I have never seen this used in Germany. We would need to be able to turn this off like "Disable Rounded Total" function in Setup -> Global Defaults.

anandpdoshi:
You can hide In Words field for now.

Best,
Anand Doshi

ERPNext

dottenbr:
Can you assign this to Makis and have him fix it? It's quite similar to the
rounded values configuration he should fix it for everyone.

Am Montag, 8. Februar 2016 schrieb Anand Doshi :

> You can hide In Words field for now.
>
> Best,
> Anand Doshi
>
> ERPNext
>
> > On 07-Feb-2016, at 11:22 PM, Dominik <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > The "In Words" like "EUR Three Thousand, Six Hundred only." for "€
> 3.599,75" is a cool feature but I have never seen this used in Germany. We
> would need to be able to turn this off like "Disable Rounded Total"
> function in Setup -> Global Defaults.
> >
> > —
> > Reply to this email directly or view it on GitHub <
> https://github.com/frappe/erpnext/issues/4754>.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/frappe/erpnext/issues/4754#issuecomment-181205037>.
>


-- 

Mit freundlichen Grüßen / All Best,

*Dominik Ottenbreit*
Geschäftsführer ESO Germany
Director of Business Development

Tel.: ​+49 6752 9112-55
Mobil: ​+49 160 90813466
Fax: ​+49 6752 1524998

www.eso-electronic.com

ESO Electronic Service Ottenbreit GmbH
Hahnenbachstr. 21
55606 Hahnenbach, Germany

Geschaeftsfuehrer (CEO and President): Dominik Ottenbreit
Prokurist (Authorized Signatory):Bjoern Ottenbreit
Gerichtsstand (Jurisdiction): Bad Kreuznach, Germany (HRB 3460)
USt-IdNr. (Tax ID): DE 812 004 832
NCAGE: CH591, DUNS: 314797515, WEEE-Reg-Nr.: DE 84 800 814

anandpdoshi:
@dottenbr makis can send a pull request

[...]

PR #4923, opened 2016-03-03
Made In Words configurable

emakis00:
As per the discussion in https://github.com/frappe/erpnext/issues/4754 I have made it configurable for everybody to turn off the "In Words" in the Global Defaults just like it is done with Rounded.

screenshot 2016-03-03 16 33 28


rmehta:
:+1:

anandpdoshi:
Thanks @emakis 

> merged manually after a minor fix.

PR #4933, opened 2016-03-06
Update de.csv

ci2014:
"Umsatzrendite" is the wrong translation for "Sales Return". The correct translation is "Rücklieferung".

"Umsatzrendite" means something like the return of capital of your revenue. "Rücklieferung" means exactly what "Sales Return" means in the context of ERPNext.


anandpdoshi:
Hi. Can you update this using: [https://translate.erpnext.com](https://translate.erpnext.com). We update translations into erpnext every week.

Issue #4686, opened 2016-01-25
while selecting "Mode of payment" other than Cash ,we can not paid .showing message pop up as "Missing Values Required Amount Paid"


ShraddhaRanjane:
Steps To Reproduce:-
1. Go to "POS" module
2.Select "Sales Invoice" and Start
3.select Customer and Item
4.click on "Pay"button
5.select 'Mode of payment' as- Credit Card,Bank Draft (i.e except "Cash")
6.click on "Pay" button again
7.Message showing -
Missing Values Required
"Amount Paid"

retroneo:
Workaround:   If you switch "mode of Payment" to "Cash" first and then switch "mode of payment" to the desired payment method, there is no error and the sale will complete.

PR #4938, opened 2016-03-07
Paypal Here Credit Card Reader Basic Support

a81j:
Basic support for PayPal Here credit card reader on iOS and Android.


retroneo:
Issue #4686 is still present.
It would be ideal if payment methods were plugins - more modular.

Issue #4828, opened 2016-02-17
Unable to pay in ERPNext POS


olbanana:
Everything is working fine and as required. But when I go to pay using the POS screen I don't get the option to pay off a transaction.

When I click on pay after doing my selections, it says missing values required and the popup after that shows up without any button to pay off the transaction.

Is this an issue with ERPNext or is it something with my configuration. Will highly appreciate any input on this as currently my implementation is live on premises and this is causing huge issues.

![screen shot 2016-02-17 at 8 18 51 pm](https://cloud.githubusercontent.com/assets/605724/13114107/e169edd8-d5b3-11e5-876a-a1ea5bd2ec81.png)
![screen shot 2016-02-17 at 8 19 00 pm](https://cloud.githubusercontent.com/assets/605724/13114108/e1c72282-d5b3-11e5-8833-9bee5c91ce86.png)

rmehta:
Do you see anything in your JS console?

Did you add "Amount Paid"?

olbanana:
Actually, based on my understanding of the issue... this function is not working altogether... as in nothing happens here.

As you can see the total amount is empty when this should have the actual value of the transaction it normally have.

Then even if I enter the amount paid it doesn't calculate the change as it normally does.

My JSConsole does throughs an error but I don't think thats the cause of it. The error attached generates when I click on pay.


![screen shot 2016-02-18 at 2 11 39 pm](https://cloud.githubusercontent.com/assets/605724/13138622/0f049b30-d64a-11e5-8198-dacbbd85d88e.png)
![screen shot 2016-02-18 at 2 13 23 pm](https://cloud.githubusercontent.com/assets/605724/13138623/0f53a144-d64a-11e5-8b67-c58fb02ba833.png)

nabinhait:
Can't replicate the issue. Are you using the latest version?

retroneo:
Can replicate here if select non-cash as default mode of payment in POS profile.

For me, error disappears if switch to cash and then back to other payment method.

Issue #4289, opened 2015-11-08
POS Payment not possible


robertbecht:
The amount is prefilled in the amount te be paid (old request :) but than blocks, and gives errors msg that amount needs to be entered

![image](https://cloud.githubusercontent.com/assets/1532439/11170350/e5e76d8a-8bd1-11e5-8f09-355e0f2cb596.png)

chabito79:
I'm having the same problem! :(

retroneo:
Many with same problem ( See #4686 / #4828 ) - still present in  v6.24.5

PR #4939, opened 2016-03-07
PayPal Here credit card reader - basic support

a81j:
Basic support for PayPal Here credit card reader on iOS and Android

### DIFF
--- a/erpnext/public/js/pos/pos.js
+++ b/erpnext/public/js/pos/pos.js
@@ -492,6 +492,8 @@ erpnext.pos.PointOfSale = Class.extend({
 		dialog.set_primary_action(__("Pay"), function() {
 			var values = dialog.get_values();
 			var is_cash = values.mode_of_payment === __("Cash");
+			var is_paypal_here = values.mode_of_payment === __("PayPal Here");
+			
 			if (!is_cash) {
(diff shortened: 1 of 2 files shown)


retroneo:
The long-standing major issue still found in v6.24.5  #4686 / #4828 / #4289 is still present.

It would be ideal if payment methods were plugins - more modular.

erpnext/public/js/pos/pos.js
```diff
@@ -505,6 +507,37 @@ erpnext.pos.PointOfSale = Class.extend({
 			me.frm.set_value("write_off_amount", me.frm.doc.grand_total - paid_amount);
 			me.frm.set_value("outstanding_amount", 0);
 
+            if (is_paypal_here) {
+                var invoice = {
+                    "merchantEmail": "test@test.com",
```
anandpdoshi: Should not be hardcoded

anandpdoshi:
> The long-standing major issue still found in v6.24.5 #4686 / #4828 / #4289 is still present.
Thanks for pointing out these issues. I've fixed this issue.

rmehta:
Again POS payment extensions can be moved to a separate app. @saurabh6790

PR #4942, opened 2016-03-08
Whitelist get_month_details

javierwonghf:
This is useful when manually pulling the salary structure in custom scripts for processing of other customized deductables or perhaps contributions. There is a need to get month details using just the fiscal year and month.


anandpdoshi:
cherry picked the final comment and merged manually

PR #4947, opened 2016-03-09
[minor]added logging for status change of issue

RobertSchouten:
added logging and required resolution details also removed buttons as they seemed redundant

### DIFF
--- a/erpnext/support/doctype/issue/issue.js
+++ b/erpnext/support/doctype/issue/issue.js
@@ -4,9 +4,9 @@ frappe.ui.form.on("Issue", {
 	},
 
 	"refresh": function(frm) {
-        last_status = frm.doc.status;
+		last_status = frm.doc.status;
 	},
-    "status": function(frm) {dialog(frm)}
+	"status": function(frm) {dialog(frm)}
 });
 function dialog(frm){
     var d = new frappe.ui.Dialog({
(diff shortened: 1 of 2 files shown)


erpnext/support/doctype/issue/issue.js
```diff
@@ -4,16 +4,45 @@ frappe.ui.form.on("Issue", {
...
-				frm.set_value("status", "Open");
-				frm.save();
-			}, __("Status"));
-		}
-	}
+        last_status = frm.doc.status;
+	},
+    "status": function(frm) {dialog(frm)}
 });
+function dialog(frm){
```
rmehta: bad idea to declare globals.

erpnext/support/doctype/issue/issue.js
```diff
@@ -4,16 +4,45 @@ frappe.ui.form.on("Issue", {
...
+                        },
+                        {
+                            "fieldtype": "Button",
+                            "label": __("Submit")
+                        }
+                    ]
+                });
+                d.set_value("resolution_details",cur_frm.doc.resolution_details)
+                d.get_input("submit").on("click", function () {
+                    if (d.get_value("resolution_details")===""){
```
rmehta: Just set `reqd: 1` in the field declaration

erpnext/support/doctype/issue/issue.js
```diff
@@ -4,16 +4,45 @@ frappe.ui.form.on("Issue", {
...
+    var d = new frappe.ui.Dialog({
+                    title: __("Details for "+ frm.doc.status),
+                    fields: [
+                        {
+                            "fieldtype": "Text",
+                            "label": __("Resolution Details"),
+                            "reqd": 1,
+							"name":"resolution_details"
+                        },
+                        {
```
rmehta: set `primary_action` instead of a button. The current UI pattern is for the submit button on the toolbar not on the form!

rmehta:
Also would like to make this optional. Comment should be optional, like in the current pattern.

PR #4959, opened 2016-03-10
Improvements about Item Prices Report

ShashaQin:
1. add Item Group column
2. widen Sales Price List and Purchase Price List
3. swap price_list_rate with price_list in Sales Price List and Purchase Price List
4. replace currency with currency symbol

### DIFF
--- a/erpnext/stock/report/item_prices/item_prices.py
+++ b/erpnext/stock/report/item_prices/item_prices.py
@@ -45,10 +45,9 @@ def get_item_details():
 
 	item_map = {}
 
-	for i in frappe.db.sql("select it.item_group as item_group, it.name as name, item_name, it.description as description, \
-		stock_uom from tabItem it, `tabItem Group` itg \
-		where it.item_group = itg.name and it.disabled = 0\
-		order by it.item_group, item_code", as_dict=1):
+	for i in frappe.db.sql("select name, item_group, item_name, description, \
+		stock_uom from tabItem \
+		order by item_code, item_group", as_dict=1):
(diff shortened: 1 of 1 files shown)


ShashaQin:
5. exclude disabled item

erpnext/stock/report/item_prices/item_prices.py
```diff
@@ -58,9 +59,9 @@ def get_price_list():
 	rate = {}
 
 	price_list = frappe.db.sql("""select ip.item_code, ip.buying, ip.selling,
-		concat(ip.price_list, " - ", ip.currency, " ", ip.price_list_rate) as price
-		from `tabItem Price` ip, `tabPrice List` pl
-		where ip.price_list=pl.name and pl.enabled=1""", as_dict=1)
+		concat(cu.symbol, " ", round(ip.price_list_rate,2), " - ", ip.price_list) as price
```
nabinhait: Why if currency symbol does not exists? It is not mandatory.

erpnext/stock/report/item_prices/item_prices.py
```diff
@@ -45,9 +45,10 @@ def get_item_details():
 
 	item_map = {}
 
-	for i in frappe.db.sql("select name, item_name, description, \
-		stock_uom from tabItem \
-		order by item_code", as_dict=1):
+	for i in frappe.db.sql("select it.item_group as item_group, it.name as name, item_name, it.description as description, \
```
nabinhait: item_group field already exists in Item, no need to use join query

ShashaQin:
fixed.

PR #4974, opened 2016-03-14
Simplify Time Log

ShashaQin:
Hi,
Our staff need to record the hours spent in an easier way on a daily basis.
We'd like to simplify time log by using a check box named 'Simplified Time Log', if checked, the workflow goes to our easier way, otherwise can still keep current way if unchecked. We'd like to use date_worked (new created) and hours to record our time log, this way is the Microsoft Project way. Being too precise consume too much time on the project management side. We think it's not worth to stick to the seconds when representing actual hours.
We adjusted calendar, gantt chart and summary report accordingly.


ShashaQin:
Simplified Time Log.

rmehta:
@ShashaQin thanks! Will be real cool if you an also add a few screenshots in your PR. Otherwise we have to manually merge and migrate a local account to see what changes you have done.

ShashaQin:
Thank you for your feedback. The following are screenshots when 'Simplified Time Log' checked.
project setting page
time log form
calendar
daily time log summary

rmehta:
@ShashaQin why not make a button for "Complete" and then save the diff between started and completed?

ShashaQin:
Sorry, I'm not very understand about 'Complete' button exactly. This simply way replace 'from_date' and 'to_date' with 'date_worked'.

rmehta:
@ShashaQin instead of this pattern, would love to find a way to make the current time log simpler. "Simplified Time Log" does not clearly say what is the difference.

In the current design you can choose the start time and add the number of hours, so its just 2 entries.

What could be a simple design is to have a start and complete button so that when you start work, you can click on "start" and then  when you complete, click on a "complete" button. That way you do not need to enter time manually.

ShashaQin:
Thanks your reply, but I think the real purpose of my simple way is removing the concept of recording precise time. My new filed date_worked just logs date (yyyy-mm-dd), not logging time (hh-mm-ss) like from_date and to_date. Staff can log multiple records on same day. So the design of the start time and complete button is still logging time which is different with my way.

ShashaQin:
We also implemented a report to supervise hours spent on projects and issues by using the simplified time log. The followings are the screenshots of this report:
report
report-2
report-3

rmehta:
report looks great. Can you rebase your branch?

ShashaQin:
However, now this report data is just based on simplified Time log case which is using date_worked record time log instead of from_date and to_date. And this report assumes time log include Issue field.

ShashaQin:
would you like more examples to show the way we use it?

rmehta:
@ShashaQin I like the report. But I don't like "Simplified Time Log".

Can you just post the report?

ShashaQin:
Branch rebased and updated check box name from 'Simplified Time Log' to 'Date Only'

rmehta:
needs to be updated for new Timesheet design

PR #4976, opened 2016-03-14
[fixes] set amount in PR by checking party_account_currency field

saurabh6790:
fetch value from grand total field while creating payment request from si

### DIFF
--- a/erpnext/accounts/doctype/payment_request/payment_request.py
+++ b/erpnext/accounts/doctype/payment_request/payment_request.py
@@ -211,7 +211,7 @@ def get_amount(ref_doc, dt):
 		grand_total = flt(ref_doc.grand_total) - flt(ref_doc.advance_paid)
 		
 	if dt == "Sales Invoice":
-		grand_total = flt(ref_doc.outstanding_amount)
+		grand_total = flt(ref_doc.grand_total)
 		
 	if grand_total > 0 :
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -211,7 +211,10 @@ def get_amount(ref_doc, dt):
 		grand_total = flt(ref_doc.grand_total) - flt(ref_doc.advance_paid)
 		
 	if dt == "Sales Invoice":
-		grand_total = flt(ref_doc.outstanding_amount)
+		if ref_doc.party_account_currency == ref_doc.currency:
+			grand_total = flt(ref_doc.grand_total)
```
nabinhait: grand_total = flt(ref_doc.outstanding_amount)

erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -211,7 +211,10 @@ def get_amount(ref_doc, dt):
...
+		else:
+			grand_total = flt(ref_doc.grand_total) / ref_doc.conversion_rate
```
nabinhait: grand_total = flt(ref_doc.outstanding_amount) / ref_doc.conversion_rate

Issue #4235, opened 2015-10-27
Filter user by role


deep1991:
Right now, In HR module in Employee doctype, for leave approver child table we get all user in drop down and not just user who has role "Leave Approver". But it should be just those user who has this role of "Leave Approver" and not all the user of the system.

vjFaLk:
I cannot replicate this on my build.

vjFaLk:
My bad, I was checking in the Leave Application form. I can replicate the issue.

vjFaLk:
Fixed

deep1991:
Thanks for fixing the issue. :+1:

PR #4979, opened 2016-03-14
Improved Leave Application / Approver feature

vjFaLk:
- [x] When assigning Leave Approver to an Employee, only Users with the Leave Approval Role will show up (https://github.com/frappe/erpnext/issues/4235)
- [x] If an Employee is not assigned a Leave Approver, all the Leave Approvers will be listed. 

[Minor] Leave Approver Name doesn't get set to current user if no Leave Approver is selected in the Leave Application form.

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.py
+++ b/erpnext/hr/doctype/leave_application/leave_application.py
@@ -9,6 +9,7 @@ from frappe.utils import cint, cstr, date_diff, flt, formatdate, getdate, get_li
 from erpnext.hr.utils import set_employee_name
 from erpnext.hr.doctype.leave_block_list.leave_block_list import get_applicable_block_dates
 from erpnext.hr.doctype.employee.employee import get_holiday_list_for_employee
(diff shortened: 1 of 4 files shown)


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -247,12 +247,23 @@ def get_approvers(doctype, txt, searchfield, start, page_len, filters):
...
+	employee_user = frappe.get_value("Employee", filters.get("employee"), "user_id")
+
+	approvers_list = frappe.db.sql("""select user.name, user.first_name, user.last_name from
 		tabUser user, `tabEmployee Leave Approver` approver where
 		approver.parent = %s
 		and user.name like %s
 		and approver.leave_approver=user.name""", (filters.get("employee"), "%" + txt + "%"))
 
+	if not approvers_list:
+		approvers_list = frappe.db.sql("""select user.name, user.first_name, user.last_name from
```
rmehta: commonify

Issue #4823, opened 2016-02-17
"Invalid Date" in Employee Allocation Tool


vjFaLk:
![Tool](http://i.imgur.com/NlMfXgKl.jpg)

Traceback : Traceback (innermost last):
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/app.py", line 57, in application
    response = frappe.handler.handle()
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/handler.py", line 20, in handle
    execute_cmd(cmd)
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/handler.py", line 37, in execute_cmd
    ret = frappe.call(method, **frappe.form_dict)
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/__init__.py", line 801, in call
    return fn(*args, **newargs)
  File "/home/frappe/press/benches/1602110817/apps/erpnext/erpnext/hr/doctype/employee_attendance_tool/employee_attendance_tool.py", line 24, in get_employees
    filters={"att_date": date}):
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/__init__.py", line 922, in get_list
    return frappe.model.db_query.DatabaseQuery(doctype).execute(None, *args, **kwargs)
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 50, in execute
    result = self.build_and_run()
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 58, in build_and_run
    args = self.prepare_args()
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 73, in prepare_args
    self.build_conditions()
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 196, in build_conditions
    self.build_filter_conditions(self.filters, self.conditions)
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 214, in build_filter_conditions
    conditions.append(self.prepare_filter_condition(f))
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/model/db_query.py", line 245, in prepare_filter_condition
    value = getdate(f.value).strftime("%Y-%m-%d")
  File "/home/frappe/press/benches/1602110817/apps/frappe/frappe/utils/data.py", line 39, in getdate
    return parser.parse(string_date).date()
[...]

PR #5019, opened 2016-03-20
[fixes] set todays date on employee attendance tool

saurabh6790:
- Issue/WN-SUP17706


vjFaLk:
Same issue : https://github.com/frappe/erpnext/issues/4823

PR #5020, opened 2016-03-20
[formatting] Added between in trends.py

rohitwaghchaure:


nabinhait:
Quotation, Sales Order and Purchase Order Trends report breaks... did not find posting_date field.. it is transaction_date in those doctype

rohitwaghchaure:
Fixed issue of posting_date not found in Sales Order, Quotation, Purchase Order Trends report.

PR #5023, opened 2016-03-21
[fix][balance sheet] Check if precious fiscal  year is closed

nabinhait:


nabinhait:
screen shot 2016-03-21 at 2 39 32 pm

PR #5025, opened 2016-03-21
added barcode and qr code min.js

gangadharkadam:
added barcode and qr code min js files and added same in build.json of erpnext.min.js.

Now user can add following sample code in any custom print format to print barcode image
`
<script> $("#barcode").JsBarcode("{{ doc.name | string }}" ,{  width: 2,  height: 80}) </script>`

Following are issue , threads for barcode image.
1. https://discuss.erpnext.com/t/feature-request-barcode-generator/4497
2. https://discuss.erpnext.com/t/rendering-a-barcode-using-html/4563/3
3. https://discuss.erpnext.com/t/printing-barcoded-fields-on-custom-print-format/4521


rmehta:
Add it to POS print format? What is the point of adding the lib without using it anywhere

rmehta:
suggest a standard pattern. Why is it needed in POS format

Or better,

Just write a help article for those who want to use it.

Issue #5006, opened 2016-03-17
Purchase Receipt with $0 amount


DevenK:
There's an issue with zero amount Purchase Orders\Receipts. If an invoice is created for these items, they don't change from a "To Bill" status to a "Completed" status.

Please see this post for more information: [Purchase Receipt with $0 amount - To Bill](https://discuss.erpnext.com/t/purchase-receipt-with-0-amount-to-bill/11241)

Thanks.

DevenK:
I really hope this issue can be fixed. I've imported 387 Purchase Receipts and 333 Delivery Notes, all with a zero amount, and all of them have a "To Bill" status. This means that there are a lot of notifications on our ERPNext system with no way to get rid of them.

PR #5033, opened 2016-03-21
Fix: Set billing status of $0 to Billed

bcornwellmott:
Checks to see if ref_doc_qty is 0, and if so, sets per_billed to 100

RE: [Purchase Receipt with $0 amount #5006](https://github.com/frappe/erpnext/issues/5006)


bcornwellmott:
Didn't work as expected...

PR #5050, opened 2016-03-23
Fixes Translations

### LINKED ISSUE #4955: Error message for required "remarks" in journal entry does not show proper translation in spanish.
Working with the platform with es selected as a language, one goes to
Account> Journal Entry > New

You fill up all the fields, except  "User Remarks"(en) = "Observaciones" (es).  When you click save, an error message telling you that "User Remarks is required"  pops up.  Once "User Remarks" is properly filled, the error message goes away. This issue specifically refers to the text shown in spanish for this error message.  It shows  "_Remarks_ es necesario" instead of "_Observaciones_ es necesario".  I checked both es_PE.csv  and es.csv  and both seem to have complete translations properly setup,  so I am inclined to think this error message is simply some variable error in the software.   I have not tested other languages to see if the problem replicates.

neilLasrado:
fixed #4955

### DIFF
--- a/erpnext/accounts/doctype/c_form/c_form.py
+++ b/erpnext/accounts/doctype/c_form/c_form.py
@@ -18,17 +18,17 @@ class CForm(Document):
 					`tabSales Invoice` where name = %s and docstatus = 1""", d.invoice_no)
 
 				if inv and inv[0][0] != 'Yes':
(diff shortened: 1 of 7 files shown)


Tropicalrambler:
If this check failed, it means the translation changes are not OK?

rmehta:
No - its just that all the tests did not pass at that point.

PR #5092, opened 2016-03-30
[Minor fix] Email URL

KanchanChauhan:
Email invitation URL fixed and minor cleanup

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -124,9 +124,9 @@ class Project(Document):
 			from `tabPurchase Invoice Item` where project = %s and docstatus=1""", self.name)
 
 		self.total_purchase_cost = total_purchase_cost and total_purchase_cost[0][0] or 0
-				
+			
 	def send_welcome_email(self):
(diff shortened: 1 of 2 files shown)


erpnext/projects/doctype/project/project.py
```diff
@@ -126,7 +126,7 @@ def update_purchase_costing(self):
 		self.total_purchase_cost = total_purchase_cost and total_purchase_cost[0][0] or 0
 				
 	def send_welcome_email(self):
-		url = get_url("/project/{0}".format(self.name))
+		url = get_url("/project?{0}".format(self.name))
```
anandpdoshi: `'/project/?name={0}'.format(self.name)` vs `'/project/?{0}'.format(self.name)`

which one works? Ideally query string parameters are key value pairs.

KanchanChauhan: My bad, it should be `/project/?name={0}'.format(self.name)`

rmehta:
@KanchanChauhan can you rebase and push again? The branch as conflicts.

PR #5095, opened 2016-03-30
[Enhancement] Added currency symbol on RFQ form of supplier portal, currency validation for party

rohitwaghchaure:


erpnext/templates/pages/rfq.py
```diff
@@ -29,3 +29,15 @@ def check_supplier_has_docname_access(supplier):
 		where supplier = '{supplier}'""".format(supplier=supplier)):
 		status = False
 	return status
+
+def unauthrized_user(supplier):
```
nabinhait: spelling?

rohitwaghchaure: Done

PR #5104, opened 2016-03-31
[minor] Fixed validation message for disabled employee & changed customer in lead to no-copy

neilLasrado:


erpnext/hr/doctype/employee/employee.py
```diff
@@ -133,11 +133,11 @@ def validate_status(self):
...
 		if not self.status == 'Active':
 			return
-		enabled = frappe.db.sql("""select name from `tabUser` where
-			name=%s and enabled=1""", self.user_id)
+		enabled = frappe.db.get_value("User", self.user_id, "enabled")
 		if not enabled:
-			throw(_("User {0} is disabled").format(
-				self.user_id), EmployeeUserDisabledError)
+			frappe.throw(_("User {0} does not exist.").format(self.user_id))
+		if enabled == 0:
```
vjFaLk: Neilji, how will it return 0?

Just do frappe.get_value() to get enabled and pass User as filter.

erpnext/hr/doctype/employee/employee.py
```diff
@@ -133,11 +133,11 @@ def validate_status(self):
 	def validate_for_enabled_user_id(self):
 		if not self.status == 'Active':
 			return
-		enabled = frappe.db.sql("""select name from `tabUser` where
-			name=%s and enabled=1""", self.user_id)
+		enabled = frappe.db.get_value("User", self.user_id, "enabled")
 		if not enabled:
-			throw(_("User {0} is disabled").format(
```
rmehta: `if enabled is None:`

anandpdoshi:
merged manually

PR #5114, opened 2016-04-01
[Minor] "Join" URL fixed for website invitation

KanchanChauhan:


erpnext/templates/includes/project_row.html
```diff
@@ -0,0 +1,28 @@
+{% if doc.status=="Open" %}
```
rmehta: I moved this to `template/includes/project/project_row.html`

PR #5120, opened 2016-04-01
[fix] received and delivered items to be billed report based on percentage billed

nabinhait:


anandpdoshi:
Change `project_name` to `project`

PR #5121, opened 2016-04-02
Add User Remark to Quick Entry Form in Journal Entry

kaflesudip:
User Remark is a required field in Journal entry. It is not provided in Quick Entry due to which we again need to add it via the main form. Including User Remark will make it easier.

Old Form:

quick_entry_old

New Form:

new_quick_entry


anandpdoshi:
Thanks!

PR #5133, opened 2016-04-04
[Fixes]Portal, Supplier can view their addresses, default buying price list …

rohitwaghchaure:
…issue

### DIFF
--- a/erpnext/shopping_cart/cart.py
+++ b/erpnext/shopping_cart/cart.py
@@ -24,7 +24,7 @@ def set_cart_count(quotation=None):
 
 @frappe.whitelist()
 def get_cart_quotation(doc=None):
-	party = get_customer()
+	party = get_party()
 
 	if not doc:
 		quotation = _get_cart_quotation(party)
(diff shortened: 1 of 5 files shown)


erpnext/shopping_cart/cart.py
```diff
@@ -276,20 +276,24 @@ def set_taxes(quotation, cart_settings):
...
+def get_party(user=None):
 	if not user:
 		user = frappe.session.user
 
-	customer = frappe.db.get_value("Contact", {"email_id": user}, "customer")
+	party = frappe.db.get_value("Contact", {"email_id": user}, ["customer", "supplier"], as_dict=1)
+
+	if party:
+		party, doctype = get_party_detail(party)
+
```
anandpdoshi: This could be simply written as:

```
party = frappe.db.get_value("Contact", {"email_id": user}, ["customer", "supplier"], as_dict=1)
if party:
    party_doctype = 'Customer' if party.customer else 'Supplier'
    party = party.customer or party.supplier

...

    if party:
        return frappe.get_doc(party_doctype, party)

    else:
        ...
```

rohitwaghchaure: @anandpdoshi 

Done

PR #5141, opened 2016-04-06
[Minor Fixes]

KanchanChauhan:
1. Added "Add multiple rows" to Production Planning Tool
2. Make button disabled for completed Purchase orders
3. Fixed "Get last purchase rate" in case where it is not null
4. Website URL in welcome Email

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -103,7 +103,7 @@ class PurchaseOrder(BuyingController):
 					d.price_list_rate = d.base_price_list_rate / conversion_rate
 					d.rate = d.base_rate / conversion_rate
 				else:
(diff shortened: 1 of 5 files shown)


erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -103,10 +103,7 @@ def get_last_purchase_rate(self):
 					d.price_list_rate = d.base_price_list_rate / conversion_rate
 					d.rate = d.base_rate / conversion_rate
 				else:
-					# if no last purchase found, reset all values to 0
-					for field in ("base_price_list_rate", "base_rate",
-						"price_list_rate", "rate", "discount_percentage"):
-							d.set(field, 0)
+					msgprint("Last purchase rate not found")
```
rmehta: add strings with _ so they become translatable like

`msgprint("Last purchase rate not found")`

PR #5149, opened 2016-04-07
[Fix] Show a message while user tries to create Material request with BOM

shreyasp:
- Validation message, when a user tries to create material request from BOM, which contains only one item which is non-stocked.

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.js
+++ b/erpnext/stock/doctype/material_request/material_request.js
@@ -56,7 +56,7 @@ erpnext.buying.MaterialRequestController = erpnext.buying.BuyingController.exten
 				if(doc.material_request_type === "Purchase")
 					cur_frm.add_custom_button(__("Supplier Quotation"),
 					this.make_supplier_quotation, __("Make"));
-				
+
 				if(doc.material_request_type === "Manufacture" && doc.status === "Submitted")
 					cur_frm.add_custom_button(__("Production Order"),
 					this.raise_production_orders, __("Make"));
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/material_request/material_request.js
```diff
@@ -127,14 +127,19 @@ erpnext.buying.MaterialRequestController = erpnext.buying.BuyingController.exten
...
-					$.each(r.message, function(i, item) {
-						var d = frappe.model.add_child(cur_frm.doc, "Material Request Item", "items");
-						d.item_code = item.item_code;
-						d.description = item.description;
-						d.warehouse = values.warehouse;
-						d.uom = item.stock_uom;
-						d.qty = item.qty;
-					});
+					if(!r.message){
+						frappe.throw(__("BOM does not contain any stocked items."))
```
anandpdoshi: should be `stocked item`. Also remove the dot at the end.

erpnext/stock/doctype/material_request/material_request.js
```diff
@@ -127,14 +127,19 @@ erpnext.buying.MaterialRequestController = erpnext.buying.BuyingController.exten
...
+					}
+					else{
```
anandpdoshi: put else in the previous line and a space after `else`

shreyasp:
@anandpdoshi Done the requested changes

anandpdoshi:
merged manually after changing 'stocked' to 'stock'

PR #5170, opened 2016-04-09
[enhancement] website: optionally show products as list

anandpdoshi:
image

@rmehta 
There is a dilemma here: `home_page_is_products` setting is part of frappe's website settings.

As a temporary measure, I added `products_as_list` in website settings: https://github.com/frappe/frappe/pull/1688

Ideally, we will need to add these 2 fields as Custom Fields via fixtures and handle home page logic about products.


anandpdoshi:
merged manually

PR #5177, opened 2016-04-11
[Fix] Default letter head

rohitwaghchaure:


erpnext/public/js/controllers/transaction.js
```diff
@@ -308,7 +308,14 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		if (this.frm.doc.posting_date) var date = this.frm.doc.posting_date;
 		else var date = this.frm.doc.transaction_date;
-		set_party_account(set_pricing);
+		
+		if (frappe.meta.get_docfield(this.frm.doctype, "shipping_address")){
```
nabinhait: shipping_address field is there in sales cycle as well, but it should be called only for purchase cycle. Use if condition based on doctype.

PR #5219, opened 2016-04-15
Simple Stock Balance Report

cgpurbaugh:
Created a simple stock balance report by copying and modifying the default stock balance report.


fderyckel:
Any chance we could do the same for stock ledger?
Thanks

fderyckel:
Any chance we could do the same for stock ledger?
Thanks

cgpurbaugh:
@fderyckel If this works, then it should be fairly easy.  But that's if it works.  I'm like a monkey with a typewriter right now.

rmehta:
Will be done by @saurabh6790

cgpurbaugh:
Thanks @rmehta & @saurabh6790 !

PR #5261, opened 2016-04-25
Moved Legacy Print Formats to Documentation

vjFaLk:


anandpdoshi:
@vjFaLk sorry 😦 we don't want to encourage the use of js print formats

PR #5268, opened 2016-04-26
Moved email salary slip to employee under HR Settings

neilLasrado:
Moved send email checkbox from salary slip to a global setting under hr settings


anandpdoshi:
@neilLasrado test case?

PR #5305, opened 2016-05-03
[enhancement]Add an option to ignore projected qty toward material requests

chaoyee:
If checked, Projected qty will be ignored. If material requests are generated based on sales orders, the qty reserved by sales orders will be included in projected qty, which causes the qty deduction of the material requests. Thus, ignorance of projected qty should be checked in this situation.


erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.json
```diff
@@ -646,6 +646,33 @@
...
+   "depends_on": "create_material_requests", 
+   "description": "If checked, Projected qty will be ignored. If material requests are generated based on sales orders, the qty reserved by sales orders will be included in projected qty, which causes the qty deduction of the material requests. Thus, ignorance of projected qty should be checked in this situation.", 
+   "fieldname": "ignore_projected_qty", 
+   "fieldtype": "Check", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_list_view": 0, 
+   "label": "Ignore Projected Qty",
```
nabinhait: "Ignore Projected Qty" label does not explain the usage of it. How about "Create Material Requests for All Required Qty"?

chaoyee: Sounds great!
Let's do it!

nabinhait: Please make the changes and resent the PR

chaoyee: "ignore_projected_qty" has been changed to "create_material_requests_for_all_required_qty".

erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py
```diff
@@ -394,7 +394,12 @@ def get_requested_items(self):
 			total_qty = sum([flt(d[0]) for d in so_item_qty])
 			if total_qty > item_projected_qty.get(item, 0):
 				# shortage
-				requested_qty = total_qty - flt(item_projected_qty.get(item))
+				if self.create_material_requests_for_all_required_qty:
```
anandpdoshi: Should this depend on the condition in line 395: `if total_qty > item_projected_qty.get(item, 0):`?

nabinhait:
Added the feature via https://github.com/frappe/erpnext/pull/5327

erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py
chaoyee: Either ways, both can reach the same result.

The "minimum order qty" is processed at:
[https://github.com/frappe/erpnext/blob/develop/erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py#L400-L401]

chaoyee:
Thanks!

PR #5306, opened 2016-05-03
[Fix] Disallow leave allocation for inactive employees

shreyasp:
- Inactive employees will not allocated leaves while using leave allocation tool.


erpnext/hr/doctype/leave_control_panel/leave_control_panel.py
```diff
@@ -21,16 +21,25 @@ def get_employees(self):
 					condition += " and " + l[1]+ "= '" +l[0] +"'"
 				flag = 1
 		emp_query = "select name from `tabEmployee` "
+
 		if flag == 1:
 			emp_query += condition
+
+		# Get only those employees which are active. If you have 'where' clause in the query append
+		# employee status, else append condition in 'where' clause
+		if 'where' in condition:
```
anandpdoshi: `where` is there anyway as per line 14. Also no need for comment about this!

shreyasp:
@anandpdoshi Please review, done requested changes in the function.

anandpdoshi:
@shreyasp your frappe seems to be in develop!

erpnext/hr/doctype/leave_control_panel/leave_control_panel.json
```diff
@@ -2,6 +2,7 @@
  "allow_copy": 1, 
  "allow_import": 0, 
  "allow_rename": 0, 
+ "beta": 0,
```
anandpdoshi: Use master branch of `frappe` so that `beta`, `track_seen`, etc. does not appear in hotfix branch

anandpdoshi:
Needs a test case

PR #5340, opened 2016-05-12
[Fix] Sales Invoice shows default date while no default mentioned

rohitwaghchaure:


nabinhait:
Did you check the default value of posting date in all transactions? It should be set as "Today" in all forms.

And also there is a trigger function on posting_date, which sets due date in invoice.

Check if that is triggered if default value loaded based docfield default

mkumar9699:
I have already removed default today for posting date in "Sales Invoice  Customized form" and no custom script to initialize for posting date. We are on cloud.


Thanks

Mithilesh

________________________________

rmehta:
@rohitwaghchaure send this on hotfix ?

PR #5353, opened 2016-05-16
Small improvement in maintenance scheduler

MaxMorais:
This improvement is intended to start disable blocking Maintenance Schedule submit, when the `Sales Person` does not have a `user_id` in the Employee!

The old logic try to use the `sales_person` email_id or the `owner` to create the event, but due some fails in the code, it dont occurs!

### DIFF
--- a/erpnext/support/doctype/maintenance_schedule/maintenance_schedule.py
+++ b/erpnext/support/doctype/maintenance_schedule/maintenance_schedule.py
@@ -49,26 +49,28 @@ class MaintenanceSchedule(TransactionBase):
 
 			if d.sales_person not in email_map:
 				sp = frappe.get_doc("Sales Person", d.sales_person)
-				email_map[d.sales_person] = sp.get_email_id()
+				try:
+					email_map[d.sales_person] = sp.get_email_id()
+				except frappe.ValidationError:
+					pass
 
 			scheduled_date = frappe.db.sql("""select scheduled_date from
 				`tabMaintenance Schedule Detail` where sales_person=%s and item_code=%s and
 				parent=%s""", (d.sales_person, d.item_code, self.name), as_dict=1)
 
 			for key in scheduled_date:
-				if email_map[d.sales_person]:
-					description = "Reference: %s, Item Code: %s and Customer: %s" % \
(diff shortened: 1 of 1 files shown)


erpnext/support/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -49,26 +49,28 @@ def on_submit(self):
 
 			if d.sales_person not in email_map:
 				sp = frappe.get_doc("Sales Person", d.sales_person)
-				email_map[d.sales_person] = sp.get_email_id()
+				try:
+					email_map[d.sales_person] = sp.get_email_id()
+				except frappe.ValidationError:
+					pass
```
rmehta: Maybe it should send a msgprint saying the email ID is not valid and then pass. If it fails silently then the user might expect the event to be made.

MaxMorais: @rmehta you are right I was thiking the same!

MaxMorais: @rmehta Done!

erpnext/support/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -47,28 +47,37 @@ def on_submit(self):
...
 				sp = frappe.get_doc("Sales Person", d.sales_person)
-				email_map[d.sales_person] = sp.get_email_id()
+				try:
+					email_map[d.sales_person] = sp.get_email_id()
+				except frappe.ValidationError:
+					no_email_sp.append(d.sales_person)
+					
+			if no_email_sp:
+				frappe.msgprint(
+					frappe._("Setting Events to <b>{0}</b>, since the Employee attached to the below Sales Persons does not have a User ID<br>{1}").format(
```
rmehta: Cool, lets leave out the HTML `<b>` `<br>` tags in the message. Not good for translations.

MaxMorais: :+1:

PR #5359, opened 2016-05-17
Inserted a blank section break in Designation DocType to allow custom…

kxt5258:
Current: Doesn't allow custom fields above "Designation" field
--Inserted a blank section break as the first field in Designation DocType to allow custom fields above designation field


nabinhait:
@kxt5258 Do you really required those custom fields before designation field? This problem exists in all document, you cannot add a custom field before first field in the document. And another thing is that, designation field gets hidden after saving the designation record, hence all your custom fields will appear at the top automatically after saving. Closing the pull request.

PR #5363, opened 2016-05-18
[Fix] Restrict access to setup wizard page if setup is already completed

rohitwaghchaure:


erpnext/setup/setup_wizard/setup_wizard.py
```diff
@@ -53,6 +54,14 @@ def setup_complete(args=None):
 				frappe.message_log.pop()
 
 			pass
+			
+def update_setup_wizrd_access():
```
nabinhait: spelling?

erpnext/setup/setup_wizard/setup_wizard.py
```diff
@@ -53,6 +54,14 @@ def setup_complete(args=None):
...
+	setup_wizard = frappe.get_doc('Page', 'setup-wizard')
+	for roles in setup_wizard.roles:
+		if roles.role == 'System Manager':
+			roles.role = 'Administrator'
```
nabinhait: why so?

PR #5365, opened 2016-05-18
[Translation] [Fix] Exception in setup wizard, item attributes Large and Extra Large has same translation

rohitwaghchaure:


anandpdoshi:
@rohitwaghchaure  why is it showing full file changed? Also did you fix this in the translator portal too?

PR #5379, opened 2016-05-20
[Fix] during making of material request from sales order, system fetching product bundle items only if so has both product bundle and purchase item

rohitwaghchaure:


nabinhait:
Send to hotfix?

PR #5398, opened 2016-05-26
Improvements about recurring documents

ShashaQin:
1. removed the label of column break under recurring 
2. removed the restriction of doc.recurring_id === doc.name on some recurring field which can give more flexibility to end user. (The current version is too strict and there is no way to recover from mistakes or backlogs)
3. make Next Date be able to be changed and calculated by from date value.


RobertSchouten:
@nabinhait ill be taking over from sasha but im playing catchup whats the reason for the dont-merge?

nabinhait:
@RobertSchouten Actually we have a plan to redesign the recurring settings with a separate doctype. Thats why I tagged it as "Don't merge". We will look into your PR while redesigning it.

rmehta:
@RobertSchouten will be awesome if you want to help us build it :)

The goal is to have a separate document "Subscription" that will contain the origin document (Invoice or whichever) and then add the frequency, duration of the contract, email information in that. The documents created will only have link to the "Subscription ID"

This way we don't duplicate the fields in all the documents.

We can make detailed mockups if you like.

RobertSchouten:
@rmehta we are interested in the changes we are currently in the process of migrating to v7
after that if budgeting allows will be happy to do

rmehta:
Please see #6313

PR #4898, opened 2016-02-29
General Ledger improvement for cost/management accountants

ShashaQin:
Fixed https://github.com/frappe/erpnext/pull/4747.

PR #5400, opened 2016-05-26
Added Project in GL Entry and General Ledger report

nabinhait:
Against #4898


IAGAdmin:
Just wondering why Support tickets/Issues are not included. 

Staff spends time and/or uses material to handle post sales issues. Related costs need to be reported to the boss as well :)

nabinhait:
@IAGAdmin Not a generic requirement, you can track that using cost center or customize yourself.

IAGAdmin:
That's ok.

We have already been using cost centres for tracking those costs broadly as we don't see creating new cost centres for every specific issue being an helpful workflow.

Moreover we have been adding customisations on the selling/purchasing cycle for tracking material, expenses and labour's cost in Projects as well as Supporting tickets/Issues like this one Ref. #4898. Btw thanks for partially merge it.

PR #5408, opened 2016-05-27
Fixed error in Gross Profit report

neilLasrado:


nabinhait:
Send the PR to hotfix

Issue #5348, opened 2016-05-13
Ability to setup tax rules based upon both county & city. 


jasonbrowne:
We recently finished our taxes and all of our taxes had to be organised by county.
Currently ErpNext only allows for creation of tax rules by city.

Feature request:
Ability to setup tax rules based upon both county & city.

jasonbrowne:
As all of our taxes had to be done by county I am sure there are several other US based companies that would benefit from, and need this feature as well.

rmehta:
@jasonbrowne are you a developer? We can help you build this.

rmehta:
Fixed

PR #5416, opened 2016-05-30
Tax Rule based on County

nabinhait:
Against #5348


jasonbrowne:
The tax rule based upon county works for us, in our testing so far. We did run into a  problem finding the specific tax rules we need to change as the tax rule list does not show city county and state. As a result of the city, county and state not showing we had to hunt through the tax rule list to find the rules we needed to change. 

Please see the attached jpg.
tax-rule-list

jasonbrowne:
The tax rule based upon county works for us, in our testing so far. We did run into a problem finding the specific tax rules we need to change as the tax rule list does not show city county and state. As a result of the city, county and state not showing we had to hunt through the tax rule list to find the rules we needed to change.
tax-rule-list

anandpdoshi:
Maybe we could use a title field and set it automatically based on what values are available

jasonbrowne:
Anything that makes it easy to identify the Tax Rules that need to be 
changed. If it showed the State, City, and County it would work good.

jasonbrowne:
Team is this part of ERPNext7 yet or will we have to add it to 7?

rmehta:
@jasonbrowne can you specify what issues you are facing?

jasonbrowne:
Sorry, It turns out the issue is with tax rules not importing was due to user error in not having the same settings on both systems. So nothing is wrong.

PR #5427, opened 2016-06-01
Update production_planning_tool.py

ShashaQin:
We found material request for non stock items can be generated when creating new material request, but not in Production Planning Tool. Production Manager has to be able to ask for non stock material as well. So we removed the condition 'item.is_stock_item = 1' when trying to get raw materials in production planning tool.


nabinhait:
`get_raw_materials()` function is also called for "Download Raw Materials" functionality. In the downloaded csv file, non-stock item should not be added.

nabinhait:
May be we should add a checkbox "Include non-stock items" in Production Planning Tool while raising material requests.

rmehta:
@ShashaQin can you add the checkbox @nabinhait is suggesting?

ShashaQin:
sure, @rmehta I'll try to add it soon.

nabinhait:
@ShashaQin Any update?

PR #5446, opened 2016-06-06
[Fixes] Purchase register, showing wrong value for total amount if tax has deducted.

rohitwaghchaure:


rohitwaghchaure:
Don't Merge

PR #5458, opened 2016-06-09
[Fix] POS page loading issue, if system has only one item

rohitwaghchaure:


rmehta:
Did you also fix the open day suggestions? Like warn the user before exiting the page?

rohitwaghchaure:
Sent pull request

PR #5459, opened 2016-06-10
Status updater for quotation/lost to lead, quotation lost to oppt, an…

aruizramon:
…d to set conversion as priority in lead

cleanup
squashed

### DIFF
--- a/erpnext/controllers/status_updater.py
+++ b/erpnext/controllers/status_updater.py
@@ -13,15 +13,17 @@ def validate_status(status, options):
 
 status_map = {
 	"Lead": [
-		["Converted", "has_customer"],
+		["Lost Quotation", "lost_quotation"],
 		["Opportunity", "has_opportunity"],
+		["Quotation", "has_quotation"],
+		["Converted", "has_customer"],
 	],
 	"Opportunity": [
 		["Quotation", "has_quotation"],
 		["Converted", "has_ordered_quotation"],
 		["Lost", "eval:self.status=='Lost'"],
+		["Lost", "has_lost_quotation"],
 		["Closed", "eval:self.status=='Closed'"]
-
 	],
 	"Quotation": [
 		["Draft", None],
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -80,6 +80,22 @@ class Lead(SellingController):
 	def has_opportunity(self):
 		return frappe.db.get_value("Opportunity", {"lead": self.name, "status": ["!=", "Lost"]})
 
+	def has_quotation(self):
+		return frappe.db.get_value(
+			"Quotation", {
+				"lead": self.name,
+				"status": ["!=", "Lost"],
+				"status": ["!=", "Cancelled"]
+				})
+
+	def lost_quotation(self):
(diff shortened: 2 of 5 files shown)


aruizramon:
Thought it would be good to have the set_status method applied to both quotations/lost quotations in Lead, and lost quotations made from opportunities. 

Also made it so that Converted is the first priority status to set in Lead - before checking of there is an opportunity, etc.

erpnext/controllers/status_updater.py
```diff
@@ -13,15 +13,17 @@ def validate_status(status, options):
 
 status_map = {
 	"Lead": [
-		["Converted", "has_customer"],
+		["Lost Quotation", "lost_quotation"],
 		["Opportunity", "has_opportunity"],
+		["Quotation", "has_quotation"],
+		["Converted", "has_customer"],
```
nabinhait: This should be first option, because if there is existing customer, status should not other than Converted. It is irrelevant to set lead's status as Opportunity / Quotation.

erpnext/crm/doctype/lead/lead.py
```diff
@@ -80,6 +80,22 @@ def has_customer(self):
 	def has_opportunity(self):
 		return frappe.db.get_value("Opportunity", {"lead": self.name, "status": ["!=", "Lost"]})
 
+	def has_quotation(self):
+		return frappe.db.get_value(
+			"Quotation", {
+				"lead": self.name,
+				"status": ["!=", "Lost"],
+				"status": ["!=", "Cancelled"]
```
nabinhait: you cannot set same key twice in a dictionary, first one will be ignored. You can write like this `"status": ["not in", ["Lost", "Cancelled"]]`

nabinhait: Add another condition for submitted docs only `"docstatus": 1`

erpnext/crm/doctype/lead/lead.py
```diff
@@ -80,6 +80,22 @@ def has_customer(self):
...
+				})
+
+	def lost_quotation(self):
+		return frappe.db.get_value(
+			"Quotation", {
+				"lead": self.name,
+				"status": ["=", "Lost"]
```
nabinhait: Add another condition for submitted docs only `"docstatus": 1`

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -13,9 +13,14 @@
 }
 
 class Quotation(SellingController):
+	def after_insert(self):
+		self.update_lead()
+
 	def validate(self):
 		super(Quotation, self).validate()
 		self.set_status()
+		self.update_opportunity()
```
nabinhait: Status of Lead or Opportunity should not be updated from a draft quotation, it should be only updated on submission of Quotation.

aruizramon:
Thanks for the feedback. I can make changes as suggested if you think the feature still has value!

erpnext/controllers/status_updater.py
aruizramon: `status_map` gets reversed later on - so converted is evaluated first.

This could be changed/removed to help with readability - priority is set on order instead of reverse order.

PR #5461, opened 2016-06-10
[POS] Confirm to reload page if internet connection is not available

rohitwaghchaure:


erpnext/accounts/page/pos/pos.js
```diff
@@ -39,6 +42,20 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
...
+		if(this.connection_status == false && frappe.get_route()[0] == "pos"){
+			e = e || window.event;
+
+			// For IE and Firefox prior to version 4
+			if (e) {
+			    e.returnValue = 'You are in offline mode,';
+			}
+
+			// For Safari
+			return 'You are in offline mode,';
```
rmehta: Can this be more descriptive?

Also needs to be translated `__("You are in offline mode. You will not be able to reload until you have network.")`

PR #5463, opened 2016-06-10
[bug fix] fixed typecasting error in customer portal rendering

shubham7saxena:
gave exception handling to the Doctype loading in post_process function for listing the portal items.

### DIFF
--- a/erpnext/controllers/website_list_for_contact.py
+++ b/erpnext/controllers/website_list_for_contact.py
@@ -90,7 +90,15 @@ def post_process(doctype, data):
 			doc.set_indicator()
 
 		doc.status_display = ", ".join(doc.status_display)
-		doc.items_preview = ", ".join([d.item_name for d in doc.items])
+		"""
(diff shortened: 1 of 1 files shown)


erpnext/controllers/website_list_for_contact.py
```diff
@@ -90,7 +90,15 @@ def post_process(doctype, data):
...
-		doc.items_preview = ", ".join([d.item_name for d in doc.items])
+		"""
+		tries to join with the item_name of received object 
+		but if any other data type other than int or str is 
+		received, it type casts it to string
+		"""
+		try:
+			doc.items_preview = ", ".join([d.item_name for d in doc.items])
+		except:
+			doc.items_preview = ", ".join([str(d.item_name) for d in doc.items])
```
anandpdoshi: When is item_name something other than string?

rmehta:
@shubham7saxena can you fix as per @anandpdoshi 's comments

PR #5474, opened 2016-06-13
[POS] Enhancement to set default payment as cash, functionality to clear amount value, renamed doctype Payments and some fixes.

rohitwaghchaure:


erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -87,6 +88,10 @@ def update_multi_mode_option(doc, pos_profile):
 
 		doc.append('payments', payment_mode)
 
+def get_mode_of_payment(doc):
+	return frappe.db.sql(""" select mpa.default_account, mpa.parent, mp.type as type from `tabMode of Payment Account` mpa,
```
rmehta: use `frappe.get_all` ?

rohitwaghchaure:
I have to get data from two tables therefore used frappe.db.sql.

PR #5481, opened 2016-06-15
Create Manufacture MRs for BOM Items

bcornwellmott:
Corrects an issue where items with BOMs have Purchase material request submitted instead of Manufacture material requests.

Based off of issue:
[https://discuss.erpnext.com/t/production-planning-tool-bug/13385](https://discuss.erpnext.com/t/production-planning-tool-bug/13385)

### DIFF
--- a/erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py
+++ b/erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py
@@ -449,9 +449,12 @@ class ProductionPlanningTool(Document):
 					"transaction_date": nowdate(),
 					"status": "Draft",
 					"company": self.company,
-					"requested_by": frappe.session.user,
-					"material_request_type": "Purchase"
+					"requested_by": frappe.session.user
 				})
+				if item_wrapper.default_bom:
+					material_request.update({"material_request_type": "Manufacture"}) 
+				else:
+					material_request.update({"material_request_type": "Purchase"})
 				for sales_order, requested_qty in items_to_be_requested[item].items():
 					material_request.append("items", {
 						"doctype": "Material Request Item",


bcornwellmott:
To clarify the issue with an example:

Lets say I have an assembly structure as follows:
image

If I'm going to manufacture this MasterAssembly, I need to do the following steps.
1. Order my raw materials (2x Part1, 4x Part2, 4x Part3)
2. Manufacture my subassembly (Assembly1), then manufacture my MasterAssembly.

The Production Planning Tool should be able to initiate all of these tasks through Purchase Material Requests and Production Orders, but it currently does not.
Right now, it can do one of two things depending on whether the Multi-Level BOM checkbox is ticked. Either: (a) create material request to purchase 2x Assembly1 and 4x Part3; or, (b) create a material request to purchase 2x Part1, 4x Part2 and 4x Part3. In both cases, it will create a Production Order for MasterAssembly. 

**The problem with the current system**
In case (a), the Material Request for Assembly1 is created for Purchase, when it should be Manufacture (we know it should be Manufacture because it has a BOM attached). 
image


In case (b), the Production Order for Assembly1 isn't created, so we will never know to schedule and perform manufacturing for that assembly.

**Proposed Change**
When the Mutli-Level BOM checkbox is not ticked, instead of creating Material Requests with type "Purchase" for ALL items (Assembly1 and Part3 in this case), look to see if a default BOM is selected for the item. If so, create a Material Request with type "Manufacture" instead. When you run the Production Planning Tool, the result will be as follows:

_First Pass_ (Selected Item from MasterAssembly)
1. Create Production Order for MasterAssembly
2. Create Material Request for 2x Assembly1 (Manufacture) and 4x Part3 (Purchase)

_Second Pass_ (Get Items for open Material Request for Manufacture)
3. Create Production Order for Assembly1
4. Create Material Request for 2x Part1 (Purchase) and 4x Part2 (Purchase)

In this case, it does take two passes of running the Production Planning Tool, but you get all the purchase material requests and production orders issued correctly.

bcornwellmott:
Any feedback on this? If there are concerns about automatically using Manufacture Material Requests if there is a default BOM, we could add a check box that requests this:

Create Manufacture Material Requests for BOM items

Personally, I don't think it is necessary....

rmehta:
assigning to @nabinhait. We are all working on v7 issues, so will merge it before we launch v7.

PR #5495, opened 2016-06-16
Add "Get items from Material Request"

bcornwellmott:
Allows users to pull items from multiple Material Requests into a Request for Quotation


bcornwellmott:
Any feedback on this? It should be relatively straightforward...

erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
```diff
@@ -103,6 +103,22 @@ frappe.ui.form.on("Request for Quotation Supplier",{
 erpnext.buying.RequestforQuotationController = erpnext.buying.BuyingController.extend({
 	refresh: function() {
 		this._super();
+		if (this.frm.doc.docstatus===0) {
```
nabinhait: docstatus should be 1

Issue #5520, opened 2016-06-21
Delivery Note List - Percent billed shows dollar sign instead of percent sign


cgpurbaugh:
see title

![screenshot dollarsign](https://cloud.githubusercontent.com/assets/13136405/16243288/84bcc0c4-37c4-11e6-8510-ad97ada3721b.jpg)

bcornwellmott:
Addressed in https://github.com/frappe/erpnext/pull/5527

PawanMeh:
@cgpurbaugh kindly close the issue, if it is fixed for you

PR #5527, opened 2016-06-22
Fix Percent Billed to show Percentage

bcornwellmott:
Previously was showing Currency


bcornwellmott:
Fixes issue: https://github.com/frappe/erpnext/issues/5520

PR #5529, opened 2016-06-23
[fix] previous document links

saurabh6790:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -663,6 +672,9 @@ def validate_fixed_asset_account(self):
 
 	def on_recurring(self, reference_doc):
 		self.due_date = None
+	
+	def get_link_filters(self, for_doctype):
```
nabinhait: Don't need to define this function in all individual files. Transaction Base class already inherited, so it is already available.

saurabh6790:
@nabinhait fixed.

Issue #5251, opened 2016-04-21
[Item] appears multiple times in Price List issue while creating any SO/SI etc


sbkolate:
Hello,

We are always getting this error.
`Item Upper Bearing Plate appears multiple times in Price List Standard Selling`
![sales_inv](https://cloud.githubusercontent.com/assets/6648915/14699551/575e7d22-07b6-11e6-9474-e5d850bc7230.png)

I have reproduced this error on your demo server.
Steps to reproduce.
1) Go to Item Price List Standard Selling
2) Select Item 'Upper Bearing Plate' and add rate 0
3) Now, create Sales Invoice with Item 'Upper Bearing Plate' and rate '100'
4) ERPNext finds price is zero and tries to add it into Price List Standard Selling.
5) Click on Save, it will throws above error. `Item Upper Bearing Plate appears multiple times in Price List Standard Selling`

Thanks,
Sambhaji

rmehta:
@sbkolate you should be fixing such bugs yourself!

PR #5534, opened 2016-06-24
Item Price update in Price List

KanchanChauhan:
https://github.com/frappe/erpnext/issues/5251 fixed

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -239,7 +239,6 @@ def insert_item_price(args):
 		if frappe.has_permission("Item Price", "write"):
 			price_list_rate = args.rate / args.conversion_factor \
 				if args.get("conversion_factor") else args.rate
-			if not price_list_rate: return
 		
 			item_price = frappe.get_doc({
 				"doctype": "Item Price",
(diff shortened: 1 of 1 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -237,20 +237,30 @@ def insert_item_price(args):
 	if frappe.db.get_value("Price List", args.price_list, "currency") == args.currency \
 		and cint(frappe.db.get_single_value("Stock Settings", "auto_insert_price_list_rate_if_missing")):
 		if frappe.has_permission("Item Price", "write"):
-
 			price_list_rate = args.rate / args.conversion_factor \
 				if args.get("conversion_factor") else args.rate
-
+			if not price_list_rate: return
+		
 			item_price = frappe.get_doc({
```
nabinhait: This should be in else block

erpnext/stock/get_item_details.py
```diff
@@ -237,20 +237,30 @@ def insert_item_price(args):
...
 				"price_list": args.price_list,
 				"item_code": args.item_code,
 				"currency": args.currency,
 				"price_list_rate": price_list_rate
 			})
-			item_price.insert()
-			frappe.msgprint(_("Item Price added for {0} in Price List {1}").format(args.item_code,
-				args.price_list))
+			
+			name = frappe.db.get_value('Item Price', {'item_code': args.item_code, 'price_list': args.price_list}, 'name')
```
nabinhait: Also consider currency while checking for existing item price record

erpnext/stock/get_item_details.py
KanchanChauhan: I realized, this line was not needed at all. Removed it.

erpnext/stock/get_item_details.py
KanchanChauhan: Added currency to filter.

PR #5553, opened 2016-06-27
[minor][fix] throw if payment modes are not set

saurabh6790:


erpnext/accounts/page/pos/pos.js
```diff
@@ -876,14 +876,22 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
...
 		if(this.frm.doc.items.length == 0){
 			frappe.throw(__("Select items to save the invoice"))
 		}
 	},
-
+	
+	validate_mode_of_payments: function(){
+		if (this.frm.doc.payments.length === 0){
+			frappe.throw(__("Payment Mode is not configured.\
+				Please check, whether account has been set on Mode of Payments or on POS Profile."))
```
rmehta: Don't break translatable messages into multiple lines

saurabh6790:
@rmehta  fixed

PR #5572, opened 2016-06-28
[fix] warehouse tree patch fix

saurabh6790:
- check for parent account with type stock in patch if not exists get parent account from existing warehouse accounts
- check auto accounting for stock in warehouse tree patch
- https://discuss.erpnext.com/t/error-on-update-account-group/13693


erpnext/patches/v7_0/create_warehouse_nestedset.py
```diff
@@ -45,4 +49,18 @@ def create_default_warehouse_group(company):
...
+def validate_parent_account_for_warehouse(company):
+	if cint(frappe.defaults.get_global_default("auto_accounting_for_stock")):
+		parent_account = frappe.db.sql("""select name from tabAccount
+			where account_type='Stock' and company=%s and is_group=1
+			and (warehouse is null or warehouse = '')""", company.name)
+		if not parent_account:
+			current_parent_accounts_for_warehouse = frappe.db.sql("""select parent_account from tabAccount
+				where account_type='Warehouse' and (warehouse is not null or warehouse != '') """)
+			if current_parent_accounts_for_warehouse:
+				doc = frappe.get_doc("Account", current_parent_accounts_for_warehouse[0][0])
```
nabinhait: Simply use set_value to set account_type

saurabh6790:
@nabinhait Fixed

PR #5604, opened 2016-06-29
[Fixes] Time sheet fixes and changed timesheets to time_logs

rohitwaghchaure:


erpnext/projects/doctype/time_sheet/time_sheet.py
```diff
@@ -261,7 +261,7 @@ def get_activity_cost(employee=None, activity_type=None):
 	if not rate:
 		rate = frappe.db.get_values("Activity Type", {"activity_type": activity_type},
 			["costing_rate", "billing_rate"], as_dict=True)
-
+	frappe.errprint(rate[0])
 	return rate[0] if rate else {}
```
rmehta: errprint

PR #5605, opened 2016-06-29
while creating warehouse tree check for non company warehouse and is_group default value fix

saurabh6790:


saurabh6790:
- https://discuss.erpnext.com/t/warehouse-stock-entry-problem-with-version-7/13750
- https://discuss.erpnext.com/t/all-warehouses-are-now-group/13749/5

PR #5615, opened 2016-06-30
[RFQ Enhancement] Added button on supplier row to download PDF for particular supplier

rohitwaghchaure:


rohitwaghchaure:
https://www.youtube.com/watch?v=Z5nkQNmaW7Q&feature=youtu.be

erpnext/buying/doctype/request_for_quotation/request_for_quotation.py
```diff
@@ -226,3 +227,16 @@ def create_rfq_items(sq_doc, supplier, data):
 		"request_for_quotation_item": data.name,
 		"request_for_quotation": data.parent
 	})
+
+@frappe.whitelist()
+def download_pdf(doctype, name, supplier_idx):
+	doc = get_rfq_doc(doctype, name, supplier_idx)
+	if doc:
+		frappe.get_attr("frappe.www.print.download_pdf")(doctype, name, doc=doc)
```
rmehta: why get_attr? This can be directly imported

rohitwaghchaure: from frappe.www.print import download_pdf
I have tried this, but it giving me an error invalid syntax at print
Seems like it's not allowing to import with filename print

rmehta: Yeah! Do this - move all the logic to another file, maybe `frappe/utils/print_format.py` and in print.py just keep the `get_context` function.

rohitwaghchaure: OK

PR #5628, opened 2016-07-04
[minor][fix] reload warehouse before executing patches

saurabh6790:
@rmehta 
https://discuss.erpnext.com/t/version-upgrade-from-6-to-7/13305/15


rmehta:
can you add it in your existing patch?

PR #5641, opened 2016-07-04
Rename prevdoc field in purchase order and purchase receipt doctype

rohitwaghchaure:


erpnext/patches/v7_0/rename_prevdoc_fields.py
```diff
@@ -0,0 +1,22 @@
...
+	frappe.reload_doctype('Purchase Order Item')
+	frappe.reload_doctype('Purchase Receipt Item')
+	update_po_fields()
+	rename_pr_fields()
+
+def update_po_fields():
+	for data in frappe.db.sql(""" select prevdoc_docname, prevdoc_detail_docname, name, prevdoc_doctype
+		from `tabPurchase Order Item` where prevdoc_doctype is not null""", as_dict=True):
+		if data.prevdoc_doctype == 'Material Request':
+			frappe.db.set_value("Purchase Order Item", data.name, "material_request", data.prevdoc_docname)
```
nabinhait: Add `update_modified=False`

We should not change modified date of any documents.

erpnext/patches/v7_0/rename_prevdoc_fields.py
```diff
@@ -0,0 +1,22 @@
...
+			frappe.db.set_value("Purchase Order Item", data.name, "material_request_item", data.prevdoc_detail_docname)
+		elif data.prevdoc_doctype == 'Sales Order':
+			frappe.db.set_value("Purchase Order Item", data.name, "sales_order", data.prevdoc_docname)
+			frappe.db.set_value("Purchase Order Item", data.name, "sales_order_item", data.prevdoc_detail_docname)
```
nabinhait: As for PO, we are not using `rename_field` function, we should at-least update Property Setters, Print Formats and Report records for those fields. Check if we can call some functions from rename_field().

PR #5653, opened 2016-07-05
[fix] task visibility fix #5601

### LINKED ISSUE #5601: [V7] Project - Tasks visibility
Steps:

1. Open existing Project.
2. Edit Status for Task.
3. Save Project.

After saving Project, none of the Task is visible in the Project. After a refresh, Tasks are visible again.

project-task-visibility

saurabh6790:


rmehta:
what is the use case?

saurabh6790:
Due to `if not self.get('__unsaved')` condition its not executing `self.load_tasks()` when you save form after modifications to it.

rmehta:
mm that was to fix another problem.

With your update, try remove all the tasks from the list.

saurabh6790:
Ya, big issue. It's not working if you delete all tasks. Let me check...

saurabh6790:
fixed

Issue #4341, opened 2015-11-20
Journal Entry - Activating Multi Currency


PrakashH:
While creating Journal Entry for multi currency account, I was unable to select USD account, whereas base currency is INR. Actually it needed checking Multi Currency. But it was not visible before selecting account, hence not intuitive.

Multi Currency field below Journal Entry Accounts table was visible, specially when form from child table is open. I suggest that Multi Currency field should be placed in the very first section of Journal Entry form. Perhaps below Posting Date field itself. It's important to update value in this field before selecting Accounts in the child table.

![screen shot 2015-11-20 at 11 55 42 am](https://cloud.githubusercontent.com/assets/4570016/11293231/22f3fbbc-8f7e-11e5-89eb-f69d467f3b5f.png)

Also we can set visibility for this field by checking Accounts for that Company. Only if Company has Account in foreign currency, this field should be visible, else hidden.

nabinhait:
Most of the users don't use multi-currency feature, thats why we placed it in the bottom.

amandadtp:
If fact, we are one of the users to use multi-currency feature everyday.  

Limitation: 
**Not showing the currency symbol in full** such as "$" instead of "US$".  It's rather confusing if we are using CAD, AUD, HKD etc. they all use "$" as the currency symbol.

Showing full currency US$ or USD should always be the practice in accounting system.   
**PLEASE HELP TO ADD THIS FEATURE IN!**

This is one of the limitation that my accountant reported to me.  I will update on this conversation once I hear more feedback from the team.

cgpurbaugh:
@amandadtp You can change the symbol to whatever you desire.  Just go to the currency list and change the symbol
![money symbols](https://cloud.githubusercontent.com/assets/13136405/16597816/e3ce21d4-42c8-11e6-862e-1261a6f4261d.PNG)

![money symbols2](https://cloud.githubusercontent.com/assets/13136405/16597842/f7bed972-42c8-11e6-8180-c2dd6af0899d.PNG)

cgpurbaugh:
@amandadtp Also, you should have made a new issue.  This issue is closed.

PR #5656, opened 2016-07-05
Moved multi-currency

cgpurbaugh:
For #4341


cgpurbaugh:
(Didn't realize this was a closed issue)

amandadtp:
Thanks

amandadtp:
So which role permission I need to have in order to edit the Currency List?

cgpurbaugh:
System Manager, by default.  It can be changed though.

amandadtp:
I have System Manager role permission, but I am not able to change the currency list.  Any idea why would that be the case?

cgpurbaugh:
Not sure, check the role permission manager.

amandadtp:
It says "For System Manager at level 0 in Currency in row 1: Cannot set Assign Amend if not Submittable"

Not too sure what that means though.

cgpurbaugh:
Amend is only for submittable documents, like Sales Order.  Currency, all you should need is read & write.

amandadtp:
I have checked the Read, Write, Create boxes but I still cannot edit the currency list

cgpurbaugh:
have you tried reloading?

amandadtp:
Still the same.  The currency page, for example, HKD is gray out and I am not able to click on anything except the ERPNext home button and my Login menu.

amandadtp:
FYI. I have just rebooted my computer

cgpurbaugh:
Seems like a bug.  Can you check your console?  (F12)

Does it say "c.sender is undefined"?

I'm getting this error on my V7 build.

amandadtp:
I am not too sure on this (I am not from tech background).  Need to check with IT...

amandadtp:
If this is a bug, where/who I can report this to?

cgpurbaugh:
Make a new issue here.

cgpurbaugh:
@amandadtp I created an issue: #5687

PR #5722, opened 2016-07-12
PPT looks at default MR type in Item to set MR type.

bcornwellmott:
Currently, the Production Planning tool decides whether a Material Request is generated as a Purchase or Manufacture material request based on if there is a Default BOM associated with the Item. This change adds a field in Item called Default Material Request Type, which is used to control what type of Materil Request is generated through the Production Planning tool.

Also, the Use Multi-BOM checkbox is shown by default when the form loads (there is no reason to hide this if you manually want to add production items).


rmehta:
You can also set this when MR is created at time of automatic re-order. Can you fix it?

bcornwellmott:
Updated - now the Item Reorder has a read-only property for material request type which is pulled from the Default Pull Request Type. It now accommodates Purchase, Transfer, Issue and Manufacture requests,

I'm not sure if it should be read-only because you might want to be able to adjust that, but it is an easy fix if we do. Right now I validate to make sure the material request matches the default one set in the item (in case they don't set it).

rmehta:
@rohitwaghchaure can you test and merge?

bcornwellmott:
@rohitwaghchaure, the check failed because the default material request type and the re-order request type didn't match. I'm not sure we need them to match, so if you'd like I can remove the validation (maybe use a warning instead?)

bcornwellmott:
@rohitwaghchaure , I think you addressed all these changes in another pull request, correct? Can I close?

nabinhait:
fixed via #5925

PR #5751, opened 2016-07-14
Add Notes to timesheet_detail.json

superlack:


rohitwaghchaure:
Hi @superlack ,
We have already added note in timesheet doctype, please take latest update and check

superlack:
Thanks for the response  I see that notes are available in Timesheet as a whole.   I currently have a business, and used to be employed by one, that for accountability and transparency reasons, would deliver reports on activity attached to each billable time; pre-emptive justification to prevent dispute. 

With my apologies, I hastily submitted that pull request in hopes of giving myself a spring board to work with the current system, but in a project-based ecosystem (to which the timesheet system doesn't discriminate; ie, multiple projects billed within a TS when making an invoice.)

Will continue to brainstorm on this..  Again, thanks for the reply!

Issue #5774, opened 2016-07-15
Error : Executing erpnext.patches.v7_0.convert_timelogbatch_to_timesheet


revant:
```
Executing erpnext.patches.v7_0.convert_timelogbatch_to_timesheet in cc.in (8a26013f3b)
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/revant/frappe-bench-v7/apps/frappe/frappe/utils/bench_helper.py", line 79, in <module>
    main()
  File "/home/revant/frappe-bench-v7/apps/frappe/frappe/utils/bench_helper.py", line 16, in main
    click.Group(commands=commands)(prog_name='bench')
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/home/revant/frappe-bench-v7/env/local/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/revant/frappe-bench-v7/apps/frappe/frappe/commands/__init__.py", line 24, in _func
    ret = f(frappe._dict(ctx.obj), *args, **kwargs)
  File "/home/revant/frappe-bench-v7/apps/frappe/frappe/commands/site.py", line 182, in migrate
[...]

PR #5775, opened 2016-07-15
Issue #5774

revant:


erpnext/patches.txt
```diff
@@ -280,6 +280,8 @@ erpnext.patches.v6_20x.repost_valuation_rate_for_negative_inventory
 erpnext.patches.v7_0.re_route #2016-06-27
 erpnext.patches.v7_0.system_settings_setup_complete
 erpnext.patches.v7_0.set_naming_series_for_timesheet
+execute:frappe.reload_doc('projects', 'doctype', 'project', force=True)
```
nabinhait: Remove force=True, it resets the permissions table.

PR #5776, opened 2016-07-16
[fix] for setting warehouse in account head

saurabh6790:
screen shot 2016-07-16 at 11 16 27 am

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -165,7 +165,7 @@ class Account(Document):
 		if not cint(frappe.defaults.get_global_default("auto_accounting_for_stock")):
 			return
 			
-		if self.account_type == "Stock" and not self.is_group:
+		if self.account_type == "Stock" and not cint(self.is_group):
 			if not self.warehouse:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/account/account.py
```diff
@@ -165,7 +165,7 @@ def validate_warehouse_account(self):
 		if not cint(frappe.defaults.get_global_default("auto_accounting_for_stock")):
 			return
 			
-		if self.account_type == "Stock" and not self.is_group:
+		if self.account_type == "Stock" and not cint(self.is_group):
```
nabinhait: It's ok. But why the warehouse was not set in account on creation of warehouse?

nabinhait: Also check if `cint` required in any other places?

saurabh6790:
@nabin actually its settings warehouse while creating account head for a warehouse. But as  self.is_group value is Unicode, the output for condition `if self.account_type == "Stock" and not cint(self.is_group):` is `false` perhaps it's executing next  code block ```elif self.warehouse: self.warehouse = None``` and settings warehouse to '' in account.

PR #5806, opened 2016-07-20
[Fix] Stock entry difference amount issue.

rohitwaghchaure:


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -234,18 +234,20 @@ def get_stock_and_rate(self):
...
-		self.set_basic_rate(force)
+	def calculate_rate_and_amount(self, force=False, enable_validate=True):
+		self.set_basic_rate(force, enable_validate)
 		self.distribute_additional_costs()
 		self.update_valuation_rate()
 		self.set_total_incoming_outgoing_value()
 		self.set_total_amount()
 
-	def set_basic_rate(self, force=False):
+	def set_basic_rate(self, force=False, enable_validate=True):
```
nabinhait: Rename `enable_validate` to `update_finished_item_rate`

rohitwaghchaure: @nabinhait Done

Issue #5807, opened 2016-07-20
[V7] Mutiple Item Prices / Create Item Prices directly from Supplier Quotation


bcornwellmott:
Based on discussion in forums:
[https://discuss.erpnext.com/t/feature-create-item-prices-directly-from-supplier-quotations/14176](https://discuss.erpnext.com/t/feature-create-item-prices-directly-from-supplier-quotations/14176)
- [ ] Add the ability to use multiple item prices per item per price list as long as min quantities are different
- [ ]  Make sure item price lookups get the correct item price based on closest qty that is less than requested amount
- [x] Automatically populate the item prices when a Supplier Quotation is created (warn if previous item prices exist)

I'll definitely be able to do some of the work myself here.

PR #5808, opened 2016-07-21
Supplier Quotation adds Item Prices

bcornwellmott:
Addressing point #3 of [https://github.com/frappe/erpnext/issues/5807](Issue #5807).

Adds a check box in the Buying Setting to allow the user to select if they want Item Prices to be created from Supplier Quotations. When submitted, it checks to see if there are duplicate item codes, if so it skips adding item prices altogether (giving a warning). If not, it will add or update the Item Prices from the Quotation Price List.


rmehta:
this feature already exists in `get_item_details.py` why replicate it?

bcornwellmott:
Sorry - didn't know about the feature in get_item_details.py

Are we ok if I modify this code to call get_item_details.insert_item_price for each item on Supplier Quote submission (if the checkbox is checked off)?

rmehta:
> Are we ok if I modify this code to call get_item_details.insert_item_price for each item on Supplier Quote submission (if the checkbox is checked off)?

Yes. let us know when this is done!

PR #5824, opened 2016-07-21
Demo records for projects+tasks+billing, pos invoice

rohitwaghchaure:


erpnext/demo/setup_data.py
```diff
@@ -316,6 +318,18 @@ def setup_salary_structure_for_timesheet():
 		ss_doc.hour_rate = flt(random.random() * 10, 2)
 		ss_doc.save(ignore_permissions=True)
 
+def setup_mode_of_payment():
+	account_dict = {'Cash': 'Cash - WPL', 'Bank': 'National Bank - WPL'}
+	for account in frappe.get_all('Mode of Payment', fields = ["*"]):
```
nabinhait: Always avoid selecting *

PR #5828, opened 2016-07-21
Added Guardian

neilLasrado:


erpnext/schools/doctype/interest/interest.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
```
rmehta: should not be in schools module

neilLasrado: I will move it to HR

PR #5829, opened 2016-07-21
Added new report for Supplier Quotation comparisons

bcornwellmott:
This is a very simple report that allows a user to compare the prices from multiple suppliers at the same time. I'm not sure if anything more than these files is needed - does anything else need to happen to include the Report?


strixaluco:
That's great. Thanks a lot for contributing.

PR #5839, opened 2016-07-22
Payment request demo

saurabh6790:
#5736


erpnext/accounts/doctype/payment_request/payment_request.js
```diff
@@ -31,5 +31,21 @@ frappe.ui.form.on("Payment Request", "refresh", function(frm) {
 			});
 		});
 	}
+	
+	if(!frm.doc.payment_gateway_account && frm.doc.status == "Initiated") {
+		frm.add_custom_button(__('Mark Payment Entry'), function(){
```
nabinhait: Make Payment Entry

PR #5844, opened 2016-07-22
Added C3.JS Plot to compare quotes in Print view

bcornwellmott:
This will display a Google Chart of a price comparison between different supplier quotations for a given item.

It will look something like this:
image


rmehta:
@bcornwellmott lets use the new charting engine c3.js

Generally we should avoid using proprietary APIs in this project.

bcornwellmott:
Understood. I have modified the html file to use the c3.js charting tool instead.

bcornwellmott:
Any way to have this move forward?

bcornwellmott:
@nabinhait / @rmehta Still waiting for this to be merged or some feedback on how to improve it. Please let me know what I can do.

PR #5865, opened 2016-07-25
[mionr] fixes and cleanups

rohitwaghchaure:


rmehta:
@rohitwaghchaure  merging. But pls add a test case too.

PR #5872, opened 2016-07-26
[Fix] Set warehouse to stock account and fixed gl entries against transactions.

rohitwaghchaure:


rmehta:
@nabinhait can you check?

PR #5877, opened 2016-07-26
[minor] Fixes

rohitwaghchaure:


erpnext/patches/v7_0/calculate_total_costing_amount.py
```diff
@@ -11,5 +11,6 @@ def execute():
 			ts.update_cost()
 			ts.calculate_total_amounts()
 			ts.flags.ignore_validate = True
+			ts.flags.ignore_mandatory = True
 			ts.flags.ignore_validate_update_after_submit = True
 			ts.save()
```
rmehta: why not just `ts.db_set()` ?

PR #5901, opened 2016-07-27
Fix Timesheet employee query to show names

bcornwellmott:
The current Employee lookup only shows the employee number. This fix will also show the employee name to make it easy for employees to find their name.

### DIFF
--- a/erpnext/projects/doctype/timesheet/timesheet.py
+++ b/erpnext/projects/doctype/timesheet/timesheet.py
@@ -295,5 +295,5 @@ def get_activity_cost(employee=None, activity_type=None):
 def get_employee_list(doctype, txt, searchfield, start, page_len, filters):
 	return frappe.db.sql("""select distinct employee, employee_name
 		from `tabSalary Structure` where salary_slip_based_on_timesheet=1
-		and employee like %(txt)s limit %(start)s, %(page_len)s""", 
-		{'txt': "%%%s%%"% txt, '_txt': txt.replace("%", ""), 'start': start, 'page_len': page_len})
+		and employee like %(txt)s or employee_name like %(txt)s) limit %(start)s, %(page_len)s""", 
+		{'txt': "%%%s%%"% txt, 'start': start, 'page_len': page_len})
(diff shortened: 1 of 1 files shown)


erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -293,7 +293,7 @@ def get_activity_cost(employee=None, activity_type=None):
 
 @frappe.whitelist()
 def get_employee_list(doctype, txt, searchfield, start, page_len, filters):
-	return frappe.db.sql("""select distinct(employee) as employee
+	return frappe.db.sql("""select distinct employee, employee_name
 		from `tabSalary Structure` where salary_slip_based_on_timesheet=1
 		and employee like %(txt)s limit %(start)s, %(page_len)s""", 
-		{'txt': "%%%s%%"%(txt), 'start': start, 'page_len': page_len})
+		{'txt': "%%%s%%"% txt, '_txt': txt.replace("%", ""), 'start': start, 'page_len': page_len})
```
rohitwaghchaure: @bcornwellmott 
What is use of _txt?

erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -293,7 +293,7 @@ def get_activity_cost(employee=None, activity_type=None):
 
 @frappe.whitelist()
 def get_employee_list(doctype, txt, searchfield, start, page_len, filters):
-	return frappe.db.sql("""select distinct(employee) as employee
+	return frappe.db.sql("""select distinct employee, employee_name
 		from `tabSalary Structure` where salary_slip_based_on_timesheet=1
 		and employee like %(txt)s limit %(start)s, %(page_len)s""",
```
rohitwaghchaure: Add (employee like %(txt)s or employee_name like %(txt)s)

erpnext/projects/doctype/timesheet/timesheet.py
bcornwellmott: Honestly, not sure. I duplicated what was on Salary Structure.

erpnext/projects/doctype/timesheet/timesheet.py
bcornwellmott: Will change

bcornwellmott:
@rohitwaghchaure, the fixes you requested should be complete.

PR #5919, opened 2016-07-28
Timesheets update on date changes in Salary Slip

bcornwellmott:
Currently, if the dates in the Salary Slip are updated, this does not change the timesheets list. This commit resolves this issue.


nabinhait:
But I did not found how `set_time_sheet` function is called on change of dates. You should add the event on salary_slip.js

bcornwellmott:
It gets called as part of cscript.fiscal_year in salary_slip.js

PR #5924, opened 2016-07-29
Formula based Salary Structure

KanchanChauhan:
- Condition and Formula based Salary Components
- Multiple employees against same Salary Structure


fikrut:
if it helps you to design the payroll structure based on formula here is our country rule to make salary income tax deductions from basic salary


Salary Range (ETB) Tax Rate Deduction (ETB) 
 1. 0 - 600 Birr Non-Taxable - 
 2. 601-1,650 Birr 10% 60 Birr 
 3. 1,651 - 3,200 Birr 15%      -142.50 Birr 
 4. 3,201 - 5,250 Birr 20%       -302.50 Birr 
 5. 5,251 - 7,800 Birr 25%        -565 Birr 
 6. 7,801 - 10,900 Birr 30%       -955 Birr 
 7. Over 10,900 Birr 35%           -1,500 Birr 
 so in our country income tax calculation is (Salary Range*Tax rate)-deduction like this assume your salary is 5000 then the calculation will be the range will fall in no 4 so (5000*20%)-302.5=697.5 so the income tax is calculated so when you develop this module take this in to consideration. Thanks

erpnext/hr/doctype/salary_component/salary_component.py
```diff
@@ -7,4 +7,21 @@
...
+	def validate_abbr(self):
+		if not self.salary_component_abbr:
+			self.salary_component_abbr = ''.join([c[0] for c in self.salary_component.split()]).upper()
+
+		self.salary_component_abbr = self.salary_component_abbr.strip()
+
+		if self.get('__islocal') and len(self.salary_component_abbr) > 5:
+			frappe.throw(_("Abbreviation cannot have more than 5 characters"))
+
+		if not self.salary_component_abbr.strip():
```
rmehta: make it mandatory in the doctype. This is not reqd!

nabinhait:
@KanchanChauhan Is it ready to merge?

KanchanChauhan:
@nabinhait yes it is, just do the code review once before merging.

PR #5925, opened 2016-07-29
PPT looks at default MR type in Item to set MR type.

rohitwaghchaure:


erpnext/patches.txt
```diff
@@ -303,4 +303,5 @@ erpnext.patches.v7_0.migrate_schools_to_erpnext
 erpnext.patches.v7_0.remove_administrator_role_in_doctypes
 erpnext.patches.v7_0.rename_fee_amount_to_fee_component
 erpnext.patches.v7_0.calculate_total_costing_amount
-erpnext.patches.v7_0.fix_nonwarehouse_ledger_gl_entries_for_transactions
\ No newline at end of file
+erpnext.patches.v7_0.fix_nonwarehouse_ledger_gl_entries_for_transactions
+execute:frappe.db.sql("update `tabItem` set default_material_request_type='Purchase'")
```
nabinhait: set default_material_request_type = 'Purchase' if there is no BOM, otherwise set it as 'Manufacture'

It will not break existing workflow

nabinhait: Also set material_request_type value in Item Reorder table

PR #6024, opened 2016-08-08
[Fix] Get account of the mode of payment on sales invoice and pos profile

rohitwaghchaure:


nabinhait:
Commonify the function, you can place it in controllers/accounts.js

rohitwaghchaure:
Done

PR #6054, opened 2016-08-10
[Fix] Debit and credit not equal for purchase invoice, if update stock is yes and taxes is defined

rohitwaghchaure:


rmehta:
@rohitwaghchaure that seems like a critical issue. I am merging now. Maybe @nabinhait can check once he is back.

PR #6070, opened 2016-08-11
Integration broker

saurabh6790:
- Payment Request Fix for new integration broker
- hooks to handle payment gateway account creation


erpnext/accounts/utils.py
```diff
@@ -570,3 +570,62 @@ def get_children():
 				each["balance_in_account_currency"] = flt(get_balance_on(each.get("value")))
 
 	return acc
+
+def create_payment_gateway_and_account(gateway):
+	"""If ERPNext is installed, create Payment Gateway and Payment Gateway Account"""
+	if "erpnext" not in frappe.get_installed_apps():
+		return
```
rmehta: You are already in erpnext! :wow:

erpnext/config/setup.py
```diff
@@ -1,121 +0,0 @@
-from __future__ import unicode_literals
```
rmehta: deleted by mistake?

saurabh6790:
https://github.com/frappe/erpnext/pull/6097

PR #6124, opened 2016-08-19
[POS] Fixed issue of paid amount, write off account

rohitwaghchaure:


erpnext/accounts/page/pos/pos.js
```diff
@@ -616,13 +616,23 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 			? this.item_serial_no[this.child.item_code][0] : '');
 	},
 
+	update_paid_amount_status: function(update_paid_amount){
+		update_paid_amount_dict = {true: false, false: true};
```
nabinhait: LOL! what is this?

erpnext/accounts/page/pos/pos.js
```diff
@@ -616,13 +616,23 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
...
+		if(this.name){
+			update_paid_amount = update_paid_amount_dict[update_paid_amount]
```
nabinhait: It can be written like this:

update_paid_amount = update_paid_amount ? false : true;

rohitwaghchaure:
@nabinhait Fixed

PR #6152, opened 2016-08-22
[Report] Profitability Analysis

rohitwaghchaure:
@nabinhait 
Based on Cost Center
screen shot 2016-08-24 at 7 13 43 pm

Based on Project
screen shot 2016-08-24 at 7 13 54 pm


erpnext/accounts/report/financial_statements.py
```diff
@@ -320,6 +315,22 @@ def set_gl_entries_by_account(company, from_date, to_date, root_lft, root_rgt, g
...
+	additional_conditions = []
+
+	if ignore_closing_entries:
+		additional_conditions.append("ifnull(voucher_type, '')!='Period Closing Voucher'")
+
+	if from_date:
+		additional_conditions.append("posting_date >= %(from_date)s")
+
+	if filters:
+		for key in filters:
```
nabinhait: Why looping through all the filters? Instead you can loop only for cost center and project.

```
for key in ['cost_center', 'project']:
    if filters.get(key):
         // logic
```

erpnext/accounts/report/profitability_analysis/profitability_analysis.py
```diff
@@ -0,0 +1,188 @@
...
+		},
+		{
+			"fieldname": "expense",
+			"label": _("Expense"),
+			"fieldtype": "Currency",
+			"options": "currency",
+			"width": 120
+		},
+		{
+			"fieldname": "total",
```
nabinhait: Use fieldname as per label "gross_profit_loss", it will help a lot to debug in the future.

rohitwaghchaure: @nabinhait Done

erpnext/accounts/report/financial_statements.py
rohitwaghchaure: @nabinhait Done

PR #6176, opened 2016-08-25
Update Lead.json

### LINKED ISSUE #6167: Cannot Create Quotation from Lead
We tried directly creating quotation from Lead by clicking from MAKE button.
but below error is coming:

error message - making quotation from lead

we have customize our source selection and shop is included in that list.
But this error is coming...
Can anyone check what is the problem?

On the other hand, we can make quotation directly from document type quotation and selecting the same lead.

thank you....

cgpurbaugh:
Fixes #6167 

Changed the options of 'Source' in lead to match the options for 'Source' in Quotation.

### DIFF
--- a/erpnext/crm/doctype/lead/lead.json
+++ b/erpnext/crm/doctype/lead/lead.json
@@ -208,7 +208,7 @@
    "no_copy": 0,
    "oldfieldname": "source",
    "oldfieldtype": "Select",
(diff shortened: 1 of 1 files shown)


rmehta:
@cgpurbaugh I will make a new master for this, so this problem gets solved!

PR #6181, opened 2016-08-26
Create ISSUE_TEMPLATE

strixaluco:


rmehta:
@strixaluco we tried this before, but since most people who post issues are "non-technical", they don't follow the rules.

strixaluco:
That's rather expected, but at least they could learn how it should be done, while more technical people will still provide more detailed info.

PR #6202, opened 2016-08-29
'Type' field added to Salary Component

KanchanChauhan:


erpnext/patches/v7_1/update_component_type.py
```diff
@@ -0,0 +1,13 @@
+import frappe
+from frappe.utils import flt
+
+def execute():
```
nabinhait: This patch can be written in one line:

frappe.db.sql("""update `tabSalary Component` set type=if(parentfield='earnings', 'Earning', 'Deduction')""")

erpnext/hr/doctype/salary_structure/salary_structure.js
```diff
@@ -13,6 +13,23 @@ cur_frm.cscript.onload = function(doc, dt, dn){
 }
 
 frappe.ui.form.on('Salary Structure', {
+	onload: function(frm) {
+		frm.set_query("salary_component", "earnings", function() {
+			return {
+				filters: {
+					type: "earning"
```
nabinhait: type is case sensitive?

erpnext/patches/v7_1/update_component_type.py
KanchanChauhan: 'parentfield' is in Salary Detail actually.

erpnext/hr/doctype/salary_structure/salary_structure.js
KanchanChauhan: No

erpnext/patches/v7_1/update_component_type.py
nabinhait: Sorry I missed that.

PR #6287, opened 2016-09-05
Proposal: Updates to Currency Exchange DocType and Exchange Rate Determination

ghost:
Changes to allow exchange rates on a specific day to be stored in Currency Exchange and to be pulled by the necessary transactions based on the posting date.

image
_Field "Date" added to Currency Exchange doctype_

image
_Exchange rate of a currency pair on a particular date can now be maintained manually by Accounting or Treasury user._

image
_Exchange rate between USD and NGN on 08/08/2006_

image
_Duplicate date and currency pair are detected and prevented_

https://youtu.be/ezLdX5wjPN0
_Usage of proposed functionality in Purchase Invoice_


manqala:
Still a few regressions to fix.

nabinhait:
@manqala Is it ready to merge?

manqala:
Hi. No, not yet. Should be done by the end of the weekend.

nabinhait:
Ok. No issue, take your time 👍

manqala:
@nabinhait I think we're done now. Some checks failed in Travis but they seem unrelated to the work done on Currency Exchange.

manqala:
@nabinhait @rmehta Hi guys, anyone had a look at this yet?

nabinhait:
@manqala sorry, did not get time to look into it. Will definitely check within tomorrow.

manqala:
Ok

manqala:
hi @nabinhait just following up on this.

nabinhait:
The code looks ok. Just few observations:

- Use tabs instead spaces
- Get exchange rate on change of posting/transaction date
- Fetch exchange date from fixer.io for the specific date
- After fetching from fixer.io for a date, create a Currency Exchange record for it. That way, it will not call the API repetitively.

manqala:
Hi @nabinhait 

Thanks for the feedback.

- Re Tabs, will do.
- Re get exchange rate on change of posting/transaction date, will do.
- Re fetching rate from fixer.io first, I differ in opinion on this point. There are many organisations with special exchange rate policies which are not compatible with getting a rate from an online resource in realtime. For example, some organisations as a policy, set their operating exchange rates on a monthly or weekly basis. This is why I think that the fetching of the exchange rate should be from the exchange rate table only and only if it fails should there be a lookup from fixer.io

In cases where daily rates from fixer.io are required to be entered in the exchange rate table, perhaps we can do a daily job to lookup fixer and insert into the exchange rate doctype once a day. This can be activated with a new setting that specifies if rates should be maintained manually or pulled at daily (or other) intervals from fixer.io

What do you think?

nabinhait:
> Re fetching rate from fixer.io first, I differ in opinion on this point. There are many organisations with special exchange rate policies which are not compatible with getting a rate from an online resource in realtime. For example, some organisations as a policy, set their operating exchange rates on a monthly or weekly basis. This is why I think that the fetching of the exchange rate should be from the exchange rate table only and only if it fails should there be a lookup from fixer.io

I also have the same opinion, we should look into Currency Exchange record first. If the system does not found the rate there, we should look online. But while looking into fixer.io, should we look for current date or the posting date? That is my concern.

>In cases where daily rates from fixer.io are required to be entered in the exchange rate table, perhaps we can do a daily job to lookup fixer and insert into the exchange rate doctype once a day. This can be activated with a new setting that specifies if rates should be maintained manually or pulled at daily (or other) intervals from fixer.io

This is also a great idea. We can add a setting "Fetch exchange rate daily from fixer.io". Based on that we can decide, what does user wants? If activated, we can add a scheduler job or fetch the rate while "get_exchange_rate" called for the first time.

nabinhait:
@manqala Any update? Otherwise, please let us know, we can make the necessary changes and merge it.

manqala:
Hi Nabin, sadly not been able to work on this yet. Will  have some time
next week but if you can, would appreciate if you could update and merge.

nabinhait:
@rohitwaghchaure Can you please make the changes and merge?

joezsweet:
Not sure if i'm late ..btw would be useful to add a bank commission on exchange rate, it's about 2-2.5% depending on the bank; you will never get that value so transaction entry will be affected by that.

nabinhait:
@manqala Sent another pull request (https://github.com/frappe/erpnext/pull/7162) after modifications. Fixed some issues and added patch and test cases.

@joezsweet Please create a separate issue with more details about the requirement.

PR #6306, opened 2016-09-07
Added requirement for employee to be filtered

bcornwellmott:
The current script allows any user to see all the timesheet entries for every user when it first loads, even if user permissions are set. This checks to see that an employee has been selected before showing hours.

### DIFF
--- a/erpnext/projects/doctype/timesheet/timesheet.py
+++ b/erpnext/projects/doctype/timesheet/timesheet.py
@@ -308,17 +308,22 @@ def get_events(start, end, filters=None):
 	:param end: End date-time.
 	:param filters: Filters (JSON).
 	"""
+	
 	filters = json.loads(filters)
+	
 
-	conditions = get_conditions(filters)
-	return frappe.db.sql("""select `tabTimesheet Detail`.name as name, `tabTimesheet Detail`.parent as parent,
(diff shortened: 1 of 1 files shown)


erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -308,17 +308,22 @@ def get_events(start, end, filters=None):
...
-	conditions = get_conditions(filters)
-	return frappe.db.sql("""select `tabTimesheet Detail`.name as name, `tabTimesheet Detail`.parent as parent,
-		from_time, hours, activity_type, project, to_time from `tabTimesheet Detail`, 
-		`tabTimesheet` where `tabTimesheet Detail`.parent = `tabTimesheet`.name and `tabTimesheet`.docstatus < 2 and
-		(from_time between %(start)s and %(end)s) {conditions}""".format(conditions=conditions),
-		{
-			"start": start,
-			"end": end
-		}, as_dict=True, update={"allDay": 0})
+	if filters.get('employee'):
```
nabinhait: I think it should be based on "match_conditions", then it should honour user permissions. You can check implementation of match_condition in https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/queries.py

nabinhait:
@rohitwaghchaure has already fixed this.

PR #6325, opened 2016-09-09
fix allow negatives in leave type check

RobertSchouten:


nabinhait:
Changes for adding field "allow_negative" in leave_allocation.json is missing.

Also, need changes in leave_application where system validates with balance leave, it should also be based on "allow_negative".

RobertSchouten:
the allow negative isnt coming from leave allocation is checking leave type
also doesnt leave application allready validates negative balance doesnt it?
https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/leave_application/leave_application.py#L138

nabinhait (approved):

Issue #5945, opened 2016-08-01
GL missing


ShashaQin:
Hi,
We found GL Entry will be lost in some case.
This is one sales invoice, including one finished goods item, and update stock ticked.
![sales_invoice](https://cloud.githubusercontent.com/assets/12147007/17284327/5a049cb0-57ec-11e6-9b3b-23f84e70dd92.png)
This is the Accounting Ledger of above sales invoice, looks some account records not generated. It should have one 'finished goods' account and one 'cost of goods sold' account like the third picture.
![incorrect_gl](https://cloud.githubusercontent.com/assets/12147007/17284328/5bf3352c-57ec-11e6-9d84-c58d8b347068.png)
![correct_gl](https://cloud.githubusercontent.com/assets/12147007/17284331/5dc5e87c-57ec-11e6-8e99-331972c1b98f.png)

I'm not sure if this GL missing is caused by sales invoice created before stock entry, but the third picture is tested on version 7, that sales invoice is created before stock entry as well(Picture 1 and 2 are history data generated in the earlier version than version 7). I'm wondering if the version 7 fixed this issue, but not create a patch to fix the missing data.

rohitwaghchaure:
Hi @ShashaQin 
Can you check, what is the valuation rate of the item **item1** before the date 09/09/2015.

ShashaQin:
@rohitwaghchaure thank you for your reply. The valuation rate of the item1 is 0. 
And I tested again on version 7, I found this problem still exists. The following two scenario will cause GL missing problem:
1. Stock entry created/submitted before sales invoice, but posting date is after the posting date of sales invoice. 
2. Stock entry created/submitted after sales invoice.

ShashaQin:
This is stock ledger for item1:
![stock ledger](https://cloud.githubusercontent.com/assets/12147007/17394589/926d03ae-5a5c-11e6-8ded-5f2c1ffdd988.png)

ShashaQin:
This is GL report, the circled one couldn't find matched gl record that should be generated when submit SINV-12122 (as you can see the second picture)
![gl](https://cloud.githubusercontent.com/assets/12147007/17395067/07e28a94-5a5f-11e6-8a5e-56d22f4d29ac.png)

ShashaQin:
This will leave finish good account with amounts that won't be credited even if the correspondent Sales invoice or a Delivery Note is submitted.

nabinhait:
@ShashaQin 

[...]

PR #6335, opened 2016-09-12
[fix] warning for creating stock entry after sales invoice (causing missing gl entries)

RobertSchouten:
gives warnings and fix's for issue #5945

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -36,7 +36,6 @@ class StockEntry(StockController):
 			self.pro_doc = frappe.get_doc('Production Order', self.production_order)
 
 		self.validate_posting_time()
-		self.validate_posting_date()
 		self.validate_purpose()
 		self.validate_item()
 		self.set_transfer_qty()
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -65,6 +66,38 @@ def on_cancel(self):
...
+	def validate_posting_date(self):
+		allow_negative_stock = cint(frappe.db.get_value("Stock Settings", None, "allow_negative_stock"))
+		items, warehouses = self.get_items_and_warehouses()
+		previous_stock_vouchers = self.get_previous_stock_vouchers(self.posting_date, self.posting_time, warehouses,
+		                                                           items)
+
+		if allow_negative_stock and previous_stock_vouchers:
+			for voucher_type, voucher_no in previous_stock_vouchers:
+				frappe.msgprint(
+					_("The posting date is after {0}: {1}, this will cause missing data in the General Ledger").format(
```
nabinhait: Why this validation is required?

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -65,6 +66,38 @@ def on_cancel(self):
...
+		condition = ""
+
+		if for_items:
+			condition += " and item_code in ({})".format(", ".join(["%s"] * len(for_items)))
+			values += for_items
+
+		for d in frappe.db.sql("""select distinct sle.voucher_type, sle.voucher_no
+			from `tabStock Ledger Entry` sle
+			where timestamp(sle.posting_date, sle.posting_time) <= timestamp(%s, %s) {condition}
+			and warehouse = "Finished Goods - IAG"
```
nabinhait: hardcoded warehouse?

RobertSchouten: appologies am commiting others work will re-evaluate

erpnext/stock/doctype/stock_entry/stock_entry.py
RobertSchouten: the validation prevents stock entries being posted after there relevant sales invoice's so when the stock entry is updated the sales invoice's gl entries are also updated

nabinhait:
Thanks for the PR. Removed posting date validation and merged, did not understand the reason for that.

PR #6336, opened 2016-09-12
[fix] display unclosed values for trial/balance/sheet

RobertSchouten:
displays profit and loss as closed values like profit and loss report and adds unclosed value as total
image

### DIFF
--- a/erpnext/accounts/report/trial_balance/trial_balance.py
+++ b/erpnext/accounts/report/trial_balance/trial_balance.py
@@ -87,8 +87,7 @@ def get_opening_balances(filters):
 
 
 def get_rootwise_opening_balances(filters, report_type):
-	additional_conditions = " and posting_date >= %(year_start_date)s" \
-		if report_type == "Profit and Loss" else ""
+	additional_conditions = ""
 
 	if not flt(filters.with_period_closing_entry):
 		additional_conditions += " and ifnull(voucher_type, '')!='Period Closing Voucher'"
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/trial_balance/trial_balance.py
```diff
@@ -87,8 +87,7 @@ def get_opening_balances(filters):
 
 
 def get_rootwise_opening_balances(filters, report_type):
-	additional_conditions = " and posting_date >= %(year_start_date)s" \
-		if report_type == "Profit and Loss" else ""
+	additional_conditions = ""
```
nabinhait: any reason for removing this?

RobertSchouten: removing those lines appears to fix unclosed opening balance for income +expense on trial balance sheet so they behave like liability and equity and assets

nabinhait: I think it is not a standard expectation, users can be confused, if previous year's balance is included in P&L opening balance.

Instead, may be we can show last year's profit/loss in a separate row, just like Balance Sheet, it will be much clearer.

RobertSchouten: i think considering the trial balance sheet has the period closing entry check is it not more confusing if you have only have half the accounts showing opening balance and half not showing

nabinhait: Accountants in India don't expect this value in Trial Balance. Thats why I made it optional by adding a filter. Added this commit:
https://github.com/frappe/erpnext/commit/437b32d32c524efdc4930068204f696ca29e00fe

PR #6372, opened 2016-09-15
Patch 1

dulhaver:
tried to bring more clarity about what separates a 'Has Variants' item from a 'regular' item


vrms:
@nabinhait just confirming should such a PR always be against 'develop', not 'master'?

nabinhait:
In most cases, PR should go in develop branch. If it's a urgent fix, the PR should go on hotfix branch.

PR #6374, opened 2016-09-15
print formatting for monthly Salary register

RobertSchouten:
- id field

### DIFF
--- a/erpnext/hr/report/monthly_salary_register/monthly_salary_register.html
+++ b/erpnext/hr/report/monthly_salary_register/monthly_salary_register.html
@@ -1,0 +1,39 @@
+<div style="margin-bottom: 7px;" class="text-center">
+	{%= frappe.boot.letter_heads[filters.letter_head || frappe.defaults.get_default("letter_head")] %}
+</div>
+<h2 class="text-center">{%= __(report.report_name) %}</h2>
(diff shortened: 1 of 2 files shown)


erpnext/hr/report/monthly_salary_register/monthly_salary_register.html
```diff
@@ -0,0 +1,39 @@
...
+	{%= frappe.boot.letter_heads[filters.letter_head || frappe.defaults.get_default("letter_head")] %}
+</div>
+<h2 class="text-center">{%= __(report.report_name) %}</h2>
+<h5 class="text-center">Fiscal Year: {%= filters.fiscal_year %}</h5>
+<h5 class="text-center">Month: {%= filters.month %}</h5>
+<hr>
+<table class="table table-bordered">
+	<thead>
+		<tr>
+			{% for(var i=1, l=report.columns.length; i<l; i++) { %}
```
rmehta: You can also write jinja style `for` in js templates now 😉
```
{% for col in report %}
  <th class="text-right">{{ col.label }}</th>
{% endfor %}
```

RobertSchouten: sweet though in this case the print view is removing the row number and is starting at 1 not 0

nabinhait (approved):

PR #6375, opened 2016-09-15
[fix] calc working day on salary slip when outside default holiday list

RobertSchouten:
fix's issue created when default holiday list is changed to outside the date range of the salary slip

### DIFF
--- a/erpnext/hr/doctype/salary_slip/salary_slip.py
+++ b/erpnext/hr/doctype/salary_slip/salary_slip.py
@@ -209,10 +209,13 @@ class SalarySlip(TransactionBase):
 			self.set_month_dates()
 
 		if not joining_date:
+			get_type = None
 			joining_date, relieving_date = frappe.db.get_value("Employee", self.employee,
 				["date_of_joining", "relieving_date"])
+		else:
+			get_type = "working"
 
-		holidays = self.get_holidays_for_employee(self.start_date, self.end_date)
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -252,17 +255,45 @@ def get_payment_days(self, joining_date, relieving_date):
...
-				and holiday_date >= %(start_date)s
+				holiday_date >= %(start_date)s
 				and holiday_date <= %(end_date)s''', {
-					"holiday_list": holiday_list,
 					"start_date": start_date,
 					"end_date": end_date
-				})
+				})[0]
+
+		if count == d_count:
```
nabinhait: What if there are multiple holiday lists but each one has separate dates. In that case, it should take holiday list from Employee.

nabinhait: What if any of the holiday list does not apply to a specific employee?

RobertSchouten: have been thinking about this issue and i think the holiday lists need to be grouped up so you can assign a new holiday list while the old one stays active

rmehta:
@RobertSchouten any update on this?

RobertSchouten:
@rmehta to fix the addressed issues the holiday list requires a rewrite which i dont have time for atm

PR #6393, opened 2016-09-18
Against Voucher Type and Against Voucher added to GL

### LINKED ISSUE #5745: Add "Reference Doctype" and "Reference No" columns in General Ledger report

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/5745


erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -137,6 +139,9 @@ def get_conditions(filters):
 
 	if filters.get("voucher_no"):
 		conditions.append("voucher_no=%(voucher_no)s")
+		
+	if filters.get("against_voucher"):
+		conditions.append("voucher_no=%(against_voucher)s")
```
nabinhait: There is no filters for "against_voucher", hence it is not needed.

nabinhait (changes requested):

erpnext/accounts/report/general_ledger/general_ledger.py
KanchanChauhan: Removed

nabinhait (approved):

PR #6395, opened 2016-09-19
[minor] adding Payroll date field to process payroll and salary slip

RobertSchouten:


erpnext/hr/doctype/salary_slip/salary_slip.json
```diff
@@ -40,6 +40,31 @@
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
+   "fieldname": "payroll_date",
```
nabinhait: Can we call it posting_date, to maintain integrity to other doctypes?

RobertSchouten: @nabinhait done

nabinhait (approved):

PR #6399, opened 2016-09-19
[Fix] POS, mode of payment issue

rohitwaghchaure:


erpnext/patches/v7_0/update_mode_of_payment_type.py
```diff
@@ -0,0 +1,27 @@
...
+			mode_of_payment = [d.mode_of_payment for d in si_doc.payments if flt(d.amount) > 0]
+			if mode_of_payment != set(mode_of_payment):
+				for payment_data in si_doc.payments:
+					if payment_data.idx != 1 and payment_data.amount == si_doc.grand_total:
+						remove_payment.append(payment_data)
+						frappe.db.sql(""" delete from `tabSales Invoice Payment` 
+							where name = %(name)s""", {'name': payment_data.name})
+
+			if len(remove_payment) > 0:
+				[si_doc.remove(d) for d in remove_payment]
```
nabinhait: If you are not using the result, avoid using list comprehension. Instead use normal for loop, for better readability.

nabinhait (changes requested):

erpnext/patches/v7_0/update_mode_of_payment_type.py
rohitwaghchaure: Completed

nabinhait (approved):

PR #6419, opened 2016-09-21
[feature] preferred email for payroll

RobertSchouten:
adds a dropdown menu to employee that allows you to specify a preferred email_address to send to for payroll and newsletter
also adds description in hr settings
patch is a fix for bug created by having event on select field which changes from null to empty onload causing event to run with values from previous form


erpnext/hr/doctype/employee/employee.js
```diff
@@ -40,6 +40,28 @@ erpnext.hr.EmployeeController = frappe.ui.form.Controller.extend({
...
+	},
+	company_email:function(frm){
+		frm.events.update_contact(frm)
+	},
+	user_id:function(frm){
+		frm.events.update_contact(frm)
+	},
+	update_contact:function(frm){
+		for(var i =0, j=frm.fields.length;i<j;i++){
+			if (frm.fields[i].df.label===frm.doc.prefered_contact_email){
```
nabinhait: Label can be customized, instead implement it using fieldname.

`frm.fields[i].df.fieldname===frappe.model.scrub(frm.doc.prefered_contact_email)`

nabinhait (changes requested):

erpnext/hr/doctype/employee/employee.js
RobertSchouten: i thought about that but it doesn't allow for customisation but if u think it should b will change

erpnext/patches.txt
```diff
@@ -330,5 +330,8 @@ execute:frappe.db.sql("update `tabTimesheet` ts, `tabEmployee` emp set ts.employ
 erpnext.patches.v7_1.update_lead_source
 erpnext.patches.v7_1.fix_link_for_customer_from_lead
 erpnext.patches.v7_0.update_mode_of_payment_type
+
+execute:frappe.reload_doctype('Employee')
+execute:frappe.db.sql("update `tabEmployee` set prefered_contact_email = IFNULL(prefered_contact_email,'') ")
```
nabinhait: why this?
may be we should set "Company Email" as prefered_contact_email and company_email field value as preferred_email.

RobertSchouten: the reason for it is due to when a select field is add and has a frappe.ui.form.on hook it changes the value from null to empty causing the event trigger to trigger before the page has loaded properly resulting in the child field updating with the same value as the previous open'd employee (only occurs when you change from one employee to the other with refreshing)

if we set to company email n patch so has correct data then you cant select a blank option for opt out (for the people without email)

PR #6421, opened 2016-09-21
larger col for charge description in print

hiousi:
More room for charges and taxes descritpions in print format


hiousi:
I close this pull request because it creates more problems than it resolves.

Description needs more space in Charges and Taxes Table, but if you change col-xs-5 to col-xs-7 . The table is going to be unaligned with Total and Grand total rows.

PR #6426, opened 2016-09-21
[Enhancement] Supplier quotation added on supplier portal

rohitwaghchaure:
Supplier Quotation from supplier portal
screen shot 2016-09-21 at 4 57 54 pm
screen shot 2016-09-21 at 4 58 07 pm


umairsy:
👍

Issue #5928, opened 2016-07-29
Integration Broker


rmehta:
- [x] **Integration Service** 
  - Service Name (e.g. PayPal, RazorPay, Shopify)
  - Parameters (table) (API keys, test keys, account id etc) (make a template for each integration service)
- [x] **Integration Event Type** (e.g. Payment Initiation, Payment Complete, New Order, New Customer etc)
- [x] **Integration Request**
  - Integration Type
  - Integration Request
  - Status (queued, completed, failed)
  - Data
  - Output

PR #6449, opened 2016-09-23
Payment Request fix to call payment controller

saurabh6790:
https://github.com/frappe/frappe/pull/2087


saurabh6790:
https://github.com/frappe/erpnext/issues/5928

PR #6458, opened 2016-09-23
[FIX] Issue with customer doctype not allowing name series past 10

DeveloperMetal:
Hi Team!

This is a fix for an issue I found when creating more than 10 customers with the same name on the default name series setup, see below:

Query used to fetch mast number in series breaks due to sql MAX function handing string and number ordering differently. This causes the last "[name] - 9" series to be picked up as last in the series instead of "[name] - 10". Fix is to cast value to unsigned to MAX handles ordering correctly.

### DIFF
--- a/erpnext/selling/doctype/customer/customer.py
+++ b/erpnext/selling/doctype/customer/customer.py
@@ -53,7 +53,7 @@ class Customer(TransactionBase):
 
 	def get_customer_name(self):
 		if frappe.db.get_value("Customer", self.customer_name):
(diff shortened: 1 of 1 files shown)


icloakbuilder:
Mmm... That ci build report says that "Salary Structure" not found. Is that a side effect remnant of a failed build? I don't believe my change has anything to do with salary structure. @rmehta

rmehta:
@forellana-digithinkit yeah, we made a mistake by merging a breaking PR (will be more careful in the future) we will fix it by Monday for sure

PR #6462, opened 2016-09-24
Total and difference added for amount and qty

### LINKED ISSUE #4738: Column for Total (amount) in Stock Reconciliation Item table
When we update stock for multiple (many) items, it is very important to know that which item is contributing more to "Difference Amount". We suggest to have Total amount column similar to almost all child tables (like SO, PO, et al). 

Alternatively, Formula field type also can be added (so that it can be customized without use of any code). You may also suggest if any simple alternative method.

### LINKED ISSUE #6473: Stock Reconciliation Currency Fields
As discussed @KanchanChauhan

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/4738
Fixed https://github.com/frappe/erpnext/issues/6473


nabinhait:
merged via command line

Issue #6471, opened 2016-09-27
[BUG] Pricing Rule Currency Conversion


bcornwellmott:
I have found a bug where the currency conversion done with a pricing rule is being done when it shouldn't be. 

I've created a video to demonstrate:
https://www.youtube.com/watch?v=ritcLb_PwAw

Effectively, I have a Pricing Rule created to set an Item price in USD. My company currency is CAD.
If I try to create a PO for that item with that Supplier, the price grabbed by the Pricing Rule is placed into the CAD currency, and the price that will get charged to the customer is converted from that price. In this case, the actual price (in USD) is $1.00, but the PO shows $USD 0.76. The Price List and Supplier both have currencies set to USD.

This is a pretty major bug, as I can't issue POs without manually fixing all the prices.

bcornwellmott:
I think the issue is to do with this:
https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L162

I'm not sure why we are converting the currency here. The conversion_rate here is to convert from the company currency to the supplier/price list currency, except the pricing rule price shouldn't be in the company currency if it is tied to a Supplier (and not a Supplier Type).

rohitwaghchaure:
Hi @bcornwellmott 

At the time of making the pricing rule their is no option to set the currency against the price, so system consider the company's default currency for it. So when you created pricing rule it has currency CAD and when you are making PO for the supplier whose currency is USD, system converts the price from CAD to USD.

bcornwellmott:
It seems to me that the pricing rule (if applicable for a supplier) would inherently be in the suppliers currency. I can't imagine a scenerio where you wanted a pricing rule for a specific supplier to be in a different currency than the one they use. 

The fix I propose treats the price as being in the supplier currency of the applicable_for is set to supplier.

nabinhait:
https://github.com/frappe/erpnext/issues/6630

PR #6472, opened 2016-09-27
Only use conversion rate if using Supplier Type

bcornwellmott:
Related to issue #6471 

Only uses the conversion factor for Pricing Rules when the Applicable For field is set to Supplier Type. When set to Supplier, it will not convert the price_rate based on the conversion_rate.


nabinhait:
This is not the correct fix. We should introduce a Currency field in Pricing Rule. Based on that we should apply or convert pricing rule based on conversion rate. The supplier might have invoices in multiple currencies.

Added a github issue for that (https://github.com/frappe/erpnext/issues/6630).

PR #6477, opened 2016-09-28
[fix] move linked communications with timesheet

RobertSchouten:
moves the references on communications from timelogs to timesheets and removes any references that weren't moved

### DIFF
--- a/erpnext/patches/v7_0/convert_timelog_to_timesheet.py
+++ b/erpnext/patches/v7_0/convert_timelog_to_timesheet.py
@@ -30,7 +30,12 @@ def execute():
 		time_sheet.flags.ignore_validate = True
 		time_sheet.flags.ignore_links = True
 		time_sheet.save(ignore_permissions=True)
-
+		frappe.db.sql("""update tabCommunication 
+				set reference_doctype = "Timesheet",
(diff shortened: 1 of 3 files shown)


erpnext/patches/v7_0/convert_timelog_to_timesheet.py
```diff
@@ -30,7 +30,12 @@ def execute():
 		time_sheet.flags.ignore_validate = True
 		time_sheet.flags.ignore_links = True
 		time_sheet.save(ignore_permissions=True)
-
+		frappe.db.sql("""update tabCommunication
```
nabinhait: This should be a new patch, it should also be run for those accounts where `convert_timelog_to_timesheet.py` has already been executed.

And please send this pull requests to `hotfix` branch.

Thanks for this pull request 👍

nabinhait (changes requested):

erpnext/patches/v7_0/convert_timelog_to_timesheet.py
RobertSchouten: @nabinhait oh forgot about this one ive allready improved so can match the links and do timestamps will update tommorrow

PR #6480, opened 2016-09-28
Remove time log, time log batch from Doc Field.

rohitwaghchaure:


erpnext/patches/v7_0/remove_doctypes_and_reports.py
```diff
@@ -5,6 +5,22 @@ def execute():
...
+			'doctype': 'Custom Field',
+			'label': data.label,
+			'dt': 'Timesheet Detail',
+			'fieldname': data.fieldname
+		}).insert(ignore_permissions=True)
+
+	frappe.db.sql("""delete from `tabCustom Field` where dt = 'Time Log'""")
+	frappe.reload_doc('projects', 'doctype', 'timesheet')
+	frappe.reload_doc('projects', 'doctype', 'timesheet_detail')
+	frappe.reload_doc('custom', 'doctype', 'custom_field')
```
nabinhait: What is the need of reloading custom_field? If needed it should be in frappe.

rohitwaghchaure: Fixed

PR #6485, opened 2016-09-28
[Enhancement] Added allocate payment amount checkbox in Payment Entry

rohitwaghchaure:


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -511,6 +511,17 @@ frappe.ui.form.on('Payment Entry', {
 		});
 	},
 
+	allocate_payment_amount: function(frm) {
```
nabinhait: if payment type is internal transfer, then return

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -511,6 +511,17 @@ frappe.ui.form.on('Payment Entry', {
...
+		if(frm.doc.references.length == 0){
+			frm.events.get_outstanding_documents(frm);
+		}
+
+		if(frm.doc.payment_type == "Pay")
+			frm.events.allocate_party_amount_against_ref_docs(frm, frm.doc.received_amount);
+		else
```
nabinhait: if..else is not required, as we are returning if payment type is Internal Transfer

erpnext/accounts/doctype/payment_entry/payment_entry.json
```diff
@@ -777,6 +777,33 @@
    "bold": 0, 
    "collapsible": 0, 
    "columns": 0, 
+   "default": "1", 
+   "fieldname": "allocate_payment_amount",
```
nabinhait: Set depends on `eval:in_list(['Pay', 'Receive'], doc.payment_type)`

erpnext/accounts/doctype/payment_entry/payment_entry.js
nabinhait: if allocate checkbox is unchecked, set payment amount against all ref doc to zero.

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.json
rohitwaghchaure: Done

erpnext/accounts/doctype/payment_entry/payment_entry.js
rohitwaghchaure: Added return statement if payment type is internal transfer.

PR #6488, opened 2016-09-28
[Fix] Timesheet detail records not deleted with company transactions deletion

rohitwaghchaure:


erpnext/setup/doctype/company/delete_company_transactions.py
```diff
@@ -70,14 +69,6 @@ def delete_bins(company_name):
 	frappe.db.sql("""delete from tabBin where warehouse in
 			(select name from tabWarehouse where company=%s)""", company_name)
 
-def delete_time_sheets(company_name):
```
nabinhait: Why have you removed it? I think we should delete both timesheet and timesheet details records for the company.

PR #6513, opened 2016-09-30
Group similar items in PO

### LINKED ISSUE #6041: Feature Development: Merge repeat items in Purchase Orders generated by Material Request
Hi ERPNext team,

At the moment when a material request is generated by the production planning tool there can be multiple lines for the same item if it appears in multiple sales orders or has an order minimum. When the material request is converted into a purchase order the purchase order also has multiple lines for the same item. If the lines are added up into one in the purchase order then the system does not register the complete purchase of the material request.

When a purchase order is generated by a material request, we'd like the repeated items in the material request to be added together into one line at the purchase order level. A supplier does not need to receive a purchase order with two or more lines for the same item. 

Can you help us with this please?

thanks,
Sofia

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6041


erpnext/buying/doctype/purchase_order/purchase_order.json
```diff
@@ -2215,6 +2301,59 @@
...
+   "columns": 0, 
+   "description": "", 
+   "fieldname": "group_same_item", 
+   "fieldtype": "Check", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_list_view": 0, 
+   "label": "Group same item",
```
rmehta: should be "Group Same Items" -> labels in title case

erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -155,6 +155,30 @@ def check_modified_date(self):
...
+				if item.item_code in group_item_qty:
+					item.qty = group_item_qty[item.item_code]
+					item.amount = group_item_amount[item.item_code]
+					del group_item_qty[item.item_code]
+				else:
+					duplicate_list.append(item)
+					
+			for item in duplicate_list:
+				self.remove(item)
+
```
rmehta: add a simple test case?

rmehta (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.py
KanchanChauhan: Done

PR #6531, opened 2016-10-03
[fix] currency symbol for total amount in JE

saurabh6790:


erpnext/accounts/doctype/journal_entry/journal_entry.json
```diff
@@ -506,8 +527,36 @@
   }, 
   {
    "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "currency",
```
nabinhait: Don;t use field name "currency" for this, the field is only for specific use-case and printing, use "total_amount_currency" instead.

erpnext/patches/v7_1/set_total_amount_currency_in_je.py
```diff
@@ -0,0 +1,28 @@
...
+	for je in frappe.get_all("Journal Entry", fields=["name", "multi_currency"]):
+		if cint(je.multi_currency):
+			doc = frappe.get_doc("Journal Entry", je.name)
+			pay_to_recd_from, bank_account_currency, party_account_currency = None, None, None
+			
+			for d in doc.get('accounts'):
+				if d.party_type and d.party:
+					if not pay_to_recd_from:
+						pay_to_recd_from = frappe.db.get_value(d.party_type, d.party,
+							"customer_name" if d.party_type=="Customer" else "supplier_name")
```
nabinhait: not needed

erpnext/patches/v7_1/set_total_amount_currency_in_je.py
```diff
@@ -0,0 +1,28 @@
...
+
+					party_account_currency = d.account_currency
+
+				elif frappe.db.get_value("Account", d.account, "account_type") in ["Bank", "Cash"]:
+					bank_account_currency = d.account_currency
+
+			if pay_to_recd_from:
```
nabinhait: self.pay_to_recd_from

erpnext/patches/v7_1/set_total_amount_currency_in_je.py
```diff
@@ -0,0 +1,28 @@
+import frappe
+from frappe.utils import cint
+
+def execute():
+	for je in frappe.get_all("Journal Entry", fields=["name", "multi_currency"]):
```
nabinhait: add filters = {'multi_currency': 1, 'pay_to_recd_from': ['!=', '']}

nabinhait: above for loop, set default company currency if not multi-currency and pay_to_recd_from has value

erpnext/patches/v7_1/set_total_amount_currency_in_je.py
```diff
@@ -0,0 +1,28 @@
...
+							"customer_name" if d.party_type=="Customer" else "supplier_name")
+
+					party_account_currency = d.account_currency
+
+				elif frappe.db.get_value("Account", d.account, "account_type") in ["Bank", "Cash"]:
+					bank_account_currency = d.account_currency
+
+			if pay_to_recd_from:
+				if bank_account_currency:
+					frappe.db.set_value("Journal Entry", je.name, "currency", bank_account_currency,
```
nabinhait: instead of if... else, you can write like:

`frappe.db.set_value("Journal Entry", je.name, "currency", bank_account_currency or party_account_currency, update_modified=False)`

nabinhait (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -381,6 +381,7 @@ def set_print_format_fields(self):
 
 	def set_total_amount(self, amt, currency):
 		self.total_amount = amt
+		self.currency = currency
```
nabinhait: total_amount_currency

erpnext/patches/v7_1/set_total_amount_currency_in_je.py
```diff
@@ -0,0 +1,24 @@
...
+
+def execute():
+	frappe.reload_doc("accounts", "doctype", "journal_entry")
+
+	frappe.db.sql(""" update `tabJournal Entry` set total_amount_currency = %s
+		where ifnull(multi_currency, 0) = 0
+		and (pay_to_recd_from is not null or pay_to_recd_from != "") """, get_default_currency())
+
+	for je in frappe.get_all("Journal Entry", fields=["name", "multi_currency"],
+		filters={'multi_currency': 1, 'pay_to_recd_from': ['!=', '']}):
```
nabinhait: this will return where pay_to_recd_from is null

Issue #6540, opened 2016-10-04
Option to make payment via Journal Entry


nabinhait:
- "Make Payment via Journal Entry" checkbox in Accounts Settings.
- If checked, "Make Payment" button in invoice will map to Journal Entry as before.

PR #6550, opened 2016-10-05
[Enhancement] Option to make payment via Journal Entry

rohitwaghchaure:
#6540


nabinhait:
This should be implemented for SO, PO, SI and PI

rohitwaghchaure:
Done

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -49,6 +49,9 @@ def set_indicator(self):
 			self.indicator_color = "green"
 			self.indicator_title = _("Paid")
 
+	def onload(self):
+		self.get("__onload").journal_entry = frappe.db.get_single_value('Accounts Settings', 'make_payment_via_journal_entry')
```
nabinhait: Can you rename the flag to `make_payment_via_journal_entry `? It will be more explanatory in the client side usage. 

Try to commonify the `onload` (in accounts_controller.py?). 
And there is an existing `onload` in selling_controller.py, take care of that.

rohitwaghchaure: Done

PR #6572, opened 2016-10-07
[fix] update timestamps and relink communication for timesheet

RobertSchouten:


nabinhait:
Removed updation of creation/modified part, as there is already a patch (https://github.com/frappe/erpnext/blob/hotfix/erpnext/patches/v7_0/update_missing_employee_in_timesheet.py) for that.

RobertSchouten:
@nabinhait note the where clause
https://github.com/frappe/erpnext/blob/hotfix/erpnext/patches/v7_0/update_missing_employee_in_timesheet.py#L13
causes the update to not occur as only updates when employee is missing

nabinhait:
@RobertSchouten Thanks for pointing out, will include it again. Keep up the great work 👍

PR #6575, opened 2016-10-07
[Fix] Allow user to save Notification Control without selecting trans…action

### LINKED ISSUE #6558: critial : cannot save at email Notification Control Not Saved
the update button is not working
will not save the setting

shreyasp:
Problem: If user does not select any transaction in Notification Control form to customize then, user was not allowed to save the form

Solution: If no transaction is selected just save the form to update values for notification prompts.

Fixes #6558

### DIFF
--- a/erpnext/setup/doctype/notification_control/notification_control.js
+++ b/erpnext/setup/doctype/notification_control/notification_control.js
@@ -9,7 +9,9 @@ frappe.ui.form.on("Notification Control", {
 		frm.set_value("custom_message", frm.doc[frm.events.get_fieldname(frm)]);
 	},
 	set_message: function(frm) {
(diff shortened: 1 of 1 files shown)


hoeichia:
partly fix...
the 'Prompt for Email on Submission of' checkbox part,
will not save and update alone if you check or uncheck the selection
if the Customize the Notification is empty, will prompt "Field _message not found."

shreyasp:
@hoeichia I have updated the code so that notification control page can be saved. But after saving, user will have to "reload" from the desk, for changes to take effect.

hoeichia:
if the "Customize the Notification" selection is empty
the "Field _message not found. " will be prompt when click on update,
you have to select something then only you can save

Issue #6567, opened 2016-10-06
Material Request


rohitwaghchaure:
- [x] Add latest images 
- [x] Add RFQ in the flowchart
- [x] Add information about RFQ

PR #6579, opened 2016-10-08
[fix] #6567

PawanMeh:
Documentation fixes for Material Request


erpnext/docs/user/manual/en/stock/material-request.md
```diff
@@ -25,6 +25,8 @@ A Material Request can be of type:
 * Material Issue - If the requested material is to be Issued.
 * Manufacture - If the requested material is to be Produced.
 
+You can also raise a [RFQ]({{docs_base_url}}/user/manual/en/buying/request-for-quotation.html) against a material request by using the 'Make' button.
```
rmehta: Must be full "Request for Quote"

Might be helpful to put more content here rather than just a one line. The user can also click "Make" and see!

Also all document types "Material Request" must be capital and bold, like **Material Request**

PR #6626, opened 2016-10-17
No Leave allocation for Leave Without Pay

### LINKED ISSUE #4110: No Leave allocation for Leave Type for which Leave Without Pay is checked
Leave allocation should not be possible for Leave Type for which "Leave Without Pay" is checked.

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/4110


erpnext/hr/doctype/leave_allocation/leave_allocation.js
```diff
@@ -11,6 +11,13 @@ frappe.ui.form.on("Leave Allocation", {
 			return {
 				query: "erpnext.controllers.queries.employee_query"
 			}
+		});
+		frm.set_query("leave_type", function() {
```
KanchanChauhan: @nabinhait Done

PR #6628, opened 2016-10-17
Posting Date cannot be future date in Purchase Receipt

### LINKED ISSUE #3859: Purchase Receipts with future date
When creating a Purchase Receipt, the system allows a future date which is not correct.

prec-date issue

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/3859

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -4,9 +4,9 @@
 from __future__ import unicode_literals
 import frappe
 
-from frappe.utils import flt, cint
+from frappe.utils import flt, cint, nowdate
 
-from frappe import _
+from frappe import throw, _
 import frappe.defaults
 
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -58,6 +58,10 @@ def validate(self):
 
 		pc_obj = frappe.get_doc('Purchase Common')
 		self.check_for_closed_status(pc_obj)
+		
+		if self.posting_date > nowdate():
+			throw(_("Posting Date cannot be before Current Date"))
```
nabinhait: after

nabinhait (changes requested):

erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
KanchanChauhan: @nabinhait Done

PR #6633, opened 2016-10-17
Added Employee Loan Doctype

adityaduggal:
-Emp Loan only doctype added with validations


erpnext/hr/doctype/employee_loan/employee_loan.py
```diff
@@ -0,0 +1,110 @@
...
+from __future__ import unicode_literals
+from frappe.utils import getdate
+import frappe
+from frappe.model.document import Document
+import math
+
+class EmployeeLoan(Document):
+	def validate(self):
+		self.total_loan = 0
+		for i in self.employee_loan_detail:
```
vjFaLk: Similar feedback here, i is not a great variable name. maybe use `employee`

erpnext/hr/doctype/employee_loan/employee_loan.js
```diff
@@ -0,0 +1,41 @@
...
+// For license information, please see license.txt
+
+frappe.ui.form.on('Employee Loan', {
+	refresh: function(frm) {
+
+	}
+});
+
+frappe.ui.form.on('Employee Loan Detail', "loan_amount", function(frm, cdt, cdn){
+	var d = locals[cdt][cdn]
```
rmehta: `frappe.get_doc(cdt, cdn)` also does it, but `locals` is fine :)

rmehta:
Closing this, since this is now part of #6808

PR #6653, opened 2016-10-19
[Minor] Do not allow change in credit limit

### LINKED ISSUE #3511: Suggestion for Customer Credit Limit
Example:
If Customer entry created with  below data
Customer Name=Geetanjali
Credit Limit =500

For above customer ,
1.If sales Order created with Amount =500 but not billed  

2.Again Change the Credit Limit  as 200.
3.Try to create the Sales Order with Amount =200 then error message displaying as 
“Credit limit has been crossed for customer Geetanjali 500.0/200.0”.

My suggestion  is 
At the time of Step 2 , Error message should display as
“Credit Limit Already Crossed.
 Here You can set Credit Limit as greater than Current Credit Limit Or Make Payment for Previous       Credit.”

shreyasp:
- If customer outstanding amount is greater than new credit limit, then do not allow user to change to credit limit.
- [x] Test Case

Fixes #3511

### DIFF
--- a/erpnext/selling/doctype/customer/test_customer.py
+++ b/erpnext/selling/doctype/customer/test_customer.py
@@ -8,6 +8,7 @@ import unittest
 
 from frappe.test_runner import make_test_records
 from erpnext.exceptions import PartyFrozen, PartyDisabled
+from frappe.utils import flt
 
 test_ignore = ["Price List"]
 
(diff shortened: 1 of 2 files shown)


erpnext/selling/doctype/customer/test_customer.py
```diff
@@ -111,6 +112,52 @@ def test_duplicate_customer(self):
...
+
+	def test_customer_credit_limit_on_change(self):
+		from erpnext.selling.doctype.customer.customer import get_credit_limit, get_customer_outstanding
+		from erpnext.selling.doctype.sales_order.test_sales_order import make_sales_order
+
+		outstanding_amt = get_customer_outstanding('_Test Customer', '_Test Company')
+		credit_limit = get_credit_limit('_Test Customer', '_Test Company')
+
+		customer = frappe.get_doc("Customer", '_Test Customer')
+		customer.credit_limit = flt(outstanding_amt - 100)
```
nabinhait: What if outstanding_amt less than or equal to zero?

In that case, first make a sales order with `mod(outstanding_amt) + 100`

shreyasp:
@nabinhait Done the requested changes, Please review and merge

PR #6657, opened 2016-10-19
[Feature] Provision to allow unlink the payment against the invoice

### LINKED ISSUE #3460: Please add an option to prevent cancelling paid invoices (Sales & Purchase) without cancelling the payment voucher(s) first
Please help to add and option/setting (in the setup maybe) to prevent cancelling paid invoices (partially paid or fully paid in Sales Invoices & Purchase Invoices) without cancelling the related payment voucher(s) first.
This is very important feature because it is very hard to force the users to do that.
Maybe it could be a similar feature like "Balance Must Be" option in Account Setup.
Thanks

rohitwaghchaure:
Fixed #3460


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -194,6 +195,32 @@
    "search_index": 0, 
    "set_only_once": 0, 
    "unique": 0
+  }, 
+  {
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "unlink_payment_on_cancellation_of_invoice",
```
nabinhait: Set default value as 1
And write a patch to check it in existing accounts

rohitwaghchaure: Done

Issue #6363, opened 2016-09-14
Project module enhancements


PawanMeh:
- [x] Allow to add weights for tasks.
- [x] Add percent completion method at project level, to allow for calculation of % complete of project based on weighted average of tasks.

OneiricIT:
Can I also add(?):

- [ ] Sales Quotation reference for Projects

- [ ] Add Sales Quotation in Project Related Documents

PawanMeh:
I think project is created after a sales order, the customer may not accept your sales quotation , why would you create a project in the system to execute it?

OneiricIT:
We have many situations where we need to quote on additional work and/or products within an existing project, so in order to properly manage and track the sales flow we also needed references to quotations created from the project.

PawanMeh:
You can add a custom field to Sales Quotation for tracking project in that case and get it via a report OR you could develop the feature and contribute OR you could request it as part of GitHub issue 6326, which is raised for related documents.

OneiricIT:
Thanks. Already doing 1st point at present, will also jump onto 6326.

PawanMeh:
If you have developed some custom solution for it as part of #1, I think you should try to contribute it back to the community.

OneiricIT:
Haven't done custom report or anything worth noting, just added the custom field so not much to contribute sorry.

mhbu50:
+1

PR #6662, opened 2016-10-20
[fix] #6363

PawanMeh:
1. Added ability to add 'Weights' against Tasks related to a project
2. Added calculation methods for % Tasks Completed based on weights, percentage & completion
3. Added documentation changes


erpnext/projects/doctype/project/project.json
```diff
@@ -121,6 +127,35 @@
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
+   "columns": 0, 
+   "default": "Task Completion", 
+   "fieldname": "pct_comp_method",
```
rmehta: Always use full names `percent_complete_method`

Also maybe this should be in the "Tasks" section rather than the header section

PawanMeh: 1. I have made the changes regarding the renaming of the field to percent_complete_method.

1. For the % Complete Method and %Tasks Completed, I feel they need to remain on the top where the user can see them without scrolling. The reasons being that %Complete Method usually is set up when the project is set up initially (you may or may not know all tasks at this point) and once the project starts a key parameter is %Tasks Completed which should be available upfront without scrolling.

Also both the fields with a vertical scroll in between may mess up the Grid alignment.

You can let me know if you still want #2 to be changed and I will re-push the changes accordingly.

erpnext/projects/doctype/project/project.json
```diff
@@ -967,13 +1034,18 @@
...
  "idx": 29, 
+ "image_view": 0, 
  "in_create": 0, 
  "in_dialog": 0, 
  "is_submittable": 0, 
  "issingle": 0, 
  "istable": 0, 
  "max_attachments": 4, 
- "modified": "2016-04-22 03:15:39.635420", 
+<<<<<<< 6ef2c9faa43ba40bef00abc29cd9a5de980b2037
```
rmehta: You need to remove this merge conflict

rmehta (changes requested):

PR #6688, opened 2016-10-24
Update domainify.py

umairsy:


umairsy:
@nabinhait please let me know when reviewing. Ref.: WN-SUP21215

Issue #3720, opened 2015-07-24
BOM: Material list shows wrong currency sign in rate and amount field.


chaoyee:
In item price, the price for this item has been set as "49.6" in currency USD.  
![2015-07-25 12 55 19](https://cloud.githubusercontent.com/assets/2167909/8880171/4135f516-3269-11e5-8b85-0e2fbdd659a8.png)

In BOM operation, however, "Material" list shows wrong currency sign (NT$)  but the same price (49.6) in rate and amount field.
![2015-07-25 12 54 20](https://cloud.githubusercontent.com/assets/2167909/8880105/c63e64a6-3268-11e5-8db5-aea0dedb03e4.png)

The currency exchange rate for TWD to USD has been set.  If "Material" list shows "NT$", the rate and amount should be converted through the exchange rate.  But it does not happen.

rmehta:
Thanks @chaoyee this needs to be fixed.

PR #6699, opened 2016-10-24
[Enhancement] Currency added in the BOM

rohitwaghchaure:
#3720


erpnext/docs/user/manual/en/manufacturing/bill-of-materials.md
```diff
@@ -31,6 +31,11 @@ To make a new BOM:
 
 <img class="screenshot" alt="Update Cost" src="{{docs_base_url}}/assets/img/manufacturing/bom-update-cost.png">
 
+* User can select the currency in the BOM 
+* System calculates the costing based on the price list currency
+
+<img class="screenshot" alt="Update Cost" src="{{docs_base_url}}/assets/img/manufacturing/price-list-based-currency-bom.png">
+
```
rmehta: Please put only the browser page in the screenshot, not the dock and the browser tabs!

rmehta (changes requested):

PR #6701, opened 2016-10-25
[Fix] Make purchase order item reqd_date modifiable after submit

### LINKED ISSUE #3852: Schedule date in Purchase Order Item should be modifiable after Submit
Hi,

It is very common for us to face delays/changes in delivery schedule dates from our suppliers much after the Purchase Order was placed. 

We would like to change the date in the PO (after submit) so that we are able to keep track of the delivery dates. Currently we cannot update "Schedule Date" in Purchase Order Item as it is a Standard field that cannot be marked "Allow on Submit".

Regards,

Priyanka Kaushal

shreyasp:
Fixes #3852


bcornwellmott:
I'm not sure this is the best fix. See the discussion here: https://discuss.erpnext.com/t/material-request-required-by-date-vs-purchase-order-expected-date/16931

The required by date does not change regardless of when your supplier is expected to deliver. I think adding a new field for expected delivery date is a better solution.

nabinhait (approved):

PR #6702, opened 2016-10-25
Documentation update

russiang:


vjFaLk:
@russiang Something has gone awry here, the commit author seems to be @netchampfaris

netchampfaris:
what just happened? :|

PR #6705, opened 2016-10-25
Fetch serial no by default

### LINKED ISSUE #6645: Serial Number by default FIFO for Production Order

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6645

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -379,60 +379,6 @@ cur_frm.fields_dict['items'].grid.get_field('batch_no').get_query = function(doc
 	}
 }
 
-cur_frm.cscript.item_code = function(doc, cdt, cdn) {
-	var d = locals[cdt][cdn];
-	if(d.item_code) {
-		args = {
-			'item_code'			: d.item_code,
-			'warehouse'			: cstr(d.s_warehouse) || cstr(d.t_warehouse),
-			'transfer_qty'		: d.transfer_qty,
-			'serial_no	'		: d.serial_no,
-			'bom_no'			: d.bom_no,
-			'expense_account'	: d.expense_account,
-			'cost_center'		: d.cost_center,
-			'company'			: cur_frm.doc.company
-		};
-		return frappe.call({
-			doc: cur_frm.doc,
-			method: "get_item_details",
-			args: args,
-			callback: function(r) {
-				if(r.message) {
-					var d = locals[cdt][cdn];
-					$.each(r.message, function(k, v) {
-						d[k] = v;
-					});
-					refresh_field("items");
-				}
-			}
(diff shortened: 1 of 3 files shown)


bcornwellmott:
I'm concerned about having this happen automatically without any user interaction. Users could forget this is a serial numbered part and transfer incorrect serial numbers without any warning or prompt. Could we have this trigger on a button press or a checbox in the Stock Settings?

KanchanChauhan:
@bcornwellmott It will happen only if "Automatically Set Serial Nos based on FIFO" is checked in Stock Settings.

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -409,6 +430,15 @@ cur_frm.cscript.item_code = function(doc, cdt, cdn) {
 	}
 }
 
+
+cur_frm.cscript.qty = function(doc, cdt, cdn) {
```
nabinhait: Always use `frappe.ui.form.on`

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -305,6 +305,25 @@ erpnext.stock.StockEntry = erpnext.stock.StockController.extend({
 
 		this.frm.set_value("total_additional_costs", flt(total_additional_costs, precision("total_additional_costs")));
 	},
+
+	set_serial_no: function(doc, cdt, cdn) {
+		var d = locals[cdt][cdn];
```
nabinhait: return if no `s_warehouse && item_code && qty`

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -493,6 +494,9 @@ def get_item_details(self, args=None, for_update=False):
 		stock_and_rate = args.get('warehouse') and get_warehouse_details(args) or {}
 		ret.update(stock_and_rate)
 
+		if args.get('fetch_serial_no'):
```
nabinhait: Actually it does not make any sense, as qty is always zero.

erpnext/stock/get_item_details.py
```diff
@@ -339,7 +339,7 @@ def get_pos_profile(company):
 	return pos_profile and pos_profile[0] or None
 
 
-def get_serial_nos_by_fifo(args, item_doc):
+def get_serial_nos_by_fifo(args, item_doc = None):
```
nabinhait: `item_doc` is not used

nabinhait (changes requested):

erpnext/stock/get_item_details.py
KanchanChauhan: All changes done

nabinhait (approved):

PR #6738, opened 2016-10-28
Different name on Check

### LINKED ISSUE #6604: Printing other than party name on cheque
Is there any way to pint different name instead of supplier name on cheque. As some small companies they don't have company account they accept cheques on personal names.

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6604

If "Name on Check" given, that will be printed on check, else Party Name. Only if Mode of Payment is Check.


erpnext/accounts/doctype/payment_entry/payment_entry.json
```diff
@@ -282,6 +282,34 @@
   {
    "allow_on_submit": 0, 
    "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "depends_on": "eval:doc.mode_of_payment==\"Check\"", 
+   "description": "If name different from Party name", 
+   "fieldname": "name_on_check",
```
nabinhait: Instead of adding a new field, allow editing party_name after submit and remove the field from hidden.

nabinhait (changes requested):

nabinhait (approved):

PR #6780, opened 2016-11-01
Change fieldtype from float to currency

rohitwaghchaure:


erpnext/selling/doctype/sales_order_item/sales_order_item.json
```diff
@@ -563,7 +563,7 @@
    "columns": 0, 
    "depends_on": "eval:doc.margin_type && doc.price_list_rate", 
    "fieldname": "total_margin", 
-   "fieldtype": "Float", 
+   "fieldtype": "Currency",
```
nabinhait: @rohitwaghchaure You have to change fieldtype for margin in all other doctypes as well.

rohitwaghchaure: Done

nabinhait:
@rohitwaghchaure why you did not change fieldtype for "Margin Rate or Amount" field? Also same in Pricing Rule?

rohitwaghchaure:
@nabinhait 
"Margin Rate or Amount" field is used as a percentage / amount. So user may get confused if fieldtype is currency and value is percentage

PR #6781, opened 2016-11-01
Change fieldtype from float to currency

rohitwaghchaure:


erpnext/selling/doctype/sales_order_item/sales_order_item.json
```diff
@@ -542,7 +542,7 @@
    "collapsible": 0, 
    "depends_on": "eval:doc.margin_type && doc.price_list_rate", 
    "fieldname": "total_margin", 
-   "fieldtype": "Float", 
+   "fieldtype": "Currency",
```
nabinhait: @rohitwaghchaure You have to change fieldtype for margin in all other doctypes as well.

nabinhait (changes requested):

erpnext/selling/doctype/sales_order_item/sales_order_item.json
rohitwaghchaure: Done

PR #6784, opened 2016-11-01
[fix] calendar range fixes

RobertSchouten:
shows entries that start before the calendar and end after

### DIFF
--- a/erpnext/projects/doctype/timesheet/timesheet_calendar.js
+++ b/erpnext/projects/doctype/timesheet/timesheet_calendar.js
@@ -1,7 +1,7 @@
 frappe.views.calendar["Timesheet"] = {
 	field_map: {
-		"start": "from_time",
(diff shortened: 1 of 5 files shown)


erpnext/projects/doctype/timesheet/timesheet_calendar.js
```diff
@@ -1,7 +1,7 @@
 frappe.views.calendar["Timesheet"] = {
 	field_map: {
-		"start": "start_date",
-		"end": "end_date",
+		"start": "from_time",
```
bcornwellmott: This was changed to accommodate the gantt chart. Please look at the history.

bcornwellmott:
I've added a comment...

RobertSchouten:
@bcornwellmott your right i missed that. 
fixed

netchampfaris (approved):

Issue #4946, opened 2016-03-09
Report: Bin - Item Shortage Report -- Should have Item Name


Devens:
It should be feasible to get "Item Name" from Pick Column for above mentioned report. 

![image](https://cloud.githubusercontent.com/assets/10671911/13625356/f1d36bca-e5dc-11e5-85e2-776460943912.png)

manassolanki:
Fixed

PR #6786, opened 2016-11-01
Report: Bin - Item Shortage Report -- Should have Item Name #4946

manassolanki:
New report is build including a new column Item name


nabinhait (changes requested):
@manassolanki Congrats for the first PR 👍 
- Also add description column
- Place both item name and description columns at the end

KanchanChauhan:
👍

Issue #5504, opened 2016-06-17
Journal Entry - Remove Account if it has no Debit or Credit value


umairsy:
Journal Entry - Remove Account if it has no Debit or Credit value.

![screen shot 2016-06-17 at 12 20 14 pm](https://cloud.githubusercontent.com/assets/910238/16142777/ee6b417c-3485-11e6-880d-755272671dc6.png)

nabinhait:
For reference

https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/payment_entry/payment_entry.py#L280

PR #6813, opened 2016-11-03
Journal Entry Issue #5504

manassolanki:
Remove Account if it has no Debit or Credit value

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -39,6 +39,7 @@ class JournalEntry(AccountsController):
 		self.set_account_and_party_balance()
 		if not self.title:
 			self.title = self.get_title()
+		self.clear_zero_debit_credit_row()
 
 	def on_submit(self):
 		self.check_credit_limit()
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -527,6 +528,10 @@ def set_account_and_party_balance(self):
 			d.account_balance = account_balance[d.account]
 			d.party_balance = party_balance[(d.party_type, d.party)]
 
+	def clear_zero_debit_credit_row(self):
+		self.accounts = [account for account in self.accounts if not (account.debit_in_account_currency==0.0 and account.credit_in_account_currency==0.0)]
```
nabinhait: Break the line after 90 chars

erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -527,6 +528,10 @@ def set_account_and_party_balance(self):
...
+		frappe.db.sql("""delete from `tabJournal Entry Account` where parent = %s and debit = 0.0 and credit = 0.0""", self.name)
```
nabinhait: use `debit_in_account_currency` instead of `debit`

nabinhait (changes requested):

nabinhait (commented):
Test case is failing. If you need help to understand test system, let me know.

PR #6819, opened 2016-11-04
cleanup and fixes

rohitwaghchaure:


erpnext/patches.txt
```diff
@@ -291,7 +291,7 @@ erpnext.patches.v7_0.rename_prevdoc_fields
 erpnext.patches.v7_0.rename_time_sheet_doctype
 execute:frappe.delete_doc_if_exists("Report", "Customers Not Buying Since Long Time")
 erpnext.patches.v7_0.make_is_group_fieldtype_as_check
-execute:frappe.reload_doc('projects', 'doctype', 'timesheet') #2016-09-12
+execute:frappe.reload_doc('projects', 'doctype', 'timesheet', force=True) #2016-11-04
```
nabinhait: Never do `force=True` while reloading, it resets permissions.

nabinhait (changes requested):

erpnext/patches.txt
rohitwaghchaure: Shall I remove?

nabinhait: You need to add those permission via a separate patch. If the new role is not exists, add it via patch. But we can't reset the whole permissions, it will remove customization if any.

rohitwaghchaure: Done please check

erpnext/patches/v7_1/add_account_user_role_for_timesheet.py
```diff
@@ -0,0 +1,22 @@
...
+def execute():
+	if not frappe.db.get_value('DocPerm', {'parent': 'Timesheet', 'role': 'Accounts User', 'permlevel': 1}):
+		doc = frappe.get_doc('DocType', 'Timesheet')
+		for perm in [0, 1]:
+			doc.append('permissions', {
+				'role': "Accounts User",
+				'permlevel': perm,
+				'read': 1,
+				'write': 1,
+				'create': 1,
```
nabinhait: In permlevel 1 (>0), only read and write is applicable.

rohitwaghchaure: Done

nabinhait (approved):

PR #6823, opened 2016-11-04
Item Inspection before Delivery

### LINKED ISSUE #5307: Quality Inspection for Manufactured Items.
Another minor issue I've found... if an Item is manufactured there is no catch if "Quality Inspection Required" is active. This seems to only actually require the inspection record for Purchase Receipt. I think the best place to put another catch is in the Delivery Note. If a product is on a Delivery Note and has not completed its required quality inspection, the Delivery Note should be rejected.

KanchanChauhan:
- Added check in Item if "Inspection required before delivery"
- Message on saving Delivery Note if no Quality Inspection selected.

Fixed https://github.com/frappe/erpnext/issues/5307


erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -6,16 +6,48 @@
 frappe.provide("erpnext.stock");
 frappe.provide("erpnext.stock.delivery_note");
 
-frappe.ui.form.on('Delivery Note', 'onload', function(frm) {
-	frm.set_indicator_formatter('item_code',
-		function(doc) {
-			return (doc.docstatus==1 || doc.qty<=doc.actual_qty) ? "green" : "orange"
+frappe.ui.form.on("Delivery Note", {
+	onload: function(frm) {
+		var qa_no = frappe.meta.get_docfield("Delivery Note Item", "qa_no");
```
rmehta: this should be on `setup` event instead of `onload`

`onload` is called everytime a new delivery note is loaded, and `setup` is called when the form object (DOM) is created (only once)

rmehta (changes requested):

erpnext/stock/doctype/delivery_note/delivery_note.js
KanchanChauhan: Done

rmehta (approved):

PR #6833, opened 2016-11-05
Credit Note, Debit Note, Excise Entry info added

### LINKED ISSUE #6512: Journal Entry
- [x] Add details about debit note, credit note, excise entry details with example 
- [ ] Add video of any use case

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6512

### DIFF
--- a/erpnext/docs/user/manual/en/accounts/journal-entry.md
+++ b/erpnext/docs/user/manual/en/accounts/journal-entry.md
@@ -68,4 +68,44 @@ of each year reducing its value by a certain percentage.
 > Note: There may be regulations in your country that define by how much
 amount you can depreciate a class of Assets.
 
+#### Credit Note
+
+"Credit Note" is made for a Customer against a Sales Invoice when the 
+company needs to adjust a payment for returned goods. When a Credit Note
+is made, the seller can either make a payment to the customer or adjust 
+the amount in another invoice.
+
+  * Debit: Sales Return Account
+  * Credit: Customer Account
+  
+#### Debit Note
+
+"Debit Note" is made for a Supplier against a Purchase Invoice or accepted 
+as a credit note from supplier when a company returns goods. When a Debit
+Note is made, the company can either receive a payment from the supplier or 
+adjust the amount in another invoice.
+
+  * Debit: Supplier Account
+  * Credit: Purchase Return Account 
+
+#### Excise Entry
+
+When a Company buys good from a Supplier, company pays excise duty
+on these goods to Supplier. And when a company sells these goods to Customers, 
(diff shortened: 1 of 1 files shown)


erpnext/docs/user/manual/en/accounts/journal-entry.md
```diff
@@ -68,4 +68,38 @@ of each year reducing its value by a certain percentage.
...
+#### Debit Note
+
+Debit note is a note which is given to the supplier or accepted as a credit note 
+from our supplier when we return goods. By giving this note, we try to tell that 
+we are debiting his account with returned goods amount. 
+
+  * Debit: Supplier Account
+  * Credit: Purchase Return Account 
+
+#### Excise Entry
```
rmehta: This is only for India, maybe this should be a separate article (maybe we should have separate sections for country-wise pages)

erpnext/docs/user/manual/en/accounts/journal-entry.md
```diff
@@ -68,4 +68,38 @@ of each year reducing its value by a certain percentage.
 > Note: There may be regulations in your country that define by how much
 amount you can depreciate a class of Assets.
 
+#### Credit Note
+
+Credit note is a note which is given to the customer when we get his returned
```
rmehta: Try and use gender neutral language, also documentation is best written in 3rd person.

`"Credit Note" is made for a Customer against a Sales Invoice when the company needs to refund or adjust a payment. When a Credit Note is made, the seller can either make a payment to the customer or adjust the amount in another invoice.`

erpnext/docs/user/manual/en/accounts/journal-entry.md
```diff
@@ -68,4 +68,38 @@ of each year reducing its value by a certain percentage.
...
 
+#### Credit Note
+
+Credit note is a note which is given to the customer when we get his returned 
+goods. By giving this note to customer, we try to tell that we are crediting his 
+account with his returned goods amount.
+
+  * Debit: Sales Return Account
+  * Credit: Customer Account
+
```
rmehta: Screenshot would be better!

rmehta (changes requested):

erpnext/docs/user/manual/en/accounts/journal-entry.md
KanchanChauhan: Sales Invoice or Sales Return?

erpnext/docs/user/manual/en/accounts/journal-entry.md
KanchanChauhan: Added a note, since option is available for other countries as well.

nabinhait (approved):

PR #6842, opened 2016-11-07
have profit and loss icon hidden by default

RobertSchouten:


rmehta:
Any specific reason? Profit and Loss is critical to a company. Also will only be shown based on permission

RobertSchouten:
@rhmeta Well that's the issue then users without permission having the icon

PR #6850, opened 2016-11-07
Clear rows from Journal Entry if no debit/credit

nabinhait:


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -37,8 +37,14 @@ def validate(self):
 		self.validate_credit_debit_note()
 		self.validate_empty_accounts_table()
 		self.set_account_and_party_balance()
+		self.clear_zero_debit_credit_row()
 		if not self.title:
 			self.title = self.get_title()
+			
+	def clear_zero_debit_credit_row(self):
```
rmehta: why delete from database?

Can be cleared in `self.accounts`

rmehta (changes requested):

PR #6860, opened 2016-11-08
Added status field in sales invoice and purchase invoice

rohitwaghchaure:


nabinhait (changes requested):
Add another status for negative outstanding amount, may be "Credit/Debit Note Issued"?

rohitwaghchaure:
Done

erpnext/patches/v7_1/update_invoice_status.py
```diff
@@ -0,0 +1,32 @@
...
+
+def execute():
+	frappe.reload_doc('accounts', 'doctype', 'sales_invoice')
+	frappe.reload_doc('accounts', 'doctype', 'purchase_invoice')
+
+	frappe.db.sql(""" 
+		update 
+			`tabPurchase Invoice` 
+		set 
+			status = (Case When outstanding_amount = 0 and docstatus = 1 then 'Paid'
```
nabinhait: set status as return if is_return checked

nabinhait (changes requested):

nabinhait (approved):

PR #6898, opened 2016-11-10
[Minor] Shopping Cart Fix

KanchanChauhan:


erpnext/templates/includes/navbar/navbar_items.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'frappe/templates/includes/navbar/navbar_items.html' %}
 
 {% block navbar_right_extension %}
-	<li class="shopping-cart hidden">
+	<li class="shopping-cart">
```
rmehta: Will it show by default?

PR #6903, opened 2016-11-11
[Fix] Date of Joining/Releiving  filter fixed for Process Payroll

### LINKED ISSUE #6870: Salary not being created from Process Payroll for new employee
This is in addition to issue #6834 

The last employee joined on 05-Oct. Now when we create his salary slip from Process Payroll for October, the system shows there is no such employee. But it successfully creates salary for November, for the same employee.

It seems the system is unable to create Salary Slip for the employee who joins after start of month like in my case the employee joined on 5th. Probably the code needs attention for this peculiar situation.

Thanks!

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6870


nabinhait (commented):
@KanchanChauhan Again it should go into hotfix branch.

PR #6920, opened 2016-11-13
[Fix] Fetch Serial No by default for  Delivery Note Item

### LINKED ISSUE #6706: Fix Serial Number by default FIFO for Delivery Note

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6706

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -11,6 +11,7 @@ from frappe import throw, _
 from frappe.utils import flt, cint
 from frappe.model.document import Document
 
+
 class MultiplePricingRuleConflict(frappe.ValidationError): pass
 
 class PricingRule(Document):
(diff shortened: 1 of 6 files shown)


erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -288,3 +288,38 @@ if (sys_defaults.auto_accounting_for_stock) {
 		}
 	}
 }
+
+frappe.ui.form.on('Delivery Note', {
+	set_serial_no: function(doc, cdt, cdn) {
+		var d = frappe.model.get_doc(cdt, cdn);
+		d.qty||(d.qty=1)
```
nabinhait: ???

erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -288,3 +288,38 @@ if (sys_defaults.auto_accounting_for_stock) {
...
+		if(!d.item_code && !d.warehouse && !d.qty) return;
```
nabinhait: all the 3 fields are mandatory right? In that case, condition should be based on "OR"

erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -288,3 +288,38 @@ if (sys_defaults.auto_accounting_for_stock) {
 		}
 	}
 }
+
+frappe.ui.form.on('Delivery Note', {
+	set_serial_no: function(doc, cdt, cdn) {
```
nabinhait: use `frm` instead of `doc` in args

nabinhait (changes requested):

KanchanChauhan:
Changes done

KanchanChauhan:
Call set_serial_no from server side

nabinhait (approved):

PR #6930, opened 2016-11-14
[Fix]Product thumbnail in Shopping Cart

KanchanChauhan:


rmehta (changes requested):
can you send this to hotfix - must be breaking in current too.

PR #6944, opened 2016-11-15
[Fix] Status updater issue for sales order in purchase order

rohitwaghchaure:


rmehta:
please add some comments so that it will help while merging.

PR #6947, opened 2016-11-15
[fix] delink jv from advance table

saurabh6790:
on JV cancel remove linkings from the advance table of Sales/Purchase Invoice

### DIFF
--- a/erpnext/utilities/transaction_base.py
+++ b/erpnext/utilities/transaction_base.py
@@ -129,11 +129,16 @@ class TransactionBase(StatusUpdater):
 	def delink_advance_entries(self, jv):
 		total_allocated_amount = 0
 		for adv in self.advances:
+			consider_for_total_advance = True
 			if adv.reference_name == jv:
 				frappe.db.sql("""delete from `tab{0} Advance`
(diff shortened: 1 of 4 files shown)


erpnext/utilities/transaction_base.py
```diff
@@ -125,6 +125,17 @@ def get_link_filters(self, for_doctype):
...
+	def delink_advance_entries(self, jv):
+		total_allocated_amount = 0
+		for adv in self.advances:
+			if adv.reference_name == jv:
+				adv.reference_type = ''
+				adv.reference_name = ''
+				adv.allocated_amount = 0.0
+				adv.db_update()
+			total_allocated_amount += flt(adv.allocated_amount, adv.precision("allocated_amount"))
+		self.db_set("total_advance", total_allocated_amount)
```
rmehta: will this also affect outstanding? should all totals be recalculated?

rmehta (changes requested):

rmehta (changes requested):
add test case

saurabh6790:
Added

rmehta (approved):

PR #6959, opened 2016-11-16
Group same item in SO print

KanchanChauhan:
screen shot 2016-11-16 at 2 44 01 pm


nabinhait (commented):
Put `group_same_items()` function under accounts_controller.py inside the Class. That way it will be available for both sales and purchase order and you don't have to pass doc.

vladucoju (approved):

vladucoju (approved):

KanchanChauhan:
Changes done

nabinhait (approved):

Issue #6689, opened 2016-10-24
Set the POS to show just product items and hide the services items


dufani1:
please I suggest to add the feather to POS which allows to show / hide the group of products.
Example: 
 Show in POS  just product items and hide the services items to the certain user.
Regards,

mjomane:
yes, this is a must-have

dufani1:
thanks nice work

mjomane:
thank you. feature works smoothly

PR #6969, opened 2016-11-16
User can able to add multiple item group, customer group in the POS profile.

rohitwaghchaure:
#6689 

screen shot 2016-11-17 at 12 39 59 am


erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -129,6 +129,26 @@ def get_items(doc, pos_profile):
 
 	return item_list
 
+def get_items_list(pos_profile):
+	cond = "1=1"
```
rmehta: Add docstrings

rmehta (changes requested):
Add test cases

rohitwaghchaure:
Done

nabinhait (approved):

PR #6971, opened 2016-11-17
Update naming-series.md

ShraddhaRanjane:
Updated User manual for Naming series with custom field


sbkolate (commented):
Need to add YY.MM.DD option as per following discussion
https://discuss.erpnext.com/t/you-can-use-custom-field-in-naming-series-along-with-yy-mm-dd-option/10865/9

nabinhait (approved):

PR #6995, opened 2016-11-18
Added Student Batch Name & Batch Creation Tool in Schools Module.

neilLasrado:


nabinhait (commented):
@neilLasrado 
Why doctype name are such? 
Batch Name? 
Batch Creation Tool?

nabinhait (approved):

PR #6998, opened 2016-11-20
[Fix]Workstation holiday list

### LINKED ISSUE #6977: Workstation switching to "not saved" and forcing a Holiday List
So I open a Workstation to view it. The strange thing is that it changes to "not saved" as if I had made a change but I haven't.

bildschirmfoto 2016-11-17 um 10 40 12

Then the next thing I see it overwrites the Holliday List with what seems to be the standard at the company. Now the standard is "Germany 2016" but if I just rename the workstation for example it correctly remembers it's specific Holiday List. So in this case it was "Romania 2016" which seems to have kept the information despite incorrectly trying to force "Germany 2016" in the previous view.

bildschirmfoto 2016-11-17 um 10 40 23

However the next time I open it it shows "Germany 2016" again. Thats a bit messy.

bildschirmfoto 2016-11-17 um 10 40 31

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/6977

### DIFF
--- a/erpnext/manufacturing/doctype/workstation/workstation.js
+++ b/erpnext/manufacturing/doctype/workstation/workstation.js
@@ -1,17 +1,19 @@
 // Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 // License: GNU General Public License v3. See license.txt
 
-
-
-//--------- ONLOAD -------------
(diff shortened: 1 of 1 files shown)


nabinhait (commented):
@KanchanChauhan every error fix should go to hotfix branch.

Issue #6560, opened 2016-10-06
Budget against Project


nabinhait:

nabinhait:
- In Budget add a new dropdown field "Budget Against" - Options are "Cost Center", "Project"
- Add another new field for "Project"
- If budget against project, expenses should be validated against budget for the project.

PR #7036, opened 2016-11-22
[Enhancement] Budget Against Project

manassolanki:
Budget Allocation for project and validate before every Journal Entry with linked project
#6560


erpnext/accounts/doctype/budget/budget.py
```diff
@@ -14,20 +14,34 @@ class DuplicateBudgetError(frappe.ValidationError): pass
...
+			self.name = make_autoname(self.project + "/" + self.fiscal_year + "/.###")
+	
 	def validate(self):
-		self.validate_duplicate()
+		if not self.cost_center and not self.project:
+			frappe.throw(_("Budget should be allocated against either Cost Center or Project."))
+		if self.cost_center:
+			self.validate_duplicate("cost_center")
+		if self.project:
+			self.validate_duplicate("project")
```
nabinhait: Call `self.validate_duplicate()` only once and set budget_against inside the function (there is already a if...else)

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -14,20 +14,34 @@ class DuplicateBudgetError(frappe.ValidationError): pass
...
 		self.validate_accounts()
 
-	def validate_duplicate(self):
-		existing_budget = frappe.db.get_value("Budget", {"cost_center": self.cost_center,
+	def validate_duplicate(self, budget_against):
+		budget_against_type = None
```
nabinhait: Variable names should be `budget_against_field` and `budget_against`

```
budget_against_field = frappe.scrub(self.budget_against)
budget_against = self.get(budget_against_field)
```

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -51,56 +65,124 @@ def validate_accounts(self):
...
+				from 
+					`tabBudget` b, `tabBudget Account` ba
+				where
+					b.name=ba.parent and b.fiscal_year=%s and ba.account=%s and b.docstatus=1
+					and exists(select name from `tabProject` where name=b.project)
+			""",(args.fiscal_year, args.account), as_dict=True)
+			validate_budget_records(args, budget_records_against_project)
+
+		if args.cost_center:
+			budget_records_against_cost_center = frappe.db.sql("""
```
nabinhait: duplicate query, you can only pass the additional condition

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -51,56 +65,124 @@ def validate_accounts(self):
...
+			and exists(select name from `tabCost Center`
+				where lft>=%(lft)s and rgt<=%(rgt)s and name=gle.cost_center)
+			and gle.fiscal_year=%(fiscal_year)s
+			and gle.company=%(company)s
+			and gle.docstatus=1
+			{condition}
+	""".format(condition=condition), (args))[0][0])
+
+
+def get_actual_expense_for_project(args):
```
nabinhait: Again duplicate function, can be done within existing function passing appropriate condition

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -51,56 +65,124 @@ def validate_accounts(self):
...
+			if monthly_action in ["Stop", "Warn"]:
+				budget_amount = get_accumulated_monthly_budget(budget.monthly_distribution,
+					args.posting_date, args.fiscal_year, budget.budget_amount)
+
+				args["month_end_date"] = get_last_day(args.posting_date)
+				
+				if budget.cost_center:
+					compare_expense_with_budget(args, "Cost Center", budget.cost_center,
+					budget_amount, _("Accumulated Monthly"), monthly_action)
+				elif budget.project:
```
nabinhait: if...elf can be eliminated

nabinhait (changes requested):

nabinhait (changes requested):
Add test cases for budget against project

erpnext/accounts/doctype/budget/budget.py
manassolanki: @nabinhait

Issue #7034, opened 2016-11-22
Chart of Accounts - Root Account should be set automatically


umairsy:
I was creating new Bank Account in the Chart of Accounts. In the dialog box, it should not show Root Account field. Based on the Root Type of a Parent Account, Root Type for the new Account should be set automatically.

![screen shot 2016-11-22 at 5 18 36 pm](https://cloud.githubusercontent.com/assets/910238/20522797/242988c2-b0d8-11e6-95f9-3a2fd7c6cacf.png)

umairsy:
When is it expected to release?

PR #7045, opened 2016-11-23
[fix] Root type fix

saurabh6790:
- remove root type selection while creating the new node, https://github.com/frappe/erpnext/issues/7034
 - hooks call to initialize payment gateway


erpnext/accounts/doctype/account/account_tree.js
```diff
@@ -26,8 +26,6 @@ frappe.treeview_settings["Account"] = {
 			description: __("Name of new Account. Note: Please don't create accounts for Customers and Suppliers")},
 		{fieldtype:'Check', fieldname:'is_group', label:__('Is Group'),
 			description: __('Further accounts can be made under Groups, but entries can be made against non-Groups')},
-		{fieldtype:'Select', fieldname:'root_type', label:__('Root Type'),
```
nabinhait: @saurabh6790 you should not delete this field, it should be shown conditionally, only for root account.

PR #7094, opened 2016-11-28
[Enhancement] Budget Against Project #7036

manassolanki:
Budget Allocation for project and validate before every Journal Entry with linked project
#6560


erpnext/accounts/doctype/budget/budget.py
```diff
@@ -14,20 +14,24 @@ class DuplicateBudgetError(frappe.ValidationError): pass
 
 class Budget(Document):
 	def autoname(self):
-		self.name = make_autoname(self.cost_center + "/" + self.fiscal_year + "/.###")
+		self.name = make_autoname(self.get(frappe.scrub(self.budget_against)) + "/" + self.fiscal_year + "/.###")
```
rmehta: use `-` as a separator for naming series instead of `/`

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -14,20 +14,24 @@ class DuplicateBudgetError(frappe.ValidationError): pass
...
 
 	def validate(self):
+		if not self.get(frappe.scrub(self.budget_against)):
+			frappe.throw(_("{0} is mandatory").format(self.budget_against))
 		self.validate_duplicate()
```
rmehta: Why not make the field mandatory?

rmehta (changes requested):

erpnext/accounts/doctype/budget/budget.py
manassolanki: This is because we have to make budget either against project or cost center mandatory, not for both of them.

erpnext/accounts/doctype/budget/budget.py
manassolanki: Actually there is  -  already in cost center and fiscal year naming so using that again make it very hard to understand the name.
Ex.  Mumbai - MPC/2016-2017/001

nabinhait (approved):

PR #7103, opened 2016-11-30
[Fix] Disallow same items to be fetched multiple times in sales invoice

### LINKED ISSUE #5016: If invoice is in draft mode duplicate delivery note should not be displayed in get item from delivery note
If invoice is in draft mode duplicate delivery note should not be displayed in get item from delivery note

Thanks
Mithilesh Kumar

bhupennewalkar1337:
Fixed #5016


vladucoju:
Totally agree with you.

nabinhait (approved):

PR #7107, opened 2016-12-01
calc timesheet dates on validate not just submit

RobertSchouten:


nabinhait:
@rohitwaghchaure Can you review this PR?

erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -28,6 +28,7 @@ def validate(self):
 		self.update_cost()
 		self.calculate_total_amounts()
 		self.calculate_percentage_billed()
+		self.set_dates()
```
rohitwaghchaure: Adding function under validate will execute on every save. Start date and end date is used in salary slip while making it from timesheet

RobertSchouten: the idea behind this is to keep a consitancy seems wierd to not have on save only submit and i wanted to use for a customisation but i could easily add another way

rohitwaghchaure:
@nabinhait 
Reviewed, please merge

nabinhait (approved):

PR #7118, opened 2016-12-01
[Fix] Incorrect annual billing amount, unpaid amount in customer

rohitwaghchaure:


vladucoju:
This fix is important!

erpnext/selling/doctype/customer/customer.py
```diff
@@ -24,13 +24,18 @@ def onload(self):
 
 	def load_dashboard_info(self):
 		billing_this_year = frappe.db.sql("""
-			select sum(debit_in_account_currency) - sum(credit_in_account_currency)
-			from `tabGL Entry`
-			where voucher_type='Sales Invoice' and party_type = 'Customer'
-				and party=%s and fiscal_year = %s""",
+			select
+				sum(`tabGL Entry`.debit_in_account_currency / `tabSales Invoice`.conversion_rate) -
```
nabinhait: Why is it needed? Instead, we can use debit and credit column of gl entry.

erpnext/selling/doctype/customer/customer.py
```diff
@@ -24,13 +24,18 @@ def onload(self):
...
+			from
+				`tabGL Entry`, `tabSales Invoice`
+			where
+				`tabGL Entry`.voucher_type='Sales Invoice' and `tabGL Entry`.party_type = 'Customer'
+				and `tabGL Entry`.party=%s and `tabGL Entry`.fiscal_year = %s
+				and `tabGL Entry`.against_voucher = `tabSales Invoice`.name""",
 			(self.name, frappe.db.get_default("fiscal_year")))
 
-		total_unpaid = frappe.db.sql("""select sum(outstanding_amount)
+		total_unpaid = frappe.db.sql("""select sum((outstanding_amount / conversion_rate))
```
nabinhait: It will return wrong result receivable_account is in base currency, but invoice currency is in another currency.
Instead, we should calculate outstanding based on debit and credit amount from gl entry.

nabinhait (changes requested):

PR #7139, opened 2016-12-05
[Fix] Currecny symbol for customer's total annual billing, total unpaid amount

rohitwaghchaure:


erpnext/selling/doctype/customer/customer.py
```diff
@@ -24,7 +25,8 @@ def onload(self):
 
 	def load_dashboard_info(self):
 		billing_this_year = frappe.db.sql("""
-			select sum(debit_in_account_currency) - sum(credit_in_account_currency)
+			select sum(debit_in_account_currency) - sum(credit_in_account_currency), 
+			(select account_currency from tabAccount where name = account)
```
nabinhait: Use `account_currency` field from GL Entry itself.

rohitwaghchaure: Done

nabinhait (approved):

PR #7181, opened 2016-12-12
Added "Enabled" to Sales Person

### LINKED ISSUE #6876: Feature Request: Disable "Sales Person"
Currently the Sales Person list is constantly updated with new employees. To be able to "Disable" the sales person after they leave the company so that their name would no longer appear when searching.

KanchanChauhan:
- Whenever Employee leaves, Sales Person's "enabled" unchecked.

Fixed https://github.com/frappe/erpnext/issues/6876


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.js
```diff
@@ -120,3 +120,9 @@ cur_frm.cscript.generate_schedule = function(doc, cdt, cdn) {
 cur_frm.fields_dict.customer.get_query = function(doc,cdt,cdn) {
 	return { query: "erpnext.controllers.queries.customer_query" }
 }
+
+cur_frm.fields_dict['items'].grid.get_field('sales_person').get_query = function(doc, cdt, cdn) {
+	return{
+		filters: [['Sales Person', 'is_active', '=', 1]]
+	}
```
rmehta: If you use the fieldnames `enabled` or `disabled`, then the framework automatically does make these fields selectable, so you don't need to add `get_query` methods

Maybe use `enabled` and keep it checked by default

rmehta (changes requested):

nabinhait (approved):

PR #7201, opened 2016-12-13
validate negative stock against accounting entries

RobertSchouten:
prevents missing gl enties on sales invoices and delivery notes
follow up on incorrect removed validation from #6335
fix's for issue #5945

the actual use case of missing gle is, if you deliver/issue an item at the very first time, without any inward entry for that item, in that case system does not found any valuation rate for that item and causing missing gl entry. But if you post any inward entry, and after that item goes into negative inventory, system still maintains the same valuation rate for that item, hence it does not cause any missing gl entry.

So, might be right solution is, if negative inventory is enabled, we should block outward entry, until we make an inward entry first with proper valuation rate.

@nabinhait i think the fix is good though not sure if order should b changed for performance

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -36,6 +36,7 @@ class StockEntry(StockController):
 			self.pro_doc = frappe.get_doc('Production Order', self.production_order)
 
 		self.validate_posting_time()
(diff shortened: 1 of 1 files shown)


nabinhait:
Made a separate PR on this (https://github.com/frappe/erpnext/pull/7234).

I have blocked the outgoing transaction itself, if there is no incoming entries before delivery, for the item and warehouse.

PR #7204, opened 2016-12-13
Revert "email cc test fixes for frappe/frappe#2441"

nabinhait:
Reverts frappe/erpnext#7199

PR #7232, opened 2016-12-15
Revert "Revert "email cc test fixes for frappe/frappe#2441""

nabinhait:
Reverts frappe/erpnext#7204


nabinhait:
@RobertSchouten Reverted, thanks for reminding 👍

PR #7234, opened 2016-12-15
Validate negative stock on delivery while missing valuation rate for booking expenses

nabinhait:


RobertSchouten:
@nabinhait validation is only required on
if frappe.defaults.get_global_default('allow_negative_stock') and frappe.defaults.get_global_default('auto_accounting_for_stock')

nabinhait:
@RobertSchouten This code will only be executed if negative inventory and perpetual inventory enabled.

RobertSchouten:
@nabinhait fails automated tests please address

RobertSchouten:
@nabinhait this also blocks stock entries from being amended before invoicing

nabinhait:
Fixing test cases.
Is the Stock Entry for outward transactions like Material Issue? In that case, it should block that transaction as well.

RobertSchouten:
@nabinhait the stock entry before the delivery note or sales invoice that is trying to b amended.
should currently it means that invoices also have to b cancelled to amend stock entries

PR #7279, opened 2016-12-20
Create User button in Employee if no User ID selected

### LINKED ISSUE #7173: Automate Employee creation with User creation or vice versa
Currently users and employees need to be created separately in the system even when they are the same person. Please allow employee creation with user creation or vice versa so that the work is not duplicated. This will also negate the requirement to have duplicate fields like birth dates, phone, etc. from both screens.

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/7173


erpnext/hr/doctype/employee/employee.json
```diff
@@ -2234,7 +2263,7 @@
  ], 
  "hide_heading": 0, 
  "hide_toolbar": 0, 
- "icon": "fa fa-user", 
+ "icon": "icon-user",
```
nabinhait: why icon has been overwritten? Seems like, your local sites has not migrated.

erpnext/hr/doctype/employee/employee.py
```diff
@@ -250,3 +257,20 @@ def deactivate_sales_person(status = None, employee = None):
...
+			"field_map": {
+				"employee_name": "first_name",
+				"prefered_email": "email",
+				"gender": "gender",
+				"date_of_birth": "birth_date",
+				"cell_number": "phone",
+				"bio": "bio"
+			}
+		}}, user)
+	user.save()
```
nabinhait: user should be enabled?

Also, why use mapper for this? May be `new_doc` will be sufficient and cleaner in this case.

nabinhait (changes requested):

erpnext/hr/doctype/employee/employee.json
KanchanChauhan: This is some issue, every time you change something in doctype, it changes back to icon-user.

erpnext/hr/doctype/employee/employee.py
KanchanChauhan: Done

nabinhait (approved):

PR #7280, opened 2016-12-20
Added mark as present to student leave application

neilLasrado:


nabinhait (changes requested):
Leave Application should be based on from date and to date

nabinhait (approved):

PR #7284, opened 2016-12-21
Report : Batch Item Expiry Status

revant:
- Select From Date and To Date
- Lists Item Batches about to expire or already expired
- Expiry Status column shows number of days in which item/batch will expire

batch_item_expiry_status


erpnext/stock/report/batch_item_expiry_status/batch_item_expiry_status.py
```diff
@@ -0,0 +1,94 @@
...
+		expiry_date_unicode = frappe.db.get_value('Batch', d.batch_no, 'expiry_date')
+		qty_dict.expires_on = expiry_date_unicode
+
+		exp_date = frappe.utils.data.getdate(expiry_date_unicode)
+		qty_dict.expires_on = exp_date
+
+		expires_in_days = (exp_date - frappe.utils.datetime.date.today()).days
+
+		if expires_in_days > 0:
+			qty_dict.expiry_status = "Expires in {} day(s)".format(expires_in_days)
```
rmehta: translatable ? `_(""Expires in {} day(s)")`

rmehta (changes requested):

erpnext/stock/report/batch_item_expiry_status/batch_item_expiry_status.py
revant: columnrename

How about renaming the column to (already translatable) "Expiry (In Days)" and data as plain int?

rmehta: Cool that works better!

Btw there is a better way using `get_columns` using JSON 

https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/report/profitability_analysis/profitability_analysis.py#L121

nabinhait (approved):

mdwala:
Isn't available quantity per batch is most required field for all?

PR #7285, opened 2016-12-21
[Fix] Warehouse is mandatory for packing materials on the sales invoice

rohitwaghchaure:


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -403,9 +403,10 @@ def validate_item_code(self):
 	def validate_warehouse(self):
 		super(SalesInvoice, self).validate_warehouse()
 
-		for d in self.get('items'):
-			if not d.warehouse and frappe.db.get_value("Item", d.item_code, "is_stock_item"):
-				frappe.throw(_("Warehouse required at Row No {0}").format(d.idx))
+		for d in self.get_item_list():
+			if frappe.db.get_value("Item", d.item_code, "is_stock_item") == 1:
```
nabinhait: Check `not d.warehouse` condition first, to avoid the query for all the rows.

rohitwaghchaure: Done

nabinhait (approved):

PR #7287, opened 2016-12-21
[Fix] POS issue for serialized items

rohitwaghchaure:


erpnext/accounts/page/pos/pos.js
```diff
@@ -481,18 +481,24 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
...
+			me.items = me.get_items(item_code);
+			me.validate_serial_no()
 			me.update_qty_rate_against_item_code(item_code, "qty", $(this).val());
 		})
 
 		$(this.wrapper).find("[data-action='increase-qty']").on("click", function(){
 			var item_code = $(this).parents(".pos-bill-item").attr("data-item-code");
 			var qty = flt($(this).parents(".pos-bill-item").find('.pos-item-qty').val()) + 1;
+			me.items = me.get_items(item_code);
+			me.validate_serial_no()
```
nabinhait: The above 4 lines are repeated for qty change, increase_qty, decrease_qty. Put it in a function.

rohitwaghchaure: Done

nabinhait (approved):

PR #7329, opened 2016-12-26
[SMS] Fix success_list + utf-8 support + optional success message

ghost:
- fix success_list if http status == 200 (otherwise no more SMS Log)
- utf-8 support
- optional success message, default behavior is not changed (msgprint is called)


nabinhait (approved):

nabinhait:
merged manually

PR #7409, opened 2017-01-09
proposed usage of daterange picker

RobertSchouten:
how the daterange control was designed to work and now does not function


nabinhait (approved):

RobertSchouten:
@nabinhait this was a how it should b used for the fixing of the daterange picker and is currently non functional https://github.com/frappe/frappe/issues/2538
not sure that it should b merged yet

nabinhait:
@RobertSchouten Thanks, reverted the PR.

PR #7438, opened 2017-01-10
Fixes counting of in stock parent assemblies and formatting of csv

bcornwellmott:
Currently, the PPT includes the children of in stock parents when generating material requests, despite the fact that you don't need those parts. This fixes that issue.

Also, a minor change to the way items have no stock entries are shown in the csv. It used to be that the Totals row was not displayed. Currently, a Totals row with 0,0,0 for stock quantities is displayed (easier to sort through with more consistent formatting).


bcornwellmott:
@rmehta, is there any way to move this forward?

nabinhait:
merged manually

PR #7455, opened 2017-01-12
Adds button to find MRs linked to Supplier

bcornwellmott:
Creates a button on the Make drop down menu of the Request for Quotation doc which pops up a dialog. The dialog asks the user to pick a supplier. After clicking the OK button with a Supplier picked, the system searches through all the Open material requests determines which material request items have an Item Supplier (supplier part number) associated with Item. It then adds all of them into the current Request for Quotation.

This feature is badly needed, as it is extremely tedious to manually have to search through every Material Request in order to prepare Request for Quotations. Purchase Orders have the feature which allows Material Requests to be pulled based on the default supplier, but at the RFQ stage, a default supplier hasn't been set. This way, you can add possible suppliers to each Item, then pull the correct items into an RFQ with just a few clicks.

### DIFF
--- a/erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
+++ b/erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
@@ -134,6 +134,46 @@ erpnext.buying.RequestforQuotationController = erpnext.buying.BuyingController.e
 						}
 					})
 				}, __("Get items from"));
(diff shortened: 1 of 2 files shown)


bcornwellmott:
@rmehta, is there any way to move this forward?

nabinhait:
@bcornwellmott Merged this after cleaning up some code. Please send the PR from your own fork, instead of create a branch on frappe.

PR #7467, opened 2017-01-12
[minor] Show Actual Item quantity in warehouse set query

suyashphadtare:
Currently Applicable For Sales Order/Delivery Note Item child table warehouse link.

image

Note - If there is no entry for specified item in BIN doctype then warehouse link will not show any warehouses.


nabinhait:
We should show all warehouses in the Sales Order irrespective of it has qty available or not. Because, after getting SO, we can make or purchase those items and deliver later.

nabinhait (changes requested):
We should show all warehouses in the Sales Order irrespective of it has qty available or not. Because, after getting SO, we can make or purchase those items and deliver later.

suyashphadtare:
I have made changes so all warehouses will be shown in Sales Order.
For Delivery Note, iI have kept same sql query so all warehouses will show in drop-down.

erpnext/controllers/queries.py
```diff
@@ -348,3 +349,46 @@ def get_expense_account(doctype, txt, searchfield, start, page_len, filters):
...
+	conditions, bin_conditions = [], []
+	filter_dict = get_doctype_wise_filters(filters)
+
+	sub_query = """ select round(`tabBin`.actual_qty, 2) from `tabBin`
+					where `tabBin`.warehouse = `tabWarehouse`.name
+					{bin_conditions} """.format(
+					bin_conditions=get_filters_cond(doctype, filter_dict.get("Bin"), bin_conditions))
+
+	response = frappe.db.sql("""select `tabWarehouse`.name,
+								CONCAT_WS(" : ", "Actual Qty", ifnull( ({sub_query}), 0) ) as actual_qty
```
rmehta: so much whitespace on the left causes un-necessary scrolling

suyashphadtare:
I have removed unnecessary whitespace.

nabinhait:
@suyashphadtare Can you please rebase your commits? There are lot of unwanted commits included into this PR.

suyashphadtare:
@nabinhait I have rebased my commits.

nabinhait (approved):

PR #7473, opened 2017-01-13
Patch to add Start and End Date in Salary Slips

KanchanChauhan:
Salary slips generated before April does not have Start and End date hence does not show up in Monthly Salary Register Report.

### DIFF
--- a/erpnext/patches/v7_2/update_salary_slips.py
+++ b/erpnext/patches/v7_2/update_salary_slips.py
@@ -1,0 +1,16 @@
+import frappe
+from erpnext.hr.doctype.process_payroll.process_payroll import get_month_details
+
+def execute():
+	salary_slips = frappe.db.sql("""select fiscal_year, month, name from `tabSalary Slip` 
+				where (month is not null and month != '') 
+				and (fiscal_year is not null and fiscal_year != '') and
+				(start_date is null  or start_date = '') and 
+				(end_date is null  or end_date = '') and docstatus != 2""")
+
+	for salary_slip in salary_slips:
+		get_start_end_date = get_month_details(salary_slip.fiscal_year, salary_slip.month)
+		start_date = get_start_end_date['month_start_date']
+		end_date = get_start_end_date['month_end_date']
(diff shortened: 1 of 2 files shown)


erpnext/patches/v7_2/update_salary_slips.py
```diff
@@ -0,0 +1,17 @@
+import frappe
+from erpnext.hr.doctype.process_payroll.process_payroll import get_month_details
+
+def execute():
+	salary_slips = frappe.db.sql("""select fiscal_year, month, name from `tabSalary Slip` 
+				where (month is not null or month != '') 
+				and (fiscal_year is not null or fiscal_year != '') and
+				(start_date is null  or start_date = '') and 
+				(end_date is null  or end_date = '') and docstatus != 2 and 
+				posting_date between '2016-01-01' and '2016-06-30'""")
```
nabinhait: Why this date conditions? There can be such salary slips beyond this period, depends on deployment date.

erpnext/patches/v7_2/update_salary_slips.py
```diff
@@ -0,0 +1,17 @@
+import frappe
+from erpnext.hr.doctype.process_payroll.process_payroll import get_month_details
+
+def execute():
+	salary_slips = frappe.db.sql("""select fiscal_year, month, name from `tabSalary Slip` 
+				where (month is not null or month != '')
```
nabinhait: this should be and condition

erpnext/patches/v7_2/update_salary_slips.py
```diff
@@ -0,0 +1,17 @@
+import frappe
+from erpnext.hr.doctype.process_payroll.process_payroll import get_month_details
+
+def execute():
+	salary_slips = frappe.db.sql("""select fiscal_year, month, name from `tabSalary Slip` 
+				where (month is not null or month != '') 
+				and (fiscal_year is not null or fiscal_year != '') and
```
nabinhait: this one too should be `and` condition

nabinhait (changes requested):

nabinhait (approved):

PR #7504, opened 2017-01-16
BoM Report

umaepoch:
Hi,

This is my second attempt at creating a Pull request. Hope I am doing it right this time. :-) I want you to review the BoM Item Report, which shows all the BoMs with corresponding stock in all the warehouses. The BoM can be shown with either the Items or the Exploded Items, depending upon the selection.

Please review and upload, if found suitable.

Thanks
Uma


nabinhait:
@umaepoch Thanks for the PR, but you have unnecessarily complicated the report and its logic. Also logics are correct to compute balance qty. 

@bcornwellmott has proposed a pull request for exactly same purpose, which is under review. Will try to merge that after making requested changes.

Closing your pull request.

PR #7509, opened 2017-01-16
Add Full Name as option in HR Setting Employee creation option

kickapoo:
Its useful for small companies that use ERPnext to have their  Employee Full Name in any Link reference across the application.


erpnext/hr/doctype/hr_settings/hr_settings.json
```diff
@@ -295,8 +305,8 @@
  "issingle": 1, 
  "istable": 0, 
  "max_attachments": 0, 
- "modified": "2016-12-21 18:52:03.633251", 
- "modified_by": "Administrator", 
+ "modified": "2017-01-16 14:01:31.183485", 
+ "modified_by": "anastasiadis.st00@gmail.com",
```
revant: Save the doctype as 'Administrator`

nabinhait (approved):

PR #7537, opened 2017-01-19
[Fix] Partial Purchase return, Fixes #7390

### LINKED ISSUE #7390: Partial Purchase return error
Hello,

I tried to return some items from a Purchase Invoice, but I can't return it only if I leave the same quantity, otherwise it appears “Item A has already been returned”.

e.g in purchase invoice I have item A-10 pieces and item B-5 pieces. In the Purchase return I want to have only item A with -3 pieces which will be returned to my supplier and the rest of 7 pieces will remain in my stock.

Is this possible with ERPNext?

emadshaaban92:
In a Purchase Receipt changing an item accepted qty will trigger changing rejected qty to complement received qty, this is the normal behavior but it's not the expected behavior when dealing with Return Purchase Receipt.

It makes partial purchase return looks like this (trying to return 1 from 5)

before

rejected qty in a purchase return doesn't make sense and causes errors when validating quantity 

It should look like this instead

after

### DIFF
--- a/erpnext/buying/doctype/purchase_common/purchase_common.js
+++ b/erpnext/buying/doctype/purchase_common/purchase_common.js
@@ -148,7 +148,12 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 			frappe.model.round_floats_in(item, ["qty", "received_qty"]);
 			item.rejected_qty = flt(item.received_qty - item.qty, precision("rejected_qty", item));
 		}
-
+		
+		if(doc.is_return){
+			item.received_qty = item.qty;
+			item.rejected_qty = 0;
+		}
+		
 		this._super(doc, cdt, cdn);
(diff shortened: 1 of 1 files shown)


erpnext/buying/doctype/purchase_common/purchase_common.js
```diff
@@ -148,7 +148,12 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 			frappe.model.round_floats_in(item, ["qty", "received_qty"]);
 			item.rejected_qty = flt(item.received_qty - item.qty, precision("rejected_qty", item));
 		}
-
+		
+		if(doc.is_return){
+			item.received_qty = item.qty;
+			item.rejected_qty = 0;
```
rohitwaghchaure: You can not make it as Zero, user can also return the quantity from the rejected warehouse

PR #7565, opened 2017-01-23
[Fix] Permissions to view Leave Balance Report

KanchanChauhan:
https://discuss.erpnext.com/t/leave-balance-report/19610

### DIFF
--- a/erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
+++ b/erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
@@ -30,28 +30,31 @@ def get_columns(leave_types):
 	return columns
 	
 def get_data(filters, leave_types):
-
+	user = frappe.session.user
 	allocation_records_based_on_to_date = get_leave_allocation_records(filters.to_date)
 
 	active_employees = frappe.get_all("Employee", 
 		filters = { "status": "Active", "company": filters.company}, 
-		fields = ["name", "employee_name", "department"])
+		fields = ["name", "employee_name", "department", "user_id"])
 	
 	data = []
 	for employee in active_employees:
-		row = [employee.name, employee.employee_name, employee.department]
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -211,7 +210,7 @@ def validate_leave_approver(self):
 			frappe.throw(_("{0} ({1}) must have role 'Leave Approver'")\
 				.format(get_fullname(self.leave_approver), self.leave_approver), InvalidLeaveApproverError)
 
-		elif self.docstatus==1 and len(leave_approvers) and self.leave_approver != frappe.session.user:
+		elif self.docstatus==1 and (len(leave_approvers) or self.leave_approver) and self.leave_approver != frappe.session.user:
```
nabinhait: Why so? May be it should be and condition

erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
```diff
@@ -39,19 +39,22 @@ def get_data(filters, leave_types):
...
 	for employee in active_employees:
-		row = [employee.name, employee.employee_name, employee.department]
-
-		for leave_type in leave_types:	
-			# leaves taken
-			leaves_taken = get_approved_leaves_for_period(employee.name, leave_type, 
-				filters.from_date, filters.to_date)
+		leave_approvers = [l.leave_approver for l in frappe.db.sql("""select leave_approver from `tabEmployee Leave Approver` where parent = %s""",
+							(employee.name),as_dict=True)]
+		if (len(leave_approvers) and user in leave_approvers) or (user=="Administrator") or ("HR Manager" in frappe.get_roles(user)):
```
nabinhait: Employee should also be able to see his/her leave balance.

`user in ["Administrator", employee.user_id]`

nabinhait (changes requested):

PR #7576, opened 2017-01-25
[fix] compact print format not longer ignores format builder

RobertSchouten:
part 2 of 2 see frappe for part 1
before
image
after with a print format setup to show more fields
image


RobertSchouten:
part 1
https://github.com/frappe/frappe/pull/2633

nabinhait (approved):

PR #7606, opened 2017-01-30
[Fix] Allow data import tool for cost center

### LINKED ISSUE #7604: Export / Import Cost Center is missing [BUG]
When using the data export tool in ERPNext, the available items for export does not show "Cost Centers". This is a bug.

rohitwaghchaure:
Fixed #7604


nabinhait (changes requested):
@rohitwaghchaure To allow a document in Data Import Tool, `allow_import` property should be checked. `in_create` is to allow creation using new form view.

rohitwaghchaure:
Hi @nabinhait 
Currently allow import is true still user can not able to use cost center in data import tool. For tree view on click of new button system redirect to the tree list view instead of new form view

PR #7614, opened 2017-01-30
[Fix] Account filter in Process Payroll and Salary Component

KanchanChauhan:


erpnext/hr/doctype/salary_component/salary_component.js
```diff
@@ -1,8 +1,13 @@
 // Copyright (c) 2016, Frappe Technologies Pvt. Ltd. and contributors
 // For license information, please see license.txt
 
-frappe.ui.form.on('Salary Component', {
-	refresh: function(frm) {
-
+cur_frm.set_query("default_account", "accounts", function(doc, cdt, cdn) {
```
rmehta: never use `cur_frm` !

add this in `setup` event

rmehta (changes requested):

erpnext/hr/doctype/salary_component/salary_component.js
KanchanChauhan: Done

rmehta (approved):

erpnext/hr/doctype/salary_component/salary_component.js
```diff
@@ -2,7 +2,17 @@
 // For license information, please see license.txt
 
 frappe.ui.form.on('Salary Component', {
-	refresh: function(frm) {
-
+	setup: function(frm) {
+		frm.set_query("default_account", "accounts", function(doc, cdt, cdn) {
+			var d = locals[cdt][cdn];
+			var account_types = ["Bank", "Cash"];
```
nabinhait: Account type linked to salary company is not bank / cash, it should be Expense account (root_type="Expense")

nabinhait (changes requested):

erpnext/hr/doctype/salary_component/salary_component.js
KanchanChauhan: Added both account_type and root_type

nabinhait (approved):

PR #7624, opened 2017-01-31
Code Refractor.

neilLasrado:


KanchanChauhan (commented):
Pulled onto local and checked.

PR #7643, opened 2017-02-01
[Minor]ROI fix in Employee Loan

KanchanChauhan:


nabinhait (commented):
@KanchanChauhan Can you please explain the usecase?

KanchanChauhan:
@nabinhait We are already checking repayment_method before calling the function, so it was repetitive

rmehta (approved):

PR #7654, opened 2017-02-02
[Fix] Set default tax on quotation when it's making from the lead

### LINKED ISSUE #7626: Quotation from Lead - Sales Tax issue
There is an issue when we create a Quotation for Leads:
(1) If we create the quotation from the Lead Form by using "Make - Quotation", then the quotation does not automatically calculate and add the Sales Tax into the Sales Tax table of the Quote
(2) If we create the quotation from the Quotation List form (ie. by selection "NEW"), then it works perfectly (ie. Sales tax is added to the quote automatically to for Leads).

rohitwaghchaure:
Fixed #7626

### DIFF
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -154,6 +154,9 @@ def make_quotation(source_name, target_doc=None):
 			}
 		}}, target_doc)
 	target_doc.quotation_to = "Lead"
+	target_doc.run_method("set_missing_values")
+	target_doc.run_method("set_other_charges")
+	target_doc.run_method("calculate_taxes_and_totals")
 
 	return target_doc
 


erpnext/controllers/selling_controller.py
```diff
@@ -40,6 +40,8 @@ def set_missing_values(self, for_validate=False):
 		super(SellingController, self).set_missing_values(for_validate)
 
 		# set contact and address details for customer, if they are not mentioned
+		# set default taxes
+		self.set_other_charges()
```
nabinhait: @rohitwaghchaure Why so? This code applies to all sales transactions which is not intended.

nabinhait (changes requested):

erpnext/controllers/selling_controller.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #7668, opened 2017-02-04
Production Analytics

bcornwellmott:
This is a start of a report for the Manufacturing module. It shows a report of the production orders, how many were started on time, how many were pending, how many were completed, and how many were left in process.

### DIFF
--- a/erpnext/config/manufacturing.py
+++ b/erpnext/config/manufacturing.py
@@ -112,13 +112,18 @@ def get_data():
 					"is_query_report": True,
 					"name": "Completed Production Orders",
 					"doctype": "Production Order"
+				},{ 
+					"type": "page",
+					"name": "production-analytics",
+					"label": _("Production Analytics"),  
+					"icon": "fa fa-bar-chart",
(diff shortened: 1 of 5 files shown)


bcornwellmott:
I'm having an issue with getting the grid view to work properly. I want the date columns to add cumulatively, but I haven't been able to figure out how to do that (I can't figure out how the columns currently concatenate when not using the "Daily" range setting. Basically, I want each column to include the sum of all previous columns, instead of just showing the ones that are in the current date range.

bcornwellmott:
I've fixed a lot of the issues I was having before, so it works as intended now. 
I think some of the date comparison code is not done correctly.  I just don't know how to do it the best way- suggestions would help.

bcornwellmott:
Here is a screenshot of the report:

image

bcornwellmott:
@nabinhait, can I get some help reviewing how to handle the dates? I was getting errors with the formats before and I fixed it but I don't think I'm not doing it efficiently.

PR #7671, opened 2017-02-06
Created button to link buying items to material requests

bcornwellmott:
I've found several people in my company often create RFQs, supplier quotations and Purchase Orders for items that they know they have to order. These often get done even when a Material Request has already been created, but they don't go through the Material Request to create the RFQ, SQ and POs. This creates a problem where the Material Requests stay open even though the items have been purchased. 

This button (added to all three docs) goes through each item and checks to see if any Material Requests exist, and then assigns those material requests to the items in the given PO (or RFQ, SQ). As long as the user clicks the button, all the links will be created and the Material Requests will be fulfilled.


erpnext/buying/doctype/purchase_common/purchase_common.js
```diff
@@ -221,6 +221,42 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+				frappe.call({	
+					method: "erpnext.buying.doctype.purchase_common.purchase_common.get_linked_material_requests", 
+					args:{
+						items: my_items						
+					}, 
+					callback: function(r) { 
+						for (var i in cur_frm.doc.items) {
+							var qty = cur_frm.doc.items[i].qty;
+							(r.message || []).forEach(function(d) {
+								if (d[0].qty > 0 && qty > 0 && cur_frm.doc.items[i].item_code == d[0].item_code)
```
nabinhait: What if same item appears multiple items?

erpnext/buying/doctype/purchase_common/purchase_common.js
```diff
@@ -221,6 +221,42 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+								{
+									cur_frm.doc.items[i].material_request = d[0].mr_name;
+									cur_frm.doc.items[i].material_request_item = d[0].mr_item;
+									my_qty = Math.min(qty, d[0].qty);
+									qty = qty - my_qty;
+									d[0].qty = d[0].qty  - my_qty;
```
nabinhait: Need to update ordered_qty field.

In the same line, we need to update "Requested Qty" in Bin as well.

nabinhait (changes requested):
As a whole, I don't think it's a good idea to link to material requests in such way. It can break the data integrity.

erpnext/buying/doctype/purchase_common/purchase_common.js
bcornwellmott: I don't think that would be a problem based on the code here. The nested loop goes through each of the items in the item table and then loops through all the material requests. If the same item shows up multiple times, the quantities already assigned to the first instance of that item are accounted for and the MR used has it's quantity already reduced.

erpnext/buying/doctype/purchase_common/purchase_common.js
bcornwellmott: I'm not sure why this needs to be updated here. Doesn't that happen automatically through the status updater when on_submit/on_cancel is called?

bcornwellmott:
I'm not sure I understand how it breaks data integrity. It is just filling in linked document references. Effectively it does the same thing as Get Items from Material Request after the line item has already been added to the item list (albeit in a different way).

nabinhait:
@bcornwellmott 
My all the above comments were based on wrong assumption. I was thinking the button "Link to Material Requests" will be visible after submission of PO. Good job 👍 

Now, I can see only one minor scenario where it can be improved.

MR: There are 2 rows with same item with qty 2 and 1.
screen shot 2017-02-24 at 3 52 44 pm

PO: A single row with qty 3.
screen shot 2017-02-24 at 3 54 24 pm

In this case, it just links the PO to the first row of the MR. And after submission, the ordered qty only updated against the first row, and MR remains open. 
Also, Requested Qty in BIN is reduced by 2 instead of 3.

May be in PO, we will split the row with qty 2 and 1 automatically and link to both rows in MR.

What's your view on this? Apart from this, I think it does a great job.

And you can add a feedback message on the "Link to Material Requests" button.

bcornwellmott:
@nabinhait, I have resolved the issue you identified (splitting rows for multiple MRs) and moved the button to below the table (I think it makes more sense to be there). I also fixed a bug with how I went through the item list.

nabinhait (approved):

PR #7672, opened 2017-02-06
From date, to date moved from Salary Structure to Salary Structure Employee

KanchanChauhan:


nabinhait (changes requested):
@KanchanChauhan 
I think the problem is much bigger. There is no point having `from_date` and `to_date` fields in the parent, as validity of the Salary Structure will be different for different employees. There is some logic in Salary slips as well for getting salary structure, which also fails.
We should move `from_date` and `to_date` fields to the child table for Employee.

erpnext/hr/doctype/salary_structure/salary_structure.py
```diff
@@ -29,12 +27,6 @@ def get_ss_values(self,employee):
 	def validate_amount(self):
 		if flt(self.net_pay) < 0 and self.salary_slip_based_on_timesheet:
 			frappe.throw(_("Net pay cannot be negative"))
-
-	def validate_joining_date(self):
```
nabinhait: keep the function and update it based on from and to date on salary structure employee table

erpnext/patches/v7_2/move_dates_from_salary_structure_to_employee.py
```diff
@@ -0,0 +1,9 @@
+import frappe
+
+def execute():
+	frappe.reload_doc('hr', 'doctype', 'salary_structure_employee')
+	salary_structures = frappe.db.sql("""select name, to_date, from_date from `tabSalary Structure`""", as_dict=True)
+
+	for salary_structure in salary_structures:
+		frappe.db.sql(""" update `tabSalary Structure Employee` set from_date = %s, to_date = %s
+			where parent = %s """, (salary_structure.from_date, salary_structure.to_date or 'null', salary_structure.name))
```
nabinhait: remove `or 'null'`

nabinhait (changes requested):

nabinhait (approved):

PR #7677, opened 2017-02-06
[Fix] Serial No if is_return

### LINKED ISSUE #7508: Wrong Serial Number getting fetched in Sales Return
Hi,

We've encountered an issue where the wrong serial number is being fetched in Sales Return invoices. You can replicate this if you have serials that are created at different times for an item in the same warehouse (In our case, the serials were created via Stock Entries of type 'Material Receipt' at different times). Select a more recent serial in a Sales Invoice and Submit. After that, do a Sales Return. An older serial number for the same item (in the same warehouse) is what automatically appears in the serial number field!

There seems to be some logic that makes the oldest serial number (by Creation Date) in the warehouse automatically appear in the Serial Number field instead of the Serial Number from the original invoice

Kindly help check this

Regards

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/7508

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -123,12 +123,13 @@ def apply_pricing_rule(args):
 	
 def get_serial_no_for_item(args):
 	from erpnext.stock.get_item_details import get_serial_no
+
 	item_details = frappe._dict({
 		"doctype": args.doctype,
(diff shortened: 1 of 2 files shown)


nabinhait (changes requested):
Also need to check in https://github.com/frappe/erpnext/blob/develop/erpnext/stock/get_item_details.py#L77 ?

nabinhait (approved):

PR #7704, opened 2017-02-09
Accrual Payroll System

KanchanChauhan:


nabinhait (changes requested):
Write patch for Arrear and Leave Encashment, if value exists
- Create salary component 
- Add in Earning table

nabinhait (approved):

PR #7712, opened 2017-02-10
[Fix] Valuation method changing issue in the item

rohitwaghchaure:


erpnext/stock/doctype/item/item.py
```diff
@@ -454,7 +454,7 @@ def cant_change(self):
 
 			if vals:
 				for key in to_check:
-					if self.get(key) != vals.get(key):
+					if self.get(key) != vals.get(key) and vals.get(key):
```
nabinhait: `and vals.get(key)` will defeat the purpose for Check fields.

For valuation_method, if existing value in item is blank, the existing value should be replaced from Stock Settings. Then it should be compared.

nabinhait (changes requested):

erpnext/stock/doctype/item/item.py
rohitwaghchaure: Fixed

PR #7739, opened 2017-02-14
[Fix] Payment Entry currency symbol issue

rohitwaghchaure:


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -136,6 +136,10 @@ frappe.ui.form.on('Payment Entry', {
 
 		frm.set_currency_labels(["total_allocated_amount", "unallocated_amount"], party_account_currency);
 
+		currency_field = (frm.doc.payment_type=="Receive") ? "paid_from_account_currency" : "paid_to_account_currency"
```
nabinhait: Use `party_account_currency` as per line 134

nabinhait (changes requested):

PR #7742, opened 2017-02-14
BOM Web-view

pratu16x7:


nabinhait:
@pratu16x7 Please attach a screenshot

pratu16x7:
screencapture-erpnext-dev-8000-bom-bearing-assembly-008-1487217209722

user002:
Some items missing for a complete product design description are:

1.  Item -- need item record with all details
2.  BOM -- need all BOM details shown as would be required to order or fully understand the components of the BOM
3.  work instruction -- for the future should ERPNext expand to have wiki or work instruction functionality; the operations list is helpful but does not provide enough data to build / assemble / construct th item
4.  design / support files -- not required to be stored within ERPNext but there needs to be an ability to list CAD, PDF, text, etc. files that further explain the product design and how to make it -- could just be HTML so not a major item

Other features that would be nice to have:

A.  export of data
B.  linking of data -- if I click on a BOM component, I get taken to the item page for that BOM component (example)

The BOM view itself is nice although "big".  Imagine a BOM with 100 items -- a more condensed version with no photos and more data would be nice.  Otherwise, users are doing a lot (a lot) of scrolling around to see the BOM and everything else.

rmehta:
@user002 there is a free text section in the BOM where you can enter those details. All other Item, BOM details are there, what else do you need?

user002:
@rmehta, putting the work instruction information in the BOM might work yet seems unusual.  Does not allow for good change management (who made the change on what date and what was the change).  Same thing with the design files.  Why not put that information in the Item record?  The Item record is also missing on the page as I stated above.

The above could work and thank you to everyone for making it happen.

Issue #7722, opened 2017-02-12
Unable to Create a fully depreciated asset


addymuliady:
Based on @nabinhait  advice:
https://discuss.erpnext.com/t/create-a-fully-depreciated-asset/20270

I am doing an opening asset entry.
I have an existing asset which already has depreciated fully, i.e. value of 0. It is still deemed an asset in the company.
How do I create this asset?

Purchased in July 2011
Gross Purchase = $5000
Depreciated within 60 mths at a frequency of 1 mth
Value after life = $0
In ERPNext, the Next Depreciation Date cannot be a past date. For the above, the Next Depreciation Date should be Aug 2011. This is a big problem and there is no way to create this type of Asset.

nick9822:
@addymuliady 
I did the same entry and it went through with the said warning. But I was able to create asset. Any other issue in it other than warning.

addymuliady:
Hi @nick9822 

What is Next Depreciation Date that you keyed in?

For my case, it doesn't go through at all. I have to put a future date in Next Depreciation Date.

nick9822:
August 1st, 2011
See this screenshot
![image](https://cloud.githubusercontent.com/assets/10139291/22890124/656ef3cc-f231-11e6-8d0e-911d5376bff2.png)

addymuliady:
Hi @nick9822 

What version of Frappe and ERPNext are you on?

Does the depreciation graph shows the same?

Regards

Mulyadi

nick9822:
@addymuliady 
ERPNext: v7.2.19
Frappe Framework: v7.2.15

> Does the depreciation graph shows the same?
Yes

addymuliady:
Can the warning be taken instead? Confusing.

Thanks @nick9822 for checking!

nick9822:
@addymuliady @nabinhait @rmehta 
I guess next depreciation date should be compared with the purchase date instead current date.

tundebabzy:
Is this issue still persisting? I thought it was fixed in #7751

PR #7751, opened 2017-02-15
Next Depreciation Date is entered as past date

nick9822:
#7722 
Changed validation of next_depreciation date against purchase_date instead of nowdate.

Thanks

### DIFF
--- a/erpnext/accounts/doctype/asset/asset.py
+++ b/erpnext/accounts/doctype/asset/asset.py
@@ -80,8 +80,11 @@ class Asset(Document):
 			if cint(self.number_of_depreciations_booked) > cint(self.total_number_of_depreciations):
 				frappe.throw(_("Number of Depreciations Booked cannot be greater than Total Number of Depreciations"))
 						
+		if self.next_depreciation_date and getdate(self.next_depreciation_date) < getdate(nowdate()):
+			frappe.msgprint(_("Next Depreciation Date is entered as past date"), title=_('Warning'), indicator='red')
+		
 		if self.next_depreciation_date and getdate(self.next_depreciation_date) < getdate(self.purchase_date):
-			frappe.msgprint(_("Next Depreciation Date is entered as past date"))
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/asset/asset.py
```diff
@@ -80,7 +80,7 @@ def validate_asset_values(self):
 			if cint(self.number_of_depreciations_booked) > cint(self.total_number_of_depreciations):
 				frappe.throw(_("Number of Depreciations Booked cannot be greater than Total Number of Depreciations"))
 						
-		if self.next_depreciation_date and getdate(self.next_depreciation_date) < getdate(nowdate()):
+		if self.next_depreciation_date and getdate(self.next_depreciation_date) < getdate(self.purchase_date):
```
nabinhait: This is a warning message when user selects the next depreciation date before today. It does not stop to create /submit the asset.

You can add another validation for purchase date, which should block saving the asset.

nick9822: Yes I got that warning. But why the next depreciation is validated against nowdate? I guess should be validated against purchase date.

Please advice.

nabinhait: If next depreciation date is before nowdate, that means you are creating an existing asset and its depreciation entry is not booked manually via Journal Entry. But we believe for most existing assets, their past depreciation entry should be booked via Journal Entry or in other accounting system (if user migrating to erpnext). There is a huge chance to enter a past date mistakenly. Thats why we added the warning message, maybe we should a "Warning" label before the message.

But we should surely add another validation for purchase date, it definitely can't be before purchase date.

nick9822: Thanks for the insight.
Yes you are right, and i guess there should be some indication in asset after saving that their past depreciation entry are not booked.

So i will add "Warning!" text to it and frappe.throw for self.purchase_date as a seperate validation.

nabinhait (approved):

PR #7762, opened 2017-02-16
[minor] timestamp fixes for heatmap

mbauskar:
Unix timestamp date and javascript timestamp date does not match after changing the system's timezone.

e.g. 
heatmap issue


mbauskar:
need to merge https://github.com/frappe/frappe/pull/2730 first, added get_timestamp method in data.py

nabinhait (approved):

PR #7764, opened 2017-02-16
[Fix] Cleanup supplied items if supply raw material is set as No

rohitwaghchaure:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -359,6 +359,7 @@ frappe.ui.form.on("Purchase Invoice", {
 	},
 
 	is_subcontracted: function(frm) {
+		frm.set_value("supplied_items", [])
```
nabinhait: Don't make `supplied_items` blank here.

Instead, set blank on the server side https://github.com/frappe/erpnext/blob/master/erpnext/controllers/buying_controller.py#L163

nabinhait (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
rohitwaghchaure: Done

nabinhait (approved):

PR #7768, opened 2017-02-17
[minor] auto close opportunity after 7 days

mbauskar:


nabinhait (changes requested):
Can you make the days configurable after which opportunity and support ticket will be closed? It actually depends upon the nature of business.

mbauskar:
@nabinhait,

Added Support Settings doctype and Close Opportunity After Days field in Selling Settings.
Also included the patch to set the Close Opportunity After Days to 15 days.

nabinhait (approved):

PR #7791, opened 2017-02-20
Build Qty considering quantity for which bom is created

nabinhait:


nabinhait:
@bcornwellmott Can you please review the changes, specifically item_list query. Considered quantity of FG item for which BOM is designed.

bcornwellmott (approved):
Looks good to me!

PR #7818, opened 2017-02-22
[Fix] Disappearing Dropdown fixed

### LINKED ISSUE #7429: Product bundle, link item field drop down view cut
Hi there,

i've got an overlap problem on Prroduct bundle, see pic:

schermata 2017-01-09 alle 19 21 15

Can you please have a look? 

Thx

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/7429
screen shot 2017-02-22 at 3 37 09 pm


nabinhait:
@KanchanChauhan Is this a hack? Please fix the root cause in frappe.

KanchanChauhan:
@nabinhait This was the description which was not visible anyways so I had put it in HTML type field. So this can be merged. Will look into root cause as well.

PR #7842, opened 2017-02-24
Add Sales Orders to Project costing and billing section 

superlack:
To provide the same level of "dashboard" financial tracking for projects to work alongside external clients that use PO's


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -167,10 +167,20 @@ def on_cancel(self):
...
 		self.update_reserved_qty()
-
+		self.update_project()
 		self.update_prevdoc_status('cancel')
 
 		frappe.db.set(self, 'status', 'Cancelled')
+		
+	def update_project(self):
+		project_list = []
+		for d in self.items:
```
nabinhait: "Project" field does not exists in Sales Order Item table, it is in the parent table.

nabinhait (changes requested):
Travis failing

superlack:
Will fix, originally tested with "total" fields in SO.  Thanks

superlack:
I fixed the errors (to the best of my knowledge), but Travis is showing errors on the following
`test_overlap_with_half_day (erpnext.hr.doctype.leave_application.test_leave_application.TestLeaveApplication) ... ERROR
test_overlap_with_half_day_not_applicable (erpnext.hr.doctype.leave_application.test_leave_application.TestLeaveApplication) ... ERROR`

erpnext/projects/doctype/project/project.py
```diff
@@ -176,6 +176,13 @@ def update_purchase_costing(self):
 			from `tabPurchase Invoice Item` where project = %s and docstatus=1""", self.name)
 
 		self.total_purchase_cost = total_purchase_cost and total_purchase_cost[0][0] or 0
+		
+	def update_sales_costing(self):
+		total_sales_cost = frappe.db.sql("""select sum(total)
```
nabinhait: Should not it be `grand_total`?

superlack: You're right, I was thinking about how the billing hours don't show taxes, but those are further malleable unlike SO/SINV

superlack:
I hope that's everything

nabinhait (approved):

PR #7889, opened 2017-02-28
[Fix] Sales Return Invoice (Is POS) doesn't show correct payment data

### LINKED ISSUE #7515: Sales Return Invoice (Is POS) doesn't show correct Payment Entry
Hi,

In the case of a Sales Return in which the money is being refunded to the customer, the payment entry doesn't come up as a negative value. Steps to replicate:

1. Submit POS Sales Invoice
2. Click make => Sales Return
3. Check 'Is POS' checkbox

You will observe that the amount in the item table is negative (which is correct) but the payment amount in the Mode of Payment section is not negative! This would lead to wrong postings if the user doesn't take notice and manually make corrections

Regards

rohitwaghchaure:
Fixed #7515

### DIFF
--- a/erpnext/controllers/sales_and_purchase_return.py
+++ b/erpnext/controllers/sales_and_purchase_return.py
@@ -184,7 +184,7 @@ def make_return_doc(doctype, source_name, target_doc=None):
 		doc.return_against = source.name
 		doc.ignore_pricing_rule = 1
 		if doctype == "Sales Invoice":
-			doc.is_pos = 0
+			doc.is_pos = source.is_pos
 
 			# look for Print Heading "Credit Note"
 			if not doc.select_print_heading:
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

erpnext/controllers/sales_and_purchase_return.py
```diff
@@ -198,6 +198,10 @@ def set_missing_values(source, target):
 			if tax.charge_type == "Actual":
 				tax.tax_amount = -1 * tax.tax_amount
 
+		if doc.doctype == 'Sales Invoice' and doc.is_return:
```
nabinhait: Is this applicable for Purchase Invoice as well?

rohitwaghchaure: Did changes please check

nabinhait (approved):

PR #7923, opened 2017-03-04
Adds UOM Conversion to BOM Items

bcornwellmott:
This adds a UOM conversion to the BOM Items document. This allows users to create BOMs with units other than the stock UOM. 

All the conversion happens in the BOM document and the qty and stock_uom fields still hold the same values, so no document that interfaces with the BOM should be affected. We probably need a patch to fill in the unit_qty and uom fields (copy in the qty and stock_uom fields) - I wanted to get this OK'd before adding that.


bcornwellmott:
@nabinhait, I'd really appreciate your review here!

bcornwellmott:
Any comments?

kodybaker:
What's holding this pull request up? It's causing some major issues with BOM costing in our system.

nabinhait:
@bcornwellmott Thanks for the PR and sorry for the delay in reviewing.

There are some minor issues in the calculation like it does not calculate qty on change of conversion factor, there is no need to fetch the conversion factor from Item in the server side, it should respect conversion factor if manually entered. But those are ok, can be fixed very easily.

But in overall, I think it should be implemented in the same way as it is in the other module. What I mean is, instead of "unit_qty" and "qty" fields, the fields should be like "qty" and "stock_qty". In that way, it will be easy to maintain the code in the future without looking into the field-name every time. 
screen shot 2017-03-24 at 1 39 18 pm

If you look in the above screenshot, "Rate" is visible after "Unit Qty", so there is a huge chance that users will enter the rate in "Unit Qty UOM" instead of Stock UOM.So, we have to show Stock Qty column before Rate column.

nabinhait:
Closing this as per https://github.com/frappe/erpnext/wiki/Pull-Request-Guidelines

PR #7952, opened 2017-03-08
Multiple Salary Fixes

### LINKED ISSUE #7924: Deductions not applied to timesheet-based salary slips
This may be user error, but when trying to process payroll for an hourly, part-time employee it appears that automatic deductions are not added to the salary slip.

Steps to reproduce:

- Create "Hourly" salary component, abbreviation H, type "Earning."
- Create "Test" salary structure and assign employee. Base and Variable are both set to $0. 
- On "Test," check "Salary Slip based on Timesheet," set salary component to Hourly and rate to $15.
- On "Test," add a deduction row for Income Tax. Set formula to `H * .05`. 
- Create Timeslip for employee and submit.
- Under Human Resources > Process Payroll, check "Salary Slip based on Timesheet" and select "Create Salary Slip."

Following this process, the Salary Slip will be created. Under Earnings, a row will be created with the Hourly component that is the sum of the hours in the Timesheet(s). However, the deductions row will be empty.

Is this a bug, or am I missing a step in the process?

KanchanChauhan:
- [x] Formula and condition takes changes in Salary Slip into consideration
- [x] Same employee cannot be in multiple active Salary Structure for same time period
- [x] Calculations done based on joining date and relieving date

Fixed https://github.com/frappe/erpnext/issues/7924

### DIFF
--- a/erpnext/hr/doctype/salary_slip/salary_slip.py
+++ b/erpnext/hr/doctype/salary_slip/salary_slip.py
@@ -49,12 +49,12 @@ class SalarySlip(TransactionBase):
 			self._salary_structure_doc = frappe.get_doc('Salary Structure', self.salary_structure)
 
 		data = self.get_data_for_eval()
+
 		for key in ('earnings', 'deductions'):
 			for struct_row in self._salary_structure_doc.get(key):
 				amount = self.eval_condition_and_formula(struct_row, data)
 				if amount:
 					self.update_component_row(struct_row, amount, key)
-
 
 	def update_component_row(self, struct_row, amount, key):
 		component_row = None
@@ -83,6 +83,7 @@ class SalarySlip(TransactionBase):
 					amount = eval(d.formula, None, data)
 			if amount:
 				data[d.abbr] = amount
+
 			return amount
 
 		except NameError as err:
@@ -97,17 +98,20 @@ class SalarySlip(TransactionBase):
 		'''Returns data for evaluating formula'''
 		data = frappe._dict()
 
-		for d in self._salary_structure_doc.employees:
-			if d.employee == self.employee:
-				data.update(frappe.get_doc("Salary Structure Employee", {"employee": self.employee}).as_dict())
+		data.update(frappe.get_doc("Salary Structure Employee", {"employee": self.employee}).as_dict())
 
 		data.update(frappe.get_doc("Employee", self.employee).as_dict())
 		data.update(self.as_dict())
 
 		# set values for components
 		salary_components = frappe.get_all("Salary Component", fields=["salary_component_abbr"])
-		for salary_component in salary_components:
-			data[salary_component.salary_component_abbr] = 0
+		for sc in salary_components:
+			data.setdefault(sc.salary_component_abbr, 0)
+
+		for key in ('earnings', 'deductions'):
+			for d in self.get(key):
+				data[d.abbr] = d.amount
+
 		return data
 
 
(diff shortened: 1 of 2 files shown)


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -173,13 +177,17 @@ def check_sal_struct(self, joining_date, relieving_date):
...
 
 		if self.salary_slip_based_on_timesheet:
 			self.salary_structure = self._salary_structure_doc.name
 			self.hour_rate = self._salary_structure_doc.hour_rate
 			self.total_working_hours = sum([d.working_hours or 0.0 for d in self.timesheets]) or 0.0
-			self.add_earning_for_hourly_wages(self._salary_structure_doc.salary_component)
+			wages_amount = self.hour_rate * self.total_working_hours
+
+			self.add_earning_for_hourly_wages(self, self._salary_structure_doc.salary_component, wages_amount)
+			self.add_earning_for_hourly_wages(self._salary_structure_doc, self._salary_structure_doc.salary_component, wages_amount)
```
nabinhait: This is not required now, because we have updated salary slip earnings and deductions into "data".

erpnext/hr/doctype/salary_structure/salary_structure.py
```diff
@@ -29,12 +29,24 @@ def validate_amount(self):
...
+			joining_date, relieving_date = frappe.db.get_value("Employee", employee.employee,
+				["date_of_joining", "relieving_date"])
+			if employee.from_date and getdate(employee.from_date) < joining_date:
 				frappe.throw(_("From Date {0} for Employee {1} cannot be before employee's joining Date {2}")
-					    .format(e.from_date, e.employee, joining_date))	
+					    .format(employee.from_date, employee.employee, joining_date))
+
+		st_name = frappe.db.sql("""select parent from `tabSalary Structure Employee`
+			where employee=%s and ((%s between from_date and ifnull(to_date, '2199-12-31'))
+			or (%s between from_date and ifnull(to_date, '2199-12-31')))
```
nabinhait: There should be another condition ` or from_date between %(from_date)s and %(to_date)`

nabinhait (changes requested):
Check inline comments

erpnext/hr/doctype/salary_slip/salary_slip.py
KanchanChauhan: Done

erpnext/hr/doctype/salary_structure/salary_structure.py
```diff
@@ -29,12 +29,34 @@ def validate_amount(self):
...
+
+		st_name = frappe.db.sql("""select parent from `tabSalary Structure Employee`
+			where
+			employee=%(employee)s
+			and (
+				(%(from_date)s between from_date and ifnull(to_date, '2199-12-31'))
+				or (%(to_date)s between from_date and ifnull(to_date, '2199-12-31'))
+				or (from_date between %(from_date)s and ifnull(%(to_date)s, '2199-12-31'))
+			)
+			and parent in (select name from `tabSalary Structure`
```
nabinhait: "in" query becomes very slow if there are too many values in sub query.

Instead you can write like this:
`and exists(select name from `tabSalary Structure` where name=`tabSalary Structure Employee`.parent and is_active = 'Yes') and parent != %(salary_struct)s`

nabinhait (changes requested):

Issue #7989, opened 2017-03-12
[Bug] Purchase Receipt - Item Qty error


umairsy:
In the Purchase Receipt, item's Accepted Qty should not be more than Received Qty.

![imagea69a5f](https://cloud.githubusercontent.com/assets/910238/23832560/c4d4f01a-075c-11e7-9bf2-5fcb7b97b407.png)

PR #7995, opened 2017-03-13
[fix] validate negative quantity while creating purchase receipt

saurabh6790:
https://github.com/frappe/erpnext/issues/7989


nabinhait (changes requested):
@saurabh6790 Qty is negative if it is a Return document

saurabh6790:
@nabinhait @rohitwaghchaure Fixed!!

rohitwaghchaure (approved):

PR #8041, opened 2017-03-16
Removed single quotes on the titles/totals export

RicardoJohann:
Removed single quotes on the titles/totals when export to csv


erpnext/accounts/report/cash_flow/cash_flow.py
```diff
@@ -137,7 +137,7 @@ def get_start_date(period, accumulated_values, company):
 def add_total_row_account(out, data, label, period_list, currency):
 	total_row = {
 		"account_name": "'" + _("{0}").format(label) + "'",
-		"account": "'" + _("{0}").format(label) + "'",
+		"account": _("{0}").format(label),
```
nabinhait: It was a hack to remove the hyperlink from such labels.

RicardoJohann: ok i've fixed on tools.js

PR #8046, opened 2017-03-16
[Fix] Cost Center from Sales Invoice to Delivery Note

### LINKED ISSUE #8036: Cost Center Not Getting Pulled in Delivery Note
Hi,

I just noticed that the Cost Center field isn't getting properly updated in the Delivery Note document 

Steps:

1. Set Cost Center, Income Account, and Expense Account at Item Group level
2. Create Sales Invoice for Item (Income Account and Cost Center are correctly pulled)
3. From Sales Invoice, Make => Delivery
4. In Delivery Note document, Expense Account is correctly pulled but Cost Center is not!

Please check

Kind regards,

KanchanChauhan:
It was pulling Company's default Cost Center

Fixed https://github.com/frappe/erpnext/issues/8036

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -810,7 +810,8 @@ def make_delivery_note(source_name, target_doc=None):
 				"parent": "against_sales_invoice",
 				"serial_no": "serial_no",
 				"sales_order": "against_sales_order",
-				"so_detail": "so_detail"
+				"so_detail": "so_detail",
+				"cost_center": "cost_center"
 			},
 			"postprocess": update_item,
 			"condition": lambda doc: doc.delivered_by_supplier!=1


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -810,7 +810,8 @@ def update_item(source_doc, target_doc, source_parent):
 				"parent": "against_sales_invoice",
 				"serial_no": "serial_no",
 				"sales_order": "against_sales_order",
-				"so_detail": "so_detail"
+				"so_detail": "so_detail",
+				"cost_center": "cost_center"
```
nabinhait: It should not set cost center in DN based on sales invoice. In sales invoice, default cost center is set based on default selling cost center in item and it is used for booking income. But in Delivery note, it should use default buying cost center from Item, as it is used to book COGS (expense).

nabinhait (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
KanchanChauhan: Ok. Closing PR.

Issue #8000, opened 2017-03-13
Asset Image is not showing in the Image View


nabinhait:
<img width="1280" alt="screen shot 2017-03-13 at 6 45 02 pm" src="https://cloud.githubusercontent.com/assets/836784/23855866/6d904df6-081d-11e7-821d-28487a894b8c.png">
<img width="1280" alt="screen shot 2017-03-13 at 6 45 09 pm" src="https://cloud.githubusercontent.com/assets/836784/23855869/6fc7c586-081d-11e7-9361-8e059bb12371.png">

netchampfaris:
https://github.com/frappe/erpnext/pull/8075

PR #8075, opened 2017-03-21
[fix] asset image in image view

netchampfaris:


netchampfaris:
https://github.com/frappe/erpnext/issues/8000

nabinhait (approved):

PR #8076, opened 2017-03-21
Show qty in Shopping Cart

KanchanChauhan:
Set in Shopping Cart Settings:
screen shot 2017-03-21 at 4 28 30 pm
Website view:
screen shot 2017-03-21 at 4 28 40 pm


erpnext/shopping_cart/product.py
```diff
@@ -19,7 +19,8 @@ def get_product_info(item_code):
 	qty = 0
 	cart_quotation = _get_cart_quotation()
 	template_item_code = frappe.db.get_value("Item", item_code, "variant_of")
-	in_stock = get_qty_in_stock(item_code, template_item_code)
+	in_stock = get_qty_in_stock(item_code, template_item_code).in_stock
+	stock_qty = get_qty_in_stock(item_code, template_item_code).stock_qty
```
nabinhait: `get_qty_in_stock` function should not be called twice.

nabinhait (changes requested):

erpnext/shopping_cart/product.py
KanchanChauhan: Done

robulik:
Nice. I would suggest binding this info to customer. I mean, I would be able to choose customer, which will be able to see this info in website view.

PR #8079, opened 2017-03-21
[fix] selecting new address in cart webpage unchecks previously checked

robulik:
With more than one address in shipping or billing addresses there was an issue:
On checking other address the previously checked address didn't uncheck. Leaving user in doubt whether the request was succesfull, because it left both new and old address checked.
Now, this is a fix for it. It is not probably nice code, but it works.


erpnext/templates/includes/cart.js
```diff
@@ -24,6 +24,16 @@ $.extend(shopping_cart, {
 			if($(this).prop("checked")) {
 				var me = this;
 
+				// freeze ui and wait for response:
+				frappe.freeze();
```
mbauskar: @robulik,

instead of using frappe.freeze() explicitly you can pass <code>freeze: true</code> in frappe.call it will handle the freeze and unfreeze event

robulik: Lot cleaner. Thank you for pointing this out.

mbauskar (approved):

nabinhait (approved):

PR #8081, opened 2017-03-21
Update global-defaults.md

grynn-in:
To improve, i have identified a few question that needs to be answered. I dont know the answers, but as I do, I will answer them.

### DIFF
--- a/erpnext/docs/user/manual/en/setting-up/settings/global-defaults.md
+++ b/erpnext/docs/user/manual/en/setting-up/settings/global-defaults.md
@@ -9,3 +9,8 @@ Whenever a new document is created, these values will be set as default.
 <img class="screenshot" alt="Global Defaults" src="{{docs_base_url}}/assets/img/setup/settings/global-defaults.png">
 
 {next}
+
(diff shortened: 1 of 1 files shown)


nabinhait:
This should not be a pull request. There are no meanings of those questions in the user manual, if we don't provide the answers.
Please ask those questions on discuss.erpnext.com. Closing this PR.

grynn-in:
Hi Nabi, 

How is it possible to improve a page if one cannot describe what is missing on the page? For example, if we have a bug, you can fix it or create an issue. When it gets fixed you pull it in the code base. But When an answer is given it never makes it to the documentation. 

You need to see it from a reader's perspective & see what questions pops up in the minds when the read it.  Maybe soon someone has an answer....? Maybe I will figure it out myself. And it will  help me re-visit.  Let me know if you have another idea. 

In any case, i will probably fork the docs and start annotating.  Maybe that is a better idea.

Thanks
Deepak

nabinhait:
In that case, create a github issue to update documentation with the questions.

Issue #8065, opened 2017-03-20
Allow to specify Income account in Item for each Company


ccfiel:
For now ERPNext does not allow to specify Income account in Item for each Company. You can only do this in the first company but the succeeding company you can only used one income account to all items. See discussion here https://discuss.erpnext.com/t/item-income-account-per-company/

ccfiel:
Hello @nabinhait @rmehta :)

I have created a pull request for this. It just a simple solution. :) hope this will be in develop. Thanks!

Chris

kennethsequeira:
Solved by Item Defaults section in v11 and onwards:

![image](https://user-images.githubusercontent.com/33246109/58045583-7c0fd500-7b60-11e9-9de0-60700ff1b9b9.png)

![image](https://user-images.githubusercontent.com/33246109/58045609-892cc400-7b60-11e9-9ec1-ca77e46b3cb6.png)

PR #8083, opened 2017-03-22
Allow to specify Income account in Item for each Company

ccfiel:
for simplicity if the income account specify in the transaction is not the same company in the item income account set in Item Doctype. It will try to find the same account name and account type in the Account Doctype with the same company of the transaction. If it exist this will be used as the income account.

https://github.com/frappe/erpnext/issues/8065

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -191,8 +191,9 @@ def get_basic_details(args, item):
 
 def get_default_income_account(args, item):
 	item_account = None
-	account = frappe.get_doc("Account", item.income_account)
-	if account.company != args.company:
+        if item.income_account:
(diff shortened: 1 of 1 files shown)


nabinhait:
The pull request solves it's purpose, but same should be implemented for default expense account and cost centres. I will love to implement it in more explicit way. May be by adding a common child table in Item for all the default accounts and cost centers, which will have following columns:
- Company
- Default Income Account
- Default Expense Account
- Default Selling Cost Center
- Default Buying Cost Center

What do you say?

ccfiel:
@nabinhait yes that would be a better solution :) We can the Child DocType "Item Default Account".

nabinhait:
@ccfiel Closing this PR

PR #8103, opened 2017-03-22
Localdev

dgsol:
DocType Item --> Custom Field is_product_bundle added
Validation done for product Bundle cannot be an asset or a stock Item


dgsol (commented):
OK

PR #8126, opened 2017-03-23
Add link field Package Code (fixes #8094)

### LINKED ISSUE #8094: Add Device Package Code to Item
- [ ] Add doctype Device Package Code with "Code" and "Description" as data fields
- [ ] Add the Device Package Code as link field to item

kickapoo:


kickapoo:
@nabinhait  From the fail on the Travis integration (and as far as I can understand) the test fail to something else. 
Can you please check

nabinhait:
@kickapoo Yes, I have already pushed a fix for it.

kickapoo:
@nabinhait So I have to rebase for the pull ?

nabinhait:
Yes I think.

nabinhait:
By the way, can we call it "Package Code" only? It should be more generic in nature, can be used for non-device items.

kickapoo:
@dottenbr  Please check the namespace issue @nabinhait  states on the above message

dottenbr:
I am fine with Package Code or Device Package Code ... both are fine with me

dottenbr:

erpnext/stock/doctype/package_code/package_code.json
```diff
@@ -0,0 +1,122 @@
...
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "if_owner": 0,
+   "import": 0,
+   "permlevel": 0,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "System Manager",
```
nabinhait: Instead of System Manager, the permission should be applied for "Stock User" and "Stock Manager"

nabinhait (changes requested):

kickapoo:
@nabinhait recheck please

nabinhait (approved):

rmehta:
@kickapoo @dottenbr @nabinhait  I think this was an example of a bad merge. We need correct use case, documentation, test cases and the feature needs to be linked to the overall workflow.

Let us not do half-baked features. This is where the problem starts.

I am reverting this PR

PR #8134, opened 2017-03-24
Manufacturer childtable

### LINKED ISSUE #8089: Make Manufacturer Part Numbers a child table in Item instead of a data field
- [ ] Add new Section "Manufacturer Details" and move the relevant fields out of "Supplier Details"
- [ ] Manufacturer and Manufacturer Partnumber should be editable grid child table
- [ ] write patch
- [ ] update documentation

dottenbr:
This fixes https://github.com/frappe/erpnext/issues/8089

Changed the concept from having one manufacturer and one manufacturer part number in an item to being able to have multiple. Also moved manufacturer out of supplier details section where I dont think it belongs.

bildschirmfoto 2017-03-24 um 15 06 50

Tested and created patch for existing entries.


dottenbr:

donalturrentine:
Hello all, this looks like exactly what we need in our ERPNext setup, any tips on how to get it into the user interface? Thanks!

vrms:
@nabinhait isn't it a bit strange that even though this has been merged in march 2017 it is still not a reality on the current v 10.1.75, neither v11.0.3-beta.31? Both of them still have the old manufacturer part number logic (data field in SUPPLIER DETAILS section, not independent child table) in the item master?
image
@dottenbr you are running `develop`, right? is the merge applied there?

vrms:
I have looked just a little bit into the code and it seems that the merge never happened

example: [variant.py](https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/item_variant.py)

line 45 **before** the merge:
` 	variant.manufacturer = manufacturer`

line 45 **after** the merge should have been replaced with:
` 	variant.append("manufacturers", {`

but the string ` 	variant.append("manufacturers", {` does not exist in the current `item_variant.py` instead the old ` 	variant.manufacturer = manufacturer` is still there

donalturrentine:
Bizarre that the merge never happened. We also need this for our environment

joolsr:
Yes I noticed this too. On purchase orders you can only get supplier codes to show up on po forms but really we need the mfr part numbers instead which you can't easily add to the form

vrms:
there was a test that failed, maybe that's what caused the merge never going through?

donalturrentine:
It looks like the code changes were reverted:

https://discuss.erpnext.com/t/manufacturer-and-manufacturer-part-numbers/40922/4

Also, perhaps never complete to begin with. For our use case, we need full capabilities with Suppliers and Manufacturers. Our items can be fulfilled by multiple items from multiple manufacturers, all of which are purchased from suppliers...

vrms:
can you please clarify @nabinhait whether [this merge from April 2018](https://github.com/frappe/erpnext/pull/8571) has removed the functionality of independent table for manufacturer/manufacturer part number in the `item` which had been added here again? 

That would at least explain the fact it is not there.

donalturrentine:
Hoping to get some movement on this, the discussion @frappe is here: https://discuss.erpnext.com/t/manufacturer-and-manufacturer-part-numbers/40922/19

joolsr:
Yes this could do with fixing. Its an oddity that our company has to work around. Its very confusing at the moment

PR #8136, opened 2017-03-24
Add Customs Tariff Number

### LINKED ISSUE #8092: Create Customs Tariff Number and add it to Item as a link field
- [ ] Add new Section "Foreign Trade Details" to Item
- [ ] Add field "Country of Origin" and link it to Country DocType
- [ ] Link field "Customs Tariff Number" New DocType "customs tariff number" with "Number" and "Description"

dottenbr:
Fixes https://github.com/frappe/erpnext/issues/8092

In foreign trade we need to add Country of Origin and Customs Tariff Numbers. This is to share information with customs agencies. Ultimately this has to be also put on delivery notes etc. but that's a second step.

I have added the new section in Item:

bildschirmfoto 2017-03-24 um 15 48 18

You have a quick edit for a new item:

bildschirmfoto 2017-03-24 um 15 48 37

The user rights are set to stock manager for CRUD and stock user for R. That should be good.


kickapoo:
Description should be small text field, not a data field.

dottenbr:
I disagree, it should remain a data field

dottenbr:

PR #8179, opened 2017-03-28
Add Short Code field (fixes #8099)

### LINKED ISSUE #8099: Add Customer Short Code to Customer
- [ ] Add 3-digit short code that can be used in many ways for the customer
- [ ] Create auto-rendering function based on customer name such as Frappe Inc. -> FRI
- [ ] Blacklist certain offensive names

kickapoo:
Please @nabinhait  review :)


erpnext/selling/doctype/customer/customer.py
```diff
@@ -231,3 +250,48 @@ def get_credit_limit(customer, company):
 		credit_limit = frappe.db.get_value("Company", company, "credit_limit")
 
 	return flt(credit_limit)
+
+
+@frappe.whitelist()
+def short_code_generator(customer_name):
+	new_sc = ''
```
nabinhait: put this in short_code.py (files this long should be separated)

nabinhait:
- [ ] How is short code used? This is just creating the short code, not linked to any workflow yet
- [ ] Test cases?

A broader discussion, can we do without short codes completely? Is this global search in v8 is pretty good!

nabinhait (changes requested):

kickapoo:
Short Code is self explained. Customer has a short code like: Frappe Inc --> FRI. 
Please @nabinhait provide documentation file path so @dottenbr can take over and add the proper explanation. 
I will add Test Cases. 
@nabinhait Blacklist can't be bootstraped, mayde someone want KKK as short code. Its up to the "someone".

rmehta:
@kickapoo  we still need workflow linked (where is this used???) No point having just made a feature that is unconnected.

kickapoo:
@dottenbr  please take over on @rmehta comment.

dottenbr:
Customer Short Code is used for searching in quotations, sales orders. It will also be used at ESO to generate assembly numbers (item numbers). If you don't see this as useful to others we can make it a customization. We should decide these things before we implement it though.

rmehta:
This point is not whether its useful, whether its complete.

kickapoo:
@rmehta if you aren't donna use it there is not reason to complete it and spend time on it. I can write the Unit Tested that are requested for this.

So please clearify the final call with won't merge flag or not so we can schedule our eso-internal features.

rmehta:
Its not about me using it. At this point, you are not using it either 😅

dottenbr:
It is not complete but we have a path to how to use it. We are using it 100% and we need it but we don't have it because we designed it as a contribution rather than a customization.

Do you want this in or out @rmehta?

Why was this closed @kickapoo?

nabinhait:
Closing this as per https://github.com/frappe/erpnext/wiki/Pull-Request-Guidelines. Please reopen after implementing workflow.

PR #8211, opened 2017-03-29
Read permissions to Employee on Employee Loan and Loan Type

### LINKED ISSUE #8184: Employee Loan Application - Restrict Employee Role based on Employee
This permission restriction should be available by default.

For Employee Loan Application and Employee Loan, user with role Employee should access, if Employee matches.

loan-appliaction

Also, on Loan Type, Employee should have Read permission. Currently, only HR Manager has this permission.

KanchanChauhan:
Fixed https://github.com/frappe/erpnext/issues/8184


erpnext/hr/doctype/employee_loan/employee_loan.json
```diff
@@ -942,6 +942,26 @@
    "share": 1, 
    "submit": 1, 
    "write": 1
+  }, 
+  {
+   "amend": 0, 
+   "apply_user_permissions": 0,
```
nabinhait: User permissions should be applied for employee role

nabinhait (changes requested):

nabinhait (approved):

PR #8220, opened 2017-03-30
add delivery date to print format Fix #8111

### LINKED ISSUE #8111: Add Delivery Date to Sales Order Print
The Delivery Date

grafik

Should definitely show up in the sales order print which is used as an order confirmation

grafik

Again this is probably generic enough to be a change in the core.

emakis00:


emakis:
Build error is unrelated with the issue!

nabinhait (approved):

PR #8245, opened 2017-03-30
Prevent Variants from Manufacturer Update

bcornwellmott:
The variants of items with a manufacturer part number but no manufacturer currently get updated then don't pass validation. This will prevent them from getting updated.


nabinhait:
Thanks for the pull request, but I found more efficient way, to do the patch. It will bypass both validate and on_update method. Check https://github.com/frappe/erpnext/commit/c040015bb5e17b233c343702b82a040b5d43f5df

Closing this.

PR #8254, opened 2017-03-31
changed accumulated header

RicardoJohann:
Changes the header on P&L, Balance Sheet and Cash Flow

image

### DIFF
--- a/erpnext/accounts/report/balance_sheet/balance_sheet.py
+++ b/erpnext/accounts/report/balance_sheet/balance_sheet.py
@@ -8,11 +8,14 @@ from frappe.utils import flt, cint
 from erpnext.accounts.report.financial_statements import (get_period_list, get_columns, get_data)
 
 def execute(filters=None):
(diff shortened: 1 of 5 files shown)


erpnext/accounts/report/balance_sheet/balance_sheet.py
```diff
@@ -8,11 +8,14 @@
 from erpnext.accounts.report.financial_statements import (get_period_list, get_columns, get_data)
 
 def execute(filters=None):
-	period_list = get_period_list(filters.from_fiscal_year, filters.to_fiscal_year, filters.periodicity, filters.company)
+	period_list = get_period_list(filters.from_fiscal_year, filters.to_fiscal_year, filters.periodicity, filters.accumulated_values)
```
nabinhait: Why removed filters.company? Is it very important to have company to get correct fiscal year based on a date.

nabinhait (changes requested):

PR #8262, opened 2017-03-31
Update autoname option

kickapoo:


kickapoo:
this has conflict an unmerge PR https://github.com/frappe/erpnext/pull/8263

PR #8266, opened 2017-03-31
Rename DocType LetterHead (fixes #8110)

### LINKED ISSUE #8110: Rename Letter Head to Letterhead
Letterhead is missspelled as "Letter Head" and should be renamed?

grafik

kickapoo:


dottenbr:
LetterHead should be Letterhead

nabinhait (changes requested):
- You need to search and replace in the code.
- While renaming a doctype using `rename_doc` function, it will replace all the field's "Options" value wherever Letterhead is linked to. But it will not change teh underline json files. You need to save those doctype manually or using script to update the json.

kickapoo:
@nabinhait 
If I understand correct. Rename_doc as patch still remains but I have to manually search and replace the name to ? frappe and erpnext folders? only erpnext ?

kickapoo:
@nabinhait Recheck

erpnext/patches/v8_0/rename_letterhead_doctype.py
```diff
@@ -0,0 +1,7 @@
+import frappe
+
+def execute():
+    old_name = 'LetterHead'
```
nabinhait: old_name is "Letter Head"

erpnext/patches/v4_0/reset_permissions_for_masters.py
```diff
@@ -10,7 +10,7 @@ def execute():
 		"Communication", "Company", "Contact Us Settings",
 		"Country", "Currency", "Currency Exchange", "Deduction Type", "Department",
 		"Designation", "Earning Type", "Event", "Feed", "File", "Fiscal Year",
-		"HR Settings", "Industry Type", "Leave Type", "Letter Head",
+		"HR Settings", "Industry Type", "Leave Type", "LetterHead",
```
nabinhait: When this patch will run, it is still "Letter Head"

erpnext/accounts/doctype/journal_entry/journal_entry.json
```diff
@@ -1124,10 +1124,10 @@
    "in_global_search": 0, 
    "in_list_view": 0, 
    "in_standard_filter": 0, 
-   "label": "Letter Head", 
+   "label": "LetterHead",
```
nabinhait: You need to change "modified" date-stamp as well in all json, otherwise it will not be reloaded with `bench update`.

nabinhait (changes requested):
- You need to manually correct the files for Letterhead doctype. You can save the letterehead doctype, it will create the directory. Then you can copy the codes from existing old directory, if code exists. And then delete the old directory.
- I also think, we should call it "Letterhead", instead of "LetterHead". @dottenbr @rmehta Can you suggest?

kickapoo:
I closing this one. I am not in to replacing file names and copy pasting directories. Its seems to me that is a process that can take the core team 5 minutes to resolve.

Issue #8232, opened 2017-03-30
V8 Project Name in Kanban


almeidapaulopt:
Name of the project isn't shown on kanban
![screenshot from 2017-03-30 15-34-45](https://cloud.githubusercontent.com/assets/22657407/24509637/cde89e62-155e-11e7-8f1d-cf9a1f1b439b.png)

PR #8268, opened 2017-03-31
[fix] Show project title in kanban view

netchampfaris:
- frappe/erpnext#8232


SettupillaiPrabhakaran:
I have a Doctype **Project Management**
In this Doctype there is a **select Field** name **Kanban Status**

And I create **Two kanban Board** using Project management doctype
one is **PM Board** and **TL Board** , If I create a Column in the board name was **Open**
in both boards, if I create a card in that Open column the card was showing in both board,
But I don't want that its should be shown in only one board. the card created board only

PR #8281, opened 2017-04-02
[fix] Special character issue in pos profile data

rohitwaghchaure:


erpnext/stock/get_item_details.py
```diff
@@ -327,9 +327,9 @@ def get_pos_profile_item_details(company, args, pos_profile=None):
 
 @frappe.whitelist()
 def get_pos_profile(company):
-	condition = "and company = '%s'"%(company) if company else ''
+	condition = "and company = %s" if company else ''
 	pos_profile = frappe.db.sql("""select * from `tabPOS Profile` where user = %s
-		{cond}""".format(cond=condition), (frappe.session['user']), as_dict=1)
+		{cond}""".format(cond=condition), (frappe.session['user'], company), as_dict=1)
```
nabinhait: this will break if company is blank

nabinhait (changes requested):

erpnext/stock/get_item_details.py
rohitwaghchaure: Company is mandatory in pos profile, so it will never be blank

nabinhait:
Pushed a commit, closing this.

PR #8287, opened 2017-04-03
Calculate item rate based on stock qty if item uom and stock uom is not equal

mbauskar:
https://discuss.erpnext.com/t/v8-multiple-uoms-in-selling-amount-issue/21960

### DIFF
--- a/erpnext/controllers/taxes_and_totals.py
+++ b/erpnext/controllers/taxes_and_totals.py
@@ -50,16 +50,21 @@ class calculate_taxes_and_totals(object):
 	def calculate_item_values(self):
 		if not self.discount_amount_applied:
 			for item in self.doc.get("items"):
+				price_list_rate = 0.0
 				self.doc.round_floats_in(item)
 
 				if item.discount_percentage == 100:
 					item.rate = 0.0
 				elif not item.rate:
-					item.rate = flt(item.price_list_rate *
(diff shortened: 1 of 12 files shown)


erpnext/controllers/taxes_and_totals.py
```diff
@@ -50,16 +50,21 @@ def validate_conversion_rate(self):
...
 		if not self.discount_amount_applied:
 			for item in self.doc.get("items"):
+				price_list_rate = 0.0
 				self.doc.round_floats_in(item)
 
 				if item.discount_percentage == 100:
 					item.rate = 0.0
 				elif not item.rate:
-					item.rate = flt(item.price_list_rate *
+					price_list_rate = item.price_list_rate if item.uom == item.stock_uom \
```
nabinhait: Why setting in a temporary variable? It should be set in price_list_rate field itself. Sometimes, user prints price_list_rate, in that case, it should be displayed as per sales uom.

nabinhait (changes requested):
As we are setting price_list_rate, it should be fetched on change of sales uom as well.

Also, check the Pricing Rule logic, currently it is based on stock_uom. But it should be applied after multiplying conversion factor.

nabinhait:
@mbauskar What's the status?

PR #8296, opened 2017-04-03
Add `image` field to fetch in Customer List

netchampfaris:


nabinhait (approved):

basdereus84:
Shouldn't this be added to the Supplier list and Contact list as well?

PR #8335, opened 2017-04-05
Added school settings doctype

manassolanki:


nabinhait (changes requested):
- Add screenshots
- The fields added in settings is not used anywhere, will merge the PR after full implementation.

PR #8359, opened 2017-04-06
Making jv from employee loan dont copy party type and employee

RandyLowery:


nabinhait:
Currently Party and Party Type is applicable in Journal Entry only if account type is "Receivable". Is "employee loan account" type of receivable?

PR #8385, opened 2017-04-10
[fix] Vehicle expense report not working

rohitwaghchaure:


erpnext/hr/report/vehicle_expenses/vehicle_expenses.js
```diff
@@ -22,18 +22,6 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 						query_report.trigger_refresh();
 					});
 				}
-			},
```
nabinhait: As you have removed from_date and to_date filters, there is no point of setting those filter values on change of fiscal _year.

erpnext/hr/report/vehicle_expenses/vehicle_expenses.py
```diff
@@ -3,50 +3,47 @@
...
-	if filters.to_date:
-		conditions += " and date <= %(to_date)s"
+	fy = frappe.db.get_value('Fiscal Year', filters.get('fiscal_year'), ['year_start_date', 'year_end_date'], as_dict=True)
 	data = frappe.db.sql("""select vhcl.license_plate as "License",vhcl.make as "Make",vhcl.model as "Model",
-							vhcl.location as "Location",log.name as "Log",log.odometer as "Odometer",log.date as "Date",
-							log.fuel_qty as "Fuel Qty",log.price as "Fuel Price"
-							from `tabVehicle` vhcl,`tabVehicle Log` log
-							where vhcl.license_plate = log.license_plate and log.docstatus = 1 %s
-							order by date""" % (conditions,),filters, as_dict=1)
+		vhcl.location as "Location",log.name as "Log",log.odometer as "Odometer",log.date as "Date",
```
nabinhait: Use space after comma. In this case, use separate line for each column.

nabinhait (changes requested):

erpnext/hr/report/vehicle_expenses/vehicle_expenses.js
rohitwaghchaure: fy = frappe.db.get_value('Fiscal Year', filters.get('fiscal_year'), ['year_start_date', 'year_end_date'], as_dict=True)
Fetched start date and end date from the fiscal year

erpnext/hr/report/vehicle_expenses/vehicle_expenses.py
rohitwaghchaure: Fixed

PR #8399, opened 2017-04-11
[minor] don't throw error for buying transactions if price list is not selected

mbauskar:


nabinhait:
@mbauskar Any specific reason?

mbauskar:
@nabinhait, in Buying transaction documents price_list is not mandatory, so if the price_list kept blank then system does not fetch the item details

nabinhait (approved):

PR #8401, opened 2017-04-11
[enhance] Stock qty added in purchase invoice

rohitwaghchaure:


erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -424,6 +455,35 @@
...
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Stock Qty", 
+   "length": 0, 
+   "no_copy": 1,
```
nabinhait: If no_copy, it will not be copied from PO/PR.

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -340,6 +341,36 @@
...
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Stock UOM", 
+   "length": 0, 
+   "no_copy": 1,
```
nabinhait: why no_copy?

nabinhait (changes requested):
A patch is required to set stock_uom and stock_qty in existing documents.

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
rohitwaghchaure: Fixed

nabinhait (approved):

Issue #8416, opened 2017-04-12
Sales Invoice has wrong Print Format as Default


umairsy:
For the Sales Invoice, Print Format called Sample Print is auto-created and set as default Print. In the new account, when I created new Sales Invoice and checked preview, received following error message. Sample Print Format was auto-created and not manually by me.

![screen shot 2017-04-12 at 11 58 55 am](https://cloud.githubusercontent.com/assets/910238/24944186/b0030fe0-1f77-11e7-8063-fef27e7eb3ee.png)

PR #8419, opened 2017-04-12
[fix] Default Print Format for Sales Invoice

netchampfaris:
- frappe/erpnext#8416


erpnext/accounts/doctype/sales_invoice/sales_invoice.json
```diff
@@ -7,7 +7,7 @@
  "beta": 0, 
  "creation": "2013-05-24 19:29:05", 
  "custom": 0, 
- "default_print_format": "Sample Print", 
+ "default_print_format": "",
```
rohitwaghchaure: Keep default "Standard"

Issue #8247, opened 2017-03-31
[v8] POS - Stock Item level not showing


OneiricIT:
In new v8 POS, the stock item level which was shown in v7 is no longer visible next to the item. Similar case occurred in prior issue below.

https://github.com/frappe/erpnext/issues/3140

PR #8420, opened 2017-04-12
[POS] Show stock quantity of items

netchampfaris:
image

- frappe/erpnext#8247


nabinhait:
@netchampfaris "Stock:" is not a good label. 
It can be "In Stock" / "Stock Qty" / "Available Qty" or "9 available".
What's say?

netchampfaris:
Let's go with 'Stock Qty'

nabinhait (approved):

PR #8429, opened 2017-04-12
Statistical Components for Salary Structure

ckosiegbu:
Today, any component included in a Salary Structure will contribute to the calculation of Gross Pay and Total Deductions. To make the calculation of components easier, we have introduced the ability to flag components as Statistical. This behaves something like a throwaway variable that can hold the results of calculations but does not contribute to the final calculation of the Gross Pay and Total Deductions. It allows for incremental calculation of Components and reuse of values across many steps.

image
Field "Statistical Component" is added. Selecting this option allows the Salary Detail to be used in calculations without appearing in the Salary Slip and without adding to the sum total or earnings or deductions (something like a transient variable)

image
Earnings section of Salary Structure show items that are statistical

image
Deductions section of Salary Structure show items that are statistical

image
Create Salary Slip

image
Salary Slip created with only non-statistical Components in Earnings and Deductions. Statistical Components do not get included in the final Salary Slip

Regards,
Chude


rmehta:
@ckosiegbu tabs please 😅

ckosiegbu:
@rmehta Oops! Sorry. Converted to tabs now. I'll try to remember next time :-)

nabinhait (approved):

sumitdas4u:
but it should reflect in the Accural Journal Entry at least

PR #8430, opened 2017-04-13
[minor] fetch supplier address, address_display on supplier trigger

mbauskar:
fetch supplier address and address display on supplier field trigger

stock_entry


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -229,6 +229,11 @@ erpnext.stock.StockEntry = erpnext.stock.StockController.extend({
 		this.frm.set_indicator_formatter('item_code',
 			function(doc) { return (doc.qty<=doc.actual_qty) ? "green" : "orange" })
 
+		this.frm.add_fetch("purchase_order", "supplier", "supplier");
+		this.frm.add_fetch("supplier", "supplier_name", "supplier_name");
+
+		frappe.dynamic_link = { doc: this.frm.doc, fieldname: 'supplier', doctype: 'Supplier' }
```
nabinhait: where is it used?

nabinhait (changes requested):
Patch missing to copy value form "supplier_address" to "supplier_address_display" and to set "supplier_address" as blank.

mbauskar:
@nabinhait added patch to move the supplier_address to address_display field

PR #8470, opened 2017-04-17
Fix for 8466

PawanMeh:
Use Case

1. Track conversion ratios from Lead to Opportunity, Opportunity to Quotation, Quotation to Order and Order value generated by Leads. The analysis generally is done by Campaign and Lead Owner to determine efficiency of Campaigns and Leads.

campaign_efficiency

lead owner efficiency

### DIFF
--- a/erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
+++ b/erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
@@ -1,0 +1,69 @@
+# Copyright (c) 2013, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
+
+from __future__ import unicode_literals
+import frappe
+from frappe import _
+from frappe.utils import flt,cstr
+from erpnext.accounts.report.financial_statements import get_period_list
+
+def execute(filters=None):
+	columns, data = [], []
+	columns=get_columns()
+	data=get_lead_data(filters)
+	return columns, data
+	
+def get_columns():
+	columns = [_("Campaign Name") + ":data:130", _("Lead Count") + ":Int:80",
+				_("Opp Count") + ":Int:80",
+				_("Quot Count") + ":Int:80", _("Order Count") + ":Int:100",
+				_("Order Value") + ":Float:100",_("Opp/Lead %") + ":Float:100",
+				_("Quot/Lead %") + ":Float:100",_("Order/Quot %") + ":Float:100"
+	]
+	return columns
+
+def get_lead_data(filters):
+	conditions=""
+	if filters.from_date:
+		conditions += " and date(creation) >= %(from_date)s"
+	if filters.to_date:
+		conditions += " and date(creation) <= %(to_date)s"
+	data = frappe.db.sql("""select campaign_name as "Campaign Name", count(name) as "Lead Count" from `tabLead` where 1 = 1 %s group by campaign_name""" % (conditions,),filters, as_dict=1)
+	dl=list(data)
+	for row in dl:
+		is_quot_count_zero = False
+		row["Quot Count"]= get_lead_quotation_count(row["Campaign Name"])
+		row["Opp Count"] = get_lead_opp_count(row["Campaign Name"])
+		row["Order Count"] = get_quotation_ordered_count(row["Campaign Name"])
+		row["Order Value"] = get_order_amount(row["Campaign Name"])
+		row["Opp/Lead %"] = row["Opp Count"] / row["Lead Count"] * 100
+		row["Quot/Lead %"] = row["Quot Count"] / row["Lead Count"] * 100
+		#Handle div by zero and reset count to zero
+		if row["Quot Count"] == 0:
+			row["Quot Count"] = 1
+			is_quot_count_zero = True
+		row["Order/Quot %"] = row["Order Count"] / row["Quot Count"] * 100
+		if is_quot_count_zero ==  True:
+			row["Quot Count"] = 0
+	return dl
+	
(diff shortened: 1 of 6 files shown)


erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
```diff
@@ -0,0 +1,69 @@
...
+			row["Quot Count"] = 1
+			is_quot_count_zero = True
+		row["Order/Quot %"] = row["Order Count"] / row["Quot Count"] * 100
+		if is_quot_count_zero ==  True:
+			row["Quot Count"] = 0
+	return dl
+	
+def get_lead_quotation_count(campaign):
+	quotation_count = frappe.db.sql("""select count(name) from `tabQuotation` 
+										where lead in (select name from `tabLead` where campaign_name = %s)""",campaign)
```
rmehta: style: can we have consistent tabbing for mulitline -- looks like hanging fruits in a tree. I think having a consistent whitespace will make the code look a lot cleaner.

erpnext/crm/report/lead_owner_efficiency/lead_owner_efficiency.js
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2016, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+frappe.require("assets/erpnext/js/financial_statements.js", function() {
```
rmehta: i don't think this require is needed (if you are not using any special JS functions (?)).

PawanMeh:
Made changes requested

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
```diff
@@ -0,0 +1,70 @@
...
+			row["Quot Count"] = 1
+			is_quot_count_zero = True
+		row["Order/Quot %"] = row["Order Count"] / row["Quot Count"] * 100
+		if is_quot_count_zero ==  True:
+			row["Quot Count"] = 0
+	return dl
+	
+def get_lead_quotation_count(campaign):
+	quotation_count = frappe.db.sql("""select count(name) from `tabQuotation` 
+	where lead in (select name from `tabLead` where campaign_name = %s)""",campaign)
```
nabinhait: This sub-query has been called repeatedly in all the functions and those functions are called for campaign. You should create a dictionary like `{'campaign1': ['lead1', 'lead4'], 'campaign2': ['lead5', 'lead 7']}`, outside the `for` loop, then use it.

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
```diff
@@ -0,0 +1,70 @@
...
+	return flt(opportunity_count[0][0]) if opportunity_count else 0
+	
+def get_quotation_ordered_count(campaign):
+	quotation_ordered_count = frappe.db.sql("""select count(name) from `tabQuotation` where status = 'Ordered'
+	and lead in (select name from `tabLead` where campaign_name = %s)""",campaign)
+	return flt(quotation_ordered_count[0][0]) if quotation_ordered_count else 0
+	
+def get_order_amount(campaign):
+	ordered_count_amount = frappe.db.sql("""select sum(base_net_amount) from `tabSales Order Item`
```
nabinhait: Do you want to show the amount without taxes? Otherwise, you should use `base_grand_total` value.

erpnext/crm/report/lead_owner_efficiency/lead_owner_efficiency.py
```diff
@@ -0,0 +1,72 @@
...
+			row["Quot Count"] = 1
+			is_quot_count_zero = True
+		row["Order/Quot %"] = row["Order Count"] / row["Quot Count"] * 100
+		if is_quot_count_zero ==  True:
+			row["Quot Count"] = 0
+	return dl
+	
+def get_lead_quotation_count(leadowner):
+	quotation_count = frappe.db.sql("""select count(name) from `tabQuotation` 
+	where lead in (select name from `tabLead` where lead_owner = %s)""",leadowner)
```
nabinhait: You should use the same approach as above.

nabinhait (changes requested):
@PawanMeh Please let me know if you need any kind of help.

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
PawanMeh: As these reports are for judging sales performance, taxes should not be in consideration. So I think the logic to use net amount is fine.

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
PawanMeh: Executing a query in for loop repeatedly for each lead may reduce the performance benefit got from indexing using the in list. However, if you still think this needs to be changed, let me know, I can make the changes.

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
nabinhait: sorry! `base_net_amount` is correct

erpnext/crm/report/campaign_efficiency/campaign_efficiency.py
nabinhait: You did not get my point, I am telling to do the query only once before the "for" loop. The query will give all the leads grouped by campaign. I can refactor the code if you want, after pulling your changes locally.

PR #8471, opened 2017-04-17
[fix] Pricing rule for pos

rohitwaghchaure:


erpnext/accounts/page/pos/pos.js
```diff
@@ -1779,6 +1779,13 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		})
 	},
 
+	validate_item_condition: function (data, item) {
+		var apply_on = frappe.model.scrub(data.apply_on);
+
+		return (data.apply_on == 'Item Group'
```
nabinhait: IG 0
    - IG 01
        - IG 011
        - IG 012
    - IG 02

For the above Item group tree, if an item is belong to IG 012, the the pricing rule applicable to IG0 and IG01 should be evaluated.

nabinhait (changes requested):

erpnext/accounts/page/pos/pos.js
rohitwaghchaure: Fixed

erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -147,21 +147,26 @@ def get_items_list(pos_profile):
 		""".format(cond=cond), tuple(item_groups), as_dict=1)
 
 def get_item_group(pos_profile):
```
nabinhait: get_item_groups ?

nabinhait (changes requested):

erpnext/accounts/doctype/sales_invoice/pos.py
rohitwaghchaure: To get the list of item groups defined in the POS profile or all item groups

nabinhait: I am just suggesting to rename the function

rohitwaghchaure: Fixed

nabinhait (approved):

nabinhait (approved):

PR #8474, opened 2017-04-17
Pull non ordered material requests using supplier on purchase order

### LINKED ISSUE #8451: V8 - Purchase Order missing option to fetch items based on Supplier
In the Purchase Order, there was an option to fetch Items based on Supplier. This option is missing in version 8. In the new Purchase Order, we had button "For Supplier", but it is missing now.

https://frappe.github.io/erpnext/user/manual/en/buying/articles/pull-items-in-purchase-order-based-on-supplier6

rohitwaghchaure:
Added the feature to pull the non ordered material requests using supplier on purchase order
Fixed https://github.com/frappe/erpnext/issues/8451

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.js
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.js
@@ -180,7 +180,6 @@ erpnext.buying.PurchaseOrderController = erpnext.buying.BuyingController.extend(
 					}
 				})
 			}, __("Add items from"));
-
 	},
 
 	tc_name: function() {


erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -181,6 +181,18 @@ erpnext.buying.PurchaseOrderController = erpnext.buying.BuyingController.extend(
 				})
 			}, __("Add items from"));
 
+		cur_frm.add_custom_button(__('For Supplier'),
+			function() {
+				erpnext.utils.map_current_doc({
```
nabinhait: There is already a function for this, just need to bind it to the custom button
https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order/purchase_order.js#L94

nabinhait (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.js
rohitwaghchaure: Fixed!

erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -94,7 +94,7 @@ erpnext.buying.PurchaseOrderController = erpnext.buying.BuyingController.extend(
 	get_items_from_open_material_requests: function() {
 		erpnext.utils.map_current_doc({
 			method: "erpnext.stock.doctype.material_request.material_request.make_purchase_order_based_on_supplier",
-			source_name: this.frm.doc.supplier,
+			source_doctype: "Supplier",
```
nabinhait: Why this? It needs `source_name` which is selected supplier in the PO, not `source_doctype`.

nabinhait (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -181,6 +182,12 @@ erpnext.buying.PurchaseOrderController = erpnext.buying.BuyingController.extend(
 				})
 			}, __("Add items from"));
 
+		cur_frm.add_custom_button(__('For Supplier'),
```
rmehta: `cur_frm` !!!!

Should be a button below supplier field only shown when document is not saved (__islocal)

rmehta (changes requested):

PR #8477, opened 2017-04-17
Multi Currency Exchange rate

arpitjain06:
1. if Customer has default currency (USD) apart from company currency(INR),  and Default Debtors accounts set as debtors(usd) in accounting section in iCustomer master.
2. During the creation of sales invoice , with 
   a. exchange rate: -  64.43   (1 USD = [?] INR), 
   b. Posting Date:- 17-04-2017
   c.  debit_to:   debtors(usd) in  accounting section
   d Grand Total  :- 40 $ USD

So  if customer pays the invoice amount after 2-3 month ,if  exchange rate increase to 65 or decrease 63  , So Customer will only pays  40 $, with current exchange rate not with Sales invoice Exchange rate i.e   64.43   (1 USD = [?] INR

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -323,6 +323,9 @@ class JournalEntry(AccountsController):
 		for d in self.get("accounts"):
 			if d.account_currency == self.company_currency:
 				d.exchange_rate = 1
+			elif d.exchange_rate and d.account_currency != self.company_currency:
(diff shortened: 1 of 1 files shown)


nabinhait:
In Journal Entry, the exchange rate against Debtors row will be 64.43 only. And against bank account (USD), it should be 63. And there will be another row to book "Exchange Gain / Loss" for 57.20 INR.

If we set exchange rate as 63 in the Debtors row, then there will always an outstanding amount of 57.20 in INR but in USD, there will be no outstanding and INR outstanding will never get cleared.

PR #8509, opened 2017-04-20
[enhance] automatic batch creation, move and split

rmehta:
Pending

- [x] Automatic batch selection for outgoing (Stock Entry, Delivery Note)

### Setting in Item to automatically create new Batches

item_setup_for_batch

### Batch Move

batch-move

### Batch Split

batch-split


jof2jc:
@rmehta, what about new incoming batches on purchase receipt/invoice and stock entry? Will the system auto-create it like serial no?

nabinhait:
@rmehta Should I check and test this PR now? Or I should wait for your pending task.

erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
```diff
@@ -1040,11 +1040,11 @@
    "unique": 0
   }, 
   {
-   "allow_on_submit": 0, 
+   "allow_on_submit": 1, 
    "bold": 0, 
    "collapsible": 0, 
    "columns": 0, 
-   "fieldname": "allow_zero_valuation_rate", 
+   "fieldname": "is_sample_item",
```
nabinhait: "is_sample_item" has been renamed to "allow_zero_valuation_rate" last week.

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
rmehta: done! Good catch

PR #8511, opened 2017-04-20
Improving Item Price DocType (fixes #8456)

### LINKED ISSUE #8456: Improving upon Item Price
We are buying and selling a lot of items of which the price varies based on quantity and customer or supplier that we buy from or sell to. Currently the only way this can be done is by pricing rules which we think are more to discount prices per customer or supplier group, time period (sales) etc. So in our mind using pricing rule to achieve quantity based pricing is a hack and not an ideal solution.

We propose to improve upon Item Price by adding the fields:

- Supplier (Link:Supplier) [if:buying]
- Supplier Part Number (Child Table entry) [if:buying]
- Customer (Link:Customer) [if:selling]
- Minimum Qty (Float)
- UOM (Link: UOM)
- Packing Unit (Int, quantity that must be bought or sold per UOM)
- Rate per UOM (rename Rate to that to be specific)
- Lead Time in days (Int)
- Valid From (Date)
- Valid Upto (Date)
- Note (Text)
- Disable (Checkbox)

If this is possible then based on that also pricing rules can apply to further apply discounts etc.

So for us we would put in 1-10 price breaks based on quantity for an item that we buy from up to 15 suppliers.

kickapoo:


kickapoo:
screenshot 2017-04-20 13 03 16

rmehta:
- [ ] let us not keep 3 columns - we have not used it anywhere
- [ ] this should also impact selection of the price: https://github.com/frappe/erpnext/blob/8f3cc81302d9bde8aa4cbb8ff1c51074b0979fc2/erpnext/stock/get_item_details.py#L218

kickapoo:
screenshot 2017-04-21 16 34 37

kickapoo:
itemprice

erpnext/stock/get_item_details.py
```diff
@@ -217,16 +217,18 @@ def get_default_cost_center(args, item):
...
 	meta = frappe.get_meta(args.parenttype or args.doctype)
-
 	if meta.get_field("currency"):
 		validate_price_list(args)
 		validate_conversion_rate(args, meta)
 
-		price_list_rate = get_price_list_rate_for(args.price_list, item_doc.name)
-
+		price_list_rate = get_price_list_rate_for(args.price_list,
+											  	  item_doc.name,
```
rmehta: please use consistent indentation - not hanging in one corner

erpnext/stock/get_item_details.py
```diff
@@ -257,25 +259,52 @@ def insert_item_price(args):
...
+		:param price_list: str (Standard Buying or Standard Selling)
+		:param item_code: str, Item Doctype field item_code
+		:param qty: Derised Qty
+	"""
+	price_list_rate = frappe.db.sql("""
+		select price_list_rate
+		from `tabItem Price`
+		where item_code=%s
+			and price_list=%s
+			and min_qty<=%s order by min_qty desc""",
```
rmehta: date validation?

rmehta: Test cases missing for this - make a purchase order and see different prices are selected for different qtys

Also show UI screenshot to show it works from the UI

rmehta (changes requested):

nabinhait:
Please resend the PR after completion, closing for now.

PR #8537, opened 2017-04-22
Add optional Account Number field (partial fix #8194)

### LINKED ISSUE #8194: Add Account Number
In ``Accounts`` we have account numbers. We use the German [SKR03](http://www.willer-partner.de/fileadmin/images/aktuelles/SKR03_mit_neu_eingefuehrten_und_geaenderten_Konten.pdf) for which every accountant ant tax consultant typically doesn't refer to the full account name but instead the account number e.g. "4955" would be "Cost related to Accounting".

- [ ] Add ``Account Number`` to ``Account``
- [ ] Show ``Account Number`` in ``Account Tree``
- [ ] Make ``Account Number`` searchable in link field and enable it for quick entry

kickapoo:
This is a partial PR for Account Number for the issue #8194 

This PR adds an optional field Account Number that if is set can be show in the Tree List view of Account Tree (not in Report View - read below).
This PR doesn't effect the current state of the system as its only a render if Account Number is set. 

There is the suggestion of updating the `autoname` function of account.py with `{Account Number} - {Account Name}`.  Please request change if the choice of autoname is the one and I will update the PR with the proper patch and Unit Test.

Also, some pointers/help is needed for sorting Account Tree based on Account Number field.

accountnumber


nabinhait (changes requested):
- This will just show the account number in the chart of accounts. But until it is used in the actual transactions while selection of account, it is not that useful.
- And for that, you have to use this number in the naming of account.
- There should be an option to define account number while adding an account from Chart of Accounts.
- The accounts should be appear in COA, ordered by the account number.

nabinhait:
@kickapoo Sorry, I did not read the descriptions in pull request properly and made the above comments. Those are the some points to go ahead. It is the most easiest way to intrude numbering system, if we bind it to the ID/name.

kickapoo:
@nabinhait Ty I will close it and reopen it with the proper PR.

bghayad:
Hello;
Is there any PR has been created that is related to the choice of autoname?
And what about sorting the Account Tree based on Account Number field?
One more thing: what is the required to add the issue #8194 to the master?
Regards
Bilal

bghayad:
@kickapoo Did u ope PR related to this fix?
There is also a bug in the fix: after adding the account number, if we need to do rename, then the new name appears in the tree as the modified name with the account number until we do refresh. 
Also I am interested in the PR that will do the autoname and to appear in the reports.
Regards
Bilal

PR #8544, opened 2017-04-24
Add bank guarantee doctype with documentation

ishanloya:
Add bank guarantee doctype to accounts module which lets a company track the bank guarantees (BG) given to customers


rmehta:
👍 please add screenshots to the documentation and also the PR too! Screenshots are very important to check if the fields have been labelled correctly etc.

ishanloya:
Ok. Thanks @rmehta 🙂. Here's a screenshot. Will update in documentation as well.
screen shot 2017-04-24 at 8 09 18 pm

ishanloya:
Done. Thanks again.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
```diff
@@ -0,0 +1,368 @@
...
+   "delete": 1, 
+   "email": 1, 
+   "export": 1, 
+   "if_owner": 0, 
+   "import": 0, 
+   "permlevel": 0, 
+   "print": 1, 
+   "read": 1, 
+   "report": 1, 
+   "role": "System Manager",
```
nabinhait: Why system manager? I think it should permitted to Accounts User and Accounts Manager role.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
```diff
@@ -0,0 +1,368 @@
...
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "account_name",
```
nabinhait: we always try to use same fieldname as doctype name for link field. In this case, would prefer `account` instead of `account_name` field.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
```diff
@@ -0,0 +1,368 @@
...
+ "read_only": 0, 
+ "read_only_onload": 0, 
+ "search_fields": "customer", 
+ "show_name_in_global_search": 0, 
+ "sort_field": "modified", 
+ "sort_order": "DESC", 
+ "title_field": "customer", 
+ "track_changes": 0, 
+ "track_seen": 0
+}
```
nabinhait: Add another field for remarks/notes?

nabinhait (changes requested):
Add the document in config/accounts.py, to show it under Accounts module.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
ishanloya: Didn't realise that I hadn't changed permissions. Will keep in mind next time.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
ishanloya: Ok. Done.

erpnext/accounts/doctype/bank_guarantee/bank_guarantee.json
ishanloya: Done

nabinhait:
@ishanloya Thanks for the doing requested changes so promptly. You forgot to add the document in config/accounts.py, to show it under Accounts module.

ishanloya:
Thanks @nabinhait

PR #8553, opened 2017-04-24
Changes in the assessment plan and tool

manassolanki:


erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
```diff
@@ -7,17 +7,19 @@ cur_frm.add_fetch("assessment_plan", "student_batch", "student_batch");
 frappe.ui.form.on('Assessment Result Tool', {
 
 	onload: function(frm) {
+		frm.trigger("refresh");
```
nabinhait: Why this? `refresh` is always called after `onload`.

erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
```diff
@@ -7,17 +7,19 @@ cur_frm.add_fetch("assessment_plan", "student_batch", "student_batch");
...
+	},
+
+	refresh: function(frm) {
 		if (frappe.route_options) {
 			frm.doc.assessment_plan = frappe.route_options.assessment_plan;
+			frm.trigger("assessment_plan");
```
nabinhait: and why you are forcefully triggering this on every refresh?

As I can see, it resets the table. Hence, if you go to other tabs after entering some data and come back to the same tab, it will call refresh and eventually will clear the data.

nabinhait (changes requested):

erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
manassolanki: Actually if we are using the back button, it is rendering the table from the cache, that's why I have to forcefully trigger it to render the table again

erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
manassolanki: Yup, will remove it. Got confused with caching.

erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
nabinhait: it's not a good solution, it will reset the table values, if you just returning back from any other tabs or document without any changes.

may be you can use earlier `route_options` and on `refresh` check if routing from any other student group than selected, then trigger the "assessment_plan". And always clear the route options.

nabinhait (changes requested):

PR #8562, opened 2017-04-25
Change in the assessment tool

manassolanki:


erpnext/schools/doctype/assessment_result_tool/assessment_result_tool.js
```diff
@@ -1,23 +1,19 @@
...
 
 cur_frm.add_fetch("assessment_plan", "student_group", "student_group");
 cur_frm.add_fetch("assessment_plan", "student_batch", "student_batch");
 
 frappe.ui.form.on('Assessment Result Tool', {
-
-	onload: function(frm) {
+	refresh: function(frm) {
 		if (frappe.route_options) {
-			frm.doc.assessment_plan = frappe.route_options.assessment_plan;
```
nabinhait: Why you removed this? Then whats the use of route_options?

manassolanki: When we route to assessment result tool, updating the assessment plan field doesn't trigger the table, which gives a false perception to the user that there is some bug as assessment plan has value but the table isn't being rendered. The better approach for the time being short is let the user manually select the assessment plan and in between can find a solution.

nabinhait (approved):

PR #8615, opened 2017-04-28
Caluculate total interest and payable amount in Loan Application if fixed amount per period

nabinhait:
**Before:**

screen shot 2017-04-28 at 4 29 29 pm

**After:**

screen shot 2017-04-28 at 4 29 04 pm


nabinhait:
Test case pending

PR #8629, opened 2017-04-29
[minor] disabled the instructor role if the company domain is not Education

mbauskar:
`before`

School module doctypes should be hidden from the User if the Company domain is not Education.

screen shot 2017-04-29 at 3 46 38 pm

`after`

screen shot 2017-04-29 at 4 05 13 pm


nabinhait (changes requested):
Add Instructor role to the allow_roles list for Education domain.

PR #8656, opened 2017-05-02
Force round-up on rounded total

vishdha:
Summary:

1. check box in company doctype: "Round Up total in Sales Invoice".
2. modified taxes_and_totals.py


nabinhait:
Thanks for the PR, but I could not understood the requirement clearly.

- First thing, currently system does rounding based on "Smallest Currency Fraction Value" if it is mentioned in the Currency. So, if it is set as 0.1 there, then the system rounds up 3.34 to 3.30 and 3.68 to 3.70
- If "Smallest Currency Fraction Value" is not defined, then also system rounds up. But in this case, it rounds up 3.34 to 3.0 and 3.68 to 4.0

Then why we need this settings?

And you are always doing upper rounding, why so? If required, it should based on a settings in Selling / Buying Settings. The label can be "Always use upper rounding (without fraction) for rounding Grand Total"

Closing the PR for now.

Issue #8698, opened 2017-05-04
Error when trying to Preview Salary Slip


Olawale1:
Hi,

The following error pops up when trying to Preview Salary Slip

```
Traceback (most recent call last):
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/app.py", line 55, in application
    response = frappe.handler.handle()
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/handler.py", line 52, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/__init__.py", line 907, in call
    return fn(*args, **newargs)
  File "/home/frappe/benches/bench-2017-05-04/apps/erpnext/erpnext/hr/doctype/salary_structure/salary_structure.py", line 85, in make_salary_slip
    }, target_doc, postprocess, ignore_child_tables=True)
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/model/mapper.py", line 98, in get_mapped_doc
    postprocess(source_doc, target_doc)
  File "/home/frappe/benches/bench-2017-05-04/apps/erpnext/erpnext/hr/doctype/salary_structure/salary_structure.py", line 75, in postprocess
    target.run_method('process_salary_structure')
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/model/document.py", line 667, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/model/document.py", line 892, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/model/document.py", line 875, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
  File "/home/frappe/benches/bench-2017-05-04/apps/frappe/frappe/model/document.py", line 661, in 
    fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
  File "/home/frappe/benches/bench-2017-05-04/apps/erpnext/erpnext/hr/doctype/salary_slip/salary_slip.py", line 198, in process_salary_structure
    self.calculate_net_pay()
[...]

PR #8700, opened 2017-05-05
[minor] check if date of joining is available before creating Salary Structure & Salary Slip

mbauskar:
fixes for https://github.com/frappe/erpnext/issues/8698

`before`

hr-before
`after`

hr


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -321,9 +321,13 @@ def check_existing(self):
 	def sum_components(self, component_type, total_field):
 		joining_date, relieving_date = frappe.db.get_value("Employee", self.employee,
 			["date_of_joining", "relieving_date"])
+		
 		if not relieving_date:
 			relieving_date = getdate(self.end_date)
 
+		if not joining_date:
```
nabinhait: Why this validation is here?
Make joining_date field mandatory instead.

Here only validate if joining_date exists.

nabinhait (approved):

PR #8746, opened 2017-05-09
Added Address HTML for Dynamic Link detection

bcornwellmott:
Currently, there is no way to add Company as a Dynamic Link for Addresses. This prevents users from selecting their own company address in as the ship-to destination for POs and other buying documents (as well as a bunch more other things.

This adds the Address HTML field that is the key for adding a doctype to the Dynamic Link query.


bcornwellmott:
Got resolved by another pr

Issue #8769, opened 2017-05-11
Improvements in Return / Credit Note (Sales Invoice)


umairsy:
- [x] In the original Sales Invoice, if Update Stock was not checked, then in the Return Invoice, it should not auto-checked, even from POS.
- [ ] Improve naming of next steps. Current it shows "Return / Credit Note". Show it based on value, like if Update Stock is checked, then Return will appear. Same way, only if Sales Invoice is Paid (Partially as well), then only Credit Note option should be visible. If Sales Invoice is not paid, then in the case of Return, Sales Invoice can be canceled.
- [ ] It will be very intuitive for the User to be able to make a future adjustment against the Credit Note, and not against the old Sales Invoice linked to that Credit Note.

joezsweet:
👍

PR #8773, opened 2017-05-11
1st point #8769

nick9822:
Improvements in Return / Credit Note (Sales Invoice)


nabinhait:
Please send the PR from your **fork** and to the `develop` branch .

And this does not solve the problem if you make the Return against a POS invoice (without Update Stock).

PR #8775, opened 2017-05-11
Item Rate in Sales Return must be less than or equal to reference doc

nabinhait:


manasan3010:
@nabinhait can you please explain, why is this check necessary? 
Is there any accounting problem, if the returned value of an item was higher than the sold price?

PR #8789, opened 2017-05-12
Update lead.py

hereabdulla:
Import datetime to compare next contact date with now()

### DIFF
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -4,6 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 from frappe import _
+import datetime from datetime
 from frappe.utils import (cstr, validate_email_add, cint, comma_and, has_gravatar, now)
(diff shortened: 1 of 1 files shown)


rmehta:
What is the use case?

hereabdulla:
resolves the below issue
https://discuss.erpnext.com/t/error-in-data-import-tool-while-uploading-lead/19626

erpnext/crm/doctype/lead/lead.py
```diff
@@ -45,7 +46,7 @@ def validate(self):
 
 			self.image = has_gravatar(self.email_id)
 
-		if self.contact_date and self.contact_date < now():
+		if self.contact_date and self.contact_date < datetime.now():
```
nabinhait: datetime.now() can create problem with user date format.

Instead use `nowdate()` function from `frappe.utils`

nabinhait (changes requested):

PR #8794, opened 2017-05-12
Get from BOM in Production Planning Tool

### LINKED ISSUE #8028: [Enhancement] Production Planning Tool get items from BOM
Hi,

Would it be possible to add a third option to the Production Planning Tool that gets items from a BOM? If we have a multi-level BOM where we need to manufacture the child items instead of purchasing them, this enhancement would enable us to create production orders for such child items project-wise.

[Forum link](https://discuss.erpnext.com/t/production-planning-tool-get-items-from-bom/21343)

Thanks

ishanloya:
Resolves #8028 
https://discuss.erpnext.com/t/production-planning-tool-get-items-from-bom/21343


ishanloya:
screen shot 2017-05-12 at 10 40 09 am

bcornwellmott:
I'm not sure this is the most effective solution to the problem. It seems that the problem trying to be resolved is that when generating material requests, it uses the default BOM for each item instead of looking at the BOM that is linked in the master item. This change still doesn't resolve that, but just adds another way of adding a BOM that doesn't use the default BOM, right?

It seems to me that the correct solution isn't to add another "Get Items From" type, but to fix the current system so that it looks up the BOM referenced in the parent BOM instead of the default?

ishanloya:
The aim of this is to be able to generate production orders without having to create material requests first, which is what needs to be done right now. There's no quick way to create production orders directly from BOMs. Also, it does address the problem we're having of not having a default BOM for items since the BOM is project specific for us. 

If there was a way to override the default BOM option to consider project BOM first, that would solve one of our problems. But MRs would still need to be created before Production Orders are created, which is time-consuming.

ishanloya:
Any suggestions on how I can change or improve this? Otherwise, I can add it to my custom app instead.

Thanks

erpnext/manufacturing/doctype/production_planning_tool/production_planning_tool.py
```diff
@@ -178,6 +216,36 @@ def get_mr_items(self):
...
 
+	def get_bom_items(self):
+		bom_list = [d.bom for d in self.get('boms') if d.bom]
+		if not bom_list:
+			msgprint(_("Please enter BOM in the above table"))
+			return []
+
+		"""get child bom items for which production order does not exist or production order
+		qty does not match bom qty"""
+		items = frappe.db.sql("""select distinct bom_item.item_code, bom_item.bom_no,\
```
rmehta: inconsistent indentation.

rmehta:
@ishanloya. Thanks for the contribution!

You can directly create Production Orders without going with the Production Planning Tool for specific Items. How does this help?

Also, at minimum you need to add test cases and update the documentation along with this. I know might be extra work, but the contribution needs all of this to be complete.

PR #8802, opened 2017-05-12
[fix] Payments table on the invoice is showing in the print, even if IS POS option is disabled

rohitwaghchaure:
Issue
screen shot 2017-05-12 at 6 23 43 pm

Fixed
screen shot 2017-05-12 at 6 24 32 pm

### DIFF
--- a/erpnext/patches/v8_0/make_payments_table_blank_for_non_pos_invoice.py
+++ b/erpnext/patches/v8_0/make_payments_table_blank_for_non_pos_invoice.py
@@ -1,0 +1,15 @@
+# Copyright (c) 2013, Web Notes Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	frappe.reload_doctype('Sales Invoice')
+
+	frappe.db.sql("""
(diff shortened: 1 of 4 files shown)


erpnext/patches/v8_0/make_payments_table_blank_for_non_pos_invoice.py
```diff
@@ -0,0 +1,15 @@
...
+import frappe
+
+def execute():
+	frappe.reload_doctype('Sales Invoice')
+
+	frappe.db.sql("""
+		delete from 
+			`tabSales Invoice Payment` 
+		where 
+			parent in (select name from `tabSales Invoice` where is_pos = 0)
```
nabinhait: To be in safe side, add another condition for payment amount

nabinhait (changes requested):

erpnext/patches/v8_0/make_payments_table_blank_for_non_pos_invoice.py
rohitwaghchaure: Fixed

rohitwaghchaure: Revert the changes

PR #8830, opened 2017-05-15
View Attachments in portal

chdecultot:
Hi,

This PR is a proposal for the addition of a new checkbox in portal menu items to allow the display of public attachments in the portal. It will work only for order display for now. 

It is linked to PR 3289 in Frappe.

Description:
If the box is checked, the system fetches all public attachments in the doctype and displays them at the bottom of the page.

image

### DIFF
--- a/erpnext/templates/pages/order.html
+++ b/erpnext/templates/pages/order.html
@@ -86,6 +86,7 @@
 			{% endif %}
 		{% endif %}
 	</div>
+	
 {% if attachments %}
 <div class="order-item-table">
 	<div class="row order-items order-item-header text-muted">
@@ -93,21 +94,16 @@
 			{{ _("Attachments") }}
 		</div>
 	</div>
-
 	<div class="row order-items">
 		<div class="col-sm-12">
 			{% for attachment in attachments %}
 			<p class="small">
-				<a href="{{ attachment.file_url }}" target="blank">
-      {{ attachment.file_name }}
-    		</a>
+				<a href="{{ attachment.file_url }}" target="blank"> {{ attachment.file_name }} </a>
 			</p>
 			{% endfor %}
 		</div>
 	</div>
 </div>
 {% endif %}
-
 </div>
-
 {% endblock %}
--- a/erpnext/templates/pages/order.html
+++ b/erpnext/templates/pages/order.html
@@ -86,6 +86,28 @@
 			{% endif %}
 		{% endif %}
 	</div>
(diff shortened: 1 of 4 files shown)


erpnext/templates/pages/order.html
```diff
@@ -86,6 +86,28 @@
...
+			{{ _("Attachments") }}
+		</div>
+	</div>
+
+	<div class="row order-items">
+		<div class="col-sm-12">
+			{% for attachment in attachments %}
+			<p class="small">
+				<a href="{{ attachment.file_url }}" target="blank">
+      {{ attachment.file_name }}
```
rmehta: messed up whitespace?

rmehta (changes requested):

rmehta:
Implement this via Shopping Cart Settings

chdecultot:
Hi @rmehta ,

I have no problem implementing this via Shopping Cart Settings but I don't really understand why it would be better.

The idea here is to be able to display the public attachments linked to a particular doctype.
For example I can decide to display the public attachments linked to invoices, so that my customers can see a PDF of their invoice and all their contract document on the portal. 
But maybe, at the same time, I don't want to show the public attachments linked to a purchase invoice to my suppliers.

If we implement it via shopping cart settings, public attachments will be displayed in any case if the doctype uses the order template. So we loose some flexibility.
In addition if we don't activate the shopping cart (we only want to use the portal as a display account for our customers/suppliers), then could we still activate this option ? I think it will be also misleading.

Or maybe I should check Hooks.py and show the checkbox only in rows displaying the order template ?

We can also start with a global option and try to refine it later.

I am very interested in your thoughts on that matter.

Thank you!

rmehta:
@chdecultot adding this to portal items is arbitrary, no one would imagine that is where they have to go to enable attachments. Also it does not work for any other list (like Issue). So best to keep is specific to where it is implemented, for shopping cart, hence **Shopping Cart** settings seems the most relevant place to put it as of now. Maybe there should be "Portal Settings" specific to ERPNext but till we have that, Shopping Cart settings seems fine.

chdecultot:
Hi @rmehta,

Ok no problem I totally understand your point.

I will make the necessary changes today or tomorrow and push it.

Thank you!

chdecultot:
Hi @rmehta,

I have added a field in shopping cart settings as requested and modified the logic.

Please let me know if that is ok with you.

Thank you!

PR #8844, opened 2017-05-15
link the different doctype in the school module

manassolanki:


nabinhait (changes requested):
Group these buttons under "Make" dropdown

manassolanki:
I have provided buttons in the program, course and student group doctype. Now program and course are the master and student group is one time per academic year/term. Currently, most of these buttons open the list view. so if you could suggest me whether I should link these buttons contextually or in list view or open a new doc.

rmehta:
All these should be in dashboard

PR #8848, opened 2017-05-16
Batch/Serial Number entry modal

pratu16x7:
Depends on frappe/frappe#3298

Triggers on selecting item_code after warehouse.

- Adds initial batches by default, and prefills qty sorting batches by oldest.
- Serialized items have a link to text-area adder (No smart behaviour related to item qty).
- Modal item quantity updates as batches or serial numbers are added; setting it manually doesn't do anything (flow-breaking?)
batch2
Though the new row values are set, they don't seem to be triggered(indicator stays orange) even after `refresh_field("items")`. Hints?

For fetching available quantities, the modal has a prerequisite of setting the warehouse before item_code:
batch1
As such, delivery note implementation turned out to be clunky; the warehouse field is in deep inside the grid form.
screen shot 2017-05-23 at 1 54 31 pm


pratu16x7:
Added warehouse selection:
batch3

mbauskar (changes requested):
@pratu16x7,

Serial Nos can be added multiple times, also sometime after selection of serial number the serial number is added twice.

Also, on material receipt, serial number selection should be disabled

stock-entry

Available Qty in batch does not reflect in child table

batch

pratu16x7:
- Added duplicate check for serial no (and removed serial no from material receipt)
serial_no_check

- fixed available qty re-update
serial_no_check

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -83,7 +83,9 @@ def set_transfer_qty(self):
 				frappe.throw(_("Row {0}: Qty is mandatory").format(item.idx))
 			if not flt(item.conversion_factor):
 				frappe.throw(_("Row {0}: UOM Conversion Factor is mandatory").format(item.idx))
-			item.transfer_qty = flt(item.qty * item.conversion_factor, self.precision("transfer_qty", item))
+			print type(float(item.qty))
```
rmehta: print.....

rmehta (changes requested):

Issue #6468, opened 2016-09-26
Serious Bug in sales invoice, quotation, sales order and probably on purchase grids too


ioweb-gr:
When adding a product to the item grid in sales invoice and quotation it's added with the standard pricelist rate. However there are cases when you want to charge it a bit more or a bit less according to the customer. 

If you want to charge less it's fine however if you want to overcharge them on purpose the item is added with negative discount %. That's very very wrong in my opinion.

```
colums: pricelist rate    discount      rate     
case1:    50€                50%        25€
case2:    50€               -100%       100€
```

![image](https://cloud.githubusercontent.com/assets/20220341/18837999/7c7df482-840e-11e6-85f8-c0222e9aa14d.png)

If I saw such a quotation as the customer I would personally be offended and it's not something that the system should allow. 

What I would propose is first check if the discount percentage is a negative and if it is, set discount to 0 and just show the custom rate the user inserted. 
Secondly, like when adding a rate from the grid for the first time it adds the rate to the standard pricelist, it would be great if the system could ask me if I want to update the rate in the pricelist. Or better yet have a small button to add the value to the pricelist manually next to the rate so that the user isn't overcome with the many popups.

PawanMeh:
I would suggest you can write a custom JS for validating cases of discount in case you business rules don't allow it.

ioweb-gr:
I currently addressed the problem in the Print Format for the final printout as a workaround but I really think that it is a workaround and indeed this is a real problem because the very concept of a negative discount is not proper. Also a user could select the standard print format by mistake and still send it out. 

By definition a discounted price means that the final price is less than the original. 
Dictionary Definition: Discount (noun) = a deduction from the usual cost of something
A negative discount though is actually an addition to the cost of something and although it's proper math, it's not proper linguistically. 

English is a very convenient language and I can't know exactly how it would sound to a native speaker so I'd like some input on this from a native english speaking person as well but I can for sure say that if I translated it to my language it's not proper Greek to say you're discounting something and then be increasing the price. 

As it is currently, the only real choice is to hide and replace this standard field.

PR #8856, opened 2017-05-16
[fix] if rate is greater than price_list_rate, set margin instead of discount. Fixes frappe/erpnext#6468

rmehta:
When rate is below price list rate, set discount

screenshot from 2017-05-16 14-55-25

---

When rate is above price list rate, set margin

screenshot from 2017-05-16 15-00-24


rmehta:
did not find a server side equivalent.

nabinhait (approved):

PR #8869, opened 2017-05-16
Issue 8842

### LINKED ISSUE #8842: Warehouse Renaming Issue
Hi,

There's an issue when trying to rename a warehouse to something that has a hyphen like 'New - Name'. Let's assume company abbreviation is ABC. The new ID is supposed to be New - Name - ABC but instead, the system names it as New - ABC

Kindly help resolve

Thanks

tundebabzy:
Fixes #8842. I also added a test for the issue and `encode_company_abbr`


requirements.txt
```diff
@@ -1,2 +1,3 @@
 frappe
 unidecode
+six
```
nabinhait: why another library? why not using python's inbuilt `range` or `xrange`?

nabinhait (changes requested):

tundebabzy:
I forgot to remove that while working on python 3 migration. I'm removing it now

tundebabzy:
I feel it's important to write python 3 compatible code since python 3 compatibility is in view. The swap of MySQLdb with mysqlclient passes all tests. It's much easier to maintain python 3 support with six

rmehta:
@tundebabzy since you have added six in frappe, you don't need to add it here again.

erpnext/tests/test_init.py
```diff
@@ -0,0 +1,41 @@
...
+		company.company_name = "New from Existing Company For Test"
+		company.abbr = "NFECT"
+		company.default_currency = "INR"
+		company.save()
+
+		abbr = company.abbr
+
+		names = [
+			"Warehouse Name", "ERPNext Foundation India", "Gold - Member - {a}".format(a=abbr),
+		    " - {a}".format(a=abbr), "ERPNext - Foundation - India",
```
rmehta: whitespace (?)

erpnext/tests/test_init.py
```diff
@@ -0,0 +1,41 @@
...
+			"ERPNext - Foundation - India - {a}".format(a=abbr),
+			"ERPNext Foundation India - {a}".format(a=abbr), "No-Space-{a} - {a}".format(a=abbr),
+			"- Warehouse - {a}".format(a=abbr)
+		]
+
+		for i in range(len(names)):
+			enc_name = encode_company_abbr(names[i], company.name)
+			self.assertTrue(
+				enc_name == expected_names[i],
+			    "{enc} is not same as {exp}".format(enc=enc_name, exp=expected_names[i])
```
rmehta: whitespace?

rmehta (changes requested):

tundebabzy:
I'll reduce the whitespace

PR #8879, opened 2017-05-17
Fix the demo for the schools

manassolanki:


erpnext/demo/user/schools.py
```diff
@@ -8,62 +8,82 @@
...
 
 def work():
 	frappe.set_user(frappe.db.get_global('demo_schools_user'))
 	for d in xrange(20):
 		approve_random_student_applicant()
 		enroll_random_student(frappe.flags.current_date)
-	if frappe.flags.current_date.weekday()== 0:
-		make_course_schedule(frappe.flags.current_date, frappe.utils.add_days(frappe.flags.current_date, 5))	
+	# if frappe.flags.current_date.weekday()== 0:
+	# 	make_course_schedule(frappe.flags.current_date, frappe.utils.add_days(frappe.flags.current_date, 5))
```
nabinhait: Why commented?

erpnext/demo/user/schools.py
```diff
@@ -8,62 +8,82 @@
...
 	frappe.set_user(frappe.db.get_global('demo_schools_user'))
 	for d in xrange(20):
 		approve_random_student_applicant()
 		enroll_random_student(frappe.flags.current_date)
-	if frappe.flags.current_date.weekday()== 0:
-		make_course_schedule(frappe.flags.current_date, frappe.utils.add_days(frappe.flags.current_date, 5))	
+	# if frappe.flags.current_date.weekday()== 0:
+	# 	make_course_schedule(frappe.flags.current_date, frappe.utils.add_days(frappe.flags.current_date, 5))	
 	mark_student_attendance(frappe.flags.current_date)
+	# make_assessment_plan()
```
nabinhait: Why commented?

nabinhait (changes requested):

erpnext/demo/user/schools.py
manassolanki: Needed to be changed after assessment result tool. That's why commented

PR #8897, opened 2017-05-17
fixes #8784 by sorting the returned list of dictionaries

### LINKED ISSUE #8784: Chart of Accounts wrong Order
Based on Erpnext discussion https://discuss.erpnext.com/t/chart-of-accounts-wrong-order/23653
please have a look at the ordering when adding a new Chart of Account ...

tundebabzy:
#8784


rmehta:
I don't think this is the right fix. check the sort_root_accounts function

PR #8928, opened 2017-05-19
Prompt for mandatory batch number in POS

chdecultot:
Hi,

In the current POS, when an item in a batch item, users have to search for the batch number in the item search in order to select it.
If they just click on the item icon, the following error will be displayed and there is no way to add the item directly in the cart:
image

Based on the script proposed by "vivek" on the [forum](https://discuss.erpnext.com/t/pos-selecting-batch/22216), the idea here is to open a dialog and allow the user to select the corresponding batch number.
It will improve the user experience and avoid having to know all batch numbers by heart.

Here is what will be displayed when selecting a batch item:

image

Of course let me know if you think there is a better way to achieve this.

Thank you.

### DIFF
--- a/erpnext/accounts/page/pos/pos.js
+++ b/erpnext/accounts/page/pos/pos.js
@@ -1758,13 +1758,13 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		var me = this;
 		if (this.items[0].has_batch_no && !this.item_batch_no[this.items[0].item_code]) {
 			frappe.prompt([
-             {'fieldname': 'batch', 'fieldtype': 'Select', 'label': __('Batch No'), 'reqd': 1, 'options':this.batch_no_data[this.items[0].item_code]}
-                ],
-             function(values){
-             me.item_batch_no[me.items[0].item_code] = values.batch;
-                },
-             __('Select Batch No'))
-		}
+				{'fieldname': 'batch', 'fieldtype': 'Select', 'label': __('Batch No'), 'reqd': 1, 'options':this.batch_no_data[this.items[0].item_code]}
+				],
+				function(values){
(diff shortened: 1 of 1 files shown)


erpnext/accounts/page/pos/pos.js
```diff
@@ -1757,9 +1757,13 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 	mandatory_batch_no: function () {
 		var me = this;
 		if (this.items[0].has_batch_no && !this.item_batch_no[this.items[0].item_code]) {
-			frappe.throw(__(repl("Error: Batch no is mandatory for item %(item)s", {
-				'item': this.items[0].item_code
-			})))
+			frappe.prompt([
+             {'fieldname': 'batch', 'fieldtype': 'Select', 'label': __('Batch No'), 'reqd': 1, 'options':this.batch_no_data[this.items[0].item_code]}
```
rmehta: whitespace :cry:

Please try and fix these issues atleast.

what editor you are using? try using VS Code - it will autodetect whether the file uses tabs or spaces.

rmehta (changes requested):

erpnext/accounts/page/pos/pos.js
chdecultot: Hi @rmehta ,

Sorry about these, it only shows up when pushing to Github.

I am using Atom as an editor.

I will try to avoid it in the future...

almeidapaulopt: @chdecultot Change tab type to hard in atom preferences / editor. It helps too if you enable "show invisibles"

chdecultot: Thanks a lot for your help Guys! :+1:

PR #8936, opened 2017-05-19
Added flt to convert string to float, fixes erpnext/issues#8831

MCD-50:
There was an error while multiplying payment amount with conversion rate. Since the value of amount was of type string it was flagging can't multiply sequence by non-int of type 'float'. Converted payment amount to floating number before multiplying.


nabinhait (changes requested):
Rebase from hotfix branch

Issue #8436, opened 2017-04-13
Allow updates to raw material and manufactured quantity in production order


sathishpy:
Raised as requested from https://discuss.erpnext.com/t/updating-manufactured-quantity-in-production-order/22377/1

In summary, some verticals like box manufacturing won't be able to stick to the listed quantity in BOM. As a result, they are likely to over produce than required manufacturing quantity. This request is to allow updating raw material and manufactured quantity based on a new setting or current 'Over Production' setting.

sathishpy:
I don't see this issue anymore in my usecase, however I don't think the issue is fixed. I took a quick look and the code is still the same, so thinking the changes done at my custom app is taking a different path for production order, there by masking this issue. I will close the issue as it is not more affecting my usecase.

tecnofoodmx:
Same issue.
We use ERP Next in a food manufacturing business.
Generally we have variable "raw material losses", which results in a slight variation on the final units produced vs the planned. 
In the case thre's an excess we cannot register the exceeded units. 
On the other hand, when we produce a lower amount vs planned quantity, the production order cannot be closed. Also, the raw materials are consumed proportionally to the amounts produced according to BOM. I would like to be able to assign the remaining raw materials to "material loss" in order to keep track of losses and precise tracking of stock levels. instead of using "stock recinciliation" every time.

Thanks for your support!

federicocalvo:
Hi @tecnofoodmx , i have the same situtation whit a food industry. 

at the end of production, more or less may have been produced than originally established.

First of all, the overproduction option doesn't work for me. And secondly, I have the same problem when less is produced, they are open.

Were you able to work it out somehow?

Regards!

PR #8943, opened 2017-05-20
https://github.com/frappe/erpnext/issues/8436: Allow updates to raw m…

sathishpy:
…aterial and manufactured quantity in production order
Consider over production allowance in production order and stock entry


sathishpy:
well, this is my first fix, so getting used to the system. I have made the changes and done basic testing with and without 'over production allowance' in manufacturing settings

sathishpy:
Closing to re-open

rmehta:
@sathishpy thanks for the contribution!

Can you try adding a test case, since the community is concerned about stability we need to enforce test cases more rigidly.

Closing this for now, please reopen after adding the test case.

PR #8957, opened 2017-05-22
fix the issue where image not saved even after attaching file

manassolanki:
In the lead doctype image isn't saved.

may-22-2017 12-56-28


erpnext/crm/doctype/lead/lead.py
```diff
@@ -43,7 +43,8 @@ def validate(self):
 			if self.email_id == self.contact_by:
 				frappe.throw(_("Next Contact By cannot be same as the Lead Email Address"))
 
-			self.image = has_gravatar(self.email_id)
+			if self.get("__islocal"):
```
nabinhait: Should it be like `if self.is_new() or not self.image:`?

nabinhait (changes requested):

nabinhait (approved):

PR #8966, opened 2017-05-22
Minor change: Alert to msgprint

nick9822:


erpnext/stock/doctype/material_request/material_request.js
```diff
@@ -21,7 +21,7 @@ frappe.ui.form.on("Material Request Item", {
 	"qty": function(frm, doctype, name) {
 			var d = locals[doctype][name];
 			if (flt(d.qty) < flt(d.min_order_qty)) {
-				alert(__("Warning: Material Requested Qty is less than Minimum Order Qty"));
+				msgprint(__("Warning: Material Requested Qty is less than Minimum Order Qty"));
```
netchampfaris: Please use `frappe.msgprint` instead of `msgprint`, global methods like these will be deprecated.

netchampfaris (changes requested):

erpnext/stock/doctype/material_request/material_request.js
nick9822: Thanks for the information. I will apply that.

nick9822:
@netchampfaris 
Please review new commit pushed as dictated.

nabinhait (approved):

PR #8968, opened 2017-05-22
fixes #8941: Better error message for duplicate items

### LINKED ISSUE #8941: Better error message for the duplicate items
If in the transaction, an item is selected multiple times, then error message related to the same is as following: "same item has been entered multiple times"

screen shot 2017-05-20 at 11 13 41 am

We can improve this error message by indicating Item Code and Row No., so that user can make correction faster.

tundebabzy:


mbauskar:
@tundebabzy,

can you add the before and after screenshot? It will be easier to review the PR

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -244,6 +244,23 @@ def set_conversion_rate(self):
...
+			seen_add = seen.add
+			for item in lst:
+				if item.item_code in seen or seen_add(item.item_code):
+					yield item
+
+		def get_duplicate_item(lst):
+			try:
+				item = get_duplicates(lst).next()
+				if item:
+					return item
```
mbauskar: instead of returning single item return list of duplicate items, e.g. if raw material has items A, B, B, C, C then return list [B, C] as multiple items

mbauskar: instead of returning single item return list of duplicate items, e.g. if raw material has items A, B, B, C, C then return list [B, C] as multiple items

mbauskar (changes requested):

tundebabzy:
@mbauskar Sorry about that. Noted.
**Before**
screenshot from 2017-05-25 11-47-20

**After**
screenshot from 2017-05-25 11-46-24

tundebabzy:
Changed to show all duplicates. They are displayed in an `ol`.
** Before**
screenshot from 2017-05-25 11-47-20

**After**
screenshot from 2017-05-25 12-25-20

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -252,10 +260,16 @@ def validate_materials(self):
...
-		unique_chk_list = set(check_list)
-		if len(unique_chk_list)	!= len(check_list):
-			frappe.throw(_("Same item has been entered multiple times."))
+			check_list.append(m)
+
+		duplicate_items = list(get_duplicates(check_list))
+		if duplicate_items:
+			li = []
+			for i in duplicate_items:
+				li.append("<li>{0} on row {1}".format(i.item_code, i.idx))
```
mbauskar: close the li tag `<li>{0} on row {1}</li>`

Also can we keep the message simple? Just join the list of items with <br>?

mbauskar (changes requested):

erpnext/manufacturing/doctype/bom/bom.py
tundebabzy: ok. That means no need for the `<li>` tags?

tundebabzy:
screenshot from 2017-06-06 23-09-32

PR #8974, opened 2017-05-23
Paid amount + Write Off Amount can not be greater than Grand Total in return POS

nabinhait:


mbauskar (commented):
need to delete the Grading Structure, Grade Interval JSON files
update: deleted and pushed changes https://github.com/frappe/erpnext/pull/8991

mbauskar (approved):

PR #8975, opened 2017-05-23
Milestones on Project Gantt

almeidapaulopt:
Add milestones on gantt
milestone
milestone1

### DIFF
--- a/erpnext/projects/doctype/task/task.json
+++ b/erpnext/projects/doctype/task/task.json
@@ -1,5 +1,6 @@
 {
  "allow_copy": 0, 
+ "allow_guest_to_view": 0, 
  "allow_import": 1, 
  "allow_rename": 1, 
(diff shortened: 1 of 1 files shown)


netchampfaris:
@almeidapaulopt We also need to add validation for milestones tasks, since other tasks cannot depend on Milestones.

almeidapaulopt:
hi @netchampfaris,

I'm not following you in why shouldn't be dependant. I'll try to make my point:

I have a several tasks in almost all my project projects that depends on milestones. That's the way that all reference project manager software work (ms project etc) and that a normal behaviour in project management.

So you can have a milestone that depend from tasks and tasks that depends from milestones.

For example: building a house, the task "Make roof" depends o the milestone "stucture done" that depends the tasks "foundations", "structure 1º floor" and "structure 2º floor".

netchampfaris (approved):

PR #8977, opened 2017-05-23
fixes #8965: Get item inside Request for Quote

### LINKED ISSUE #8965: Get item inside Request for Quote
When we create an RFQ we save it for later so we can add more items to it

Now, when we  open the RFQ and choose "Get Items from Material Request" then SAVE options has gone, and we can only submit. But we need to be able to save again before submitting

Save is available:

image

Choose get items from material request

image

Save has now gone

image

tundebabzy:


nabinhait (changes requested):
I think it should be fixed in `sync.js`.
Otherwise, we will get the issue in all the places it is used.

tundebabzy:
Actually, the issue is everywhere. 
For example:
1. Create an invoice for a customer for and save it pending when manager returns to approve it.
2. Create a sales order for the same customer for emergency supplementary supplies.
3. Open the saved invoice and add items from the new sales order.
The 'Save' button will be missing at this point.

That's the reason why I made the change in `utils.js` as long as information is gotten from the 'Get items from` button, the form should be flagged as dirty

nabinhait:
@tundebabzy Yes, it absolutely does the job while mapping. But I am thinking, we should make the form `dirty` whenever we call `frappe.model.sync()`, for any usecase. So, may be it is better to fix it directly in `frappe/model/sync.js`

tundebabzy:
Oh ok. I'll check that out

tundebabzy:
fix more appropriate on frappe. See PR on frappe/frappe#3382

PR #8980, opened 2017-05-23
[fix] add round off difference to last row in landed cost voucher

sagarvora:
Before:
selection_014

After:
selection_016

### DIFF
--- a/erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.js
+++ b/erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.js
@@ -102,9 +102,17 @@ erpnext.stock.LandedCostVoucher = erpnext.stock.StockController.extend({
 				total_item_cost += flt(d[based_on])
 			});
 
+			var total_charges = 0.0;
 			$.each(this.frm.doc.items || [], function(i, item) {
 				item.applicable_charges = flt(item[based_on]) * flt(me.frm.doc.total_taxes_and_charges) / flt(total_item_cost)			
+				item.applicable_charges = roundNumber(item.applicable_charges, 4)
+				total_charges += item.applicable_charges
 			});
+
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.js
```diff
@@ -102,9 +102,17 @@ erpnext.stock.LandedCostVoucher = erpnext.stock.StockController.extend({
 				total_item_cost += flt(d[based_on])
 			});
 
+			var total_charges = 0.0;
 			$.each(this.frm.doc.items || [], function(i, item) {
 				item.applicable_charges = flt(item[based_on]) * flt(me.frm.doc.total_taxes_and_charges) / flt(total_item_cost)			
+				item.applicable_charges = roundNumber(item.applicable_charges, 4)
```
nabinhait: use `flt` and `precision` of the field

erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.js
```diff
@@ -102,9 +102,17 @@ erpnext.stock.LandedCostVoucher = erpnext.stock.StockController.extend({
...
+				total_charges += item.applicable_charges
 			});
+
+			if (total_charges != this.frm.doc.total_taxes_and_charges){
+				var diff = this.frm.doc.total_taxes_and_charges - total_charges
+				this.frm.doc.items.slice(-1)[0].applicable_charges += diff
```
nabinhait: why slice? It can be like `this.frm.doc.items[-1]. applicable_charges `

nabinhait (changes requested):

nabinhait (changes requested):
Add a test case to assure validation is not raised for rounding issue

PR #8989, opened 2017-05-24
[fix] add round off difference to last row in landed cost voucher

sagarvora:


erpnext/stock/doctype/landed_cost_voucher/test_landed_cost_voucher.py
```diff
@@ -133,7 +133,24 @@ def test_landed_cost_voucher_for_serialized_item(self):
...
+			"item_code": "_Test Item",
+			"warehouse": "_Test Warehouse - _TC",
+			"cost_center": "_Test Company - _TC",
+			"qty": 5,
+			"rate": 50
+		})
+		pr.submit()
+
+		submit_landed_cost_voucher("Purchase Receipt", pr.name, 123.22)
+
```
nabinhait: Add a `assertEqual` statement as well, to compare the distributed charges value.

nabinhait (changes requested):

erpnext/stock/doctype/landed_cost_voucher/test_landed_cost_voucher.py
```diff
@@ -133,7 +133,27 @@ def test_landed_cost_voucher_for_serialized_item(self):
...
+			"cost_center": "_Test Company - _TC",
+			"qty": 5,
+			"rate": 50
+		})
+		pr.submit()
+
+		lcv = submit_landed_cost_voucher("Purchase Receipt", pr.name, 123.22)
+		
+		self.assertEquals(lcv.items[0].applicable_charges, 41.07)
+		self.assertEquals(lcv.items[2].applicable_charges, 41.07999999999999)
```
nabinhait: It should be 42.08. What is the precision of the applicable_charges field?

nabinhait (changes requested):

nabinhait (approved):

PR #8991, opened 2017-05-24
[minor] removed the Grade Interval and Grading Structure doctype

mbauskar:
- Deleted the `Grading Structure` and `Grade Interval` files
- Grading Structure has been renamed to Grading Scale
- Grade Interval has been renamed to Grading Scale interval


nabinhait (changes requested):
You need to write a patch as well to delete those doctypes from db

mbauskar:
@nabinhait added patch to delete the doctyps

nabinhait (approved):

PR #8995, opened 2017-05-24
Incorrect item translations

chdecultot:
Hi,

This is an amendment to the previous commit fixing translations.

If "Qty", "Price" and "Disc" are translated in the id also, then all three buttons are not able to work correctly anymore.

Therefore only the displayed text must be translated.

This leads to a question: how will these three items be picked up by the translator if they are not flagged directly as translatable ? Can we add them manually to the translation file ?

Thank you.


nabinhait (changes requested):
Thanks for the fix. Can you send this PR to hotfix branch after rebasing from hotfix branch?

chdecultot:
Humm... tried to rebase and send but it added a few other commits in the PR.

I prefer to close it and I will send a new one in a few minutes.

PR #8997, opened 2017-05-24
Allow multiple item in purchase receipt linked with same purchase order

manassolanki:


nabinhait (changes requested):
The same is also applicable for Purchase Invoice, Delivery Note and Sales Invoice.

manassolanki:
Purchase Invoice, Delivery Note, and Sales Invoice already have this functionality. Trying to figure any specific reason why earlier this is not included in the purchase receipt, but all the validations seem to working fine.

nabinhait (approved):

Issue #9000, opened 2017-05-24
Job Application - Won't save without Email ID


umairsy:
In the Job Applicant, Email ID is not a mandatory field. If we try saving Job Applicant without Email ID, then this is the validation message.

![screen shot 2017-05-24 at 5 57 55 pm](https://cloud.githubusercontent.com/assets/910238/26403394/b4faa01c-40ab-11e7-90ed-ee450fde3d41.png)

PR #9009, opened 2017-05-24
Email Id mandatory for job applicant

manassolanki:
https://github.com/frappe/erpnext/issues/9000


umairsy:
What if Jon Application are being auto-created via Email Account? What is someone is recruiting a candicate who doesn't have email id, like for the position of office helpers?

Email ID field should be non-mantodatory, and there should be no error message as indicated in #9000

manassolanki:
@umairsy Done, check email id only if provided in the form.

nabinhait:
@manassolanki send it to hotfix

PR #9034, opened 2017-05-26
Document Change Notice

bcornwellmott:
This is a document change notice doctype, which allows a user to indicate that certain documents have changed and have those changes approved by multiple people. 

This is different from workflow because the main purpose it to keep track of several documents changing at once and explain what changed and why. As well, the approvals are customizable each time, as opposed to workflow which is document based.

The document automatically creates todo tasks for each department manager that needs to approve.

This type of change request / notice is essential for a quality management system.


bcornwellmott:
I think I need some help with this. I'm having two major issues:

- [ ] Employee number fills in in unwanted areas
- [ ] Can't change visibility of grid fields without Reloading page

bcornwellmott:
I'm going to hold off on this - better idea on how to implement it

PR #9027, opened 2017-05-25
Fixes in the patch

manassolanki:

PR #9035, opened 2017-05-26
Reloaded Student Group Instructor and Student Group Student

nabinhait:


erpnext/patches/v8_0/merge_student_batch_and_student_group.py
```diff
@@ -3,15 +3,15 @@
...
-from frappe.model.mapper import get_mapped_doc
+from frappe.model.utils.rename_field import rename_field
 
 
 def execute():
 
 	# for converting student batch into student group
-	frappe.reload_doctype("Student Group")
-
+	for doctype in ("student_group", "student_group_student", "student_group_instructor"):
```
manassolanki: We can also add "Student Attendance" here instead of in the end.

manassolanki:
Done this in https://github.com/frappe/erpnext/pull/9027

PR #9036, opened 2017-05-26
set student roll no if not given

manassolanki:


mbauskar:
@manassolanki screenshots?

Can you also add the test-case to check roll no is set or not

mbauskar (changes requested):
@manassolanki screenshots?

Can you also add the test-case to check the roll no

manassolanki:
@mbauskar yeah sure... 
Closing since have to merge in hotfix

Issue #8963, opened 2017-05-22
Batch selection not filtering based on Item in Purchase Invoice with Update stock checked


mdwala:
In "Purchase Invoice" with checked "Update Stock",
Batches showing for all items instead of specific selected item.

Once above done, need to remove this(as not require anymore)
"BATCHID is not a valid Batch Number for Item ITEMNAME"

PR #9038, opened 2017-05-26
[minor] added filter for batch no field in purchase transaction

mbauskar:
fixes for https://github.com/frappe/erpnext/issues/8963

`before`
batch-before

`after`
batch-after

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -98,6 +98,13 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 			frm.cscript.calculate_taxes_and_totals();
 		});
+
+		var me = this;
+		if(this.frm.fields_dict["items"].grid.get_field('batch_no')) {
+			this.frm.set_query("batch_no", "items", function(doc, cdt, cdn) {
+				return me.set_query_for_batch(doc, cdt, cdn)
+			});
+		}
 	},
 	onload: function() {
 		var me = this;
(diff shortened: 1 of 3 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -1137,4 +1144,26 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+			frappe.throw(__("Please enter Item Code to get batch no"));
+		} else {
+			filters = {
+				'item_code': item.item_code,
+				'posting_date': me.frm.doc.posting_date || frappe.datetime.nowdate(),
+			}
+			if(item.warehouse) filters["warehouse"] = item.warehouse
+
+			return {
+				query : "erpnext.controllers.queries.get_batch_no",
```
nabinhait: `get_batch_no`  only returns batches if the batch has qty. But in Purchase receipt, we need to allow all batches, even if it has no qty.

https://github.com/frappe/erpnext/blob/master/erpnext/controllers/queries.py#L267

nabinhait (changes requested):

erpnext/public/js/controllers/transaction.js
mbauskar: @nabinhait,

filtered the batch no based on item_code if doctype is Purchase Receipt

erpnext/public/js/controllers/transaction.js
```diff
@@ -1137,4 +1144,30 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+
+	set_query_for_batch: function(doc, cdt, cdn) {
+		// Show item's batches in the dropdown of batch no
+
+		var me = this;
+		var item = frappe.get_doc(cdt, cdn);
+
+		if(!item.item_code) {
+			frappe.throw(__("Please enter Item Code to get batch no"));
+		} else if (doc.doctype == "Purchase Receipt") {
```
nabinhait: also applicable to Purchase Invoice with update_stock checked.

nabinhait (changes requested):

nabinhait (approved):

Issue #8711, opened 2017-05-05
Prepare ERPNext for GST India


revant:
**Phase 1**
- [x] State Doctype to have state code and state name https://bitbucket.org/mntechnique/gst_api/raw/3e9af570eb86600bd0a81c8ebad2c103dcb88c63/gst_api/fixtures/gst_india_state.json
- [x] Add link to State doctype in Address, Warehouse Address
   - [ ] Copy Address from selected link to address_display (small text/html)
- [x] Add HSN Code Doctype to record Category and code https://bitbucket.org/mntechnique/gst_api/raw/3e9af570eb86600bd0a81c8ebad2c103dcb88c63/gst_api/fixtures/hsn_code.json 
- [x] Link to HSN code in Item master
- [x] PAN to be added for all customers / suppliers https://github.com/frappe/erpnext/issues/8373
- [x] GSTIN to be added for company, Address and Warehouse
- [ ] Default tax account and template for SGST/CGST/IGST/UGST for Intra/Inter State
- [ ] In transactions, based on based on first 2 digits of source and target GSTIN, identify intra/inter state and accordingly set tax template.
- [x] Prepare data for GSTR1 Sales Invoice (Item-wise Sales Register)
- [x] Prepare data for GSTR2 Purchase Invoice (Item-wise Purchase Register)
- [ ] Prepare data for GSTR6 Input service distributors (Purchase Invoice for services)
- [ ] Prepare data for GSTR8 Annual Returns
- [ ] Reconciliation Tool for GSTR2A and Item-wise Purchase Register
- [ ] Manual Export of data via reports so that users can export xls, csv and submit it to their Auditors/Tax consultants
- [ ] GST on charges like transport, packing etc (Maintain Item for such charges as it has HSN codes and should be submitted in GSTR1 as line item)
- [x] Validation for GSTIN
     - should be alphanumeric, 15 digits and first 2 digits should be state code
- [x] Add HSN Code on Sales/Purchase Invoice Item and item-wise sales/purchase register
- [ ] GST for stock transfer
    - via Stock Entry
    - Introduce tax table, item-wise tax calculation
    - Fetch Addresses from Warehouse
    - Validate that source and target warehouse belong to one state respectively

- [x] Patch
    - Remove standard tax_id field from Supplier and create custom field where data exists
    - Create State record based on value in Address
    - Remove address fields from warehouse and patch to create address records

**Phase 2**
[...]

PR #9057, opened 2017-05-27
[fix] #8711

PawanMeh:
1. Added GST Fields to Company & PAN to Supplier, Customer
2. Modified Purchase register and Sales Register to show PAN fields
3. Added HSN Doc type
4. Added HSN to Item Master
5. Added State Code field to Warehouse address (State Code doc type sent as separate PR in Frappe Repo since Address doc type belongs to Geo module, if this needs to be merged under ERPNext module let me know)

This is initial PR related to GST


nabinhait:
cherry-picked some commits

Issue #9061, opened 2017-05-29
Update Warranty and AMC status of Serial No automatically after expiry date


nabinhait:
Call `set_maintenance_status` function from scheduler.

PR #9066, opened 2017-05-29
updating serial no status from scheduler

manassolanki:
#9061

### DIFF
--- a/erpnext/stock/doctype/serial_no/serial_no.py
+++ b/erpnext/stock/doctype/serial_no/serial_no.py
@@ -324,3 +324,12 @@ def update_serial_nos_after_submit(controller, parentfield):
 						update_rejected_serial_nos = False
 						if accepted_serial_nos_updated:
 							break
+
+def update_maintenance_status():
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/serial_no/serial_no.py
```diff
@@ -324,3 +324,13 @@ def update_serial_nos_after_submit(controller, parentfield):
 						update_rejected_serial_nos = False
 						if accepted_serial_nos_updated:
 							break
+
+def update_maintenance_status():
+	last_date = frappe.utils.add_days(frappe.utils.nowdate(), -1)
```
nabinhait: Why finding last_date? Instead use `<` sign in query

nabinhait (changes requested):

erpnext/stock/doctype/serial_no/serial_no.py
manassolanki: In the last logic, updating docs having the last date as expiry. Changed now.

nabinhait (approved):

Issue #8044, opened 2017-03-16
Payment Entry - Draft Sales Invoices should not be selectable


umairsy:
![screen shot 2017-03-16 at 3 19 44 pm](https://cloud.githubusercontent.com/assets/910238/23990512/54ddc6b2-0a5c-11e7-8a92-c54f567d90ee.png)

PR #9072, opened 2017-05-29
Payment Entry - Draft Sales Invoices should not be selectable #8044

ashwinisave35:


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -78,6 +78,15 @@ frappe.ui.form.on('Payment Entry', {
 				filters: { "name": ["in", doctypes] }
 			};
 		});
+
+		frm.set_query("reference_name", "references", function () {
+			return {
+				filters: {
+					docstatus: 1,
+					title: frm.doc.party
```
mbauskar: @ashwinisave35,

filter document based on `party type` instead of `title`. e.g. if Party type is Customer then filter based on customer

mbauskar (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -78,6 +78,24 @@ frappe.ui.form.on('Payment Entry', {
...
+				}
+			} else if (frm.doc.party_type=="Supplier") {
+				return {
+					filters: {
+						docstatus: 1,
+						supplier: frm.doc.party
+					}
+				}
+			}
+		});
```
mbauskar: @ashwinisave35,

This code can be reduced.
e.g.

```
frm.set_query("reference_name", "references", function () {
	party_field = frappe.model.scrub(frm.doc.party_type)
	return {
		filters: {
			docstatus: 1,
			party_field: frm.doc.party
		}
	}
});
```

mbauskar:
@ashwinisave35, Please push the changes and reopen the PR

PR #9097, opened 2017-05-31
The employee is added to dropdown for  trial_balance_for_party.

sumitdas4u:
The employee is added to dropdown for  trial_balance_for_party. it's easy to find the account balance of employee after paying the expense 

image

### DIFF
--- a/erpnext/accounts/report/trial_balance_for_party/trial_balance_for_party.py
+++ b/erpnext/accounts/report/trial_balance_for_party/trial_balance_for_party.py
@@ -19,7 +19,7 @@ def execute(filters=None):
 	return columns, data
 	
 def get_data(filters, show_party_name):
-	party_name_field = "customer_name" if filters.get("party_type")=="Customer" else "supplier_name"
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/trial_balance_for_party/trial_balance_for_party.py
```diff
@@ -19,7 +19,7 @@ def execute(filters=None):
 	return columns, data
 	
 def get_data(filters, show_party_name):
-	party_name_field = "customer_name" if filters.get("party_type")=="Customer" else "supplier_name"
+	party_name_field = "customer_name" if filters.get("party_type")=="Customer" else "supplier_name" if filters.get("party_type")=="Supplier" else "Employee"
```
rmehta: Please add a multi-line `if` statement since there are more than one ifs, or make a dict of fieldnames and use the correct fieldname.

rmehta (changes requested):
Closing for now, please fix and re-open

PR #9103, opened 2017-05-31
[Fix] Fieldtype in Serial No Rename

akshay83:
In some DocTypes fieldtype for serial_no is "Small Text", in some it is "Text"

### DIFF
--- a/erpnext/stock/doctype/serial_no/serial_no.py
+++ b/erpnext/stock/doctype/serial_no/serial_no.py
@@ -160,7 +160,7 @@ class SerialNo(StockController):
 	def after_rename(self, old, new, merge=False):
 		"""rename serial_no text fields"""
 		for dt in frappe.db.sql("""select parent from tabDocField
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/serial_no/serial_no.py
```diff
@@ -160,7 +160,7 @@ def before_rename(self, old, new, merge=False):
 	def after_rename(self, old, new, merge=False):
 		"""rename serial_no text fields"""
 		for dt in frappe.db.sql("""select parent from tabDocField
-			where fieldname='serial_no' and fieldtype='Text'"""):
+			where fieldname='serial_no' and fieldtype like '%%Text%%'"""):
```
mbauskar: @akshay83, can you add the `fieldtype in ('Text', 'Small Text')` condition instead?

mbauskar (changes requested):

mbauskar:
@akshay83, thanks for the PR, can you send the PR to hotfix branch ??

akshay83:
Ok

PR #9106, opened 2017-06-01
Set website route field as No Copy field

netchampfaris:
Two Item's should not have the same route.


erpnext/stock/doctype/item/item.json
```diff
@@ -3143,8 +3143,8 @@
  "issingle": 0, 
  "istable": 0, 
  "max_attachments": 1, 
- "modified": "2017-05-15 11:49:47.525859", 
- "modified_by": "Administrator", 
+ "modified": "2017-05-31 14:10:19.976457", 
+ "modified_by": "faris@erpnext.com",
```
mbauskar: @netchampfaris,

can you change the modified by to Administrator?

mbauskar (changes requested):

mbauskar (approved):

nabinhait (approved):

PR #9114, opened 2017-06-01
Fix Empty Doctype naming series error when updationg

ahmadRagheb:
when update naming series list for empty doctype naming series it show erros like in the picture :+1: 

  File "/home/ahmad/benches-workspace/testing-benchs/test1/apps/erpnext/erpnext/setup/doctype/naming_series/naming_series.py", line 124, in check_duplicate
    options = self.scrub_options_list(self.set_options.split("\n"))
AttributeError: 'NoneType' object has no attribute 'split'

### DIFF
--- a/erpnext/setup/doctype/naming_series/naming_series.py
+++ b/erpnext/setup/doctype/naming_series/naming_series.py
@@ -14,6 +14,7 @@ from frappe.permissions import get_doctypes_with_read
 class NamingSeriesNotSetError(frappe.ValidationError): pass
 
 class NamingSeries(Document):
+	x=""
 	def get_transactions(self, arg=None):
 		doctypes = list(set(frappe.db.sql_list("""select parent
 				from `tabDocField` df where fieldname='naming_series'""")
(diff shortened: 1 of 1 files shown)


ahmadRagheb:
naming

erpnext/setup/doctype/naming_series/naming_series.py
```diff
@@ -131,9 +136,13 @@ def validate_series_name(self, n):
 			throw(_('Special Characters except "-", "#", "." and "/" not allowed in naming series'))
 
 	def get_options(self, arg=None):
+		global x
```
mbauskar: use the local variables instead of global, please use the readable variable name

erpnext/setup/doctype/naming_series/naming_series.py
```diff
@@ -14,6 +14,7 @@
 class NamingSeriesNotSetError(frappe.ValidationError): pass
 
 class NamingSeries(Document):
+	x=""
```
mbauskar: variable name should be meaningful and readable

mbauskar:
Please push the changes and reopen the PR

PR #9120, opened 2017-06-02
Fetch query for assessment groups

manassolanki:


erpnext/schools/doctype/assessment_plan/assessment_plan.js
```diff
@@ -6,7 +6,18 @@ cur_frm.add_fetch("examiner", "instructor_name", "examiner_name");
 cur_frm.add_fetch("supervisor", "instructor_name", "supervisor_name");
 
 frappe.ui.form.on("Assessment Plan", {
-        refresh: function(frm) {
+    onload: function(frm) {
+        frm.set_query("assessment_group", function(doc, cdt, cdn) {
+            return{
+                query: "erpnext.schools.doctype.assessment_plan.assessment_plan.get_assessment_group",
```
nabinhait: No need to call this method, just pass the filters.

nabinhait (changes requested):

mbauskar (approved):

PR #9122, opened 2017-06-02
Serial datatype

akshay83:
As suggested sending in on  Hotfix branch 

In some DocTypes fieldtype for serial_no is "Small Text", in some it is "Text".


mbauskar:
opened a new PR https://github.com/frappe/erpnext/pull/9126

PR #9134, opened 2017-06-04
Fix Supplier Addresses And Contacts  label to accept translation

ahmadRagheb:
Supplier Addresses And Contacts doesn't accept translation


mbauskar:
@ahmadRagheb,

We do not translate the **name** field as it maps the report instead translate the label

PR #9136, opened 2017-06-05
[minor] Taxes and Totals.js

fahimalizain:
Some minor clearing and checking the absolute value for `total_amount_to_pay` so this can be used as a base class for custom return transactions (like a custom sales return)


faztp12:
Some of them are fixed already.

PR #9137, opened 2017-06-05
[enhancement] Allow on Submit for Sales Order Customer PO and Customer PO Date

javierwonghf:
Allow on Submit for Sales Order Customer PO and Customer PO Date

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.json
+++ b/erpnext/selling/doctype/sales_order/sales_order.json
@@ -14,6 +14,7 @@
  "engine": "InnoDB", 
  "fields": [
   {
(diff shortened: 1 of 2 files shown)


mbauskar:
@bohlian,

Can you please explain the use case?

You can also make the customer po field can be also set as Allow after submit from Customize form

bohlian:
@mbauskar in many parts of the world, customers send POs to businesses alot later than they should. In many of these cases customers request that their orders be processed first and therefore SOs need to be processed first before a Customer's PO is received. 

You can't change "Allow on Submit" properties for standard fields. You get this error.

image

I remember someone else requesting for this although I can't seem to find the issue anymore.

PR #9159, opened 2017-06-06
GST India basics, #8711

nabinhait:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -817,6 +817,37 @@
    "bold": 0, 
    "collapsible": 0, 
    "columns": 0, 
+   "fieldname": "supplier_gstin",
```
rmehta: should be custom field!!

rmehta (changes requested):

erpnext/accounts/report/item_wise_purchase_register/item_wise_purchase_register.py
```diff
@@ -35,7 +39,8 @@ def execute(filters=None):
 
 		expense_account = d.expense_account or aii_account_map.get(d.company)
 		row = [d.item_code, d.item_name, d.item_group, d.parent, d.posting_date, d.supplier,
-			d.supplier_name, d.credit_to, d.mode_of_payment, d.project, d.company, d.purchase_order,
+			d.supplier_name, d.supplier_gstin, supplier_details.get(d.supplier, frappe._dict()).tax_id,
```
rmehta: how is this relevant to other countries??? We need separate reports for GST that will only be installed if it is India

Issue #7302, opened 2016-12-22
Sales Invoice: Payments are not being recorded when I use advance payments


juliancms:
Payments are not being recorded in the general ledger for the following scenario:
1 Creation of Journal Entry as advance payment
2 Creation of Sales Invoice without clicking "Is POS" and click on "Get Advances Received". Doing this the option to add payments appears.
3 Add a cash payment, save and submit.
Then I check in the General Ledger and only the advance payment is registered but the payment does not appears.

In the permissions of "Payments Section" says: eval:doc.is_pos===1||(doc.advances && doc.advances.length>0) . If I understand well that means it should work with advance payments greater than 0.

rmehta:
@Arundhatii can you replicate?

tundebabzy:
If `isPos` was not selected, you shouldn't have been able to add the cash payment as ERPNext is designed such that you can only enter a payment amount when `isPos` is selected.....or this has been fixed.

I tried to replicate the issue but I was not able to successfully add a payment before clicking save as expected (the payment table should not even appear since it would not be honoured). Closing this for now.

PR #9174, opened 2017-06-07
hides `payments` field when `is_pos` is checked

tundebabzy:
Related to  #7302 where although `is_pos` is not selected, the Payments table is still visible. The sales invoice does reject anything that is entered into the payment table so no harm is done but this is a source of confusion.

### Before
peek 2017-06-07 10-54

### After
peek 2017-06-07 11-02


faztp12:
but the whole section depends on is_pos already.. why wont this work then ? 😯

https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/sales_invoice/sales_invoice.json#L2603

faztp12:
ah ok @tundebabzy, you have advances created..

mbauskar (approved):

Issue #9171, opened 2017-06-07
Customer_Name when creating Quotation from Lead.


Arundhatii:
Reported via https://frappe.erpnext.com/desk#Form/Issue/WN-SUP25584

(1) When you MAKE >> QUOTATION from the Lead form, it inserts the lead field "lead_name" value into the "customer_name" field on the quotation.

(2) However, when you create a new quotation from the Quotation List screen, and choose the Lead >>Opportunity

 from within the quotation form, it inserts the lead field "company_name" value into the "customer_name" field on the quotation.

Preferably we want the "company_name" value to reflect on quotations (not the "lead_name" value. But we want to be able to MAKE-QUOTATIONS directly from the lead form.

![lead to quotation](https://user-images.githubusercontent.com/20164097/26873496-a54a0ca6-4b97-11e7-85e9-b91743d626b4.gif)

manassolanki:
Fixed

PR #9188, opened 2017-06-08
make quotation against lead's company name

manassolanki:
https://github.com/frappe/erpnext/issues/9171

### DIFF
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -150,8 +150,7 @@ def make_quotation(source_name, target_doc=None):
 		{"Lead": {
 			"doctype": "Quotation",
 			"field_map": {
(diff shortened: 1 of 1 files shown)


erpnext/crm/doctype/lead/lead.py
```diff
@@ -151,7 +151,7 @@ def make_quotation(source_name, target_doc=None):
 			"doctype": "Quotation",
 			"field_map": {
 				"name": "lead",
-				"lead_name": "customer_name",
+				"lead_name": "company_name",
```
nabinhait: In quotation, there is no field called "company_name".

The correct fix will be, remove the line `"lead_name": "customer_name",`. `set_missing_values` does the mapping correctly.

nabinhait (changes requested):

Issue #9180, opened 2017-06-08
Payment Entry - Latest Exchange Rate not fetched


addymuliady:
ERPNext: v8.0.44 (master)
Frappe Framework: v8.0.64 (master)

Currency Exchange Rate Setup

![image](https://user-images.githubusercontent.com/21187126/26912385-7f0018d8-4c46-11e7-8b29-1a2a80a40df0.png)

Payment Entry Set Up

Company Currency is in USD. 
However, we want to pay from a foreign currency account.
As you can see below, exchange rates are not fetched when I specify a foreign currency account.

![image](https://user-images.githubusercontent.com/21187126/26912515-89a0edfc-4c47-11e7-827d-22d973af9583.png)

![image](https://user-images.githubusercontent.com/21187126/26912443-f4bc2bb6-4c46-11e7-8546-b76807c2f5f1.png)

tundebabzy:
Hello @addymuliady. Thanks for reporting this. 

Let me go through the issue as I understand it. Your company default currency is USD. You have a bank account in a foreign currency, say SGD. When you try to make a payment from the SGD bank, ERPNext fails to retrieve the latest exchange rate for SGD - USD. Am I correct?

addymuliady:
Hi @tundebabzy 

Yes, that is correct.

tundebabzy:
In my case, the exchange rate is fetched although it's the average exchange rate that is fetched.

addymuliady:
Thanks @tundebabzy

tundebabzy:
Fixed

addymuliady:
@tundebabzy can you please claim your bounty? lol
https://www.bountysource.com/issues/46005757-payment-entry-latest-exchange-rate-not-fetched

However, now I have another issue. Not sure if you touched script for posting date.
https://github.com/frappe/erpnext/issues/9491

tundebabzy:
> @tundebabzy can you please claim your bounty? lol

Gladly. I wasn't even aware of it. Thanks. I think the posting date issue has been fixed but I'll investigate.

PR #9193, opened 2017-06-08
Payment Entry - Latest Exchange Rate not fetched

### LINKED ISSUE #8919: Journal Entry not fetching currency exchange rate set in Currency Exchange List master
ERPNext: v8.0.24 (master)

Frappe Framework: v8.0.42 (master)

-------------------------------------------------------

Steps to replicate

1) Set Exchange Rate
For example, this is set on 31 January 2017

image

image

2) Create Journal Entry

3) Select Posting Date as 31 January 2017

4) Row 1 Journal Entry Detail - Select a Foreign Currency Account. Click to view detail.
The exchange rate isn't fetched. It is still 1. As such the amount in Company Currency is wrong.

image

Expected Behaviour:
Exchange rate to fetch the latest from Posting Date. This will correctly calculate amount in company currency.

tundebabzy:
This fixes #8919, #9180

These two issues arise from the fact that "Pay" payment entries retrieve the average exchange rate used for transactions to the specified bank account. The side effect of this is that ERPNext can present different exchange rates for two different accounts with the same currency (#8919). Also, ERPNext will not be able to provide any exchange rate if the bank account balance is zero (#9180).

Technically, [IAS 21](https://www.iasplus.com/en/standards/ias/ias21#link8) does not allow for averaging exchange rates. Rather each transaction should be converted at the closing exchange rate on the day the transaction took place. Thus, the exchange rate automatically drawn should be the last saved exchange rate.

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -873,7 +873,7 @@ def get_exchange_rate(posting_date, account=None, account_currency=None, company
 
 		# The date used to retreive the exchange rate here is the date passed
 		# in as an argument to this function.
-		elif account_currency and posting_date:
+		elif (not exchange_rate or exchange_rate==1) and account_currency and posting_date:
 			exchange_rate = get_exchange_rate(account_currency, company_currency, posting_date)
 	else:
 		exchange_rate = 1
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -871,14 +871,9 @@ def get_exchange_rate(posting_date, account=None, account_currency=None, company
 		if reference_type in ("Sales Invoice", "Purchase Invoice") and reference_name:
 			exchange_rate = frappe.db.get_value(reference_type, reference_name, "conversion_rate")
 
-		elif account_details and account_details.account_type == "Bank" and \
-			((account_details.root_type == "Asset" and flt(credit) > 0) or
-				(account_details.root_type == "Liability" and debit)):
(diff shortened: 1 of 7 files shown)


tundebabzy:
Here's the difference

### Before
peek 2017-06-08 16-20

### After
peek 2017-06-08 16-35

nabinhait (changes requested):
The same issue is there in Journal Entry when bank account is credited, it takes average there as well.

erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -871,14 +871,9 @@ def get_exchange_rate(posting_date, account=None, account_currency=None, company
...
 
-		elif account_details and account_details.account_type == "Bank" and \
-			((account_details.root_type == "Asset" and flt(credit) > 0) or
-				(account_details.root_type == "Liability" and debit)):
-			exchange_rate = get_average_exchange_rate(account)
-
 		# The date used to retreive the exchange rate here is the date passed
 		# in as an argument to this function.
-		if not exchange_rate and account_currency and posting_date:
+		elif account_currency and posting_date:
```
nabinhait: Why removed `not exchange_rate`? User will not be able to set exchange_rate manually.

nabinhait (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.py
tundebabzy: i'll ammend the commit

addymuliady:
Greetings @nabinhait 

Is it possible to merge this into hotfix as it is critical to my accounting. I cannot overwrite the average exchange rate and am stuck with my work until this fix is merged to master.

For your kind perusal please. Thanks!

Issue #8343, opened 2017-04-05
Add more Contact fields like (Title /  Gender)


kickapoo:
In many cases like print format, email the following fields are needed:

- Title: A custom data field where user can enter values like Mr. / Dr. / Prof. (Before first_name)
- Gender: A select field with option Male / Female (After email)

kickapoo:
@nabinhait  Please validate/provide:
 - Proposed Field name are ok ? Yes/No
 - The documentation filepath that needs to update for this addition
 -  Are UnitTest needed in this case?

ishanloya:
https://github.com/frappe/erpnext/issues/8145

ishanloya:
https://github.com/frappe/erpnext/issues/8143

kickapoo:
@nabinhait Please check this one.

KanchanChauhan:
@kickapoo Already done, see PR https://github.com/frappe/erpnext/pull/8351 and https://github.com/frappe/frappe/pull/3026

me2ne:
better to add some extra email ad field , some contact have two email address

dottenbr:
great, looking forward to the merge so we can update our contacts

Issue #8143, opened 2017-03-25
[Enhancement] Contact name when lead is converted


ishanloya:
From the [forum](https://discuss.erpnext.com/t/contact-name-when-lead-is-converted/8947):

> Say I have a lead, named John Doe. While he is a lead, his full name is stored in person_name. When I convert to a customer, the lead is converted into a contact and the full lead name is dumped into the contact's first name.
> 
> Does it make sense to either de-concatenate the name (not sure if that's the right word) when the contact is added? Or perhaps change contact name to be full name?
>

kennethsequeira:
The system splits first and last names in contact when a lead is converted to customer

PR #9199, opened 2017-06-09
Salutation and Gender in Lead and Customer

### LINKED ISSUE #8145: [Enhancement] Salutation in lead & contact name
Suggestion:
- Add salutation field to lead
- Add salutation field to contact
- Map salutation lead field to contact when lead is converted to customer

Benefit:
- Field can be used in sales quotations (skipping salutation might seem a little too direct in some cultures, esp. when first name is not available for a contact)
- Could potentially be used for analytics (filter by gender for instance)
- Field can be used in purchase orders, RFQ

KanchanChauhan:
Depends on https://github.com/frappe/frappe/pull/3079

Fixed #8145, #8143 and https://github.com/frappe/erpnext/issues/8343

image


rmehta:
merging this for now, but there should be a follow up PR showing the title where applicable (i.e. Print Formats)

PR #9216, opened 2017-06-09
Create accounts_receivable.py

bobzz-zone:
reduce sql by not call party map every time


nabinhait:
It does not execute sql query every time, sql query is executed only once. Check the `hasattr` condition in `get_party_map` function.

Closing this PR.

Issue #8911, opened 2017-05-18
Once primary key is generated, do not allow changing Series


umairsy:
Raised via WN-SUP25185

Once the Sales Invoice is generated, Series field is not visible. But one can still edit value in the Series field from Report Builder, or (I suspect) from Data Import Tool as well.

If a primary key is generated for a document, then value in the Series field should freeze.

tundebabzy:
Hello @umairsy 
Please follow progress on this issue in frappe/frappe#3460

tundebabzy:
Fixed

PR #9235, opened 2017-06-12
Once primary key is generated, do not allow changing Series #8911

tundebabzy:
DocFields containing `naming_series` are presently have `set_only_once`as 0. Instead of adding more LOC, I think it's a better and easier option to set the `DocFields` with `naming_series` having `set_only_once` to 1.

### Before
peek 2017-06-10 13-57

### After
peek 2017-06-12 14-18


mbauskar:
@tundebabzy,

You will need to set the set_only_once field for all the doctype containing the field `naming_series`.

Closing the PR for now please reopen the once done.

tundebabzy:
@mbauskar 
The query does that and affects 46 Doctypes. `tabDocType` doesn't have fields `naming_series` and `set_only_once`.

rmehta:
@tundebabzy you need to update all the relevant JSON files otherwise it will not work for new installs. The patches don't run on fresh installs.

tundebabzy:
I forgot to update erpnext/patches.txt and that's probably why the tests were failing.

tundebabzy:
> @tundebabzy you need to update all the relevant JSON files otherwise it will not work for new installs. The patches don't run on fresh installs.

Oh ok. Will work on that now.

rmehta:
quick way of doing that is via a simple python script :)
```
for d in frappe.db.sql('select parent from tabDocField where fieldname="naming_series"):
  doctype = frappe.get_doc('DocType', d[0])
  for f in doctype.fields:
    if f.fieldname=='naming_series': f.set_only_once = 1
    doctype.save()
```

mbauskar:
@tundebabzy, can you rebase and push the changes again?

Please reopen the PR once conflicts are removed

tundebabzy:
My branch for this pull was based on hotfix. Accidentally pushed it to develop and that's the reason for the conflicts. The rebase is now very ugly so i'll squash the changes and push them in a new PR

Issue #9236, opened 2017-06-12
Maintain stock checkbox should reset serial nos and other checkboxes


mobentec:
We were having issues with product bundle parents needing a batch number.  We found out that the product had "batch number" originally selected, but when we un-checked "maintain stock", the batch number section within the item disappeared.  This caused us issues where we had to undo all orders and make the change within the product bundle parent.

nick9822:
Fixed via #9382

PR #9238, opened 2017-06-12
#9236 fix: Resetting batch and serial nos checks

nick9822:
#9236 
Resetting batch and serial nos checks on maintain_stock

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -123,10 +123,11 @@ frappe.ui.form.on("Item", {
 	},
 	
 	is_stock_item: function(frm) {
-		if(!frm.doc.is_stock_item)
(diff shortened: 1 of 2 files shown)


mbauskar:
@nick9822,

you will need to enclose your code in `{``}` brackets for if condition. Also please go through https://github.com/frappe/erpnext/wiki/Contribution-Guidelines

Please add screenshot/gif and description in PR body

nick9822:
Okay, please check description below.

While creating a new item with maintain stock, user checks all three serial and batch nos checkboxes.
capture 79
okay then user goes on and unchecking maintain stock.
capture 80
Now the previous action should reset the serial and batch nos checkboxes, but it does not. You will find it by again clicking on maintain stock checkbox and it will have previous checks as it is.
capture 81

And this pull request fixed the bug for us.

PR #9263, opened 2017-06-13
Production Order Enhancements

### LINKED ISSUE #8731: Production Order enhancements related to Warehouse
**Source Warehouse**
- [ ]  Add "Source Warehouse" field in the "Production Order Item" child table and it can be a group warehouse.
- [ ] Default Source Warehouse will be fetched from BOM, add source warehouse field in the BOM Item table.
- [ ] Reserve raw materials on it's own "source warehouse", even if it's a group.

**WIP Warehouse**
- [ ] Allow group warehouse in the WIP Warehouse
- [ ] If group, while making transfer entry, don't set target warehouse as WIP warehouse, leave it blank

**Reference Qty**
-  [ ] Add "Available Qty at Source Warehouse" and "Available Qty at WIP Warehouse" fields in Item table.

nabinhait:
Fixes #8731

  - Show required items child table
  - Source warehouse for each raw materials, in Pro Order Item and BOM Item table
  - Group warehouse allowed for source and wip warehouse
  - Patch to populate required items, to fix status and reserved qty for stopped pro order
  - Cleaned up existing codes
  - Test cases
screen shot 2017-06-12 at 5 53 45 pm


erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -475,6 +478,7 @@ def get_bom_items_as_dict(bom, company, qty=1, fetch_exploded=1, fetch_scrap_ite
 				item.default_warehouse,
 				item.expense_account as expense_account,
 				item.buying_cost_center as cost_center
+				{select_conditions}
```
rmehta: these are not conditions, these are columns, better name would be `select_columns`

rmehta:
@nabinhait can you also add an animated GIF to show UX?

Issue #9253, opened 2017-06-13
UX changes in program enrollment, student group and assessment plan


manassolanki:

tundebabzy:
merged

PR #9264, opened 2017-06-13
Minor changes in the UX and courses automatically fetched upon saving in program enrollment

manassolanki:
https://github.com/frappe/erpnext/issues/9253

image


erpnext/schools/doctype/program_enrollment/program_enrollment.js
```diff
@@ -46,6 +46,7 @@ frappe.ui.form.on("Program Enrollment", {
 					}
 				}
 			});
+			frm.events.get_courses(frm);
```
nabinhait: This should be inside `callback`

nabinhait (changes requested):

erpnext/schools/doctype/program_enrollment/program_enrollment.js
manassolanki: This is a different function from getting the fee schedule. Any advantage in writing inside the callback.

nabinhait: Both the functions is calling server side functions and executed asynchronously. In `get_courses` function, you are passing `doc` to the server, where still fees schedule is not there. Hence, if return of get_courses function comes later, it will overwrite the fees_schedule, which is earlier fetched from the first function.

PR #9265, opened 2017-06-13
Once primary key is generated, do not allow changing Series #8911

tundebabzy:
This is a PR for #8911 and continuation of #9235


nabinhait (changes requested):
As json has been modified, the patch is not needed.

Issue #8055, opened 2017-03-17
Purchase Return for Non Stock item


Arundhatii:
While creating a Purchase return entry for Non Stock (Service item), it says "Warehouse is Mandatory"
![return](https://cloud.githubusercontent.com/assets/20164097/24045775/3d8fa788-0b46-11e7-9f70-c2e1c2bfca4a.gif)

umairsy:
Purchase Return entry should be allowed for the stock item only. Even if Purchase Receipt was made for the non-stock item (which is debatable), in the Return entry, only stock items should be allowed.

Incase of services, they can simply issue Debit Note to the Supplier which can be adjusted in the future.

bohlian:
Guys, can I take this up? I am actually doing a bunch of fixes for Credit Notes and Debit Notes. Let me know.

PR #9282, opened 2017-06-14
[minor] Check if item is stock item before validating Warehouse while Purchase Return

sagarvora:
**Reference:** #8055 
---

**Before**

before 8055

---
**After**

after 8055

### DIFF
--- a/erpnext/controllers/sales_and_purchase_return.py
+++ b/erpnext/controllers/sales_and_purchase_return.py
@@ -71,7 +71,7 @@ def validate_returned_items(doc):
 	already_returned_items = get_already_returned_items(doc)
 
 	# ( not mandatory when it is Purchase Invoice or a Sales Invoice without Update Stock )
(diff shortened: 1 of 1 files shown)


erpnext/controllers/sales_and_purchase_return.py
```diff
@@ -101,8 +101,9 @@ def validate_returned_items(doc):
 								frappe.throw(_("Row # {0}: Serial No {1} does not match with {2} {3}")
 									.format(d.idx, s, doc.doctype, doc.return_against))
 
-				if warehouse_mandatory and not d.get("warehouse"):
-					frappe.throw(_("Warehouse is mandatory"))
+				if warehouse_mandatory and frappe.db.get_value("Item", d.item_code, "is_stock_item") \
```
nabinhait: Can you check how `warehouse_mandatory` has been set?

I think it has an issue. It should be `warehouse_mandatory = not ((doc.doctype=="Purchase Invoice" or doc.doctype=="Sales Invoice") and not doc.update_stock))`

nabinhait (changes requested):

nabinhait (approved):

PR #9295, opened 2017-06-14
[UX] Error in purchase transaction - Ignore instead of Prompt #7766

### LINKED ISSUE #7766: [UX] Error in purchase transaction - Ignore instead of Prompt
Steps:

1. In a new Purchase Order, selected Service Item.
2. Applied taxes.

On saving Purchase Order, getting error message shown below.

screen shot 2017-02-16 at 8 09 54 pm

Suggestion: Instead of forcing the user to change tax type, we should ignore valuation for the service item, and just show a notification to the user. Referring to the error message, User doesn't understand what to change/correct.

tundebabzy:
Fixes #7766 
If all items are non-stock items, the tax category is changed to 'Total' and a message is displayed on the page to notify the user of the change.

### Before
peek 2017-06-14 17-34

### AFter
peek 2017-06-14 17-46

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -73,10 +73,14 @@ class BuyingController(StockController):
 
 	def validate_stock_or_nonstock_items(self):
 		if self.meta.get_field("taxes") and not self.get_stock_items():
-			tax_for_valuation = [d.account_head for d in self.get("taxes")
+			tax_for_valuation = [d for d in self.get("taxes")
 				if d.category in ["Valuation", "Valuation and Total"]]
+
 			if tax_for_valuation:
-				frappe.throw(_("Tax Category can not be 'Valuation' or 'Valuation and Total' as all items are non-stock items"))
+				for d in tax_for_valuation:
+					d.db_set("category", 'Total')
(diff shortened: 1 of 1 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -73,10 +73,14 @@ def set_supplier_from_item_default(self):
...
 		if self.meta.get_field("taxes") and not self.get_stock_items():
-			tax_for_valuation = [d.account_head for d in self.get("taxes")
+			tax_for_valuation = [d for d in self.get("taxes")
 				if d.category in ["Valuation", "Valuation and Total"]]
+
 			if tax_for_valuation:
-				frappe.throw(_("Tax Category can not be 'Valuation' or 'Valuation and Total' as all items are non-stock items"))
+				for d in tax_for_valuation:
+					d.db_set("category", 'Total')
+				msgprint("Tax Category has been changed to \"Total\" because all Items in the Purchase Order "
```
rmehta: all messages have to be translatable _("Tax ...")

Also translatable messages cannot be broken into separate lines, and you cannot escape `\` inside translatable messages.

PR #9299, opened 2017-06-15
[fix] Payment Entry - draft sales invoices should not be selectable

### LINKED ISSUE #8044: Payment Entry - Draft Sales Invoices should not be selectable

rohitwaghchaure:
Fixed #8044

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.js
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.js
@@ -78,6 +78,20 @@ frappe.ui.form.on('Payment Entry', {
 				filters: { "name": ["in", doctypes] }
 			};
 		});
+
+		frm.set_query("reference_name", "references", function(doc, cdt, cdn) {
+			child = locals[cdt][cdn];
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -78,6 +78,19 @@ frappe.ui.form.on('Payment Entry', {
 				filters: { "name": ["in", doctypes] }
 			};
 		});
+
+		frm.set_query("reference_name", "references", function(doc, cdt, cdn) {
+			child = locals[cdt][cdn];
+			filters = { "docstatus": 1 };
```
nabinhait: Also add company filter

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.js
rohitwaghchaure: Fixed

PR #9302, opened 2017-06-15
Total sales cost in Project should be updated based on base_grand_total

nabinhait:


rmehta:
Screenshot was missing. Next-time won't merge!

Issue #8085, opened 2017-03-22
BOM - Raw-material Item's table. Reset value in BOM field on selection of Item Code


umairsy:
Steps:

1. While creating BOM, I selected a raw-material item which also had it's own BOM.
2. Then I edited Item Code and selected another raw-material Item (without BOM).
3. When saving BOM, I get the error message that Select BOM not available in the BOM selected with it.

This is because, when I replaced raw-material Item with one without BOM, value in the BOM field (in the BOM Item table) was not reset.

sagarvora:
Merged.

PR #9305, opened 2017-06-15
[minor] Fixed BOM value not updating on changing item

sagarvora:
**Reference:** #8085 

**Before**

before 8085

**After**

after 8085

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -113,7 +113,7 @@ class BOM(WebsiteGenerator):
 		item = self.get_item_det(args['item_code'])
 		self.validate_rm_item(item)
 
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -113,7 +113,7 @@ def get_bom_material_detail(self, args=None):
 		item = self.get_item_det(args['item_code'])
 		self.validate_rm_item(item)
 
-		args['bom_no'] = args['bom_no'] or item and cstr(item[0]['default_bom']) or ''
+		args['bom_no'] = item and cstr(item[0]['default_bom']) or ''
```
rmehta: but what if the item is same, but the user wants a non-default BOM?

this must be handled at the UI. When item is changed, clear the BOM and then reset it.

rmehta (changes requested):

PR #9309, opened 2017-06-15
leaderboard of customers, items, suppliers and sales partner

MCD-50:
Made a page under utilities modules.

Features:
 - return top 10 results based on doctype.
 - sort results based on the current filter selected
 - return results for a week, a month, a quarter and a year based on selected time.

Tested with demo entries.

screenshot from 2017-06-15 15-42-13


rmehta:
- numbers should be right aligned
- currency should be correctly formatted.

rmehta:
Please re-open with new screenshots and animated GIF when done.

Issue #9012, opened 2017-05-24
Fraction in UOM not allowed while Selling, but allowed on Purchase! 


mdwala:
In Sales Invoice, Fraction in UOM not allowed, 
Even when - nos is whole after fraction | UOM Case is not Checked as whole number | Already accepting Fraction while Purchasing

Example, 1 case = 100, 0.5 case(50 nos) not allowed.

https://discuss.erpnext.com/t/fraction-in-uom-not-allowed-while-selling-but-allowed-on-purchase/24166?u=mdwala

mdwala:
As nabinhait assigned ashwinisave35, how much time it will take?

umairsy:
Can you please share the screenshots to help this case. An screenshot of UoM, and validation message received in the Sales Invoice.

PR #9311, opened 2017-06-15
[minor] fix logic of quantity validation

sagarvora:
**Reference:** #9012 

**Changes:**

`stock_qty` is validated with `stock_uom`

and

`qty` is validated with `uom`

**Before**

before 9012

**After**

after 9012


mdwala:
Just came to say THANKS.....

PR #9329, opened 2017-06-16
[UX] Error in purchase transaction - Ignore instead of Prompt #7766

tundebabzy:
Corrections from #9295.
Reference issue (#7766) 

Notice is now translatable and no longer multi-line


mbauskar:
@tundebabzy ,

Previous PR was reverted https://github.com/frappe/erpnext/pull/9320

Closing this PR for now

rmehta:
@tundebabzy we don't use `db_set` in validate, because validate is called before saving (insert/update).

Issue #7582, opened 2017-01-25
POS Cash reconciliation 


ahmad18189:
Hello ,
POS cash reconciliation for sales daily is very important feature for POS module .
in the end of day there should by a reconciliation for each POS employee where it show 
the sales invoice made by him and the actual cash or transaction.

please if some one have more details needed for this feature i am willing to develop it and make a pull request for it 
best regards
Ahmad

mhbu50:
+1

Olawale1:
+1

Very Very Important

Olawale1:
Hi @ahmad18189 

Trust you're doing great. Any progress here?

Regards

pifiu:
Please see this: https://github.com/frappe/erpnext/issues/2731

xdelios:
I need this as well.
Any news on the progress?

rmehta:
Start a bounty! Get the developers attention.

heejit:
You can try following

1. We need to track start tran number and end tran number
   for each cashier.

2. We need to tag transaction as cash, card or credit

2. All sales will be debited to one customer a/c
   WALK-IN-CUSTOMER A/C

3. When you want to close the account 
   read all transaction for a cashier with start and end 
   transaction number 

4. Tally the cash, card and credit transaction

5. Post the JV
   CASH   DR
   CARD   DR
   CREDIT DR
     WALK-IN-CUSTOMER A/C CR
   
   End of the day walk-in-customer a/c must be zero 
   always.
   
6. Mark Start and end tran is closed for each cashier

chdecultot:
Hi @ahmad18189 ,

Have you been able to finish your alpha version ?

I am also volunteering to help develop it.

chdecultot:
Hi Everyone,

Since this topic seems to be on hold, I would like to propose the following and have started working on it.

I propose to divide the work in several parts. I didn't finish the spec for all steps, but I will update this ticket as soon as possible.

**Step 1**

In the scope of generating an X and a Z report, each transaction must be linked to a particular cashier.

My proposal is therefore to add a new field in the sales invoice Doctype, named for example "pos_profile" and to register the POS Profile making the transaction when it is submitted.

Every transaction should therefore be linked to a unique cashier.

It means that each POS Profile should only be applicable to one particular User. So we need to also change the POS Profile doctype to enable the parameter "Set Only Once" for a User.


I just pushed PR#9339 for this step.  


**Step 2**

Each new transaction should be flagged with a particular transaction number.
[...]

Issue #6308, opened 2016-09-07
POS - Allow Multiple POS Profiles for a User


felixvarghese:
Potential Use Cases:
- Currently, if a company has multiple sales outlets with users working at both, there is no good way to have a POS Profile for each
- Tracking for Individual POS Checkouts within a single location
- Working at different registers (returns/service vs sales)

Proposal: Allow multiple POS Profiles per user. Right now, if you try to create a second profile for a user/company combination, it throws an error.

Implementation
1. Allow multiple profiles to be created for a user/company combination - this may require the addition of a "Profile Name" to help people identify profiles
2. When opening the POS, If a user only has one profile created or valid, load the profile automatically. If a user has multiple profiles, allow a user to select which profile they want to use.

graphskill:
This would be a really useful addition

olamide2:
This is a good and much desired feature, I have just one issue to add..how can we ensure that when a staff is in a particular physical store all transactions he does will devolve into that particular store without HQ needing to do anything extra.

xctheo:
+1

mhbu50:
+1

felixvarghese:
I'm adding the comment I made on #7582 here.

With regards to POS Profile, one idea may be the following. Fundamentally, a user will create a POS Profile mostly based on a location. So, if you have 5 locations and 10 users, that can be 50 profiles. However, what if you simply assigned users to a POS Profile? So, if you have 5 locations, create 5 profiles. Then assign users to those profiles.

This would be similar to assigning employees to salary structures.

So, have a POS Profile, then have a child table of users. In the child table of users, you can assign users.

As for a unique identifier, each record in the user child table has a hash which can be used in Sales Invoice.

With this comment, implementation steps look like the following

- [ ] Create a POS Profile Users Child table - it has one field, a linked field to User
- [ ] Add the POS Profile Users child table to POS Profile
- [ ] Add a "POS Profile Name" field to POS Profile - this becomes the new unique identifier for POS Profile
- [ ] When user opens POS, allow them to select the correct POS Profile (I realize this point will need to be broken down into more parts for implementation)
- [ ] Write a patch to deal with existing implementation. The patch would do two things - (1)Populate the POS Profile Name field. Use the "Applicable for User" field as the value for this. (2)Populate the new Users child table in POS Profile. In this implementation, the "Applicable for User" User would be added as a line to the child table for each POS Profile.
- [ ] Deprecate the "Applicable for User" field in POS Profile

[...]

PR #9339, opened 2017-06-16
New fields for POS Profile in Sales Invoice and POS Profile Modification

chdecultot:
Hi,

This PR is the first of a bigger project to implement X and Z reports in the POS.
It is linked to issue #7582 

I have completed the issue to propose an approach and I would be very grateful if the Frappé Team could give some inputs before I continue the developments.

This first PR is only the addition of a new field called "pos_profile" in the sales invoice to register the name of the person behind the cash register.
I therefore also modified the pos.js file to register that information:

image

Of course the field is hidden but I just took a screenshot to demonstrate the result in the invoice document.

In order to have a consistent information, I also propose to modify the POS Profile doctype to set the linked user only once. Indeed we should be able to know for sure who was the cashier at a certain point and with the current system we can change it afterwards.

Don't hesitate to let me know your thoughts about this proposal.

Thank you very much.

Best regards,

Charles-Henri


felixvarghese:
To find out the cashier, can you just use the already existing "created by" field? 

That requires no modification to Sales Invoice and also no modification to POS Profile.

chdecultot:
Hi @felixvarghese,

That's a good point. I think it is good to link it to the POS profile, especially since it is in the roadmap to have multiple POS profiles per users in Q4 2017 (https://github.com/frappe/erpnext/issues/6308 that you created 😉).

As for the modification of the POS Profile, even if it is not agreed to have the information linked to the sales invoice, I don't see any use case where a POS profile should be linked to different users. In my opinion it should be unique to avoid fraudulent use cases. 
If needed it can be duplicated and a new one can be created for a new user.

If you are working with the POS system a lot, don't hesitate to share your use cases, I'm very interested to try to enhance it. 😊

felixvarghese:
If multiple POS profiles come into play, then its true that linking POS Profile in Sales Invoice will be important to generate X and Z reports. Good point.
 
It's currently not possible to link the same POS Profile to multiple users. And in the rough spec I mentioned for multiple profiles, it still won't be possible. Of course, if the spec changes, then that's a different story.

rmehta:
Multiple POS profile is a popular requirement, so will be good to plan for it. 👍

PR #9341, opened 2017-06-16
send email on update and on update after submit

### LINKED ISSUE #8898: Training doc sends emails on save
The training document currently sends emails to all the Open attendees on Save of the document, when it should wait until the document is submitted.

manassolanki:
fixes #8898

### DIFF
--- a/erpnext/hr/doctype/training_event/training_event.py
+++ b/erpnext/hr/doctype/training_event/training_event.py
@@ -8,20 +8,19 @@ from frappe import _
 from frappe.model.document import Document
 
 class TrainingEvent(Document):
-	def validate(self):
-		if self.event_status == "Scheduled":
-			self.invite_employee()
+	def on_update(self):
+		self.invite_employee()
 	
 	def on_update_after_submit(self):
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/training_event/training_event.py
```diff
@@ -8,20 +8,19 @@
...
-			.format(self.type, self.event_name, self.start_time, self.end_time, self.location))
-	
-		for emp in self.employees:
-			if emp.status== "Open":
-				frappe.sendmail(frappe.db.get_value("Employee", emp.employee, "company_email"), \
-					subject=subject, content= self.introduction)
-				emp.status= "Invited"
\ No newline at end of file
+		if self.event_status == "Scheduled" and self.send_email:
+			subject = _("""You are invited for to attend {0} - {1} scheduled from {2} to {3} at {4}."""\
```
rmehta: Lets do email alert for this? Its a much more elegant solution since the user can design the content of the email. You can add it to the fixtures and then call it when `send_email` is true.

rmehta (changes requested):

Issue #6510, opened 2016-09-30
Purchase Invoice


rohitwaghchaure:
- [ ] IS Paid feature's information is missing 
- [ ] Update Stock feature's information is missing 
- [ ] Remove old images and add latest

PR #9347, opened 2017-06-18
Purchase Invoice #6510

tundebabzy:
#6510


rmehta:
@tundebabzy thanks for going into the docs :)

Currently our practice for the docs is not to have field level documentation (this should be obvious to the user) but a functional documentation. 

What was missing as per #6510 is the explanation for "Is Paid" and "Update Stock" features. Maybe a couple of paragraphs explaining that would be good.

Closing this for now. Please re-open when you fix it!

PR #9348, opened 2017-06-18
Fix and improve shopping cart

ghost:
- Fix login page redirect when putting product in cart
- Products ordered by item name (after weightage)
- Fix shopping cart crash when removing the last product from cart
- Don't display tax in cart if amount is zero
- Translations

shopping-cart

### DIFF
--- a/erpnext/public/js/shopping_cart.js
+++ b/erpnext/public/js/shopping_cart.js
@@ -36,7 +36,6 @@ $.extend(shopping_cart, {
 	},
 
 	update_cart: function(opts) {
-		var full_name = frappe.session.user_fullname;
 		if(!full_name || full_name==="Guest") {
 			if(localStorage) {
 				localStorage.setItem("last_visited", window.location.pathname);
--- a/erpnext/public/js/shopping_cart.js
+++ b/erpnext/public/js/shopping_cart.js
@@ -35,10 +35,8 @@ $.extend(shopping_cart, {
 		});
 	},
 
-	update_cart: function(opts) {
-		var full_name = frappe.session && frappe.session.user_fullname;
-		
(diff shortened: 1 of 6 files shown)


rmehta:
@britlog thanks for the PR. Can you add an animated GIF to show that this works. Since the users are concerned about stability, we want to ensure that user testing is done at your end. You can use LiceCAP to take an animated GIF.

britlog:
@rmehta Done, but I can't reopen it because it was not closed by myself, could you please do it ? Thanks

rmehta:
Thanks! The strawberries look delicious 🍓

erpnext/public/js/shopping_cart.js
```diff
@@ -36,7 +36,6 @@ $.extend(shopping_cart, {
 	},
 
 	update_cart: function(opts) {
-		var full_name = frappe.session.user_fullname;
```
britlog: I 've noticed frappe.session.user_fullname was empty, so i just revert to the last version which worked but indeed use global. Thanks for review.

erpnext/public/js/shopping_cart.js
```diff
@@ -35,10 +35,8 @@ $.extend(shopping_cart, {
 		});
 	},
 
-	update_cart: function(opts) {
-		var full_name = frappe.session && frappe.session.user_fullname;
-		
-		if(!full_name || full_name==="Guest") {
+	update_cart: function(opts) {		
+		if(fraappe.session.user==="Guest") {
```
britlog: There is a typing error on frappe word  ` if(fraappe.session.user==="Guest")`

rmehta:
Thanks fixed.

sophatvathana (commented):
Ok

PR #9350, opened 2017-06-19
[minor] Fixed BOM value not updating on changing item

sagarvora:
**Reference:** #8085 

Made changes requested in #9305 

**Before**

before 8085

**After**

after 8085


rmehta:
better UX is that it should select the default BOM instead of keeping it empty.

So when it is changed, pass a key `changed` and then if it is changed, either set the new default BOM or make it blank.

sagarvora:
Hi, the `get_bom_material_detail` function on line 75 does select default BOM. 

Proof:

proof

PR #9355, opened 2017-06-19
[Enhance] Company wise perpetual inventory settings

rohitwaghchaure:
Added perpetual inventory setting in the company
27227460-2f977c10-52c1-11e7-80f8-8c9288eb28b3

#9400


erpnext/__init__.py
```diff
@@ -53,4 +54,12 @@ def encode_company_abbr(name, company):
 
 	return " - ".join(parts)
 
+def is_perpetual_inventory_enabled(company):
+	if not company: return None
+
+	is_enabled = frappe.cache().hget("enable_perpetual_inventory", company) or None
```
rmehta: 👎 what if auto inventory is off? This will always be false!

Local caching means caching in the memory. Like in `frappe.local`

```
def is_perpetual_inventory_enabled(company):
    if not getattr(frappe.local, 'enable_perpetual_inventory'):
        frappe.local.enable_perpetual_inventory = frappe.db.get_value("Company", company, "enable_perpetual_inventory")
  return frappe.local.enable_perpetual_inventory
```

rmehta (changes requested):

PR #9362, opened 2017-06-19
Inspection required before delivery isn't working for item variants

### LINKED ISSUE #9332: Inspection required before delivery isn't working for Item Variants.
Reported via https://discuss.erpnext.com/t/how-to-creating-inspection-criteria-items-in-the-variant/24875

I defined a template with variants. On the other hand, I checked "Inspection Required before Delivery" and stored some values in the table. My expectation was generate these items inside variants automatically when I create them. However it didn't happen.

tundebabzy:
fixes #9332

`copy_attributes_to_variant` method ignores "Tables" when doing its heavy lifting. Tables in templates shouldn't be ignored


rmehta:
Have you tested this works? Animated GIF missing.

Tables might to be copied manually in python (I am not sure moving by reference will work)

Closing this for now. Lets keep our PR standards high!

tundebabzy:
### Before
peek 2017-06-19 21-41

### After
peek 2017-06-19 21-29

mbauskar:
@tundebabzy, 

Travis build is failing can you please fix the test case? Please reopen the PR once the test case is fixed

tundebabzy:
Finally found the problem.

It seems like at a point in time, changes to a template automatically changes variants created from the template... at least `_test_auto_material_request` seemed to be working based on that premise which is no longer true. I changed the test to work directly with the variant or item not the template from which the variant was created.

PR #9372, opened 2017-06-20
Serial no/batch entry dialog

pratu16x7:
#8848


erpnext/stock/doctype/batch/batch.json
```diff
@@ -412,8 +425,8 @@
  "issingle": 0, 
  "istable": 0, 
  "max_attachments": 5, 
- "modified": "2017-04-20 03:22:19.888058", 
- "modified_by": "Administrator", 
+ "modified": "2017-05-21 21:00:11.096038", 
+ "modified_by": "prateeksha@erpnext.com",
```
mbauskar: modified_by should be Administrator

mbauskar (changes requested):

rmehta:
@pratu16x7 try using object references instead of jquery as we discussed.

pratu16x7:
Now depends on frappe/frappe#3540

rmehta:
👍 also needs to happen for delivery note, purchase receipt, sales invoice (if update stock is checked) and purchase invoice (if update stock is checked) 

Add this in a separate file and import it the various documents

erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -301,6 +301,24 @@ def update_billing_status(self, update_modified=True):
 
 		self.load_from_db()
 
+	def get_batched_serialized_details(self, item_code):
```
pratu16x7: So Stock Entry simply had a comprehensive `get_item_details` method that got many things. Delivery Note and the Invoices don't have such method. This is the way I fetched whether the item is batched or serialised here. But doing the same thing for the invoices seems too repetitive and specialised.

erpnext/public/build.json
```diff
@@ -30,6 +30,7 @@
         "public/js/payment/payment_details.html",
         "public/js/templates/item_selector.html",
         "public/js/utils/item_selector.js",
+        "public/js/utils/serial_no_batch_selector.js",
```
rmehta: don't add this here... this will add to the initial load.

use `frappe.require` in the files you need just JS. The file will be loaded in background by the time the user comes to the item selection

rmehta (changes requested):

erpnext/public/js/controllers/transaction.js
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 // License: GNU General Public License v3. See license.txt
 
+frappe.require("assets/erpnext/js/utils/serial_no_batch_selector.js");
```
pratu16x7: `frappe.require` usage without callback dependent on frappe/frappe#3564

rmehta: require is async, you have to use the callback

erpnext/public/js/controllers/transaction.js
```diff
@@ -265,14 +267,22 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	item_code: function(doc, cdt, cdn, from_barcode) {
 		var me = this;
 		var item = frappe.get_doc(cdt, cdn);
+		var update_stock = 0, show_batch_dialog = 0;
+
+		if(['Sales Invoice', 'Purchase Invoice'].includes(this.frm.doc.doctype)) {
```
rmehta: also delivery note and purchase receipt (if return)

erpnext/public/js/controllers/transaction.js
```diff
@@ -265,14 +267,22 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+			update_stock = cint(me.frm.doc.update_stock);
+			show_batch_dialog = update_stock;
+		} else if(this.frm.doc.doctype === 'Delivery Note') {
```
rmehta: also purchase receipt (if return) `is_return`

erpnext/public/js/controllers/transaction.js
```diff
@@ -310,6 +320,23 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
 							me.frm.script_manager.trigger("price_list_rate", cdt, cdn);
 							me.toggle_conversion_factor(item);
+							if(show_batch_dialog) {
+								var d = locals[cdt][cdn];
+								$.each(r.message, function(k, v) {
+									if(!d[k]) {
+										d[k] = v;
+									}
+								});
+								let serial_no_batch_selector = new erpnext.SerialNoBatchSelector({
```
rmehta: create a method `erpnext.show_serial_batch_selector` and put `frappe.require` inside it.

erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -0,0 +1,340 @@
...
+		if(this.has_batch) {
+			this.values.batches.map((batch, i) => {
+				let row = (i !== 0) ? this.frm.add_child("items") : this.item;
+				refresh_field("items");
+				if(i !== 0) {
+					// let index = i + me.item.idx - 1;
+					// let grid_row = me.frm.fields_dict.items.grid.grid_rows[index];
+					// grid_row.toggle_editable_row();
+					// grid_row.on_grid_fields_dict.item_code.set_value(this.item_code);
+					row.item_code = this.item_code;
```
pratu16x7: Discovered that `refresh_field("items")` after this line (setting item_code value) does little to trigger the relevant fields and turn the indicator green.
One way is probably using `set_value()`, but it is too expensive for multiple rows.

screen shot 2017-06-26 at 3 33 45 pm

rmehta: you need to do
```
frappe.model.set_value(row.doctype, row.name, 'item_code', this.item_code);
```

erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -156,6 +161,11 @@ erpnext.SerialNoBatchSelector = Class.extend({
 			row.warehouse = values.warehouse || warehouse;
 		}
 
+		if(item_code_field) {
+			item_code_field.set_value(this.item_code);
```
pratu16x7: So yes, `set_value()`, the only way one trigger the remaining fields.
Seems it still has a problem, completely setting only the last row in the group (the first row is already set initially), leaving all the others partially filled:
set_value_miss
This probably has to do with it being called in bulk for all the items.

erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -0,0 +1,349 @@
...
+		if(this.warehouse_details.type === 'Source Warehouse') {
+			row.s_warehouse = values.warehouse || warehouse;
+		} else if(this.warehouse_details.type === 'Target Warehouse') {
+			row.t_warehouse = values.warehouse || warehouse;
+		} else {
+			row.warehouse = values.warehouse || warehouse;
+		}
+
+		if(item_code_field) {
+			item_code_field.set_value(this.item_code);
```
pratu16x7: So yes, `set_value()`, the only way one trigger the remaining fields.
Seems it still has a problem, completely setting only the last row in the group (the first row is already set initially), leaving all the others partially filled:
set_value_miss
This probably has to do with it being called in bulk for all the items.

pratu16x7:
Now depends on frappe/frappe#3570

Issue #9390, opened 2017-06-20
[Error Report] make_variant_item_code() takes exactly 3 arguments (2 given)


saurabh6790:
```
Traceback (most recent call last):
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/desk/form/save.py", line 22, in savedocs
    doc.save()
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 230, in save
    return self._save(*args, **kwargs)
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 253, in _save
    self.insert()
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 187, in insert
    self.set_new_name()
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 333, in set_new_name
    set_new_name(self)
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/naming.py", line 38, in set_new_name
    doc.run_method("autoname")
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 666, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 892, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 875, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
  File "/home/frappe/benches/bench-2017-06-15/apps/frappe/frappe/model/document.py", line 660, in 
    fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
  File "/home/frappe/benches/bench-2017-06-15/apps/erpnext/erpnext/stock/doctype/item/item.py", line 40, in autoname
    self.item_code = make_variant_item_code(self.variant_of, self)
TypeError: make_variant_item_code() takes exactly 3 arguments (2 given)
```

PR #9280, opened 2017-06-14
[Minor] fix in the item for the function make_variant_item_code

manassolanki:
Fix for the issue https://github.com/frappe/erpnext/issues/9279

PR #9391, opened 2017-06-20
[minor][fix] pass template item name to make_variant_item_code

saurabh6790:
https://github.com/frappe/erpnext/issues/9390


manassolanki:
https://github.com/frappe/erpnext/pull/9280

PR #9410, opened 2017-06-21
Patch for the PR 8754, fixes #9011

### LINKED ISSUE #9011: Existing company record address not converted on upgrade
This PR - https://github.com/frappe/erpnext/pull/8754 - got rid of the address text field and moved to a related address record. While that is great, the upgrade process not take into account existing values.

Old (I obfuscated the address there on purpose):
image

New:
image

I would have expected the upgrade process to have parsed the address and created an associated address record on upgrade. Losing data during an upgrade is not a good practice.

manassolanki:
patch for removing the address field from the company and creating an address doc.

fixes https://github.com/frappe/erpnext/issues/9011

### DIFF
--- a/erpnext/patches/v8_0/create_address_doc_from_address_field_in_company.py
+++ b/erpnext/patches/v8_0/create_address_doc_from_address_field_in_company.py
@@ -9,7 +9,8 @@ def execute():
 	# so here is the patch for moving the address details in the address doc
 	company_list = []
 	if 'address' in frappe.db.get_table_columns('Company'):
-		company_list = frappe.db.sql('''select name, address from `tabCompany` where address is not null''', as_dict=1)
+		company_list = frappe.db.sql('''select name, address from `tabCompany` 
+			where address is not null and address != ""''', as_dict=1)
 
 	for company in company_list:
 		add_list = company.address.split(" ")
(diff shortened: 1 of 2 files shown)


erpnext/patches/v8_0/create_address_doc_from_address_field_in_company.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2017, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	# new field address_html is created in place of address field for the company's address
+	# patch for moving the address details in the address doc
+	company_list = frappe.db.sql('''select name, address from `tabCompany` where address is not null''', as_dict=1)
```
mbauskar: @manassolanki,

Please check if the address column is available or not then use `frappe.db.sql`

mbauskar (changes requested):

mbauskar:
please push the changes and reopen the PR

nabinhait:
Create the address record only if there is already no address record for the same company.

To check address column in the table, you can use `if 'address' in frappe.db.get_table_columns('Company')`.

manassolanki:
@nabinhait ,
This is the patch for the PR https://github.com/frappe/erpnext/pull/8754

Issue #9415, opened 2017-06-21
Incorrect payment accounts in purchase invoice


mobentec:
We are entering in a bill for a purchase of raw goods.  We create a Purchase Invoice and select the "Is Paid" checkbox.

Under "Payments" subheading, we select mode of payment "Credit Card".

For Cash/Bank Account, we need to select our credit card (American Express account xxxx) which is a liabilities account, but ERPNext is only listing Asset accounts from within it's drop down.

The way we were able to get around this, is to uncheck "Is paid", and then make a Payment Entry.  
When going through Payment Entry, it allows us to select the correct liabilities account (Amex Credit Card account).

![2017-06-21 at 09-39-39](https://user-images.githubusercontent.com/6663120/27387053-99dbf054-5665-11e7-9c41-457e747b2d65.png)

tundebabzy:
I don't think the "Is Paid" check box was meant to treat non-asset accounts though

nick9822:
Fixed via #9418

PR #9418, opened 2017-06-21
Removed "Asset" filter on payment account

nick9822:
Fix toward #9415 
Purchase Invoice should also have "Liability" root type, so removed the current "Asset" filter. 
So now user can choose any account.
capture 86


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -224,7 +224,6 @@ cur_frm.fields_dict.cash_bank_account.get_query = function(doc) {
 	return {
 		filters: [
 			["Account", "account_type", "in", ["Cash", "Bank"]],
-			["Account", "root_type", "=", "Asset"],
```
rohitwaghchaure: Add filter report type as "Balance Sheet"

nick9822: @rohitwaghchaure 
Okay, its done. Please check latest commit.

rohitwaghchaure (approved):

PR #9419, opened 2017-06-22
ERPNext Logo to show on Email Template

ashwinisave35:


rmehta:
should be erpnext-logo, not erp-logo

PR #9442, opened 2017-06-23
BOM Stock Report to allow Multi Level Warehouse Setup

creamdory:
In the current BOM Stock Report, we are only allowed to view current stock in a specific warehouse. It would be much helpful if we can view all current stock in all child warehouses of a Parent. I have revised the report from a Query report to a Script Report to cater this usecase, using the original Query from the previous query report, I added condition for the warehouse filter. If warehouse is a group, it will show all qty under its Child warehouse.

Here is a sample Warehouse Structure:

warehouse structure

Here are the current stock of my component items of my BOM:

stock summary

Here is the Report from my PR:
bom stock report

### DIFF
--- a/erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
+++ b/erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
@@ -5,13 +5,11 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _
 
-
 def execute(filters=None):
     if not filters: filters = {}
     columns = get_columns()
     data = get_bom_stock(filters)
     return columns, data
-
 
 def get_columns():
     """return columns"""
@@ -24,7 +22,6 @@ def get_columns():
     ]
 
     return columns
-
 
 def get_bom_stock(filters):
     conditions = ""
@@ -57,5 +54,4 @@ def get_bom_stock(filters):
             WHERE
     	        bom_item.parent = '%s'
 
-            GROUP BY bom_item.item_code""" %
-                         (conditions, bom))
+            GROUP BY bom_item.item_code""" % (conditions, bom))
--- a/erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
+++ b/erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
@@ -1,0 +1,61 @@
+# Copyright (c) 2013, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
+
+from __future__ import unicode_literals
+import frappe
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
```diff
@@ -0,0 +1,66 @@
...
+from frappe import _
+
+
+def execute(filters=None):
+    if not filters: filters = {}
+    columns = get_columns()
+    bom_stock = get_bom_stock(filters)
+    data = []
+
+    for row in bom_stock:
```
mbauskar: @creamdory,

No need to iterate over bom_stock, just remove the as_dict=1 from `get_bom_stock` and directly assign the result to data

erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
```diff
@@ -0,0 +1,66 @@
...
+
+def get_bom_stock(filters):
+    conditions = ""
+    bom = filters.get("bom")
+
+    if filters.get("warehouse"):
+        warehouse_details = frappe.db.get_value("Warehouse", filters.get("warehouse"), ["lft", "rgt"], as_dict=1)
+        if warehouse_details:
+            conditions += " and exists (select name from `tabWarehouse` wh \
+        				where wh.lft >= %s and wh.rgt <= %s and ledger.warehouse = wh.name)" % (warehouse_details.lft,
```
mbauskar: can you please fix the intendation ?

erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
```diff
@@ -0,0 +1,66 @@
...
+            FROM
+    	        `tabBOM Item` AS bom_item
+    	        LEFT JOIN `tabBin` AS ledger
+    		    ON bom_item.item_code = ledger.item_code
+    		    %s
+            WHERE
+    	        bom_item.parent = '%s'
+
+            GROUP BY bom_item.item_code""" %
+                         (conditions, bom), as_dict=1)
```
mbauskar: remove this as_dict so that we don't need to iterate through items

mbauskar (changes requested):

erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
creamdory: Sure. I'll make the necessary changes @mbauskar

creamdory:
Done with the changes @mbauskar , thanks!

PR #9455, opened 2017-06-24
redirect to previous page on save purchase entry

vishdha:
Redirect to previous purchase invoice page on save payment entry


rmehta:
redirect only if last route is in references

PR #9490, opened 2017-06-27
keep the batch no always viewable

### LINKED ISSUE #9246: Batch number number is shown in Item table of Sales Invoice is not update If "Update stock" is unchecked. Also the issue prevails if the user starts a new Invoice
Reported via https://frappe.erpnext.com/desk#Form/Issue/WN-SUP25629 

Now in this scenario there are two types of sales Invoice in our list - some with update stock checked and some with update stock not checked, but both having batches. While viewing a sales Invoice with Update Stock checked, we see the batches displayed.

While viewing a sales Invoice with Update Stock not checked, we don't see the batches displayed although the batches are selected. Now that is the scenario when I open a invoice just after refreshing ERPNext or fresh login. But when you switch between two different types of Invoices, then the display is not refreshed and you see the display form for the first Invoice you opened. i.e - If I open Invoice 1 (with Update stock)  first and then opened Invoice 2 (w/o Update Stock) I would see batches displayed for Invoice 2 also and vice versa.

manassolanki:
Leave this on the user how batch no and serial no should be handled for the different items.

Fixes https://github.com/frappe/erpnext/issues/9246

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
@@ -324,10 +324,13 @@ cur_frm.cscript.hide_fields = function(doc) {
 		}
 	}
 
+	// Keep all field viewable in item child table
+	/*
 	var item_fields_stock = ['batch_no', 'actual_batch_qty', 'actual_qty', 'expense_account',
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
Send fix into the hotfix branch

PR #9495, opened 2017-06-28
Supplier image

vishdha:
Summary
shows supplier image in desk page


mbauskar:
Thanks for contributing!
Our contribution guidelines require that you add a screenshot showing that your proposed change works with an example. This pull request will be closed, but you can re-open it after adding the requested screenshot.

PR #9509, opened 2017-06-28
[Fix] General ledger report shows null voucher type in other language than english

### LINKED ISSUE #9394: General Ledger Report Shows Null Voucher Type in Other Lenguage
https://discuss.erpnext.com/t/general-ledger-report-shows-null-voucher-type-in-other-lenguage/25115

In General Ledger report the field voucher type that reference the voucher number dissapear when the user language is other than english, Spanish my case, making imposible to access the voucher via the quick link as it throws a doctype not found error.

rohitwaghchaure:
**Issue**
sales_invoice_issue

**Fixed**
fixed_general_ledger_issue

Fixed https://github.com/frappe/erpnext/issues/9394

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.py
+++ b/erpnext/accounts/report/general_ledger/general_ledger.py
@@ -89,7 +89,7 @@ def get_columns(filters):
 		]
 
 	columns += [
-		_("Voucher Type") + "::120", _("Voucher No") + ":Dynamic Link/"+_("Voucher Type")+":160",
+		"Voucher Type" + "::120", _("Voucher No") + ":Dynamic Link/Voucher Type:160",
 		_("Against Account") + "::120", _("Party Type") + "::80", _("Party") + "::150",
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -89,7 +89,7 @@ def get_columns(filters):
 		]
 
 	columns += [
-		_("Voucher Type") + "::120", _("Voucher No") + ":Dynamic Link/"+_("Voucher Type")+":160",
+		"Voucher Type" + "::120", _("Voucher No") + ":Dynamic Link/Voucher Type:160",
```
mnatalia: don't you think that column label "Voucher Type" will not be translatable? I think you should leave it as _("Voucher Type")
I suppose you just don't need it in "Dynamic Link/Voucher Type:160"
btw, how about line 95, does everything work?
and sorry, I use ERPNext: v8.0.36 and I cannot replicate this problem

nabinhait:
Seems like this is not the correct fix. Something, you need to fix in the core.

Issue #9166, opened 2017-06-07
Unable to create Manual Depreciation of Asset due to Error


addymuliady:
ERPNext: v8.0.40 (master)
Frappe Framework: v8.0.60 (master)

In reference to : https://discuss.erpnext.com/t/depreciation-entry-manual-adjustments/24468/2

Replicate

1. Create Depreciation Entry via Journal Entry
2. Select "Accumulated Depreciation" Type Account
3. Select Reference Type as Asset
4. Try select Asset
5. System ask for Party (which shouldn't be the case)

![image](https://user-images.githubusercontent.com/21187126/26863043-860f96f4-4b82-11e7-8f2d-d8d602320729.png)

addymuliady:
Update: I have selected Party Type and Party but still receiver the following error on my console log.

jquery.min.js:4 GET http://erp.asonic-aviation.com/?txt=&doctype=Asset&filters=%5B%5B%22Asset%2…stries+Pte+Ltd%22%5D%5D&cmd=frappe.desk.search.search_link&_=1496812320710 500 (INTERNAL SERVER ERROR)

rohitwaghchaure:
Hi @addymuliady 

No need to define the party type and party name for asset depreciation.
![screen shot 2017-06-07 at 3 29 29 pm](https://user-images.githubusercontent.com/8780500/26873051-339b32fc-4b96-11e7-9c30-b3738110e17e.png)

addymuliady:
Hi @rohitwaghchaure 

Appreciate your reply. So I assume this is a bug? I am self hosted.

Party Type and Party fields are shown by default on my side.

Thanks!

tundebabzy:
fixed.

PR #9510, opened 2017-06-28
Unable to create Manual Depreciation of Asset due to Error #9166

tundebabzy:
Reference: #9166 

### Before
peek 2017-06-28 13-53

### After
peek 2017-06-28 13-57


nabinhait:
This is not the correct fix. The issue will arise for all reference types except Sales/Purchase Order and Sales / Purchase Invoice.

The correct fix will be:
1. Check for account if reference type is Sales / Purchase Invoice
1. Check for party type and party name if reference type is Sales / Purchase Order
3. Don't check for anything if reference_type is other than 1 & 2.

Issue #9530, opened 2017-06-29
Can't set Start and End Dates in Salary Slip #9513


tundebabzy:
Reference: #9513 
Presently, start date for salary slips and process payroll can only be set to the start of the period in the fiscal year. This problematic for companies that do not have their accounting period starting at the beginning of the month. For example, if accounting period runs from 16th - 15th, you can only create slips with pay period starting on the 16th and ending on the 30th.

I'm proposing we change this as follows:
- [x] Create a new Document - 'Payroll Cycle' which holds information about the company's payment cycle. It will be tied to `Designation`s.
- [x] Create a new Document - 'Payroll Period' to be set by the user. `Payroll Period` contains the start date, end date and name of the Payroll Period. It will be a child of `Payroll Cycle`.
- [x] Create a new Document - 'Payroll Cycle Designation' which is a link to 'Designation' and added as a child to 'Payroll Cycle'
- [x] Show status in list view (Is Active/Not Active)
- [ ] Add Payroll Cycle validation:
  - [ ] Frequency cannot be changed once saved
  - [ ] Number of Payroll Periods must be consistent with the frequency chosen. e.g if frequency is 'Monthly', there should be not more than one Payment Period per calendar month; there should be no more than 12 Payment Periods per fiscal year.
  - [ ] Dates must not overlap.
  - [ ] Gaps between dates not allowed.
  - [ ] Period dates cannot be set to dates to past periods e.g If today's date is 01-07-2017 and Payroll Cycle frequency is "Monthly", you cannot add period 01-06-2017 - 30-06-2017 to the cycle.
- [ ] Add 'Payroll Cycle' as a link field to `Process Payroll` and `Salary Slip`.
- [ ] Populate `start_date` and `end_date` from the selected `Payroll Cycle`.
- [ ] In salary slip validation, ensure that salary slip pay periods do not overlap.
- [ ] Remove `payment_frequency` field

rmehta:
Maybe **Payroll Period** instead of **Payroll Frequency**

tundebabzy:
> Maybe Payroll Period instead of Payroll Frequency

Yap. it's more explanatory. OP edited

Issue #9513, opened 2017-06-28
Can't set Start and End Dates in Salary Slip


theopen-institute:
When creating new salary slips for any of my monthly employees, I am not able to set my desired Start and End Dates. No matter what I put, the Start and End dates automatically revert to the 16th and last days of whatever month I've selected (respectively).

My salary structure seems to be set correctly, and all of my employees have start dates before the current month. Something, however, is causing the dates to change as soon as I enter them.

tundebabzy:
Hi @theopen-institute ,
Does your fiscal year start on the 16th?

theopen-institute:
Thanks for taking this on @tundebabzy 

Good catch, yes it does. Is that the root of the problem? Even if I were to shift my payment cycle onto 16th to 15th, I'm still not able to change the end date to anything other than the last day of the month.

tundebabzy:
@theopen-institute 
> Is that the root of the problem?

Yes. The behaviour is wrong though. The start date should be exactly the same as the start date on the salary structure. I can fix this today and hopefully into will make into into production in tomorrow's release.

theopen-institute:
Thanks @tundebabzy

tundebabzy:
Hello @theopen-institute 
Please follow progress on this issue on #9530 .

PR #9559, opened 2017-06-30
Can't set Start and End Dates in Salary Slip - 1

tundebabzy:
Reference - #9530, #9513 

Here is the first PR that creates Doctypes towards making creating Salary Slips and Process Payroll more flexible as well as fixing bugs like #9513 - Payroll Cycle, Payroll Period and Payroll Cycle Designation. Both Payroll Period and Payroll Cycle Designation are child tables of Payroll Cycle. I also added a test case. 

The target is that for each employee, ERPNext would check the Payment Cycle attached to the employee's designation to get pay period information needed for payroll runs.


tundebabzy:
### Screenccast

peek 2017-06-30 23-33

tundebabzy:
### Completed task 4
screenshot from 2017-07-01 13-43-40

erpnext/hr/doctype/payroll_cycle/test_payroll_cycle.py
```diff
@@ -0,0 +1,45 @@
...
+                "doctype": "Designation"}
+
+period = {"end_date": "2017-06-30", "start_date": "2017-06-01",
+          "doctype": "Payroll Period", "idx": "1"}
+
+cycle_dict = {"cycle_name": '_Test Payment Cycle All Employees', "company_name": '_Test Company',
+              "payroll_frequency": "Monthly", "doctype": "Payroll Cycle"}
+
+
+def get_document(dt, designatione):
```
rmehta: put this in `def setUp` method of the test case

erpnext/hr/doctype/payroll_cycle/test_payroll_cycle.py
```diff
@@ -0,0 +1,45 @@
...
+	if not frappe.db.exists(dt, designation.name):
+		designation.insert()
+	return designation
+
+get_document('Designation', designations)
+
+
+class TestPayrollCycle(unittest.TestCase):
+	def test_create_payment_cycle(self):
+		payroll_cycle = frappe.get_doc(cycle_dict)
```
rmehta: just add the dict here. its much more readable!

rmehta (changes requested):

tundebabzy:
Trying another approach

magic-overflow:
Have you been working on another approach?

PR #9578, opened 2017-07-03
New field "valid through" in Quotation to check uptill when it is valid

### LINKED ISSUE #9577: New field in Quotation "Valid Through" to check uptill when it is valid

KanchanChauhan:
Be default set for 1 month from Transaction Date
screen shot 2017-07-03 at 11 12 55 am

Fixed https://github.com/frappe/erpnext/issues/9577


rmehta:
"Valid Till" is more clear. Please update

screen shot 2017-07-10 at 12 18 14 pm

PR #9585, opened 2017-07-03
[Fix] Opening Entry, Unknown column 'warehouse' in 'where clause'

rohitwaghchaure:
**Issue**

78c469b12919dff986bee09c3d8c4cc5e3c9ee42

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -736,7 +736,7 @@ def get_opening_accounts(company):
 			is_group=0 and report_type='Balance Sheet' and company=%s and
 			name not in(select distinct account from tabWarehouse where
 			account is not null and account != '')
-		order by name asc""", company)
+		order by name asc""", frappe.db.escape(company))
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -733,10 +733,8 @@ def get_opening_accounts(company):
 	accounts = frappe.db.sql_list("""select
 			name from tabAccount
 		where
-			is_group=0 and
-			report_type='Balance Sheet' and
-			ifnull(warehouse, '') = '' and
-			company=%s
+			is_group=0 and report_type='Balance Sheet' and company=%s and
+			name not in(select distinct account from tabWarehouse where account is not null)
```
nabinhait: account is not null and account != ''

nabinhait (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.py
rohitwaghchaure: Fixed!

PR #9427, opened 2017-06-22
POS Sessions

chdecultot:
Hello,

In the scope of issue #7582, I started to work on the creation of POS Sessions, essential to generate X and Z reports.

This first PR is for the creation of a basic local session. It will need to be enhanced to allow multi-location sessions and to be linked to a particular user/pos profile.
The idea is to propose small PR to facilitate the work of the reviewers and also to generate some feedback going forward. So please don't hesitate to comment and give suggestions regarding the best way to move forward :smiley: 

**Changes in this PR**

The new workflow proposed in this PR is the following:
1. A user navigates to the POS module.
2. If a session is already open, he will be directed to the main POS page. 
If no session is opened, he will be directed to a page proposing him to start a new session.
3. The session start date is registered in LocalStorage and passed to all invoices generated through the POS.
4. At the end of the day, the  user can decide to close his session by clicking on "End Session" in the menu.
5. All invoices containing the POS start date of the current session will be updated with the POS end date and the POS session start date will be cleared from LocalStorage.

__pos.py__

The python file contains a new sql query to update all invoices generated during the current session with the pos session end date.

__sales_invoice.js__

Two new fields are created to store the pos session start and end date.
I also changed the pos_profile field to make it non editable (in case the invoice is passed as draft).

__pos.js__

Modification of onload to redirect to the pos_session_start if no session if currently opened.
Creation of a new pos session start date if none currently exists.
Addition of a menu button to close the session.

__build.json__

Addition of the pos_session_start.html file

__pos_session_start.html__

Creation of a new page to initialize a new session.
For now it present the user image, but it will need to be enhanced later to propose all available pos profiles linke to an user (if he has multiple pos profile).

**Workflow screenshots**

1. A user connects to the POS without any open session
image

2. A new POS session is created
image

3. The invoice generated contains the POS session start date
image

4. The user ends the current session
image

image

pos_session_start is removed:
image

5. All invoices from this session are updated with the POS session end date
image

**Next Steps**

- Link a session to a single user/pos profile
- Allow multi-location sessions

This PR is linked to a PR in Frappe for a new css class: "avatar-xl"

Best regards,

Charles-Henri

PR #9600, opened 2017-07-03
POS Sessions

chdecultot:
Note: This is a second version of PR #9427
---
Hello,

In the scope of issue #7582, I started to work on the creation of POS Sessions, essential to generate X and Z reports.

This first PR is for the creation of a basic local session. It will need to be enhanced to allow multi-location sessions and to be linked to a particular user/pos profile.
The idea is to propose a first PR to facilitate the work of the reviewers (sorry it was difficult to make it smaller without breaking the functionality) and also to generate some feedback going forward. So please don't hesitate to comment and give suggestions regarding the best way to move forward :smiley: 

**Changes in this PR**

The new workflow proposed in this PR is the following:
1. A user navigates to the POS module.
2. If a session is already open, he will be directed to the main POS page. 
If no session is opened, he will be directed to a page proposing him to start a new session.
3. The session start date is registered in LocalStorage and used to generate a new POS Session that will store links to all invoices registered during that session.
4. At the end of the day, the  user can decide to close his session by clicking on "End Session" in the menu.
5. The current POS session will be updated with the POS end date and both the POS session start date and POS session name will be cleared from LocalStorage.

__pos.py__

The python file contains three new functions to start a new session, end an existing session and update an existing session.

__sales_invoice.js__

A new field is created to store the name of the POS Session.
I also changed the pos_profile field to make it non editable (in case the invoice is passed as draft).

__pos.js__

Modification of onload to redirect to the pos_session_start if no session if currently opened.
Creation of a new pos session start date and pos session name if none currently exists.
Addition of a menu button to close the session.

__build.json__

Addition of the pos_session_start.html file

__pos_session_start.html__

Creation of a new page to initialize a new session.
For now it present the user image, but it will need to be enhanced later to propose all available pos profiles linke to an user (if he has multiple pos profile).

**Workflow screenshots**

1. A user connects to the POS without any open session
image

2. A new POS session is created
image

3. The invoice generated is added to the POS Session during the synchronization
image

4. The user ends the current session
image

image

pos_session_start is removed:
image

5. The POS  session is updated with the POS session end date
image

**Next Steps**

- Link a session to a single user/pos profile
- Allow multi-location sessions

Best regards,

Charles-Henri


felixvarghese:
1. If no session is open, is there a need to ask the user to explicitly click "New Session" - why not just create a new session in the background? Or show a message "Creating New Session" for 1-2 seconds before opening the POS. And if there is already a session open, show a "Continuing Existing Session Started on ________" message for 1-2 seconds.
2. Maybe change the "Submit" and "Close" text in the End POS Session Modal to "No" and "Yes" respectively - it makes more semantic sense.

chdecultot:
Hi @felixvarghese ,

Thanks for your feedback.

Regarding the first point, as of now your proposal makes sense since each user can only be linked to one POS Profile.
But since multiple POS Profiles will be implemented in a near future, users will be required to choose the adequate POS Profile, so this feature will not be able to execute automatically in the background.
Additionally, since it is an important action (opening and closing sessions is a legal requirement in France for example) I would prefer to have users make a deliberate decision. But it is of course only my humble opinion.

The message "Continuing Existing Session..." is an excellent idea. I will add it in a second PR (this one is already too big but I couldn't really break it in smaller parts... :disappointed: )

I will change immediately the "Close" and "Submit" buttons, you are totally right.

felixvarghese:
Regarding the Start POS Screen, with multiple POS Profiles, the screen is actually not mandatory. If a user only has the option of one profile, then ideally the screen wouldn't appear and the only profile is selected by default. And I think a majority of users actually would only have one POS Profile, so the idea would be to make the multiple profiles option as transparent as possible for them. Only users with multiple profiles would ever see the start screen.

So, for people with one profile, the start session can be executed in the background. And for people with multiple profiles, they would have to select their profile and start session.

The next question will be, should sessions be required for all companies/organizations using ERPNext POS. Some will care about sessions, and some will not need them. Are we OK with forcing sessions on everyone.

One option would be to add a setting in "Selling Settings" which enables or disables sessions. I don't like this option because "adding a setting" is often a lazy solution, but the reason this may be the only option is that certain reports like X&Z which would depend on sessions could also be disabled if a user doesn't select the option of using sessions.

I realize both these ideas add complexity, but they also add value by hiding complexity when not needed. Where the balance is, I don't know.

It might be good for others to comment and give their ideas as well...

chdecultot:
I do agree that the "open a new session" screen is not mandatory and is an additional step for the user, therefore not great for user experience. 
To be honest I gave a lot of thoughts to the process and can't find a way to have a smooth process without compromising with the quality of the data. Same as you, I don't know where is the balance :wink: 

Opening a new session must be limited to real sessions: what if a user clicks on the POS icon on Desk, starts automatically a new session without noticing it and don't close it ? 
In this case you will have a session opened for a few days maybe before it is closed. In term of data registration it is not very rigorous..

In regular POS systems, a cashier sometimes doesn't even have the rights to open or close a session... We are not there yet but must be careful with how these sessions are started and closed.

And again, sorry if it feels very self-centered :wink: , but I must insist on the fact that opening and closing a POS session has legal implications in France and I guess in some other countries also.

Regarding the possibility to activate POS Sessions for certain implementations or not, again I do agree with you that it could be an option.
Again, only because of legal requirements (a company should not be able to disable this option), I would recommend to have this option in site_config.json or somewhere not accessible by a system manager (an administrator would be fine since it can be a third party that certifies the software is in line with the legislation).

erpnext/accounts/doctype/sales_invoice/sales_invoice.json
```diff
@@ -347,7 +347,37 @@
...
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "pos_session_name",
```
rmehta: `pos_session` should be a link field ?

erpnext/accounts/doctype/pos_session_invoices/pos_session_invoices.json
```diff
@@ -0,0 +1,108 @@
+{
```
rmehta: Why child table in POS Session? Since an Invoice can have only one POS Session, it makes sense to link the POS Session to the invoice and not add a child table. Maybe a button in POS Session can be added to view invoices linked to this session.

Also doctype names are always in singular, like "POS Session Invoice" not invoices. Since now row represents only one invoice :)

erpnext/accounts/page/pos/pos.js
```diff
@@ -90,6 +137,38 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 			me.make_payment();
 		}).addClass('visible-xs');
 
+		this.page.add_menu_item(__("End Session"), function () {
+					var d = new frappe.ui.Dialog({
```
rmehta: lets stick to one-indent instead of hanging indents.

rmehta (changes requested):
Please see comments. Child table not required.

This is great improvement on the earlier PR 👍

rmehta:
Closing for now. Please re-open!

erpnext/accounts/doctype/sales_invoice/sales_invoice.json
chdecultot: You are totally right, that's a mistake. I'll correct it.

erpnext/accounts/doctype/pos_session_invoices/pos_session_invoices.json
chdecultot: Yes, the idea behind the child table was to have some sort of control log. 
For compliance reasons, we must ensure that if an invoice is deleted by an user, we have a trace of that deletion. Therefore I thought of having a link in the invoice and a log of all registered invoices in the POS Session. 
It was only a first step, since this log must also contain a SHA-2 key to guarantee it's inalterability.

I will delete the child table in this table for now and will make another PR for this log issue. Please let me know if you have any better idea of doing it.

erpnext/accounts/page/pos/pos.js
chdecultot: :+1:

erpnext/accounts/doctype/pos_session_invoices/pos_session_invoices.json
rmehta: - Anything that is deleted goes in the table "Deleted Document". 
- Also having a numbering series is also a good way to identify missing invoices!

Depends on what is "good enough". Is there a definition for this?

chdecultot: In a normal use case it would be largely good enough in my opinion. The problem here is that the compliance requirement doesn't consider this good enough at all.

We must ensure that people who want to fraud can't do it, which is much harder :wink: 
And the french anti-fraud department considers that it is the role of the software to ensure, people can't fraud.

As of now, when one uses the POS, he can easily delete the invoice after the transaction by clearing LocalStorage. We should have a cryptographic log ensuring that if they do it, at least we can prove it.

A system manager can also get the authorization to delete items in the "Deleted Document" which is not considered acceptable by the government.

Numbering series are a good way to identify missing invoices and are also a legal requirement, the only issue here is that people can rename their invoices if they want to fraud. So we need to ensure that if they do so, we have a trace...

I would prefer people didn't fraud in the first place :smile:

PR #9601, opened 2017-07-04
Adds Whitelist Method for Advance Payment Entry

Olawale1:
This Pull Request adds a whitelist method to the expense_claim.py file that helps in making a payment entry for Advance Payments (i.e. Imprest) made to employees

Basically, it creates a new Journal Entry and sets the Employee's Advance Account (custom field) in the debit entry while the default bank/cash account is set on the credit side

Kind regards,

### DIFF
--- a/erpnext/hr/doctype/expense_claim/expense_claim.py
+++ b/erpnext/hr/doctype/expense_claim/expense_claim.py
@@ -247,3 +247,37 @@ def get_expense_claim_account(expense_claim_type, company):
 	return {
 		"account": account
 	}
+
+@frappe.whitelist()
+def make_advance_entry(docname):
+	from erpnext.accounts.doctype.journal_entry.journal_entry import get_default_bank_cash_account
+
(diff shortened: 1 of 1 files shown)


rmehta:
@Olawale1 thanks for the contribution :)

1. Where is this called from?
2. Test Cases

Olawale1:
Hi @rmehta 

Thanks for your feedback. I finally figured it out :)

1. Currently it's being called using a Custom Button and Custom Script
2. Please see below:
advance payment patch1
advance payment patch2
advance payment patch3

Kindly let me know if anything further is required and how to proceed

Thanks again

Olawale1:
By the way, I'd be happy to add the js code as well if you think it's something that can be made available to all users at this stage. My only concern is how to include the custom button and custom field since those were added from the UI

Or we could just leave this hidden for now 

Thanks

rmehta:
I don't see the JS part in the PR. Please fix and re-open

Issue #9574, opened 2017-07-02
Test Coverage for Selling


rmehta:
- [x] Lead
- [x] Opportunity
- [x] Quotation
  - [x] Address and Contact Selection (select a new address / contact)
  - [x] Price List Selection
  - [x] Manually setting price
  - [x] Tax Template
  - [x] Terms
  - [x] Multi Currency
  - [x] Flow: Make Sales Order
  - [x] Save and Edit
  - [x] Submit, Cancel and Amend
  - [x] Print Format
- [x] Sales Order
  - [x] Reservation

PR #9607, opened 2017-07-04
Added Lead fixture, Added test for Lead, Opportunity

scmmishra:
For #9574 

test_7_july


rmehta:
rebase and push again so tests pass

Issue #9630, opened 2017-07-05
Print Format - Multiple Letter Heads fetching


umairsy:
In the Print Format of Report, the Letter Head of Company selected in the Report is always fetched, even if we don't check Letter Head. And if we check Letter Head, then selected Letterhead, as well as companies default Letter Head is fetched.

![report-print](https://user-images.githubusercontent.com/910238/27856996-9ef246be-618d-11e7-9567-01770d02d0ea.gif)

PR #9644, opened 2017-07-05
[Fix] Multi letter head showing in the print

rohitwaghchaure:
27856996-9ef246be-618d-11e7-9567-01770d02d0ea
For the print formats
Accounts Receivable
Accounts Payable
General Ledger

Fixed https://github.com/frappe/erpnext/issues/9630


me2ne:
It's great idea

PR #9652, opened 2017-07-05
Added lead time field to supplier quotation

bcornwellmott:
Used to track how long it takes to get something from your supplier.
image


nabinhait:
To maintain integrity with "Lead Time" field in Item master, it will be better to keep one field "Lead Time (Days)" instead of 2 fields "Lead Time" and "Lead Time Unit".

bcornwellmott:
Updated to match Item master.

nabinhait:
@bcornwellmott Will it be a good idea if we fetch default lead_time from Item master?

bcornwellmott:
I would say no. The idea is to get the value from the supplier, as each supplier will have a different lead time. 

I think there's a missing path to setting the default supplier for an item, and part of filling that could be setting the lead time, but right now none of that is there. Furthermore, the Item Price probably could have a lead time associated with it (instead of using a generic one for the item), but I'd say that's a separate thing.

nabinhait:
@bcornwellmott Ok, merging the PR

PR #9658, opened 2017-07-06
Invoice copy, state code in gst print format, hsn code in other sales/purchase docs

nabinhait:
screen shot 2017-07-06 at 11 34 59 am


sumitdas4u:
when will it be released?

rohitwaghchaure (approved):

PR #9690, opened 2017-07-07
[fix] reset flag after promise

rmehta:


erpnext/public/js/controllers/transaction.js
```diff
@@ -69,31 +69,30 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
-			delete frm.via_discount_percentage;
+			frm.set_value("discount_amount", discount_amount)
+				.then(() => delete frm.via_discount_percentage);
 		});
 
 		frappe.ui.form.on(this.frm.doctype, "discount_amount", function(frm) {
 			frm.cscript.set_dynamic_labels();
 
 			if (!frm.via_discount_percentage) {
-				frm.via_discount_amount = true;
```
nabinhait: You have removed this, but it is used in line 66

nabinhait (changes requested):

erpnext/public/js/controllers/transaction.js
rmehta: fixed

PR #9723, opened 2017-07-07
Sales Goal by Company

pratu16x7:
Depends on frappe/frappe#3641

screen shot 2017-07-10 at 10 21 18 am

------------------------------------------------------
Settable in Company:
screen shot 2017-07-07 at 11 51 16 pm

------------------------------------------------------
Calculated by Invoices:
screen shot 2017-07-10 at 4 59 35 am

------------------------------------------------------
Asked in Setup Wizard:
screen shot 2017-07-07 at 10 50 35 pm


rmehta:
- [ ] Show line graph for last 12 months... daily sales makes no sense.
- [ ] keep consistent font size...

pratu16x7:
screen shot 2017-07-12 at 10 54 35 am

Graph is shown regardless of goal setting; goal stats will be set to zero(no field value) by default.
Is a patch required to set a default goal based on past data?

pratu16x7:
Travis fail due to frappe dependency

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -138,6 +138,7 @@ def on_submit(self):
 			self.update_against_document_in_jv()
 
 		self.update_time_sheet(self.name)
+		update_company_monthly_sales(self.company)
```
rmehta: don't call this after every submit - this will be expensive for large data steps... let targets be calculated only by the scheduler.

erpnext/setup/doctype/company/company_dashboard.py
```diff
@@ -0,0 +1,40 @@
+from frappe import _
+
+def get_data():
+	return {
+		'heatmap': True,
+		'heatmap_message': _('This is based on transactions against this Company. See timeline below for details'),
+
+		'graph': True,
+		'graph_method': "frappe.utils.goal.get_monthly_goal_graph_data",
```
rmehta: cache this!

rmehta (changes requested):

erpnext/patches/v8_3/update_company_total_sales.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2017, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+from erpnext.accounts.doctype.sales_invoice.sales_invoice import update_company_monthly_sales
+
+def execute():
+	'''Update company monthly sales field based on sales invoices this month'''
+	companies = [d['name'] for d in frappe.get_list("Company")]
```
mbauskar: @pratu16x7,

Need to reload the company doctype as `update_company_monthly_sales` update's the `total_monthly_sales` field in company

Also add the patch entry in patches.txt

erpnext/setup/doctype/company/company.py
```diff
@@ -310,3 +310,12 @@ def get_name_with_abbr(name, company):
 		parts.append(company_abbr)
 
 	return " - ".join(parts)
+
+def cache_companies_monthly_sales_history():
+	'''Cache past year monthly sales of every company based on sales invoices'''
+	from frappe.utils.goal import get_monthly_results
+	companies = [d['name'] for d in frappe.get_list("Company")]
+	for company in companies:
+		filter_str = 'company = "_Test Company" and status != "Draft"'
```
mbauskar: _Test Company?

erpnext/setup/doctype/company/company.py
```diff
@@ -310,3 +310,12 @@ def get_name_with_abbr(name, company):
...
+		month_to_value_dict = get_monthly_results("Sales Invoice", "grand_total", filter_str, "sum")
+		frappe.cache().hget("month_to_value_dict:Sales Invoice", "Company:" + company, month_to_value_dict)
```
mbauskar: hget is use to get the cache data it should be hset

mbauskar (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -918,6 +919,27 @@ def update_item(source_doc, target_doc, source_parent):
...
+			company = "{1}"
+		group by
+			month_year;
+	''').format(current_month_year, company), as_dict = True)
+
+	monthly_total = results[0]['total'] if len(results) > 0 else 0
+
+	frappe.db.sql(('''
+		update tabCompany set total_monthly_sales = {0} where name="{1}";
+	''').format(monthly_total, company))
```
mbauskar: escape company name using `frappe.db.escape`

mbauskar (changes requested):

pratu16x7:
Travis fail due to frappe dependency

mbauskar (approved):

pratu16x7:
Now with frappe/frappe#3711

pratu16x7:
Another fix: frappe/frappe#3720

vladucoju:
Hi, i like this feature. I would suggest to add fields for month goals for separate insert. Goals for each month doesn't have same value. Or maybe last year sales line to be option on check box. I also must say that i don't find graphs usable in ERPNEXT because we don't have option to have them on report. The good option will be when we print in PDF to have check option to include chart, like we have for print header, but that is not part of this topic.

PR #9746, opened 2017-07-10
don't send deleted doctype for updating naming series

### LINKED ISSUE #9665: Error when opening Naming series document
Reported via https://frappe.erpnext.com/desk#Form/Issue/WN-SUP26178
naming series error

manassolanki:
If a table isn't deleted from the database and doctype is deleted then don't show the table name in option and for updating the naming series.

naming series

Fixes https://github.com/frappe/erpnext/issues/9665

### DIFF
--- a/erpnext/setup/doctype/naming_series/naming_series.py
+++ b/erpnext/setup/doctype/naming_series/naming_series.py
@@ -25,14 +25,17 @@ class NamingSeries(Document):
 		prefixes = ""
 		for d in doctypes:
 			options = ""
+			exist_flag = True
 			try:
 				options = self.get_options(d)
 			except frappe.DoesNotExistError:
+				exist_flag = False
 				frappe.msgprint('Unable to find DocType {0}'.format(d))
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/naming_series/naming_series.py
```diff
@@ -25,14 +25,17 @@ def get_transactions(self, arg=None):
 		prefixes = ""
 		for d in doctypes:
 			options = ""
+			exist_flag = True
```
rmehta: better name for this is `found_all = True`

Still did not understand the use case?

rmehta (changes requested):

nabinhait:
This is not a correct fix. Actual issue is much severe than you see.

The doctype "Student Batch" is appearing in naming series because "Custom DocPerm" was not deleted while deleting the doctype using `delete_doc` function.

Sent a PR to fix the actual issue (https://github.com/frappe/frappe/pull/3654)

PR #9759, opened 2017-07-10
POS Sessions

chdecultot:
Hi,
Reopening PR following comments on #9600 

Child table in POS Session has been removed (as discussed, an alternative solution to log submitted invoices needs to be found).

Indent has been corrected.

Merge conflicts after rebase have been fixed.


erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -480,3 +480,23 @@ def save_invoice(e, si_doc, name):
 		si_doc.flags.ignore_mandatory = True
 		si_doc.due_date = si_doc.posting_date
 		si_doc.insert()
+
+@frappe.whitelist()
+def start_pos_session(pos_session_start):
+	new_session = frappe.get_doc({
+					"doctype" : "POS Session",
```
mbauskar: please use the single tab for indentation

mbauskar (changes requested):
@chdecultot,

Thanks for the PR, please also add the Documentation?

chdecultot:
Hi @mbauskar ,

I have updated the documentation, rebased and corrected the conflict in the test file and corrected the indent issue in pos.py

Let me know if that's ok! :smiley: 

Thanks!

rmehta:
@chdecultot sorry dropped the ball on this one. 

@netchampfaris is working on major refactoring of the POS. One of the features we will add is disabling the Offline feature, so that should take care of the security problem your PR is trying to handle. The user will not be able to print before it is inserted in the database. Does that satisfy your requirement?

chdecultot:
Hi @rmetha,

Thanks and no problem! I know you have a lot on your plate right now 😊
Adding the possibility to disable the offline POS will be a great thing for the french compliance issue. 
It doesn't fulfill all regulatory requirements (Damn french laws! 😉) but is a necessary step forward. Thanks for taking the users feedback into consideration and for working on it!

The POS Session is meant to enable the generation of X and Z reports which are mandatory in a lot of countries and can be useful for all POS users regardless of the country laws. So we will need to add this feature anyway after the refactoring of the POS.

Therefore, I propose to close this PR and I will make another proposal after @netchampfaris has merged the refactored POS. Is it fine for you ?

PR #9776, opened 2017-07-10
[fix] validate multiple billing

YuriyGural:
Solution issue about error message while creating Purchase/Sales invoice with the revised rate.
Discuss:
https://discuss.erpnext.com/t/stock-setting-error-message-while-creating-purchase-invoice/21452


nabinhait:
Currently the system validates over-billing based on amount. We can't directly changed it based on qty. There is other github issue for making configurable whether it should be based on qty or amount. Currently "Billed Qty" is also not updated in SO or DN, which should also be fixed with this PR. 

Closing this, we will fix the behaviour very soon.

PR #9785, opened 2017-07-11
Test lead

utkarsh-goswami:


erpnext/tests/ui/test_sell.js
```diff
@@ -0,0 +1,33 @@
+/*QUnit.module('sales');
+
+QUnit.test("test quotation", function(assert) {
```
rmehta: why commented?

erpnext/tests/ui/test_sellling.js
```diff
@@ -117,4 +117,89 @@ QUnit.test("test: quotation", function (assert) {
...
+			assert.ok(cur_frm.doc.enquiry_from=='Lead');
+			assert.ok(cur_frm.doc.enquiry_type=='Sales')
+		},
+		() => done()
+	]);
+});
+
+// Test for creating query report
+
+QUnit.test("test building report", function(assert) {
```
rmehta: this should be in frappe (test_report.js)

erpnext/tests/ui/test_sellling.js
```diff
@@ -117,4 +117,89 @@ QUnit.test("test: quotation", function (assert) {
...
+			assert.ok(cur_frm.doc.report_type=='Query Report');
+		},
+		() => done()
+	]);
+});
+
+
+//Test for generating report with the help of writing query
+
+QUnit.test("test query report", function(assert) {
```
rmehta: tests cannot be dependent on each other.. make the report as a test_fixture

rmehta (changes requested):

PR #9797, opened 2017-07-11
Test cases for naming series

rohitwaghchaure:
Test cases are failing because of dependent PR https://github.com/frappe/frappe/pull/3661


rmehta:
Tests failing...

just check the naming series creation.. no need to test with full Sales Invoice

PR #9809, opened 2017-07-12
[fix] Don't add newline at the end, onchange of serial_no and filter warehouse based on company and is_group

nabinhait:


rmehta:
Would have been nice to see an animated gif as a proof of working....

PR #9811, opened 2017-07-12
[Fix] Manual discount is not applying in the transaction if discount amount is zero in the pricing rule

### LINKED ISSUE #9664: Pricing Rule - #Bug Discount is mandatory
Hi,
This is my first issue on Github. (My ERP system is hosted at frappe, so don't have any admin rights)
- Objective = Apply a % margin for a specific set of products and limit the **Max** discount.

   I tried using pricing rule, in this particular instance I want to apply a 15% (on price list) margin for a particular product range.
The system is mandating me to insert a Price/discount, and when I put 0%, then it overrides any user applied discount to back to 0%
image

If we take the regular use case,
- User should be able to select either Margin or Discount or both.

- For discounts, either it could be a fixed discount or a limit on the max discount

Can this issue be resolved?

Best regards

rohitwaghchaure:
**ISSUE**

Objective = Apply a % margin for a specific set of products and limit the Max discount.

I tried using pricing rule, in this particular instance I want to apply a 15% (on price list) margin for a particular product range.
The system is mandating me to insert a Price/discount, and when I put 0%, then it overrides any user applied discount to back to 0%

Fixed https://github.com/frappe/erpnext/issues/9664

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -185,7 +185,7 @@ def get_pricing_rule_for_item(args):
 				"discount_percentage": 0.0
 			})
 		else:
-			item_details.discount_percentage = pricing_rule.discount_percentage
+			item_details.discount_percentage = pricing_rule.discount_percentage or args.discount_percentage
 	elif args.get('pricing_rule'):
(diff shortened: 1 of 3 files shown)


mbauskar (approved):

drthilina:
Linked with #9811 .
      Is it possible to set a "Max Discount" as a third option?

PR #9812, opened 2017-07-12
Changed state codes and added state

adityaduggal:
-Andaman 35 added
-UP changed to 09 from 35
-Uttrakhand changed to 05 from 36

### DIFF
--- a/erpnext/regional/india/__init__.py
+++ b/erpnext/regional/india/__init__.py
@@ -1,5 +1,6 @@
 states = [
  '',
+ 'Andaman and Nicobar Islands',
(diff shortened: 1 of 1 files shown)


adityaduggal:
@nabinhait Can we merge this to the master branch since i am unable to run bench update since I have made the changes in my production server for this and due to this we are facing a problem with JS scripting at the user end.
I have tried `bench update --patch --no-backup` and then `bench build` but the the documents are stuck at saving and I presume its because I have not run `bench update` as it asks me to stash my changes to the erpnext app.

nabinhait:
released

PR #9849, opened 2017-07-13
[fix] correctly choice 'Credit Limit'

YuriyGural:
If 'Credit Days Based On' is empty in 'Supplier Type/Customer Group' then value 'Due Date' in 'Purchase/Sales Invoice' doesn't taken from the default settings.


erpnext/accounts/party.py
```diff
@@ -282,12 +282,14 @@ def get_credit_days(party_type, party, company):
...
-				frappe.db.get_value("Customer Group", customer_group, ["credit_days_based_on", "credit_days"]) \
-				or frappe.db.get_value("Company", company, ["credit_days_based_on", "credit_days"])
+				frappe.db.get_value("Customer Group", customer_group, ["credit_days_based_on", "credit_days"])
 		else:
 			credit_days_based_on, credit_days = \
-				frappe.db.get_value("Supplier Type", supplier_type, ["credit_days_based_on", "credit_days"])\
-				or frappe.db.get_value("Company", company, ["credit_days_based_on", "credit_days"] )
+				frappe.db.get_value("Supplier Type", supplier_type, ["credit_days_based_on", "credit_days"])
+
+	if not credit_days_based_on:
```
nabinhait: `if not (credit_days_based_on and credit_days):`?

nabinhait (changes requested):

nabinhait:
Closing, please reopen after fixing.

PR #9852, opened 2017-07-13
Timesheets on Web Portal

vishdha:
Summary
Timesheet web portal created to show timesheets details.
screenshot from 2017-07-13 23-24-04


erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -389,7 +390,27 @@ def get_conditions(filters):
...
 				dt = 'tabTimesheet Detail'
-				
+
 			conditions.append("`%s`.%s = '%s'"%(dt, key, filters.get(key)))
 
 	return " and {}".format(" and ".join(conditions)) if conditions else ""
+
+
+
+def get_t_list(doctype, txt, filters, limit_start, limit_page_length=20, order_by="modified"):
```
rmehta: bad function name

erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -389,7 +390,27 @@ def get_conditions(filters):
...
+	user = frappe.session.user
+	return frappe. db.sql('''SELECT A.name, B.activity_type, A.status, A.total_billable_hours, B.sales_invoice, B.project  FROM \
```
rmehta: instead of A and B, use `ts`, `tsd` as aliases

rmehta (changes requested):

rmehta:
Please add permissions

PR #9874, opened 2017-07-14
Renamed report Support Hours to Support Hours Distribution

rohitwaghchaure:
New look of report

**Graph**
screen shot 2017-07-14 at 8 04 44 pm

**Report**

screen shot 2017-07-14 at 7 34 58 pm


erpnext/support/report/support_hour_distribution/support_hour_distribution.py
```diff
@@ -23,26 +23,29 @@ def execute(filters=None):
...
-	data = get_data(filters)
-	return columns, data
+	data, timeslot_wise_count = get_data(filters)
+	chart = get_chartdata(timeslot_wise_count)
+	return columns, data, None, chart
 
 def get_data(filters):
 	start_date = getdate(filters.from_date)
 	data = []
+	timeslotwise_total_count = {}
```
rmehta: why not `time_slot_wise_total_count` ?

erpnext/support/report/support_hour_distribution/support_hour_distribution.py
```diff
@@ -70,4 +73,25 @@ def get_columns():
...
 		})
 
-	return columns
\ No newline at end of file
+	return columns
+
+def get_chartdata(timeslot_wise_count):
+	x_interval = ['x']
+	total_count = ['Total']
+	timeslots = ['12AM - 3AM', '3AM - 6AM', '6AM - 9AM',
```
rmehta: will be more readable as `['12a - 3a', '3a - 6a' ... ]`

rmehta (changes requested):

erpnext/support/report/support_hour_distribution/support_hour_distribution.py
rohitwaghchaure: Have used '12AM - 3AM' in the report column, need to remove from there too?

PR #9878, opened 2017-07-15
Make 'Financial Year' translatable in Setup wizard

strixaluco:


strixaluco:
Just realized I'm pushing from erpnext:localize. Apologize.

PR #9891, opened 2017-07-17
Allow to select asset account in the payable in the expense claim for imprest management

### LINKED ISSUE #9867: Payable Account filteration
As per the current validation, Payable Account in the Expense Claim filtered based on Root Type as well.

https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/expense_claim/expense_claim.js#L229

Can we remove this validation and allow selection of all the Payable Accounts. That will be helpful for the Company who use ERPnext for imprest management as well.

rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/issues/9867


umairsy:
We have made changes from our end which will be released in your ERPNext account by Wednesday.

PR #9902, opened 2017-07-17
[Fix] Sales invoice serial no validation issue

### LINKED ISSUE #9887: Sales Invoice - Serial No. validation error
WN-SUP26535

Steps:

1. Create Delivery Note for the 2 qty of a Serialized Item.
2. Created Sales Invoice for 1 Qty of item. However, in the Sales Invoice Item table, both the Serials Nos. are referenced.
3. When try creating another Sales Invoice for the last one item, getting an error message as follows.

screen shot 2017-07-17 at 10 49 43 am

Suggestion to resolve:

Perhaps we should ignore this validation when Update Stock is not checked in the Sales Invoice. Or let Serial No. field always be visible and editable, so that user can update values accordingly.

rohitwaghchaure:
**Issue**
Steps:

1. Create Delivery Note for the 2 qty of a Serialized Item.

1. Created Sales Invoice for 1 Qty of item. However, in the Sales Invoice Item table, both the Serials Nos. are referenced.

1. When try creating another Sales Invoice for the last one item, getting an error message as follows.

sales_invoice_sn_issue

Fixed https://github.com/frappe/erpnext/issues/9887


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -812,9 +812,22 @@ def validate_serial_numbers(self):
 		"""
 			validate serial number agains Delivery Note and Sales Invoice
 		"""
+		self.set_serial_no_against_delivery_note()
 		self.validate_serial_against_delivery_note()
 		self.validate_serial_against_sales_invoice()
 
+	def set_serial_no_against_delivery_note(self):
```
nabinhait: The serial nos should not be set forcefully on validate. The user should be able to change the serial nos. It should be set via mapper while creating Invoice from DN.

And serial nos will validated against DN, to ensure the serial no exists in DN.

nabinhait (changes requested):

nabinhait:
Closing , reopen after fixing.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
rohitwaghchaure: Fixed

erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -390,6 +391,9 @@ def set_missing_values(source, target):
 
 	def update_item(source_doc, target_doc, source_parent):
 		target_doc.qty = source_doc.qty - invoiced_qty_map.get(source_doc.name, 0)
+		if source_doc.serial_no and source_parent.per_billed >= 1:
```
nabinhait: source_parent.per_billed > 0

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -826,14 +834,18 @@ def validate_serial_against_delivery_note(self):
...
 
 			serial_nos = item.serial_no or ""
-			si_serial_nos = set(serial_nos.split("\n"))
+			si_serial_nos = set(get_serial_nos(serial_nos))
 
 			if si_serial_nos - dn_serial_nos:
 				frappe.throw(_("Serial Numbers in row {0} does not match with Delivery Note".format(item.idx)))
 
+			if item.serial_no and cint(item.qty) != len(si_serial_nos):
+				frappe.throw(_("Row {0}: Serial numbers {1} required for Item {2}. You have provided {3}.".format(
```
nabinhait: Serial numbers 5 required or 5 serial numbers required?

erpnext/stock/doctype/delivery_note/test_delivery_note.py
```diff
@@ -170,6 +170,10 @@ def test_serialized(self):
 			"delivery_document_no": dn.name
 		})
 
+		si = make_sales_invoice(dn.name)
+		si.insert(ignore_permissions=True)
+		self.assertEquals(dn.items[0].serial_no, si.items[0].serial_no)
+
```
nabinhait: write a test case for partial invoicing and if number of serial nos does not match with qty

nabinhait (changes requested):

erpnext/stock/doctype/delivery_note/test_delivery_note.py
rohitwaghchaure: Fixed

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
rohitwaghchaure: Fixed

Issue #9880, opened 2017-07-15
Strange Behavior with Serial No Selector


akshay83:
Please see: 
![serial entry](https://user-images.githubusercontent.com/7766272/28237509-ca19350e-695e-11e7-8aa7-f25795c8986e.gif)

Also getting the following error in the console:

```
get @ desk.min.js:1035
(anonymous) @ VM431:119
callback @ core.js:92
callback @ desk.min.js:1205
_ @ desk.min.js:1229
(anonymous) @ desk.min.js:1325
i @ jquery.min.js:2
fireWith @ jquery.min.js:2
z @ jquery.min.js:4
(anonymous) @ jquery.min.js:4
desk.min.js:4030 Uncaught TypeError: Promise.resolve is not a constructor
    at Class.validate_and_set_in_model (desk.min.js:4030)
    at Class.parse_validate_and_set_in_model (desk.min.js:4025)
    at HTMLInputElement.<anonymous> (desk.min.js:5348)
    at HTMLInputElement.dispatch (jquery.min.js:3)
    at HTMLInputElement.$event.dispatch (report.min.js:2942)
    at HTMLInputElement.r.handle (jquery.min.js:3)
    at HTMLInputElement.trigger (jquery.min.js:3)
    at Object.trigger (jquery.min.js:4)
    at HTMLInputElement.<anonymous> (jquery.min.js:4)
    at Function.each (jquery.min.js:2)
validate_and_set_in_model @ desk.min.js:4030
parse_validate_and_set_in_model @ desk.min.js:4025
(anonymous) @ desk.min.js:5348
dispatch @ jquery.min.js:3
$event.dispatch @ report.min.js:2942
r.handle @ jquery.min.js:3
trigger @ jquery.min.js:3
trigger @ jquery.min.js:4
(anonymous) @ jquery.min.js:4
each @ jquery.min.js:2
each @ jquery.min.js:2
trigger @ jquery.min.js:4
n.fn.(anonymous function) @ jquery.min.js:4
(anonymous) @ VM4052:203
dispatch @ jquery.min.js:3
$event.dispatch @ report.min.js:2942
r.handle @ jquery.min.js:3
i.fire @ libs.min.js:2
select @ libs.min.js:2
mousedown @ libs.min.js:2
```

P.S.: Using Licecap for the first time, the GIF would have some glitches.

PR #9905, opened 2017-07-17
[fix] serial no modal fixes

pratu16x7:
Some cleanup, with a fix for frappe/erpnext#9880
serial_no qty


nabinhait:
merged manually

Issue #9804, opened 2017-07-12
Hide "Make" button from form toolbar if there is no next action.


rmehta:
For a completed Purchase Order, the make button is shown, but the drop down is empty.

umairsy:
![screen shot 2017-07-14 at 5 23 51 pm](https://user-images.githubusercontent.com/910238/28211243-38391adc-68b9-11e7-9ddb-adba1f08d7d8.png)

rmehta:
@tundebabzy did you work on a generic version for this?

tundebabzy:
working on it

tundebabzy:
Fixed in frappe/frappe#3880

PR #9979, opened 2017-07-19
Hide "Make" button from form toolbar if there is no next action (#9804)

tundebabzy:
Reference: #9804 

### Before
28211243-38391adc-68b9-11e7-9ddb-adba1f08d7d8

### After
screenshot from 2017-07-19 20-25-45


rmehta:
@tundebabzy this problem also exists in other documents. 

Idea: Can we make a generic update that the button group will be cleared on refresh so that it will show up only if custom buttons are added?

tundebabzy:
for sure

rmehta:
let me know if you need help

PR #10015, opened 2017-07-21
Test to check if a billing cost is correctly calculated for a project

utkarsh-goswami:


erpnext/projects/doctype/project/test_project.js
```diff
@@ -0,0 +1,53 @@
...
+						{task: task.name},
+						{billable: '1'},
+						{billing_rate: 10},
+						{costing_rate: 8}
+					]
+				]}
+			]);
+		},
+		() => {
+			// To check if a correct billable and costing amount is calculated
```
rmehta: this should be in `test_timesheet`

erpnext/projects/doctype/project/test_project.js
```diff
@@ -0,0 +1,53 @@
...
+	let done = assert.async();
+	frappe.run_serially([
+		() => {
+			// Creating project with task
+			return frappe.tests.make('Project', [
+				{ project_name: 'Test App'},
+				{ expected_start_date: '2017-07-22'},
+				{ expected_end_date: '2017-09-22'},
+				{ estimated_costing: '10,000.00'},
+				{ tasks:[
```
rmehta: try with multiple tasks (2-3 at least)

rmehta (changes requested):

PR #10019, opened 2017-07-21
[Fix] Negative amount showing in the grand total for multicurrency if discount has applied

### LINKED ISSUE #10013: Getting negative multi currency value in Sales Order on setting discount on grand total.
Steps to Reproduce: 
  1. Create Sales Order with Customer name, item_code, delivery_date, qty, margin_type: percentage, discount_amount,  taxes and charges, and then discount on grand total, after that if change currency from INR to USD it show negative value on base_net_total, base_total
but after save it shows normal corrected value.

Version Number: 
erpnext(develop)
frappe(develop)

anim

rohitwaghchaure:
**Issue**
Discount amount is not converted after changing of currency
screen shot 2017-07-21 at 5 39 54 pm

**After Fixed**
screen shot 2017-07-21 at 5 30 25 pm

Fixed https://github.com/frappe/erpnext/issues/10013

### DIFF
--- a/erpnext/public/js/controllers/taxes_and_totals.js
+++ b/erpnext/public/js/controllers/taxes_and_totals.js
@@ -40,8 +40,10 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 	},
 
 	calculate_discount_amount: function(){
-		if (frappe.meta.get_docfield(this.frm.doc.doctype, "discount_amount"))
+		if (frappe.meta.get_docfield(this.frm.doc.doctype, "discount_amount")) {
+			this.set_discount_amount();
 			this.apply_discount_amount();
+		}
 	},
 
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -41,6 +41,7 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 
 	calculate_discount_amount: function(){
 		if (frappe.meta.get_docfield(this.frm.doc.doctype, "discount_amount"))
```
rmehta: no brackets???

rmehta (changes requested):

nabinhait:
Please send to hotfix branch.

PR #10061, opened 2017-07-24
Required On date for Material Request not auto populating after first instance (#9980)

### LINKED ISSUE #9980: Required On date for Material Request not auto populating after first instance
I used to be able to add the "Required On" date for the first line, and then all below it would automatically populate.  It is no longer doing it......  This is a major issue when I have to enter the date for a 200 line material request.

capture

tundebabzy:
Fixed #9980 

### Before
peek 2017-07-24 12-57

### After
peek 2017-07-24 12-43

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.js
+++ b/erpnext/stock/doctype/material_request/material_request.js
@@ -122,6 +122,18 @@ erpnext.buying.MaterialRequestController = erpnext.buying.BuyingController.exten
 
 	},
 
+	schedule_date: function(doc, cdt, cdn) {
+		var val = locals[cdt][cdn].schedule_date;
+		if(val) {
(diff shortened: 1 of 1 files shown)


nabinhait (changes requested):
I think this function was removed purposefully by @rmehta to cleanup the code. In that PR, he also added a new block of code https://github.com/frappe/erpnext/blob/36311020dee131969f1d5df88da4ebcb872eba7b/erpnext/stock/doctype/material_request/material_request.js#L39-L42

But forgot to write replacement of removed code. Can you write this using the standard function like below.

https://github.com/frappe/erpnext/blob/develop/erpnext/selling/doctype/sales_order/sales_order.js#L39-L43

PR #10062, opened 2017-07-24
[enhance] Provision to set default mode of payment for the POS in POS profile

### LINKED ISSUE #9855: POS - Possibility to choose the default mode of payment
Hi,

In the current POS, 'cash' is hard coded as the default mode of payment.
In reality, some shops will register more 'credit card' payments than 'cash' payments, or even sometimes 'cheques'.

It could be nice to configure the defaut mode of payment for each POS Profile for example.

I can make the changes if that's ok, but just one question, would it be easily possible to select only one checkbox in the 'Sales Invoice Payment' child table for example or do I need to make a custom JS script ?

#### For example if the default mode of payment is 'Cheque'

image

Thanks !

rohitwaghchaure:
**POS Profile**
screen shot 2017-07-24 at 5 53 20 pm

**POS**
screen shot 2017-07-24 at 5 55 22 pm

Fixed https://github.com/frappe/erpnext/issues/9855


erpnext/accounts/doctype/pos_profile/pos_profile.py
```diff
@@ -48,6 +49,15 @@ def validate_duplicate_groups(self):
 		if len(customer_groups) != len(set(customer_groups)):
 			frappe.throw(_("Duplicate customer group found in the cutomer group table"), title = "Duplicate Customer Group")
 
+	def check_default_payment(self):
+		if self.payments:
+			default = [d.default for d in self.payments if d.default]
```
nabinhait: default is a python keyword, use 'default_mode_of_payment'?

nabinhait (changes requested):

erpnext/accounts/doctype/pos_profile/pos_profile.py
rohitwaghchaure: Fixed!

mbauskar (approved):

mbauskar (changes requested):
@rohitwaghchaure ,

Also add the documentation and test cases !!

rohitwaghchaure:
Fixed!

PR #10068, opened 2017-07-24
Add RFQ email sent check

bcornwellmott:
Allows the user to see if an email has already been sent to a supplier for an RFQ. This will allow filtered lists and reports to show this, whereas the only way to check now is by looking at the communications feed.
rfq_email_sent


dufani1:
Thanks @bcornwellmott your changes amazing

PR #10096, opened 2017-07-26
Improve email format for training event, add ui test

bcornwellmott:
Currently, the training module is fairly limited. My main complaint has been that there's no easy way for an attendee to acknowledge if they are going to attend. They get an email saying they've been invited but there's no link to the document, minimal information about when/where it is (unless it's explained in the introduction) and if they eventually find the document on ERPNext they still need to find their name in the list, select the right field, then update it.

To this end, I've made the following changes:
1. Change the sendmail to a mail template (for more customization of the message)
2. Include event details in the email (including a link to the event)
3. Add buttons to automatically confirm or withdraw from the event right from the email
4. Add a new Self-Study type that allows you to simply confirm attendance (for simple watch/read this and then you are done training)
5. Allow attachments to the Training Event to be appended to the email (for forwarding training documents)

I've also created a basic UI test (not sure if I did it right), and updated the documentation.

* I'm really not sure if using a GET is the right way to grab data off the form when linking through the email, but I couldn't find another way to do it. If there's a better way, please let me know.

Here is an example where I send an email then accept the invitation:
te_new_features

Here is an example where I do a self-study and confirm completion (with an attached document):
te_new_features2


rmehta:
👍 for UI Tests! 😄

erpnext/hr/doctype/training_event/training_event.py
```diff
@@ -6,21 +6,79 @@
...
+		sender = frappe.session.user not in STANDARD_USERS and frappe.session.user or None
+		message = frappe.get_template(template).render(args)
+		attachments = self.get_attachments()
+
+		self.send_invitation_email(data, sender, subject, message, attachments)
+
+	def send_invitation_email(self, data, sender, subject, message, attachments):
+		email = frappe.db.get_value("Employee", data.employee, "company_email")
+		if email:
+			make(subject = subject, content=message,recipients=email,
```
netchampfaris: Why not use `frappe.sendmail`? It also supports templates directly

erpnext/hr/doctype/training_event/training_event.py
```diff
@@ -6,21 +6,79 @@
...
+		return attachments
+
+@frappe.whitelist(allow_guest=True)
+def set_response(event, response):
+	doc = frappe.get_doc('Training Event Employee', event)
+
+	if doc:
+		doc.status = response
+		doc.save()
+		frappe.msgprint("Status for this training event as been updated")
```
netchampfaris: Translated string

erpnext/hr/doctype/training_event/training_event.js
```diff
@@ -17,5 +17,37 @@ frappe.ui.form.on('Training Event', {
...
+				frappe.msgprint(__(frm.doc.name + ": Status for " + newTemp[0].employee_name +
+					" is updated to " + newTemp[0].status));
+				frappe.route_options = {};
+				frappe.set_route("List", "Training Event");
+			}
+		}
 	}
 });
+
+function getSearchParameters() {
```
netchampfaris: camelCase 😢

erpnext/hr/doctype/training_event/training_event.js
```diff
@@ -17,5 +17,37 @@ frappe.ui.form.on('Training Event', {
...
+	onload: function(frm) {
+		var params = getSearchParameters();
+		if (params.hasOwnProperty('employee') && params.hasOwnProperty('status')) {
+			var newTemp = frm.doc.employees.filter(function(obj) {
+				return obj.name == params.employee;
+			});
+			if (newTemp) {
+				newTemp[0].status = params.status;
+				frm.refresh_field("employees");
+				frappe.msgprint(__(frm.doc.name + ": Status for " + newTemp[0].employee_name +
```
netchampfaris: `frappe.msgprint(__('{0}: Status for {1} is updated to {2}', [arg0, arg1, arg2]))` works better for translated messages

erpnext/hr/doctype/training_event/training_event.js
```diff
@@ -17,5 +17,37 @@ frappe.ui.form.on('Training Event', {
 				frappe.set_route("List", "Training Feedback");
 			});
 		}
+	},
+	onload: function(frm) {
+		var params = getSearchParameters();
```
netchampfaris: I am not sure if this is the correct way to pass parameters from email.
Let's see what @rmehta has to say.

netchampfaris (dismissed):

erpnext/hr/doctype/training_event/training_event.py
bcornwellmott: I copied from the RFQ document, which I though was a good way to do it. Any reason not to follow that example?

bcornwellmott:
I've addressed some of the concerns mentioned by netchampfaris. I didn't change from make to sendmail, because I'm matching what the RFQ document is doing. If that's not the right way, I'll change it but I'm not clear on why that would be wrong.

I also didn't address getting parameters from email. I did it the only way I know how, so if there's another way then please let me know.

rmehta:
@bcornwellmott will test and merge this tomorrow! I am inkling towards a web form against the hash hack. Let me look at this and get back.

erpnext/templates/emails/training_event.html
```diff
@@ -0,0 +1,21 @@
+<h3>{{_("Training Event")}}</h3>
```
rmehta: this should be sent via email alert so the user can edit it.

rmehta:
check my comments here: https://github.com/frappe/erpnext/pull/10379

Not happy with my version either, but seems more standardized. What do you think?

PR #10101, opened 2017-07-26
Add erpnext-grey.png, add logo in footer

netchampfaris:
image


rmehta:
Unsubscribe is still very prominent. Our goal is not to make it obvious!

netchampfaris:
image

PR #10108, opened 2017-07-26
[minor] enable all roles and domain before running tests cases

mbauskar:
enable all the role and domain and add the roles to `test@erpnext.com` user

### DIFF
--- a/erpnext/setup/utils.py
+++ b/erpnext/setup/utils.py
@@ -114,10 +114,10 @@ def enable_all_roles_and_domains():
 	if not domains:
 		return
 
-	domain_settigns = frappe.get_doc("Domain Settings", "Domain Settings")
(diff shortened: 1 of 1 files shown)


erpnext/setup/utils.py
```diff
@@ -94,3 +95,29 @@ def get_exchange_rate(from_currency, to_currency, transaction_date=None):
...
+
+	# add all roles to users
+	user = frappe.get_doc("User", "test@erpnext.com")
+	user.add_roles(*[role.get("name") for role in roles])
+
+	domains = frappe.get_list("Domain")
+	if not domains:
+		return
+
+	domain_settigns = frappe.get_doc("Domain Settings", "Domain Settings")
```
rmehta: spelling...

PR #10120, opened 2017-07-26
Add disable check for Activity Type

bcornwellmott:
Currently, there is no way to disable or delete an Activity Type if you decide that activity is no longer valid (can't delete because of linked documents). Adding the checkbox prevents it from showing up in the list view, and there's now a validation to make sure they aren't disabled.

image


nabinhait:
Permission of Activity Type has been modified, please restore the permission and send the PR again.

PR #10125, opened 2017-07-27
Closed/completed projects are not shown in the portal#9342

ashwinisave35:
Now Completed project also shown in Project list:

screen shot 2017-07-27 at 12 00 04 pm

screen shot 2017-07-27 at 12 01 09 pm


nabinhait (changes requested):
In Sales Order and Delivery Note, completed project should not be shown. It uses same query for SO/DN as well.

PR #10129, opened 2017-07-27
[UI Test] Sales order with multi uom, multi tax, multi currency, tax template"

vishdha:


erpnext/selling/doctype/sales_order/test_sales_order.js
```diff
@@ -0,0 +1,107 @@
+$.extend(frappe.test_data, {
```
rmehta: add this in `test_item.js` (or use those items)

rmehta (changes requested):

PR #10156, opened 2017-07-28
[UI Test Salary] Test to check if the salary is correctly calculated for an employee

utkarsh-goswami:
Project Timesheet test updated
- Test Employee has been used. 

[UI Test Salary Structure]
- Created a Salary structure for test employee
- Checked if all the values of earning, deductions are correctly inserted.

[UI Test Salary Slip]
- Created a Salary Slip for test employee.
- Checked if all the amounts are correctly calculated.


erpnext/projects/doctype/project/project_timesheet.js
```diff
@@ -54,6 +59,15 @@ QUnit.test("test project", function(assert) {
 	};
 	frappe.run_serially([
 		() => {
+			// Creating Employee
+			return frappe.tests.make('Employee', [
```
rmehta: use test employee

rmehta (changes requested):

PR #10161, opened 2017-07-28
[ui test] expense claim type and expense claim in HR

uyadav212:


mbauskar:
@uyadav212,

Can you please check the and fix the travis issues and push the fix?

Issue #10163, opened 2017-07-28
Timesheet Company Issue


umairsy:
Steps:

1. Have two company, A and B in an account. A is the default.
1. Created Production Order for Company B.
1. In the Timesheet, auto-created on submission of Production Order has Company A updated in it.

Issue: In the Timesheet, Company should be picked from Production Order

PR #10165, opened 2017-07-28
[Fix] Timesheet Company Issue

rohitwaghchaure:
Steps:

Have two company, A and B in an account. A is the default.
Created Production Order for Company B.
In the Timesheet, auto-created on submission of Production Order has Company A updated in it.

Fixed https://github.com/frappe/erpnext/issues/10163

### DIFF
--- a/erpnext/manufacturing/doctype/production_order/production_order.py
+++ b/erpnext/manufacturing/doctype/production_order/production_order.py
@@ -273,7 +273,7 @@ class ProductionOrder(Document):
 		timesheets = []
 		plan_days = frappe.db.get_single_value("Manufacturing Settings", "capacity_planning_for_days") or 30
 
-		timesheet = make_timesheet(self.name)
+		timesheet = make_timesheet(self.name, self.company)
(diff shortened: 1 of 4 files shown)


erpnext/manufacturing/doctype/production_order/production_order.py
```diff
@@ -579,6 +579,7 @@ def make_timesheet(production_order):
 	timesheet = frappe.new_doc("Timesheet")
 	timesheet.employee = ""
 	timesheet.production_order = production_order
+	timesheet.company = frappe.db.get_value('Production Order', production_order, 'company')
```
nabinhait: Pass the company as an argument, instead of getting from db.

nabinhait (changes requested):

erpnext/manufacturing/doctype/production_order/production_order.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #10170, opened 2017-07-28
Stock in items list

ghost:
Hi,

This PR contains
- stock status in item list (in stock or not in stock)
- change the order by to display "in stock" items first, it is still possible to overload by using weightage in item master
- 4 items in each row, instead of 3, better UX for me

Thanks

item-list-stock

### DIFF
--- a/erpnext/setup/doctype/item_group/item_group.py
+++ b/erpnext/setup/doctype/item_group/item_group.py
@@ -60,7 +60,7 @@ class ItemGroup(NestedSet, WebsiteGenerator):
 
 	def get_context(self, context):
 		context.show_search=True
-		context.page_length = 6
+		context.page_length = 8
 		context.search_link = '/product_search'
 
 		start = int(frappe.form_dict.start or 0)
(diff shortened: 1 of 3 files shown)


mbauskar:
@britlog,

Thanks for contributing, can you also add the test user cases for the same, closing this PR for now, but you can reopen it.

britlog:
@mbauskar Hi, could you please explain ? Or give me one exemple. Thanks

britlog:
@mbauskar @rmehta What means "test user cases for the same" ? Please explain, I have planned to work on this point tonight. Thanks

rmehta:
@britlog A test case is a script that will test that the modifications you made actually work.

Here is what I would do:

1. add a file `test_product_search.py` (maybe in `erpnext/tests`)
2. write a test case which calls the method with some parameters and asserts that the requested data was returned.

Read this up for a primer:

https://frappe.io/docs/user/en/guides/automated-testing/unit-testing

Automated testing guarantees (upto a point) that your patches actually work and also if someone changes the behaviour in the future, they will also call that out.

Feel free to post any queries on the forum regarding this.

Thanks!

PR #10194, opened 2017-07-31
[ui-test] modified test_item to include child table assertions

codingCoffee:
- modified production_order to include child table assertions
- tested the entire production process


mbauskar:
@codingCoffee,

please check and fix the Travis issues

Issue #10205, opened 2017-08-01
Tax Rounding Issue


akshay83:
Please refer to issue faced: https://github.com/frappe/erpnext/issues/7918

I thought this might have been fixed via: #10135 

Therefore updated (I Know, I was in a Hurry). And unfortunately facing the following issue:

```
Uncaught (in promise) TypeError: this.set_in_company_currency is not a function
    at Object.<anonymous> (erpnext.min.js:1224)
    at Function.each (jquery.min.js:2)
    at Object.<anonymous> (erpnext.min.js:1189)
    at Function.each (jquery.min.js:2)
    at Class.calculate_taxes (erpnext.min.js:1187)
    at Class._calculate_taxes_and_totals (erpnext.min.js:1034)
    at Class.calculate_taxes_and_totals (erpnext.min.js:1006)
    at erpnext.min.js:1581
    at runner (form.min.js:2554)
    at form.min.js:2569
```
With my limited knowledge I changed `this.set_in_company_currency` to `me.set_in_company_currency`. But then I got the following error:

```
Uncaught (in promise) TypeError: this.frm.doc.get is not a function
    at Class.set_cumulative_total (erpnext.min.js:1236)
    at Object.<anonymous> (erpnext.min.js:1222)
    at Function.each (jquery.min.js:2)
    at Object.<anonymous> (erpnext.min.js:1189)
    at Function.each (jquery.min.js:2)
    at Class.calculate_taxes (erpnext.min.js:1187)
    at Class._calculate_taxes_and_totals (erpnext.min.js:1034)
    at Class.calculate_taxes_and_totals (erpnext.min.js:1006)
    at Class.price_list_rate (eval at setup (form.min.js:2615), <anonymous>:472:8)
    at runner (form.min.js:2552)
```

akshay83:
@mbauskar 

Thanks. But I think the issue still persists..the second error is still there.

```
Uncaught (in promise) TypeError: this.frm.doc.get is not a function
    at Class.set_cumulative_total (erpnext.min.js:1236)
    at Object.<anonymous> (erpnext.min.js:1222)
    at Function.each (jquery.min.js:2)
    at Object.<anonymous> (erpnext.min.js:1189)
    at Function.each (jquery.min.js:2)
    at Class.calculate_taxes (erpnext.min.js:1187)
    at Class._calculate_taxes_and_totals (erpnext.min.js:1034)
    at Class.calculate_taxes_and_totals (erpnext.min.js:1006)
[...]

PR #10209, opened 2017-08-01
[minor] TypeError: this.set_in_company_currency is not a function fixes

mbauskar:
fixes for https://github.com/frappe/erpnext/issues/10205


nabinhait:
Closing, fixed via https://github.com/frappe/erpnext/pull/10210

PR #10227, opened 2017-08-02
print format for the assessment report

manassolanki:
Depends on https://github.com/frappe/frappe/pull/3841

image


nabinhait:
@manassolanki Should Assessment Plan be comma separated? How the layout will look if there more assessment plan?

manassolanki:
image

PR #10230, opened 2017-08-02
Fix column width in GST Tax Breakup

netchampfaris:
Before:
image

After:
image


mbauskar:
@netchampfaris,

Travis test cases failed please check

PR #10248, opened 2017-08-02
[Enhancement] Notify Expense Claim Approver whenever Expense Claim is created

KanchanChauhan:


erpnext/hr/doctype/expense_claim/expense_claim.py
```diff
@@ -36,6 +36,10 @@ def validate(self):
 		if self.task and not self.project:
 			self.project = frappe.db.get_value("Task", self.task, "project")
 
+	def on_update(self):
+		if self.exp_approver and self.approval_status == "Draft":
+			self.notify_expense_claim_approver()
```
nabinhait: So, this email will go on every save of the Expense Claim, which is not so good. Add a button instead to send the email.

nabinhait (changes requested):

nabinhait:
Reopen after fixing.

KanchanChauhan:
Button added to notify approver.

screen shot 2017-08-03 at 11 34 21 am

nabinhait:
@KanchanChauhan Please add a custom button in the toolbar, it will be shown only in draft mode and if status is Draft.

KanchanChauhan:
screen shot 2017-08-03 at 12 39 44 pm

mbauskar:
@KanchanChauhan,

Thanks for PR.

Travis checks failed please check and fix the test cases. Closing the PR for now but you can reopen it.

rmehta:
Why not Email Alert?

nabinhait:
@rmehta Email Alert will go every time on saving the Expense Claim, which is not desired.

erpnext/hr/doctype/expense_claim/expense_claim.js
```diff
@@ -173,6 +173,17 @@ frappe.ui.form.on("Expense Claim", {
 			}, __("View"));
 		}
 
+		if (frm.doc.docstatus===0 && frm.doc.approval_status=="Draft"
+			&& frm.doc.exp_approver) {
+			frm.add_custom_button(__('Notify Expense Claim Approver'),
```
nabinhait: Can we change the label to "Notify Approver"?

nabinhait (changes requested):
Also please add a screenshot of the email

rmehta:
@nabinhait I have added a feature to add a flag in Email Alert (like "Email Sent"). If its true, then it will not re-sent 😄

nabinhait:
@rmehta Thats great 👍 But in this case, user can add more expenses after first saving, we really don't know when it is finalised.

rmehta:
Then set a property and trigger the email alert. the message should be configurable...

Issue #9511, opened 2017-06-28
Feature Missing - View Respective Ledger from the Account Heads


fasilkaks:
Hi
View ledger from the Account head feature is missing with the past few updates. This was an excellent feature which saved lot of our time. Please re enable this with the next update. 

Kindest Regards
Fasil

PR #10254, opened 2017-08-02
[Fix] View ledger button not disaplying in the head of an account

rohitwaghchaure:
**issue**
screen shot 2017-08-02 at 7 28 42 pm
**Fixed**
screen shot 2017-08-02 at 7 32 37 pm

As recently we have enabled the option User can not search for GL Entry doctype therefore user can not able to see View Ledger button

**GL Entry**
screen shot 2017-08-02 at 7 35 00 pm
 
Fixed https://github.com/frappe/erpnext/issues/9511

### DIFF
--- a/erpnext/accounts/doctype/account/account.js
+++ b/erpnext/accounts/doctype/account/account.js
@@ -68,7 +68,8 @@ frappe.ui.form.on('Account', {
 					}
 				});
 			});
-		} else if (cint(frm.doc.is_group) == 0) {
+		} else if (cint(frm.doc.is_group) == 0
+			&& frappe.boot.user.can_read.indexOf("GL Entry") !== -1) {
 			cur_frm.add_custom_button(__('Ledger'), function () {
 				frappe.route_options = {
 					"account": frm.doc.name,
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/account/account_tree.js
```diff
@@ -56,7 +56,7 @@ frappe.treeview_settings["Account"] = {
 	toolbar: [
 		{
 			condition: function(node) {
-				return !node.root && frappe.boot.user.can_read.indexOf("GL Entry") !== -1
```
nabinhait: `can_read` is false because of "User cannot search", which was enabled for gl entry for dynamic link validation. We can remove the permission check, the all the balances will be open to all. We need to check permissions in some other way.

nabinhait (changes requested):

erpnext/accounts/doctype/account/account_tree.js
rohitwaghchaure: Can I add Account permission, View Ledger button only display if user has permission to read account data

PR #10265, opened 2017-08-03
[fix] redirect to payment url if paying via web portal

saurabh6790:
*. While paying from web portal system not redirecting to payment url
- Before
payment_error
- After
fix_payment_error

---

*. If website warehouse is not defined in Item master then on item selection 
 - Before
screen shot 2017-08-03 at 2 36 35 pm
 - After
screen shot 2017-08-03 at 2 36 48 pm


mbauskar:
@saurabh6790 ,

Travis checks failed can you please check and fix the broken test cases? closing the PR for now

nabinhait (approved):

PR #10267, opened 2017-08-03
Added option to amend submitted orders in the POS

rohitwaghchaure:
**Issue**

1. Added minimum discount as 15% in one item 

1. Created an order against the item using POS, but order has not synced due to an error "Maxiumm discount for Item Mobile Headphone is 15.0%"

**Fixed**

1. Added provision in the POS profile to allow to amend an orders

1. User can able to amend the orders if there is an errors in the submitted order

cancel_amend_invoice


chdecultot:
Hi @rohitwaghchaure,

I humbly think your PR needs to be modified to be in line with the behavior of a standard POS system.
If you give the ability to modify a submitted invoice, whether it's in the POS or not, it needs to be recorded as such.
Here, it could only be the cancellation of the first invoice and the creation of an amended document which will be the final invoice (it should be similar to the behavior of the Sales Invoice module).
It would not be very rigorous to avoid tracing any change as soon as a document is submitted.

I really hope you will consider this remark as it could be a major issue for a lot of users.

Thanks a lot for your understanding 😊

rohitwaghchaure:
Hi @chdecultot 
All right so you meant we need to provide the cancel and amend both option for the POS and has to captured the modification did by the user after the order has cancelled.

chdecultot:
Hi @rohitwaghchaure,

Yes my point is only that any submitted document must be cancelled and a new amended document should be created if the user needs to make modifications. 

In this case you can leave the option to amend the invoice in the POS profile, it is only the behavior that needs to be changed.

Moreover, the proposed behavior will be easier to implement in the case the system doesn't allow offline POS. Once the invoice will be submitted, if it needs to be amended it will first need to be cancelled (like any normal invoice)

I hope I am clear enough 😊

grynn-in:
If transaction is hanging in POS : it will stay there forever, if validation on server side does not pass. Because the transaction has happened long before it got submitted. So 2 choices, loose the entire transaction OR change it with an option to write off / error entry **to correct later.** 

For example: If discount was not allowed and given, then the only way it will pass server validationis if we either change the master data or change the discount. Perhaps give a lower discount and write off difference. (in essence, you loose revenue either way) & perhaps pay higher VAT on that transaction. OR create additional credit note and also deduct the VAT accordingly. This is part of controlling. 

As a bug : Some validation during offline transaction should be disabled..unfortunately..this can lead to POS agent giving more discounts than allowed. 

OR disable offline POS.

grynn-in:
btw: this doesnt solve the problem of orders being in a queue and if one order hangs all following orders hang. It is like fast moving cars on a highway, if one crashes, everyone behind him crash into him. 

If you wish to avoid, the cars behind should avoid him (go around). In physical world it is not possible, but in software it should be. You cant prevent crashes, but you can in software world prevent pile ups. 

So what you did here doesn't solve the problem, but solves problem in one instance if a crash happens in one way. Not sure if I used the right metaphor :)

Hope that helps.

olamide2:
Hope am not too late to comment, 

In a proper retail environment, the print process is very critical, when a transaction is printed in retail, it signals the successful completion of the transaction. Where a transaction still has issues (wrong discount/credit limit exceeded etc) but the cashier is able to print and then issue a receipt to the customer is a BIG problem in retail.

We need to incorporate this in the process flow. Print should not be possible unless the transaction is validated and accepted. I am a retail user and not a programmer so I really do not know how difficult this will be in ERPNext PoS module, but it is a standard workflow in all Retail Systems I have used.

Olamide

Olawale1:
Hi @olamide2 

Trust you're doing great. I believe the emphasis is on validation before 'submission' (in ERPNext parlance) and not actually printing (a la regular offline POS systems) because once a document is successfully submitted, it can (and should) be available for printing

The ERPNext POS does validate before submission (local) so I think what would help is to provide a list of criteria (for example, max discount) which ought to be included in the validations but are currently missing

Cheers

Olawale1:
Also, as rightly pointed out by @grynn-in, this doesn't solve the problem of transactions being held up in a queue but at least it gives the end-user some ability to un-block the queue by resolving the issue

Cheers!

nabinhait:
Closing this, will add it with new redesign

PR #10269, opened 2017-08-03
BOM - Currency symbol issue (#10099)

### LINKED ISSUE #10099: BOM - Currency symbol issue
bom page 1 as usd heading but cny actual

tundebabzy:
Fixed #10099
Also depends on frappe/frappe#

The OP raised currency issues in different parts of ERPNext:
### 1. BOM:
I could not replicate it.

#### Issue screen shot:
screenshot from 2017-08-03 12-07-12

#### After fix
screenshot from 2017-08-03 11-09-53

#### Issue screenshot
peek 2017-08-03 12-03

#### After fix
peek 2017-08-03 11-18


mbauskar:
@tundebabzy,

Test case is missing. Can you add the simple test case to check if the currency is fetched properly or not?

PR #10131, opened 2017-07-27
Allow Sidebar Menu search using text from Item Name of a record

creamdory:
From PR in frappe repo https://github.com/frappe/frappe/pull/3798#pullrequestreview-52562113 as requested changes.

For web pages in Menu Sidebars, the common column for doctypes with "items" child table include item name's summary as below:

item names

It would be best to allow having this column to look at when using the page's **Search Bar.**

item name filter

This is provided that "items" field is included in the doctype's search fields:

search field

PR #10281, opened 2017-08-04
[Enhancement] Allow Sidebar Menu record search using text from Item Name of a record

creamdory:
From PR https://github.com/frappe/erpnext/pull/10131 with requested changes.

For web pages in Menu Sidebars, the common column for doctypes with "items" child table include item name's summary as below:
item names

It would be best to allow having this column to look at when using the page's Search Bar.
item name filter

This will only take effect for Menus using "website_list_for_contact.py" for generated list.

Test case generates 2 separate submitted quotation with different items to test search.

### DIFF
--- a/erpnext/controllers/website_list_for_contact.py
+++ b/erpnext/controllers/website_list_for_contact.py
@@ -19,7 +19,6 @@ def get_list_context(context=None):
 	}
 
 def get_transaction_list(doctype, txt=None, filters=None, limit_start=0, limit_page_length=20, order_by="modified"):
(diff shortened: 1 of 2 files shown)


nabinhait (changes requested):
The PR looks ok. Only thing is you have used "spaces" instead of "tabs" in writing `get_list_for_transactions`. Please fix and reopen the PR.

creamdory:
Hi @nabinhait , please see edit here. https://github.com/frappe/erpnext/pull/10324 Can't re-open due to forced push while close. Thanks!

PR #10286, opened 2017-08-04
[Feature] Project/Task Details Template

KanchanChauhan:
Details/Notes template for Project and Task. Sometimes there is a specific standard format for Scope of project/task, Terms and Conditions, Steps to follow to complete task. etc. which can be defined here.

Template Example:
screen shot 2017-08-04 at 12 39 16 pm

Usage:
screen shot 2017-08-04 at 12 39 05 pm

Fixed https://github.com/frappe/frappe/issues/3712


mbauskar:
@KanchanChauhan,

Thanks for the PR,

Can you please check and fix the codecy issues? closing the PR for now but you can reopen it.

erpnext/projects/doctype/project/project.js
```diff
@@ -97,6 +97,20 @@ frappe.ui.form.on("Project", {
...
+	notes_template: function(frm) {
+		if (frm.doc.notes_template){
+			return frappe.call({
+				method: 'erpnext.projects.doctype.project_and_task_details_template.project_and_task_details_template.get_details',
+				args: {
+					template_name: frm.doc.notes_template,
+					doc: frm.doc
+				},
+				callback: function(r) {
+					cur_frm.set_value("notes", r.message);
```
rmehta: cur_frm ??

Also you can simple set the value in options of `notes`, like `notes_template.notes`

rmehta:
I don't like the doctype name. A better design would be to just use "Note" from Frappe for this!

KanchanChauhan:
@rmehta Can you suggest some better name. Note cannot be used here unless we add "disabled" field to it. Also "Note" on template selection does not make sense.

rmehta:
There is also "Standard Reply" which does the same thing.

Maybe rename "Standard Reply" to "Text Template" and use the same wherever it is needed?

Otherwise we will just keep on adding DocTypes like this.

PR #10309, opened 2017-08-06
Supplier Delivery Note field added to Purchase Receipt, good to have for references

KanchanChauhan:
screen shot 2017-08-06 at 11 06 05 pm


joezsweet:
Any chance to allow option to check if DN number is unique, like in purchase invoice?

PR #10320, opened 2017-08-08
[Fix] Getting an error cost center is not defined in the purchase receipt, even if it's defined in the company

rohitwaghchaure:
**Issue**
screen shot 2017-08-08 at 1 18 06 pm

Issue still exist even if the cost center added into the company

Added cost center

screen shot 2017-08-08 at 8 30 15 pm

### DIFF
--- a/erpnext/__init__.py
+++ b/erpnext/__init__.py
@@ -28,6 +28,16 @@ def get_default_currency():
 	if company:
 		return frappe.db.get_value('Company', company, 'default_currency')
 
+def get_default_cost_center(company):
+	'''Returns the default cost center of the company'''
+	if not company:
+		return None
+
+	if not frappe.flags.company_cost_center:
+		frappe.flags.company_cost_center = {}
+	if not company in frappe.flags.company_cost_center:
+		frappe.flags.company_cost_center[company] = frappe.db.get_value('Company', company, 'cost_center')
+	return frappe.flags.company_cost_center[company]
 
 def get_company_currency(company):
 	'''Returns the default company currency'''
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -83,10 +83,12 @@ class BuyingController(StockController):
 
 	def set_landed_cost_voucher_amount(self):
 		for d in self.get("items"):
-			lc_voucher_amount = frappe.db.sql("""select sum(applicable_charges)
+			lc_voucher_data = frappe.db.sql("""select sum(applicable_charges), cost_center
 				from `tabLanded Cost Item`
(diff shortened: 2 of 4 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -192,7 +192,7 @@ def get_gl_entries(self, warehouse_account=None):
 					gl_entries.append(self.get_gl_dict({
 						"account": warehouse_account[d.warehouse]["account"],
 						"against": stock_rbnb,
-						"cost_center": d.cost_center,
+						"cost_center": d.cost_center or erpnext.get_default_cost_center(self.company),
```
nabinhait: Why setting default cost center here? Instead, in validate we should set default cost center if missing.

nabinhait (commented):
The error message is coming while reposting GLE from landed cost voucher. Why it did not come on the first place on submission of PR?

nabinhait:
closing, please reopen after fixing

erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
rohitwaghchaure: Fixed!

rohitwaghchaure:
In Purchase receipt no entry has been made against the expense head.
During submission of landed cost voucher the amount has credited against expense account expenses_included_in_valuation and use has not defined the cost center in PR and therefore system not allowing to submit the landed cost voucher

erpnext/accounts/doctype/gl_entry/gl_entry.py
```diff
@@ -63,6 +62,9 @@ def check_mandatory(self):
 
 	def pl_must_have_cost_center(self):
 		if frappe.db.get_value("Account", self.account, "report_type") == "Profit and Loss":
+			if not self.cost_center:
+				self.cost_center = erpnext.get_default_cost_center(self.company)
```
nabinhait: I don't think it is a good idea to set a default cost center if not specified in the form. We should force users to set cost center in the transactions itself. We are already setting default cost center there while create a new transaction.

In current problem's context, we should edit the error message to direct user to take necessary action. In this case, PR should be amended and cost center should be entered there.

In the below message, we can change like:

```frappe.throw(_("{0} {1}: Cost Center is required for 'Profit and Loss' account {2}. Please edit/amend the {3} to set the Cost Center.")
  					.format(self.voucher_type, self.voucher_no, self.account, self.voucher_type))
```

nabinhait (changes requested):

nabinhait (approved):

PR #10325, opened 2017-08-08
[new] Modified .travis.yml for setup wizard befre UI TEST

vishdha:


rmehta:
You also need to activate all roles and domains. See how it is done while running tests.

Issue #7105, opened 2016-11-30
Email Alert


graphskill:
We have set up an email alert to email when the projected quantity of a product is less than the safety stock as shown here

![image](https://cloud.githubusercontent.com/assets/925619/20747555/663f2e6e-b6e2-11e6-87e8-2ea0fd25d3db.png)

Two problems

1) Ideally we should be able to also select the Re-Order level rather than stock level
2) If the projected quantity is less than 0 it does not seem to work. We have a product with a projected quanitty of -178 and a safeyty stock of 200 but the email was not generated

graphskill:
We have tried on the forum, but no result. Must be a bug?

https://discuss.erpnext.com/t/email-alerts-error/18063

PR #10329, opened 2017-08-08
[hotfix] set total projected qty using db_set to trigger onchange event for email alerts

mbauskar:
fixes for https://github.com/frappe/erpnext/issues/7105

`total_projected_qty` field in the item is updated from `tabBin` using update query which does not trigger the `on_change` event for the email alert if the email alert is based on value change of `total_projected_qty`  field.


nabinhait:
@mbauskar Travis is failing due to the change, please check.

PR #10196, opened 2017-07-31
Added Quote Status to RFQ Suppliers with No Quote

bcornwellmott:
Rebasing https://github.com/frappe/erpnext/pull/10143

PR #10333, opened 2017-08-08
Added Quote Status to RFQ Suppliers with No Quote

bcornwellmott:
Tried to rebase https://github.com/frappe/erpnext/pull/10196 and it screwed things up.

@nabinhait, the no_quote checkbox is so that the select field for quote_status can be read only.
I though it was a cleaner and easier to understand user experience if you have a checkbox to indicate whether the supplier has informed you they will not quote the RFQ, then you have the read-only select field that gives you a status that you can review easily in a report. 

If you had just the select field, I think it would be confusing because sometimes it would be read-only and sometimes it wouldn't be (depending on if a quote has been received).


nabinhait:
@bcornwellmott Please fix the merge conflict

nabinhait:
merged using command line

PR #10349, opened 2017-08-09
mark holiday in the employee monthly attendance sheet

### LINKED ISSUE #10071: Monthly Attendance Report - Indicate Holidays (H)
Currently, columns for the holidays (like Saturday and Sunday) appears blank in the Monthly Attendance Report. If the date is included in the Holiday List, then we could indicate it as H (holiday). We can still allow marking attendance for the date added in the Holiday List, for the case of compensatory off.

In the following screenshot, 1 and 2 day of the month and subsequent weekends are added in the Holiday List already, but columns appear as blank in the report.

screen shot 2017-07-25 at 10 57 59 am

manassolanki:
Fixes https://github.com/frappe/erpnext/issues/10071

image

### DIFF
--- a/erpnext/hr/report/monthly_attendance_sheet/monthly_attendance_sheet.py
+++ b/erpnext/hr/report/monthly_attendance_sheet/monthly_attendance_sheet.py
@@ -15,6 +15,12 @@ def execute(filters=None):
 	att_map = get_attendance_list(conditions, filters)
 	emp_map = get_employee_details()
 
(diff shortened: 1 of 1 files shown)


tundebabzy:
Hi
I think Half Day should be changed to HD perhaps so H can be for holiday. The boldened H isn't clear and unique enough IMHO

manassolanki:
image

PR #10368, opened 2017-08-11
While fetching items from another doctype create blank row at item child table in Material Request #10339

### LINKED ISSUE #10339: While fetching items from another doctype create blank row at item child table in Material Request
stock

tundebabzy:
Fixed #10339 

peek 2017-08-11 01-17

### DIFF
--- a/erpnext/public/js/controllers/buying.js
+++ b/erpnext/public/js/controllers/buying.js
@@ -347,6 +347,12 @@ erpnext.buying.get_items_from_product_bundle = function(frm) {
 			freeze: true,
 			callback: function(r) {
 				if(!r.exc && r.message) {
+					if($.isArray(frm.doc.items) && frm.doc.items.length > 0) {
+						// remove first item row if empty
+						if(!frm.doc.items[0].item_code) {
+							frm.doc.items = frm.doc.items.splice(1);
+						}
+					}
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/buying.js
```diff
@@ -347,6 +347,12 @@ erpnext.buying.get_items_from_product_bundle = function(frm) {
 			freeze: true,
 			callback: function(r) {
 				if(!r.exc && r.message) {
+					if($.isArray(frm.doc.items) && frm.doc.items.length > 0) {
+						// remove first item row if empty
+						if(!frm.doc.items[0].item_code) {
```
mbauskar: Item Code is not mandatory for Purchase Invoice, Invoice can be created with item_name, This will cause to delete the first row if item was added with item_name and other details

mbauskar (commented):
Please see my comments

mbauskar:
@tundebabzy,

Can you also add simple UI test case for the PR? Closing the PR for now, but you can reopen the PR

erpnext/public/js/controllers/buying.js
tundebabzy: This is actually a material request. However, at second glance, I'm not too comfortable with the fix. I'm thinking of another fix. I'll reopen once I cook up something nicer.

Issue #9973, opened 2017-07-19
Better Management of Currency Exchange Rates


Olawale1:
Hi ,

For foreign currency transactions, ERPNext fetches the last exchange rate defined (prior to posting date) regardless of how old it is. let's say posting date is 17th July and the last exchange rate before then is 13th January, the system automatically uses this rate (13th January). This is however not best practice as exchange rates could have fluctuated significantly within the period

Ideally, the system should throw an error message requesting that an exchange rate be entered for the posting date (i.e. 17th July in the example above). If however there are multiple exchange rates existing on the system for 17th July, the latest one can be automatically selected. This principle should actually apply to all foreign currency transaction documents

Thanks

umairsy:
Hi,

If you wish to have latest exchange rates auto-fetched, then ensure that no Currency Exchange record exists between these currencies.

Olawale1:
Hi @umairsy

Trust you're doing great. Not very sure what you mean but my point is that it should be mandatory to enter an exchange rate for the exact date (Posting Date) of the transaction. Its This is the standard practice in ERP systems like SAP

Please let me know if further clarification is required 

Thanks

tundebabzy:
Perhaps we can implement this as such:
1. If the document has never been saved, we leave the exchange rate field blank and make it compulsory.
2. For (1), when the document is saved, the exchange rate is also saved to Currency Exchange as a new record.
3. If the document has been saved before, the latest exchange rate is retrieved into the exchange rate field.
4. In (3), if the exchange rate is changed, the exchange rate is also saved to Currency Exchange as a new record.

What's your reaction?
cc: @addymuliady

joolsr:
Could I possibly request that a factor can also added as an option to use a slightly different rate ? This is because we have to pay £30 GBP whenever we accept another currency so we normally use a slightly worse exchange rate to our customers to make up for this charge to us.

Just a thought as you may not always want the absolute defined exchange rate.

Julian

⁣Sent from BlueMail ​

tundebabzy:
The exchange rate will be user defined not automatically defined (for now). Does that address your concern @joolsr

joolsr:
Yes though getting auto rates that can have a factor added on would be better

Julian

⁣Sent from BlueMail ​

Olawale1:
Hi @tundebabzy 

[...]

Issue #10145, opened 2017-07-28
Payment Entry - Multicurrecy Payment - Set Exchange Gain/Loss should be in company currency


addymuliady:
ERPNext: v8.5.4 (master)
Frappe Framework: v8.5.7 (master)

1. Create New Payment Entry
2. Select a Foreign Account
3. Allocate to all outstanding invoices
4. There should be a Difference Amount
5. Click on Set Exchange Gain Loss
6. ERPNext will automaticall create a difference entry in account currency instead of company curency
7. Click on Save and Submit

Prompt:
![image](https://user-images.githubusercontent.com/21187126/28698834-8120179a-7378-11e7-8b43-4f74652c6792.png)

tundebabzy:
Seems like there is more to this
![peek 2017-07-28 11-25](https://user-images.githubusercontent.com/818803/28713936-1491f5a6-7389-11e7-8d2d-07b4c637c901.gif)

@addymuliady can you confirm this behaviour that changes to allocated amount are not converted when updating the allocated amount field

addymuliady:
@tundebabzy yes, I can confirm that too. ;(

tundebabzy:
Tasks:
1. The bug in this issue stems from the fact that we are mixing two currencies when calculating unallocated amount and difference amount. Exchange gain/loss should be separated from deductions
- [ ] Create another read only child table for Exchange Gain/Loss. If user needs to do any exotic splits, a journal entry will suffice.
- [ ] Make accounts selectable from Deductions table to be only those in transaction currency.

iamkhalidbashir:
@nabinhait 
This issue is not resolved
"ERPNext will automatically create a different entry in account currency instead of company currency"
This was the actual issue addressed by @addymuliady first, the latter issue is solved only

Only company currency deductions can be made but not account currency

PR #10369, opened 2017-08-11
Enhance Currency Exchange Management

tundebabzy:
There was a discussion in #9973 about managing exchange rates.

My conclusion was to create a Settings doctype for Currency Exchange where the behaviour for fetching exchange rates can be tweaked. Presently, ERPNext cannot help users enforce IAS 21 standards of choosing exchange rates which recommends that the exchange rate used should be the one in effect on the transaction date. According to IAS 21 standards, exchange rates become stale at the end of the day it takes effect. With this new setting, this can be enforced.

screenshot from 2017-08-11 01-50-31

screenshot from 2017-08-11 01-54-06

"Allow Stale Exchange Rates" will be the default and doesn't change ERPNext's present behaviour. If unchecked, the user can then specify how many days to consider as stale (1 by default).

With this change, we can now add something like this to the `get_exchange_rate` function:
```(python)
stale_date = None
if not Currency_Settings.allow_stale:
    stale_date = get_datetime_str(add_days(transaction_date, days=-Currency_Settings.stale_days))
filters = [
    ["from_currency", "=", from_currency],
    ["to_currency", "=", to_currency]
]
if stale_date:
    filters.append(["date", ">", stale_date])
else:
    filters.append(["date", "<=", get_datetime_str(transaction_date)])

...
if not exchange_rates:
    frappe.msgprint("Please save the current exchange rate in Currency Exchange")
...
```


mbauskar:
@tundebabzy,

Can you please check and fix the codecy issues?, Closing the PR for now, Please reopen the PR once fixed.

erpnext/setup/doctype/currency_exchange_settings/test_currency_exchange_settings.py
```diff
@@ -0,0 +1,30 @@
...
+# See license.txt
+from __future__ import unicode_literals
+
+import frappe
+import unittest
+
+
+class TestCurrencyExchangeSettings(unittest.TestCase):
+	def test_change_setting(self):
+		settings_doc = frappe.get_doc("Currency Exchange Settings", "Currency Exchange Settings")
```
rmehta: @tundebabzy such tests are not needed. You should only write tests to test code in your own controller.

Also this can also be part of "Accounts Settings"

rmehta (changes requested):

tundebabzy:
TASKS:
- [x] Currency Exchange Settings Doctype
- [x] validation
- [x] In `get_exchange_rate` check if settings allows for stale exchange rate
- [x] In forms, make conversion rate field read only if settings does not allow stale exchange rates
- [ ] Add documentation

Currency Settings fields:
1. Allow Stale Exchange Rate
2. Stale exchange rate days

tundebabzy:
Please review but don't merge yet. #10410 first has to be merged and #10145 needs to be fixed in order to proceed.

PR #10378, opened 2017-08-11
While fetching items from another doctype create blank row at item child table in Material Request (#10339)

### LINKED ISSUE #10339: While fetching items from another doctype create blank row at item child table in Material Request

tundebabzy:
Fixed #10339 
peek 2017-08-11 01-17

### DIFF
--- a/erpnext/public/js/controllers/buying.js
+++ b/erpnext/public/js/controllers/buying.js
@@ -346,7 +346,21 @@ erpnext.buying.get_items_from_product_bundle = function(frm) {
 			},
 			freeze: true,
 			callback: function(r) {
+				const first_row_is_empty = function(child_table){
+					if($.isArray(child_table) && child_table.length > 0) {
+						return !child_table[0].item_code;
+					}
+					return false;
+				};
+
+				const remove_empty_first_row = function(frm){
+				if (first_row_is_empty(frm.doc.items)){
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/buying.js
```diff
@@ -366,3 +367,16 @@ erpnext.buying.get_items_from_product_bundle = function(frm) {
 	});
 	dialog.show();
 }
+
```
rmehta: why global functions? add the functions in the same scope

rmehta (changes requested):

erpnext/public/js/controllers/buying.js
tundebabzy: Moved the function to the callback scope

erpnext/public/js/controllers/buying.js
```diff
@@ -346,7 +346,21 @@ erpnext.buying.get_items_from_product_bundle = function(frm) {
...
 			callback: function(r) {
+				const first_row_is_empty = function(child_table){
+					if($.isArray(child_table) && child_table.length > 0) {
+						return !child_table[0].item_code;
+					}
+					return false;
+				};
+
+				const remove_empty_first_row = function(frm){
+				if (first_row_is_empty(frm.doc.items)){
```
rmehta: inconsitent whitespace...

PR #10417, opened 2017-08-16
[fix] Update monthly sales value on Company

nabinhait:
- Update monthly sales value on cancellation of invoice
- update modified timestamp of company  to avoid overwriting data on saving of already opened company records
- Update figures on deleting company transactions.


alfredhy:
Does it fix already? Since I still have found the Total Monthly Sales contains value. 
default

Do I need to click the detele all transactions again?

nabinhait:
No, it is merged in hotfix branch, will be merged to master mostly on Monday.

PR #10448, opened 2017-08-17
improvement for self-postbacks

joelios:
In the product_page.js is a code wich shall prevent of self-postbacks, this code does not allow that variants of an article has an article number like a substring of the other variants article number.
For example, Variant # 1 = 12345 and variant # 2 = 12345-1 causes errors in the webshop.

This PR would be a more elegant solution.

### DIFF
--- a/erpnext/templates/includes/product_page.js
+++ b/erpnext/templates/includes/product_page.js
@@ -73,7 +73,7 @@ frappe.ready(function() {
 			}
 		}
 
-		if (window.location.search.indexOf(item_code)!==-1) {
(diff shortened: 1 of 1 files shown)


erpnext/templates/includes/product_page.js
```diff
@@ -73,7 +73,7 @@ frappe.ready(function() {
 			}
 		}
 
-		if (window.location.search.indexOf(item_code)!==-1) {
+		if (window.location.search == ("?variant=" + item_code)) {
```
netchampfaris: Why not handle both the cases?

PR #10456, opened 2017-08-18
BOM rates as per selected currency and UOM

### LINKED ISSUE #10429: Auto fetched rate of raw materials in BOM should be converted as per selected currency
As valuation rate/ last purchase rate / price list rate is maintained in company currency, while fetching rate, it should be converted using exchange rate.

nabinhait:
Fixes #10429


rohitwaghchaure:
Merge from terminal https://github.com/frappe/erpnext/commit/2967d7ed01712cbe4eee4699059f9c14818daa1e

PR #10482, opened 2017-08-21
Enhance Currency Exchange Management

tundebabzy:
This is a continuation of #10369

TASKS:
- [x] Currency Exchange Settings Doctype
- [x] validation
- [x] In `get_exchange_rate` check if settings allows for stale exchange rate
- [x] In forms, make conversion rate field read only if settings does not allow stale exchange rates
- [x] Add documentation

Currency Settings fields:
1. Allow Stale Exchange Rate
2. Stale exchange rate days


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -286,6 +286,99 @@
...
+   "length": 0, 
+   "no_copy": 0, 
+   "permlevel": 0, 
+   "precision": "", 
+   "print_hide": 0, 
+   "print_hide_if_no_value": 0, 
+   "read_only": 0, 
+   "remember_last_selected_value": 0, 
+   "report_hide": 0, 
+   "reqd": 1,
```
nabinhait: How a checkbox can be mandatory?

erpnext/accounts/doctype/accounts_settings/accounts_settings.py
```diff
@@ -5,10 +5,20 @@
...
 	def on_update(self):
-		pass
\ No newline at end of file
+		pass
+
+	def validate(self):
+		self.validate_stale_days()
+
+	def validate_stale_days(self):
+		if not self.allow_stale and cint(self.stale_days) <= 0:
```
nabinhait: `if allow_stale`?

erpnext/accounts/doctype/accounts_settings/test_accounts_settings.py
```diff
@@ -0,0 +1,22 @@
...
+	def tearDown(self):
+		# Just in case `save` method succeeds, we need to take things back to default so that other tests
+		# don't break
+		cur_settings = frappe.get_doc('Accounts Settings', 'Accounts Settings')
+		cur_settings.allow_stale = 1
+		cur_settings.save()
+
+	def test_stale_days(self):
+		cur_settings = frappe.get_doc('Accounts Settings', 'Accounts Settings')
+		cur_settings.allow_stale = 0
```
nabinhait: `cur_settings.allow_stale = 1`?

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -286,6 +286,99 @@
...
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "default": "1", 
+   "depends_on": "eval:doc.allow_stale==0",
```
nabinhait: I did not understand why `stale_days` should be visible if `allow_stale` is unchecked. And I checked all the logic is also based on that. In my opinion, stale_days is applicable only if `allow_stale` is checked.

nabinhait (changes requested):

erpnext/public/js/controllers/transaction.js
```diff
@@ -531,6 +532,12 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			}
 
 		}
+		// Make read only if Accounts Settings doesn't allow stale rates
+		frappe.model.get_value("Accounts Settings", null, "allow_stale",
+			function(d){
+				me.set_df_property("conversion_rate", "read_only", cint(d.allow_stale) ? 0 : 1);
```
nabinhait: Why read only? User should be able to change the conversion rate manually. Maybe `fixer.io` is not available and user is offline, using local system.

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
nabinhait: OK, but it is little bit confusing and contradictory. If we set `allow_stale==0` that means we are not allowing stale, so system should take only current exchange rate. Why it should still allow some stale records based on some days? Is it not confusing?

erpnext/public/js/controllers/transaction.js
nabinhait: ok

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
nabinhait: I think now I understood this, I did not know the meaning of stale properly. It means "No longer fresh" and it always associated with a time period, after how many days it becomes stale.

PR #10519, opened 2017-08-23
Python 3 compatibility syntax error fixes

adityahase:
erpnext port

while running `bench --site site1.local install-app erpnext` few syntax errors occur this PR fixes that

### DIFF
--- a/erpnext/accounts/doctype/shipping_rule/shipping_rule.py
+++ b/erpnext/accounts/doctype/shipping_rule/shipping_rule.py
@@ -54,13 +54,15 @@ class ShippingRule(Document):
 			d.idx = i + 1
 
 	def validate_overlapping_shipping_rule_conditions(self):
-		def overlap_exists_between((x1, x2), (y1, y2)):
+		def overlap_exists_between(num_range1, num_range2):
 			"""
-				(x1, x2) and (y1, y2) are two ranges
-				if condition x = 100 to 300
-				then condition y can only be like 50 to 99 or 301 to 400
+				num_range1 and num_range2 are two ranges
+				ranges are represented as a tuple e.g. range 100 to 300 is represented as (100, 300)
+				if condition num_range1 = 100 to 300
+				then condition num_range2 can only be like 50 to 99 or 301 to 400
 				hence, non-overlapping condition = (x1 <= x2 < y1 <= y2) or (y1 <= y2 < x1 <= x2)
 			"""
+			(x1, x2), (y1, y2) = num_range1, num_range2
 			separate = (x1 <= x2 <= y1 <= y2) or (y1 <= y2 <= x1 <= x2)
 			return (not separate)
 
--- a/erpnext/patches/v4_0/reset_permissions_for_masters.py
+++ b/erpnext/patches/v4_0/reset_permissions_for_masters.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 # License: GNU General Public License v3. See license.txt
 
-from __future__ import unicode_literals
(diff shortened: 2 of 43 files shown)


erpnext/patches/v4_0/reset_permissions_for_masters.py
```diff
@@ -16,5 +16,5 @@ def execute():
 		try:
 			reset_perms(doctype)
 		except:
-			print "Error resetting perms for", doctype
+			print("Error resetting perms for", doctype)
 			raise
```
tundebabzy: You left this out. `raise` in this form is not python 3 compatible. The python 3 compatible form is also not compatible with python 2. However, six is a dependency so use six.reraise

tundebabzy (changes requested):

erpnext/accounts/doctype/shipping_rule/shipping_rule.py
```diff
@@ -54,13 +54,14 @@ def sort_shipping_rule_conditions(self):
 			d.idx = i + 1
 
 	def validate_overlapping_shipping_rule_conditions(self):
-		def overlap_exists_between((x1, x2), (y1, y2)):
+		def overlap_exists_between(x, y):
 			"""
 				(x1, x2) and (y1, y2) are two ranges
```
adityahase: Done

adityahase: Done.
How can I make it better than it currently is?

erpnext/patches/v4_0/reset_permissions_for_masters.py
adityahase: Um. I read accepted answer from [this Stack Overflow question](https://stackoverflow.com/q/2052390). I'm confused. 

Anyways. That doesn't really prevent me from running `bench --site site1.local install-app erpnext` as of now. I'll tackle that when it becomes a problem. Cool?

erpnext/accounts/doctype/shipping_rule/shipping_rule.py
```diff
@@ -54,13 +54,14 @@ def sort_shipping_rule_conditions(self):
...
 				if condition x = 100 to 300
 				then condition y can only be like 50 to 99 or 301 to 400
 				hence, non-overlapping condition = (x1 <= x2 < y1 <= y2) or (y1 <= y2 < x1 <= x2)
 			"""
+			(x1, x2), (y1, y2) = x, y
 			separate = (x1 <= x2 <= y1 <= y2) or (y1 <= y2 <= x1 <= x2)
 			return (not separate)
```
adityahase: I guess those are necessary there. I can rewrite that as

```
x1, x2 = x
y1, y2 = y
```

But not as

```
(x1, x2), (y1, y2) = x, y
```

erpnext/patches/v4_0/reset_permissions_for_masters.py
tundebabzy: Oops. You are right. I was wrong. There's nothing wrong with that raise statement

erpnext/accounts/doctype/shipping_rule/shipping_rule.py
tundebabzy: I meant in the return statement --> `return not separate` instead of `return(not separate)`. It's just a code clean up similar to removing the extra spaces

tundebabzy (approved):

erpnext/accounts/doctype/shipping_rule/shipping_rule.py
adityahase: Oh. Will create a separate PR soon for these kind of clean ups.

PR #10520, opened 2017-08-23
naming setting for the instructor

### LINKED ISSUE #10242: Remove Naming Series from Instructor

manassolanki:
Fixes https://github.com/frappe/erpnext/issues/10242

image


mbauskar:
@manassolanki,

Can you add the test cases and update the documents as well?

PR #10522, opened 2017-08-23
[fix] Decode Chart of Accounts filenames in UTF8

emcniece:
This PR fixes a crash during setup due to unicode filesystem conflicts.

During setup we encountered an error around step 5:

```
  File "/home/frappe/frappe-bench/installed/apps/frappe/frappe/app.py", line 57, in application
  File "/home/frappe/frappe-bench/installed/apps/frappe/frappe/handler.py", line 22, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/frappe-bench/installed/apps/frappe/frappe/handler.py", line 53, in execute_cmd
  File "/home/frappe/frappe-bench/installed/apps/frappe/frappe/__init__.py", line 922, in call
    return fn(*args, **newargs)
  File "/home/frappe/frappe-bench/installed/apps/erpnext/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py", line 108, in get_charts_for_country
    if (fname.startswith(country_code) or fname.startswith(country)) and fname.endswith(".json"):
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 19: ordinal not in range(128)
```

It appears that filenames are being read with unicode problems. I did see a warning about filesystem encoding:

```
/home/frappe/frappe-bench/installed/env/local/lib/python2.7/site-packages/werkzeug/filesystem.py:63: BrokenFilesystemWarning: Detected a misconfigured UNIX filesystem: Will use UTF-8 as filesystem encoding instead of 'ANSI_X3.4-1968'
8/22/2017 9:41:35 PM04:41:35 web.1            |   BrokenFilesystemWarning)
```

... but as [mentioned over here](https://stackoverflow.com/questions/34515331/werkzeug-raises-brokenfilesystemwarning), this is only a warning and can be mitigated by setting an environment variable (which we have).

To prevent this crash, several filenames iterated over in `os.listdir()` can be decoded. The following changes resolve this bug.

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
@@ -78,7 +78,7 @@ def get_chart(chart_template, existing_company=None):
 		for folder in folders:
 			path = os.path.join(os.path.dirname(__file__), folder)
 			for fname in os.listdir(path):
-				if fname.endswith(".json"):
+				if fname.decode("utf-8").endswith(".json"):
 					with open(os.path.join(path, fname), "r") as f:
 						chart = f.read()
 						if chart and json.loads(chart).get("name") == chart_template:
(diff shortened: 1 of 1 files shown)


mbauskar:
fixed via https://github.com/frappe/erpnext/pull/10620 & https://github.com/frappe/erpnext/pull/10601

PR #10538, opened 2017-08-24
[fix] warehouse query fixes https://github.com/frappe/erpnext/issues/10537

rmehta:


erpnext/controllers/queries.py
```diff
@@ -367,31 +367,30 @@ def warehouse_query(doctype, txt, searchfield, start, page_len, filters):
...
-			%(start)s, %(page_len)s
+			{start}, {page_len}
 		""".format(
 			sub_query=sub_query,
 			key=frappe.db.escape(searchfield),
 			fcond=get_filters_cond(doctype, filter_dict.get("Warehouse"), conditions),
-			mcond=get_match_cond(doctype)
-		),
-		{
-			"txt": "%%%s%%" % frappe.db.escape(txt),
```
rmehta: wrong way to escape!

PR #10547, opened 2017-08-24
Pos scanner fix

revant:
POS Fix

- Search item based on key press instead of 1sec wait 
- Fast Scanning needs this to avoid 1sec delay

@bkm888 please add details of your test.


bkm888:
Prior to these changes a user with a barcode scanner would be forced to wait 1 second or more between items scans. Anything less that this would pile up scan data on the search bar and force a manual clearing. MN Technique was able to find and correct the user interface issue. When entering data into the search bar, a key press (Enter Key in my case) will retrieve the search results. Scanner is programmed to append Enter key to every scan. Works great.

rohitwaghchaure:
Hi @revant 

Kindly send the PR on hotfix branch

bkm888:
Hello,
 
I do not know what "PR" means. I am only an observer on GitHub and have no idea how to really use it. I am not a developer. I am an implementer.
 
BKM
 
-----Original Message-----

agritheory:
@bkm888 - "PR" = Pull Request. Rohit asked that Revant submit it on a different branch than this one. I am very pleased that you sponsored this improvement, by the way.

revant:
https://github.com/frappe/erpnext/pull/10559

PR #10559, opened 2017-08-26
[Fix] POS Scanner

revant:
Search item based on key press instead of 1sec wait
Fast Scanning needs this to avoid 1sec delay

### DIFF
--- a/erpnext/accounts/page/pos/pos.js
+++ b/erpnext/accounts/page/pos/pos.js
@@ -426,11 +426,16 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		});
 
 		this.serach_item.make_input();
+		
 		this.serach_item.$input.on("keypress", function (event) {
-			if((me.serach_item.$input.val() != "") || (event.which == 13)){
-				me.items = me.get_items();
-				me.make_item_list();
-			}
+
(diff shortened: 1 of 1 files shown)


revant:
travis-ci

```
Checking if passed "erpnext/hr/doctype/employee_attendance_tool/test_employee_attendance_tool.js"
FAIL
```

erpnext/accounts/page/pos/pos.js
```diff
@@ -426,11 +426,11 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		});
 
 		this.serach_item.make_input();
-		this.serach_item.$input.on("keyup", function () {
-			setTimeout(function () {
+		this.serach_item.$input.on("keypress", function (event) {
+			if((me.serach_item.$input.val() != "") && (event.which == 13)){
```
rohitwaghchaure: why event.which == 13?
What if user not using barcode scanner to search an item?  
After entered the search value in the search box, user needs to click on enter button every times.
Can we put or(||) condition instead of and(&&)?

revant:
travis-ci
```
Checking if passed "erpnext/hr/doctype/leave_application/test_leave_application.js"
FAIL
```

PR #10586, opened 2017-08-29
Fixed rounding issue in Process Payroll, fixes #9300

### LINKED ISSUE #9300: Rounding error while making bank entry from Process Payroll
**Error message:**
"Total Debit must be equal to Total Credit. The difference is -0.0100000000093"

**Discussion:**
https://discuss.erpnext.com/t/error-when-making-bank-entry-in-payroll-processing/20943

nabinhait:


hfaridgit:
The journal entry is wrong because there is a mistake in line number 287:
for acc, amt in deductions.items():       <=== should be ===>        for acc, amount in deductions.items():

PR #10595, opened 2017-08-29
[Enahance] Update variants fields defined in the Item Variant Settings, if template updated

### LINKED ISSUE #9482: Variants not Updated when Item Master is Updated
If Item Variants are not updated when you update Item Template there is no need to use Variants anymore.
Are you willing to reintroduce this function? Why did you remove it?
I chose ERPNext instead of Odoo also because I preferred the way variants were managed.
Last I think you should let also users of Hosted Plan to be able to decide which fields can be set as NoCopy.

rohitwaghchaure:
- [x] Default values on Patch + Install
- [x] Rename "Variant Fields" to "Copy Fields to Variant"
---
**Issue**
Variants data not updated if user update the data in the temaplte

**Fixed**
Select the fields in the  Item Variant Settings which want to update in the variants master from the template
screen shot 2017-09-11 at 12 44 00 pm

Fixed https://github.com/frappe/erpnext/issues/9482


GiorgioKN:
Hi,
thank you very good news.
Could you please tell me when this will be available for the cloud version?

rohitwaghchaure:
Hi @GiorgioKN 

Currently it's under review and we'll release it once PR is merged!

erpnext/config/stock.py
```diff
@@ -73,6 +73,11 @@ def get_data():
 				},
 				{
 					"type": "doctype",
+					"name": "Item Variant Settings",
```
rmehta: this should be in settings!

rmehta (changes requested):

erpnext/config/stock.py
rmehta: also maybe add a button in Item (if it is a variant) for Variant Settings

rmehta:
patch + default settings on install.

rmehta (approved):

nabinhait:
merged from console

PR #10605, opened 2017-08-30
Show thumbnail image in daily work summary

### LINKED ISSUE #10201: User Image formatting for Email Format for Daily Work Summary
screen shot 2017-08-01 at 10 56 03 am

netchampfaris:
Fixes https://github.com/frappe/erpnext/issues/10201

image

Depends on: https://github.com/frappe/frappe/pull/4057

### DIFF
--- a/erpnext/hr/doctype/daily_work_summary/daily_work_summary.py
+++ b/erpnext/hr/doctype/daily_work_summary/daily_work_summary.py
@@ -52,6 +52,18 @@ class DailyWorkSummary(Document):
 			d.sender_name = emp[0].employee_name if emp else d.sender
 			d.image = emp[0].image if emp and emp[0].image else None
 
+			# make thumbnail image
+			try:
+				if d.image:
+					file_name = frappe.get_list('File', {'file_url': d.image})
+
+					if file_name:
+						file_name = file_name[0].name
+						file_doc = frappe.get_doc('File', file_name)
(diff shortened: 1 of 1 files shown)


mbauskar (changes requested):
@netchampfaris,

Please check and fix the codecy issues

erpnext/hr/doctype/daily_work_summary/daily_work_summary.py
```diff
@@ -52,6 +52,18 @@ def get_message_details(self):
...
 
+			# make thumbnail image
+			try:
+				if d.image:
+					file_name = frappe.get_list('File', {'file_url': d.image})
+
+					if file_name:
+						file_name = file_name[0].name
+						file_doc = frappe.get_doc('File', file_name)
+						d.image = file_doc.make_thumbnail(set_as_thumbnail=False, width=100, height=100)
```
rmehta: aspect ratio?

Also `make_thumbnail` will always make a new thumbnail. Make a new method `get_thumbnail_url` that will make a thumbnail if missing...

rmehta (changes requested):

erpnext/hr/doctype/daily_work_summary/daily_work_summary.py
netchampfaris: The thumbnail created here is only for the email, `get_thumbnail_url` would return the default thumbnail of 300x300

PR #10616, opened 2017-08-31
[Enahance] Provision to select customer, item code while making an invoice from timesheet

### LINKED ISSUE #9555: Improve billing from Timesheet
1. Currently when we create Sales Invoice from timesheet, the timesheet billable hours, total billable amount is available in the timesheet section of the sales invoice.

2. The information above, is sufficient to invoice a customer but we still have to manually create an Item, add quantity & hours.

Suggestion:

1. When creating Invoice from timesheet, or pulling multiple timesheet for Sales Invoice, copy the  
    total billable hours to quantity field & billable amount from timesheet to Amount into the Item grid. 
    Project it available, can also be copied.

2. User can then select just the Item on the Invoice directly without having to worry to enter the total hours, amount, project or modal window can be given for selection of item before invoice creation.

This will simplify billing from timesheet.

rohitwaghchaure:
make_invoice_from_timesheet

Fixed https://github.com/frappe/erpnext/issues/9555


erpnext/projects/doctype/timesheet/timesheet.js
```diff
@@ -57,10 +57,37 @@ frappe.ui.form.on("Timesheet", {
...
+		let dialog = new frappe.ui.Dialog({
+			title: __("For Item"),
+			fields: [
+				{"fieldtype": "Link", "label": __("Item Code"), "fieldname": "item_code", "options":"Item"},
+				{"fieldtype": "Link", "label": __("Customer"), "fieldname": "customer", "options":"Customer"},
+				{"fieldtype": "Button", "label": __("Make Sales Invoice"), "fieldname": "make_sales_invoice", "cssClass": "btn-primary"},
+			]
+		});
+
+		dialog.fields_dict.make_sales_invoice.$input.click(function() {
```
mbauskar: @rohitwaghchaure ,

Instead of adding Custom button can we use Primary Button to create Sales Invoice
e.g. `dialog.set_primary_action(__("Make Sales Invocie"), function() {` ?

mbauskar (changes requested):

erpnext/projects/doctype/timesheet/timesheet.js
rohitwaghchaure: Fixed!

PR #10661, opened 2017-09-04
Add excluded fields to item variants

Alchez:
Changes:

1. **Show in Website (Variant)** - To avoid unchecking the field in the variant since the fieldname doesn't exist for the template item.

2. **Standard rate** - To avoid the standard rates for the variants to change if they are different from the template item.


rmehta:
@Alchez thanks for the contribution! With #10595 you will be able to select which fields to update. 

Closing this, since now you can configure which fields to map.

PR #10663, opened 2017-09-04
[fix] wizard UI test order

pratu16x7:


erpnext/setup/setup_wizard/utils.py
```diff
@@ -1,15 +0,0 @@
-import json, os
```
rmehta: don't delete this! I use it all the time for testing 😜

rmehta (changes requested):

erpnext/setup/setup_wizard/utils.py
pratu16x7: okay :)

PR #10678, opened 2017-09-05
Sorting in Production Order BOM Items

creamdory:
In most of the Manufacturing companies, BOM Items in BOM Setup are arranged because of important reasons. When Production Order is done for a specific BOM, it currently don't follow the same arrangement in the setup. It would be best to have it set exactly like it for efficiency:

Suggested Feature:

**BOM Setup**

bom 1

**Production Order**

prod 1


kickapoo:
@nabinhait Can you please push a review for this ?

creamdory:
Hi All! My Travis Ci Build failing is due to a  test case which I believe not due to the PR. If somebody can check and make feedback for any changes would be really great. Thanks!

nabinhait:
Merged manually after minor cleanup.

https://github.com/frappe/erpnext/commit/77b225e02184546cd8f517bf141fc373755f12ca

PR #10689, opened 2017-09-06
[fix] unset default customer group

rmehta:


mbauskar:
@rmehta,

Test cases for shopping cart are failing can you please check and fix the same?

PR #10691, opened 2017-09-06
Customer Satisfaction Survey

KanchanChauhan:
- Created a simple Customer Satisfaction Survey with some basic questions
- Based on that doctype created a web form
- Very basic and simple nothing fancy, others can customize to add more questions
- Whenever you need to turn the survey on/off just add it to the Portal menu and enable/disable
- Since surveys are anonymous so no user information is being captured

Any suggestions for improvement are welcome. 

screencapture-dev-newmatik-8000-customer-satisfaction-survey-1504679910885


rmehta:
This needs to be more generic

- [ ] **Survey Template**
- [ ] **Survey Result**
- [ ] Ability to send Survey to an Email Group (via Newsletter ?)
- [ ] Summary of results (report / graph)

PR #10698, opened 2017-09-06
Fix for company address in case of multiple company

PawanMeh:
Note this PR has dependency on Frappe PR#4076.

Testing video below

multiple company issue test


erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -516,6 +516,24 @@ frappe.ui.form.on('Sales Invoice', {
...
+					if (r.message){
+						frm.set_value("company_address",r.message)
+					}
+					else {
+						frm.set_value("company_address","")
+					}
+				}
+			})
+		}
+	},
```
rohitwaghchaure: This will override existing company trigger defined in the [transaction.js](https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/transaction.js#L375-L403) file. Can you move your code from sales_invoice.js to transaction.js

PawanMeh: The default company trigger does not seem to be firing on change of the company field. You can check from your end.

rmehta:
@PawanMeh you will have to start a new one. Seems the branch is not merge-able

PR #10706, opened 2017-09-07
Added ability to disable auto creation of Sales Invoice when customer pays against a payment request

neilLasrado:
In many cases, you won't want to create an invoice directly against a payment; but you want to receive the payment as an advance against the Sales Order and later make the Invoice once the product is delivered.

The following workflow is achieved by this change:

Sales Order -> Payment Request (From Sales Order) -> Payment Entry (From Sales  Order; against Payment Request) -> Delivery Note (from Sales Order) -> Sales Invoice (from Delivery Note)

the same also works for orders placed through the shopping cart.

A check has been added under 'Accounts Settings' to toggle this.

image


rmehta:
If payment request is against Sales Order, it should not automatically create an invoice. No setting required. Can you fix it? 

cc @saurabh6790

nabinhait:
@neilLasrado Reverted after merging, please fix as per Rushabh's comment

neilLasrado:
https://github.com/frappe/erpnext/pull/10860

neilLasrado:
@rmehta also wanted to highlight that this affects Cart Orders. Right now Sales Invoices are created against orders places from shopping cart directly. It must not be the case since in many companies the Invoice is generated after the actual Item is delivered.

PR #10744, opened 2017-09-11
Enhancement to handle Employee Cash Advances

PawanMeh:
Testing Video

expenses cash advance


Olawale1:
Hi @PawanMeh 

Great job! The flow seems to still require some elements though:

1. Usually an employee requests for advance payment based on estimates because the actual expenses haven't occurred yet. After approval of those estimates (i.e. Expense Claim still in Draft), advance payment is released. When the employee comes to retire, the Expense Claim is then submitted and the difference between Total Claimed Amount and Advances Paid (whether positive or negative) can then be booked to the employee's Payable account 

2. There needs to be a way to track Total Advances Paid against each Expense Claim (preferably within the Expense Claim itself)

3. It's helpful to directly create Advance Payment entries from within the Expense Claim to keep track of paid/approved amounts and automatically link the payments to the Expense Claim document

4. Actual expenses should not be booked before advance is paid else, it is no longer 'Advance'

5. I'm not really sure how the 'Payable Account' field was used for Employee Advances (which is actually a current asset account) but I think it should be a different field. Not also sure about the logic behind 'Payable account not having sufficient cash advance amount'

I believe the PR below by @nabinhait already covers most of the scenarios above

https://github.com/frappe/erpnext/pull/10632

Maybe you might want to consider tweaking that if there are some elements you think should be added or adjusted

Cheers!

Olawale1:
Hi @PawanMeh 

Going through again, I think I now understand your logic better. I'm assuming that the solution here is for scenarios where an advance payment has already been made to an employee (an Imprest account, actually) and then the employee makes expenses against that account

If that's the case, then it's fine but I think there still needs to be a different field for the Employee Advance account though

Cheers!

PawanMeh:
@Olawale1 , yes this is to knock off advance against employee ledger, the payable account would show asset accounts when "Against Advance" is checked, reason being employee advance is usually treated as a prepaid expense which is an asset account.

Olawale1:
Hi @PawanMeh 

> the payable account would show asset accounts when "Against Advance" is checked

I think you should consider adding a separate field called 'Advance Account' which shows up when "Against Advance" is checked rather than using the payable account field. It makes for a better and clearer user experience

> yes this is to knock off advance against employee ledger

Guessed as much 👍 . The PR by @nabinhait takes care of Advance against Expense Claim while this PR takes care of Expense Claim against Advance. I really hope both PRs get merged soon. It'll make expense claims usable in a number of organizations especially now that expenses get booked immediately a claim is submitted 

Cheers!

nabinhait:
@PawanMeh Closing this as per our discussion. Please check the feature developed in #10632, merging that PR.

PR #10752, opened 2017-09-12
Initialized module Agriculture

vishalseshagiri:


erpnext/agriculture/doctype/land_asset/land_asset.js
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2017, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Land Asset', {
```
rmehta: Land Unit ?

Issue #10759, opened 2017-09-13
AttributeError: 'DeliveryNote' object has no attribute 'recurring_print_format'


mbauskar:
```
Traceback (most recent call last):
  File "/Users/makarandbauskar/workspace/frappe/dev-bench/apps/erpnext/erpnext/subscription/doctype/subscription/subscription.py", line 120, in create_documents
    send_notification(doc, data.print_format, data.recipients)
  File "/Users/makarandbauskar/workspace/frappe/dev-bench/apps/erpnext/erpnext/subscription/doctype/subscription/subscription.py", line 178, in send_notification
    print_format = print_format or new_rv.recurring_print_format
AttributeError: 'DeliveryNote' object has no attribute 'recurring_print_format'
```

Issue #10760, opened 2017-09-13
Subscription Cleanup


rmehta:
- [x] Move to "Accounts"
- [x] Link in `module/Accounts`
- [x] Move documentation in Accounts
- [x] Status = 'Active' / 'Stopped'
- [x] Custom button for 'Stop', 'Resume'
- [x] Remove Cancel Permission

PR #10772, opened 2017-09-13
Move subscription module to accounts

rohitwaghchaure:
- [x] Move to "Accounts"
- [x] Link in `module/Accounts`
- [x] Move documentation in Accounts
- [x] Status = 'Active' / 'Stopped'
- [x] Custom button for 'Stop', 'Resume'
- [x] Remove Cancel Permission

Fixed https://github.com/frappe/erpnext/issues/10759, https://github.com/frappe/erpnext/issues/10760


erpnext/templates/emails/recurring_document_failed.html
```diff
@@ -2,11 +2,10 @@
 
 <p>An error occured while creating recurring {{ type }} <b>{{ name }}</b> for <b>{{ party }}</b>.</p>
```
mnatalia: just a question, don't you think it should be possible to translate this message?

rohitwaghchaure: Thanks for your review, will fix it!

mnatalia: sorry again but I think all sentences in the recurring_document_failed.html should have possibility to be translated

rohitwaghchaure: Can you please send PR after merging of this PR?

nataliamm: ok

mnatalia: ok

PR #10792, opened 2017-09-14
[BUG] Fixes make Production Order from Sales Order

neilLasrado:
Not sure why this was implemented the way it was. 

1. Sql must not include cancelled documents
2. Checking logic in js is complicated and wierd - !r.message.every(function(d) { return !!d.pending_qty })
3. get_production_order_items must only return items in the sales order which need to be produced

This PR fixes two issues - 

1. In case you raise PO against the SO and cancel the PO, the SO won't allow you to raise the PO again since the SQL did not check Cancelled docs; You had to delete the PO to do this

2. In case you have 2 items in the SO and you make a PO against one while trying to make the PO against the second item; the system did not allow it.

----------Update ---------------

The second commit enables tracking between the Sales Order Line Item and the Production Order.

Imagine a case where a Sales Order Has the same item twice in 2 lines instead of the Item having Qty 2.


nabinhait (changes requested):
Write a patch to set `sales_order_item` in existing Production Orders.

rmehta:
@neilLasrado can you add test cases for this?

neilLasrado:
@nabinhait the patch can be dangerous; since you really can't match the Sales Order Item with the Production Order accurately (In case you have a Sales Order with the same item in multiple lines)

neilLasrado:
@rmehta should I write a test case to check if this < get_production_order_items() > function returns the correct values ?

rmehta:
@neilLasrado write a full test of the scenario. with multi-item sales orders and production orders. Thanks!

neilLasrado:
@rmehta @nabinhait Not sure how to go forward with this. Test Cases just refuse to run. 

This is a bug in the system tough and needs urgent attention. the more the delay, the more the documents affected. Also, there is no way to write an accurate patch to fix it.

neilLasrado:
@rmehta Test Cases won't run locally. How do I write test cases without being sure if the test cases I wrote work? Do you expect me to write and commit test cases without testing them? 

If this is the case, let me know, will commit my code.
Else if test cases are being fixed, will wait till they are fixed; please update with some estimated timeline. Thanks!

rmehta:
You can run your specific test cases with `bench run-tests --doctype "Production Order"`

Get them to work first. Try and make it work, don't complain 😅

neilLasrado:
Hi! What's the update on this?

nabinhait:
merged manually

PR #10817, opened 2017-09-17
[patch] remove employee name salary structure

saurabh6790:
salary slip5e9541

### DIFF
--- a/erpnext/patches/v8_9/remove_employee_from_salary_structure_parent.py
+++ b/erpnext/patches/v8_9/remove_employee_from_salary_structure_parent.py
@@ -1,5 +1,5 @@
 import frappe
 
 def execute():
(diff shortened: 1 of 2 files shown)


erpnext/patches/v8_9/remove_employee_from_salary_structure_parent.py
```diff
@@ -0,0 +1,5 @@
+import frappe
+
+def execute():
+	if frappe.get_meta('Salary Structure').has_field('employee'):
```
nabinhait: Don't check in meta, instead check in database table.
`if 'employee' in frappe.db.get_table_columns("Salary Structure")`

nabinhait (changes requested):

PR #10834, opened 2017-09-18
[UI Test] Tests for payment entry against purchase invoice 

utkarsh-goswami:
#10668 
Tests to check if the invoice number is correctly mapped in payment entry.

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/tests/test_payment_against_purchase_invoice.js
+++ b/erpnext/accounts/doctype/payment_entry/tests/test_payment_against_purchase_invoice.js
@@ -1,0 +1,42 @@
+QUnit.module('Payment Entry');
+
+QUnit.test("test payment entry", function(assert) {
+	assert.expect(5);
+	let done = assert.async();
+
+	frappe.run_serially([
+		() => {
+			return frappe.tests.make('Payment Entry', [
+				{payment_type:'Pay'},
+				{company:'For Testing'},
+				{party_type:'Supplier'},
+				{party:'Test Supplier'},
+				{paid_from:'Cash - FT'}
+			]);
+		},
+		() => cur_frm.save(),
+		() => frappe.tests.click_button('Submit'),
+		() => frappe.tests.click_button('Yes'),
+		() => frappe.timeout(2),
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/payment_entry/tests/test_payment_against_purchase_invoice.js
```diff
@@ -0,0 +1,42 @@
...
+		},
+		() => cur_frm.save(),
+		() => frappe.tests.click_button('Submit'),
+		() => frappe.tests.click_button('Yes'),
+		() => frappe.timeout(2),
+		() => frappe.timeout(3),
+		() => {
+			assert.equal(cur_frm.doc.party, 'Test Supplier',
+				'supplier set in payment entry');
+			assert.equal(cur_frm.doc.paid_amount, 590,
```
nabinhait: paid_amount can be anything, it depends on outstanding purchase invoices. If we add/edit some test cases which changes outstanding amount  of purchase invoices, it will fail.

nabinhait: Try to rewrite this like https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/payment_entry/tests/test_payment_against_invoice.js

nabinhait (changes requested):

nabinhait:
Closing, check comments.

erpnext/accounts/doctype/payment_entry/tests/test_payment_against_purchase_invoice.js
utkarsh-goswami: @nabinhait Okay.

utkarsh-goswami:
@nabinhait 
Test for purchase invoice is already done. So can't I use this purchase invoice to test the payment entry?
https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/purchase_invoice/test_purchase_invoice.js
Or should I write a test to make a independent purchase invoice and then test a payment entry of it?
I thought as if a test of purchase invoice is already created then I can use it in any other test.

PR #10842, opened 2017-09-19
Auto Reorder: Propagate projected_qty all the way to the root warehouse

ppd:
### Problem
"Check in (group)" in item auto reoder does only entail checking the group's immediate children warehouses for quantities of the item to be ordered. This is problematic if the user requires a more complex warehouse topology with deeper nested warehouses.

### Solution
Propagate projected_qty from the leaf warehouses ("no group") up to the root warehouse instead of just one level up.
This makes "Check in (group)" work for non-trivial warehouse topologies

For further info see https://discuss.erpnext.com/t/auto-reorder-check-in-group-does-not-check-children-warehouses/28270


ppd1990:
Do you need something else? An issue, documentation...?

PR #10860, opened 2017-09-20
Removed auto creation of Sales Invoice when customer pays against a payment request

neilLasrado:
Changes were done as per conversation at https://github.com/frappe/erpnext/pull/10706


nabinhait:
@neilLasrado Please add an option of disabling creation of sales invoice for shopping cart in Shopping Cart Settings. And it will be an option only for Shopping Cart. If payment request made against SO manually, then don't create invoice automatically.

Closing this PR, please reopen after changing.

PR #10874, opened 2017-09-22
Fetching Document ID also in Email Subject while clicking "Send Supplier Emails" button

AravindPranera:
If we normally send email, it will fetch the document id in subject. But if we send email via "Send Supplier Emails" button it doesn't fetch currently. This code will solve that issue.


AravindPranera:
Here is a screenshots.
mail
send supplier email button

PR #10877, opened 2017-09-22
Revert "Advance against expense claim"

nabinhait:
Reverts frappe/erpnext#10632


nabinhait:
@olawale1 I have reverted the pull request #10632. The design is not very intuitive, mainly keeping Expense Claim in Draft mode for making advance against it and submitting on actual claim. I will redesign it soon.

Olawale1:
Hi @nabinhait 

I also wasn't thoroughly satisfied with that part but the current implementation of Expense Claims makes it a bit difficult to work around. What do you have in mind (design-wise)?

Thanks

Olawale1:
Hi @nabinhait 

Trust you're doing great. Any ideas for new design yet? We've been doing a lot of thinking about this and it seems the most elegant solution might be:

1.  Have a separate DocType 'Advance Request' which is similar to the Expense Claim document
2.  The approval flow for this document should be similar to that of Expense Claim
3.  On submission of the Advance Request form, buttons for 'Make Advance Payment' and 'Make Expense Claim' should become available
4.  Clicking on 'Make Expense Claim' should create an Expense Claim document with all entries automatically mapped. These can then be adjusted for actual values
5.  Settlement (Payment) for the Expense Claim should be against the employee 'Advance Account'

I think you get the general idea. This is obviously a longer way around so if you have any suggestions on how to improve that part of the flow within the Expense Claim document itself (without needing a separate DocType), that would be splendid

Thanks

nabinhait:
@Olawale1 Yes, the similar design is also in my mind. We are thinking to develop the feature on the first day of ERPNext Conference.

Olawale1:
Hi @nabinhait 

That's great. Look forward to updates day after tomorrow :))

Kind regards,

PR #10897, opened 2017-09-25
Getting last purchase price of an item

shreyashah115:
Summary:
1. Added a new column field `last_purchase_rate` to purchase order item child table.
2. On adding a new item to the purchase order, last purchase rate of the particular item gets fetched.
3. Removed the button `Get last purchase rate`, which previously did the same job, but on clicking the button.
4. Added a test case for the same.

anim


erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -24,6 +24,19 @@ frappe.ui.form.on("Purchase Order", {
 	},
 });
 
+frappe.ui.form.on("Purchase Order Item", {
+	item_code: function(frm) {
+		frappe.call({
+			"method": "get_last_purchase_rate",
+			"doc": cur_frm.doc,
+			callback: function(r, rt) {
+				cur_frm.dirty();
```
netchampfaris: Avoid using `cur_frm` in code. Use `frm` instead of `cur_frm`.

netchampfaris (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.js
shreyashah115: Done!

Issue #7389, opened 2017-01-05
Payroll Feature Request - Option to Exclude Earning/Deduction from Total


OneiricIT:
**The option to exclude an earning or deduction entry from being calculated in the total**
Perhaps like a check box? Example being some earnings may not be directly deposited into the employees account (i.e. Superannuation in Australia).

KanchanChauhan:
https://discuss.erpnext.com/t/payroll-employer-contribution/19456/5

jwrober:
This will be a great feature if you all can add it in. In the US there are a number of employer paid items that should show on the payslip like health insurance.. Then there are other items like social security tax that both sides pay and typically the employer side does not show on the slip, but is paid at the same time.

OneiricIT:
Hi @KanchanChauhan,
Would this feature be on the road map for v8 by chance?

KanchanChauhan:
It will need more information like how the accounting entry will be handled for such salary components.

jwrober:
We talk about that in the forum discussion linked above. It would work essentially the same as the other component types, but would not come out of the total salary.

OneiricIT:
@KanchanChauhan Can you confirm whether this has been addressed in the below addition to v8.0.11? 

> Introduce the ability to specify in a Salary Structure that a component is statistical. This allows components to be used in calculations without being added/deducted from earnings deductions.

[https://github.com/frappe/erpnext/releases](url)

KanchanChauhan:
@OneiricIT Unfortunately not yet. But another wave of changes in Payroll coming soon so might incorporate it in that.

OneiricIT:
@KanchanChauhan Great, glad to see it is still on the map.

ckosiegbu:
Does this PR https://github.com/frappe/erpnext/pull/8429 address this issue?

jwrober:
For the most part yes. The only thing else I would like to see is the ability to flag some of them to be shown on the payslip as "employer contributions" just to show employees what we are paying for.

OneiricIT:
If I am not mistaken (please correct me if I am wrong), it does not appear to actually contribute to payroll journal entries. So these are only statistical information that you can play with which are not reflected in any CoA. So I do not believe it addresses this request.

Also that appears to be feature I referenced on 18th April which @KanchanChauhan advised was not applicable to this.

utkarsh-goswami:
#10900

PR #10900, opened 2017-09-25
Option added to Exclude Earning/Deduction from Total in Payroll 

utkarsh-goswami:
#7389

### DIFF
--- a/erpnext/hr/doctype/salary_detail/salary_detail.json
+++ b/erpnext/hr/doctype/salary_detail/salary_detail.json
@@ -12,6 +12,7 @@
  "editable_grid": 1, 
  "fields": [
   {
+   "allow_bulk_edit": 0, 
    "allow_on_submit": 0, 
    "bold": 0, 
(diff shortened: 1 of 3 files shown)


utkarsh-goswami:
finl

erpnext/hr/doctype/salary_detail/salary_detail.json
```diff
@@ -314,6 +324,37 @@
...
+   "fieldname": "exclude", 
+   "fieldtype": "Check", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Exclude",
```
rmehta: `Exclude` is not clear. This should be named as `Do not include in total`

rmehta (changes requested):

PR #10903, opened 2017-09-25
Properly redirect supplier after quotations submit

brownharryb:
As it is now, suppliers responding to RFQ, after submitting are redirected to an error page. The supplier will think the quote has not been submitted and raise an alarm.

supplier_redirect_error

This redirects them to the submitted quote.


nabinhait:
👍 Thanks for the first PR

Issue #10908, opened 2017-09-25
If you have done Setup Progress in English and later you switch language, it is throwing error


KanchanChauhan:
Looks like encoding or translation issue. Unfortunately cannot dig deeper at this moment.

<img width="1031" alt="screen shot 2017-09-26 at 12 19 30 am" src="https://user-images.githubusercontent.com/16913064/30833743-9c77cdba-a250-11e7-8c4d-4d0de3b81805.png">

Sent in a PR https://github.com/frappe/erpnext/pull/10909. It solved the issue but not sure if it is the right way to do it. Please review.

KanchanChauhan:
@pratu16x7 Any progress on this, all our users use German language and everyone is facing this issue. It is kind of urgent.

pratu16x7:
#10911

applepipe:
thank you @KanchanChauhan this solved my issue https://github.com/frappe/erpnext/issues/10835

PR #10909, opened 2017-09-25
[Minor] Looks like action_name does not need to be translated in Setup Progress

KanchanChauhan:
For more info look at issue https://github.com/frappe/erpnext/issues/10908


pratu16x7:
Thanks for the fix! It needed patch fixes as well, so sent another one on #10911. :)

PR #10920, opened 2017-09-26
Number format based on locale for supplier quotations

brownharryb:
As it is now, 
When a supplier responds to an rfq, the quantity and rate inputs on the rfq form does not take into account the decimal for different locales, thereby giving erratic grand totals
supplier_quote_err

This makes use of the numeral.js library to get the correct locale number for the inputted string on both qty and rate
http://numeraljs.com

### DIFF
--- a/erpnext/templates/includes/rfq.js
+++ b/erpnext/templates/includes/rfq.js
@@ -30,9 +30,9 @@ rfq = Class.extend({
 	change_qty: function(){
 		var me = this;
 		$('.rfq-items').on("change", ".rfq-qty", function(){
-			me.idx = parseFloat($(this).attr('data-idx'));
-			me.qty = parseFloat($(this).val()) || 0;
-			me.rate = parseFloat($(repl('.rfq-rate[data-idx=%(idx)s]',{'idx': me.idx})).val());
+			me.idx = parse_float_with_locale($(this).attr('data-idx'));
+			me.qty = parse_float_with_locale($(this).val()) || 0;
+			me.rate = parse_float_with_locale($(repl('.rfq-rate[data-idx=%(idx)s]',{'idx': me.idx})).val());
 			me.update_qty_rate();
 			$(this).val(format_number(me.qty, doc.number_format, 2));
 		})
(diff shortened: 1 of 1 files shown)


netchampfaris:
@brownharryb 
We already have formatting capabilities in frappe. Loading another library just for this seems overkill.

What is the language you set which causes the error ? Let's discuss an alternate solution.

brownharryb:
Hi @netchampfaris 
The issue is when a value is entered into qty or rate that is already formatted (precisely with a comma) and the input box loses focus
supplier_err_1

... the `parseFloat` js function converts it to a float and strips the comma (since it's not a valid float)
supplier_err_2

... this is what is used to calculate the actual grand total which is misleading to the supplier.

brownharryb:
Hi @netchampfaris 
Please can you back to me on this
Thanks

revant:
@brownharryb 
Genuine use case and must be handled. Thanks for PR.

Can you check this out? https://remysharp.com/2016/12/13/format-numjsor-es6. New JS is powerful!

because I also agree with @netchampfaris. Adding library just to use it in one function is overkill.

brownharryb:
Hi @netchampfaris 
Thanks for the reply.
I have checked out the link, they emphasize the use of toLocaleString() which is fine, but basically, this converts a number to a locale string. We require the opposite of this function in this use case ie converting a locale string to a valid float (used in subsequent calculations), pretty much these 

https://stackoverflow.com/questions/7571553/javascript-parse-float-is-ignoring-the-decimals-after-my-comma 
https://stackoverflow.com/questions/12694455/javascript-parsefloat-in-different-cultures

The solutions on SO suggest replacing commas with periods, but this also won't work in some regions that use commas to separate thousands.
I'm open to any more suggestions.

The numeraljs lib seems to handle this well and I think there will be future cases(if there aren't already) where frappe will need to deal with this issue not just this one.

What do you think?

revant:
Even if we add it to library? 
You need to add library in build.json and public/js. Generally libraries are part of assets and not cdn link. 

refer 
https://discuss.erpnext.com/t/build-json-and-custom-css/27138

brownharryb:
Hi @revant
Done
https://github.com/frappe/frappe/pull/4528

revant:
I really don't know whether this should be merged immediately or closed or rebased and merged when required?

Now that if someone plans to make heavy usage of numerals.js for core, this becomes reference point 

Thanks for your effort @brownharryb

netchampfaris:
If we are using a library, might as well use it in `frappe.format` for Currency type fields ?

rmehta:
@brownharryb closing this for now. This will be fixed when web-forms become 1st class

PR #10924, opened 2017-09-27
[hotfix] Wrong calculation of total in taxes and totals

rohitwaghchaure:


nabinhait:
@rohitwaghchaure I have merged this as it is in hotfix and urgent. But check why the bug was not identified in client side testcases. If the specific test is not there, add the same.

PR #10940, opened 2017-09-27
[fix] Non working hours of workstation added into the timesheet scheduling on submission of production order.

### LINKED ISSUE #3917: Timesheet against Production Order should be made for BOM qty (not Production Order qty)
- [ ] Timesheet should be generated based on BOM Qty not total qty. For example if BOM qty is 1 and Production Order qty is 3, there should be 3 timesheets (for 1 each)

On submitting Production Order, following is the error message:

screen shot 2015-08-20 at 6 43 39 pm

This issue arises if Working Hours table in the Workstation master is blank. After updating value in the said table, Production Order gets submitted.

We should either improve validation message, so that user knows what needs to be fixed. Or we can make Working Hours table mandatory in the Workstation master.

### LINKED ISSUE #10845: Issue with the Working hours of a Work station.
Reported via https://frappe.erpnext.com/desk#Form/Issue/WN-SUP27962

We can add working hours for a work station but when a time sheet is created for a workstation after production order it considers non working house as working hours.

working hours1
working hours 2

rohitwaghchaure:
**Issue**
We can add working hours for a work station but when a time sheet is created for a workstation after production order it considers non working house as working hours.
workstation_timings

Showing non working hours in the timesheet
timesheet_issue
Fixed https://github.com/frappe/erpnext/issues/10845
fixed https://github.com/frappe/erpnext/issues/3917

### DIFF
--- a/erpnext/manufacturing/doctype/production_order/production_order.py
+++ b/erpnext/manufacturing/doctype/production_order/production_order.py
@@ -11,7 +11,8 @@ from erpnext.manufacturing.doctype.bom.bom import validate_bom_no, get_bom_items
 from dateutil.relativedelta import relativedelta
 from erpnext.stock.doctype.item.item import validate_end_of_life
 from erpnext.manufacturing.doctype.workstation.workstation import WorkstationHolidayError
-from erpnext.projects.doctype.timesheet.timesheet import OverlapError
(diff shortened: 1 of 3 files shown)


nabinhait (changes requested):
@rohitwaghchaure Add test cases and reopen the PR

Issue #9209, opened 2017-06-09
Trial Balance Order not correct 


Helkyd:
Hi,
How can i fix this as you can see on the two images the order is not correct and when i click on Account to sort it make the TOTAL as the first and the rest on the second row ...while the rest is in the correct order.
Based on the https://discuss.erpnext.com/t/trial-balance-order-not-correct/24703
you can find the attachments there.

addymuliady:
Balance sheet is also an issue

![image](https://user-images.githubusercontent.com/21187126/27796971-7a722098-603e-11e7-987c-2ebc4ca8c804.png)

Profit and Loss is an issue too

![image](https://user-images.githubusercontent.com/21187126/27797067-d9e3b7d0-603e-11e7-9987-ea3a15d5061b.png)

Helkyd:
Hi,
I can see that this was not fixed on the new version 
ERPNext: v8.11.2 (master)
Frappe Framework: v8.10.5 (master)

Still showing me on the wrong order

![image](https://user-images.githubusercontent.com/20396898/30482436-ecfc22e2-9a1a-11e7-8754-f375905c99d5.png)

Helkyd:
**Fix for Balance_sheet.py**
def get_accounts(company, root_type):
	return frappe.db.sql("""select name, parent_account, lft, rgt, root_type, report_type, account_name from `tabAccount`
		where company=%s and root_type=%s order by account_name, lft""", (company, root_type), as_dict=True)


**fix for Trial_balance.py**
def get_data(filters):
	accounts = frappe.db.sql("""select name, parent_account, account_name, root_type, report_type, lft, rgt
		from `tabAccount` where company=%s order by account_name, lft""", filters.company, as_dict=True)
	company_currency = erpnext.get_company_currency(filters.company)


This is under 
erpnext 8.11.6
frappe 8.10.9

tundebabzy:
Hi @Helkyd 
We'll really love for you to send this as a PR :)

lasalesi:
+1 chart of accounts (and other documents) must be sorted by account number (if there is no account number by account name)

PR #10945, opened 2017-09-27
Trial Balance, Balance Sheet and Profit and Loss Statement order Fix

Helkyd:
Based on this issue https://github.com/frappe/erpnext/issues/9209#issuecomment-332634058

Fix for Balance_sheet.py
def get_accounts(company, root_type):
return frappe.db.sql("""select name, parent_account, lft, rgt, root_type, report_type, account_name from tabAccount
where company=%s and root_type=%s order by account_name, lft""", (company, root_type), as_dict=True)

fix for Trial_balance.py
def get_data(filters):
accounts = frappe.db.sql("""select name, parent_account, account_name, root_type, report_type, lft, rgt
from tabAccount where company=%s order by account_name, lft""", filters.company, as_dict=True)
company_currency = erpnext.get_company_currency(filters.company)

This is under
erpnext 8.11.6
frappe 8.10.9


nabinhait:
> order by account_name, lft

@bohlian It should not be like that, it will not work if there is no account number.

Also, you have implemented it in develop branch. You cannot send the pull requests to v8.x.x.

Issue #10955, opened 2017-09-28
Allow users to only save text-only Item descirptions as it messes up the print formats


rmehta:
- [ ] Create a new DocType "Item Settings"
- [ ] "Convert description to text" checkbox
- [ ] Checked by default in new installs
- [ ] On save, convert all existing items to text

PR #10957, opened 2017-09-28
[fix] add item settings

rmehta:
fixes https://github.com/frappe/erpnext/issues/10955


nabinhait:
@rmehta Please fix conflicts and check travis error.

Issue #10960, opened 2017-09-28
GST section under Accounts should be visible only if country is India, other countries do not need it


KanchanChauhan:
<img width="670" alt="screen shot 2017-09-28 at 3 21 02 pm" src="https://user-images.githubusercontent.com/16913064/30960515-dc3bc674-a460-11e7-875e-a09b55aa8f9a.png">

tundebabzy:
#10965

PR #10965, opened 2017-09-28
GST section under Accounts should be visible only if country is India, other countries do not need it (#10960)

tundebabzy:
As pointed out in #10960, ERPNext will only show GST links if it finds a Company that has its country set as India in the Accounts module page.

I needed to be able to test `get_data` so I changed the signature thus: `get_data(**kwargs).

Test case added.

### DIFF
--- a/erpnext/config/accounts.py
+++ b/erpnext/config/accounts.py
@@ -4,7 +4,13 @@ import frappe
 from frappe import _
 
 
-def get_data():
+def get_data(**kwargs):
+	"""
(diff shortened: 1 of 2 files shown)


nabinhait:
It will be difficult to maintain config changes based on region. Instead it should be handled using Roles. By default no roles should be there for regional doctypes and reports. Permissions should be applied via regional setup.

Actually, by mistake the permissions for GST doctypes were introduced. I have reverted it.
https://github.com/frappe/erpnext/pull/10984

Closing this.

Issue #6242, opened 2016-09-01
Weakness when receiving email in ERPNext


ai-erp:
There is weakness in ERPNext for active incoming email and Append To Opportunity. When sender is already a customer in ERPNext, the system doesn’t understand that the sender is already customer and it is creating new lead for that. The system should first check who is the sender, and if the sender is already in the contact of customer or supplier or any other contact, it should not add new lead for the sender.

PR #10976, opened 2017-09-29
[minor] don't create lead if customer contact is already created against contact_email

mbauskar:
fixes for https://github.com/frappe/erpnext/issues/6242

- [x] Set customer instead of creating new Lead if the customer contact is already available
- [x] fix or add test cases for Opportunity.


joolsr:
So in this case as described will any emails create a new opportunity which will automatically be set against the existing customer?

mbauskar:
@joolsr,

Yes, New Email from existing contact (linked with Customer) will create new Opportunity against the Customer instead of Lead

PR #10982, opened 2017-09-29
[Enhancement] Allow Multiple users in POS Profile

### LINKED ISSUE #6308: POS - Allow Multiple POS Profiles for a User

netchampfaris:
closes https://github.com/frappe/erpnext/issues/6308

image

image


rohitwaghchaure:
@netchampfaris 

Write patch to update pos profile user

emushene:
Does anyone know how we can implement this for multiple warehouses in one POS Profile? I have tried Table-Multiselct, but this won't allow me to change the docType, editing the warehouse from link to table multi-select. 

Looking forward to suggestions.

PR #11012, opened 2017-10-02
[minor edits] BOM Stock Report

creamdory:
We needed to show filters in PDF in BOM Stock Report and we believe it is best to have it in the core using an HTML template. Since filters are both Mandatory, it make sense to have it in PDF for additional information. See below:

Current : 

bom stock report default

Proposed:

propposed

Also, I have added BOM Stock Report in Manufacturing Config under Report:

manufacturing

Regards,

Dori


creamdory:
Travis error is unrelated to the commit. Pls check and let me know if there is any changes. Thanks!

PR #11028, opened 2017-10-03
[Minor] Employee name in standard filter

KanchanChauhan:
I could have customized the form and set it for ourself but it makes sense to be able to search for employee in employee list rather than using global search.

screen shot 2017-10-03 at 10 31 36 am


nabinhait:
@KanchanChauhan Rebase with develop branch and send the PR again.

KanchanChauhan:
@nabinhait opened a new one https://github.com/frappe/erpnext/pull/11043

PR #11048, opened 2017-10-04
Add Items Variants Attributes columns to Stock Balance Report

brownharryb:
Show Item attributes in stock balance report for item variants if they exist.

stock_balance_report

### DIFF
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -15,6 +15,7 @@ def execute(filters=None):
 	item_map = get_item_details(filters)
 	item_reorder_detail_map = get_item_reorder_details(filters)
 	iwb_map = get_item_warehouse_map(filters)
+	variants_attributes = get_variants_attributes()
 
 	data = []
 	for (company, item, warehouse) in sorted(iwb_map):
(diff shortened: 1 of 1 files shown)


nabinhait:
What if there are many attributes (lets say 50 or more)? The report will not look good. How about adding one single column which will show all the attribute values for that item.

ckosiegbu:
@brownharryb Or perhaps a filter at the top that lets us narrow down using select fields for the attribute and attribute value respectively.

nabinhait:
@brownharryb Let's make it optional by adding a filter checkbox "Show Item Variant Attributes". If checked, then add those columns, default should be unchecked.

brownharryb:
@nabinhait 
Valid point, but say a client wants to filter based on multiple attributes eg "I want to see all items in my warehouse with colour=blue and length=456m", putting all in a single column might not be that practical in terms of filtering

I have two proposals

- Add a single input box that displays attributes columns based on comma separated attributes
comma_sep_attr
comma_sep_attr2

- Make it a frappe list style filtering
comma_sep_attr3

Would appreciate a feedback
Regards

nabinhait:
@brownharryb I understand your point, basic purpose of those columns is filtering. So, keep separate columns for each attributes. But those columns will be added only if checked "Show Item Variant Attributes" filter.

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -36,8 +37,14 @@ def execute(filters=None):
 			qty_dict.bal_val, qty_dict.val_rate,
 			item_reorder_level,
 			item_reorder_qty,
-			company
-		])
+			company,
+
+		]
+		report_data += [get_variant_attr_val(item,i) for i in variants_attributes]
```
nabinhait: Doing a query for each row is not good. Instead create a dict for all item attribute values outside for loop and then use that dict to get attribute value.

```
{
  item1: {
    'color': 'blue',
    'size': 'XL'
  },
  item2: {
    'size': 'L',
    'pattern': 'Stripe'
  }
}
```

For example, check https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/report/sales_register/sales_register.py#L28

nabinhait (changes requested):

brownharryb:
Hi @nabinhait 
This has been redone 
https://github.com/frappe/erpnext/pull/11278

PR #11063, opened 2017-10-05
[minor] Add "Attendance" in grid view for Training Event Employee

creamdory:
It would be best to show "Attendance" in Training Event Employee Child table. This is an important information that needs to be viewed in one look. Also, we still have enough space for a column in the list view.

training event employee


creamdory:
Travis error not related to PR. Let me know if there are any issues. Thanks!

PR #11067, opened 2017-10-05
Training Events: Add multirows in  Attendees selection

kickapoo:


dufani1:
good idea what if we add it to email list when sending an email to select contacts

kickapoo:
@dufani1 Use Email Alerts for this.

dufani1:
@kickapoo thanks for quick reply.. please see what I mean in the following issuse
https://github.com/frappe/erpnext/issues/11193

PR #11093, opened 2017-10-06
Add has_certificate and level fields in Training Event

creamdory:
Training Events have different Types (Seminar,Theory,Workshop,Conference,Exam,Internet,Self-Study). It makes sense that for some of these types, there is a **Level** (Beginner, Expert Advance) and a tag if the training **Has Certificate** .

**Level** depends on _Seminar, Workshop and Exam_ while **Has Certificate** depends on _Seminar, Workshop, Conference and Exam_.

added fields


creamdory:
Travis error is not related to PR. Please check and let me know if there are any changes. Thanks!

PR #11112, opened 2017-10-09
Replace c3

pratu16x7:
'depends-on': frappe/frappe#4228

prod_anal


Olawale1:
Hi @pratu16x7 

This looks great! Does it also display information when cursor hovers on it like the current charts?

Kind regards,

PR #11113, opened 2017-10-09
[Enhancement] Shipping rule for Buying

KanchanChauhan:
Right now Shipping Rules are just in Selling Module. Applied them in Buying Module and enhanced Shipping Rules.
Shipping Rule Changes:

- Added type for Buying and Selling
- Added Fixed Amount
- Fixed worldwide shipping
- Added documentation
- Patch to update current shipping rules to type Selling since all are selling right now

sr

Added Shipping Rules to Supplier Quotation, Purchase Order, Purchase Receipt and Purchase Order:

Supplier Quotation
screen shot 2017-10-09 at 9 57 47 am

Purchase Order
screen shot 2017-10-09 at 9 58 59 am

Purchase Receipt
screen shot 2017-10-09 at 9 59 28 am


erpnext/controllers/buying_controller.py
```diff
@@ -420,4 +420,39 @@ def validate_schedule_date(self):
 				if d.schedule_date and getdate(d.schedule_date) < getdate(self.transaction_date):
 					frappe.throw(_("Expected Date cannot be before Transaction Date"))
 		else:
-			frappe.throw(_("Please enter Schedule Date"))
\ No newline at end of file
+			frappe.throw(_("Please enter Schedule Date"))
+
+	def apply_shipping_rule(self):
```
rmehta: Please be a DRY programmer

https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/selling_controller.py#L70

rmehta (changes requested):

erpnext/controllers/buying_controller.py
rmehta: Move this to a common method in `shipping_rule.py`

KanchanChauhan: @rmehta Done. But I have put the common code in transaction.js and accounts_controller.py

KanchanChauhan:
Tests which are failing are unrelated to this PR.

erpnext/controllers/selling_controller.py
```diff
@@ -67,38 +67,6 @@ def set_price_list_and_item_details(self, for_validate=False):
 		self.set_price_list_currency("Selling")
 		self.set_missing_item_details(for_validate=for_validate)
 
-	def apply_shipping_rule(self):
```
rmehta: where is this method added back???

rmehta (changes requested):

erpnext/controllers/selling_controller.py
KanchanChauhan: Looks like last time I rebased something went wrong, it is in accounts_controller.py file.

rmehta: Thats why we need test cases! can you add a few test cases to ensure the Shipping rule works for both Sales and Purchase ?

KanchanChauhan: @rmehta Test cases done for both selling and buying but travis is failing again for some other reasons. :(

rmehta: 👍 will check other issues

rmehta:
https://github.com/frappe/erpnext/pull/11628

PR #11122, opened 2017-10-09
Add Bundle Description Field To Product Bundle

tundebabzy:
Bundle Description field is set with `in_list_view: 1` so that users can know what is in a bundle without having to first open the product bundle.

screenshot from 2017-10-09 09-49-13


erpnext/selling/doctype/product_bundle/product_bundle.json
```diff
@@ -72,6 +75,37 @@
    "unique": 0
   }, 
   {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "product_bundle_description",
```
rmehta: how about just `description` this is the Product Bundle DocType, so it will be called as `product_bundle.description`

rmehta (changes requested):

Issue #11102, opened 2017-10-06
[ENH] HR Injury or illness incidents


almeidapaulopt:
This is a proposal to create a "Injury or illness incidents" management and reporting to HR.

TODO: 

- [ ] Add "Injury or illness incidents" Doctype
- [ ] add "Injury or illness type" Doctype (injury, hearing, Poisoning,etc)
- [ ] add "Body Parts" Doctype (arm, leg, head, etc)
- [ ] add "Outcome Type" Doctype (days off, Medical, transfer, death, etc)
- [ ] add report "Incidents by Body Part"
- [ ] add report "Incidents on Last Year"
- [ ] add report "Incidents by type"

rmehta:
Best to add a clean namespace to these DocTypes, for example "Outcome Type" could mean anything.

- "HR Incident Report"
- "HR Incident Type"
- "HR Incident Body Part"
- "HR Incident Outcome"

almeidapaulopt:
@umairsy I don't think this should be in labeled to heathcare domain. This is something for all companies in their HR. (In several countries this is a law compliance obligation).

bcornwellmott:
This is a great idea that is definitely needed.

almeidapaulopt:
@rmehta , you propose in https://github.com/frappe/erpnext/pull/11128 to call it "Employee Incident" but this shouldn't be applied only to employees.
For example, in a Construction Site, there are workers subcontracted so they aren't employees but in some countries you have to record all work incidents, employees or not. So I propose

  "Worker incidents" Doctype
  "Worker Incident Type" Doctype
  "Worker Incident Body Part" Doctype
  "Worker Incident Outcome" Doctype
  "Worker Incident Severity level" Doctype

rmehta:
"Safety Incident" ?

almeidapaulopt:
What we will call to the reports "Injury and Illness Report Reports?" :smile: 
what do you think about "Injury and Illness Case" or my first proposal ""Injury and Illness Incident"?

rmehta:
Lets go with "Injury or Illness Case" then!

(too many I's in Injury or Illness Incident" !! 😄

PR #11124, opened 2017-10-09
HR Incident Body Part Doctype

almeidapaulopt:
PR for: https://github.com/frappe/erpnext/issues/11102

image


rmehta:
See comment https://github.com/frappe/erpnext/pull/11128

PR #11128, opened 2017-10-09
HR Incident Type Doctype

almeidapaulopt:
pr for https://github.com/frappe/erpnext/issues/11102

image


rmehta:
@almeidapaulopt sorry I had suggested these names, but I just realised that we should not use abbreviations

How about "Employee Incident"? Since its mostly static DocTypes, it should not be too hard!

Can you reopen all in one PR. Thanks!

PR #11131, opened 2017-10-09
HR Incident Outcome Doctype

almeidapaulopt:
pr for https://github.com/frappe/erpnext/issues/11102

image


rmehta:
https://github.com/frappe/erpnext/pull/11128

PR #11135, opened 2017-10-09
Payment entry showing wrong difference amount for party type supplier and payment type Pay (#11110)

### LINKED ISSUE #11110: Payment entry showing wrong difference amount for party type supplier and payment type Pay
screen shot 2017-10-08 at 12 20 53 pm

Paid Amount = 10000
Allocated = 7000
Bank Chanrges = 3000
Difference amount should be zero

tundebabzy:
Fixes #11110 
screenshot from 2017-10-09 15-29-58


rohitwaghchaure:
Fix causing an issue in the sales flow
sales_flow_issue

rohitwaghchaure (approved):

nabinhait:
@tundebabzy This is not a correct fix, it will introduce the old problem stated on #10145 

Let me work on the issue.

PR #11145, opened 2017-10-10
Error in the Address fetched in Sales Order.(#11129)

### LINKED ISSUE #11129: Error in the Address fetched in Sales Order.
Reported via https://frappe.erpnext.com/desk#Form/Issue/WN-SUP28330
If customer is having two address Shipping and Billing. The billing address in fetched in the sales order document. This works fine if the Shipping address is checked as preferred Shipping address.
Refer the gif attached. 

billing address issue

tundebabzy:
Fixed #11129 
Depends on frappe/frappe#4298 to work. Now, Sales Order will only try to retrieve Shipping Address For the party and leave blank if not found

### Before
peek 2017-10-10 16-57

### After
peek 2017-10-10 17-00

The function returns with considerations in this order:
- Return Address with `is_shipping_address = 1` and `address_type='Shipping'`
- Return Address with `address_type = 'Shipping`
- Return Address with `is_shipping_address = 1`
- Return ''

Test cases included

### DIFF
--- a/erpnext/accounts/utils.py
+++ b/erpnext/accounts/utils.py
@@ -740,32 +740,3 @@ def create_payment_gateway_account(gateway):
 	except frappe.DuplicateEntryError:
 		# already exists, due to a reinstall?
 		pass
-
-
-def get_party_shipping_address(doctype, name):
-	"""
-	Returns an Address name (best guess) for the given doctype and name for which `address_type == 'Shipping'` is true.
-	and/or `is_shipping_address = 1`.
-
-	It returns an empty string if there is no matching record.
-
-	:param doctype: Party Doctype
-	:param name: Party name
-	:return: String
-	"""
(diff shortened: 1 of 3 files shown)


rmehta:
@tundebabzy Lets keep these methods in erpnext, shipping is an erpnext functionality. maybe in accounts/utils.py

erpnext/accounts/utils.py
```diff
@@ -740,3 +740,32 @@ def create_payment_gateway_account(gateway):
 	except frappe.DuplicateEntryError:
 		# already exists, due to a reinstall?
 		pass
+
+
+def get_party_shipping_address(doctype, name):
```
saurabh6790: Keep this function in party.py itself. accounts/utils.py is specific to accounting functions

PR #11172, opened 2017-10-12
Employee Advance feature

nabinhait:
To manage advance payment against Expense Claim

screen shot 2017-10-12 at 2 50 47 pm


Olawale1:
Hi @nabinhait 

Trust you're doing great. This looks promising but I think there will need to be a table similar to what we have in the Expense Claim document. Many times, the approver needs to see a breakdown of estimates that make up the total figure. In cases where this is not necessary, then the Employee can use 1 row only. Actually, it would be best if this table could map to the Expense Claim details table

Also, I believe buttons to make Journal Entry (for advance payment) and Expense Claim (for settlement) will be included as well

Kind regards,

nabinhait:
@Olawale1 Lets not complicate this by introducing a child table. In most cases, user will not know the exact amount for each type of expenses. They will only know while doing actual claim. anyway, they can write the bifurcation of amounts in purpose.

Yes, there will be button for payment /Journal entry. And also workflow for expense claim is pending. Actually, we could not complete the feature in the conference. I will complete it for sure within next week.

Closing it for now.

Olawale1:
Hi @nabinhait 

> anyway, they can write the bifurcation of amounts in purpose.

Yes, you're right

> Actually, we could not complete the feature in the conference. I will complete it for sure within next week.

Thanks a lot Nabin, looking forward to this getting completed

Cheers!

Olawale1:
Hi @nabinhait 

Trust you're doing great. Hope we're still on course to finish the advance feature this week. I know you're really busy so I understand if things get shifted a bit

Cheers!

Olawale1:
Hi @nabinhait 

Trust you're doing great. Looking forward to some update on this

Thanks plenty

PR #11181, opened 2017-10-13
Add Training Program DocType, Script, Path and Doc

creamdory:
(**EDIT** : We have renamed the new Doctype to **Training Program**)

In relation to discuss https://discuss.erpnext.com/t/feature-training-event/29205 , our team presented improvement for Training Event by adding **Training Doctype** as a grouping for Training Events. The believe that it make sense to use **Training** instead of **Training Module** as the name for the new Doctype because the word "Module" can be misleading as we have Modules also in our ERP structure. Anyway, below are the improvements:

1. Adding **Training** under HR > Training
training

1. New Training Doctype with the following fields (Not Submittable but with Status), with dashboard and **Training Event** button to create New Training Event from it.
new training

1. Mandatory **Training** link field in **Training Event**.
training event

1. Patch to create Training in every existing Training Event (We assume that current record has 1:1 relationship. 1 Training Event in every Training created) **[To Follow once PR is merged.]**

training to training event


kickapoo:
@creamdory Update  commit message with  Add new DocType Training Program

erpnext/config/hr.py
```diff
@@ -178,6 +178,10 @@ def get_data():
 			"items": [
 				{
 					"type": "doctype",
+					"name": "Training"
```
kickapoo: Update name to 'Training Program'

kickapoo:
@creamdory very cool way with `frappe.model.mapper.get_mapped_doc` and `frappe.model.mapper.make_mapped_doc`.

erpnext/hr/doctype/training_program/training_program.py
```diff
@@ -0,0 +1,30 @@
...
+from frappe.model.mapper import get_mapped_doc
+
+class TrainingProgram(Document):
+	pass
+
+def set_missing_values(source, target):
+	target.run_method("set_missing_values")
+
+@frappe.whitelist()
+def make_training_event(source_name, target_doc=None):
```
kickapoo: Just add a minor doctstring like 
```
def make_training_event(source_name, target_doc=None):
      Returns a Training Event based on parent Training Program Fields
``

Just for future reference

File: erpnext/config/hr.py (hunk 1 above)
Comment (creamdory) [reply erpnext/config/hr.py]: Fixed @kickapoo

File: erpnext/hr/doctype/training_program/training_program.py (hunk 2 above)
Comment (creamdory) [reply erpnext/hr/doctype/training_program/training_program.py]: Added as comment @kickapoo Thanks!

**Comment by creamdory** [comment]
Travis error not related to the PR. Please check and let me know if there are any changes. Thanks!

**Comment by creamdory** [comment]
Hi @nabinhait , @rmehta Please review and let us know. Thanks!

File: erpnext/hr/doctype/training_program/training_program.js [hunk 3]
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2017, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Training Program', {
+	refresh: function(frm) {
+		if(frm.doc.status == 'Scheduled') {
```
rmehta: Never hardcode status.

You should also be able to see training events linked to this. Try adding a `dashboard` its pretty simple. (See `material_request_dashboard.py`

erpnext/hr/doctype/training_program/training_program.js
```diff
@@ -0,0 +1,18 @@
...
+			frm.add_custom_button(__("Training Event"), function () {
+				frm.trigger('make_training_event');
+			});
+		}
+	},
+	make_training_event: function() {
+		frappe.model.open_mapped_doc({
```
rmehta: You don't need a server method for this. If you use dashboard, the training program will automatically be linked based on the field.

rmehta (changes requested):

erpnext/hr/doctype/training_program/training_program.js
creamdory: I have already added a dashboard @rmehta

erpnext/hr/doctype/training_program/training_program.js
creamdory: I see. I should just Delete the button as the Dashboard is already sufficient. Thanks!

creamdory:
Hi @rmehta , done with the changes asked. Please check. Thanks!

rmehta:
Add the docs!

creamdory:
Yes of course @rmehta . Thanks!

creamdory:
@rmehta Added/Edited docs as requested. Thanks!

creamdory:
Hi @rmehta , we have added a test case and a patch too for existing Training Events. Please check and let us know. Thanks!

erpnext/patches/v9_0/create_training_program_for_training_event.py
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2017, Frappe and Contributors
```
rmehta: Lets not patch old Training Events. Lets keep Training Program Optional.

We will have to make this backward compatible, else we will have to change the version number.

rmehta (changes requested):
Lets just remove the patch!

erpnext/patches/v9_0/create_training_program_for_training_event.py
creamdory: Ok @rmehta

creamdory:
@rmehta Patch deleted and make Training Program link in Training Event set to not mandatory

creamdory:
Hi @rmehta ! Let us know if there are needs to be done before Merge. Thanks!

PR #11182, opened 2017-10-13
[hotfix] fixed Permissions Error while fetching allow_stale field value from Accounts Settings

mbauskar:
Permissions Error if the Purchase User try to create the `Purchase Receipt`
`before`
pr

`after`
pr

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -542,7 +542,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		}
 		// Make read only if Accounts Settings doesn't allow stale rates
-		this.frm.set_df_property("conversion_rate", "read_only", erpnext.allow_stale_rate());
+		this.frm.set_df_property("conversion_rate", "read_only", erpnext.stale_rate_allowed());
 	},
 
 	set_actual_charges_based_on_currency: function() {
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -542,11 +542,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		}
 		// Make read only if Accounts Settings doesn't allow stale rates
-		frappe.model.get_value("Accounts Settings", null, "allow_stale",
-			function(d){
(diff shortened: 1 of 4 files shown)


erpnext/public/js/utils.js
```diff
@@ -37,6 +37,10 @@ $.extend(erpnext, {
 		}
 	},
 
+	allow_stale_rate: () => {
```
tundebabzy: Let's name it `stale_rate_allowed`. `allow_stale_rate` can lead readers to think allow_stale_rate is being enabled

erpnext/public/js/utils.js
```diff
@@ -37,6 +37,10 @@ $.extend(erpnext, {
...
+		return cint(frappe.boot.sysdefaults.allow_stale) || 0;
```
tundebabzy: allow_stale should is 1 by default

erpnext/startup/boot.py
```diff
@@ -19,6 +19,8 @@ def boot_session(bootinfo):
 			'territory')
 		bootinfo.sysdefaults.customer_group = frappe.db.get_single_value('Selling Settings',
 			'customer_group')
+		bootinfo.sysdefaults.allow_stale = frappe.db.get_single_value('Accounts Settings',
+			'allow_stale') or 0
```
tundebabzy: allow_stale should be 1 by default

erpnext/public/js/controllers/transaction.js
```diff
@@ -542,11 +542,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		}
 		// Make read only if Accounts Settings doesn't allow stale rates
-		frappe.model.get_value("Accounts Settings", null, "allow_stale",
-			function(d){
-				me.set_df_property("conversion_rate", "read_only", cint(d.allow_stale) ? 0 : 1);
-			}
-		);
+		this.frm.set_df_property("conversion_rate", "read_only", erpnext.allow_stale_rate());
```
tundebabzy: let's rename `erpnext.allow_stale_rate` to `erpnext.stale_rate_allowed`

tundebabzy (changes requested):

tundebabzy (approved):

Issue #11225, opened 2017-10-16
Duplicate Sales Invoice Issue in POS V9


Olawale1:
Hi ,

Steps to replicate:

1. Create a POS sale for an item that is not in stock
2. Wait for Invoice to get synced (Invoice will be in Draft due to shortage of stock)
3. Make a new sale with the same item

The following error pops up and stops the sync from progressing:

"Duplicate name Sales Invoice SINV-xxxxx"

Please help fix asap

Thanks

rohitwaghchaure:
Hi @Olawale1 

Thanks for reporting an issue, we have identified and fixed the issue and soon we'll release it

Olawale1:
Hi @rohitwaghchaure 

I just noticed that this issue affects invoices even when they are not in Draft! Hope the fix takes care of that scenario too!

By the way, hope this will be merged very soon as it's holding up sync of sales!

Thanks

PR #11230, opened 2017-10-17
[Fix] Getting an error duplicate name while making an invoice in draft mode from offline POS

rohitwaghchaure:
Steps to replicate:

Create a POS sale for an item that is not in stock
Wait for Invoice to get synced (Invoice will be in Draft due to shortage of stock)
Make a new sale with the same item
The following error pops up and stops the sync from progressing:

"Duplicate name Sales Invoice SINV-xxxxx"

Fixed https://github.com/frappe/erpnext/issues/11225


Olawale1:
Hi,

This fix is for a critical bug affecting retail sales transactions in production setups. Please help merge on high priority

Thanks

PR #11261, opened 2017-10-20
Make Variant does not copy UOM properties (#10887)

### LINKED ISSUE #10887: Make Variant does not copy UOM properties [BUG]
When you have created a master and configured it accordingly, upon making a variant, many field will copy over, however the UOM items specified in the Master Template  do not copy over and must be manually set.  
For example, in Item Master Template, Stock UOM is
Unit  = 1
Box = 10
Box has 10 Units.

Behavior should be:
On Item Master with Variants > Press Make Button > Select Variant > Choose Variant name
New Item variant sheet opens
UOM should show:
Unit = 1
Box = 10

Manual modification of item variant is possible, but unrelated to this issue. This simply streamlines the creation of individual Item Variants for users.

tundebabzy:
Fixes #10887 
Bug stemmed from the fact that `uoms` was set as no_copy. `uoms` should also be in Item Variant Settings by default but isn't.

This PR fixes both issues

### DIFF
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -451,3 +451,4 @@ erpnext.patches.v8_9.update_billing_gstin_for_indian_account
 erpnext.patches.v9_0.fix_subscription_next_date
 erpnext.patches.v9_0.add_healthcare_domain
 erpnext.patches.v9_0.set_variant_item_description
+erpnext.patches.v9_0.set_uoms_in_variant_field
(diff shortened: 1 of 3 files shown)


nabinhait (changes requested):
Why so many changes in item.json?

tundebabzy:
Honestly I have no idea why frappe did all the other changes. Now that you mentioned, I guess it's safe to revert the extra changes. Will do that ASAP

Issue #9824, opened 2017-07-12
Updating costing in the Task and Project, even if "Bill" is not checked


dufani1:
Could we add feature of updating costing in the Task and Project, even if "Bill" is not checked.

![untitled](https://user-images.githubusercontent.com/23026263/28133287-c7c11810-673f-11e7-88f9-e7cbd45717cf.png)

umairsy:
Hello,

This feature of updating cost in the Task and Project is already independent "Bill" field in the Timesheet. Please ensure that you have updated Cost in the Activity Type. In the Timesheet, Cost per hour will be fetched from Activity Type selected and total cost will be calculated based on actual hours.

agerken:
@umairsy This issue is most definitely not resolved, I have the exact same issue. 
![task non bill](https://user-images.githubusercontent.com/18719200/28431509-9870fa94-6d52-11e7-835f-5d835f05d0c1.PNG)
![task billable](https://user-images.githubusercontent.com/18719200/28431514-99748028-6d52-11e7-96f5-beb21dd7ce83.PNG)

PawanMeh:
Fix is merged into develop, closing this issue for now.

PR #11310, opened 2017-10-24
[fix] #9824 - Calculate costing amount even if billable is unchecked

PawanMeh:
Testing Video

timesheet_costing_test


PawanMeh:
1. Travis server side tests have passed.
2. Client side test error does not seem related to this PR

Issue #11209, opened 2017-10-16
Bug - Incorrect Total and Qty in Online POS Cart


jay-parikh:
Adding items to cart by clicking quickly and multiple times causing incorrect Qty and Total.

Steps as below:

1. Add items to cart by clicking quickly and multiple time which causes Qty field not getting updated correctly.
2. Grand Total showing incorrect value.
3. Remove all items from cart one by one clicking (-) minus sign in Qty field.
4. Check Grand Total. Which should be non zero but showing some value.
5. Goto Menu and open form view, which is showing items which have been removed from Cart.

![qty](https://user-images.githubusercontent.com/3703266/31616023-25df725c-b2a9-11e7-9ec0-9625718949ee.gif)
![formview](https://user-images.githubusercontent.com/3703266/31616039-2e049cc8-b2a9-11e7-8976-a52361a36737.gif)

jay-parikh:
@rohitwaghchaure @umairsy 
I got the solution and it's working. Sending PR - [https://github.com/frappe/erpnext/pull/11335](https://github.com/frappe/erpnext/pull/11335) for this issue.

Here is working gif:
![bug - incorrect total and qty in online pos cart 11209](https://user-images.githubusercontent.com/3703266/31998462-9b368044-b9ad-11e7-9f81-9218c5e0acd1.gif)

jay-parikh:
@rohitwaghchaure Issue still exist and not resolved. I have tested with your fix today and still facing an issue. See gif
![image](https://user-images.githubusercontent.com/3703266/32219298-b9f72984-be53-11e7-863a-0294be3b2b15.png)

tundebabzy:
#11470

PR #11335, opened 2017-10-25
Bug - Incorrect Total and Qty in Online POS Cart #11209

jay-parikh:
Steps to reproduce bug before fix:
1. Add items to cart by clicking quickly and multiple time which causes Qty field not getting updated correctly.
2. Grand Total showing incorrect value.
3. Remove all items from cart one by one clicking (-) minus sign in Qty field.
4. Check Grand Total. Which should be non zero but showing some value.
5. Goto Menu and open form view, which is showing items which have been removed from Cart.

Steps to Verify bug after fix applied:
1. Add items to cart by clicking quickly and multiple time which causes Qty field to be updated one at a time.
2. Grand Total showing correct value.
3. Remove all items from cart one by one clicking (-) minus sign in Qty field.
4. Check Grand Total. Which should be zero.
5. Goto Menu and open form view, which doesn't have any items.

Here is Corrected behavior:
bug - incorrect total and qty in online pos cart 11209

### DIFF
--- a/erpnext/selling/page/point_of_sale/point_of_sale.js
+++ b/erpnext/selling/page/point_of_sale/point_of_sale.js
@@ -738,19 +738,24 @@ class POSCart {
 	bind_events() {
 		const me = this;
 		const events = this.events;
+		var flag = false;
 
 		// quantity change
 		this.$cart_items.on('click',
 			'[data-action="increment"], [data-action="decrement"]', function() {
-				const $btn = $(this);
-				const $item = $btn.closest('.list-item[data-item-code]');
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@jay-parikh 

Add timeout is not the proper solution to fix the issue. I have sent PR https://github.com/frappe/erpnext/pull/11361
Please test it on your account and share your feedback in the PR

jay-parikh:
Thanks @rohitwaghchaure 
Sure. I will share feedback soon.

PR #11361, opened 2017-10-27
[Fix] POS, fast click on item not update the qty proper in the cart

rohitwaghchaure:
**Issue**
Steps as below:

- Add items to cart by clicking quickly and multiple time which causes Qty field not getting updated correctly.

- Grand Total showing incorrect value.

- Remove all items from cart one by one clicking (-) minus sign in Qty field.

- Check Grand Total. Which should be non zero but showing some value.

- Goto Menu and open form view, which is showing items which have been removed from Cart.
https://github.com/frappe/erpnext/issues/11209

**After Fix**

pos_fast_click_issue

### DIFF
--- a/erpnext/selling/page/point_of_sale/point_of_sale.js
+++ b/erpnext/selling/page/point_of_sale/point_of_sale.js
@@ -169,6 +169,10 @@ erpnext.pos.PointOfSale = class PointOfSale {
 				value = item.serial_no + '\n'+ value;
 			}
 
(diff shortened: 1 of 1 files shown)


noetico:
I still have this issue after updating just today, its not fully fixed, I think caching should be totally dropped for the online pos.

jay-parikh:
@rohitwaghchaure Issue still exist and not resolved. I have tested with your fix today and still facing an issue. See gif
click issue

PR #11363, opened 2017-10-27
set bank account only if there is single bank type account

### LINKED ISSUE #11353: Set bank/cash account in the bank journal entry only if there is single account.

manassolanki:
Fixes https://github.com/frappe/erpnext/issues/11353

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -569,14 +569,18 @@ def get_default_bank_cash_account(company, account_type=None, mode_of_payment=No
 		if account_type=="Bank":
 			account = frappe.db.get_value("Company", company, "default_bank_account")
 			if not account:
-				account = frappe.db.get_value("Account",
-					{"company": company, "account_type": "Bank", "is_group": 0})
+				account_list = frappe.get_all("Account", filters = {"company": company,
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -569,14 +569,18 @@ def get_default_bank_cash_account(company, account_type=None, mode_of_payment=No
 		if account_type=="Bank":
 			account = frappe.db.get_value("Company", company, "default_bank_account")
 			if not account:
-				account = frappe.db.get_value("Account",
-					{"company": company, "account_type": "Bank", "is_group": 0})
+				account_list = frappe.get_all("Account", filters = {"company": company,
+					"account_type": "Bank", "is_group": 0})
+				if len(account_list) == 1:
```
rmehta: Should be

```
if len(account_list) > 0:
```

Also please show proof

rmehta (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.py
manassolanki: Actually the issue is that if there are multiple bank type account, firstly created account shouldn't be chooosed (in Journal Entry). If there is only one bank type account we can set it.

Issue #11354, opened 2017-10-26
POS profile details not fetching in the Sales invoice Report.


Arundhatii:
If the Sales invoice is created by POS profile and Report is generated from Sales invoice list the POS profile column is blank. 
![pos](https://user-images.githubusercontent.com/20164097/32052780-e06e8480-ba76-11e7-93ff-889f09c28103.png)

khushalti:
Seems issue have been fixed on develop

PR #11364, opened 2017-10-27
set the pos profile in the sales invoice

manassolanki:
Fixes https://github.com/frappe/erpnext/issues/11354


rohitwaghchaure (approved):

manassolanki:
**POS Profile Doctype**

pos

PR #11367, opened 2017-10-27
Injury or Illness Case Type

almeidapaulopt:
https://github.com/frappe/erpnext/issues/11102


rmehta:
Same PR for everything ?

almeidapaulopt:
Just trying to follow 30 lines code rule, but already sent new PR

PR #11381, opened 2017-10-29
Update issue.js

ci2014:
Make the close / reopen button more flexible for customized user status.

For example, if the user has several status, it will now show the Close button every time the ticket is not closed! It will show Reopen only, if the ticket is closed, and not if it's on hold for example.

This will make it more flexible.


rmehta:
@ci2014 will be great if you add better titles to your PRs 'update issue.js' is not good enough. Please help us in reviewing better. 😄

ci2014:
@rmehta I will in future, thank you for merging.

PR #11387, opened 2017-10-30
Use account_head instead of description while fetching tax details in item-wise reports

sagarvora:
Sales register uses `account_head` from the table of `Sales Taxes and Charges` while generating report. To maintain consistency, and even ideally, same should be used by Item-wise reports instead of `description`.

Before:
selection_046

After: 
selection_045


sagarvora:
Needs change to fix calculation of charges.

PR #11392, opened 2017-10-31
Getting Quotation Document in Sales Invoice

AravindPranera:
Here is a screenshot.
image
image
image


nabinhait:
merged from terminal

PR #11401, opened 2017-10-31
'_Test Project' issue travis fix

KanchanChauhan:
screen shot 2017-10-31 at 6 37 09 pm


nabinhait:
No need to call create_project from everywhere, it should be created based on test_records. The problem is "_Test Project" is not created because of some duplicate task. 

I have sent a fix, https://github.com/frappe/erpnext/pull/11445

Closing this.

KanchanChauhan:
@nabinhait yeah I think rebuild_tree in task.py was causing this issue. I think we are good with your PR.

PR #11413, opened 2017-11-01
Checkbox has_expiry_date in Item to make sure Expiry Date is entered in Batch

KanchanChauhan:
- Added checkbox `has_expiry_date` in Item which is visible if Item has Batch.
- If checked, it makes Expiry Date mandatory field while creating batch for that Item.
screen shot 2017-11-01 at 4 40 56 pm
screen shot 2017-11-01 at 4 41 15 pm


erpnext/stock/doctype/batch/batch.js
```diff
@@ -24,6 +24,13 @@ frappe.ui.form.on('Batch', {
 			frm.trigger('make_dashboard');
 		}
 	},
+	item: (frm) => {
+		frappe.db.get_value('Item', {name: frm.doc.item}, 'has_expiry_date', (r) => {
+			if (r.has_expiry_date == 1) {
+				frm.toggle_reqd('expiry_date', true)
+			}
```
nabinhait: Instead of this `if` condition, write like `frm.toggle_reqd('expiry_date', r.has_expiry_date)`. Otherwise, you also have to write `else` condition. 
Currently, if an item has expiry date, it will make expiry_date mandatory and then if you change the item which does not have expiry date, still the date will remain mandatory.

nabinhait (changes requested):

erpnext/stock/doctype/batch/batch.js
KanchanChauhan: @nabinhait You are right, thanks for the tip. Fixed it.

PR #11419, opened 2017-11-01
sub_item into Quotation

Simeon-Banov:
A sub item is a price rule ignore, but for a single item, not the document.

Enable Items in Quotation to be able to be marked as a Sub Item, meaning that it is a part of the Item above, so it has no price. The print format may also be changed not to show the price of the Sub Item. A button may be placed on Quotation Item table to mark multible Items as Sub Item.

There is a need for this for companies which produce Items from other Items. Showing the parts used in creating them in the print format, so marking them as a Sub Item is a nice way to go. The Quotation is the most important place to have that, other doctypes are not that important in this process.


rmehta:
@Simeon-Banov thanks for your first PR!

I think Product Bundles also achieve the same goals and is consistent across documents. I am not sure how to evaluate your PR. Do you have screenshots or animated GIF to share?

nabinhait:
I also believe this is not the ideal way to handle the sub-items, it will lead to confusion in Quotation Item table, if we don't maintain separate table for sub-items. And please explain why you not using Product Bundle for this. 

Closing the PR for now.

PR #11423, opened 2017-11-02
[Fix] Packing Slip creation if custom field does not have a value

Alchez:
**Issue:**

If you try to create a Packing Slip from a Delivery Note with custom fields, it breaks when trying to retrieve fields that don't actually exist in the database, such as those with fieldtype "Section Break".

### DIFF
--- a/erpnext/stock/doctype/packing_slip/packing_slip.py
+++ b/erpnext/stock/doctype/packing_slip/packing_slip.py
@@ -2,11 +2,13 @@
 # License: GNU General Public License v3. See license.txt
 
 from __future__ import unicode_literals
-import frappe
-from frappe.utils import flt, cint
-from frappe import _
 
+import frappe
+from frappe import _
+from frappe.model import no_value_fields
 from frappe.model.document import Document
+from frappe.utils import cint, flt
+
 
 class PackingSlip(Document):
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/packing_slip/packing_slip.py
```diff
@@ -84,11 +84,14 @@ def get_details_for_packing(self):
 			* No. of Cases of this packing slip
 		"""
 
-		# also pick custom fields from delivery note
 		rows = [d.item_code for d in self.get("items")]
 
-		custom_fields = ', '.join(['dni.`{0}`'.format(d.fieldname) for d in \
-			frappe.get_meta("Delivery Note Item").get_custom_fields()])
+		# also pick custom fields from delivery note
+		exclude_fieldtypes = ["Column Break", "Section Break"]
```
nabinhait: You can use `no_value_fields` defined in frappe/model/__init__.py

https://github.com/frappe/frappe/blob/develop/frappe/model/__init__.py#L10

nabinhait (changes requested):

Alchez:
@nabinhait, I've pushed a commit with the requested changes.

PR #11431, opened 2017-11-02
Render patient detail html using template

### LINKED ISSUE #11295: Medical record/Consultation is not translated.
I am testing Healthcare Domain. But some things are not translated, like the Medical Record. 
I think that the csv file is not taking the lines to be translated.

/apps/erpnext/erpnext/healthcare/page/medical_record/medical_record.js
/apps/erpnext/erpnext/healthcare/doctype/consultation/consultation.js 

some lines like
` if(data.email) details += "<br><b>Email :</b> " + data.email;`
`if(data.mobile) details += "<br><b>Mobile :</b> " + data.mobile;`
need to be fixed.

And when I click Heathcare Icon in desk, In the "Master" section the "Physician"(Only this) is not getting translated.

rkurungadam:
fix https://github.com/frappe/erpnext/issues/11295


erpnext/public/js/healthcare/patient_details.html
```diff
@@ -0,0 +1,60 @@
...
+    <div class="col-md-6" style="padding-bottom:20px;"><b>{{ __("Personal and Social History") }}</b>
+      {% if data.occupation %} <br> {{ __("Occupation") }}:   {{ data.occupation }} {% endif %}
+      {% if data.marital_status %} <br> {{ __("Marital Status") }}:   {{ data.marital_status }} {% endif %}
+      <br>
+    </div>
+  {% endif %}
+  {% if data.tobacco_past_use || data.tobacco_current_use || data.alcohol_past_use || data.alcohol_current_use ||
+    data.surrounding_factors || data.other_risk_factors %}
+    <div class="col-md-6" style="padding-bottom:20px;"><b>{{ __("Risk Factors") }}</b>
+      {% if data.tobacco_past_use %} <br> {{ __("Tobacco Past Use") }}:   {{ data.tobacco_past_use }} {% endif %}
```
netchampfaris: Shouldn't we use `for` loop here?

rkurungadam: and use frappe.meta.get_label? does this trigger a server side call?

netchampfaris: @rkurungadam No, you have to pass the labels in an array

rmehta: Ideally this should be configurable from the doctype

rkurungadam: how about rendering all document fields server side based on Section break into separate html div?

netchampfaris:
@rkurungadam Hey, can you add some screenshots? There is a lot of boilerplate code, and use of `cur_frm` which will be deprecated. Run your code through eslint, or setup your editor to use ESLint to find code style errors.

netchampfaris:
So I tried this on my local:
image

Is the **Patient Details** really required? There is already a link field, and the user can view Patient Information there.

rkurungadam:
@netchampfaris This is usablility issue, we couldn't yet convince a physician to traverse across documents while doing a consultation. Right now we are working on a transcription tool to record everything from a single page along with option to see past consultations and graph of vital signs. Will push this once its ready to review, so may be this is not needed in consultation. BTW is there a method to get document as HTML? this would help in similar scenarios.
Thanks and regards

Issue #11447, opened 2017-11-03
`get_batch_no` returning smallest batch instead of oldest batch


tundebabzy:
When selling items in retail or wholesale, it is the common and sensible practice to push the oldest stock off the shelves first. That way, you don't find yourself with expired stock or obsolete stock. `get_batch_no` on the other hand returns the batch with the smallest quantity that is able to fulfil the transaction. 

`get_batch_no` should actually return a batch number using FEFO (First Expiring First Out). This means it will decide which batch to return by considering options in this order:
- Batch is the oldest of the soonest to expire and is of sufficient quantity
- Batch is the oldest and is of sufficient quantity
- Batch is of sufficient quantity

tundebabzy:
#11454

Issue #11004, opened 2017-10-01
Auto fetch batches based on quantity on POS


mdwala:
In reference to
https://github.com/frappe/erpnext/issues/11003
and
https://discuss.erpnext.com/t/pos-change-selected-batch/28884
Batch selection should be automatic(optional may be) on “First Expiry First Out” based on required selling qty, I believe this is standard procedure for dealing in Batches for all businesses. Attached example snapshot shows require field to auto fetch batches. 
Thanks…
![4](https://user-images.githubusercontent.com/24455908/31053811-8717b84e-a6b6-11e7-9334-a5496d5a5251.png)

tundebabzy:
- [ ] Add new setting in POS Settings so that user can choose for ERPNext to automatically choose batch
- [ ] In Batch Selector Dialog, expired batches should not be included
- [ ] Automatically batch number which will be the oldest in stock as quantity is not known before hand

- Automatically selected batch number means user will only need to enter "Qty" in Batch Selector Dialog
- If the automatically selected batch is not sufficient, user will have to either change the batch manually or add another batch to satisfy the order. 
- The Batch Selector Dialog will always appear.

mdwala:
Hi. @tundebabzy 
If quantity is higher than single batch will auto batch selection be multiple batches?

tundebabzy:
@mdwala
No. User will have to adjust for the remaining manually. I chose this design because of programming simplicity as I think the latter situation will not be frequent enough to be a headache. However, take my thoughts with a pinch of salt. That's the reason why I put the design requirements here for others to see and critique before I start coding. If it is important that ERPNext should fetch remaining batches, then I can add that....

tundebabzy:
Just discovered that the POS cannot accept from multi batches at the moment. So present design will stand while my colleagues work on redesigning the POS.

tundebabzy:
POS will now automatically fetch batch number into the Batch Selector Dialog in POS. #11004

PR #11454, opened 2017-11-04
Default batch number selection

### LINKED ISSUE #10562: Default batch number selection
How can we use the FIFO method for batch number while doing the sales invoice? 

Current Scenario:
In case of Pharma/FMCG vertical, complete purchase and sales is dependent on batch and expiry date. Currently, ERPNext have the provision to setup batch number against each item while receiving the product. So from purchase perspective this is good. While doing sales invoice, we have to manually select the batch and add the quantity in it.

Expected Scenario:
While doing the sales invoice, item should automatically fetch the batches as per FIFO method. In case if the first available batch have less quantity it should pickup next available batch and so on. 

Can we do this?

tundebabzy:
Fixed #10562 #11447 

It is a standard and wise practice for those selling items with batches to use First Expiring First Out (FEFO) which is a variant of FIFO just that earliest item to expire is given priority. This avoids the seller having expired stock or obsolete items.

This PR changes ERPNext present behaviour when it comes to auto-fetching batch numbers. In sales invoice, the batch number is not fetched. In other places, it returns the smallest batch that can fulfil the order.

### Before
screenshot from 2017-11-04 22-08-35

### After
screenshot from 2017-11-04 22-01-51


nabinhait (changes requested):
`transaction.js` is inherited in both selling and buying cycle. But `set_batch_number` should only be called in selling cycle, more specifically in DN and sales invoice (only if update_stock is checked)

Also same functionality should be implemented for `stock_entry.js`, it does not inherit `transaction.js`.

tundebabzy:
`stock_entry.js` calls `erpnext.stock.get_item_details.get_batch_no` directly so it's already taken care of

nabinhait:
@tundebabzy I did not find the call. Can you send the link of github?

tundebabzy:
There is something wrong with the batch selector dialog. It's either the stock entry requirement is added in a separate PR or we fix the selector dialog first.

peek 2017-11-20 15-54

tundebabzy:
Dialog selector fixed in frappe/frappe#4517. Awaiting merge

tundebabzy:
@nabinhait Stock entry presents a batch selector dialog immediately after an Item with has_batch:1 is set. With this work flow, there's no clean way to add batch number automatically as the batch selector is shown before the user has a chance to specify quantity.

That said, I think we can push this and then work on auto fetching into the batch selector dialog in a fresh issue.

tundebabzy:
#11004 for auto fetching batches in batch selector

nabinhait (approved):

elvise:
@tundebabzy i'm really interested to batch based fifo do you have any planning about this point ?

PR #11480, opened 2017-11-07
[Website] Items list and stock

ghost:
Hi,

I add stock availability in items list (in stock / not in stock) and a parameter to set the number of products per page, it was hardcoded to 6, I keep this value by default.
Please look at screenshots below.

image

image


britlog:
I 've just added a new commit to substract reserved_qty for website stock.
Indeed, when an order is done, the website stock needs to decrease to avoid over-orders.
available_qty = actual_qty - reserved_qty 

Thanks

PR #11483, opened 2017-11-08
[New Feature] Delivery

creamdory:
From the discussion here https://discuss.erpnext.com/t/contribution-showcase-delivery-service-tour/29442/24 .

Our team has initiated the development of Delivery Feature addition in Selling Module to track delivery stops. It has links to Driver and Vehicle. As discussed, here are the basic functions:

- Create a new Delivery
- Select Date, Vehicle, Departure Time and Driver
- Add multiple Delivery Stops
- Notes for delivery driver or customer (what to pay attention to, what to pick up or deliver)
- Automatic Calculation of arrival time with Google Maps (rounded to full 15 minutes)
- Adding submitted delivery notes
- Sending email notification to customer

delivery


giuseppelamatrice:
Great!

rmehta:
"Delivery" will be mistaken for "Delivery Note".

Let us call this "Delivery Trip"

ekpott:
I agree with @rmehta on the Delivery naming

ekpott:
Let it be "Delivery Trip"

vishdha:
Hi @creamdory.  I checked your pull request few more thing to discuss
1. Show Distance and its UOM like 10 KM/ 10 Miles etc.

2. Currently its just show the distance from origin address to its subsequent destination address. 
for eg., city X to Y distance is 100 and Y to Z are 50 so Total Distance is 150 which is not optimized.
I think optimization using algorithm would be the best I looking for some resource to Optimize distance path.
for eg., City X to Z distance is 60  and city Z to Y distance is 50 so Total distance is 110 Optimized one.

creamdory:
Hey @vishdha thanks for testing this out and for your suggestion! Though this will be very helpful, this is a limitation for this PR. Once this is merge, your team or others can improve this one and integrate an algorithm to select the shortest route and time calculation. :D Really appreciate your suggestion. Thanks!

vladucoju:
Great work @creamdory. 
I will suggest to put routes in a future steps. Delivery manager could create delivery throw created routes. For example he will create couple routes. New route will have options to insert multiple city. 
Route1: city1, city2, city3
Route2: city4
Route3: city5, city6

When create new delivery hi will choose:
Root
Date of delivery (delivery note date)
Vehicle
Driver
After this is done in window will be all delivery notes in that route on that day. Manager will choose what items will be packed in that vehicle.

creamdory:
Hi @vladucoju ! Thank you for your suggestion! Though the current design is that Delivery trips are created after Delivery Note creation though it can handle multiple stops in a Trip. This makes sense since we prepare first Delivery Notes before going on a trip. Also, this is somehow detached yet from the Selling workflow in a since that this is still optional. Though once this is merged, we can improve upon this with added features. Thanks for your suggestions!

nabinhait:
@creamdory I have made some cleanups before merging, please check those.

creamdory:
Hi @nabinhait , Thanks!

PR #11489, opened 2017-11-08
Treeview fix for task

Zlash65:
task-tree-fix-2
Summary:-
1. Task series restored
2. Doctype rename disabled
3. Treeview visible on the basis of subject
4. Test case updated

Dependent on [#4449](https://github.com/frappe/frappe/pull/4449)


erpnext/projects/doctype/task/task_tree.js
```diff
@@ -45,9 +48,33 @@ frappe.treeview_settings['Task'] = {
...
+					}
+				});
+				d.show();
+			}
+		},
+		{
+			label:__("Rename"),
+			click: function(node) {
+				var d = new frappe.ui.Dialog({
+					'title': __("Rename Subject"),
```
nabinhait: We can simply remove the Rename option from tree view. User can change the subject using Edit button. "Rename" button is confusing, as it is actually not renaming.

nabinhait (changes requested):

erpnext/projects/doctype/task/task_tree.js
```diff
@@ -1,5 +1,5 @@
 frappe.provide("frappe.treeview_settings");
-
+var glob;
```
nabinhait: Don't use global variable

erpnext/projects/doctype/task/task_tree.js
```diff
@@ -25,8 +25,10 @@ frappe.treeview_settings['Task'] = {
 		return node.data.value;
 	},
 	onload: function(me){
+		glob = me;
 		me.make_tree();
 		me.set_root = true;
+		me.allow_rename = false;
```
nabinhait: not required, it should be visible based on permission

nabinhait (changes requested):

PR #11492, opened 2017-11-08
[minor][fix] condition check

saurabh6790:


nabinhait:
fixed by https://github.com/frappe/erpnext/pull/11501

PR #11510, opened 2017-11-10
Add item directly in the cart only if serial no, batch, barcode has enter in the search field

rohitwaghchaure:
**After Fix**
pos_search_issue


noetico:
Brilliant Boss! Thanks!

Issue #11518, opened 2017-11-10
Default Tax template auto-updating on Save


umairsy:
WN-SUP28931

On creating new sales transaction, default tax template is fetched automatically. However, even if the user removes the default template and saves the transaction, the default template is re-updated on save. 

![tax](https://user-images.githubusercontent.com/910238/32658175-c75c9af8-c63e-11e7-82d4-b235a731d3e6.gif)

Default template should be applied to new transaction, but not on save and submit,

PR #11540, opened 2017-11-13
[Fix] Default taxes appeared on save even if removed

shreyashah115:
#11518 

anim


nabinhait:
Closing this, fixed by https://github.com/frappe/erpnext/pull/11544

PR #11542, opened 2017-11-13
Post GL Entry based on rounded total in Sales/Purchase Invoice

nabinhait:


sathishpy:
Great. Looks very different from my initial work. Accounting is certainly a complex module and needs a lot of experience to make modifications.

nabinhait:
@sathishpy Have you test it?

sathishpy:
Can you tell me whether it is ok to 'git pull' from https://github.com/nabinhait/erpnext/tree/develop.
It says 'This branch is 1208 commits behind frappe:develop.'. Not sure whether there is a better way to update the code from merge requests.

sathishpy:
Great Job. I just updated to latest release and did some basic testing. It works as expected. It might be better to update the fields dynamically when 'disable rounded total' is selected. However people are not going to use it often, so I don't really care about it much.

PR #11557, opened 2017-11-14
Allow print before pay in POS

harshmehta16:
In Pos System currently one can print invoice only once the payment is done. but we have following two use cases for POS system.

Case 1: For super markets where the payment is recieved first and then the print of invoice is generated and given to customer. In this case the system works perfect.

Case 2: For Restaurants where the invoice print is to be generated first and given to client then customer decides to pay via cash or card. In this case the current POS system is not usefull. As there is no Print option untill you submit the payment.

So We have added option "print before pay" in Pos profile and if user have checked that option in pos profile then user will be able to print Bill without pay.

Thanks


erpnext/accounts/doctype/pos_profile/pos_profile.json
```diff
@@ -3,7 +3,7 @@
  "allow_guest_to_view": 0, 
  "allow_import": 0, 
  "allow_rename": 0, 
- "autoname": "field:pos_profile_name",
```
rohitwaghchaure: Why removed?

erpnext/accounts/doctype/pos_profile/pos_profile.json
```diff
@@ -1531,7 +1375,7 @@
  "show_name_in_global_search": 0, 
  "sort_field": "modified", 
  "sort_order": "DESC", 
- "title_field": "pos_profile_name",
```
rohitwaghchaure: Removed?

erpnext/accounts/page/pos/pos.js
```diff
@@ -1614,17 +1648,18 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 
 	create_invoice: function () {
 		var me = this;
-		var invoice_data = {};
+		var invoice_data = {}
 		this.si_docs = this.get_doc_from_localstorage();
-		if (this.frm.doc.offline_pos_name) {
```
rohitwaghchaure: removed?

rohitwaghchaure:
@harshmehta16 
You have removed lot of latest code in this commit, I am closing the PR, kindly clean you PR and reopen it

Issue #11571, opened 2017-11-14
POS - On submission, show Dialog to Print invoice or create New Invoice


umairsy:
On submission of POS Invoice, show the dialog to Print invoice or create New POS Invoice. It was available in the earlier version.

PR #11575, opened 2017-11-14
[fix] Added primary action in the modal which will show on the submission in the pos

rohitwaghchaure:
**Added Modal**
screen shot 2017-11-14 at 6 43 33 pm

Fixed

https://github.com/frappe/erpnext/issues/11571


nabinhait:
@rmehta What do you think about this? Is the modal needed, as we already have **Print** and **New** button as a part of standard toolbar?

tundebabzy:
This is another place where frappe/frappe#4203 would make sense

PR #11596, opened 2017-11-15
[Minor] Ability to select Operation in Production Order Required Items

KanchanChauhan:
Currently in Production there is no way to know which required item (raw material) is needed for which operation.
Added Operation field to Production Order - Required Items which fetches the Operations only from that Production Order BOM.

screen shot 2017-11-15 at 5 14 59 pm


KanchanChauhan:
Closing this one, routing is a better way to do it. Will work on that.

PR #11608, opened 2017-11-16
Multiple variant creation dialog

pratu16x7:
Depends on: https://github.com/frappe/frappe/pull/4503

Creates variants as per all combinations of selected attributes. Pushes to background, does not create if exists.

multivariantdialog

After above gif, created 6 additional ones apart fro 'L' variant:
screen shot 2017-11-16 at 5 52 52 am


joezsweet:
Would be nice to be able to choose attribute from group of attributes, I.e. for size you can have UK, US, FR, IT.
If US: S,M,L,XL attribute will be listed
If IT: 40, 42, 44, 46, 48, 50 etc

elvise:
@joezsweet +1

PR #11610, opened 2017-11-16
[enhance] by pass credit limit

ashish-greycube:
All comments of @nabinhait  are incorporated


erpnext/selling/doctype/sales_order/tests/test_sales_order_with_bypass_credit_limit_check.js
```diff
@@ -0,0 +1,54 @@
...
+		// save form
+		() => cur_frm.save(),
+		() => frappe.timeout(1),		
+
+		() => {
+			return frappe.tests.make('Sales Order', [
+				{customer: 'Test Customer 5'},
+				{items: [
+					[
+						{'delivery_date': frappe.datetime.add_days(frappe.defaults.get_default("year_end_date"), 1)},
```
rmehta: can you write these tests on the server side too?

ashish-greycube: Rushabh, I did give a try for server side i.e. test_sales_order.py but the issue is it has make_sales_order function which doesn't use submit logic of sales_order.py. Server side does submit via frappe.new_doc and submit directly to DB. 
sales_order.py --> bypass_credit_limit_check_at_sales_order logic is in check_credit_limit function called before click of submit button.
Hence i have written the test case using QUnit. thanks

rmehta:
@ashish-greycube please write better title for your PRs

ashish-greycube:
Hi @nabinhait , please advise if there is anything pending on my part for PR! Thanks! for all the hand holding, I hope i will be able to do better next time.

nabinhait:
@ashish-greycube 
Made some cleanups before merging the pull request (https://github.com/frappe/erpnext/commit/68768252ce6d5e52f022d29804855b189213bbcb), please check once.

ashish-greycube:
@nabinhait ,
Your logic will fail when say "by pass CL" flag is true and partial advance is paid against that sales order. Report will show incorrect Credit Limit. Excel file is attached explaining it with a case.
image

[advance.xlsx](https://github.com/frappe/erpnext/files/1501406/advance.xlsx)

ashish-greycube:
@nabinhait , following is cosmetic change, i guess it would make more sense to end user if credit limit and its checkbox are group together and a col separator is kept for terms template.
image

turner-james:
@nabinhait when will this feature be available on the hosted version?

nabinhait:
@ashish-greycube 

Sorry for the delayed response.

So, you are talking that in "Customer Credit Balance" report, if there is any advance against a Sales Order, that Sales Order amount should be considered for calculating outstanding amount, even if "Bypass credit limit at SO" is checked. So, we are bypassing Sales Orders conditionally, only if there are no advance against it. Am I correct?
I am just thinking, why there should be 2 different logic in 2 places for calculating outstanding?

jigneshpshah:
@nabinhait 

The whole use case is , As explained by @turner-james is that he want the system to allow him to book sales order for a customer and should prohibit him only at the time of delivery/invoicing, as the risk of the credit limit given to the customer takes place when he delivers the goods, There is no risk for the credit limit when he takes Sales Order, the product produced or packed via Sales Order can be used to supply other customers if the said customer differs.

Credit Limit check is bypassed at SO based on only one condition i.e. if it is checked in the filed as below

image


Credit Balance amount  should reflect considering the net outstanding.

Case 1) For Customer if the credit limit check is bypassed at SO,
After Booking the Sales Order , an Advance is paid by the customer against the Sales Order

Credit Limit = 1000

Credit Balance = 1000
Sales Order = 1200
Outstanding Amount = 1200

Credit Balance = 1000 (for the customers who have credit limit check bypassed as SO, for the rest the above scenario wont happen as system will block the user from booking the Sales Order itself)

Now, 
Advance Paid Against Sales Order = 400
Outstanding Amount = 1200-400 = 800
New Credit Balance  =1000-800 = 200.

The same should reflect in the Customer Credit Balance Report.

the [excel image](https://user-images.githubusercontent.com/29812965/33207485-e9c60cbe-d133-11e7-8212-dedf7c375a91.png) reflects the said logic.

if you can explain your concern through data , it will help a lot.

jigneshpshah:
ERPNext: v10.0.17 (master)
Frappe Framework: v10.0.16 (master)

Current error in the logic is explained  through gif below 
credit limit checkbypass so nabin

turner-james:
My other concern is that currently, we have excluded Sales Orders from the Credit Balance report UNLESS there is an advance against the sales order. I understand and agree with that. HOWEVER, if Sales Orders are excluded from Outstanding Balance and Credit Balance on the Credit Balance Report, then DELIVERY NOTES must be included on the Credit Balance Report calculations. AT the moment it appears that Delivery Notes are excluded from the calculation of Credit Balances. Surely that is not right? Please advise.

PR #11614, opened 2017-11-16
[fix] Valuation rate in stock entry and code cleanup

rohitwaghchaure:


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -104,20 +104,146 @@ frappe.ui.form.on('Stock Entry', {
...
+		const args = {
+			'item_code'			: item.item_code,
+			'posting_date'		: frm.doc.posting_date,
+			'posting_time'		: frm.doc.posting_time,
+			'warehouse'			: cstr(item.s_warehouse) || cstr(item.t_warehouse),
+			'transfer_qty'		: item.transfer_qty,
+			'serial_no	'		: item.serial_no,
+			'bom_no'			: item.bom_no,
+			'expense_account'	: item.expense_account,
+			'cost_center'		: item.cost_center,
```
nabinhait: Are all those args required? I think expense_account, cost_center, bom_no, transfer_qty no needed.

nabinhait (changes requested):

PR #11616, opened 2017-11-16
[fix] do not translate domains while creating db record

saurabh6790:
- Import non-translated domain while installing fixtures
 - Patch to rename domain in English and to reset domain settings for activated domain
 - Moved **resave_domain_settings** and **revert_domain_settings** patch from frappe to erpnext. 
     - The domains specified in both patches only belongs to ERPNext. If ERPNext is not installed then these patches will break as they are not found the domain records.
     -  Handled patch re-execution by checking  into Patch Log table 
         ```if frappe.db.get_value("Patch Log", {'patch': 'frappe.patches.v9_1.resave_domain_settings'})```
   

Dependent PR: https://github.com/frappe/frappe/pull/4506

### DIFF
--- a/erpnext/patches/v9_2/rename_translated_domains_in_en.py
+++ b/erpnext/patches/v9_2/rename_translated_domains_in_en.py
@@ -1,0 +1,26 @@
+import frappe
+from frappe import _
+
+def execute():
+	language = frappe.get_single("System Settings").language
+
+	if language.startswith('en'): return
+
+	all_domains = frappe.get_hooks("domains")
+
+	for domain in all_domains:
+		translated_domain = _(domain, lang=language)
+		if frappe.db.exists("Domain", translated_domain):
+			frappe.rename_doc("Domain", translated_domain, domain, ignore_permissions=True, merge=True)
+
+	domain_settings = frappe.get_single("Domain Settings")
+	active_domains = [d.domain for d in domain_settings.active_domains]
+	
+	for domain in active_domains:
+		domain = frappe.get_doc("Domain", domain)
+		domain.setup_domain()
+
+		if int(frappe.db.get_single_value('System Settings', 'setup_complete')):
+			domain.setup_sidebar_items()
+			domain.setup_desktop_icons()
+			domain.set_default_portal_role()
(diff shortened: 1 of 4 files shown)


nabinhait (changes requested):
If you move a patch from frappe to erpnext, it will run again. Lets keep those patches in frappe.

saurabh6790:
@nabinhait they will only run if there is no entry for frappe's patches in Patch Log

```
if frappe.db.get_value("Patch Log", {'patch': 'frappe.patches.v9_1.revert_domain_settings'}):
```
---
```
if frappe.db.get_value("Patch Log", {'patch': 'frappe.patches.v9_1.resave_domain_settings'}):
```


```
if frappe.db.get_value("Patch Log", {'patch': 'frappe.patches.v9_1.resave_domain_settings'}):
```

erpnext/patches/v9_2/rename_translated_domains_in_en.py
```diff
@@ -0,0 +1,24 @@
+import frappe
+from frappe import _
+
+def execute():
+	all_domains = frappe.get_hooks("domains")
+	language = frappe.get_single("System Settings").language
+
```
rmehta:
```
if language.startswith('en'): return
```

erpnext/patches/v9_2/rename_translated_domains_in_en.py
```diff
@@ -0,0 +1,24 @@
...
+	for domain in all_domains:
+		translated_domain = _(domain, lang=language)
+		if frappe.db.exists("Domain", translated_domain):
+			frappe.rename_doc("Domain", translated_domain, domain, ignore_permissions=True, merge=True)
+
+	domain_settings = frappe.get_single("Domain Settings")
+	active_domains = [d.domain for d in domain_settings.active_domains]
+	
+	for domain in active_domains:
+		if domain in ('Education', 'Healthcare', 'Hospitality'):
```
rmehta: setup all active domains?

rmehta (changes requested):

PR #11621, opened 2017-11-16
Employee treeview

Zlash65:
employee_tree
Summary:-
1. Employee DocType -> added a check field 'has_subordinates' as a means to know if an Employee has any people working under him i.e has any other employee selected the Employee in its reports_to.
2. employee_tree.js -> Filter based on company. Toolbar overridden to show only edit option. New Employee creation option added in Menu.
3. employee.js -> To handle scenario in which reports_to which is set as some employee is then set as blank or some different employee - 'has_subordinates' checkbox should be appropriately toggled for new as well as old employee that was set in the field.
4. employee.py -> get_children() method added. Used reports_to for parent relation. Used hidden checkbox 'has_subordinates' for expandable nodes. Added validation to avoid cyclic settings of reports_to. Added a set method to enable/disable checkbox based on if its being linked to any employee or not.


joezsweet:
What about using Nestes set?

Zlash65:
@joezsweet Thought about that. But then again, its a simple view for Organization's Employee hierarchy and felt like NestedSet calculations were not really needed or so.

joezsweet:
@Zlash65 well, would be nice to have all trees views using Nestedset, thought

nabinhait:
I agree with @joezsweet , as all tree views are using nestedset model, why not for it?

erpnext/hr/doctype/employee/employee.json
```diff
@@ -2418,6 +2418,36 @@
...
    "set_only_once": 0, 
    "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "has_subordinates",
```
rmehta: no need for has subordinates. You should be able to add Employees against any other employee

rmehta (changes requested):

PR #11624, opened 2017-11-17
[New Feature] Sample Retention from a Batch of item received

KanchanChauhan:
Most of the manufacturing firm likes to store some of the raw material from the batch as a sample for future reference or quality check or some reasons unknown to me.

It will possible in ERPNext now. To do so:

**Set Sample Retention Warehouse in Stock Settings** (preferably new warehouse just for samples)

sample-warehouse

**Enable Retain Sample in Item master**

retain-sample

**Stock Entry**

Whenever a Stock Entry is created with the purpose as Material Receipt, for items which have Retain Sample enabled, the Sample Quantity can be set during that Stock Entry. Sample quantity cannot be more than the Maximum sample quantity set in Item Master.

material-receipt-sample

On submission of this Stock Entry, button 'Make Retention Stock Entry' will be available to make another Stock Entry for the transfer of sample items from the mentioned batch to the retention warehouse set in Stock Settings. On clicking this button it will direct you to new Stock Entry with all the information, verify the information and click Submit.

material-transfer-sample

- [x] Documentation
- [x] Test cases


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -803,6 +805,45 @@ def validate_batch(self):
 						if getdate(self.posting_date) > getdate(expiry_date):
 							frappe.throw(_("Batch {0} of Item {1} has expired.").format(item.batch_no, item.item_code))
 
+	def move_sample_to_retention_warehouse(self):
+		if self.purpose == "Material Receipt":
+			for item in self.get("items"):
+				if item.sample_quantity and item.batch_no:
+					retention_warehouse = frappe.db.get_single_value('Stock Settings', 'sample_retention_warehouse')
```
nabinhait: This should be executed outside for loop

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -803,6 +805,45 @@ def validate_batch(self):
...
+def create_sample_material_transfer(company, item, retention_warehouse, sample_serial_nos):
+	stock_entry = frappe.new_doc("Stock Entry")
+	stock_entry.company = company
+	stock_entry.purpose = "Material Transfer"
+	stock_entry.append("items", {
+		"item_code": item.item_code,
+		"s_warehouse": item.t_warehouse,
+		"t_warehouse": retention_warehouse,
+		"qty": item.sample_quantity,
+		"basic_rate": item.basic_rate,
```
nabinhait: Should be `"basic_rate": item.valuation_rate,`

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -803,6 +805,45 @@ def validate_batch(self):
...
+		"conversion_factor": 1.0,
+		"serial_no": sample_serial_nos,
+		'batch_no': item.batch_no
+	})
+	stock_entry.insert()
+	stock_entry.submit()
```
nabinhait: Creating and submitting stock_entry through code is never a good idea. Because, if users added some mandatory custom fields in Stock Entry, it will fail. And the user will never be able to submit the stock entry.

What we can do, instead of creating stock entry automatically, we can add a button "Make Retention Stock Entry". On click of button, it will prepare a new stock entry for all the items combined and load it.

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -186,9 +186,33 @@ frappe.ui.form.on('Stock Entry Detail', {
...
+var validate_sample_quantity = function(frm, cdt, cdn) {
+	var d = locals[cdt][cdn];
+	if (d.sample_quantity && frm.doc.purpose == "Material Receipt") {
+		frappe.call({
+			method: 'erpnext.stock.doctype.stock_entry.stock_entry.validate_sample_quantity',
+			args: {
+				batch_no: d.batch_no,
+				item_code: d.item_code,
+				sample_quantity: d.sample_quantity,
+				qty: d.qty
```
nabinhait: Should be `d.transfer_qty` which is in Stock UOM.

nabinhait (changes requested):

nabinhait:
@KanchanChauhan Please reopen after fixing requested changes.

KanchanChauhan:
@nabinhait Reopened after making all the changes you asked for. Now it will just redirect to the Stock Entry for sample retention and user can submit it after verifying the details.

erpnext/patches.txt
```diff
@@ -461,7 +461,7 @@ execute:frappe.delete_doc_if_exists("DocType", "Program Fee")
 erpnext.patches.v9_0.set_pos_profile_name
 erpnext.patches.v9_0.remove_non_existing_warehouse_from_stock_settings
 execute:frappe.delete_doc_if_exists("DocType", "Program Fee")
-erpnext.patches.v8_10.update_gl_due_date_for_pi_and_si
```
nabinhait: Why removed it?

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -86,6 +86,12 @@ frappe.ui.form.on('Stock Entry', {
 		if(frm.doc.company) {
 			frm.trigger("toggle_display_account_head");
 		}
+
+		if(frm.doc.docstatus==1 && frm.doc.purpose == "Material Receipt") {
+			frm.add_custom_button(__('Make Retention Stock Entry'), function () {
```
nabinhait: Should "Make Retention Stock Entry" be available in Purchase Receipt as well?

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.js
KanchanChauhan: Yup makes sense, added it to Purchase Receipt as well.

erpnext/patches.txt
KanchanChauhan: Sorry about that.

Issue #11631, opened 2017-11-17
Syntax Error: while bench get-app erpnext https://github.com/frappe/erpnext !!


asharamseervi:
I'm trying to setup new bench, and getting new erpnext into that bench. While using command `bench get-app erpnext https://github.com/frappe/erpnext` gives some Syntax Errors on Terminal, as follows:

```
root@ip-xx-xx-xx-xx:/home/frappe/frappe-bench/new-bench# bench get-a                                                                                        pp erpnext https://github.com/frappe/erpnext
INFO:bench.app:getting app erpnext
INFO:bench.utils:git clone https://github.com/frappe/erpnext  --depth 1 --origin                                                                                         upstream
Cloning into 'erpnext'...
remote: Counting objects: 5833, done.
remote: Compressing objects: 100% (5193/5193), done.
remote: Total 5833 (delta 1040), reused 3080 (delta 489), pack-reused 0
Receiving objects: 100% (5833/5833), 172.18 MiB | 1.27 MiB/s, done.
Resolving deltas: 100% (1040/1040), done.
Checking connectivity... done.
Checking out files: 100% (5813/5813), done.
('installing', u'erpnext')
INFO:bench.app:installing erpnext
INFO:bench.utils:./env/bin/pip install -q  -e ./apps/erpnext --no-cache-dir
Wrote css/frappe-web.css - 65.11 KB
Wrote js/frappe-web.min.js - 132.05 KB
Wrote js/control.min.js - 72.03 KB
Wrote js/dialog.min.js - 116.34 KB
Wrote css/desk.min.css - 309.4 KB
Wrote css/frappe-rtl.css - 32.49 KB
Wrote js/libs.min.js - 1.13 MB
Wrote js/desk.min.js - 459.7 KB
Wrote css/module.min.css - 2.08 KB
Wrote css/form.min.css - 4.47 KB
Wrote js/form.min.js - 195.7 KB
Wrote css/list.min.css - 15.04 KB
Wrote js/list.min.js - 149.61 KB
Wrote css/report.min.css - 7.89 KB
Wrote js/report.min.js - 259.97 KB
Wrote js/web_form.min.js - 247.55 KB
Wrote css/web_form.css - 24.42 KB
Wrote js/print_format_v3.min.js - 23.39 KB
Wrote css/erpnext.css - 8 KB
Wrote js/erpnext-web.min.js - 3.79 KB
[...]

PR #11633, opened 2017-11-17
[Minor][Bug] Fix syntax error

KanchanChauhan:


asharamseervi:
I also reported the same issue to Github: https://github.com/frappe/erpnext/issues/11631, which was resolved in this commit. Please close the issue.

So, basically many issues are still pending because they not noticed and some known contributor feels trouble there, they pull it and fix it. Hence, the issue is still pending. Kindly have some policy for this.

rmehta:
Please raise issues and tag them as "bug"

PR #11667, opened 2017-11-20
Employee tree

Zlash65:
employee-nested
Without nestedset - [#11621 ](https://github.com/frappe/erpnext/pull/11621)


Olawale1:
Hi,

Good job. Hope all branches will be expanded by default. Would be odd to open an Employee tree only to find the names of top management and then having to click each name to view subordinates 

Kind regards,

Zlash65:
@Olawale1 No it does not by default as of now but it could be done, I'll ask around and see what majority has to say and will implement it 👍

rohitwaghchaure:
@Zlash65 
Patch is required to update lft and rgt of existing employee

erpnext/hr/doctype/employee/employee.py
```diff
@@ -300,4 +302,26 @@ def get_employee_emails(employee_list):
...
\ No newline at end of file
+	return employee_emails
+
+@frappe.whitelist()
+def get_children(doctype, parent=None, company=None, is_root=False, is_tree=False):
+	condition = ''
+
+	if is_root:
+		parent = ""
+	if parent and company and parent!=company:
```
rohitwaghchaure: Why checked company with parent?

erpnext/hr/doctype/employee/employee.py
```diff
@@ -52,7 +53,11 @@ def validate(self):
 				frappe.permissions.remove_user_permission(
 					"Employee", self.name, existing_user_id)
 
+	def update_nsm_model(self):
+		frappe.utils.nestedset.update_nsm(self)
+
 	def on_update(self):
+		self.update_nsm_model()
```
rohitwaghchaure: Add in on_trash too!

rohitwaghchaure:
Closing PR, kindly make changes and reopen it

erpnext/hr/doctype/employee/employee.py
```diff
@@ -154,13 +159,13 @@ def validate_reports_to(self):
 			throw(_("Employee cannot report to himself."))
 
 	def on_trash(self):
+		self.update_nsm_model()
```
nabinhait: Instead of calling `update_nsm_model`, call super function (on_trash of NestedSet class)

nabinhait (changes requested):

nabinhait (changes requested):
- Renaming of employee breaks
- Merging also breaks
The error comes from `before_rename` and `after_rename` function of nestedset.py

PR #11672, opened 2017-11-21
[feature] Shipping Rule Tax calculated based on Total Net Weight

vishdha:
In Shipping Rule select calculate_based_on: Net Weight and Select condition from value to value.
In Sales Invoice Item Weight, total_weight and weight_uom field are added. 
based on qty weight are calculated for eg. weight=20 and qty = 5, total_weight = 100
In Sales invoice on select shipping rule, it will calculate tax based on total_net_weight.
weight


erpnext/accounts/doctype/sales_invoice_item/sales_invoice_item.json
```diff
@@ -405,6 +405,97 @@
    "set_only_once": 0, 
    "unique": 0
   }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "weight",
```
nabinhait: `weight_per_unit` will be more clearer

nabinhait (changes requested):

vishdha:
anim

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -343,6 +343,9 @@ cur_frm.cscript.select_print_heading = function(doc,cdt,cdn){
 		cur_frm.pformat.print_heading = __("Purchase Invoice");
 }
 
+cur_frm.add_fetch('item_code','weight_per_unit','weight_per_unit');
+cur_frm.add_fetch('item_code','weight_uom','weight_uom');
```
rohitwaghchaure: Wrote same code in every file. Add this code into transaction.js and check child table(sales order item) has field weight_per_unit

erpnext/controllers/taxes_and_totals.py
```diff
@@ -70,6 +70,7 @@ def calculate_item_values(self):
 				item.net_rate = item.rate
 				item.amount = flt(item.rate * item.qty,	item.precision("amount"))
 				item.net_amount = item.amount
+				item.total_weight = flt(item.weight_per_unit * item.qty)
```
rohitwaghchaure: Total weight will calculate only if there is taxes? I think this has to be done on qty change trigger?

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -92,6 +92,7 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 				item.amount = flt(item.rate * item.qty, precision("amount", item));
 				item.net_amount = item.amount;
 				item.item_tax_amount = 0.0;
+				item.total_weight = flt(item.weight_per_unit * item.qty)
```
rohitwaghchaure: this has to be done on qty change trigger?

rohitwaghchaure:
Duplicate https://github.com/frappe/erpnext/pull/11770

Issue #11163, opened 2017-10-11
Add slide for creating letter head in User Progress


umairsy:
- [ ] Remove brand in slides
- [x] Add letter head in user progress

PR #11679, opened 2017-11-21
Letterhead in user progress

shreyashah115:
#11163 
letterheadd

### DIFF
--- a/erpnext/utilities/user_progress.py
+++ b/erpnext/utilities/user_progress.py
@@ -81,6 +81,26 @@ def get_slide_settings():
 				}
 			]
 		),
+
+		frappe._dict(
+			action_name='Add Letterhead',
+			domains=('Manufacturing', 'Services', 'Retail', 'Distribution', 'Education'),
+			title=_("Add Letterhead"),
+			help=_("Upload your letter head. (you can edit it later)."),
+			fields=[
+				{"fieldtype":"Attach Image", "fieldname":"letterhead",
+					"label":_("Letterhead"),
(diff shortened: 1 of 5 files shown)


erpnext/utilities/user_progress.py
```diff
@@ -81,6 +81,26 @@ def get_slide_settings():
...
+
+		frappe._dict(
+			action_name='Add Letterhead',
+			domains=('Manufacturing', 'Services', 'Retail', 'Distribution', 'Education'),
+			title=_("Add Letterhead"),
+			help=_("Upload your letter head. (you can edit it later)."),
+			fields=[
+				{"fieldtype":"Attach Image", "fieldname":"letterhead",
+					"label":_("Letterhead"),
+					"description": _("Keep it web friendly 900px (w) by 100px (h)"),
```
pratu16x7: Let's just skip the label and description altogether. Will also help in keeping the image control in the centre. Instead, we'll change help to "Upload your letter head (Keep it web friendly as 900px by 100px)". Once inside, they'll already be aware that things can be changed later on. (The edit line was while in the setup wizard).

pratu16x7 (changes requested):

pratu16x7:
Also update `install_fixtures.py`.

rmehta:
Also update Setup Wizard test!

pratu16x7:
@rmehta Agree, if we are also removing the brand info in this PR (right now there are no changes in Setup Wizard)

nabinhait:
@shreyashah115 Merged this PR. Remove brand slide from setup-wizard via a separate PR.

PR #11685, opened 2017-11-22
 Updated images in help pages for fleet management and print format

Arundhatii:


erpnext/docs/user/manual/en/schools/Attendance/index.txt
```diff
@@ -1,3 +1,8 @@
+21Nov
+Student_Attendance
```
rmehta: files not added. Also please use lower case for filenames

rmehta (changes requested):

Issue #10786, opened 2017-09-14
Permission - Unable to view reports based on company


addymuliady:
ERPNext: v8.11.2 (master)
Frappe Framework: v8.10.5 (master)

Unable to view reports due to permission issue

**Setup**
![image](https://user-images.githubusercontent.com/21187126/30423680-384ea094-9976-11e7-8992-791a23facd5c.png)

**Error**
![image](https://user-images.githubusercontent.com/21187126/30423739-66b863c0-9976-11e7-9d12-5d4773a1e936.png)

**Finding**
![image](https://user-images.githubusercontent.com/21187126/30423609-f4f7d432-9975-11e7-8371-1026abad05ea.png)

**Report Role**
![image](https://user-images.githubusercontent.com/21187126/30423881-d1d06072-9976-11e7-9af1-583f7ba6c47f.png)

**User Role**
![image](https://user-images.githubusercontent.com/21187126/30423970-109bbffe-9977-11e7-8022-e4a13605bd44.png)

umairsy:
Is this issue occurring in your hosted ERPNext account? If yes, please report this issue on support@erpnext.com. Also, share the login credentials of the user facing this issue. Since this issue is related to specific configuration in your account, we will need to access your account remotely locate the cause of the issue.

If it is self-hosted (with no support subscription), request you to post it on ERPNext community forum.

umairsy:
Please also ensure that in the Company master, you have checked Apply User Permission, with the condition on the Company field. After this, your permissions should work fine.

![screen shot 2017-09-21 at 12 46 51 pm](https://user-images.githubusercontent.com/910238/30683425-0d0f212e-9ecb-11e7-9c8c-1f567e0b42e1.png)

addymuliady:
Hi @umairsy 

Have posted this on forum 10 days ago. And with no respond, I posted this into GitHub.
https://discuss.erpnext.com/t/permission-issue-unable-to-view-company-in-report-filter/28309

I did as required.

Added a Company field in the Company Master too.

![image](https://user-images.githubusercontent.com/21187126/30796847-5a1742ec-a206-11e7-9188-86bda633c0ee.png)

PR #11689, opened 2017-11-22
filter reports based on Company with User Permission Applied

ashish-greycube:
In context to [defect](https://erpnext.org/bounties/solve-issue---unable-to-filter-reports-based-on-company-with-user-permission-applied)

Defect : At present company dropdown(ex. Trial balance report) , the default company is shown as initially created company for all user irrespective of user has permission to that company.

ERPNext, boot.py is corrected to check value from 'tabUser Permission'. Note if user has permission to more than 1 company then the first created company is set as default. The fix works for all the report/form where company dropdown is present.

After fix
1) Creat an account user and out of say 3 company assign it access to one of the company using User Permission. Note, "Apply for all Roles for this User" has to be checked
2) Login with newly created user and go to Trial balance report
3)Note the company dropdown defaults to the company to which it has access to i.e 1 company. Also options donot show name of other 2 companies.

1)Login with admin user and check the default company in trial balance report. It should be the initially created company

### DIFF
--- a/erpnext/startup/boot.py
+++ b/erpnext/startup/boot.py
@@ -19,7 +19,7 @@ def boot_session(bootinfo):
 		res = get_user_default_from_user_permission("company",frappe.session['user'])
 		if res==None:
 			res = get_user_default_from_default_value("company")
-			bootinfo.sysdefaults.company = res			
+			bootinfo.sysdefaults.company = res
 		else:
 			bootinfo.user.defaults.company = res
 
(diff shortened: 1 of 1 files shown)


addymuliady:
https://github.com/frappe/erpnext/issues/10786

addymuliady:
@umairsy kindly take a look at this ... thanks!

nabinhait:
This should be a generic fix for all the User Permissions, not only for company. Closing it for now.

ashish-greycube:
Both the functions to get default value are generic
-- get_user_default_from_user_permission(key, user=None )
-- get_user_default_from_default_value(key)

We are just passing company as key as that is the key for which we want default.
There are many other key for which default can be set by looping such set.

1)In future it can be easily extended
2)If you have such set then we can loop it right away. Thanks

PR #11693, opened 2017-11-22
Default Payment Mode on POS when on Draft

Helkyd:
Default Payment Mode set when retrieving POS on Draft


Helkyd:
@nabinhait 
Wonder if those are valid changes to implement....
Default payment mode selected when POS was drafted and allow no ZERO payment to be processed in order to avoid user forgetting the payment type done by the customer.

erpnext/accounts/page/pos/pos.js
```diff
@@ -1639,6 +1639,14 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		this.si_docs = this.get_doc_from_localstorage();
 		if (this.name) {
 			this.update_invoice()
+			//to retrieve and set the default payment
+			invoice_data[this.name] = this.frm.doc;
+			invoice_data[this.name].payments[0].amount = this.frm.doc.net_total
+			invoice_data[this.name].payments[0].base_amount = this.frm.doc.net_total
```
rohitwaghchaure: This is already implemented
https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/taxes_and_totals.js#L614-L617

erpnext/public/js/payment/payments.js
```diff
@@ -27,8 +27,14 @@ erpnext.payments = erpnext.stock.StockController.extend({
...
 	
 		this.dialog.set_primary_action(__("Submit"), function() {
-			me.dialog.hide()
-			me.submit_invoice()
+			//Allow no ZERO payment
+			$.each(me.frm.doc.payments, function (index, data) {
+				if (data.amount != 0) {
+					me.dialog.hide();
+					me.submit_invoice();
+					return;
```
rohitwaghchaure: This should be configurable, what if user wants to submit an invoice with partial payment?

rohitwaghchaure:
Closing PR, kindly make the changes and reopen

erpnext/public/js/payment/payments.js
Helkyd: Assuming that he is paying on POS partial payments on this don't think should be !!! But still a partial payment would be different than Zero ... Or can we make it configurable so that is up to user to accept Zero Payments when on POS or not ?

Issue #11226, opened 2017-10-16
[refactor] Party Type


Olawale1:
- [ ] Extend Dynamic Link to accept Party Type as the first link (this can be done via hooks)
- [ ] Validate "Party Type" must be "DocType" (party type cannot be created by user)
- [ ] Change Dynamic Links (DocType -> Party Type)
- [ ] Remove un-necessary get_query type filters
- [ ] "Receivable or Payable"
- [ ] Check workflow in Payment Entry

This needs to be replaced by "Party Type" doctype

----
Hi,

Steps to replicate:

1. Open Journal Entry list
2. Filter for 'Party (Journal Entry Account)' equals...

you'll notice that options are not being pulled. Also, if you repeat the same steps but selecting 'Party Type (Journal Entry Account)' in step 2, you'll find all sorts of doctypes appearing in the options list instead of just Party Types (i.e. Customer, Supplier, Employee)

Thanks

PR #11702, opened 2017-11-23
v8.x.x - Setup Category field for Party Type for Buyer and Seller

brownharryb:
This allows custom doctypes to be set as a Party Type specifying whether it's a seller or buyer and also enables Payment Reconciliation to work for these doctypes.

party_type_category


nabinhait:
@brownharryb Thanks for the pull request. Closing the pull request for the following reasons.

- We also have a plan to work on the same issue but to remove all the hard-coded codes for party type, like in Payment Entry / Journal Entry etc. Created a github issue for the same and it is included in the December milestone (#11226).
- The category "Buyer" and "Seller" is not so generic to define the category of party type.

PR #11716, opened 2017-11-24
Remove Company from Patient - Patches Added

pjamsheer:
Remove Company Refernce from Patient 
Patches Added to remove company from table Patient
Update patient DOB field label to Date of birth
Patient marital status default null


nabinhait (approved):
Many companies does not want to set default company purposefully, to force selecting company while creating transactions, I have removed the exception.

Issue #8427, opened 2017-04-12
Subcontracting purchase order - feature suggestion


somicoito:
When I make a subcontracting purchase order this should have impact on the stock of raw material items as it is for production orders.
Basically a subcontracting purchase order is like a production order, so they should affect the warehouse in the same way.

Devens:
+1

chanyork:
+1

PawanMeh:
PR - 11719 sent

PR #11719, opened 2017-11-24
[fix] #8427 - Add material reservation to sub contract

PawanMeh:
Testing Video

sub_contracting_po


PawanMeh:
Checked Travis errors, they do not seem related to this PR, assigning for review

erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -187,6 +187,8 @@ def on_submit(self):
 		self.update_prevdoc_status()
 		self.update_requested_qty()
 		self.update_ordered_qty()
+		if self.is_subcontracted == "Yes":
+			self.update_reserved_qty_for_subcontract()
```
nabinhait: It should be called on `on_cancel` as well

erpnext/stock/doctype/bin/bin.py
```diff
@@ -94,6 +94,28 @@ def update_reserved_qty_for_production(self):
...
+			required_qty = frappe.db.sql('''select sum(itemsup.required_qty)
+				from `tabItem` item, `tabPurchase Order` po, `tabPurchase Order Item Supplied` itemsup
+				where
+					item.name = itemsup.rm_item_code
+					and po.name = %s
+					and itemsup.rm_item_code = %s
+					and itemsup.parent = po.name
+					and po.docstatus = 1
+					and po.is_subcontracted = 'Yes'
+					and item.default_warehouse = %s''', (po_name, self.item_code, self.warehouse))[0][0]
```
nabinhait: We cannot rely on default warehouse set in Item master. It can be changed after submitting PO, and if the user cancelled thereafter what will happen? We need to introduce a new field for "Reserve Warehouse" inside "PO Item"/ "Item Supplied" table, we can set the default value based on item.

Also this type of function should not be dependent on args like po_name, transferred_qty, transaction_type. The function should be self dependent, so that we can call the function from other places like patches etc.
So, basically, it should recalculate reserved qty for the item and warehouse based on all submitted Purchase Orders and Subcontracted Stock Entries.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -64,6 +65,8 @@ def on_submit(self):
 		update_serial_nos_after_submit(self, "items")
 		self.update_production_order()
 		self.validate_purchase_order()
+		if self.purchase_order and self.purpose == "Subcontract":
+			self.update_purchase_order_supplied_items()
```
nabinhait: Should be called on_cancel

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -803,6 +806,30 @@ def validate_batch(self):
...
+		#Get PO Supplied Items Details
+		po_doc =  frappe.get_doc("Purchase Order",self.purchase_order)
+		po_supplied_items = po_doc.get("supplied_items")
+		items = list(set([d.rm_item_code for d in po_supplied_items]))
+		item_wh = frappe._dict(frappe.db.sql("""select item_code as "item_code", default_warehouse as "warehouse"
+			from tabItem where name in ({0})""".format(", ".join(["%s"] * len(items))), items))
+		#Update reserved sub contracted quantity in bin based on Supplied Item Details
+		for d in po_supplied_items:
+			warehouse = item_wh.get(d.rm_item_code)
+			transferred_qty = materials_transferred.get(d.rm_item_code + warehouse)
```
nabinhait: While calculating transferred qty against a PO, we should not consider source_warehouse in the Stock Entry. Whatever the Source Warehouse is, we have to reduce reserved qty from the warehouse where it is actually booked via PO.

nabinhait (changes requested):

nabinhait:
Test case missing

nabinhait:
Closing it for now, please reopen after fixing.

erpnext/stock/doctype/bin/bin.py
somicoito: sorry @nabinhait  but when I make a Production Order the material is reserved in the "Source Warehouse", you mean a new field "Source Warehouse" shall be created also in Subcontracted Purchase Order? 
Because in Production Order we actually don't have the "Reserve Warehouse" field.
Anyway if I actually make a Production Order and reserve material in a specific warehouse I still can transfer material from another warehouse and reserved material is deduced the same.

PawanMeh:
Would make the changes and resend the PR.

PawanMeh:
Made changes as suggested and re-opened PR (https://github.com/frappe/erpnext/compare/develop...PawanMeh:fixes_8427)

nabinhait:
Closing, reopen after making requested changes as per our discussion on phone.

PawanMeh:
Made changes as suggested and re-opened PR(https://github.com/frappe/erpnext/compare/develop...PawanMeh:fixes_8427)

PawanMeh:
Any update on this?

somicoito:
@nabinhait @PawanMeh that's fantastic!! Thanks to both of you! When this will be available?

PawanMeh:
@somicoito, the PR would next get into the staging branch and then to the production branch. I think in another week or two weeks, it should get into master branch.

Issue #8894, opened 2017-05-17
Suggestion for lead contacts 


me2ne:
Hello
I have suggestion for lead ,
in ERPNEXT we can’t add other contact details  for a lead,  if lead is company with different contact details we have to add different lead and it’s big problem ,specially  when people want to convert this type of lead to customer . my suggestion is  change this scenario ,  so  people can add lead for company with one primary contact and they can add more contact details for one lead ( company)  .
so when lead converted to customer all contact also link to that customer . 
Regards

PR #11724, opened 2017-11-24
Allowing the creation of an Organization as Lead

chdecultot:
Hi,

This pull request is linked to issue: https://github.com/frappe/erpnext/issues/8894

In the current workflow of ERPNext, a lead is in all cases a physical person.

But in the B2B world, sales persons try to get in touch within a company through different people and still consider the organization as the Lead.

We could create a new lead for each contact, but when the customer is created, we would need to re-create all contacts associated to this company manually.

Therefore my proposal is the following:

 We don't change the actual behavior, which is great for individual leads, but add an option to set the Lead as an organization.

In this case, it becomes possible to add several contacts to the same Lead.

When the Lead is converted, all contacts are automatically linked to the new customer.

_Lead Creation:_
image

_Customer:_
image

This PR is dependant on https://github.com/frappe/frappe/pull/4534


joolsr:
Really nice feature . Great work. I'm on the Community CRM team and this is a great addition to the CRM module. If you'like to join us you are more than welcome. Much of our initial emphasis will be on improving email handling which although not strictly CRM plays a big part in it and does need some work to bring it up to scratch compared to some other CRM tools.

chdecultot:
Thanks @joolsr,

Yes I'll be happy to join the community CRM team.
That's great if you are improving email handling because it is a much used feature that needs a lot of improvement.
I have also started working on an integration with Mautic that could bring a lot of powerful marketing automation features to businesses using ERPNext. Here is a very early stage alpha version pulling Contacts and Companies from Mautic to ERPNext: https://github.com/DOKOS-IO/mautic_integration

I plan to finish a working beta version before christmas and then push it to the core beginning of next year.

Issue #8192, opened 2017-03-28
Only show Sales Orders that have the Item to Manufacture in them in Production Order


dottenbr:
When selecting the ``Sales Order`` in a ``Production Order`` please only list those ``Sales Orders`` that have the ``Item to Manufacture`` in them as ``Sales Order Lines``.

Devens:
+1

Regards,
Deven Shah
098250 60116

dottenbr:
not as simple as it seems because it's a child of the doctype to be filtered

kickapoo:
@rmehta @nabinhait 

Where is an example to filter parent Doctype from Child table value ?

rmehta:
You can write a `DISTINCT` query on `parent` for Sales Order Item. That should be easy?

kickapoo:
@rmehta the way I know to make custom query is like 👍 
```
cur_frm.fields_dict['items'].grid.get_field('fieldname').get_query = function(doc, cdt, cdn) {
  var row = frappe.get_doc(cdt, cdn)
	return {
		filters: [
                 ['parent', '=', row.item_code ]
             ]
	}
 }
```

I dont know where to write the custom sql query.

kickapoo:
Found it

```
cur_frm.fields_dict[child table's fieldname].grid.get_field("item_code").get_query = function(doc, cut, cdn) {
return your query
}

```

kickapoo:
@dottenbr  
Please `Item to Manufacture` your mean in Item 'default_material_request_type' == 'Manufacture' ?

Sorry but I cant find an clear indicator for `Item to Manufacture`.

creamdory:
@kickapoo indicators for Item to manufacture would be if an item has a BOM setup.

PR #11726, opened 2017-11-24
Set query for sales_order based on selected production_item

Zlash65:
sales-order-query
[#8192](https://github.com/frappe/erpnext/issues/8192)

### DIFF
--- a/erpnext/manufacturing/doctype/production_order/production_order.py
+++ b/erpnext/manufacturing/doctype/production_order/production_order.py
@@ -634,3 +634,9 @@ def stop_unstop(production_order, status):
 	pro_order.notify_update()
 
 	return pro_order.status
+
+@frappe.whitelist()
+def query_sales_order(production_item):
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/production_order/production_order.py
```diff
@@ -634,3 +634,9 @@ def stop_unstop(production_order, status):
 	pro_order.notify_update()
 
 	return pro_order.status
+
+@frappe.whitelist()
+def query_sales_order(production_item):
+	out = frappe.db.sql_list("select distinct so.name from `tabSales Order` so, `tabSales Order Item` so_item\
```
nabinhait: Consider both `Sales Order Item` and `Packed Item` table for matching items to find out sales orders.

nabinhait (changes requested):

nabinhait:
Closing for now, please reopen after fixing.

PR #11740, opened 2017-11-27
[minor] Set Rename for Training Program

creamdory:
Allow rename for Training Program.

rename


creamdory:
Travis error is not related.

PR #11763, opened 2017-11-28
Fetch items of item groups defined in the pos profile

### LINKED ISSUE #11751: Offline POS won't fetch items from item groups selected in POS Profile
Hello,

I’ve been dealing with an issue for a while. The POS wouldn’t fetch items from item groups selected in POS Profile witch would result in no products displayed if the user would not manually define a Item Group. After trying many options, the issue was solved by re-enabling the Offline POS…

Seems like there is an issue with the online POS (using latest version and a few previous versions since over a month)

Hope this helps you find where the problem is with the Online POS which I prefer to use.

Thanks!

G.

rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/issues/11751


erpnext/accounts/doctype/pos_profile/pos_profile.py
```diff
@@ -95,6 +96,17 @@ def set_defaults(self, include_current_pos=True):
...
+def get_item_groups(pos_profile):
+	item_groups = []
+	pos_profile = frappe.get_doc('POS Profile', pos_profile)
+
+	if pos_profile.get('item_groups'):
+		# Get items based on the item groups defined in the POS profile
+		for data in pos_profile.get('item_groups'):
+			item_groups.extend(["'%s'"%d.name for d in get_child_nodes('Item Group', data.item_group)])
+
+	return item_groups
```
nabinhait: return list(set(item_groups))

nabinhait (changes requested):

erpnext/accounts/doctype/pos_profile/pos_profile.py
rohitwaghchaure: Fixed!

gbard:
The issue is back in v10.0.4 (master)

PR #11766, opened 2017-11-28
Minor changes around maintenance task syncing

KanchanChauhan:
- Asset Repair listview changes
- If values in Maintenance Task changes, it should reflect in Asset Maintenance Log also
- Auto create next Asset Maintenance Log to know what Maintenance in Planned for future (reflects in Calendar as well)


erpnext/assets/doctype/asset_maintenance/asset_maintenance.py
```diff
@@ -16,26 +16,44 @@ def validate(self):
 				throw(_("Start date should be less than end date for task {0}").format(task.maintenance_task))
 			if getdate(task.next_due_date) < getdate(nowdate()):
 				task.maintenance_status = "Overdue"
-			if not self.get("__islocal"):
-				if not task.assign_to:
-					task.assign_to = self.maintenance_manager
-				if task.assign_to:
-					self.assign_tasks(task)
+
+	def save(self):
```
nabinhait: Why you overwritten `save` method of document.py? You can call `sync_maintenance_tasks` method in `after_insert` or `on_update`.

nabinhait (changes requested):

PR #11798, opened 2017-11-30
[patch] Patch added to rename net_weight field

vishdha:


erpnext/patches/v9_2/rename_net_weight_in_item_master.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import unicode_literals
+import frappe
+from frappe.model.utils.rename_field import rename_field
+
+def execute():
+	frappe.reload_doc("stock", "doctype", "item")
+	if frappe.db.has_column('Item', 'net_weight'):
+		rename_field("Item", "net_weight", "weight_per_unit")
```
rohitwaghchaure: Patch will break if field weight_per_unit is exists!

rohitwaghchaure: if exists copy data from net_weight to weight_per_unit else rename_field("Item", "net_weight", "weight_per_unit")

joezsweet:
What's the goal of this?

vishdha:
Rename field net_weight to weight_per_unit in Item Master.

joezsweet:
Got it, i mean what's the goal of renaming?

vishdha:
We added shipping rule based on Net Weight. so, selling and buying module fetch weight data from the item master and calculate shipping charges based on that weight.

joezsweet:
@vishdha ok, thx for explanation

PR #11804, opened 2017-11-30
On changing assign_to assign maintenance task in Asset Maintenance

KanchanChauhan:
- Also Depreciation schedule and Accounts should be set only if `calculate_epreciation` is checked.
- Some doctype permission Fixes


erpnext/assets/doctype/asset/asset.py
```diff
@@ -17,12 +17,12 @@ def validate(self):
 		self.validate_item()
 		self.set_missing_values()
 		self.validate_asset_values()
-		self.make_depreciation_schedule()
-		self.set_accumulated_depreciation()
+		if self.calculate_depreciation:
```
nabinhait: Is this a new field?

nabinhait (changes requested):

erpnext/assets/doctype/asset/asset.py
KanchanChauhan: @nabinhait Yes, it is a checkbox. We set it in Asset if Depreciation should be calculated or not. Some assets does not depreciate like Leased assets or Land or softwares.
screen shot 2017-11-30 at 4 54 00 pm

PR #11812, opened 2017-11-30
Portal print format correction

chdecultot:
Hi,

On the portal, we have the possibility to print orders and invoices, but the default print format set by the user is not correctly fetched, even though the link to print the PDF includes it.

This PR is meant to correct this behavior.

**Logic:**

If a default print format has been setup for a particular doctype, then the PDF will be printed with this print format, else it will use the standard print format.

Thanks.

### DIFF
--- a/erpnext/templates/pages/order.py
+++ b/erpnext/templates/pages/order.py
@@ -24,9 +24,9 @@ def get_context(context):
 
 	context.enabled_checkout = frappe.get_doc("Shopping Cart Settings").enable_checkout
 
-	default_print_format = frappe.db.sql("""select value from `tabProperty Setter` where property='default_print_format' and doc_type='{0}'""".format(frappe.form_dict.doctype), as_dict=True)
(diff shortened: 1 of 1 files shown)


erpnext/templates/pages/order.py
```diff
@@ -24,6 +24,12 @@ def get_context(context):
 
 	context.enabled_checkout = frappe.get_doc("Shopping Cart Settings").enable_checkout
 
+	default_print_format = frappe.db.sql("""select value from `tabProperty Setter` where property='default_print_format' and doc_type='{0}'""".format(frappe.form_dict.doctype), as_dict=True)
```
rmehta: `frappe.db.get_value('Property Setter', dict(property='default_print_format', doc_type=frappe.form_dict.doctype)`

chdecultot: Thanks @rmehta. I have corrected it!

PR #11813, opened 2017-11-30
Batch Stock Items, having serial number can't be moved without inserting Serial Number (#11792)

### LINKED ISSUE #11792: Batch Stock Items, having serial number can't be moved without inserting Serial Number
An item, having serial number & batch, can't be moved to other warehouses, because of no input of field for serials.

To reproduce, Create an item with serials & batch. Now using Stock Entry, Material Receipt in any one warehouse. Let's assume 100 serial items are at X Warehouse, and have to move 50 Serial Items to Y Warehouse.

This will work well, while again using Stock Entry, and Material Transfer to Y Warehouse from X. It'll ask about Batch No. if item is having Batch, and as selected FIFO, first 50 Serials will be transferred to Y Warehouse. Perfect.

But, Same feature is there, in `Stock > Batch` and select same item having bath & serials. There will be shown, 50 serials in Y Warehouse, while another 50 in X. But, using MOVE button, it only asks for Warehouse where to transfer/move. We'll select warehouse, but `**MandatoryError: Row #1: Please specify Serial No for Item XYZ**` will be prompt, _only because Batch/Serial not selected._

tundebabzy:
Fixed #11792 

### Before
peek 2017-11-30 23-10

### After
peek 2017-11-30 23-11

### DIFF
--- a/erpnext/stock/doctype/batch/batch.js
+++ b/erpnext/stock/doctype/batch/batch.js
@@ -65,41 +65,38 @@ frappe.ui.form.on('Batch', {
 					// move - ask for target warehouse and make stock entry
 					rows.find('.btn-move').on('click', function() {
 						var $btn = $(this);
-						frappe.prompt({
-							fieldname: 'to_warehouse',
-							label: __('To Warehouse'),
-							fieldtype: 'Link',
-							options: 'Warehouse'
-						},
-						(data) => {
-						    const args = {
-							    item_code: frm.doc.item,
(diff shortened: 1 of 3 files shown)


KanchanChauhan:
@tundebabzy @nabinhait I think stock entry goes into the function `process_serial_numbers` even if Item just have batch and no serial numbers. It is messing our some customization, can you please look into this.
screen shot 2017-12-13 at 12 48 04 pm

PR #11836, opened 2017-12-04
Checkbox in Price list for Price not UOM dependant

shreyashah115:
#2798 
Summary:
1. Added a checkbox in Price List DocType `price_not_uom_dependant`.
2. When the checkbox is checked, any changes made in UOM and its conversion factor won't affect the price list.
3. If the checkbox is not checked, changes made to the UOM or conversion factor will be reflected in the Price List of that particular form.
4. Added a test case for the same.

uom


elvise:
+1

Issue #11845, opened 2017-12-04
Gross profit report [Paid Development]


imansoltani:
Please consider the following case:

I buy an item for CHF 27.00 (including VAT) and I sell it for CHF 28.50 (including VAT).

And when I get my Gross Profit Report, it shows a lost of CHF 0.61 which is not correct! This is because the system calculates the selling price excluding VAT (CHF 26.39) and deducts it from the buying price (CHF 27.00).

But in this case, the buying price must also be calculated without the VAT (CHF 25.00). If so, I will have a gross profit of CHF 1.39

Please notice when I buy the item for CHF 27.00, I pay also CHF 2.00 for the VAT which is included in the price.

tundebabzy:
VAT should not be included in the Item valuation

tundebabzy:
@imansoltani 
Actually, this is not a bug. You need to change the "Consider Tax or Charge For" field to "Total". The other options are "Total and Valuation" and "Valuation" and both will add the tax amount to the Item's valuation.

![peek 2017-12-05 13-24](https://user-images.githubusercontent.com/818803/33607291-ea3f2e74-d9c0-11e7-9e03-4f04577eb837.gif)

imansoltani:
My field "Consider Tax or Charge For" was already set to "Total"!

The only difference with your example is that my buying prices are VAT lncluded.

imansoltani:
The rates in the column of "The average buying rates" of the gross profit report are given by **including the VAT**. And this is precisely where the error comes.

For me the issue is not yet resolved.

tundebabzy:
Did you check the "Is this Tax Included in Basic rate" check box?

umairsy:
If you select "Total" as a category for the tax, the value of tax will not be included in the item's valuation. Only in the category for the tax or charges is selected as "Valuation" or "Valuation and Total", then an only value of that tax/charge is added to the item's valuation.

Generally, the tax amount is not included in the value of the item purchases, because after taking input credit on the tax, the tax amount payable is less. But if you still want tax amount to be added to the item's valuation, then ensure that you select "Valuation" or "Valuation and Total" as a category for the tax account.

Hope this helps.

imansoltani:
@tundebabzy 
Yes, the box is checked but the the VAT is not deducted from the buying rate in the report.

imansoltani:
@umairsy 
Here is the TAX setup for a purchase invoice:

[...]

PR #11857, opened 2017-12-05
Change Purchase Taxes and Charges default field for category to 'Total'

tundebabzy:
Inspired from #11845 
Most taxes that will go into a purchase invoice will mostly be recoverable taxes which should not be added to the Item's valuation. The default for `category` under `Purchase Taxes and Charges` is "Valuation and Total". This PR changes it to "Total" for user convenience.


joezsweet:
I think would be better to differentiate Taxes from Charges, as all charges like shipping, logistic cost has to be added to valuation while taxes are recoverable so, as discussed hasn't to be included in valuation. 

Other than that charges would be per supplier, as different supplier could add different charges.

Issue #11872, opened 2017-12-07
Failed switching to payment entry from paid expense claim form via View->Bank Entries


szufisher:
Steps to reproduce the error
1. set up the default account for a expense claim type
2. create a new expense claim, choose your own user account as approver, then SAVE, select Approved as approve status, click submit
3. click make payment button, input reference, and date, save and submit the payment
4. go back to the expense claim form, reload the page, the expense status is paid now, then click view->Bank entries button
5. system routed to journal entry/list instead of payment entry/list which failed switch to the desired payment entry

Here the extracted code from /erpnext/hr/doctype/expense_claim/expense_claim.js line 87 ~ 95
if (cint(doc.total_amount_reimbursed) > 0 && frappe.model.can_read("Journal Entry")) {
				cur_frm.add_custom_button(__('Bank Entries'), function() {
					frappe.route_options = {
						"Journal Entry Account.reference_type": me.frm.doc.doctype,
						"Journal Entry Account.reference_name": me.frm.doc.name,
						company: me.frm.doc.company
					};
					frappe.set_route("List", "Journal Entry");
				}, __("View"));

need to be changed as 
			if (cint(doc.total_amount_reimbursed) > 0 && frappe.model.can_read("Payment Entry")) {
				cur_frm.add_custom_button(__('Bank Entries'), function() {
					frappe.route_options = {
						"Payment Entry Reference.reference_doctype": me.frm.doc.doctype,
						"Payment Entry Reference.reference_name": me.frm.doc.name,
						company: me.frm.doc.company
					};
					frappe.set_route("List", "Payment Entry");
				}, __("View"));

already successfully tested in my own site www.erpnext.xin

as I am not real developer, have almost no successful experience contribute code via Github, can anyone help to do the pull?

umairsy:
Issue replicating.

shreyashah115:
@umairsy Fixed via #11874

PR #11874, opened 2017-12-07
Route to Payment Entry instead of Journal Entry in paid Expense Claim

shreyashah115:
#11872

### DIFF
--- a/erpnext/hr/doctype/expense_claim/expense_claim.js
+++ b/erpnext/hr/doctype/expense_claim/expense_claim.js
@@ -83,15 +83,26 @@ cur_frm.cscript.refresh = function(doc) {
 
 		if (doc.docstatus===1 && doc.approval_status=="Approved") {
 			/* eslint-disable */
-			// no idea how `me` works here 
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/expense_claim/expense_claim.js
```diff
@@ -84,14 +84,14 @@ cur_frm.cscript.refresh = function(doc) {
...
 				cur_frm.add_custom_button(__('Bank Entries'), function() {
 					frappe.route_options = {
-						"Journal Entry Account.reference_type": me.frm.doc.doctype,
-						"Journal Entry Account.reference_name": me.frm.doc.name,
+						"Payment Entry Reference.reference_doctype": me.frm.doc.doctype,
+						"Payment Entry Reference.reference_name": me.frm.doc.name,
 						company: me.frm.doc.company
 					};
-					frappe.set_route("List", "Journal Entry");
+					frappe.set_route("List", "Payment Entry");
```
nabinhait: Set route based on the value of "Make Payment via Journal Entry" defined in Accounts Settings.

nabinhait (changes requested):

PR #11886, opened 2017-12-08
[Minor] Update Status based on Asset Maintenance and Asse Repair

KanchanChauhan:
If asset required maintenance and if today it is in Maintenance, status should be "In Maintenance".

If Asset is in repair, status should be "Out of Order"

screen shot 2017-12-08 at 2 45 40 pm


erpnext/assets/doctype/asset/asset.js
```diff
@@ -31,6 +31,9 @@ frappe.ui.form.on('Asset', {
 		frm.events.make_schedules_editable(frm);
 
 		if (frm.doc.docstatus==1) {
+			if (frm.doc.maintenance_required) {
+				frm.trigger("update_maintenance_status");
```
nabinhait: Don't call this type of functions in `refresh`. Instead you should write a function to update the asset status and call it via Scheduler using hooks.

nabinhait (changes requested):

erpnext/assets/doctype/asset/asset.js
KanchanChauhan: @nabinhait Done via hooks. Please check if ok.

PR #11896, opened 2017-12-08
Payment Entry Returned (#8659)

### LINKED ISSUE #8659: Payment Entry Returned
Use case:
When we do a payment entry for receipt of payment, then usually the entry is adjusted against outstanding invoices. Now if incase the cheque is RETURNED then we have to cancel the existing payment entry and UN-ALLOCATE it so that it can be adjusted against the RETURN entry.

Ideally what should happen, is that we should have a button on a payment entry and click on RETURN so it should create a REVERSING ENTRY for that Payment entry and all the allocations for Sales Invoices should be replaced with the entry so that there is no need to cancel the payment entry.

tundebabzy:
This PR attempts to close #8659.

With this PR, users can have ERPNext automatically create a Journal Entry against a returned payment as well as a repayment.

When a Payment Entry is created, a "Return" button will be added. Behind the scenes, ERPNext will try to verify which button to display ("Return" or "Redeposit") or won't display the button at all if ERPNext is not sure.

screenshot from 2017-12-08 11-01-49

peek 2017-12-08 12-50

peek 2017-12-08 11-41

I've tried to track return and redeposit journal entries by saving the generated journal entry with Reference number as the Payment Entry name and Reference Date as Payment Entry posting date. The journal entry for returns also have Entry Type as Contra Entry while for redeposit its "Bank Entry"

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.js
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.js
@@ -112,7 +112,7 @@ frappe.ui.form.on('Payment Entry', {
 		frm.events.set_dynamic_labels(frm);
 		frm.events.show_general_ledger(frm);
 
-		if(frm.doc.docstatus === 1) {
+		if(frm.doc.docstatus === 1 && frm.doc.payment_type === 'Receive') {
 			frm.add_custom_button(
 				__('Return'),
 				function() {
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.js
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.js
@@ -115,12 +115,59 @@ frappe.ui.form.on('Payment Entry', {
 		if(frm.doc.docstatus === 1) {
 			frm.add_custom_button(
 				__('Return'),
-				console.log('Button has been added'),
+				function() {
+					frm.events.show_return_journal_dialog(frm);
+				},
 				__('Make')
 			);
 		}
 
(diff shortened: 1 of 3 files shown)


tundebabzy:
requires frappe/#4569

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -111,6 +111,141 @@ frappe.ui.form.on('Payment Entry', {
...
+			},
+			__(primary_button_name)
+		);
+	},
+
+	get_journal_dialog: function(frm) {
+		return new frappe.ui.Dialog({
+			fields: [
+				{
+					fieldtype:'Date', reqd:1, label:'Posting Date',
```
rmehta: label should be translatable

`__('Posting Date')`

Note: Fix all labels.

rmehta (changes requested):
Please add / update related documentation and add test cases.

rmehta:
@tundebabzy please reopen after fixing the issues.

vrms:
@tundebabzy does this feature satisfy the requirements for this open bounty? https://www.bountysource.com/issues/39242671-pdc-management-suggestion-full-feature-bounty
if so, I suggest to claim the funds, so the bounty can be closed.

vrms:
sorry, I was mislead. Thought this had been merged (as there was a reference on another PR for this being the better approach). So maybe my comment from above is irrelevant or is there a way to proceed with this and close both (this PR AND the bounty)?

PR #11905, opened 2017-12-08
Items should not be used from Sample Retention warehouse

KanchanChauhan:
Samples should not be taken out of retention warehouse for transactions.
Filter on from warehouse and source warehouse in Stock Entry

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -4,7 +4,6 @@ frappe.provide("erpnext.stock");
 
 frappe.ui.form.on('Stock Entry', {
 	setup: function(frm) {
-
 		frm.set_query('production_order', function() {
 			return {
 				filters: [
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -15,6 +14,25 @@ frappe.ui.form.on('Stock Entry', {
 			}
 		});
 
+		frappe.db.get_value('Stock Settings', {name: 'Stock Settings'}, 'sample_retention_warehouse', (r) => {
+			if (r.sample_retention_warehouse) {
+				frm.set_query("from_warehouse", function() {
```
nabinhait: There is already another setup_query for warehouse fields in `stock_controller.js`. It will overwrite that function.

https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/stock_controller.js#L10

nabinhait (changes requested):

nabinhait:
closing this for now

erpnext/stock/doctype/stock_entry/stock_entry.js
KanchanChauhan: @nabinhait Only way I can think of is to have same filters in stock_entry.js since we need them only for source warehouse. Cannot change the core filters on warehouse since it is being used in other doctypes as well. Let me know if it ok.

PR #11907, opened 2017-12-09
Hotels

revant:
Feature built during Gandhinagar Technical Training Event


elvise:
👍🏻

tundebabzy (changes requested):
The new doctypes are not accessible. 'You don't have permission'

revant:
Pushed some changes.

Select Domain

> Domain Settings > Hospitality

olamide2:
Good Job guys,

See my observations and ideas below

1. We need a PoS like interface to be able to use the hotel module properly
2. On the PoS interface all rooms will be displayed (we should create different room types like item groups so we can filter the room list by room type)
3. Looking at the PoS interface we should be able to see the status of each room, status for each room should be one of Available/Occupied/Dirty/Reserved.
4. Only rooms that are Available can have reservations made on them
5. Once a room is booked status becomes "occupied"
6. We should have a "check out" functionality for when a guest is checking out
7. Once a guest checks out, the status changes to "dirty"
8. Staff will need to manually change status from "dirty" to "Available" once the room is cleaned and ready for the next guest

If we can create this, it would be a start. Still a lot of things to include but if the 8 above is not in place…we cannot even start.

ryuliantoro:
Isn't the Domain Setting > Hospitality having Restaurant in it?
I think hospitality is more for Hotel than Restaurant.

rmehta:
dumping this for now, so that we can keep working on it in future meetups 😉

PR #11908, opened 2017-12-09
Chaining of sales and payment transactions

chdecultot:
Hi,

This PR is linked to https://github.com/frappe/frappe/pull/4595.

The goal of this implementation is to allow the creation of a chained log for sales invoices and payment entries if the option is switched on in site_config.json

If switched on, each submission of sales invoice or payment entry will be recorded in a specific log, where entries are chained together to be able to determine if an entry has been deleted fraudulently from the system at some point.


chdecultot:
Sorry, branch push error. Closed and to be reopened

Issue #9990, opened 2017-07-20
Project Gross Margin


dufani1:
Gross Margin dose not work with Sales Order, the Sales Order ( like the contract with customer) and has the total of project quoted amount, It should when make any payment through the sales order, this payment should be updated as income for this project not cost.
Please let us improve the cost tracking of projects and I am available for any help.

![untitled](https://user-images.githubusercontent.com/23026263/28413946-be7d4aca-6d49-11e7-84dd-4b37b3966345.png)

superlack:
Admittedly I did this, but didn't follow through with the rest of it.   Margin should also reflect employee salaries based chargeout, but I think it's currently set up that it would be based off of activity per employee.  An overhaul using the employees salary at time of time log instead of activity cost per employee is a better scenario

dufani1:
Hi superlack
lets enplane the following example:
Ex :
Project X  with Sales Order (contract of project) = 120000 $
Duration of project= 90 days
customer is Beta Company
project start date 1-June-2017
Until now, we gained 60000 $ from customer Beta by submitting partial sales invoice against the **Sales order** and gained 5000 $ from customer by submitting Sales Invoice against **Timesheets**.
Until now, we spent 25000 $ against **Purchase Invoices** and 10000 $ against **Expenses Claims** and 3000 $ against **Timesheets**.  
So the Project Gross Margin now is (60000+5000)-(25000+10000+3000)=27000 $ (Profit of project X) 
It hope clear

superlack:
Absolutely it's clear.  

What I had intended with having the Sales Orders in the project document as a way to view budgets.  In example: Deliver estimate or quotation, client accepts, becomes Sales Order which is the "Budget" for the job.  Any change orders associated are added (also through the use of Sales Order), and manager is able to keep track of labour, purchase, and expense cost.  The margin should definitely take into account a lot more factors than it currently does.  

For your proposal:

The calculation of $5000 - $3000 through timesheets.  I believe that should reflect different pay rates of employees (in HR/Salary module).  Right now it is handled as "Activity Cost", and drilling down "Activity Cost per Employee".  I think it's more common to have margins handled as a direct relation between employee cost and employee chargeout rate.  If there are different rates for activities, it should be formulaic, as the salary slip is. 

There are a lot of discussions recently on bolstering the Project module, especially to handle different billing formats, and tracking budgeting and profit margins.   

[...]

Issue #5393, opened 2016-05-24
Gross margin for projects


cyberdee:
Hi,

How to add manually gross margin calculation? I want to add a formula which is below:

Budget Estimate Cost vs Actual Cost
Actual Invoice Received vs Actual Cost
Sales Order vs Actual Cost

For example:

BUDGET ESTIMATE VS ACTUAL COST
Estimated Cost: Rp 250.000.000
Total Costing Amount (via Time Logs): Rp 88.966,67
Total Expense Claim (via Expense Claims): Rp 177.349.700,00
Total Purchase Cost (via Purchase Invoice): Rp 13.177.200,00

Margin = ESTIMATE - COSTING AMOUNT - EXPENSE CLAIM - PURCHASE COST = Rp 59.384.133,33

---

Sales Order: Rp 365.000.000
Total Costing Amount (via Time Logs): Rp 88.966,67
Total Expense Claim (via Expense Claims): Rp 177.349.700,00
Total Purchase Cost (via Purchase Invoice): Rp 13.177.200,00

Margin = ORDER - COSTING AMOUNT - EXPENSE CLAIM - PURCHASE COST = Rp 174.384.133,33

---

ACTUAL INVOICE RECEIVED VS ACTUAL COST

Actual Invoice Received: Rp 355.000.000
Total Costing Amount (via Time Logs): Rp 88.966,67
Total Expense Claim (via Expense Claims): Rp 177.349.700,00
Total Purchase Cost (via Purchase Invoice): Rp 13.177.200,00

Margin = INVOICE RECEIVED - COSTING AMOUNT - EXPENSE CLAIM - PURCHASE COST = Rp 164.384.133,33

Can ERP next automatically create those formula? Or should we add in the core?

Please advise

Thanks

rmehta:
This should be in the projects module. Yes you can add it in the core.

cyberdee:
Thanks @rmehta if you can add this in the core module that would be excellent!

rmehta:
If you need it immediately, you will have to contribute it. This is not on our priority list.

PR #11911, opened 2017-12-09
Project Margin Calculation Improvement

chdecultot:
Hi,

This PR is meant to correct some of the inconsistencies in the margin calculation for projects spotted in issues #5393 and #9990

Here are the corrections:

- Margin is now calculated as follow:
margin = total billing (Sales Invoice) - (total cost from timesheets + total cost from expense claims + total cost from purchase invoices)

- "Total Sales Cost" is renamed to "Total Sales Amount" as it described the sales against this project

- "Total Billing Amount" is renamed to "Total Billable Amount" for more consistency with the timesheet module denomination

- A new field "Total Billed Amount" displays the actual amount billed based on Sales Invoices.

- Time Log has been renamed to Timesheets for consistency with the timesheet module

- The project costing is now update when the document loads also, to avoid having some field with incorrect data.

Output:
project_costing

Thanks


nabinhait (approved):

Olawale1:
Hi @chdecultot 

Trust you're doing great. Thanks a lot for the great work here. I'm however re-evaluating the accuracy of using Purchase Invoices as a cost element in this calculation. Here's why:

A Purchase Invoice only captures the intent of the purchase. Imagine I purchased 100 units of Item Y for a project but somewhere along the line only used 50 and maybe diverted the other 50 to some other urgent project (this is a very common scenario in practice). This would make the Gross Margin calculation inaccurate

If however, we take that cost element from the actual consumption against the project (i.e. Delivery Note, Stock Entry), we would have a more accurate picture. Since those documents (Delivery Note, Stock Entry) don't contain the actual valuation rate of the items, it might be best to use outgoing Stock Ledger Entries for this purpose

What do you think?

Kind regards,

dufani1:
Hi @chdecultot nice work well done

cyberdee:
Thanks for the info!

PR #11924, opened 2017-12-11
[fix] Create Variant field position changed UX Issue

vishdha:
item


pratu16x7:
@vishdha Please check comments on original issue.

Issue #11902, opened 2017-12-08
Stock reconciliation issue when quantity was set to zero.


fresthy:
How to reproduce:
1. Create a new stock reconciliation
2. Set the quantity to any number other than zero: difference amount will be shown correctly
3. Set the quantity to zero: difference amount will be empty

https://discuss.erpnext.com/t/stock-reconciliation-gl-entries-dont-match-stock-reconciliation-difference-amount/24705
https://discuss.erpnext.com/t/stock-reconciliation-updating-items-difference-amount-issue/28575

Issue still persist in version
ERPNext: v9.2.21 (master)
Frappe Framework: v9.2.22 (master)

PR #11925, opened 2017-12-11
[Fix] Stock Reconciliation difference amount

shreyashah115:
#11902 
Difference amount was calculated as 0 whenever `item_qty == 0` as it considered `qty` instead of `item_qty` and calculated the amount.

qty


fresthy:
is it possible to release this as hotpatch for version 9?

PR #11930, opened 2017-12-11
[fix] Sales funnel text overlap

### LINKED ISSUE #10711: Sales funnel text overlap
Hi,
    In the sales funnel the top text are overlapping, please see below,
image

Is this a bug or can I change it myself?

BR

rohitwaghchaure:
**Before Fix**
screen shot 2017-12-11 at 3 15 10 pm

**After Fix**
screen shot 2017-12-11 at 3 13 47 pm

Fixed https://github.com/frappe/erpnext/issues/10711


nabinhait (changes requested):
@rohitwaghchaure Why there should minimum width in the funnel if it is zero? It should be fixed only for labels.

PR #11935, opened 2017-12-11
[Fix] Trigger shipping rule and tax recalculation

Zlash65:
fix-shipping


erpnext/public/js/controllers/transaction.js
```diff
@@ -355,6 +355,11 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 		}
 	},
 
+	weight_per_unit: function() {
+		this.calculate_taxes_and_totals();
+		this.shipping_rule();
```
nabinhait: `shipping_rule` trigger already executes `calculate_taxes_and_totals`, no need to call here

nabinhait (changes requested):

Issue #11956, opened 2017-12-12
Batch ID not fetched based on Expiry Date


umairsy:
In Batch Master:

![screen shot 2017-12-12 at 3 28 35 pm](https://user-images.githubusercontent.com/910238/33878468-8707253c-df51-11e7-9f6d-728ac5d9d383.png)

In Delivery Note:

![screen shot 2017-12-12 at 3 30 07 pm](https://user-images.githubusercontent.com/910238/33878478-8bcf883e-df51-11e7-86ff-63dd79d3810b.png)

PR #11960, opened 2017-12-12
[Fix] Batch selection in delivery Note

Zlash65:
Issue:- [#11956](https://github.com/frappe/erpnext/issues/11956)
screen shot 2017-12-12 at 4 59 01 pm
fix-batch-fetch
Batch 2 correctly fetched which is expiring before batch 1.


tundebabzy (approved):

umairsy:
@Zlash65 was 2 created after batch no. 1? Please ensure that it actually worked on Batch Expiry, and not based on FIFO.

Issue #11962, opened 2017-12-12
Chart of Accounts - add child to root node


Zlash65:
![issue-coa](https://user-images.githubusercontent.com/11695402/33882630-bcabbbdc-df5e-11e7-9b9e-fdc90d779be3.gif)
Adding child to root node should be disabled for Chart of Accounts

PR #11971, opened 2017-12-13
[Fix] Add child for root node should be disabled

Zlash65:
Issue:- [#11962](https://github.com/frappe/erpnext/issues/11962)
issue-coa

Fix:- 
fix-tree-root


nabinhait (changes requested):
@Zlash65 This is not the right fix. We need "Add Child" button for Company label as well in Chart of Accounts and Chart of Cost Centers. Because users are allowed to create new root accounts and cost centers.

Zlash65:
I was under the impression that root nodes cannot be added for them - since all the necessary root are pre-built. Will update the pr accordingly 👍

Issue #11959, opened 2017-12-12
Payment Entry - Party Name not fetching


umairsy:
On selection of Party, Party Name (Customer / Supplier) is not fetching in the relevant field.

![party-name](https://user-images.githubusercontent.com/910238/33881649-69fa6c88-df5b-11e7-8472-ca968d78e3fd.gif)

PR #11985, opened 2017-12-13
Fetch party name in Payment Entry

shreyashah115:
#11959 

party

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -617,10 +617,13 @@ def get_party_details(company, party_type, party, date):
 
 	account_currency = get_account_currency(party_account)
 	account_balance = get_balance_on(party_account, date)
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -617,10 +617,12 @@ def get_party_details(company, party_type, party, date):
 
 	account_currency = get_account_currency(party_account)
 	account_balance = get_balance_on(party_account, date)
+	party_name = frappe.db.get_value(party_type, party, party_name)
```
nabinhait: Is this working? It should give error `party_name is not defined`

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.py
shreyashah115: My bad! Fixed it.

PR #11986, opened 2017-12-13
Incoming rate fixes

rohitwaghchaure:
1. On save of stock entry, fetched valuation rate from previous stock ledger entry if valuation rate is zero.

1. Validate zero valuation rate if allow zero valuation rate is disabled on save of stock entry


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -213,6 +214,19 @@ def check_duplicate_entry_for_production_order(self):
...
+		for d in self.items:
+			if d.s_warehouse:
+				args = self.get_args_for_incoming_rate(d)
+				d.basic_rate = get_incoming_rate(args)
+			elif d.t_warehouse and not d.basic_rate:
+				warehouse = d.s_warehouse or d.t_warehouse
+				d.basic_rate = get_valuation_rate(d.item_code, warehouse,
+					self.doctype, d.name, d.allow_zero_valuation_rate,
+					currency=erpnext.get_company_currency(self.company))
+			elif d.allow_zero_valuation_rate and not d.s_warehouse:
```
nabinhait: This `elif` condition should be in 2nd position. If there is s_warehouse this elif will not be executed

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -213,6 +214,19 @@ def check_duplicate_entry_for_production_order(self):
 				frappe.throw(_("Stock Entries already created for Production Order ")
 					+ self.production_order + ":" + ", ".join(other_ste), DuplicateEntryForProductionOrderError)
 
+	def set_incoming_rate(self):
+		for d in self.items:
+			if d.s_warehouse:
+				args = self.get_args_for_incoming_rate(d)
+				d.basic_rate = get_incoming_rate(args)
+			elif d.t_warehouse and not d.basic_rate:
+				warehouse = d.s_warehouse or d.t_warehouse
```
nabinhait: why d.s_warehouse or d.t_warehouse ? If there is s_warehouse this `elif` will not be executed

nabinhait (changes requested):

PR #11997, opened 2017-12-13
Patch to copy old field data to new field

Zlash65:


erpnext/patches/v10_0/copy_projects_renamed_fields.py
```diff
@@ -0,0 +1,11 @@
+import frappe
+
+def execute():
+    """ copy data from old fields to new """
+    frappe.reload_doc("projects", "doctype", "project")
+
+    frappe.db.sql("""update `tabProject`
```
nabinhait: Don't update query to update the values.

Use `rename_field` function instead, it will replace the field in many other places.

Also execute `rename_field` only if column 'total_sales_cost' exists in the db.

nabinhait (changes requested):

PR #12019, opened 2017-12-14
Make Journal entry from Invoice based on Payment Terms (due date) #11989

### LINKED ISSUE #11989: Make Journal entry from Invoice based on Payment Terms (due date)
While making Journal Entry  from Sales/Purchase Invoice, add multiple rows in Journal Entry based on due dates.

tundebabzy:
peek 2017-12-14 18-07

Journal Entry can now be split according to invoice due date

Fixes #11989


rebin2002:
@tundebabzy thats exactly what I need, but I didn't find Reference Due Date in Journal Entry! 
and in your case, if you pay only first amount, then in AR report it shows the other 2 dates with due amounts?

PR #12047, opened 2017-12-15
Make Journal entry from Invoice based on Payment Terms (due date) #11989

### LINKED ISSUE #11989: Make Journal entry from Invoice based on Payment Terms (due date)

tundebabzy:
Correctly fixes #11989 this time around. Also fixes a bug where `reference_due_date` options were missing after loading the Journal Entry

peek 2017-12-15 13-08

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -687,12 +687,7 @@ def get_payment_entry(ref_doc, args):
 		"remark": args.get("remarks")
 	})
 
-	options_ref_name_list = [
-		d.due_date.strftime(DATE_FORMAT) for d in ref_doc.payment_schedule
-		if d.get('due_date')
-	]
-
-	for payment_term in ref_doc.payment_schedule:
+	if not ref_doc.payment_schedule:
 		je.append("accounts", {
 			"account": args.get("party_account"),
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -687,26 +687,39 @@ def get_payment_entry(ref_doc, args):
...
-		"is_advance": args.get("is_advance"),
-		"reference_type": ref_doc.doctype,
-		"reference_name": ref_doc.name
-	})
+	options_ref_name_list = [
+		d.due_date.strftime(DATE_FORMAT) for d in ref_doc.payment_schedule
+		if d.get('due_date')
+	]
+
+	for payment_term in ref_doc.payment_schedule:
```
nabinhait: What if no payment_schedule?

nabinhait (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.py
tundebabzy: Will work. I thought we migrated Sales/Purchase orders to automatically create Payment Schedule

PR #12078, opened 2017-12-18
[fix] Promise issue, in cart against item displaying zero value

### LINKED ISSUE #12074: POS item added to the cart displays price as zero
Hello,
 
Items added to the cart on POS displays price as zero, the total is not displayed with correct value too.

screen shot 2017-12-17 at 1 58 23 pm

Changing qty or applying discount makes the price appear.

rohitwaghchaure:
**Issue**
pos_before_fix

**After Fix**
pos_after_fix

Fixed https://github.com/frappe/erpnext/issues/12074


arafatmkd:
dear brother, I need this solution, I also have this problem. please give me the solution

Issue #7864, opened 2017-02-25
Final Manufacture Entry


dineshpanchal93:
Hi all,

We need solution of Multiple Stock Entry "Material Transfer to Manufacture" of against one Production order(Qty 1 ). and final we make Manufacture entry of that Production order, it will fetched all the items from Stock entry against related to Production order.

Right now, we can able to create only one Stock entry against production order.  
when we trying to add more stock entry against production order. gives me an error 
**"Material Transferred for Manufacturing (2.0) cannot be greater than planned quantity (1.0) in Production Order PRO-0002"**

Please do suggest me a solution.

"

aptech-admin:
+1

Really needed improvement

dineshpanchal93:
Yes we Solution for that. Please support

dineshpanchal93:
Any Update on this ?

djpalshikar:
+1
This is affecting us also, if there are items are added to the stock entry apart from the ones present in the BOM, we get the same message.

dineshpanchal93:
This issue again arrived.

Please kindly look into this asap.

https://discuss.erpnext.com/t/manufacturing-issues-in-version-11/47349

dineshpanchal93:
Allow to add items in manufacturing process which are not part of bom items #12110

Please update on the same.

CloudERPng:
This appears n ot to be resolved yet

mrodriguez614:
Hi everyone!
We're using ERPNext v13 and we also have the same issue above. Are there any updates on this?

sharmanakul:
similar to #28377

PR #12110, opened 2017-12-19
Allow to add items in manufacturing process which are not part of bom items

rohitwaghchaure:
**Issue**
We need solution of Multiple Stock Entry "Material Transfer to Manufacture" of against one Production order(Qty 1 ). and final we make Manufacture entry of that Production order, it will fetched all the items from Stock entry against related to Production order.

Right now, we can able to create only one Stock entry against production order.
when we trying to add more stock entry against production order. gives me an error
"Material Transferred for Manufacturing (2.0) cannot be greater than planned quantity (1.0) in Production Order PRO-0002"

https://github.com/frappe/erpnext/issues/7864

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -49,6 +49,9 @@ class StockEntry(StockController):
 		self.validate_with_material_request()
 		self.validate_batch()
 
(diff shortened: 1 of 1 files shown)


dineshpanchal93:
This issue was resolved.. but after updating in version v11 it wont work..

Issue #11044, opened 2017-10-04
[New] Production Plan


rohitwaghchaure:
- [ ] Make new doctype Production Plan.

- [ ] Add fields get items from sales order/material request and filters like item. customer, from date, to date.

- [ ] Add child table to show data of sales order/material request.

- [ ] Add child table with columns item code, default bom no, planned qty, pending qty and pull items data from the sales order/material request.

- [ ] Add button Make Production Order, create PO against an items available in the child table.

- [ ] Add button Make Material Request, create MR against an items in the child table.

- [ ] Progress bar of the production orders with status (Pending, Work In Progress, Completed)

- [ ] Documentation and Test Cases (UI & Server)

LifeP:
May I suggest consolidation of multiple sales orders for same product?

applepipe:
adding my 2cents:

- Create production orders for shortage qty, not the total qty on order (option)
- Filter orders by delivery date

hokgt:
Hi @rohitwaghchaure , 

I want to make suggestions to the Production module: 

**Better production module for multi-level BOM product** 
- Integrated ID for each multi-level BOM item Production Order to detect Production Order relation
- Automatically generate child Production Order from top level of multi-level Production Order
- BOM versioning : User may change BOM version for same item in one Production Order for production result

**Production Schedule and Management**
- Drag and drop Production Order in calendar view
- Create down time (time consumed for non-production activities such as machine maintenance etc), loss time (time consumed for non-productive activities). Down time and loss time created can be plug to production schedule.
- Calculate production productivity of each workstation and site productivity
- Calculate OEE (Overall Equipment Effectiveness) for site based on workstation availability to result

**Managing production by site**
Site has purpose to manage production result movement and stages of production (production routing). Similar with workflow but its including stock movement. 
- Merge workstation and operation
- Each site may have multiple workstations
- Material reservation for next route
- Production result from one site to replenish next site

**Simplified production item**
- Production Order based on item variants (user don't to create multiple item for different variants for production)

Thank you and if you need discussion regard to the suggestions, please contact me.

Issue #7947, opened 2017-03-07
Production Plan DocType -> Redesign Production Planning Tool


KanchanChauhan:
- [ ] Production Plan Doctype
  - [ ] From / To Date
  - [ ] Items child table (set quantities)
  - [ ] For Warehouse
  - [ ] Status (% complete)
- [ ] Add Production Plan link in Material Request
- [ ] Add Production Plan link in Production Order

bcornwellmott:
What is the purpose of benefit of this?

rmehta:
@bcornwellmott see updated description

bcornwellmott:
Thank you,

I'm wondering how this is going to work. 

For example, with status:
There are two major components for the PPT - Purchasing and Production. I'm not sure how you can characterize a % status with both these tools.

With MR and PrO documents, adding the Production Plan link can make sense, but what happens with items that are already in stock? Those won't be assigned to the production plan, so you will still have a disconnect between some items being allocated to the plan and others that aren't.

I guess I'm trying to understand the added value in having a Production Plan document that the current tool doesn't provide.

sharmanakul:
^ as mentioned in the connected link, the production plan is already enhanced to incorporate many of the features mentioned above. 
I'm not sure, how value adding % status feature is. A descriptive status, such as not started, in progress, completed, is already provided for the production plans.

Providing links in the material requests and work orders can be userful and i'll create a separate GitHub issue for that.

PR #12160, opened 2017-12-25
[enhance] Production Plan

rohitwaghchaure:
# Production Plan

Production plan helps user to plan production aginst the multiple sales orders or the material requests and also plan for the purchase of the raw materials which will be used in the production.

To use the Production Plan, go to:

> Manufacturing > Production > Production Plan

production_plan

#### Production Against Sales Orders

* Select option as Sales Order from the drop down list of get items from. System will show the filters, using that we can pull the sales orders for the production.

sales_order_filter

* Click on Get Sales Orders to fetch sales orders based on above filters

sales_order_items

* Click on Get Items for Production Order to fetch the items from the above sales orders.

sales_order_items

#### Production Against Material Requests

* Select option as Material Request from the drop down list of get items from. System will show the filters, using that we can pull the material requests for the production.

material_request_filter

* Click on Get Material Request to fetch material requests based on above filters

material_requests

* Click on Get Items for Production Order to fetch the items from the above material requests.
material_request_items

## Planning for Material Requests
* Click on get raw materials for production button to fetech raw materials required in the production.

material_request_plan

  * Include Non Stock Items :- To add non stock items in the material request planning.
  * Include Subcontracted Items :- To add subcontracted item's raw materials if use multi level bom is disabled
  * Ignore Existing Ordered Quantity :- If enabled then system will not check the projected quantity to make material request.

# Options To Make Production Order and Material Request
make_po_mr

# Production Plan List
screen shot 2017-12-25 at 2 27 07 pm

https://github.com/frappe/erpnext/issues/11044
https://github.com/frappe/erpnext/issues/7947


sachinhub:
Hi @rohitwaghchaure , What is the difference between existing production planning tool and this one?

rohitwaghchaure:
@sachinhub 

1. Using production plan, you can maintain the history of your production which is not possible in the production planning tool.

1. In production planning tool if you set Use Multi Level BOM then it applies to all the items selected for the production where as in production plan you can enabled/disabled Use Multi Level BOM against each production item

1. You can see the list of raw materials in the production plan to make material requests. You can also add/delete these items.

tundebabzy:
@rohitwaghchaure
Conflicts

sachinhub:
Ok @rohitwaghchaure will give it a try. We were also working on some enhacements to production planning tool. Mainly following - 

1. Ability to start and finish multiple production orders at a time.
2. Create a consolidated Material Request for all the Raw Materials needed for all Pending Production orders. A single material request will be created which will then used for stock transfer from raw material warehouse.
3. Automatically create stock entries (submit stock entry) along with pre-configured batch number format

rmehta:
The column "Use Multilevel BOM" is confusing... How about "Include Exploded Items" ?

rohitwaghchaure:
@rmehta Sure, I'll make change

rohitwaghchaure:
Fixed!

PR #12180, opened 2017-12-26
Payment schedule due date on orders should consider delivery date #12060

### LINKED ISSUE #12060: Payment schedule due date on orders should consider delivery date
Due dates are calculated using posting date, but It would be more accurate if it considered the expected delivery date on orders or quotation (sales and purchase), which will probably match the invoice date. Customers get confused with the dates on sales orders sent to approval.

tundebabzy:
Fixes #12060 
Summary:
- Purchase Orders, will use `schedule_date` and fallback to `transaction_date`
- Sales Orders will use `delivery_date` and fallback to `transaction_date`.

### Purchase Order -before
screenshot-2017-12-26 new purchase order 1 - new purchase order 1 2

### Purchase Order - after
screenshot-2017-12-26 new purchase order 1 - new purchase order 1 1

### DIFF
--- a/erpnext/buying/doctype/purchase_order/test_purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/test_purchase_order.py
@@ -18,6 +18,13 @@ class TestPurchaseOrder(unittest.TestCase):
 		pr = create_pr_against_po(po.name)
 		self.assertEquals(len(pr.get("items")), 1)
 
+	def test_purchase_order_payment_terms_due_date(self):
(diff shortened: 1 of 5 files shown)


nabinhait:
What will happen if there is a payment term like "Advance payment 50% within 5 days of confirmation of the order". In this case, it should be based on transaction date.

I also feel that this PR will complicate the process more. Delivery date is also not absolutely correct, it's just can give better result in some cases. It totally depends on which kind of terms are there. Let user give to edit the due date manually. I am closing this PR.

Issue #12129, opened 2017-12-21
'Standard Selling Rate' in Item Master Does Not Update Selling Standard Price 


kennethsequeira:
The Standard Selling Rate Field in Item master is a text field which gives the user the impression that the pricing can be changed from there. But change has to be made from **Price List**

The text field serves no purpose other than entering the price while creating the item and even that is for just creating the valuation rate. If you enter valuation rate, the selling price field is rendered obsolete.

![item-standard-selling](https://user-images.githubusercontent.com/33246109/34245059-4c861cca-e64e-11e7-90fc-ac7f7293f352.gif)

Proposed Solutions:
* link the field to Standard Selling Price for the Item so user can change the price from Item Master itself

* make it a 'View Only' and pull the rate from the Price List. 

* Change the label to something else which makes more sense.

https://discuss.erpnext.com/t/item-standard-selling-rate-vs-selling-price-list/31812

nabinhait:
Made the standard rate field hidden after saving the item record.

kennethsequeira:
Issue still not resolved. Let the issue remain open until then.

https://github.com/frappe/erpnext/issues/12744

aleoses:
Hi @kennethsequeira does this feature would be done at all?

heatherpotter:
This is badly needed. Has it been completed yet? There should not be a field on Item master that isn't ever updated.

KenyanGeek:
This is a big problem i had to do every item manually wasting a lot of time @kennethsequeira the workarounds with Data import are not helpful

PR #12232, opened 2017-12-28
Item std selling rate

vinayak--:
Possible solution for issue #12129  

Standard selling rate updated every time an item is modified.

peek 2017-12-28 20-12


nabinhait:
@vinayak-- Thanks for the PR 👍 

But if we update all price list rate of the item based on standard rate, it defeats the purpose of multiple price lists. And even if we update only default "Standard Selling" price list rate, it does not fix the issue entirely because many user deletes/rename that price list and give some more meaningful name. In that case, we don't know which price list the standard selling rate is linked to.

As per our current design, it is just a helper field to create item price record automatically while creating the new item. To remove the confusion later, I will hide the field after saving the item record.

Issue #12272, opened 2017-12-31
Program Enrollment Tool Showing Rejected & Pending Applicants


witedusa:
Hi ERPNext Team,
Program Enrollment Tool, is showing also Rejected Applicants in the list, and when i click enroll students, its enrolling them all.
System should not enroll rejected or pending approval applicants.
Please help us to solve this. ASAP

Below are the images to show:
![error6](https://user-images.githubusercontent.com/34868046/34463898-063197aa-ee7c-11e7-8e9e-064073085619.jpg)
![error6_1](https://user-images.githubusercontent.com/34868046/34463901-0ff1b3ce-ee7c-11e7-8daf-aff48f612815.jpg)

Issue also posted at discussion of ERPNext
[https://discuss.erpnext.com/t/program-enrollment-tool-showing-rejected-pending-applicants/32163](https://discuss.erpnext.com/t/program-enrollment-tool-showing-rejected-pending-applicants/32163)

Issue #12262, opened 2017-12-30
Addition of Batch field in Program Enrollment Tool Or Batch Creation Tool Needed


witedusa:
As batch is an important thing for Student Group Creation Tool. we have to mention batch for each program enrollment one by one. It will be great if we can have batch field in Program Enrollment tool. so that we can create batch in one go.
Another way is to give batch creation tool, to attach batches to program enrollments
Thanks

Issue #12206, opened 2017-12-27
Addition of Academic Term in Program Enrollment Tool


witedusa:
Our institue have short courses or long courses starting from two Quarters to ten Quarters. So we want student to be enrolled term (Quarter) based, when student finish first term, we can enroll him to the next term.
here are the issues with system:

- We are unable to enroll same student for different Quarters for a single year. using Program Enrollment

- We are unable to re-enroll already enrolled students to next term using Program Enrollment Tool

already posted on discussions of ERPNext:
[https://discuss.erpnext.com/t/program-enrollment-issue/31969](https://discuss.erpnext.com/t/program-enrollment-issue/31969)

Please help and fix.

PR #12291, opened 2018-01-02
Fix + Enhancement in Program Enrollment Tool

manassolanki:
- Academic Term in the Program Enrollment Tool
- Only approved student applicant can be enrolled
- Students from a specified batch can be fetched
- Batch can be also be set for individual students
- Setting the batch for the bulk creation

Fixes https://github.com/frappe/erpnext/issues/12272
Fixes https://github.com/frappe/erpnext/issues/12262
Fixes  https://github.com/frappe/erpnext/issues/12206


witedusa:
Great work, very nice fixings.
I tried to play with new changes and came up with a small problem.
Students are able to enroll in an Academic Term which is not Term of an Academic year, means we can enroll student or applicant in year 2016-17, term 2017-8 (Semester 1).
error7

manassolanki:
Hi @witedusa Please create an issue for the same.

Issue #10657, opened 2017-09-03
Made to Order in Shopping Cart


raghuzor:
I am trying to set up shopping cart. I manufacture 'made to order' Items. I have unchecked “maintain stock” in the items, I also set the website warehouse blank and with that I was hoping the Shopping cart would allow ordering but I see that it still displays “Not it stock”.

rmehta:
For non stock items, stock should not be seen

mghoneem10:
what about this issue coz i`m also trying to implement the same module 'made to order'  will it be included in v10 or check for another scenario  ?

PR #12294, opened 2018-01-02
fix for the non stock items for the shopping cart

manassolanki:
If there is some non-stock item, always show in stock for that.

Fixes https://github.com/frappe/erpnext/issues/10657


MichaelPinkowski:
This says that the issue was Fixed and Merged 10 days ago (~ Jan 3rd).  Does that mean that it was fixed in Ver 10 only and not in Ver 9.x?

nabinhait:
@MichaelPinkowski Yes, it has been fixed in v10 only. It is very difficult to push all fixes in the earlier versions with such small teams. Though if community sent any PR to the earlier versions for any bug fix, we generally merge them :)

Issue #12273, opened 2017-12-31
Cashier shift closing


mhbu50:
1.    A cashier can close his shift, so he will get the amount of cash that should be available in a drawer, and he can insert amount of cash that is actually available in the drawer.
2.    Print cash reconciliation report with shift duration in hours, shift starting date time, shift ending date and time, store name and username of the cashier, cashier name, and total sales.

ashish-greycube:
https://github.com/frappe/erpnext/pull/12334

bghayad:
Hello @mhbu50
There is not any progress about this issue?
Regards
Bilal

mansonkibe:
There is not any progress about this issue?

PR #12334, opened 2018-01-04
POS Cashier Shift Closing Logic against issue 12273

ashish-greycube:
Related #12273 

**pos_shift_detail** : You can define shift for cashiers example Day_9_18 with start time 9 and end time 6 PM
image

**pos_cashier_closing** :
 

1. Cashier can start his day by entering the amount he has in opening cash. He also needs to select user/start date/shift and opening cash
2. Cashier can close his day by entering the amount he has in closing cash. When he presses reconciliation button or save button. He is updated with POS system cash and Reconciliation Difference if any.
3. Cashier has the ability to open POS cashier closing doctype from POS itself with two menu options
image
image

**pos_sales_reconciliation report** :It is provided so that cashier can see all sales invoice generated by him for particular date and shift
image


olamide2:
Good idea ..but what of other payment methods like Credit cards ? I think the cashier shift closing should account for this too

bghayad:
Very important feature and really I dont know how POS is used now without having shift start and shift close.

Fredoluka:
The Reconcilliation should include all mode of payments and if there is  difference it should be made possible to channel it to the cashier pending investigation

nabinhait:
- "POS Shift Detail" doctype should be renamed to a generic name "Shift Work". It can be used for production / salary processing as well in the future
- "POS Cashier Closing" doctype should be renamed to "POS Session"
- Add tests
- Add documentation

Closing this for now, please reopen after fixing.

PR #12349, opened 2018-01-05
Chaining of sales and payment transactions

chdecultot:
EDIT: Sorry had to close and reopen after an error with git. Previously #11908

Hi,

This PR is linked to frappe/frappe#4595.

The goal of this implementation is to allow the creation of a chained log for sales invoices and payment entries if the option is switched on in site_config.json

If switched on, each submission of sales invoice or payment entry will be recorded in a specific log, where entries are chained together to be able to determine if an entry has been deleted fraudulently from the system at some point.


tundebabzy:
@chdecultot 
> where entries are chained together to be able to determine if an entry has been deleted *fraudulently* from the system at some point.

Do you mean deleted fraudulently or when an entry is deleted?

tundebabzy (approved):
Looks good thus far. Still waiting for Travis though

chdecultot:
Hi @tundebabzy,

Thanks for your review!
Actually, to answer your question, the purpose is to detect both. 
In France we are not allowed to delete sales transaction: If they are kept in the "Deleted Documents" folder it's fine but they should never disappear from the system.
Therefore if an entry is completely deleted from the DB by any mean it can be considered as a fraudulent deletion.
Hope it clarifies.

tundebabzy:
I forgot to add.... Transaction Log does not catch invoices from offline POS. Is that ok will regulators?

chdecultot:
Hi @tundebabzy ,

Very good question indeed. Of course that's not OK with regulator.

I am figuring out a way to completely remove the Offline POS option for french users. That would solve our problem, because I don't see any way to make it compliant as of now... (Local Storage cannot be trusted).

PR #12358, opened 2018-01-06
Added Payment Details to Sales Payment Summary Report

PawanMeh:
sales_payment_summary


erpnext/accounts/report/sales_payment_summary/sales_payment_summary.js
```diff
@@ -35,6 +35,11 @@ frappe.query_reports["Sales Payment Summary"] = {
 			"fieldname":"is_pos",
 			"label": __("POS?"),
 			"fieldtype": "Check"
-		}
+		},
+		{
+			"fieldname":"payment_detail",
+			"label": __("Payment Detail?"),
```
nabinhait: We generally avoid this type of label with question mark. Change it to "Show Payment Details"? Same in case of "POS?", it can be "Show only POS"

nabinhait (changes requested):

PawanMeh:
As requested, label changes have been made

Issue #4779, opened 2016-02-09
Delivery Note to be treated as GR and direct finance posting for delivery note itself


mkumar9699:
It's transport company and please find below the requirement and suggest.

They received goods through GR (Goods Receipt/Bilty form) from Consignor and send to Consignee.
No Stock updation, because it is a service industry
There are Three types of GR (Paid / To Pay / To Be Billed)
In the case of Paid and To Pay GR- They want to post financial entries to directly
But in the case of To Be Billed, they want to raise invoice for the collection of GR for a customer
No Rate requred for Item because it is a service item
One GR contains multiple Charges to be collected (.i.e Item 1 - 500 boxes Freight Charegs,Barrier charges,Door Delivery Charges Door Pickup Charges, GR Charges, Loading Charges, Unloading Charges to be collected from Consignor)
Please tell me how can I change Sales Tax template amount at the time time of generation of Invoice/delivery note. All Charges collected for a GR as an Income to be post in finance directly for Paid/To Pay customer Please suggest what to do. It should be fit in delivery note after slight modification

rmehta:
This is very complicated. Can you explain your use case with mockups?

jwrober:
Could workflow be used to aid this process?

PR #12370, opened 2018-01-08
[Fix] Delete root node for account

Zlash65:
[#4779](https://github.com/frappe/frappe/pull/4779)


jasonbrowne:
This would be a great feature the ability to delete root accounts.
Just ran an update this morning but was not able to get the command to work.
frappe@ERPNext-Demo2 frappe-bench] > bench delete root node allow 
Usage: bench [OPTIONS] COMMAND [ARGS]...

Error: No such command "delete".

jasonbrowne:
The delete button in the gui does not work for root accounts.
root-del

PR #12395, opened 2018-01-09
Set due date in purchase invoice based on suppler invoice date

rohitwaghchaure:


erpnext/controllers/accounts_controller.py
```diff
@@ -646,7 +646,7 @@ def group_similar_items(self):
 			self.remove(item)
 
 	def set_payment_schedule(self):
-		posting_date = self.get("posting_date") or self.get("transaction_date")
+		posting_date = self.get("biil_date") or self.get("posting_date") or self.get("transaction_date")
```
tundebabzy: Typo. bill_date not biil_date

erpnext/controllers/accounts_controller.py
```diff
@@ -646,7 +646,7 @@ def group_similar_items(self):
...
 		date = self.get("due_date")
 		due_date = date or posting_date
```
tundebabzy: `bill_date` is overwritten here except `due_date` is not set.

tundebabzy (changes requested):

tundebabzy:
PR doesn't work correctly:
peek 2018-01-09 13-59

erpnext/controllers/accounts_controller.py
rohitwaghchaure: Fixed

cosimot:
Hello! Any news on this?

Issue #12397, opened 2018-01-09
[non-profit]Donor Image and List view shows email addresses (user ids) instead of names


kennethsequeira:
![screen shot 2018-01-09 at 5 40 36 pm](https://user-images.githubusercontent.com/33246109/34720370-76910064-f564-11e7-960d-f6e2ab91d010.png)
![screen shot 2018-01-09 at 5 41 49 pm](https://user-images.githubusercontent.com/33246109/34720373-76bb6ce6-f564-11e7-8518-0747e7cc2844.png)

PR #12442, opened 2018-01-11
[non profit][minor] Allow name to be listed in image view

vishdha:
https://github.com/frappe/erpnext/issues/12397


tundebabzy:
Screenshots?

Issue #12280, opened 2018-01-01
Shopping Cart - Payment Terms error


umairsy:
When clicking on Add to Cart, it shows error related to Payment Term

![0eb2220f3f777fc16b9c8c1e2a3b02174df3197e_1_690x371](https://user-images.githubusercontent.com/910238/34467333-728a046c-ef14-11e7-8cd0-514b92c6739e.png)

https://discuss.erpnext.com/t/erpnext-v10-shopping-cart-issue/32207

lasalesi:
I confirm this bugs reproduces on several systems and seems to be caused by a hardcoded payment terms entry in _erpnext\shopping_cart\cart.py_

    "payment_terms_template": "_Test Payment Term Template",

Affects v10.0.5

lasalesi:
Quickfix: add a "Payment Terms Template" with name "_Test Payment Term Template" 😄

PR #12443, opened 2018-01-11
[fix] Issue Add to cart error

vishdha:
https://github.com/frappe/erpnext/issues/12280


tundebabzy (changes requested):
Please pull into hotfix

PR #12445, opened 2018-01-11
[Fix] Dropdown for Item Variant Settings does not show 'has_serial_no' field

Zlash65:
fix-has-serial-no
has_serial_no field wasn't visible before


tundebabzy (changes requested):
Looks good. You still have to write a patch to add `has_serial_no` in Item Variant Settings

saurabh6790:
@Zlash65 Please rebase and push again!

nabinhait:
@Zlash65 What was the actual issue?

This is designed in such a way, the field "has_serial_no" can only be set once. It is to avoid the changing value of has_serial_no after making some transactions against the same item. So, even if you change it in template, it will give a validation while setting in variants.
And other thing is, how you change "has_serial_no" field's value in the template? Because, as template item is non-stock item, the section in which has_serial_no field is there, is not visible.

somicoito:
@nabinhait the issue is that actually if I have an item master with "has serial No" set I can't make multiple Item Variants as per new feature.
Also a big problem is that when you make a variant you have one chance only to mark "has serial no" before saving, if you forget you must restart everything... not very good for data entry users.
If we are able to add this field in Item Variant Settings we shall solve these 2 problems.

somicoito:
@nabinhait now in item template "Has Serial No" is visible, so this should not be a problem

tundebabzy:
has_serial_no is no longer set_only_once as per #12614. I think this is ready to merge. @nabinhait  what do you think?
@Zlash65 conflicts

rohitwaghchaure:
This PR will update enabled, disabled variants has_serial_no even if stock entry exists. To fix this issue have disabled property set only once for has serial no field and field editable if there is no stock entries.
Fixed 
https://github.com/frappe/erpnext/pull/12614

cosimot:
@Zlash65  is this going to be merged?

PR #12462, opened 2018-01-12
PDC Enhancements

rohitwaghchaure:
- **Added PDC Details in accounts receivable report**
screen shot 2018-01-12 at 4 51 52 pm

- Added Customer PO No and PO Date in Sales Invoice

- Append multiple Customer PO No if one delivery note made from multiple sales order


Mishaal1021:
Can we add one column for Amount Remaining after PDC value? 

Also, does the report provide aging summary on the top? Can we see Print Version of this to ensure everything fits in the portrait version.

rohitwaghchaure:
@Mishaal1021 
Can we add one column for Amount Remaining after PDC value?
Do you meant outstanding amount - PDC/LC Amount, please correct me if I'm wrong

Also, does the report provide aging summary on the top? 
Can you please elaborate more, currently there is graph in the report

Can we see Print Version of this to ensure everything fits in the portrait version.
screen shot 2018-01-14 at 8 53 37 pm

Mishaal1021:
Dear Rohit,

1) Its balance remaining after the PDC amount. Please see attached excel version we use

2) Please see attached sample - this is paid customization report which we had done - the code is shared with you on GitHub already

3) It would be really good if we can limit the columns per the attachments, and keep this report as clean as possible. i.e. Ref column should only include voucher #, Party is only in report header - as its 1 report per customer, Paid Amt & Credit Note not required. Also, ideally it should be 1 row, since sometimes these can be 3 pages long with single rows. We don't want customers being distracted by too much info when asking them for money : )

4) Will the report include PDCs in draft. The way our customization was drafted was if posting date is less than cheque date, it would include it as PDC - regardless of status being draft or submitted. This is because we generally enter PDC in system when cheque is received, and then its submitted when sent to the bank - a lot of times customers will ask us to put cheque on hold. 
[Radiant - SOA Jan 6, 2018.pdf](https://github.com/frappe/erpnext/files/1631094/Radiant.-.SOA.Jan.6.2018.pdf)
[Radiant SOA upto 31-Dec-2017.pdf](https://github.com/frappe/erpnext/files/1631098/Radiant.SOA.upto.31-Dec-2017.pdf)

rohitwaghchaure:
@Mishaal1021 
Added remaining balance in the report
screen shot 2018-01-16 at 3 00 57 pm

Added summary in the print
screen shot 2018-01-16 at 3 02 52 pm


Regarding removing of few columns to show report clean, I am discussing with the team and soon we'll fix this

Mishaal1021:
Just a suggestion if we can give users options to select the columns to be printed, that would work equally well.

Mishaal1021:
Hi Rohit,

Any update on this?

rohitwaghchaure:
@Mishaal1021 

This PR is under review, we'll release after completion of review.

rohitwaghchaure:
**Added filter Show PDC in Print**
screen shot 2018-02-01 at 11 30 07 am

**With PDC**
screen shot 2018-02-01 at 11 30 38 am

**Without PDC**
screen shot 2018-02-01 at 11 30 24 am

Mishaal1021:
Rohit,

This looks great. Again, only issue is that each row takes space for 3 items. If we can remove the remarks columns, and remove "Sales Invoice" in the Ref column, it would be really appreciated. As mentioned, most companies may have more than 30 open invoices at a time which would result to a 3 page SOA with the current format. 

Looking forward to your positive response. 

Regards,

Mishal

Issue #12148, opened 2017-12-23
[Paid Development][Feature Request] BOM Stock Report for Exploded BOM


djpalshikar:
We regularly use BOM Stock Report. It is very useful. Would like to have a check BOX which shows the stock report of the exploded BOM. I am willing to sponsor it.

asharamseervi:
@djpalshikar Kindly write your expectation in detail, so anyone will easily pick and contribute, might you get it without payments. So, Kindly write complete expectation, what & how will happen, and how it will be useful for others. Thanks.

djpalshikar:
As such it is pretty straightforward. The following screenshot should help, let me know if there is a specific doubt:
![bom_stock_report_text](https://user-images.githubusercontent.com/23133635/34319117-91d1bce6-e7ff-11e7-8396-8ac7b5db264b.png)

PR #12475, opened 2018-01-14
Exploded view for BOM Stock Report

gaurav-naik:
Added a simple checkbox to show exploded view for BOM Stock Report

Issue for reference: https://github.com/frappe/erpnext/issues/12148

exploded_bom_stock_report


erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
```diff
@@ -6,52 +6,74 @@
...
+				bom_item.parent = '%s' and bom_item.parenttype='BOM'
+
+			GROUP BY bom_item.item_code""" % (conditions, bom))
+
+
+def get_exploded_bom_stock(filters):
+	root_bom_name = filters.get("bom")
+	root_bom = frappe.get_doc("BOM", root_bom_name)
+
+	bom_tree_list = root_bom.traverse_tree()
```
nabinhait: No need to traverse tree and execute the same query for all the sub BOMs in a for loop. There is a separate child table "BOM Explosion Item" inside BOM, which has all the exploded raw materials. So, in the above query, you just need to change the table from `tabBOM Item` to `tabBOM Explosion Item` in case of exploded view.

nabinhait (changes requested):

nabinhait:
@gaurav-naik Closing this for now, please reopen after making the requested changes.

erpnext/manufacturing/report/bom_stock_report/bom_stock_report.py
gaurav-naik: Pushed changes in PR https://github.com/frappe/erpnext/pull/12506.
I couldn't reopen https://github.com/frappe/erpnext/pull/12475 . Not sure how to do this.

PR #12494, opened 2018-01-15
Run country fixtures if country changed in the company

rohitwaghchaure:


erpnext/setup/doctype/company/company.py
```diff
@@ -80,9 +81,12 @@ def on_update(self):
 		if not frappe.db.sql("""select name from tabAccount
 				where company=%s and docstatus<2 limit 1""", self.name):
 			if not frappe.local.flags.ignore_chart_of_accounts:
+				frappe.flags.country_change = True
 				self.create_default_accounts()
 				self.create_default_warehouses()
-				install_country_fixtures(self.name)
+
+		if frappe.flags.country_change:
+			install_country_fixtures(self.name)
```
tundebabzy: Can we have a default fixture for countries that are ok with defaults? If I change from Nigeria to India, I get India specific fields. If I change back to Nigeria, India Specific fields are still there. If I change once again to Saudi Arabia, I now have India and Saudi Arabia fields

### Nigeria
screenshot-2018-1-15 new sales invoice 1 - new sales invoice 1

### Switched to India
screenshot-2018-1-15 new sales invoice 1 - new sales invoice 1 1

### Back to Nigeria
screenshot-2018-1-15 new sales invoice 1 - new sales invoice 1 1

### Switched to Saudi Arabia
screenshot-2018-1-15 new sales invoice 1 - new sales invoice 1 2

tundebabzy (changes requested):

erpnext/setup/doctype/company/company.py
rmehta: I think we should not delete fields, maybe there should be a page where you can check which countries' you want to setup.

PR #12498, opened 2018-01-15
fix documentation image links

crossxcell99:
Change image links from ".../docs/assets/img/..." to ".../assets/erpnext_docs/assets/img/..." to fix broken image links in documentation

Broken Image Link


tundebabzy (changes requested):
erpnext_docs is a symlink to docs. Why the change? I don't think this will fix the broken links.

crossxcell99:
it fixed it for me
screenshot from 2018-01-18 10-14-28

I guess the original link is not from the /www or /templates/pages folder so it does not open properly.
Regardless, this also takes care of cases where two apps have different files with the same file names in the same folder names, the symlink makes their links unique

crossxcell99:
It has been fixed

PR #12500, opened 2018-01-15
fixed image links in Users and Permissions

kennethsequeira:
* Fixed image links in role-based permissions
* Fixed image link in user permissions
* Fixed spelling mistake in user permissions


tundebabzy (changes requested):
This should be on hotfix branch so that it can be released ASAP

PR #12507, opened 2018-01-16
Added Assets link in the main index. Was missing till now.

umairsy:


tundebabzy (changes requested):
@umairsy 
I think this should go to hotfix branch. This is more of a fix rather than a new document.

manassolanki:
@tundebabzy I think it should go with `develop` because docs are created from the `develop` and published on erpnext.org

tundebabzy:
Oh OK. Cool. Merging...

PR #12519, opened 2018-01-17
Fix total monthly sales

ci2014:
We need to commit the db queries, before we call methods which update other doctypes, which rely on SQL queries, because otherwise the calculation of them is wrong and doesn't consider the current submitted or cancelled sales invoice.

I only saw problems with the current_month_sales. However, update_project relies on a direct SQL query as well. I don't know if update_time_sheet is affected, but it doesn't do any harm to commit the database query before calling the other methods.

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -140,6 +140,8 @@ class SalesInvoice(SellingController):
 		if not cint(self.is_pos) == 1 and not self.is_return:
 			self.update_against_document_in_jv()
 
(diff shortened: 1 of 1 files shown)


nabinhait:
Thanks for reporting the issue and sending PR.

Never commit in between, it creates problem if we get some error after commit statement. In that case, it does not rollback the entire process.
It is not considering current sales invoice sometimes, because `update_total_sales` method is executed via background jobs and can be executed sometime before completion of current sales invoice submission process. Otherwise, in on_submit method, you will get current sales invoice as submitted doc.

I have created another pull request to fix the issue. https://github.com/frappe/erpnext/pull/12522

PR #12532, opened 2018-01-18
ERPNext POS CSS

kennethsequeira:
* Uncommitted changes added
* Instructions from Achilles


rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/pull/12533

Issue #12322, opened 2018-01-04
Shopping Cart Enhancements


nabinhait:
- [ ] Mode of Payments selection
     - Cash on Delivery option
     - Option to choose the payment gateway
- [ ] Delivery Schedule
- [ ] Coupons / Gift Cards
     - [ ] Coupon Code Master
     - [ ] Based on the amount/quantity
     - [ ] Flat or percentage based
     - [ ] Underlying linked with the Pricing Rule
     - [ ] Max number of uses
     - [ ] Expiration 
- [ ] Loyalty Points
    - [x] Customer Enrollment in the Loyalty Program
    - [x] Collection from the Sales Invoices
    - [x] Expiry Date
    - [ ] Redemption

Olawale1:
Hi,

This is a very good enhancement. I trust that these features will be built into the relevant DocTypes so that they are not only limited to the Shopping Cart. Features like Coupons / Gift Cards and Loyalty Points should be available via POS / Sales Invoice as well

Thanks

rmehta:
https://github.com/frappe/erpnext/issues/12786

https://github.com/frappe/erpnext/issues/12787

dawoodjee:
+1 for cash on delivery, much more useful in places without readily usable Payment Gateways

PR #12559, opened 2018-01-19
[ADD]shopping_cart_coupon_code: feature of applying price rule on eco…

Solufyin:
@revant 
Feature to apply pricing rule using coupons mechanism on e-commerce before payment.

shopping_cart_coupon_code

shopping_cart_coupon_code_detail_in_so


revant:
Inviting 
@PawanMeh, @nabinhait as they had discussed a model for related issue.

@manassolanki for related issue for milestone

https://github.com/frappe/erpnext/issues/12322

manassolanki:
@Solufyin Thanks for the first contribution 👍 

Few feedbacks:
1. There is no need to create a separate module "Shopping Cart Coupon Code".
2. Do all the changes in the core doctype instead of creating the custom one.
3. Make a new doctype "Coupon Code" instead of the custom doctype "Shopping Cart Coupon Code". Since this is going to be generic which can be used inside the desk and POS.

Also please go through the given contribution guidelines which provide the some best practice approach and design related advice.

erpnext/modules.txt
```diff
@@ -20,4 +20,5 @@ Healthcare
 Restaurant
 Agriculture
 ERPNext Integrations
-Non Profit
\ No newline at end of file
+Non Profit
+Shopping Cart Coupon Code
```
revant: I think no need to add one more module. All the features can be added under already existing `Shopping Cart` module

erpnext/shopping_cart_coupon_code/custom/price_list.json
```diff
@@ -0,0 +1,80 @@
+{
```
revant: in case of PR don't use `custom`. you can make changes directly to doctypes.

manassolanki:
Hi @Solufyin , Let us know if you are still working on the same.

Solufyin:
Hello, @manassolanki and @revant,
we removed "Shopping Cart Coupon Code" Doctype And custom doc fields and make changes directly into doctypes.

erpnext/selling/doctype/quotation/quotation.json
```diff
@@ -2867,6 +2897,66 @@
...
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "total_amount_without_discount",
```
manassolanki: Why two extra field `total_amount_without_discount` and `discount`? These values can be calculated from the item tables in the client side itself (if your purpose is to show the discount amount in the shopping cart)

erpnext/selling/doctype/sales_order/sales_order.json
```diff
@@ -3516,6 +3516,96 @@
...
+   "set_only_once": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "total_amount_without_discount",
```
manassolanki: Same as the above comment

erpnext/shopping_cart/cart.py
```diff
@@ -48,6 +48,67 @@ def get_cart_quotation(doc=None):
 		"shipping_rules": get_applicable_shipping_rules(party)
 	}
 
+@frappe.whitelist()
+def apply_coupon_code(applyed_code):
```
manassolanki: Please also write generic method for the same so that coupon code can be used inside desk, pos or shopping cart. If you write generic one, you don't need to write this method, `get_cart_quotation` can be used for the same then.

Check the following for more reference
1. Frappe call for getting the item details. Also includes auto apply of the pricing rule.
https://github.com/frappe/erpnext/blob/59c915f5998533a3bb14407caafc9b11f1f798a9/erpnext/public/js/controllers/transaction.js#L315
2. Python method for applying the pricing rule
https://github.com/frappe/erpnext/blob/59c915f5998533a3bb14407caafc9b11f1f798a9/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L79

erpnext/shopping_cart/cart.py
```diff
@@ -48,6 +48,67 @@ def get_cart_quotation(doc=None):
...
+    if applyed_code:
```
manassolanki: Spelling `applied_code`

manassolanki (changes requested):

manassolanki:
Hi @Solufyin 
Please check my comments.

erpnext/selling/doctype/quotation/quotation.json
Solufyin: Hello @manassolanki 
we take both field "total_amount_without_discount" and "discount" to keep the history and also purpose to show the discount amount in the shopping cart. it will become helpful for further and future enhancement. 

Quotation and Sales Order both views are separate and when the shopping cart page is rendered that type it will get the values from the doc (database) and display on screen through the .html template.

Thanks

Solufyin:
Hello @manassolanki 
We improved the code and added generic method "get_coupon_code_discount" to get discount based on the price list and coupon code.

manassolanki:
Hi @Solufyin , Can we have a call over Skype to discuss this? You can contact me at manas@erpnext.com

manassolanki:
Hi, as per our call, please check the following todo's
- [ ] A new master named "Coupon"
image

- [ ] A new link field for the coupon in the item child table of the "Quotation", "Sales Order", "Delivery Note" and "Sales Invoice"

- [ ] Applying/removing coupon code should apply/remove the corresponding price list.

rmehta:
@Solufyin This PR has multiple issues and is incomplete design. Closing this for now.

Muzzy73:
@Solufyin May I ask for update on this? This looks good and would be essential for shopping cart

parsimonyERP:
@Muzzy73 This functionality is being included in a set of features called "Loyalty Code".  I think it is being released in v11.

PR #12564, opened 2018-01-19
[Enhancement] Searchable address in online POS, Customer Link field

saurabh-bhosale:
**Description:** 
Added few fields in _'Primary Address and Contact Details'_ Section in Customer Doctype to allow the user to select/use a primary address for a specific customer. This selected primary address becomes searchable in Customer Link fields just like the newly added searchable mobile_no feature in ERPNext version10

screenshot

**Use case:** 
1. Helps identify a customer in case of multiple customers with the same name.
2. There can be two types of POS Users: stationary and non-stationary. While already having a functionality of searchable mobile_no field in _"Customer"_ Link field, this enhancement of searchable address would be useful for POS Users who use POS while they are mobile(non-stationary) whereas stationary POS Users can use the searchable mobile_no functionality.
3. Be it POS or any other place, searchable address and mobile_no will both complement each other while adding up to the Customer Link field's usability and functionality.

searchable_address

P.S. Has a dependency on Frappe pull https://github.com/frappe/frappe/pull/4883


bkm888:
This is to support several medium sized businesses that make their own products and and employ a fleet of sales trucks to sell the products directly to their specialized customers. This business model is similar to large tool manufactures such as Snap-on tools here in the USA. They make mechanics tools and sell them directly to the mechanics at auto repair facilities from the back of their trucks. In this case they would have a several customers with the exact same name such as Deacon Jones Auto Sales which has 7 different locations. Each location has the exact same name but is differentiated only by their unique address. The same truck driver/salesman may have 3 of those Deacon Jones Auto Sales locations on his route and must be able to tell the difference between them when he stops at each one to sell tools. Most other systems include the address of a location for this purpose. The mobile salesman plans their day from a map of the city. Destination points are how they work, so addresses are important to them.

This is also true for the bakery that sells muffins to all of the Starbucks coffee locations. They are all called Starbucks, but they each have a different address and management team. They are all billed independently as well. Yet they all go by the exact same name. In this case the bakery truck salesman needs to find each Starbucks by their address in the POS customer lookup.

I am currently working to take over many route based businesses with new ERP systems to better manage their manufacturing and distribution. This is a critical feature for route managers to be able to properly identify the customers they sell and invoice during the course of their sales day.

bkm@jmibarcodes.com

tundebabzy (changes requested):
I can't select anything for Customer Primary Address:
screenshot-2018-1-25 address

peek 2018-01-25 09-06

Why are you query dynamic link table?

saurabh-bhosale:
@tundebabzy I was not able to recreate the issue on my local system. Can you please recheck and let me know how shall I recreate it so that I can rectify it?

I tried the following different approaches, like:
1. Adding already existing addresses to an existing Customer and then checking the correctness of the link field options.
2. Creating a new customer and adding new addresses for that customer on the go.

Also, I've tried testing it with the dependency branch's changes as well as without it, the link field options are shown correctly in both the cases.

Approach 1:
approach1

Approach 2:
approach2

Reason to use Dynamic Link Table in the query is, _'Address'_ Doctype has the _'Dynamic Link'_ childtable and using it in the where clause of query helps to pick the primary addresses for that specific customer only.

tundebabzy:
@saurabh-bhosale 
Cool. @rohitwaghchaure will test it further. My setup is probably messed up

erpnext/selling/doctype/customer/customer.js
```diff
@@ -33,7 +33,27 @@ frappe.ui.form.on("Customer", {
...
 	},
+	customer_primary_address: function(frm){
+		if(!frm.doc.customer_primary_address){
+			frm.doc.primary_address = "";
+		}
+	},
+	customer_primary_contact: function(frm){
+		if(!frm.doc.customer_primary_contact){
+			frm.doc.mobile_no = "";
+			frm.doc.email_id = "";
```
rohitwaghchaure: Use
```
frm.set_value('mobile_no', '')
frm.set_value('email_id', '')
```

rohitwaghchaure:
@saurabh-bhosale 
Kindly make suggested changes in the [dependent PR](https://github.com/frappe/frappe/pull/4883), so I can merge this PR.

rohitwaghchaure (approved):

saurabh-bhosale:
@rohitwaghchaure Have tried to remove the dependency on Frappe PR. Also, found a bug and I've tried to fix that as well. When an existing contact is edited for a customer this change isn't reflected in the Primary Details section's mobile_no field if that same contact is selected(user has to choose the same option again to have the correct data in the search field) 

Bug Before Fix:
before_fix

After Fix:
after_fix

erpnext/selling/doctype/customer/customer.py
```diff
@@ -20,6 +20,7 @@ def onload(self):
 		"""Load address and contacts in `__onload`"""
 		load_address_and_contact(self)
 		self.load_dashboard_info()
+		self.fetch_primary_address_and_contact_detail()
```
rohitwaghchaure: Why this method? If user want to update the value of mobile no, email id  then he will reselect the contact in the customer form

saurabh-bhosale: Exactly. The User has to know about it or there must be some way by which the form gets dirty and user comes to know that he/she has to reselect the option. Just a thought if something can be done for this or I'll revert the changes but at some later point one might consider it as a bug.

saurabh-bhosale:
@rohitwaghchaure Have refactored the code a bit and have removed the dependency on 'complete_address' field. I will close the dependant PR, if this is good to go with.

saurabh-bhosale:
@tundebabzy  Can you try testing it again now and update the _'waiting-for-fixes'_ tag accordingly?

Issue #12583, opened 2018-01-22
Total Field don't get updated if item row is deleted - V10


ushashmi:
After upgrade to version 10 the total field dont get recalculated if any item is deleted from the item list. we have to hit the save button every time to recalculate the total. 

This issue exist on almost all the doctypes like purchase order, purchase receipt, Sales Invoice etc.

shreyashah115:
Thanks for reporting the issue. 
Fixed via #12590

PR #12584, opened 2018-01-22
[Fix] Calculate taxes when shipping rule not present at item removed.

shreyashah115:
#12583 
Issue:
items_remove_error
Fix:
items_remove_fix

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -648,6 +648,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				}
 			}).fail(() => this.frm.set_value('shipping_rule', ''));
 		}
+		else {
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -648,6 +648,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				}
 			}).fail(() => this.frm.set_value('shipping_rule', ''));
 		}
+		else {
```
vishdha: No need to call calculate_taxes_and_totals here.

vishdha:
In your first image.  _Test_FG_Item may not contain weight details. Add weight_per_unit at item master.

vishdha (approved):

vishdha:
change base from develop to hotfix.

shreyashah115:
Will create a new PR.

PR #12601, opened 2018-01-22
Add min/max amount for selecting pricing rule

sunnyakaxd:
Often there is a requirement of selecting discounts when amount of a particular item exceeds a specific amount. This PR adds a small modification to provide that functionality.

Added two fields min_amt and max_amt, on which an additional filtering happens for selecting the rule.

screenshot from 2018-01-22 16-34-11

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -111,8 +111,8 @@ def apply_pricing_rule(args):
 
 	item_list = args.get("items")
 	args.pop("items")
-	
-	set_serial_nos_based_on_fifo = frappe.db.get_single_value("Stock Settings", 
+
+	set_serial_nos_based_on_fifo = frappe.db.get_single_value("Stock Settings",
 		"automatically_set_serial_nos_based_on_fifo")
 
 	for item in item_list:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -291,6 +291,11 @@ def filter_pricing_rules(args, pricing_rules):
 		pricing_rules = filter(lambda x: (flt(stock_qty)>=flt(x.min_qty)
 			and (flt(stock_qty)<=x.max_qty if x.max_qty else True)), pricing_rules)
 
+		amt = args.get('qty') * args.get('conversion_factor', 1) * args.get('price_list_rate', 0)
+
+		pricing_rules = filter(lambda x: (flt(amt)>=flt(x.min_amt)
+			and (flt(amt)<=x.max_amt if x.max_amt else True)), pricing_rules)
```
rohitwaghchaure: Seems you have added min_amt and max_amt using custom field in your local account and therefore it's not showing in the pricing_rule.json file, add these fields in the pricing rule doctype and push the pricing_rule.json file changes.

rohitwaghchaure:
@sunnyakaxd 
User can add the price in the pricing rule, if user has set min amt as 0, max amt as 50 and rate as 100 in the pricing rule. In transaction the amount is 40, does system will apply the above pricing rule because rate is 100 which is more than max amount?

rohitwaghchaure:
Closing the PR, make the requested changes and reopen it again

PR #12602, opened 2018-01-22
Select multiple raw material items for Sub Contract Stock Entries

PawanMeh:
Testing Video

subcontract testing

1. Allow selection of multiple items from Sub Contract PO based on BOM.
2. Group raw material stock entries by BOM Finished Goods and create the stock entries.

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -34,6 +34,12 @@ class PurchaseOrder(BuyingController):
 			'overflow_type': 'order'
 		}]
 
+	def onload(self):
+		super(PurchaseOrder, self).onload()
+
+		self.set_onload('disable_fetch_last_purchase_rate',
+		cint(frappe.db.get_single_value("Buying Settings", "disable_fetch_last_purchase_rate")))
+
 	def validate(self):
 		super(PurchaseOrder, self).validate()
 
@@ -393,6 +399,7 @@ def make_purchase_invoice(source_name, target_doc=None):
 @frappe.whitelist()
 def make_stock_entry(purchase_order, item_code):
 	purchase_order = frappe.get_doc("Purchase Order", purchase_order)
+
 	stock_entry = frappe.new_doc("Stock Entry")
 	stock_entry.purpose = "Subcontract"
 	stock_entry.purchase_order = purchase_order.name
@@ -437,6 +444,7 @@ def make_rm_stock_entry(purchase_order, rm_items):
 			stock_entry.supplier_address = purchase_order.supplier_address
 			stock_entry.address_display = purchase_order.address_display
 			stock_entry.company = purchase_order.company
+			stock_entry.docstatus = 0
 			stock_entry.from_bom = 1
 			po_item = [d for d in purchase_order.items if d.item_code == item_code][0]
 			stock_entry.fg_completed_qty = po_item.qty
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -435,30 +435,29 @@ def make_rm_stock_entry(purchase_order, rm_items):
 		item_wh = frappe._dict(frappe.db.sql("""select item_code, description
 										from `tabItem` where name in ({0})""".
 										format(", ".join(["%s"] * len(item_code_list))), item_code_list))
+		stock_entry = frappe.new_doc("Stock Entry")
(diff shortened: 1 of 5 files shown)


PawanMeh:
Checked Travis errors, do not seem to be related to this change. Assigning for review.

erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -415,6 +415,54 @@ def make_stock_entry(purchase_order, item_code):
 	stock_entry.get_items()
 	return stock_entry.as_dict()
```
rohitwaghchaure: There is two methods make_stock_entry, make_rm_stock_entry for the same functionality, can you commonify it?

erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -415,6 +415,54 @@ def make_stock_entry(purchase_order, item_code):
...
+			for rm_item_data in rm_items_list:
+				if rm_item_data["item_code"] == item_code:
+					items_dict = {rm_item_data["rm_item_code"]:
+					{"item_name":rm_item_data["item_name"],
+					"description":item_wh.get(rm_item_data["rm_item_code"]),
+					'qty':rm_item_data["qty"],
+					'from_warehouse':rm_item_data["warehouse"],
+					'stock_uom':rm_item_data["stock_uom"]}}
+					stock_entry.add_to_stock_entry_detail(items_dict)
+			stock_entry.save()
```
rohitwaghchaure: Why save separate stock entry for each item, you can make single stock entry against selected items and no need to save. On click of transfer button open single unsaved stock entry with the selected items.

erpnext/buying/doctype/purchase_order/purchase_order.py
nabinhait: I think `make_stock_entry` does not required anymore.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -64,6 +64,14 @@ def validate(self):
...
 
 	def on_submit(self):
+		if self.purpose == "Subcontract":
+			for d in self.items:
+				if not d.t_warehouse:
+					if self.to_warehouse:
+						d.t_warehouse = self.to_warehouse
+					else:
+						frappe.throw(_("Target warehouse is mandatory for row {0}").format(d.idx))
+
```
rohitwaghchaure: No need to make these changes

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -157,7 +165,9 @@ def validate_warehouse(self):
 				if self.to_warehouse:
 					d.t_warehouse = self.to_warehouse
 				else:
-					frappe.throw(_("Target warehouse is mandatory for row {0}").format(d.idx))
+					#move validation for sub contract to submit
+					if self.purpose != "Subcontract":
+						frappe.throw(_("Target warehouse is mandatory for row {0}").format(d.idx))
```
rohitwaghchaure: No need to make these changes

erpnext/buying/doctype/purchase_order/purchase_order.py
PawanMeh: If we don't want to maintain old way of stock entry creation for one finished good, we can just change if condition in JS to call the new modal in all cases and then remove the old method.

rohitwaghchaure: Have checked the code and found that BOM No is used for reference only. Can we create the single stock entry against all the items without BOM No?

PawanMeh: In that case, what should we set in stock entry (finished good completed quantity, should it be sum of all finished goods item completed quantities? or should it be zero?)
stock_entry.fg_completed_qty = po_item.qty

PawanMeh:
Made changes to make single stock entry instead of multiple & update bom no at stock entry detail level

erpnext/buying/doctype/purchase_order/purchase_order.py
rohitwaghchaure: Can you remove the old method make_stock_entry, we can use method make_rm_stock_entry if one item is there.

PawanMeh:
Made changes, travis error does not seem to be related to this. Manual testing video attached

subcontract stock entry

PawanMeh:
Removed old stock entry method and rewrote test case accordingly.

rohitwaghchaure (approved):

PR #12606, opened 2018-01-22
[Website] Price in items list and sales uom management

ghost:
Hi,

I add the possibility to display items price in product template (data available in context) and the price management based on the sales UOM.

For instance, I sell carrots with a stock UOM in Kg and a sales UOM in 500g.
So I have a conversion factor of 0.5.
The standard price rate is 2.5€ (for stock UOM = 1 Kg) and the sales price is 1.25€ (for sales UOM = 500g).

image

I've also taken into account the UOM conversion factor to display the website stock in sales UOM.
For instance in the exemple below, there is 8.5 Kg of carrots in stock UOM, so 17 Kg in sales UOM (500g).

image

### DIFF
--- a/erpnext/templates/pages/product_search.py
+++ b/erpnext/templates/pages/product_search.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe.utils import cstr, nowdate, cint
 from erpnext.setup.doctype.item_group.item_group import get_item_for_list_in_html
+from erpnext.shopping_cart.product_info import get_product_info_for_website
 
 no_cache = 1
 no_sitemap = 1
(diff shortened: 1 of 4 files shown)


erpnext/templates/pages/product_search.py
```diff
@@ -42,5 +43,17 @@ def get_product_list(search=None, start=0, limit=12):
 		"today": nowdate()
 	}, as_dict=1)
 
+	for item in data:
+		product_info = get_product_info_for_website(item.item_code)
```
rmehta: this is duplicate code. Please convert to a function!

rmehta (changes requested):

nabinhait:
@britlog Seems like you did not push the changes in the template. I did not find any code using `price_sales_uom`  or `formatted_price_sales_uom`. Closing it for now.

erpnext/templates/pages/product_search.py
britlog: @rmehta Thanks for the review, function added in PR #13335

britlog:
@nabinhait Thanks for the review. Indeed, I haven't included the templates, as it is often customized but  I've added them now for product list and product page in #13335

PR #12359, opened 2018-01-06
Print Functionality in Online POS using allow print before pay check

deepak-mnt:
1. "_Allow print before pay_" functionality in Online POS
2. Test Cases to check the print before pay functionality in Online POS
3. Handles the Sales Invoice update/creation in both cases(print before and after submit)

jmi_pr_1

PR #12620, opened 2018-01-23
Print Functionality in Online POS using allow print before pay check

deepak-mnt:
**Description:**
1. "Allow print before pay" functionality in Online POS
2. Handles the Sales Invoice update/creation for Online POS in both cases(print before and after submit)

**Note -** Have made the changes suugested in https://github.com/frappe/erpnext/pull/12359 and re-opened it

jmi_pr_1


erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -88,11 +88,15 @@ def get_conditions(item_code, serial_no, batch_no, barcode):
 	return '%%%s%%'%(frappe.db.escape(item_code)), condition
 
 @frappe.whitelist()
-def submit_invoice(doc):
+def submit_invoice(doc,is_saved):
 	if isinstance(doc, basestring):
 		args = json.loads(doc)
 
-	doc = frappe.new_doc('Sales Invoice')
+	if(int(is_saved) == 1):
```
rohitwaghchaure: Use cint instead of int

rohitwaghchaure (approved):

PR #12633, opened 2018-01-24
[fix] Sales order link to purchase order not working fixed

vishdha:
Create sales order.
Create material request from the sales order.
Create purchase order from the material request.

Purchase Order Link visible in Sales Order.
Ordered Qty update in a sales order.


snvora:
Hi,
This is great. 

Will it work if a quotation is first created from the material request and then purchase order is made?

vishdha:
Do you mean from sales order -> material_request -> supplier_quotation -> purchase order.

snvora:
yes

vishdha:
not yet,
I will do it for supplier quotation too.
just need to add docfield in supplier_quotation

tundebabzy (changes requested):
Works well thus far. Only thing remaining is that already created Sales Orders remain as is because the fix only works for new Sales Orders

rohitwaghchaure:
Write patch to update sales order in the purchase order item for the existing purchase orders

rohitwaghchaure (approved):

PR #12637, opened 2018-01-24
Organize ERPNext JS files according to webpack

netchampfaris:
Based on https://github.com/frappe/frappe/pull/4902


achillesrasquinha:
😭 Can't believe this is coming to life!

Issue #12630, opened 2018-01-24
AR: New Columns and filters Suggestions in Accounts Receivable Report


adityaduggal:
Hi,
This is an enhancement for the AR report and AR summary report as below:

- [x] Column for Sales person should be there, the only issue is that since Sales Person is a separate table in the Customer, the issue could be with multiple Sales Person's for a particular customer. Would need help from someone on this column (We can also drop this as we already have a filter for the same)

- [x] There should be a filter for Sales person as well and the filter should be able to consider the TREE view as well

- [x] There should be a column for sales partner as well.(Dropping the plan for a column since it would create many columns which are not always used hence only filter is there)
- [x] Filter for Sales partner as well.
- [x] Filter for TERRITORY should also be there and it should consider the TREE structure of the TERRITORY.

PR #12638, opened 2018-01-24
Resolves Issue #12630 -Added Filters to AR

adityaduggal:
- Added Territory Filter to AR report
- Added Sales partner filter
- Added Sales person filter
-Terrtory filter is needed since if you need to see AR in North Inda you get with filter not with column
-Similarly there was a need for Sales Partner and sales person filter in the AR report which is being fulfilled now.

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.js
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.js
@@ -29,6 +29,24 @@ frappe.query_reports["Accounts Receivable"] = {
 			"options": "Payment Terms Template"
 		},
 		{
+			"fieldname":"territory",
+			"label": __("Territory"),
+			"fieldtype": "Link",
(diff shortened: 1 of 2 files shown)


tundebabzy (changes requested):
Thanks for this @adityaduggal 

Please let's see your changes in action using animated gifs

adityaduggal:
Hi @tundebabzy 
I am currently providing screen shots since I always work with my company data and on GIF all data is visible, also how would you know if the filter for Sales partner and Sales person is working since there is NO COLUMNS for the same as its not needed.

Here is the screen shot for the AR report with Territory Filter "NORTH INDIA" which would show all the territories under NORTH INDIA since its a node territory:

image

Let me know if this suffices. Since it is showing the filter for the Sales partner and Sales Person as well.

rohitwaghchaure (approved):

PR #12629, opened 2018-01-24
Fixes issue #12512

vinayak--:
@rohitwaghchaure
Changes made as per said and re-opened the pull request https://github.com/frappe/erpnext/pull/12544.

solved-issue

Issue #12512, opened 2018-01-16
Item - Stock Entry creation permission


umairsy:
In the Item master, there is open to add stock in the Warehouse. If user is not given permission on the creation of Stock Entry, then option to add stock in the Warehouse should not work. Currently, it does show the pop-up to enter details and then shows the error message.

![paste_1516043588](https://user-images.githubusercontent.com/910238/34987365-e86500d8-fae0-11e7-888e-cf26983d2df0.png)

Jacksoneasyway:
Hi: Umair

This is Jackson
still waiting for your suggestion on above issue
How could we restrict the window of 'add' show up.

Jacksoneasyway:
for the user where i did the screen shot, I didn't even give him the permission on the stock entry document.
please support to solve this last problem, before we go-ahead for subscription.

looking forward to your comment.

Jacksoneasyway:
Hi: Guys
I really don't understand how are you Guys supporting your USER, 

No body give a word for this matter? 

what happens if we have some problem during the work after we pay?

tundebabzy:
Hi @Jacksoneasyway 
This will be fixed ASAP

tundebabzy:
#12642

PR #12642, opened 2018-01-24
Fix to issue #12512

vinayak--:
@rohitwaghchaure
Made the suggested changes and re-opened the pull request https://github.com/frappe/erpnext/pull/12629

solved-issue

### DIFF
--- a/erpnext/stock/dashboard/item_dashboard_list.html
+++ b/erpnext/stock/dashboard/item_dashboard_list.html
@@ -39,8 +39,8 @@
 					</span>
 				</span>
 			</div>
+			{% if can_write %}
 			<div class="col-sm-2 text-right" style="margin-top: 8px;">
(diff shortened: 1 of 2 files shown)


erpnext/stock/dashboard/item_dashboard_list.html
```diff
@@ -40,18 +40,20 @@
 				</span>
 			</div>
 			<div class="col-sm-2 text-right" style="margin-top: 8px;">
```
rohitwaghchaure: Add if condition ``` {% if can_write %}``` before div

gaurav-naik: If can_write is false, col-sm-2 will not render and it may disrupt the 12 column layout. Thoughts?

rohitwaghchaure (approved):

PR #12657, opened 2018-01-25
Minor - Healthcare Desktop UI

system19:
Minor - Healthcare Desktop UI
Spelling error, Links fixed

### DIFF
--- a/erpnext/config/desktop.py
+++ b/erpnext/config/desktop.py
@@ -293,7 +293,7 @@ def get_data():
 		},
         {
 			"module_name": "Patient",
-			"color": "#FF888B",
+			"color": "#6BE273",
 			"icon": "fa fa-user",
 			"doctype": "Patient",
 			"type": "link",
@@ -303,7 +303,7 @@ def get_data():
         },
         {
 			"module_name": "Patient Appointment",
-			"color": "#FF888B",
+			"color": "#934F92",
 			"icon": "fa fa-calendar-plus-o",
 			"doctype": "Patient Appointment",
 			"type": "link",
@@ -313,7 +313,7 @@ def get_data():
         },
         {
 			"module_name": "Consultation",
-			"color": "#FF888B",
+			"color": "#2ecc71",
 			"icon": "fa fa-stethoscope",
 			"doctype": "Consultation",
 			"type": "link",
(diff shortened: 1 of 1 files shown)


nabinhait (changes requested):
Any specific reason for changing the colour of the icons?

system19:
No not really. I don't actually like colour. I just returned them all to the same colour they were before the last healthcare icon related PR. I like the idea of the healthcare Module being mostly one colour though.
The main part of this PR is that I found several Icons that didn't link. Clicking them would do nothing.

raghukamath:
> No not really. I don't actually like colour. I just returned them all to the same colour they were before the last healthcare icon related PR. I like the idea of the healthcare Module being mostly one colour though.

Colour is subjective matter, some people like colours while others don't, however in this situation I guess we can keep different colour to make it a bit more consistent, because our desktop in other domains has various colours for icon background. adding color also makes it easy to spot icons and breaks the monotonous looks.

Can you please add a screenshot showing how it looks now. 

Thank you.

erpnext/config/desktop.py
```diff
@@ -283,46 +283,53 @@ def get_data():
...
-			"color": "#2ecc71",
-			"icon": "fa fa-stethoscope",
-			"doctype": "Consultation",
-			"type": "link",
-			"label": _("Consultationt")
-        },
         {
 			"module_name": "Patient",
-			"color": "#6BE273",
+			"color": "#FF888B",
```
raghukamath: Can you keep the colors as it is?

ghost (changes requested):

erpnext/config/desktop.py
```diff
@@ -283,46 +283,53 @@ def get_data():
...
+			"label": _("Patient Appointment"),
+			"hidden": 1
+        },
+        {
+			"module_name": "Consultation",
+			"color": "#FF888B",
+			"icon": "fa fa-stethoscope",
+			"doctype": "Consultation",
+			"type": "link",
+			"link": "List/Consultation",
```
raghukamath: thanks for adding the links

nabinhait:
@system19 As Raghu suggested, to keep consistency with other icons, I am merging this PR keeping the multiple colour.

system19:
Sure, No problem :-)

Issue #12658, opened 2018-01-25
Only Sales Invoice Should Try To Set Batch Number For Its Items


tundebabzy:
![peek 2018-01-25 19-13](https://user-images.githubusercontent.com/818803/35404637-f80b9798-0203-11e8-9aa1-f7f0be9a72f3.gif)

PR #12659, opened 2018-01-25
set batch number only in Sales Invoice and Delivery Note

tundebabzy:
### Before
peek 2018-01-25 19-13

### After
peek 2018-01-25 19-26


tundebabzy:
Reference #12658

PR #12670, opened 2018-01-26
Report presentation currency

tundebabzy:
This PR makes it easy for those operating multiple currencies to make sense out of the reports. It follows IAS 21standards too.

### Present General Ledger Report
screenshot-2018-1-25 general ledger

Let's filter by a USD denominated account and use the "Print In Account Currency"
screenshot-2018-1-25 general ledger 1

However, when the report contains more than one currency, things start getting hard
screenshot-2018-1-25 general ledger 2

And it's the same for the financial statement reports.

### New General Ledger Report
There will now be an option to pick a presentation currency from the filters.
screenshot from 2018-01-26 23-23-13

screenshot-2018-1-26 general ledger

So if you need to view your reports in a foreign currency, simply change the currency
screenshot-2018-1-26 general ledger 1

..or you run a business and source all your raw materials from abroad. You will probably want to see what your account looks like in the foreign currency...
screenshot-2018-1-26 profit and loss statement

screenshot-2018-1-27 profit and loss statement

screenshot-2018-1-27 balance sheet

screenshot-2018-1-27 balance sheet 1

Notes:
- You can use any currency you like as long as you have saved the exchange rate or it's one of the supported currencies in fixer.io. 
- IAS 21 standard is respected when you select any currency that is not the functional currency.


tundebabzy:
Yet to test if printing works properly due to #12671. frappe/#frappe#4930 needs to be merged

tundebabzy:
Print currencies work as expected
### USD
screenshot-2018-1-31 general ledger

### NGN
screenshot-2018-1-31 general ledger 1

### PDF
screenshot from 2018-01-31 13-11-02

tundebabzy:
@nabinhait 
Ready for final review

Issue #12535, opened 2018-01-18
Website Cart Should Not Show Expired Items As 'In Stock'


tundebabzy:
### Web page 
![screenshot-2018-1-18 diced pork](https://user-images.githubusercontent.com/818803/35093715-77551ffe-fc43-11e7-9df5-1403717af4b6.png)

### Batch 
![screenshot-2018-1-18 batch](https://user-images.githubusercontent.com/818803/35093734-82ab1214-fc43-11e7-8b72-f05a9c38d322.png)

![screenshot-2018-1-18 diced pork - 00002 1](https://user-images.githubusercontent.com/818803/35093806-b5c469a2-fc43-11e7-95c0-82cd8c2d8394.png)

tundebabzy:
#12677

PR #12677, opened 2018-01-27
Website Cart Should Not Show Expired Items As 'In Stock' #12535

tundebabzy:
Reference #12535 

Initially, the website used to include expired items. This PR fixes that.

screenshot-2018-1-27 diced pork - 00002

### Before
screenshot-2018-1-27 diced pork

screenshot-2018-1-27 products

### After
screenshot-2018-1-27 diced pork 1

screenshot-2018-1-27 products 1


erpnext/utilities/product.py
```diff
@@ -4,25 +4,70 @@
...
+	if stock_qty[0][0]:
+		in_stock = stock_qty[0][0] > 0 and 1 or 0
+
+	if stock_qty:
 			in_stock = stock_qty[0][0] > 0 and 1 or 0
 
 	return frappe._dict({"in_stock": in_stock, "stock_qty": stock_qty, "is_stock_item": is_stock_item})
 
+
+def adjust_for_expired_items(item_code, stock_qty, warehouse):
```
saurabh6790: `adjust_qty_for_expired_items`  maybe more specific ?

erpnext/utilities/product.py
```diff
@@ -4,25 +4,70 @@
...
+	batches = frappe.get_list('Batch', filters=[{'item': item_code}], fields=['expiry_date', 'name'])
+	expired_batches = get_expired_batches(batches)
+	stock_qty = [list(item) for item in stock_qty]
+	
+	if expired_batches:
```
saurabh6790: I think no need to check `if expired_batches` as `get_expired_batches` will return the empty list if expired batch not found

erpnext/setup/doctype/item_group/item_group.py
```diff
@@ -104,8 +106,26 @@ def get_product_list_for_group(product_group=None, start=0, limit=10, search=Non
 
 	data = frappe.db.sql(query, {"product_group": product_group,"search": search, "today": nowdate()}, as_dict=1)
 
+	data = adjust_for_expired_items(data)
+
 	return [get_item_for_list_in_html(r) for r in data]
 
+
+def adjust_for_expired_items(data):
```
saurabh6790: `adjust_qty_for_expired_items` maybe more specific ?

saurabh6790 (changes requested):

erpnext/utilities/product.py
tundebabzy: yeah

tundebabzy:
@saurabh6790 
fixed

saurabh6790:
@tundebabzy Also remove `frappe.has_permission('Batch', throw=True)` from batch.py

tundebabzy:
@saurabh6790 done

saurabh6790 (approved):

erpnext/utilities/product.py
```diff
@@ -4,25 +4,70 @@
...
+
 	if not warehouse and template_item_code and template_item_code != item_code:
 		warehouse = frappe.db.get_value("Item", template_item_code, item_warehouse_field)
 
 	if warehouse:
 		stock_qty = frappe.db.sql("""select GREATEST(actual_qty - reserved_qty, 0) from tabBin where
 			item_code=%s and warehouse=%s""", (item_code, warehouse))
-		if stock_qty:
+
+	if stock_qty[0][0]:
```
saurabh6790: If warehouse is not found @L21 then stock_qty will remain 0 and flow will break for stock_qty[0][0]

```
  File "<stdin>", line 1, in <module>
TypeError: 'int' object has no attribute '__getitem__'
```

so just check for stock_qty instead of stock_qty[0][0]

saurabh6790 (changes requested):

tundebabzy:
@saurabh6790 
Fixed

saurabh6790 (approved):

PR #12697, opened 2018-01-29
Trial Balance sort by Account Name

Helkyd:
erpnext_trial_balance_sort1

Fixing Trial Balance report.


rohitwaghchaure:
@Helkyd 
The accounts order is not showing properly
screen shot 2018-02-01 at 6 50 56 pm

Helkyd:
@rohitwaghchaure 
The issue is solved ... but only happens to account_name starting with numbers.
Will PR for your review.

PR #12700, opened 2018-01-29
GSTR2 report

vishdha:
Gstr2 Report with Tax Amount Details
screenshot from 2018-01-29 23-03-15

### DIFF
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -491,3 +491,4 @@ erpnext.patches.v10_0.update_assessment_result
 erpnext.patches.v10_0.added_extra_gst_custom_field
 erpnext.patches.v10_0.workflow_leave_application #2018-01-24
 erpnext.patches.v10_0.set_default_payment_terms_based_on_company
+erpnext.patches.v10_0.added_extra_gst_custom_field_in_gstr2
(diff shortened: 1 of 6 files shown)


snvora:
Will GSTR 1 and 2 report work where people are using "charges" like freight charges/other charges, in Taxes and charges template?

vishdha:
Hi @snvora, I don't think so, taxable value for gstr calculated based on cgst, sgst and igst. other taxes not included in GST tax.

deepeshgarg007:
@vishdha I am unable to understand the use of these fields. 

image

Are they only used to see the tax amounts in GSTR-2 Report? Then it could have been captured by "Purchase Taxes and Charges Table" as well". Can you explain the use case in which these fields are used

test2a:
@deepeshgarg007 from what i could understand, these are useful in cases for example in a seller of both taxable as well as non taxable items, ITC of capital goods are to be taken based on a formula, like over 5 years and all that. 
I think the boxes mean you can have a total ITC of say Rs. 100,000.00 in a purchase invoice but after doing the calculation, you will only avail 10% or Rs. 10,000.00 this time around

deepeshgarg007:
@test2a So in the "Purchase Taxes and Charges" table taxes for only taxable items will be there, tax amount for non taxable items will be anyways zero and since this is captured only in GSTR-2 report these details can be captured from "Purchase Taxes and Charges" Table. My question is why do we need separate fields for this purpose.

PR #12741, opened 2018-02-01
Update Trial balance

Helkyd:
Fix to order only if account starts with Number...
update_trial_balance_order_1


erpnext/accounts/report/trial_balance/trial_balance.py
```diff
@@ -52,7 +52,7 @@ def validate_filters(filters):
 
 def get_data(filters):
 	accounts = frappe.db.sql("""select name, parent_account, account_name, root_type, report_type, lft, rgt
-		from `tabAccount` where company=%s order by lft""", filters.company, as_dict=True)
+		from `tabAccount` where company=%s order by account_name, lft""", filters.company, as_dict=True)
```
rohitwaghchaure: @Helkyd 
account_number field is available in the account, get the account_number from the above query and use that to check it has account_number instead of if accounts[0].account_name[0].isnumeric():

Helkyd: @rohitwaghchaure Assuming users will fill the account_number.

PR #12753, opened 2018-02-03
Enhancement and bug fix in batch

sivankar:
### Enhancement
added search by manufacturing date while selecting a batch, as many of the products have manufacturing date printed on the product with shelf life, so it would be easy for an user to search batch by manufacturing date.
Example:
batch with manufacturing date

added support of jinja template so in stock settings batch naming series one can use jinja template to create batch with item name
Stock Setting:
stock setting
Batch is created as:
batch

### Bug Fix
In `serial_no_batch_selector.js` renamed item_code to item as `erpnext.controllers.queries.get_batch_numbers` expects item because we refer items in batch by `item` and not `item_code` so to maintain the naming convention renamed it to item.

Added a validation if item is configured with `has expiry date` then do not allow batch without expiry date.


erpnext/stock/doctype/batch/batch.js
```diff
@@ -25,9 +25,17 @@ frappe.ui.form.on('Batch', {
 		}
 	},
 	item: (frm) => {
-		frappe.db.get_value('Item', {name: frm.doc.item}, 'has_expiry_date', (r) => {
-			frm.toggle_reqd('expiry_date', r.has_expiry_date);
-		});
+		// frappe.db.get_value('Item', {name: frm.doc.item}, 'has_expiry_date', (r) => {
```
tundebabzy: You can delete this since it's not valuable

erpnext/stock/doctype/batch/batch.py
```diff
@@ -120,6 +110,28 @@ def item_has_batch_enabled(self):
...
+	def before_save(self):
+		has_expiry_date, shelf_life_in_days = frappe.db.get_value('Item', self.item, ['has_expiry_date', 'shelf_life_in_days'])
+		if not self.expiry_date and has_expiry_date and shelf_life_in_days:
+			self.expiry_date = add_days(self.manufacturing_date, shelf_life_in_days)
+
+		if has_expiry_date and not self.expiry_date:
+			frappe.throw('Expiry date is mandatory for selected item')
+			frappe.msgprint('Set items shelf life in days, to set expiry based on manufacturing_date plus self life ')
+
+	def get_name_from_naming_series(self):
```
tundebabzy: `get_name_from_naming_series` was actual written away from the object so it can be reusable. Is there any other reason why you decided to turn it to a method?

tundebabzy (changes requested):

tundebabzy:
+1 for first contribution.
+1 for small commits at a time
+1 for descriptive commit messages

erpnext/stock/doctype/batch/batch.py
sivankar: that is because, i have added support for jinja, user can refer to any attribute it would be available in self. that is why added to class method.

erpnext/controllers/queries.py
```diff
@@ -270,13 +270,14 @@ def get_batch_no(doctype, txt, searchfield, start, page_len, filters):
 	}
 
 	if args.get('warehouse'):
-		batch_nos = frappe.db.sql("""select sle.batch_no, round(sum(sle.actual_qty),2), sle.stock_uom, batch.expiry_date
+		batch_nos = frappe.db.sql("""select sle.batch_no, round(sum(sle.actual_qty),2), sle.stock_uom, concat('MFG-',batch.manufacturing_date), concat('EXP-',batch.expiry_date)
```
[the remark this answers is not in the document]
sivankar: how do we know which date is manufacturing and which one is expiry?

erpnext/stock/doctype/batch/batch.py
tundebabzy: cool

PR #11054, opened 2017-10-04
Attendance check in

shridarpatil:
Activate checkin / checkout in attendance control panel / HR settings.
On login, if not checked in, show popup to allow checkin

https://github.com/frappe/erpnext/issues/428

Issue #428, opened 2013-06-04
[HR] Checkin and Checkout buttons for Attendance


rmehta:
Add check-in and check-out buttons for Attendance control.
- [ ] On login, if not checked in, show popup to allow checkin
- [ ] On logout, show popup to confirm checkout
- [ ] Activate checkin / checkout in attendance control panel / HR settings.

gohyc:
Just wondering anyone working on this enhancement?  We have a requirement to use this function now.

cgpurbaugh:
This does look like it would be a good feature to have.

tareq07:
i need feature.

rmehta:
This system can be easily gamed (Employee can easily login from anywhere to get attendance marked). It is too cumbersome to track this via IP addresses.

centenojeremyjc:
Hi folks, has this feature already been implemented?

tundebabzy:
It's a wontfix

rmehta:
@tundebabzy re-opening, so people can upvote!

indolering:
> This system can be easily gamed (Employee can easily login from anywhere to get attendance marked). It is too cumbersome to track this via IP addresses.

There are any number of ways to manipulate time-sheets (buddy punching being the most common) and anyone caught doing this would be fired.

I'm interested in building the front-end for this functionality if ERPNext is capable of tracking time by minutes instead of daily attendance.  @rmehta know of any seasoned developers willing to collaborate with me on this?

wojosc:
@rmehta this is surely something that has now been updated in v12. I'm not sure which of your initial checks are now implemented. Please provide some info on the status and maybe link to documentation. I'm happy to do some testing on this.

kennethsequeira:
Closing this the base requirement is essentially satisfied by the check-in checkout doctypes + device syncing feature.

PR #12755, opened 2018-02-03
Mark attendance on login.

shridarpatil:
Activate checkin / checkout in attendance control panel / HR settings.
On login, if not checked in, show popup to allow checkin

Closed PR: https://github.com/frappe/erpnext/pull/11054
issue: https://github.com/frappe/erpnext/issues/428


shridarpatil:
I think the build is getting failed because of this patch https://github.com/frappe/erpnext/blob/develop/erpnext/patches.txt#L492


image

erpnext/hr/doctype/attendance/attendance.py
```diff
@@ -54,3 +54,61 @@ def validate(self):
 		self.validate_attendance_date()
 		self.validate_duplicate_record()
 		self.check_leave_record()
+
+
+def employee():
```
nabinhait: `employee` is not a good method name, `get_employee()` ???

erpnext/public/js/checkin_checkout_attendance.js
```diff
@@ -0,0 +1,39 @@
...
+						show_popup();
+					}
+				}
+			});
+		}
+	});
+
+
+function show_popup() {
+	var dialog = new frappe.ui.Dialog({
```
nabinhait: If user does not logged out till the next date, what will happen? May be check the server time every few minutes and if day changes, show the popup again.

nabinhait (changes requested):
- Add few screenshots
- Add documentation

nabinhait:
Closing this for now, please reopen after fixing.

PR #12805, opened 2018-02-07
docs-fixes

elba7r:


elba7r:
this pull request for resolving the images that does not appear in the documentation due to wrong src path for the images

PR #12831, opened 2018-02-08
Add Total Qty to online POS

PawanMeh:
Testing video
pos testing


erpnext/accounts/doctype/sales_invoice/sales_invoice.json
```diff
@@ -4579,6 +4579,36 @@
...
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Total Qty", 
+   "length": 0, 
+   "no_copy": 0, 
+   "permlevel": 0, 
+   "precision": "", 
+   "print_hide": 0, 
+   "print_hide_if_no_value": 1, 
+   "read_only": 0,
```
rohitwaghchaure: can you make it read_only true

erpnext/accounts/doctype/sales_invoice/sales_invoice.json
```diff
@@ -4579,6 +4579,36 @@
...
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Total Qty", 
+   "length": 0, 
+   "no_copy": 0, 
+   "permlevel": 0, 
+   "precision": "", 
+   "print_hide": 0,
```
rohitwaghchaure: Enable print hide, those required they will disable using Customize Form

erpnext/public/css/pos.css
```diff
@@ -174,3 +174,6 @@ input[type=number]::-webkit-outer-spin-button {
 .rounded-total-value {
   font-size: 18px;
 }
+.quantity-total {
+  font-size: 18px;
```
rohitwaghchaure: add in pos.less, not in pos.css

PawanMeh: Made suggested changes to sales_invoice.json, however when I add similar code to css.less I do not get similar styling and font for Total Qty. Refer screen shot below
total_qty

PawanMeh:
Thanks for the review, requested changes have been made

erpnext/public/css/pos.css
```diff
@@ -1,176 +0,0 @@
...
-}
-.grand-total .list-item {
-  height: 60px;
-}
-.grand-total .grand-total-value {
-  font-size: 18px;
-}
-.rounded-total-value {
-  font-size: 18px;
-}
```
rohitwaghchaure: Why removed?

PawanMeh: removed this since it was added in pos.less and when we compile it automatically adds it to pos.css, let me know if I am missing something here.

rohitwaghchaure (approved):

PR #12833, opened 2018-02-08
fix the status updater for multiple sources

manassolanki:


erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -232,12 +231,17 @@ def on_update(self):
...
 			'target_field': 'ordered_qty',
-			"join_field": "sales_order_item",
-			"target_parent_field": ''
+			'target_parent_dt': 'Sales Order',
+			'target_parent_field': '',
+			'join_field': 'sales_order_item',
+			'source_dt': 'Purchase Order Item',
+			'target_ref_field': 'stock_qty',
+			'source_field': 'stock_qty',
+			'percent_join_field': 'material_request'
```
nabinhait: should be `sales_order`

nabinhait (changes requested):
Add few test cases

nabinhait:
closing this, reopen after adding tests

PR #12887, opened 2018-02-13
Remove total_projected_qty

netchampfaris:


nabinhait:
@netchampfaris send the PR to hotfix

PR #12902, opened 2018-02-13
[ Enhance ] Production to Work Order

Zlash65:
1. Changed name from Production Order to Work Order
2. Removed hard coded occurrences
3. Reports updated accordingly.
4. Patch added to rename and reload and shift update data.
5. Naming series changed from PRO- to WO- 

On a data of 1400+ entries, patch took roughly 1 minute to execute.


dottenbr:
Can I ask why you would want to change the name from "production order" to "work order"? This seems counter-intuitive since all ERP systems call it "production order".

rmehta:
@dottenbr If you want to use ERPNext for Project based execution, the name "Production Order" is not correct. By making it a more generic "Work Order", it can be used for construction and other project based planning.

With this, we will also allow the Work Order to be created from Tasks.

dufani1:
Nice idea... similar the following

https://docs.oracle.com/cd/E59116_01/doc.94/e58795/create_work_ord_proj.htm#WEAWO163

erpnext/patches/v11_0/rename_production_order_to_work_order.py
```diff
@@ -0,0 +1,29 @@
...
+    frappe.reload_doc('manufacturing', 'doctype', 'work_order_item')
+
+    rename_doc('DocType', 'Production Order Operation', 'Work Order Operation', force=True)
+    frappe.reload_doc('manufacturing', 'doctype', 'work_order_operation')
+
+    frappe.reload_doc('projects', 'doctype', 'timesheet')
+    frappe.reload_doc('stock', 'doctype', 'stock_entry')
+
+    # fetch all linked fields
+    old, new = 'Production Order', 'Work Order'
```
nabinhait: Use rename_field for this

nabinhait (dismissed):

Issue #12874, opened 2018-02-12
Invoice Due Date calculated from posted date instead of Supplier Invoice Date


dboobisMF:
The due date for Purchase Invoices is being calculated from the posted date of the document and not the invoice date. In the Payment Terms doctype, the wording for "Due Date Based On" is precisely "Days After Invoice Date" but the current behaviour does not do this.

There is scope to incorporate the current behaviour as an additional drop down option, but primarily this is a bug and needs to be fixed.

Given a supplier invoice has a document date
And payment terms are net of this date
When entering the invoice into the ERPNext system
Then the automatically calculated due date should be based on the invoice date

umairsy:
Hi,

By invoice date, are you referring to Supplier Invoice Date? If yes, it cannot be made to work as expected. The Due Date will be set based on the Posting Date of the invoice only. The Supplier's Invoice Date is added there just for the reference and not considered in the calculation of due date. If needed, you can create Post Dated Purchase Invoice as well by amending Posting Date of the invoice.

dboobisMF:
This is unacceptable, that's not how international accounting works. If the system can't handle setting the due date off the supplier invoice date then the system is broken. 

The accruals principle requires that you can post invoices into a different period than they are dated for. It's a fundamental tenet of International Accounting Standards. For the user to have to manually calculate the due date for every invoice is unacceptable.

How would I best go about getting this resolved?

umairsy:
Can you please mention which field are you referring to from Invoice Date? Is it Supplier Invoice Date?

dboobisMF:
Yes, supplier invoice date is the field.

nabinhait:
@dboobisMF I agree that due date should be calculated based on supplier invoice date instead of posting date. We will make necessary changes.

PR #12913, opened 2018-02-14
Calculate due date in Purchase Invoice on the basis of Supplier invoice date 

shreyashah115:
#12874 
Summary:

- Calculate due date on the basis of Supplier invoice date i.e. `bill_date` if present else do the calculations on the basis of posting date.
- Similar calculations for due dates in payment_schedule table in Purchase invoice.
- If due date calculated on the basis of `bill_date` is less than `posting_date` then set `due_date` = `posting_date`.

due_date_in_pi


manasan3010:
`If due date calculated on the basis of bill_date is less than posting_date then set due_date = posting_date`

For what use case is this logic used for?

PR #12948, opened 2018-02-15
[new] WIP Leader board Analysis

vishdha:
WIP: Leader board analysis

ls

### DIFF
--- a/erpnext/utilities/page/leaderboard/leaderboard.js
+++ b/erpnext/utilities/page/leaderboard/leaderboard.js
@@ -17,7 +17,7 @@ frappe.Leaderboard = Class.extend({
 		this.$sidebar_list = this.page.sidebar.find('ul');
 
 		// const list of doctypes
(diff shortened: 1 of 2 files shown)


erpnext/utilities/page/leaderboard/leaderboard.js
```diff
@@ -141,7 +142,7 @@ frappe.Leaderboard = Class.extend({
 						],
 						labels: graph_items.map(d=>d.name)
 					},
-					colors: ['light-green'],
+					colors: ['orange'],
```
rmehta: green is success, orange is warning - lets go with green 👍

rmehta (changes requested):

erpnext/utilities/page/leaderboard/leaderboard.js
vishdha: :+1:

PR #12981, opened 2018-02-17
Sales Payment Summary - fix for change amount

PawanMeh:


erpnext/accounts/report/sales_payment_summary/sales_payment_summary.py
```diff
@@ -118,7 +118,7 @@ def get_mode_of_payment_details(filters):
 	invoice_list_names = ",".join(['"' + invoice['name'] + '"' for invoice in invoice_list])
 	if invoice_list:
 		inv_mop_detail = frappe.db.sql("""select a.owner, a.posting_date,
-			ifnull(b.mode_of_payment, '') as mode_of_payment, sum(b.base_amount) as paid_amount
+			ifnull(b.mode_of_payment, '') as mode_of_payment, sum(b.base_amount - a.base_change_amount) as paid_amount
```
nabinhait: As `base_change_amount` is part of parent, if there are multiple mode_of_payments, then it will be deducted multiple times, which is wrong.

nabinhait (changes requested):

erpnext/accounts/report/sales_payment_summary/sales_payment_summary.py
PawanMeh: currently in POS invoices, is it possible to have multiple mode of payments?

olamide2: Yes it is possible to have multiple mode of payments

PawanMeh: @olamide2, in case of multiple mode of payments for one transaction, we would be assuming that the change amount needs to be adjusted against the cash transaction only or in other words is it fair to assume, that change amount is applicable to Cash Transactions only?

olamide2: Yes you are right  ..change is applicable to cash alone

PawanMeh: Change has been made to apply change amount adjustment to Cash mode of payment.

olamide2:
Hello @PawanMeh 

We just noticed that the report does not allow for filtering by cost center. This was one of the requirements stated from the beginning, would you mind doing an update for this?

Regards

PawanMeh:
Hi Olamide,

As replied to you on discuss thread, since the cost center is at Item level and mode of payment is not maintained at the same level, it does not seem technically feasible to add it to this report. Hence it was added filters were added to the sales register report via https://github.com/frappe/erpnext/pull/11600 so the any such required information can be extracted out from there. 

Thanks,
Pawan

PR #12985, opened 2018-02-17
Code cleanup online POS

rohitwaghchaure:
- savesubmit() method to submit invoice and removed python code

- Removed set_value

Many users has reported that system has submitted sales invoice automatically, therefore have made changes in the code of submission of sales invoice

Dependent PR https://github.com/frappe/frappe/pull/5028

### DIFF
--- a/erpnext/accounts/doctype/pos_profile/pos_profile.py
+++ b/erpnext/accounts/doctype/pos_profile/pos_profile.py
@@ -32,7 +32,7 @@ class POSProfile(Document):
 					.format(res[0][0], row.user), raise_exception=1)
 			elif not row.default and not res:
 				msgprint(_("User {0} doesn't have any default POS Profile. Check Default at Row {1} for this User.")
-					.format(row.user, row.idx), raise_exception=1)
+					.format(row.user, row.idx))
 
 	def validate_all_link_fields(self):
 		accounts = {"Account": [self.income_account,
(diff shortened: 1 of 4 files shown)


olamide2:
Thanks Rohit

Is it possible to see a gif of the new flow if there is a change in workflow?

olamide2:
While we are on code clean-up is it possible to have a look at the following issues 

. For some reasons cashiers can give discounts on offline mode even if the ¨do not give discount option is turned on in PoS profile, this is a bug

4. Need an option in the PoS profile that will prevent a cashier from submitting an invoice if payment via payment methods is not equal to the invoice total. So if invoice total is 10,000 then the sum of payment methods must be 10,000 else submit button should be grayed out.

5. Once the cashier clicks on submit and confirms the submission.  the PoS should automatically submit and print the receipt and open a new invoice. This should bypass the ¨Print New"option button. in effect we are asking that the submit, print and new buttons should become one. Right now cashiers have to click 5 times to complete a transaction  and open a new invoice when 2 is all that's needed.

rmehta:
Proof? Animated GIF?

olamide2:
@rmehta is that request meant for me or @rohit?

Regards

rohitwaghchaure:
### For non serialized items 
non_serialized

### For serialized items 
serialized_items


### For batch items
batch_items


### For barcode items
barcode_items

Olawale1:
Hi @rohitwaghchaure 

What of offline POS? I'm not sure what you mean by 'system has submitted sales invoice automatically' but the issue we're experiencing is simply that the ERP is creating multiple Sales Invoices for a single transaction

This is happening in both Offline and Online POS. Please clarify is this is meant to address the issue in both options of POS

Thanks

PR #13008, opened 2018-02-20
Gocardless integration

chdecultot:
Hi,

This PR is linked to a Frappe PR for the integration of GoCardless in ERPNext.

GoCardless allows companies within the SEPA zone to collect wire transfers easily without having to process it manually with the bank or through a protocol like EBICS.

Since a SEPA payment can be made directly without a user approval if a mandate is valid, I added a method in the payment request to determine if an email should be send or  upon submission.

Thanks in advance for your help merging this PR.

EDIT: Linked with 5038 in Frappe


kennethsequeira:
@chdecultot are the docs added for the same?

PR #13010, opened 2018-02-20
Add total row in all trends reports

shreyashah115:


nabinhait:
@shreyashah115 The changes is applicable for all the `trends` report.

Issue #12286, opened 2018-01-01
BOM update tool does not update Grandparent's Exploded BOM


djpalshikar:
![screenshot-2018-1-1 bom update tool](https://user-images.githubusercontent.com/23133635/34468180-7a9bd634-ef29-11e7-887c-9aed447a720c.png)

The BOM update tool only updates all BOMs where Current BOM is explicitly mentioned. It does not trigger a change further up above the Grandparent.

**EDIT**: Also while Great Grand Parents Tree when expanded gets updated, the exploded BOM table does not get updated.

I don't see any valid use case for this behaviour.

ERPNext v10.0.4 (master)
Frappe Framework: v10.0.5 (master)

umairsy:
Didn't replicate it in my test build. Closing issue for now.

umairsy:
As received from Dhananjay:

I have edited the issue. There was a mistake I made. The problem occurs with the Great-Grand-Parent's BOM.

Please consider re-opening the issue. If you can access Sapcon's account, I have created a BOM for testing to show you the same: BOM-GGP-001.

djpalshikar:
@rohitwaghchaure can you confirm if this has been released to Frappe hosted accounts

PR #13026, opened 2018-02-21
[Fix] BOM Update Tool not update grandparent's exploded BOM

rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/issues/12286


djpalshikar:
@rohitwaghchaure could you confirm if this PR has made to the release ERPNext v10.1.15 
I am on Frappe cloud

PR #13043, opened 2018-02-22
Provision to include exploded items while making transfer entry of raw materials for subcontrtacted item

### LINKED ISSUE #11967: [Paid Development] Subcontracting with Multi-Level BOM does not fetch Exploded BOM
If a subcontracted Item has a multilevel BOM, the "Raw Materials Supplied" table is not populated with Multi-Level BOM. 
subcontracting

I suggest there should be  a **new Check Box** in the Items Table  with the label "Use Multi Level BOM for Material Transfer". This should fetch the "Exploded BOM table" in the Purchase Order.

Since this is essential for us, I would like to pay for the feature.

rohitwaghchaure:
Added provision to Include Exploded Items in the purchase order item for subcontracted item
screen shot 2018-02-22 at 3 02 07 pm

Fixed https://github.com/frappe/erpnext/issues/11967

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -195,7 +195,11 @@ class BuyingController(StockController):
 			self.set('supplied_items', [])
 
 	def update_raw_materials_supplied(self, item, raw_material_table):
-		bom_items = self.get_items_from_bom(item.item_code, item.bom)
+		exploded_item = 1
+		if hasattr(item, 'include_exploded_items'):
+			exploded_item = item.get('include_exploded_items')
+
+		bom_items = get_items_from_bom(item.item_code, item.bom, exploded_item)
 		raw_materials_cost = 0
 		items = list(set([d.item_code for d in bom_items]))
 		item_wh = frappe._dict(frappe.db.sql("""select item_code, default_warehouse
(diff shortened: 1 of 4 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -174,6 +174,7 @@ def validate_for_subcontracting(self):
 					item.bom = None
 
 	def create_raw_materials_supplied(self, raw_material_table):
+		self.set(raw_material_table, [])
```
nabinhait: Don't make it blank every time. Because user enters serial nos and batch nos in the raw material table, those will get blank.

erpnext/patches/v11_0/enable_include_exploded_items.py
```diff
@@ -0,0 +1,17 @@
...
+import frappe
+
+def execute():
+	doctypes = {'Purchase Order': 'Buying', 'Purchase Invoice': 'Accounts', 'Purchase Receipt': 'Stock'}
+
+	for doctype, module in doctypes.items():
+		child_doc = doctype + ' Item'
+		frappe.reload_doc(frappe.scrub(module), 'doctype', frappe.scrub(child_doc))
+
+		frappe.db.sql(""" update `tab{child_doc}` as child_doc, `tab{parent_doc}` as parent_doc
```
nabinhait: Why `include_exploded_items` should be set as 1 in existing transactions. Previously it was taking raw materials from BOM Item, so was not considering exploded items.

nabinhait (changes requested):

erpnext/patches/v11_0/enable_include_exploded_items.py
rohitwaghchaure: Removed patch!

Issue #12150, opened 2017-12-23
[Feature Request][Paid Development] Material Transfer for Subcontracting should not be restricted as per BOM


djpalshikar:
We give out Subcontracting Jobs for machining raw materials items like Steel Rods which have same operations but have different lengths. 
E.g.
Subcontracted item: SR-L**X**-MCN 
Raw Material item: SR-RAW , Qty varies as per **X** 

As of now we have to create a number of items everyday(around 20-30) and their BOMs for awarding subcontracting orders for each possible item.
Proposed Solution:
 

- A generic practice can be followed:
A Single Subcontracted Item: SR-L-CMN 
Single BOM: SR-RAW  where Qty could be the default(most common) required qty.

- From here on the transfer of Material should **not** be restricted by the BOM.
- Backflush of Raw Materials should also be based on actual stock entries made similar to the Production Order.

Would like to hear if someone has a better suggestion and is useful to others. 
I am willing to sponsor.

djpalshikar:
#8540 may affect the issue.

nabinhait:
@djpalshikar As of now, the design seems ok. We can refine it more on the pull requests. And you are correct, #8540 will affect this, which is already under development and there is a open pull request.

epochc:
How about we change the Sub Contracting feature to make this possible:

User Clicks Supply Raw materials in PO.

Clicking Supply Raw Materials will serve a different Sub Contracting PO format to user.

Now user has two options clicking (this is actually a Purchase Order Item field, it works so that if you use one option, that option gets auto-populated in each PO Line item - Like Default Source/Destination Warehouse, with the only exception that user can change the option for each line item. Let's say there are 10 items in the PO, 9 items have a definitive relationship in consumption of raw material, one doesn't. Then the user picks B, but just goes to the specific line item and changes that to B):

A. There is a definite relationship between the Item(s) I receive back from the Supplier and the Raw Materials issued to the Supplier
B. There are manufacturing variabilities and the quantities of raw materials consumed varies

Then the PO template has the following options:

The user has to specify for each line item that is on the PO, the raw material(s) that is/are needed to make that item and the quantities. If the user has a BOM ready for this relationship, the user picks the BOM. If there is no BOM, the user enters in the raw materials and the estimated quantities required.

Then the user saves and submits the PO. The Supplier Warehouse can be specified in the PO, in fact it can be made as a part of the Supplier Master and this can be auto populated in transactions.

[...]

PR #13051, opened 2018-02-22
[Feature] Supplier warehouse link field on purchase order

harsh0214:
Development based on https://github.com/frappe/erpnext/issues/12150  
Changes Made:
1. Supplier warehouse link field on purchase order for sub-contracting.
2. Auto select supplier warehouse while making purchase receipt from purchase order.

po1


rohitwaghchaure:
@harsh0214 
While making transfer entry for raw materials of subcontracted item from purchase order, set default to warehouse in stock entry as the supplier warehouse which is defined in the purchase order.

harsh0214:
Have made changes suggested by @rohitwaghchaure 

sub

rohitwaghchaure (approved):

Issue #13048, opened 2018-02-22
Member ID should be set on Member's Name of Naming Series


umairsy:
Currently, Member's ID is generated based on Email ID of the member. However, there could be a member who won't have email id.

- [x] Member naming based on Naming Series

kothagunda:
Very useful for farmer members who don't have emails

**Will there still be any mandatory link from member to user doctype?**

Currently when a member is chosen as party type in a JE, the email shows up as the party.  And we can filter GL entry table based on email. 
Now i guess with this change we should be able to filter GL entry using the naming series, right?
Find attached the current screenshots during JE and GL entry. 

![je](https://user-images.githubusercontent.com/30433779/36656398-3e9b0088-1aee-11e8-9fbc-1cf7f00b5f3c.png)
![gl](https://user-images.githubusercontent.com/30433779/36656402-442ee0a0-1aee-11e8-8c95-c272d8ceada9.png)

PR #13056, opened 2018-02-23
Naming in Member DocType on the basis of naming series

shreyashah115:
#13048 

naming series in mmeber


umairsy:
Will Naming Series for this doctype be configurable from Naming Series tool?

shreyashah115:
@umairsy Yep, it will be.

PR #13068, opened 2018-02-23
[Fix] Optimize update Territory/Customer_Group patch

Zlash65:


erpnext/patches/v10_0/update_territory_and_customer_group.py
```diff
@@ -1,13 +1,13 @@
 import frappe
-from frappe.model.rename_doc import update_linked_doctypes
+from frappe.model.rename_doc import update_linked_doctypes, get_fetch_fields
 
 def execute():
-	customers = frappe.get_all('Customer')
+	customers = frappe.get_all('Customer', fields={"name", "territory", "customer_group"})
```
nabinhait: should be ["name", "territory", "customer_group"]?

nabinhait (dismissed):

PR #13113, opened 2018-02-27
[Hotfix] Shopping cart order items by modified

shreyashah115:


britlog:
@shreyashah115 Hi, could you please tell me why ? I changed it to order product list by item name because it is much more logical.
For customer, it is stranged to get the list ordered by modified date, this date is a desk side field for system user, not website user. When a customer searches for an item in a list, he naturally searches by name...
Thanks

PR #12379, opened 2018-01-08
WIP: Alternative BOM Lines

kickapoo:
This the initial Add Alternative Item under Manufacturing Section in Item Doctype.
 It based on community design proposal by newmatik team. (https://discuss.erpnext.com/t/new-feature-bom-components-alternative/31915)

Please help me to make the child table more readable to end users:
https://discuss.erpnext.com/t/contribution-help-for-alternative-bom-lines-change-link-field-display-look/32444

screenshot 2018-01-12 10 45 42

screenshot 2018-01-12 10 45 17

screenshot 2018-01-12 10 44 09

TODO for [Level 1 - Data entry setup]

- [x] An alternative can be a “Two-way” or a “one-way” type. “Two-way” alternative means the parent component is also an alternative of the set alternative items. If it is a “One-way” type, the parent Item is not an alternative of the set alternative item.  A script must be added in Item Doctype "On Save" to handle this.
- [x] Add Naming Series
- [x] Sync List deleting with Two-Way
- [x] Change Table Naming following the Frappe Guidelines.
- [ ] Unit Tests @creamdory 

TODO for [Level 2 - "BOM" Alternative Set-up]

- [ ] If Item select check if Alternative exists and denote row
- [ ] (not final) Set option in bom "with Alternative" so the above validation is valid
- [ ]  List alternative item options in child table

PR #13121, opened 2018-02-27
[enhance] Item Alternative

### LINKED ISSUE #10499: Alternate/Substitute Line Items in BOM for sub contract manufacturing cycle (Paid)
Please let me know the quotation for the said paid feature.
 
Use case is for alternate line items in subcontracting in manufacturing 

Alternate/Substitute Line Item is an item which is alternate/substitute aginst a particular line item for a BOM (based on application/end usage).

Example : BOM (Finished Good) = FG PC-Corei5, Monitor 18.5inch
BOM Line Items - qty 
1) Intel Processor Corei5- 1
2) Monitor Samsung 18.5 (Alternate Items - LG, Dell ) 
2) Keyboard - HP (Alternate items TVS Gold, Microsoft) 
3) Mouse Optical -Logitech ( Alternate Items iBall, Dell)

While adding line items in BOM we need to add alternate line items against a particular line item and that group of alternate should be linked by BOM Line Item Number# (e.g. BL2), but its effect should be handled across all relevant doctypes, reports, modules for e.g.  BOM costing ,BOM Browser, Purchase Order, Material transfer for sub contracting, Purchase Receipt 

Please find the images/screenshot attached depicting the desired changes for the alternate item feature in some of the prominent screens of the cycle.

**1) BOM**
1_a_bom_sc

**2) BOM Browser**
1_b_bombrowser-alternate

**3) BOM Stock Report**
1_c_bomstockreport-alternate

**4) Purchase Order to subcontracting supplier**
2_po_rawmaterial_sc

**5) Material Transfer for PO**
3_material trasfer_to default_samsung_sc

**6) Material Transfer after changing from the substitute item mentioned in the BOM.**

3_material trasfer_to_changedell_sc

**7) Validation while doing material transfer of the alternate line item.**
4_materialtransfer_validation_error_dell_sc

rohitwaghchaure:
First cut of an item alternative feature, kindly share your feedback

### New doctype Item Alternative
screen shot 2018-02-28 at 12 01 41 am

### Item Alternative List
screen shot 2018-02-27 at 11 28 50 pm

### Added checkbox Allow Alternative Item in BOM, Stock Entry
screen shot 2018-02-27 at 11 50 37 pm

### Added custom button(Alternate Item)
screen shot 2018-02-27 at 11 54 04 pm

### Select alternative item
screen shot 2018-02-27 at 11 54 20 pm

### Subcontract example
subcontract

- [x] Test cases

- [x] Documentation

Fixed https://github.com/frappe/erpnext/issues/10499


joezsweet:
Would be interesting to allow it in item master not only in BOM

bohlian:
@rohitwaghchaure you think it would be possible to have multiple items as an alternative? In manufacturing which I believe will be most useful there are many occasions where there are multiple alternatives.

creamdory:
@rohitwaghchaure this is really a great feature! We have open a PR for this here https://github.com/frappe/erpnext/pull/12379 but has been on pause. From our suggested design, we can also add the following features:

- Would be best to be attached in the Item master level (not only BOM level)
- User must be able to choose if it is a "two-way" of "one-way" alternative. "Two-way" means the parent item can be an alternative to a child alternative and vice versa. "One-way" means alternative is strictly for parent to child only.
- Alternative can be set in the BOM and Production Order/Work Order level. It means that set-up from the BOM level can be changed in Production Order level as well.

Cheers,

Dori

dottenbr:
Awesome @rohitwaghchaure thank you for this improvement!

rohitwaghchaure:
@joezsweet @creamdory @dottenbr @bohlian 
Thanks for your feedback!

- @bohlian User can create multiple alternative items against the item, user has to create new records each time

- @creamdory @joezsweet We'll add the Allow Alternative Item checkbox in Item, Production Order. System will check first in Item > Production Order > BOM.

- @creamdory Regarding UOM > if the alternative item has different stock UOM then user will add conversion factor. Because if user has selected different UOM for Original Item while transaction then there is no point to restrict alternative item based on Stock UOM

PawanMeh:
@rohitwaghchaure , in sub-contracting when we receive the finished goods, does this feature also remove the alternative items accordingly from the supplier warehouse?

rohitwaghchaure:
@PawanMeh 
Yes, it will deduct the material which has been transferred. If user has transferred alternate item then while receiving finished good, system will deduct that alternate item.

nabinhait:
@rohitwaghchaure Can you please fix the PR based on Production Order -> Work Order renaming?

rohitwaghchaure:
Fixed!

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -120,6 +120,22 @@ frappe.ui.form.on('Stock Entry', {
 			});
 		}
 
+		if (frm.doc.docstatus == 0 && frm.doc.__onload &&
+			frm.doc.__onload.allow_alternative_item) {
```
nabinhait: What if user added an item later after loading the stock entry, which have allows alternative items?

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.js
rohitwaghchaure: Fixed!

nabinhait (approved):

leonpollak:
Excuses - are there any planned dates to see the V11 going out? This feature is highly required in my case...
Or may be I can incorporate these changes directly into my VM version 10?

lamphannguyen:
Please , show me how to use item alternative in version 10. I saw it merge but I can not find where to use this option

mallesree:
How do we use this item alternative in V11?

mallesree:
>     * User can create multiple alternative items against the item, user has to create new records each time

@rohitwaghchaure Can you please help in the process, how to create new records so that we can create multiple alternative items against the item? Currently I do not get option to do so in V11.

Issue #12222, opened 2017-12-28
[BUG][Critcal] Sales/Purchase return for Different UOM


KT-snorlax:
Stock UOM is Unit and conversion rate for boxes is given as 10.

If you sell in boxes (10unit for each) , then return comes in unit you cannot make return,
even though in actual sold 10 units by selling 1 box as per stock UOM , i cannot make return for 4 units as invoice only sees sold qty and not actual qty.

Return should be considered on the **actual** quantity sold/ purchased as per Stock irrespective of the UOM.

refer 2nd part of https://discuss.erpnext.com/t/sales-purchase-return-for-different-uom-and-bulk-return/32005

Hope to see a quick resolution ,
Thanks

KT-snorlax:
@nabinhait , Any update on this?

KT-snorlax:
@umairsy , Any update on this please

KT-snorlax:
@nabinhait 
extremely sorry to keep insisting , could you please address this issue with a little more priority 
, other requirement are also stuck because of this.
i have to do a very long workaround for this.

Thanks for the efforts so far.

rohitwaghchaure:
@KT-snorlax 

We have [fixed](https://github.com/frappe/erpnext/pull/13132) this issue and currently fix is under review and we'll release the fix once the PR is merged.

PR #13132, opened 2018-02-28
Multi-UOM for sales/purchase return

rohitwaghchaure:
**Issue**

- Sale in 1 BOX of an item _Test Purchase Return For Multi-UOM

- 1 box = 10 Unit

- While making sales return for 4 unit getting below error

screen shot 2018-02-28 at 5 25 04 pm

Fixed https://github.com/frappe/erpnext/issues/12222


ushashmi:
Can we also do this for purchase receipt as well? If the Purchase Order is in BOX and we received item in units then system should allow to receive in units aswell.

PR #13146, opened 2018-03-02
New subscription Management

### LINKED ISSUE #12797: Subscriptions Management Enhancements
As per @umairsy request, I'm creating this issues as a feature request.

Originally posted in ERPNext forum: https://discuss.erpnext.com/t/manage-subscriptions-as-a-subscription-not-as-a-recurring-invoice/32105

The current system is messy, why not something closer to what programs like WHMCS, ClientExec, Blest, PayPlans or OSMembership do?

They all use the same core method which can be resumed in this way.

Client -> Subscriptions
Each client can have multiple subscriptions
Each subscription keeps an history of invoices.

How subscriptions are done.
-> Create a service
(options like)
-- Is it recurring?
-- Fee per billing term
-- Billing terms possible (x days, monthly, year, lifetime)
-- Trial term? (if yes, define the timeline)

How to Manually update a service
-> Client -> Subscription -> details (change whatever as an overwrite)

tundebabzy:
This PR aims to add a full featured Doctype for managing subscriptions. Once a Subscription has been saved, it doesn't require user interaction as ERPNext will automatically update the Subscription as changes occur.

subscription-1

##### Create a Subscriber
subscriber

##### Create a Subscription Plan
subscription-plan

##### Create the subscription
- Add the Subscriber
- Add required dates and quantity as needed
- Add Subscription Plan(s)
- Add Sales Taxes and Charges Template if required
- Add Discount information if required
- Save

##### After that...
ERPNext will constantly check each saved Subscription to change their status and generate new invoices as needed.

### Subscription Status
There are five:
- Trialling: Subscription in trial period
- Active: Subscription that has no uncleared invoice
- Past Due Date: Subscription that has an overdue invoice
- Unpaid: Subscription that has an invoice overdue beyond the grace period
- Canceled: Subscription that has an invoice overdue beyond the grace period and is canceled

### Manually Updating A Subscription
Simply open the required Subscription and make changes. Note that changes will only reflect in subsequent invoices. Past invoices will remain the same.

If the Subscription seems not to be in sync with reality, simply click the 'Fetch Subscription Updates' button so ERPNext can immediately update it.

### Subscription Cancellation
If you need to immediately cancel a Subscription, open the Subscription and click the 'Cancel Subscription' button. If the subscription is active at the time of cancellation, ERPNext will generate a final invoice that will be prorated by default. If you want the invoice to be for the full amount, you can adjust this in Subscription Settings

### Restarting Subscriptions
You can restart a canceled Subscription. Open the Subscription you want to restart and click the 'Restart Subscription' button. Note that restarting a Subscription will cause you to lose the history of invoices generated by the Subscription. This doesn't mean that the actual invoices will be lost, the Subscription will only 'forget' the invoices it has generated before the Subscription got restarted.

This closes #12797


olamide2:
Cool Concept,
Tunde, when are we going to start looking at integrating with paystack?

tundebabzy:
Hmmm. Never ever given it a thought. You can open an issue for it and let's see how far we can take it from there

Olawale1:
Hi @tundebabzy 

Great Job.... Two thumbs up!

Kind regards,

ryuliantoro:
Is this the existing Subscription feature (for the reccuring invoices etc)? 
or the new discussed Subscription (for services)?

tundebabzy:
This is a new Subscription. It is quite different in functionality from the present Subscription module and the present Subscription app will be renamed.

rmehta:
- [ ] Rename current "Subscription" to "Auto Repeat" and move it to frappe
- [ ] Rename this to "Subscription"

nabinhait:
Use column breaks in Subscription doctype

erpnext/accounts/doctype/subscriber/subscriber.json
```diff
@@ -0,0 +1,126 @@
+{
```
nabinhait: Do we need a new Subscriber doctype? Maybe we can directly use Customer.

tundebabzy: Actually we need the Subscriber doctype to capture things that are subscription related as time goes on. For example, i'll soon add the ability to use coupons in subscriptions. Coupons can added on subscription basis or on subscriber basis. If the user chooses to add a coupon on subscriber basis, Customer doctype doesn't seem the right place to add it. Another potential addition is account balance. User's should be able to view the Subscriber's balance on account without going through the subscriptions. Customer doctype will also not be a good place for it.

Olawale1: Hi @tundebabzy 

Trust you're doing great. I see your point here but I hope the Subscriber concept doesn't get a bit confusing for users. My first inclination was also 'Why doesn't the Subscriber field link straight to the Customer doctype?' After your explanation though, it makes sense

Please include 'Member (Non-Profit Domain)' in the Subscriber DocType too!

Cheers!

tundebabzy:
Depends on frappe/frappe#5183 which moves the present Subscription into frappe as Auto Repeat

erpnext/accounts/doctype/subscription/subscription.js
```diff
@@ -1,75 +1,78 @@
...
-			frm.add_custom_button(__(label),
-				function() {
-					frappe.route_options = {
-						"subscription": frm.doc.name,
-					};
-					frappe.set_route("List", frm.doc.reference_doctype);
-				}
-			);
+		if(!frm.is_new()){
+			if(frm.doc.status !== 'Canceled'){
```
nabinhait: Check spelling "Cancelled"

erpnext/patches/v10_1/transfer_subscription_to_auto_repeat.py
```diff
@@ -0,0 +1,31 @@
...
+					'Delivery Note', 'Purchase Invoice', 'Quotation', 'Sales Order',
+					'Purchase Receipt', 'Supplier Quotation']
+
+	frappe.reload_doc('accounts', 'doctype', 'sales_invoice')
+	frappe.reload_doc('accounts', 'doctype', 'purchase_invoice')
+	frappe.reload_doc('accounts', 'doctype', 'payment_entry')
+	frappe.reload_doc('accounts', 'doctype', 'journal_entry')
+	frappe.reload_doc('buying', 'doctype', 'purchase_order')
+	frappe.reload_doc('buying', 'doctype', 'supplier_quotation')
+	frappe.reload_doc('desk', 'doctype', 'Auto Repeat')
```
nabinhait: auto_repeat

erpnext/patches/v10_1/transfer_subscription_to_auto_repeat.py
```diff
@@ -0,0 +1,31 @@
...
+
+	for doctype in to_rename:
+		rename_field(doctype, 'subscription', 'auto_repeat')
+
+	subscriptions = frappe.db.sql('select * from `tabSubscription`', as_dict=1)
+
+	for doc in subscriptions:
+		doc['doctype'] = 'Auto Repeat'
+		auto_repeat = frappe.get_doc(doc)
+		auto_repeat.insert()
```
nabinhait: We can use `db_insert` here, to make it faster

erpnext/patches/v8_1/gst_fixes.py
```diff
@@ -39,8 +39,8 @@ def add_custom_fields():
 
 	custom_fields = {
 		'Address': [
-			dict(fieldname='gst_state_number', label='GST State Number',
-				fieldtype='Int', insert_after='gst_state'),
+			dict(fieldname='auto_repeat', label='Auto Repeat',
```
nabinhait: Is it done by mistake?

erpnext/patches/v10_1/drop_old_subscription_records.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import unicode_literals
+import frappe
+
+
+def execute():
+	subscriptions = frappe.db.sql('SELECT name from `tabSubscription`', as_dict=True)
```
nabinhait: May be we can delete directly using ```frappe.db.sql('delete from `tabSubscription`')```

nabinhait (changes requested):

erpnext/patches/v10_1/drop_old_subscription_records.py
tundebabzy: True

erpnext/patches/v8_1/gst_fixes.py
tundebabzy: Definitely a mistake. I have no idea why I did this. Reverting

erpnext/accounts/doctype/subscription/subscription.js
tundebabzy: @nabinhait I actually mistakenly used the American spelling (VS Code spell checker caused it)... and it's not just here. Should I leave it as 'canceled' since it's technically a correct spelling (American)?

tundebabzy:
@nabinhait changes made

tundebabzy:
spelling corrected

chdecultot:
Hi @tundebabzy, thanks for your great work with this refactored module!

Just one question, why did you add a cost field in the plan doctype instead of just proposing to select a pricing list ?

I have the feeling that it adds an extra configuration step, whereas the price could be directly handled through the price list with potential pricing rules... It's just a thought and I'm interested to know if that has been done on purpose ? :smiley:

tundebabzy:
Hi @chdecultot
> Just one question, why did you add a cost field in the plan doctype instead of just proposing to select a pricing list ?

Actually it was done on purpose because my mental model of the price is *price of the plan*, not *price of the item(s) in the plan*.

tundebabzy:
@agritheory 
> This is great. Are you still open to working in holiday and date management?

Sorry for taking so long to reply. You mean improvement in holiday and date management?

agritheory:
@tundebabzy My thought on this was to have the subscription dates be pushed around weekends and holidays using the company's holiday schedule or a designated alternative. I had a PR pulled in that added these manually, but I think the holiday list is the best place to manage this.

cudokmark:
Hi all,

In what version is this available. We run ERPNext: v10.1.48 (master) and Frappe Framework: v10.1.45 (master). Thank you.

henwoodluke:
When will this be available?

PR #13147, opened 2018-03-02
Add Ends-On field in the Lead doctype

Zlash65:
lead-enhance

Earlier if a contact_date was set for Lead, an event was added in the calendar. The ends_on date of the event added had to be manually changed by going to the Event doc created for the Lead.
If the contact_date was changed from the Lead, the events linked with it were all deleted and new event would be created. Which means manually going to the Event doc and changing the ends_on for the event.

Now - ends_on can directly be set from Lead itself and will be mapped with the event.

### DIFF
--- a/erpnext/crm/doctype/lead/lead.js
+++ b/erpnext/crm/doctype/lead/lead.js
@@ -75,6 +75,14 @@ erpnext.LeadController = frappe.ui.form.Controller.extend({
 		if (this.frm.doc.organization_lead == 1) {
 			this.frm.set_value("lead_name", this.frm.doc.company_name);
 		}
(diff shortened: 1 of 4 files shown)


nabinhait:
@Zlash65 Why do we need ends_on for the event? As the event is for contacting someone, should not it be a one day event?

Zlash65:
If there is no ends_on time for an event - this is how it looks in calendar
screen shot 2018-03-07 at 10 33 21 pm
It becomes an all day event.

There was also a support issue for this [WN-SUP30989](https://frappe.erpnext.com/desk#Form/Issue/WN-SUP30989)

Issue #12967, opened 2018-02-16
System not allows to rename the Item Attribute Value If variant is exist


saurabh6790:
![notabletorenameitemattributevalue](https://user-images.githubusercontent.com/3784093/36297262-bde3d29a-1317-11e8-9549-b48d9712f241.gif)

vishdha:
Hi @saurabh6790 It is not allowed because it's linked with variants. Even if I set allow rename in variant settings if the checkbox is 1 it will allow renaming item attribute value for a new variant but for the previous variant item whose stock ledger generated cannot be modified.

cosimot:
Hi @vishdha ! I am actually allowed to change abbreviation but not attribute value, it is strange as abbreviation has more impact than attribute value, which is a description in fact...

cosimot:
I think If changing abbreviation gives no problem also changing attribute value should not be a problem

vishdha:
Hi @cosimot Why do you want to rename. You can directly add new attribute value.

cosimot:
Hi @vishdha ! Correct typos or (in my case) changing from uppercase to lowercase. Can’t create a new attribute with same abbreviation. People can change their mind/needs or make mistakes!

PR #13161, opened 2018-03-05
Allow Item variant Rename 

vishdha:
https://github.com/frappe/erpnext/issues/12967

item_val


vishdha:
https://github.com/frappe/erpnext/issues/12967

PR #13165, opened 2018-03-05
Data import Tool bug fix

### LINKED ISSUE #13064: Data import tool does not should correct message for mandatory naming_series
How to replicate:

- try importing Employee with naming series set
- don't set naming series

What happens?

No error message, data import tool goes to "success"

surajshetty3416:
fixes #13064

### DIFF
--- a/erpnext/hr/doctype/employee/employee.py
+++ b/erpnext/hr/doctype/employee/employee.py
@@ -24,6 +24,8 @@ class Employee(NestedSet):
 			throw(_("Please setup Employee Naming System in Human Resource > HR Settings"))
 		else:
 			if naming_method == 'Naming Series':
+				if not self.naming_series:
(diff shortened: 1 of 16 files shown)


erpnext/hr/doctype/employee/employee.py
```diff
@@ -24,6 +24,8 @@ def autoname(self):
 			throw(_("Please setup Employee Naming System in Human Resource > HR Settings"))
 		else:
 			if naming_method == 'Naming Series':
+				if not self.naming_series:
```
rmehta: create a common function `set_name_by_naming_series` and commonify.

rmehta (changes requested):

nabinhait:
@surajshetty3416 Travis is failing because now naming is based on ##### instead of ####.

PR #13166, opened 2018-03-05
[fix] use join instead of subquery

saurabh6790:


amitsaxena01:
Error in sql statement : 
frappe.db.sql('''update `tabPurchase Order Item` poi, `tabSales Order Item` soi
		set poi.sales_order_item = soi.name
		where ifnull(poi.sales_order, "")!="" and soi.parent=poi.sales_order
		and soi.item_code=poi.item_code where docstatus = 1
	''')
SQL statements has two wheres. I believe correct should be 

frappe.db.sql('''update `tabPurchase Order Item` poi, `tabSales Order Item` soi
		set poi.sales_order_item = soi.name
		where ifnull(poi.sales_order, "")!="" and soi.parent=poi.sales_order
		and soi.item_code=poi.item_code and poi.docstatus = 1
	''')

saurabh6790:
@amitsaxena01 Thanks!

Issue #13049, opened 2018-02-22
Set Customer's Item Code based on Customer Group


umairsy:
In the item master, we can also track Customer's Item Code.

There could be some Item Code which would be similar for the group of Customer. Hence, with the selection of Customers, we should also give scope for the selection of Customer Group in this table.

![screen shot 2018-02-22 at 7 00 12 pm](https://user-images.githubusercontent.com/910238/36541086-c9c7f29a-1802-11e8-82b5-07dd48da9608.png)

PR #13198, opened 2018-03-06
[Enhance] Fetch Customer's Item code based on Customer Name or Customer Group

Zlash65:
Issue:- https://github.com/frappe/erpnext/issues/13049

Item Master :- 
screen shot 2018-03-06 at 9 24 41 pm

customer-item-code

If customer_name is already set, then changing the customer_group for it will be prohibited.


joezsweet:
Would be possible to do this on supplier side as well?

erpnext/stock/doctype/item/item.js
```diff
@@ -170,6 +170,36 @@ frappe.ui.form.on('Item Reorder', {
...
+	var grid_row = cur_frm.fields_dict['customer_items'].grid.grid_rows_by_docname[row.name];
+
+	if (!row.customer_name) {
+		return false;
+	}
+
+	frappe.model.with_doc("Customer", row.customer_name, function() {
+		var customer = frappe.model.get_doc("Customer", row.customer_name);
+		row.customer_group = customer.customer_group;
+		grid_row.refresh_field("customer_group");
```
nabinhait: no need for `grid_row` obj here. You can use `set_value` function, it also refresh fields after setting the value.

erpnext/stock/get_item_details.py
```diff
@@ -396,8 +396,16 @@ def validate_conversion_rate(args, meta):
...
+		out.customer_item_code = None
 		customer_item_code = item_doc.get("customer_items", {"customer_name": args.customer})
-		out.customer_item_code = customer_item_code[0].ref_code if customer_item_code else None
+
+		if customer_item_code:
+			out.customer_item_code = customer_item_code[0].ref_code
+		else:
+			customer_group = frappe.db.get_value("Customer", args.customer, "customer_group")
+			customer_group_item_code = item_doc.get("customer_items", {"customer_group": customer_group})
+			if customer_group_item_code and not customer_group_item_code[0].customer_name:
```
nabinhait: This `if` condition is not required.

erpnext/stock/doctype/item_customer_detail/item_customer_detail.json
```diff
@@ -33,24 +38,62 @@
...
+   "no_copy": 0,
+   "options": "Customer Group",
+   "permlevel": 0,
+   "precision": "",
+   "print_hide": 0,
+   "print_hide_if_no_value": 0,
+   "read_only": 0,
+   "remember_last_selected_value": 0,
+   "report_hide": 0,
+   "reqd": 1,
```
nabinhait: It should not be mandatory

nabinhait (dismissed):

PR #13211, opened 2018-03-07
Healthcare - Patient Appointment - Service unit based scheduling and booking

### LINKED ISSUE #13016: Healthcare Patient Appointment Save Button Issue
appointment-error

When I first create an appointment, there is no Save button.  However, if I clear the cache by using the "reload" option, the Save button appears.  Although the save button appears, I don't seem to be able to save.  Any changes I make to the doc are not persisted.

Arrived is a custom field.  Notes is not.  Both seem to save without error until I clear the cache at which point their values are gone.  Since Notes is a built in field and is not marked as readonly, I would expect my values to persist after clicking save.

It seems to me that the bug could be one of two things: First, maybe the doctype is meant to be readonly and the save button should never appear.  Otherwise, it seems that the save button is not working correctly.

I am running a fresh install of ERPNext on Ubuntu 16.04 installed via the install script.  I made a couple customizations to Patient Appointment, but before making the screencast, I "reset to defaults" on the customize page of Patient Appointment.

pjamsheer:
Patient Appointment - Service Unit based scheduling and booking
Availability check and book appointment based on service unit, appointment invoice creation optimized
Patient Appointment - issue fixed  #13016 
Consultation - Remove validation on submit, Label changed from Drug Prescription to Medication
Physician - schedule based on patient service unit
Patient Service Unit


nabinhait:
@pjamsheer Travis is failing

pjamsheer:
@nabinhait Fixed.

PR #13212, opened 2018-03-07
removed tether between monthly dates and fiscal year

### LINKED ISSUE #12732: Wrong dates created for salary slip
When using the new Payroll Entry doctype, ERPNext currently creates incorrect start dates for my employee's Salary Slips. For example, when I create a Payroll Entry for the period 01-Jan-2018 to 31-Jan-2018, the salary slips come back starting 16-Jan-2018 (incorrect)  and ending 31-Jan-2018 (correct).

Based on previous discussions (see here: https://github.com/frappe/erpnext/issues/9513), this seems to be related to the fact that my fiscal year begins on the 16th of July.

ghost:
(Fixes https://github.com/frappe/erpnext/issues/12732)

The `Payroll Entry` doctype currently relies on Fiscal Year details to set `start_date` and `end_date` of a payroll period. There are two problems with the implementation:

1. It forces users to start their pay period on the same day of the month that their fiscal year starts on, which is proper in many cases perhaps but limiting in others.

2. It truncates the pay period to the end of the current calendar month for users whose fiscal year starts on something other than the 1st.

This pull request removes the link between the pay period and the fiscal year, making monthly and bimonthly payroll periods function more similarly to weekly and biweekly.

### DIFF
--- a/erpnext/hr/doctype/payroll_entry/payroll_entry.py
+++ b/erpnext/hr/doctype/payroll_entry/payroll_entry.py
@@ -6,7 +6,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe.model.document import Document
 from dateutil.relativedelta import relativedelta
-from frappe.utils import cint, flt, nowdate, add_days, getdate, fmt_money, add_to_date, DATE_FORMAT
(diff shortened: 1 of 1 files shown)


theopen-institute:
Ah, thanks. Since the issue was still open, I didn't realize it had been fixed. Glad it's done now.

Issue #13220, opened 2018-03-07
Default VAT Switzerland not correct in 2018


lasalesi:
When the setup wizard creates tax rates, this is no longer correct in 2018 (new taxe rates apply).

Tax rates should be 2.5% reduced VAT, 7.7% normal VAT and 3.7% lodging VAT. External reference: https://www.estv.admin.ch/estv/en/home/mehrwertsteuer/fachinformationen/steuersaetze/entwicklung-mwst.html

PR #13221, opened 2018-03-07
corrected setup wizard tax rates for Switzerland 2018 (#13220)

lasalesi:
This is a correction to the setup wizard for the correct taxes valid as of 1st of January 2018. Details, see #13220 

Also, the % sign in the name sometimes leads to issues (e.g. when trying to rename, there is an error resource not found), therefore clear names without % signs used.

This will require an update of the DE/FR/IT language file, which should be implemented through https://translate.erpnext.com/


nabinhait:
@lasalesi 👍 for the first time contribution.

PR #13237, opened 2018-03-08
Nepal localization

ghost:
Added regional hooks to prevent deletion of sales invoices if location is Nepal

Ongoing discussion here:
https://discuss.erpnext.com/t/compliance-requirement-that-must-be-fulfilled-to-get-approval-from-tax-authority-in-nepal/34532


theopen-institute:
Hmm...I must be misunderstanding how these tests work if they’re failing, since I do have a function returning “overridden”. I’ll take a look at it and see if I can figure out the problem, but any tips are appreciated. Thanks!

PR #13239, opened 2018-03-08
allow assets to cancel even if it is linked with PI

manassolanki:
If the Asset is created against the Purchase Invoice, then it creates a cyclic validation and then Asset or Purchase Invoice can't be cancelled.


nabinhait:
@manassolanki Please make the changes as per our discussion

PR #13246, opened 2018-03-09
prevent deletion of sales invoices if country is Nepal

ghost:
The first of several compliance steps for Nepal. Prevents deletion of Sales Invoices in Nepal (using France's code as a model).

See this thread: https://discuss.erpnext.com/t/compliance-requirement-that-must-be-fulfilled-to-get-approval-from-tax-authority-in-nepal/34532/3

### DIFF
--- a/erpnext/hooks.py
+++ b/erpnext/hooks.py
@@ -201,7 +201,7 @@ doc_events = {
 	"Sales Invoice": {
 		'validate': 'erpnext.regional.india.utils.set_place_of_supply',
 		"on_submit": "erpnext.regional.france.utils.create_transaction_log",
-		"on_trash": "erpnext.regional.france.utils.check_deletion_permission"
+		"on_trash": ["erpnext.regional.france.utils.check_deletion_permission", "erpnext.regional.nepal.utils.check_deletion_permission"]
 	},
 	"Payment Entry": {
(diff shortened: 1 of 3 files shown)


theopen-institute:
Hmm...my apologies, but I don't understand why the unit test is failing. I've got a test method in utils.py  returning "overridden", but the test is still getting "original" with flag "Nepal". Any tips?

erpnext/hooks.py
```diff
@@ -201,7 +201,7 @@
 	"Sales Invoice": {
 		'validate': 'erpnext.regional.india.utils.set_place_of_supply',
 		"on_submit": "erpnext.regional.france.utils.create_transaction_log",
-		"on_trash": "erpnext.regional.france.utils.check_deletion_permission"
+		"on_trash": ["erpnext.regional.france.utils.check_deletion_permission", "erpnext.regional.nepal.utils.check_deletion_permission"]
```
rmehta: @theopen-institute thanks for the PR!

just make a common function and just add to the list of countries.

`erpnext.regional.check_delete_permission`

rmehta (changes requested):

nabinhait:
@theopen-institute I have fixed the PR and merged it. By the way, travis was failing because you did not mention `regional_overrides` in hooks.

theopen-institute:
Thanks @nabinhait and @rmehta! I'm relatively new to frappe and python, but hopefully my next PR will have fewer problems :)

PR #13255, opened 2018-03-09
Don't allow to set negative quantity if transaction is not return entry

rohitwaghchaure:
Added validation if user set the negative quantity
screen shot 2018-03-09 at 4 58 07 pm

### DIFF
--- a/erpnext/controllers/status_updater.py
+++ b/erpnext/controllers/status_updater.py
@@ -156,6 +156,9 @@ class StatusUpdater(Document):
 
 			# get unique transactions to update
 			for d in self.get_all_children():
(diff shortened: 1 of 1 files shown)


gbard:
Is there an option to deactivate this? We are used to process item exchange by creating a Sales Invoice with negative quantities for items returned and positive quantities for items that are exchanged for the returned items (on the same Sales Invoice), Ex: Item A | Qty: -1 | Amount -9.99$ > Item B | Qty: 1 | Amount 9.99$ > TOTAL: 0$.

vcavel:
This check bothers us a lot as well.

We sell beers in returnable containers and this check makes it very difficult to do so ! It is not practical to create a credit note every time !

PR #13257, opened 2018-03-09
Add Health Insurance Doctype and link to Employee

creamdory:
Health Insurance is a generic information that should be added as an Employee information. In a company, it can be one Insurance provider or several.

In this case, I have added "Health Insurance" doctype and added "Health Insurance Provider" link field in Employee Doctype with "Health Insurance No" data field that depends on it.

screen shot 2018-03-09 at 14 17 17

insurance no


erpnext/hr/doctype/health_insurance/health_insurance.json
```diff
@@ -0,0 +1,147 @@
...
+ "editable_grid": 1, 
+ "engine": "InnoDB", 
+ "fields": [
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "name1",
```
nabinhait: Use `health_insurance_name` instead

erpnext/hr/doctype/health_insurance/health_insurance.json
```diff
@@ -0,0 +1,147 @@
...
+   "translatable": 0, 
+   "unique": 0
+  }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "company",
```
nabinhait: Why is it required? It can be global records for every company.

nabinhait (changes requested):

creamdory:
@nabinhait I have edited request changes. Also, I have added it under "Set-up" in HR module.

screen shot 2018-03-13 at 08 37 56

screen shot 2018-03-13 at 08 30 12

PR #13263, opened 2018-03-10
[Feature] Advance Account Balance in Employee Advance

### LINKED ISSUE #13054: Advance Account Balance in Employee Advance
Hi,

Please add a field in Employee Advance to show the balance on the Employee's Advance Account. This is information that the approver usually needs to see before making a decision to approve or not

Thanks

deepak-mnt:
fixes #13054 
advance_balance

### DIFF
--- a/erpnext/hr/doctype/employee_advance/employee_advance.py
+++ b/erpnext/hr/doctype/employee_advance/employee_advance.py
@@ -53,9 +53,6 @@ class EmployeeAdvance(Document):
 		self.set_status()
 		frappe.db.set_value("Employee Advance", self.name , "status", self.status)
 
-	def get_due_advance_amount(self):
-		employee_due_amount = frappe.get_all("Employee Advance",filters={"employee":self.employee,"docstatus":1,"posting_date":("<=",self.posting_date)},fields=["advance_amount","paid_amount"])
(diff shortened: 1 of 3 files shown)


erpnext/hr/doctype/employee_advance/employee_advance.py
```diff
@@ -53,6 +53,10 @@ def set_total_advance_paid(self):
 		self.set_status()
 		frappe.db.set_value("Employee Advance", self.name , "status", self.status)
 
+	def get_due_advance_amount(self):
```
nabinhait: Make it a whitelisted function. If we write it inside the class, then we need to pass the entire doc to the server, which is unnecessary. We only need employee and posting date.

erpnext/hr/doctype/employee_advance/employee_advance.py
```diff
@@ -53,6 +53,10 @@ def set_total_advance_paid(self):
...
+		employee_due_amount = frappe.get_all("Employee Advance",filters={"employee":self.employee,"docstatus":1,"posting_date":("<=",self.posting_date)},fields=["advance_amount","paid_amount"])
```
nabinhait: Break this line in multiple line, for better readability. Also, use space after comma 😉

nabinhait (changes requested):

Issue #4182, opened 2015-10-16
Purchase Invoice must update Stock Valuation


jof2jc:
I found 'serious issue' which is....Purchase Invoice doesn't update stock valuation. There's some business facts we must understand:
1. Purchase receipt rate is not always same with invoice rate...due to rate changes, correction, final discount from supplier etc
2. Additional costs that may occur at invoice time.

We can do all above in the purchase invoice if allowance percent is allowed, but unfortunately it doesn't update valuation per stock item. This issue definitely will cause wrong valuation that affects wrong calculation of stock assets value, COGS & profit/loss calculation

I just wondering why no ERPNext customers who aware on this issue

See thread here https://discuss.erpnext.com/t/different-rate-between-purchase-receipt-and-purchase-invoice/7488/6

Please other users inputs...Thanks

jof2jc:
Any comment from other users on this issue?

nabinhait:
@jof2jc Thanks for your input, your point is very much valid. If there is any additional costs in tax table, that can be managed via Landed Cost Voucher. But as we are allowing to change rate in Purchase Invoice, there should be a way to update item valuation as per changed rate.

But as there is a multiple-to-multiple relationship between PO, PR and PI and those documents can be made in any order, it is a very difficult problem to solve. Anyway, we will discuss on it internally, and let you know how we can solve the issue.

jof2jc:
@nabinhait Have you found any solutions for this? Thanks

nabinhait:
@jof2jc We have discussed it internally, but did not found any easy solution yet, to fix the issue. Will update you as soon as we found any solution.

jof2jc:
@nabinhait thanks for this, waiting for the solution....btw, sorry if this one is slightly out of this topic, BUT it's still related to wrong costing calculation. I'm using ERPNext for quite sometime for live environment so now I'm getting wrong stock calculation due to this issue as well. I believe many other customers don't aware too..

If you don't mind to get a chance to take a look at this too:
https://discuss.erpnext.com/t/fifo-does-not-work-on-repack-entry-for-batch-items/7963

noetico:
[...]

PR #13265, opened 2018-03-10
Update stock valuation with Purchase Invoice

mokhito:
Our company receives raw material every week. Upon arrival, a Purchase Receipt is issued with zero valuation rate (or previous rate) for the raw material item because its price is not yet determined.
At the end of the month, our supplier sends us an invoice with the price for all deliveries of the month.

We need to be able to update the stock value of our raw material when submitting the Purchase Invoice.

This Pull Request is our implementation of this functionality.

It adds a check button to update the valuation rate of each item in a Purchase Invoice that was imported from a Purchase Receipt. This button is visible only if we make a Purchase Invoice from a Purchase Receipt or if we use the import items tool in a Purchase Invoice.

Note: we don't use perpetual inventory.

Related to issue #4182.


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -298,12 +301,40 @@ def on_submit(self):
...
 		self.update_project()
 		self.update_fixed_asset()
 
+	def update_sl_valuation(self):
+		for item in self.get("items"):
+			pr = item.purchase_receipt
+			pr_doc = frappe.get_doc("Purchase Receipt", pr)
+		
+			for pr_item in pr_doc.get("items"):
+				if item.item_code == pr_item.item_code:
```
nabinhait: Instead of `item_code`, update it based on `pr_detail` value which holds the value of name of the PR Item row.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -298,12 +301,40 @@ def on_submit(self):
...
+					pr_item.valuation_rate = item.rate
+
+			# save will update item valuation rate in PR
+			pr_doc.save()
```
nabinhait: Instead of saving, use `db_update` method to avoid unwanted validations and mandatory checks. Check for reference, https://github.com/frappe/erpnext/blob/master/erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.py#L114

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -1238,6 +1238,69 @@
...
    "unique": 0
   }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "default": "0", 
+   "fieldname": "has_purchase_receipt",
```
nabinhait: This field is not required as we can check this via code itself.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -1238,6 +1238,69 @@
...
+   "fieldname": "update_stock_valuation", 
+   "fieldtype": "Check", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Update Stock Valuation",
```
nabinhait: The label should be more explanatory like 'Update Valuation Rate on Purchase Receipts', as we removed the above field.

nabinhait (changes requested):
- Write unit test cases
- Update docs

nabinhait:
@mokhito Thanks for the PR 👍, this is really a great feature addition. Please check my comments above. Let me know if you need any help. 
Closing this for now, please reopen the PR after fixing.

rvbd:
Did this ever got another PR? Cause this functionality would be really handy.

ishanloya:
Wondering if this was implemented. Seems like a really important feature

PR #13284, opened 2018-03-12
Physician Schedule fixes

### LINKED ISSUE #11582: HealthCare - Cannot Book More Than One Appointment For A Day For A Doctor
The Physician has entered 15 minutes per patient as an entry. Hence, if an appointment is booked for 12 PM, the next appointment slot available should be starting from 12:15 and so on.

However, once a patient has booked a slot for 12 PM, no new appointments can be made for the day. 

peek 2017-11-15 12-29

Ideally, I should be able to select any slot listed in the Physician's schedule from 12:15:00 to 13:30:00 (as that is the schedule for the day)

image

rkurungadam:
- fix https://github.com/frappe/erpnext/issues/11582
- ability to add schedule for multiple days in a shot

### DIFF
--- a/erpnext/healthcare/doctype/patient/patient.json
+++ b/erpnext/healthcare/doctype/patient/patient.json
@@ -44,6 +44,7 @@
    "reqd": 0, 
    "search_index": 0, 
    "set_only_once": 0, 
+   "translatable": 0, 
    "unique": 0
   }, 
   {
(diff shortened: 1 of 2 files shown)


nabinhait (changes requested):
@rkurungadam "Add Time Slots" button is looking really good 👍, we can implement it many more places.
Few observations:
1. The dialog does not hide automatically after we select some slots.
2. It allows same time slots (for same day) multiple times.

rkurungadam:
- Dialog is left open to add more
- There were no validation earlier, will update and raise another PR
Thanks

PR #13290, opened 2018-03-13
Set po nos in DN and SI

nabinhait:


adityaduggal:
@nabinhait This is creating a lot of issues with our system. Please note that there should be an option to not PULL the po_no since it has virtually stopped all our invoicing refer:  WN-SUP31654 and  WN-SUP31660

I think the code should be like this on https://github.com/frappe/erpnext/blob/master/erpnext/controllers/selling_controller.py#L332
```
	def set_po_nos(self):
		if self.doctype in ("Delivery Note", "Sales Invoice"):
			if self.doctype == 'Sales Invoice' and self.is_pos != 1:
				ref_fieldname = "against_sales_order" if self.doctype == "Delivery Note" else "sales_order"
				sales_orders = list(set([d.get(ref_fieldname) for d in self.items]))
			else:
				ref_fieldname = "against_sales_order" if self.doctype == "Delivery Note" else "sales_order"
				sales_orders = list(set([d.get(ref_fieldname) for d in self.items]))
			if sales_orders:
				po_nos = frappe.get_all('Sales Order', 'po_no', filters = {'name': ('in', sales_orders)})
				self.po_no = ', '.join(list(set([d.po_no for d in po_nos if d.po_no])))
```

PR #13296, opened 2018-03-13
[Feature] Add default duration and color to appointment type

tomtom5152:
When creating a new `Patient Appointment`, it is useful if that appointment has a default duration, and extends off #13211.

In addition, it is useful to see the appointment type in the calendar view by color, giving the physician/reception staff an idea of what the appointment might involve at a glance.

Default duration by Appointment Type screencap

Color by Appointment Type screencap


rkurungadam:
@tomtom5152 please update the modified time in json, the field added is not getting reflected.

tomtom5152:
Modified to the timestamp of the original commit.

rkurungadam:
@tomtom5152 appointment_type.json too.

tomtom5152:
My apologies, I'm very much still getting the hang of frappe so the little nuances like this are getting missed at this stage. Should all now be in order but I haven't had time to revert back and test it applies correctly still.

system19:
The ability to colour code an appointment type is great.

rkurungadam:
@nabinhait please merge, looks like the travis fails is not related. thanks @tomtom5152

shreyashah115:
@tomtom5152 , why are the color and duration fields hidden?

PR #13303, opened 2018-03-15
unordered types None vs int

achillesrasquinha:


adityaduggal:
@achillesrasquinha Why not use `if flt(task.task_weight) > 0:` instead of `if task.task_weight or 0 > 0:`
just asking for increasing my knowledge

achillesrasquinha:
Hi @adityaduggal, the issue seems to be something else. Python 2 is able to compare `None` with `int`s, not true for Python 3. Just removing out past ambiguities now that we're upgrading. 😄

PR #13316, opened 2018-03-16
[hotfix] Added 'Paid Amount' and 'Change Amount' in POS print formats

ghost:
Added 'Paid Amount' and 'Change Amount' in
1. GST POS Invoice.
2. POS Invoice.

**GST POS Invoice:**

**Before:**
gst_pos_before

**After:**
gst_pos_after

**POS Invoice:**

**Before:**
pos_invoice_before

**After:**
pos_after


nabinhait:
@Siddhisawant8 👍for the first PR.
Please add a `if` condition for change amount, only show if there is change amount.

Siddhisawant8:
@nabinhait Have made the said changes in the new commits.

PR #13320, opened 2018-03-17
[hotfix] Autonaming of stock item.

mokhito:
This fixes a bug introduced in commit f2d28ebd6a8bd7a6d94e36128378797a739c0fc9.

The Item would not get an item code on saving and an error would show saying "Item Code is required."


nabinhait:
@mokhito Thanks for sending the PR. Your change will create problem is item naming is done based on item_code. I have sent another fix to solve the issue. 
https://github.com/frappe/erpnext/commit/d97c5a04b2e289ab73738597279ba7fa69a6c5f5

PR #13321, opened 2018-03-17
Copy source project into Purchase Order Item

tukutela:
When using "Make -> Purchase Order" from a drop shipped Sales Order, if the Sales Order has part of a Project, then copy that into the Purchase Order Item doctype. 

screen shot 2018-03-18 at 10 15 50 am

An incredibly simple addition, but very powerful if you need it.


nabinhait:
@tukutela 👍for the first contribution.

tukutela:
Thankyou, I hope to do more.

Issue #13197, opened 2018-03-06
[Bug][Education] Error when showing Gantt view in Course Schedule


ilyesderp:
Hi,
In Education module in Course Schedule when we switch from List view to Gantt view we get the following error:
![capture du 2018-03-06 14 58 59](https://user-images.githubusercontent.com/10146310/37040432-1d89bffc-215a-11e8-8d2d-987f621d17a6.png)
We are using ERPNext version 10.1.5

PR #13338, opened 2018-03-20
[Fix] Course Schedule gantt-view fix

Zlash65:
Issue:- https://github.com/frappe/erpnext/issues/13197

fix:- gantt-fix

dependent on - https://github.com/frappe/frappe/pull/5233


erpnext/education/doctype/course_schedule/course_schedule_calendar.js
```diff
@@ -8,6 +8,7 @@ frappe.views.calendar["Course Schedule"] = {
 		"allDay": "allDay"
 	},
 	gantt: false,
+	order_by: "from_time",
```
manassolanki: Shouldn't be `schedule_date` handier for the user? otherwise irrespective of different dates all the course schedule will come on the same day since from_time is a Time field.

PR #13346, opened 2018-03-20
Student Report Card

manassolanki:
- Optional Letter Head
- Grade or Marks can be shown
- Option to show only final grades or accordingly to the assessment groups.
- Assessment details or brief of grading scale to be printed for information. 

shishuvan_8000_desk 1


nabinhait:
@manassolanki Travis is failing.

Check `update_student_name_in_linked_doctype` function. `student_name` should not be updated in single type doctype.

Actually you can use this common function to update it.
https://github.com/frappe/erpnext/blob/master/erpnext/selling/doctype/customer/customer.py#L79

PR #13363, opened 2018-03-21
Timer in Timesheets!

shreyashah115:
screen shot 2018-03-21 at 10 25 01 pm

timer-final

Summary:
- On clicking, 'Start Timer', a dialog pops up and starts the timer for already present activity for which checkbox `completed` is unchecked.
- If no activities are present, fill up the activity, expected hours or project in the dialog itself, on clicking 'Start', a new row is added into the Timesheet Details child table and timer begins.
- On clicking, 'Complete', the hours and to_time fields are updated for that particular activity.
- At any point of time, if the dialog is closed without completing the activity, on opening the dialog again, the timer resumes by calculating how much time has elapsed since `from_time` of the activity.
- If any activities are already present in the Timesheet with `completed` unchecked, clicking on 'Resume Timer' fetches the activity and starts its timer.
- If the time exceeds the `expected_hours`, an alert box appears.

screen shot 2018-03-21 at 10 34 28 pm
Thanks @pratu16x7 and @netchampfaris for helping out with the UI!


pratu16x7:
What is the behaviour of the timer on manual entry (specifically `from_time`) in the child table?

erpnext/projects/doctype/timesheet/timesheet.html
```diff
@@ -0,0 +1,11 @@
+<div class="stopwatch">
```
pratu16x7: Looks like this belongs in `timer.js` itself, as a helper function returning the template, and not in `timesheet`.

erpnext/public/js/projects/timer.js
```diff
@@ -0,0 +1,135 @@
...
+			'project': row.project,
+			'expected_hours': row.expected_hours
+		});
+	}
+	dialog.get_field("timer_html").$wrapper.append(frappe.render_template("timesheet"));
+	control_timer(frm, dialog, row, timestamp);
+	dialog.show();
+};
+
+var control_timer = function(frm, dialog, row, timestamp=0) {
```
pratu16x7: Will cause this to be left hanging as a global. Let's namespace it with `erpnext.timesheet` as well :)

rmehta:
Also update timesheet docs

shreyashah115:
@pratu16x7 On a manual entry, the timer will start whenever `from_time` = `now_time`.

KanchanChauhan:
It will be good idea to keep it minimized so that you can continue with your other work but still look at the timer.

shreyashah115:
@KanchanChauhan The timer can be minimized i.e. until you click 'Complete', the timer will keep running, even if it is closed. It will always resume by calculating the elapsed time.

rmehta:
@shreyashah115 please also add field to edit Task in the modal.

shreyashah115:
@rmehta done!

PR #13374, opened 2018-03-22
[Fix] General Ledger Report

shreyashah115:
Whenever `party type` == Student or Shareholder, on selecting `party`, it tries to set a field `party_name` by appending "_name" to `party` which won't be fetched as there is no field student_name or shareholder_name in the respective party masters and hence, throws an error.

Issue:
report-issue
Fix:
report-fix

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.js
+++ b/erpnext/accounts/report/general_ledger/general_ledger.js
@@ -85,12 +85,10 @@ frappe.query_reports["General Ledger"] = {
 					frappe.query_report_filters_by_name.party_name.set_value("");
 					return;
 				}
-				if (party_type != "Student" && party_type != "Shareholder") {
-					var fieldname = frappe.scrub(party_type) + "_name";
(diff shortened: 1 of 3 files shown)


erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -82,11 +85,12 @@ frappe.query_reports["General Ledger"] = {
 					frappe.query_report_filters_by_name.party_name.set_value("");
 					return;
 				}
-
-				var fieldname = frappe.scrub(party_type) + "_name";
-				frappe.db.get_value(party_type, party, fieldname, function(value) {
-					frappe.query_report_filters_by_name.party_name.set_value(value[fieldname]);
-				});
+				if (party_type != "Student" && party_type != "Shareholder") {
```
rmehta: such type of hardcoding should be avoided. Let there be a global object that will return the `_name` property of the given party.

rmehta (dismissed):

Issue #13369, opened 2018-03-22
New Column “Consumed Qty” in Production Order


creamdory:
## Current State

In Production Order, there is no record in which we can take a look QTY of consumed components (from Stock Entry) as it doesn't allow "Manufacture" entry without finished product. In this issue https://github.com/frappe/erpnext/issues/13368 , we proposed to allow this functionality. Part of this is to record consumption in Production Order.

## Goal
- [x] Add "Consumed Qty" in Production Order Item doctype.
- [x] The "Consumed Qty" column must be updated if there is a "Material Consumption for Manufacture"  or "Manufacture" Stock Entries against the Production Order. Please see details here https://github.com/frappe/erpnext/issues/13368.

Issue #13368, opened 2018-03-22
New Stock Entry Purpose: "Material Consumption for Manufacture"


creamdory:
## Current State

Currently the only way you can book raw materials against a production order (consuming them) is by finishing a full or partial quantity of a production order. This does not reflect reality in which you will have to be able to consume raw materials against a production order while not yet finishing the product and keeping it as work in progress.

Instead of finishing a set quantity of an item within a production order and consuming the necessary raw materials (1:1), it would be best to allow feature to consume materials for manufacturing of a production order without booking a finished quantity. This may be necessary if certain operations are finished and require for materials to be consumed without creating a final product yet.

## Goal

- [x] Add new Stock Entry Purpose : "Material Consumption for Manufacture". 
This purpose acts like "Material Issue" but with attached Production Order. The consumed QTY must be recorded as well. Please see issue related here https://github.com/frappe/erpnext/issues/13369
- [x] Alter Stock Entry Validation for this purpose.
- [x] Add "Material Consumption" button in "Production Order/Work Order". Must appear if the production order is started. When the button is click, it should create Stock Entry with purpose "Material Consumption for Manufacture". User can use whether to click "Material Consumption" button or "Finish" button.

djpalshikar:
When the user finishes the production order eventually, all the "Material Consumption for Manufacture" entries should affect "Stock valuation" of the manufactured item in some way.

PR #13384, opened 2018-03-24
[Enhancement] Work Order Material Consumption

creamdory:
Made the enhancements from these issues:

https://github.com/frappe/erpnext/issues/13368
https://github.com/frappe/erpnext/issues/13369
https://github.com/frappe/erpnext/issues/13368

BOM:

bom

### Usecase:
**1 transfer, 1 consumption and 1 Finish from a Work Order**

- Work Order

work order 1

- Material Transfer

full material transfer

- Material Consumption

full consumption

- Work Order Completion

fulll finish

### Usecase:
**Multiple Transfer, Consumption and Finish from a Work Order.**

- Work Order
work order 2

- **Partial Stock Transfer** suggest only "Required Qty" * "Qty for Material Transfer for Manufacture". If we click "Finish" before any "Material Consumption", must suggest Qty to consumed in the Manufacture Stock Entry. If we create "Material Consumption for Manufacture" before "Finish", must take into account the "Consumed Qty" during Finish.

1 transfer 1 consumption 2 finish

### Usecase:
**Work Order with Partial transfer, consumption and completion.**

work order

- Material Transfer : Suggest the total Stock to be transferred from the "Required Qty" column and the "Qty for Material Transfer for Manufacture".

stock transfer

- Material Consumption : Suggest and allow only consumption less than or equal to the "Transferred Qty".

material consumption

- Finish Production : Suggest consumption of materials less than the "Consumed Qty" and base total consumption qty from the "Qty for Manufacture".

finish

## Validations

1) Doesn't allow consumption more than Transferred Qty.

dont allow consumption more than transfer

2) Allow Transfer more than the Required Qty. Allow Consumption more than required Qty.
3) Doesn't allow cancellation of "Material Consumption for Manufacture" when Work Order is Completed.

cancel consumption validation

## Notes

- Please help test, write feedbacks and report issues.
- Test Case Creation is on going.
- Documentation will be after Alternative BOM line and Work Order renaming Documentation is done.


dottenbr:
**Let me quickly explain why we didn’t yet contribute documentation on this.** 

We have not updated the documentation since the Frappe team is working on that currently due to the Production Order to Work Order transition. Whoever works on that could include this in the documentation if the design gets accepted or we can contribute the documentation after the updated documentation is available.

dottenbr:
Could you let us know if there is anything you need from us to get the PR merged ASAP? We are on a tight schedule. We are migrating on Friday this week and need this in.

creamdory:
@nabinhait we bumped into an issue with this design since it records "issuance" accounting in the new purpose "Material Consumption for Manufacture". This creates a problem in making/calculating Valuation Rate of the Finish product. Can you suggest how to proceed so we can change code right away? Issue was pointed out in this discuss post https://discuss.erpnext.com/t/material-consumption-for-manufacture/35076.

creamdory:
@nabinhait , here is my suggestion. We should have the idea of recording "WIP" accounts.

So in the current design, we have these accounting entries for direct material:

"Manufacture" Stock Entry:

DEBIT:       Finish Goods      $ 1.00
CREDIT:                  Components         $ 1.00

If we will be adding the idea of "Material Consumption for Manufacture", we can have these accounting entries:

"Material Consumption for Manufacture" Stock Entry:

DEBIT:       WIP Direct Material      $ 1.00
CREDIT:                  Components         $ 1.00

"Manufacture" Stock Entry:

DEBIT:       Finish Goods      $ 1.00
CREDIT:                  WIP Direct Material         $ 1.00

**OR**

"Manufacture" Stock Entry:

DEBIT:       Finish Goods      $ 1.00
CREDIT:                  WIP Direct Material         $ 0.50
                                Components                    $ 0.50

So the components will be exhausted during the consumption using a WIP account. This will also show in the Balance sheet. We can add a column/field in the "Company" doctype as "WIP Direct Material" account to set-up the default account for this scenario. Same is true with Direct Material 

Let us know what you think.

djpalshikar:
>    Doesn't allow consumption more than Transferred Qty.
Currently if the WIP warehouse enough stock, "Finishing is allowed", is there is specific reason why you would like to disallow consumption ?
> We should have the idea of recording "WIP" accounts.
Can you should how the stock ledger will look like after consumption and completion?

creamdory:
@djpalshikar can you pls put feedback from my recent suggestion about WIP accounts?

creamdory:
@djpalshikar

> Currently if the WIP warehouse enough stock, "Finishing is allowed", is there is specific reason why you would like to disallow consumption ?

This is the idea. If the WIP has enough stock it can consume. But if WIP only has 50 and consumption is 60 then this is not allowed.

> We should have the idea of recording "WIP" accounts.

In my recent comment, you can look at the accounting entries. For the Stock Ledger, here it is:

"Material Consumption for Manufacture" Stock Entry

Item                              Warehouse                      Qty             Incoming Rate            Valuation Rate
Component 1               WIP Warehouse              - 1                                                    $1
Component 2               WIP Warehouse             - 1                                                     $1

"Manufacture" Stock Entry

Item                              Warehouse                      Qty             Incoming Rate            Valuation Rate
Finish Goods 1             FG Warehouse                 1                $2                                $2

For partial consumption with "Manufacture" Stock Entry

"Material Consumption for Manufacture"

Item                              Warehouse                      Qty             Incoming Rate            Valuation Rate
Component 1               WIP Warehouse              - 1                                                    $1

"Manufacture" Stock Entry

Item                              Warehouse                      Qty             Incoming Rate            Valuation Rate
Component 2               WIP Warehouse             - 1                                                     $1
Finish Goods 1             FG Warehouse                  1                $2                                $2

nabinhait:
Regarding accounting entry for "Material Consumption for Manufacture", let leave it to the users. Basically, user should be able to enter "Difference Account" manually and it can be either expense or asset account.
I will suggest to skip accounting entries for "Material Consumption for Manufacture". So, to do that we will set "Difference Account" as the WIP warehouse account. Hence, the system debit and credit the same account and resulting no accounting entries. And in "Manufacture" entry as well, we will set "Difference Account" as WIP warehouse. So, system will debit FG Warehouse account and credit WIP warehouse account.
If the user want to book expense while consumption, they can change it to "Cost of Goods Sold" account or any other. 
If the user want to maintain another intermediate asset account, they can manually select that account in the consumption entry.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -79,15 +81,26 @@ def on_submit(self):
...
-		self.update_stock_ledger()
-		self.update_work_order()
+
 		if self.purchase_order and self.purpose == "Subcontract":
 			self.update_purchase_order_supplied_items()
+
+		if self.work_order and self.purpose == "Material Consumption for Manufacture":
+			self.validate_work_order_status()
+		else:
+			self.update_stock_ledger()
```
nabinhait: `update_stock_ledger` should be outside `else` condition.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -126,11 +139,41 @@ def validate_item(self):
 				frappe.throw(_("Row #{0}: Please specify Serial No for Item {1}").format(item.idx, item.item_code),
 					frappe.MandatoryError)
 
+	def validate_qty(self):
+		manufacture_purpose = ["Manufacture", "Material Consumption for Manufacture"]
+
+		if self.purpose in manufacture_purpose and self.work_order:
+			if not frappe.get_value('Work Order', self.work_order, 'skip_transfer'):
+				item_code = []
+				for item in self.items:
```
nabinhait: This loop should only be run for raw materials. Add a if condition `if cstr(item.t_warehouse) = '':`

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -126,11 +139,41 @@ def validate_item(self):
...
+									and sed.item_code=%s and se.work_order= %s and ifnull(sed.t_warehouse, '') != ''
+							""", (item.item_code, self.work_order), as_dict=1)
+
+					stock_qty = flt(item.qty)
+					trans_qty = flt(transferred_materials[0].qty)
+					if req_items:
+						if stock_qty > trans_qty:
+							item_code.append(item.item_code)
+				if item_code:
+					frappe.throw(_("Consumption quantity cannot be more than transferred quantity for item/s {0}.").format(', '.join(item_code)), OverConsumptionError)
```
nabinhait: Should we do this validation? Because many cases, some unused raw materials (from earlier Work Orders) are there in WIP warehouse, which gets used. And it can lead to more consumption than transferred qty.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -506,6 +549,18 @@ def _validate_work_order(pro_doc):
 				if self.purpose == "Manufacture":
 					pro_doc.run_method("update_planned_qty")
 
+			if (self.purpose == "Material Consumption for Manufacture" or self.purpose == "Manufacture"):
+				for items in self.items:
+					if self.docstatus == 1:
+						frappe.db.sql("""Update `tabWork Order Item` set consumed_qty = consumed_qty + %s
```
nabinhait: Always avoid codes like this which does "+" and "- for updating qty. Use `sum` of all consumed qty against the work order, similar to https://github.com/frappe/erpnext/blob/develop/erpnext/manufacturing/doctype/work_order/work_order.py#L498

nabinhait (changes requested):

creamdory:
@nabinhait , what about the calculation of the Finish product valuation rate ?

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -727,20 +785,48 @@ def get_transfered_raw_materials(self):
...
+				if qty >= req_qty:
+					qty = (req_qty/trans_qty) * flt(self.fg_completed_qty)
+				else:
+					qty = qty - consumed_qty
+
+				if self.purpose == 'Manufacture':
+					if consumed_qty != 0:
+						remaining_qty = consumed_qty - (produced_qty * req_qty_each)
+						exhaust_qty = req_qty_each * produced_qty
+						if remaining_qty > exhaust_qty :
```
nabinhait: Can you add some docstrings in this whole calculation, to explain the use case?

nabinhait (changes requested):

nabinhait:
> This means the stock is consumed and deducted in the system but with zero accounting entries? And how do we make all the entries during finish? Should we get component cost from all the "Material Consumption for manufacture" for a work order?

Yes, it will not going to work, it is not that simple. What about if we book normal expenses while consumptions? And while updating finished goods, raw material expenses will be reverted, if we use same difference account.
Now, the main challenge is updating valuation rate for FG item. Even if take consumed material costs, we don't know against how many FG items, those raw materials are consumed.

creamdory:
@nabinhait , made the changes you requested. And about below:

> Yes, it will not going to work, it is not that simple. What about if we book normal expenses while consumptions? 

Currently, it is done like this now. "Material Consumption for Manufacture" behaves like a "Material Issue" only with tagged Work Order.

> And while updating finished goods, raw material expenses will be reverted, if we use same difference account. And expenses will be booked only for additional costs. Now, the main challenge is updating valuation rate for FG item. Even if it take consumed material costs, we don't know against how many FG items, those raw materials are consumed.

Can't be done since there is an option to do consumption in the "Manufacture" stock entry too when booked "Material Consumption for manufacture" is not done fully for the fg_completed_qty. And yes it will be challenging in the valuation rate calculation that is why I have suggested a WIP account that will be reverted in the "Manufacture" stock entry.

nabinhait:
@dottenbr @creamdory Documentation is updated for renaming Production Order to Work Order. YOu can start documenting this changes.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -727,20 +770,47 @@ def get_transfered_raw_materials(self):
...
+			consumed_qty = flt(req_items[0].consumed_qty)
+
+			if trans_qty and manufacturing_qty >= (produced_qty + flt(self.fg_completed_qty)):
+				if qty >= req_qty:
+					qty = (req_qty/trans_qty) * flt(self.fg_completed_qty)
+				else:
+					qty = qty - consumed_qty
+
+				if self.purpose == 'Manufacture':
+					# If Material Consumption is booked, must pull only remaining components to finish product
```
nabinhait: The same calculation is applicable if "Backflush Raw Materials Based On" is set as "BOM" in Manufacturing Settings. It should consider Consumption entry and pull only remaining items.

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -313,6 +313,11 @@ erpnext.work_order = {
 				if ((flt(doc.produced_qty) < flt(doc.material_transferred_for_manufacturing))
 						&& frm.doc.status != 'Stopped') {
 					frm.has_finish_btn = true;
+
+					frm.add_custom_button(__('Material Consumption'), function() {
```
nabinhait: It should not be shown if all items  are consumed

nabinhait (changes requested):
This entire functionality should be hidden by default. And can be enabled via a setting in Manufacturing Settings.

erpnext/stock/doctype/stock_entry/stock_entry.py
creamdory: @nabinhait this is what this code means.

nabinhait: But I think it does not work if "Backflush Raw Materials Based On" is set as "BOM" in Manufacturing Settings. I think `get_transfered_raw_materials` is not called for that use case.

creamdory:
@nabinhait 

Material Consumption in "Manufacturing Settings"
screen shot 2018-03-28 at 12 24 15

screen shot 2018-03-28 at 12 26 57

consumption

nabinhait:
@creamdory Change the label "Material Consumption" to "Allow Multiple Material Consumption".

creamdory:
@nabinhait 

screen shot 2018-03-28 at 13 57 46

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -656,10 +641,13 @@ def get_items(self):
...
 					self.add_to_stock_entry_detail(item_dict)
 
-				elif self.work_order and (self.purpose == "Manufacture" or self.purpose == "Material Consumption for Manufacture") and \
+				elif self.work_order and self.purpose == "Manufacture" and \
 					frappe.db.get_single_value("Manufacturing Settings", "backflush_raw_materials_based_on")== "Material Transferred for Manufacture":
 					self.get_transfered_raw_materials()
 
+				elif self.work_order and self.purpose == "Material Consumption for Manufacture":
+					self.get_transfered_raw_materials()
+
```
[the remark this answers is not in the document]
nabinhait: It is not right. If "backflush_raw_materials_based_on" is based on BOM, then it should not consider transferred materials. It will simply assume all the items in the BOM is in WIP warehouses.

creamdory:
@nabinhait requested changes made.

creamdory:
@nabinhait currently, documentation is broken. https://discuss.frappe.io/t/bug-documentation-broken/601

creamdory:
@nabinhait made the documentation. Please check.

djpalshikar:
@creamdory is there decision on the valuation then?

creamdory:
@djpalshikar currently not automated. Nabin says to change accounts during Stock Entries. Though we will think about it and make some enhancements. Currently, this feature can be enabled or disabled for this case.

PR #13399, opened 2018-03-26
[Minor] Project in Purchase Order, Purchase Invoice and Purchase Receipt

KanchanChauhan:


dottenbr:
We noticed that the Project Dashboard has those DocTypes in it but the DocTypes don't have the Project field to link them. We believe they should!

screen shot 2018-03-26 at 3 58 41 pm

nabinhait:
Project field is already there in Item table of those doctypes. We just have to fix the linking in the Project dashboard.

KanchanChauhan:
@nabinhait Why is it on Item since it is on doctype level in other doctypes like Sales Order, Production Order. May be it can be on doctype level as well.

nabinhait:
Because in purchase cycle, we can buy materials together for multiple projects for getting better deal from supplier. But in sales cycle, it is in our hand.

PR #13432, opened 2018-03-29
[Fix] Allow zero valuation rate if previous sle has zero valuation rate

rohitwaghchaure:


schilgod:
@rohitwaghchaure Is this change only for FIFO valuation method?

PR #13434, opened 2018-03-29
Link Share Transfer to Account

### LINKED ISSUE #13415: [Feature request]Link Share Transfer to Account
Can Share transfer doctype have two fields to show the accounts(Equity and Assets) related data for the account entries. During a share transfer, corresponding accounting entries take place which affect the balance sheet. We can add the necessary accounts on this doctype on which basis we can go ahead and make a JV. It should also be submittable because it is a transactional entry

If transfer_type = "Issue" or "Purchase" show both fields one with Equity accounts related data and other for Assets accounts related data

If transfer_type = "Transfer" show only one field with Equity accounts related data

deepak-mnt:
Fixes #13415
share_transfer1


erpnext/accounts/doctype/share_transfer/share_transfer.js
```diff
@@ -16,6 +16,22 @@ frappe.ui.form.on('Share Transfer', {
 				};
 			};
 		});
+		frm.set_query("equity_account", function() {
+			return {
+				filters: {
+					"root_type":"Equity",
```
nabinhait: Many times, we don't have any accounts with root_type "Equity" (For example, in India people mostly add equity related accounts under Liability). Hence, it will be better if we allow both Equity and Liability accounts here.

erpnext/accounts/doctype/share_transfer/share_transfer.js
```diff
@@ -16,6 +16,22 @@ frappe.ui.form.on('Share Transfer', {
...
+					"company": frm.doc.company
```
nabinhait: Also add a filter for "is_group: 0"

erpnext/accounts/doctype/share_transfer/share_transfer.js
```diff
@@ -16,6 +16,22 @@ frappe.ui.form.on('Share Transfer', {
...
+				}
+			};
+		});
+		frm.set_query("asset_account", function() {
+			return {
+				filters: {
+					"root_type":"Asset",
+					"company": frm.doc.company
```
nabinhait: is_group: 0

erpnext/accounts/doctype/share_transfer/share_transfer.py
```diff
@@ -12,7 +12,7 @@
 class ShareDontExists(ValidationError): pass
 
 class ShareTransfer(Document):
-	def before_save(self):
+	def before_submit(self):
 		if self.transfer_type == 'Issue':
 			company_doc = self.get_shareholder_doc(self.company)
```
nabinhait: Can you rename the variable "company_doc" to "shareholder", because it is actually not company_doc

nabinhait (changes requested):

deepak-mnt:
@nabinhait  Have made the suggested changes

erpnext/accounts/doctype/share_transfer/share_transfer.js
Tropicalrambler: @nabinhait 's proposal could be helpful for the future when one wishes to convert bonds issued to stock, or vice versa. I agree, both types should be allowed. Let the user choose, just add a configuration to allow to set the default account type for the filter.

deepak-mnt:
closing this because some file related to this pull request has been changed

PR #13467, opened 2018-04-01
[Fix] chart of accounts, root type is mandatory issue in charts.erpnext.com

### LINKED ISSUE #13126: Bugs in charts.erpnext.com
Reported by user via Support Issue /WN-SUP31248

- edit the chart of accounts: nothing happens when we click on the button start or fork.
- download chart of accounts: nothing happens when you click on download.
- create a new chart of accounts: nothing happens when you click on create a new chart of accounts

rohitwaghchaure:
**Issue**
before_fix_coa

**After Fix**
after_fix_coa

Fixed https://github.com/frappe/erpnext/issues/13126

### DIFF
--- a/erpnext/accounts/utils.py
+++ b/erpnext/accounts/utils.py
@@ -252,16 +252,11 @@ def add_ac(args=None):
 	if not ac.parent_account:
 		ac.parent_account = args.get("parent")
 
-	if getattr(ac, 'is_root', None):
-		ac.parent_account=''
-
 	ac.old_parent = ""
 	ac.freeze_account = "No"
 	if cint(ac.get("is_root")):
 		ac.parent_account = None
 		ac.flags.ignore_mandatory = True
-	else:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/utils.py
```diff
@@ -252,16 +252,14 @@ def add_ac(args=None):
...
-	if getattr(ac, 'is_root', None):
+	if cint(ac.is_root):
 		ac.parent_account=''
 
 	ac.old_parent = ""
 	ac.freeze_account = "No"
 	if cint(ac.get("is_root")):
 		ac.parent_account = None
 		ac.flags.ignore_mandatory = True
-	else:
```
nabinhait: Don't change here, instead pass is_root from js to fix the issue in charts.erpnext.com

nabinhait:
@krnkris No, charts.erpnext.com portal still does not support numbers. We will soon implement it.

PR #13470, opened 2018-04-01
Reqd Date column in Sales Order Item

creamdory:
Added "Reqd Date" in Sales Order Item in order to track the date when the customer wants the order to be delivered. Delivery Date is the date when the order is really delivered.

screen shot 2018-04-01 at 15 49 53


nabinhait:
The current delivery date is to track expected delivery date, which sometimes user uses to track "Reqd Date by customer" or "Expected delivery date by ourself". It is used for both purpose as per the need. I think adding another field will complicate the process for many users. Hence, will request you to add the field as a custom field in your custom app.

PR #13509, opened 2018-04-04
[fix] [ux] allow account to be created via form

rmehta:


Zlash65:
Working
account-create-from-form

Issue #13536, opened 2018-04-05
[Feature] Heatmap style (Discrete or Continuous) should can be set in Global Settings and transactions in heatmap of (see comment) should reflect the linked records from dashboard


joelios:
- [x] customer
 - [x] item
 - [x] company
 - [x] supplier
 - [x] Employee
 - [x] Project
 - [x] Member
- [x] Student

joelios:
implemented with #13537

PR #13537, opened 2018-04-05
[Heatmap] Implement feature #13536

joelios:
For this feature to work properly, the [Frappe pull request](https://github.com/frappe/frappe/pull/5387)  must also be merged!

status:
 - [x] customer
 - [x] item
 - [x] company
 - [x] supplier
 - [x] Employee
- [x] Project
 - [x] Member
- [x] Student


joelios:
belongs to feature #13536

erpnext/buying/doctype/supplier/supplier.py
```diff
@@ -48,3 +48,74 @@ def on_trash(self):
...
+		timestamp = get_timestamp(date)
+		out.update({ timestamp: count })
+
+	'''purchase order'''
+	items = dict(frappe.db.sql('''select transaction_date, count(*)
+		from `tabPurchase Order` where supplier=%s
+			and transaction_date > date_sub(curdate(), interval 1 year)
+			group by transaction_date''', name))
+
+	for date, count in iteritems(items):
```
pratu16x7: Thanks for the contribution :) While the finer control and overriding in every Doctype is great, I'm slightly worried about code duplication; for example, this `for` functionality that takes in `items` and updates `out` could be commonified as a utility function `def update_timeline_data(items, out)` that is always called with the most recently updated `out`. There are 3-4 occurrence of these same 5 lines across 8 files, I'm sure they can be reduced :)

pratu16x7:
Can you check the Codacy review?

erpnext/buying/doctype/supplier/supplier.py
pratu16x7: Another immediate side effect is having to use a `for` to update `out`; there are 3-4 occurrence of these same 5 lines of the loop across 8 files.

nabinhait:
Closing this as of now. Please reopen after fixes as per @pratu16x7 's suggestion.

erpnext/projects/doctype/project/project.py
```diff
@@ -267,13 +267,129 @@ def update_dependencies_on_duplicated_project(self):
...
+
+	for date, count in iteritems(items):
+		timestamp = get_timestamp(date)
+		if not timestamp in out:
+			out.update({timestamp: count})
+		else :
+			out.update({timestamp: out[timestamp] + count})
+
+	'''issue'''
+	items = dict(frappe.db.sql('''select creation, count(*)
```
rmehta: also might be a good idea to move this into functions as this query has a standard pattern.

rmehta:
also move this into a single function, it will shrink the volume of code by almost 80% 😉 

```
+	'''purchase receipt'''
+	items = dict(frappe.db.sql('''select posting_date, count(*)
+		from `tabPurchase Receipt` where supplier=%s
+			and posting_date > date_sub(curdate(), interval 1 year)
+			group by posting_date''', name))
+
+	for date, count in iteritems(items):
+		timestamp = get_timestamp(date)
+		if not timestamp in out:
+			out.update({timestamp: count})
+		else :
+			out.update({timestamp: out[timestamp] + count}) 
```

bghayad:
Hello @joelios 
Hope you are doing fine.
Is it possible to continue the PR so it can be merged and used instead of keep modifying the codes manually?
Regards
Bilal

PR #13558, opened 2018-04-06
Supplier Part selectable in Purchase Order Item

KanchanChauhan:
In some cases Supplier have different Supplier Part Number for the same item because of different brand so it should be selectable. Right now it auto sets top one.

screen shot 2018-04-06 at 1 48 25 pm
screen shot 2018-04-06 at 1 52 18 pm


joezsweet:
61f2dfe7b2 ..not that human readable ..old issue on link field

rmehta:
I agree with @joezsweet this is not good UX. I think you can use autosuggest to return relevant supplier parts directly. There is a new `Autocomplete` control now!

joezsweet:
@rmehta any example on how new Autocomplete control works?

PR #13564, opened 2018-04-07
Bank Statement Import

sathishpy:
Ok, I have managed to complete the doctype renaming. I haven't added the link from Account tab yet, will add once it is through the review process.
Apart from earlier changes, I have created a simple Bank doctype. The Statement Management is associated with the Bank Accounts, but the statement settings themselves are specific to a bank rather than account. So I have kept the similar structure with new Bank doctype. Ideally it should be added to Account doctype, so that we don't have to select both account and bank.


sathishpy:
My bad, I realised just now that I haven't updated the SQL queries as per rmehta's comment earlier. Perhaps will take a look in parallel to naming changes review.

sathishpy:
While submitting the payment requests, I got the error 'You are not authorized to add or update entries'. I didn't pay much attention thinking it is something to do with the fiscal year closing. I haven't done a upgrade in couple of months, so not really sure what change resulted in this. I would appreciate if someone throws some light instead of I trying to debug this which doesn't happen on our production srever.

sathishpy:
A quick look at the code and setting 'Frozen Account Modifier' role resolves that. We are having ERPNext: v10.0.7 on production and most of the development I did was during that time. It certainly felt that version was more robust than the current one. I had a tough time renaming the doctype(just didn't work). Just noticed submitting partial entries were modifying the the original table with some strange data(basically header). Not something I witnessed in 30+ transactions on production alone :(

sathishpy:
move_reconciled_entries() function moves the reconciled entries from initial transaction list to reconciled transaction list after submitting the payment entries. This is done by removing the entry from original and pushing it to the reconciled list. There is perhaps better way of handling this, but it looks like latest version isn't playing well with list entry removal. I will perhaps work on a better way of doing this next week.

sathishpy:
@nabinhait Let me know if you have any high level comments on this. I am looking at sql injection comment and other similar ones today. Earlier I asked if there is a way to log something important like creating a payment entry. At present I have some console print statements, which I would like to replace with some logger alternative

Issue #13527, opened 2018-04-04
Refactor Shopify


revant:
Related Links :

Woocommerce PR, https://github.com/frappe/erpnext/pull/13217

Webhook, verification and API documentation https://help.shopify.com/api/reference/webhook

saurabh6790:
- [ ] Move Shopify Settings and Tax Settings to ERPNext Integrations
- [ ] Create webhooks for Order Sync via private app on setup
- [ ] Pull orders from webhooks
- [ ] If Item not exists create one
- [ ] If custom not exists create the one

PR #13588, opened 2018-04-10
Refactor shopify

saurabh6790:
https://github.com/frappe/erpnext/issues/13527


erpnext/erpnext_integrations/connectors/shopify_connection.py
```diff
@@ -0,0 +1,283 @@
...
+					frappe.throw(_("Unverified Webhook Data"))
+			frappe.set_user(shopify_settings.modified_by)
+
+		return fn
+
+	return innerfn
+
+@frappe.whitelist(allow_guest=True)
+@validate_webhooks_request()
+def sync_salse_order(order=None):
```
revant: correction `sync_sales_order`

Issue #13596, opened 2018-04-11
Exchange Rate type for Selling or Buying


rmehta:
- In exchange rate add checkboxes for "for buying, for selling"
- In buying doctypes, use buying rate in selling rate, use selling rate
- By default both are checked
- Patch for creating and checking fields

PR #13644, opened 2018-04-13
Exchange Rate type for Selling or Buying

saurabh-bhosale:
#13596


nabinhait (changes requested):
Implement the actual usage.

In Selling module, fetch exchange rate only if selling is checked. In buying, fetch only if buying is checked.

Write patch to set "buying" and "selling" in all existing Exchange rate records.

PR #13649, opened 2018-04-13
Group by party in general ledger report

rohitwaghchaure:
Commonified Group By 
group_by_party

Fixed https://github.com/frappe/erpnext/issues/13556


nabinhait (changes requested):
Commonify "Group By" field, use a dropdown

nabinhait (approved):

Issue #13620, opened 2018-04-12
Attendance Request


rmehta:
- [x] Employee
- [x] From Date
- [x] To Date
- [x] Reason (Work From Home, On Duty)
- [x] Explanation

---

- on submission, create Attendance

pjamsheer:
- add attendaces for all days between from to dates (except for dates in holiday list, approved leaves)
- add Request reference in Attendance 
- On Cancel Cancel all attendace with reference to Request

pjamsheer:
#13662 - PR

Issue #13613, opened 2018-04-12
Parental Leave


rmehta:
### Leave Type

- [x] Check (is parental leave)
- [x] Max Leaves Allowed
- [x] Applicable After (Working Days)

- [x] Validate on Leave Application based on "Applicable After" and Number of Leaves

PR #13662, opened 2018-04-16
Attendance Request

pjamsheer:
- #13620 
- Add Company in Attendance Request
- Add attendaces for all days between from to dates (except for dates in holiday list, approved leaves)
- Add Request reference in Attendance 
- On Cancel Cancel all attendace with reference to Request
- #13613


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -41,6 +42,67 @@ def validate(self):
 	def on_submit(self):
 		self.validate_back_dated_application()
 
+	def validate_parental_leave(self):
+		if self.leave_type:
+			leave_type = frappe.get_doc("Leave Type", self.leave_type)
+			if leave_type.is_parental_leave == 1:
+				# Check on Applicable After
+				if leave_type.applicable_after > 0:
```
nabinhait: validate applicable_after and max_leaves_allowed for all leave_types, don't restrict to parental leave

erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -340,6 +402,36 @@ def get_approved_leaves_for_period(employee, leave_type, from_date, to_date):
 
 	return leave_days
 
+def get_total_approved_leaves_for_period(employee, from_date, to_date):
```
nabinhait: Why rewriting the same code twice?

It is same as `get_approved_leaves_for_period`.  Make leave_type args optional.

nabinhait (changes requested):

nabinhait (changes requested):
Add test cases

erpnext/hr/doctype/attendance_request/attendance_request.py
```diff
@@ -5,6 +5,62 @@
...
+
+	def on_cancel(self):
+		attendance_list = frappe.get_list("Attendance", {'employee': self.employee, 'attendance_request': self.name})
+		if attendance_list:
+			for attendance in attendance_list:
+				attendance_obj = frappe.get_doc("Attendance", attendance['name'])
+				attendance_obj.cancel()
+
+	def validate_date(self):
+		date_of_joining = frappe.db.get_value("Employee", self.employee, "date_of_joining")
```
nabinhait: We can do this in one query:
date_of_joining, relieving_date = frappe.db.get_value("Employee", self.employee, ["date_of_joining", "relieving_date"])

PR #13669, opened 2018-04-16
[fix] accounts receivable report fix

codingCoffee:
Print pdf for accounts receivable report used to use the customer doc
name instead of the customer_name

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.html
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.html
@@ -19,7 +19,7 @@
 
 
 <h2 class="text-center">{%= __(report.report_name) %}</h2>
-<h4 class="text-center">{%= filters.customer || filters.supplier %} </h4>
+<h4 class="text-center">{%= report.context_data.customer_name || filters.supplier %}</h4>
 <h6 class="text-center">
 		{% if (filters.tax_id) { %}
 		{%= __("Tax Id: ")%}	{%= filters.tax_id %}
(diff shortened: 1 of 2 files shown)


codingCoffee:
screenshot_20180416_160642

awesomescreenshot-2018-04-16t10-37-20-091z

screenshot_20180416_160959

erpnext/accounts/report/accounts_receivable_summary/accounts_receivable_summary.js
```diff
@@ -14,7 +14,14 @@ frappe.query_reports["Accounts Receivable Summary"] = {
 			"fieldname":"customer",
 			"label": __("Customer"),
 			"fieldtype": "Link",
-			"options": "Customer"
+			"options": "Customer",
+			on_change: () => {
+				var customer = frappe.query_report_filters_by_name.customer.get_value();
```
nabinhait: This change should be in accounts_receivable.js

erpnext/accounts/report/accounts_receivable/accounts_receivable.html
```diff
@@ -19,7 +19,7 @@
 
 
 <h2 class="text-center">{%= __(report.report_name) %}</h2>
-<h4 class="text-center">{%= filters.customer || filters.supplier %} </h4>
+<h4 class="text-center">{%= report.context_data.customer_name || filters.supplier %}</h4>
```
nabinhait: if filters.customer then `(report.context_data.customer_name || filters.customer)`
And the same is applicable for supplier as well in accounts_payable report

nabinhait (changes requested):

Issue #13599, opened 2018-04-11
Inter company Journal Entry


rmehta:
- [x] In account, add check for "Inter Company Account"
- [x] Type of Journal Entry "Inter Company Journal Entry"
- [x] Only allow selection of Inter Company Accounts
- [x] Once Inter company JE is submitted, show button for "Make Inter Company Journal Entry"
  - [x] Link both Journal Entries (doubly linked)
  - [x] Ensure sub of debits/credits is equal in both journal entries
  - [x] Once paired entry is made, don't show button
- [x] Before submitting inter-company entry validate that a paired entry exists in the other company in draft or submitted state.

snvora:
Hello,

I see the objective of this as to link inter-company transactions and validate them.

But a lot of transactions may happen between related companies including purchase, sale, lending, borrowing, i.e. through payment entries.

Do we plan on implementing that?

PR #13683, opened 2018-04-17
Inter Company Journal Entry

shreyashah115:
Summary:
- New voucher type 'Inter Company Journal Entry'.
- Checkbox `internal_company_account` in Accounts master.
- On submission of an Inter Company Journal Entry, button 'Make Inter Company Journal Entry' will appear.
- On clicking, select company for which Journal Entry is to be made.
- Route and map all the relevant fields (If same account names and same account currency), else skip the accounts child table.
- Link field `inter_company_journal_entry_reference` on both the linked Entries, will be unlinked if cancelled.
- Validation if total debit/ credit amounts differ while account currencies are same.

inter company jv
Fixes: #13599


nabinhait (changes requested):

PawanMeh:
Typically in an inter-company transaction, payable account in one company would be mapped to receivable account in another company. 

In the video above, it appears that we are just reversing the entry in the new company. Just want to make sure if the scenario stated above is handled.

shreyashah115:
Hi @PawanMeh, according to the video, the Salary account is being debited in the first entry, while it gets credited in the second entry, isn't that the scenario we need?

nabinhait:
@PawanMeh That means we can't actually populate the accounts automatically in the 2nd Journal Entry.
Or another option is to maintaining a mapping of inter-company accounts in account record.

snvora:
I am sorry, but I beg to differ.

Linking of accounts should be necessary for.
- understanding where make inter-company entry button is required
- setting off accounts on the consolidated balance sheet (if we ultimately plan in that direction).

Example how linking would probably work:

**Scenario**
Suppose there are two related companies A and B.

In books of A, there will be Account of B which should be linked to company B and vice versa in books of B.

So when Journal Entry is made with B's Account in books of A, then make inter-company entry button would appear.

Possible ways in which company A and B be related: So say in books of A

- From Chart of accounts --> Account of B
- B as customer or supplier of A.

In some scenarios, where related companies above i.e A and B are individual of a family,
- B could be an employee or sales commission agent.

Also, multiple accounts of B are possible in books of A
i.e. as a capital/loan account (from the chart of accounts) and as a Customer/Supplier

Hope it helps

nabinhait (changes requested):
- [ ] Rebase
- [ ] Add tests
- [ ] Add docs

nabinhait (approved):

Issue #13704, opened 2018-04-18
Alternative Item in the BOM and Work Order Level


creamdory:
## Current State

Alternative Items are set in the "Item Alternative" list. In the Manufacturing workflow for the current design, the alternative can only be chosen in the Stock Entry Level. This means that the BOM and the Work Order is already created. In reality, this is not the case. These are the problem we have encountered in our business flow:

1) There are some "Customer" specific alternatives and in some cases the original item and the alternative item is not even with the same technical description. Thus, you cannot set it as an official alternative for the item in the "Item Alternative" list. Therefore, the selection should not be limited only from the general "Item Alternative" setup.

2) Alternatives are BOM specific and not general. Though we have a general "Item Alternative" list, this must still be override in the BOM level. This alternative set-up will be for the specific BOM only.

3) BOM Alternative set-up should be the basis of the alternative the user can select in the Wok Order.

4) "Allow Alternative Item" must be per BOM line and should not be for the entire BOM to limit users to only allow alternative for specific lines.

## Goal

- [x] In the BOM, "Allow Item Alternative" must be set per Item and not in the entire BOM. It means that, we need to transfer below checkbox in the line:

<img width="1183" alt="bom1" src="https://user-images.githubusercontent.com/21003054/38918698-c98673da-42ee-11e8-98cd-c76bd9e0b06e.png">

- [x] Setting of item alternative in the BOM Line with the following features:

- Must be able to select from the alternative of the item already exist in the "Item Alternative" list.
- If preferred alternative is not in the list, must have the choice to select from the Item Doctype list.
- Alternative in this case is One-way.
- Alternative set in BOM WILL NOT be auto-saved in "Item Alternative" list.
- Indicator that the BOM Item has alternative setup.

- [x] Selection of Alternative in Work Order with the following features:

- User can select from the alternative of the item set-up in BOM
- Must have indicator that the Work Order Item is an alternative and not the original BOM item.

creamdory:
Hi @nabinhait  and @rohitwaghchaure , we are developing an enhancement of Item Alternative feature in ERPNext. Can you check the requirements above if this is clear? We are working on the PR for this.

kickapoo:
First level of BOM implementation 
 - In BOM level if Allow Alternative Item is set when is allowed to set Alternative Items in BOM Line (BOM Item)
- If selected Item "has alternative" in Item when ready only field in BOM Item is updated to indicate this and user can select from Alternative Item List as well
[...]

PR #13705, opened 2018-04-18
[WIP] Alternative items

creamdory:
From this issue https://github.com/frappe/erpnext/issues/13704 , we have this as an initial PR. We have duplicated and implemented current design from Stock Entry in both BOM and Work Order level. But this needs more enhancements from the description of the Issue above. The following features are needed:

- [ ] In the BOM, "Allow Item Alternative" must be set per Item and not in the entire BOM. It means that, we need to transfer below checkbox in the line:

bom1

- [ ] Setting of item alternative in the BOM Line with the following features:

- Must be able to select from the alternative of the item already exist in the "Item Alternative" list.
- If preferred alternative is not in the list, must have the choice to select from the Item Doctype list.
- Alternative in this case is One-way.
- Alternative set in BOM WILL NOT be auto-saved in "Item Alternative" list.
- Indicator that the BOM Item has alternative setup.

- [ ] Selection of Alternative in Work Order with the following features:

- User can select from the alternative of the item set-up in BOM
- Must have indicator that the Work Order Item is an alternative and not the original BOM item.

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -68,6 +68,15 @@ frappe.ui.form.on("BOM", {
 		frm.toggle_enable("item", frm.doc.__islocal);
 		toggle_operations(frm);
 
+		frm.set_indicator_formatter('item_code',
+			function(doc) {
+				if (doc.original_item){
+					return (doc.item_code != doc.original_item) ? "orange" : ""
+				}
+				return ""
+			}
+		)
+
 		if (!frm.doc.__islocal && frm.doc.docstatus<2) {
 			frm.add_custom_button(__("Update Cost"), function() {
 				frm.events.update_cost(frm);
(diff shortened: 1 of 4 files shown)


generare:
Could this same logic of "Alternative Items" be used for ecommerce product pages? It is common to have webstores suggesting alternative items on the product pages.

When the alternative items are something that the end-user would be interested in buying, e.g. while buying a tablet, would also like to buy a protective case for it, the admin could have set up this "protective case" as an alternative item below the product item page. These are normally called cross-selling items in different ecommerce platforms.

The other situation that alternative items should be visible is when they are up-selling products, e.g. better versions, more expensive versions of the item that the buyer is viewing.

Third options is that the alternative items are based on the client previous behaviour, so they are recommended products for the user.

In all of these cases the alternative items would be shown below the product info on the product page.

creamdory:
@rmehta sorry, we have to update this PR before merging. We have the full issue description here updated by @kickapoo  https://github.com/frappe/erpnext/issues/13704

rmehta:
Ok send another one. Please tag as "dont-merge" or close the PR if you are sure you don't want it merged.

We want to move towards a zero open PR policy.

creamdory:
Understood @rmehta . Yes we will send another one. Thanks!

PR #13709, opened 2018-04-18
Update manual-feedback-request.md

joolsr:
Changed wording to reflect current wording of Feedback Request rather than 'Ask a Feedback'


joolsr:
@rmehta Thanks. I fixed the docs when I started using these triggers ;-)

Issue #13697, opened 2018-04-18
In the Company master, Currency field should be up in the order


umairsy:
In the Company master, Currency field should be up in the order

PR #13718, opened 2018-04-18
Department tree improv and moved Currency field up the list in Company

Zlash65:
https://github.com/frappe/erpnext/issues/13697

Company :-
screen shot 2018-04-18 at 6 17 17 pm

Department :-
department-tree


nabinhait (dismissed):
- Department naming based on company abbr
- Patch for existing user
- Filter Parent Department based on company

PR #13737, opened 2018-04-19
fixes in trail balance report

codingCoffee:
- show account_number + account_name in trail balance report


erpnext/accounts/report/trial_balance/trial_balance.py
```diff
@@ -179,7 +179,7 @@ def prepare_data(accounts, filters, total_row, parent_children_map, company_curr
 	for d in accounts:
 		has_value = False
 		row = {
-			"account_name": d.account_name,
+			"account_name": "{} {}".format(d.account_number, d.account_name),
```
rmehta: only if account_number exists!

rmehta (changes requested):

PR #13746, opened 2018-04-20
[Enhance] Rename Supplier Type to Supplier Group - tree structure implemented

Zlash65:
https://github.com/frappe/erpnext/issues/13720

### Supplier Group

screen shot 2018-04-20 at 12 33 01 pm

### Supplier Group Tree

supplier-group-tree


rohitwaghchaure:
Hi @Zlash65 
Supplier group node pricing rule no applying on the child supplier group

1. Have created supplier group ABC with is group enabled and created one more supplier group ABC1 under group ABC

1. Set supplier group ABC1 in supplier Rakesh S

1. Created pricing rule for supplier group ABC(which is group node)

1. While making purchase order against the supplier Rakesh S, pricing rule is not applied

PR #13754, opened 2018-04-21
Patch fixes

shreyashah115:


amitsaxena01:
There is an error in the patch move_leave_approvers_from_employee.py
The believe sql should be 
frappe.db.sql("""update `tabDepartment Approver` app set app.parenttype = '{0}',
                                app.parent = '{1}' and parentfield = 'leave_approver' where app.approver = '{2}'"""
                                .format(_('Department'), record.department, record.leave_approver))

essadek:
Looks like the patch misses other fix. This is what the installation leads to after the update applied : 
`ImportError: Module import failed for POS Profile (erpnext.accounts.doctype.pos_profile.pos_profile Error: No module named cache_manager)`

PR #13775, opened 2018-04-24
Ageing Based on Supplier Invoice Date in Accounts Payable report

AravindPranera:
In Accounts Payable report, we can see the ageing days based on Supplier invoice date in Purchase invoice document.


nabinhait (changes requested):
Please rebase you branch with frappe/develop

PR #13776, opened 2018-04-24
[HotFix] Taxes not adding in POS sales invoice

rohitwaghchaure:


erpnext/controllers/accounts_controller.py
```diff
@@ -232,14 +232,19 @@ def set_taxes(self):
...
-		if self.is_new() and not self.get("taxes"):
+		if (self.is_new() or self.update_taxes()) and not self.get("taxes"):
 			if self.company and not self.get("taxes_and_charges"):
 				# get the default tax master
 				self.taxes_and_charges = frappe.db.get_value(tax_master_doctype,
 					{"is_default": 1, 'company': self.company})
 
 			self.append_taxes_from_master(tax_master_doctype)
 
+	def update_taxes(self):
```
nabinhait: The name `update_taxes` works if it is a flag. As it is a method, `is_pos_profile_changed` is better name.

erpnext/patches/v10_0/taxes_issue_with_pos.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	for d in frappe.get_all('Sales Invoice', fields=["name"],
+		filters = {'is_pos':1, 'docstatus': 1, 'creation': ('>', '2018-04-23')}):
+		doc = frappe.get_doc('Sales Invoice', d.name)
+		if (not doc.taxes and doc.taxes_and_charges and doc.pos_profile and
+			frappe.db.get_value('POS Profile', doc.pos_profile, 'taxes_and_charges') == doc.taxes_and_charges):
```
nabinhait: frappe.db.get_value('POS Profile', doc.pos_profile, 'taxes_and_charges', cache=True)

nabinhait (changes requested):

nabinhait (approved):

PR #13800, opened 2018-04-25
Production Plan, do not consider the sales orders against which work order has been created

### LINKED ISSUE #13279: Sales Order already processed (full/ partial) shows again in Production Planning Tool
[Prod Pln Ord Issue.pdf](https://github.com/frappe/erpnext/files/1802349/Prod.Pln.Ord.Issue.pdf)
[Prod Pln Ord Issue.pdf](https://github.com/frappe/erpnext/files/1802352/Prod.Pln.Ord.Issue.pdf)

e.g. If I created Sales Order by xyz Product Code with 4 Qty

then made Production order through Planning tool for 4 Qty

and again if i go to Production tool shows full Sales with Qty. It should not show  if Production entry done for full qty and should show only pending qty if Production done for partial qty.

rohitwaghchaure:
**Added work order quantity in sales order item**
screen shot 2018-04-25 at 7 33 25 pm

Fixed https://github.com/frappe/erpnext/issues/13279


nabinhait (changes requested):
Add tests

rohitwaghchaure:
@nabinhait 

Added Test Cases

PR #13808, opened 2018-04-26
Cost center number

### LINKED ISSUE #13717: Cost Center Number
Similar to account number

deepak-mnt:
fixes #13717

cost_center


deepak-mnt:
@nabinhait ,i have write the methods in such a way ,if we want same functionality on other tree too, then we can use this methods

PR #13820, opened 2018-04-28
Add Employee Grade

ahmad18189:
Employee Grade : 
- Make the base Employee Grade doctype 
-Add Employee Grade and Level into Employee
-In Salary Slip when adding employee:
--if employee does not have Grade:  the data from Add employees selected in base and varable 
--if employee is assigned to grade: the base will be equal  to Grade base salary + The Sum of the levels value (the salary calculation is factorial  means nth salary is equal to nth-1 * level percentage)
-in salary structure, you can fetch the salary details  value from grade by custom dialogue 

Employee Grade is used in other docs in the future 
-Business trips 
-over time application 
-End of service 
-Health Insurance


erpnext/hr/doctype/employee_grade/employee_grade.js
```diff
@@ -0,0 +1,71 @@
...
+				}
+			}
+		});
+	},
+	toggle_fields: function(frm) {
+		frm.toggle_display(['salary_component', 'hour_rate'], frm.doc.salary_slip_based_on_timesheet);
+		frm.toggle_reqd(['salary_component', 'hour_rate'], frm.doc.salary_slip_based_on_timesheet);
+		frm.toggle_reqd(['payroll_frequency'], !frm.doc.salary_slip_based_on_timesheet);
+	}
+	//~ ,
```
ahmadRagheb: remove commented lines no need for them

erpnext/hr/doctype/employee_grade/employee_grade.js
```diff
@@ -0,0 +1,71 @@
...
+	//~ level_percent : function(frm) {
+		//~ if (frm.doc.base >0)
+		//~ {
+			//~ frm.set_value("level_value",frm.doc.level_percent*frm.doc.base/100)
+		//~ }
+	//~ }
+});
+
+
+cur_frm.cscript.custom_level_value = function(doc, dt, dn) {
```
ahmadRagheb: cscript ,, lets but it in stander way .. 
custom_level_value : function(doc, dt, dn) { 
inside frappe.ui.form.on('Employee Grade', {

erpnext/hr/doctype/salary_structure/salary_structure.py
```diff
@@ -72,7 +72,35 @@ def strip_condition_and_formula_fields(self):
...
 			row.condition = row.condition.strip() if row.condition else ""
 			row.formula = row.formula.strip() if row.formula else ""
-
+	
+	def get_grade_info(self,employee_grade):
+		self.deductions = []
+		self.earnings = []
+		grade_doc = frappe.get_doc("Employee Grade",employee_grade)
+
+		for e in  grade_doc.earnings:
```
ahmadRagheb: lets use more meaningful name than e  as a variable

ahmadRagheb (changes requested):

Issue #13552, opened 2018-04-06
Show subscription schedule in Subscription (as HTML)


rmehta:

Siddhisawant8:
This Subscription schedule functionality should be similar to 'Maintenance Schedule'?

PR #13836, opened 2018-04-30
Subscription Schedule as HTML field

ghost:
#13552 

sub_schedule


rmehta:
The goal was to show schedule for probably the next one year based on the frequency! Also this has been moved to Auto-Repeat in develop. Closing this, you might have to re-do this.

cc @revant

PR #13837, opened 2018-04-30
Serial No. and Batch No. linking

### LINKED ISSUE #12635: Serial No. and Batch No. linking
Currently, Serial No. and Batch Nos. are not interlinked.

Use-case:

If an Item is both Serialized and batched, then in the Delivery Note and Sales Invoice, on the selection of Batch ID, all the relevant Serial Nos. (based on FIFO) should be auto-fetched.

shreyashah115:
Added `batch_no` field in Serial No DocType.

screen shot 2018-04-30 at 3 34 07 pm

In Delivery Note and Sales Invoice, serial nos. will be fetched on the basis of batch if item has `has_batch_no` and `has_serial_no` checked in its master.

serial batch

Serial nos. for which batch is not set will be fetched as its currently done, but serial nos. having a different batch than the one selected in the Delivery Note/ Sales Invoice will not be allowed to make the transaction.

serial batch 2

Closes: #12635

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -198,7 +198,6 @@ class AccountsController(TransactionBase):
 
 					if self.get("is_subcontracted"):
 						args["is_subcontracted"] = self.is_subcontracted
-
 					ret = get_item_details(args)
 
 					for fieldname, value in ret.items():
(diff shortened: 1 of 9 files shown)


pratu16x7 (approved):

federicocalvo:
Hi @shreyashah115, I've been trying beta.erpnext.com and I don't see the changes mentioned above.

This is provided for the V11?

Thanks!

shreyashah115:
@federicocalvo Yes, it will be provided in v11. We have fixed the bug. #14559

PR #13853, opened 2018-05-01
[Enhancement] UOM Conversion Factors

### LINKED ISSUE #5032: Feature Request - Conversion Factor Table - predefined table
I think that the UOM table is fantastic.  I like the fact that ERPNext supports PER ITEM conversion factors. But I consider that a great addition would be a generic, default table. This way, ERPNext could use the generic conversion factor as priority #1, and where a PER ITEM conversion exists, as is currently the case, use the PER ITEM conversion factor as priority #1. This would allow any conversion factor for a BOM, stock entry, or when entering a purchase receipt.
For example, two units of measure are defined in ERPNext. 
1. "Foot" referring to the English Foot measurement
2. "Metre" referring to the SI unit.

The conversion table would thus house that a "Foot", defined in terms of "Metre" is equivalent to 0.3048 Metre
Source: https://en.wikipedia.org/wiki/Foot_(unit)

Electrical wire is purchased from supplier, usually, by "Metre".   But the warehouse or field operator requires amount in "Foot" in Material Request (due to cultural and idiosincratyc behavioral habits that die hard, or is impractical to request change). When the purchase order is printed, the measurement is shown in "Metre", and stated that amount originally was expressed in "Foot".

The problem is that having a PER ITEM definition forces you to find the conversion factor each time, where some factors would be ideally pre-loaded into the system, with the SI and worldwide standards pre-loaded by ERPNext and the capability to define or customize the units.

shreyashah115:
Summary:

New DocType: UOM Category

screen shot 2018-05-01 at 5 25 41 pm

New DocType: UOM Conversion Factor

screen shot 2018-05-01 at 5 26 20 pm

- More than 240 conversions bootstrapped.
- In the item master, on selection of a UOM in the child table `UOM Conversion Detail`, on the basis of stock UOM, the conversion factor will be fetched from UOM Conversion Factor if both the UOMs belong to the same category else it will remain blank.

uom conversion factor

Closes: #5032


erpnext/stock/doctype/item/item.py
```diff
@@ -880,3 +880,25 @@ def check_stock_uom_with_bin(item, stock_uom):
 	if not matched:
 		frappe.throw(
 			_("Default Unit of Measure for Item {0} cannot be changed directly because you have already made some transaction(s) with another UOM. You will need to create a new Item to use a different Default UOM.").format(item))
+
+@frappe.whitelist()
+def get_uom_conv_factor(uom, stock_uom):
```
rmehta: test case for this?

rmehta (dismissed):
Add tests and docs!

erpnext/stock/doctype/item/test_item.py
```diff
@@ -241,6 +240,30 @@ def test_item_merging(self):
 		self.assertTrue(frappe.db.get_value("Bin",
 			{"item_code": "Test Item for Merging 2", "warehouse": "_Test Warehouse 1 - _TC"}))
 
+	def test_uom_conversion_factor(self):
+		if frappe.db.exists('Item', 'Test Item UOM'):
+			frappe.delete_doc('Item', 'Test Item UOM')
+		item_doc = make_item("Test Item UOM", {
+			"stock_uom": "Gram",
+			"uoms": [
```
rmehta: `'uoms': [dict(uom='Carat'), dict(uom='Kg')]`

nabinhait (approved):

nabinhait (approved):

PR #13902, opened 2018-05-03
[fix] consider user permissions in leave application calendar view

saurabh6790:
## Before

Showing all records though, the user permissions are applied
screen shot 2018-05-03 at 1 57 11 pm

---
## After

Showing records by considering user permissions
screen shot 2018-05-03 at 1 56 55 pm


erpnext/patches.txt
```diff
@@ -499,4 +499,5 @@ erpnext.patches.v10_0.update_hub_connector_domain
 erpnext.patches.v10_0.set_student_party_type
 erpnext.patches.v10_0.update_project_in_sle
 erpnext.patches.v10_0.fix_reserved_qty_for_sub_contract
-erpnext.patches.v10_0.taxes_issue_with_pos
\ No newline at end of file
+erpnext.patches.v10_0.taxes_issue_with_pos
+execute:frappe.db.set_value("HR Settings", None, "show_leaves_of_all_department_members_in_calendar", 1)
```
nabinhait: It will break, as HR Settings is not reloaded.

nabinhait (changes requested):

saurabh6790:
@nabinhait fixed

Issue #13833, opened 2018-04-30
Cannot make Purchase Order on demo.erpnext.com


netchampfaris:
On saving Purchase Order on demo.erpnext.com

```
Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/save.py", line 22, in savedocs
    doc.save()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 260, in save
    return self._save(*args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 283, in _save
    self.insert()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 222, in insert
    self.run_before_save_methods()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 869, in run_before_save_methods
    self.run_method("validate")
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 765, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1040, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1023, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 759, in 
    fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/buying/doctype/purchase_order/purchase_order.py", line 44, in validate
    super(PurchaseOrder, self).validate()
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/controllers/buying_controller.py", line 35, in validate
    super(BuyingController, self).validate()
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/controllers/stock_controller.py", line 17, in validate
    super(StockController, self).validate()
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/controllers/accounts_controller.py", line 41, in validate
    self.set_missing_values(for_validate=True)
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/controllers/buying_controller.py", line 70, in set_missing_values
    self.set_missing_item_details(for_validate)
[...]

Issue #13682, opened 2018-04-17
Default Price List is not auto-fetching in Sales Invoice 


PrakashH:
Default Price List is not auto-fetching in Sales Invoice. 

Steps: 
1) Open New Sales Invoice form
2) Select Customer
3) Scroll and Check CURRENCY AND PRICE LIST section

![screen shot 2018-04-17 at 2 24 12 pm](https://user-images.githubusercontent.com/4570016/38859186-1b67b51e-424b-11e8-83c8-c374b462ff4c.png)

Zlash65:
https://github.com/frappe/erpnext/pull/13923

PR #13926, opened 2018-05-04
[Minor] Python 2&3 related fix in use of filter function

Zlash65:
In Python 2, the filter() function returned a list, the result of filtering a sequence through a function that returned True or False for each item in the sequence. In Python 3, the filter() function returns an iterator, not a list.

Related Issues - https://github.com/frappe/erpnext/issues/13682 & https://github.com/frappe/erpnext/issues/13833


Zlash65:
fixed by https://github.com/frappe/erpnext/pull/13923

Issue #13632, opened 2018-04-12
Employee Onboarding and Offboarding


rmehta:
Employee Onboarding and Offboarding will be managed via tasks

### Employee Onboarding Template

- [x] Company
- [x] Grade
- [x] Designation
- [x] Activities (child table) (activity, user, role, required_for_employee_creation)

### Employee Onboarding

- [x] Employee Name
- [x] Job Applicant
- [x] Designation
- [x] Grade
- [x] Joining Date
- [x] Status
- [x] Project (created automatically from template, with Tasks)
- [x] Employee (after creation)
- [x] Make Employee (after pre-requisite activities are complete)

### Employee Offboarding Template

- [x] Company
- [x] Grade
- [x] Designation
- [x] Activities (child table) (activity, user, role, required_for_employee_creation)

### Employee Offboarding

- [x] Employee
- [x] Status
- [x] Project (created automatically from template, with Tasks)
- [x] Exit Interview Summary

### Project

- [x] Add "Department" in Project and Task
- [x] Create Project and Tasks based on activity
- [x] Set "Assign To" in Task based on user and role defined against activity. Check permissions based on User Permissions for department and company.
- [x] Validate Employee record based on required_for_employee_creation

PR #13984, opened 2018-05-10
[feature] added the employee onboarding and separation doctype

manassolanki:
Fixes https://github.com/frappe/erpnext/issues/13632

employee boarding


erpnext/hr/utils.py
```diff
@@ -6,11 +6,100 @@
...
+	'''
+	def validate(self):
+		# remove the task if linked before submitting the form
+		if self.amended_from:
+			for activity in self.activities:
+				activity.task = ''
+
+	def on_submit(self):
+		# create the project for the given employee onboarding
+		project_name = self.doctype + " for " + self.employee_name
```
rmehta: don't use "for" as it is not translatable, use ":" as a separator, also translate the doctype while naming

erpnext/hr/utils.py
```diff
@@ -6,11 +6,100 @@
...
+				"company": self.company
+			}).insert(ignore_permissions=True)
+		self.db_set("project", project.name)
+
+		# create the task for the given project and assign to the concerned person
+		for activity in self.activities:
+			task = frappe.get_doc({
+					"doctype": "Task",
+					"project": project.name,
+					"subject": activity.activity_name + " for " + self.employee_name,
```
rmehta: not translatable

erpnext/hr/doctype/employee_onboarding/employee_onboarding.js
```diff
@@ -0,0 +1,59 @@
...
+frappe.ui.form.on('Employee Onboarding', {
+	setup: function(frm) {
+		frm.add_fetch("employee_onboarding_template", "company", "company");
+		frm.add_fetch("employee_onboarding_template", "department", "department");
+		frm.add_fetch("employee_onboarding_template", "designation", "designation");
+		frm.add_fetch("employee_onboarding_template", "employee_grade", "employee_grade");
+	},
+
+	refresh: function(frm) {
+		if (frm.doc.employee) {
```
rmehta: dashboard

rmehta (changes requested):

PR #13951, opened 2018-05-08
Attendance Request, Leave Period, Compensatory Leave Request, Leave Policy

pjamsheer:
- #13620
- Add Company in Attendance Request
- Add attendaces for all days between from to dates (except for dates in holiday list, approved leaves)
- Add Request reference in Attendance
- On Cancel Cancel all attendace with reference to Request
- #13618
- #13612
- Leave Period - Grant Leave Allocation, set to_date based on from_date and Validate Overlap and Dates
- Leave Allocation - new field - link to Leave Period and Compensatory Leave Allocation
- #13613
- Validate on Leave Application based on "Applicable After"
- #13619
- Leave Type - new field - is_compensatory
- Leave Allocation - link to Compensatory Leave Request, validate max_leaves_allowed in leave type
- Compensatory Leave Request - filter is_compensatory on leave type, validate overlap and create leave allocation on submit

PR #13741, opened 2018-04-19
Attendance Request, Leave Period, Compensatory Leave Request, Leave Policy

pjamsheer:
- #13620
- Add Company in Attendance Request
- Add attendaces for all days between from to dates (except for dates in holiday list, approved leaves)
- Add Request reference in Attendance
- On Cancel Cancel all attendace with reference to Request
- #13618
- #13612
- Leave Period - Grant Leave Allocation, set to_date based on from_date and Validate Overlap and Dates
- Leave Allocation - new field - link to Leave Period
- #13613
- Validate on Leave Application based on "Applicable After" and Number of Leaves
- #13619
- Leave Type - new field - is_compensatory
- Leave Allocation - link to Compensatory Leave Request, validate max_leaves_allowed in leave type
- Compensatory Leave Request - filter is_compensatory on leave type, validate overlap and create leave allocation on submit

PR #13985, opened 2018-05-10
add get leave period method

saurabh6790:
```
ImportError: Module import failed for Leave Allocation (erpnext.hr.doctype.leave_allocation.leave_allocation Error: cannot import name get_leave_period)
```

### DIFF
--- a/erpnext/hr/utils.py
+++ b/erpnext/hr/utils.py
@@ -49,3 +49,9 @@ def update_employee(employee, details, cancel=False):
 			new_data = get_datetime(new_data)
 		setattr(employee, item.fieldname, new_data)
 	return employee
+
+
+def get_leave_period(from_date, to_date, company):
+	return frappe.db.sql("""select name from `tabLeave Period`
+		where is_active=1 and company=%s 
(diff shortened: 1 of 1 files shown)


rkurungadam:
Hi @saurabh6790 this would fix the import error, but the optional leave validation would still fail. issue came up with PR #13931 which had a dependancy on #13741 and I have mentioned too, and its not yet merged, jamsheer has closed and raised a new one #13951, would be great if you can review this.
Thanks

erpnext/hr/utils.py
```diff
@@ -49,3 +49,9 @@ def update_employee(employee, details, cancel=False):
 			new_data = get_datetime(new_data)
 		setattr(employee, item.fieldname, new_data)
 	return employee
+
+
+def get_leave_period(from_date, to_date, company):
```
nabinhait: this missing function is defined in https://github.com/frappe/erpnext/pull/13951/files. Please copy it from there.

nabinhait (changes requested):

saurabh6790:
Fixed!

PR #13987, opened 2018-05-11
[Fix] Sales Person Reports

shreyashah115:
Show quantity and amounts based on the delivered quantities if the Sales Order has been closed and all quantities weren't considered for further transactions.


nabinhait (dismissed):
delivered_qty is maintained in sales uom, you have to multiply it with conversion_factor

Issue #13551, opened 2018-04-06
Budget validation on Material Request


rmehta:
### Material Request
- [x] Add rate and amount in Material Request
- [x] Select Account and Cost Center (default based on item / item group / company)
- [ ] Fetch item rate based on last purchase rate (if exists) or default buying price list

### Budget
- [ ] Budget on parent accounts and cost center
- [ ] Control Section
     - [ ] Applicable on Material Request
     - [ ] Applicable on Purchase Order
     - [ ] Applicable on booking actual expenses
     - [ ] Separate Action fields for Material Request, PO and actual.
     - [ ] Exception Role for Material Request, PO and actual.

---

### Validation
- [ ] If Material Request is applicable, force to apply budget on PO and actual
- [ ] If applicable on Material Request -> expense = (Requested Amount + Ordered Amount + Actual amount)
- [ ] If applicable on Purchase Order but not applicable on MR -> Expense = (Ordered Amount + Actual amount)
- [ ] If only applicable on actual -> Expense = Actual amount

- Requested Amount  = Material Requested Amount pending to be ordered
- Ordered Amount  = Purchase Ordered Amount pending to be billed
- Actual Amount = sum(debit) - sum(credit) from GL Entry for cost center / project for the period

deepak-mnt:
@rmehta Budget validation should be on item quantity or item price because there is no item price factor on Material Request?

nabinhait:
It will be based on price. We have to introduce item price in Material Request.

PR #14020, opened 2018-05-14
Budget validation on material request

deepak-mnt:
#13551


rohitwaghchaure:
https://github.com/frappe/erpnext/pull/14061

PR #14081, opened 2018-05-16
Department Analytics Line Chart.

Solufyin:
Hello @rmehta , @revant

[ADD] Added Department Analytics Line Chart Features.

Added menu in Human Resources > Analytics > Department Analytics. See the below screenshot.

dept_analytics

emp

Department Analytics Chart that shows the department wise records.

chart2

Thanks.


Olawale1:
Hi, good job. Wouldn't this be better rendered as a Bar Chart?

Kind regards,

Solufyin:
@Olawale1 
We have also developed Department Analytics Chart rendered as Bar chart. Please see below snapshot.

hr1

hr2

Thanks.

erpnext/config/hr.py
```diff
@@ -9,136 +9,113 @@ def get_data():
...
-			"label": _("Leaves and Holiday"),
+			"label": _("Recruitment"),
 			"items": [
 				{
 					"type": "doctype",
-					"name": "Leave Period",
-				},
-				{
-					"type": "doctype",
-					"name": "Leave Policy",
```
nabinhait: Why removed?

erpnext/hr/report/department_analytics/department_analytics.py
```diff
@@ -0,0 +1,52 @@
...
+def get_employees(filters):
+	conditions = get_conditions(filters)
+	return frappe.db.sql("""select name, employee_name, date_of_birth,
+	branch, department, designation,
+	gender, company from `tabEmployee` where status = 'Active' %s""" % conditions, as_list=1)
+	
+def get_conditions(filters):
+	conditions = ""
+	if filters.get("department"): conditions += " and department = '%s'" % \
+		filters["department"].replace("'", "\\'")
```
nabinhait: Use `frappe.db.escape`

erpnext/hr/report/department_analytics/department_analytics.py
```diff
@@ -0,0 +1,52 @@
...
+	return conditions
+
+def get_chart_data(data):
```
nabinhait: The chart you implemented does not make much sense to me. Why employees are in X-axis and what are the Y-value means for a employee?
Actually it will be better if show the chart only if data is not filtered by any specific department. And in chart, we can show number of employees based on department. So, on X-axis, there will be department names and in Y-axis number of employees per department.
Also, it will look better in BAR chart.

nabinhait (changes requested):
Closing this, please reopen after fixing.

PR #14114, opened 2018-05-17
Total Quantity in Sales and Purchase DocTypes

### LINKED ISSUE #13199: Feature Request - Total Quantity in Documents with Item Table
Hi,
This is a pretty basic feature request 😅 

Use Case 
A read only field which shows the total quantity of the items in Sales Invoice, Purchase Invoice, Delivery Note, Purchase Receipt and other documents which has an Item Table with Quantity. 

Additionally, In Global Defaults have an option to Disable Total Quantity, like Disable Rounded Total or Disable in words

I do understand this can be achieved by using Custom script, I feel a lot of users ( who have no coding experience ) can benefit with this as a ERPNext Standard Feature, specially when available in print format builder for the user to show it in multipage print outs / pdfs

### LINKED ISSUE #14019: Total Qty field in all the sales and purchase transactions
Add "Total Qty" field in all the sales and purchase transactions.

shreyashah115:
Closes #14019 
Closes #13199 

total-qty


erpnext/patches/v11_0/update_total_qty_field.py
```diff
@@ -0,0 +1,27 @@
...
+	frappe.reload_doc('selling', 'doctype', 'quotation')
+	frappe.reload_doc('stock', 'doctype', 'delivery_note')
+	frappe.reload_doc('stock', 'doctype', 'purchase_receipt')
+	frappe.reload_doc('accounts', 'doctype', 'sales_invoice')
+	frappe.reload_doc('accounts', 'doctype', 'purchase_invoice')
+
+	doctypes = ["Sales Order", "Sales Invoice", "Delivery Note",\
+		"Purchase Order", "Purchase Invoice", "Purchase Receipt", "Quotation", "Supplier Quotation"]
+
+	for doctype in doctypes:
```
nabinhait: This patch might take too much time. The key of optimization is reducing number of queries. Here we can avoid sub-query, instead can get total_qty of all records of a doctype in a single query using GROUP BY. Then to update we can use CASE statement to update multiple records in a query.

nabinhait (changes requested):
Entry in patches.txt is missing

PR #14118, opened 2018-05-17
Alternative Item Enhancement

creamdory:
Full detail is here https://github.com/frappe/erpnext/issues/13704 .

Also related to this PR https://github.com/frappe/erpnext/pull/13705


rmehta (changes requested):

creamdory:
@rmehta we are currently fixing the request changes and the documentation. Thanks!

PR #14122, opened 2018-05-17
Fix in marking attendance from Leave Application

shreyashah115:


ashish-greycube:
Hi @shreyashah115 
Any particular reason why 'Mark attendance only if to_date <= nowdate in Leave Application'
I mean, say today is 15th Dec and I apply for leave on 18th Dec which is approved. So my understanding is that there should attendance record with 'on leave' for 18th Dec. Why we are stopping from doing that?Thanks!

PR #14154, opened 2018-05-20
Set leave approver in the leave application

rohitwaghchaure:


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -530,3 +530,8 @@ def add_holidays(events, start, end, employee, company):
 				"title": _("Holiday") + ": " + cstr(holiday.description),
 				"name": holiday.name
 			})
+
+@frappe.whitelist()
+def get_leave_approver_data(employee):
+	return frappe.db.get_value("Employee Leave Approver",
```
nabinhait: I think the table "Employee Leave Approver" has been deprecated. Check https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/department_approver/department_approver.py#L14

nabinhait (changes requested):

PR #14160, opened 2018-05-21
Fix patches for Travis

shreyashah115:


erpnext/patches/v11_0/rename_field_max_days_allowed.py
```diff
@@ -1,7 +1,10 @@
...
 
 def execute():
-	frappe.reload_doc("hr", "doctype", "leave_type")
+	frappe.db.sql("""
+		UPDATE `tabLeave Type`
+		SET max_days_allowed = '0'
+		WHERE trim(coalesce(max_days_allowed, '')) = ''
+	""")
 	frappe.db.sql_ddl("""ALTER table `tabLeave Type` modify max_days_allowed int(8) NOT NULL""")
-	rename_field("Leave Type", "max_days_allowed", "max_continuous_days_allowed")
```
nabinhait: `rename_field` is required. Old `max_days_allowed ` field has been renamed to "max_continuous_days_allowed" and a new field with same name introduced.

nabinhait (changes requested):

erpnext/patches/v11_0/rename_field_max_days_allowed.py
Zlash65: Could not find 'max_continuous_days_allowed' in any of the json files. Where was it introduced ? There is still a 'max_days_allowed' in leave_type.json

PR #14179, opened 2018-05-22
Project Budget does not filter per Project

jcbdigitalhub:
Project Budget does not filter per Project and always return actual expense amount of all entries tagged to a project.

### DIFF
--- a/erpnext/accounts/doctype/budget/budget.py
+++ b/erpnext/accounts/doctype/budget/budget.py
@@ -227,7 +227,7 @@ def get_actual_expense(args):
 			where lft>=%(lft)s and rgt<=%(rgt)s and name=gle.cost_center)"""
 	
 	elif args.budget_against_field == "Project":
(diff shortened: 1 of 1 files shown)


osstechnolab:
Can this be merged? Error was not related to the patch.

erpnext/accounts/doctype/budget/budget.py
```diff
@@ -227,7 +227,7 @@ def get_actual_expense(args):
 			where lft>=%(lft)s and rgt<=%(rgt)s and name=gle.cost_center)"""
 	
 	elif args.budget_against_field == "Project":
-		condition2 = "and exists(select name from `tabProject` where name=gle.project)"
+		condition2 = "and exists(select name from `tabProject` where name=gle.project and gle.project = %(budget_against_field)s)"
```
nabinhait: Thanks for the fix but it should be `%(budget_against)s`

nabinhait (changes requested):

PR #14249, opened 2018-05-28
Stocking out items with expired batches

### LINKED ISSUE #12843: Stocking Out Expired Batched Items
As per the current design, once item is expired, then no stock transaction is allowed for it. We will have to explicitly provide an option for the expired batched items to be stocked-out, so that expense could be booked for them. 

May be allow selection of Expired batches only in the Stock Entry should help?

shreyashah115:
stockout

Closes: #12843

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -155,6 +155,10 @@ frappe.ui.form.on('Stock Entry', {
 					}
 				})
 			}, __("Get items from"));
+
+			frm.add_custom_button(__('Expired Batches'), function() {
+
+			}, __("Get items from"));
 		}
 
 		if (frm.doc.company) {
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -170,6 +170,7 @@ frappe.ui.form.on('Stock Entry', {
 								d.qty = element.qty;
 								d.uom = element.stock_uom;
 								d.conversion_factor = 1;
+								d.batch_no = element.batch_no;
 								d.transfer_qty = element.qty;
 								frm.refresh_fields();
 							});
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1153,6 +1153,15 @@ def get_uom_details(item_code, uom, qty):
 		}
 	return ret
 
+@frappe.whitelist()
+def get_expired_batch_items():
+	return frappe.db.sql("""select b.item, sum(sle.actual_qty) as qty, sle.warehouse, sle.stock_uom\
```
nabinhait: select batch_no as well, it is required to set in stock_entry

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1153,6 +1153,15 @@ def get_uom_details(item_code, uom, qty):
...
+	from `tabBatch` b, `tabStock Ledger Entry` sle
+	where b.expiry_date <= %s
+	and b.expiry_date is not NULL
+	and b.batch_id = sle.batch_no
+	group by sle.warehouse""",(nowdate()), as_dict=1)
```
nabinhait: group by sle.warehouse, sle.item_code, sle.batch_no

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -156,6 +156,28 @@ frappe.ui.form.on('Stock Entry', {
...
+							frm.set_value("items", []);
+							r.message.forEach(function(element) {
+								let d = frm.add_child("items");
+								d.item_code = element.item;
+								d.s_warehouse = element.warehouse;
+								d.qty = element.qty;
+								d.uom = element.stock_uom;
+								d.conversion_factor = 1;
+								d.transfer_qty = element.qty;
+								frm.refresh_fields();
```
nabinhait: set batch_no

nabinhait (dismissed):

nabinhait (approved):

PR #14266, opened 2018-05-29
[UX Improvement] Removed uri encoding from search results on front end

sushantnadkar:
Issue: The search strings on the website were encoded and displayed, for example, maple "leaf", displayed as maple%20%22leaf%22
image

Solution: Remove uri encoding form search string.
image

Related to: [https://github.com/frappe/frappe/pull/5626](https://github.com/frappe/frappe/pull/5626
)


rmehta (changes requested):
use `html2text` see comment in frappe

sushantnadkar:
@rmehta made the changes, please check.

PR #14270, opened 2018-05-29
POS closing voucher

chdecultot:
Hi,

Still trying to solve the issue of daily cash reconciliation for the POS in ERPNext.
The Sales Payment Summary is a great report but it doesn't set in stone the state at the end of a particular day.
It is a legal requirement in France and some other countries as far as I understand (https://github.com/frappe/erpnext/issues/7582), so I'm proposing a new approach with a POS closing voucher:

At the end of each day (or period), the sales manager can review its POS sales and the split between all payment methods + taxes before submitting it:

image
image

This first version assumes that the sales manager makes the subsequent accounting entries manually (difference postings + cash out), but if this functionality is integrated in ERPNext I will work towards automating the postings. We should also add a POS opening voucher to register the cash level at the beginning of day also.

Thanks in advance for your feedback!


olamide2:
Great concept, but I think this closing voucher should also be done on a per user level and not just on a Pos Profile level.

Great idea though and the automated postings will be an awesome icing to the cake

chdecultot:
Hi @olamide2 ,

Thanks for your feedback.
Yes you are right, we should consider doing it on a per user basis for companies working with multiple POS profile/user.
It will be more complicated in terms of UX though, because a user can have several POS profiles corresponding to different point of sales counters that he/she needs to close separately. Maybe we should add some indications about which counter has been closed and which has not.
Let me see how I can modify it.

olamide2:
Hello @chdecultot

In my opinion it should be easy to conceptualize.

Let's have the option to filter by PoS Profile and/or also by user.

If you choose user alone then the result will be all transactions by the user across all PoS Profiles .

If you choose PoS Profile alone then the result will be all transactions done by all users assigned that PoS profile.

If you filter by PoS Profile and a particular user then the result will be transactions done by that user using that particular PoS Profile.

Am not a programmer so I have no idea how difficult or easy this should be. So forgive me if I sound naive.

Regards

chdecultot:
Hi @olamide2 ,

Sorry if I was not clear enough, but IMHO, the issue is not technical but functional in order to make it easy to close the POS counter.

I do agree with your conceptualization. Let's take the three cases you present:
1. _If you choose PoS Profile alone then the result will be all transactions done by all users assigned that PoS profile._
That is the case implemented above: for a given counter, each cashier has an individual POS Profile and starts with some cash on hand. At the end of the day, he/she can close its counter by entering the detail of each sum for each payment method.

For the sales manager: we should have a dashboard indicating which counter has been closed and which has not.

2. _If you filter by PoS Profile and a particular user then the result will be transactions done by that user using that particular PoS Profile._
This case assumes that you have one POS Profile for several users.
Each POS Profile corresponds to a particular POS counter, so by filtering by POS profile and then by user, you get the amounts collected for each payment method for this particular user on each different POS counter.
The difficulty here is to not miss any combination of POS Profile/User when closing the POS counters at the end of the day.

For the sales manager: same as n°1. We need a dashboard indicating which combinations have been closed and which have not.

3. _If you choose user alone then the result will be all transactions by the user across all PoS Profiles ._
I guess that if you have several POS Profiles, it means that you have different POS counters. Given that each counter should be closed separately (at least in France), in my understanding it is similar to option n°2.

An option could be to make a report showing the total sales per user (already available in ERPNext).


Please don't hesitate to tell me if you are handling things differently ! Especially if you really need to make one closing per user using several POS Profile.
In the meantime, I'll start implementing option n°2.

Thanks!

olamide2:
We are a multi location retail business. And each location has multiple pos users

We use Pos Profile to distinguish the users in each location.

So we set up a PoS profile for each location and assign the users to that profile.

This is why in our situation it would be convenient to be able to close per uer,  by pos profile and a combination of the two.

This is our use case.

Regards

chdecultot:
Hi @olamide2 ,

Thanks for your feedback.
Your business case seems in fact quite common.

What I don't understand is why you would close counters in different locations at the same time ? Correct me if I'm wrong, but is it not safer to ask a cashier to close his/her counter at the end of his/her shift ?

Are the cash funds in location A not accounted differently from the cash funds in location B, even if it's the same user working in the morning in the location A and in the afternoon in location B ?
Or is the user taking his/her cash drawer with him/her to location B ?

olamide2:
Hi @chdecultot

Cashiers in each location run shifts.

If we have 6 cashiers in a location they would all have the same PoS profile. But we may need to close for all cashiers at different times.

This is why we would need to be able to filter for user when closing and not just PoS profile.

Hence the need for the user filter option and not just the PoS profile option.

chdecultot:
Hi @olamide2,

Great! So it will be fine with the implementation of the case n°2.
You will be able to select a POS Profile + a User to make more detailed closing vouchers.

My concerns were only related to the need for case n°3.
If I'm still missing a point please let me know!

olamide2:
Great!

All is well.

Thanks for the good job 

Olamide

chdecultot:
Based on the feedback from @olamide2 , please find screenshots of the updated PR:

image
image


The voucher is not edited per POS Profile and Cashier (User).

olamide2:
Perfect!

nabinhait (approved):

Muzzy73:
Hi @chdecultot. Thanks for the PR. It looks good. You have mentioned POS opening voucher but I could find anything in your PR. My question I how do cashier inputs the amount when opening a session. Lets say starts with $10 as change in the cash drawer. Makes cash sales of $100 (ignoring tax for here). Hence at closing the session cash should be $110 in the cash drawer. Thanks for the help.

chdecultot:
Hi @Muzzy73,

> We should also add a POS opening voucher to register the cash level at the beginning of day also.

This is not implemented in this PR, but it would be indeed great to implement and integrate it with the POS closing voucher. As you state it is necessary to implement a POS opening voucher and follow the amount of cash in the drawer at the beginning and end of the day.
I have it on my todo list, but I have no idea when I will have the bandwidth to develop it.

Muzzy73:
Thank you. I have asked my developer to work on opening cash based on your design. If successful then will send PR.

olamide2:
Hello @chdecultot 
Just noticed that you can actually create multiple period closing vouchers for the same day for the same user. I am of the opinion that this should not be possible. What do you think ?

olamide2:
I actually have no idea why the Cashier closing feature was introduced and merged. I tested it and in my opinion it is not as well structured as your PoS Closing Voucher. Your Concept is almost at the finish line. If you do have the time consider adding the following features

1. The duplication check as stated above
2. The ability to have a ¨blind closing¨ where the cashier simply keys in the on hand cash and credit card figures without knowing what the system balances are, a more senior officer can then review the shortages/overages

Like I said a lot of thought went into your design and I think you should try and finish it off.

Good job

f-9t9it:
@chdecultot
Kindly check the [#15273](https://github.com/frappe/erpnext/pull/15273) 
I have updated some things like opening and closing cash.

PR #14282, opened 2018-05-30
Documentation to Fetch value in Child Table for a field

jigneshpshah:
Sample Script to fetch value in child table


Olawale1:
Thanks @jigneshpshah 

We need as many useful sample scripts as possible. I look forward to many more getting added

Much appreciated

nabinhait (approved):

PR #14288, opened 2018-05-30
Apply GST based on Origin and Place of supply GST Code

### LINKED ISSUE #13466: need item wise and state wise auto application of GST based on HSN /SAC codes
GST rate Calculation Based on Item's HSN / SAC code
SGDT CGST IGST Prediction Based on Origin and Place of supply GST Code

shreyashah115:
Added a checkbox `is_inter_state` in Sales Taxes And Charges Template and Purchase Taxes And Charges Template.
On selection of a customer in Sales Invoice or Delivery Note, if the GST codes of place of supply and customer/ shipping address doesn't match, inter state template will be set as the taxes template and in the opposite scenario, default taxes template will be set.
Same is the process for Purchase Invoice, on selection of Supplier, the aforementioned procedure takes place.

Inter State GST:
gst-auto
Intra State GST:
gst-auto-2

Closes: #13466


erpnext/regional/india/utils.py
```diff
@@ -61,7 +62,7 @@ def get_itemised_tax_breakup_data(doc):
 
 	return hsn_tax, hsn_taxable_amount
 
-def set_place_of_supply(doc, method):
+def set_place_of_supply(doc, method=None):
 	if not frappe.get_meta('Address').has_field('gst_state'): return
```
nabinhait: Call `get_place_of_supply` here

nabinhait (dismissed):

nabinhait (approved):

Issue #13719, opened 2018-04-18
Supplier master - Tax ID should have Unique Property enabled


umairsy:
Make Tax ID field unique in the Supplier master for the validation to prevent duplication of supplier entry.

If reasonable, we can enable Unique check for the Customers as well.

Also, we can make it configurable where user can select a validation criteria:

- [ ] Supplier ID + Tax ID + Pan No.
- [ ] Tax ID + Pan No.
- [ ] Pan No.
- [ ] Tax ID

User should be able to configure it from Buying Settings.

PR #14313, opened 2018-06-01
Supplier master :validation to prevent duplication of supplier entry

deepak-mnt:
Closes: #13719
new_buying_setting
new_supplier
new_selling_setting
new_customer


rmehta:
@deepak-mnt you can't hardcode this.

If you want to do this right, in frappe

1. Create a new doctype for `Uniqueness Rule`
 - fields: doctype, child table of fields

2. On save of the rule, update database constraints.

PR #14358, opened 2018-06-05
Salary Structure, Slip - make fields readonly

rkurungadam:


erpnext/hr/doctype/salary_structure/salary_structure.js
```diff
@@ -61,6 +61,11 @@ frappe.ui.form.on('Salary Structure', {
 				frappe.set_route('Form', 'Salary Structure Assignment', doc.name);
 			});
 		}
+		frm.set_df_property("is_tax_applicable", "read_only", 1, "Salary Detail", "earnings");
+		let fields_read_only = ["is_tax_applicable", "is_flexible_benefit", "variable_based_on_taxable_salary"];
```
nabinhait: Why not used `read_only` property of docfield?

nabinhait (changes requested):

erpnext/hr/doctype/salary_structure/salary_structure.js
rkurungadam: its marked readonly, should we allow form customization to change it? line 64 is by mistake, removing it.

nabinhait (approved):

PR #14372, opened 2018-06-07
Cancel attendance if leave application is cancelled

shreyashah115:


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -138,6 +139,13 @@ def update_attendance(self):
 						doc.insert(ignore_permissions=True)
 						doc.submit()
 
+	def cancel_attendance(self):
+		if self.docstatus == 2:
+			attendance = frappe.db.sql("""select name from `tabAttendance` where employee = %s\
+				and (attendance_date between %s and %s) and docstatus < 2""",(self.employee, self.from_date, self.to_date), as_dict=1)
```
nabinhait: and status is in ('On Leave', 'Half Day')

nabinhait (dismissed):

nabinhait (approved):

PR #14423, opened 2018-06-08
Add Employee Status Temporary Leave

creamdory:
What type of a PR is this? 
- [x ] Changes to Existing Features

Description:

There are situations that Employees will be on a long leave like Maternity and Sabbatical leaves. The employee is still active but it would be efficient if we will know at one look that the Employee is on leave for long and is temporarily inactive.

In this PR, we have added "Temporary Leave" as an option in the Employee Status.

screen shot 2018-06-08 at 2 10 00 pm

screen shot 2018-06-08 at 2 10 08 pm


nabinhait (approved):

pm-at:
The change seems to have been revoked in employee doctype in commit d3877c83c6d72b6d4c73df47c26b2b238febbe8e / PR #14758. Hope, that was intentional? 

If yes, then need to remove from below too
https://github.com/frappe/erpnext/blob/096b943b0cab3807777dee933a725299fc810766/erpnext/hr/doctype/employee/employee_list.js#L6
https://github.com/frappe/erpnext/blob/9002a6c1951efaf970e67297963dcb456bebcb0f/erpnext/hr/doctype/employee/employee.py#L40

Issue #14429, opened 2018-06-08
[Bug] loan application rate of interest doesnot update on select of loan type


shridarpatil:
Issue: Bug report

Rate of interest is not updating on select of loan type, it shows 0%

**To Reproduce** :chart_with_downwards_trend:
Steps to reproduce the behaviour:
1. Loan application
2. Click on create new 
3. Enter details and save 
4. Check Rate of Interest after save it shows 0%

**Expected behaviour** :chart_with_upwards_trend:
Rate of interest should be interest which is added while creating loan type.
 
**Screenshots** :crystal_ball:
issue_template
![image](https://user-images.githubusercontent.com/11792643/41161683-538a37a8-6b51-11e8-809f-e466d430aa61.png)

**Version Information**
- Which branch are you on?
    - [ ] `develop` :fire:
- ERPNext Version:  v11.x.x-develop (19ce904) (develop)

PR #14431, opened 2018-06-08
Loan application get rate of interest an update

shridarpatil:
https://github.com/frappe/erpnext/issues/14429


nabinhait:
Seems like this has already been fixed via https://github.com/frappe/erpnext/commit/33df0b4fbfdbe9225b5392bf2f5438249aa47fa4#diff-d6c1795a777eeac0537b691aa9de4f27R404

PR #14492, opened 2018-06-13
Fix conflict for taxable value having same item in GSTR-1 Report

finbyz:
In the current GRSR-1 report, if there are multiple same items in single parent, then it does not adds to the invoice_items values which also creates problem with Taxable Value.

demo1

So the revised code with different logic which resolves the above issue.

demo2

### DIFF
--- a/erpnext/regional/report/gstr_1/gstr_1.py
+++ b/erpnext/regional/report/gstr_1/gstr_1.py
@@ -144,7 +144,7 @@ class Gstr1Report(object):
 		""" % (self.doctype, ', '.join(['%s']*len(self.invoices))), tuple(self.invoices), as_dict=1)
 
 		for d in items:
-			if not self.invoice_items or d.item_code not in self.invoice_items.values():
+			if d.item_code not in self.invoice_items.get(d.parent, {}):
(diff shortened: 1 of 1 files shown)


erpnext/regional/report/gstr_1/gstr_1.py
```diff
@@ -144,17 +144,8 @@ def get_invoice_items(self):
...
-
-			if d.parent in self.invoice_items:
-				parent_dict = self.invoice_items[d.parent]
-				if d.item_code in parent_dict:
-					item_details[d.item_code] += parent_dict[d.item_code]
-				else:
-					item_details.update(parent_dict)
-
-			self.invoice_items[d.parent] = item_details
+			if not self.invoice_items or d.item_code not in self.invoice_items.values():
```
nabinhait: I think `self.invoice_items.values()` is not correct.

It should be `self.invoice_items.get(d.parent, {})`

nabinhait (changes requested):

erpnext/regional/report/gstr_1/gstr_1.py
finbyz: True. Your requested changes is done.

nabinhait (approved):

PR #14495, opened 2018-06-13
Fetch details from Journal Entry in Accounts Payable report

shreyashah115:
Before:

screen shot 2018-06-13 at 12 32 38 pm

Now:

screen shot 2018-06-13 at 12 24 48 pm


PawanMeh (commented):
Instead of "null as due_date" maybe it would be better to use ifnull(due_date, ''). I think same would apply to bill_date as well.

shreyashah115:
@PawanMeh Done. But, `bill_date` column does exist in Journal Entry?

shreyashah115:
@PawanMeh My bad! I thought `due_date` doesn't exist in Journal Entry.

nabinhait (approved):

PR #14532, opened 2018-06-15
Renamed asset adjustment to asset value adjustment

rohitwaghchaure:


erpnext/patches/v11_0/rename_asset_adjustment_doctype.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+import frappe
+from frappe.model.rename_doc import rename_doc
+
+def execute():
+	rename_doc('DocType', 'Asset Adjustment', 'Asset Value Adjustment', force=True)
```
nabinhait: if Asset Adjustment exists, then only rename

nabinhait (changes requested):

erpnext/patches/v11_0/rename_asset_adjustment_doctype.py
rohitwaghchaure: Fixed!

nabinhait (approved):

PR #14551, opened 2018-06-18
[Fix]Check for existing permission before adding

KanchanChauhan:
- Every time a change is made to the User record, the User Permission are added again making duplicate permission records.
screen shot 2018-06-18 at 9 50 00 am

- Added code to check for existing permissions.

### DIFF
--- a/erpnext/hr/doctype/employee/employee.py
+++ b/erpnext/hr/doctype/employee/employee.py
@@ -68,8 +68,10 @@ class Employee(NestedSet):
 		if not self.create_user_permission: return
 		if not has_permission('User Permission', ptype='write'): return
 
-		add_user_permission("Employee", self.name, self.user_id)
-		set_user_permission_if_allowed("Company", self.company, self.user_id)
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/employee/employee.py
```diff
@@ -68,8 +68,10 @@ def update_user_permissions(self):
 		if not self.create_user_permission: return
 		if not has_permission('User Permission', ptype='write'): return
 
-		add_user_permission("Employee", self.name, self.user_id)
-		set_user_permission_if_allowed("Company", self.company, self.user_id)
+		if not frappe.db.exists({'doctype':'User Permission','user':self.user_id, 'allow':'Company', 'for_value': self.company}):
+			set_user_permission_if_allowed("Company", self.company, self.user_id)
+		if not frappe.db.exists({'doctype':'User Permission','user':self.user_id, 'allow':'Employee', 'for_value': self.name}):
```
nabinhait: This is already checked within `add_user_permission`. There was an error in `add_user_permission `, thats why it was creating duplicate. We have fixed that, also written a patch to remove the duplicate.

Issue #14404, opened 2018-06-07
v11][test] Enhanced Fixed Asset Management


nabinhait:
- CWIP Accounting
- Serialised asset
- Improved workflow for managing assets
- Multiple depreciation schedule based on finance book
- Asset value adjustment
- Improved Asset Movement between location or employee
	- Issue
	- Receive
	- Movement

PawanMeh:
#14463

PawanMeh:
#14522

PR #14572, opened 2018-06-18
Added test cases for asset accounting, asset value adjustment

rohitwaghchaure:
Added test cases for 

- CWIP Accounting

- Asset received but not billed accounting

- Asset value adjustment test cases

- Asset movement among the locations and the employees

- Added validation that user must create the asset from the Purchase Receipt or Purchase Invoice with update stock

Issue https://github.com/frappe/erpnext/issues/14404


erpnext/assets/doctype/asset/asset.py
```diff
@@ -71,8 +71,19 @@ def validate_asset_values(self):
 		if not flt(self.gross_purchase_amount):
 			frappe.throw(_("Gross Purchase Amount is mandatory"), frappe.MandatoryError)
 
+		if not (self.purchase_receipt or self.purchase_invoice):
```
nabinhait: These validations will be only applicable for non-existing assets

nabinhait (approved):

erpnext/assets/doctype/asset/asset.py
rohitwaghchaure: Fixed!

nabinhait (approved):

PR #14584, opened 2018-06-19
improve performance of stock_value retrieval using tabBin instead of …

sachinhub:
Currently to retrieve stock value, queries are done on `tabStock Ledger Entry` table which results in entire table scan.

Refer discuss thread @ 

https://discuss.erpnext.com/t/erpnext-performance-warehouse-tree-fetches-stock-value-of-all-children/37524/3

### DIFF
--- a/erpnext/stock/utils.py
+++ b/erpnext/stock/utils.py
@@ -11,21 +11,47 @@ from six import string_types
 
 class InvalidWarehouseCompany(frappe.ValidationError): pass
 
+def get_stock_value (warehouse=None, item_code=None):
+
+	values = {}
+	conditions = ""
+	if warehouse:
+		conditions += """ and warehouse in (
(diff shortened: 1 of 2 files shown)


erpnext/stock/utils.py
```diff
@@ -11,21 +11,47 @@
 
 class InvalidWarehouseCompany(frappe.ValidationError): pass
 
+def get_stock_value (warehouse=None, item_code=None):
```
rmehta: `get_stock_value_from_bin` ?

rmehta (changes requested):

erpnext/stock/utils.py
sachinhub: Ok. Done.

sachinhub (commented):
Okay. Done

PR #14588, opened 2018-06-19
[security][fix] Sanitize searchfields to avoid sql injection

saurabh6790:
Test-cases pending...


nabinhait (commented):
Instead of sanitizing here, we can do it where it is called on frappe (search.py ?).

It will be very hard to remember to add `@sanitize_searchfields` for every new function.

PR #14609, opened 2018-06-20
Uniqueness Rule

### LINKED ISSUE #13719: Supplier master - Tax ID should have Unique Property enabled

deepak-mnt:
Closes: #13719
uniqueness_rule

new_msg


deepak-mnt:
@shreyashah115 have made suggested changes

rmehta:
This must be part of frappe !

PR #14634, opened 2018-06-22
Employee benefit - Last payroll period benefit

pjamsheer:
- Employee Benefit - Last payroll period benefit calculation

### DIFF
--- a/erpnext/hr/doctype/employee_benefit_claim/employee_benefit_claim.py
+++ b/erpnext/hr/doctype/employee_benefit_claim/employee_benefit_claim.py
@@ -115,7 +115,7 @@ def get_benefit_claim_amount(employee, start_date, end_date, salary_component):
 		return claimed_amount
 	return False
 
-def get_total_benefit_despensed(employee, sal_struct, sal_slip_start_date, payroll_period):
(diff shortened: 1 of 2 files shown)


erpnext/hr/doctype/employee_benefit_claim/employee_benefit_claim.py
```diff
@@ -91,23 +71,111 @@ def get_pro_rata_amount_in_application(self, payroll_period):
...
 	}, as_dict = True)
-
 	if benefit_claim_details:
 		claimed_amount = 0
 		for claim_detail in benefit_claim_details:
 			claimed_amount += claim_detail.claimed_amount
 		return claimed_amount
 	return False
+
+def get_total_benefit_despensed(employee, sal_struct, sal_slip_start_date, payroll_period):
```
nabinhait: dispensed?

nabinhait (changes requested):

erpnext/hr/doctype/employee_benefit_claim/employee_benefit_claim.py
pjamsheer: Method name refactored

nabinhait (approved):

PR #14670, opened 2018-06-25
Add configurable frequency in Selling Settings for update of project …

sunnyakaxd:
…and company

Pull-Request

- [X] Have you followed the guidelines in our Contributing document?
- [X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [X] Have you lint your code locally prior to submission?
- [X] Have you successfully run tests with your changes locally?
- [X] Does your commit message have an explanation for your changes and why you'd like us to include them?
- [X] Docs have been added / updated
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Did you modify the existing test cases? If yes, why?

---

What type of a PR is this? 

- [X] Changes to Existing Features
- [ ] New Feature Submissions
- [ ] Bug Fix
- [ ] Breaking Change

--- 

Creating a Sales Order / Sales Invoice used to update Project and Company documents. If 100s of invoices are being generated per minute, this becomes a bottleneck as the update reads all the Sales Orders and Sales Invoices to calculate the new value.  This could be scheduled at Daily/Monthly triggers to reduce the overhead on system for creation of each document.
(Discuss Reference: [Monthly sales calculation on every Sales Invoice is slowing down ErpNext](https://discuss.erpnext.com/t/monthly-sales-calculation-on-every-sales-invoice-is-slowing-down-erpnext/37168) )

For which, this PR adds a configurable setting in "Selling Settings" to defer the billing and sales calculation at midnight instead of each transaction. However, the default will still stay at Each Transaction for new setups, which can be changed later.

The scheduler code is only added for project as functionality is already present to daily calculate company monthly sales.

image

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -412,22 +412,29 @@ def create_project_update(project):
 			)
 	return data
 
-def update_project_daily():
-	if frappe.db.get_single_value("Selling Settings", "sales_update_frequency") != "Daily":
+def update_project_sales_billing():
+	sales_update_frequency = frappe.db.get_single_value("Selling Settings", "sales_update_frequency")
+	if sales_update_frequency == "Each Transaction":
 		return
-	for project in frappe.get_all('Project'):
-		project = frappe.get_doc('Project', project.name)
-		project.update_sales_amount()
-		project.update_billed_amount()
-		project.save()
+	elif (sales_update_frequency == "Monthly" and frappe.utils.now_datetime().day != 1):
+		return
 
-def update_project_monthly():
-	if frappe.db.get_single_value("Selling Settings", "sales_update_frequency") != "Monthly":
-		return
-	for project in frappe.get_all('Project'):
-		project = frappe.get_doc('Project', project.name)
-		project.update_sales_amount()
-		project.update_billed_amount()
+	#Else simply fallback to Daily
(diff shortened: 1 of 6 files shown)


erpnext/projects/doctype/project/project.py
```diff
@@ -412,11 +412,29 @@ def create_project_update(project):
 			)
 	return data
 
+def update_project_daily():
+	if frappe.db.get_single_value("Selling Settings", "sales_update_frequency") != "Daily":
+		return
+	for project in frappe.get_all('Project'):
```
rmehta: why update all projects??? only those that have invoices should be updated

erpnext/projects/doctype/project/project.py
```diff
@@ -412,11 +412,29 @@ def create_project_update(project):
...
+		project = frappe.get_doc('Project', project.name)
+		project.update_sales_amount()
+		project.update_billed_amount()
+		project.save()
+
+def update_project_monthly():
+	if frappe.db.get_single_value("Selling Settings", "sales_update_frequency") != "Monthly":
+		return
+	for project in frappe.get_all('Project'):
```
rmehta: this is same as the earlier function, please make a common function for the project update

rmehta (changes requested):

PR #14672, opened 2018-06-25
[Fix] While saving project getting timeout error

rohitwaghchaure:
**Issue**
Getting timeout error while saving project
screen shot 2018-06-25 at 5 51 40 pm

**Fix**

- Added indexing for project, subject in Task doctype
- Added indexing for task_id in Project Task doctype
- On rename of the project update the copied_from if before rename copied_from and project name is same

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -4,7 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 
-from frappe.utils import flt, getdate, get_url
+from frappe.utils import flt, getdate, get_url, now
 from frappe import _
 
 from frappe.model.document import Document
@@ -60,6 +60,7 @@ class Project(Document):
 		self.validate_weights()
 		self.sync_tasks()
 		self.tasks = []
+		self.load_tasks()
 		self.send_welcome_email()
 
 	def validate_project_name(self):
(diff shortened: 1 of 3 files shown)


erpnext/projects/doctype/project_task/project_task.py
```diff
@@ -7,3 +7,6 @@
 
 class ProjectTask(Document):
 	pass
+
+def on_doctype_update():
+	frappe.db.add_index("Project Task", ["task_id"])
```
nabinhait: To create index based on single field, don't need to write code. It can be added based on docfield property

erpnext/projects/doctype/project/project.py
```diff
@@ -227,9 +253,10 @@ def send_welcome_email(self):
 				user.welcome_email_sent=1
 
 	def on_update(self):
-		self.load_tasks()
-		self.sync_tasks()
-		self.update_dependencies_on_duplicated_project()
+		# self.load_tasks()
+# 		self.sync_tasks()
```
nabinhait: Is this purposefully commented?

erpnext/patches/v10_0/add_index_in_project_task.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+import frappe
+
+def execute():
+	for dt in ["Project Task", "Task"]:
+		frappe.get_doc("DocType", dt).run_module_method("on_doctype_update")
```
nabinhait: No patch needed if we add index via docfield property

nabinhait (changes requested):

erpnext/projects/doctype/project/project.py
rohitwaghchaure: Yes added in validate and divide sync_tasks function, 
added update_costing_and_percentage_complete function which is added in on_update function

nabinhait (approved):

PR #14688, opened 2018-06-26
Membership

deepak-mnt:
Correct the label and field name of Membership DocType


rmehta:
@deepak-mnt labels must be title case too :)

PR #14000, opened 2018-05-12
allow cost center in entry of balance sheet account

sanjay-hercules:
https://github.com/frappe/erpnext/issues/12596
@rmehta as we discussed in last meeting, I have done changes to allow cost center in entry of balance sheet.
Please review.

PR #14743, opened 2018-06-29
Allow cost center in entry of balance sheet accounts

sanjay-hercules:
Pull-Request

- [x] Have you followed the guidelines in our Contributing document? Yes
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change? Yes
- [ ] Have you lint your code locally prior to submission?
- [x] Have you successfully run tests with your changes locally? Test failing for Agriculture module
- [x] Does your commit message have an explanation for your changes and why you'd like us to include them? Yes
- [x] Docs have been added / updated? Yes
- [ ] Tests for the changes have been added (for bug fixes / features) NA
- [ ] Did you modify the existing test cases? If yes, why? No

---

What type of a PR is this? 

- [ ] Changes to Existing Features
- [x] New Feature Submissions
- [ ] Bug Fix
- [ ] Breaking Change

--- 

- Related Issue: 
- Screenshots (if applicable, remember, a picture tells a thousand words): 

**Please don't be intimidated by the long list of options you've to fill. Try to fill out as much as you can. Remember, the more the information the easier it is for us to test and get your pull request merged** :grin:


sanjay-hercules:
@rmehta , could you please review and merge these changes?

erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -844,14 +844,14 @@ def get_outstanding(args):
 		}
 
 @frappe.whitelist()
-def get_party_account_and_balance(company, party_type, party):
+def get_party_account_and_balance(company, party_type, party, cost_center):
```
nabinhait: `cost_center` should be optional

nabinhait (commented):
- Cost Center should be added in Sales / Purchase Invoice for gl entry against debtors/creditors.
- Cost center against party in payment entry
- Add tests

erpnext/accounts/doctype/journal_entry/journal_entry.py
sanjay-hercules: Thank you @nabinhait for your review.
* cost_center should be optional - Sure, will make it optional.
* Cost Center should be added in Sales / Purchase Invoice for gl entry against debtors/creditors. - I thought we will be making this change as part of Branch implementation as we discussed in the meeting. I will make required change in Sales and Purchase Invoice to incorporate this change. Party GL entries will be based on item to consider multiple cost centers in one invoice.
* Cost center against party in payment entry - I have added it earlier but removed after changes suggested in #14000. I will make available again.
* Add tests - I will add required tests

nabinhait: @sanjay-hercules 
>Party GL entries will be based on item to consider multiple cost centers in one invoice.

Don't split GL Entries based on multiple cost centers. For Debtors, add another field for cost center in parent (sales invoice).

> Cost center against party in payment entry - I have added it earlier but removed after changes suggested in #14000. I will make available again.

In your last implementation, I think main issue was introducing 2 fields `cost_center_from` and `cost_center_to` in Payment Entry. I think we can add one field in PE. And for the scenario, where we need 2 cost centers, we can manage those entries via Journal Entry.

sanjay-hercules: >Don't split GL Entries based on multiple cost centers. For Debtors, add another field for cost center in parent (sales invoice).

So you are now okay to add cost center field in parent document which will get enabled only if **maintain cost center in entry of balance sheet account** option is checked in Account Settings. We need to do same in Purchase Invoice. **It would be better to consider Delivery Note and Purchase Receipt documents as well for cost center.**

>I think we can add one field in PE. And for the scenario, where we need 2 cost centers, we can manage those entries via Journal Entry.

Sure, I will

sanjay-hercules:
Created new PR : #14972

PR #14770, opened 2018-07-02
on delete contact update issue

shridarpatil:
fixes https://github.com/frappe/frappe/issues/5724


erpnext/custom_hooks/contacts.py
```diff
@@ -0,0 +1,11 @@
+#!/bin/env python
```
rmehta: put this in `issue.py`

rmehta (changes requested):

PR #14772, opened 2018-07-02
[Fix] Incorrect assignment of Serial No in POS Invoice

Alchez:
Pull-Request

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
- [x] Have you lint your code locally prior to submission?
- [ ] Have you successfully run tests with your changes locally?
- [x] Does your commit message have an explanation for your changes and why you'd like us to include them?
- [ ] Docs have been added / updated
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Did you modify the existing test cases? If yes, why?

---

What type of a PR is this? 

- [ ] Changes to Existing Features
- [ ] New Feature Submissions
- [x] Bug Fix
- [ ] Breaking Change

--- 

**Notes:** Although it works, this might not be the most optimal fix, considering how the Controller might affect other documents, hence comments and suggestions are appreciated.

**Problem:** Currently in the POS, if you search an item using an existing Serial No and complete the transaction, the wrong Serial No gets added to the Sales Invoice - it finds the first available Serial No and uses that instead.

**Cause:** In code, this switch happens because the Sales Invoice tries to validate the amount of Serial Nos added to the POS with the item's `stock_qty`, which is 0 by default and is set at a later step. Due to the resulting mismatch, an arbitrary Serial No for that item is set.

**Solution:** Instead of comparing the amount of Serial Nos with the `stock_qty`, the controller now uses the `qty` field for the comparison.

---

- Screenshots (if applicable, remember, a picture tells a thousand words):

1. Adding the Serial No (725272730836) in the POS.
image
1. Checking out with 1 quantity of the associated item.
image
1. After syncing the offline invoice, the item has picked up the wrong Serial No (WB20180020).
image
1. The actual Serial No (725272730836) is still in Stores.
image

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -239,8 +239,8 @@ class AccountsController(TransactionBase):
 								item.set(fieldname, value)
 
 							elif fieldname == "serial_no":
-								stock_qty = item.get("stock_qty") * -1 if item.get("stock_qty") < 0 else item.get("stock_qty")
-								if stock_qty != len(get_serial_nos(item.get('serial_no'))):
+								item_qty = abs(item.get("qty"))
+								if item_qty != len(get_serial_nos(item.get('serial_no'))):
 									item.set(fieldname, value)
 
 					if ret.get("pricing_rule"):
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -239,7 +239,8 @@ class AccountsController(TransactionBase):
 								item.set(fieldname, value)
 
 							elif fieldname == "serial_no":
-								item_qty = abs(item.get("qty"))
+								item_qty = abs(item.get("qty")) * item.get("conversion_factor")
+
 								if item_qty != len(get_serial_nos(item.get('serial_no'))):
 									item.set(fieldname, value)
 


Olawale1:
Hi @Alchez

Trust you're doing well. We used to experience a similar issue only to discover it was because the 'Automatically Set Serial Nos Based on FIFO' option was checked under 'Stock Settings'. Unchecking that option solved the issue. Just want to be sure this PR doesn't disable an actual feature which may be useful for someone else :)

Cheers

erpnext/controllers/accounts_controller.py
```diff
@@ -237,8 +237,8 @@ def set_missing_item_details(self, for_validate=False):
 								item.set(fieldname, value)
 
 							elif fieldname == "serial_no":
-								stock_qty = item.get("stock_qty") * -1 if item.get("stock_qty") < 0 else item.get("stock_qty")
-								if stock_qty != len(get_serial_nos(item.get('serial_no'))):
+								item_qty = abs(item.get("qty"))
```
nabinhait: Serial Nos are maintained based on Stock UOM, hence it will not work if sales UOM is different than stock UOM. May be you can use `item_qty = abs(item.get("qty")) * item.get("conversion_factor")`

nabinhait (changes requested):

Alchez:
@Olawale1 - thanks for the info! I tried it out and that does set the right Serial No, however, I feel that's not the right design for this. The POS should pick up the right serialized item, regardless of the FIFO stock setting.

@nabinhait - I've added the CF on the quantity check. Anything else I need to consider due to this change?

nabinhait (approved):

Issue #14778, opened 2018-07-02
Option to filter for Draft, Submitted salary slips in salary register


PawanMeh:
1. Current Salary register shows only submitted salary slips.

2. An option is needed to show Draft salary slips too as it helps in having a consolidated view for many employees before submitting all the salary slips.

![salary_register](https://user-images.githubusercontent.com/20460498/42156569-083f0704-7e09-11e8-93fe-5edde8bc6d36.gif)

PR #14780, opened 2018-07-02
Add filters to show Draft salary slips in salary register

PawanMeh:
Issue: #14778 

Testing video

salary_register

### DIFF
--- a/erpnext/hr/report/salary_register/salary_register.py
+++ b/erpnext/hr/report/salary_register/salary_register.py
@@ -74,13 +74,23 @@ def get_columns(salary_slips):
 def get_salary_slips(filters):
 	filters.update({"from_date": filters.get("from_date"), "to_date":filters.get("to_date")})
 	conditions, filters = get_conditions(filters)
-	salary_slips = frappe.db.sql("""select * from `tabSalary Slip` where docstatus = 1 %s
+	salary_slips = frappe.db.sql("""select * from `tabSalary Slip` where %s
 		order by employee""" % conditions, filters, as_dict=1)
 
 	return salary_slips or []
 
 def get_conditions(filters):
 	conditions = ""
+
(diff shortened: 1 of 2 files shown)


erpnext/hr/report/salary_register/salary_register.py
```diff
@@ -74,13 +74,23 @@ def get_columns(salary_slips):
...
-	salary_slips = frappe.db.sql("""select * from `tabSalary Slip` where docstatus = 1 %s
+	salary_slips = frappe.db.sql("""select * from `tabSalary Slip` where %s
 		order by employee""" % conditions, filters, as_dict=1)
 
 	return salary_slips or []
 
 def get_conditions(filters):
 	conditions = ""
+
+	if filters.get("docstatus"):
```
Zlash65: Maybe we can skip the if-elif-elif block ?
```
doc_status = {"Draft": 0, "Submitted": 1, "Cancelled": 2}

if filters.get("docstatus"):
	conditions += "docstatus = {0}".format(doc_status(filter.get("docstatus")))
```

PawanMeh: Tried to use the suggested code and got the below error. 

conditions += "docstatus = {0}".format(doc_status(filter.get("docstatus")))
AttributeError: 'builtin_function_or_method' object has no attribute 'get'

I think we will have to loop through the dict after getting the filter value and then use the value in format? Is that better than if-elif-elif?

Zlash65: My bad, it should be filters.get("docstatus") - missed an s there

PawanMeh: conditions += "docstatus = {0}".format(doc_status(filters.get("docstatus")))
TypeError: 'dict' object is not callable

PR #14782, opened 2018-07-02
[Fix] BOM replace timeout issue

rohitwaghchaure:
**Issue**
System throwing timeout error while replacing the BOM, when the respective BOM is linked with 100 BOM Items

**Solution**
Enqueued the BOM replace operation


erpnext/manufacturing/doctype/bom_update_tool/bom_update_tool.js
```diff
@@ -22,6 +22,24 @@ frappe.ui.form.on('BOM Update Tool', {
...
+				method: "erpnext.manufacturing.doctype.bom_update_tool.bom_update_tool.enqueue_replace_bom",
+				freeze: true,
+				args: {
+					args: {
+						"current_bom": frm.doc.current_bom,
+						"new_bom": frm.doc.new_bom
+					}
+				},
+				callback: function() {
+					frappe.msgprint(__("BOM replaced successfully"));
```
nabinhait: As we have enqueued the task, this message is wrong. Actually the message is not needed, you have added a msgprint in server side.

erpnext/manufacturing/doctype/bom_update_tool/bom_update_tool.py
```diff
@@ -17,12 +18,16 @@ def replace_bom(self):
...
 			updated_bom = bom_obj.update_cost_and_exploded_items(updated_bom)
 			bom_obj.calculate_cost()
 			bom_obj.update_parent_cost()
 			bom_obj.db_update()
+			if (getattr(bom_obj.meta, 'track_changes', False)
+				and bom_obj._doc_before_save and not bom_obj.flags.ignore_version):
+				bom_obj.save_version()
 
-		frappe.msgprint(_("BOM replaced"))
+		# frappe.msgprint(_("BOM replaced"))
```
nabinhait: remove

nabinhait (changes requested):

nabinhait (approved):

Issue #14622, opened 2018-06-21
creation of too many item variants causes Internal Error 


codingCoffee:
site goes down and `frappe` and `erpnext` "seem" uninstalled

PR #14790, opened 2018-07-02
Prevent creation of more than 500 item variants

codingCoffee:


codingCoffee:
https://github.com/frappe/erpnext/issues/14622

PR #14826, opened 2018-07-06
Allow bulk edit for the sales invoice item

rohitwaghchaure:


nabinhait:
@rohitwaghchaure allow bulk edit for all the sales and purchase transactions.

rohitwaghchaure:
For a while can we add this in the sales invoice and then later on add this option in customize form?

nabinhait (approved):

PR #14839, opened 2018-07-08
Modify renaming process of Account

shreyashah115:
- Removed Allow Rename from Account master
- Account name and number could only be renamed by clicking on `Update Account Number` button.

account number

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -217,8 +217,3 @@ def update_account_number(name, account_name, account_number=None):
 	if name != new_name:
 		frappe.rename_doc("Account", name, new_name, ignore_permissions=1)
 		return new_name
-
-def get_name_with_number(new_account, account_number):
-	if account_number and not new_account[0].isdigit():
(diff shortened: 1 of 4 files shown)


rohitwaghchaure:
Hi @shreyashah115 

- If account number is on the form then set the account number as default in the popup.
- Rename update account number to Update Account Number / Account Name

shreyashah115:
@rohitwaghchaure Done.

erpnext/accounts/doctype/account/account.py
```diff
@@ -188,7 +188,7 @@ def before_rename(self, old, new, merge=False):
 
 		return new_account
 
-	def after_rename(self, old, new, merge=False):
+	def after_not(self, old, new, merge=False):
```
nabinhait: `after_not` ?

nabinhait (changes requested):
- Still need to fix the original issue of splitting in `after_rename` method.
- Add some tests around renaming account

nabinhait (approved):

Issue #13806, opened 2018-04-26
[GST India] Address Template for India Does Not have GSTIN by default


kennethsequeira:
GSTIN was added to the template via a patch. This is not reflected in new accounts that were created.

![image](https://user-images.githubusercontent.com/33246109/39298022-b6185672-4962-11e8-939a-f22c6b440cf2.png)

PR #14855, opened 2018-07-09
[Fix] GSTIN not displaying in the address print

rohitwaghchaure:
Before
screen shot 2018-07-10 at 1 22 13 am

After
screen shot 2018-07-10 at 1 25 20 am

Fixed via https://github.com/frappe/erpnext/issues/13806


nabinhait (approved):

siddhpant:
I still don't get this working... I'm on latest version.

PR #14861, opened 2018-07-10
[Enhance][Fix] Healthcare - DocType and Fields renaming

pjamsheer:
- Rename doctypes and fields in Healthcare Domain
- Patient - customer creation fix
- Vital Signs - New Fields - Abdomen, Tongue and Reflexes
- Web Form - Patient - Personal Details


erpnext/patches/v11_0/rename_healthcare_doctype_and_fields.py
```diff
@@ -0,0 +1,60 @@
...
+	"Physician Service Unit Schedule": "Practitioner Service Unit Schedule",
+	"Consultation": "Patient Encounter",
+	"Physician": "Healthcare Practitioner"
+}
+
+def execute():
+	domain_settings = frappe.get_doc('Domain Settings')
+	active_domains = [d.domain for d in domain_settings.active_domains]
+
+	if "Healthcare" in active_domains:
```
nabinhait: The patch should be run even if Healthcare domain is not activated.

nabinhait (changes requested):

erpnext/patches/v11_0/rename_healthcare_doctype_and_fields.py
pjamsheer: Will update as soon as possible

pjamsheer: Patch Updated

nabinhait (approved):

PR #14873, opened 2018-07-11
Add patch to skip user perm check for doctype

surajshetty3416:
Skip user permission check for doctypes where
department link field was recently added

https://github.com/frappe/erpnext/pull/14121


erpnext/patches/v11_0/skip_user_permission_check_for_department.py
```diff
@@ -0,0 +1,28 @@
...
+        filters=[['allow', '=', 'Department']],
+        fields=['name', 'skip_for_doctype'])
+
+    doctypes_to_skip = []
+
+    for doctype in ['Appraisal', 'Leave Allocation', 'Expense Claim', 'Instructor', 'Salary Slip',
+                    'Attendance', 'Training Feedback', 'Training Result Employee',
+                    'Leave Application', 'Employee Advance', 'Activity Cost', 'Training Event Employee',
+                    'Timesheet', 'Sales Person', 'Payroll Employee Detail']:
+        if frappe.db.exists('Custom Field', { 'dt': doctype, 'fieldname': 'department'}): return
```
nabinhait: Is `return` works in this case? I think it should be `continue`?

nabinhait (changes requested):

nabinhait (approved):

PR #14874, opened 2018-07-11
Defaults in Item Group

shreyashah115:
- Added defaults table in Item Group.
- Removed defaults fields from Item Group.
- Modified files to fetch defaults from Item Group if not found in the defaults of Item.
- Added Patch to move already existing defaults from Item Group to child table Defaults.


nabinhait (dismissed):
Check Travis

nabinhait (changes requested):

nabinhait (approved):

PR #14888, opened 2018-07-12
[ Improv ] Loyalty Program more fixes and test case

Zlash65:
- Loyalty Program cost center and account was set on js side, so if created through api or so, those fields weren't set and it led to error.

- if no redemption was done, then making loyalty point entry caused error because of redemption=None

- return was getting added to the original's amount while recalculating loyalty points

- Test case added - for single tier, multiple tier, cancel si, return based si

- Changing Loyalty Program should reset the assigned tier

- loyalty based fields should be non copy-able


nabinhait (dismissed):
Travis failing

PR #14891, opened 2018-07-12
[enhance] Allow transfer for manufacture in the BOM

rohitwaghchaure:
If you disabled Allow transfer for manufacture then system will not consider the respective item in the stock entry with purpose as "Material Transfer for Manufacture" which we creates against the work order. But the item will be added in the Manufacture stock entry at the time of completion of the work order

Added non stock items costing into the additional charges table
screen shot 2018-07-12 at 5 52 32 pm


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -663,9 +664,29 @@ def make_stock_entry(work_order_id, purpose, qty=None):
 			additional_costs = get_additional_costs(work_order, fg_qty=stock_entry.fg_completed_qty)
 			stock_entry.set("additional_costs", additional_costs)
 
+		add_additionl_cost(stock_entry, work_order)
+
 	stock_entry.get_items()
 	return stock_entry.as_dict()
 
+def add_additionl_cost(stock_entry, work_order):
```
nabinhait: spelling?

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -663,9 +664,29 @@ def make_stock_entry(work_order_id, purpose, qty=None):
 			additional_costs = get_additional_costs(work_order, fg_qty=stock_entry.fg_completed_qty)
 			stock_entry.set("additional_costs", additional_costs)
 
+		add_additionl_cost(stock_entry, work_order)
```
nabinhait: This function should be in BOM and also called from `get_items` method in Stock Entry.

nabinhait (changes requested):
- Add tests

rohitwaghchaure:
@nabinhait Fixed the spelling and added test cases

PR #14905, opened 2018-07-15
Update batch.py

yimesgen:
Performance improvements for pos stock availability checks on a particular warehouse due to the slowing down effects of the where clauses combined by the "and".

### DIFF
--- a/erpnext/stock/doctype/batch/batch.py
+++ b/erpnext/stock/doctype/batch/batch.py
@@ -251,8 +251,8 @@ def get_batch_no(item_code, warehouse, qty=1, throw=False):
 
 def get_batches(item_code, warehouse, qty=1, throw=False):
 	batches = frappe.db.sql(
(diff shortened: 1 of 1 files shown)


nabinhait:
@yimesgen I did not understand why it will slowing down the query. We specifically added that index to make this query faster. In one of our customer's database, Stock Entry was getting very slow due to this query and after adding the index it makes it almost 100 times faster.

nabinhait (approved):

yimesgen:
hi team

the merge is welcome...but i was wondering what made you change your
mind. the reason for adding the index was clear.

Thanks

nabinhait:
Actually I forgot to add index on SLE on standard product, after adding it in customer database. So, I added it today. https://github.com/frappe/erpnext/commit/0c0c34567874dc126367a8a9196302f1c8767544

And today, I realised you were ignoring index on item and warehouse individual column, not on custom made indexes.

PR #14910, opened 2018-07-16
Make last purchase rate button visible until submitted

shreyashah115:


erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -38,6 +38,13 @@ frappe.ui.form.on("Purchase Order", {
 				frm.doc.__onload.disable_fetch_last_purchase_rate);
 		}
 	},
+
+	refresh: function(frm) {
+		if (frm.doc.__onload) {
```
rmehta: why not based on `docstatus` ?

rmehta (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.js
shreyashah115: Because need to fetch this `frappe.db.get_single_value("Buying Settings", "disable_fetch_last_purchase_rate")` from purchase_order.py, and the button's visibility will depend on that value.

PR #14918, opened 2018-07-16
[Fix] Not able to update the current value of the prefixes which are defined in the autoname field

rohitwaghchaure:
#### **Issue**
Item Price has the prefix ITEM-PRICE which is not showing in the naming series and therefore user is not able to update the current value of the respective prefix 
screen shot 2018-07-16 at 7 13 44 pm

#### **After Fix**
screen shot 2018-07-16 at 7 05 20 pm

### DIFF
--- a/erpnext/setup/doctype/naming_series/naming_series.py
+++ b/erpnext/setup/doctype/naming_series/naming_series.py
@@ -34,7 +34,14 @@ class NamingSeries(Document):
 			if options:
 				prefixes = prefixes + "\n" + options
 		prefixes.replace("\n\n", "\n")
-		prefixes = "\n".join(sorted(prefixes.split("\n")))
+		prefixes = prefixes.split("\n")
+
+		custom_prefixes = frappe.get_all('DocType', fields=["autoname"],
+			filters={"name": ('not in', doctypes), "autoname":('like', '%.#%'), 'module': ('not in', ['Core'])})
+		if custom_prefixes:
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/naming_series/naming_series.py
```diff
@@ -34,7 +34,14 @@ def get_transactions(self, arg=None):
 			if options:
 				prefixes = prefixes + "\n" + options
 		prefixes.replace("\n\n", "\n")
-		prefixes = "\n".join(sorted(prefixes.split("\n")))
+		prefixes = prefixes.split("\n")
+
+		custom_prefixes = frappe.get_all('DocType', fields=["autoname"],
+			filters={"name": ('not in', doctypes), "autoname":('like', '%.#%')})
+		if custom_prefixes:
+			prefixes = prefixes + [d.autoname.rsplit('.')[0] for d in custom_prefixes]
```
nabinhait: `d.autoname.rsplit('.')[0]` will fail if autoname is like `SO.customer_name.####` or `SO.YY.####`

nabinhait (changes requested):

erpnext/setup/doctype/naming_series/naming_series.py
rohitwaghchaure: Fixed!

nabinhait (approved):

PR #14926, opened 2018-07-17
Created Eway Bill Report

finbyz:
The Eway Bill report is created in accordance with the format of Eway Bill prescribed by the Government of India on site https://docs.ewaybillgst.gov.in/html/formatdownload.html ->JSON Preparation Tools ->[E-Waybill JSON Preparation](https://docs.ewaybillgst.gov.in/Documents/bulkewb/EWB_Preparation_Tool.xlsm)

This excel file is as below, which is capable of generating Required JSON file:-

screenshot 15

Thus, by using this report, one can export their data to Excel from ERP and can directly copy that data to the Eway Bill Preparation Tool without making any modifications.

This can be the first step for managing Ewaybill, next we are working on generating JSON directly from Delivery Note of ERP.


erpnext/stock/doctype/delivery_note/delivery_note.json
```diff
@@ -2620,6 +2650,67 @@
...
+   "fieldname": "distance", 
+   "fieldtype": "Float", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Distance",
```
nabinhait: Distance (Km)?

erpnext/stock/doctype/delivery_note/delivery_note.json
```diff
@@ -2587,6 +2587,36 @@
    "set_only_once": 0, 
    "unique": 0
   }, 
+  {
+   "allow_bulk_edit": 0, 
+   "allow_on_submit": 0, 
+   "bold": 0, 
+   "collapsible": 0, 
+   "columns": 0, 
+   "fieldname": "transporter_id",
```
nabinhait: Lets rename the field to `transporter`, we don't use `id` anywhere.

erpnext/stock/doctype/delivery_note/delivery_note.json
```diff
@@ -2650,6 +2741,67 @@
...
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Vehicle Type", 
+   "length": 0, 
+   "no_copy": 0, 
+   "options": "\nRegular\nODC",
```
nabinhait: Avoid using short form (ODC)

erpnext/stock/report/eway_bill/eway_bill.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2016, FinByz Tech Pvt. Ltd. and contributors
+// For license information, please see license.txt
+/* eslint-disable */
+
+frappe.query_reports["Eway Bill"] = {
+	"filters": [
+		{
+			'fieldname': 'id',
```
nabinhait: Let's call it `delivery_note`

erpnext/stock/report/eway_bill/eway_bill.py
```diff
@@ -0,0 +1,374 @@
...
+		set_address_details(row, special_characters)
+		
+		row.posting_date = '/'.join(str(row.posting_date).replace("-", "/").split('/')[::-1])
+		row.lr_date = '/'.join(str(row.lr_date).replace("-", "/").split('/')[::-1])
+
+		row.item_name = re.sub(special_characters, " ", row.item_name)
+		row.description = row.item_name
+
+		row.uom = unit.get(row.uom, row.uom)
+		row.customer = re.sub(special_characters[:-1] + "&0-9" + "]", "", row.customer)
```
nabinhait: Can you add some comments here what it does?

erpnext/stock/report/eway_bill/eway_bill.py
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	special_characters = "[$%^*()+\\[\]{};':\"\\|<>.?]"
+
+	for row in data:
+		set_defaults(row)
+		set_taxes(row, filters)
+		set_address_details(row, special_characters)
+		
+		row.posting_date = '/'.join(str(row.posting_date).replace("-", "/").split('/')[::-1])
```
nabinhait: add some comments

nabinhait (changes requested):
- Is the feature only applicable for India? In that case all the fields and reports should be part of regional module.
- Add some screenshots
- Add documentation in https://github.com/erpnext/foundation repo

erpnext/stock/doctype/delivery_note/delivery_note.json
finbyz: In the excel provided by Government for generating Json file, these are only two options which are acceptable.

regular odc eway bill

test2a:
Great to see the eway bill being developed. Although I haven't checked it working, but looking at the code I have a couple of things that might cause issues once in production. I just want to know if you are aware of these things and have you made any checks for that. 
1. Name of the buyer. It should include a "Urp" which simply removes requirement and checking of gstin later in the process.
2. HSN code. I ran into a bug in the department website yesterday where I was simply unable to use "2523" HSN code which is for Portland cement when using from a sub user. Tried the primary user and it got through but the issue here is there should be a check for the HSN code, if it is of the minimum 4,6,8 length.
3. For e way bills below 50 km, i think there is no need to fill part B of the bill. There should be a check for that. If the user wants that. 
4. Bill to ship to.
5. There should be an option, in sales that right after invoice at least it should ask to create an e way bill and if yes, should pick up figure from that invoice.
6. Once we get one e way bill successfully generated, bulk should be easier.

I am seriously considering adopting this software in production in my own office and for that im willing to provide my inputs.

nabinhait (approved):

nabinhait:
@finbyz Can you please add some replies on the above comments from @test2a .

nabinhait:
@finbyz Also, can you please rebase your branch from frappe/hotfix and push again?

finbyz:
Hi @test2a 

1. As of now there is no directly link with GSTIN, making GSP App Integration will be all together different project.
2.I agree there is no check on HSN code which is coming from Items table in delivery note. But this is part of original design we haven't changed it. Making It Link field Like Item would be a good Idea, but then it has to be same in all documents like Sales Order Item/Sales Invoice Item.
3. Fields are added for all, but they are not made Mandatory, as Delivery note is document used for purposes other than generating Eway bill.
4. We are using Bill To Ship To of Delivery Note
5. Currently we have not generated new document for Eway bill, We have just added few additional fields required for Eway bill in Delivery Note. As It's not 
6. We have created one report which generates data from Delivery note in the format given by Gov for generating Eway bill. This will also serve for Bulk Exports.

If you use it for Production, please share your feedback. We have two clients using it successfully from past one month.
We also got requirement for building similar thing from Sales Invoice document as well, please let me know if it will add value to you as well.

test2a:
@finbyz 
thanks for responding to my queries. I am unable to check this right now because of some issues i have on my semi production bitnami vm is not working correctly. 
for 1, my original query was that there are two types of customers in GST. One is a registered person and another is unregistered. Eway bill needs some name in the recepient portion and thats why automatic thing on the portal is to either have a gstin or "URP". we dont want to waste time checking the gstin on the website and all that, just a simple GSTIN checksum validity checker and if there isnt any, just use URP. thats that. you can easily check the formulae for GSTIN online. 

How difficult would it be to generate JSON file, i'm sure it has a well defined structure and all the data checks should be done before file generation level. 

going forward, i would suggest adding transporter information tab during invoice creation level and if data is inserted, it should ask if we want to generate JSON. if yes,it should download the json and we would only have to upload that file on the portal and thats it. if we select no, thats it for now and say if we want bills in bulk, we would then go to the bulk option, select date and bulk json would be created.

I am going to try this out myself on a dummy client so i will be giving you feedback in a couple of days.  My clients either use tally or just use the portal, but i will get at least a few of them to try this out.

PR #14947, opened 2018-07-18
[Portal] Display issues raised by all users of a customer

shreyashah115:


erpnext/support/doctype/issue/issue.py
```diff
@@ -94,10 +94,15 @@ def get_list_context(context=None):
...
+	contact = frappe.db.get_value('Contact', {'user': user}, 'name')
+	if contact:
+		contact_doc = frappe.get_doc('Contact', contact)
+		customer = contact_doc.get_link_for('Customer')
+
 	ignore_permissions = False
 	if is_website_user():
 		if not filters: filters = []
-		filters.append(("Issue", "raised_by", "=", user))
+		filters.append(("Issue", "customer", "=", customer)) if customer else filters.append(("Issue", "raised_by", "=", user))
```
nabinhait: It will give error `customer not defined`, if contact not found.

nabinhait (changes requested):

PR #14950, opened 2018-07-18
[Minor] Moved mandatory Company field from More Information section to top

KanchanChauhan:
Company is a mandatory field and it is way down below in Work Order in More Information section. 
It makes sense if it is at the top like in every doctype where Company is valid.

screen shot 2018-07-18 at 12 43 39 pm

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.json
+++ b/erpnext/manufacturing/doctype/work_order/work_order.json
@@ -14,6 +14,7 @@
  "fields": [
   {
    "allow_bulk_edit": 0, 
+   "allow_in_quick_entry": 0, 
    "allow_on_submit": 0, 
    "bold": 0, 
    "collapsible": 0, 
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/work_order/work_order.json
```diff
@@ -3,7 +3,7 @@
  "allow_guest_to_view": 0, 
  "allow_import": 1, 
  "allow_rename": 0, 
- "autoname": "naming_series:", 
+ "autoname": "Prompt",
```
nabinhait: Why Prompt?

nabinhait (changes requested):

erpnext/manufacturing/doctype/work_order/work_order.json
KanchanChauhan: Strange, I did not do that, let me fix it.

nabinhait (approved):

PR #14972, opened 2018-07-20
Allow cost center in entry of balance sheet accounts

sanjay-hercules:
This feature will facilitate user to maintain balance sheet accounts at cost center level.

Future Enhancement:
* Allow user to maintain warehouse at cost center level


sanjay-hercules:
@nabinhait , please review and let me know if any change is required.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -317,7 +318,8 @@ frappe.ui.form.on('Payment Entry', {
 				method: "erpnext.accounts.doctype.payment_entry.payment_entry.get_account_details",
 				args: {
 					"account": account,
-					"date": frm.doc.posting_date
+					"date": frm.doc.posting_date,
+					"cost_center": frm.doc.cost_center
```
nabinhait: I don't think party outstanding balance should be fetched based on selected cost center. Party has nothing to do with your cost center. So, mostly they will pay against total outstanding amount.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -859,3 +861,10 @@ frappe.ui.form.on('Payment Entry Deduction', {
 		frm.events.set_unallocated_amount(frm);
 	}
 })
+frappe.ui.form.on('Payment Entry', {
```
nabinhait: As per my above comment, if balance comes as per total outstanding, this piece of code is not needed, I think. Also  this will cause asynchronous issue, as there are multiple ajax call involved.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -520,4 +520,12 @@ frappe.ui.form.on("Purchase Invoice", {
 		}
 		frm.toggle_reqd("supplier_warehouse", frm.doc.is_subcontracted==="Yes");
 	}
-})
\ No newline at end of file
+})
+frappe.ui.form.on('Purchase Invoice', {
+	cost_center: function(frm, cdt, cdn) {
+		erpnext.utils.copy_parent_value_in_all_row(frm.doc, cdt, cdn, "items", "cost_center", "cost_center");
```
nabinhait: Why copying parent cost center to item rows? Expenses can be booked against multiple cost center and it should be based on item's default cost center.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -658,7 +665,7 @@ def make_write_off_gl_entry(self, gl_entries):
 					"credit": flt(self.base_write_off_amount),
 					"credit_in_account_currency": self.base_write_off_amount \
 						if write_off_account_currency==self.company_currency else self.write_off_amount,
-					"cost_center": self.write_off_cost_center
+					"cost_center": self.cost_center or self.write_off_cost_center
```
nabinhait: priority should be given to `self.write_off_cost_center`

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -673,7 +680,7 @@ def make_gle_for_rounding_adjustment(self, gl_entries):
 					"against": self.supplier,
 					"debit_in_account_currency": self.rounding_adjustment,
 					"debit": self.base_rounding_adjustment,
-					"cost_center": round_off_cost_center,
+					"cost_center": self.cost_center or round_off_cost_center,
```
nabinhait: priority should be given to `round_off_cost_center `

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -783,3 +783,11 @@ var select_loyalty_program = function(frm, loyalty_programs) {
 
 	dialog.show();
 }
+frappe.ui.form.on('Sales Invoice', {
+	cost_center: function(frm, cdt, cdn) {
+		erpnext.utils.copy_parent_value_in_all_row(frm.doc, cdt, cdn, "items", "cost_center", "cost_center");
+		erpnext.utils.copy_parent_value_in_all_row(frm.doc, cdt, cdn, "taxes", "cost_center", "cost_center");
```
nabinhait: same as purchase invoice comment?

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -754,13 +755,14 @@ def make_loyalty_point_redemption_gle(self, gl_entries):
...
-					"against_voucher_type": self.doctype
+					"against_voucher_type": self.doctype,
+					"cost_center": self.cost_center
 				})
 			)
 			gl_entries.append(
 				self.get_gl_dict({
 					"account": self.loyalty_redemption_account,
-					"cost_center": self.loyalty_redemption_cost_center,
+					"cost_center": self.cost_center or self.loyalty_redemption_cost_center,
```
nabinhait: priority should be given to `loyalty_redemption_cost_center`

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -852,7 +859,7 @@ def make_write_off_gl_entry(self, gl_entries):
 					"debit": self.base_write_off_amount,
 					"debit_in_account_currency": self.base_write_off_amount \
 						if write_off_account_currency==self.company_currency else self.write_off_amount,
-					"cost_center": self.write_off_cost_center or default_cost_center
+					"cost_center": self.cost_center or self.write_off_cost_center or default_cost_center
```
nabinhait: `self.write_off_cost_center or or self.cost_center or default_cost_center`

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -867,7 +874,7 @@ def make_gle_for_rounding_adjustment(self, gl_entries):
 					"against": self.customer,
 					"credit_in_account_currency": self.rounding_adjustment,
 					"credit": self.base_rounding_adjustment,
-					"cost_center": round_off_cost_center,
+					"cost_center": self.cost_center or round_off_cost_center,
```
nabinhait: `round_off_cost_center or self.cost_center`

nabinhait (changes requested):
- Rebase your PR please

FinForceConsulting:
@sanjay-hercules "this feature will facilitate user to maintain balance sheet accounts at cost center level"

What is the Use Case of showing Balance Sheet items on Cost Center Level? Cost Center or Profit Center is used to slice and dice the P&L situation and generally, BalanceSheet is looked into the whole as of a particular day.

Splitting Payable and Receivable at a Cost Center level is not quite right unless there is a separate entity of a Customer or a Supplier is dealing with a particular Cost Center.

Regards
CA Liyakat

agritheory:
> What is the Use Case of showing Balance Sheet items on Cost Center Level?

@FinForceConsulting Liyakat, this allows you to calculate an ROI or ROA per cost center, which is very useful if you have cost centers set up as physical locations within one business. There's a less strong case for it if you're measuring departments or a product line, but it's still good to have receivables and payables in there so you can get an accrual profitability out of your cost center.

FinForceConsulting:
@Tyler, do you have a sample report analysed in that way to calculate ROI. And it used BS items for it. I would love to see that.

See, I know QBO does this split of BS items onto Class (kind of cost center) but I have never used it.

_Liyakat

erpnext/accounts/doctype/payment_entry/payment_entry.js
sanjay-hercules: In ideal scenario each branch has its own bank account and maintains branch wise receivable and payable except the case where a company has centralized finance and purchase.
So if company A is having head office in Mumbai and branches in Delhi, Kolkatta and Chennai. Then in this case the branch receivable and payable will settled by branch itself not by Head Office.

Tagging Cost Center in Balance Sheet accounts is similar to maintaining branch wise financial books (e.g.. Trial Balance, P&L and Balance Sheet).

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
sanjay-hercules: If user opted to tag cost center in Balance Sheet Accounts then round_off amount should also go to Cost Center selected in parent document otherwise trial balance wouldn't match.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
sanjay-hercules: If user opted to tag cost center in Balance Sheet Accounts then write-off amount should also go to Cost Center selected in parent document otherwise trial balance wouldn't match.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
sanjay-hercules: If user opted to tag cost center in Balance Sheet Accounts then loyalty_redemption amount should also go to Cost Center selected in parent document otherwise trial balance wouldn't match.

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
sanjay-hercules: The purpose of copying parent cost center to item is to book sales against cost center selected in parent document to maintain cost center wise receivable.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
sanjay-hercules: If user opted to tag cost center in Balance Sheet Accounts then rounded-off amount should also go to Cost Center selected in parent document otherwise trial balance wouldn't match.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
sanjay-hercules: If user opted to tag cost center in Balance Sheet Accounts then write-off amount should also go to Cost Center selected in parent document otherwise trial balance wouldn't match.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
sanjay-hercules: The purpose of copying parent cost center to item is to book purchase against cost center selected in parent document to maintain cost center wise payable.

erpnext/accounts/doctype/payment_entry/payment_entry.js
sanjay-hercules: >As per my above comment, if balance comes as per total outstanding, this piece of code is not needed, I think. 

This is required to show party balance of selected cost center. If no cost center is selected, then show overall balance at company level.

>Also this will cause asynchronous issue, as there are multiple ajax call involved.

Agreed. Will enhance this.

erpnext/accounts/doctype/payment_entry/payment_entry.js
jk2081: I think we would require both pieces of information of cc-wise outstanding and total outstanding. We have cases where customers pay based on total as Navin has commented.

erpnext/accounts/doctype/payment_entry/payment_entry.js
jk2081: Again, I think if cost centre is selected, perhaps show cost centre and total for entire company.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
jk2081: Once again, my comment here of being able to see the total payable even when viewing CC wise accounts. The idea is that we would CC for product lines similar to classes in QBO so same branch may have more than one CCs. This is the scenario why visibility of both CC-level and total outstanding data would be useful.

jk2081:
A general comment is that even one branch may have multiple cost centers for ex. based on product lines. And the cost centers may share vendors and customers. In this scenario, it is important to view the total amounts and cost center level amounts of either payables or receivables since customers do pay based on total outstanding regardless of our definition of cost center.

erpnext/accounts/doctype/payment_entry/payment_entry.js
nabinhait: Ok, as cost center is optional here, we can fetch outstanding balance as per selected cost center. If user wants to process against total outstanding, then they will leave cost center blank.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
nabinhait: Let's don't copy parent CC to child items. Keep them separate. We have a plan to introduce "Department" and "Branch" inside item table, so that we can get P&L analysis based on branch / department.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
nabinhait: ok

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
nabinhait: ok

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
nabinhait: as per purchase invoice

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
nabinhait: ok

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
nabinhait: ok

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
nabinhait: ok

nabinhait (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
sanjay-hercules: > Let's don't copy parent CC to child items. Keep them separate. We have a plan to introduce "Department" and "Branch" inside item table, so that we can get P&L analysis based on branch / department.

Okay, I will remove function to copy parent CC to child.

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
sanjay-hercules: Okay, I will remove function to copy parent CC to child.

sanjay-hercules:
@nabinhait Done below changes:
1. Remove copy parent cost_center to child from Sales and Purchase Invoice
2. Improve update party and account balance functionality on change of cost_center in Payment Entry
3. Add cost_center filter to get_outstanding_documents in payment entry
Please review.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -859,3 +862,38 @@ frappe.ui.form.on('Payment Entry Deduction', {
...
+				callback: function(r, rt) {
+					if(r.message) {
+						frappe.run_serially([
+							() => {
+								frm.set_value("paid_from_account_balance", r.message.paid_from_account_balance);
+								frm.set_value("paid_to_account_balance", r.message.paid_to_account_balance);
+								frm.set_value("party_balance", r.message.party_balance);
+							},
+							() => {
+								if(frm.doc.payment_type != "Internal") {
```
nabinhait: Internal Transfer

nabinhait (changes requested):
Please rebase and push your code. We have fixed unrelated test cases which were failing on travis.

erpnext/accounts/doctype/payment_entry/payment_entry.js
sanjay-hercules: Getting outstanding documents is not required for Internal  Transfer.

sanjay-hercules: Rebase and push code.

jk2081:
Looking forward to this. @sanjay-hercules and @nabinhait thank you for your work. Not a small PR.

PR #15000, opened 2018-07-24
Trial balance and payable accounts reports and sales return fixes

elba7r:
…summary reports

adding supplier type filter to payable accounts and payable accounts
summary reports as customer group at receivable accounts and summary
reports as its important and easier to put this filter
also  there is a fix for the opening and closing values at the trial balance report as the deleted operation was causing incorrect values and not equal to the total row values at opening and closing so now its correct and equal to the total row values


erpnext/accounts/report/accounts_payable/accounts_payable.js
```diff
@@ -28,6 +28,12 @@ frappe.query_reports["Accounts Payable"] = {
 				});
 			}
 		},
+		{
+			"fieldname":"supplier_type",
+			"label": __("Supplier Type"),
+			"fieldtype": "Link",
+			"options": "Supplier Type"
```
nabinhait: "Supplier Type" has been renamed to "Supplier Group" in develop branch.

erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -378,6 +378,14 @@ def prepare_conditions(self, party_type):
 				conditions.append("""party in (select parent
 					from `tabSales Team` where sales_person=%s and parenttype = 'Customer')""")
 				values.append(self.filters.get("sales_person"))
+
+		if party_type_field=="supplier":
+   		    if self.filters.get("supplier_type"):
+		        suptype = self.filters.get("supplier_type")
+    		        conditions.append("""party in (select name from tabSupplier
+		                where supplier_type=%s)""")
```
nabinhait: `supplier_type` field also renamed to `supplier_group`

nabinhait (changes requested):
Can you add a attachment showing the problem in trial balance?

nabinhait:
@elba7r Check if https://github.com/frappe/erpnext/pull/14990 solves the problem in trial balance.

erpnext/accounts/report/accounts_receivable/accounts_receivable.py
elba7r: thats ok i have applied this requests and going to apply the pull request again

elba7r:
dus to trial balance the applied pull request which is mentioned in your comment not fixing the problem but it makes it worse and destroy the meaning of trial balance report.
that way of showing the trial balance is giving incorrect values and cause misunderstanding for the real numbers in the accounts .also it is not fix the opening values.
my pull request is fixing the problem from the root as you should not make any operations on the opening and closing ( debit - credit ) columns and you should show the real amounts on the accounts and this is what my pull request do :)

erpnext/accounts/report/accounts_payable/accounts_payable.js
elba7r: thats ok i have applied this requests

elba7r:
Here is the print screen for trial balance before my edits
trial_balance_before-edits

here is the correct trial balance after my edits

trial_balance_after-edits

FinForceConsulting:
@elba7r - "here is the correct trial balance after my edits' 
This is grossly incorrect as per the accounting conventions. Trial balance is a summary of 'Net balance' in the ledger account. And one Ledger account can have either a Dr or a Cr balance. What is see in the 'before' picture seems fine to me as it is showing only the Net balances. But in the 'After' it shows total of all the Dr entries and total of the Cr entries, which is incorrect.

Showing total of Dr & Cr line items in the Transaction column (i.e. column 3 & 4) is fine, but for the opening Balances and for the closing balances, it should ONLY show one balance, either Dr or Cr, as applicable appropriately.

"that way of showing the trial balance is giving incorrect values and cause misunderstanding for the real numbers in the accounts .also it is not fix the opening values."

The #14990 fixes the problem of Sum, that is good. But your opening balance showing total of Dr separately and Cr separately is giving no meaning. That is not the way an Accountant looks at the report.

@nabinhait, Unless we have a strong Use Case for this PR request of advantanges of showing totals of Dr & Cr separataly in the Opening & Closing balances, this PR should not be merged. Thanks.

Regards
CA Liyakat

elba7r:
@FinForceConsulting  
here is a print screen for the current trial balance without my changes
printscreen-showing the current error

i believe showing the full values CR and DR is much better than showing the report in this way as if you focus you will see that total row is not correct if you filtered and period during the year.

please remove the do not merge flag until we find a better way showing the trial balance with correct totals in your way

elba7r:
@FinForceConsulting  @nabinhait 

Now i fixed the trial balance at the correct way and the values and totals are correct and here is a print screen after correction

correct-trial-balance

elba7r (commented):
i have applied all required requests

FinForceConsulting:
@elba7r, Great! So, how **different** your contribution will make the Trial Balance from the existing Trial Balance? Sorry! I'm not a coder, so you have to explain to me the functional side of it.

@nabinhait, Let's discuss anything from the accounting side before you may decide to take this into the merge. Thanks.

_Liyakat

elba7r:
@FinForceConsulting @nabinhait 
The current trial balance opening and closing was looking for which field is bigger if dr > cr so dr = dr-cr and so on and this was causing some values does not appear in totals and cause confusion, but my change making this operation depends on the root type of the account i put screen shots for the same period for the same company before and after the edits which may explain the changes more clear : 

Here is the print screen for trial balance before my changes : 

current-trial-balance-1

current-trial-balance-2

current-trial-balance-3

current-trial-balance-4

current-totals


Here is the same period and company after fixing : 

fixed-trial-balance-1

fixed-trial-balance-2

fixed-trial-balance-3

fixed-trial-balance-4

fixed-totals

Best Regards

elba7r:
@nabinhait  @FinForceConsulting 
Today The stale put a flag inactive and the pull request will be closed automatically, so please any news about this pull request as i still have important feature for the pos requested from the community ( cashier closing )  i wants to push, and also a fix for the pos duplicates invoices, but i'm waiting until this pull request to be merged.
So please inform me !

Best Regards

nabinhait (approved):

PR #15015, opened 2018-07-26
Item Alternate Image Features in e-commerce site.

Solufyin:
Hello @nabinhait
Added Item Alternate Image Features in e-commerce site.

As per your changed request #12709 We have developed Item Alternate Image Features.
We keep both features in single doctype. Slideshow and Item alternate image functionality shown in e-commerce site.

**The User can use two alternative choices:**
1) If user is select slideshow option in slideshow dropdown then user is able to see **slideshow functionality** in e-commerce site.

slideshow1

2) If user is not select slideshow option in slideshow dropdown then a user is able to see **Item Alternate Image functionality** in e-commerce site.

We configured images in slideshow item doctype in below snapshot:
con

**Please find attached video for more information:**

[item_alternate_image.mp4.zip](https://github.com/frappe/erpnext/files/2231523/item_alternate_image.mp4.zip)

Thanks

PR #15044, opened 2018-07-30
Item Alternate Image Features in e-commerce site.

Solufyin:
Hello @nabinhait
Added **Item Alternate Image Features** in e-commerce site.
We request to you closed previous pull request #15015.
As per your changed request #15015 We have developed Item Alternate Image Features.
We keep both features in single doctype. Slideshow and Item alternate image functionality shown in e-commerce site.

**The User can use two alternative choices:**

**1) If user is select slideshow option in slideshow drop down then user is able to see **slideshow functionality** in e-commerce site.**

**2) If user is not select slideshow option in slideshow drop down then a user is able to see **Item Alternate Image** functionality in e-commerce site.**

**We configured images in slideshow item doctype in below snapshot:**

con

**Please find attached video for more information:**

[item_alternate_image.mp4.zip](https://github.com/frappe/erpnext/files/2239954/item_alternate_image.mp4.zip)

Thanks.


erpnext/public/css/website.css
```diff
@@ -268,3 +268,21 @@
 .bom-spec {
   margin-bottom: 20px;
 }
+/* For item alternate image */
+div.list_all_images img:hover {
```
netchampfaris: Put these styles in website.less

erpnext/templates/generators/item.html
```diff
@@ -3,122 +3,152 @@
...
+                        <div id="item-update-cart" style="display: none;">
+                            <a href="/cart" class='btn btn-sm btn-default'>
+                                <i class='octicon octicon-check'></i>
+                                {{ _("View in Cart") }}</a>
+                        </div>
+                    </div>
+                </div>
+            </div>
+         </div>
+        <div class="row item-website-description" style="margin-top:30px; margin-bottom:20px">
```
netchampfaris: Put these inline styles in website.less ?

netchampfaris (changes requested):
I think we should replace the Image Slideshow altogether and set this mode as default.

Solufyin:
Dear @netchampfaris ,

We Improved code as per your suggestion.

netchampfaris:
@Solufyin Hey, as I mentioned in my review, we should remove the old item slideshow and replace it with the new browsing mode you have added. It should become the default mode of browsing multiple images of the item.

gmplab:
i think Solufyin's solution is better. The user can make their own decisions, use the new way or the old way.

Solufyin:
@gmplab, "i think Solufyin's solution is better. The user can make their own decisions, use the new mode or the old mode. You can set the new mode as default."

Yes, The user can make their own choice that's why we keep both functionality slide show and item alternate image in eCommerce site.

Solufyin:
Dear @netchampfaris , 

"@Solufyin Hey, as I mentioned in my review, we should remove the old item slideshow and replace it with the new browsing mode you have added. It should become the default mode of browsing multiple images of the item."

We removed old slideshow and added new item alternate image functionality set as default mode as per your review.

netchampfaris:
item-alternative-image

netchampfaris (approved):

generare:
This look awesome! All ok with it now?

generare:
Hello again! We were checking the Alternative Images functionality and it will be absolutely nice to have this functionality but it seems to have slightly many steps to input the alternate images if understood correctly. It would be nice if we would not need to jump to a Slideshow interface but rather upload the images on the same item interface page.

We borrowed the images from @Solufyin and edited those a bit to show what we mean.
alternate-images-backend

PR #15055, opened 2018-07-31
Related to PR:15061 for MWS Connector

PawanMeh:
1. Related to PR: https://github.com/frappe/erpnext/pull/15061


nabinhait:
It is already included in #15061

PR #15056, opened 2018-07-31
[enhance] Payment order

rohitwaghchaure:


agritheory:
What's a payment order? Is there an issue for this?

rohitwaghchaure:
@agritheory 
Payment order added to do bulk payment against the purchase orders or purchase invoices. Also we have added feature to make payment request for the purchase advance amount

PR #15078, opened 2018-08-02
Fix] Task's email alert on status change not working if we change the task status from the project form

rohitwaghchaure:


codingCoffee:
@rohitwaghchaure merge-able ??

PR #15154, opened 2018-08-15
Purchase Invoice revalue Purchase Receipt Item valuation tax amount

SaiFi0102:
**Problem in one line**: Purchase Invoice does not update the stock valuation in Purchase Receipts.

**Problem**: If the tax amounts (included in valuation) in Purchase Invoice is different from it's Purchase Receipt tax amounts then although the Purchase Invoice will debit the Expense Account heads but the Expenses Included in Valuation credited in Purchase Receipts would not match the debits since Purchase Receipt does not know about the difference in values between the 2 documents.

**Solution**:
Add 2 additional fields in Purchase Receipts Item:
* invoiced_item_tax_amount: Value of item_tax_amount calculated from Purchase _Invoice_ Items
* invoiced_item_tax_portion: Ratio of the Qty of Purchase Invoice Item to the Qty of Purchase Receipt Item

Revalue Purchase Receipts on submitting or cancelling Purchase Invoice
* Call pr_docs.update_receipts_valuation on Submitting/Canceling Purchase Invoice
* update_receipts_valuation will recalculate the valuation rate considering the invoiced_item_tax_amount and if the Invoiced Qty is less than the Received Qty then it will distribute the tax amounts proportional to 
invoiced_item_tax_portion between PINV tax amount and PR tax amount.


SaiFi0102:
-Maybe I should revert the visual changes in Purchase Invoice Items?
-Maybe Purchase Invoice should also override the rate of items???

SaiFi0102:
Any thoughts on this? Would it work with Purchase Returns without any extra changes?

SaiFi0102:
Reopening. Please provide feedback.

PR #15218, opened 2018-08-24
Revert "Fix shipping_address for DropShipping"

MaxMorais:
Reverts frappe/erpnext#15202


MaxMorais:
@nabinhait it introduced a little bug when you try to create an PO directly!

I'll investigate more, and do more tests, to send the PR.
Sorry for the inconvenince

MaxMorais:
@nabinhait also, there is something weird into the revert :(

PR #15221, opened 2018-08-26
Adding "Include UOM" in Reports with Qty and Rates

SaiFi0102:
**Problem**: Stock reports only show in stock UOM and does not show other UOMs
**Solution**: Include a filter field "Include UOM" and add extra columns for converted rate and qty columns using conversion_factor

**Thoughts for the future**: Maybe use MultiSelect for Include UOM field.
**Missing out on**: Due to the complexity of getting conversion_factors from UOM Conversion Factor DocType, I did not consider those conversion factors. Should the reports also convert using UOM Conversion Factor?

**Implemented on reports**: Stock Ledger, Stock Balance, Stock Projected Qty 
**Further implementation**: Since qty and rate fields are included in many reports, I will continue updating reports to include the "Include UOM" field whenever the need arises. I would encourage the community to update the rest of the reports.

### DIFF
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -14,7 +14,8 @@ def execute(filters=None):
 
 	validate_filters(filters)
 
-	columns = get_columns()
+	include_uom = filters.get("include_uom")
+	columns, qty_columns, rate_columns = get_columns(include_uom)
 	items = get_items(filters)
 	sle = get_stock_ledger_entries(filters, items)
 
@@ -26,6 +27,13 @@ def execute(filters=None):
 	item_map = get_item_details(items, sle, filters)
 	item_reorder_detail_map = get_item_reorder_details(item_map.keys())
 
+	def update_converted_qty_rate(row, conversion_factor):
+		if include_uom and conversion_factor:
+			for c in qty_columns:
+				row[c + "_alt"] = flt(row[c] / conversion_factor)
+			for c in rate_columns:
+				row[c + "_alt"] = flt(row[c] * conversion_factor)
+
 	data = []
 	for (company, item, warehouse) in sorted(iwb_map):
 		if item_map.get(item):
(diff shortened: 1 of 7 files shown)


erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -72,17 +79,17 @@ def get_columns():
 		_("Description")+"::140",
 		_("Warehouse")+":Link/Warehouse:100",
 		_("Stock UOM")+":Link/UOM:90",
-		_("Opening Qty")+":Float:100",
+		{"label": _("Opening Qty"), "fieldname": "opening_qty", "fieldtype": "Float", "width": 100, "convertible": "qty"},
```
rmehta: +1 for using the new format, but why not refactor all the columns, so that it looks consistent

erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -103,6 +111,14 @@ def get_item_details(items, sl_entries):
 		""".format(', '.join(['"' + frappe.db.escape(i,percent=False) + '"' for i in items])), as_dict=1):
 			item_details.setdefault(item.name, item)
 
+	if include_uom:
+		for item in frappe.db.sql("""select item.name, ucd.conversion_factor from `tabItem` item
```
rmehta: is this the same as above? (commonify)

erpnext/stock/utils.py
```diff
@@ -248,3 +248,35 @@ def validate_warehouse_company(warehouse, company):
 def is_group_warehouse(warehouse):
 	if frappe.db.get_value("Warehouse", warehouse, "is_group"):
 		frappe.throw(_("Group node warehouse is not allowed to select for transactions"))
+
+def update_included_uom_in_report(columns, result, include_uom, conversion_factors):
+	if not include_uom or not conversion_factors:
+		return
+
+	convertible_cols = {}
+	for iCol in reversed(range(0, len(columns))):
```
rmehta: snake_case in Python! `col_idx`

rmehta (changes requested):

SaiFi0102:
@rmehta can you open the PR again? I've already completed the required changes.

PR #15238, opened 2018-08-28
[Bug Fix] Added missing methods used in Delivery Trip which were deleted on renaming Google Maps

KanchanChauhan:
Error:
`ImportError: Module import failed for Delivery Trip (erpnext.stock.doctype.delivery_trip.delivery_trip Error: No module named google_maps.google_maps)`


vjFaLk:
@KanchanChauhan I've removed the imports and using a helper function to get the gmaps client, that import is not needed anymore. Update your erpnext repo to get the latest code.

KanchanChauhan:
Fixed by @nabinhait already https://github.com/frappe/erpnext/commit/04bca58fb7d2293425ac8fb664a1cc089378a166

PR #15245, opened 2018-08-28
fix for fields names at sales invoice return payments and verifying the payment amount

elba7r:
the payments amount and base amount fields names fixed at sales and purchase return py controller file  added verification for the payment amount at sales invoice return to be sure its negative


nabinhait (approved):

codingCoffee:
Please check for indentation in PRs henceforth @elba7r !

PR #15249, opened 2018-08-29
Disallow renaming of root accounts

shreyashah115:
Hide rename and merge account buttons for root accounts.

screen shot 2018-08-29 at 4 41 55 pm

### DIFF
--- a/erpnext/accounts/doctype/account/account.js
+++ b/erpnext/accounts/doctype/account/account.js
@@ -47,18 +47,6 @@ frappe.ui.form.on('Account', {
 			// show / hide convert buttons
 			frm.trigger('add_toolbar_buttons');
 		}
-
-		if(!frm.doc.__islocal) {
-			frm.add_custom_button(__('Update Account Name / Number'), function () {
-				frm.trigger("update_account_number");
-			});
-		}
-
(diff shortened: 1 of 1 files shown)


agritheory:
This is not a good idea. Not being able to rename " Application of Funds (Assets)" to just "Assets" in banal. Wanting to name it things besides that could be technically incorrect, but I think this is a freedom you _don't_ want to take away.

shreyashah115:
@agritheory Well, this was just something missed out during some changes which were made few weeks back. Renaming of root accounts is not allowed in the default design. Sure, we could have a discussion if it should be further continued or not.

agritheory:
I think there's a better fix here, where title field is changed but the root account `account.name` namespace is not.

parsimonyERP:
I don't understand the implications of renaming root accounts in terms of code, queries, and other structural issues, but in this case, just "Assets" is strongly preferred to "Application of Funds (Assets)".   The same is true for Source of Funds (Liabilities).  I am +1 for the customary brevity and clarity of ERPNext that should favor just "Assets" and "Liabilities".

nabinhait:
I agree, it should be allowed to rename, closing.

PR #15253, opened 2018-08-29
Moved daily event (update_latest_price_in_all_boms) to daily_long since it was timing out

KanchanChauhan:
Issue:
screen shot 2018-08-29 at 9 52 02 pm

Solution:
Moved method from daily to daily_long


rmehta:
Good opportunity to profile an optimize this

https://github.com/frappe/erpnext/wiki/ERPNext-Performance-Tuning

PR #15266, opened 2018-08-30
Added ability to create a Return Sales Invoice/ Credit Note against a delivery note of type return

neilLasrado:
- Added a checkbox 'issue_credit_note' that is visible only if the Delivery Note has 'is_return' checked

If checked it will create a return sales invoice/credit note 

screen shot 2018-08-30 at 6 12 30 pm


neilLasrado:
Any reviews on this?

PR #15289, opened 2018-09-01
[fix] email digest date range problem

RicardoJohann:
**Email Digest tested on Monthly, Weekly and Daily summary across Fiscal Years**

To test it I have used current fiscal year 01-07-2018 to 30-06-2019 and I made one sales invoice of tests on June-18 and one sales invoice in July 2018 (same invoice amounts, but one for last fiscal year and one for current fiscal year).

The example make was Monthly Summary values were expected only from July 2018, but then I got the wrong results for “New Income” like the amount of July 2018 subtracting everything in the past fiscal year (June 2018). So, the result was zero:

image

**AFTER FIX:**

After fix it shows correctly just amounts for this financial year, in this case only the invoice I made in July 2018 is showing. For Weekly and Daily Summaries occurs the same error.

image

### DIFF
--- a/erpnext/setup/doctype/email_digest/email_digest.py
+++ b/erpnext/setup/doctype/email_digest/email_digest.py
@@ -340,14 +340,11 @@ class EmailDigest(Document):
 		balance = past_balance = 0.0
 		count = 0
 		for account in accounts:
-			balance += (get_balance_on(account, date = self.future_to_date)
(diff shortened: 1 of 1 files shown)


nabinhait:
@RicardoJohann It was not correct fix, I have sent another PR https://github.com/frappe/erpnext/pull/15320 to fix the issue. Closing the PR.

PR #15329, opened 2018-09-06
[New Feature]Scan Item Barcode label to quick create/update records in items child table

szufisher:
How It Works

- Maintain barcode field in Item doctype
barcode item master

- Scan barcode in target docs in form view on New / change mode 
material request barcode scan

target docs should have
- field items, field type as Table type
- items child table have fields: item_code and qty
 A new scan barcode field above the items table will be added, scan the item barcode label,( ensure the carriage RETURN is attached as suffix in the scanner setting), or manual key in barcode then press ENTER), the program will auto fetch the item based on barcode, when first time scanning the material, it auto creates a new record at top of the items table with item_code and qty as 1, otherwise, the existing record will be changed by adding 1 to existing qty. no impact to opening existing submitted doc with non editable status.

Currently this new feature will be automatically enabled for the following standard doctypes ,

- Material Request
- Purchase Order
- Sales Order
- Purchase Receipt
- Delivery Note
- Stock Entry
- Purchase Invoice
- Sales Invoice

discussion thread
https://discuss.erpnext.com/t/new-feature-proposal-barcode-scan-to-create-update-records-in-child-table-useful-for-stock-movement-and-orders/33695/5

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -256,10 +256,10 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			}
 			else{
 				this.frm.$wrapper.find('.field_barcode') && this.frm.$wrapper.find('.field_barcode').remove();
-		    }
+			}
 	      }
 	},
-	
+
 	scan_barcode: function(e){
 		var barcode = e.target.value;
 		if (barcode){
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -313,7 +313,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				}
 			});
 			e.target.value="";
-	    }
+		}
 		return false;
 	},
 
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -222,15 +222,95 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			this.setup_item_selector();
 		}
 	},
-
 	refresh: function() {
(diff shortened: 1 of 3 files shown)


olamide2:
I wish this could be merged ASAP

rmehta:
👍 sorry for the late review. Looks good

You need to add the "Scan Barcode" field in the JSON of all stock related documents.

erpnext/public/js/controllers/transaction.js
```diff
@@ -229,6 +229,92 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 		this.set_dynamic_labels();
 		this.setup_sms();
 		this.setup_quality_inspection();
+		this.setup_barcode_scan();
+		this.frm.$wrapper.find('div.frappe-control.field_barcode p.help-box').text('');
```
rmehta: where is field_bardcode ?

Also bad idea to pick fields via jquery

rmehta (changes requested):

szufisher:
> 👍 sorry for the late review. Looks good
> 
> You need to add the "Scan Barcode" field in the JSON of all stock related documents.

Considering the barcode field only exists for capturing the scanned barcode text, not meant to be saved to backend database, in other word it is a kind of dummy front end field only, so instead of adding it to the json file, the barcode field is dynamically created by js at front end only.

erpnext/public/js/controllers/transaction.js
szufisher: > where is field_bardcode ?
> 
> Also bad idea to pick fields via jquery

the barcode field is a dynamically created field, not accessible via field_dict.

Zlash65:
Have sent another PR with some clean up and changed extended from this

PR #15365, opened 2018-09-08
Fixed PDC data in AR/AP report

SaiFi0102:
I found a bug in AR/AP report's PDC columns that if there is a payment with posting_date=reference_date then the amount of that payment would be included in BOTH pdc amount and paid amount on that posting/reference_date. This led to an incorrect value in the "Remaining Balance" column.

I wasn't complete sure of the intention of the PDC fields so I made a guess and I made the following changes:
-Get PDC date from posting_date instead of reference date since posting_date is the date at which the post dated cheque can be cleared or paid. Incase the payment isn't a PDC date (reference_date=posting_date), it still serves to show as future payments. I have no idea what the LC in PDC/LC means. 
-Get PDC data from submitted documents only

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
@@ -453,7 +453,7 @@ def get_pdc_details(party_type, report_date):
 		on
 			(pref.parent = pent.name)
 		where
-			pent.docstatus=1 and pent.posting_date > %s
+			pent.docstatus < 2 and pent.posting_date > %s
 			and pent.party_type = %s
 			group by pent.party, pref.reference_name""", (report_date, party_type), as_dict=1):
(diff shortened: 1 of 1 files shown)


SaiFi0102:
Keepin it alive

SaiFi0102:
...

rmehta:
@SaiFi0102 sorry for the late review. Can you add a screenshot of the fixed report. Also add some explanation why `posting_date` should be considered instead of `reference_date`

SaiFi0102:
### Payment Entry:
Reference Date (cheque received on): 4-10-2018
Posting Date (date on cheque): 5-10-2018
image

### Accounts Payable (some columns hidden):
On 3-10-2018:
image

On 4-10-2018
image

On 5-10-2018
image

SaiFi0102:
Bump

SaiFi0102:
Please include this in staging-fixes too

PR #15374, opened 2018-09-10
[Fix] Stock difference between gl entry and stock ledger entry booked in stock adjustment

rohitwaghchaure:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -396,8 +403,28 @@ def make_item_gl_entries(self, gl_entries):
 					val_rate_db_precision = 6 if cint(item.precision("valuation_rate")) <= 6 else 9
 
 					# warehouse account
+					net_amt_precision = item.precision("base_net_amount")
 					warehouse_debit_amount = flt(flt(item.valuation_rate, val_rate_db_precision)
-						* flt(item.qty)	* flt(item.conversion_factor), item.precision("base_net_amount"))
+						* flt(item.qty)	* flt(item.conversion_factor), net_amt_precision)
+
+					# Stock ledger value is not matching with the warehouse amount
```
rmehta: move this to a separate function. Average function size should be 10 lines

rmehta (changes requested):

schilgod:
@rohitwaghchaure @rmehta Possible to provide some explain why this change was necessary? We are having issues with PI having COGS entries, we cant figure out why COGS was credited with the difference amount. thanks.

lxnow:
I'd like to understand too, particularly in the Purchase receipt level generating COGS entries

A7medAbdin:
The same here  too, is there is any update

PR #15380, opened 2018-09-11
ERPNext Quickbooks Migrator

adityahase:
Still isn't ready


netchampfaris (changes requested):

vjFaLk:
I understand that this is a WIP, however, after doing a quick review of the PR so far, I feel like it's not designed very well. 

Please ignore this entirely if I've spoken too soon and you plan to implement this 

Here's some pointers -:

* Most of your code should belong in the class itself, as it's specific to the QuickBooks Migrator DocType
* The first thing you wanna do is use an existing library instead of using requests, I believe there is a python library to access the QB API 
* Assuming that you did already check that, and felt that you'd be better off rolling your own implementation, you should've started with a separate configuration DocType and created a wrapper around the request API 
* Even the data that comes from the API could be handled in a different place
* You don't want to constantly be accessing corners of a dictionary across all your code, makes it harder to read. 
* Your code does one thing, and one thing only, try to make the foundations more generic, so anyone else can use this QB Wrapper to do anything else they would like in the context of ERPNext 
* There's too much code in one single file, you want to separate it out

That's some of the things off my head, I hope I don't sound like an ass, I just want to help :)

PR #15382, opened 2018-09-11
[Staging] Stock Reconciliation

Zlash65:
- updated query [more optimized]
- filter added from dropdown

### DIFF
--- a/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
+++ b/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
@@ -271,7 +271,8 @@ class StockReconciliation(StockController):
 
 @frappe.whitelist()
 def get_items(warehouse, posting_date, posting_time, company):
-	items = [d.item_code for d in frappe.get_list("Bin", fields=["item_code"], filters={"warehouse": warehouse})]
+	items = frappe.db.sql('''select i.name, i.item_name from `tabItem` i, `tabBin` bin where i.name=bin.item_code
+		and i.disabled=0 and bin.warehouse=%s''', (warehouse), as_dict=True)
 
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
```diff
@@ -273,26 +273,24 @@ def cancel(self):
...
+		and id.default_warehouse=%s and id.company=%s group by i.name''', (warehouse, company), as_dict=True)
 
 	res = []
-	for item in set(items):
-		stock_bal = get_stock_balance(item[0], warehouse, posting_date, posting_time,
+	for item in items:
+		stock_bal = get_stock_balance(item.name, warehouse, posting_date, posting_time,
 			with_valuation_rate=True)
 
-		if frappe.db.get_value("Item",item[0],"disabled") == 0:
```
vjFaLk: I don't think this was supposed to be removed? Also this can be written in a better way : 

```
if not frappe.db.get_value("Item", item.name, "disabled"):
	...
```

erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
```diff
@@ -273,26 +273,24 @@ def cancel(self):
...
-				"qty": stock_bal[0],
-				"item_name": frappe.db.get_value('Item', item[0], 'item_name'),
-				"valuation_rate": stock_bal[1],
-				"current_qty": stock_bal[0],
-				"current_valuation_rate": stock_bal[1]
-			})
+		res.append({
+			"item_code": item.name,
+			"warehouse": warehouse,
+			"qty": stock_bal[0],
```
vjFaLk: It would also be cleaner if the index values were put into a more aptly named variable

vjFaLk (changes requested):

PR #15405, opened 2018-09-14
Bank statement import corrections

chdecultot:
Hello,

I have been testing the new bank statement import tool and I have to say that it looks very promising. Thank you!

While testing I have faced some blocking issues for which I propose the following corrections + some UX corrections. It took me a while to understand how it works, so I guess some additional love should be provided to help end users.

1. The header mapping was not working and the original headers where hard coded, whereas there is a mapping table.
I corrected it to allow using different headers (as intented in the initial design)
image

2. I added the safe_decode function to all description references to avoid ASCII errors (very common with french).

3. I made the party_type field editable for manual entries in the invoice table and made a mapping between the party type and the invoice type to avoid errors with the query filter:
image

Thanks in advance for reviewing this PR

**PS: Personal thoughts on things to improve**

- Display error messages when the headers are not found. It is always confusing when the system just does nothing when there is an error

- Add error/display messages for the buttons when there is no transaction to generate. Maybe provide additional information below to explain what they are meant to do.

- Improve the reconciliation of already existing payment entries: having to select two elements + the payment entry number (without context) from a dropdown is horrible... I understand it's not the initial intent of this functionality, but it would be a wonderful addition to have a simple drag'n'drop or selectable table to easily match bank entries with payment entries manually.

### DIFF
--- a/erpnext/accounts/doctype/bank_statement_transaction_entry/bank_statement_transaction_entry.js
+++ b/erpnext/accounts/doctype/bank_statement_transaction_entry/bank_statement_transaction_entry.js
@@ -85,7 +85,7 @@ frappe.ui.form.on('Bank Statement Transaction Invoice Item', {
 		} else if (row.party_type == "Account") {
 			row.invoice_type = "Journal Entry";
 		}
-		refresh_field("invoice_type", row.name, "payment_invoice_items")
+		refresh_field("invoice_type", row.name, "payment_invoice_items");
 
 	},
 	invoice_type: function(frm, cdt, cdn) {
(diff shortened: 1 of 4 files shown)


nabinhait:
@chdecultot Thanks for testing and fixes. I agree with you that there are a lot of scope for improvements on this feature. Please feel free to improve the functionality based on your thoughts.

chdecultot:
Hi @nabinhait,

Thanks a lot for the merge. I am definitely thinking about how to improve this functionality and will get back to you over the next weeks with a proposal (I have pending work on the CRM module to finish first).

Have a great evening!

PR #15410, opened 2018-09-16
update staff plan when add or delete job-opening

Alsum:
update staff plan when add or delete job-opening
and edit validation of current_vacancies to make it
based on planned_vacancies in staff plan

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ erpnext/docs/current
 *.swo
 __pycache__
 *~
+.idea/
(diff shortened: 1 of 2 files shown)


nabinhait:
@Alsum It will not work if Staffing Plan is defined against parent company and job opening is created against child company. Also if we update balance in staffing plan you need to write a patch to update opening balance for the accounts who are already using the feature. I think it is better to keep the code  in existing way, unless there are any bug.

Alsum:
@nabinhait okay thank you, but in current code if there is valid number of planned_vacancies in staff plan
and number of employees that has same designation bigger than planned_vacancies 
i can not create new job-opening for that planned_vacancies

neelbhanushali:
@nabinhait any updates regarding this?
@Alsum did you figure out a work around for this?

PR #15426, opened 2018-09-18
fix(translations): Replace translations by keyword with indexed version

netchampfaris:
Keywords in the translation also gets translated which results in an error because python cannot find the key

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -753,8 +753,11 @@ def validate_item_type(doc, fieldname, message):
 		""".format(item_list, fieldname), as_list=True)]
 
 	if invalid_items:
-		frappe.throw(_("Following item {items} {verb} marked as {message} item.\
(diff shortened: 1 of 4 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -753,8 +753,11 @@ def validate_item_type(doc, fieldname, message):
...
-			You can enable them as {message} item from its Item master".format(
-				items = ", ".join([d for d in invalid_items]),
-				verb = _("are not") if len(invalid_items) > 1 else _("is not"),
-				message = message)))
+		items = ", ".join([d for d in invalid_items])
+
+		if len(invalid_items) > 1:
+			error_message = _("Following items {0} are not marked as {2} item. You can enable them as {2} item from its Item master".format(items, message))
+		else:
+			error_message = _("Following item {0} is not marked as {2} item. You can enable them as {2} item from its Item master".format(items, message))
```
Alchez: This will error out. Change {2} to {1}.

PR #15437, opened 2018-09-19
[Enhance] Deferred Expense

Zlash65:
## Item Master

screen shot 2018-09-05 at 10 53 07 am

## Company Master

screen shot 2018-09-05 at 10 54 01 am

## Purchase Invoice

screen shot 2018-09-05 at 10 55 36 am

- Validation added for date intervals
- Scheduler event added to book expense every month

### Commonified Code in deferred_revenue.py for both Sales Invoice and Purchase Invoice
- also dates are adjusted based on last gl entry so that no days are skipped if manually triggered


PawanMeh (commented):
Deferred Expense Account should be an Asset Account instead of Liability, eg: Prepaid Expense

PR #15445, opened 2018-09-19
[CRM Enhancement] Addition of a "Call" Doctype

chdecultot:
Hi,

This PR is related to PR https://github.com/frappe/frappe/pull/6113 in Frappe.

It adds a new Doctype called "Call" in the CRM of ERPNext and make available to be added from any other doctype as an interaction.

image

image

Any feedback is welcome!


chdecultot:
After some afterthoughts and discussions with users, it feels better to leverage the "Event" doctype in order to centralize all types of events (and especially to have a common calendar view).
I am therefore closing this PR and will update the PR in Frappe with enhancements to the "Event" doctype.

PR #15466, opened 2018-09-21
[Refactored] Asset depreciation for WDV method and pro rata calculation

rohitwaghchaure:


rmehta:
@rohitwaghchaure we seriously need tests for this.

PR #15490, opened 2018-09-25
Contact and email fields in Payment Entry

shreyashah115:
screen shot 2018-09-25 at 2 01 13 pm


nabinhait (dismissed):
Map contact details when payment entry is created from invoice using "Make Payment" button

PR #15491, opened 2018-09-25
[Feature] Assign a default return warehouse for Goods received via Sales Return

neilLasrado:
Added ability to have a default return warehouse for Sales Returns

- Added a link field 'Default Return Warehouse' in 'Stock Settings' 

screen shot 2018-09-25 at 4 01 39 pm

- While making a sales return against delivery note the system checks if a default returns warehouse exists and replaced the item warehouse else uses the existing warehouse in the source delivery note.

This feature is desirable because in many cases goods once sold and received as returns cannot be resold as new by law. This helps track such goods.


rmehta:
Merging, but generally I don't like the pattern of having too many settings. They are not discoverable.

Issue #15508, opened 2018-09-26
Enhancement - No Lead Information field present


joolsr:
Issue seen on using current Master or V11 Staging and verified on beta.erpnext.com

Using the Leads functionality, there is not a field to record rough notes on a potential requirement or interest. When we exhibited at a show recently, we entered all contact with show visitors as Leads but we could only add a quick note of their interest in the Communication rather than on the Lead. 
This meant that it was difficult to get a report with Lead, Company, and potential interest as the recording of the potential interest was made in the Communication field linked to the Lead and not on the Lead doctype itself.

This is a very simple enhancement where one new Text field is required, and the Leads Report can be tweaked to include this too. I am not a developer so cannot create a PR myself.
 @chdecultot Would you be able to help here after your sterling work to improve the CRM capabilities for events, and meetings ?

chdecultot:
Hi @joolsr,

I do agree that a "Note" or "Information" text field is missing in the lead. I have had the same experience as you.
It is true we could use the "comment" section but it is, in my opinion, great for quick contextual comments, but not great for general comments regarding the Lead itself.

What about adding a new text editor field in a new collapsible section below the follow-up section ?
Something like that:

![image](https://user-images.githubusercontent.com/4903591/46090628-6bfe4c00-c1b1-11e8-824a-878cb952d072.png)


Let's see how it could be included in the lead details report if the text is quite long, but it should be possible.

Let me know your thoughts and I'll be happy to make a PR for this.

joolsr:
Thanks Charles for your ideas. This was more or less my thinking too re the text editor field. When we recorded notes on Leads they were generally quite short, say a few lines long at most. If you could use a smaller text editor it may work. But your plan with the collapsible section is fine. 
The report needs to show this info where it may be wrapped in a column so that it should fit nicely on a A4 page ..

chdecultot:
Hi @joolsr,

I just submitted a PR.
Because of the multiple use cases, I prefer to propose the usage of a text editor instead of a small text.
It shouldn't be an issue for you I guess.

Please take a look at the report. I used a big text, but you can imagine how it would be with a few lines.

Let me know if that's ok for you.

Have a nice evening!

joolsr:
Yes this looks great as it is. many thanks again

joolsr:
I think we can close this now

PR #15510, opened 2018-09-26
Lead notes

chdecultot:
Hi,

This PR follows the discussion in issue https://github.com/frappe/erpnext/issues/15508.

A field for adding notes regarding a lead is obviously missing in the lead doctype.
We can use comments, but comments are more time-related, whereas most of the time one need to register general notes regarding a lead that should be kept in one place.

Our proposal is therefore to add a new collapsible section in the Lead called "Notes":
image

This section will have a text editor, thus allowing any sales person to make a formatted note regarding a lead.

I have also updated the sales details report to add this field:
image

(Maybe it would be good to have the possibility to make the text wrappable in a future update of the datatable)

When exporting in PDF, the format is kept as defined (this test example is extreme):
image

Thanks in advance for your review and feedback.

**Technical note**
I have moved the report from the "Selling" module to the "CRM" module for consistency, because the Lead doctype is in the CRM module.


rmehta:
@chdecultot send v11 pulls to `staging-fixes`

PR #15523, opened 2018-09-27
[Fix] Attendance tool

rohitwaghchaure:
**Before Fix**
screen shot 2018-09-27 at 7 45 39 pm

**After Fix**
screen shot 2018-09-27 at 7 45 18 pm

If User has created attendance in draft and later on he is updating the attendance from the data import tool then submit the draft mode attendance, previously system was creating the new attendance. So one remains in submit and one in draft

Dependency
https://github.com/frappe/frappe/pull/6156

### DIFF
--- a/erpnext/hr/doctype/upload_attendance/upload_attendance.py
+++ b/erpnext/hr/doctype/upload_attendance/upload_attendance.py
@@ -5,7 +5,7 @@
 
 from __future__ import unicode_literals
 import frappe
-from frappe.utils import cstr, add_days, date_diff
+from frappe.utils import cstr, add_days, date_diff, getdate
 from frappe import _
 from frappe.utils.csvutils import UnicodeWriter
 from frappe.model.document import Document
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/upload_attendance/upload_attendance.py
```diff
@@ -114,6 +115,11 @@ def upload():
 		if not row: continue
 		row_idx = i + 5
 		d = frappe._dict(zip(columns, row))
+
+		if (d.get("name") and
+			frappe.db.get_value('Attendance', d.get('name'), 'docstatus') == 1):
```
rmehta: this should throw an error if the attendance value is different

rmehta (changes requested):

erpnext/hr/doctype/upload_attendance/upload_attendance.py
rohitwaghchaure: Fixed!
Have removed this code and added except AttributeError: pass to avoid the update_after_submit_error

Before 
screen shot 2018-10-01 at 12 30 09 pm


After
screen shot 2018-10-01 at 12 30 42 pm

PR #15524, opened 2018-09-27
[Enhancement] Improvement to the sales pipeline

chdecultot:
Hi,

The current CRM in ERPNext is lacking the possibility to see a proper sales pipeline taking into account the following data:

- A prospective sales amount
- A percentage of realization of this opportunity
- Different sales stages

In order to solve this issue, my proposal is the following:

1. Add the above fields in the opportunity DocType:
image

2. Add two new charts reports in the sales funnel page (basic reports in most CRM systems):

sales_pipeline

The first chart provides a quick view of all opportunities by stage and compounded amount (amount set in the opportunity x probability of realization)

The second gives an overview of all opportunities by source.

I have also added a filter by company for the sales funnel, as it was missing.

Please let me know if you have comment/feedback. I'll be happy to improve this PR.

Best regards,

Charles-Henri


erpnext/selling/page/sales_funnel/sales_funnel.py
```diff
@@ -1,35 +1,80 @@
...
+	df = pd.DataFrame(cp_opportunities).groupby(['source', 'sales_stage'], as_index=False).agg({'compound_amount': 'sum'})
+
+	result = {}
+	result['labels'] = list(set(df.source.values))
+	result['datasets'] = []
+	
+	for s in set(df.sales_stage.values):
+		result['datasets'].append({'name': s, 'values': [0]*len(result['labels']), 'chartType': 'bar'})
+
+	for i, row in df.iterrows():
```
barredterra: Replace `i` with `_` to show that it will not be used.

erpnext/selling/page/sales_funnel/sales_funnel.js
```diff
@@ -186,4 +256,44 @@ erpnext.SalesFunnel = Class.extend({
...
+		});
+	}
+
+	render_pipeline() {
+		let me = this;
+		let currency = frappe.defaults.get_default("currency");
+
+		let chart_data = me.options.data ? me.options.data : null;
+
+		const parent = me.elements.funnel_wrapper[0]
```
barredterra: Add semicolon.

erpnext/selling/page/sales_funnel/sales_funnel.py
chdecultot: Thanks for this review @barredterra, I just committed a correction.

erpnext/selling/page/sales_funnel/sales_funnel.js
chdecultot: Thanks! added in a new commit.

erpnext/selling/page/sales_funnel/sales_funnel.py
Alchez: Actually avoid using `_` as a placeholder, since Frappe already uses that for translations.

chdecultot: Humm nice catch! I replaced it with "dummy". Pylint should be ignoring "dummy" according to the regex available in the documentation: Default: _+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_

Alchez: Ah, that's interesting.

I was looking into iteration over a `Dataframe`, and I found this: https://stackoverflow.com/questions/16476924/how-to-iterate-over-rows-in-a-dataframe-in-pandas

I don't know enough about `pandas` or `Dataframe`, but based on that thread, `Dataframe.itertuples` seems to provide better performance. Might be significant when someone's dealing with large datasets.

Just my 2 cents.

chdecultot: Thanks for pointing that out @Alchez !
Actually I decided to use Pandas because of it's ability to handle large datasets. It will be very useful for other reports as well.

I have moved to itertuples as suggested by the stackoverflow discussion.
I guess the dataset needs to be quite huge to see any significant difference between the two methods, but it's always better to implement the fastest method from the beginning.

Thanks again!

Alchez: No problem, glad to help! 😄

barredterra: > Actually avoid using `_` as a placeholder, since Frappe already uses that for translations.

Of course, sorry, i didn't think of that.

erpnext/crm/doctype/opportunity/opportunity.json
```diff
@@ -1,1324 +1,1520 @@
...
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Sales Stage", 
+   "length": 0, 
+   "no_copy": 0, 
+   "options": "Prospecting\nQualification\nNeeds Analysis\nValue Proposition\nIdentifying Decision Makers\nPerception Analysis\nProposal/Price Quote\nNegotiation/Review",
```
rmehta: Convert this into a master too!

rmehta (changes requested):
Convert Sales Stage into a doctype (don't forget to add default fixtures both for install and also via a patch)

chdecultot:
Hi @rmehta,

Thanks for your review.
I just created a new doctype called "Sales Stage" and the corresponding fixtures/patch.

Let me know if you see any additional improvement.

Have a nice day!

chdecultot:
Oh my apologies... 😯

adityahase:
@chdecultot Why so much duplication in `erpnext/erpnext/selling/page/sales_funnel/sales_funnel.js` ?

chdecultot:
Hi @adityahase,

No particular reason except that I should have made cleaner code...

PR #15525, opened 2018-09-27
Lead to customer UX enhancement

chdecultot:
Hi,

This is a small enhancement in the customer DocType.

Today we have the ability to set a market segment and an industry in the Lead, but these two values are not present in the Customer, where they could be very helpful for reporting purposes.

I propose to add these two fields in the Customer doctype so these values can be automatically populated when the customer is created from a lead.

I propose to add them in the section "More information":
image

Thanks in advance for your review and feedback!


rmehta:
Create a new master (DocType) for "Market Segment" - Users will want to customize this for their business, so a configurable master is the best option.

rmehta (changes requested):
Create new DocType "Market Segment"

chdecultot:
Hi @rmehta,

I created a new DocType called Market Segment as requested.
Thanks a lot for your review.

Have a nice day!

erpnext/patches/v11_0/add_market_segments.py
```diff
@@ -0,0 +1,17 @@
...
+import frappe
+from frappe import _
+from erpnext.setup.setup_wizard.operations.install_fixtures import make_fixture_records
+
+def execute():
+	frappe.reload_doc('crm', 'doctype', 'market_segment')
+
+	frappe.local.lang = frappe.db.get_default("lang") or 'en'
+
+	records = [
```
rmehta: make a separate function for this and just call it?

chdecultot: Of course, I've juste made the split.

PR #15532, opened 2018-09-29
[Minor Fix] Force toggle display Net Total field when default taxes are fetched

SaiFi0102:
**To reproduce the problem:**
* Create Taxes and Charges Template with some Inclusive Taxes
* Set it as the default Template so that it will be fetched on transactions automatically
* Create a new transaction (ex Purchase Order)
* You will see that Net Total, Net Rate and Net Amount fields are not shown since there is a condition in set_dynamic_fields() that is preventing the AJAX call to toggle its display.

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -123,6 +123,7 @@ class PurchaseInvoice(BuyingController):
 	def set_missing_values(self, for_validate=False):
 		if not self.credit_to:
 			self.credit_to = get_party_account("Supplier", self.supplier, self.company)
+			self.party_account_currency = frappe.db.get_value("Account", self.credit_to, "account_currency", cache=True)
 		if not self.due_date:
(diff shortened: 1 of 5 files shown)


rmehta:
👍  @SaiFi0102 Thanks for fixing this issue.

Please include an animated GIF / screenshots for your PRs. This will greatly help us in reviewing.

Please do through this once:

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist

SaiFi0102:
Net Total does not show up when an inclusive tax is added in taxes table:
nettotal
After fix:
nettotalfix

My custom feature to create Purchase Invoice from Sales Order shows that item.discount_amount is not calculated in taxes_and_totals.py 
discount
It only needed one line to fix this so I'm skipping the GIF of the fix here

Since party_account_currency isn't set from set_missing_values, clicking Make -> Sales/Purchase Invoice shows undefined in place of currency for Outstanding Amount
party_account_currency
After fix:
party_account_currency_fix

I also removed the condition in set_dynamic_labels because then it was creating more problems trying to force reset labels for only some conditions. Like it would show (Company Currency) fields, even though the transaction would be in the company currency:
companycurrency

SaiFi0102:
This PR is ready for review

PR #15534, opened 2018-09-29
Fixed a problem set_item_default which led to filling up Item's Item Default table

SaiFi0102:
set_item_default was referencing a variable company of "company" which caused:

image


SaiFi0102:
I think I set the branch incorrectly. It should be included in v11 too: staging

PR #15541, opened 2018-09-30
Adding "Include UOM" in Reports with Qty and Rates

SaiFi0102:
Reopened https://github.com/frappe/erpnext/pull/15221

**Problem**: Stock reports only show in stock UOM and does not show other UOMs
**Solution**: Include a filter field "Include UOM" and add extra columns for converted rate and qty columns using conversion_factor

**Thoughts for the future**: Maybe use MultiSelect for Include UOM field.
**Missing out on**: Due to the complexity of getting conversion_factors from UOM Conversion Factor DocType, I did not consider those conversion factors. Should the reports also convert using UOM Conversion Factor?

**Implemented on reports**: Stock Ledger, Stock Balance, Stock Projected Qty 
**Further implementation**: Since qty and rate fields are included in many reports, I will continue updating reports to include the "Include UOM" field whenever the need arises. I would encourage the community to update the rest of the reports.

### DIFF
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -14,7 +14,8 @@ def execute(filters=None):
 
 	validate_filters(filters)
 
-	columns = get_columns()
+	include_uom = filters.get("include_uom")
+	columns, qty_columns, rate_columns = get_columns(include_uom)
 	items = get_items(filters)
 	sle = get_stock_ledger_entries(filters, items)
 
@@ -26,6 +27,13 @@ def execute(filters=None):
 	item_map = get_item_details(items, sle, filters)
 	item_reorder_detail_map = get_item_reorder_details(item_map.keys())
 
+	def update_converted_qty_rate(row, conversion_factor):
+		if include_uom and conversion_factor:
+			for c in qty_columns:
+				row[c + "_alt"] = flt(row[c] / conversion_factor)
+			for c in rate_columns:
+				row[c + "_alt"] = flt(row[c] * conversion_factor)
+
 	data = []
 	for (company, item, warehouse) in sorted(iwb_map):
 		if item_map.get(item):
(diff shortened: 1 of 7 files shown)


rmehta:
@SaiFi0102 please include screenshots of the reports as proof that it works

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -54,36 +57,40 @@ def execute(filters=None):
...
-		_("In Value")+":Float:80",
-		_("Out Qty")+":Float:80",
-		_("Out Value")+":Float:80",
-		_("Balance Qty")+":Float:100",
-		_("Balance Value")+":Float:100",
-		_("Valuation Rate")+":Float:90",
-		_("Reorder Level")+":Float:80",
-		_("Reorder Qty")+":Float:80",
-		_("Company")+":Link/Company:100"
+		{"label": _("Item"), "fieldname": "item_code", "fieldtype": "Link", "options": "Item", "width": 100},
```
rmehta: 👍 for refactoring columns to new style

SaiFi0102:
image

SaiFi0102:
I plan to add the feature to merge (at client side) UOM columns with qty and rate columns when there is only ONE uom in the whole report output

SaiFi0102:
This PR is ready for review

shreyashah115 (approved):

erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -88,19 +99,26 @@ def get_items(filters):
...
+	if include_uom:
+		cf_field = ", ucd.conversion_factor"
+		cf_join = "left join `tabUOM Conversion Detail` ucd on ucd.parent=item.name and ucd.uom=%(include_uom)s"
+
 	for item in frappe.db.sql("""
-		select name, item_name, description, item_group, brand, stock_uom
-		from `tabItem`
-		where name in ({0})
-		""".format(', '.join(['"' + frappe.db.escape(i,percent=False) + '"' for i in items])), as_dict=1):
+		select item.name, item.item_name, item.description, item.item_group, item.brand, item.stock_uom{cf_field}
```
codingCoffee: Hey, great PR! Just write the SQL queries this way henceforth: https://github.com/frappe/erpnext/blob/develop/erpnext/stock/report/stock_ledger/stock_ledger.py#L116

This would help standardize the SQL queries and also since we want the SQL queries to be DB agnostic.

SaiFi0102: Alright thanks @codingCoffee, I'll keep it in mind... the PostgreSQL commits were too sudden for me to adapt earlier

PR #15542, opened 2018-09-30
fix: Payment System Enhancements - Generalizing invoices and payments

SaiFi0102:
Reopened #15362 since it got mixed up with other commits and since I'm changing the base branch.

# Problem

Journal Entry's advance payment balance is calculated from `Journal Entry Account` which can result in an invalid unallocated amount. When a Payment Entry or reverse Journal Entry is made against a Journal Entry that debits Creditors or credits Debtors (makes a payment), the unallocated amount is not updated.

To fix this problem of being able to reverse payable/receivable entry I had to modify the payment system such that Journal Entry's balance is calculated from the GL Entries made by and against the Journal Entry. To do that I had to change and generalize payable/receivable entries from voucher documents.

# Modifications

#### Voucher Documents (the ones that credit/debit payables/receivables):
* Previously all voucher documents other than Journal Entry were setting `against_voucher` to reference themselves. In my opinion Journal Entry follows a more logical scheme as it sets `against_voucher` as empty rather than referencing itself when not referencing another voucher.
* Made `GL Entry:against_voucher`'s values consistent with all voucher documents such that `against_voucher` is set as _empty_ when making a party payable/receivable entry and is set to reference another voucher _only_ when adjusting a previously made payable/receivable GL Entry (for payments and returns etc). This simplifies and generalizes queries for calculating invoices and unallocated payments.
* Added a generalized utility function `get_balance_on_voucher` that calculates a balance for a `(voucher_type, voucher_no, account, party_type, party)` tuple.

#### GL Entry and General Ledger:
* Indexed `GL Entry:against_voucher_type` since invoice/payment queries filter through this field often.
* Calling `update_outstanding_amt` from `save_gl_entries` and `delete_gl_entries` instead of `on_update_with_args` so that one query is required to update `outstanding_amount` rather than calling a query for _each and every_ GL Entry.
* Added validation for GL Entry so that either `party_type` and `party` both must be set or both must be unset. Same for `against_voucher` and `against_voucher_type`

#### Journal Entry:
* Changed logic of calculating balance/unallocated payments such that GL Entry balance is queried to get the balance for the whole (Journal Entry, Account, Party Type, Party) tuple instead of individual Journal Entry Accounts. This prevents overallocating the balance of Journal Entry and prevents losing information about it's outstanding amount.
* Modified `update_reference_in_journal_entry` so that it updates references in multiple Journal Entry Accounts if the first entry's amount is less than the `allocated_amount`
* Removed `is_advance` field from Journal Entry Account as it becomes impossible to get `is_advance` field when calculating advance payments from GL Entries. And also because in my opinion an unallocated payment should always be allocatable from Advance Payments in Purchase/Sales Invoice.
* Removed `reference_due_date` from Journal Entry Account since it wasn't being used for anything.
* Force set `exchange_rate` for entries referencing Purchase Invoice, Sales Invoice or Journal Entry.
* (Journal Entry, Account, Party Type, Party) tuple's exchange rate is calculated as the average of all exchange rates in a (Journal Entry, Account, Party Type, Party) tuple.
* I tried to make `exchange_rate` read-only when referencing voucher documents, however, frappe doesn't allow setting a specific row in a child table as read-only, so if the user changes exchange rate manually for such entries, he/she could see errors for debit != credit!
* Some bug fixes: setting remarks as empty string to prevent a copied value, remarks for Purchase Order using debit amount instead of credit amount

#### Payment Entry:
* Get negative invoices only when Receiving from Payable or Paying to Receivable
* Reload reference details before validating allocated amounts. Print details if `allocated_amount > outstanding_amount` since the client may not show the updated outstanding amount.
* Some bug fixes: auto set `paid_amount` instead of `received_amount` and vice-versa when receiving currency and paying currency is different, replaced an incorrect payment_type == 'Internal Transfer' with 'Receive', display depends on for `exchange_rate` field when it's not 1

#### Payment Reconciliation:
* Using utils functions instead of it's own query to get advance payments
* Include remarks
* Only auto set allocated amount if not manually set
* Updated `get_advance_payment_entries` to filter by `against_account` for Bank/Cash account field

#### Exchange Rate fields
* Set all `exchange_rate` field's precision to 9

# What's Left?

#### Setting read-only on specific child table _row_ field
* If the user changes exchange rate in Journal Entry Account referencing a voucher document, the server reloads the exchange rate from the database so there may be an validation error that debit != credit. But the user will not see the updated exchange rate. In this case, the exchange rate should be read only
 
#### Validation fail due to old outstanding balance not updated in client:
* Payment Entry reloads outstanding amount and then validates if allocated amount <= outstanding amount. If it fails the server-side script will print a validation error (with outstanding and allocated amounts), however, it will not update the UI's outstanding_amount and probably confusing the user.
* Maybe there should be a feature to push changes forcefully to the client

#### Bank Statment Transaction Entry
* I ignored Bank Statment Transaction Entry's use of `reconcile_against_document` because I have no idea how it works.
* Would it require changes to use the new `reconcile_against_document`?

#### Use precision based threshold to check if monetary value > 0 or < 0
* I saw multiple absolute values used to check if a monetary value is significantly greater than 0.
* Some values: > 0.005, >= 0.005, > 0.009
* I suggest replacing them all with values like unless there is a specific reason for using 0.005:
* * `>= 1.0 / (10 ** (precision("field")))`
* * `< 1.0 / (10 ** (precision("field")))`
* Or better yet create a utility function for this purpose
* It would allow currencies with high precision like cryptocurrencies to function properly


SaiFi0102:
I will update on this PR soon

SaiFi0102:
To explain my pull request better and to give a sense of the rationale behind why I thought these changes were necessary here are some screenshots:

## (1) Reference Journal Entry
Advance payment to Supplier: General Party
paymenttosupplier

## (2) Attempt to create a reverse Journal Entry with reference to the above JV
This is because of the validation function `journal_entry.validate_against_jv` which prevents referencing another journal entry while debiting an Asset account or crediting a Liability account. This validation isn't necessary but the current staging branch doesn't have the mechanism to handle this properly, which I'll explain below
2-crdrvalidationproblem

## (3) Bypassing the validation and partially reversing Journal Entry in (1)
Submitted JV (Note amount is 500 rather than the complete 1000, ALSO NOTE, this JV references the JV in (1))
3-bypassedreversejv

## (4) Effect of the reverse Journal Entry in (3)
First: Accounts Payable shows the CORRECT unallocated amount (outstanding balance) of 500
accountspayableok

BUT: Advance Payments section of Purchase Invoice and Payment Reconciliation Tool shows incorrect unallocated amount of 1000 because both these tables query the Journal Entry table rather than the GL Entry table. Advance Payments Section of Invoices and Payment Reconciliation Tool only consider the unallocated amount in the original Journal Entry BUT do not consider the GL Entries that are made against it (to reverse it's outstanding balance)
paymentreconciliationnotok
advancepaymentsnotok

## (5) The problem! Overallocated Journal Entry
The result is that an Invoice can over allocate a Journal Entry. In this case, JV-CT-2018-00001 should ONLY be allocate 500 to any invoice... but since ERPNext doesn't validate this, an incorrect amount is allocated.

Note that accounts payable information about the Journal Entries is lost and JV-CT-2018-00001 has been overallocated!

pinvnotok

accountspayablenotok

## Solution
Calculate outstanding balance/unallocated amount of a Journal Entry by querying GL Entries like: (1: GL Entries that invoice or are unallocated payments) - (GL Entries made against 1:)

## Summary
All the changes in this PR is mostly to fix this problem, then there are some improvements and changes that builds upon trying to fix the issue of reversing a advance payment type entries.

agritheory:
Hey @SaiFi0102 This is a pretty complex fix and I think I understand most of it, but it's hard to know since I 'm not personally familiar with the problem you're facing . One of my action items from the conference is to organize an Accounts module working group and it seems like you'd have a lot to contribute to that. @FinForceConsulting Check this out

SaiFi0102:
Not inactive

SaiFi0102:
Waiting for review

clarkejj:
Hi Saif I viewed your changes - since there are so many plus they are non-trivial, it is hard to know how to debug the travis results -  Just 5x test errors - 2x sales invoice 3X purchase invoice and fortunately all on the same validation code (I don't think the error message is accurate, that this a case here of an inmemory object has changed so now it is inconsistent with what is stored in the database.)  What I suggest to troubleshoot - Starting with check_if_advance_entry_modified(), back out and back in your changes (incrementally by method or by file?) and run the 5x invoice tests to identify exactly what change broke these tests. 
Another option is to back out all or a large part of your changes until the tests pass, then reintroduce your changes in series to identify where the logic error resides. Good luck and thanks for contributing this, John

SaiFi0102:
Thanks @clarkejj I really appreciate it. I figured out what the problem was.

agritheory:
Some comments, as this has been open for a while and I believe should now be merged:
1) The Accounts working group has reviewed these changes and approves. Last week we talked through some issues, @clarkejj offered to help @SaiFi0102 chase some issues in the tests and they were resolved. 
2) The codacy errors are bogus and the new changes _improve_ SQL injection security.

SaiFi0102:
Created a patch

FinForceConsulting:
Screenshot Exist. Need a review for this PR @nabinhait

sagarvora:
@SaiFi0102 is the complete now? add semantic title please.

PR #15565, opened 2018-10-03
Fix 'transporter info' section in delivery note

sagarvora:
- create is_transporter checkbox in supplier doctype
- new field transporter (supplier with is_transporter checked)
- removed GST India fields from core doctype
- removed readonly & allow_on_submit from fields (delivery trip applicable only for in-house delivery)
- read_only transporter_name field (kept for backwards compatibility)
- made necessary changes in delivery_trip implementation
- hidden section's fields from print (preserving v10 convention)
- added transporter field to 'Driver' doctype
- added queries for transporter and driver fields
- GST India fields in regional module
- Fix e-waybill report

@nabinhait Please review.

Screenshot after changes made:

screenshot from 2018-10-04 10-05-35


Alchez:
@sagarvora 

What's the idea / use-case behind splitting the Driver and Transporter doctypes?

I ask because it seems like transporters are a party type that you're subcontracting deliveries through. It might be better handled through Supplier, maybe.

sagarvora:
Hi, @Alchez, thanks for the review. Yes, Transporter DocType was created specifically for third-party transport companies, e.g. 'Fast Transport Pvt Ltd' has driver 'John Doe'. 

I like the idea you're proposing. Will add a supplier type 'Transporter' and create a filtered link field in Delivery Note... sounds good? Also, kindly approve changes in the Delivery Trip DocType.

Alchez:
Sounds good, but I would suggest making a Discuss post to get the community's views on this proposal. Better to design it the right way right now, than to hit obstacles or resistance later on.

Regarding the Delivery Trip changes, they seem fine, but just double-check that it's not suppressing any unwanted validations.

erpnext/hr/doctype/driver/driver.js
```diff
@@ -2,4 +2,13 @@
 // For license information, please see license.txt
 
 frappe.ui.form.on('Driver', {
+	setup: function(frm) {
+		frm.set_query('transporter', function(){
+			return {
+				filters: [
+					['Supplier', 'supplier_group', 'in', ['Transporter', 'transporter']]
```
nabinhait: Don't use hard-coded value "Transporter" in code. The user will never know until they read code / docs.
Let's allow all suppliers.

nabinhait (changes requested):

erpnext/hr/doctype/driver/driver.js
sagarvora: @nabinhait I've made the discussed changes.

Issue #15571, opened 2018-10-03
Add provision for prebilled subscription invoices


joeirimpan:
## Feature Goals:
* Ability to generate invoices for customer who paid for the subscription beforehand. Usually service subscriptions are prepaid. This change account for that case.

## Extra fields:
* `prebilled` boolean field

## Discovery:
* Easily discoverable from `SUBSCRIPTION PERIOD` section

### Field
![prebilled_sub](https://user-images.githubusercontent.com/11286580/46406893-a70cfc00-c72a-11e8-96d0-da8d829d7a0a.png)
### Invoice generated using prebilled flag
![prebilled_sub2](https://user-images.githubusercontent.com/11286580/46407076-43370300-c72b-11e8-891e-6b98c593f693.png)

## Links to existing DocTypes:
[Subscription](https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/subscription/subscription.py#L14)

PR #15572, opened 2018-10-03
feat: Add provision for prebilled subscription invoices

joeirimpan:
Discussion: https://github.com/frappe/erpnext/issues/15571


chdecultot:
HI @joeirimpan,

Nice to see some additions to the subscription module!

I am trying to understand the business goal of your PR but am not getting it, I'm sorry...
Can you please explain what is the difference for a subscription if you click on the `prebilled` field and if not ?

The tests seem to show that the next invoice will only be generated by clicking on the `prebilled` button. Am I understanding it correctly ? Does that mean that if you don't click it, the next invoice will not be generated ?

It seems that this invoice is still not automatically paid, whereas the initial goal seemed to generate paid invoices automatically ? 

> Ability to generate invoices for customer who paid for the subscription beforehand.

Maybe the button should be "Prepaid" instead of "Prebilled" and it should not change the invoice generation system but add the payment within the invoice during the generation if that's what you meant.


I have the feeling I am missing the point here and I would be glad if you could explain it to me please.

Thanks a lot!

joeirimpan:
@chdecultot 
Purpose of this feature is the ability to generate prepaid invoices. Lets say a user subscribes to a plan for 30 days and make the payment right there, there is no way for generating an invoice then and there if i m understanding correctly. After 30 days, initiate another payment and collect it at the beginning of the period and so on.

The way subscription invoices are generated now are based on the invoice end date.  So, there is no concept of prepaid invoice here. But i think this might be on the [plan](https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/subscription/subscription.py#L221).

`prebilled`/`prepaid` is merely a boolean flag marking this as prepaid subscription. If you do not mark it a `prebilled`/`prepaid`, invoice will only be generated at the end of invoice date i.e, `current_invoice_end`.

Can you take a quick look into https://stripe.com/docs/billing/invoices/subscription?
Open for suggestions and corrections.

Thanks!

chdecultot:
Hi @joeirimpan,

Thanks a lot for your explanation.

So to recap you are proposing to generate the invoice at the beginning of the period instead of the end ? Sorry for not getting it earlier.

May I suggest using a more self explanatory word in your checkbox like "Generate the invoice at the beginning of the period" or something similar ?

Additionally I would not use the word "paid" in your variables as they is not linked with any payments at this stage (the invoice is only submitted immediately, then it's up to the user to generate the corresponding payment if I understand your test case correctly).
I may be wrong but generating an invoice at the beginning of the period is not necessarily linked to a prepayment: you can bill the 1st of the month and be paid only the 5th, so both need to be decoupled.

A future enhancement could be to add the possibility to automatically mark the invoice as paid (generate a corresponding payment entry or add the payment in the invoice) for prepaid invoices.

Just my 2cts !  :wink:

joeirimpan:
> May I suggest using a more self explanatory word in your checkbox like "Generate the invoice at the beginning of the period" or something similar ?

Makes sense

> the invoice is only submitted immediately, then it's up to the user to generate the corresponding payment if I understand your test case correctly

Yes. With prepaid subscriptions, invoice is generated in the first processing. thats it. Its up to user for paying the invoice and submitting. I added creating payment record in the test case to describe the flow.

I ll amend the changes. Thanks for the review. :thumbsup:

joeirimpan:
@chdecultot I ve updated the PR if you haven't noticed. Can you please take a look and see if it is good to go?

nabinhait:
@joeirimpan Can you please send this PR to staging-fixes branch?

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#which-branch

joeirimpan:
Closing in favour of https://github.com/frappe/erpnext/pull/15700

PR #15589, opened 2018-10-04
Server side script report for sales analytics 

deepeshgarg007:
Server side script report for sales analytics
Screenshots:

By Item Group:

screen shot 2018-10-04 at 7 10 12 pm
By Territory based on value:

screen shot 2018-10-04 at 7 10 47 pm

By Territory based on quantity:

screen shot 2018-10-04 at 7 11 30 pm

By Customer:
screen shot 2018-10-04 at 7 08 24 pm


adnanmunir83:
in my point of view
in the above report, as you can see in pictures, by customer and by Territory, it's set for codes and in analysis report it should show Names as well for batter analysis.

rmehta:
Thanks @adnanmunir83 for the comment.

@deepeshgarg007 Concat customer ID + customer name (item code + item name) in both reports.

PR #15596, opened 2018-10-05
Payroll improvements

nabinhait:
**Deduct Salary for Previous Leaves**
Based on prev leaves the equivalent amount is calculated and added in salary slip deductions

screen shot 2018-10-05 at 2 57 26 pm

**Salary Component: Prorated based on attendance**
If checked, salary component will be calculated based on attendance (any kind of leaves)

screen shot 2018-10-05 at 2 59 09 pm

**Bulk Leave Applications**
screen shot 2018-10-05 at 2 58 34 pm

**Recurring Additional Salary**
screen shot 2018-10-05 at 3 04 44 pm


rmehta (changes requested):
Update docs and code comments!

PR #15608, opened 2018-10-07
[Fix] Healthcare - Inpatient Record - Test fix

pjamsheer:
- Inpatient Record - Test fix


rmehta:
Please can we have better commit / PR subjects https://conventionalcommits.org/

PR #15628, opened 2018-10-09
Copy Payment terms from Sales Order when creating Sales Invoice from SO or DN and also update due date

KanchanChauhan:
Issue:
On creating the Sales Invoice from Sales Order or Delivery Note, the payment schedule is not created. On save the payment schedule does appear based on Payment Terms Template but the manually added payment terms or edited payment terms does not. And also there is a need to recalculate the due date as it is based on Sales Invoice date and in case of Delivery Note, amount can be different too in case of partial delivery so needs recalculation.
screen shot 2018-10-02 at 11 31 57 pm

Solution:
In Sales Order just copy over the payment schedule and calculate the new due date.
In Delivery Note, since it does not have the payment schedule, check if items are against SO, if yes, then get payment schedule from SO and re-calculate.

Sales Order payment terms with manual edits:
screen shot 2018-10-09 at 2 40 16 pm

Sales Invoice payment terms created when Sales Invoice created from Sales Order:
screen shot 2018-10-09 at 2 40 46 pm

Sales Invoice payment terms created from Delivery Note on making partial delivery.
screen shot 2018-10-09 at 2 57 04 pm

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.py
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.py
@@ -6,6 +6,7 @@ from __future__ import unicode_literals
 import frappe
 import frappe.defaults
 from erpnext.controllers.selling_controller import SellingController
+from erpnext.controllers.accounts_controller import get_due_date
 from erpnext.stock.doctype.batch.batch import set_batch_nos
 from erpnext.stock.doctype.serial_no.serial_no import get_delivery_note_serial_no
 from frappe import _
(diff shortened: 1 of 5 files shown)


erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -410,6 +411,17 @@ def set_missing_values(source, target):
 		target.update(get_company_address(target.company))
 		if target.company_address:
 			target.update(get_fetch_values("Sales Invoice", 'company_address', target.company_address))
+		against_sales_order = target.get("items")[0].sales_order
+		if against_sales_order:
```
rmehta: needs test coverage

rmehta (dismissed):

erpnext/stock/doctype/delivery_note/delivery_note.py
KanchanChauhan: Test Case: https://github.com/frappe/erpnext/pull/15628/files#diff-0a3847ae36b2ba240c1b291b370aef43R640

KanchanChauhan: @rmehta test cases added.

dottenbr:
Waiting on merge by @nabinhait

rohitwaghchaure:
@KanchanChauhan 

Sometimes user wants, system has to calculate the due date based on the invoiced date(Posting Date), how we will manage this scenario here. Because there might be chances that sales order date and sales invoice date is different. Also how it could be handle in the case of making one invoice from multiple sales order.

I think better solution will be we need order date on the invoice and user will be able to choose from which date he wants to calculate the due date like from order date or from the posting date. Please share your comments

PR #15632, opened 2018-10-09
 Eway Bill for Sales Invoice

finbyz:
Currently, the Eway Bill Report is generated only the Delivery Note. But some companies generate Eway Bill report directly through the Sales Invoice. So this report can now be generated for Delivery Note as well as Sales Invoice through filters.


test2a:
Nice. haven't tested this out but does this version or the latest version support for "other" column in the away bill as updated in the portal?

Like if there are items other than the stock, the away bill should only show taxable value of goods and not other line items which should show up in the other column?

test2a:
Nice. haven't tested this out but does this version or the latest version support for "other" column in the away bill as updated in the portal?

Like if there are items other than the stock, the away bill should only show taxable value of goods and not other line items which should show up in the other column?

sagarvora:
Hi @finbyz, 

Thanks for the contribution!

- Please check https://github.com/frappe/erpnext/pull/15565 for changes made to your implementation. Please update your PR accordingly. 

- Please ensure that fields exclusive to GST India users are added using the regional doctype. See #15565 for example.

- It might be prudent to change the branch to `staging-fixes` so that the report can be tested before it is used in production instances.

- Kindly rename the report to "GST e-Way Bill" or something more descriptive.

- If you are modifying a JSON file, it is a good idea to highlight the changes made in your PR description so that a reviewer can easily locate the same.

finbyz:
Hi @sagarvora 

So you are saying I should create a PR in `staging-fixes` branch rather that hotfix branch?
And should I add `eway_bill_fields` variable in Sales Invoice in regional setup.py rather than creating fields for Transporter Info Section in Sales Invoice?

sagarvora:
Hi @finbyz,

- Yes, use `staging-fixes`.
- Transporter Info section has both standard and india-specific fields. If standard fields may be useful worldwide, we can add those to sales invoice doctype. @nabinhait What's your opinion on this?

In either case, regional fields of Delivery Note have already been added to setup.py (pull latest staging-fixes) so you can use them in sales invoice too, by modifying setup.py

If you need further clarification, let me know.

PR #15636, opened 2018-10-09
Validation to check duplicate student leave application

rohitwaghchaure:
screen shot 2018-10-09 at 6 39 58 pm

### DIFF
--- a/erpnext/education/doctype/student_leave_application/student_leave_application.py
+++ b/erpnext/education/doctype/student_leave_application/student_leave_application.py
@@ -4,7 +4,29 @@
 
 from __future__ import unicode_literals
 import frappe
+from frappe import _
+from frappe.utils import get_link_to_form
 from frappe.model.document import Document
 
 class StudentLeaveApplication(Document):
-	pass
+	def validate(self):
+		self.validate_duplicate()
+
+	def validate_duplicate(self):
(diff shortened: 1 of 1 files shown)


erpnext/education/doctype/student_leave_application/student_leave_application.py
```diff
@@ -4,7 +4,28 @@
...
+				name != %(name)s and student = %(student)s and docstatus < 2
+		""", {
+			'from_date': self.from_date,
+			'to_date': self.to_date,
+			'student': self.student,
+			'name': self.name
+		}, as_dict=1)
+
+		if data:
+			link = "<a href='#Form/Student Leave Application/{0}'>{1}</a>".format(data[0].name, data[0].name)
```
Alchez: You can use `get_link_to_form` to generate this.

erpnext/education/doctype/student_leave_application/student_leave_application.py
```diff
@@ -4,7 +4,28 @@
...
+			frappe.throw(_("Leave application {0} is already exists for the same period against the student {1}")
```
Alchez: This is minor, but it should read `{0} already exists` and `against student {1}`.

Alchez (changes requested):

PR #15637, opened 2018-10-09
File api

Thunderbottom:
remove file_manager stuff and migrate to file-api


nabinhait:
depends on https://github.com/frappe/frappe/pull/6231

PR #15640, opened 2018-10-10
[Fix]Deleted Unnecessary Column 

deepeshgarg007:


erpnext/selling/report/customer_credit_balance/customer_credit_balance.py
```diff
@@ -40,7 +40,6 @@ def execute(filters=None):
 
 def get_columns(customer_naming_type):
 	columns = [
-		_("Name") + ":Link/Customer:120",
```
nabinhait: You should also remove the data

PR #15644, opened 2018-10-10
Update Items in Submit state can add new row

kickapoo:
In Sales Order and Purchase Order when docstatus is submitted user can
use Update Item btn to add new child Items

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -1083,8 +1083,6 @@ def set_purchase_order_defaults(parent_doctype, parent_doctype_name, child_docna
 	"""
 	p_doctype = frappe.get_doc(parent_doctype, parent_doctype_name)
 	child_item = frappe.new_doc('Purchase Order Item', p_doctype, child_docname)
-	conversion_rate = flt(p_doctype.get('conversion_rate')) or 1.0
-
 	item = frappe.get_doc("Item", item_code)
 	child_item.item_code = item.item_code
 	child_item.item_name = item.item_name
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -6,6 +6,8 @@ import frappe, erpnext
 import json
 from frappe import _, throw
 from frappe.utils import today, flt, cint, fmt_money, formatdate, getdate, add_days, add_months, get_last_day, nowdate
+from erpnext.stock.doctype.item.item import get_last_purchase_details
+from erpnext.stock.get_item_details import get_conversion_factor
 from erpnext.setup.utils import get_exchange_rate
 from erpnext.accounts.utils import get_fiscal_years, validate_fiscal_year, get_account_currency
 from erpnext.utilities.transaction_base import TransactionBase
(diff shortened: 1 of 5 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -1050,24 +1052,79 @@ def get_supplier_block_status(party_name):
...
 		if child_item.billed_amt > (flt(d.get("rate")) * flt(d.get("qty"))):
 			frappe.throw(_("Row #{0}: Cannot set Rate if amount is greater than billed amount for Item {1}.")
 						 .format(child_item.idx, child_item.item_code))
 		else:
 			child_item.rate = flt(d.get("rate"))
 		child_item.flags.ignore_validate_update_after_submit = True
-		child_item.save()
+		if new_child_flag:
+			child_item.insert()
+			frappe.db.commit()
```
nabinhait: Don't commit in between, it the thread breaks later, it will not rollback.

erpnext/controllers/accounts_controller.py
```diff
@@ -1050,24 +1052,79 @@ def get_supplier_block_status(party_name):
...
+	child_item = frappe.new_doc('Purchase Order Item', p_doctype, child_docname)
+	conversion_rate = flt(p_doctype.get('conversion_rate')) or 1.0
+
+	item = frappe.get_doc("Item", item_code)
+	child_item.item_code = item.item_code
+	child_item.item_name = item.item_name
+	child_item.description = item.description
+	child_item.schedule_date = p_doctype.schedule_date
+	child_item.uom = item.stock_uom
+	child_item.company = p_doctype.company
```
nabinhait: "Company" is not a standard field in Purchase Order Item

erpnext/controllers/accounts_controller.py
```diff
@@ -1050,24 +1052,79 @@ def get_supplier_block_status(party_name):
...
+	return child_item
+
+
+def set_purchase_order_defaults(parent_doctype, parent_doctype_name, child_docname, item_code):
+	"""
+	Returns a Purchase Order Item child item containing the default values
+	"""
+	p_doctype = frappe.get_doc(parent_doctype, parent_doctype_name)
+	child_item = frappe.new_doc('Purchase Order Item', p_doctype, child_docname)
+	conversion_rate = flt(p_doctype.get('conversion_rate')) or 1.0
```
nabinhait: not used

erpnext/controllers/accounts_controller.py
```diff
@@ -1050,24 +1052,79 @@ def get_supplier_block_status(party_name):
...
+
+	item = frappe.get_doc("Item", item_code)
+	child_item.item_code = item.item_code
+	child_item.item_name = item.item_name
+	child_item.description = item.description
+	child_item.schedule_date = p_doctype.schedule_date
+	child_item.uom = item.stock_uom
+	child_item.company = p_doctype.company
+	child_item.conversion_factor = get_conversion_factor(item_code, item.stock_uom).get("conversion_factor") or 1.0
+	child_item.base_rate = 1 # Initiallize value will update in parent validation
```
nabinhait: Did not understand why is it needed?

nabinhait (changes requested):
Add tests and docs

erpnext/controllers/accounts_controller.py
kickapoo: @nabinhait You are right and I saw that, but my issue ia that if I dont commit in that stage I dont have "ID" of the record to relate to parent.  I dont know how else to pass it in the parent without commit.

erpnext/controllers/accounts_controller.py
kickapoo: As you said in the previous comment, "Don't commit in between, it the thread breaks later, it will not rollback." because I am inserting I put init value for those fields as they are mandatory and their values are updated to correct ones as the parent doc is saved and re-validates child items.

Alchez:
@kickapoo 

Just wanted some clarifications on this.
- What workflow is this a solution for? As opposed to simply cancelling and amending the documents?
- How would this affect the stock, accounting, manufacturing, etc. elsewhere in the system?

kickapoo:
@Alchez 

There is no addition to a new workflow. Currently, there is the `Update Items` btn that allow you to make basic changes to qty/rate when Doctype is submitted.  This PR just add the ability to add new as well. 
This has the same effect as the already `Update Items` btn.

erpnext/controllers/accounts_controller.py
kickapoo: I have research this. So the reason that I am putting the company is due other fields options like 
https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order_item/purchase_order_item.json#L786 .
With company can "fetch" the correct values.

erpnext/controllers/accounts_controller.py
kickapoo: Without it I am getting.
screenshot 2018-10-16 15 31 11

erpnext/controllers/accounts_controller.py
```diff
@@ -1050,24 +1052,79 @@ def get_supplier_block_status(party_name):
...
 	data = json.loads(trans_items)
+
 	for d in data:
-		child_item = frappe.get_doc(parent_doctype + ' Item', d.get("docname"))
-		child_item.qty = flt(d.get("qty"))
+		new_child_flag = False
+		if not d.get("docname"):
+			new_child_flag = True
+			if parent_doctype == "Sales Order":
+				child_item  = set_sales_order_defaults(parent_doctype,
```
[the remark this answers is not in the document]
kickapoo: https://github.com/frappe/erpnext/pull/15644/commits/58dcb2211abec126d8761563a6b83598ca75c07c

kickapoo:
@nabinhait your assignment on this means that you will deal with Unit Tests? If not, can you please point me some similar code lines, for example I can not find the "Update Items" btn UnitTest to add a new Test.

nabinhait:
@kickapoo I think there are no tests available for `update_child_qty_rate`.

Please add a fresh one. Basically what you have to do, in tests you need check the outcome of `update_child_qty_rate` function. Like whether the item is added to doc, reserved_qty is increased or not. If the SO is already delivered, after addition of items, the status of the SO updated or not etc.

kickapoo:
I did found a test to update
```
#apps/erpnext/erpnext/selling/doctype/sales_order/test_sales_order.py
	def test_update_child_qty_rate(self):
		so = make_sales_order(item_code= "_Test Item", qty=4)
		create_dn_against_so(so.name, 4)
		make_sales_invoice(so.name)

		existing_reserved_qty = get_reserved_qty()

		trans_item = json.dumps([{'item_code' : '_Test Item', 'rate' : 200, 'qty' : 7, 'docname': so.items[0].name}])
		update_child_qty_rate('Sales Order', trans_item, so.name)

		so.reload()
		self.assertEqual(so.get("items")[0].rate, 200)
		self.assertEqual(so.get("items")[0].qty, 7)
		self.assertEqual(so.get("items")[0].amount, 1400)
		self.assertEqual(so.status, 'To Deliver and Bill')

		self.assertEqual(get_reserved_qty(), existing_reserved_qty + 3)

```

kickapoo:
@nabinhait  I am trying to add the Unit Test but I can not even run the current Frappe Tests.

Please remember I have custom app along with frappe and erpnext

Based on the docs: https://frappe.io/docs/user/en/guides/automated-testing/unit-testing

`bench --verbose run-tests --doctype "Sales Order` fails of before_hooks with 
```
frappe.exceptions.LinkValidationError: Could not find Parent Item Group: All Item Groups
```

Running with `bench --verbose run-tests --doctype "Sales Order" --skip-before-tests`
getting 
```
frappe.exceptions.LinkValidationError: Could not find Parent Cost Center: _Test Company - _TC
```

So,  I can not update Unit Tests if I can not run the current ones... 

I can tell that the custom app is affecting this but on the other hand how I can run them ?
May I assume that the only way with a seperate site database ?

kickapoo:
In process

kickapoo:
@nabinhait The build is failing with unrelated issue from this PR.

dottenbr:
@nabinhait what's your status on this one?

creamdory:
Any status on this?

PR #15691, opened 2018-10-15
[Fix] patch - change healthcare desktop icons

pjamsheer:
- Fix - patch - change healthcare desktop icons


nabinhait:
@pjamsheer This is already fixed via https://github.com/frappe/erpnext/pull/15689/files

But you can add the line for returning the patch if healthcare domain is not enabled.

PR #15695, opened 2018-10-15
Deprecating subscriber doctype

deepeshgarg007:
- Deprecated subscriber doctype
- Subscriptions are now directly linked to customer master

screen shot 2018-10-15 at 5 59 34 pm


erpnext/accounts/doctype/subscription/subscription.json
```diff
@@ -1,9 +1,10 @@
 {
  "allow_copy": 0, 
+ "allow_events_in_timeline": 0, 
  "allow_guest_to_view": 0, 
  "allow_import": 0, 
  "allow_rename": 0, 
- "autoname": "ACC-SUB-.YYYY.-.#####", 
+ "autoname": "naming_series:",
```
nabinhait: naming_series field not needed

nabinhait (changes requested):

PR #15699, opened 2018-10-16
Fix Duplicated Sales Invoice After Syncing Invoices

tundebabzy:

This is related to the issues previously reported about invoices being duplicated while syncing offline POS invoices (see [this discussion](https://discuss.erpnext.com/t/duplicated-sales-invoice-when-using-pos/41736/11) ).

The issue still exists even though reproducing it has been elusive. The duplication problem happens in high traffic multiple client environment. Here are some points that have inspired this PR:
- For single client usage of the POS, the POS is mostly bulletproof. The offline pos is named as the timestamp of the invoice's creation which is fairly bulletproof if you ignore the fact that you cannot trust the computer's time to be correct everytime. However, things are different when you have more than one client.
- Particularly in high traffic usage, the probability of having a clash increases as clients work in parallel. Also remember that speed of execution also comes into play.
- The offline pos name is technically as good as the invoice number so should be unique. However, ERPNext only tries to preserve the data integrity for this field at the code layer and not the database layer. 
- In the case where two different payloads are sent to ERPNext to create invoices (`make_invoice` in `pos.py`), if ERPNext finds an offline pos name clash, the latter payload received will be ignored which can potentially lead to loss of genuine data.

In summary, 
- [x] The offline POS name needs to be a _real unique identifier_.
- [X] ERPNext _must_ add the unique constraint at the database layer in case.

This PR does the latter and the first point needs to be implemented as well to add more resiliency in race conditions.

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
@@ -314,7 +314,7 @@
    "in_standard_filter": 0,
    "label": "Offline POS Name",
    "length": 0,
-   "no_copy": 0,
+   "no_copy": 1,
    "permlevel": 0,
    "precision": "",
    "print_hide": 1,
@@ -325,7 +325,7 @@
    "reqd": 0,
    "search_index": 0,
    "set_only_once": 0,
-   "unique": 0
+   "unique": 1
   },
   {
    "allow_bulk_edit": 0,
@@ -4773,7 +4773,7 @@
  "istable": 0,
  "max_attachments": 0,
  "menu_index": 0,
- "modified": "2018-07-17 13:46:52.449142",
+ "modified": "2018-10-17 18:28:52.449142",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Sales Invoice",
--- a/erpnext/patches/v10_0/deduplicate_offline_pos.py
+++ b/erpnext/patches/v10_0/deduplicate_offline_pos.py
@@ -1,0 +1,26 @@
+import frappe
+
+
+def execute():
+    duplicates = frappe.db.sql(
+        'select name, offline_pos_name from `tabSales Invoice` where '
+        'offline_pos_name is not null group by '
+        'offline_pos_name having count(*) > 1'
+    )
+
+    for duplicate in duplicates:
+        keep, offline_pos_name = duplicate
+        duplicate_list = frappe.db.sql(
+            'select name from `tabSales Invoice` where '
+            'offline_pos_name=%s',
+            offline_pos_name
+        )
+
+        counter = 0
+        for item in duplicate_list:
+            name, = item
+            if name != keep:
+                si = frappe.get_doc('Sales Invoice', name)
+                counter += 1
+                new_name = '{0}-{1}'.format(si.offline_pos_name, counter)
+                si.db_set('offline_pos_name', new_name)
--- a/erpnext/patches/v10_0/deduplicate_offline_pos.py
+++ b/erpnext/patches/v10_0/deduplicate_offline_pos.py
@@ -3,16 +3,20 @@ import frappe
 
 def execute():
     duplicates = frappe.db.sql(
-        'select name, offline_pos_name from `tabSales Invoice` where '
-        'offline_pos_name is not null group by '
-        'offline_pos_name having count(*) > 1'
+        """
+        SELECT name, offline_pos_name FROM `tabSales Invoice`
+        WHERE offline_pos_name IS NOT NULL
+        GROUP BY offline_pos_name having count(*) > 1
+        """
     )
 
     for duplicate in duplicates:
         keep, offline_pos_name = duplicate
         duplicate_list = frappe.db.sql(
-            'select name from `tabSales Invoice` where '
-            'offline_pos_name=%s',
+            """
+            SELECT name FROM `tabSales Invoice` WHERE
+            offline_pos_name=%s
+            """,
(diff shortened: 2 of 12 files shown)


rohitwaghchaure:
@tundebabzy 
Did you able to replicate the issue locally?

erpnext/patches/v10_0/deduplicate_offline_pos.py
```diff
@@ -0,0 +1,26 @@
...
+    duplicates = frappe.db.sql(
+        'select name, offline_pos_name from `tabSales Invoice` where '
+        'offline_pos_name is not null group by '
+        'offline_pos_name having count(*) > 1'
+    )
+
+    for duplicate in duplicates:
+        keep, offline_pos_name = duplicate
+        duplicate_list = frappe.db.sql(
+            'select name from `tabSales Invoice` where '
```
rohitwaghchaure: Use quote  """ """"
For reference check below link
https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v10_0/update_user_image_in_employee.py#L10-L19

erpnext/patches/v10_0/deduplicate_offline_pos.py
```diff
@@ -0,0 +1,26 @@
...
+        )
+
+        counter = 0
+        for item in duplicate_list:
+            name, = item
+            if name != keep:
+                si = frappe.get_doc('Sales Invoice', name)
+                counter += 1
+                new_name = '{0}-{1}'.format(si.offline_pos_name, counter)
+                si.db_set('offline_pos_name', new_name)
```
rohitwaghchaure: Can you provide some information about this patch, based on code it seems that you are

erpnext/accounts/doctype/sales_invoice/sales_invoice.json
```diff
@@ -1,4875 +1,4875 @@
...
- "beta": 0,
- "creation": "2013-05-24 19:29:05",
- "custom": 0,
- "default_print_format": "",
- "docstatus": 0,
- "doctype": "DocType",
- "document_type": "",
- "editable_grid": 0,
- "engine": "InnoDB",
+ "allow_copy": 0,
```
rohitwaghchaure: Because of space, file showing lot of changes, please remove the space so that we can see the actual changes

tundebabzy:
@rohitwaghchaure The issue is not reproducible on my local setup but I have personally seen where this bug is affecting a real business. The issue seems to be as a result of multiple clients syncing offline invoices at the same time. Its just a hypothesis but I'm pretty sure we are having offline pos name clashes by using timestamp.

erpnext/patches/v10_0/deduplicate_offline_pos.py
tundebabzy: This patch will do the following:
- check for invoices that have the same offline_pos_name. This will give us a list of invoices which have been duplicated.
- If there are any duplicates, append a number to them. e.g, we find invoices SINV-0001 and SINV-0002 have the same offline POS name of 1234. SINV-0001 offline pos name remains as 1234 and SINV-0002 becomes 1234-1

erpnext/accounts/doctype/sales_invoice/sales_invoice.json
tundebabzy: It was autogenerated by frappe. Maybe I'll do it manually

tundebabzy:
In order to help with testing the patch to see that it works as expected, you can simply duplicate any is_pos invoice and save it. It will have a new invoice number but the offline pos name will be the same as the one from which you duplicated it. Like I said in the review comment, when the patch runs, one of the invoices will be appended with 1 essentially making sure that all offline pos name in the database are unique before frappe adds the unique constraint to the column

tundebabzy:
After investigation, some clients are sending in the same data multiple times. We were able to find an instance where the same data was send 4 times. Sometimes the requests are coming in at almost the same time so they are able to slip into the database. The unique constraint at the database layer has ensured that no more duplicates are saved.
I'm closing for now to open a cleaner PR. 

PS: some more feedback on this PR is will be greatly appreciated. I wouldn't want to spend time preparing another PR only to find out that there is no intention to merge this

tundebabzy:
After investigation, some clients are sending in the same data multiple times. We were able to find an instance where the same data was send 4 times. Sometimes the requests are coming in at almost the same time so they are able to slip into the database. The unique constraint at the database layer has ensured that no more duplicates are saved.
I'm closing for now to open a cleaner PR. 

PS: some more feedback on this PR is will be greatly appreciated. I wouldn't want to spend time preparing another PR only to find out that there is no intention to merge this

tundebabzy:
After investigation, some clients are sending in the same data multiple times. We were able to find an instance where the same data was send 4 times. Sometimes the requests are coming in at almost the same time so they are able to slip into the database. The unique constraint at the database layer has ensured that no more duplicates are saved.
I'm closing for now to open a cleaner PR. 

PS: some more feedback on this PR is will be greatly appreciated. I wouldn't want to spend time preparing another PR only to find out that there is no intention to merge this

tundebabzy:
After investigation, some clients are sending in the same data multiple times. We were able to find an instance where the same data was send 4 times. Sometimes the requests are coming in at almost the same time so they are able to slip into the database. The unique constraint at the database layer has ensured that no more duplicates are saved.
I'm closing for now to open a cleaner PR. 

PS: some more feedback on this PR is will be greatly appreciated. I wouldn't want to spend time preparing another PR only to find out that there is no intention to merge this

PR #15700, opened 2018-10-16
feat: Add provision for prebilled subscription invoices

joeirimpan:
(cherry picked from commit db33e6304d3167d0a2c0c9cee4a9405f5833120e)

Discussion: #15571
Prev PR: https://github.com/frappe/erpnext/pull/15572


barredterra:
Hey @joeirimpan, what are `is_postpaid_to_invoice()` and `is_prepaid_to_invoice()` supposed to do? They're causing much trouble for me, because in [this case](https://github.com/frappe/erpnext/blob/19e12cfb448aac4edb1c752bcb328cc430109bae/erpnext/accounts/doctype/subscription/test_subscription.py#L683-L714) (#29626) both methods return `False` at the same time, breaking the subscription.

Why is `is_postpaid_to_invoice` not the opposite of `is_prepaid_to_invoice`?

PR #15706, opened 2018-10-16
Create a Delivery Trip from multiple Delivery Notes in list view

Alchez:
**Screenshots / GIFs:**

dn-to-dt-erpn

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note_list.js
+++ b/erpnext/stock/doctype/delivery_note/delivery_note_list.js
@@ -10,5 +10,47 @@ frappe.listview_settings['Delivery Note'] = {
 		} else if (doc.grand_total === 0 || flt(doc.per_billed, 2) == 100) {
 			return [__("Completed"), "green", "per_billed,=,100"];
 		}
+	},
+	onload: function (doclist) {
+		const action = () => {
+			const selected_docs = doclist.get_checked_items();
+			const docnames = doclist.get_checked_items(true);
+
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/delivery_note/delivery_note_list.js
```diff
@@ -10,5 +10,47 @@ frappe.listview_settings['Delivery Note'] = {
...
+							method: "frappe.model.mapper.map_docs",
+							args: {
+								"method": "erpnext.stock.doctype.delivery_note.delivery_note.make_delivery_trip",
+								"source_names": docnames,
+								"target_doc": cur_frm.doc
+							},
+							callback: function (r) {
+								if (!r.exc) {
+									frappe.model.sync(r.message);
+									cur_frm.dirty();
```
nabinhait: Don't use `cur_frm` in new code, we are in process of deprecating it in near future.

nabinhait (changes requested):

erpnext/stock/doctype/delivery_note/delivery_note_list.js
Alchez: I tried avoiding that, but I wasn't sure how to get the `frm` object for the new Delivery Trip instance that is created. Maybe the `new_doc` Promise can return it?

Alchez:
@nabinhait, @netchampfaris 

Any suggestions on how to proceed with this?

PR #15723, opened 2018-10-17
Update Delivery Trip status based on visited stops

Alchez:
**Changes:**

- Schema additions:
  - **Delivery Trip:** `status` (Draft, Scheduled, Cancelled, In Transit, Completed)
  - **Delivery Stop:** `visited` (0 or 1)
- Add indicator in Delivery Stop child table based on whether the stop was visited or not.
- Update Delivery Trip status based on number of visited stops.
- Added patch to update old Delivery Trip statuses.
- Updated tests.

<details>
<summary>Previous Note</summary>
While trying to run tests (`bench run-tests --doctype "Delivery Trip"`), I get the following error (even though I'm not changing the Departure Time):

	UpdateAfterSubmitError: Not allowed to change Departure Time after submission
			
The error is being thrown because the Trip object's `departure_time` value is different from the `departure_time` value stored in the database.
			

	test_delivery_trip_status_completed
	OBJECT: datetime.datetime(2018, 10, 22, 2, 29, 30, 610796)
	DATABASE: datetime.timedelta(0, 8970, 610796)
			
	test_delivery_trip_status_in_transit
	OBJECT: datetime.datetime(2018, 10, 22, 2, 29, 30, 725078)
	DATABASE: datetime.timedelta(0, 8970, 725078)
			
Can someone please guide on how to fix this?
			
@rmehta @nabinhait 
</details>


nabinhait:
@Alchez Use print statement and try to find out which function is changing `departure_time` value.

Alchez:
@nabinhait 

Just an update. The issue was being caused because I was initally trying to set a `datetime.datetime` object in a `Time`-type field, which actually expects a `datetime.timedelta` object.

This was causing the database to be correct, but the Delivery Trip object to have the wrong format.

Pushed a fix right now.

Alchez:
@nabinhait, @rmehta 

What's needed on this?

erpnext/patches/v11_0/update_delivery_trip_status.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	frappe.reload_doc('stock', 'doctype', 'delivery_trip')
+	frappe.reload_doc('stock', 'doctype', 'delivery_stop', force=True)
+
+	for trip in frappe.get_all("Delivery Trip"):
+		trip_doc = frappe.get_doc("Delivery Trip", trip.name)
+
```
rmehta: this code is almost duplicate of `update_status` in `delivery_trip.py`

rmehta (changes requested):

erpnext/patches/v11_0/update_delivery_trip_status.py
Alchez: Yeah, this is intentional, since I don't want to update the `modified` dates of the old trips, and have them pushed up in the list view.

PR #15724, opened 2018-10-17
Do not consider current leave application while submitting leave

shreyashah115:


rmehta:
@shreyashah115 lets get into the habit of writing tests, even for fixes!

PR #15725, opened 2018-10-17
Account creation after insert customer and supplier

mhbu50:
image


rmehta:
Test Cases!

rmehta:
A separate account is not needed since Party is already a dimension of the ledger. Please explain the use case in a proposal.

mhbu50:
@rmehta Most accountants in Saudi Arabia they prefer to create a separate account for each Customers & suppliers and employees.

And I think the use case is already there but I just need to make it automated:

image

This PR will create new account and add it to customer or supplier on creation.

PR #15727, opened 2018-10-17
fix(hr) calculate net pay base on  working days of the payroll frequency

TaiLouis:

Hi there!
I got a bug with net pay when change start_date or end_date in salary slip. It dont calculate salary base on real working days.

Issue:
My employee get $2000 per month. He had left at 15/10. So I change end_date to 15/10 but it still calculate $2000 for 15 working days instead of $2000 /30 * 15

Breaking changes:
Before:
if with lwp
     amount = amount * payment_days / total_workingdays
else 
     amount = default
After:
if with lwp
     amount = amount  * payment_days / total_working_days_in_paroll_frequency
else 
     amount = amount * total_workingdays / total_working_days_in_paroll_frequency

    &: total_working_days_in_paroll_frequency = 30
    &: total_workingdays = 15


nabinhait:
@TaiLouis 👍 for first contribution.

The problem here is you are setting "End Date" as 15/10. Instead you should process salary for the entire month as you have monthly structure. And as the employee left on 15/10, you should set "relieving_date" on employee master. Then system will automatically reduce payment days to 15 days and working_days will still remain as 30.

Please let me know if I am correct.

TaiLouis:
@nabinhait Thank for quick reply!
In case: In a month an employee have two salary structure assignments as probation and official. So I want to create 2 salary slips in a month for this employee with exactly working days. 
I have some ideas for this situation.
+ allow user select salary structure when creating salary slip
+ add field to_date for salary structure assignment
+ add total_working_days _of_payroll_frequency
Im on the way improving all of features. So let me know if it no need to do. Thank so much.

TaiLouis:
I think the main problem here is you assume just 1 salary structure assignment for a payroll and the system work well. But now the system is a little bit rigid. Why you allow to set start_date and end_date without any changes?

thuanpq:
> @TaiLouis 👍 for first contribution.
> 
> The problem here is you are setting "End Date" as 15/10. Instead you should process salary for the entire month as you have monthly structure. And as the employee left on 15/10, you should set "relieving_date" on employee master. Then system will automatically reduce payment days to 15 days and working_days will still remain as 30.
> 
> Please let me know if I am correct.

@nabinhait the problem here is not about the relieving date but about the changing of ssa, suppose the emp A who has been promoted for a new ssa (may be from probation to fulltime or just increase salary) so that on the month of ssa updated , this emp A will have 2 salary slip for each SSA (old ssa and new ssa)

nabinhait:
@TaiLouis Sorry for the delayed response. For the use case, the solution seems correct. Can you please add a few tests and send the PR to staging-fixes branch?

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist

deepeshgarg007:
@TaiLouis can you provide an example for this use case.
In some cases salary component amount is fixed and does not depend on ```total_working_days_in_payroll_frequency``` or ```total_working_days```, have you considered that in you PR

sagarvora:
Closed due to inactivity.

PR #15731, opened 2018-10-18
[New Feature]Scan Item Barcode label to quick create/update records in items child table

Zlash65:
https://github.com/frappe/erpnext/pull/15329

Scan Barcode field added in
- Material Request
- Purchase Order
- Sales Order
- Delivery Note
- Stock Entry
- Purchase Invoice
- Sales Invoice

Code cleaned up.


erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -119,6 +109,23 @@ def get_items(start, page_length, price_list, item_group, search_value="", pos_p
 
 	return res
 
+@frappe.whitelist()
+def search_serial_or_batch_or_barcode_number(search_value):
+	# search serial no
+	serial_no_data = frappe.db.get_value('Serial No', search_value, ['name as serial_no', 'item_code'], as_dict=True)
+	if serial_no_data:
```
szufisher: Any reason why the searching sequence is serial number , then batch, finally barcode, not the other way around?
There is also barcode field in Item table, it is already obsolete and replaced by the Item barcode child table?

rmehta (changes requested):

erpnext/public/js/controllers/transaction.js
```diff
@@ -253,6 +253,63 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+							add_row_index = d.idx;
+						}
+					});
+
+					if(add_row_index == -1) {
+						child = frappe.model.add_child(this.frm.doc, cur_grid.doctype, "items", add_row_index);
+						show_description(child.idx, r.message.item_code, child.item_code);
+					} else {
+						child = cur_grid.get_grid_row(add_row_index-1).doc;
+						show_description(child.idx, r.message.item_code, child.item_code);
```
rmehta: this line is common, can be outside the `if` block !

rmehta (dismissed):

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -25,20 +22,13 @@ def get_items(start, page_length, price_list, item_group, search_value="", pos_p
...
-
-		if not serial_no and not batch_no:
-			barcode_data = frappe.db.get_value('Item Barcode', {'barcode': search_value}, ['parent', 'barcode'])
-			if barcode_data:
-				item_code, barcode = barcode_data
+	item_code = data.get("item_code") if data.get("item_code") else search_value
+	serial_no = data.get("serial_no") if data.get("serial_no") else ""
+	batch_no = data.get("batch_no") if data.get("batch_no") else ""
+	barcode = data.get("barcode") if data.get("barcode") else ""
+	frappe.errprint([item_code, serial_no, batch_no, barcode])
```
nabinhait: remove `errprint`

nabinhait (changes requested):

PawanMeh:
Some possible gaps, scan field not available on Purchase receipt and also does the scanned bar code get updated in inventory?

FinForceConsulting:
@Zlash65 Zarrar, this PR merged only created field to capture Barcode number at the Item level and related docs.

Could you share a gif Screencast of how this will work for multiple items and the item table gets created automatically with scan. If this works like retail stores, it would be a great functionality

PR #15736, opened 2018-10-18
Update 'cart count' cookie when shopping cart is updated

johnhkelley1:
Currently, when the website's shopping cart is updated by either adding an item or changing the quantity of an existing item, the 'cart count' cookie isn't updated until the page is refreshed. This results in an incorrect cart count in the shopping cart dropdown on the nav bar. 

For instance, if there is 1 item in the shopping cart, and you click 'Add to Cart' on an item page so that there are now 2 items in the shopping cart, the cart dropdown in the nav bar will still say there is 1 item in the shopping cart. 

This fix resets the 'cart count' cookie after the cart is updated so that the correct cart count is displayed in the nav bar without needing to reload the page.

### DIFF
--- a/erpnext/shopping_cart/cart.py
+++ b/erpnext/shopping_cart/cart.py
@@ -112,9 +112,11 @@ def update_cart(item_code, qty, with_items=False):
 	quotation.payment_schedule = []
 	if not empty_card:
 		quotation.save()
+		cart_count = cstr(len(quotation.get("items")))
 	else:
 		quotation.delete()
 		quotation = None
+		cart_count = 0
 
 	set_cart_count(quotation)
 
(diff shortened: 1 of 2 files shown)


erpnext/shopping_cart/cart.py
```diff
@@ -112,9 +112,11 @@ def update_cart(item_code, qty, with_items=False):
 	quotation.payment_schedule = []
 	if not empty_card:
 		quotation.save()
+		cart_count = cstr(len(quotation.get("items")))
```
Alchez: Is this supposed to be a string? Or did you mean to use `cint`?

Alchez: Is this supposed to be a string? Or did you mean to use `cint`?

johnhkelley1: Good catch, I think cint is better since I'm casting to string in the callback. I just made that change.

Alchez: Ah, no problem. Glad to help.

I had another note: although your way is more explicit, Javascript is a weakly typed language, so you wouldn't need to do the string conversion in order to do the concatenation.

johnhkelley1: True, good point. I'll make that change as well

PR #15738, opened 2018-10-18
Location default area in square meters

Tropicalrambler:
For consistency, location now shows default area UOM, which is square meters when drawn from the map. In future functionality, we will add interchangeability of areas. For now this permits clarity.
agmods-location-area-default

It also lists one location only for Crop Cycle.  Each crop cycle is one location only. This maintains ag procedures organized, and respects the nature of agricultural work: It is preferable to have many crop cycles start the same day in several fields and then be able to harvest each field when it is possible or ready, and keep track of work done on each field. This allows farmer to best estimate yields.  
agmods-crop-cycle-single


tooclean2320:
I don’t know how to close my account... i no long work with gethub... please close it for me

PR #15749, opened 2018-10-20
Add the total by location and the total by all locations in the tree view in square meters

monroy95:
Modification to tree view to show the area of each location and the total for all locations

location_tree_view_2


erpnext/assets/doctype/location/location.js
```diff
@@ -11,7 +11,14 @@ frappe.ui.form.on('Location', {
 			};
 		});
 	},
-
+	refresh: function (frm) {
```
Zlash65: Why is this needed?
`All Locations` is just a label node displayed on top and each group node under `All Locations` would not have parent node and will be regarded as `root_node` which is kinda wrong. There should only be one root node.

Zlash65: Why is this needed?
`All Locations` is just a label node displayed on top and each group node under `All Locations` would not have parent node and will be regarded as `root_node` which is kinda wrong. There should only be one root node.

Zlash65: Why is this needed?
`All Locations` is just a label node displayed on top and each group node under `All Locations` would not have parent node and will be regarded as `root_node` which is kinda wrong. There should only be one root node.

Zlash65: Why is this needed?
`All Locations` is just a label node displayed on top and each group node under `All Locations` would not have parent node and will be regarded as `root_node` which is kinda wrong. There should only be one root node.

erpnext/assets/doctype/location/location_tree.js
```diff
@@ -27,6 +27,31 @@ frappe.treeview_settings["Location"] = {
 			condition: 'frappe.boot.user.can_create.indexOf("Location") !== -1'
 		}
 	],
+	onrender: function (node) {
```
Zlash65: Instead of appending it right after the name, we can make it somewhat like how `Account tree` looks like where the area will be set at the right most side of the label?

Take a look at `account_tree.js` file to see how its done there.

erpnext/assets/doctype/location/utils_location.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2018, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+@frappe.whitelist()
+def get_total_location():
```
Zlash65: This would show wrong result if location filter is applied.
Also no need for new file, we can add this method in `location.py` itself.

erpnext/assets/doctype/location/location_tree.js
Zlash65: Instead of appending it right after the name, we can make it somewhat like how `Account tree` looks like where the area will be set at the right most side of the label?

Take a look at `account_tree.js` file to see how its done there.

Zlash65 (changes requested):

erpnext/assets/doctype/location/utils_location.py
```diff
@@ -0,0 +1,13 @@
...
+    # Get the total of all locations in square meters
+    total_location = frappe.db.sql('''SELECT sum(round(area, 3)) FROM `tabLocation` WHERE is_group = 0''')
```
Alchez: You can use the ORM to do this as well.

```python
total_location = frappe.db.get_value("Location", {"is_group": 0}, "sum(round(area, 3))")
```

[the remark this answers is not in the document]
monroy95: Great. Thank you.

erpnext/assets/doctype/location/location_tree.js
monroy95: @Zlash65 New view for location tree: 
location_tree_view

erpnext/assets/doctype/location/utils_location.py
monroy95: Suggested changes in: e72bd3c

erpnext/assets/doctype/location/location.js
Zlash65: Why is this needed?
All Locations is just a label node displayed on top and each group node under All Locations would not have parent node and will be regarded as root_node which is kinda wrong. Moreover, `parent_location` isn't a mandatory field either.

erpnext/assets/doctype/location/location.py
```diff
@@ -223,3 +223,10 @@ def add_node():
 
 def on_doctype_update():
 	frappe.db.add_index("Location", ["lft", "rgt"])
+
+@frappe.whitelist()
+def get_total_location():
+	# Get the total of all locations in square meters
+	total_location = frappe.db.get_value("Location", {"is_group": 0}, "sum(round(area, 3))")
```
Zlash65: Open the tree view for location and select a location in the filter above. Notice that the tree view will change and the selected filter will become the root of the structure but its area would be wrongly calculated.

Zlash65 (changes requested):

erpnext/assets/doctype/location/location.js
monroy95: I added it because I understood that each of the nodes groups of all locations are a representation of what they contain and there should be no need to edit the node group and thus not modify the total area per group. Do you suggest to remove it?

monroy95: I remain pending the suggestion to push the changes with the solution.

erpnext/assets/doctype/location/location.py
monroy95: I'm sorry, it's true. I solved it:

solved_tree_view_location

monroy95: The suggested changes have been applied to the following  5705af8

PR #15769, opened 2018-10-23
[Minor] Remove hard coded status

Zlash65:
- Allows flexibility to add more status type from Customize Form


rmehta:
@Zlash65 this is already done via `document.py`. This is hard-coded here since there is a business logic impact for each of these options and they should not be editable by the user.

Closing this for now.

PR #15808, opened 2018-10-25
Allow Rename in Customs Tariff Number

dottenbr:
This is a non-urgent change which allows renaming of Customs Tariff Numbers. Since those tariff numbers change regularly you don't only need to add new ones but might have to rename to re-classify your tariff numbers.


Zlash65:
Hi Dominik, there seems to be no changes made in your PR ( except for the change in modified date), and allow rename has been enabled for Custom Tariff Number doctype since 2 months.

PR #15811, opened 2018-10-25
Fix Purchase Order Issue in Update Items

dottenbr:
Exception handling for possible zero division.

Internal ESO Note: if merged delete fork hotfix https://github.com/frappe/erpnext/commit/ca6c7a21299951e6887ad78816edf69b09cff590

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -295,9 +295,9 @@ class PurchaseOrder(BuyingController):
 		for item in self.items:
 			received_qty += item.received_qty
 			total_qty += item.qty
-		try:
+		if total_qty:
 			self.db_set("per_received", flt(received_qty/total_qty) * 100, update_modified=False)
-		except:
+		else:
 			self.db_set("per_received", 0, update_modified=False)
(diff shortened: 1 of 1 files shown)


dottenbr:
@rohitwaghchaure are you able to review this?

erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -295,7 +295,10 @@ def update_receiving_percentage(self):
 		for item in self.items:
 			received_qty += item.received_qty
 			total_qty += item.qty
-		self.db_set("per_received", flt(received_qty/total_qty) * 100, update_modified=False)
+		try:
+			self.db_set("per_received", flt(received_qty/total_qty) * 100, update_modified=False)
+		except:
```
Alchez: This will catch all exceptions - even ones you didn't intend to. To avoid unexpected behavior, you should catch the `ZeroDivisionError` explicitly.

Alternatively, if you're only expecting problems for zero-quantity purchase orders, it might be more readable to replace the `try..except` with an `if..else` check on the `total_qty`.

Alchez (changes requested):

PR #15815, opened 2018-10-25
[fix] - GL entry not creating on PR if rejected qty mentioned

khushalti:

Khushal Trivedi
Indictrans Technologies


khushalti:
Purchase Receipt subssion with only accepted qty creates GL Entry correctly,but if we mention rejected qty on purchase receipt does not create GL entry for accepted qty as well

khushalti:
Should there be Test cases for this ?

rmehta:
@khushalti yes a test case would be good. But this looks like a serious bug. @nabinhait @rohitwaghchaure ?

Issue #15458, opened 2018-09-20
If taxes table is empty, discount amount isn't printed in purchases/sales


kardmode:
Actual behavior: Addional Discount amount is not displayed when the taxes table is empty in purchases/sales.

Cause: Discount display depends on the taxes table field being rendered. 

Expected behavior: Discount should show even if no taxes are set.

Steps to Reproduce: 
1. Create new purchase order with no taxes and set a discount amount. Print.
2. Add tax to purchase order. Print

![1](https://user-images.githubusercontent.com/8419064/45843272-280cd200-bced-11e8-8b75-692e4ca7d9d7.gif)

ERPNext: v10.1.37 (master)
Frappe Framework: v10.1.36 (master)

PR #15817, opened 2018-10-25
[fix] conditional discount_amount print_hide

patilsangram:
Discount Amount field is part of taxes.html. So, the discount amount is not visible on print if taxes are absent. So, I have added conditional print_hide for Discount Amount field. So, it will show on Print if taxes are not available.

Ref : [15458](https://github.com/frappe/erpnext/issues/15458)


patilsangram:
Reference PR - [15812](https://github.com/frappe/erpnext/pull/15812)

Issue #15814, opened 2018-10-25
[Bug V11] "Payment Terms Template" could not be fetched if sales order is created from Quotation


asanodaria:
Steps to reproduce:
1. Create Quotation with Payment Terms Template.
2. Go to Make button click on "Sales Order".
Result: All the fields are gets fetched from Quotation except the "Payment Terms Template". See video

![pt-from-quotation](https://user-images.githubusercontent.com/36321437/47483212-56cf1880-d856-11e8-911f-32fac268f2ec.gif)

Zlash65:
After hitting `Make Sales Order` the template does get copied. If you save the Sales Order, according to the date set, the child table will auto populate.

deepeshgarg007:
@asanodaria  Closing this issue. If the issue still persists in the latest version-13 or above then please reopen a new issue

PR #15776, opened 2018-10-24
fix: Minor bug fixes related to Payment Terms

Zlash65:
### https://github.com/frappe/erpnext/issues/15744
- Payment Terms not copied from Quotation to Sales Order

### https://github.com/frappe/erpnext/issues/15696
- Duplicating Sales Invoices doesn't re-calculate due date fix, payment terms will be set afterwards and triggered accordingly.
ptt

PR #15822, opened 2018-10-25
[FIX][#15814]['Payment Terms Template' will fetched on sales order]

jsukrut-indic:
Payment Terms Template will be fetched on sales order


shreyashah115:
@jsukrut I guess this is already fixed here - https://github.com/frappe/erpnext/pull/15776

jsukrut:
Fixed ..!!

PR #15825, opened 2018-10-25
fix(gle): Post gl entry for booking COGS from Sales Invoice even if grand_total is zero

### LINKED ISSUE #15459: COGS entries not passed by ERPNext if Sales Invoice has zero Grand total
I just discovered that if the total value of a sales invoice is zero, the Cost Of Goods Sold (COGS) entries are not passed by ERPNext.

This scenario can occur if a company gives 100% discount to a customer on a Sales Invoice. Though the total Sales value is zero, the sales invoice still has items that have cost and this cost should be debited to the cost of goods sold account once the goods are sold.

ERPNext does not pass this entry unless the Sales Invoice has a total value no matter how small. But if the total value is zero, COGS entry is not passed.

This is a bug as it negates all accounting rules. This is for Version 10. It needs to be corrected

Olamide

nabinhait:
Fixes #15459


rmehta:
What is the use case? Please explain!

PR #15827, opened 2018-10-25
First Run Fees Component Account heads patch

joshiparthin:
As discussed with @scmmishra and @Thunderbottom, a proposal for a facility to take different income accounts in the ledger for different fees component.

Let's say there are different fees components:

Tuition Fees,
Library Fees,
Sports Fees, etc.
Currently, all fees go in one account head.

A proposed PR for considering different account heads if account heads are mentioned in the Fees Component.

Please Review it in the totality of the accounts flow.

Also kindly validate the use case and test in your sample data.

Regards,

Parth Joshi


joshiparthin:
Ok Shreya. Thanks for the reply…

erpnext/education/doctype/fees/fees.py
```diff
@@ -107,6 +110,52 @@ def make_gl_entries(self):
 		from erpnext.accounts.general_ledger import make_gl_entries
 		make_gl_entries([student_gl_entries, fee_gl_entry], cancel=(self.docstatus == 2),
 			update_outstanding="Yes", merge_entries=False)
+	
+	def check_for_individual_heads(self):
+		linked_fs = frappe.get_doc("Fee Structure", self.fee_structure)
```
Alchez: You don't need to get the entire document to check the value. It'll be faster to just do:

```
return frappe.db.get_value("Fee Structure", self.fee_structure, "process_fees_category_account_heads")
```

Since the value will be `False` if the box is unchecked.

Alchez (changes requested):

joshiparthin:
Sure. Thanks for the tip. I have to send pr to staging-changes anyways. So will accommodate then.  

—
Sincerely, 

Parth Joshi.

rmehta:
Please add tests and update docs!

PR #15833, opened 2018-10-25
[Fix] Website conversion factor query

ghost:
Hi,

The uom conversion factor query returns multiple results because of a bad join.

Thanks


britlog:
Hi @shreyashah115 ,

Can I use `Hotfix `for V10 instead ? As it is a bugfix.
Or you definitely prefer `staging-fixes` ? :)

Thanks

shreyashah115:
@britlog If it's based on v10, please send it on hotfix.

PR #15835, opened 2018-10-25
Add Task Description to autofill Description

brownharryb:
Description field in Tasks can be auto-filled from Task Description Template
out


Zlash65:
Instead of rewriting the code for task_description, maybe you can add it in the utils.js file [ just like get_terms for invoices, etc ] and call it from project and task? Take a look at how terms and conditions are fetched for transactions.

brownharryb:
Hi @Zlash65, thanks for the comment.
Please check latest commit.
Getting description has been made a little more generic and anyone should be able to fetch description whether is tailored for jinja or not.

Zlash65:
PR seems to be good after changes. Can you fix the Codacy's red cross and add a test case to create a task description using some jinja in it and making a task where you can check that jinja is properly rendered.

brownharryb:
Code updated

rmehta (changes requested):
Rename "Task Description" to "Task Template"

PR #15836, opened 2018-10-26
Staging fixes

agritheory:
This changes all the doctypes in the Ag module that were not "restricted to domain" to be now "restricted to domain" plus a naming change to the Weather DocType that was connected to the date which is no longer allowed.

Second commit carries the same type of change to "unrestricted" doctypes in Education, Healthcare and Non Profit domains. 

Calling this a bug fix, because the industry-specific doctypes that are not restricted to their domains are annoying. To V11 staging-fixes. 

@shreyashah115 Is this the correct branch structure? I'm not sure I'm following your instructions correctly.


agritheory:
@shreyashah115 Please cancel this PR, I only want the two commits I made on here, this is too much to review.

shreyashah115:
Hi Tyler, kindly do a `git reset --hard upstream/staging-fixes` create a new branch, apply your changes and push it. You will have to open a new PR after these changes and set the base-branch as `staging-fixes`. Hope this clears your doubts!

agritheory:
@shreyashah115 I think I've messed up badly. I'm trying to pull from upstream and I see my commits there. It seems like  I may have committed to this branch directly? Is that even possible?
I plan to visit the office Monday if we need to clear this up any further. Once I get the hang of this I'll stop being such a pain, I promise.

shreyashah115:
@agritheory If you are still confused, I can open a PR for you as a co-author if that's okay with you?

PR #15837, opened 2018-10-26
Option for Packing List from Sales Order to Delivery Note/Sales Invoice

RicardoJohann:
When creating a Delivery Note or a Sales Invoice (with Update Stock ticked) from a Sales Order which has a Product Bundle that has been changed in the meanwhile, the system populates the Packing List with the lastest item Bundle list even though the Sales Order has a different packing list.
Add a check option "Maintain Product Bundle Items during Selling Cycle" in Selling Settings to let Administrator decide whether retrieve bundled items from Sales Order or Product Bundle

checkbox_bundle


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -577,6 +577,15 @@ def update_item(source, target, source_parent):
 		}
 	}, target_doc, set_missing_values)
 
+	if not cint(frappe.db.get_default('maintain_packed_items_list')):
+		if hasattr(target_doc, 'packed_items'):
```
rmehta: duplicate code, please add it in `transaction_base.py`

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -18,6 +18,12 @@ def get_product_bundle_items(item_code):
 		from `tabProduct Bundle Item` t1, `tabProduct Bundle` t2
 		where t2.new_item_code=%s and t1.parent = t2.name order by t1.idx""", item_code, as_dict=1)
 
+def get_cur_product_bundle_items(item_code, name):
```
rmehta: don't use abbreviations

rmehta (changes requested):

rmehta:
This should not be a setting. If the bundle has changed, it should become a new product code if there are existing orders that need to be shipped.

https://github.com/frappe/erpnext/wiki/Model-Design-Guidelines#make-contextual-settings

PR #15842, opened 2018-10-27
feat: Journal Entry Multiple Reference/Clearance No/Date and Cleaner GL Entry Remarks

SaiFi0102:
This pull requests include the following somewhat-related changes:
* **Journal Entry:** Multiple Reference Numbers, Reference Dates and Clearance Dates in child table to allow cases like payment into multiple banks in one Journal Entry
* **Payment Entry:** Added user remarks in child tables
* **SOs, DNs, POs, PRECs:** Added remarks field in More Information section
* **GL Entry:** Included `reference_no` and `reference_date` fields
* **GL Entry Remarks:** include only user remarks in GL Entries instead of computer generated remarks to make it more readable
* **General Ledger Report:** Added more filters, reordered columns according to their significance and added reference number and reference date columns
* **General Ledger Print Format:** enhanced print format

## Journal Entry
You can enter reference number and reference date in the parent form which will set it for each child table AND you can set it from child table form which will override what ever reference is set in the parent form.

image

image

So this kind of a print format becomes possible:
image

## General Ledger Report
Reordered Filters and Added Against Voucher and Reference Number in filters:
image

Reordered Columns:
image

Print Format:
The links are clickable so that the user can chose to view the general ledger in this format
image

Print Format with Against Column:
image

Print Format also shows reference document in the remarks column automatically:
image

## Patch
The patch simply copies the old fields (`cheque_no`, `cheque_date`, `clearance_date`) from Journal Entry to all it's Journal Entry Account child table rows. With the exception for `clearance_date` it is only copied for Bank and Cash type accounts.


SaiFi0102:
@Zlash65 added description and screenshots. What kind of test do you want? I haven't really introduced any new feature here.

PR #15844, opened 2018-10-28
Selling rate in Stock Ledger Entry

sivankar:
Added new field selling rate in stock ledger entry, since we have incoming and valuation rate in SLE it would be good to have selling rate as well, so all the reporting can be done from one table.


adnanmunir83:
it's a good idea to have this field.  will help to build complex reports quickly.

erpnext/controllers/stock_controller.py
```diff
@@ -250,6 +250,7 @@ def get_sl_entries(self, d, args):
 			"voucher_type": self.doctype,
 			"voucher_no": self.name,
 			"voucher_detail_no": d.name,
+			"selling_rate": d.get('base_rate', 0) if not self.doctype.startswith("Purchase") else 0,
```
nabinhait: It will be better to have explicit doctypes from where we can get selling rate.
In this case, `self.doctype in ('Delivery Note', 'Sales Invoice')`

nabinhait (changes requested):

erpnext/controllers/stock_controller.py
sivankar: change request is done

sivankar:
Sales Invoice.
sales invoice rate

Stock Ledger Entry.
stock ledger si

erpnext/controllers/selling_controller.py
```diff
@@ -191,6 +191,7 @@ def get_item_list(self):
 				frappe.throw(_("Row {0}: Qty is mandatory").format(d.idx))
 
 			if self.has_product_bundle(d.item_code):
+				qty = frappe.db.sql('Select sum(pbi.qty) as qty from `tabProduct Bundle Item` pbi where pbi.parent = %(pb)s ', {'pb': d.item_code}, as_dict=1)[0].qty
```
Alchez: You can use `db.get_value` instead.

`qty = frappe.db.get_value("Product Bundle Item", filters={"parent": d.item_code}, fieldname="sum(qty)")`

Alchez (changes requested):

sivankar:
@nabinhait when ever you get time can you please review this code

PR #15845, opened 2018-10-28
[Fix] Conversion factor in offline POS

ghost:
Hi,

In POS, the item uom is the default selling uom, but the conversion factor is not applied, so there is pricing problems.

For instance, you sell eggs with stock uom in unit and selling uom in box of 6.
Adding the eggs in the POS will fetch the price for 1 unit but in the invoice, the uom will be the box of 6.

It works already in online POS, this fix only applied to offline POS.

Thanks.


olamide2:
Hello @britlog 

I have been scratching my head about UOM in PoS and am happy someone is doing something about it.

I am thinking we should make it possible to be able to pick different UOM at point of sale. For example if we call up eggs at PoS, then both the unit and the box should come up. The user can then choose which option she wants to sell at that point.

Is this what your fix does?

Regards

britlog:
Hi @olamide2, 

It fixes the price by applying conversion factor.
I add the field uom and use it while invoices are created, so stock uom and sales uom can be managed but I have not added user interface to select the uom. For now, the uom will be the sales uom if exists, else stock uom. 
I think it is better to let like that by default (to avoid a popup each time you add an item with multiple uom) and add a button to change the default uom and price accordingly. If you want to manage all uoms for the items, you need to do more changes because only the conversion factor between stock uom and sales uom is present. But I think these 2 uoms will be enough in the POS. 

Regards

britlog:
I add the field but it is set by default to sales uom, or stock uom if sales uom is not defined :
`this.child.uom = this.items[0].sales_uom || this.items[0].stock_uom;`

There is no user interface to change it dynamically from the POS.
You may develop it or ask any service provider to do it.
I can analyse and quote this dev, contact me on my website if you want so.

Regards

olamide2:
Email sent

nabinhait:
@britlog Can you add a screenshots / GIF to show how it is working.

olamide2:
Hi @britlog 
Why is this closed ??

britlog:
Hi Olamide,

I've just reopened it on staging-fixes branch for V11 as develop seems to
be for V12.
The new PR :
https://github.com/frappe/erpnext/pull/15867

Regards,

Frédéric NIHOUARN
09 72 61 29 43
06 17 44 22 36
www.britlog.com

Issue #15846, opened 2018-10-28
[bug V11] Cannot delete department


Yemeni:
**Description:**
When I try to delete  department I get an error.

**Steps to reproduce:**
1. make new erpnext installation
2. complete the setup
3.  go to
https://site.example:8000/desk#List/Department/List
or https://site.example:80/desk#List/Department/List
4. create a department
5. delete the department 

_the bug happens (also with an Administrator account)_

**tried in version:**
ERPNext: v11.0.3-beta.10 () (staging)
ERPNext: v11.0.3-beta.12 () (staging)
ERPNext: v11.0.3-beta.13 () (staging)
ERPNext: v11.0.3-beta.15 () (staging)

on my Linode Ubuntu 16 LTS server
and on https://demo.erpnext.com

Related article 
https://discuss.erpnext.com/t/error-cannot-delete-department

**error log:** 
`Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/app.py", line 61, in application
    response = frappe.handler.handle()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 56, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/__init__.py", line 1007, in call
    return fn(*args, **newargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/client.py", line 237, in delete
    frappe.delete_doc(doctype, name, ignore_missing=False)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/__init__.py", line 752, in delete_doc
    ignore_permissions, flags, ignore_on_trash, ignore_missing)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/delete_doc.py", line 78, in delete_doc
    doc.run_method("on_trash")
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 772, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1048, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py", line 1031, in runner
[...]

PR #15854, opened 2018-10-29
[ Minor] Fix Events deletion

Zlash65:
Before - 
issue-event

After -
fix-event


Yemeni:
Is it a fix for this?
https://github.com/frappe/erpnext/issues/15846

Zlash65:
@Yemeni Wasn't able to replicate the issue. But the fix that i sent could have caused trouble nonetheless if left unchecked.

rmehta:
Should this not be in the framework?

nabinhait:
@Zlash65 Yes, move to the frappe

sagarvora:
Closing, to be moved.

PR #15865, opened 2018-10-30
feat: Enhanced UOM Conversions Table

SaiFi0102:
## Problem:
For the example of a paint mixing business:
* Stock UOM is in Cans
* BOM is entered in Grams
* Purchase is done in Cans
* Sale is done in either Cans or Packs
* Analysis is done in Litre

The problems are:
- We can not enter how many Litres a Can has directly, we have to enter the conversion factor as 1/size_of_can.
- We can't enter the density of the paint: how many Grams are in a Litre (has an indirect conversion to the stock UOM which is Can), we have to do more calculation for that
- We can only enter how many Cans are in a Pack

Another problem:
* If a Can has 3 Litres in it, the conversion_factor becomes 0.333
* If we put 3 Litres in any transaction, ERPNext would calculate the Stock Qty as 0.999, even though it should be 1.

My solution is to increase the conversion_factor's precision.

## Solution
The new UOM Conversion Graph child table acts as an _adjacency list_ for a **graph** with the set of UOMs as vertices and conversion_factors between UOMs as edges. To get the conversion_factors from any UOM to stock UOM we can run a Depth First Search to find all conversion paths for that UOM to the Stock UOM and the product of the weights will be the `conversion_factor`. 

The only change to focus on is in `item.py` the rest is mostly modification to conversion_factor fields precision to 9 to prevent the small fractional errors in Qty and to reflect the precision in `UOM Conversion Factor` DocType. Or they are updates to the test cases to manipulate the new `UOM Conversion Graph` child table instead of the `UOM Conversion Detail` child table.

image

## Patch
Since there are multiple ways to input the new UOMs table, I have created a patch that suits my problem best that creates the new UOMs table from the old UOMs table:
**From Qty:** 1
**From UOM:** doc.stock_uom
**To Qty:** 1/row.conversion_factor
**To UOM:** row.uom
It think it may be better not to take the inverse of the conversion factor, however, that decision is up to the maintainers.

## Future possibility
UOM Conversion Factors can be cached and a graph of UOMs conversions can be made in the cache as well. This will improve performance of calculating conversion factors. And this will also allow conversions like Can -> Milliliter or Can -> Fluid Ounce or any conversion that is somehow linked in the graph. (Given the item has a conversion factor for Can -> Litre EVEN IF there is no conversion factor for Litre -> Milliliter or Litre -> Fluid Ounce)


SaiFi0102:
The test `test_multiple_uom_in_selling` in `Sales Invoice` is expecting a conversion_factor of 10.0 without ever setting it explicitly anywhere. Is there a problem with this test case?

SaiFi0102:
Waiting for response on test's error

sagarvora:
Travis failing.

PR #15866, opened 2018-10-30
[Fix] Website Conversion factor query

ghost:
Hi,

The uom conversion factor query returns multiple results because of a bad join.
Same as #15833, I open a new PR to change branch to `staging-fixes`.

Thanks


britlog:
To keep this PR alive

PR #15867, opened 2018-10-30
[Fix] Conversion factor in offline POS

ghost:
Hi,

Same as #15845 but on `staging-fixes`.
In POS, the item uom is the default selling uom, but the conversion factor is not applied, so there is pricing problems in the invoice.

For instance, you sell eggs with stock uom in unit and selling uom is box of 6.
Adding the eggs in the POS will fetch the price for 1 unit but in the invoice, the uom will be the box of 6.

UI is not changed, but you could see on the gif that price is 0.40€ for 1 egg : the box of 6 added in cart is sold 2.40€. Before this PR it was 0.40€ for the box in the cart, it works already in online POS, this fix only applied to offline POS.

pos-uom

Thanks.


olamide2:
Hello @britlog 

You and I spoke offline, but am still confused.

Why is this a bug? what if the cashier actually is trying to sell 1 egg for 0.40 as shown in the item image page ?

I think this pull request just replaces one problem with another without solving the underlying problem

britlog:
Hi @olamide2 

Because for now, POS doesn't manage multiple UOM but always sell in sales UOM (or stock UOM if no sales UOM is defined). So conversion factor need to be applied.

Without this fix, the price in the POS is for stock UOM (= 1 egg) but the invoice will be created in sales UOM (= box), so the final invoice will be Qty 1, Unit Box and Price 0.40€ : bug.

With this fix, offline POS will work like online POS in sales UOM and prices are good.
The next step will be to add the possibility to select different UOMs on POS.

Regards

olamide2:
Great !!!

britlog:
Till alive !

PR #15872, opened 2018-10-31
fix: Mandatory offline name

tundebabzy:

This is a repackaged PR based on #15699.

It doesn't address the reason for the duplication experienced when the POS is under heavy use. However, it  strengthens the POS' design by taking the responsibility of ensuring there's only one Sales Invoice per offline_pos_name from the application layer to the database layer which is more resilient to the race condition bug.

PS: While I haven't been able to reproduce this, I have gone through databases with loads of such duplicates. I have also analysed logs confirming the POS sending the same data more than 4 times to ERPNext for syncing. This should be treated as critical

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
@@ -314,7 +314,7 @@
    "in_standard_filter": 0,
    "label": "Offline POS Name",
    "length": 0,
-   "no_copy": 0,
+   "no_copy": 1,
    "permlevel": 0,
    "precision": "",
    "print_hide": 1,
@@ -325,7 +325,7 @@
    "reqd": 0,
    "search_index": 0,
    "set_only_once": 0,
-   "unique": 0
+   "unique": 1
   },
   {
    "allow_bulk_edit": 0,
@@ -4773,7 +4773,7 @@
  "istable": 0,
  "max_attachments": 0,
  "menu_index": 0,
- "modified": "2018-07-17 13:46:52.449142",
+ "modified": "2018-10-17 18:28:52.449142",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Sales Invoice",
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -507,3 +507,4 @@ erpnext.patches.v10_0.set_discount_amount
 erpnext.patches.v10_0.recalculate_gross_margin_for_project
 erpnext.patches.v10_0.delete_hub_documents
 erpnext.patches.v10_0.update_user_image_in_employee
+erpnext.patches.v10_0.deduplicate_offline_pos
--- a/erpnext/patches/v10_0/deduplicate_offline_pos.py
+++ b/erpnext/patches/v10_0/deduplicate_offline_pos.py
@@ -1,0 +1,26 @@
+import frappe
+
+
+def execute():
+    duplicates = frappe.db.sql(
+        'select name, offline_pos_name from `tabSales Invoice` where '
+        'offline_pos_name is not null group by '
+        'offline_pos_name having count(*) > 1'
+    )
+
+    for duplicate in duplicates:
+        keep, offline_pos_name = duplicate
+        duplicate_list = frappe.db.sql(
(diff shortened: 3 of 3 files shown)


Olawale1:
I hope there's a plan to merge this soon. Anything that can improve the current state of the POS is welcome

Thanks

rohitwaghchaure:
Hi @tundebabzy 

I am still confused about the patch, if two invoices has the same offline name value then one is valid and second should be removed but in your patch you're updating the offline name of the second then how it could be the duplicate because both have different sales invoice number

rohitwaghchaure:
@tundebabzy 
Removed the patch and just keep the JSON changes, we'll merge and check whether it will fix the duplicate invoice creation issue

tundebabzy:
Hi
There are two reasons I chose to add the patch.
1. There's no way to add a unique constraint if all the values in the
column are not unique.
2. Some users might not fancy data loss even if its wrong.

rohitwaghchaure:
Hi @tundebabzy 

See my comments inline

1) There's no way to add a unique constraint if all the values in the
column are not unique.
**What about the sales invoices which are not pos invoices and has the offline_pos_name as null, I think unique constraint is not proper solution for this. We have to find out the exact cause of the issue, I have tried many times to replicate this issue but not able to replicate it**

2) Some users might not fancy data loss even if its wrong.
Either it should be correct or wrong, I think better we will find out the main cause of the issue and solved it.

I am closing this PR, @Olawale1 how many sales invoices have you create simultaneously using the POS and how many simultaneous users works on the POS.

tundebabzy:
Hi @rohitwaghchaure 
> Either it should be correct or wrong, ...

That is true but all those multiple invoices have reached the GL, P&L reports... Deleting the duplicate invoices means that someone may come back to look at a previous financial report and discover that figures have changed. And because ERPNext deletes GL entries, it will be untraceable. We should never delete user data without their consent.

> What about the sales invoices which are not pos invoices and has the offline_pos_name as null, I think unique constraint is not proper solution for this. ...

Apart from  SQL Server versions prior to SQL Server 2008 (which ERPNext does not support and probably never will), there's nothing wrong with having a unique constraint on a nullable columns. Under ANSI standards, a unique constraint *_must_* disallow duplicate _non-null values_ but _allow multiple null values_.

I think you should look at it from this angle and for now ignore what is possible from the code. Should we allow 2 different sales invoices to have the same `offline_pos_name`? (Me says No). Should `offline_pos_name` be unique? (I think the answer is Yes).

Now let's look at code, in the `make_invoice` function in `pos.py`:
https://github.com/frappe/erpnext/blob/e3136c38e1c1ee10211a7606da4e15f97336347b/erpnext/accounts/doctype/sales_invoice/pos.py#L318 the code is trying to ensure that an invoice with a given offline pos name does not exist. So even if you say you do not agree that the column should be unique, your code says it should. You may then say 'the code is prevent duplicates from being stored'. I totally agree but it's safer and less error prone to allow the database do the gate keeping not the code.

> We have to find out the exact cause of the issue, I have tried many times to replicate this issue but not able to replicate it

> I think better we will find out the main cause of the issue and solve it.

I totally agree. I was not able to reproduce it locally as well neither was I able to find the root cause but I stopped trying when it occurred to me gate keeping from the application layer is not as effective as gate keeping from the database layer. We have not experienced a single duplicate invoice after adding the unique constraint.

In summary, please reconsider this patch. The summary of what it is trying to achieve is this:
- Move unique checks from the application layer to the database layer
- To achieve that, ensure that all the data in the offline_pos_name column are unique
- Add the unique constraint

With my argument, I'm hoping you will reconsider this PR. I'll reopen it one last time and I will respect your opinion if you feel it doesn't need a second look.

Olawale1:
Hi @rohitwaghchaure

Trust you're doing well. I'm not sure about simultaneous invoices but simultaneous POS users is about 10

While I understand the concerns you raised about this PR, I think it will help to more easily identify where the issue is coming from. Like you, we would all rather identify the actual cause of the issue but this at least helps to improve things in the meantime 

Thanks

rmehta:
@rohitwaghchaure will this work?

I think raising an exception will help us trace this issue.

rohitwaghchaure:
@tundebabzy Can you rebase?

PR #15879, opened 2018-10-31
Add Serial Number button in Delivery Note and Sales Invoice's Packing List

RicardoJohann:
serial_no_button

### DIFF
--- a/erpnext/selling/doctype/quotation/quotation.js
+++ b/erpnext/selling/doctype/quotation/quotation.js
@@ -206,7 +206,7 @@ cur_frm.cscript['Declare Order Lost'] = function(){
 
 }
 
-frappe.ui.form.on("Quotation Item", "items_on_form_rendered", function(frm, cdt, cdn) {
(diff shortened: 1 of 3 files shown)


erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -206,7 +206,7 @@ cur_frm.cscript['Declare Order Lost'] = function(){
 
 }
 
-frappe.ui.form.on("Quotation Item", "items_on_form_rendered", function(frm, cdt, cdn) {
+frappe.ui.form.on("Quotation Item", "items_on_form_rendered", "packed_items_on_form_rendered", function(frm, cdt, cdn) {
```
rmehta: I don't think this works, please show an animated screenshot that this works.

rmehta (changes requested):

erpnext/selling/doctype/quotation/quotation.js
RicardoJohann: **Before had no serial button:**


noserialbutton


After coding:


gif

PR #15887, opened 2018-11-01
[Fix] Asset date validation

rohitwaghchaure:
User not able to put in use date for back dated entries.
For example user has created purchase receipt with date as 20-10-2018 and after 10 days he is submitting the asset in which he has added the in use date as 21-10-2018. But because of validation use not able to submit the asset because current date is greater than in use date

After fix:
Now we are checking in use date with the date of purchase receipt or purchase invoice(update stock enabled) and not checking with current date


nabinhait:
@rohitwaghchaure Please describe the problem and add a screenshot.
https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#explanation

rohitwaghchaure:
@nabinhait Added description

PR #15899, opened 2018-11-04
Modification to tree view to show the area of each location and the total for all locations

monroy95:
**Displaying default unit of measure in square meters:**

uom default

**Showing total square meters by location and total by all locations:**

tree view location


erpnext/assets/doctype/location/location_tree.js
```diff
@@ -27,6 +27,44 @@ frappe.treeview_settings["Location"] = {
...
 		}
 	],
+	onrender: function (node) {
+
+		if (!node.is_root) {
+			frappe.db.get_value("Location", node.data.value, "area")
+				.then((r) => {
+					// Place the location area on the right side, tree accounts style.
+					$('<span class="balance-area pull-right text-muted small">'
+						+ (`${((flt(r.message.area)).toLocaleString('en') + ' Square Meters')}`)
```
rmehta: Strings must be translatable

```js
_('{0} Sq m', flt(r.message.area).toLocaleString('en'));
```

(us abbreviation for narrow views)

rmehta (changes requested):

erpnext/assets/doctype/location/location_tree.js
monroy95: Understood, I've applied the changes to show **Sq m** in 175838e

erpnext/assets/doctype/location/location_tree.js
```diff
@@ -27,6 +27,44 @@ frappe.treeview_settings["Location"] = {
...
 		}
 	],
+	onrender: function (node) {
+
+		if (!node.is_root) {
+			frappe.db.get_value("Location", node.data.value, "area")
+				.then((r) => {
+					// Place the location area on the right side, tree accounts style.
+					$('<span class="balance-area pull-right text-muted small">'
+						+ (`${((flt(r.message.area)).toLocaleString('en') + ' Sq m')}`)
```
nabinhait: Why hard-coded "Sq m"? As we already have a dedicated field "Area UOM".

nabinhait (changes requested):

erpnext/assets/doctype/location/location_tree.js
Tropicalrambler: Was an oversight, we are aware it needs to be translatable and will make those changes!

Tropicalrambler: We have made the changes expect a PR today.

PR #15926, opened 2018-11-07
Update sales_invoice.py : Allow other currency in invoice then ordered

rasos:
Fixes use cases as described here https://discuss.erpnext.com/t/use-different-sales-invoice-currency-than-sales-order-currency/33473/4

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -451,7 +451,7 @@ class SalesInvoice(SellingController):
 		super(SalesInvoice, self).validate_with_previous_doc({
 			"Sales Order": {
 				"ref_dn_field": "sales_order",
-				"compare_fields": [["customer", "="], ["company", "="], ["project", "="], ["currency", "="]]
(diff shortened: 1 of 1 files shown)


SaiFi0102:
@rasos I believe you will also have to modify the logic that updates the previous document status since Sales Order and Delivery Note stores `Invoiced Amount` (rather than Invoiced Qty) which would lead to an incorrect `Invoiced Amount` and `% Billed`

rasos:
@SaiFi0102 our use case is a manual change of the currency in the invoice only for some customers, who decide to pay with a local LETS currency, with CHF or with FairCoins. Sales Order or Delivery Note can remain in the currency of the order - which is even preferred for controlling purposes.

nabinhait:
@SaiFi0102 is correct. As we update "Billed Amount" in Sales Order based on the invoice amount in invoice currency, it will update incorrect value in SO. Also there can be other similar issues like that.

In my opinion, if you want to raise invoice in other currency, then don't link the invoice with Sales Order or amend the Sales Order to change the currency. It is really not a good idea to create an invoice in different currency.

Closing the PR.

PR #15973, opened 2018-11-13
Accounts receivable report based on payment terms

deepeshgarg007:
screenshot 2018-11-13 at 3 04 02 pm
screenshot 2018-11-13 at 3 04 30 pm


erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -57,6 +57,21 @@ def get_columns(self, party_naming_by, args):
 
 		credit_or_debit_note = "Credit Note" if args.get("party_type") == "Customer" else "Debit Note"
 
+		if self.filters.based_on_payment_terms:
+			columns.append({
```
Alchez: Use `columns.extend()` instead of two `append` calls.

nabinhait (approved):

PR #16003, opened 2018-11-16
Validations for multiple trip records for the same driver or Delivery Note

Alchez:
Changes:
- Whenever a user enters the driver, the system checks for existing trips today for that driver, and if one is found, prompts the user if they want to modify the old one, or continue with the new one.
- Whenever a Delivery Trip is created with stops, and if some/all of the Delivery Notes have already been added in another trip, the system prompts the user asking if they want to continue with the added Delivery Note.

<details>
<summary>Screenshots / GIFs</summary>

Creating a Delivery Trip for a driver with an existing trip on the same day.
- Clicking on **Yes** will take the user to the existing Delivery Trip.
- Clicking on **No** will let the user continue as normal.

existing-driver

<hr>

Creating a Delivery Trip with Delivery Notes that are already present in other trips.
- Clicking on **Yes** will let the user continue as normal.
- Clicking on **No** will take the user back to the list view.

existing-dn

</details>

### DIFF
--- a/erpnext/stock/doctype/delivery_trip/delivery_trip.js
+++ b/erpnext/stock/doctype/delivery_trip/delivery_trip.js
@@ -85,11 +85,11 @@ frappe.ui.form.on('Delivery Trip', {
 							}
 							existing_trips += "</ul>"
 
-							let confirm_message = `The following deliveries already exist in recent
-								trip(s): ${existing_trips} Do you still want to continue?`;
+							let confirm_message = __(`The following deliveries already exist in recent
+								trip(s): ${existing_trips} Do you still want to continue?`);
 
 							frappe.confirm(
-								__(confirm_message),
+								confirm_message,
 								() => { resolve(); },  // If "Yes" is selected
 								() => { frappe.set_route("List", frm.doc.doctype); }  // If "No" is selected
 							);
@@ -109,10 +109,10 @@ frappe.ui.form.on('Delivery Trip', {
 				departure_time: [">", frappe.datetime.nowdate()]
 			}, "name", (r) => {
 				if (r) {
-					let confirm_message = `${frm.doc.driver_name} has already been assigned a Delivery Trip
-											today (${r.name}). Do you want to modify that instead?`;
+					let confirm_message = __(`${frm.doc.driver_name} has already been assigned a Delivery Trip
+						today (${r.name}). Do you want to modify that instead?`);
 
-					frappe.confirm(__(confirm_message), function () {
+					frappe.confirm(confirm_message, function () {
 						frappe.set_route("Form", "Delivery Trip", r.name);
 					});
 				};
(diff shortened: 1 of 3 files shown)


erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -345,6 +347,22 @@ def get_directions(route, optimize):
...
 
+@frappe.whitelist()
+def validate_unique_delivery_notes(delivery_stops):
+	delivery_stops = json.loads(delivery_stops)
+	delivery_notes = [stop.get("delivery_note") for stop in delivery_stops if stop.get("delivery_note")]
+
+	if not delivery_notes:
+		return {}
+
+	existing_trips = frappe.get_all("Delivery Stop",
```
rmehta: indentation

erpnext/stock/doctype/delivery_trip/delivery_trip.js
```diff
@@ -66,6 +66,60 @@ frappe.ui.form.on('Delivery Trip', {
...
+
+	driver: function (frm) {
+		if (frm.doc.driver) {
+			frappe.db.get_value("Delivery Trip", {
+				docstatus: ["<", 2],
+				driver: frm.doc.driver,
+				departure_time: [">", frappe.datetime.nowdate()]
+			}, "name", (r) => {
+				if (r) {
+					let confirm_message = `${frm.doc.driver_name} has already been assigned a Delivery Trip
```
rmehta: translations

rmehta (changes requested):

erpnext/stock/doctype/delivery_trip/delivery_trip.js
Alchez: Already done on line 115.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
Alchez: Fixed.

erpnext/stock/doctype/delivery_trip/delivery_trip.js
Alchez: Moved translations to string generation.

erpnext/stock/doctype/delivery_trip/delivery_trip.js
```diff
@@ -66,6 +66,60 @@ frappe.ui.form.on('Delivery Trip', {
...
+					callback: (r) => {
+						if (!$.isEmptyObject(r.message)) {
+							// create an unordered list of existing note <-> trip mappings
+							let existing_trips = "<ul>";
+							for (let trip in r.message) {
+								existing_trips += `<li><strong>${trip}</strong> exists in <strong>${r.message[trip]}</strong></li>`
+							}
+							existing_trips += "</ul>"
+
+							let confirm_message = __(`The following deliveries already exist in recent
```
rmehta: Translations won't work unless you use `__("...")` or `__('...')`

rmehta (changes requested):

rmehta (dismissed):
please fix message strings

Alchez:
Fixed in both places.

PR #16006, opened 2018-11-16
fix/feat: Allow extension of barcode types (without validation)

gbm001:
Currently, it is difficult to add new custom barcode types for two reasons, both of which relate to validate_barcode in item.py:
- There is a bug where barcode types with a space in, such as Code 128, are split in two (so barcode_type is checked against 'Code' and '128' rather than 'Code 128'). This is fixed by splitting the Options field against a newline, instead of spaces.
- All barcodes are validated against the stdnum.ean library. This only handles EAN-8, EAN-13 and UPC-12 barcodes and any other barcode will fail. Barcodes with no type will continue to not be checked. Barcodes with the default barcode_types of EAN, UPC will continue to be checked. The non-default barcode_types of EAN-13 and EAN-8 will also be checked. The barcode_type is cast to upper case before this check is made so ean, upc, ean-13 and ean-8 will also be validated.

This allows people to add their own barcode types, such as Code 128 and QR codes. Users can add custom validation of these barcodes using the usual hooks, but they cannot remove the standard validation.


gbm001:
Added some unit testing for Item Barcodes, even though they don't test what I have added, because why not.

nabinhait (approved):

PR #16016, opened 2018-11-19
[Fix] Tax Withholding, TDS is applying on the tax

rohitwaghchaure:
In cumulative case, system adding TDS on the previous invoices with including tax, it should only consider the previous invoices amount without tax

**Issue**
screen shot 2018-11-19 at 1 16 40 pm

**After fix**
image

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -830,6 +830,10 @@ class PurchaseInvoice(BuyingController):
 			return
 
 		tax_withholding_details = get_party_tax_withholding_details(self)
+
+		if not tax_withholding_details:
+			return
+
 		accounts = []
 		for d in self.taxes:
 			if d.account_head == tax_withholding_details.get("account_head"):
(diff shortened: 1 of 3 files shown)


PawanMeh:
In general the logic should be, if there is TDS applied on any past invoice, it should be applied on future invoices irrespective of thresholds. Thresholds should be applied only if no past invoice exists with TDS applied.

rohitwaghchaure:
## Cumulative
Single          Cumulative   Tax Rate
0          		30,000       10%

1st INV   10,000
2nd INV   10,000
3rd INV   10,000 -> Add 10% TDS on 30,000 amount
4th INV   5,000  -> Add 10% TDS on 5,000 amount

## Single
Single          Cumulative   Tax Rate
20,000          0                  10%

1st INV   20,000 -> Add 10% TDS on 20,000 amount
2nd INV   10,000 -> Add 10% TDS on 10,000 amount

## Single
Single          Cumulative       Tax Rate
20,000          30,000             10%

1st INV   10,000 -> 
2nd INV   10,000 -> Add 10% TDS on 20,000 amount

PawanMeh:
@rohitwaghchaure , tried testing this fix but I am unable to see witholding information at supplier level after the latest update. Please see screenshot attached

unable_to_see_witholding_category

rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/pull/16184/files

PawanMeh:
@rohitwaghchaure , I tried the below scenario & it seems to not be working properly in the below case

1. Create Supplier and apply TDS rule with cumulative threshold of 30,000.
2. Create one invoice of 20,000 - no TDS applied
3. Create second invoice of 10,000 - correct TDS applied @2%
4. Change the amount in the 2nd invoice to 5,000, the TDS applied does not change.

Attached is the video

tds_recalc

rohitwaghchaure:
@PawanMeh Fixed

PawanMeh:
@rohitwaghchaure , ran a few tests, looks good now

PR #16018, opened 2018-11-19
feat: Shortcut Popover

surajshetty3416:
shortcut popover

- Shortcut popover for quick navigation
- Ability to disable the feature
- Added **Shortcut Setting** and **Shortcut Setting Item** doctypes

This requires https://github.com/frappe/frappe/pull/6485


rmehta:
Can we have a better way to add shortcuts?

rmehta:
Also, move this to Frappe!

Issue #15936, opened 2018-11-08
Serial No : Validation in Stock Entry


kalaisaranA:
Need  serial number validation in stock entry

kennethsequeira:
@kalaisaranA please elaborate on what kind of validation you require and the use case.

nxdevelopment:
Stock Entry allows same Serial Number items (duplication of serial no). Even though the product is sold (contain Delivery Note),again the another product with same serial number is given it generate the new product with previous serial_no.In this Serial No list the creation document no is rewritten. So it allows repeation/duplication of serial number. It is one of issue. Detailed description is given below
Initally on work order:
![screenshot_2018-11-19 pdt - mfg-wo-2018-00014](https://user-images.githubusercontent.com/44566832/48706346-fa2b0780-ec21-11e8-99de-d6666e29fc0d.png)
2.when Product started to Manufacturing .Serial no created for 1st product-
![screenshot_2018-11-19 manufacture - mat-ste-2018-00049 1](https://user-images.githubusercontent.com/44566832/48706468-64dc4300-ec22-11e8-8f18-ea7022dc0b84.png)
3.The deliverynote  and document creation  in Serial no A55-

![screenshot_2018-11-19 nxweb - mat-dn-2018-00007 1](https://user-images.githubusercontent.com/44566832/48707786-8d663c00-ec26-11e8-8e2c-0a1a43fa2dce.png)

![screenshot_2018-11-19 a555](https://user-images.githubusercontent.com/44566832/48706592-b8e72780-ec22-11e8-8e4c-91f12e5aabab.png)
4.Again the another product started to manufactured with same serial number -
![screenshot_2018-11-19 manufacture - mat-ste-2018-00050](https://user-images.githubusercontent.com/44566832/48706772-43c82200-ec23-11e8-8dcf-56abff72cadc.png)
5.The Generation of same Serial No in list -
![screenshot_2018-11-19 a555 1](https://user-images.githubusercontent.com/44566832/48706838-73772a00-ec23-11e8-8c41-c05da6d444c0.png)

How it is possible same serial Number for 2products??????

These are detailed explanation of serial number duplication .Kindly refer the screenshot. 
Requirement : when same serial number is repeated in stock entry it must throw msg as " Already Referenced in Document " and not allow to rewrite in document.

Modified file :https://github.com/frappe/erpnext/pull/16020/commits/b5170a46381c8d8b27567b61e3d2104b9455e801

After modification  :  RESULT 
[...]

PR #16020, opened 2018-11-19
Serial Number validation in stock entry

nxdevelopment:
Stock Entry allows  same Serial Number items (duplication of serial no). Even though the product is sold (contain Delivery Note),again the  another product with same serial number is given it generate the new product with previous serial_no.In this Serial No list the creation document no is rewritten. So it allows repeation/duplication of serial number.In order to avoid the duplication the file is modified.

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -72,6 +72,13 @@ class StockEntry(StockController):
 		self.set_incoming_rate()
 		self.set_actual_qty()
 		self.calculate_rate_and_amount(update_finished_item_rate=False)
+		self.validate_serial_numbers()
+	def validate_serial_numbers(self):
+		"""
+			validate serial number stock
+		"""
+		
+		self.validate_serial_against_stock_entry()
 
 	def on_submit(self):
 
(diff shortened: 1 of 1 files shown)


nxdevelopment:
kindly refer this in Issues  "Serial No : Validation in Stock Entry #15936 " where detailed description for Serial Number validation in stock entry is explained.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -286,6 +293,18 @@ def validate_work_order(self):
 				self.check_duplicate_entry_for_work_order()
 		elif self.purpose != "Material Transfer":
 			self.work_order = None
+	def validate_serial_against_stock_entry(self):
```
rmehta: please add test case

rmehta (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -72,6 +72,13 @@ def validate(self):
 		self.set_incoming_rate()
 		self.set_actual_qty()
 		self.calculate_rate_and_amount(update_finished_item_rate=False)
+		self.validate_serial_numbers()
+	def validate_serial_numbers(self):
+		"""
```
rmehta: such a docstring is meaningless because it is the function name

erpnext/stock/doctype/stock_entry/stock_entry.py
rmehta: there must be an empty line above the function

rmehta (changes requested):

sagarvora:
Reopen after changes.

Olawale1:
Any update on this? Hope it's been fixed now

Issue #14211, opened 2018-05-24
Problem with delivery note 


alwsiela23:
I have problem with delivery note ,after I make delivery note against sales invoice , I can make another delivery note with quantity equal 0 against the same sales invoice ,why I can make it? i wont to deny make delivery note with quantity equal 0

Anurag810:
fixed by #14211

Anurag810:
fixed by #16061 will be merged soon

Anurag810:
Fixed will be merged soon

PR #16061, opened 2018-11-22
[fix] delivery note with quantity equal 0 #14211

Anurag810:
fixed the delivery note which can not be have quantity less than zero


deepeshgarg007:
Rebase your branch from upstream staging-fixes and then push again

erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -119,6 +119,12 @@ def validate(self):
 		self.update_current_stock()
 
 		if not self.installation_status: self.installation_status = 'Not Installed'
+	
+	def validate_qty(self):
+		data = frappe.get_all("Delivery Note Item", filters={"parent" : self.name}, fields=["qty"])
```
Alchez: The `self` object already has the items data, so you don't need the `get_all`. Simply do:

```python
for item in self.items:
	if not item.qty:
		frappe.throw("Item quantity can not be zero")
```

Also, shouldn't this also be called in the `validate` method?

Alchez (changes requested):

Issue #14949, opened 2018-07-18
Customer/ Supplier Dashboard shows incorrect amounts


shreyashah115:
The dashboard shows incorrect values for a multi company setup.

Eg. 
An invoice with an outstanding amount of $ 250 and the company used during that transaction had a default currency as INR, so ₹ 12,500.

<img width="952" alt="screen shot 2018-07-18 at 11 56 17 am" src="https://user-images.githubusercontent.com/17617465/42863724-aac0954a-8a81-11e8-9188-5dd2c8f170fd.png">

Now, my default company's currency is USD. But, on the supplier's dashboard, we get $ 12500 which is incorrect.

<img width="1025" alt="screen shot 2018-07-18 at 11 55 17 am" src="https://user-images.githubusercontent.com/17617465/42863723-aa80a2f0-8a81-11e8-8a12-baf5343e3b0d.png">

PR #16065, opened 2018-11-23
Customer/Supplier dashboard fix for multi company setup

deepeshgarg007:
-Annual Billing and Total Unpaid can be seen for each company in a multi company setup
-Should Fix:https://github.com/frappe/erpnext/issues/14949

screenshot 2018-11-27 at 5 07 08 pm


erpnext/accounts/party.py
```diff
@@ -459,38 +459,61 @@ def get_timeline_data(doctype, name):
...
+
+	company_wise_grand_total = frappe._dict(frappe.db.sql("""
+		select company, sum(grand_total)
+		from `tab{0}`
+		where {1}=%s and docstatus=1 and posting_date between %s and %s
+		group by company"""
+		.format(doctype, party_type.lower()),
+		(party, current_fiscal_year.year_start_date, current_fiscal_year.year_end_date)))
+
+	company_wise_base_grand_total = frappe._dict(frappe.db.sql("""
```
nabinhait: Why separate query for base_grand_total?

nabinhait (changes requested):

PR #16105, opened 2018-11-26
Account creation after insert customer and supplier

mhbu50:
Most accountants in Saudi Arabia they prefer to create a separate account for each Customers & suppliers and employees.

image
This PR will add account by defualt in non-standard account:
image


mhbu50:
@rohitwaghchaure Could you reviwe my PR, Please.

FinForceConsulting:
@mhbu50 Mohammed: "create a separate account for each customer & suppliers and employees"
This is not a good idea as a creation of separate Ledger account is a concept derived from non-database driven systems like Tally. In most ERPs there is a concept of MASTERS & LEDGER ACCOUNTS.

In ERPNext, Customer master holds a list of each customer, which is tagged to Debtors A/c and so on.

For the case of Non-standard account tagging, that is to be used when there is a change in Currency, else it's not useful.

**I do not see any benefit of Creating individual ledger (COA entity) for each Customer/Supplier/Employee.**

sagarvora:
Travis is failing. Please fix and reopen.

PR #16112, opened 2018-11-27
[Fix] Accounts receivable summary print is not working

rohitwaghchaure:
**Issue**
Showing zero values
image

**After Fix**
image


ERPtrismart (approved):
Kindly Approve

PR #16117, opened 2018-11-29
De-clutter, Regionalize and Domainify Accounts Module Configuration (on staging-fixes)

agritheory:
**Problem 1**:
The "Billing" Section of the Accounts config contains items that are 1) not related to billing and 2) has too many entries. 

image

Fix:
image

**Problem 2**:
In the config file, it is possible to positively select entires based on country, but not to hide sections, notably the GST section in the accounts module. This section only effects users in India and should be hidden if there is not an Indian company record.
image

**Problem 3**: 
There are enough Retail functions on this page that they should be consolidated:
image

**Problem 4**:
Subscriptions are way down at the bottom and this ins inconvenient for service companies. If the Service Domain is active it should be moved closer to the top. Likewise Retail. 

accounts-new-layout-services

Retail-optimized layout. 
accounts-new-layout-retail

### DIFF
--- a/erpnext/config/accounts.py
+++ b/erpnext/config/accounts.py
@@ -1,8 +1,10 @@
 from __future__ import unicode_literals
 from frappe import _
+import frappe
+
 
 def get_data():
-	return [
+	config = [
 		{
 			"label": _("Billing"),
 			"items": [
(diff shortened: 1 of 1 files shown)


shreyashah115 (approved):

rmehta:
Merged but should have gone to `develop`

agritheory:
@rmehta Agreed and noted. I originally opened there and this was what Shreya suggested when Travis would not play nice. Thanks for pulling it in.

PR #16131, opened 2018-11-30
Update package total on old delivery trips

Alchez:
**Problem:**

Package total wasn't being updated if a user added or deleted Delivery Notes in the Trip.


Alchez:
Hey, this is absolutely my bad. I completely forgot to contribute the field itself. The patch will break on a migrate, since that field doesn't exist. Can you please revert this commit? @rmehta, @nabinhait

rmehta:
Release done. Send another patch asap

PR #16138, opened 2018-12-03
[Fix] Permissions issue while making item variants

rohitwaghchaure:
We have recently [fixed](https://github.com/frappe/frappe/pull/6509) vulnerability issue in which we always check the passed parent is actual parent of the child doctype or not 

**Issue**
screen shot 2018-12-03 at 11 44 19 am

**After fix**
image


vrms:
nice, thanks @rohitwaghchaure

amr444555:
thanks a lot , really appreciate your effort

PR #16141, opened 2018-12-03
[Fix] While making sales invoice from delivery note, system not remove the returned qty

rohitwaghchaure:
**Issue**

- Created sales order and delivery note for quantity 2.
- Created sales return against the delivery note for 1 quantity
- Now when trying to make sales invoice against the original delivery note, system fetching the quantity as 2 in the sales invoice instead of 1(because 1 is already returned)


rohitwaghchaure:
Added test cases
image

PR #16142, opened 2018-12-03
Linking company address to pos profile

deepeshgarg007:
Link company address mentioned in pos profile to sales invoice

screenshot 2018-12-03 at 2 47 40 pm
screenshot 2018-12-03 at 2 48 14 pm


erpnext/accounts/doctype/pos_profile/pos_profile.js
```diff
@@ -42,18 +42,32 @@ frappe.ui.form.on('POS Profile', {
 			frm.toggle_display('offline_pos_section', is_offline);
 			frm.toggle_display('print_format_for_online', !is_offline);
 		});
+
+		frm.set_query('company_address', function(doc) {
+			if(!doc.company) {
+				frappe.throw(_('Please set Company'));
```
Alchez: Frappe uses double-underscores `__()` for translations in JS. The single variant is for server-side.

Alchez (changes requested):

PR #16145, opened 2018-12-03
[Enhance] Added total quantity row in the batch wise balance history report

rohitwaghchaure:
#### **Before**
image

#### **After** 
image


Mishaal1021:
> Shouldn't this be automatically achieved by setting the "Add Total Row" in the report? Are the fieldtypes not compatible with the totals calculation?

It is, but since this is a standard report, other users can not enable the "Add Total Row" feature. This is an issue with most standard reports.

Alchez:
The "Add Total Row" is enabled for this report in the `staging-fixes` branch.

https://github.com/frappe/erpnext/blob/3ecb995485171503540982f2ac7137dd4052507c/erpnext/stock/report/batch_wise_balance_history/batch_wise_balance_history.json#L1-L3

rmehta:
@rohitwaghchaure is this required?

PR #16154, opened 2018-12-04
Set operations for Work Order when created from Material Request

Alchez:
**Describe the bug**
When you create a Work Order from a Material Request, the BOM operations for that item do not get copied over, which doesn't activate Job Cards and breaks the manufacturing workflow.

**To Reproduce**
1. Create, save and submit a Material Request set to "Manufacture" for any item that has a BOM with operations.
1. Click on "Make" on the top-right section.
1. Click on "Work Order" from the dropdown menu.
1. The newly created Work Order will not have any operations.

**Expected behavior**
Create Work Orders from Material Request for items that need operations.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser: Chrome
 - Version: 70

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.py
+++ b/erpnext/stock/doctype/material_request/material_request.py
@@ -441,31 +441,40 @@ def raise_work_orders(material_request):
 	errors =[]
 	work_orders = []
 	default_wip_warehouse = frappe.db.get_single_value("Manufacturing Settings", "default_wip_warehouse")
+
 	for d in mr.items:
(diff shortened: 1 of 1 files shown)


Alchez:
I'm trying to run tests for Material Request, but it keeps failing at trying to fetch BOMs, saying they don't exist.

There is a line in there saying test_dependencies = ["Currency Exchange", "BOM"], which I thought would create it. Do I need to setup these fixtures separately?

rmehta:
Merging for now, but next time #NoTestNoMerge

PR #16176, opened 2018-12-09
feat: Item Contents/Alternate UOM (requires: Enhanced UOM Conversions Table)

SaiFi0102:
Requires https://github.com/frappe/erpnext/pull/15865

For example: a Can of paint (Stock UOM) has 5 Litres (Contents UOM) of paint. The print format of transactions should show quantities in both UOMs

## Allow setting a Contents UOM from the Item master

image

image

## Calculate Contents Quantity in Transaction (Taxes and Totals) Documents
image

In the case when an Item does not have an Contents UOM set. It will not show in the Transaction's Item child table. However, it will calculate Contents Qty (`alt_uom_qty`) = Stock Qty (`stock_qty`)

## This allows the print format to include both quantities
image


sagarvora:
A suggestion: 
- Please fix existing PRs for semantic title, codacy errors & travis errors.
- Reinitate the conversation for stale PRs.
- Certain changes are big. Although your clients may have tested it, they still need to be tested by a wider range of users before being merged into production systems. Please rethink this and rebase to the `develop` branch where required.

PR #16182, opened 2018-12-10
feat: option for packing list from sales order to delivery note/sales invoice

RicardoJohann:
When creating a Delivery Note or a Sales Invoice (with Update Stock ticked) from a Sales Order which has a Product Bundle that has been changed in the meanwhile, the system populates the Packing List with the lastest item Bundle list even though the Sales Order has a different packing list. Add a check option "Maintain Product Bundle Items during Sales Cycle" in Sales Order to let User decide whether retrieve bundled items from Sales Order or Product Bundle

packeditemsalesorder


RicardoJohann:
Ok @rmehta. Please have a look to the following tests with checkbox ticked and not ticked:


checkbox_ticked_


checkbox_not_ticked_

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -63,38 +69,65 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
...
+			if [d.parent_item, d.parent_detail_docname] not in parent_items:
+				# mark for deletion from doclist
+				delete_list.append(d)
 
 	if not delete_list:
 		return doc
 
 	packed_items = doc.get("packed_items")
 	doc.set("packed_items", [])
+	c = 1
```
Alchez: Can you please add a comment, or use a better variable name to note why you are doing this step exactly?

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -63,38 +69,65 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
...
 
-def make_packing_list(doc):
+def make_packing_list(doc, doctype=None, name=None, fieldname=None):
 	"""make packing list for Product Bundle item"""
 
 	if doc.get("_action") and doc._action == "update_after_submit": return
 
+	maintain_packed_items_list = 0
+	if doctype is not None and doctype == "Sales Order":
+		maintain_packed_items_list = cint(frappe.get_value(doctype, name, fieldname))
```
Alchez: Based on the variable name, seems like you're checking for a very specific value here.

Why not just assume its consistent here, rather than have to maintain this code in so many different places (wherever you're calling `update_packing_list_suggested_from_sales_order_invoice`)?

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -63,38 +69,65 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
 	bin = get_bin_qty(packing_item_code, pi.warehouse)
 	pi.actual_qty = flt(bin.get("actual_qty"))
 	pi.projected_qty = flt(bin.get("projected_qty"))
+	packed_items_list.append(pi.idx)
 
-def make_packing_list(doc):
+def make_packing_list(doc, doctype=None, name=None, fieldname=None):
```
Alchez: Maybe use different names for the new variables, to distinguish them from `doc`.

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -63,38 +69,65 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
...
-def cleanup_packing_list(doc, parent_items):
+def cleanup_packing_list(doc, parent_items, packed_items_list, maintain_packed_items_list):
 	"""Remove all those child items which are no longer present in main item table"""
 	delete_list = []
 	for d in doc.get("packed_items"):
-		if [d.parent_item, d.parent_detail_docname] not in parent_items:
-			# mark for deletion from doclist
-			delete_list.append(d)
+		if maintain_packed_items_list:
+			if d.idx not in packed_items_list:
```
Alchez: IDX match seems unreliable. Child tables usually create a hash as the object name - maybe use that?

Alchez (changes requested):

erpnext/stock/doctype/packed_item/packed_item.py
RicardoJohann: Because this function is shared between sales invoice and sales order.

erpnext/stock/doctype/packed_item/packed_item.py
RicardoJohann: Since in function cleanup_packing_list idx is updated by c variable it becomes reliable.

erpnext/stock/doctype/packed_item/packed_item.py
Alchez: Yeah, that's true. But my point is remove the `fieldname` argument and directly use `maintain_packed_items_list`, since it's too specific to ever change.

More code to maintain means more scope for error.

RicardoJohann: Ok I got it

rmehta:
@RicardoJohann can we please have a test case for this?

RicardoJohann:
@rmehta I have added a test case

RicardoJohann:
@sagarvora I have fixed semantic-title.

Issue #6599, opened 2016-10-11
A lot of forms are cached, not able to return selection unless F5 refresh 


hoeichia:
eg 
i click setup print format builder, and selected a doc to edit
after edit , i click save > setup > print format builder again, last document show up instead of choose document screen, i have to press F5 to get there
this also apply to alot of pages which i forgot

![image](https://cloud.githubusercontent.com/assets/22652544/19255305/1a9a9446-8f8f-11e6-95c6-b5d620aa53fe.png)

paulator:
Same issue here with Release v7.1.20.

rmehta:
This is fixed

PR #16190, opened 2018-12-10
[Staging] Travis Fix: bench reinstall allow setting MariaDB root user/pass #6599

SaiFi0102:
For: https://github.com/frappe/erpnext/pull/16173

Travis fails at the command `bench reinstall --yes` because it prompts for keyboard input (for root user/pass) and times out. This PR allow passing Travis's MySQL root user/pass to `bench reinstall`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_script:
   - cd ~/frappe-bench
   - bench get-app erpnext $TRAVIS_BUILD_DIR
   - bench use test_site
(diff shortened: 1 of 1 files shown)


codingCoffee (approved):

SaiFi0102:
Maybe the the reinstall command should also have --admin-password? Or would it work without it?

codingCoffee:
@SaiFi0102 we have it in the test site's conifg.json, so it is not needed

https://github.com/frappe/erpnext/blob/staging-fixes/test_sites/test_site/site_config.json#L8

PR #16197, opened 2018-12-12
[ADD] Added functionality for Item Wise Tax applied on sale order line.

Solufyin:
Hello @nabinhait,

Added functionality for Item Wise Tax applied on sale order line.

screen_shot

**Please find attached video for more information.**

[Applied_tax_on_saleorder_line.zip](https://github.com/frappe/erpnext/files/2671175/Applied_tax_on_saleorder_line.zip)


rmehta (changes requested):
This needs to be part of the entire sales / purchase cycle. No point adding it in invoice only.

Also tests are missing

rmehta:
This is a very large project. You need to send a proposal first! @Solufyin

PR #16209, opened 2018-12-13
Rename BOM and Work order fields

rohitwaghchaure:
1. Rename field allow_transfer_for_manufacture to include_item_in_manufacturing in BOM, Work Order
1. Removed field Transfer Material Against Job Card from BOM and Work Order and added select field Transfer Material Against with option as Work Order/Job Card


erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -131,15 +134,28 @@ def set_transferred_qty(self):
...
 
-		self.set_status()
+		self.set_status(update_status)
+
+	def set_status(self, update_status=False):
+		self.status = {
+			"0": "Open",
+			"1": "Submitted",
+			"2": "Cancelled"
+		}[cstr(self.docstatus or 0)]
```
Alchez: `cstr` seems unnecessary. Just use integers as the keys?

Alchez (changes requested):

rmehta:
Patch breaks

rohitwaghchaure:
Fixed!

rmehta:
Merging for now, but next time #NoTestNoMerge

PR #16218, opened 2018-12-15
[Bug-fix] Analytics report graph skipping quarter values bug fix

deepeshgarg007:
screenshot 2018-12-15 at 3 31 24 pm

### DIFF
--- a/erpnext/selling/report/sales_analytics/sales_analytics.py
+++ b/erpnext/selling/report/sales_analytics/sales_analytics.py
@@ -276,7 +276,11 @@ class Analytics(object):
 
 	def get_chart_data(self):
 		length = len(self.columns)
-		labels = [d.get("label") for d in self.columns[2:length-1]]
+
+		if self.filters.tree_type in ["Customer", "Supplier", "Item"]:
+			labels = [d.get("label") for d in self.columns[2:length-1]]
+		else:
+			labels = [d.get("label") for d in self.columns[1:length-1]]
 		self.chart = {
 			"data": {
 				'labels': labels,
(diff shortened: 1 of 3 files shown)


erpnext/selling/report/sales_analytics/sales_analytics.py
```diff
@@ -276,7 +276,11 @@ def get_supplier_parent_child_map(self):
 
 	def get_chart_data(self):
 		length = len(self.columns)
-		labels = [d.get("label") for d in self.columns[2:length-1]]
+
+		if self.filters.tree_type in ["Customer", "Supplier", "Item"]:
+			labels = [d.get("label") for d in self.columns[2:length-1]]
```
Alchez: You don't need to provide `length-1`. You can list slice by using `self.columns[2:]` as well.

Although, if you're using list slicing, good practice would be to add a comment to say exactly why you are doing this, since it's not exactly clear why you're skipping some columns.

Alchez (changes requested):

erpnext/selling/report/sales_analytics/sales_analytics.py
deepeshgarg007: using ```self.columns[2:]``` will only slice first two columns where as I need to slice last column as well, thats why I have used ```length-1```.These columns are non-numeric columns which doesn't need to be displayed in graph.

Alchez: `length-1` will not slice the last column. You will need to do `length-2` for that. Or `self.columns[2:-1]`.

And add that description in code as a comment.

SaiFi0102:
I have already created a fix for this problem in my PR https://github.com/frappe/erpnext/pull/16179

PR #16220, opened 2018-12-15
feat: Party Ledger Summary Report

SaiFi0102:
This report summarizes Amounts Invoiced, Paid/Received, Debit/Credit Note and Write Offs/Discounts Allowed in a duration of time.

image

The Write Off Amount is calculated by querying all the GL Entries in a voucher that has BOTH a receivable/payable account and expense/income account and then calculating the amounts that went against expense (or income in the case of Supplier Ledger Summary) accounts.

You must select a Write Off account and Discount Allowed/Received account in the Company Master for the amounts to be shown in the correct column, otherwise it will show in the "Other Adjustments" columns

image


rmehta (changes requested):
Looks good 👍 

Please add a test for this!

sagarvora:
Please review codacy.

PR #16235, opened 2018-12-18
Fix user permission checks

surajshetty3416:
Fix user permission check 

to support 
https://github.com/frappe/frappe/pull/6397


erpnext/accounts/party.py
```diff
@@ -151,10 +151,8 @@ def get_default_price_list(party):
 
 def set_price_list(out, party, party_type, given_price_list):
 	# price list
-	price_list = filter(None, get_user_permissions()
-		.get("Price List", {})
-		.get("docs", []))
-	price_list = list(price_list)
+	price_list = [d.get('doc') for d in get_user_permissions().get('Price List', []) \
```
rmehta: can you move this to a separate function? `get_permitted_documents('Price List')`

(it is also used below)

rmehta (changes requested):

PR #16253, opened 2018-12-20
[Fix] Filter cancelled and draft payments in sales payment summary

chdecultot:
Hi,

The sales payment summary report is currently incorrect if you have cancelled or draft payments.
These should be filtered for the report to be accurate.

I have also replaced some dots by tabs in the code.

Thanks!


rmehta:
@chdecultot can we add a test case for this? Really hard to verify if this fix will work

chdecultot:
Sure I'll add some test cases!

chdecultot:
Test cases added (painfully, thanks Travis!... :smile: )

nabinhait (approved):

PR #16256, opened 2018-12-20
[Fix] Pos bugs

chdecultot:
Hi,

This PR aims at correcting a few things on the POS:

1. The item group "All Item Groups" was hardcoded, whereas users may want to change the name of that group.
I have remove the hardcode and added a request to fetch the main group in the hierarchy.

2. The "Change POS profile" button was not working, because the fields in the dialog had no names. I have just added fieldnames for both fields.

3. I have corrected a typo in the "get_prompt_fields" function name.

4. I have added the campaign to the online POS because it is useful for some users (I have clients using the POS only on fairs and they want to categorize their POS invoices). Since the field exists for the offline POS in the POS profile, it should also be available in the online POS I guess.

Please let me know if you have any remark.

Thanks!

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -324,7 +324,8 @@ class SalesInvoice(SellingController):
 			return {
 				"print_format": print_format,
 				"allow_edit_rate": pos.get("allow_user_to_edit_rate"),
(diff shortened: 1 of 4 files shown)


olamide2:
Hi

Is the Change PoS functionality available for offline use ??

chdecultot:
Hi @olamide2,

No I don't think this functionality exists in the offline POS. I'm not sure it would be feasible anyway, since it is offline...
The above changes are only for the online POS (except for the default Item Group that I have also corrected on the offline POS also).

PR #16267, opened 2018-12-23
adding account type (payable) to "Payroll Payable" (2120) account

dulhaver:
The "Payroll Payable" (2120) account is missing the account type (payable).
Hope this PR fixes the matter

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/verified/standard_chart_of_accounts.py
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/verified/standard_chart_of_accounts.py
@@ -147,7 +147,9 @@ def get():
 			        _("Creditors"): {
 			            "account_type": "Payable"
 			        },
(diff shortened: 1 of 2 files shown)


sagarvora:
Hello, please try to check why travis might be failing:
https://travis-ci.com/frappe/erpnext/jobs/166957496

vrms:
@sagarvora thanks for the pointer. will check and see whether I can eliminate the problem.

vrms:
sorry @sagarvora can't draw any sense from the travis report, neither can I see any obvious mistake in the line I have added.

That being said ... I do not now python, so the reason may be obvious to anyone who does. Shall I close the PR as it's not mergable as is and I am lacking the expertize to make it work?

sagarvora:
The reason why this is failing is because it might be used as a receivable account in tests. 

I'm closing this for now.

Issue #15493, opened 2018-09-25
[Proposal] Support Enhancements and Service Levels (SLA)


rmehta:
- Support Team (Employee Group)
    - Issue Criticality (new doctype)
        - Keywords (for auto selection)
    - Service Level
        - Issue Criticality 
        - Employee Group
        - Applicable Hours (from / to)
        - For holiday or working day 
        - Holiday list 
        - Response time (time / unit) example (3, days) or (3, hours)
        - Resolution time (time / unit) example (3, days) or (3, hours) 
    - Support Contract Template
        - Service Level (child table)
    - Support Contract
        - Default Contract (check)
        - Customer (if not default)
        - Support Contract Template
        - Service Level (child table)
        - Start / End Date
    - Issue 
        - Issue Criticality (select default based on keywords, or global default)
        - Service Level (auto selection basis of criticality and customer)
        - Service level status (fulfilled / not fulfilled)
        - Hours to respond (auto calculate every few mins)
        - Hours to resolve 
---

netchampfaris:
https://github.com/frappe/erpnext/pull/16828

PR #16268, opened 2018-12-23
feat: Support Enhancement and Service Levels (SLA)

hrwX:
Support Enhancement and Service Levels according to the proposal [#15493](https://github.com/frappe/erpnext/issues/15493)

- Employee Group (Support Team)
emp-group

- Service Level
    - Employee Group
    - Holiday list
    - Priority
    - Response time (time / unit) example (3, days) or (3, hours)
    - Resolution time (time / unit) example (3, days) or (3, hours) 
    - Support and Resolution (Child Table)
        - Workday
        - Start Time
        - End Time
serv-lvl

- Service Level Agreement
    - Default Service Level Agreement (check)
    - Customer (if not default)
    - Service Level
    - Start / End Date
    - Contract Status
        - Active
        - Expired
serv-lvl-agr

- Issue (Updated Existing)
    - Service Level (auto selection on basis customer)
    - Service level status (fulfilled / not fulfilled)
    - Hours to respond (auto calculated)
    - Hours to resolve (auto calculated)
    - Priority
iss


Alchez:
Looks great!

@hrwX One question though: there already is a DocType called Contract, so should Support Contract be integrated into that, as a sub-type? Having both would seem confusing.

hrwX:
@Alchez An agreement can include any understanding between multiple parties about what they promise to do whereas a contract follows a more formal outline than an agreement. I will rename the DocType to avoid confusion if any, so Service Level Agreement -> Service Level Agreement Template and Support Contract ->Service Level Agreement.
Sounds Good ?

> Looks great!
> 
> @hrwX One question though: there already is a DocType called Contract, so should Support Contract be integrated into that, as a sub-type? Having both would seem confusing.

erpnext/support/doctype/service_level/test_service_level.py
```diff
@@ -0,0 +1,106 @@
...
+from __future__ import unicode_literals
+from erpnext.hr.doctype.employee_group.test_employee_group import make_employee_group
+from erpnext.hr.doctype.holiday_list.test_holiday_list import make_holiday_list
+
+import frappe
+import unittest
+
+class TestServiceLevel(unittest.TestCase):
+
+	def test_service_level(self):
```
surajshetty3416: @hrwX These tests are not useful. Here you are checking if you are getting the inserted data... why do we want to test that!

We need tests to check the SLA functionalities with different cases. 

eg.test to check Response time allocation if there's a holiday in between, or test to check if the status is changed accordingly based on the response time, or test check if only working hours are considered while calculating the response time, etc.

erpnext/support/doctype/issue/issue.py
```diff
@@ -97,6 +98,123 @@ def split_issue(self, subject, communication_id):
 			doc.save(ignore_permissions=True)
 		return replicated_issue.name
 
+	def before_insert(self):
+		week = ['Monday', 'Tuesday', 'Wednesday', 'Thursday','Friday', 'Saturday', 'Sunday']
```
surajshetty3416: You can make a common function to get weekdays... since it's used/created twice in this PR...

erpnext/support/doctype/issue/issue.py
```diff
@@ -97,6 +98,123 @@ def split_issue(self, subject, communication_id):
...
+	#	Loop starts counting from current weekday and iterates till
+	#	time_set_flag is set indicating the time has been calculated.
+
+	while time_set_flag != 1:
+		for count, weekday in enumerate(week):
+
+			#	Initially time_added_flag is zero and the code will only start
+			#	executing if today and weekday is the same and keep executing
+			#	henceforth as time_add is incremented.
+
```
surajshetty3416: Please do not explain the code in your comments, if possible explain the intent of the code instead. Maybe add some examples if the code is getting too complicated.

You should also break the code to reduce the code complexity and[ for other reasons](https://github.com/frappe/erpnext/wiki/Coding-Standards#break-functionsmethods-more-than-10-lines-of-code)

surajshetty3416 (changes requested):

erpnext/support/doctype/service_level/test_service_level.py
hrwX: Have checked the test functionalities in DocType Issue tests

surajshetty3416: @hrwX Yes, but still those tests are bit weak and are of the similar kind... need to cover more scenarios

surajshetty3416 (changes requested):

PR #16271, opened 2018-12-24
Rename Accounts to Accounting

surajshetty3416:
After internal discussion, it was decided that `Accounts` should be renamed as `Accounting` since it is the widely accepted term used for "accounts" related work

screenshot 2018-12-24 at 12 17 07 pm

Pending tasks
- [ ] Update docs


rmehta:
Woah!

- Can't you just rename the label?
- In Staging fixes?

PR #16273, opened 2018-12-24
General UI fixes and changes

surajshetty3416:
- Replace 'Make' with 'Create'
(Since 'Create' seems to be more contextually correct than 'Make')

- Replace desktop icon for Project list with the icon for Projects module

- Rename Accounts -> Accounting 
After internal discussion, it was decided that Accounts should be renamed as Accounting since it is the widely accepted term used for "accounts" related work
(Simply renamed the label of a desktop icon since module view uses desktop icon label to display the page heading)

- Rename Learn to Help


rmehta:
@surajshetty3416 patch failing

nabinhait (approved):

Issue #16212, opened 2018-12-13
Loyalty point issue throwing error


jaqobk:
Hi

am facing an error in my system, am trying to open the customer and am getting this error whenever I try to open at a specific customer and invoices too throw this error though they are opening.

> Traceback (most recent call last):
> File “/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/load.py”, line 33, in getdoc
> run_onload(doc)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/load.py”, line 203, in run_onload
> doc.run_method(“onload”)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py”, line 772, in run_method
> out = Document.hook(fn)(self, *args, **kwargs)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py”, line 1048, in composer
> return composed(self, method, *args, **kwargs)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py”, line 1031, in runner
> add_to_return_value(self, fn(self, *args, **kwargs))
> File “/home/frappe/frappe-bench/apps/frappe/frappe/model/document.py”, line 766, in 
> fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
> File “/home/frappe/frappe-bench/apps/erpnext/erpnext/selling/doctype/customer/customer.py”, line 23, in onload
> self.load_dashboard_info()
> File “/home/frappe/frappe-bench/apps/erpnext/erpnext/selling/doctype/customer/customer.py”, line 29, in load_dashboard_info
> info[“loyalty_point”] = loyalty_point_details.loyalty_points
> TypeError: list indices must be integers or slices, not str
> 
> Traceback (most recent call last):
> File “/home/frappe/frappe-bench/apps/frappe/frappe/app.py”, line 61, in application
> response = frappe.handler.handle()
> File “/home/frappe/frappe-bench/apps/frappe/frappe/handler.py”, line 21, in handle
> data = execute_cmd(cmd)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/handler.py”, line 56, in execute_cmd
> return frappe.call(method, **frappe.form_dict)
> File “/home/frappe/frappe-bench/apps/frappe/frappe/init.py”, line 1007, in call
> return fn(*args, **newargs)
[...]

PR #16277, opened 2018-12-25
fix: Loyalty program issue fix in customer dashboard 

deepeshgarg007:
Should Fix:https://github.com/frappe/erpnext/issues/16212

screenshot 2018-12-25 at 2 32 31 pm


erpnext/accounts/party.py
```diff
@@ -506,12 +517,18 @@ def get_dashboard_info(party_type, party):
 
 		total_unpaid = flt(company_wise_total_unpaid.get(d.company))
 
+		if loyalty_point_details:
```
saurabh6790: Initialize loyalty_point_details as an empty list before `if party_type == "Customer":` Line No: 482.

Without this for other party types, this will result as

screen shot 2018-12-26 at 10 35 15 am

saurabh6790 (dismissed):

erpnext/accounts/party.py
deepeshgarg007: @saurabh6790  Done

jaqobk:
Hi

what the status of this issue?

Issue #15719, opened 2018-10-17
[bug] update qty in Sales Order [v11]


snvora:
Issue:

- Create Sales Order
- Create Delivery Note
- Update Qty in the Sales order.
- Qty update is allowed for qty less than delivered qty. 
- Qty update should check both delivered and invoiced qty.

Below is a screenshot from the beta version.

Before qty update (item 2 is delivered),

![capture](https://user-images.githubusercontent.com/10496564/47068169-345f4e80-d209-11e8-8d97-86f95571a77f.PNG)

Qty update is allowed even though the item is delivered.

![capture_1](https://user-images.githubusercontent.com/10496564/47068207-4ccf6900-d209-11e8-8049-1aa8a5fbd8d4.PNG)

asanodaria:
I have tried to reproduce this bug. I can't be reproduced it. Pls see attached GIF

![increase_qty_after_delivery_note](https://user-images.githubusercontent.com/36321437/47409790-78aa9b80-d781-11e8-91b0-211ce8bfeea7.gif)

snvora:
I think you did not get the issue right.

This issue is about updating qty and rate. New feature in V11. You get an option to update items on top, next to status and make button. You can update qty and rate using that feature.

I changed Qty of SO-0023 to 1. It should have been restricted to 2 units.

Hope you get it.

asanodaria:
> I think you did not get the issue right.
> 
> This issue is about updating qty and rate. New feature in V11. You get an option to update items on top, next to status and make button. You can update qty and rate using that feature.
> 
> I changed Qty of SO-0023 to 1. It should have been restricted to 2 units.
> 
> Hope you get it.

Ohh Yes I got it now.. I think you should mention clear and detail steps while writing bug as it would be better to understand..

PR #16296, opened 2018-12-27
fix: Validation while updating items in sales order

deepeshgarg007:
Should Fix:https://github.com/frappe/erpnext/issues/15719

If an item is delivered then update items button should not allow to set qty less than delivered qty

2 qty of Item "Iphone Xs" is fully delivered. So if try to set it to 1 then following validation will be thrown.
screenshot 2018-12-27 at 2 08 24 pm
screenshot 2018-12-27 at 2 08 46 pm
screenshot 2018-12-27 at 2 09 03 pm


nabinhait (changes requested):
test case missing

nabinhait (approved):

PR #16299, opened 2018-12-27
fix(assign_to): retain "assign_to " even after any issue or task is closed

sahil28297:
Earlier the assign_to field for any task or issue was set as blank when that issue/task was closed. Changed this behaviour and now the assign_to field retains its assignees even after the issue/task is closed.
Before :
image

After : 
image


joolsr:
I think the understanding is that if the task or issue is closed then any assignations would normally also hence be closed.

Does your patch still close the assignment or just keep the users name against it?

PR #16304, opened 2018-12-28
fix(website): Pagination in child item groups

netchampfaris:
Item groups pages which had child item groups were not paged at all despite having next, prev buttons


saurabh6790 (changes requested):
@netchampfaris screenshot (before/after)?

PR #16315, opened 2019-01-02
[Fix] HR - Leave allocation issue & Dashboard at Leave Application

pjamsheer:
- Regarding Leave allocation issue & Dashboard at Leave Application
- Related Discuss post: https://discuss.erpnext.com/t/regarding-leave-allocation-issue-dashboard-at-leave-application/43934

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.py
+++ b/erpnext/hr/doctype/leave_application/leave_application.py
@@ -369,7 +369,7 @@ def get_leave_details(employee, date):
 		date = allocation.to_date
 		leaves_taken = get_leaves_for_period(employee, d, allocation.from_date, date, status="Approved")
 		leaves_pending = get_leaves_for_period(employee, d, allocation.from_date, date, status="Open")
-		remaining_leaves = allocation.total_leaves_allocated - leaves_taken - leaves_pending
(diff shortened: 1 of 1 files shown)


sagarvora:
Hello. Leaves can be encashed only after year-end, isn't it?
So how can the same be subtracted from leaves allocated for the current year?

erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -369,7 +369,7 @@ def get_leave_details(employee, date):
 		date = allocation.to_date
 		leaves_taken = get_leaves_for_period(employee, d, allocation.from_date, date, status="Approved")
 		leaves_pending = get_leaves_for_period(employee, d, allocation.from_date, date, status="Open")
-		remaining_leaves = allocation.total_leaves_allocated - leaves_taken - leaves_pending
+		remaining_leaves = allocation.total_leaves_allocated - leaves_taken - leaves_pending - allocation.total_leaves_encashed
```
nabinhait: @pjamsheer Encashed leave should be considered based on encashment date, calculate directly from Leave Encashment form.

nabinhait (changes requested):

PR #16323, opened 2019-01-02
fix: leave application and timesheet tests

sagarvora:
Summary of changes made:
- increase Leave Period to ensure leave application doesn't fall outside it
(now - 7 days used to fail for first 7 days of year) 
- reset date of joining in timesheet test to prevent output of other tests to interfere with it


sagarvora:
Test cases run successfully, to fix patch testing merge: https://github.com/frappe/frappe/pull/6708

PR #16329, opened 2019-01-03
fix: fallback to stock uom if uom is not defined

rohitwaghchaure:
**Issue**

If UOM is not added then getting below error
```
  File "/home/frappe/benches/bench-2018-11-15/apps/frappe/frappe/app.py", line 62, in application
  File "/home/frappe/benches/bench-2018-11-15/apps/frappe/frappe/handler.py", line 22, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-2018-11-15/apps/frappe/frappe/handler.py", line 53, in execute_cmd
  File "/home/frappe/benches/bench-2018-11-15/apps/frappe/frappe/__init__.py", line 939, in call
    return fn(*args, **newargs)
  File "/home/frappe/benches/bench-2018-11-15/apps/erpnext/erpnext/stock/get_item_details.py", line 44, in get_item_details
    out = get_basic_details(args, item)
  File "/home/frappe/benches/bench-2018-11-15/apps/erpnext/erpnext/stock/get_item_details.py", line 261, in get_basic_details
    out.stock_qty = out.qty * out.conversion_factor
TypeError: can't multiply sequence by non-int of type 'float'
```

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -281,7 +281,7 @@ def get_basic_details(args, item):
 		out.conversion_factor = 1.0
 	else:
 		out.conversion_factor = args.conversion_factor or \
-			get_conversion_factor(item.item_code, args.uom).get("conversion_factor") or 1.0
+			get_conversion_factor(item.item_code, args.uom).get("conversion_factor")
 
 	args.conversion_factor = out.conversion_factor
 	out.stock_qty = out.qty * out.conversion_factor
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -651,7 +651,7 @@ def get_conversion_factor(item_code, uom):
 	if not conversion_factor:
 		stock_uom = frappe.db.get_value("Item", item_code, "stock_uom")
 		conversion_factor = get_uom_conv_factor(uom, stock_uom)
-	return {"conversion_factor": conversion_factor}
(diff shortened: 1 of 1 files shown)


sagarvora:
Please review PR title. It's a bug fix and same must be reflected.

erpnext/stock/get_item_details.py
```diff
@@ -281,7 +281,7 @@ def get_basic_details(args, item):
 		out.conversion_factor = 1.0
 	else:
 		out.conversion_factor = args.conversion_factor or \
-			get_conversion_factor(item.item_code, args.uom).get("conversion_factor") or 1.0
+			get_conversion_factor(item.item_code, args.uom or item.get('stock_uom')).get("conversion_factor", 1) or 1.0
```
sagarvora:
```suggestion
			get_conversion_factor(item.item_code, args.uom or item.get('stock_uom')).get("conversion_factor") or 1.0
```

seeing as how we already have `or 1.0`

sagarvora:
Why does the error try to multiply a sequence. Pls check if `out.qty` is a sequence, not float.

sagarvora:
@rohitwaghchaure I think that it was a non-empty string. Considering that if it were an empty string, it should execute the code after `or` and default to `1.0`
Let's use `flt` to solve this.

saurabh6790 (approved):

PR #16342, opened 2019-01-05
fix: rounded total if available while caclulating advance paid

finbyz:
Currently, if there is rounded total in orders and is greater that grand total if rounding is greater than .50, then while setting advance paid of rounded total value, then, it throws the error. So we have resolved by comparing rounded total if it is present.


sagarvora:
Please use conventional commits/PRs:
https://github.com/frappe/erpnext/wiki/Coding-Standards#commit-messages

Please try to ensure codacy passes.

PR #16356, opened 2019-01-07
chore: Updated DocType JSON Files to Clean Up Git Diff of Other PRs

SaiFi0102:
No changes! Only ran `frappe.get_doc("DocType", dt).save()` to update JSON files so that other PR's git diff will show only relevant changes and make rebasing easier.


scmmishra:
[Base64 Display Bug](https://trello.com/c/s0469oal/13-base64-display-bug)

PR #16358, opened 2019-01-07
feat(naming): Deferred naming for SLE and GL Entry

adityahase:
To allow fast insertions, use hash-based auto naming for Stock Ledger Entry and GL Entry.

An hourly run, scheduled job will rename the temporarily named docs using auto name options.

`to_rename` flag on both DocTypes is used to keep track of temporarily named docs.

**Note:** All occurences of `order by name` for both DocTypes is replaced by `order by creation`


erpnext/accounts/doctype/gl_entry/gl_entry.py
```diff
@@ -6,14 +6,21 @@
...
 
 exclude_from_linked_with = True
-
 class GLEntry(Document):
+	def autoname(self):
+		"""
+		Temporarily name doc for fast insertion
+		name will be changed using autoname options (in a scheduled job)
+		"""
+		self.name = frappe.generate_hash(txt="", length=10)
```
rohitwaghchaure: No need of this, keep auto name field blank in the GL Entry doctype.

erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -16,6 +16,13 @@ class StockFreezeError(frappe.ValidationError): pass
 exclude_from_linked_with = True
 
 class StockLedgerEntry(Document):
+	def autoname(self):
+		"""
+		Temporarily name doc for fast insertion
+		name will be changed using autoname options (in a scheduled job)
+		"""
+		self.name = frappe.generate_hash(txt="", length=10)
```
rohitwaghchaure: Keep auto name field as blank in the stock ledger entry doctype

rmehta: Autoname can't be blank since its primary key (frappe will default it hash as primary)

rohitwaghchaure (approved):

nabinhait (approved):

PR #16359, opened 2019-01-07
Fix: Errors in the Period closing voucher tests

Don-Leopardo:
Travis test, don't merge.


erpnext/accounts/doctype/period_closing_voucher/test_period_closing_voucher.py
```diff
@@ -13,6 +13,16 @@ class TestPeriodClosingVoucher(unittest.TestCase):
 	def test_closing_entry(self):
 		year_start_date = get_fiscal_year(today(), company="_Test Company")[1]
 
+		frappe.db.sql("""
```
rmehta: beats the whole point of the test!

Don-Leopardo: I thought so

PR #16308, opened 2018-12-28
fix: errors fixed in Supplier Scorecard

Don-Leopardo:
Resolves #16255  
An extra error in the creation of Supplier Scorecard Period:
captura de pantalla 2018-12-28 a la s 10 58 33

And some small bug in javascript and front end (related to fetches and console logs).

PR #16376, opened 2019-01-09
fix: unused argument in sql query

finbyz:
Removed unused keyword argument 'supplier' in SQL query of **make_all_scorecards** function.


Don-Leopardo:
In this PR #16308 I fix this problem and some others in Supplier Scorecard and some related DocTypes.

PR #16377, opened 2019-01-09
feat: Tax Category and Item Tax Template (v12)

SaiFi0102:
Requires the PR in frappe: https://github.com/frappe/frappe/pull/6772

<details><summary><b>Moved Item Tax Rates from Item Master to Item Tax Template Master:</b></summary>

image

</details>
<br />

<details><summary><b>You can create Tax Category to apply different taxes based on the Tax Category:</b></summary>

image

</details>
<br />

<details><summary><b>Item Taxes are set like this:</b></summary>

image

</details>
An empty Tax Category means that this Item Tax will be applied when the Tax Category in transactions is also empty
<br /><br />

<details><summary><b>You can also set Item Tax Templates on Item Group:</b></summary>

image

</details>
Item Tax Templates applied on an Item Group will be effective on all its descendants too unless a descendant Item Group overrides it OR an Item overrides it
<br /><br />

<details><summary><b>You can select Tax Category from any transaction which will set Item Tax Templates automatically:</b></summary>

image

</details>
<br />

<details><summary><b>You can also manually select Item Tax Template for each Item in a transaction:</b></summary>

image

</details>
<br />

<details><summary><b>You can also assign a Tax Category to a Customer, Supplier or a specific Address so that the Tax Category is automatically selected:</b></summary>

image

image

</details>
<br />

<details><summary><b>You can even create a Tax Rule with a Tax Category filter so that Taxes and Charges Template is automatically selected:</b></summary>

image

</details>
<br />

<details><summary><b>You can configure these new features:</b></summary>

image

</details>


erpnext/patches/v12_0/move_item_tax_to_item_tax_template.py
```diff
@@ -0,0 +1,80 @@
+import frappe
+import json
+from six import iteritems
+
+old_item_taxes = {}
+item_tax_templates = {}
+rename_template_to_untitled = []
```
nabinhait: Avoid global variables in python

https://stackoverflow.com/questions/19158339/why-are-global-variables-evil

nabinhait (changes requested):

nabinhait (changes requested):
@SaiFi0102 I have tested manually, it is working quite well. A few observation:
- `"Automatically Add Taxes and Charges from Item Tax Template"` checkbox should be checked by default and it should be default behaviour.
- `Item Tax Template` should be shown in default grid view (with item_code, qty, rate and amount) in transactions.

PR #16394, opened 2019-01-14
fix: Code clean-up and added validation for accounting period

rohitwaghchaure:
Accounting Period

screen shot 2019-01-14 at 11 47 06 pm

Added Validation
screen shot 2019-01-14 at 11 47 32 pm

### DIFF
--- a/erpnext/accounts/doctype/gl_entry/gl_entry.py
+++ b/erpnext/accounts/doctype/gl_entry/gl_entry.py
@@ -4,7 +4,7 @@
 from __future__ import unicode_literals
 import frappe, erpnext
 from frappe import _
-from frappe.utils import flt, fmt_money, getdate, formatdate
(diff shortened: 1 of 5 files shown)


erpnext/accounts/doctype/gl_entry/gl_entry.py
```diff
@@ -97,6 +100,30 @@ def validate_account_details(self, adv_adj):
 			frappe.throw(_("{0} {1}: Account {2} does not belong to Company {3}")
 				.format(self.voucher_type, self.voucher_no, self.account, self.company))
 
+	def validate_accounting_period(self):
```
nabinhait: This function should be added in `general_ledger.py` and should only be **called once** for a transaction.

nabinhait (changes requested):

PR #16407, opened 2019-01-16
fix: fix tests - report row order changes

clarkejj:
Basic test of General Ledger, Balance Sheet, Trial Balance, Cash Flow, Profit and Loss, Accounts Payable, Accounts Receivable

1. full 'run-tests' test suite _may_ need to run first (for the data a report test expects to assert on and pass)
2. tests are intended to assert on master doctype data since transaction doctype data is keyed (for eg voucher_no field _T-Journal Entry-00006 and so on) and these values are not fixed and may change on each test run so these values are not subject to assert.
3. Report coverage (ie the example data any given report reports on) is imported from the data in test_records.json test setup fixture input data (test_records = frappe.get_test_records('Report')
4. The report data content results have not been audited or 'certified' as accurate or correct!
5. No change or impact to production code, this is strictly test side code
6. The order of rows in a given report may change from one run to the next? 

tests moved to erpnext repos; in original PR tests in frappe repos https://github.com/frappe/frappe/pull/6594

### DIFF
--- a/erpnext/accounts/report/test_report.py
+++ b/erpnext/accounts/report/test_report.py
@@ -1,0 +1,272 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# See license.txt
+
+from __future__ import unicode_literals
+import frappe
+import json
+import os
+import unittest
+import datetime
+
+test_records = frappe.get_test_records('Report')
+
+
+class TestReport(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


nabinhait:
@clarkejj Thanks for the pull request.
But as the tests are dependent on other test records which can change anytime, it is not the correct way to test the reports. For report testing, the data must be fully independent from other PRs.

May be we can create a new company, create some data against it and then test the reports against that data.

PR #16412, opened 2019-01-17
fix(setup): disable rounded total by default

rmehta:


sagarvora:
I think rounded total should be default.

The Japanese Yen for example doesn't have anything similar to cents or paise. So decimals in that currency must be rounded off.

GST rules also require rounding off the tax amount.

PR #16433, opened 2019-01-19
fix: Gst checksum fix

hrwX:
Removing GST Checksum calculation as edge cases breaks in case of '27NSKG13429D1DR'.


sagarvora:
For the GSTIN you entered, the validation works just fine.

PR #16458, opened 2019-01-22
fix: NoneType object has no attribute "gstin"

finbyz:
In the latest update, if no Party GSTIN is provided, then it throws the following error:-

screenshot 10

So, we have fixed this error by using `get` method.


sagarvora:
Thanks for the PR! Merging..

PR #16488, opened 2019-01-24
fix(sales_invoice): fetch customer price list if available before pos price list

deepeshgarg007:
Fetch customer's ```default_price_list``` instead of POS profile ```selling_price_list``` in case of POS sales invoice


nabinhait (changes requested):

erpnext/accounts/party.py
```diff
@@ -149,12 +151,20 @@ def get_default_price_list(party):
...
 
 	if price_list:
 		price_list = price_list[0]
+	elif pos and party_type == 'Customer':
+		customer_price_list = frappe.get_value('Customer', party.name, 'default_price_list')
+
+		if customer_price_list:
+			price_list = customer_price_list
+		else:
+			pos_price_list = frappe.get_value('Pos Profile', pos, 'selling_price_list')
```
Alchez: `Pos Profile` -> `POS Profile`, otherwise this will always return `None`.

Maybe add a test for this.

Alchez (changes requested):

PR #16317, opened 2019-01-02
feat: AR/AP and Stock Ledger Report Group By Filter

SaiFi0102:
* Converted data to list of dicts
* Added Group By filter in AR/AP
* Added Group Totals

# AR/AP

screenshot from 2019-01-18 18-03-18
image

# Stock Ledger

stock-ledger-group-by-optinos
image

PR #16491, opened 2019-01-24
fix(ar-summary): Changes to column list to match recent changes

adityahase:
PR https://github.com/frappe/erpnext/pull/16447 is useless after https://github.com/frappe/erpnext/pull/16406. Reverting.

Before
screenshot 2019-01-24 at 7 16 30 pm

After
screenshot 2019-01-24 at 7 15 25 pm


SaiFi0102:
This is definitely a needed workaround, however, a permanent fix is included in my PR https://github.com/frappe/erpnext/pull/16317. AR/AP Summary is susceptible to errors because AR/AP returns a list-of-list rather than a list-of-dicts and that is interpreted with assumptions about which index refers to which column.

PR #16499, opened 2019-01-25
fix: ascii codec unicode issue

rohitwaghchaure:


adityahase:
@rohitwaghchaure 	https://github.com/frappe/erpnext/pull/16456

PR #16503, opened 2019-01-27
POS fix: configuration to remove zero qty item from the cart in pos profile

rohitwaghchaure:
In v10, if user wants to remove the item from the cart he was set the qty as zero. This feature has [removed](https://github.com/frappe/erpnext/pull/13471) from the v11 to fix the decimal qty issue. Now some users want this feature, so to fix both issue we have added the checkbox "Remove Zero Qty Item From Cart" in the pos profile. If checkbox is enabled then system will remove the zero qty item from the POS cart

screen shot 2019-01-27 at 1 32 30 pm


nabinhait (changes requested):
Please change the fieldname suggested by @sagarvora

PR #16505, opened 2019-01-27
feat: Enhancement  in POS functionality

deepeshgarg007:
- Render price_list based on customers default price list in both online and offline mode
- Addition of expense and custody fields in POS Closing voucher

pos


erpnext/accounts/page/pos/pos.js
```diff
@@ -1037,7 +1039,7 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		this.numeric_keypad.show();
 	},
 
-	make_item_list: function () {
+	make_item_list: function (customer=null) {
```
sagarvora: no need for `=null`: it is `undefined` (falsy) by default

erpnext/accounts/page/pos/pos.js
```diff
@@ -1417,8 +1426,20 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		this.child.income_account = this.pos_profile_data['income_account'] || this.items[0].income_account;
 		this.child.warehouse = (this.item_serial_no[this.child.item_code]
 			? this.item_serial_no[this.child.item_code][1] : (this.pos_profile_data['warehouse'] || this.items[0].default_warehouse));
-		this.child.price_list_rate = flt(this.price_list_data[this.child.item_code] * this.child.conversion_factor, 9) / flt(this.frm.doc.conversion_rate, 9);
-		this.child.rate = flt(this.price_list_data[this.child.item_code] * this.child.conversion_factor, 9) / flt(this.frm.doc.conversion_rate, 9);
+
+		customer = this.frm.doc.customer;
+		let rate
+
+		customer_price_list = this.customer_wise_price_list[customer]
```
sagarvora: semicolon missing.
is there a possibility that customer may not exist in array?

deepeshgarg007: Added semicolon.
```customer_wise_price_list``` is an object. If customer does not exist then rate will be fetched from POS price_list i.e else part will be executed

erpnext/accounts/page/pos/pos.js
deepeshgarg007: fixed

PR #16508, opened 2019-01-28
fix(human resources): gross pay calculation issue

navdeepghai:
In Salary slip the backend(python) gross pay calculations(earnings) not matching with frontend(js).

I found this issue in sum_components(python) function.
Here's the calculation on frontend(before saving the changes):
before_save
Here's the calculation on backend(after saving the changes):
after_save

It's throwing error**(Total Debit must be equal to Total Credit. The difference is 0.0100000000002)** while saving JV on submit of Salary Slip.

I fixed the issue in this pull request:
after_changing_float


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -360,7 +360,7 @@ def sum_components(self, component_type, total_field):
 			elif not d.amount:
 				d.amount = d.default_amount
 			if not d.do_not_include_in_total:
-				self.set(total_field, self.get(total_field) + flt(d.amount))
+				self.set(total_field, self.get(total_field) + flt(d.amount, 2))
```
sagarvora: shouldn't the float precision be fetched from `System Settings` instead?

sagarvora:
Please follow the coding standards & ensure that the PR title and commit message is semantic:
https://github.com/frappe/erpnext/wiki/Coding-Standards#commit-messages

navdeepghai1:
@sagarvora  

I  read the pull request prerequisites already, but I am not familiar with Semantic Changes.
I think it just integrated before 2-3 weeks.
Right now I can't change the comments but In future I will take care of it.

sagarvora:
I've fixed it for you :slightly_smiling_face:

PR #16509, opened 2019-01-28
fix(sqli): Avoid SQL Injection with sender param

adityahase:


erpnext/templates/utils.py
```diff
@@ -16,7 +16,7 @@ def send_message(subject="Website Query", message="", sender="", status="Open"):
 
 	customer = frappe.db.sql("""select distinct dl.link_name from `tabDynamic Link` dl
 		left join `tabContact` c on dl.parent=c.name where dl.link_doctype='Customer'
-		and c.email_id='{email_id}'""".format(email_id=sender))
+		and c.email_id='%s'""", (sender))
```
surajshetty3416: quotes not required around `%s`

surajshetty3416 (changes requested):

surajshetty3416 (approved):

PR #16513, opened 2019-01-29
feat: on click of pay button and on focus out of qty field remove the zero qty items from the cart

rohitwaghchaure:
If qty is zero then system will remove the item from the cart on click of pay button

remove_item_from_pos


sagarvora (dismissed):

erpnext/accounts/page/pos/pos.js
```diff
@@ -1219,11 +1226,11 @@ erpnext.pos.PointOfSale = erpnext.taxes_and_totals.extend({
 		me.bind_delete_event()
 	},
 
-	update_qty: function (item_code, qty) {
+	update_qty: function (item_code, qty, remove_zero_qty_items='') {
```
sagarvora: minor change: 
parameters in JS are `undefined` (falsy)  by default. no need for setting a default value here below.

PR #16530, opened 2019-01-31
fix: value missing for selling settings - sales_update_frequency

sagarvora:
Refer: https://discuss.erpnext.com/t/error-value-missing-for-selling-settings-sales-update-frequency/44960
Affects .ova users. Cannot replicate on local machine, but possible fix.


sagarvora:
Changes work locally:

screenshot from 2019-01-31 08-15-10

PR #16541, opened 2019-01-31
feat(regional): Germany module with address template

barredterra:
Add regional/germany module with address template, for a start. 

Next step should be to refactor `add_address_template()` into `regional/utils.py`

Expected to be filled with Germany-specific stuff in the coming months.


erpnext/regional/germany/setup.py
```diff
@@ -0,0 +1,21 @@
+import frappe, os
+
+def setup(company=None, patch=True):
+    if not patch:
```
rmehta: project users tabs, not spaces.

rmehta (commented):
👍 for starting the module. please use tabs for consistency.

PR #16544, opened 2019-01-31
fix(get_item_details): Price calculation based on UOM

SaiFi0102:
`args.get("price_list_uom_dependant")` in `get_item_details.py` was being used incorrectly


erpnext/stock/get_item_details.py
```diff
@@ -503,7 +503,7 @@ def get_price_list_rate_for(args, item_code):
 	if item_price_data:
 		if item_price_data[0][2] == args.get("uom"):
 			return item_price_data[0][1]
-		elif not args.get('price_list_uom_dependant'):
+		elif args.get('price_list_uom_dependant'):
```
nabinhait: Why it is wrong? If price list  is uom dependent, it should only fetch rate if uom matches or no uom mentioned. It should not fetch by multiplying conversion factor.

nabinhait (changes requested):

erpnext/stock/get_item_details.py
SaiFi0102: This has caused some confusion so I'll explain it a bit elaborately.

According to the history in ERPNext's GitHub, before `price_not_uom_dependant` field in Price List DocType was introduced, the default behavior was that the price would be calculated after multiplying the conversion factor (I saw it in a GIF). But somehow it got mixed up, such that multi uom tests failed but a workaround was added to make the test pass.

Another way of seeing it: "Price Not UOM Dependant" is the same as "Price is UOM Independent", meaning do not change the price according to conversion factor. So if this field is unchecked then the price should be calculated according to the conversion factor.

Before (bugged):
uom_dependent_bug

After (fixed):
uom_dependent_fix

SaiFi0102: This PR https://github.com/frappe/erpnext/pull/11836 will show that if the Price Not UOM Dependant is checked only then the price should not be multiplied with conversion factor.

SaiFi0102:
Reported by users as a bug:
* https://discuss.erpnext.com/t/sales-invoice-purchase-invoice-uom-conversion-factor-issue-with-price-list-rate-v11/46325
* https://discuss.erpnext.com/t/changing-uom-not-work-in-sales-order-and-purchase-order/46851
* https://github.com/frappe/erpnext/issues/16938

Olawale1:
Hi @SaiFi0102 

Thanks for this. Could you also look into the related Print Format issue or should we open a separate thread for that?

https://discuss.erpnext.com/t/sales-invoice-purchase-invoice-uom-conversion-factor-issue-with-price-list-rate-v11/46325/7

Kind regards,

PR #16545, opened 2019-01-31
feat: Enhanced UOM Conversion Table (v12)

SaiFi0102:
Reopened for V12: https://github.com/frappe/erpnext/pull/15865

## Problem:
For the example of a paint mixing business:
* Stock UOM is in Cans
* BOM is entered in Grams
* Purchase is done in Cans
* Sale is done in either Cans or Packs
* Analysis is done in Litre

The problems are:
- We can not enter how many Litres a Can has directly, we have to enter the conversion factor as 1/size_of_can.
- We can't enter the density of the paint: how many Grams are in a Litre (has an indirect conversion to the stock UOM which is Can), we have to do more calculation for that
- We can only enter how many Cans are in a Pack

Another problem:
* If a Can has 3 Litres in it, the conversion_factor becomes 0.333
* If we put 3 Litres in any transaction, ERPNext would calculate the Stock Qty as 0.999, even though it should be 1.

My solution is to increase the conversion_factor's precision.

## Solution
The new UOM Conversion Graph child table acts as an _adjacency list_ for a **graph** with the set of UOMs as vertices and conversion_factors between UOMs as edges. To get the conversion_factors from any UOM to stock UOM we can run a Depth First Search to find all conversion paths for that UOM to the Stock UOM and the product of the weights will be the `conversion_factor`. 

The only change to focus on is in `item.py` the rest is mostly modification to conversion_factor fields precision to 9 to prevent the small fractional errors in Qty and to reflect the precision in `UOM Conversion Factor` DocType. Or they are updates to the test cases to manipulate the new `UOM Conversion Graph` child table instead of the `UOM Conversion Detail` child table.

image

## Patch
I have written a patch that fills the new UOM Conversion table using the logic:
**From Qty:** `row.conversion_factor if abs(row.conversion_factor) >=1 else 1`
**From UOM:** `doc.stock_uom`
**To Qty:** `1 if abs(row.conversion_factor) >=1 else 1/row.conversion_factor`
**To UOM:** `row.uom`

## Future possibility
UOM Conversion Factors can be cached and a graph of UOMs conversions can be made in the cache as well. This will improve performance of calculating conversion factors. And this will also allow conversions like Can -> Milliliter or Can -> Fluid Ounce or any conversion that is somehow linked in the graph. (Given the item has a conversion factor for Can -> Litre EVEN IF there is no conversion factor for Litre -> Milliliter or Litre -> Fluid Ounce)


erpnext/buying/doctype/purchase_order_item/purchase_order_item.json
```diff
@@ -562,7 +562,7 @@
    "oldfieldname": "conversion_factor", 
    "oldfieldtype": "Currency", 
    "permlevel": 0, 
-   "precision": "", 
+   "precision": "9",
```
sagarvora: a configurable in `Stock Settings` (that can default to the value of `9`)? Set in buying and selling controllers?

erpnext/stock/get_item_details.py
```diff
@@ -558,7 +558,7 @@ def get_price_list_rate_for(args, item_code):
 	if item_price_data:
 		if item_price_data[0][2] == args.get("uom"):
 			return item_price_data[0][1]
-		elif not args.get('price_list_uom_dependant'):
+		elif args.get('price_list_uom_dependant'):
```
sagarvora: agreed the variable naming was incorrect, but this was defaulting to have price dependent on uom. Now this has to be set everywhere.

change price list doctype to have the variable named as `price_not_uom_dependent` 
(reference: https://www.grammar-monster.com/easily_confused/dependant_dependent.htm)

use the same variable everywhere instead of `price_list_uom_dependant`.

Change this block like so:
```
elif not args.get('price_not_uom_dependent'):
	return flt(item_price_data[0][1] * flt(args.get("conversion_factor", 1)))
else:
	return item_price_data[0][1]

```

erpnext/stock/doctype/item/item.py
```diff
@@ -472,7 +577,7 @@ def validate_conversion_factor(self):
 
 			if d.uom and cstr(d.uom) == cstr(self.stock_uom) and flt(d.conversion_factor) != 1:
 				frappe.throw(
-					_("Conversion factor for default Unit of Measure must be 1 in row {0}").format(d.idx))
+					_("Conversion factor for default Unit of Measure must be 1"))
```
sagarvora: Why change to a less useful error message?

SaiFi0102: Because the user will no longer use the old Conversion Factors table. Except they will be using the new Conversion Graph table and its row # cannot be determined (since the values may be computed using multiple rows from the conversion graph).

When this error is thrown, the (old) conversion factor table will not be updated and hence the row # will only confuse the user.

erpnext/stock/get_item_details.py
SaiFi0102: `Price List` DocType's field is already named `price_not_uom_dependant`
`get_item_details.py` sets a property (arg) named `price_list_uom_dependant`

So are you saying I should refactor `price_list_uom_dependant` in get_item_details to `not price_not_uom_dependant`

sagarvora:
Adding a child table affects performance. Maybe we can do it with one table instead of two?

sagarvora:
Remove unused import (codacy)

erpnext/stock/get_item_details.py
sagarvora: No.
`price_list_uom_dependant` is actually the same as `price_not_uom_dependant` in Price List.

I'm asking you to change all instances of both these fields to the correct name:
`price_not_uom_dependent`

Subsequently, the existing logic was just fine.

erpnext/buying/doctype/purchase_order_item/purchase_order_item.json
SaiFi0102: Hmm. I would prefer if this is the default. Users can always configure field precision themselves anyway. The reason why I would rather keep `9` as default is that this should not have any breaking change but only prevent a precision loss as I explained in the description above.

erpnext/stock/get_item_details.py
SaiFi0102: Follow this specific change in PR https://github.com/frappe/erpnext/pull/16544
I'll include the reason and Sagar's proposed changed in it.

erpnext/buying/doctype/purchase_order_item/purchase_order_item.json
SaiFi0102: Another reason that I decided to use precision: 9 is that UOM Conversion Factor DocType also uses it

SaiFi0102:
@sagarvora The import is due to an auto generated file from frappe.

PR #16558, opened 2019-02-02
fix: Validation while getting raw materials for production in Production Plan

deepeshgarg007:
```
  File "/home/frappe/benches/bench-2019-01-23/apps/frappe/frappe/app.py", line 61, in application
  File "/home/frappe/benches/bench-2019-01-23/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-2019-01-23/apps/frappe/frappe/handler.py", line 56, in execute_cmd
  File "/home/frappe/benches/bench-2019-01-23/apps/frappe/frappe/__init__.py", line 1013, in call
    return fn(*args, **newargs)
  File "/home/frappe/benches/bench-2019-01-23/apps/erpnext/erpnext/manufacturing/doctype/production_plan/production_plan.py", line 571, in get_items_for_material_requests
    item_master = frappe.get_doc('Item', data['item_code']).as_dict()
KeyError: u'item_code'
```

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -571,7 +571,7 @@ def get_items_for_material_requests(doc, sales_order=None, company=None):
 			if data['item_code']:
 				item_master = frappe.get_doc('Item', data['item_code']).as_dict()
 			else:
-				frappe.throw(_("Please select items in SELECT ITEMS section"))
+				frappe.throw(_("Please select items first"))
 
 			purchase_uom = item_master.purchase_uom or item_master.stock_uom
 			conversion_factor = 0
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -567,12 +567,8 @@ def get_items_for_material_requests(doc, sales_order=None, company=None):
 				else:
 					item_details = get_subitems(doc, data, item_details, bom_no, company,
 						include_non_stock_items, include_subcontracted_items, 1, planned_qty=planned_qty)
-		else:
-			if data['item_code']:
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -568,7 +568,11 @@ def get_items_for_material_requests(doc, sales_order=None, company=None):
 					item_details = get_subitems(doc, data, item_details, bom_no, company,
 						include_non_stock_items, include_subcontracted_items, 1, planned_qty=planned_qty)
 		else:
-			item_master = frappe.get_doc('Item', data['item_code']).as_dict()
+			if data['item_code']:
+				item_master = frappe.get_doc('Item', data['item_code']).as_dict()
+			else:
+				frappe.throw(_("Please select items in SELECT ITEMS section"))
```
sagarvora: Could it be as simple as "Please select items first"? Explicit is better than implicit, but it's pretty intuitive in this case.

deepeshgarg007: @sagarvora  Done

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -568,7 +568,11 @@ def get_items_for_material_requests(doc, sales_order=None, company=None):
 					item_details = get_subitems(doc, data, item_details, bom_no, company,
 						include_non_stock_items, include_subcontracted_items, 1, planned_qty=planned_qty)
 		else:
-			item_master = frappe.get_doc('Item', data['item_code']).as_dict()
+			if data['item_code']:
```
Alchez: This won't fix the issue. You're still trying to access a key that doesn't exist. Use `data.get("item_code")` here instead.

Alchez (changes requested):

Issue #16251, opened 2018-12-20
'Expected End Date' in task should not be later than 'Expected End Date' of project


asbasawaraj:
'Expected End Date' in task should not be later than 'Expected End Date' of project

asbasawaraj:
This feature is available in v11. However the error message needs to be improved.

Currently, it says: Expected end date cannot be after Project: 'Project One' Expected end date

Pls change it to Expected End Date of the task {task name} cannot be after Project: 'Project One' Expected End Date

PR #16586, opened 2019-02-06
fix: Validate task dates in project

deepeshgarg007:
Fix for: https://github.com/frappe/erpnext/issues/16251
screenshot 2019-02-10 at 5 35 28 pm

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -81,13 +81,22 @@ class Project(Document):
 	def validate_dates(self):
 		if self.tasks:
 			for d in self.tasks:
-				if not getdate(self.expected_start_date) <= getdate(d.start_date) <= getdate(self.expected_end_date):
-					frappe.throw(_("Start date of task <b>{0}</b> should be between <b>{1}</b> expected start and end dates")
-						.format(d.title, self.name))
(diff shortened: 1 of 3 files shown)


erpnext/projects/doctype/project/project.py
```diff
@@ -79,6 +79,25 @@ def validate_project_name(self):
 			frappe.throw(_("Project {0} already exists").format(frappe.safe_decode(self.project_name)))
 
 	def validate_dates(self):
+		if self.tasks:
+			for d in self.tasks:
+				if self.expected_start_date:
+					if d.start_date and getdate(d.start_date) < getdate(self.expected_start_date):
+						frappe.throw(_("Start date of task <b>{0}</b> cannot be less than <b>{1}</b> expected start date")
```
netchampfaris: Show the actual End Date of the project in this message.

Start date of task **ABC** cannot be less than project's start date **2019-09-08**

netchampfaris (changes requested):

erpnext/projects/doctype/project/project.py
deepeshgarg007: @netchampfaris Done

netchampfaris (approved):

PR #16589, opened 2019-02-06
fix: Remove illegal character after break

hrwX:


erpnext/erpnext_integrations/doctype/woocommerce_settings/woocommerce_settings.py
```diff
@@ -28,7 +28,7 @@ def create_delete_custom_fields(self):
 
 				if not frappe.get_value("Custom Field",{"name":i[0]}) or not frappe.get_value("Custom Field",{"name":i[1]}):
 					create_custom_field_id_and_check_status = True
-					break;
```
sagarvora: FYI, not illegal as a character, just unwanted.
https://docs.python.org/3/reference/compound_stmts.html

PR #16590, opened 2019-02-06
fix(stock_balance): Items with a % sign crash Stock Balance.

gbm001:
Items with a % sign in the name cause an Exception when it reaches the SQL command unparameterized in stock_balance.py.

This pull request fixes the SQL command so it uses proper parameterization, rather
than a formatting command. This has the side benefit of protecting more against SQL injection (although an escaping method was used in the original code).

### DIFF
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -217,19 +217,22 @@ def get_item_details(items, sle, filters):
 		items = list(set([d.item_code for d in sle]))
 
 	if items:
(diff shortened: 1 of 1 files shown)


gbm001:
Just seen https://github.com/frappe/erpnext/pull/16537 by @nabinhait although that looks more complicated (to me) and still doesn't use a parametrized query.

gbm001:
The Codacy alert is a false positive - a string formatting operation is used to construct the SQL query, but placeholders for parameterized data are inserted rather than user data, so there is no risk of injection. All user data is inserted by parameterization.

PR #16595, opened 2019-02-07
feat: GSTR3B Report json creation and print format

deepeshgarg007:
Json Creation:
screenshot 2019-02-07 at 9 16 07 am
Print Format:
screenshot 2019-02-07 at 9 16 15 am


erpnext/regional/doctype/gstr_3b_report/gstr_3b_report.html
```diff
@@ -0,0 +1,245 @@
...
+		font-size: 8.0pt !important;
+		font-family: Tahoma, sans-serif;
+	}
+	.disabled {
+		background-color: #d9d9d9;
+	}
+
+</style>
+<h3 class="text-center">GSTR3B-Form</h3>
+<h5>GSTIN:&nbsp{{ data.gstin }}</h5>
```
sagarvora: One space can be added without using `&nbsp`.

erpnext/regional/doctype/gstr_3b_report/gstr_3b_report.html
```diff
@@ -0,0 +1,245 @@
...
+
+<h5>3.1&nbsp&nbspDetails of Outward Supplies and inward supplies liable to reverse charge</h5>
```
sagarvora: Better to use `&nbsp;`

And have a space after it too, helps with readability.

erpnext/patches/v12_0/set_gst_category.py
```diff
@@ -0,0 +1,43 @@
...
+		"input service": "Import Of Service",
+		"capital goods": "Import Of Capital Goods",
+		"input": "Input Service Distributor"
+	}
+
+	has_gst_fields = frappe.db.has_column('Purchase Invoice','eligibility_for_itc')
+
+	if has_gst_fields:
+		for old, new in itc_update_map.items():
+			frappe.db.sql("UPDATE `tabPurchase Invoice` SET eligibility_for_itc = replace(eligibility_for_itc, %s, %s)", (old, new))
```
karthikeyan5:
```suggestion
			frappe.db.sql("UPDATE `tabPurchase Invoice` SET eligibility_for_itc = %s where eligibility_for_itc = %s", (new, old))
```
this query would be faster to execute and is more correct in this use case.

karthikeyan5:
since the `invoice_type` field was removed in the patch can you please update the same in the following reports:

- gst_purchase_register

karthikeyan5 (changes requested):

test2a:
yay...... finally after such a long time, GSTR 3B is finally here. Good job

karthikeyan5 (dismissed):

netchampfaris (approved):

erpnext/regional/doctype/gstr_3b_report/gstr_3b_report.html
```diff
@@ -0,0 +1,297 @@
...
+		font-size: 8.0pt !important;
+		font-family: Tahoma, sans-serif;
+	}
+	.disabled {
+		background-color: #d9d9d9;
+	}
+
+</style>
+<div>
+	<h3 class="text-center">GSTR3B-Form</h3>
```
netchampfaris: Add Translated Strings

netchampfaris (changes requested):

netchampfaris (approved):

PR #16608, opened 2019-02-08
fix(work_order): Add item name in Work Order

Alchez:
**Problem:**

The manufactured item's name is not stored in the Work Order itself, just the item description. Although it's generally useful info to have on the document, one use case is that it can be used as the WO's title field, if required.

**Patch:**

image

**Form View:**

image


erpnext/patches/v12_0/add_item_name_in_work_orders.py
```diff
@@ -0,0 +1,13 @@
+import frappe
+
+
+def execute():
+	frappe.reload_doc("manufacturing", "doctype", "work_order")
+
+	for wo in frappe.get_all("Work Order"):
```
rmehta: this can be fixed in one `UPDATE` query

rmehta (dismissed):

erpnext/patches/v12_0/add_item_name_in_work_orders.py
Alchez: Updated with `UPDATE` :)

sagarvora:
Pls fix merge conflict.

Anurag810:
Please resolve merge conflicts

PR #16612, opened 2019-02-08
fix: (Sales Invoice Return) Auto Update Payment Amount for Sales Invoice Return

jay-parikh:
**_Version:_ erpnext 11.1.4 frappe 11.1.4**
Steps to verify:
**1)** Sale item from POS.
**2)** Come to the Sales Invoice form and do Return against above POS.
**3)** Decrease some QTY from item table.
**4)** According to QTY adjustment amount is updated in Payment child table.


erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -391,9 +391,27 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 			? this.frm.doc["taxes"][tax_count - 1].total + flt(this.frm.doc.rounding_adjustment)
 			: this.frm.doc.net_total);
 
+		var paid_amount = 0.0;
+		var base_paid_amount = 0.0;
+		var base_grand_total = 0.0;
```
sagarvora: This variable is used only once.
Let's use `me.frm.doc.base_grand_total` instead.

jay-parikh: Fixed.

jay-parikh:
@sagarvora Why is it not moving?

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -390,10 +390,21 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 		this.frm.doc.grand_total = flt(tax_count
 			? this.frm.doc["taxes"][tax_count - 1].total + flt(this.frm.doc.rounding_adjustment)
 			: this.frm.doc.net_total);
-
+		var payment_status = true;
 		if(in_list(["Quotation", "Sales Order", "Delivery Note", "Sales Invoice"], this.frm.doc.doctype)) {
 			this.frm.doc.base_grand_total = (this.frm.doc.total_taxes_and_charges) ?
 				flt(this.frm.doc.grand_total * this.frm.doc.conversion_rate) : this.frm.doc.base_net_total;
+			if (this.frm.doc.is_pos && this.frm.doc.is_return) {
```
nabinhait: I think this should be handled in `set_default_payment` method only (by modifying if condition). 
https://github.com/frappe/erpnext/blob/hotfix/erpnext/public/js/controllers/taxes_and_totals.js#L619

And then the method should be called from https://github.com/frappe/erpnext/blob/hotfix/erpnext/public/js/controllers/taxes_and_totals.js#L574

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -390,10 +390,21 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
...
+				$.each(this.frm.doc['payments'] || [], function (index, data) {
+					if (payment_status) {
+						data.amount = me.frm.doc.base_grand_total;
```
nabinhait: amount should be in invoice currency

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -390,10 +390,21 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
...
+						data.base_amount = flt(data.amount * me.frm.doc.conversion_rate, precision("base_amount"));
+						payment_status = false;
+					}
+				});
+				this.frm.doc.paid_amount = flt(me.frm.doc.base_grand_total, precision("paid_amount"));
```
nabinhait: paid_amount should be in invoice currency

nabinhait (changes requested):

erpnext/public/js/controllers/taxes_and_totals.js
jay-parikh: The problem here is that if we add code in `set_default_payment()` and call this in `calculate_outstanding_amount()` then `calculate_outstanding_amount()` is called when user manually change Amount in Payments Child Table. Method doesn't get triggered when Item qty is being updated in Item Child Table to update Payment Amount respectively in Payments Child Table.

Where as we want Amount in Payments Child Table to be updated as soon as Item qty is being updated in Item Child Table which is working as we have added code to `calculate_totals()`.

jay-parikh: The problem here is that if we add code in `set_default_payment()` and call this in `calculate_outstanding_amount()` then `calculate_outstanding_amount()` is called when user manually change Amount in Payments Child Table. Method doesn't get triggered when Item qty is being updated in Item Child Table to update Payment Amount respectively in Payments Child Table.

Where as we want Amount in Payments Child Table to be updated as soon as Item qty is being updated in Item Child Table which is working as we have added code to `calculate_totals()`.

erpnext/public/js/controllers/taxes_and_totals.js
jay-parikh: Does it mean to use `base_paid_amount` field instead of `paid_amount`?

erpnext/public/js/controllers/taxes_and_totals.js
jay-parikh: Does it mean to use `base_amount` field instead of `amount`?

erpnext/public/js/controllers/taxes_and_totals.js
nabinhait: ok

erpnext/public/js/controllers/taxes_and_totals.js
nabinhait: I mean it should be like `data.amount = me.frm.doc.grand_total;`

erpnext/public/js/controllers/taxes_and_totals.js
nabinhait: `this.frm.doc.paid_amount = flt(me.frm.doc.grand_total, precision("paid_amount"));`

jay-parikh: Thanks for feedback Updated.

erpnext/public/js/controllers/taxes_and_totals.js
jay-parikh: Thanks for feedback Updated.

jay-parikh:
@nabinhait can we get this merged?

jay-parikh:
@nabinhait any updates?

jay-parikh:
@sagarvora Can we merge this?

jay-parikh:
@rohitwaghchaure, @rmehta  any updates for this pull request?

rohitwaghchaure:
@jay-parikh If pos invoice paid using multiple payment options, then return invoice not showing the correct amount in the payments

not_working_for_multiple_payments

jay-parikh:
@rohitwaghchaure Can you share your erpnext & frappe version? It's working without any issues. If pos invoice paid using multiple payment options, then return invoice is showing correct amount in the payments.
See below gif for more details.
partial-return fix

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -390,10 +390,21 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
...
 			: this.frm.doc.net_total);
-
+		var payment_status = true;
 		if(in_list(["Quotation", "Sales Order", "Delivery Note", "Sales Invoice"], this.frm.doc.doctype)) {
 			this.frm.doc.base_grand_total = (this.frm.doc.total_taxes_and_charges) ?
 				flt(this.frm.doc.grand_total * this.frm.doc.conversion_rate) : this.frm.doc.base_net_total;
+			if (this.frm.doc.is_pos && this.frm.doc.is_return) {
+				$.each(this.frm.doc['payments'] || [], function (index, data) {
+					if (payment_status) {
+						data.amount = me.frm.doc.grand_total;
```
rohitwaghchaure: You have set the grand_total amount directly due to which if the original invoice has multiple mode of payments then while making sales return entry the amount of first mode of payment has been set as grand total which is wrong. For more details please check below GIF

pos_return_issue

rohitwaghchaure:
@jay-parikh 
I have pulled your code in my local account and checked on latest version, if you see the code you have set the sales invoice's grand total amount to the first mode of payment. Which will be work if original invoice has one mode of payment but it will not work properly if the invoice has multiple mode payments.

PR #16615, opened 2019-02-09
 feat(projects): auto-set employee and company in timesheet

barredterra:
#### Use-case
User creates a new `Timesheet`.

#### Old behavior
User has to enter his name and company manually.

#### New behavior
If there is an `Employee` with matching `user_id` for the session user, set `employee` and `company` automatically.

screencast

Also added semicolons, where missing.


sagarvora:
This seems like a good idea, what about the case when another person creates the Timesheet?

barredterra:
Thanks for your comment, @sagarvora .

If another person creates the Timesheet, he would have to click on the employee field and set the specific employee manually. 

At the moment everybody has to set his own employee ID when creating a timesheet. After this PR, setting the employee would only be needed in the special case you mentioned.

erpnext/projects/doctype/timesheet/timesheet.js
```diff
@@ -20,16 +20,27 @@ frappe.ui.form.on("Timesheet", {
...
 
 		frm.fields_dict['time_logs'].grid.get_field('project').get_query = function() {
 			return{
 				filters: {
 					'company': frm.doc.company
 				}
+			};
+		};
+
+		// set employee (and company) to the one that's currently logged in
```
rmehta: this will only be called the first time. If you want to add it to any new document, add it to `onload` or `refresh` trigger.

rmehta (changes requested):

erpnext/projects/doctype/timesheet/timesheet.js
barredterra: I couldn't find an example where this doesn't work as expected. Do you have one?

barredterra:
Closing this for now. I'll open a new PR after testing effects of different triggers.

PR #16623, opened 2019-02-10
feat: Chart of Accounts Importer

Zlash65:
coa-sync-1

<details>

<summary> Usage </summary>

<p>

- Select a Company for which you want to upload accounts. A valid company is one that has no transactions made for it else a validation message is thrown.
- Download the template in which you can enter the accounts details.
- Save the csv after entering the data and attach it using the `Attach` button.
- Pre-validations on accounts are run and a tree structure preview is generated.
    - Root type accounts should be atleast 4
    - Accounts that have no parent account are root accounts and they should have root type specified.
    - Root type must be one of these - ("Asset", "Liability", "Expense", "Income", "Equity")
    - Account type that are must - ["Cost of Goods Sold", "Depreciation", "Fixed Asset", "Payable", "Receivable", "Stock Adjustment"]
    - Account type for group accounts that are must - ["Bank", "Cash", "Stock"]
- If all validations are a pass, `Start Import` button can be seen else an indicator is set to notify the user that the csv has an error.
- Once import starts, progress bar is shown, wait till the progress bar hides itself and indicator shows the status of Import.
- `Reset` button can be used to reset all fields.

</p>

</details>


vrms:
I can't get over this error

> _"Please identify/create Account (Ledger) for type - Receivable , Fixed Asset , Payable , Depreciation , Cost of Goods Sold , Stock Adjustment"_

... even though my import file has these types (Receivable , Fixed Asset , Payable , Depreciation , Cost of Goods Sold , Stock Adjustment) in column E (Account Type)

image

Also there is missing the account numbers (which should exists without being mandatory.

deepeshgarg007 (approved):

PR #16625, opened 2019-02-11
please give permission to delete the Attachments

gamma05772:
please give restrictions on user to delete Attached files,if doctype is in Approved state.


sagarvora:
We are not considering merging `develop` into `master` as of now.

PR #16634, opened 2019-02-11
fix(stock_balance): Make balance qty and value columns more visible in the Stock Balance report

Alchez:
image


sagarvora:
Judging from the branch name, you intended it to be `hotfix`?

Alchez:
@sagarvora I have no strong opinion on this, it's pretty opinionated. Either target branch works :smile:

PR #16636, opened 2019-02-11
fix: Update documents to be fully billed / received / delivered if the amount / qty values based on it are zero

Alchez:
**Problem:**

When a Delivery Note return is made for a Sales Order while issuing a Credit Note, there's a field (`per_billed`) that indicates how much that order is billed, based on any linked invoices. And this field is what you check when you're trying to make the return.

However, for zero-dollar orders, that field never gets updated to a non-zero value, and the system thinks that the order hasn't been billed, which is incorrect.

**Solution:**

As a general rule, if the reference field (the Sales Order amount for this example) is zero, we set the target percentage field as 100% to indicate that it has been fulfilled. For the given example, we update the `per_billed` field in Delivery Note to 100% if it's a zero-dollar order.

### DIFF
--- a/erpnext/controllers/status_updater.py
+++ b/erpnext/controllers/status_updater.py
@@ -280,25 +280,52 @@ class StatusUpdater(Document):
 	def _update_percent_field(self, args, update_modified=True):
 		"""Update percent field in parent transaction"""
 
-		self._update_modified(args, update_modified)
-
 		if args.get('target_parent_field'):
-			frappe.db.sql("""update `tab%(target_parent_dt)s`
-				set %(target_parent_field)s = round(
-					ifnull((select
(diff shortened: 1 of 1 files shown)


sagarvora:
Hey, Travis is failing. This will also need some tests.

Alchez:
@sagarvora 

I've fixed the server-side tests (Travis is still breaking at a seemingly unrelated patch). However, Codacy's now warning about SQL injection, the fix for which led to the tests failing in the first place.

For example,

```python
frappe.db.sql("%(doctype)s query with %(arguments)s" % args)  # warns for SQL Injection
frappe.db.sql("%(doctype)s query with %(arguments)s", args)  # prevents SQL Injection, but tests fail
```

Not sure about the right fix. Any suggestions?

PR #16659, opened 2019-02-12
feat: Revalue Purchase Receipt Items from Purchase Invoice (v12)

SaiFi0102:
**Problem in one line**: Purchase Invoice does not update the stock valuation in Purchase Receipts.

**Problem**: If the tax amounts (included in valuation) in Purchase Invoice is different from it's Purchase Receipt tax amounts then although the Purchase Invoice will debit the Expense Account heads but the Expenses Included in Valuation credited in Purchase Receipts would not match the debits since Purchase Receipt does not know about the difference in values between the 2 documents.

**Solution**:
Add 2 additional fields in Purchase Receipts Item:
* invoiced_item_tax_amount: Value of item_tax_amount calculated from Purchase _Invoice_ Items
* invoiced_item_tax_portion: Ratio of the Qty of Purchase Invoice Item to the Qty of Purchase Receipt Item

Revalue Purchase Receipts on submitting or cancelling Purchase Invoice
* Call pr_docs.update_receipts_valuation on Submitting/Canceling Purchase Invoice
* update_receipts_valuation will recalculate the valuation rate considering the invoiced_item_tax_amount and if the Invoiced Qty is less than the Received Qty then it will distribute the tax amounts proportional to 
invoiced_item_tax_portion between PINV tax amount and PR tax amount.

## Purchase Receipt
prec

## Stock value before invoicing
stockvaluebeforeinvoice

## Purchase Invoice
pinv

## Stock value after invoicing
valueafterinv


sagarvora:
Why not raise a `frappe.prompt` (JS) instead of having the user find the checkbox?

SaiFi0102:
> Why not raise a `frappe.prompt` (JS) instead of having the user find the checkbox?

Because that will make it possible to see and search for Purchase Invoices that revalue their linked Purchase Receipts

sagarvora:
> Because that will make it possible to see and search for Purchase Invoices that revalue their linked Purchase Receipts

Not against the checkbox. Against having the user manually check it instead of clicking a button.

SaiFi0102:
@sagarvora Any idea how I can prompt the user from server side? The validation for revaluing Purchase Receipt is server-side.

sagarvora:
Socketio is one way:
https://github.com/resilient-tech/bank_integration/blob/274081bbd2a33e1f2dd18314dcab70ab681b98fe/bank_integration/public/js/payment_entry.js#L10

SaiFi0102:
@sagarvora I will work on making the confirmation dialog to replace the manual checking in another PR

PR #16673, opened 2019-02-14
feat(issue): Create and view Tasks from Issues

Alchez:
**Changes:**

- New "Make Task" button to create a Task from an issue.
- Pick up all issue details as well as project (if any linked to the customer) in the Task.
- Added a new Issue link in the Task doctype.
- When creating a new Issue, the system will now keep its description after a communication is added to it.
- New Issue dashboard to view all Tasks created against it.
- New Task dashboard to view all Timesheets and Expense Claims created against it.
- (Minor) Restructure the Issue and Task doctypes

<details>
<summary> Screenshots / GIFs </summary>

**Issue view:**

image

**Task view:**

image

</details>


erpnext/support/doctype/issue/issue.py
```diff
@@ -30,16 +34,18 @@ def validate(self):
 
 	def on_update(self):
 		# create the communication email and remove the description
```
sagarvora: This comment needs to be edited if we're keeping description.

I think the primary reason why description was removed was to avoid repetition as it would appear in the timeline anyway. Do we have a good reason to keep it now?

sagarvora (dismissed):

erpnext/support/doctype/issue/issue.py
sagarvora: Answering my own question, I guess you want to keep the description because a task may be created from it subsequently.

How about hiding the field on creating communication? This was we can access it and user needn't read the same thing twice.

erpnext/support/doctype/issue/issue.py
```diff
@@ -160,3 +179,18 @@ def has_website_permission(doc, ptype, user, verbose=False):
 def update_issue(contact, method):
 	"""Called when Contact is deleted"""
 	frappe.db.sql("""UPDATE `tabIssue` set contact='' where contact=%s""", contact.name)
+
+
+@frappe.whitelist()
+def make_task(source_name, target_doc=None):
+	def set_missing_values(source, target):
+		if not target.project:
+			target.project = frappe.db.get_value("Project", {"customer": source.customer})
```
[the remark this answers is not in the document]
Alchez: Hm, you're right. I'll let the user decide if this is a part of a project manually.

erpnext/support/doctype/issue/issue.py
```diff
@@ -160,3 +179,18 @@ def has_website_permission(doc, ptype, user, verbose=False):
...
+
+	doclist = get_mapped_doc("Issue", source_name, {
```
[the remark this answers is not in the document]
Alchez: Good point. Fixed. Although there's ~50 other instances where `doclist` is being used to store mapped docs. Maybe we need a fix for those as well, or get more clarity on why it's that way.

erpnext/support/doctype/issue/issue.py
Alchez: Descriptions and event timelines give separate pieces of information though.

You could keep updating the issue's description based on its communications, which would be necessary for the task as well.

I think it's fine if we have both visible to the user. What do you think?

sagarvora: I guess you're right.. 

@rmehta, @nabinhait Thoughts?

rmehta:
Instead of a "View" button, please add a "Dashboard" to issue, so you can view multiple tasks linked to it (and other documents, like timesheets etc)

Alchez:
@rmehta 

Updated with new dashboards in Issue and Task.

PR #16677, opened 2019-02-14
fix: Consider parenttype while updating total qty in transactions

nabinhait:


sagarvora:
How about this instead?

```
'''
SELECT
parent, SUM(qty) as qty
IFROM
`tab{dt} Item`
where parenttype='{dt}'
GROUP BY parent
'''.format(doctype=doctype)
```

Using `.format` and `%s` together can get confusing.

PR #16680, opened 2019-02-15
feat: Auditors print formats

Anurag810:
**For Sales-Invoice:**

sales_invoice

 **For Purchase-Invoice:**

purchase_invoice

**For Journal-Entry:**

journal_entry

**For Payment-Entry:**

payment_entry


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -52,6 +52,13 @@ def on_submit(self):
 		self.update_loan()
 		self.update_inter_company_jv()
 
+	def before_print(self):
+		gl_entries = frappe.get_list("GL Entry",filters={"voucher_type": "Journal Entry",
```
sagarvora: why create another variable?
```suggestion
		self.gl = frappe.get_list("GL Entry",filters={"voucher_type": "Journal Entry",
```

sagarvora: Also, maybe name it `self.gl_entries`?

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -70,6 +70,14 @@ def on_submit(self):
 		self.update_advance_paid()
 		self.update_expense_claim()
 
+	def before_print(self):
+		gl_entries = frappe.get_list("GL Entry",filters={"voucher_type": "Payment Entry",
+			"voucher_no": self.name} ,
+			fields=["account", "party_type", "party", "debit", "credit", "remarks"]
+		)
+		print(gl_entries)
```
sagarvora: Remove.

erpnext/accounts/print_format/bank_and_cash_payment_voucher/bank_and_cash_payment_voucher.html
```diff
@@ -0,0 +1,82 @@
+{%- from "templates/print_formats/standard_macros.html" import add_header -%}
+<style>
+    .table-bordered td.top-bottom {border-top: none !important;border-bottom: none !important;}
+    .table-bordered td.right{border-right: none !important;}
+    .table-bordered td.left{border-left: none !important;}
```
sagarvora: why are we using `table-bordered` class if we don't want the border?

erpnext/accounts/print_format/bank_and_cash_payment_voucher/bank_and_cash_payment_voucher.html
```diff
@@ -0,0 +1,82 @@
...
+            </tr>
+            {% endif %}
+            {% endfor %}
+            <tr>
+                    <td class="top-bottom" colspan="4"> </td>
+            </tr>
+            <tr>
+                    <td class="top-bottom" colspan="5"><strong>Debit</strong></td>
+            </tr>
+            {% set total_debit = 0 -%}
```
sagarvora: Why not do this in `before_print`?

sagarvora (changes requested):

PR #16704, opened 2019-02-18
feat: Auditors print formats

Anurag810:
**For Sales-Invoice:**

sales_invoice

 **For Purchase-Invoice:**

purchase_invoice

**For Journal-Entry:**

journal_entry

**For Payment-Entry:**

payment_entry

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -52,6 +52,13 @@ class JournalEntry(AccountsController):
 		self.update_loan()
 		self.update_inter_company_jv()
 
+	def before_print(self):
+		gl_entries = frappe.get_list("GL Entry",filters={"voucher_type": "Journal Entry",
+			"voucher_no": self.name} ,
+			fields=["account", "party_type", "party", "debit", "credit", "remarks"]
+		)
+		self.gl = gl_entries
+
 	def get_title(self):
 		return self.pay_to_recd_from or self.accounts[0].account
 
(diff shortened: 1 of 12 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -52,6 +52,13 @@ def on_submit(self):
 		self.update_loan()
 		self.update_inter_company_jv()
 
+	def before_print(self):
+		gl_entries = frappe.get_list("GL Entry",filters={"voucher_type": "Journal Entry",
+			"voucher_no": self.name} ,
+			fields=["account", "party_type", "party", "debit", "credit", "remarks"]
+		)
+		self.gl = gl_entries
```
rohitwaghchaure: No need of gl_entries variable, assign the value of frappe.get_list directly to self.gl_entries instead of self.gl

Anurag810: done

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -70,6 +70,12 @@ def on_submit(self):
 		self.update_advance_paid()
 		self.update_expense_claim()
 
+	def before_print(self):
+		self.gl = frappe.get_list("GL Entry",filters={"voucher_type": "Payment Entry",
```
rohitwaghchaure: No one can understand the meaning of gl, make it gl_entries (self.gl_entries)

Anurag810: done

PR #16707, opened 2019-02-18
fix: POS batches issue on decrement function of qty

navdeepghai:
Batch number is chaning on exceution of qty function in sales_common.js even if batch number is already linked with item doc.

Because of change of  batch no. it's not able to find  cart item using [data-batch-no] attribute and throwing error.

> Removed all QTY  and item still showing in cart.
after-removing-all-items

> now there is not item in Sales Invoice  form but it's still showing in pos cart. If user try to increment/decrement the value then it's throwing  this error.
freeeze
console-error

<b>Batch Number should not be changed if it's already linked with pos cart item.</b>


navdeepghai1:
No need to merge this pull request  in develop. I will send new request for hotfix

PR #16720, opened 2019-02-19
fix(regional,italy): added state code on address, updated patch

gaurav-naik:
1. Added state code field in Address
2. Updated italian localization patch to set state code 
3. Updated xml to use state code

### DIFF
--- a/erpnext/patches/v11_0/make_italian_localization_fields.py
+++ b/erpnext/patches/v11_0/make_italian_localization_fields.py
@@ -3,12 +3,26 @@
 
 from __future__ import unicode_literals
 from erpnext.regional.italy.setup import make_custom_fields, setup_report
+from erpnext.regional.italy import state_codes
 import frappe
 
+
 def execute():
+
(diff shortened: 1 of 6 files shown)


erpnext/patches/v11_0/make_italian_localization_fields.py
```diff
@@ -5,10 +5,37 @@
 from erpnext.regional.italy.setup import make_custom_fields, setup_report
 import frappe
 
+state_codes = {'Siracusa': 'SR', 'Bologna': 'BO', 'Grosseto': 'GR', 'Caserta': 'CE', 'Alessandria': 'AL', 'Ancona': 'AN', 'Pavia': 'PV',
```
sagarvora: I think this belongs to the regional module instead. Does it work for fresh installs?

sagarvora: See https://github.com/frappe/erpnext/blob/develop/erpnext/regional/india/gst_state_code_data.json for example.

PR #16739, opened 2019-02-20
fix: remove mandatory for account field in bank account

rohitwaghchaure:


nabinhait:
@rohitwaghchaure why so?

asbasawaraj:
@nabinhait The link field to Account should not be mandatory. It can be mandatory when 'Is Company Account' is checked. If I create a record for a supplier's bank account, I will not add a ledger for that.

PR #16743, opened 2019-02-21
fix(projects): Auto-set employee and company in Timesheet

barredterra:
#### Use-case
User creates a new Timesheet.

#### Old behavior
At the moment, when creating a Timesheet, the user has to enter his name and company manually.

#### New behavior
If there is an `Employee` with matching `user_id` for the session user, set `employee` and `company` automatically.

If another person creates the Timesheet, she would have to click on the employee field and set the specific employee manually. Setting the employee will only be necessary in this special case.

screencast

Also added semicolons, where missing.


erpnext/projects/doctype/timesheet/timesheet.js
```diff
@@ -20,33 +20,37 @@ frappe.ui.form.on("Timesheet", {
...
+			};
+		};
 	},
 
 	onload: function(frm){
 		if (frm.doc.__islocal && frm.doc.time_logs) {
 			calculate_time_and_amount(frm);
 		}
+
+		if (frm.doc.docstatus < 1) {
```
nabinhait: It should only set employee and company if document is unsaved.

`if(frm.is_new())`

nabinhait (dismissed):

barredterra:
Thanks for the hint, done. @nabinhait

barredterra:
Anything else I can do to get this merged, @nabinhait?

barredterra:
@sagarvora did anything happen to Nabin?

sagarvora (dismissed):

erpnext/projects/doctype/timesheet/timesheet.js
```diff
@@ -294,4 +298,17 @@ var calculate_time_and_amount = function(frm) {
...
+};
+
+// set employee (and company) to the one that's currently logged in
+const set_employee_and_company = function(frm) {
+	const options = { user_id: frappe.session.user };
+	const fields = ['name', 'company'];
+	frappe.db.get_value('Employee', options, fields).then(({ message }) => {
+		if (message) {
+			// there is an employee with the currently logged in user_id
+			frm.fields_dict['employee'].set_value(message.name);
```
sagarvora: Let's use `frm.set_value` for consistency instead?
https://github.com/frappe/erpnext/pull/16743/files#diff-bb8ecad0277db241f5eef7705751b0ceR300

barredterra: Okay, but only if you promise to get this merged within a week :stuck_out_tongue_winking_eye:

sagarvora (approved):

PR #16749, opened 2019-02-21
fix(payments): Only check for Draft or Submitted payments while cancelling payment requests

Alchez:
Trying to cancel a Payment Request with linked Payment Entries would throw an error, even if they were all in the Cancelled state. This adds a check to ignore Cancelled payments.

### DIFF
--- a/erpnext/accounts/doctype/payment_request/payment_request.py
+++ b/erpnext/accounts/doctype/payment_request/payment_request.py
@@ -214,11 +214,10 @@ class PaymentRequest(Document):
 
 	def check_if_payment_entry_exists(self):
 		if self.status == "Paid":
-			payment_entry = frappe.get_all("Payment Entry Reference",
+			if frappe.get_all("Payment Entry Reference",
 				filters={"reference_name": self.reference_name, "docstatus": ["<", 2]},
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -214,9 +214,11 @@ def set_as_cancelled(self):
 
 	def check_if_payment_entry_exists(self):
 		if self.status == "Paid":
-			payment_entry = frappe.db.sql_list("""select parent from `tabPayment Entry Reference`
-				where reference_name=%s""", self.reference_name)
-			if payment_entry:
+			payment_entry = frappe.get_all("Payment Entry Reference",
+				filters={"reference_name": self.reference_name, "docstatus": ["<", 2]},
+				fields=["distinct(parent)"])
```
sagarvora: TIL we can do this.
No need for distinct though, if it's only being used for validation (increases query time).

Alchez: Actually, `distinct` performs much better here. It's more readable, but marginally slower than `group by`.

image

PR #16762, opened 2019-02-24
fix: Selling and buying amount precision fix

deepeshgarg007:


sagarvora:
Hotfix ==> Develop on release. Closing.

PR #16764, opened 2019-02-24
fix: Selling and buying amount precision fix

deepeshgarg007:

### DIFF
--- a/erpnext/accounts/report/gross_profit/gross_profit.py
+++ b/erpnext/accounts/report/gross_profit/gross_profit.py
@@ -161,7 +161,7 @@ class GrossProfitGenerator(object):
 						new_row.qty += row.qty
 						new_row.buying_amount += flt(row.buying_amount, currency_precision)
 						new_row.base_amount += flt(row.base_amount, currency_precision)
-				new_row = self.set_average_rate(new_row, currency_precision, float_precision)
+				new_row = set_average_rate(new_row, currency_precision, float_precision)
 				self.grouped_data.append(new_row)
 			else:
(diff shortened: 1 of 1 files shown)


sagarvora:
Don't we want to fetch the float precision from `System Settings` instead?

deepeshgarg007:
@sagarvora Yes, currency_precision and float_precision should be fetched form system settings.
I have made the requested changes.

erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -106,11 +106,14 @@ def process(self):
 		self.grouped = {}
 		self.grouped_data = []
 
+		currency_precision = cint(frappe.db.get_default("currency_precision")) or 3
```
Alchez: Maybe set these as global constants in the file? You won't need to pass it as an argument.

erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -349,3 +344,11 @@ def load_product_bundle(self):
 	def load_non_stock_items(self):
 		self.non_stock_items = frappe.db.sql_list("""select name from tabItem
 			where is_stock_item=0""")
+
+def set_average_rate(new_row, currency_precision, float_precision):
+		new_row.gross_profit = flt(new_row.base_amount - new_row.buying_amount, currency_precision)
```
Alchez: This is over-indented.

Alchez (changes requested):

netchampfaris (approved):

PR #16765, opened 2019-02-24
fix: Get mode of payment while selecting payment term in payment term template

surajshetty3416:


sagarvora:
In you PR title, you mean `mode of payment` instead?

Issue #16635, opened 2019-02-11
Duplicate Payment Term at row 2 when creating payment entry template


joshiparthin:
When creating a payment entry template getting this error when row 2 is different: 

![image](https://user-images.githubusercontent.com/6594877/52557288-f5d34c80-2e14-11e9-85af-38d9b8f1fac9.png)

This is in v11 (master).

kanssfer:
The issue is about credit days. The credit days in both the rows were same. Message needs to be more Specific... Kindly change the error message mechanics...

joshiparthin:
The above issue occurs because credit days were same in both the rows. Please put appropriate error message.

joshiparthin:
Added a draft PR https://github.com/frappe/erpnext/pull/16770

doniafrizal:
V12 also same

ryuliantoro:
> The above issue occurs because credit days were same in both the rows. Please put appropriate error message.

I think it should be allowed to be the same. So the error is not the message but on the logic.

PR #16770, opened 2019-02-25
Fixing Error Message for Issue #16635 

joshiparthin:
Just a change in the message to bring better context for the issue
https://github.com/frappe/erpnext/issues/16635


doniafrizal:
This issue still happen on v12

ryuliantoro:
Too bad this issue is automatically closed. It is a useful feature yet the bug is disturbing a lot so it is almost unusable at all.

PR #16783, opened 2019-02-26
fix: set warehouse from def_warehouse_for_sales_return in sales return

sahil28297:
If defalut_warehouse_for_sales_return is set in Company doctype then in sales return, the value of warehouse will be fetched from it.

default_warehouse_for_sales_return


neilLasrado:
Please also take into consideration this PR merged to develop -> https://github.com/frappe/erpnext/pull/15491

nabinhait (changes requested):
@sahil28297 Please fix the conflict

nabinhait (approved):

nabinhait:
@Anurag810 Please fix the conflict

nabinhait (dismissed):

nabinhait (approved):

PR #16813, opened 2019-02-28
fix(regional,italy): replace decode with safe_decode

gaurav-naik:
The following error was encountered while updating the bench in a python 3.6.7 environment.

decode_error_trace

Replaced the offending `x.decode('utf-8')` with `frappe.safe_decode(x, encoding='utf-8')`


gaurav-naik:
Changes included in #16817
Closing.

PR #16860, opened 2019-03-07
Fix:for Attendance gantt change the field map

farqalit:
For Attendance Gantt Change the field map

### DIFF
--- a/erpnext/hr/doctype/attendance/attendance_calendar.js
+++ b/erpnext/hr/doctype/attendance/attendance_calendar.js
@@ -16,3 +16,20 @@ frappe.views.calendar["Attendance"] = {
 	},
 	get_events_method: "erpnext.hr.doctype.attendance.attendance.get_events"
 };
+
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/attendance/attendance_calendar.js
```diff
@@ -15,4 +15,21 @@ frappe.views.calendar["Attendance"] = {
 		}
 	},
 	get_events_method: "erpnext.hr.doctype.attendance.attendance.get_events"
-};
\ No newline at end of file
+};
+
+frappe.views.calendar["AttendanceGantt"] = {
```
netchampfaris: AttendanceGantt is not a valid doctype. You should not create another configuration for calendar, just override the above one.

netchampfaris (changes requested):

farqalit:
I tried to override that in .py
But it doesn't work as it needs start and end time which can only be
override in combined js which you used in calendar as well.
So please guide what is the proper way to override that issue ?

PR #16887, opened 2019-03-10
GSTR1 json format download 

deepeshgarg007:


karthikeyan5 (changes requested):
- [x] Bug1: the 'fp' field in json is always the current month. it should take the selected period in to account when populating 'fp'.
- [x] Bug2: when an invoice has items with more than one rate of tax, each rate of tax item gets a seperate invoice object in the array. it should be populated as array of 'itms' instead.(see screenshots below for current and expected behavior)

image
Current Behavior.


image
Expected behavior.

karthikeyan5 (approved):

PR #16895, opened 2019-03-11
fix(transaction): Fix shipping rule charge override by tax template

Alchez:
**Before:**

before-charge-fix

**After:**

after-charge-fix


Alchez:
@nabinhait, this was also made to hotfix at #16896.

PR #16896, opened 2019-03-11
fix(transaction): Fix shipping rule charge override by tax template

Alchez:
**Before:**

before-charge-fix

**After:**

after-charge-fix


Alchez:
@nabinhait, this might need a different fix, as done for develop at #17210.

rohitwaghchaure:
@Alchez due to this fix same tax adding multiple times in POS
Will add https://github.com/frappe/erpnext/pull/17210 in hotfix

PR #16950, opened 2019-03-18
Fix(Delivery Note): Fix issue credit note

hrwX:
Remove Issue Credit Note from Delivery Note


rmehta:
@hrwX explain the fix

hrwX:
> @hrwX explain the fix

The Issue Credit Note doesn't work and technically the return should be done from sales invoice and not delivery note

YY2naJc

neilLasrado:
This is not exactly a fix; rather than a feature revert.

Currently, in erpnext, we can make returns either against the Delivery Note or against the Sales Invoice 

Returns can be broadly classified into 2 types - 

A. The customer wants the product to be replaced 

B. The customer needs a refund 

In case the return is of type A, a  return delivery note is made against the delivery note to receive the inventory and a new delivery note is created against the sales order when the product is redelivered

In case the return is of type B, a  return delivery note is made against the delivery note to receive the inventory and a return sales invoice is raised (credit note) 

Now especially for large orders with returns its not possible to manually create a return sales invoice from the delivery note.

@hrwX in case you are facing an issue, we need to find the cause of the error and fix that rather than reverting the feature

PR #16984, opened 2019-03-22
fix: Show consolidated amount of payment entries in Bank reconciliation

surajshetty3416:
A bank account can have credit and debit transaction with the same reference number.
Screenshot 2019-03-22 at 2 42 34 PM

Previously, while getting entries for reconciliation, we used to get only the sum of credit or debit.
Screenshot 2019-03-22 at 2 43 41 PM

This fix gets the amount after consolidating the credit and debit transactions.
Screenshot 2019-03-22 at 2 40 04 PM


surajshetty3416:
duplicate of https://github.com/frappe/erpnext/pull/17004

PR #17000, opened 2019-03-24
fix: Task overdue status propagates to project

scmmishra:
## Issue
The task of the Project shows Open as the status but when opened from Task List, it says Overdue.
The status of the Tasks are not fetching.
LbOii6N

## Changes Made
- Used ORM instead of a SQL for set_task_as_overdue
- update_status function compares the time and marks the appropriate task as overdue
- The on_update hook makes the changes in project as well

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -160,7 +160,7 @@ class Task(NestedSet):
 		self.update_nsm_model()
 
 	def update_status(self):
-		if self.status not in ('Cancelled', 'Closed'):
+		if self.status not in ('Cancelled', 'Closed') and self.exp_end_date:
 			from datetime import datetime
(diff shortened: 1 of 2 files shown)


erpnext/projects/doctype/task/task.py
```diff
@@ -159,6 +159,14 @@ def on_trash(self):
 
 		self.update_nsm_model()
 
+	def update_status(self):
+		if self.status not in ('Cancelled', 'Closed') and self.exp_end_date:
+			from datetime import datetime
+			if self.exp_end_date < datetime.now().date():
+				print(self.subject)
+				self.status = 'Overdue'
```
scmmishra: Use self.set_db and call `update_project` manually

adityahase (approved):

PR #17006, opened 2019-03-25
fix: Ignore root company validation setting in company master

deepeshgarg007:
User is also not allowed to change the Account currency in the child companies CoA manually, It gives this error message even if the account exists in parent company.

Screenshot 2019-03-25 at 11 37 24 AM

A checkbox is added in the company master under account settings to allow user to ignore root company validation
Screenshot 2019-03-25 at 10 13 54 PM


erpnext/setup/doctype/company/company.json
```diff
@@ -1414,6 +1457,40 @@
...
+   "fieldname": "ignore_root_company_validation", 
+   "fieldtype": "Check", 
+   "hidden": 0, 
+   "ignore_user_permissions": 0, 
+   "ignore_xss_filter": 0, 
+   "in_filter": 0, 
+   "in_global_search": 0, 
+   "in_list_view": 0, 
+   "in_standard_filter": 0, 
+   "label": "Ignore Root Company Validation",
```
nabinhait: @deepeshgarg007 Very bad naming. How the users will know what is the "Root Company Validation". 
May be "Allow Account Creation against Child Company" will be a better name.

nabinhait (changes requested):

PR #17018, opened 2019-03-26
fix: Addition of third party types in FEC

chdecultot:
Hi,

In the FEC report, we only fetched the customer and supplier party types, but in some cases, Employee, Student or Member are also required.
I have therefore added theses three party types to the report.


chdecultot:
Up

nabinhait (approved):

PR #17020, opened 2019-03-26
fix: offline pos, duplicate records creating

rohitwaghchaure:


sunhoww:
Hi, are there any particular reasons **ID**s for `customer_pos_id` and `offline_pos_name` are now generated by `frappe.datetime.now_datetime` instead of `$.now`? I was concerned about this because the **ID**s are now literal datetime strings like `2019-05-19 18:39:55` instead of the earlier epoch time Integer like `1558280400090`.

PR #17023, opened 2019-03-26
fix(Salary Slip): Consider Leave without Pay for calculation

hrwX:

### DIFF
--- a/erpnext/hr/doctype/salary_slip/salary_slip.py
+++ b/erpnext/hr/doctype/salary_slip/salary_slip.py
@@ -441,7 +441,7 @@ class SalarySlip(TransactionBase):
 	def calculate_net_pay(self):
 		if self.salary_structure:
 			self.calculate_component_amounts()
-		
+
 		disable_rounded_total = cint(frappe.db.get_value("Global Defaults", None, "disable_rounded_total"))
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -452,10 +452,10 @@ def calculate_net_pay(self):
 
 		self.set_loan_repayment()
 
-		self.net_pay = flt(self.gross_pay) - (flt(self.total_deduction) + flt(self.total_loan_repayment))
+		self.net_pay = (flt(self.gross_pay) - (flt(self.total_deduction) + flt(self.total_loan_repayment))) * flt(self.payment_days / self.total_working_days)
```
Alchez: Might need a check for total working days being greater than zero, to avoid a zero-division error.

Alchez (changes requested):

erpnext/hr/doctype/salary_slip/salary_slip.py
hrwX: total working days is always going to be number of days in a month and can never be zero

Alchez: Ah, looks good then.

netchampfaris (approved):

PR #17037, opened 2019-03-28
Master

pratikcws:
In all link field when we filter records via code
if you type something that doesn't exist , the field will automatically be cleared
but if you type something that exist in database but not available in filter record , just not allowed to select, it would not be cleared

Example, if i knew the full name from another company "All Warehouses - Do"
I can type it on the warehouse field, and the system doesn't clear the field out when the input field loose focus..
so we change code  
/home/frappe/frappe-bench/apps/frappe/frappe/public/js/frappe/form/controls/link.js

[link.zip](https://github.com/frappe/erpnext/files/3017039/link.zip)

on line 167 : 
if(r.results.length==0){
     frappe.model.set_value(me.df.parent, me.df.name, me.df.fieldname, "", me.df.fieldtype);
    refresh_field(me.df.fieldname);
}


netchampfaris:
@pratikcws Hey, you should open this PR against the `hotfix` branch, also rebase your branch with `hotfix`.

nabinhait:
@pratikcws The branches are not correct in your PR. Please fix it and reopen a new PR.

PR #17079, opened 2019-04-01
syntax error causing issue upon install

tukutela:
using ubuntu 18.04, python3, this syntax error causing installation failure.


rmehta (changes requested):
Python 3 support is v11 only. Even if we fix this, there are many other places this is likely to break.

PR #17083, opened 2019-04-01
Revert "fix(Salary Slip): Consider Leave without Pay for calculation"

nabinhait:
Reverts frappe/erpnext#16926

This is a wrong fix, earning and deductions should also be calculated based on LWP, not only net pay.

@hrwX , @netchampfaris


Alchez:
@nabinhait, you may also want to revert the PR for develop then - #17023.

PR #17087, opened 2019-04-01
Party type validation in payment entry

deepeshgarg007:
Any doctype can be entered in party type since it is a dynamic link.A server error occurs if user enters a doctype which not handeled in the backend. So restricting part type to doctypes only in frappe.boot.party_account_types


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -232,6 +232,13 @@ frappe.ui.form.on('Payment Entry', {
 	},
 
 	party_type: function(frm) {
+
+		let party_types = Object.keys(frappe.boot.party_account_types);
+		if(frm.doc.party_type && !party_types.includes(frm.doc.party_type)){
+			frm.set_value("party_type", "");
+			frappe.throw(__("Party can only be one of "+ party_types.join(", ")));
```
nabinhait: Instead of giving this message, let's guide user that they need to add a new record in Party Type master first.

nabinhait (changes requested):

PR #17097, opened 2019-04-01
fix: Make Customer and lead field dynamic in quotations and opportunity

deepeshgarg007:
There are two seperate fields for Customer and Lead which causes permission issues.
Using dynamic link to resolve this.


erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -43,16 +42,9 @@ def has_sales_order(self):
...
-		if self.customer:
-			self.quotation_to = "Customer"
-			self.lead = None
-		elif self.lead:
-			self.quotation_to = "Lead"
-
 	def update_lead(self):
-		if self.lead:
-			frappe.get_doc("Lead", self.lead).set_status(update=True)
+		if self.quotation_to == "Lead":
```
Alchez: Also check if `self.customer_lead` is entered, otherwise `get_doc` will break.

Alchez (changes requested):

erpnext/selling/doctype/customer/customer.py
```diff
@@ -104,9 +104,8 @@ def update_lead_status(self):
 		if self.lead_name:
 			frappe.db.set_value('Lead', self.lead_name, 'status', 'Converted', update_modified=False)
 
-			for doctype in ('Opportunity', 'Quotation'):
-				for d in frappe.get_all(doctype, {'lead': self.lead_name}):
-					frappe.db.set_value(doctype, d.name, 'customer', self.name, update_modified=False)
+		for d in frappe.get_all('Opportunity', {'lead': self.lead_name}):
```
Alchez: This should be inside the if check. If there is no lead name, the `get_all` will return ALL opportunities.

Alchez (dismissed):

erpnext/crm/doctype/opportunity/opportunity.js
```diff
@@ -75,13 +81,25 @@ frappe.ui.form.on("Opportunity", {
...
-		} else if(frm.doc.lead) {
+		} else if(frm.doc.opportunity_from == "Lead" && frm.doc.customer_lead) {
 			frappe.dynamic_link = {doc: frm.doc, fieldname: 'lead', doctype: 'Lead'}
 		}
 	},
 
+	set_dynamic_field_label: function(frm){
+		if (frm.doc.opportunity_from == "Customer")
+		{
+			frm.set_df_property("customer_lead", "label", "Customer");
```
netchampfaris:
```suggestion
			frm.set_df_property("customer_lead", "label", frm.doc.opportunity_from);
```

erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -29,7 +29,7 @@ def validate(self):
 
 		self.make_new_lead_if_required()
 
-		if not self.enquiry_from:
+		if not self.opportunity_from:
```
netchampfaris: Make the field mandatory, no need to check here.

erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -137,10 +137,10 @@ def has_lost_quotation(self):
 			return True
 
 	def validate_cust_name(self):
-		if self.customer:
-			self.customer_name = frappe.db.get_value("Customer", self.customer, "customer_name")
-		elif self.lead:
-			lead_name, company_name = frappe.db.get_value("Lead", self.lead, ["lead_name", "company_name"])
+		if self.customer_lead and self.opportunity_from == 'Customer':
+			self.customer_name = frappe.db.get_value("Customer", self.customer_lead, "customer_name")
```
netchampfaris: Use fetch_from

erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -188,16 +188,13 @@ def validate_item_details(self):
 				if not d.get(key): d.set(key, item.get(key))
 
 	def validate_lead_cust(self):
-		if self.enquiry_from == 'Lead':
-			if not self.lead:
+		if self.opportunity_from == 'Lead':
+			if not self.customer_lead:
```
netchampfaris: Make the field mandatory, no need to check here

erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -219,8 +216,8 @@ def set_missing_values(source, target):
 		quotation = frappe.get_doc(target)
 
 		company_currency = frappe.get_cached_value('Company',  quotation.company,  "default_currency")
-		party_account_currency = get_party_account_currency("Customer", quotation.customer,
-			quotation.company) if quotation.customer else company_currency
+		party_account_currency = get_party_account_currency("Customer", quotation.customer_lead,
```
netchampfaris: Check if Quotation is from Customer

erpnext/selling/doctype/customer/customer.py
```diff
@@ -104,9 +104,8 @@ def update_lead_status(self):
 		if self.lead_name:
 			frappe.db.set_value('Lead', self.lead_name, 'status', 'Converted', update_modified=False)
 
-			for doctype in ('Opportunity', 'Quotation'):
-				for d in frappe.get_all(doctype, {'lead': self.lead_name}):
-					frappe.db.set_value(doctype, d.name, 'customer', self.name, update_modified=False)
+		for d in frappe.get_all('Opportunity', {'customer_lead': self.lead_name}):
```
netchampfaris: Why exclude Quotation?

netchampfaris (changes requested):

netchampfaris (approved):

PR #17119, opened 2019-04-03
fix: sales order status for order type 'Maintenance'

Anurag810:
For sales order If Order type i s maintenance then no need to set the docstatus to 'To deliver and bill'.
So,  updated status to 'To BILL'.


deepeshgarg007:
@Anurag810 update the description or add some screenshots

deepeshgarg007 (changes requested):
@Anurag810 Please fix test case and add description

deepeshgarg007 (approved):

PR #17124, opened 2019-04-03
fix: payroll entry child table(Payroll employee details) remove read only

Anurag810:


nabinhait:
@Anurag810 If you make the employee table editable, then you have to fix the server logic as well. In server side logic, it should always take from child table, currently it get employees based on filters.

PR #17132, opened 2019-04-04
fix: Fix validation message for re-order entry

kennethsequeira:
Duplicating https://github.com/frappe/erpnext/pull/17072 for develop branch


nabinhait:
already fixed and merged

PR #17149, opened 2019-04-05
fix(India): run patch for GST custom fields

hrwX:
run patch for newly added GST custom fields

### DIFF
--- a/erpnext/patches/v10_0/gst_hsn_fixes.py
+++ b/erpnext/patches/v10_0/gst_hsn_fixes.py
@@ -1,0 +1,14 @@
+import frappe
+from erpnext.regional.india.setup import setup
+from frappe.custom.doctype.custom_field.custom_field import create_custom_fields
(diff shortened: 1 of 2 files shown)


deepeshgarg007 (changes requested):
@hrwX please check if the country of the the company is India and then only create the custom field

erpnext/patches/v10_0/gst_hsn_fixes.py
```diff
@@ -0,0 +1,14 @@
+import frappe
+from erpnext.regional.india.setup import setup
+from frappe.custom.doctype.custom_field.custom_field import create_custom_fields
+
+def execute():
```
nabinhait: The patch should only run if any of the company belongs to India

nabinhait (changes requested):

nabinhait (approved):

PR #17154, opened 2019-04-05
fix: Calculate rate based on discount on server side, only if not rate or pricing rule applied

nabinhait:
**Before:** 
4le3vg9

**After:**

item-discount


rohitwaghchaure:
Travis is failing

PR #17161, opened 2019-04-08
fix(woocommerce integration): 403 error and adding defaults

karthikeyan5:
- Fixed 403 and 417 error being thrown by the method `erpnext.erpnext_integrations.connectors.woocommerce_connection.order`

- added defaults for "Creation User", "Warehouse", "Sales Order Series" and "UOM"
image


erpnext/patches/v11_1/woocommerce_set_creation_user.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	woocommerce_setting_enable_sync = frappe.db.sql("SELECT t.value FROM tabSingles t WHERE doctype = 'Woocommerce Settings' AND field = 'enable_sync'",  as_dict=True)
+	if len(woocommerce_setting_enable_sync) and woocommerce_setting_enable_sync[0].value == '1':
+		frappe.db.sql("""UPDATE tabSingles
+					SET value = (SELECT t.value FROM tabSingles t WHERE doctype = 'Woocommerce Settings' AND field = 'modified_by')
+					WHERE doctype = 'Woocommerce Settings'
+					AND field = 'creation_user';""")
```
rohitwaghchaure: For better readability, use ORM to update the value

PR #17188, opened 2019-04-09
Stock Balance Defaults UOM

lelsheri:
Stock Balance Defaults UOM
update and rename last report from 'Stock Balance Multi-UOM' to 'Stock Balance Defaults UOM'


lelsheri (commented):
pls. review ASAP.

rmehta (changes requested):
Code repetition. This duplicates code from `stock_balance.py`, import and use common functions.

lelsheri:
> Code repetition. This duplicates code from `stock_balance.py`, import and use common functions.
The Application does not have a report for stock by batch number , So I Think is not repeating 
pls. review carefully , it is 2 reports not one.
Regards,
Luay

PR #17210, opened 2019-04-11
fix: Taxes and charges table not clear the existing rows if the template is changed

rohitwaghchaure:
**Issue**

Taxes and charges table not removes the rows after selection of new taxes and charges template

user_has_to_remove_the_taxes_row_manually

**After fix**
after_fix


nabinhait (approved):

Alchez:
@nabinhait @rohitwaghchaure 

Should this be made to hotfix as well?

PR #17252, opened 2019-04-16
fix(Customer): validate percentage total

hrwX:
Percentage should not be more than 100 in sales team child table


rmehta (dismissed):
All validations must be server side

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#all-business-logic-and-validations-must-be-on-server-side

erpnext/selling/doctype/customer/customer.py
```diff
@@ -60,6 +60,12 @@ def validate(self):
 			if self.loyalty_program == customer.loyalty_program and not self.loyalty_program_tier:
 				self.loyalty_program_tier = customer.loyalty_program_tier
 
+		allocated_percentage = 0
+		for percentage in self.sales_team:
+			allocated_percentage += percentage.allocated_percentage
+			if allocated_percentage > 100:
```
nabinhait: I think it must be equal to 100. Any use case for less than 100?

nabinhait (dismissed):

nabinhait (approved):

PR #17254, opened 2019-04-16
fix(woocommerce): fix patch failure on some sites

karthikeyan5:
MYSQL error: 1093 was occurring on some sites when running this patch.

### DIFF
--- a/erpnext/patches/v11_1/woocommerce_set_creation_user.py
+++ b/erpnext/patches/v11_1/woocommerce_set_creation_user.py
@@ -2,9 +2,18 @@ from __future__ import unicode_literals
 import frappe
 
 def execute():
-	woocommerce_setting_enable_sync = frappe.db.sql("SELECT t.value FROM tabSingles t WHERE doctype = 'Woocommerce Settings' AND field = 'enable_sync'",  as_dict=True)
+	woocommerce_setting_enable_sync = frappe.db.sql("""SELECT t.value
+		FROM tabSingles t
+		WHERE doctype = 'Woocommerce Settings'
+		AND field = 'enable_sync'""", as_dict=True)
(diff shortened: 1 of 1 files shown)


erpnext/patches/v11_1/woocommerce_set_creation_user.py
```diff
@@ -4,7 +4,13 @@
 def execute():
 	woocommerce_setting_enable_sync = frappe.db.sql("SELECT t.value FROM tabSingles t WHERE doctype = 'Woocommerce Settings' AND field = 'enable_sync'",  as_dict=True)
 	if len(woocommerce_setting_enable_sync) and woocommerce_setting_enable_sync[0].value == '1':
-		frappe.db.sql("""UPDATE tabSingles
-					SET value = (SELECT t.value FROM tabSingles t WHERE doctype = 'Woocommerce Settings' AND field = 'modified_by')
-					WHERE doctype = 'Woocommerce Settings'
-					AND field = 'creation_user';""")
\ No newline at end of file
+		modified_by = frappe.db.sql("""SELECT t.value
```
rmehta: please no hanging indents

https://github.com/frappe/erpnext/wiki/Coding-Standards#use-simple-indenting

rmehta (changes requested):

erpnext/patches/v11_1/woocommerce_set_creation_user.py
karthikeyan5: it does look strange on github... below is how it looked on vscode:
hanging_SQL

I have updated it now...

karthikeyan5:
already fixed by https://github.com/frappe/erpnext/pull/17250

PR #17269, opened 2019-04-16
feat: woocommerce integration - added product and client webhooks

gamaraan:
Added to woocommerce integration the possibility to have items and clients webhooks, so new item/item updates and new client/client updates events can be synced to ERPNext not only on new orders, but as they occur in woocommerce site.
Webhooks are set up in a similar way with the main order created webhook:

-  for new item/item updated events, pointing to the endpoint your.erp.site/api/method/erpnext.erpnext_integrations.connectors.woocommerce_connection.item
- new client/client updated, pointing to the endpoint your.erp.site/api/method/erpnext.erpnext_integrations.connectors.woocommerce_connection.client

using same secret as set up on new order webhook.


erpnext/erpnext_integrations/connectors/woocommerce_connection.py
```diff
@@ -20,45 +24,44 @@ def verify_request():
 		not sig == bytes(frappe.get_request_header("X-Wc-Webhook-Signature").encode()):
 			frappe.throw(_("Unverified Webhook Data"))
 	frappe.set_user(woocommerce_settings.creation_user)
+	print ("Request verified successfully")
```
rmehta: print

rmehta (changes requested):
Please also update documentation for setting up webhooks

:+1: for first PR to ERPNext

erpnext/erpnext_integrations/connectors/woocommerce_connection.py
gamaraan: Should I edit the page here:
https://erpnext.com/docs/user/manual/en/erpnext_integration/woocommerce_integration
?

gamaraan (commented):
print removed

PR #17279, opened 2019-04-18
fix: Requested Items To Be Ordered report showing records even if material request is fully ordered

rohitwaghchaure:
**Issue**

Created material request with an Item ball pen which has stock uom as BOX, but created material request with UOM as Nos.
Screen Shot 2019-04-18 at 6 57 12 pm

Even after creating the fully purchase order against the material request "MREQ-00074-1" the material request is still showing in the report "Requested Items To Be Ordered"

Screen Shot 2019-04-18 at 6 57 30 pm

In the sql query I found that we are comparing the quantity with ordered quantity instead of stock quantity with ordered quantity

**After fix**

Updated quantity with stock quantity and issue is fixed
Screen Shot 2019-04-18 at 7 00 04 pm


nabinhait (changes requested):
For Qty column, you have used stock_qty but for pending qty it is taking qty. Why so?

rohitwaghchaure:
@nabinhait Fixed!

nabinhait (approved):

PR #17295, opened 2019-04-19
fix: Reopen button does not appear in delivery note

Anurag810:
Reopen button does not appear in delivery note when delivery note is closed.


erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -101,8 +101,7 @@ erpnext.stock.DeliveryNoteController = erpnext.selling.SellingController.extend(
 	refresh: function(doc, dt, dn) {
 		var me = this;
 		this._super();
-
-		if ((!doc.is_return) && (doc.status!="Closed" || doc.is_new())) {
+		if ((!doc.is_return) && (doc.status!="Closed" || doc.is_new)) {
```
deepeshgarg007: @Anurag810 doc object has no attribute is_new

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #17302, opened 2019-04-19
fix(site_sync): return more data in level

sahil28297:
DO NOT MERGE, DEPENDS ON OTHER PRs OF DIFFERENT REPOS


erpnext/utilities/activation.py
```diff
@@ -48,7 +48,27 @@ def get_level():
 	if frappe.db.sql('select name from tabUser where last_login > date_sub(now(), interval 2 day) limit 1'):
 		activation_level += 1
 
-	return activation_level
+	level = {"activation_level": 10, "sales_data": [
```
saurabh6790: Better way
```
activation_level = 0
sales_data = []
doctypes = {"Item": 5, "Sales Invoice": 2, "User": 5, ...}
for doctype, min_count in iteritems(doctypes):
      count = frappe.db.count(doctype)
      if count > min_count:
          activation_level += 1
      sales_data.append({doctype: count})
```

saurabh6790 (changes requested):

saurabh6790 (approved):

PR #17327, opened 2019-04-22
fix: (Patch) Removed page Bom-Browser

Anurag810:


erpnext/patches/v11_1/delete_bom_browser.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+    frappe.db.sql(""" Delete from `tabPage` where name='bom-browser' """)
```
netchampfaris: Use frappe.delete_doc

netchampfaris (changes requested):

erpnext/patches/v11_1/delete_bom_browser.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+    frappe.delete_doc('Page', 'bom-browser')
```
nabinhait: Use `frappe.delete_doc_if_exists`, otherwise it will fail for new instance where the page does not exist.

nabinhait (changes requested):

erpnext/patches/v11_1/delete_bom_browser.py
Anurag810: done

nabinhait (dismissed):

nabinhait (approved):

PR #17332, opened 2019-04-22
feat: Automatic inter company Sales and Purchase order

deepeshgarg007:
If a "Purchase Order" is created for company "A" with supplier as company "B":
System shall automatically create a Sales order for company "B" which customer is company "A", replicating the same data in both orders

Inter order


scmmishra:
Add a good description explaining the case of automatic inter company orders. Also update/add tests for the new class functions wherever possible

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -455,6 +460,42 @@ def validate_serial_no_based_delivery(self):
 				Item {0} is added with and without Ensure Delivery by \
 				Serial No.").format(item.item_code))
 
+def validate_inter_company_party(doctype, party, company, inter_company_order_reference):
```
nabinhait: These functions can be commonified with invoice

nabinhait (changes requested):

erpnext/selling/doctype/sales_order/sales_order.py
deepeshgarg007: Fixed

nabinhait (approved):

PR #17349, opened 2019-04-23
feat: Get invoiced item's gross margin using API

rohitwaghchaure:
Screen Shot 2019-04-23 at 6 32 03 pm


nabinhait (changes requested):
Add a filter of sales invoice in gross profit as well, the logic is already there.

nabinhait (approved):

PR #17359, opened 2019-04-24
fix: Item Attribute Order

netchampfaris:
The line removed overrides the order of values. Hence, removed.


deepeshgarg007 (changes requested):
@netchampfaris Please add description

nabinhait (approved):

PR #17361, opened 2019-04-24
feat: (Report) Subcontracted Reports

Anurag810:
It shows the transferred and pending quantity of raw material in subcontracted purchase Order.
As shown in screenshot
photo_2019-04-24_14-42-16


erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,93 @@
...
+def get_data(data, filters):
+	po = get_po(filters)
+	po_name = list(map(lambda x: {k:v for k, v in x.items() if k == 'name'}, po ))
+	sub_items = get_purchase_order_item_supplied(po_name)
+
+	from pprint import pprint
+
+	for item in sub_items:
+		for order in po:
+			print(order)
```
deepeshgarg007: Remove print

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,93 @@
...
+		{
+			"label": _("Purchase Order"),
+			"fieldtype": "Link",
+			"fieldname": "purchase_order",
+			"options": "Purchase Order",
+			"width": 150
+		},
+		{
+			"label": _("Date"),
+			"fieldtype": "date",
```
deepeshgarg007: D for date should be capital

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,93 @@
...
+		{
+			"label": _("Supplier"),
+			"fieldtype": "Link",
+			"fieldname": "supplier",
+			"options": "Supplier",
+			"width": 150
+		},
+		{
+			"label": _("Item Code"),
+			"fieldtype": "data",
```
deepeshgarg007: D for data and F for float should be capital

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,93 @@
...
+			"width": 100
+		}
+	]
+
+def get_data(data, filters):
+	po = get_po(filters)
+	po_name = list(map(lambda x: {k:v for k, v in x.items() if k == 'name'}, po ))
+	sub_items = get_purchase_order_item_supplied(po_name)
+
+	from pprint import pprint
```
deepeshgarg007: Remove unused imports

deepeshgarg007 (changes requested):

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
Anurag810: done

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
Anurag810: done

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
Anurag810: done

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
Anurag810: done

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1136,15 +1135,25 @@ def update_purchase_order_supplied_items(self):
...
 			item_code = d.get('original_item') or d.get('item_code')
+			transfered_qty[item_code] = d.qty
 			reserve_warehouse = item_wh.get(item_code)
 			stock_bin = get_bin(item_code, reserve_warehouse)
 			stock_bin.update_reserved_qty_for_sub_contracting()
 
+		po = frappe.get_doc("Purchase Order", self.purchase_order)
+
+		for items in po.supplied_items:
+			items.transfered_qty = items.transfered_qty + transfered_qty.get(items.rm_item_code)
```
rohitwaghchaure: What if user cancelled the stock entry, does transfered_qty will be reduced?

rohitwaghchaure: In BOM, user may can add same item two times. In that case same item will be added multiple times in the "Purchase Order Item Supplied" table. As you have used the item code for the reference, the transfered_qty will show wrong for the 1st place item. For example: Item A has 2 qty in first row and Item A has 4 qty in second row. Now when user creates the stock entry, as per your design transferred qty for Item A at first row and second row will be 4 which is wrong.

erpnext/patches/v12_0/set_transferred_qty_if_po_is_subcontracted.py
```diff
@@ -0,0 +1,31 @@
...
+                details["Purchase_order"] = entry.purchase_order
+                if entry.purchase_order not in records:
+                    records[entry.purchase_order] = {}
+                    records[entry.purchase_order][details.item_code] = details.qty
+                else:
+                    records[entry.purchase_order][details.item_code] = records[entry.purchase_order].get(details.item_code, 0) + details.qty
+
+    for k,v in records.items():
+        for key,value in v.items():
+            frappe.db.sql("update `tabPurchase Order Item Supplied` set transfered_qty = %s where parent = %s and rm_item_code = %s", (value, k, key))
```
rohitwaghchaure: Update query in for loop, so consider there are 100 records then 100 times update query will be called.
This can be handle using one sql query.

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,96 @@
...
+			"label": _("Pending Quantity"),
+			"fieldtype": "Float",
+			"fieldname": "p_qty",
+			"width": 100
+		}
+	]
+
+def get_data(data, filters):
+	po = get_po(filters)
+	po_name = list(map(lambda x: {k:v for k, v in x.items() if k == 'name'}, po ))
```
rohitwaghchaure: po_name = [po.name for d in po]

erpnext/buying/report/subcontracted_raw_materials_to_be_transferred/subcontracted_raw_materials_to_be_transferred.py
```diff
@@ -0,0 +1,96 @@
...
+			["is_subcontracted", "=", "Yes"],
+			["supplier", "=", filters.supplier],
+			["transaction_date", "<=", filters.to_date],
+			["transaction_date", ">=", filters.from_date],
+			["docstatus", "=", 1]
+		]
+	return frappe.get_all("Purchase Order", filters=record_filters, fields=["name", "transaction_date", "supplier"])
+
+def get_purchase_order_item_supplied(po):
+	po = ', '.join(map(str, [l['name'] for l in po if 'name' in l]))
```
rohitwaghchaure: No need to convert it to string, you can pass the list in the filters
filters = {'parent': ('in', po)}

Anurag810:
#17476

PR #17375, opened 2019-04-25
Replace explicit validation for unique email in Lead with DocType validation

vjFaLk:
The reasoning behind this is that we can override the unique validation using Customizations. Also, it's not DRY.


erpnext/crm/doctype/lead/lead.json
```diff
@@ -209,10 +215,10 @@
    "remember_last_selected_value": 0, 
    "report_hide": 0, 
    "reqd": 0, 
-   "search_index": 1, 
+   "search_index": 0, 
    "set_only_once": 0, 
    "translatable": 0, 
-   "unique": 0
+   "unique": 1
```
nabinhait: It will break for many users while setting unique in db. There are many duplicate email ids as the validations was missing at the beginning.

vjFaLk: Can we do anything to not affect existing records?

nabinhait: I can't think any solution. Closing the PR.

PR #17384, opened 2019-04-25
feat: Procurement Tracker Report

Mangesh-Khairnar:
- Tracks changes in procurement(purchase cycle):

procurement-track


rohitwaghchaure:
Date filters(from date, to date) are required otherwise this report will take lot of time to load when data is large

erpnext/buying/report/procurement_tracker/procurement_tracker.py
```diff
@@ -0,0 +1,270 @@
...
+def get_mapped_pi_records():
+	return frappe._dict(frappe.db.sql("""
+		SELECT
+			po_detail,
+			base_amount
+		FROM `tabPurchase Invoice Item`
+		WHERE
+			docstatus=1
+			AND po_detail IS NOT NULL
+		"""))
```
rohitwaghchaure: Why are you fetching the all PI records, what if the particular order is completed. Does above code will fetch their PI? I think your code should only fetch the PI records of which PO is not in the status ("Closed","Completed","Cancelled")

Mangesh-Khairnar: fixed

saurabh6790 (approved):

PR #17398, opened 2019-04-27
fix: Batch naming series hotfix

agritheory:
Call of `unicode()` on line 63 s not python 3 compatible. Fix runs on both 2.7 and 3.x


mujeerhashmi:
Hi,
You can create pull request with the final commit. Why 3 commits??

PR #17401, opened 2019-04-28
feat: adds ability to have zero quantity items on transactions

tukutela:
Includes ability to alter turn it on or off via the Account Settings page.


nabinhait (commented):
Can you please explain the use case?

erpnext/controllers/accounts_controller.py
```diff
@@ -88,8 +88,7 @@ def validate(self):
 			self.validate_paid_amount()
 
 		if self.doctype in ['Purchase Invoice', 'Sales Invoice']:
-			pos_check_field = "is_pos" if self.doctype=="Sales Invoice" else "is_paid"
-			if cint(self.allocate_advances_automatically) and not cint(self.get(pos_check_field)):
+			if cint(self.allocate_advances_automatically):
```
nabinhait: It should not be changed

nabinhait (changes requested):

tukutela:
Hi Nabin, thanks for reviewing! 

As for use-cases we have a few, but the main reason is that one of my clients is a language translation company in Chicago, a client uploads a document and they translate it. In order to quote for the work (and eventually invoice for it) they run it through a process of collating the number of words, duplicates, previous matches, and something close called a fuzzy match. Essentially you run the file through a database of something you've done for that client previously, saves a bit of money as the translation tool does the work. So the lines on a transaction is the number of words, then the different types of matches the importation process found which are the discounts. Giving you a price per language for a group of documents. (Hope all this makes sense). 

But to answer your question a little more succinctly, we need to show that the files provided by the client had zero repetition and/or fuzzy matches for some languages (eg. we haven't done any work for them in this language, but others we have). Zero line quantities are pretty important to this particular business, for new clients they appear more often, and even for long-term existing ones working in a new language, or with new subject matter.  

Re: As soon as I figure out how to do an update of accounts_controller.py within this context, I'll update it :) Thanks for picking that up, I didn't intend for that edit.

erpnext/controllers/accounts_controller.py
tukutela: you are absolutely right! my apologies.

tukutela:
Hi Nabin, could I ask that you review this again? Thankyou.

tukutela (commented):
Hoping this gets reviewed again, please....

tukutela:
keeping this from being 'stale'

CVSaritama:
> Includes ability to alter turn it on or off via the Account Settings page.

In V10 we added products with zero amount for offers, but now it is impossible. We need an alternative.

iMoshi:
Anyone has a version for v12, or v13? :) 

Allowing quantity of 0 item on transactions should be allowed. Many times users import and export quotations, opportunities etc from branches, on spreadsheets, that have no internet access. For example, from mining field.

zoharyExponent:
is this now included in V14 ?

jcagbay:
No, I just tested V14.

PR #17406, opened 2019-04-28
fix: browser hangs while making payment entry for large number of outstaning invoices

rohitwaghchaure:
**Issue**
In the New Payment Entry, Customer and Bank Account is selected. When user tries to entering the value in the Paid Amount field browser hangs and doesn't react. This issue is coming because there are lot of outstanding invoices are pending against the customer. System takes lot of time to fetch and add the outstanding invoices in the reference table of the payment entry.

**Fix**
To fix this issue added button "Get Outstanding Invoice" in the payment entry, on click of button a prompt will open in which user has to select the from date and to date and system will fetch the 100 invoices based on the dates selected in the prompt. Also added checkbox "Auto Fetch Outstanding Invoices" which will fetch the outstanding invoices without input of the dates

fetch_manual_outstanding_invoices


adnanmunir83:
in this same payment entry, there is another bug.  when it's fetching outstanding invoices, we are not sending Company parameters. and in multi company setup, it fetch all the invoices for all companies.

rohitwaghchaure:
Will move this to develop branch

PR #17423, opened 2019-05-01
fix(Customer):  Total percentage should not be greater than 100 for sales team

hrwX:


erpnext/selling/doctype/customer/customer.py
```diff
@@ -60,6 +60,12 @@ def validate(self):
 			if self.loyalty_program == customer.loyalty_program and not self.loyalty_program_tier:
 				self.loyalty_program_tier = customer.loyalty_program_tier
 
+		allocated_percentage = 0
+		for percentage in self.sales_team:
+			allocated_percentage += percentage.allocated_percentage
+			if allocated_percentage > 100:
```
nabinhait: I think it must be equal to 100. Any use case for less than 100?

nabinhait (dismissed):

erpnext/selling/doctype/customer/customer.py
```diff
@@ -60,6 +60,14 @@ def validate(self):
 			if self.loyalty_program == customer.loyalty_program and not self.loyalty_program_tier:
 				self.loyalty_program_tier = customer.loyalty_program_tier
 
+		if self.sales_team:
+			allocated_percentage = 0
+			for percentage in self.sales_team:
+				allocated_percentage += percentage.allocated_percentage
+
+			if not allocated_percentage == 100:
```
Alchez:
```suggestion
			if sum([member.allocated_percentage for member in self.sales_team]) != 100:
```

Alchez (dismissed):

nabinhait (approved):

PR #17435, opened 2019-05-01
fix(minor): chart of accounts sorting and message

rmehta:


erpnext/accounts/report/financial_statements.py
```diff
@@ -334,6 +334,9 @@ def compare_accounts(a, b):
 			if re.split('\W+', a[key])[0].isdigit():
 				# if chart of accounts is numbered, then sort by number
 				return cmp(a[key], b[key])
+			else:
```
nabinhait: No need for `if...else`

nabinhait (changes requested):

nabinhait (approved):

PR #17456, opened 2019-05-02
fix: handling case if from date and to date are equal in billing reports

Anurag810:
This is the edge case, when from_date and end_date are same


nabinhait (commented):
Refactored PR https://github.com/frappe/erpnext/pull/17460

PR #17459, opened 2019-05-02
fix: browser hangs while making payment entry for large number of outstaning invoices

rohitwaghchaure:
**Issue**
In the New Payment Entry, Customer and Bank Account is selected. When user tries to entering the value in the Paid Amount field browser hangs and doesn't react. This issue is coming because there are lot of outstanding invoices are pending against the customer. System takes lot of time to fetch and add the outstanding invoices in the reference table of the payment entry.

**Fix**
To fix this issue added button "Get Outstanding Invoice" in the payment entry, on click of button a prompt will open in which user has to select the from date and to date and system will fetch the 100 invoices based on the dates selected in the prompt. Also added checkbox "Auto Fetch Outstanding Invoices" which will fetch the outstanding invoices without input of the dates

payment_entry_enhance


erpnext/patches/v12_0/update_due_date_in_gle.py
```diff
@@ -0,0 +1,17 @@
...
+    for doctype in ["Sales Invoice", "Purchase Invoice", "Journal Entry"]:
+        frappe.reload_doc("accounts", "doctype", frappe.scrub(doctype))
+
+        frappe.db.sql(""" UPDATE `tabGL Entry`, `tab{doctype}`
+            SET
+                `tabGL Entry`.due_date = `tab{doctype}`.due_date
+            WHERE
+                `tabGL Entry`.voucher_no = `tab{doctype}`.name and `tabGL Entry`.party is not null
+                and `tabGL Entry`.voucher_type in ('Sales Invoice', 'Purchase Invoice', 'Journal Entry')
+                and account in (select name from tabAccount where account_type in ('Receivable', 'Payable') )"""
```
Mangesh-Khairnar:
```suggestion
                and account in (select name from `tabAccount` where account_type in ('Receivable', 'Payable') )"""
```

Mangesh-Khairnar (dismissed):

Mangesh-Khairnar (approved):

PR #17474, opened 2019-05-03
fix: Make Customer and lead field dynamic in quotations and opportunity

deepeshgarg007:


Anurag810:
Please add some proper description or screenshot.

PR #17477, opened 2019-05-03
fix(stock): Allow expired batches to be flushed out of the system

Alchez:
**Original fix:** #14249
Made to develop at https://github.com/frappe/erpnext/pull/17478.

<hr>

**Problem:**

Trying to issue batches out of the system fails for expired batches because of an extra validation.

**Solution:**

Remove validation for Material Issue stock entries.


nabinhait (approved):

f-9t9it:
How can we return the expired batch to the supplier using Purchase Return ??
When we raise a purchase invoice return with expired items then the expired validation doesnt allow return.

PR #17492, opened 2019-05-05
feat(Service Level Agreement): Service Level Agreement based on Customer, Group and Territory

hrwX:
- Multiple Service Level Priorities can now be defined in Service Level, enabling to change priority when needed
sl

- A Service Level Agreement is created using the above Service Level
sla

- A Service Level Agreement can be defined for a specific Customer or a Customer Group or a Territory
sla

- Priority of the Issue can now be changed and time to respond and resolve can be set as per priority
sla-issue

- Service Level Agreements linked based on Customer or Customer Group or Territory. The Service Level Agreement is selected on the following basis 
`> Customer > Customer Group > Territory > Default Service Level Agreement`

Screenshot 2019-06-10 at 12 42 47 PM

iss

- To update Service Level Agreement Status in Issues on Custom Status, use the following code in Custom Scripts
`frappe.ui.form.on('Issue', {
    status: function(frm) {
        if(frm.doc.status == "qwe"){
            frm.call('update_agreement_fulfilled_on_custom_status').then(r=>{
            frappe.msgprint("Updated");
            frm.refresh();
            })
        }
	}
})`


erpnext/support/doctype/issue/issue.js
```diff
@@ -1,13 +1,53 @@
 frappe.ui.form.on("Issue", {
 	onload: function(frm) {
 		frm.email_field = "raised_by";
+
 		if (frm.doc.service_level_agreement) {
-			set_time_to_resolve_and_response(frm);
+			frappe.call({
```
rmehta: why 2 separate calls? Get all related information in one call

erpnext/support/doctype/issue/issue.js
```diff
@@ -20,13 +60,58 @@ frappe.ui.form.on("Issue", {
...
+
+	service_level_agreement: function(frm) {
+		frappe.show_alert({
+			indicator: 'green',
+			message: __('Changing Service Level Agreement.')
+		});
+		frm.call('change_service_level_agreement_and_priority', {
+			"service_level_agreement": frm.doc.service_level_agreement
+		}).then(() => {
+			frappe.msgprint(__("Service Level Agreement changed to {0}.", [frm.doc.service_level_agreement]));
```
rmehta: `show_alert` again ?

erpnext/support/doctype/issue/test_issue.py
```diff
@@ -18,56 +18,116 @@ def test_assignment(self):
...
+		issue = make_issue(creation, "_Test Customer", 1)
 
-		# make issue with customer specific SLA
-		issue = make_issue(creation, '_Test Customer')
+		self.assertEquals(issue.response_by, datetime.datetime(2019, 3, 4, 14, 0))
+		self.assertEquals(issue.resolution_by, datetime.datetime(2019, 3, 4, 15, 0))
 
-		self.assertEquals(issue.response_by, datetime.datetime(2019, 3, 7, 18, 0))
-		self.assertEquals(issue.resolution_by, datetime.datetime(2019, 3, 9, 18, 0))
+		"""
```
rmehta: use comments instead of docstrings

erpnext/patches/v12_0/set_priority_for_support.py
```diff
@@ -0,0 +1,70 @@
+import frappe
```
rmehta: add comments, explaining what and why each function does

rmehta (dismissed):
👍 see changes and add comments

erpnext/support/doctype/issue/issue.js
hrwX: Reverted to one call to get data

erpnext/support/doctype/issue/issue.js
hrwX: reverted to only one `msgprint`on change

PR #17494, opened 2019-05-06
refactor: Make Travel Request Doctype Submittable

kennethsequeira:
- Doctype is now submittable. 
- Removes the issue of someone changing a very old travel request
- added a patch which will submit any old travel requests greater than 30 days. Newer ones will be in draft stage.


kennethsequeira:
@nabinhait since changes to v11 are forbidden, should this be closed and pushed to v12/develop?

PR #17502, opened 2019-05-06
fix: Show make sales order button only if quotation to customer

deepeshgarg007:


saurabh6790 (approved):

saurabh6790:
@deepeshgarg007 

Test:
1. after converting a lead to customer, the system should allow sales order creation against the particular quotation.

PR #17508, opened 2019-05-06
fix: set task cancelled if project is cancelled

Anurag810:
Once project is cancelled, all tasks linked to the project should get cancelled.


Anurag810:
Duplicate https://github.com/frappe/erpnext/pull/17497

PR #17537, opened 2019-05-09
fix: allow user to sync stock_value and account_balance jv if perpetual inventory is checked

Anurag810:
Before: if perpetual inventory is enabled then we are not allow to create journal entry.
now: if perpetual inventory is enabled then we are allow to create journal entry only **if stock value and account balance miss match** to balance record.


Anurag810:
@nabinhait need help

Anurag810:
It should be in develop

PR #17543, opened 2019-05-09
fix: change in task end date and start date does not reflect in project

Anurag810:
fix: change in task end date and start date does not reflect in project

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -100,11 +100,6 @@ class Task(NestedSet):
 	def update_project(self):
 		if self.project and not self.flags.from_project:
 			project = frappe.get_doc("Project", self.project)
-			for task in project.tasks:
-				if task.task_id == self.name:
-					task.start_date = self.exp_start_date
-					task.end_date = self.exp_end_date
-
 			project.update_project()
 			project.save()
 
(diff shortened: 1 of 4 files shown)


scmmishra (dismissed):

erpnext/projects/doctype/task/task.py
```diff
@@ -99,7 +99,14 @@ def update_time_and_costing(self):
 
 	def update_project(self):
 		if self.project and not self.flags.from_project:
-			frappe.get_doc("Project", self.project).update_project()
+			project = frappe.get_doc("Project", self.project)
+			for task in project.tasks:
+				if task.task_id == self.name:
+					task.start_date = self.exp_start_date
```
netchampfaris: This should be handled here

https://github.com/frappe/erpnext/blob/3866e132f71dfe98cb2dc95bf613a2b1d5de1cb7/erpnext/projects/doctype/project/project.py#L142

netchampfaris (changes requested):

erpnext/projects/doctype/task/task.py
Anurag810: but it synced when task is updated so put it into task.py

nabinhait (changes requested):
Travis tests for Task are failing.

erpnext/projects/doctype/task/task.py
Anurag810: it sync the tasks if something is updated from project side.
But Here is the opposite scenario when we update project the changes go to tasks.
But when we update the task(date) it does not go to the project.

Anurag810: done in different way

Anurag810: Now fixing travis

Issue #17547, opened 2019-05-09
Pricing Rules are not queried correctly after removing 'customer' and adding 'party_name' in Quotation


mnour:
Hi

  On our production environment we used to work with version 11.1.21, and after upgrading to version 11.1.25, we noticed that all of our pricing rules are not working correctly at all.

After digging deep we found that the main issue behind this is the removal of 'customer' field and replacing it with 'party_name' for doctype Quotation.

This affected the query constructed in 'erpnext.accounts.doctype.pricing_rule.pricing_rule' method 'get_pricing_rules', there the fields checks to add the relevant query part(s) are based on the availability of 'customer' field.

There are two places that needed to be fixed:

1. JavaScript Code: Namely 'erpnext/public/js/controllers/transaction.js', which has been fixed by ERP Next team and it has published in a new release, namely the latest release of ERP Next v11.1.28

2. Python/Backend Code: That part was still not working in our case, cause we are using ERP Next from both UI Client and via its APIs, and the APIs part was still not working. After yet another deep digging and remote debugging we found the issue and we will issue a PR.

Hope the ERP Next team takes a look at the PR, review it as soon as they can, and include it in another release.

Thanks

nabinhait:
Fixed via https://github.com/frappe/erpnext/pull/17760

PR #17550, opened 2019-05-09
Pricing Rules are not queried correctly after removing 'customer' and adding 'party_name' in Quotation #17547

mnour:
- Filling in 'customer' arg field if not already set and if 'party_name' is available
- This will fix the logic during which Pricing Rule is constructing the SQL query to list all Pricing Rules related to an item
- The reason to put it in accounts_controller is:
-# This is the palce when args are made before making call which will end using Pricing Rule logic, which means we don't have to go all over the place to add the proper arg field
-# IMHO, it is clean enough in contrast to the case where that code is added to pricing_rule.py for example, cause doing this exposes Pricing Rule to the knowledge about a field called 'party_name' while a Pricing Rule document doesn't have any


erpnext/controllers/accounts_controller.py
```diff
@@ -237,6 +237,10 @@ def set_missing_item_details(self, for_validate=False):
 				document_type = "{} Item".format(self.doctype)
 				parent_dict.update({"document_type": document_type})
 
+                        # To fix the problems cause by renaming 'customer' to 'party_name- in selling docs
+                        if not parent_dict.get("customer") and parent_dict.get("party_name"):
```
saurabh6790: The field is only renamed for Quotation document. 

Also now you can quote to a customer as well as lead, so before settings party_name value directly to a customer, first you need to check whether the quotation is against customer or lead.

saurabh6790 (changes requested):

erpnext/controllers/accounts_controller.py
mnour: Hi @saurabh6790 

   Thanks for the review, will do the updates based on your feedback.

nabinhait (approved):

Issue #17559, opened 2019-05-10
Webshop should not require address


rasos:
In webshops with virtual goods a delivery address does not make sense. Also most clients do not want to reveal a billing address and a phone number, which is currently required. This is a GDPR issue. 

I suggest to remove the mandatory address dialog in the webshop. Customers still are able to enter it.

rasos:
Could we get this merged into ERPnext 12? @rmehta

rmehta:
If you can send a correct fix with test coverage.

rasos:
We patch it manually after each upgrade and it works fine, tested.

I don't think it requires a setting, as the webshop order page always offers a dialogue to enter an address. Deleting lines 58+59 in cart.py simply does not make entering an address field mandatory.

PR #17561, opened 2019-05-10
Do not require address field in webshop

rasos:
Fixes https://github.com/frappe/erpnext/issues/17559

Mandatory address field is a GDPR issue.


rasos:
This should be an easy merge. Please make address field optional for webshop sales. We mainly sell digital services, there is no need to deliver a parcel physically.

rasos:
Still an open issue.

rmehta:
please don't comment code. Can we make this a setting?

rasos:
Yes, setting would be even better - there might be some industries which always need an address from webshop orders.

rasos:
resubmitted PR as https://github.com/frappe/erpnext/pull/19096

PR #17563, opened 2019-05-10
Fix for Chart of Account sorting problem

bghayad:
Hello;

This is fix for the chart of account sorting problem that appeared in version 11 and was not existed in version 10.
Appreciate the help to merge it into the core code to avoid finding it when doing upgrade or fresh installation for version 11.

Regards
Bilal


bghayad:
Hello;
Can we look for this to be merged if possible?
Regards
Bilal

nabinhait (approved):

PR #17589, opened 2019-05-14
fix: Do not append row if item is sold within days since last ordered

deepeshgarg007:

### DIFF
--- a/erpnext/accounts/report/inactive_sales_items/inactive_sales_items.py
+++ b/erpnext/accounts/report/inactive_sales_items/inactive_sales_items.py
@@ -96,6 +96,8 @@ def get_data(filters):
 						"qty": item_obj.qty,
 						"days_since_last_order": item_obj.days_since_last_order
 					})
+				else:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/inactive_sales_items/inactive_sales_items.py
```diff
@@ -96,6 +96,8 @@ def get_data(filters):
 						"qty": item_obj.qty,
 						"days_since_last_order": item_obj.days_since_last_order
 					})
+				else:
+					continue
```
rohitwaghchaure: Instead of this move data.append(row) in the if condition next to 98 line. But this looks like intentionally kept outside of the If condition.

deepeshgarg007: Yes it is intentional, if the item is not there in sales_invoice_date that means that item has never been sold in the respective territory. So the row should also be appended.

PR #17618, opened 2019-05-15
feat: set COGS for material issue only

rohitwaghchaure:
System set the COGS account in the expense account field even if the purpose is Manufacturing. With new changes only purpose with Material Issue will have the COGS account

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -652,35 +652,36 @@ class StockEntry(StockController):
 
 		ret = frappe._dict({
 			'uom'			      	: item.stock_uom,
-			'stock_uom'			: item.stock_uom,
+			'stock_uom'				: item.stock_uom,
 			'description'		  	: item.description,
-			'image'				: item.image,
+			'image'					: item.image,
 			'item_name' 		  	: item.item_name,
-			'expense_account'		: args.get("expense_account"),
-			'cost_center'			: get_default_cost_center(args, item, item_group_defaults),
-			'qty'				: args.get("qty"),
+			'cost_center'			: get_default_cost_center(args, item, item_group_defaults, self.company),
+			'qty'					: args.get("qty"),
 			'transfer_qty'			: args.get('qty'),
 			'conversion_factor'		: 1,
-			'batch_no'			: '',
+			'batch_no'				: '',
 			'actual_qty'			: 0,
 			'basic_rate'			: 0,
-			'serial_no'			: '',
+			'serial_no'				: '',
 			'has_serial_no'			: item.has_serial_no,
 			'has_batch_no'			: item.has_batch_no,
 			'sample_quantity'		: item.sample_quantity
 		})
-		for d in [["Account", "expense_account", "default_expense_account"],
-			["Cost Center", "cost_center", "cost_center"]]:
-				company = frappe.db.get_value(d[0], ret.get(d[1]), "company")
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -656,8 +656,7 @@ def get_item_details(self, args=None, for_update=False):
 			'has_batch_no'			: item.has_batch_no,
 			'sample_quantity'		: item.sample_quantity
 		})
-		for d in [["Account", "expense_account", "default_expense_account"],
-			["Cost Center", "cost_center", "cost_center"]]:
+		for d in [["Cost Center", "cost_center", "cost_center"]]:
```
nabinhait: `cost center`  is also not required, as now we have company specific defaults in Item master.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -666,6 +665,9 @@ def get_item_details(self, args=None, for_update=False):
 		if args.get("uom") and for_update:
 			ret.update(get_uom_details(args.get('item_code'), args.get('uom'), args.get('qty')))
 
+		if self.purpose == 'Material Issue':
+			ret["expense_account"] = frappe.get_cached_value('Company',  self.company,  "default_expense_account")
```
nabinhait: It should be set based on `get_default_expense_account(args, item, item_group_defaults)`

nabinhait (changes requested):
- In line 646, expense account should not be set. Otherwise, it will pick "COGS" from item.

erpnext/stock/doctype/stock_entry/stock_entry.py
nabinhait: cost center is also not required, as now we have company specific defaults in Item master.

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
rohitwaghchaure: Done

erpnext/stock/doctype/stock_entry/stock_entry.py
rohitwaghchaure: Not used get_default_expense_account because item.expense_account has found

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -656,17 +655,16 @@ def get_item_details(self, args=None, for_update=False):
...
-				if not ret[d[1]] or (company and self.company != company):
-					ret[d[1]] = frappe.get_cached_value('Company',  self.company,  d[2]) if d[2] else None
 
 		# update uom
 		if args.get("uom") and for_update:
 			ret.update(get_uom_details(args.get('item_code'), args.get('uom'), args.get('qty')))
 
-		if not ret["expense_account"]:
+		if self.purpose == 'Material Issue':
+			ret["expense_account"] = (item.get("expense_account") or
```
nabinhait: If expense account is not set in Item master, then it should look for Item Group and then Company record.

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
rohitwaghchaure: Done

rmehta:
not required in `develop` ?

PR #17619, opened 2019-05-15
fix: Query fixes in quotation and opportunity 

deepeshgarg007:


erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -162,16 +166,16 @@ erpnext.selling.QuotationController = erpnext.selling.SellingController.extend({
 		}
 	},
 
-	lead: function() {
+	get_lead_details: function() {
 		var me = this;
-		if(!this.frm.doc.lead) {
+		if(!this.frm.doc.quotation === "Lead") {
```
nabinhait: `quotation_to`?

nabinhait (changes requested):

erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -47,6 +47,10 @@ erpnext.selling.QuotationController = erpnext.selling.SellingController.extend({
 		erpnext.utils.get_party_details(this.frm, null, null, function() {
 			me.apply_price_list();
 		});
+
+		if(me.frm.doc.quotation=="Lead") {
```
nabinhait: `quotation_to`

nabinhait (changes requested):

nabinhait (approved):

PR #17627, opened 2019-05-16
fix: variant item description based on attribute

Anurag810:
The item description on creating the variants from template was not based on attribute.
this pull request is to fix that issue


erpnext/controllers/item_variant.py
```diff
@@ -287,19 +287,12 @@ def copy_attributes_to_variant(item, variant):
 					variant.set(field.fieldname, item.get(field.fieldname))
 
 	variant.variant_of = item.name
-	if 'description' in allow_fields:
```
netchampfaris: Include this check

netchampfaris (changes requested):

netchampfaris (approved):

PR #17632, opened 2019-05-16
fix: Allow return if delivery note or sales order is required in selling settings(v11)

Anurag810:
Before:
sales invoice return if delivery note or sales order is required in selling settings was not allowed.

After:
sales invoice return if delivery note or sales order is required in selling settings is allowed now.


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -509,7 +509,8 @@ def so_dn_required(self):
 				for d in self.get('items'):
 					if (d.item_code and frappe.get_cached_value('Item', d.item_code, 'is_stock_item') == 1
 						and not d.get(i.lower().replace(' ','_')) and not self.get(dic[i][1])):
-						msgprint(_("{0} is mandatory for Item {1}").format(i,d.item_code), raise_exception=1)
+						if not self.is_return:
```
nabinhait: Why another `if` condition, club into one if condition?

nabinhait (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
Anurag810: done

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -503,11 +503,14 @@ def validate_auto_set_posting_time(self):
...
 	def so_dn_required(self):
 		"""check in manage account if sales order / delivery note required or not."""
+		if self.is_return:
+			return
 		dic = {'Sales Order':['so_required', 'is_pos'],'Delivery Note':['dn_required', 'update_stock']}
 		for i in dic:
 			if frappe.db.get_single_value('Selling Settings', dic[i][0]) == 'Yes':
 				for d in self.get('items'):
-					if (d.item_code and frappe.get_cached_value('Item', d.item_code, 'is_stock_item') == 1
+					cached_item_code = frappe.get_cached_value('Item', d.item_code, 'is_stock_item')
```
nabinhait: Variable naming is too bad. it should define the purpose like `is_stock_item`

nabinhait (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
Anurag810: done

nabinhait (approved):

PR #17652, opened 2019-05-17
fix: Unable to create item variant based on manufacturer

deepeshgarg007:
Hotfix: https://github.com/frappe/erpnext/pull/17651

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -397,7 +397,7 @@ $.extend(erpnext.item, {
 			// call the server to make the variant
 			data.template = frm.doc.name;
 			frappe.call({
-				method:"erpnext.controllers.item_variant.get_variant",
+				method: "erpnext.controllers.item_variant.get_variant",
 				args: data,
 				callback: function(r) {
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item.js
```diff
@@ -375,10 +375,18 @@ $.extend(erpnext.item, {
 	show_modal_for_manufacturers: function(frm) {
 		var dialog = new frappe.ui.Dialog({
 			fields: [
-				{fieldtype:'Link', options:'Manufacturer',
-					reqd:1, label:'Manufacturer'},
-				{fieldtype:'Data', label:'Manufacturer Part Number',
-					fieldname: 'manufacturer_part_no'},
+				{
+					fieldtype:'Link',
```
nabinhait: Space after colon, fix everywhere

nabinhait (changes requested):

erpnext/stock/doctype/item/item.js
deepeshgarg007: fixed

nabinhait (approved):

Issue #17560, opened 2019-05-10
Website Product links


amitsaxena01:
In Version 12, for a website example.com  having an item group with route /products/ig1 and an item with route /products/ig1/item1.
If we goto the product listing for an item group , example.com/products/ig1,  it display the items there. On clicking the link to product on this page sat item1, url used is  example.com/products/ig1/products/ig1/item1 instead of example.com/products/ig1/item1. As a result a Page on Available error is shown.

gmplab:
#17916

PR #17665, opened 2019-05-17
fix: Correcting the reference to web pages from item search on item_group

amitsaxena01:

This pull request solves issue https://github.com/frappe/erpnext/issues/17560. 
Currently on clicking the link on item row, it adds the address of current page to the route rather than the base url.
The changes corrects the issues and web pages are opening correctly

### DIFF
--- a/erpnext/www/all-products/item_row.html
+++ b/erpnext/www/all-products/item_row.html
@@ -2,7 +2,7 @@
 	<div class="row no-gutters">
 		<div class="col-md-3">
 			<div class="card-body">
-				<a class="no-underline" href="{{ item.route }}">
+				<a class="no-underline" href="/{{ item.route }}">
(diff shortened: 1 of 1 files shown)


amitsaxena01:
Review requested

amitsaxena01:
request code review

ashish-greycube:
@amitsaxena01 @deepeshgarg007 , request you to open this fix. 
[a] Issue is there and it is reported correctly
[b] also, i tested this fix and it is correct

EconCode:
Hi, found the same issue and tried a similar fix. 
However, I'm not sure if the fix is the best idea as it will only allow root relative links in the item_row template. It basically hard-codes a hidden prefix that is not shown to the enduser. 
Therefore, wouldn't changing the automatic route generation in the item_group controller be a less invasive way of handling this issue? That way it is clearly shown to the enduser that the link is root-relative and they'd still have the possibility to manually change it if they so desire for whatever reason?

PR #17673, opened 2019-05-20
Fix: Precision in Appraisals

Mitchy25:
This fixes issues associated with 2 decimal places. Example: 4 questions (Each weighted 25%), will give a total result of 5.2 (If 5 achieved on every question). This is because 5*0.25 = 1.25 which is rounded to 1.3 for precision=2. When precision = 3, this is not an issue.


erpnext/hr/doctype/appraisal/appraisal.js
```diff
@@ -52,7 +52,7 @@ cur_frm.cscript.score = function(doc,cdt,cdn){
 			refresh_field('score', d.name, 'goals');
 		}
 		var total = flt(d.per_weightage*d.score)/100;
-		d.score_earned = total.toPrecision(2);
+		d.score_earned = total.toPrecision(3);
```
nabinhait: It should not be hard-coded. It should be based on field's precision.

nabinhait (changes requested):

nabinhait (dismissed):

PR #17679, opened 2019-05-20
fix(Revert): sales order status for order type 'Maintenance

Anurag810:
Reverts #17119
The PR is needed to Revert because Order type Maintenance need delivery note in some cases and is not needed in some cases. So we have decided to create a checkbox Delivery Note required in Sales Order if Order Type is Maintenance

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.js
+++ b/erpnext/selling/doctype/sales_order/sales_order.js
@@ -140,12 +140,15 @@ erpnext.selling.SalesOrderController = erpnext.selling.SellingController.extend(
 				}
 
 				// delivery note
-				if(flt(doc.per_delivered, 6) < 100 && ["Sales", "Shopping Cart"].indexOf(doc.order_type)!==-1 && allow_delivery) {
+				if(flt(doc.per_delivered, 6) < 100 && allow_delivery) {
 					this.frm.add_custom_button(__('Delivery'),
 						function() { me.make_delivery_note_based_on_delivery_date(); }, __("Make"));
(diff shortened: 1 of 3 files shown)


rohitwaghchaure:
@Anurag810 at JS side it's still hard coded
https://github.com/frappe/erpnext/blob/hotfix/erpnext/selling/doctype/sales_order/sales_order_list.js#L34

Anurag810:
> @Anurag810 at JS side it's still hardcoded
> https://github.com/frappe/erpnext/blob/hotfix/erpnext/selling/doctype/sales_order/sales_order_list.js#L34
It is a standard option, if i have to remove it so how can I distinguish between other order types.

erpnext/selling/doctype/sales_order/sales_order.js
```diff
@@ -181,6 +181,8 @@ erpnext.selling.SalesOrderController = erpnext.selling.SellingController.extend(
 				// maintenance
 				if(flt(doc.per_delivered, 2) < 100 &&
 						["Sales", "Shopping Cart"].indexOf(doc.order_type)===-1) {
+					this.frm.add_custom_button(__('Delivery'),
+						function() { me.make_delivery_note_based_on_delivery_date(); }, __("Make"));
```
rohitwaghchaure: Why double? this code is already there at line no 145

Anurag810: done

rohitwaghchaure (approved):

PR #17690, opened 2019-05-21
fix: while making the item, default warehouse not set even if the stock settings has the warehouse

rohitwaghchaure:
**Issue**

In stock settings "Default Warehouse" is set
Screen Shot 2019-05-21 at 10 40 26 am

Sill while making the item default warehouse is not set
Screen Shot 2019-05-21 at 10 42 24 am

**After Fix**
Screen Shot 2019-05-21 at 11 02 04 am


nabinhait (changes requested):
I think this should not be handled by adding defaults in item from item group / stock settings, if missing. Think if after creation of an item, if default warehouse is set on item group / stock settings, it does not updated all related items.

It should be handled, while getting default warehouse from item via `get_item_details` file. It should check in item group and stock settings, if default not found in item.

rohitwaghchaure:
@nabinhait  In the get_item_details the code was there but it was not working, have fixed that. Please review

erpnext/stock/get_item_details.py
```diff
@@ -220,17 +220,18 @@ def get_basic_details(args, item):
 	if item.variant_of:
 		item.update_template_tables()
 
-	from frappe.defaults import get_user_default_as_list
-	user_default_warehouse_list = get_user_default_as_list('Warehouse')
```
nabinhait: I think, this is for setting values based on "User Permission". Should we remove it?

nabinhait (changes requested):

erpnext/stock/get_item_details.py
rohitwaghchaure: User permissions works in the child table, why we need code for that

nabinhait (approved):

PR #17724, opened 2019-05-23
fix: Item template deletes info from its variants

### LINKED ISSUE #17664: [BUG] Saving an Item template deletes important info from its variants
### Issue:
The `reorder_levels` Item's child table disappears with no apparent reason.

### Tracked down bug:
By searching in the `tabVersion` table of the db, I detected that the child table had been in all the items at the same time, and the common denominator of all that "random" items was the Item Template. So I searched for a change made in the Item Template at that same time, and there it was, the `item_defaults` child table was modified.
I tryed changing that table to recreate the issue, and it happened. I decided to take it a step further and save the document without changing anything, it deleted the `reorder_levels` child table too.
In conclusion: When you save an Item Template, all its variants are updated and their `reorder_levels` records are deleted.

### How to fix it:
The way to fix it is to add `"reorder_levels"` to the `exclude_fields` list in the next line:

https://github.com/frappe/erpnext/blob/0692e5eb78564b4070880a234c61a0382bccf082/erpnext/controllers/item_variant.py#L262-L267

I could do it, but a lot of my Pull Requests are open and unattended, so I let this here and you can make your own Pull Request.

Don-Leopardo:
fix #17664

### DIFF
--- a/erpnext/controllers/item_variant.py
+++ b/erpnext/controllers/item_variant.py
@@ -263,7 +263,8 @@ def copy_attributes_to_variant(item, variant):
 	# copy non no-copy fields
 
 	exclude_fields = ["naming_series", "item_code", "item_name", "show_in_website",
-		"show_variant_in_website", "opening_stock", "variant_of", "valuation_rate"]
+		"show_variant_in_website", "opening_stock", "variant_of", "valuation_rate",
+		"reorder_levels"]
 
 	if item.variant_based_on=='Manufacturer':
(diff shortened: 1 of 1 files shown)


Don-Leopardo:
Well, the bug is a little bitt more complex than I thougth in first place. This can be a viable solution, but this should be discussed.
The existing functionality is to copy the `reorder_level` child table from the template to the variants. The problem is that the possibility for a variant to have its own `reorder_level` does not exists, and it's a real case.

PR #17732, opened 2019-05-23
fix: Make Quotation from Lead with no company

### LINKED ISSUE #17723: [BUG] Make Quotation from Lead
How to reproduce:
In a Lead, Make > Quotation

Captura de pantalla 2019-05-23 a la(s) 08 46 00

ERPNext: v11.1.32 (master)
Frappe Framework: v11.1.30 (master)

Don-Leopardo:
fix #17723
It was caused because there wasn't a company configured in the Global Defaults of the instance. Anyways, the error message wasn't clear and I think that if the company is required to make a Quotation from the Lead it should be required to create a Lead in the first place.
I added the required to company in Lead and bow the error message is going to be clear because it's going to say that the company field is required for Lead.


saurabh6790 (changes requested):
@Don-Leopardo Transactions are specific to companies but master data can be shared across multiple companies. Thus the field is not mandatory by default. 

If the error message is not clear on the transaction, then we have to fix the error message by adding a more descriptive message.

saurabh6790:
@nabinhait Please check this

Don-Leopardo:
@saurabh6790 The company is assigned to the lead automatically and in a field that is almost hidden. There is no difference in making it required, it only saves the case when you forget to configure the company in global configuration, that is this case.

nabinhait:
Fixed the issue #17723 via https://github.com/frappe/erpnext/pull/18292, closing this PR.

PR #17736, opened 2019-05-24
test: [WIP]Report tests

Anurag810:
This PR contains the test cases for some reports to increase the over all test coverage.


erpnext/stock/report/stock_balance/test_stock_balance.py
```diff
@@ -0,0 +1,77 @@
...
+def get_item_map(rows):
+
+	item_map = {}
+	for row in rows:
+		item_map[row[0]] = row
+
+	return item_map
+
+
+def get_item(item_code, qty):
```
rmehta, line 48: repetitive function (?)

rmehta (changes requested):
DRY

erpnext/stock/report/stock_balance/test_stock_balance.py
Anurag810: I just checked there is only one method for creating the item map.

Anurag810:
will work on this later

PR #17774, opened 2019-05-28
fix: Adding task to project from task doctype

### LINKED ISSUE #17730: [CRITICAL] Cannot add more than one Task to Project
### How to reproduce

1. Go to Projects List and create a new Project, "Test Project" and save.
2. Go to Tasks List and create a new Task "Test Task 1", link to above Project and save.
3. Open the Project "Test Project".

	Project shows one Task in dashboard and "Test Task 1" in Project Tasks List.

4. Still in "Test Project", hit Save.
5. Go to Tasks List and create a second Task, "Test Task 2", link it to "Test Project" and save.
6. Open the Project "Test Project".

	"Test Project" shows two tasks in the dashboard but **only "Test Task 1" in the Project Tasks List**.

7. Hit save again.

	**Second Task disappears from dashboard.** ("Test Task 2"'s link to "Test Project" gets removed)

### Versions

ERPNext: v11.1.31 (master)
Frappe Framework: v11.1.29 (master)

Reproduced on two independent Systems.

@nabinhait @wojosc @templis

Don-Leopardo:
fix #17730 

The error wasn't only that you couldn't add more than one task to a project, it was that you couldn't add tasks to a project from the task doctype.
With this fix, if you create a task with an assigned project, the task will apear in the project child table; furthermore, if you edit a task and change the project, it will be moved to the child table of the new project you selected and it will be deleted from the child table of the old one (it works when you select no project too).


erpnext/projects/doctype/task/task.py
```diff
@@ -101,6 +107,39 @@ def update_project(self):
 		if self.project and not self.flags.from_project:
 			frappe.get_doc("Project", self.project).update_project()
 
+	def update_project_tasks(self):
+		tasks = frappe.db.sql("""SELECT name
```
rmehta: fix indentation https://github.com/frappe/erpnext/wiki/Coding-Standards#use-simple-indenting

rmehta: Also use ORM

rmehta (commented):
- [ ] Add tests
- [ ] Move this to `develop`

rmehta (changes requested):
see comments

erpnext/projects/doctype/task/task.py
Don-Leopardo: I think that this is solved now, can you confirm?

Don-Leopardo:
> * [ ]  Add tests
> * [ ]  Move this to `develop`

Does this mean that I should close this PR or should I just open another to develop? (I prefer the to open another)

Can you provide to me some link about making tests? I had never done it and I don't know how they work. What should I make a test for?

barredterra:
@rmetha can this be treated like a bugfix? It solves the unexpected behavior described in #17730 , that makes it hard to use project management.

rodrigoinformationsys:
I've done some testing in your branch and I've found an unexpected behaviour. Maybe it is how erpnext works as default.
Testing the connection between the "Project Task" and the "Task" and everything went well. But I do use "Track Changes" in my projects and tasks so I can keep up with my team. When I change status from "Open" to "Working" of a task from project's page it doesn't show this change in the task track.

So i got some odd changes like:
image

Maybe it should be another issue but I need to know if this is the proper behaviour or not.

barredterra:
@rodrigoinformationsys this PR only syncs changes in Task to the corresponding Project, not the other way round. But please open a new issue for the bug you found :)

Don-Leopardo:
> I've done some testing in your branch and I've found an unexpected behaviour. Maybe it is how erpnext works as default.
> Testing the connection between the "Project Task" and the "Task" and everything went well. But I do use "Track Changes" in my projects and tasks so I can keep up with my team. When I change status from "Open" to "Working" of a task from project's page it doesn't show this change in the task track.
> 
> So i got some odd changes like:
> image
> 
> Maybe it should be another issue but I need to know if this is the proper behaviour or not.

That issue is caused by the line 167 in the project.py file:
https://github.com/frappe/erpnext/blob/55eca1bde2b9bfbec6cfae6c79b56df98a46b85c/erpnext/projects/doctype/project/project.py#L165-L168

The `db_update` method does not log the changes. You can try by changing it with a simple `save`. I don't know if this is a bug or not.

rmehta:
@Don-Leopardo please open another PR. There needs to be one PR for all supported versions.

Please also add tests

rmehta:
The task table should be reloaded when a new linked task is added. The is a UI problem

Don-Leopardo:
@rmehta What do you mean? Are you saying that this PR has that error in the UI or are you saying that all that we need to do is to solve that error in the UI? If it's the latest, it's not true. I've just finished a new test in the latest v11 and it stills does not create the record in the proyect's child table, even after reloading it.
If you are trying to say that this PR has that error and needs to be solved, then I get confused with the closing of it.
Can you resolve this to know if I have to keep working in the tests?

yrestom:
Hi
I do not think it's a UI problem, because when you add a task from a task doctype and link it to an existing project, the number of tasks within the project increases by one, but it does not appear within the task table within the project.
If you try to modify the project page, it will unlink the project from the newly added task.
In addition, any change to any task within the same task is not reflected on the tasks within the project, for example change the start date or end date or task name.

rmehta:
https://github.com/frappe/erpnext/pull/18059

Don-Leopardo:
@rmehta Are you planning to solve the issue in v11 too?

rmehta:
Don't think so, if you want to fix it, you will have to add test coverage for all the use cases (the tests don't exist in the first place). IMO most people won't use Task and Project as the entry point in the same session. 

But if there is a valid fix with tests, will take it.

barredterra:
> IMO most people won't use Task and Project as the entry point in the same session.

It fails even if you use only Task. My scenario in #17730 never touched the project's task table, but still Project will be unlinked from Tasks.

Issue #17776, opened 2019-05-28
[CRITICAL] Missing patch for changes in opportunity


Don-Leopardo:
There was a **big rework of the Opportunity doctype** and the corresponding patch to move the data from old fields to new fields is **missing**.
I'm talking about **fields** like `enquiry_from` that were **replaced by other fields**, in this case `opportunity_from`. **Another example** is `lead` that was replaced by `party_name` (I don't remember if there are more).

This changes were introduced in the **Release v11.1.23** (https://github.com/frappe/erpnext/releases/tag/v11.1.23).
More precisely with **the PR #17097** in **the commit https://github.com/frappe/erpnext/pull/17097/commits/af4d588f648f4c16e2228ad25fd4fc20ad6aa06b**

The **inexistence** of this patch is causing the following **problem**:
If you have an instance with **some opportunities** created using **a release older than v11.1.23** and proceed to update the version, **the opportunities will lose all the linked data** (the data of the changed fields).

PR #17777, opened 2019-05-28
fix: Patch for renaming enquiry_from to opportunity_from

deepeshgarg007:
Fixes https://github.com/frappe/erpnext/issues/17776

### DIFF
--- a/erpnext/patches/v11_1/move_customer_lead_to_dynamic_column.py
+++ b/erpnext/patches/v11_1/move_customer_lead_to_dynamic_column.py
@@ -10,5 +10,6 @@ def execute():
 	frappe.db.sql(""" UPDATE `tabQuotation` set party_name = customer WHERE quotation_to = 'Customer' """)
 
 	frappe.reload_doctype("Opportunity")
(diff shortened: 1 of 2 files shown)


erpnext/patches/v11_1/move_customer_lead_to_dynamic_column.py
```diff
@@ -10,5 +10,6 @@ def execute():
 	frappe.db.sql(""" UPDATE `tabQuotation` set party_name = customer WHERE quotation_to = 'Customer' """)
 
 	frappe.reload_doctype("Opportunity")
+	frappe.db.sql(""" UPDATE `tabOpportunity` set opportunity_from = enquiry_from where opportunity_from IS NULL""")
```
nabinhait: You cannot rerun this patch because in new opportunities, lead column will be blank and party_name will have value. Hence it will blank the party_name column.
Also in new accounts, there will be no lead column in the database. Hence, it should check whether lead column exists.

So the correct solution will be writing a new patch with the specific use case.

nabinhait (changes requested):

PR #17785, opened 2019-05-29
fix: Patch for renaming enquiry_from to opportunity_from

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/17776


Don-Leopardo (dismissed):
Good job. It seems to be a great patch. Thank you.

nabinhait (approved):

PR #17786, opened 2019-05-29
fix: trial balance opening balance not showing in the debit side for the liability account

rohitwaghchaure:
**Issue**

Screen Shot 2019-06-12 at 11 59 24 am

**After Fix**
Screen Shot 2019-06-12 at 12 29 16 pm

**At group level**

Screen Shot 2019-06-12 at 7 12 26 pm

### DIFF
--- a/erpnext/accounts/report/trial_balance/trial_balance.py
+++ b/erpnext/accounts/report/trial_balance/trial_balance.py
@@ -180,20 +180,28 @@ def calculate_values(accounts, gl_entries_by_account, opening_balances, filters,
 
 		if d["root_type"] == "Asset" or d["root_type"] == "Equity" or d["root_type"] == "Expense":
 			d["opening_debit"] -= d["opening_credit"]
-			d["opening_credit"] = 0.0
-			total_row["opening_debit"] += d["opening_debit"]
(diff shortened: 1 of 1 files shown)


nabinhait (changes requested):
@rohitwaghchaure 
I think opening / closing should be in either debit or credit column based on account type. And if the amount is negative, then switch the column.

rohitwaghchaure:
@nabinhait 

But in that case user may get confuse, for example let's say there i a liability account "ABC" in which user has credited 4000, and debited 5000 as opening entry. So with this opening credit will be -1000 for this account but after fix if we show it to debit side then it may be confusion for the user to understand how much they have debited or credited.

nabinhait (changes requested):
@rohitwaghchaure I think opening and closing columns refer balance amount, not individual dr/cr amount. Balance amount should be either in debit or credit.

It was in that way only for a long time and no user complained about it. Only thing need to fixed is, in case of negative balance the column should be switched.

rohitwaghchaure:
@nabinhait Changes are done, please check

nabinhait (approved):

PR #17790, opened 2019-05-29
fix: Customer and Lead dashboard fixes for opportunity and quotation

deepeshgarg007:
- Removed Dynamic link logic from frappe: https://github.com/frappe/frappe/pull/7591
- Handled the dynamic link logic for Lead  and Customer  Dashboard in ERPNext

### DIFF
--- a/erpnext/crm/doctype/lead/lead.js
+++ b/erpnext/crm/doctype/lead/lead.js
@@ -6,6 +6,18 @@ cur_frm.email_field = "email_id";
 
 erpnext.LeadController = frappe.ui.form.Controller.extend({
 	setup: function () {
+
+		this.frm.make_methods = {
+			'Quotation': () => erpnext.utils.create_new_doc('Quotation', {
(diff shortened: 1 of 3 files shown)


erpnext/crm/doctype/lead/lead.js
```diff
@@ -40,6 +40,32 @@ erpnext.LeadController = frappe.ui.form.Controller.extend({
 
 		if (!this.frm.doc.__islocal) {
 			frappe.contacts.render_address_and_contact(cur_frm);
+
+			let make_quotation = this.frm.dashboard.transactions_area.find('.btn-new[data-doctype="Quotation"]').unbind('click');
```
saurabh6790: Can we add handler via js hooks on `+` button? Like done for list view events ...

saurabh6790 (changes requested):

rmehta (approved):

PR #17791, opened 2019-05-29
perf: Remove MANIFEST.in

adityahase:
Faster ```pip install -e erpnext```

https://stackoverflow.com/questions/24727709/do-python-projects-need-a-manifest-in-and-what-should-be-in-it

https://github.com/frappe/frappe/pull/7360

Before this fix
```shell
time env/bin/pip install -v -e ./apps/erpnext  17.58s user 11.02s system 95% cpu 29.842 total
```

After 
```shell
time env/bin/pip install -v -e ./apps/erpnext  4.45s user 1.57s system 97% cpu 6.147 total
```


adityahase:
Also merge https://github.com/frappe/erpnext/pull/17792 and https://github.com/frappe/erpnext/pull/17793

scmmishra (approved):

PR #17794, opened 2019-05-29
fix(demo): Add System Manager role to Demo User

adityahase:
Without this ERPNext QMS cannot be accessed

Suggested by @netchampfaris


adityahase:
@jaichavan

saurabh6790 (changes requested):
We can assign any other role to QMS module or can create new roles instead of granting System Admin permissions.

PR #17801, opened 2019-05-30
fix: Dashboards for doctypes in accounts module

deepeshgarg007:
Requires: https://github.com/frappe/frappe/pull/7646


erpnext/accounts/doctype/pos_profile/pos_profile_dashboard.py
```diff
@@ -0,0 +1,19 @@
...
+
+from frappe import _
+
+
+def get_data():
+	return {
+		'fieldname': 'pos_profile',
+		'transactions': [
+			{
+				'label': _('Sales Invoices'),
```
rmehta: heading not required for single entries

rmehta (changes requested):
Add report link in dashboard (general feature)

erpnext/accounts/doctype/bank/bank_dashboard.py
```diff
@@ -0,0 +1,18 @@
...
+from frappe import _
+
+
+def get_data():
+	return {
+		'fieldname': 'bank',
+		'transactions': [
+			{
+				'label': _('Bank Deatils'),
+				'items': ['Bank Account', 'Bank Statement Transaction Entry', 'Bank Guarantee']
```
nabinhait: Remove 'Bank Statement Transaction Entry', it is not used anymore. I will write a patch to delete all related docs.

nabinhait (changes requested):

PR #17830, opened 2019-06-02
feat: provision to set multiple manufacturer against the item

rohitwaghchaure:
**Old Design**
Screen Shot 2019-06-02 at 4 08 52 pm

**New Design**
Screen Shot 2019-06-07 at 12 11 04 am

**Also added manufacturer and manufacturer_part_no fields in all the purchase related documents**


harshit-30:
@rohitwaghchaure we need to update the document for this, docs still has the older way.

Cc. @marination

PR #17834, opened 2019-06-03
fix: Serial no filtering issue for delivered items

deepeshgarg007:

### DIFF
--- a/erpnext/public/js/utils/serial_no_batch_selector.js
+++ b/erpnext/public/js/utils/serial_no_batch_selector.js
@@ -345,7 +345,7 @@ erpnext.SerialNoBatchSelector = Class.extend({
 
 		let serial_no_filters = {
 			item_code: me.item_code,
-			delivery_document_no: ["in", ""]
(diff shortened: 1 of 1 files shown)


erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -342,13 +342,24 @@ erpnext.SerialNoBatchSelector = Class.extend({
 	get_serial_no_fields: function() {
 		var me = this;
 		this.serial_list = [];
+
+		let serial_no_filters = {
+			item_code: me.item_code,
+			delivery_document_no: ["in", ""]
```
surajshetty3416: why?

can be:
```suggestion
			delivery_document_no: ""
```

deepeshgarg007: Fixed

deepeshgarg007: Fixed

nabinhait (approved):

PR #17837, opened 2019-06-03
feat: Dashboards for Human Resource module

Anurag810:
**Employee:**

Employee

**Employee Onboarding-Template:**

Employee-Onboarding-temp

**Employee Separation Template:**

Employee-seperartion-template

**Holiday list:**

Holiday list

**Job Applicant:**

Job Applicant

**Job Opening:**

Job Opening

**Leave Allocation:**

Leave Allocation

**Loan Application:**

Loan application

**Loan Type:**

Loan Type

**Payroll Period:**

Payroll Period

**Salary Structure:**

Salary Structure

**Shift Request:**

Shift Request

**Shift Type:**

Shift Type

**Staffing Plan:**

Staffing plan

**Traning Event:**

Training Event

**Vehicle:**

Vehicle

### DIFF
--- a/erpnext/hr/doctype/employee/employee_dashboard.py
+++ b/erpnext/hr/doctype/employee/employee_dashboard.py
@@ -13,23 +13,31 @@ def get_data():
 			},
 			{
 				'label': _('Payroll'),
-				'items': ['Salary Structure Assignment', 'Salary Slip', 'Timesheet']
+				'items': ['Salary Structure Assignment', 'Salary Slip', 'Additional Salary', 'Timesheet']
+			},
+			{
+				'label': _('Shift'),
+				'items': ['Shift Request', 'Shift Assignment']
 			},
 			{
 				'label': _('Expense'),
-				'items': ['Expense Claim']
+				'items': ['Expense Claim', 'Travel Request']
 			},
 			{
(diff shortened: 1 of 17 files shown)


erpnext/hr/doctype/employee/employee_dashboard.py
```diff
@@ -13,23 +13,31 @@ def get_data():
...
-				'label': _('Training'),
-				'items': ['Training Event', 'Training Result']
-			},
 			{
 				'label': _('Lifecycle'),
 				'items': ['Employee Transfer', 'Employee Promotion', 'Employee Separation']
-			}
+			},
+			{
+				'label': _('Benefit'),
```
rmehta: should be "Payroll"

rmehta (changes requested):

erpnext/hr/doctype/employee/employee_dashboard.py
Anurag810: I think there should be a different label name because payroll section already contains many links and is also larger. can we can split it ??

Anurag810: **Updated Employee dashboard: **

Screenshot 2019-06-14 at 2 07 02 PM

PR #17845, opened 2019-06-04
fix: General Ledger Group by Voucher not ordering by posting_date

### LINKED ISSUE #17707: [BUG] General Ledger Group by Voucher not ordering by posting_date
### Issue:
In the general ledger, when you use the option "Group by Voucher", the entries aren't ordered by posting_date.
Captura de pantalla 2019-05-22 a la(s) 15 03 06

### Tracking the error:
To track this error I analiced ​​the changes suffered by the `gl_entries` and the` gle_map` variables in the function `get_data_with_opening_closing`.
https://github.com/frappe/erpnext/blob/daa8db4d9584a6e6930e854ba00bd24af2b39b87/erpnext/accounts/report/general_ledger/general_ledger.py#L200-L207
The result of the analysis was that the `gl_entries` variable has the ordered data as espected. And the method `initialize_gle_map` is the one that deorders it. This method is the one that reagroup the entries by `voucher_no` but when doing so it overwrites the order gived by the SQL query.
https://github.com/frappe/erpnext/blob/daa8db4d9584a6e6930e854ba00bd24af2b39b87/erpnext/accounts/report/general_ledger/general_ledger.py#L254-L270

### Cheap solution:
The structure of `gle_map` is this:
```
{

    voucher_no: {
        'entries': list_of_entry_dicts,
        'totals': list_of_totals
        },
    ...
    ...
}
```
So we can add the smallest `posting_date` of the entries list to the voucher dict to have something like this:
```
{
    voucher_no: {
        'entries': list_of_entry_dicts,
        'totals': list_of_totals,
        'posting_date': smallest_posting_date
        },
    ...
    ...
}
```
And then proceed to reorder it by `posting_date`.

This is a cheap solution because it consists in a lot of `if` and `for` statements to determine if it's the "Group by Voucher" case or not. I think that this can be resolved in other way, maybe reestructuring the original query to include a `group by`.

Don-Leopardo:
fix #17707


rmehta (changes requested):
Please also raise corresponding PR for `develop` branch

Don-Leopardo:
@rmehta PR for develop: #18023

rohitwaghchaure:
@Don-Leopardo I have checked in my local account and found that both frappe._dict() and OrderedDict() giving the same result. Can you add the screenshot of general ledger report before and after fix.

Don-Leopardo:
@rohitwaghchaure sure. You are not seeing the difference probably because you don't have different types of documents with different dates. Here are the screenshots:
### Before
sincambios
### After
concambios

rohitwaghchaure (approved):

PR #17854, opened 2019-06-05
fix: Total amount fix in print for journal entry

deepeshgarg007:
Total amount in print in Journal Entry should be either Total Debit or Total Credit not sum of total_debit + total credit
However still not sure Total Amount makes sense in journal entry

Before:
image

After:
image

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.py
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.py
@@ -425,21 +425,10 @@ class JournalEntry(AccountsController):
 					pay_to_recd_from = frappe.db.get_value(d.party_type, d.party,
 						"customer_name" if d.party_type=="Customer" else "supplier_name")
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -425,21 +425,10 @@ def set_print_format_fields(self):
...
-
 		if pay_to_recd_from:
 			self.pay_to_recd_from = pay_to_recd_from
-			if bank_amount:
-				total_amount = bank_amount
-				currency = bank_account_currency
-			else:
-				total_amount = party_amount
-				currency = party_account_currency
+			total_amount = self.total_debit or self.total_credit
```
nabinhait: This is not right fix, let's discuss it offline.

nabinhait (changes requested):

PR #17874, opened 2019-06-07
fix: Make Travel Request Document Submittable

kennethsequeira:
- Changed doctype to submittable
- Added patch to submit entries older than 30 days


erpnext/patches/v12_0/make_travel_request_submittable.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
+# License: GNU General Public License v3. See license.txt
+
+#Patch submits Travel Requests which are older than 30 days
+from __future__ import unicode_literals
+import frappe
+
+ def execute():
+    a = frappe.get_list("Travel Request", 
+                        filters={"creation": ["<", frappe.utils.add_days(frappe.utils.datetime.datetime.now(), -30)]}) #Get Travel Requests which are older than 30 days
```
nabinhait: Submitting old records older than 30 days (some random number) is not a good idea. Let's keep existing docs in draft. User can submit those in bulk from List view.

nabinhait (changes requested):

erpnext/patches/v12_0/make_travel_request_submittable.py
kennethsequeira: done.

nabinhait (approved):

PR #17879, opened 2019-06-07
fix: Total amount fix  in print for journal entry

deepeshgarg007:
Before:
Screenshot 2019-06-05 at 2 37 28 PM

After:
Screenshot 2019-06-05 at 2 37 40 PM


sahil28297:
@frappe-bot create instance

nabinhait (approved):

PR #17882, opened 2019-06-07
feat: Upgrade Daily Work Summary to Regular Work Summary

aleksas:
**Regular Work Summary** DocType added to erpnext.hr. It was cloned from **Daily Work Summary** with the addition of frequency (daily, weekly, monthly and yearly) option. Also **Daily Work Summary Group User** has additional receive field to indicate whether to receive **Request**, **Summary** or **Both**.

[Discussed HERE](https://discuss.erpnext.com/t/how-to-make-monthly-work-summary-report/42223/16?u=aleksas)

The unit tests are taken from Daily Work Summary, but additional tests for expanded functionality are missing.
Also migration code from Daily Work Summary was not added.

image

P.S.
I may be wrong but running Daily Work Summary (or similarly Regular Work Summary) test using command `bench run-tests --module "erpnext.hr.doctype.daily_work_summary.test_daily_work_summary"` should fail instead of indicating 3 passed tests. It should depend on [presence of a "_Test Email Account 1"](https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/daily_work_summary/test_daily_work_summary.py#L44), which I presume, is created when preparing global testing environment, but since the call was just for module said email account was not created.


erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
```diff
@@ -0,0 +1,175 @@
...
+  {
+   "default": "January",
+   "depends_on": "eval:doc.send_emails_frequency == \"Yearly\"",
+   "fieldname": "send_request_emails_month",
+   "fieldtype": "Select",
+   "label": "Month",
+   "options": "January\nFebruary\nMarch\nApril\nMay\nJune\nJuly\nAugust\nSeptember\nOctober\nNovember\nDecember"
+  },
+  {
+   "fieldname": "send_summary_emails_at",
```
hrwX: Field `send_request_emails_at` has been repeated.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
```diff
@@ -0,0 +1,175 @@
...
+   "reqd": 1
+  },
+  {
+   "fieldname": "send_request_emails_at",
+   "fieldtype": "Select",
+   "label": "Time",
+   "options": "00:00\n01:00\n02:00\n03:00\n04:00\n05:00\n06:00\n07:00\n08:00\n09:00\n10:00\n11:00\n12:00\n13:00\n14:00\n15:00\n16:00\n17:00\n18:00\n19:00\n20:00\n21:00\n22:00\n23:00"
+  },
+  {
+   "default": "Friday",
```
hrwX: Any particular reason to set Friday as default Day for sending Regular Work Summary ?

erpnext/hr/doctype/regular_work_summary/test_regular_work_summary.py
```diff
@@ -0,0 +1,95 @@
...
+# Copyright (c) 2019, Frappe Technologies Pvt. Ltd. and Contributors
+# See license.txt
+from __future__ import unicode_literals
+
+import os
+import frappe
+import unittest
+import frappe.utils
+
+# test_records = frappe.get_test_records('Regular Work Summary')
```
hrwX:
```suggestion
```

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
```diff
@@ -0,0 +1,175 @@
...
+   "depends_on": "eval:doc.send_emails_frequency == \"Yearly\"",
+   "fieldname": "send_request_emails_month",
+   "fieldtype": "Select",
+   "label": "Month",
+   "options": "January\nFebruary\nMarch\nApril\nMay\nJune\nJuly\nAugust\nSeptember\nOctober\nNovember\nDecember"
+  },
+  {
+   "fieldname": "send_summary_emails_at",
+   "fieldtype": "Select",
+   "label": "Time",
```
hrwX: Why not use Time DocField rather than Select DocField ?

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: Just had an idea that Friday is likely to be a  common option for weekly reports... Do you have any policy for this sort of choices?

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: That was in Daily Work Summary at least in master branch. If I understand correctly the reason not to use Time for the field is that the scheduler has an hour frequency and consequently there is no practical reason in setting time with second precision.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: One is `send_request_emails_at`  the other is `send_summary_emails_at`

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
```diff
@@ -0,0 +1,175 @@
+{
+ "autoname": "Prompt",
+ "creation": "2019-06-01 20:33:50.300725",
+ "doctype": "DocType",
+ "editable_grid": 1,
+ "engine": "InnoDB",
+ "field_order": [
+  "enabled",
+  "select_users",
+  "users",
```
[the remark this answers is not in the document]
aleksas: This is the way it was for Daily Work Summary. Don't know original reasons. @surajshetty3416 any comment on that?

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: The time is too detailed with second being the smallest unit. The send email at field is meant for scheduler to call a trigger function. This scheduler calls the trigger function each hour. There is no reason to overwhelm user with unnecessary time precision.

aleksas:
@hrwX any particular reason Quality Review started disliking unused javascript function parameter in [HERE](https://app.codacy.com/app/SaiFi0102/erpnext/pullRequest?prid=3690034) ? It's the same as [HERE in Daily Work Summary](https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/daily_work_summary/daily_work_summary.js). How should I fix it? Deleting this js file is ok?

hrwX:
> HERE

Just comment out the refresh function, don't delete the file.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
hrwX: Let the user decide which day, make it mandatory and remove the default.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
hrwX: We have a scheduler activity that triggers every 4 mins.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: That would be wrong to make it mandatory because it's not always visible. If you choose Daily, Monthly or Yearly you don't see the week day field. Will set to Monday both fields.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: Still doesn't make much sense to have Time field with second precision and backend frequency of 4 min., that would be confusing for the end user.

erpnext/hr/doctype/regular_work_summary_group/regular_work_summary_group.json
aleksas: Unless mandatory field constraint is checked against `eval:doc.send_emails_frequency == "Weekly";` and does't raise an error for empty field or Select field can not be empty which would mean some sort of default value...

rmehta (changes requested):
@aleksas makes no sense having both these features, please remove Daily Work Summary and write a patch to convert existing ones to Regular Work Summary.

Please also add docs

rmehta:
@aleksas a better name is acutally "Status Update". If you are renaming it 😉

aleksas:
> @aleksas a better name is acutally "Status Update". If you are renaming it

@rmehta  You mean Regular Status Update? Or just Status Update? Or what exactly should be renamed to Status Update?

aleksas:
> @aleksas makes no sense having both these features, please remove Daily Work Summary and write a patch to convert existing ones to Regular Work Summary.
> 
> Please also add docs

@rmehta a link to an example path that is most similar to solving this task would be greatly appreciated :)

aleksas:
Also what should be the patching procedure?
Create new Summary docs from Daily Work Summary and delete the old.
Or rename Daily Work Summary to new name and add features?

aleksas:
IMHO status update may not be a good name in some sense. status update implies a status change for a finite process like a project with start and end at an expected point in time in future also status implies some formal states. And there is Project/tasks/boards for that. Daily Work Summary has a freedom of describing arbitrary work done within a regular time interval. It doesn't impose additional structure apart from time interval. It's just a subtle feeling from personal experience.

PR #17895, opened 2019-06-10
fix(Utils): Do not add duplicate item to items

hrwX:
Do not add duplicate item to items while creating Sales Invoice from same Delivery Note
Screenshot (86)


erpnext/public/js/utils.js
```diff
@@ -617,6 +618,26 @@ erpnext.utils.map_current_doc = function(opts) {
...
+		item_code => [qty, warehouse, batch_no]
+	*/
+	let items = [];
+
+	for (let item of cur_frm.doc.items) {
+		if (!(unique_items.has(item.item_code) && unique_items.get(item.item_code)[0] === item.qty &&
+			unique_items.get(item.item_code)[1] === item.warehouse && unique_items.get(item.item_code)[2] === item.batch_no &&
+			unique_items.get(item.item_code)[3] === item.delivery_date && unique_items.get(item.item_code)[4] === item.required_date)) {
+
+			unique_items.set(item.item_code, [item.qty, item.warehouse, item.batch_no, item.delivery_date, item.required_date]);
```
nabinhait: Rate can also be different

nabinhait (dismissed):

nabinhait (approved):

PR #17896, opened 2019-06-10
fix(Utils): Do not add duplicate items

hrwX:
Do not add duplicate item to items while creating Sales Invoice from same Delivery Note
Screenshot (86)


erpnext/public/js/utils.js
```diff
@@ -590,6 +591,20 @@ erpnext.utils.map_current_doc = function(opts) {
 	}
 }
 
+erpnext.utils.clear_duplicates = function() {
+	const unique_items = new Map();
+	let items = []
+
+	for (let i = 0; i < cur_frm.doc.items.length; i++) {
```
Alchez:
```suggestion
	for (let item of cur_frm.doc.items) {
```

You can remove the `item` assignment on the next line as well.

erpnext/public/js/utils.js
```diff
@@ -590,6 +591,20 @@ erpnext.utils.map_current_doc = function(opts) {
...
+		let item = cur_frm.doc.items[i];
+		if (!(unique_items.has(item.item_code) && unique_items.get(item.item_code) === item.qty)) {
```
Alchez: This will misbehave if you are delivering from different warehouses, or have two different batches for an item with the same quantity.

Alchez (changes requested):

erpnext/public/js/utils.js
```diff
@@ -590,6 +591,25 @@ erpnext.utils.map_current_doc = function(opts) {
...
+		Create a Map of items with
+		item_code => [qty, warehouse, batch_no]
+	*/
+	let items = [];
+
+	for (let item of cur_frm.doc.items) {
+		if (!(unique_items.has(item.item_code) && unique_items.get(item.item_code)[0] === item.qty &&
+			unique_items.get(item.item_code)[1] === item.warehouse && unique_items.get(item.item_code)[2] === item.batch_no)) {
+
+			unique_items.set(item.item_code, [item.qty, item.warehouse, item.batch_no]);
```
rmehta: in case of PO/SO `delivery_date`, `required_date` is also unique

hrwX: Added `delivery_date` and `required_date` check

Alchez:
@hrwX @rmehta 

Is this check really necessary? The condition set seems to be increasing every day, which makes me think we might be missing some other critical field(s) as well.

I think we should err on the side of having duplicate items, than forcefully removing seemingly similar item lines that are actually different.

hrwX:
> @hrwX @rmehta
> 
> Is this check really necessary? The condition set seems to be increasing every day, which makes me think we might be missing some other critical field(s) as well.
> 
> I think we should err on the side of having duplicate items, than forcefully removing seemingly similar item lines that are actually different.

I think so `frappe.throw` so do the trick, because if it would have been just name and quantity, then this fix would have been good enough, but now as you said @Alchez there are too many conditions. What do you think @rmehta @nabinhait

erpnext/public/js/utils.js
```diff
@@ -590,6 +591,26 @@ erpnext.utils.map_current_doc = function(opts) {
...
+		item_code => [qty, warehouse, batch_no]
+	*/
+	let items = [];
+
+	for (let item of cur_frm.doc.items) {
+		if (!(unique_items.has(item.item_code) && unique_items.get(item.item_code)[0] === item.qty &&
+			unique_items.get(item.item_code)[1] === item.warehouse && unique_items.get(item.item_code)[2] === item.batch_no &&
+			unique_items.get(item.item_code)[3] === item.delivery_date && unique_items.get(item.item_code)[4] === item.required_date)) {
+
+			unique_items.set(item.item_code, [item.qty, item.warehouse, item.batch_no, item.delivery_date, item.required_date]);
```
nabinhait: Rate can also be different

nabinhait (dismissed):

nabinhait (approved):

PR #17917, opened 2019-06-12
fix(stock): Remove hardcoded validation for Item

Alchez:
Made to develop at #17918.

This is mostly to remove legacy code (the logic removed here was added ~6 years ago, before form validations would occur on the front-end).


vjFaLk (approved):

Anurag810:
What if an item is created through API then there should be a validation for that.

Alchez:
The mandatory validations will still occur for a newly created document.

rmehta:
Lets avoid making changes in Version 11 unless its a bug

PR #17918, opened 2019-06-12
fix(stock): Remove hardcoded validation for Item

Alchez:
Made to hotfix at #17917.

This is mostly to remove legacy code (the logic removed here was added ~6 years ago, before form validations would occur on the front-end).


Anurag810:
What if an item is created through API then there should be a validation for that.

Alchez:
The mandatory validations will still occur for a newly created document.

PR #17925, opened 2019-06-13
feat(Sales/Purchase Order): optional to reference a Blanket Order

ronelvcabrera:
This feature is to be able to have an option to reference or not a Blanket Order when creating a Sales/Purchase Order.

We encounter a customer that doesn't want a Sales Order be under a existing Blanket Order. Currently, when creating a Sales Order that has the same details on an existing Blanket Order, system will automatically set a Blanket Order for said SO.

We added a checkbox that will cater such scenario. Ticking the check box will automatically set the Blanket Order and if not ticked, system will leave it as blank.

Feature is applied to both Purchase Order and Sales Order.

We added test case for both PO and SO to validate that the feature is working as expected.

Sales Order
sales order bo optional fix

Purchase Order
purchase order bo optional fix

### DIFF
--- a/erpnext/buying/doctype/purchase_order/test_purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/test_purchase_order.py
@@ -15,6 +15,7 @@ from erpnext.stock.doctype.material_request.test_material_request import make_ma
 from erpnext.stock.doctype.material_request.material_request import make_purchase_order
 from erpnext.stock.doctype.stock_entry.test_stock_entry import make_stock_entry
 from erpnext.controllers.accounts_controller import update_child_qty_rate
+from erpnext.manufacturing.doctype.blanket_order.test_blanket_order import make_blanket_order
 import json
 
 class TestPurchaseOrder(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -209,7 +209,8 @@ def get_basic_details(args, item):
 			project: "",
 			qty: "",
 			stock_qty: "",
-			conversion_factor: ""
+			conversion_factor: "",
+			against_blanket_order: 1/0
```
rmehta: why `1/0`

erpnext/buying/doctype/purchase_order/test_purchase_order.py
```diff
@@ -583,6 +584,25 @@ def test_advance_payment_entry_unlink_against_purchase_order(self):
...
+	def test_PO_optional_blanket_order(self):
+		"""
+			Expected result: Blanket order Ordered Quantity should only be affected on Purchase Order with against_blanket_order = 1.
+			Second Purchase Order should not add on to Blanket Orders Ordered Quantity.
+		"""
+
+		bo = make_blanket_order(blanket_order_type = "Purchasing", quantity = 10, rate = 10)
+
+		po = create_purchase_order(item_code= "_Test Item", qty = 5, against_blanket_order = 1)
+		self.assertTrue(frappe.get_doc('Purchase Order', po.get('name')))
```
rmehta: seems like a meaningless assertion

erpnext/buying/doctype/purchase_order/test_purchase_order.py
```diff
@@ -583,6 +584,25 @@ def test_advance_payment_entry_unlink_against_purchase_order(self):
 		frappe.db.set_value("Accounts Settings", "Accounts Settings",
 			"unlink_advance_payment_on_cancelation_of_order", 0)
 
+
+
+
+	def test_PO_optional_blanket_order(self):
```
rmehta: use lowercase method names `test_po_optional_blanket_order`

rmehta (changes requested):
Why is the reference added? Will there be additional validations?

rmehta:
👍 for first PR

erpnext/stock/get_item_details.py
ronelvcabrera: this is to be inline with the field type values for check box which is 1 or 0

erpnext/buying/doctype/purchase_order/test_purchase_order.py
ronelvcabrera: will check and improve this test case.

ronelvcabrera:
@rmehta No more additional validation. 

Also, updated PR with the requested change. Kindly review.

erpnext/stock/get_item_details.py
rmehta: 1/0 = ZeroDivisionError

ronelvcabrera: will change to be more clear

ronelvcabrera: Updated line. Copied line above `ignore_pricing_rule: 0/1`

PR #17929, opened 2019-06-13
feat(customer): Add report to show item prices per Customer

Alchez:
Made to develop at #17930.

<hr>

**Reason:**

Different Customers might have different prices per Item. Having an accessible Customer Price List makes it easy to export / send for reference.

**Screenshots / GIFs:**

image

image


erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
```diff
@@ -0,0 +1,101 @@
...
+
+def get_data(filters=None):
+	data = []
+	customer_details = get_customer_details(filters)
+	items = get_selling_items(filters)
+
+	for item in items:
+		price_list_rate = get_price_list_rate_for(customer_details, item.item_code) or 0.0
+		available_stock = frappe.db.sql("SELECT sum(actual_qty) FROM `tabBin` WHERE item_code = %s", item.item_code)
+		available_stock = available_stock[0][0] if available_stock else None
```
saurabh6790: Why available qty?

saurabh6790 (dismissed):

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: It helps sales reps in the field to quickly pull up a customer's price list and see if that item is available for order.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
```diff
@@ -0,0 +1,101 @@
...
+		}
+	]
+
+
+def get_data(filters=None):
+	data = []
+	customer_details = get_customer_details(filters)
+	items = get_selling_items(filters)
+
+	for item in items:
```
Anurag810: Don't use query inside the loop. Guess what will happen for 1000000000+ items.
Try to optimize that.
Suggestion:
First, get all the item code at once and then try to use IN operator of SQL.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.json
```diff
@@ -0,0 +1,43 @@
+{
+ "add_total_row": 0,
```
Anurag810: set :```"add_total_row": 0,```

Anurag810 (dismissed):

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: Fixed the query.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.json
Alchez: Why though? The sum of rates or stock is not meaningful.

PR #17930, opened 2019-06-13
feat(customer): Add report to show item prices per Customer

Alchez:
Made to hotfix at #17929.

<hr>

**Reason:**

Different Customers might have different prices per Item. Having an accessible Customer Price List makes it easy to export / send for reference.

**Screenshots / GIFs:**

image

image


erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def get_data(filters=None):
+	data = []
+	customer_details = get_customer_details(filters)
+	items = get_selling_items(filters)
+
+	for item in items:
+		price_list_rate = get_price_list_rate_for(customer_details, item.item_code) or 0.0
+		available_stock = frappe.db.sql("SELECT sum(actual_qty) FROM `tabBin` WHERE item_code = %s", item.item_code)
```
saurabh6790: Why available qty? Seems no resemblance.

saurabh6790 (dismissed):

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: It helps sales reps in the field to quickly pull up a customer's price list and see if that item is available for order.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
```diff
@@ -0,0 +1,101 @@
...
+		}
+	]
+
+
+def get_data(filters=None):
+	data = []
+	customer_details = get_customer_details(filters)
+	items = get_selling_items(filters)
+
+	for item in items:
```
Anurag810, line 70: Don't use query inside the loop. Guess what will happen for 1000000000+ items.
Try to optimize that.
**Suggestion:**
First, get all the item code at once and then try to use ```IN``` operator of SQL.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.json
```diff
@@ -0,0 +1,43 @@
+{
+ "add_total_row": 0,
```
Anurag810, line 2: Set  ```"add_total_row": 1,```

Anurag810 (changes requested):

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: Good point. I'll keep that query outside the loop.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.json
Alchez: Why though? The sum of rates or stock is not meaningful.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: Fixed the query.

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Anurag810: first, get all items and convert it into the list.
``` items = [d.item_code for d in frappe.get_all("Item", filters=item_filters, fields=["item_code", "item_name"], order_by="item_name")]```

And then you can try

```actual_qty = frappe.get_all("Bin", filters = [('item_code', 'IN', items)], fields = ["actual_qty"])```

After that sum up only two queries is required.
so first, it will bring all data.

**Review by Anurag810 — CHANGES_REQUESTED** [review]

File: erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+	customer_details = get_party_details(party=filters.get("customer"), party_type="Customer")
+	customer_details.update({
+		"company": get_default_company(),
+		"price_list": customer_details.get("selling_price_list")
+	})
+
+	return customer_details
+
+
+def get_selling_items(filters):
```
Anurag810, line 95: convert it into the list as told above

Anurag810 (dismissed):

erpnext/selling/report/customer_wise_item_price/customer_wise_item_price.py
Alchez: @Anurag810 You might be looking at an older version of the code. I already fixed this with a single `get_all` query on Bin, and then I process it later.

Anurag810: Yeah, my mistake.

surajshetty3416 (approved):

PR #17933, opened 2019-06-13
fix: column sequence and add total hours

Anurag810:
interchange column sequence and added total worked hours

Screenshot 2019-06-13 at 3 23 56 PM


rohitwaghchaure:
@Anurag810 can you added screenshots of before change and after change?

scmmishra (approved):

PR #17942, opened 2019-06-13
fix: Show product page even if cart is not enabled

netchampfaris:
If cart is disabled, the `/contact` page is shown. Instead it should just hide the **Add to Cart** button and the Cart. This PR does just that.


Anurag810:
Please add some description and screenshot.

netchampfaris:
Travis passed: https://travis-ci.com/frappe/erpnext/builds/115459999

The status is not updated here for some reason.

surajshetty3416 (approved):

PR #17943, opened 2019-06-13
fix: User MultiSelectList control for Party filter

netchampfaris:
The comma based filter wouldn't work for values that had comma in them

multiselect-list

Depends on https://github.com/frappe/frappe/pull/7698


rohitwaghchaure:
@netchampfaris Multiselect is used in multiple reports, you have made the changes in one report only. Can you do this changes in all the reports where multiselect fieldtype is used?

netchampfaris:
Depends on https://github.com/frappe/frappe/pull/7713

PR #17948, opened 2019-06-14
fix: filter reference name by cost center in journal entry account

rushparikh:
In Journal Entry Account if cost center is selected then filter reference_name by it. This is only for Sales Invoice and Purchase Invoice reference type.


rushin29:
Any updates on this??

nabinhait (approved):

PR #17975, opened 2019-06-18
feat: purchase order against sales order

Anurag810:
taskbyrohit

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.js
+++ b/erpnext/selling/doctype/sales_order/sales_order.js
@@ -129,28 +129,11 @@ erpnext.selling.SalesOrderController = erpnext.selling.SellingController.extend(
 					   me.frm.cscript.update_status('Re-open', 'Draft')
 				   }, __("Status"));
 			   }
-		    }
+			}
 			if(doc.status !== 'Closed') {
 				if(doc.status !== 'On Hold') {
-					for (var i in this.frm.doc.items) {
-						var item = this.frm.doc.items[i];
-						if(item.delivered_by_supplier === 1 || item.supplier){
-							if(item.qty > flt(item.ordered_qty)
-								&& item.qty > flt(item.delivered_qty)) {
-								allow_purchase = true;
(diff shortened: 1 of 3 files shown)


Anurag810:
Removed add row and delete buttons from table field.
Updated screenshot for that.
Screenshot 2019-06-18 at 4 18 32 PM

erpnext/selling/doctype/sales_order/sales_order.js
```diff
@@ -129,28 +129,11 @@ erpnext.selling.SalesOrderController = erpnext.selling.SellingController.extend(
...
-							if(item.qty > flt(item.delivered_qty)) {
-								allow_delivery = true;
-							}
-						}
 
-						if (allow_delivery && allow_purchase) {
-							break;
-						}
-					}
+					allow_purchase = true
```
nabinhait: remove allow_purchase

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -765,7 +765,10 @@ def get_events(start, end, filters=None):
 	return data
 
 @frappe.whitelist()
-def make_purchase_order_for_drop_shipment(source_name, for_supplier=None, target_doc=None):
+def make_purchase_order_for_drop_shipment(source_name, for_supplier=None, selected_items=[], target_doc=None):
```
nabinhait: As now PO can be made against any item in SO, remove the drop_shipment term from function naming

nabinhait (changes requested):

erpnext/selling/doctype/sales_order/sales_order.js
Anurag810: done

erpnext/selling/doctype/sales_order/sales_order.py
Anurag810: done

nabinhait (approved):

PR #17989, opened 2019-06-19
Shopping Cart Bug: Price list rate error with Pricing Rule affecting cart total 

sapna-kcs:
Ref: https://discuss.erpnext.com/t/pricing-rule-on-shopping-cart/17715/3

Pricing rule is applied over minimum and maximum qty of an item. After max qty, Price list rate should apply only on the new qty added. However, it is impacting all qty for the item and giving incorrect cart totals.

**Eg:** Item Price List rate is $100;
Pricing rule of 50% discount is applied for minimum 2 and maximum 5 qty for this Item, ie. discounted Rate = $50.

Within shopping cart, if user adds 8 qty of this item, then upto 5 pieces pricing rule rate should apply while additional 3 pieces should add at price list rate. However, after 5 qty, price list rate is applying over all 8 pieces, instead of only 3 pieces which is giving incorrect cart totals.

**Current cart total**: 8 * $100 = $800
while **Correct cart total** should be: (5 * $50) + (3 * $100) = $550

This has been resolved with solution video attached.

[Pricing rule bug - 2.zip](https://github.com/frappe/erpnext/files/3305494/Pricing.rule.bug.-.2.zip)


rmehta:
@sapna-kcs can you please rebase and reopen with your commits only

sapna-kcs:
> @sapna-kcs can you please rebase and reopen with your commits only

Its a Fix of v11 bug regards pricing rule which has been resolved. Pl suggest which option to select to make a pull in the right base.
image

PR #17993, opened 2019-06-19
feat: "Order Type" tree type for Sales Analytics report

Don-Leopardo:
### Description
This PR adds the "Order Type" option to the tree type filter of the Sales Analytics report.
Captura de pantalla 2019-06-19 a la(s) 09 32 54

### Utility
This option is very useful to see from where does the orders come and to compare, for example, how many orders came from the Shopping Cart and how many are from the Sales department.

### Possible Improvements
Today this only works for Sales Order, but maybe it can be improved to work with Sales Invoice and Delivery Note. I'm open to your ideas.
Another possible improvement is to make "Order Type" a link to a new doctype to make it self-administered, because this can be more useful with "Order Types" like "Shopify Order" or things like that.

**Note:** This PR can be discussed and some improvements can be made. Don't be afraid to speak up and express your ideas.

### DIFF
--- a/erpnext/selling/report/sales_analytics/sales_analytics.js
+++ b/erpnext/selling/report/sales_analytics/sales_analytics.js
@@ -8,7 +8,7 @@ frappe.query_reports["Sales Analytics"] = {
 			fieldname: "tree_type",
 			label: __("Tree Type"),
 			fieldtype: "Select",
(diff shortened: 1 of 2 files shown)


rmehta (changes requested):
new features in `develop` only

PR #17999, opened 2019-06-19
fix: Move links to appropriate section

netchampfaris:
Before:
image

After:
image


rmehta (changes requested):
Add one more section, "Accounting Entries" and move `Journal Entry` and `Payment Entry` there

netchampfaris:
A new section with two links?

image

nabinhait (approved):

PR #18007, opened 2019-06-20
fix(BOM): set default to 0 in include_exploded_items

sahil28297:


nabinhait (changes requested):
PR for develop branch?

sahil28297:
https://github.com/frappe/erpnext/pull/18017

PR #18009, opened 2019-06-20
fix: precision for certain calculations in buying controller

Mangesh-Khairnar:
Fixes:
calculation_problem

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -417,7 +417,6 @@ class BuyingController(StockController):
 
 	# validate accepted and rejected qty
 	def validate_accepted_rejected_qty(self):
-		float_precision = frappe.get_single("System Settings").float_precision
 		for d in self.get("items"):
 			self.validate_negative_quantity(d, ["received_qty","qty", "rejected_qty"])
 			if not flt(d.received_qty) and flt(d.qty):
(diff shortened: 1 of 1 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -428,8 +429,9 @@ def validate_accepted_rejected_qty(self):
 			elif not flt(d.rejected_qty):
 				d.rejected_qty = flt(d.received_qty) -  flt(d.qty)
 
+			val  = flt(d.qty) + flt(d.rejected_qty)
 			# Check Received Qty = Accepted Qty + Rejected Qty
-			if ((flt(d.qty) + flt(d.rejected_qty)) != flt(d.received_qty)):
+			if (flt(val, float_precision) != flt(d.received_qty)):
```
nabinhait: Rounding for `val` and `d.received_qty` should be based on same precision. I think in this we can do it based on precision of "received_qty" field instead of system settings precision.

nabinhait (changes requested):

erpnext/controllers/buying_controller.py
Mangesh-Khairnar: The received_qty has a field type float, which takes precision from system settings, on saving received qty will get rounded to the float precision from system settings. So shouldn't rounding be based on that float precision?

nabinhait (approved):

PR #18020, opened 2019-06-21
feat: Product Bundle in Purchase Cycle

bhavishyasharma:
Hi everyone!

I have implemented the bundle in Buying Cycle. The valuation of child items is calculated based on the weightage which is defined in the ProductBundle

Steps:

1. Add weightage field to Product Bundle.
image
2. Add weightage field to Packed Item.
3. Add rate and amount fields to Packed Item.
image

4. Add packed_items to Purchase Order, Purchase Receipt, Purchase Invoice.
Screenshot 2019-06-04 at 23 18 59

5. Pass stock_entry for Purchase Receipt and Purchase Invoice.


erpnext/controllers/buying_controller.py
```diff
@@ -487,6 +487,39 @@ def update_stock_ledger(self, allow_negative_stock=False, via_landed_cost_vouche
...
+								"actual_qty": flt(p.qty),
+								"serial_no": cstr(p.serial_no).strip()
+							})
+							if self.is_return:
+								original_incoming_rate = frappe.db.get_value("Stock Ledger Entry",
+																			 {"voucher_type": "Purchase Receipt",
+																			  "voucher_no": self.return_against,
+																			  "item_code": p.item_code},
+																			 "incoming_rate")
+
```
rohitwaghchaure, line 512: Remove extra spaces

bhavishyasharma:
@rohitwaghchaure please review

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -875,6 +876,12 @@ def set_tax_withholding(self):
 		# calculate totals again after applying TDS
 		self.calculate_taxes_and_totals()
 
+	def update_packing_list(self):
+		if cint(self.update_stock) == 1:
+			from erpnext.stock.doctype.packed_item.packed_item import make_packing_list
+			make_packing_list(self)
+			self.calculate_packing_list_rates()
```
rohitwaghchaure: As you have added method calculate_packing_list_rates in buying_controller.py file why you're calling this method from individual file like purchase_invoice.py, purchase_order.py, purchase_receipt.py
Call the method calculate_packing_list_rates from the validate method of the buying_controller.py and put the condition of the doctype of which you want to call this method

bhavishyasharma: I did that because other docs like MaterialRequest use BuyingContoller and then I would have to check for conditions. This will add extra logic for each document extending BuyingController. If you think that adheres to erpnext guidelines better, I can do that.

rohitwaghchaure: No need of extra logic you have to only add doctype condition like
```
If self.doctype in ['Purchase Order', 'Purchase Invoice', 'Purchase Receipt']:
    make_packing_list(self)
    self.calculate_packing_list_rates()
```

erpnext/controllers/buying_controller.py
```diff
@@ -488,6 +493,39 @@ def update_stock_ledger(self, allow_negative_stock=False, via_landed_cost_vouche
...
+								})
+							sl_entries.append(sle)
+
+						if flt(d.rejected_qty) != 0:
+							sl_entries.append(self.get_sl_entries(p, {
+								"warehouse": d.rejected_warehouse,
+								"actual_qty": flt(d.rejected_qty)*(flt(p.qty)/flt(d.qty)),
+								"serial_no": cstr(d.rejected_serial_no).strip(),
+								"incoming_rate": 0.0
+							}))
```
rohitwaghchaure, line 530: sl entries for product bundle items can be commonify with non product bundle items. 
For example
```
items = frm.doc.items + packing items
for d in items:
    create the sl entries 
```

bhavishyasharma: I avoided that for various reasons

1. Need to change AccountsController - set_stock_items method is defined in AccountsController
```
stock_items = self.get_stock_items()

for d in self.get('items'):
	if d.item_code in stock_items and d.warehouse:
```

2. There are various fields in item that are needed for packed_item and are not copied
```
conversion_factor
rejected_warehouse
rejected_qty
```
these fields are calculated while making sl entries. these fields will be need to be added to PackedItem

PR #18025, opened 2019-06-21
fix: Wrong Casual Leave Opening in Employee Leave Balance Report

Anurag810:
Before a casual opening balance in the leave balance report was not getting value correct.
Now it is fixed.

### DIFF
--- a/erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
+++ b/erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _
 from erpnext.hr.doctype.leave_application.leave_application \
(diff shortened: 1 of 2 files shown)


erpnext/hr/report/employee_leave_balance/employee_leave_balance.py
```diff
@@ -51,7 +54,8 @@ def get_data(filters, leave_types):
 					filters.from_date, filters.to_date)
 
 				# opening balance
-				opening = get_total_allocated_leaves(employee.name, leave_type, filters.to_date)
```
rohitwaghchaure: remove import get_total_allocated_leaves and also remove the method get_total_allocated_leaves from the leave_application.py file

Anurag810: done

PR #18059, opened 2019-06-25
BREAKING CHANGE: Remove anti-pattern "Project Task"

rmehta:
The child table "Project Task" was a UX hack so that the user can quickly add tasks to project. This type of 2-way syncing is an anti-pattern and source of bugs and performance problems.

#TheBestCodeIsTheOneThatIsNotWritten


Don-Leopardo:
Are you planning to solve the issue in v11 too?

erpnext/hr/doctype/employee_onboarding/test_employee_onboarding.py
```diff
@@ -12,7 +12,7 @@
 class TestEmployeeOnboarding(unittest.TestCase):
 	def test_employee_onboarding_incomplete_task(self):
 		if frappe.db.exists('Employee Onboarding', {'employee_name': 'Test Researcher'}):
```
Alchez:
```suggestion
		frappe.delete_doc_if_exists('Employee Onboarding', {'employee_name': 'Test Researcher'})
```

nabinhait (dismissed):

cjpit:
Awesome, that was driving me insane as the first example I had to erpnext ;) - was very confusing as to why it was there ;)

rodrigoinformationsys:
> Are you planning to solve the issue in v11 too?

Any news on this?

Don-Leopardo:
Sorry, I don't know how to do the tests an I'm doing other things right now so, the pr is still without tests

rodrigoinformationsys:
So everyone using V11 is facing this Issue? Because I don't know any solution in the time been. If my team updates the projects expected date it changes every task in that project.
We are facing this right now, +160 tasks got their dates changed because of the bug in V11.

Don-Leopardo:
We are not suffering it because we've apply our PR in our instance, but yes, v11's projects/task relation is failing.
You can use it too if you want. The code is here: #17774

PR #18066, opened 2019-06-25
fix(item): fix patch for barcode childtable migration in item

karthikeyan5:
fix for `item_barcode_childtable_migrate` patch


nabinhait (dismissed):

nabinhait (commented):
@karthikeyan5 Please send another PR for develop branch.

PR #18069, opened 2019-06-26
fix(item): fix patch for barcode childtable migration in item

karthikeyan5:
(cherry picked from commit b2c43ee2d91599038f7bdf2038bbc4c4e7b36285)

cherry pick for: https://github.com/frappe/erpnext/pull/18066 and https://github.com/frappe/erpnext/pull/18078


nabinhait (dismissed):
@karthikeyan5 Travis is failing because "Item Barcode" does not exists in earlier version. You should do `reload_doc` before using `get_all`. Also fix in hotfix branch

nabinhait (approved):

PR #18074, opened 2019-06-26
fix: returned qty issue while making sales invoice from dn if same item added multiple times

rohitwaghchaure:
**Issue**

1.  Added same item 3 times with qty as 2,4,10 in the delivery note and submitted
- Created sales invoice against above delivery note for the 3rd row with qty as 10
- Created delivery note return against above delivery note for the 2nd row with qty as 4
- Now when making the sales invoice for the 1st item, system throwing message that "all items are invoiced"

After debug issue identified that for the 1st row system considered that it is already retuned because in the code we get the retuned qty group by item code and in the above example same item code added multiple times.


nabinhait (approved):

nabinhait (commented):
@rohitwaghchaure Can you fix the same thing for Purchase Receipt?

PR #18086, opened 2019-06-26
fix: Order type restriction conflictive code

Don-Leopardo:
I was experimenting and added a new `order_type` via property setter, but when I saved my new order with my brand new `order_type`, it turns out that there is a hardcoded validation in the code. I was surprised, should it be there? Does a reason for it exists?
If there is no reason for it to exists then this PR can be merged to permit the users to create their own `order_type`.
If there is a valid reason, I think that we should discuss it, because is probable that there exists a better way to solve it. This is not only bad for the users that can't add a new order_type, it's bad because is code that should be maintained. For example, if tomorrow we decide to implement a new `order_type`, we should change this code too, and that's something more to remember and, because of that,  bad.
Please, share your ideas.

### DIFF
--- a/erpnext/controllers/selling_controller.py
+++ b/erpnext/controllers/selling_controller.py
@@ -134,13 +134,6 @@ class SellingController(StockController):
 		if sales_team and total != 100.0:
 			throw(_("Total allocated percentage for sales team should be 100"))
 
-	def validate_order_type(self):
-		valid_types = ["Sales", "Maintenance", "Shopping Cart"]
-		if not self.order_type:
-			self.order_type = "Sales"
-		elif self.order_type not in valid_types:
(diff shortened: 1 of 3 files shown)


rmehta:
Breaking changes can only go in `develop` branch.

Don-Leopardo:
@rmehta Is this a breaking change or is it that we can't decide if it is a useless code because we don't know why is it there?

rmehta:
This is a very specific validation so I am pretty sure it’s not useless. But I agree it’s not a good pattern, so should be fixed. But V11 users should not be affected by any such refactors.

Don-Leopardo:
To develop: https://github.com/frappe/erpnext/pull/18096

PR #18096, opened 2019-06-27
fix: Order type restriction conflictive code

Don-Leopardo:
I was experimenting and added a new order_type via property setter, but when I saved my new order with my brand new order_type, it turns out that there is a hardcoded validation in the code. I was surprised, should it be there? Does a reason for it exists?
If there is no reason for it to exists then this PR can be merged to permit the users to create their own order_type.
If there is a valid reason, I think that we should discuss it, because is probable that there exists a better way to solve it. This is not only bad for the users that can't add a new order_type, it's bad because is code that should be maintained. For example, if tomorrow we decide to implement a new order_type, we should change this code too, and that's something more to remember and, because of that, bad.
Please, share your ideas.


rmehta:
https://github.com/frappe/erpnext/blob/develop/erpnext/selling/doctype/sales_order/sales_order_list.js#L37

https://github.com/frappe/erpnext/blob/develop/erpnext/selling/doctype/sales_order/sales_order.js#L165

https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/status_updater.py#L41

Some instances where order_type is hard-coded. All of these need to be refactored too

Don-Leopardo:
@rmehta Those are fine. Those cases just use the predefined `order_types` for what they where planned, if you don't use any of the predefined `order_types` in your order, nothing happens, because those cases weren't planned for other `order_types`.

The thing is that with this PR the user can add custom `order_types` via property setter and we don't have to maintain this "hardcoded validation" when we decide to add a new `order_type`.

But yes, when we decide to add a new `order_type`, we are going to have to code its behavior. And when the user adds a new `order_type` it's not going to have a custom behavior.

rmehta:
Users can also delete order_types

Don-Leopardo:
@rmehta Yes, and if they delete `order_types` they lose the functionality of those `order_types`.

nabinhait:
@Don-Leopardo Let's allow addition of new order types, but not deletion of existing standard types.

nabinhait (changes requested):
@Don-Leopardo Let's allow addition of new order types, but not deletion of existing standard types.

Don-Leopardo:
@nabinhait This PR does not allow the user to delete existing order types. If that's a possibility, that possibility already exists before this PR. This PR only allows to create new order types, that wasn't possible before this PR.

nabinhait (approved):

Alchez:
Should this be backported to version 12?

nabinhait:
> Should this be backported to version 12?

@Alchez Can you send a PR to version-12-hotfix?

Alchez:
@nabinhait Sure, I'll send one soon

Alchez:
Made at https://github.com/frappe/erpnext/pull/21264.

PR #18104, opened 2019-06-28
fix: incorrect value booked in the accumulated depreciation account on sell of the asset

rohitwaghchaure:


clarkejj:
Please note this report that relates to this commit https://discuss.erpnext.com/t/scrapping-asset-error-indexerror-list-index-out-of-range/53204/3

Issue #7794, opened 2017-02-21
Employee Loan Application Error


umairsy:
Steps:

Human Resource > Employee Loan Management > Employee Loan Application

Error message comes as soon as click on Employee Loan Application's link.

`Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/app.py", line 56, in application
    response = frappe.handler.handle()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 19, in handle
    execute_cmd(cmd)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 42, in execute_cmd
    ret = frappe.call(method, **frappe.form_dict)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/__init__.py", line 903, in call
    return fn(*args, **newargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/load.py", line 70, in getdoctype
    docs = get_meta_bundle(doctype)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/load.py", line 81, in get_meta_bundle
    bundle = [frappe.desk.form.meta.get_meta(doctype)]
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/meta.py", line 20, in get_meta
    meta = frappe.cache().hget("form_meta", doctype, lambda: FormMeta(doctype))
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/redis_wrapper.py", line 171, in hget
    value = generator()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/meta.py", line 20, in 
    meta = frappe.cache().hget("form_meta", doctype, lambda: FormMeta(doctype))
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/meta.py", line 32, in __init__
    self.load_assets()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/meta.py", line 42, in load_assets
    self.load_templates()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/desk/form/meta.py", line 151, in load_templates
    module = load_doctype_module(self.name)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/modules/utils.py", line 179, in load_doctype_module
    raise ImportError, 'Module import failed for {0} ({1})'.format(doctype, module_name)
[...]

PR #18110, opened 2019-06-29
feat(Google Maps): Google Maps Refactor [Unified Google API Settings]

hrwX:
Merged Google Maps Integration in Delivery Trip
[Unified Google API Settings]
Dependent on PR [#7794](https://github.com/frappe/frappe/pull/7794)

- Route Calculation and Optimization:
  - Route calculation and optimization takes into consideration the address associated with the driver to carry out the calculation and optimization.
driver-op


erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -158,8 +158,11 @@ def form_route_list(self, optimize):
 			(list of list of str): List of address routes split at locks, if optimize is `True`
 		"""
 
-		settings = frappe.get_single("Google Maps Settings")
-		home_address = get_address_display(frappe.get_doc("Address", settings.home_address).as_dict())
+		settings = frappe.get_doc("Google Maps", {"entity": self.driver})
```
Alchez: If the `Google Maps` document doesn't exist for the driver, `get_doc` will raise a `DoesNotExistError` exception and fail.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -327,7 +330,7 @@ def get_directions(route, optimize):
 		(dict): Route legs and, if `optimize` is `True`, optimized waypoint order
 	"""
 
-	settings = frappe.get_single("Google Maps Settings")
+	settings = frappe.get_doc("Google Maps", {"entity": driver})
```
Alchez: Same as above.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -310,7 +313,7 @@ def sanitize_address(address):
 	return ', '.join(address[:3])
 
 
-def get_directions(route, optimize):
+def get_directions(route, optimize, driver):
```
Alchez: Update the docstring as well.

Alchez (changes requested):

Alchez:
@hrwX, you're also adding a new `macro` directory in the LMS module. Probably unintended.

hrwX:
> @hrwX, you're also adding a new `macro` directory in the LMS module. Probably unintended.

Thanks for pointing that out, didnt pay attention while pushing it.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -204,6 +200,21 @@ def rearrange_stops(self, optimized_order, start):
 
 		self.delivery_stops[start:start + len(stops_order)] = stops_order
 
+	def get_client(self):
```
Alchez: This is a very generic function, and should not be restricted to the Delivery Trip class. Otherwise, you'd need to define this for every doctype where it's needed.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -100,16 +100,13 @@ def process_route(self, optimize):
...
-			frappe.throw(_("Cannot process route, since Google Maps Settings is disabled."))
-
 		departure_datetime = get_datetime(self.departure_time)
 		route_list = self.form_route_list(optimize)
 
 		# For locks, maintain idx count while looping through route list
 		idx = 0
 		for route in route_list:
-			directions = get_directions(route, optimize)
+			directions = get_directions(self, route, optimize)
```
Alchez: If the document is required, convert the `get_directions` function into a method in the class.

Alchez (changes requested):

erpnext/stock/doctype/delivery_trip/delivery_trip.py
hrwX: Direction API Integration is just used in Delivery Trip, so making it a class function

netchampfaris (dismissed):

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -157,9 +154,10 @@ def form_route_list(self, optimize):
 		Returns:
 			(list of list of str): List of address routes split at locks, if optimize is `True`
 		"""
+		if not self.driver_address:
```
Alchez: Check on the front-end itself, and don't call `process_route` if address is missing.

Alchez (changes requested):

netchampfaris (approved):

PR #18111, opened 2019-06-29
fix: POS loyalty issues

### LINKED ISSUE #15479: Loyalty Redemption Not Showing Up in PoS page - Version 11
The Loyalty total and redemption option is not showing up on the PoS page anymore.

This makes it impossible to use loyalty for retail customers, and they are the ones that really need it.

Olamide

sunhoww:
This PR attempts to solve a few issues
- fixes #15479
- updating loyalty points was having no effect the payment amounts as intended because `set_value` calls made in `Payment.update_loyalty_points` were asynchronous.
- logical to reset `loyalty_point` to zero when `redeem_loyalty_points` is unchecked

I used `async` / `await` for the async functions. I don't know whether Promises would better for more browser compatibility, because this JS file is loaded during runtime instead of being packaged during the build process.


rohitwaghchaure (approved):

joshreeder:
Screenshot 2019-07-19 11 44 34
I am not using the loyalty program and now POS gets this error...

joshreeder:
click the image - it doesnt preview properly.

rohitwaghchaure:
@joshreeder [fixed](https://github.com/frappe/erpnext/pull/18462) loyalty program issue

PR #18122, opened 2019-07-01
fix(item): item_code should not be unique

sahil28297:


sagarvora:
@sahil28297 Item code should be unique, since the Item is named based on that:
https://github.com/frappe/erpnext/blob/046ba7b442ec9cf137ce733782a5b854fbc69842/erpnext/stock/doctype/item/item.json#L7

nabinhait:
> @sahil28297 Item code should be unique, since the Item is named based on that:
> https://github.com/frappe/erpnext/blob/046ba7b442ec9cf137ce733782a5b854fbc69842/erpnext/stock/doctype/item/item.json#L7
> 
> What is the argument for not keeping Item code unique?

Item naming is based on settings, it can be based on item_code / item_name / naming_series. Making item_code unique, we need to update item_code as name always unnecessarily.

sagarvora:
> Item naming is based on settings, it can be based on item_code / item_name / naming_series. Making item_code unique, we need to update item_code as name always unnecessarily.

Having a unique key prevents the user from creating duplicate items. Duplicate items can cause errors in identifying stock.

sagarvora:
@nabinhait Keeping item code unique allows us to avoid creating duplicate items and can prevent errors in maintaining stock. What if we set the document name as auto-generated, make item code not mandatory?

nabinhait:
> @nabinhait Keeping item code unique allows us to avoid creating duplicate items and can prevent errors in maintaining stock. What if we set the document name as auto-generated, make item code not mandatory?

Everything is based on `name` (primary key) column of the item. There is no point maintaining unique item_code for every item.

netchampfaris:
@nabinhait We should remove the `item_code` field altogether.

nabinhait:
> @nabinhait We should remove the `item_code` field altogether.

@netchampfaris We can't remove the item_code field. Because in some cases, specially while importing items from other software, it is needed to maintain old item codes. We can't enter item_code in item_name field because item_name is descriptive in nature for printing purpose.

sagarvora:
Item Code is conceptually unique as it refers to the Item's SKU. After much thought, I've come up with this solution:
- We can remove unique from `item_code` field and make it mandatory.
- We can create a stock setting labelled "Maintain Item Codes".
If checked, a validation will be done before saving an item to ensure that all item codes are unique.
If unchecked, we can set the label as "Item Name" instead of "Item Code" and hide the Item Name field (using JS). The `item_name` field will have to replicated from `item_code` in this case.

In all other DocTypes like SO Item, we can re-label "Item Code" ==> "Item".

Alternate Solution:
We can remove `unique` from `item_code` and do a validation before saving to check if `item_code` doesn't exist in item table. We can then make `item_name` mandatory (one of the fields must be) and write a patch to copy `item_code` to `item_name` where it doesn't exist.

This would however have repercussions in all DocTypes where items are being used.

nabinhait:
> Item Code is conceptually unique as it refers to the Item's SKU. After much thought, I've come up with this solution:
> 
> * We can remove unique from `item_code` field and make it mandatory.
> * We can create a stock setting labelled "Maintain Item Codes".
>   If checked, a validation will be done before saving an item to ensure that all item codes are unique.
>   If unchecked, we can set the label as "Item Name" instead of "Item Code" and hide the Item Name field (using JS). The `item_name` field will have to replicated from `item_code` in this case.
> 
> In all other DocTypes like SO Item, we can re-label "Item Code" ==> "Item".
> 
> Alternate Solution:
> We can remove `unique` from `item_code` and do a validation before saving to check if `item_code` doesn't exist in item table. We can then make `item_name` mandatory (one of the fields must be) and write a patch to copy `item_code` to `item_name` where it doesn't exist. This would however have repercussions in all DocTypes where items are being used.
> 
> In any case, I think this PR shouldn't go into v11.

@sagarvora I really can't understand why so much complexity. Item_code is just an input field to create **Item ID**. Item ID is unique and if ID is generated based on item_code, it automatically becomes unique too. But if ID is generated by series or item_name, there is no point maintaining item_code field as unique.
I can understand in all transactions, "Item Code" label is used, which might causing the confusion. May be renaming "Item Code" label to "Item" will remove the confusion.

sagarvora:
> But if ID is generated by series or item_name, there is no point maintaining item_code field as unique.

@nabinhait I do understand the need for simplification in industries like Service / Retail, but it shouldn't come at the cost of over-simplification for industries like Manufacturing / Distribution.

There are multiple reasons why companies choose to maintain item codes instead of using naming series. Here are some of the benefits of Item Codification ([as mentioned in ERPNext docs](https://erpnext.com/docs/user/manual/en/stock/item-codification)):

> 1. Standard way of naming things.
> 2. Less likely to have duplicates.
> 3. Explicit definition.
> 4. Helps to quickly find if a similar item exists.
> 5. Item names get longer and longer as more types get introduced. Codes are shorter.

One of the reasons for maintaining item codes is for easy accessing of items. Many of the manufacturing/distribution companies which use ERPNext maintain a database of lacs of items. A item codification system helps companies access items easily while creating Sales or Purchase documents.

As mentioned in the documentation:
>  When you have items that cross a certain size, life without codification is a nightmare.

The best solution I can think of is keeping this domain-specific and user-configurable. Maybe discussing this with a wider audience would help.

Nofal-Mami:
Hi,
See to make long history short, system should have an item code field mandatory and unique, entered manually or auto generated by system using naming series, another field for description, mandatory, text, entered manually and unique to avoid duplicate creation of item, the same fetched in item lov, also get printed in PO, SO, SI, PI and every where used. In the backend you can manage to have an item_id not visible to end user, a sequence generated for each new item, and this id is used as foreign key, item_id is optional approach and i suggest to use. Item name become useless. You can add item long description if required. For stocked item impossible to manage duplicate item code, all transaction are based on item code and description. For service item you can put a check box to opt for a modifiable description for the item at PO level to avoid creating lot of items.
Let me if you have any question. 
Rgds
Nofal

PR #18137, opened 2019-07-02
fix(sales-invoice): create sales invoice from quotation

Mangesh-Khairnar:
Fixes bug introduced from the change of field-name from customer to party_name


Mangesh-Khairnar:
@frappe-bot create instance

PR #18148, opened 2019-07-03
fix: reloads all newly created docs in v12

Mangesh-Khairnar:
- Fixes migration from v11 to v12
- On migrating from v11 to v12 certain doctypes are not reloaded, which is breaking the migration on adding custom fields.

### DIFF
--- a/erpnext/patches/v12_0/make_custom_fields_for_bank_remittance.py
+++ b/erpnext/patches/v12_0/make_custom_fields_for_bank_remittance.py
@@ -3,6 +3,8 @@ import frappe
 from erpnext.regional.india.setup import make_custom_fields
 
 def execute():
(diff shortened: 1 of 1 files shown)


Thunderbottom (approved):

erpnext/patches/v12_0/make_custom_fields_for_bank_remittance.py
```diff
@@ -3,6 +3,8 @@
 from erpnext.regional.india.setup import make_custom_fields
 
 def execute():
+    reload_doc("accounts", "doctype", "tax_category")
```
Thunderbottom: `frappe.reload_doc()` and not `reload_doc()`

same goes for the next line as well

Thunderbottom (changes requested):

nabinhait (approved):

PR #18174, opened 2019-07-05
chore: added check for auto-reorder in stock settings

scmmishra:
When setting Auto Re-Order levels for an item, the server would not validate if auto reorder is enabled or not.

This PR adds a check for that in `validate_auto_reorder_enabled_in_stock_settings` and shows a warning in case it is disabled. **Note:** The user is able to save the item anyway. 

---
Warning Message

Screenshot 2019-07-05 at 11 05 37 AM

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -866,8 +866,6 @@ class Item(WebsiteGenerator):
 			if not enabled:
 				frappe.msgprint(msg=_("You have to enable auto re-order in Stock Settings to maintain re-order levels."), title=_("Enable Auto Re-Order"), indicator="orange")
 
-		return
-
 def get_timeline_data(doctype, name):
 	'''returns timeline data based on stock ledger entry'''
 	out = {}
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item.py
```diff
@@ -859,6 +860,14 @@ def check_if_linked_document_exists(self, field):
 					filters={"production_item": self.name, "docstatus": 1}):
 				return True
 
+	def validate_auto_reorder_enabled_in_stock_settings(self):
+		if self.reorder_levels:
+			enabled = frappe.db.get_single_value('Stock Settings', 'auto_indent')
+			if not enabled:
+				frappe.msgprint(msg=_("You have to enable auto re-order in Stock Settings to maintain re-order levels."), title=_("Enable Auto Re-Order"), indicator="orange")
+
+		return
```
Alchez: You don't need an explicit return

Alchez (changes requested):

Anurag810 (approved):

PR #18186, opened 2019-07-05
fix: BOM Item with Operation

bhavishyasharma:
**Issue**
Same Item for different operations is merged.
Ex:
- Operation 1
    - Item 1 x 2
    - Item 2 x 1
    - Item 3 x 1
- Operation 2
    - Item 2 x 1
    - Item 3 x 1
    - Item 4 x 1

Result
- Operation 1
    - Item 1 x 2
    - Item 2 x 2
    - Item 3 x 2
- Operation 2
    - Item 4 x 1

Screenshot 2019-07-05 at 23 02 51
Screenshot 2019-07-05 at 23 03 08
Screenshot 2019-07-05 at 23 03 47

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -516,16 +516,14 @@ class BOM(WebsiteGenerator):
 		return erpnext.get_company_currency(self.company)
 
 	def add_to_cur_exploded_items(self, args):
-		if self.cur_exploded_items.get(args.item_code) and self.cur_exploded_items[args.item_code]["operation"] == args.operation:
-			if args.operation is None:
-				self.cur_exploded_items[args.item_code]["stock_qty"] += args.stock_qty
-			else:
-				self.cur_exploded_items[args.item_code+":"+args.operation]["stock_qty"] += args.stock_qty
+		key = (args.item_code)
+		if args.operation:
+			key = (args.item_code, args.operation)
+
+		if key in self.cur_exploded_items:
+			self.cur_exploded_items[key]["stock_qty"] += args.stock_qty
 		else:
-			if args.operation is None:
-				self.cur_exploded_items[args.item_code] = args
-			else:
-				self.cur_exploded_items[args.item_code+":"+args.operation] = args
+			self.cur_exploded_items[key] = args
 
 	def get_child_exploded_items(self, bom_no, stock_qty):
 		""" Add all items from Flat BOM of child BOM"""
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -516,10 +516,16 @@ def company_currency(self):
 		return erpnext.get_company_currency(self.company)
 
 	def add_to_cur_exploded_items(self, args):
-		if self.cur_exploded_items.get(args.item_code):
-			self.cur_exploded_items[args.item_code]["stock_qty"] += args.stock_qty
+		if self.cur_exploded_items.get(args.item_code) and self.cur_exploded_items[args.item_code]["operation"] == args.operation:
+			if args.operation is None:
```
rohitwaghchaure: at line number 519 you have checked self.cur_exploded_items[args.item_code]["operation"] == args.operation which means that  args.operation is not none then why you have added condition if args.operation is None at line number 520?

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -619,23 +625,31 @@ def get_bom_items_as_dict(bom, company, qty=1, fetch_exploded=1, fetch_scrap_ite
...
-			select_columns = ", bom_item.uom, bom_item.conversion_factor, bom_item.source_warehouse, bom_item.idx, bom_item.operation, bom_item.include_item_in_manufacturing")
+			select_columns = ", bom_item.uom, bom_item.conversion_factor, bom_item.source_warehouse, bom_item.idx, bom_item.operation, bom_item.include_item_in_manufacturing",
+			groupby_columns = """, bom_item.operation""")
 		items = frappe.db.sql(query, { "qty": qty, "bom": bom, "company": company }, as_dict=True)
 
 	for item in items:
-		if item.item_code in item_dict:
-			item_dict[item.item_code]["qty"] += flt(item.qty)
+		if item.item_code in item_dict and ( fetch_scrap_items or item_dict[item.item_code]["operation"] == item.operation):
+			if item.operation is None:
```
rohitwaghchaure: at line number 643 you have checked self.cur_exploded_items[args.item_code]["operation"] == args.operation which means that  args.operation is not none then why you have added condition if args.operation is None at line number 644?

erpnext/manufacturing/doctype/bom/bom.py
bhavishyasharma: In case the item does not have an operation attached to it,  both self.cur_exploded_items[args.item_code]["operation"] and args.operation will be none, so they will be equal. But in that case we dont want index to be ":item_code", just "item_code" hence the check for None.

erpnext/manufacturing/doctype/bom/bom.py
bhavishyasharma: In case the item does not have an operation attached to it,  both self.cur_exploded_items[args.item_code]["operation"] and args.operation will be none, so they will be equal. But in that case we dont want index to be ":item_code", just "item_code" hence the check for None.

erpnext/manufacturing/doctype/bom/bom.py
rohitwaghchaure: For that case you can check first args.operation is none or not at line 519, no need of extra condition in the next line 
like as below
```
self.cur_exploded_items.get(args.item_code) and args.operation and self.cur_exploded_items[args.item_code]["operation"] == args.operation:
```

bhavishyasharma: doing so overly complicated the logic
```
if self.cur_exploded_items.get(args.item_code) and args.operation and self.cur_exploded_items[args.item_code]["operation"] == args.operation:
	self.cur_exploded_items[args.item_code + ":" + args.operation]["stock_qty"] += args.stock_qty
elif self.cur_exploded_items.get(args.item_code) and not args.operation:
	self.cur_exploded_items[args.item_code]["stock_qty"] += args.stock_qty
elif args.operation:
	self.cur_exploded_items[args.item_code + ":" + args.operation] = args
else:
	self.cur_exploded_items[args.item_code] = args
```

bhavishyasharma: another solution is to use the :item_code for index when the operation is None. this will reduce the logic to simpler form.

rohitwaghchaure: Try below one

```
key = (args.item_code)
if args.operation:
    key = (args.item_code, args.operation)

if key in self.cur_exploded_items:
    self.cur_exploded_items[key]["stock_qty"] += args.stock_qty
else:
    self.cur_exploded_items[key] = args
```

rohitwaghchaure (approved):

rohitwaghchaure:
@bhavishyasharma please send the same PR on develop branch

PR #18187, opened 2019-07-05
fix: use Bank link field in Employee instead of data field (+ patch)

sagarvora:
Before Patch:
Screenshot from 2019-07-05 23-35-57

After Patch:
Screenshot from 2019-07-05 23-38-12


sagarvora:
Will reopen after fixing tests.

PR #18188, opened 2019-07-05
fix: use Bank link field in Employee instead of data field (+ patch)

sagarvora:
For `hotfix`: #18187


sagarvora:
Will reopen after fixing tests.

PR #18191, opened 2019-07-06
Hotfix

Ashraf-Elsharqawy:
Sales order (when changing uom in sales order rate still fixed with default uom 
example if uom unit and rate 0.5   and i want to sell with carton it should Change rate to rate per carton it dose not happen.


hrwX:
@Ashraf-Elsharqawy I suppose you wanted to raise an Issue and not a PR. If so create an issue in [here](https://github.com/frappe/erpnext/issues). I am closing this PR as you are merging hotfix into develop.

PR #18205, opened 2019-07-09
feat: Adjustment entry

harshil-shipmnts:
Added a new doctype for handling adjustment entries

### Problem
In any industry when a company is buying from as well as selling to the same party, then the company may have both outstanding sales and purchase invoices with that party.

In this case, ideally the company & involved party both will want to first adjust the sales and purchase invoices against each other and then pay/receive the remaining amount as per the netting.

### Solution
Introduce a new doctype called "Adjustment entry" to handle the netting logic. It allows the user to fetch all outstanding invoices with a party by selecting it in customer and vendor. The user can then auto allocate the amount against invoices or manually enter the amount to be allocated for each invoice. Adjustment entry also allows the user to settle multi-currency invoices by entering the exchange rates, it will then auto calculate the exchange gain/loss amount accordingly.

Once the user has entered the allocated amount for the invoices that they want to net against the other and the difference amount is zero, then they can submit the adjustment entry to make the general ledger entries and settle those invoices.

SoA-new

### Notes
- Please help test, write feedbacks and report issues.
- Do I need to add documentation anywhere?


harshil-shipmnts:
@nabinhait waiting for your review. Please let me know if I can help you with anything.

PR #18233, opened 2019-07-10
fix: set project link in item tables

Anurag810:
mapped project on the material transfer from
Screenshot 2019-07-12 at 11 38 06 AM
 stock Entry

### DIFF
--- a/erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
+++ b/erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
@@ -61,7 +61,8 @@
   "ste_detail",
   "column_break_51",
   "transferred_qty",
-  "reference_purchase_receipt"
+  "reference_purchase_receipt",
+  "project"
  ],
  "fields": [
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
```diff
@@ -472,11 +473,18 @@
    "label": "Reference Purchase Receipt",
    "options": "Purchase Receipt",
    "read_only": 1
+  },
+  {
+   "fieldname": "reference_project",
```
nabinhait: Why fieldname is `reference_project`? `project` is used everywhere.

nabinhait (dismissed):

erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
Anurag810: done

nabinhait (approved):

PR #18241, opened 2019-07-10
feat: added chart to cash-flow report

scmmishra:
Added missing charts for Cash Flow report

Screenshot 2019-07-10 at 8 29 07 PM

The bad labels have to be fixed in Charts library. Will be fixed in the next release of charts.


netchampfaris (changes requested):
Also return the fieldtype so that the currency is properly formatted.

https://github.com/frappe/erpnext/pull/18211/files

deepeshgarg007 (approved):

PR #18246, opened 2019-07-11
fix: Restore "allow delivery" condition checks

surajshetty3416:
- Users were not able to create delivery note because `allow_delivery` flag was false and it never used to get updated

**Before**
Screenshot 2019-07-11 at 9 46 00 AM

**After**
Screenshot 2019-07-11 at 9 43 29 AM


nabinhait (changes requested):
Instead of removing `allow_delivery`, please restore the original functionality of allow_delivery, which was removed by https://github.com/frappe/erpnext/pull/17975/

nabinhait (approved):

Issue #18215, opened 2019-07-10
 Chart of Accounts Importer - Can we have Error log with row nos.?


umairsy:
Enter Root type for which account, which Row No.?

![photo_2019-07-10 12 01 21](https://user-images.githubusercontent.com/910238/60945965-9ec19d00-a30a-11e9-9e50-496332eaceee.jpeg)

PR #18285, opened 2019-07-12
fix: Usability fixes for chart of Accounts Importer

deepeshgarg007:
- Provided Better error messages
- Row no for rows with no Account name


nabinhait (approved):

surajshetty3416:
https://github.com/frappe/erpnext/issues/18215

PR #18301, opened 2019-07-12
fix(bank-reconciliation-statement): payment entry links

Mangesh-Khairnar:
Payment entry links were referencing payment document in the bank reconciliation statement report.
Screenshot 2019-07-12 at 5 55 21 PM


erpnext/accounts/report/bank_reconciliation_statement/bank_reconciliation_statement.py
```diff
@@ -58,8 +58,8 @@ def get_columns():
 		{
 			"fieldname": "payment_entry",
 			"label": _("Payment Entry"),
-			"fieldtype": "Dynamic Link",
-			"options": "payment_document",
+			"fieldtype": "Link",
+			"options": "Payment Entry",
```
nabinhait: The column also shows Journal Entry, hence keep it Dynamic Link, just add another column 'Payment Document' to fix link issue.

nabinhait (changes requested):

erpnext/accounts/report/bank_reconciliation_statement/bank_reconciliation_statement.py
Mangesh-Khairnar: Added changes

nabinhait (approved):

nabinhait (approved):

PR #18323, opened 2019-07-15
fix: Fixes in get_balance_on() function

deepeshgarg007:


erpnext/accounts/utils.py
```diff
@@ -104,6 +104,9 @@ def get_balance_on(account=None, date=None, party_type=None, party=None, company
 		# get balance of all entries that exist
 		date = nowdate()
 
+	if account:
+		acc = frappe.get_doc("Account", account)
```
nabinhait: Use get_value instead of get_doc

nabinhait (changes requested):

erpnext/accounts/utils.py
deepeshgarg007: @nabinhait acc object is used to get report_type, company, is_root, account_currency and check permissions on that account. Keeping it get_doc for now

nabinhait (approved):

Issue #18351, opened 2019-07-16
[Projects] "Daily Project Progress" email content issue


jaichavan:
Steps to replicate:
1) Enable "Collect Progress" under "Monitor Progress" tab in a project

![image](https://user-images.githubusercontent.com/40264279/61256544-c42b2c80-a78a-11e9-8d99-79ac4f391c54.png)

2) At the scheduled time, the users who have access to the project will get an email like this:

![image](https://user-images.githubusercontent.com/40264279/61256579-e7ee7280-a78a-11e9-8bd5-8f2342a8cf19.png)

Say,  didn't reply to this email. Or maybe I did.

3) Then at midnight, another email is sent (like DWS) with the subject "Daily Project Progress for Project _project_name_" which has _this _: **user.full_name**

![image](https://user-images.githubusercontent.com/40264279/61256441-7b737380-a78a-11e9-9c99-bdfcc6b61a42.png)

netchampfaris:
This is fixed in https://github.com/frappe/erpnext/pull/18343 and merged

PR #18352, opened 2019-07-16
fix: added user.full_name to 'Fetch From'

scmmishra:
Solves https://github.com/frappe/erpnext/issues/18351

user.full_name was added to `Default` option instead of `Fetch From` causing the issue.


netchampfaris:
Already fixed it in https://github.com/frappe/erpnext/pull/18343

PR #18353, opened 2019-07-16
fix(call popup): Multiple changes

surajshetty3416:
- Remove summary from call log use comments instead
- Add contact and lead if matched to the call log
- Add received by the field with Employee Link
- DocPerm to allow Employees to have read access on call logs
- Remove unwanted code
- Add a method to get lead just by providing the phone number
- Show popup only to Employee to which the system is trying to connect.
- Remove custom code from call popup dialog and replace it with
 standard fields
Screenshot 2019-08-04 at 11 01 53 PM
Screenshot 2019-08-04 at 11 38 14 PM

- Increase timeout to hide popup after the call has been disconnected.


erpnext/communication/doctype/call_log/call_log.py
```diff
@@ -4,16 +4,59 @@
...
+from frappe import _
 from frappe.model.document import Document
-from erpnext.crm.doctype.utils import get_employee_emails_for_popup
+from erpnext.crm.doctype.utils import get_scheduled_employees_for_popup
+from frappe.contacts.doctype.contact.contact import get_contact_with_phone_number
+from erpnext.crm.doctype.lead.lead import get_lead_with_phone_number
 
 class CallLog(Document):
+	def before_insert(self):
+		number = self.get('from').lstrip('0')
```
netchampfaris, line 17: Explain this line

erpnext/crm/doctype/utils.py
```diff
@@ -3,82 +3,56 @@
...
+	if not contact and not lead: return
 
-	last_communication = {}
-	last_issue = {}
-	if reference_doc.doctype == 'Contact':
-		customer_name = ''
+	last_communication = None
+	last_issue = None
+	if contact:
 		query_condition = ''
```
netchampfaris, line 13: Build this query using `%s` or `%(name)s` for automatic escaping

erpnext/crm/doctype/utils.py
```diff
@@ -3,82 +3,56 @@
...
-
-	elif reference_doc.doctype == 'Lead':
+	if lead:
 		last_communication = frappe.get_all('Communication', filters={
-			'reference_doctype': reference_doc.doctype,
-			'reference_name': reference_doc.name,
+			'reference_doctype': 'Lead',
+			'reference_name': lead,
 			'sent_or_received': 'Received'
 		}, fields=['name', 'content'], limit=1)
```
netchampfaris: Add `order_by` explicitly

erpnext/crm/doctype/utils.py
```diff
@@ -3,82 +3,56 @@
...
-		call_log.summary = summary
-	else:
-		call_log.summary += '<br>' + summary
-	call_log.save(ignore_permissions=True)
-
-def get_employee_emails_for_popup(communication_medium):
+def get_last_issue_from_customer(customer_name):
+	issues = frappe.get_all('Issue', {
+		'customer': customer_name
+	}, ['name', 'subject', 'customer'], limit=1)
```
netchampfaris: Add `order_by`

netchampfaris (changes requested):

netchampfaris (dismissed):

netchampfaris (approved):

PR #18360, opened 2019-07-16
feat: added default dashboard fixtures and patch

scmmishra:
Create default account dashboards when installing ERPNext.

**Preview:**
Screenshot_2019-07-22 Accounts Dashboard


erpnext/setup/setup_wizard/data/dashboard_charts.py
```diff
@@ -0,0 +1,118 @@
...
+				"chart_name": "Bank Balance",
+				"timespan": "Last Year",
+				"color": "#ffb868",
+				"value_based_on": "",
+				"filters_json": json.dumps({"company": company.name, "account": bank_account}),
+				"source": "Account Balance Timeline",
+				"chart_type": "Custom",
+				"timeseries": 1,
+				"based_on": "",
+				"owner": "Administrator",
```
adityahase: Can we get rid of this?

adityahase (changes requested):

erpnext/setup/setup_wizard/operations/install_fixtures.py
```diff
@@ -475,7 +475,6 @@ def install_defaults(args=None):
 
 				frappe.db.set_value("Company", args.company_name, "default_bank_account", bank_account.name, update_modified=False)
 
-				return doc
```
adityahase: Were you high?

adityahase (approved):

erpnext/setup/setup_wizard/operations/install_fixtures.py
scmmishra: Someone else probably was.

erpnext/setup/setup_wizard/data/dashboard_charts.py
scmmishra: The empty fields? Removed them!

adityahase (approved):

nabinhait (approved):

PR #18393, opened 2019-07-18
fix: send sms not working for the quotation

rohitwaghchaure:
**Issue**

Customer field is not available in the quotation due to which below error was coming when user clicks on Send SMS button

```
  File "/home/frappe/benches/bench-11-2019-07-16/apps/frappe/frappe/app.py", line 61, in application
  File "/home/frappe/benches/bench-11-2019-07-16/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-11-2019-07-16/apps/frappe/frappe/handler.py", line 56, in execute_cmd
  File "/home/frappe/benches/bench-11-2019-07-16/apps/frappe/frappe/__init__.py", line 1036, in call
    return fn(*args, **newargs)
TypeError: get_contact_number() takes exactly 3 arguments (2 given)
```

### DIFF
--- a/erpnext/public/js/sms_manager.js
+++ b/erpnext/public/js/sms_manager.js
@@ -30,6 +30,8 @@ erpnext.SMSManager = function SMSManager(doc) {
 			this.show('', '', '', doc.mobile_no, default_msg[doc.doctype]);
 		else if (doc.doctype == 'Opportunity')
 			this.show('', '', '', doc.contact_no, default_msg[doc.doctype]);
+		else if (doc.doctype == 'Quotation')
+			this.show(doc.contact_person, doc.quotation_to, doc.party_name, '', default_msg[doc.doctype]);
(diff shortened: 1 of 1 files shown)


erpnext/public/js/sms_manager.js
```diff
@@ -30,6 +30,8 @@ erpnext.SMSManager = function SMSManager(doc) {
 			this.show('', '', '', doc.mobile_no, default_msg[doc.doctype]);
 		else if (doc.doctype == 'Opportunity')
 			this.show('', '', '', doc.contact_no, default_msg[doc.doctype]);
+		else if (doc.doctype == 'Quotation')
```
deepeshgarg007: @rohitwaghchaure quotation is already handled above on line no 25

deepeshgarg007 (changes requested):

PR #18412, opened 2019-07-19
fix: Change other_charges_calculation type from text to longtext

zeta17:
update other_charges_calculation from type text to longtext, to hold more items in Taxes and Charges Calculation, because the maximum text capacity is only 65535 characters.


nabinhait:
@zeta17 Thanks for the first contribution. Please send the Pull Request to the develop branch. And also change the fieldtype to "Long Text" for all related doctypes like PO, PR, PI, SO, DN, SI.

Closing this PR.

PR #18468, opened 2019-07-24
fix: query for payment entry

Anurag810:
before it was taking the amount, not on the basis of rounding total.
fixed the query


scmmishra:
Add a good PR name and description

Anurag810:
> Add a good PR name and description

will update once raised on all branches

nabinhait (approved):

PR #18470, opened 2019-07-24
fix: query for payment entry

Anurag810:
before it was taking the amount, not on the basis of rounding total.
fixed the query.

Suppose,
A sales order of total 1648.90 after rounding it will be 1649.00
but  on submitting it was saying that 
it can't be greater than 1648.90

port of
#18468
#18465


nabinhait (approved):

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -641,20 +641,18 @@ def get_orders_to_be_billed(posting_date, party_type, party,
 
 	# Add cost center condition
 	if voucher_type:
-		doc = frappe.get_doc({"doctype": voucher_type})
-		condition = ""
-		if doc and hasattr(doc, 'cost_center'):
-			condition = " and cost_center='%s'" % cost_center
```
nabinhait, line 647: @Anurag810 Any specific reason for removing this?

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.py
Anurag810: by mistake

Anurag810:
#18709

PR #18477, opened 2019-07-24
fix: accounting period

Mangesh-Khairnar:
Currently accounting period doesn't work. This PR adds appropriate checks to make sure that accounting entries are not made for those doctypes which have accounting period enabled(Open).


erpnext/accounts/general_ledger.py
```diff
@@ -23,6 +25,28 @@ def make_gl_entries(gl_map, cancel=False, adv_adj=False, merge_entries=True, upd
...
 
+def validate_accounting_period(gl_map):
+	accounting_periods = frappe.db.sql(""" SELECT
+			ap.name as name
+		FROM
+			`tabAccounting Period` ap, `tabClosed Document` cd
+		WHERE
+			ap.name = cd.parent
+			AND ap.company = %(company)s
+			AND ap.status = "Open"
```
nabinhait: This parent status is confusing, I think "Closed" checkbox in child is sufficient. Why not removing parent status field altogether?

nabinhait (changes requested):

erpnext/accounts/general_ledger.py
Mangesh-Khairnar: Added changes

nabinhait (approved):

PR #18488, opened 2019-07-25
perf(travis): Faster builds

adityahase:
Similar to https://github.com/frappe/frappe/pull/7964

https://frappe.erpnext.com/desk#Form/Timesheet/TS-00818


adityahase:
Thanks to @karthikeyan5

erpnext/hr/doctype/staffing_plan/test_staffing_plan.py
```diff
@@ -92,5 +92,5 @@ def make_company():
 	company.abbr = "_TC10"
 	company.parent_company = "_Test Company"
 	company.default_currency = "INR"
-	company.country = "India"
+	company.country = "Pakistan"
```
Mangesh-Khairnar, line 95: Why is this change needed?

netchampfaris: @Mangesh-Khairnar 
https://github.com/frappe/erpnext/pull/18488/commits/0f072d7d40ff820ebd237a02b96e81f629661854

deepeshgarg007 (approved):

PR #18497, opened 2019-07-26
fix(payroll-entry): show make bank entry button

Mangesh-Khairnar:
On manual submission of salary slip related to a payroll entry make bank entry button is not visible. This PR adds a check to find all the submitted entries and if all salary slips are submitted then it shows the button.


Anurag810 (dismissed):

nabinhait (approved):

nabinhait:
@Mangesh-Khairnar Can you please also fix this in v11 and v12? Also note that I have made a little changes.

PR #18613, opened 2019-08-07
fix(timesheet): Hours automatically used to set 0

surajshetty3416:
**Case:** If the company had `Standard Working Hour` set
and when the user creates a timesheet with **same date** for
"To" and "From" then the `Hours` field automatically used to
get reset to 0 after saving the form.

timesheet_issue


PawanMeh:
Hours are zero with or without the fix I think since the difference in time is zero for the same date. I am not sure what is the exact use case that is to be tackled here. If there is no difference in start time and end time, time will be zero.

Also, just check time difference on same date in timesheet to check if the value is getting calculated correctly after your fix.

surajshetty3416:
Pawan, time difference on the same date is getting calculated correctly after this fix.

Issue #18526, opened 2019-07-30
Installation fails due to conflicting pandas version


NeolithEra:
## Issue description

Hi, users are unable to run erpnext due to dependency conflict with _**pandas**_ package. 

As shown in the following full dependency graph of **_erpnext_**, **_erpnext_** requires _**pandas**_ (the latest version)，while **_frappe_** requires _**pandas==0.14.1**_.

According to pip’s “first found wins” installation strategy, _**pandas==0.25.0**_ is the actually installed version. However, _**pandas==0.25.0**_ does not satisfy _**pandas==0.14.1**_.

### Dependency tree-------
```
erpnext(version range:)
| +-frappe(version range:)
| | +-click(version range:==3.3)
| | +-django(version range:==1.7.1)
| | +-django-cors-headers(version range:==0.13)
| | +-django-docopt-command(version range:==0.2.0)
| | +-django-health-check(version range:==1.0)
| | +-djangorestframework(version range:==2.4.2)
| | +-docopt(version range:==0.6.2)
| | +-futures(version range:==2.1.6)
| | +-mozilla-logger(version range:==0.2)
| | +-mysql-python(version range:==1.2.5)
| | +-pandas(version range:==0.14.1)
| | +-python-dateutil(version range:==2.2)
| | | +-six(version range:)
| | +-python-memcached(version range:==1.53)
| | +-pytz(version range:==2014.7)
| | +-raven(version range:==5.1.1)
| | +-six(version range:==1.4.1)
| | +-uwsgi(version range:==2.0.8)
| +-unidecode(version range:)
| +-pygithub(version range:)
| +-googlemaps(version range:)
| +-python-stdnum(version range:)
| +-braintree(version range:)
| +-gocardless_pro(version range:)
| +-woocommerce(version range:)
| +-pandas(version range:)
| +-plaid-python(version range:)
| | +-requests(version range:>=2.7.0)
```

Thanks for your help.
Best,
Neolith

NeolithEra:
### Suggested solutions

1.	Fix your direct dependency to be **_pandas==0.14.1_**. I have checked this revision will not affect your downstream projects now.
2.	Ask your upstream project **_frappe_** to loose the version range to **_pandas>=0.14.1_**.

Which solution do you prefer, 1 or 2?

@rohitwaghchaure Please let me know your choice. I can submit a PR to solve this issue.

Anurag810:
[...]

Issue #18525, opened 2019-07-30
[v12] Bench update last minute


FiloSpaTeam:
```python
Executing frappe.patches.v12_0.fix_public_private_files in xxx.xxx.it (1bd3e0294da19198)
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 97, in <module>
    main()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 18, in main
    click.Group(commands=commands)(prog_name='bench')
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/__init__.py", line 25, in _func
    ret = f(frappe._dict(ctx.obj), *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/site.py", line 234, in migrate
    migrate(context.verbose, rebuild_website=rebuild_website, skip_failing=skip_failing)
[...]

PR #18614, opened 2019-08-07
Change the version constraint of pandas to be pandas==0.14.1, to fix dependency conflict #18525.

NeolithEra:
#18526 
Fix direct dependency to be pandas==0.14.1.


heyakyra:
This is too old and won't allow the app to be installed #22424

PR #18637, opened 2019-08-07
fix: division by zero in asset Depereciation

Anurag810:

### DIFF
--- a/erpnext/assets/doctype/asset/asset.py
+++ b/erpnext/assets/doctype/asset/asset.py
@@ -255,9 +255,9 @@ class Asset(AccountsController):
 		precision = self.precision("gross_purchase_amount")
 
 		if row.depreciation_method in ("Straight Line", "Manual"):
+			depreciation_left = (cint(row.total_number_of_depreciations) - cint(self.number_of_depreciations_booked))
 			depreciation_amount = (flt(row.value_after_depreciation) -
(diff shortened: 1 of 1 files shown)


erpnext/assets/doctype/asset/asset.py
```diff
@@ -255,9 +255,9 @@ def get_depreciation_amount(self, depreciable_value, total_number_of_depreciatio
 		precision = self.precision("gross_purchase_amount")
 
 		if row.depreciation_method in ("Straight Line", "Manual"):
+			depreciation_left = (cint(row.total_number_of_depreciations) - cint(self.number_of_depreciations_booked))
 			depreciation_amount = (flt(row.value_after_depreciation) -
-				flt(row.expected_value_after_useful_life)) / (cint(row.total_number_of_depreciations) -
-				cint(self.number_of_depreciations_booked))
+				flt(row.expected_value_after_useful_life)) / depreciation_left if depreciation_left else 1
 		else:
```
nabinhait, line 267: @Anurag810 If no depreciations left, it should not take default 1. It should give a validation message that as all the depreciations has been booked, "value after depreciation" should be equal to "expected value after useful life"

nabinhait (changes requested):

erpnext/assets/doctype/asset/asset.py
Anurag810: done

Issue #18308, opened 2019-07-13
Retention bonus and employee incentive not getting fetched while creating salary slip


michellealva:
Steps to replicate this issue:

1. Create an employee, create a salary structure, assign salary structure to the employee.
2. In HR > Payroll > Retention Bonus (Create a new retention bonus with a date in August).
3. In HR > Payroll > Employee Incentive (Create a new employee incentive with a date in August).
4. Run payroll entry for the month of August and generate Salary Slips.

The created employee incentive and retention bonus do not get fetched in the salary slip.

![Screenshot 2019-07-13 at 9 13 36 PM](https://user-images.githubusercontent.com/50285544/61173640-a8b9f900-a5b3-11e9-8925-efe0b703a844.png)

![Screenshot 2019-07-13 at 9 14 10 PM](https://user-images.githubusercontent.com/50285544/61173641-ac4d8000-a5b3-11e9-979f-da130e791839.png)
![Screenshot 2019-07-13 at 9 16 18 PM](https://user-images.githubusercontent.com/50285544/61173642-af487080-a5b3-11e9-9eaa-61dd23143f2b.png)

michellealva:
@ruchamahabal any progress?

ruchamahabal:
https://github.com/frappe/erpnext/pull/18647
@michellealva you can track the progress of this feature here.

PR #18647, opened 2019-08-07
feat: Employee Incentive and Retention Bonus added in Salary Slip (Additional Salary)

ruchamahabal:
fix for the issue/feature request:
https://github.com/frappe/erpnext/issues/18308

- On submission of Employee Incentive and Retention Bonus, Additional Salary records of component type 'Arrear' created, so that they are fetched while creating Salary Slips.
- Fix for a small typo in bank_dashboard.py


erpnext/hr/doctype/employee_incentive/employee_incentive.py
```diff
@@ -7,4 +7,42 @@
 from frappe.model.document import Document
 
 class EmployeeIncentive(Document):
-	pass
+	def on_submit(self):
+		company = frappe.db.get_value('Employee', self.employee, 'company')
+		additional_salary = frappe.db.exists('Additional Salary', {
+				'employee': self.employee, 
+				'salary_component': 'Arrear',
```
nabinhait: Arrear should not be hard-coded, add a field "Salary Component" inside Employee Incentive form.

erpnext/hr/doctype/retention_bonus/retention_bonus.py
```diff
@@ -10,7 +10,51 @@
...
+			frappe.throw(_('Cannot create Retention Bonus for left Employees'))
 		if getdate(self.bonus_payment_date) < getdate():
-			frappe.throw(_("Bonus Payment Date cannot be a past date"))
+			frappe.throw(_('Bonus Payment Date cannot be a past date'))
+
+	def on_submit(self):
+		company = frappe.db.get_value('Employee', self.employee, 'company')
+		additional_salary = frappe.db.exists('Additional Salary', {
+				'employee': self.employee, 
+				'salary_component': 'Arrear',
```
nabinhait: Arrear should not be hard-coded, add a field "Salary Component" inside Retention Bonus form.

erpnext/hr/doctype/retention_bonus/retention_bonus.py
```diff
@@ -10,7 +10,51 @@
...
+		else:
+			bonus_added = frappe.db.get_value('Additional Salary', additional_salary, 'amount') + self.bonus_amount
+			frappe.db.set_value('Additional Salary', additional_salary, 'amount', bonus_added)
+			self.db_set('additional_salary', additional_salary)
+
+	def on_cancel(self):
+		if self.additional_salary:
+			linked_document_exists = check_if_linked_document_exists(self)
+
+			if linked_document_exists:
```
nabinhait: Checking hard-coded linked documents is not a good idea. Tomorrow we can add another document from which additional salary can be created.
Instead, we can always deduct the amount and then if amount become zero, we can cancel the additional salary record.

nabinhait (changes requested):

nabinhait (dismissed):

PR #18655, opened 2019-08-08
fix: incorrect value of ImponibileImporto in the xml invoice

rohitwaghchaure:
If same taxes set multiple times in the invoice them ImponibileImporto showing incorrect vale
image


nabinhait (approved):

nabinhait:
@rohitwaghchaure PR for other v-12-hotfix and develop?

PR #18660, opened 2019-08-08
fix: Duplicate items check in Sales Invoice

deepeshgarg007:
In Selling Settings there is a check box to Allow same items multiple times in a transaction.
This setting was honoured in Sales Order and Delivery Note but not in Sales Invoice 
 
Screenshot 2019-08-08 at 4 58 50 PM


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -550,6 +552,26 @@ def validate_delivery_note(self):
 			if d.delivery_note:
 				msgprint(_("Stock cannot be updated against Delivery Note {0}").format(d.delivery_note), raise_exception=1)
 
+	def validate_for_items(self):
```
nabinhait: It should be a common function for all doctypes (SO, DN and SI).

nabinhait (changes requested):

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -71,25 +70,6 @@ def validate_po(self):
...
-		check_list = []
-		for d in self.get('items'):
-			check_list.append(cstr(d.item_code))
-
-			# used for production plan
-			d.transaction_date = self.transaction_date
-
-			tot_avail_qty = frappe.db.sql("select projected_qty from `tabBin` \
-				where item_code = %s and warehouse = %s", (d.item_code, d.warehouse))
-			d.projected_qty = tot_avail_qty and flt(tot_avail_qty[0][0]) or 0
```
nabinhait: @deepeshgarg007 Set `transaction_date` and `projected_qty` via a separate function, as you have removed that.

erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -165,29 +164,6 @@ def validate_proj_cust(self):
...
-				else:
-					check_list.append(e)
-			else:
-				if f in chk_dupl_itm:
-					frappe.msgprint(_("Note: Item {0} entered multiple times").format(d.item_code))
-				else:
-					chk_dupl_itm.append(f)
-			#Customer Provided parts will have zero valuation rate
-			if frappe.db.get_value('Item', d.item_code, 'is_customer_provided_item'):
-				d.allow_zero_valuation_rate = 1
```
nabinhait: @deepeshgarg007 Set `allow_zero_valuation_rate` via a separate function, as it has been removed

nabinhait (dismissed):

nabinhait (approved):

PR #18662, opened 2019-08-08
fix: Duplicate items check in sales Invoice

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/18660


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -553,6 +555,26 @@ def validate_delivery_note(self):
 			if d.delivery_note:
 				msgprint(_("Stock cannot be updated against Delivery Note {0}").format(d.delivery_note), raise_exception=1)
 
+	def validate_for_items(self):
```
nabinhait: It should be a common function for all doctypes (SO, DN and SI).

nabinhait (changes requested):

nabinhait (approved):

PR #18664, opened 2019-08-08
fix: Payment validation for pos invoices v11

Anurag810:

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -78,6 +78,7 @@ class SalesInvoice(SellingController):
 			self.so_dn_required()
 
 		self.validate_proj_cust()
+		self.validate_pos_return()
 		self.validate_with_previous_doc()
 		self.validate_uom_is_integer("stock_uom", "stock_qty")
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -199,6 +200,16 @@ def on_submit(self):
 		if "Healthcare" in active_domains:
 			manage_invoice_submit_cancel(self, "on_submit")
 
+	def validate_pos_return(self):
+
+		if self.is_pos and self.is_return:
+			total_amount_in_payments = 0
+			for payment in self.payments:
+				total_amount_in_payments = payment.amount
```
nabinhait, line 208: `+=` ?

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -199,6 +200,16 @@ def on_submit(self):
...
+
+			if -total_amount_in_payments > -self.rounded_total:
```
nabinhait, line 210: Better way of writing this condition would be `total_amount_in_payments < self.rounded_total`

nabinhait (changes requested):

PR #18686, opened 2019-08-09
fix: multiple stock entry issues for the work order

rohitwaghchaure:
- Set "Backflush Raw Materials Based On" as "Material Transferred for Manufacture" in Manufacturing Settings
     - Not able to transfer extra quantity, "All items have already been transferred for this Work Order".
     - Not able to transfer extra item, "Did not found transfered item in Work Order , the item not added in Stock Entry".

- Set "Backflush Raw Materials Based On" as "BOM"
     -  User can transferred extra raw materials quantity.


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -993,7 +995,7 @@ def get_transfered_raw_materials(self):
 				filters={'parent': self.work_order, 'item_code': item_code},
 				fields=["required_qty", "consumed_qty"]
 				)
-			if not req_items:
+			if backflush_based_on=="BOM" and not req_items:
```
Alchez: This will not be "BOM", because this function is only called when the backflush setting is set to "Material Transferred for Manufacture".

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -493,6 +499,11 @@ def update_transaferred_qty_for_required_items(self):
 						'item': d.item_code
 					})[0][0]
 
+			required_qty = d.required_qty + (allowance_percentage/100 * self.qty)
```
nabinhait: Allowance should not be checked for raw materials

erpnext/stock/doctype/stock_entry/stock_entry.py
nabinhait: Remove entire validation

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
rohitwaghchaure: Fixed

erpnext/manufacturing/doctype/work_order/work_order.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #18694, opened 2019-08-11
fix(payroll-entry): Check for submitted salary slips onload for manual submission

Mangesh-Khairnar:
Fixes: Since "modified by" details changes on dbset, the user is not able to proceed with further transactions.
Screenshot 2019-08-11 at 2 37 37 PM


erpnext/hr/doctype/payroll_entry/payroll_entry.py
```diff
@@ -19,8 +19,7 @@ def onload(self):
 		# check if salary slips were manually submitted
 		entries = frappe.db.count("Salary Slip", {'payroll_entry': self.name, 'docstatus': 1}, ['name'])
 		if cint(entries) == len(self.employees) and not self.salary_slips_submitted:
-			self.db_set("salary_slips_submitted", 1)
-			self.reload()
+			self.db_set("salary_slips_submitted", 1, update_modified=False, commit=True)
```
nabinhait: Don't commit implicitly inside the code. You can use `flags` instead of setting the value in a field.

nabinhait (changes requested):

nabinhait (approved):

PR #18695, opened 2019-08-11
fix(payroll-entry): Check for submitted salary slips onload for manual submission

Mangesh-Khairnar:
Fixes: Since "modified by" details changes on dbset, the user is not able to proceed with further transactions.
Screenshot 2019-08-11 at 2 37 37 PM


nabinhait (changes requested):
Don't commit implicitly inside the code. You can use flags instead of setting the value in a field.

nabinhait (approved):

Issue #18377, opened 2019-07-17
From Date - To date not working in leave application 


reema30:
From date - To date in the default filters is not working. It works well when to click add filter. What is the use of these filters then? GIF attached.

![Leave application](https://user-images.githubusercontent.com/45488333/61378711-9eee0980-a8c3-11e9-907f-643d783bf0ec.gif)

Regards,
Reema

reema30:
@scmmishra assigning it to you, since you will be doing solving the same one assigned in Frappe

scmmishra:
@reema30 a fix for this is a work in progress. Will be solved soon. You can follow the PR here: https://github.com/frappe/frappe/pull/8121

scmmishra:
It is possible to add a custom condition for standard filter, but it will be to implicit and might confuse users. The best course of action would be to remove those filters

scmmishra:
Solved https://github.com/frappe/erpnext/pull/18696

PR #18696, opened 2019-08-11
refactor: remove from date and to date from standard filter

scmmishra:
A fix for the issue https://github.com/frappe/erpnext/issues/18377

This is not a bug but can be considered a UX issue as discussed in this thread: https://github.com/frappe/frappe/pull/8121

The natural expectation of the standard filters is to work as `=` or `like` query. Allowing custom conditions makes it too implicit. User has no explicit way of knowing what the actual condition is. It is best to remove those from standard filters.


surajshetty3416 (approved):

nabinhait (approved):

nabinhait (changes requested):
@scmmishra PR for v11?

PR #18713, opened 2019-08-13
feat: create address and contact after lead creation (v11)

Alchez:
**Problem:**

Currently, if I add a lead in the system (through desk, data import, etc.), creating any addresses and contacts for that lead requires additional steps. It's unnecessary and leads to bad UX.

**Changes:**

- (Feature + Format) Add fields in Lead to store address information.
  - If any of the address fields are filled in, the system will create an Address document and backlink to the Lead.
  - If no address fields are selected, the system simply skips Address creation.
- (Feature) Create a Contact document based on entered info in the Lead, and backlink to the Lead.
- (Bugfix) Only change organization lead's name if a person's name isn't already set.
- (Format) Change lead's end date to auto-set 1 day after next contact date.
- (Format) Update function calls on the frontend.

**Screenshots / GIFs:**

lead-address-contact


Alchez:
Avoiding feature additions to v11

PR #18715, opened 2019-08-13
feat: create address and contact after lead creation (develop)

Alchez:
**Problem:**

Currently, if I add a lead in the system (through desk, data import, etc.), creating any addresses and contacts for that lead requires additional steps. It's unnecessary and leads to bad UX.

**Changes:**

- (Feature + Format) Add fields in Lead to store address information.
  - If any of the address fields are filled in, the system will create an Address document and backlink to the Lead.
  - If no address fields are selected, the system simply skips Address creation.
- (Feature) Create a Contact document based on entered info in the Lead, and backlink to the Lead.
- (Bugfix) Only change organization lead's name if a person's name isn't already set.
- (Format) Change lead's end date to auto-set 1 day after next contact date.
- (Format) Update function calls on the frontend.

**Screenshots / GIFs:**

lead-address-contact


erpnext/crm/doctype/lead/lead.py
```diff
@@ -115,10 +121,83 @@ def set_lead_name(self):
...
+		names = self.lead_name.split(" ")
+		if len(names) > 1:
+			first_name, last_name = names[0], " ".join(names[1:])
+		else:
+			first_name, last_name = self.lead_name, None
+
+		contact_fields = ["email_id", "salutation", "gender", "designation", "phone", "mobile_no"]
+
+		contact = frappe.new_doc("Contact")
+		contact.update({contact_field: self.get(contact_field) for contact_field in contact_fields})
```
hrwX: - Add email, phone and mobile no. to the respective child table

hrwX (changes requested):

erpnext/crm/doctype/lead/lead.py
Alchez: Fixed

PR #18721, opened 2019-08-13
feat: Added 'barcode' field in BOM and 'BOM barcode' field in Work Order

marination:
- **BOM** now includes a barcode field to register barcodes pertaining to that BOM

image

- **Work Order** has an additional **BOM Barcode** field wherein the barcode number can be inserted and the concerned BOM No. and Item code will be fetched

bom

- Inappropriate barcodes will also be taken care of
errorbom


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -436,6 +436,18 @@ def get_items_and_operations_from_bom(self):
 		self.set_work_order_operations()
 
 		return check_if_scrap_warehouse_mandatory(self.bom_no)
+	
+	def get_item_bom_from_barcode(self,barcode):
+		bom = frappe.db.get_value("BOM Barcode",barcode,"parent")
+		if bom == None :
```
deepeshgarg007, line 442:
```suggestion
		if not bom :
```

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -347,6 +347,21 @@ frappe.ui.form.on("Work Order", {
...
 	},
 
+	bom_barcode: function(frm) {
+		frm.call({
+			doc: frm.doc,
+			method: "get_item_bom_from_barcode",
+			args: {barcode: frm.doc.bom_barcode},	
+		}).then((r) => {
+			if (r.message){
+				frm.set_value("production_item","");
```
deepeshgarg007, line 357:
```suggestion
				frm.set_value("production_item", "");
```
add single space after comma everywhere

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -347,6 +347,21 @@ frappe.ui.form.on("Work Order", {
 		});
 	},
 
+	bom_barcode: function(frm) {
+		frm.call({
+			doc: frm.doc,
+			method: "get_item_bom_from_barcode",
+			args: {barcode: frm.doc.bom_barcode},	
+		}).then((r) => {
+			if (r.message){
```
deepeshgarg007, line 356:
```suggestion
			if (r.message) {
```

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -436,6 +436,18 @@ def get_items_and_operations_from_bom(self):
 		self.set_work_order_operations()
 
 		return check_if_scrap_warehouse_mandatory(self.bom_no)
+	
+	def get_item_bom_from_barcode(self,barcode):
```
deepeshgarg007, line 440: Make this function a whitelisted function
There is no need to pass the work order doc since it it not used anywhere in this function

deepeshgarg007 (changes requested):

nabinhait:
@marination Does not look like a generic requirement, make it part of custom app

PR #18727, opened 2019-08-14
fix: removed filters(not required)

Anurag810:
removed unwanted filters during sales return during validation.
#18728
#18729


deepeshgarg007:
@Anurag810 No need to raise PR for version-12-hotfix

Issue #17552, opened 2019-05-10
Update Suggestion: Mode Of payment Enable / Disable Option  


sammish:
Is it Possible Make mode of payment Enable / Disable ?
![image](https://user-images.githubusercontent.com/12432802/57511690-833e1400-7312-11e9-9ab9-041749e9e89e.png)

sammish:
Please Add this feature 

![Screenshot from 2019-05-19 11-41-56](https://user-images.githubusercontent.com/12432802/57978469-3eebfb80-7a17-11e9-8614-487809355c26.png)
![Screenshot from 2019-05-19 11-42-09](https://user-images.githubusercontent.com/12432802/57978470-3f849200-7a17-11e9-99fc-1065f80ed69d.png)
![Screenshot from 2019-05-19 11-42-15](https://user-images.githubusercontent.com/12432802/57978471-3f849200-7a17-11e9-905d-47d484590f82.png)

sammish:
@umairsy please try to add this feature

PR #18738, opened 2019-08-14
refactor: mode of payment

sammish:
Mode of Payment Enable Option 

Mode of Payment Enable Option  #17552    Effective Month / Month wise analyze #18750

image


santoshbb (dismissed):

mhbu50:
@sammish Checkbox is not getting validate in anywhere in the system!
You should write more validation script in all doctypes that will call mode of payment as Link like POS Profile, Payment Entry

Please look at this link for your reference:
https://github.com/frappe/erpnext/blob/c07e8e5b0d0f9d197864d20815e04560ebc22288/erpnext/healthcare/doctype/clinical_procedure/clinical_procedure.js#L32

sammish:
> @sammish Checkbox is not getting validate in anywhere in the system!
> You should write more validation script in all doctypes that will call mode of payment as Link like POS Profile, Payment Entry
> 
> Please look at this link for your reference:
> https://github.com/frappe/erpnext/blob/c07e8e5b0d0f9d197864d20815e04560ebc22288/erpnext/healthcare/doctype/clinical_procedure/clinical_procedure.js#L32

nabinhait:
@sammish Adding "Months" doctype as a generic requirement does not look like a good idea. It is just a work around and user should create a custom doctype for using it in accounting dimension if needed. Please create a separate PR adding "Enabled" option in mode of payment. Closing this PR.

PR #18740, opened 2019-08-14
feat: Immutable ledger

deepeshgarg007:
Requires: https://github.com/frappe/frappe/pull/8255
Requires: https://github.com/frappe/frappe/pull/8290

This PR introduces the following changes in accounting and ledger generation:
- On cancellation of any document instead of deleting the GL Entry or SLE a reverse entry will be posted against that document
- Backdated entries will only be allowed after the last stock transaction's timestamp
- Since backdated entries are disabled there is no need of reposting future entries so removed logic for reposting future entries
- User will be able to filter cancelled doc entries which might be needed for audit purposes.


deepeshgarg007:
Do not close

erpnext/patches/v12_0/update_is_cancelled_field.py
```diff
@@ -0,0 +1,17 @@
+from __future__ import unicode_literals
+import frappe
+from erpnext.regional.india.setup import make_custom_fields
```
Don-Leopardo: Unused import? And maybe you should print the exception in te execute. None should be pass? isn't it more clear?

nabinhait (changes requested):
Travis failing

erpnext/controllers/accounts_controller.py
```diff
@@ -681,6 +681,10 @@ def set_total_advance_paid(self):
 
 		if advance:
 			advance = advance[0]
+
+			if self.doctype == "Sales Order":
+				advance['amount'] = - advance['amount']
```
nabinhait: What does it mean?

erpnext/stock/stock_balance.py
```diff
@@ -241,6 +239,41 @@ def reset_serial_no_status_and_warehouse(serial_nos=None):
 			except:
 				pass
 
+def repost_all_stock_vouchers():
```
nabinhait: remove this function

erpnext/patches/v12_0/update_is_cancelled_field.py
```diff
@@ -0,0 +1,17 @@
...
+		frappe.db.sql("UPDATE `tabStock Ledger Entry` SET is_cancelled = 0 where is_cancelled in ('', NULL, 'No')")
+		frappe.db.sql("UPDATE `tabSerial No` SET is_cancelled = 0 where is_cancelled in ('', NULL, 'No')")
+
+		frappe.db.sql("UPDATE `tabStock Ledger Entry` SET is_cancelled = 1 where is_cancelled = 'Yes'")
+		frappe.db.sql("UPDATE `tabSerial No` SET is_cancelled = 1 where is_cancelled = 'Yes'")
+
+		frappe.reload_doc("stock", "doctype", "stock_ledger_entry")
+		frappe.reload_doc("stock", "doctype", "serial_no")
+	except:
+		None
```
nabinhait: pass

nabinhait (changes requested):

erpnext/controllers/stock_controller.py
```diff
@@ -34,10 +34,6 @@ def make_gl_entries(self, gl_entries=None, repost_future_gle=True, from_repost=F
 					gl_entries = self.get_gl_entries(warehouse_account)
 				make_gl_entries(gl_entries, from_repost=from_repost)
 
-			if (repost_future_gle or self.flags.repost_future_gle):
-				items, warehouses = self.get_items_and_warehouses()
-				update_gl_entries_after(self.posting_date, self.posting_time, warehouses, items,
```
nabinhait, line 39: remove update_gl_entries_after function

nabinhait (changes requested):
Simplify `build` method in stock_ledger.py

nabinhait (approved):

PR #18744, opened 2019-08-14
fix: valuation rate in stock ledger

Mangesh-Khairnar:
Last valuation rate for an item was getting fetched from the stock ledger of the same transaction. This PR fixes that by skipping the stock ledger entry for that particular transaction.


SaiFi0102:
Yes this is indeed a bug I have also encountered https://github.com/frappe/erpnext/pull/16997

nabinhait (approved):

PR #18752, opened 2019-08-15
fix: Party dashboard heatmap not capturing  sales, purchase and other activities

deepeshgarg007:
Before:
image

After:
image

### DIFF
--- a/erpnext/accounts/party.py
+++ b/erpnext/accounts/party.py
@@ -450,7 +450,9 @@ def get_timeline_data(doctype, name):
 	# fetch and append data from Activity Log
 	data += frappe.db.sql("""select {fields}
 		from `tabActivity Log`
-		where reference_doctype="{doctype}" and reference_name="{name}"
+		where (reference_doctype="{doctype}" and reference_name="{name}")
+		or (timeline_doctype in ("{doctype}") and timeline_name="{name}")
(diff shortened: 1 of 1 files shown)


erpnext/accounts/party.py
```diff
@@ -450,7 +450,9 @@ def get_timeline_data(doctype, name):
 	# fetch and append data from Activity Log
 	data += frappe.db.sql("""select {fields}
 		from `tabActivity Log`
-		where reference_doctype="{doctype}" and reference_name="{name}"
+		where (reference_doctype="{doctype}" and reference_name="{name}")
+		or (timeline_doctype in ("{doctype}") and timeline_name="{name}")
+		or (reference_doctype in ("Quotation", "Opportunity") and timeline_name="{name}")
```
nabinhait, line 455: @deepeshgarg007 Why only Quotation and Opportunity? May be we can fetch all activities where `timeline_name` matches.

deepeshgarg007: @nabinhait All the activities having name as `timeline_name = {name}` and `timeline_doctype` as "Customer" will be fetched. But since "Quotation" and "Opportunity" has dynamic links for customer the `timeline_doctype` is captured as "DocType" instead of "Customer", so fetched records for these two separately.

nabinhait (approved):

PR #18778, opened 2019-08-19
fix: group by voucher consolidated showing incorrect data for deferred entries

rohitwaghchaure:


erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -305,12 +298,22 @@ def update_value_in_dict(data, key, gle):
 			update_value_in_dict(totals, 'total', gle)
 			if filters.get("group_by") != _('Group by Voucher (Consolidated)'):
 				gle_map[gle.get(group_by)].entries.append(gle)
+			elif filters.get("group_by") == _('Group by Voucher (Consolidated)'):
+				key = (gle.get("posting_date"), gle.get("voucher_type"),
```
nabinhait: why posting_date should be part of the key?

nabinhait (changes requested):

erpnext/accounts/report/general_ledger/general_ledger.py
rohitwaghchaure: Fixed

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -305,12 +298,22 @@ def update_value_in_dict(data, key, gle):
...
 			if filters.get("group_by") != _('Group by Voucher (Consolidated)'):
 				gle_map[gle.get(group_by)].entries.append(gle)
+			elif filters.get("group_by") == _('Group by Voucher (Consolidated)'):
+				key = (gle.get("voucher_type"), gle.get("voucher_no"),
+					gle.get("account"), gle.get("cost_center"))
+				if key not in consolidated_gle:
+					consolidated_gle.setdefault(key, gle)
+				else:
+					update_value_in_dict(consolidated_gle, key, gle)
 			else:
```
nabinhait: no scope of `else` statement

nabinhait (changes requested):

erpnext/public/node_modules
```diff
@@ -0,0 +1 @@
+/Users/rohitwaghchaure/Documents/erpnext_new/frappe-bench/apps/erpnext/node_modules
```
nabinhait: remove this?

nabinhait (changes requested):

erpnext/public/node_modules
rohitwaghchaure: Done

nabinhait (approved):

PR #18794, opened 2019-08-20
fix: Enhancement in Purchase Order

deepeshgarg007:
- Added a over transfer allowance field in Buying Settings

Screenshot 2019-08-20 at 3 01 58 PM

- Supplied Qty updation from Stock Entry

image


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1157,7 +1164,11 @@ def update_purchase_order_supplied_items(self):
 			where po.name = poitemsup.parent
 			and po.name = %s""", self.purchase_order))
 
-		#Update reserved sub contracted quantity in bin based on Supplied Item Details
+		#Update Supplied Qty in PO Supplied Items
+		frappe.db.sql("""UPDATE `tabPurchase Order Item Supplied` pos, `tabStock Entry Detail` sed
+			SET pos.supplied_qty = sed.transfer_qty where pos.name = sed.po_detail""")
```
nabinhait: What if multiple transfer entry made against the same PO?

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -521,14 +521,20 @@ def validate_purchase_order(self):
...
 				item_code = se_item.original_item or se_item.item_code
 				precision = cint(frappe.db.get_default("float_precision")) or 3
-				total_allowed = sum([flt(d.required_qty) for d in purchase_order.supplied_items \
+				required_qty = sum([flt(d.required_qty) for d in purchase_order.supplied_items \
 					if d.rm_item_code == item_code])
+
+				total_allowed = required_qty + (required_qty * (qty_allowance/100))
+
 				if not total_allowed:
 					frappe.throw(_("Item {0} not found in 'Raw Materials Supplied' table in Purchase Order {1}")
```
nabinhait, line 543: This validation is for the case if item is not present in PO. How it is related to allowance or total allowed qty?

nabinhait (changes requested):

PR #18809, opened 2019-08-21
fix: pos return validation

sunhoww:
WhatsApp Image 2019-08-21 at 3 16 22 PM

The validation exception occurs for setups which have rounding disabled. This is because `rounded_total` is set to zero during such cases.

This change hopes to considers `grand_total` as well during the validation.


sunhoww:
So meant this for **version-11-hotfix**

PR #18814, opened 2019-08-22
fix: allow multiple invoices against a single serial number

Alchez:
**Problem:**

Currently, the system validates invoices that have repeated a serial number. Although there are certain scenarios where it might be valid to re-use a serial number, for example, if a services company is billing their customers for repair or warranty services against a particular item. In that case, it helps to have the serial number entered and have it show up on prints and other documents.


nabinhait (commented):
> there are certain scenarios where it might be valid to re-use a serial number, for example, if a services company is billing their customers for repair or warranty services against a particular item. In that case, it helps to have the serial number entered and have it show up on prints and other documents.

In case of repair / warranty services, the billed item can be a separate service item. And it should not use serial no field. The user can might mention serial no in item description for reference. Can you please validate the scenario why the same item/serial no should be billed twice?

Alchez:
@nabinhait 

In our client's case, we do use a separate service item for billing, but using the Serial No also helps in generating reports.

You are correct that a single item should not be billed twice, but that's not what's really happening for services. It could be a different service item being billed against a reference Serial No.

In such cases, the validation should only occur if the item code AND the serial no match with any previous invoices.

Your thoughts?

nabinhait (commented):
> In such cases, the validation should only occur if the item code AND the serial no match with any previous invoices.

Yes, I agree. If item code is different then we can allow same serial no. Created a PR for that.
https://github.com/frappe/erpnext/pull/19580

PR #18815, opened 2019-08-22
fix: allow multiple invoices against a single serial number

Alchez:
**Problem:**

Currently, the system validates invoices that have repeated a serial number. Although there are certain scenarios where it might be valid to re-use a serial number, for example, if a services company is billing their customers for repair or warranty services against a particular item. In that case, it helps to have the serial number entered and have it show up on prints and other documents.


Alchez:
@nabinhait, Travis seems to be failing on a `on_cancel` document trigger, but my code changes are only in `validate`. Can you help me out with fixing the test please?

nabinhait (commented):
> In such cases, the validation should only occur if the item code AND the serial no match with any previous invoices.

Yes, I agree. If item code is different then we can allow same serial no. Created a PR for that.
https://github.com/frappe/erpnext/pull/19581

PR #18844, opened 2019-08-27
feat: Adding Supplier Filter to the Stock Reconciliation

Mitchy25:
Apologies I dont know how to insert Test Cases!

GIF:

Supplier Stock Reconciliation

Description:

Allows the user to do Stock Reconciliation by Supplier. This is very useful for users who keep their stock in their warehouse by supplier (Which I would suggest would be a lot of users). It allows the user to only see stock by supplier, rather than all stock in their warehouse.

The code may not be perfect as it relys upon the Default_Supplier field of the Item DB (Or more precisely, the 'Dynamic Link' table), however I imagine this is where most users store this info?


Mitchy25:
Need to fix erpnext.stock.doctype.stock_reconciliation.test_stock_reconciliation.TestStockReconciliation

Mitchy25:
Reopening

PR #18854, opened 2019-08-28
fix: exception in Shopify Settings

ruchamahabal:
On save:

shopify-1


erpnext/erpnext_integrations/doctype/shopify_settings/shopify_settings.py
```diff
@@ -71,7 +71,7 @@ def unregister_webhooks(self):
 				res.raise_for_status()
 				deleted_webhooks.append(d)
 			except Exception as e:
-				frappe.log_error(message=frappe.get_traceback(), title=e.message[:140])
+				frappe.log_error(message=frappe.get_traceback(), title=e[:140])
```
hrwX:
```suggestion
				frappe.log_error(message=frappe.get_traceback(), title=e)
```

hrwX (approved):

PR #18863, opened 2019-08-28
fix: Setting expire_allocation to None

Mitchy25:
This fixes the error where there are no leave_type_records and expire_allocation doesn't get assigned any value causing the following error:

UnboundLocalError: local variable 'expire_allocation' referenced before assignment


Mangesh-Khairnar:
This is not the right fix and changes the intended functionality of the feature. This has been fixed over here #18853

PR #18867, opened 2019-08-29
fix(auto attendance): handling None case for IN/OUT Logs

karthikeyan5:
This PR Fixes the following error reported by a community member @hereabdulla .
G

An error was occurring due to an unhandled  `None` returned by a function. This has now been fixed in this PR.


nabinhait (approved):

hereabdulla:
@nabinhait  it is a fix only then it has to be merged in hotfix right?

PR #18901, opened 2019-09-03
fix: Add UOM in analytics report when viewing based on item

deepeshgarg007:
Show Stock  UOM when viewing Sales/Purchase analytics report based on item


nabinhait (approved):

SaiFi0102:
You also need to get Qty from `stock_qty` rather than Transaction's `qty` field for this to be useful

PR #18967, opened 2019-09-09
fix: proper mapping during creation of purchase order from matterial request

Anurag810:
before it was not mapping suppliers and items on creating a purchase order from material requests.
This will fix.


saurabh6790 (changes requested):
Please check comments https://github.com/frappe/erpnext/pull/18969

PR #18968, opened 2019-09-09
fix: Set Price List in case of User Permissions

netchampfaris:
- Pick Price List if there is one User Permission record


nabinhait (approved):

nabinhait:
@netchampfaris PR for develop?

netchampfaris:

PR #18969, opened 2019-09-09
fix: proper mapping during creation of purchase order from matterial request

Anurag810:
before it was not mapping suppliers and items on creating a purchase order from material requests.
This will fix.

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.py
+++ b/erpnext/stock/doctype/material_request/material_request.py
@@ -251,16 +251,17 @@ def update_status(name, status):
 
 @frappe.whitelist()
 def make_purchase_order(source_name, target_doc=None):
-
 	def postprocess(source, target_doc):
 		if frappe.flags.args and frappe.flags.args.default_supplier:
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/material_request/material_request.py
```diff
@@ -251,16 +251,17 @@ def update_status(name, status):
...
-
 	def postprocess(source, target_doc):
 		if frappe.flags.args and frappe.flags.args.default_supplier:
+			supplier = frappe.flags.args.default_supplier
 			# items only for given default supplier
 			supplier_items = []
 			for d in target_doc.items:
-				default_supplier = get_item_defaults(d.item_code, target_doc.company).get('default_supplier')
-				if frappe.flags.args.default_supplier == default_supplier:
+				supplier_data = get_item_defaults(d.item_code, target_doc.company).get("supplier_items")
```
nabinhait, line 260: `get_item_defaults` does not return supplier_items. What is the purpose of the PR?

nabinhait (changes requested):

PR #18975, opened 2019-09-09
fix: dont make dashboards if company is not set

scmmishra:


erpnext/setup/setup_wizard/data/dashboard_charts.py
```diff
@@ -3,8 +3,18 @@
 import frappe
 import json
 
+def get_company_for_dashboards():
+	company = frappe.defaults.get_defaults().company
+	if company:
+		return company
+	else:
+		company_list = frappe.get_all("Company")
```
nabinhait: It should be `get_list` to honour the permissions.

nabinhait (changes requested):

nabinhait (approved):

PR #18976, opened 2019-09-09
fix: While fetching the Sales Order in Production Plan, the Grand Total value is not being fetched properly.

Anurag810:
fix: While fetching the Sales Order in Production Plan, the Grand Total value is not being fetched properly.


nabinhait (approved):

nabinhait:
@Anurag810 Why the PR raised to `version-12` branch? It should be either `version-11-hotfix` or `develop`.

PR #18990, opened 2019-09-10
fix: pull actual transferred qty for manufacture (develop)

Alchez:
**Problem:**

In non-discrete (process) manufacturing, production of items may require more raw materials than defined in the item's BOM. If material backflush is set to be based on transferred materials, the system is not honouring the setting and is instead only picking up the required quantities for that item.

**Changes:**

- (Enhancement) If transferred raw material quantity is more than the required quantity defined in the BOM, the system now proportionally consumes them for each unit of the manufactured item.
  - For example, if 10 units of a raw material item is required for producing 10 units of the production item, but 15 units are actually transferred / consumed, then 1.5 units of the raw material will be consumed per unit of the production item.
- (Format) Made the codebase slightly easier to read.
- (Format) De-duplicated the stock queries.

<hr>

**Screenshots / GIFs:**

**Before:**

before-stock-fix

**After:**

after-stock-fix


rohitwaghchaure:
For example, if 10 units of a raw material item is required for producing 10 units of the production item, but 15 units are actually transferred / consumed, then 1.5 units of the raw material will be consumed per unit of the production item.

**If the stock uom of the raw materials is in Nos then the solution will not work. For example user trying to make finished good entry for 1 item for which system will consumed the 1.5 Nos of raw materials. But as the stock uom Nos has "Must be Whole Number" checkbox checked. System will not allowed user to consumed the 1.5 qty and it will throw an error "Quantity 1.5 cannot be a fraction"**

Alchez:
@rohitwaghchaure The example I used was only to note that the items would now be proportionally consumed with the required quantity.

If the user inputs a non-integer value for consumed / transferred qty, then the system would be correct to throw that validation, and the user would be required to change the consumed qty to an integer.

rohitwaghchaure:
@Alchez If the qty is fractional and the UOM has "must_be_whole_number" property enabled then ceil the quantity.

Alchez:
@rohitwaghchaure, I've made the change, can you have a look?

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -995,85 +995,84 @@ def get_unconsumed_raw_materials(self):
...
-			qty= item.qty
 			item_code = item.original_item or item.item_code
-			req_items = frappe.get_all('Work Order Item',
+			transferred_qty = item.qty
+			transferred_qty_each = flt(transferred_qty / qty_to_manufacture)
+
+			required_item = frappe.get_all("Work Order Item",
 				filters={'parent': self.work_order, 'item_code': item_code},
 				fields=["required_qty", "consumed_qty"]
 				)
```
rohitwaghchaure: What if user has transferred extra materials which is not there in the BOM Item?

Alchez: I've made a fix so that it defaults the consumed quantity to zero, if it doesn't exist in the BOM or WO.

rohitwaghchaure:
@Alchez While doing testing I am getting the below error
Screenshot 2020-01-01 at 1 20 40 PM

1. Created work order for 3 quantity 

1. Created "Material Transfer for Manufacture" stock entry against the work order and added extra item "PCI Slot" with quantity as 5. 

1. Then tried to create the Manufacturing entry with quantity as 1, got the below error
"Quantity (1.667) cannot be a fraction in row 5"

Alchez:
@rohitwaghchaure, sorry, I missed a place where the quantity is set. Should be fixed with the latest changes. Thanks for testing!

rohitwaghchaure:
@Alchez Are you still working on it?

Alchez:
@rohitwaghchaure, I'm not working actively on this, but I keep merging to check for status from automated tests, since this module is fairly active now.

I would also like to get feedback from you, if possible.

rohitwaghchaure:
@Alchez we are working on the new design(removing Back-flush raw materials based on from Manufacturing Settings) in which we will fix this issue. For now closing this PR

PR #19000, opened 2019-09-11
fix(asset): do not submit journal entries with active workflow

Mangesh-Khairnar:
**Fixes:** Asset creates a depreciation journal entry via the scheduler. If multiple workflow state exists corresponding to a particular docstatus, this makes it impossible to change the workflow state after the submission of a document. This PR stops the submission of a journal entry when the workflow for the same is active. The user would have to manually approve the journal entry via the corresponding workflow.


asbasawaraj:
@nabinhait Can you check this is okay? This is a breaking change. I would recommend that we set the workflow field also automatically.

nabinhait:
@rmehta Any input on this?

rmehta:
Looks okay to me. I think like deferred, we should have depreciation manual too.

nabinhait:
> Looks okay to me. I think like deferred, we should have depreciation manual too.

@rmehta Thanks for the input. We already have manual depreciation option for asset. Merging it.

nabinhait (approved):

muchai:
What of a case where the depreciation entries are many? Could there be an option to auto submit depreciation entries even if the workflow is active. Like an optional check under settings.

PR #19011, opened 2019-09-12
refactor: cost center

sammish:
This feature will help disable and enable each cost center.


erpnext/accounts/doctype/cost_center/cost_center.json
```diff
@@ -1,457 +1,172 @@
...
-  }, 
+   "fieldname": "cost_center_name",
+   "fieldtype": "Data",
+   "in_list_view": 1,
+   "label": "Cost Center Name",
+   "no_copy": 1,
+   "oldfieldname": "cost_center_name",
+   "oldfieldtype": "Data",
+   "reqd": 1,
+   "unique": 1
```
nabinhait: cost_center_name should not be unique, it's ID is already unique. There can be duplicate cost_center_name in case of multiple companies.

nabinhait (changes requested):

nabinhait (approved):

erpnext/accounts/doctype/cost_center/cost_center.json
nabinhait: removed it directly

PR #19021, opened 2019-09-12
feat: add bank account in payroll entry

Mangesh-Khairnar:
- **Fixes:** Journal Entry Account has a bank account no field which actually fetches bank account. 
- This PR also adds a bank account field in payroll entry from which payments will be processed, the payment account in payroll entry is fetched from the bank account.


erpnext/patches/v12_0/rename_bank_account_field_in_journal_entry_account.py
```diff
@@ -0,0 +1,21 @@
...
+from __future__ import unicode_literals
+import frappe
+from frappe.utils import getdate, today
+
+def execute():
+	""" Generates leave ledger entries for leave allocation/application/encashment
+		for last allocation """
+	if not frappe.get_meta("Journal Entry Account").has_field("bank_account"):
+		frappe.reload_doc("Accounts", "doctype", "Journal Entry Account")
+		update_journal_entry_account_fieldname()
```
nabinhait, line 12: Use `rename_field` function to rename the field, it will automatically migrate the data from bank_account_no to bank_account field. It will also take care of other place where the field is used like 'Property Setter', 'report', custom script etc.

nabinhait (changes requested):

erpnext/patches/v12_0/rename_bank_account_field_in_journal_entry_account.py
Mangesh-Khairnar: added changes

nabinhait (approved):

PR #18726, opened 2019-08-14
fix: only select serial no that are present in the selected batch

Alchez:
**Problem:**

If an Item is setup to have both batches and serialized inventory, and a Batch is selected in any of the related documents, the "Add Serial No" dialog doesn't filter for serials present in that Batch.

**Solution:**

If a batch is selected, only pull Serial Nos that are allocated to that batch in the selected warehouse, otherwise pull all Serial Nos for that item in the selected warehouse.

<hr>

**Screenshots / GIFs:**

**Before:**

serial-no-before

**After:**

serial-no-after

PR #19022, opened 2019-09-12
fix: pull serial numbers linked to batches + pull warehouse correctly (develop)

Alchez:
Made to hotfix at #18726.

<hr>

**Problem:**

- If an Item is setup to have both batches and serialized inventory, and a Batch is selected in any of the related documents, the "Add Serial No" dialog doesn't filter for serials present in that Batch.
- For certain documents, and in certain scenarios, the serial no selector is pulling the wrong warehouse for filters.

**Solution:**

- If a batch is selected, only pull Serial Nos that are allocated to that batch in the selected warehouse, otherwise pull all Serial Nos for that item in the selected warehouse.
- Contextually pull the right warehouse to filter serial numbers, with fallbacks.


erpnext/public/js/controllers/transaction.js
```diff
@@ -1760,14 +1760,39 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	}
 });
 
-erpnext.show_serial_batch_selector = function(frm, d, callback, on_close, show_dialog) {
+erpnext.show_serial_batch_selector = function (frm, d, callback, on_close, show_dialog) {
+	let warehouse, receiving_stock, existing_stock;
+	if (frm.doc.is_return) {
+		receiving_stock = true;
```
nabinhait: In case of purchase return, should it be false?

nabinhait (changes requested):

erpnext/public/js/controllers/transaction.js
Alchez: You're right, I've handled purchase documents now too with the latest commit.

Although during testing, I noticed that if an item is setup for both batch and serial numbers, the system always defaults to the "Select Batch" prompt. This is problematic for selecting serial numbers, since that dialog box never shows up. Not sure how we should handle such cases.

nabinhait (approved):

PR #19037, opened 2019-09-13
fix: produced_qty field hidden and not updated in sales order item

ruchamahabal:
fix for:
- produced_qty field in Sales Order Item hidden.
- When produced_qty field was updated in Work Order, it was not getting updated in linked Sales Order Item


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -212,6 +212,7 @@ def update_work_order_qty(self):
 					self.meta.get_label(fieldname), qty, completed_qty, self.name), StockOverProductionError)
 
 			self.db_set(fieldname, qty)
+			frappe.db.set_value('Sales Order Item', self.sales_order_item, 'produced_qty', self.produced_qty)
```
nabinhait: There can be multiple work order against same sales order item. In that case, it should consider sum(produced_qty)

erpnext/patches/v12_0/set_produced_qty_field_in_sales_order_for_work_order.py
```diff
@@ -0,0 +1,9 @@
+import frappe
+from frappe.utils import flt
+
+def execute():
+    frappe.reload_doctype('Sales Order Item')
+    frappe.reload_doctype('Sales Order')
+    work_orders = frappe.db.get_all('Work Order', ['name', 'produced_qty', 'sales_order_item'], {'sales_order_item': ('!=', None)})
+    for work_order in work_orders:
+        frappe.db.set_value('Sales Order Item', work_order.get('sales_order_item'), 'produced_qty', flt(work_order.get('produced_qty')))
```
nabinhait: Consider above comment

nabinhait (changes requested):

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -213,6 +213,20 @@ def update_work_order_qty(self):
 
 			self.db_set(fieldname, qty)
 
+			#for multiple work orders against same sales order item
```
nabinhait: Make it a function and put it inside sales_order.py, this way you will be able to call the function very easily to correct  the produced_qty in sales order (just like the use case of below patch)

nabinhait (changes requested):

erpnext/manufacturing/doctype/work_order/work_order.py
ruchamahabal: sales_order.py or sales_order_item.py? As a single work order is for producing one single item, so produced_qty will be updated in sales_order_item.

nabinhait: > sales_order.py or sales_order_item.py? As a single work order is for producing one single item, so produced_qty will be updated in sales_order_item.

Keep it in sales_order.py, we don't write any logic in child table's file.

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -1023,3 +1023,16 @@ def update_item_quantity(source, target, source_parent):
...
+	linked_wo_with_so_item = frappe.db.get_all('Work Order', ['produced_qty'], {
+		'sales_order_item': sales_order_item,
+		'docstatus': 1
+	})
+	if len(linked_wo_with_so_item) > 0:
+		total_produced_qty = 0
+		for wo in linked_wo_with_so_item:
+			total_produced_qty += flt(wo.get('produced_qty'))
+		frappe.db.set_value('Sales Order Item', sales_order_item, 'produced_qty', total_produced_qty)
+		frappe.db.commit()
```
nabinhait: Commit is not required. It can lead to committing partially if there is an exceptions after it executed.

nabinhait (changes requested):

nabinhait (commented):
@ruchamahabal Please check if this fix is required for version-11 as well.

erpnext/selling/doctype/sales_order/sales_order.py
ruchamahabal: In the patch it does not work if I don't have frappe.db.commit. Should I put a commit there instead?

nabinhait (approved):

PR #19042, opened 2019-09-13
fix: circular dependency issue for the asset invoice cancellation

rohitwaghchaure:
**Issue**

1. Purchase receipt created with asset in draft

1. Invoice created against the purchase receipt

1. Now when trying to cancel the invoice, system throws an error that asset is linked with the invoice


nabinhait (commented):
Closing this, as asset field will be removed from PR / PI via https://github.com/frappe/erpnext/pull/19369

PR #19043, opened 2019-09-13
fix: circular dependency issue for the asset

rohitwaghchaure:
Port https://github.com/frappe/erpnext/pull/19042


nabinhait (commented):
Closing this, as asset field will be removed from PR / PI via https://github.com/frappe/erpnext/pull/19369

PR #19046, opened 2019-09-13
feat(delivery note): make the delivery trip plus button fetch data fr…

DeeMysterio:
**Problem:** 
The customer details aren't fetched when we create Delivery trip from + sign in the Delivery Note Dashboard.

**Screenshots/Gifs:**
dt-plus-button

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.js
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.js
@@ -123,6 +123,9 @@ erpnext.stock.DeliveryNoteController = erpnext.selling.SellingController.extend(
 	setup: function(doc) {
 		this.setup_posting_date_time_check();
 		this._super(doc);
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/delivery_note/delivery_note.js
```diff
@@ -239,7 +242,7 @@ erpnext.stock.DeliveryNoteController = erpnext.selling.SellingController.extend(
 	make_delivery_trip: function() {
 		frappe.model.open_mapped_doc({
 			method: "erpnext.stock.doctype.delivery_note.delivery_note.make_delivery_trip",
-			frm: this.frm
+			frm: cur_frm
```
Anurag810: Use ```frm``` instead of ```cur_frm```

nabinhait (approved):

PR #19050, opened 2019-09-14
fix: Do not validate warehouse in stock entry if purpose is  Repack

deepeshgarg007:
The scenario here is if the items are fetched from BOM and the items in target warehouse are split into multiple rows and entered manually so that it can be repacked into multiple batches the rows that are entered manually do not have `bom_no` field set which causes the below error to arise.

image


nabinhait (commented):
@deepeshgarg007 What is the use case for having multiple batches for the same finished good item?

nabinhait (approved):

Issue #8421, opened 2017-04-12
Feature Request Submit Salary and post JV from salary slips


RandyLowery:
Based in this topic: https://discuss.erpnext.com/t/salary-slip-submitted-but-no-journal-entries/22461 When creating payroll for employee theres no way to make or submit a Jounal Voucher for the employee individualy.

jwrober:
+1

Tropicalrambler:
We just finished recently entering laggard payroll entries in our business.  Currently, Payroll Slips get all bunched up into one journal entry against a specified account. In our case: "Salaries Payable".  This is fine for us, because it registers one large sum per payroll against multiple sums for each account involved in the salary structure.   The problem arises when registering the actual cash payment.
One bank groups all the individual payment amounts into one large sum (similar to the posting against "Salaries Payable", but another one separates each payment into different discrete transactions.

The current solution is to manually create the journal entry against the payroll, for the amount required, on the date and with the reference number. This works fine for now.

Expected functionality: Have the option of creating "One Payment for the entire payroll amount" or "Individual Payment Entries for each employee".  The accounting against Salaries Payable is fine to be left in a grouped fashion as is right now. The payment entries (Journal Entries) should be created as draft, referencing the individual salary slip or employee, and then the user simply should add the confirmation numbers from the bank or change dates if desired.

vrms:
seems this problem still exists today (v11 released just some days ago)
https://discuss.erpnext.com/t/pay-salaries-without-touching-the-hr-module/43664/10
https://discuss.erpnext.com/t/v11-payroll-entry-party-type-party-required-for-payable-account-salaries-payable/44931
I actually wonder how anybody is able to create any proper accounting entries for salary payments, no matter Payroll is involved or not.

kingraphaii:
Any developments on this issue? Payroll Entry is buggy on the current version so creating them manually is the only option I have, but no way to record this in the books.

PR #19051, opened 2019-09-14
feat(Module): Module on Demand

hrwX:
- Now select modules which should be installed during new site installation.
  - All the doctypes under that module for which `create_on_install` is checked, the tables will be created in database
  - All the doctypes under disabled modules, their meta is loaded in DocType so ListView and FormView is rendered using that
- You need to click on `Enable Module` in ModuleView to create tables for the doctypes which come under that particular module
- If you try to insert any data in doctypes with not db tables, itll return `Module Disabled Error`.

Refer [#8421](https://github.com/frappe/frappe/pull/8421)


nabinhait (changes requested):
@hrwX Please fix merge conflict

PR #19055, opened 2019-09-15
fix: Create error log if something goes wrong while call log creation

surajshetty3416:
Create Error Log if something goes wrong while creating a call log for better debugging


erpnext/erpnext_integrations/exotel_integration.py
```diff
@@ -1,25 +1,30 @@
...
+		if status == 'free':
+			return
+
+		call_log = get_call_log(call_payload)
+		if not call_log:
+			create_call_log(call_payload)
+		else:
+			update_call_log(call_payload, call_log=call_log)
+	except Exception as e:
+		frappe.log_error(title=_('Error in Exotel incoming call'))
```
nabinhait, line 27: First rollback, then log_error and commit

nabinhait (changes requested):

erpnext/erpnext_integrations/exotel_integration.py
surajshetty3416: @nabinhait Why do we require to rollback?
There's only one method which is trying to create/update a log... if it fails or something fails before that, then there will be nothing to rollback...

**Note:** This is a whitelisted method which will be called through a GET request.

nabinhait: Suppose, an error raised in [on_update](https://github.com/frappe/erpnext/blob/develop/erpnext/communication/doctype/call_log/call_log.py#L22) while creating call log. The call log has already been inserted in the database at that time. If we don't rollback, it will get committed via except block.
It is always a good practice, to do rollback in `except` if we are committing at the end of `try` block.

nabinhait (changes requested):

nabinhait (approved):

PR #19068, opened 2019-09-16
fix: Decimal point issue for e-invoice

rohitwaghchaure:


nabinhait (approved):

toofun666:
I think this update just effed up everything related to decimal fields on forms!

PR #19073, opened 2019-09-16
fix: Mandatory accounting dimensions while creating asset depreciation entry

deepeshgarg007:
Add mandatory accounting dimensions while creating Asset Depreciation Entry

### DIFF
--- a/erpnext/assets/doctype/asset/depreciation.py
+++ b/erpnext/assets/doctype/asset/depreciation.py
@@ -70,12 +70,12 @@ def make_depreciation_entry(asset_name, date=None):
 			}
 
 			for dimension in accounting_dimensions:
-				if (asset.get(dimension['fieldname']) or dimension.get('mandatory_for_pl') or
-					dimension.get('mandatory_for_bs')):
+				if (asset.get(dimension['fieldname']) or dimension.get('mandatory_for_bs')):
 					credit_entry.update({
 						dimension['fieldname']: asset.get(dimension['fieldname']) or dimension.get('default_dimension')
 					})
 
+				if (asset.get(dimension['fieldname']) or dimension.get('mandatory_for_pl')):
 					debit_entry.update({
 						dimension['fieldname']: asset.get(dimension['fieldname']) or dimension.get('default_dimension')
 					})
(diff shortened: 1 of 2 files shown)


erpnext/assets/doctype/asset/depreciation.py
```diff
@@ -51,20 +54,35 @@ def make_depreciation_entry(asset_name, date=None):
...
+			}
+
+			for dimension in accounting_dimensions:
+				if (asset.get(dimension['fieldname']) or dimension.get('mandatory_for_pl') or
+					dimension.get('mandatory_for_bs')):
+					credit_entry.update({
+						dimension['fieldname']: asset.get(dimension['fieldname']) or dimension.get('default_dimension')
+					})
+
+					debit_entry.update({
```
nabinhait, line 79: debit_entry will be updated if `asset.get(dimension['fieldname']) or dimension.get('mandatory_for_pl')`

erpnext/assets/doctype/asset/depreciation.py
```diff
@@ -51,20 +54,35 @@ def make_depreciation_entry(asset_name, date=None):
...
 				"reference_type": "Asset",
 				"reference_name": asset.name,
 				"cost_center": depreciation_cost_center
-			})
+			}
+
+			for dimension in accounting_dimensions:
+				if (asset.get(dimension['fieldname']) or dimension.get('mandatory_for_pl') or
+					dimension.get('mandatory_for_bs')):
+					credit_entry.update({
```
nabinhait, line 74: credit_entry will be updated if `asset.get(dimension['fieldname']) or dimension.get('mandatory_for_bs')`

nabinhait (changes requested):

erpnext/assets/doctype/asset/depreciation.py
deepeshgarg007: @nabinhait understood accumulated_depreciation_account is "Balance Sheet" account and depreciation_expense_account is "Profit and Loss Account". Made the requested change

nabinhait (approved):

PR #19075, opened 2019-09-16
fix: show a descriptive message on submission of duplicate account

Mangesh-Khairnar:
On creation of parent account when child account with a similar name exists. This allows creation if all other states are similar.
Screenshot 2019-07-25 at 7 31 42 PM

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -126,7 +126,8 @@ class Account(NestedSet):
 					"account_currency": None,
 					"parent_account": parent_acc_name_map[company]
 				})
-				doc.save()
+				if not self.check_if_child_acc_exists(doc):
+					doc.save()
 				frappe.msgprint(_("Account {0} is added in the child company {1}")
 					.format(doc.name, company))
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/account/account.py
```diff
@@ -170,6 +171,24 @@ def validate_account_currency(self):
...
+		}, ['company', 'account_currency', 'is_group', 'root_type', 'account_type', 'balance_must_be', 'account_name'], as_dict=1)
+
+		if not info:
+			return
+
+		doc = vars(doc)
+		dict_diff = [k for k in info if k in doc and info[k] != doc[k] and k != "company"]
+		if dict_diff:
+			frappe.throw(_("Account {0} already exists in child company {1}. The following fields have different values, they should be same:<ul><li>{2}</li></ul>")
+				.format(info.account_name, info.company, '</li><li>'.join(dict_diff)))
```
deepeshgarg007, line 188: @Mangesh-Khairnar instead of the fieldname the label of the field should be shown in list

deepeshgarg007 (changes requested):

PR #19097, opened 2019-09-17
fix(Integrations): Shopify Integration

prssanna:
Shopify sync now works. Sales Orders, Sales Invoices and Delivery Notes get synced from Shopify to ERPNext:

Screenshot 2019-09-17 at 9 36 38 PM

Screenshot 2019-09-17 at 9 36 59 PM

Screenshot 2019-09-17 at 9 37 24 PM


erpnext/erpnext_integrations/connectors/shopify_connection.py
```diff
@@ -246,6 +251,7 @@ def update_taxes_with_shipping_lines(taxes, shipping_lines, shopify_settings):
 	return taxes
 
 def get_tax_account_head(tax):
+	return
```
saurabh6790: why?

erpnext/erpnext_integrations/doctype/shopify_settings/shopify_settings.py
```diff
@@ -21,37 +21,21 @@ def validate(self):
...
-			if method in created_webhooks:
-				continue
-
 			session = get_request_session()
 			try:
 				d = session.post(url, data=json.dumps({
 					"webhook": {
 						"topic": method,
-						"address": get_webhook_address(connector_name='shopify_connection', method='store_request_data'),
+						"address": frappe.utils.get_url(),
```
saurabh6790: get_webhooks_address returns url with specific endpoint
```
endpoint = "erpnext.erpnext_integrations.connectors.{0}.{1}".format(connector_name, method)

	if exclude_uri:
		return endpoint

	try:
		url = frappe.request.url
	except RuntimeError:
		url = "http://localhost:8000"

	server_url = '{uri.scheme}://{uri.netloc}/api/method/{endpoint}'.format(uri=urlparse(url), endpoint=endpoint)

	return server_url
```

frappe.utils.get_url only returns only `{uri.scheme}://{uri.netloc}`

saurabh6790 (changes requested):

erpnext/erpnext_integrations/connectors/shopify_connection.py
prssanna: I had added it while testing, forgot to remove it

saurabh6790 (approved):

PR #19101, opened 2019-09-18
chore: remove internal fields from print of purchase documents

Alchez:
This is mostly to avoid certain fields showing up in the print view, there are no functional changes.

**Changes:**
- Purchase Invoice
  - Hide "Status"
  - Hide "Print Language"
- Purchase Invoice Item
  - Hide "Include Exploded Items"
  - Show "Purchase Order"
- Purchase Order Item
  - Hide "Include Exploded Items"
  - Hide "Last Purchase Rate"
- Purchase Receipt Item
  - Hide "Include Exploded Items"
  - Show "Purchase Order"

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
+++ b/erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
@@ -591,7 +591,6 @@
    "oldfieldname": "purchase_order",
    "oldfieldtype": "Link",
    "options": "Purchase Order",
-   "print_hide": 1,
    "read_only": 1,
    "search_index": 1
   },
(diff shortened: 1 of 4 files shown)


erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -591,7 +591,6 @@
    "oldfieldname": "purchase_order",
    "oldfieldtype": "Link",
    "options": "Purchase Order",
-   "print_hide": 1,
```
nabinhait, line 594: purchase_order should be hidden from print.

nabinhait (changes requested):

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
Alchez: It's useful to have a reference for the other party to get a proper trail.

nabinhait (approved):

PR #19110, opened 2019-09-19
fix: User permission issue in "Total Stock Summary" report

surajshetty3416:
Data in "Total Stock Summary" report were not getting filtered
based on applied user permissions because some link fields had
wrong options

frontport of: https://github.com/frappe/erpnext/pull/19103

### DIFF
--- a/erpnext/stock/report/total_stock_summary/total_stock_summary.py
+++ b/erpnext/stock/report/total_stock_summary/total_stock_summary.py
@@ -15,8 +15,8 @@ def execute(filters=None):
 
 def get_columns():
 	columns = [
-		_("Company") + ":Link/Item:250",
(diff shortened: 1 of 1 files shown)


Anurag810 (approved):

jigneshpshah:
Hi @surajshetty3416 ,

it seems because of this change the report has stopped working

image

#### Error comes as below
**ProgrammingError:** (1064, u"You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'Company ABC'' GROUP BY ledger.warehouse, item.item_code' at line 13")

**MariaDB Version - 10.2.23-MariaDB-1:10.2.23+maria~bionic-log**

PR #19112, opened 2019-09-19
fix: get_bin_details_and_serial_nos() takes at least 3 arguments (4 g…

rohitwaghchaure:
**Issue**

```
  File "/home/frappe/benches/bench-version-11-2019-09-10/apps/frappe/frappe/app.py", line 61, in application
  File "/home/frappe/benches/bench-version-11-2019-09-10/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-version-11-2019-09-10/apps/frappe/frappe/handler.py", line 56, in execute_cmd
  File "/home/frappe/benches/bench-version-11-2019-09-10/apps/frappe/frappe/__init__.py", line 1036, in call
    return fn(*args, **newargs)
TypeError: get_bin_details_and_serial_nos() takes at least 3 arguments (4 given)
```

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -808,7 +808,7 @@ def get_serial_no_details(item_code, warehouse, stock_qty, serial_no):
 	return {'serial_no': serial_no}
 
 @frappe.whitelist()
-def get_bin_details_and_serial_nos(item_code, warehouse, has_batch_no, stock_qty=None, serial_no=None):
+def get_bin_details_and_serial_nos(item_code, warehouse, has_batch_no=None, stock_qty=None, serial_no=None):
 	bin_details_and_serial_nos = {}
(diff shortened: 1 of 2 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -808,7 +808,7 @@ def get_serial_no_details(item_code, warehouse, stock_qty, serial_no):
 	return {'serial_no': serial_no}
 
 @frappe.whitelist()
-def get_bin_details_and_serial_nos(item_code, warehouse, has_batch_no, stock_qty=None, serial_no=None):
+def get_bin_details_and_serial_nos(item_code, warehouse, has_batch_no=None, stock_qty=None, serial_no=None):
```
saurabh6790, line 811: instead on None, can set 0 as default arg

saurabh6790 (changes requested):

PR #19121, opened 2019-09-19
feat: Allow to make landed cost voucher against the asset item

rohitwaghchaure:
Gl Entries for an Asset after making of landed cost voucher

Screenshot 2019-09-19 at 7 13 00 PM


nabinhait:
@rohitwaghchaure Travis failing

nabinhait (approved):

PR #19122, opened 2019-09-19
fix: Add a setting to make creation of new Student User optional

ghost:
This pull request adds a setting to the Education Settings single doc that allows admins to skip User creation when a new Student document is created.

### DIFF
--- a/erpnext/education/doctype/education_settings/education_settings.json
+++ b/erpnext/education/doctype/education_settings/education_settings.json
@@ -11,6 +11,7 @@
   "validate_batch",
   "validate_course",
   "academic_term_reqd",
+  "user_creation_skip",
   "section_break_7",
   "instructor_created_by",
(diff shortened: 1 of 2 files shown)


theopen-institute:
Is there anything further I can do to improve this PR?

deepeshgarg007:
@theopen-institute can you mention one use case where user creation of a student is not required.

theopen-institute:
A use case would be any organization that wants to use the student enrollment process without creating user accounts for every student added.

For example, we have used the student applicant -> program enrollment flow for thousands of students, but now every time we do we are forced to create a new User as well. This is undesirable because (A) we are quickly adding many new User accounts that we then must delete manually, and (B) every new Student is confused when they receive an email telling them to create a password, etc., for our ERPNext system.

nabinhait (approved):

PR #19125, opened 2019-09-20
fix: Removed Address section in Customer Quick Entry Dialog Box

marination:
Removed address section with it's mandatory fields as it slowed down quick entry process

Screenshot 2019-09-20 at 6 38 04 AM


rohitwaghchaure:
@marination Why???? Instead of remove mandatory property

PR #19134, opened 2019-09-22
fix(Stock): item variant description

ruchamahabal:
Problem:
- Item variant description updated even if description field is not present in "copy fields to variant" table in  Item Variant Settings.
- Every time the description is updated in template it is appended to the description in item variant
item_variant_prob

### DIFF
--- a/erpnext/controllers/item_variant.py
+++ b/erpnext/controllers/item_variant.py
@@ -283,7 +283,7 @@ def copy_attributes_to_variant(item, variant):
 	if 'description' not in allow_fields:
 		if not variant.description:
 				variant.description = ""
-
+	else:
 		if item.variant_based_on=='Item Attribute':
 			if variant.attributes:
 				attributes_description = item.description + " "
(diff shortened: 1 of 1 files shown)


erpnext/controllers/item_variant.py
```diff
@@ -283,15 +283,15 @@ def copy_attributes_to_variant(item, variant):
...
-
+	else:
 		if item.variant_based_on=='Item Attribute':
 			if variant.attributes:
 				attributes_description = item.description + " "
 				for d in variant.attributes:
 					attributes_description += "<div>" + d.attribute + ": " + cstr(d.attribute_value) + "</div>"
 
 				if attributes_description not in variant.description:
-					variant.description += attributes_description
```
nabinhait, line 294: If attributes_description is not already there, then only it used to added in the description. The previous code was there to keep the user's manual edit in the description. If it is added multiple times, there might be some other issue.

nabinhait (changes requested):

erpnext/controllers/item_variant.py
ruchamahabal: It is getting added multiple times because on line 289:

`attributes_description = item.description + " "` 

attributes_description is also taking item.description first and then the description of the attributes. So every time the item description is updated in the template the description was getting appended to the variant description

PR #19136, opened 2019-09-23
fix: Work order dialog box 'Select Quantity' on clicking 'Finish'

marination:
- The dialog box fetched 0 as Quantity to Manufacture and Max Quantity on finishing a Work Order
- It now fetches the quantity correctly

Screenshot 2019-09-23 at 1 08 54 PM


surajshetty3416 (approved):

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -545,11 +545,14 @@ erpnext.work_order = {
 
 	get_max_transferable_qty: (frm, purpose) => {
 		let max = 0;
-		if (frm.doc.skip_transfer) return max;
-		if (purpose === 'Manufacture') {
-			max = flt(frm.doc.material_transferred_for_manufacturing) - flt(frm.doc.produced_qty);
+		if (frm.doc.skip_transfer) {
+			max = flt(frm.doc.qty) - flt(frm.doc.produced_qty);
```
nabinhait: Why max qty is needed if we are skipping transfer?

nabinhait (approved):

PR #19137, opened 2019-09-23
fix: Outstanding Amount field of Discounted Invoices Table in Invoice Discounting

marination:
It fetched **Grand Total** instead of **Outstanding Amount** from Sales Invoice for the selected discounted invoice.


erpnext/accounts/doctype/discounted_invoice/discounted_invoice.json
```diff
@@ -38,13 +38,11 @@
    "read_only": 1
   },
   {
-   "fetch_from": "sales_invoice.grand_total",
```
deepeshgarg007, line 41: No need to remove this. Just replace grand_total with outstanding_amount and check fetch if empty so that the user entered amount is not overridden . No need to write any client side code
image

erpnext/accounts/doctype/invoice_discounting/invoice_discounting.js
```diff
@@ -205,9 +204,15 @@ frappe.ui.form.on('Invoice Discounting', {
 });
 
 frappe.ui.form.on('Discounted Invoice', {
-	sales_invoice: (frm) => {
+	sales_invoice: (frm, cdt, cdn) => {
 		frm.events.calculate_total_amount(frm);
-		frm.events.refresh_filters(frm);
+		frm.events.refresh_filters(frm);	
+		
+		let row = locals[cdt][cdn];
```
deepeshgarg007: Remove this. This only works if Invoice Discounting Is done through user interface. Is doesn't work if created from server side or through API. Test Cases are failing because of it

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #19149, opened 2019-09-23
use task status closed instead of completed

lasalesi:
* Version 12 bugfix
* Issue description: https://discuss.erpnext.com/t/cannot-mark-project-completed-ver-12/52659
* Task status options have changed, but code does not cover the changes: still the removed "completed" status is used, which is invalid

### DIFF
--- a/erpnext/projects/doctype/task/task.json
+++ b/erpnext/projects/doctype/task/task.json
@@ -109,7 +109,7 @@
     "no_copy": 1,
     "oldfieldname": "status",
     "oldfieldtype": "Select",
-    "options": "Open\nWorking\nPending Review\nOverdue\nCompleted\nCancelled"
+    "options": "Open\nWorking\nPending Review\nOverdue\nClosed\nCancelled"
    },
    {
     "fieldname": "priority",
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -40,9 +40,9 @@ class Task(NestedSet):
 			frappe.throw(_("'Actual Start Date' can not be greater than 'Actual End Date'"))
 
 	def validate_status(self):
-		if self.status!=self.get_db_value("status") and self.status == "Completed":
+		if self.status!=self.get_db_value("status") and self.status == "Closed":
 			for d in self.depends_on:
-				if frappe.db.get_value("Task", d.task, "status") != "Completed":
+				if frappe.db.get_value("Task", d.task, "status") != "Closed":
 					frappe.throw(_("Cannot close task {0} as its dependant task {1} is not closed.").format(frappe.bold(self.name), frappe.bold(d.task)))
 
 			from frappe.desk.form.assign_to import clear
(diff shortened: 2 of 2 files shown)


barredterra:
https://github.com/lasalesi/erpnext/blob/9581904ecd27d6879b85ea05da22b3b54e690161/erpnext/projects/doctype/task/task.json#L112

The options still show "Completed", not "Closed" 🤔

lasalesi:
Thanks for the hint @barredterra !
Interestingly, this seems to not affect actual installations...
Vanilla ERPNext will install (Open, Working, Pending Review, Overdue, Completed, Cancelled), however, on updated systems you will have (Open, Working, Pending Review, Overdue, Closed, Done, Cancelled)

Reverting the "new" status completed back to "Closed" would resolve the issue. Otherwise, the new status "Completed" would have to replace the old status "Closed", note that many tasks are on that status. Keeping both is highly confusing. While I personally prefer "Completed", I would rather have "Closed" working than the current situation...

scmmishra:
@lasalesi what issue does this fix? 

From the discuss link I found in your PR description, I see the following is the issue I could find. 

> I confirm the issue, if you set Task > Percentage to 100%, then save, you will get
>
> Status cannot be "Completed". It should be one of "Open", "Working", "Pending Review", "Overdue", "Closed", "Done", "Cancelled", "Archiv", "Fertig"
>
> You one can manually add new status “Completed”, but this should be working out of the box…

I tried replicating this on a local setup, and tasks worked just fine.

scmmishra:
Is there a good reason for using `Closed` instead of `Completed`? Can you present a good case for it?

scmmishra (commented):
Is there a good reason for using `Closed` instead of `Completed`? Can you present a good case for it?

lasalesi:
> Is there a good reason for using Closed instead of Completed? Can you present a good case for it?

Complete is better (as mentioned), but does not work in an upgraded state. The good reason is: it does not work as it is right now.

> I tried replicating this on a local setup, and tasks worked just fine.

You need to set up an ERPNext v10 instance, then upgrade to v12 (through v11 or directly, makes no difference). Once upgraded, try to complete a project or task (set Task > Percentage to 100%, then save, you will get) and it will fail with 

    Status cannot be "Completed". It should be one of "Open", "Working", "Pending Review", "Overdue", "Closed", "Done", "Cancelled", "Archiv", "Fertig"

As is, this is not working. There is a workaround by manually adding "Completed". There might be a workaround using a patch that will add "Completed" as a valid status, and at the same time replace "Closed" with "Completed" and the remove "Closed" as a status.

scmmishra:
The status was renamed from Closed to Completed. There is a patch that already manages the change between the versions. [ref](https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v12_0/set_task_status.py).

Has the patch run on your instance?

lasalesi:
The patch "set_task_status" was executed but did not do the job (and other people report the same issue)...

nabinhait (commented):
@scmmishra Can you comment on this?

scmmishra:
@nabin I couldn't replicate the said problem, the patches run just fine. I don't see why this change is required

PR #19151, opened 2019-09-24
chore: Add standard chart of accounts template for El Salvador

deepeshgarg007:
Screenshot 2019-09-24 at 8 52 52 AM
Screenshot 2019-09-24 at 8 52 30 AM


nabinhait (approved):

nabinhait (changes requested):
@deepeshgarg007 Filename should be without spaces, replace with underscore

PR #19157, opened 2019-09-24
fix: packed items child table reset on amending docs

Anurag810:
packed items child table reset on amending docs

#19158

### DIFF
--- a/erpnext/stock/doctype/packed_item/packed_item.py
+++ b/erpnext/stock/doctype/packed_item/packed_item.py
@@ -7,7 +7,6 @@ from __future__ import unicode_literals
 import frappe, json
 from frappe.utils import cstr, flt
 from erpnext.stock.get_item_details import get_item_details
-
 from frappe.model.document import Document
 
 class PackedItem(Document):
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -52,28 +51,35 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
...
+		old_packed_items_map = get_old_packed_item_details(doc.packed_items)
+	else:
+		old_packed_items_map = False
 
 	parent_items = []
 	for d in doc.get("items"):
 		if frappe.db.get_value("Product Bundle", {"new_item_code": d.item_code}):
 			for i in get_product_bundle_items(d.item_code):
-				update_packing_list_item(doc, i.item_code, flt(i.qty)*flt(d.stock_qty), d, i.description)
+				update_packing_list_item(doc, i.item_code, flt(i.qty)*flt(d.stock_qty), d, i.description, old_packed_items_map)
```
nabinhait: why it should be passed as an argument? it can be called directly in update_packing_list_item

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -113,3 +119,9 @@ def get_items_from_product_bundle(args):
 
 def on_doctype_update():
 	frappe.db.add_index("Packed Item", ["item_code", "warehouse"])
+
+def get_old_packed_item_details(old_packed_items):
+	old_packed_items_map = {}
+	for items in old_packed_items:
+		old_packed_items_map[items.item_code] = items.as_dict()
```
nabinhait: Key should be based on `(items.item_code, items.parent_item)`, otherwise if an item appears multiple times it will create problem.

nabinhait (changes requested):

nabinhait (approved):

PR #19162, opened 2019-09-24
fix: internal issues

nextchamp-saqib:
[Bypass credit limit label is bad](https://frappe.erpnext.com/desk#Form/Internal%20Issue/FR-ISS-140760) : Just a label fix
[Project Template field not visible](https://frappe.erpnext.com/desk#Form/Internal%20Issue/FR-ISS-130259) : On creating a Project via Quick Entry removes Project Template field as it is supposed to be set only once.


Anurag810 (approved):

jaichavan:
@nextchamp-saqib can we make a field label as "Bypass credit limit check at Sales Order".. as this check is applicable for Sales Order only... Label doesn't indicate the upfront.

PR #19173, opened 2019-09-25
fix: Item Rate within Update Items in Sales order

marination:
Blocked negative discount percentage, which affected the item rate every time Update Items action was taken

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -1114,8 +1114,22 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name):
 		else:
 			child_item.rate = flt(d.get("rate"))
 		if flt(child_item.price_list_rate):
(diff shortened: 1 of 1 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -1113,9 +1113,12 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name):
 						 .format(child_item.idx, child_item.item_code))
 		else:
 			child_item.rate = flt(d.get("rate"))
+
 		if flt(child_item.price_list_rate):
-			child_item.discount_percentage = flt((1 - flt(child_item.rate) / flt(child_item.price_list_rate)) * 100.0, \
+			discount = flt((1 - flt(child_item.rate) / flt(child_item.price_list_rate)) * 100.0,
 				child_item.precision("discount_percentage"))
+			if discount > 0:
```
nabinhait: In case of negative, it should set 'margin'

nabinhait (changes requested):

nabinhait (approved):

PR #19184, opened 2019-09-26
fix(Task): Do not create/schedule task after project end

hrwX:
Earlier, the system allowed creating tasks with end dates after the Project end date. With this fix, the task end date will be validated against the project date. End date of tasks should be before the project end date.


michellealva:
Earlier, the system allowed creating tasks with end dates after the Project end date. With this fix, the task end date will be validated against the project date. End date of tasks should be before the project end date.

nabinhait (dismissed):

nabinhait (approved):

PR #19189, opened 2019-09-27
fix: Allow alternative item in stock entry while transferring material to subcontractor

deepeshgarg007:
Allow to transfer alternative item in Stock Entry while transferring material to subcontractor if "Allow Alternative Item" is checked in BOM

image


nabinhait (approved):

nabinhait (commented):
@deepeshgarg007 Please send a separate PR for version-11-hotfix branch

PR #19193, opened 2019-09-27
fix: Move add_custom_button for Gantt and Kanban Board

ci2014:
Move add_custom_button for Gantt and Kanban Board to set_buttons, because in onload it is not working.


nabinhait (approved):

wojosc:
@nabinhait it seems that the KanBan Board button does not fetch the set filter for the project. Instead it shows all tasks from all projects in kanban view.

PR #19195, opened 2019-09-28
fix: null check for valuation rate

scmmishra:
```
  File "/home/frappe/benches/bench-version-12-2019-09-24/apps/frappe/frappe/app.py", line 60, in application
...
  File "/home/frappe/benches/bench-version-12-2019-09-24/apps/erpnext/erpnext/manufacturing/doctype/bom/bom.py", line 229, in update_cost
    "conversion_factor": d.conversion_factor
  File "/home/frappe/benches/bench-version-12-2019-09-24/apps/erpnext/erpnext/manufacturing/doctype/bom/bom.py", line 179, in get_rm_rate
    rate = self.get_valuation_rate(arg) * (arg.get("conversion_factor") or 1)
TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
```

Valuation rate if null would break, added a `frappe.throw`

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -286,6 +286,9 @@ class BOM(WebsiteGenerator):
 		if not valuation_rate:
 			valuation_rate = frappe.db.get_value("Item", args['item_code'], "valuation_rate")
 
+		if not valuation_rate:
+			frappe.throw(_("Please set valuation rate for item {0}.".format(args['item_code'])))
+
 		return valuation_rate
 
 	def manage_default_bom(self):
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -173,7 +173,7 @@ class BOM(WebsiteGenerator):
 			#Customer Provided parts will have zero rate
 			if not frappe.db.get_value('Item', arg["item_code"], 'is_customer_provided_item'):
 				if arg.get('bom_no') and self.set_rate_of_sub_assembly_item_based_on_bom:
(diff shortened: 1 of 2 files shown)


nabinhait (commented):
@scmmishra Please fix travis

scmmishra:
@rohitwaghchaure @nabin this can be fixed by adding rate to the test records. However this *makes the rate mandatory* for the item when it's added to a BOM. Is this okay to have?

I am confused regarding this one thing.

rohitwaghchaure:
Looks fine

nabinhait:
> @rohitwaghchaure @nabin this can be fixed by adding rate to the test records. However this _makes the rate mandatory_ for the item when it's added to a BOM. Is this okay to have?
> 
> I am confused regarding this one thing.

I think we should not raise mandatory validation while creating BOM, as user can manually enter the rate and proceed.

scmmishra:
> I think we should not raise mandatory validation while creating BOM, as user can manually enter the rate and proceed.

But the following code will break if we use `frappe.msgprint` instead

https://github.com/frappe/erpnext/blob/8b741cfefb4421f32a18b3aa34ed9b7fd22aa8bc/erpnext/manufacturing/doctype/bom/bom.py#L179

Returning a default value, say 0, seems like a bad idea, it may confuse users into submitting incorrect data.

nabinhait:
Just use `flt(self.get_valuation_rate(arg))`, it will replace null with zero.
 I think rate is mandatory in Raw Material table. If it does not found valuation rate, users put it manually. Also, while updating cost later using "Update Cost" button, system only overwrite rate if new rate is non-zero.

scmmishra:
I've made the necessary changes. Need review!

nabinhait (approved):

PR #8515, opened 2017-04-20
Add option to skip material transfer for production orders

ishanloya:
[Forum link](https://discuss.erpnext.com/t/production-order-with-bulk-stock-transfer/21641)

PR #19204, opened 2019-09-30
fix(Global Search): add doctypes for modules

hrwX:
- Add Global Search DocTypes for modules in hooks
- Depends on [#8515](https://github.com/frappe/frappe/pull/8515)


erpnext/hooks.py
```diff
@@ -353,6 +353,7 @@
 	}
 ]
 
+# ERPNext doctypes for Global Search
 global_search_doctypes = [
```
netchampfaris: Structure should be:
```
global_search_doctypes = {
	'Default': [list of doctypes],
	'Healthcare': [list of healthcare doctypes],
	...
}
```

netchampfaris (dismissed):

surajshetty3416 (approved):

PR #19215, opened 2019-09-30
fix(Purchase Invoice): rename payment button on make to payment entry

DeeMysterio:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -62,7 +62,7 @@ erpnext.accounts.PurchaseInvoice = erpnext.buying.BuyingController.extend({
 
 		if(doc.docstatus == 1 && doc.outstanding_amount != 0
 			&& !(doc.is_return && doc.return_against)) {
-			this.frm.add_custom_button(__('Payment'), this.make_payment_entry, __('Create'));
+			this.frm.add_custom_button(__('Payment Entry'), this.make_payment_entry, __('Create'));
```
nabinhait, line 65: Let's keep it Payment only. Payment can be done via Journal Entry as well depends on selection on Accounts Settings.

PR #19222, opened 2019-09-30
feat(Sales Order): Allow to skip delivery note creation

Mangesh-Khairnar:
- Allow users to skip delivery note creation from sales order
Screenshot 2019-10-01 at 1 05 20 AM


erpnext/selling/doctype/sales_order/sales_order.json
```diff
@@ -1171,12 +1173,20 @@
    "fieldtype": "Data",
    "label": "Phone",
    "read_only": 1
+  },
+  {
+   "default": "0",
+   "fieldname": "skip_delivery_note_creation",
+   "fieldtype": "Check",
+   "in_list_view": 1,
+   "label": "Skip Delivery Note Creation",
```
nabinhait: Rename it to "Skip Delivery Note"

nabinhait (changes requested):

erpnext/selling/doctype/sales_order/sales_order.json
Mangesh-Khairnar: added changes

nabinhait (approved):

PR #19224, opened 2019-09-30
fix: 'Warehouse' filter in the Stock Ledger report

Mitchy25:
The SQL was formed incorrectly.


nabinhait (commented):
@Mitchy25 What was the problem with existing query? 
Even if you refactor using new approach, then you have to use `warehouse in ()` instead of `warehouse = ()`, to handle group warehouse filter. And `in` becomes very slow if there are huge number of children of that group warehouse.

Mitchy25:
Hi @nabinhait,

The original filter didn’t actually filter by warehouse. It instead returned all warehouses. This fixes the filter logic

Mitchy25:
Previous code didnt set the field name 'warehouse' equal to the SQL return. So the SQL returned but didnt change the row return.

nabinhait (commented):
I can't replicate the issue of filtering by warehouse. Please check attached GIF. Closing the pull request.
stock-ledger-warehouse-filter

PR #19225, opened 2019-10-01
fix(contract): make the is_signed field read-only

DeeMysterio:


erpnext/crm/doctype/contract/contract.json
```diff
@@ -50,7 +50,8 @@
    "fieldname": "is_signed",
    "fieldtype": "Check",
    "label": "Signed",
-   "no_copy": 1
+   "no_copy": 1,
+   "read_only": 1
```
deepeshgarg007, line 54: @DeeMysterio The `is_signed` field is used to toggle Signee Details Section. If it is made read only, how will you enter the signee details and how the `is_signed` field will be updated because currently there is no server side code to update it.
Contract

deepeshgarg007 (changes requested):

PR #19228, opened 2019-10-01
fix(stock-ageing): filter none values from the fifo queue

Mangesh-Khairnar:
skip None posting date in the fifo queue


erpnext/stock/report/stock_ageing/stock_ageing.py
```diff
@@ -15,6 +15,8 @@ def execute(filters=None):
 	to_date = filters["to_date"]
 	data = []
 	for item, item_dict in iteritems(item_details):
+
+		item_dict['fifo_queue'] = [item for item in item_dict if item[1]]
```
deepeshgarg007:
```suggestion
		temp_fifo = [d for d in item_dict['fifo_queue'] if d[1]]
```
'item' variable name is already used above in for loop and you need to filter item_dict['fifo_queue'] not the entire item_dict
After filtering sort the temp_fifo

deepeshgarg007 (changes requested):

erpnext/stock/report/stock_ageing/stock_ageing.py
Mangesh-Khairnar: added changes

deepeshgarg007 (approved):

Issue #19196, opened 2019-09-28
[New Feature][Sponsored] Integrate Whatsapp for business


fproldan:
Hello community.

We will develop this new feature that is initially sponsored by @idolaru (Iulian_Olaru
in discourse).

We encourage the community and the frappe team to bring up ideas of how is the better way to approach this development in order to be integrated into the core.

@yashodhank can you please add the corresponding labels to this issue and make it visible to the developers of frappe team that you consider necessary.

Below we describe our first approach 

<hr>

### New Integration
![Captura de Pantalla 2019-09-28 a la(s) 10 22 29](https://user-images.githubusercontent.com/4945122/65817232-2ec10580-e1db-11e9-8bef-81e2b289ae07.png)

Vendor to be defined:
- Twilio: https://www.twilio.com/whatsapp
- Nexmo: https://www.twilio.com/whatsapp
- Messagebird https://www.messagebird.com/whatsapp/
- Community proposals

<hr>

### [Option 1] SMS Center -> Message Center (Name to be defined)
- Message type: SMS, Whatsapp

![Captura de Pantalla 2019-09-28 a la(s) 10 04 02](https://user-images.githubusercontent.com/4945122/65817194-bf4b1600-e1da-11e9-9987-7a44e5b30dc5.png)

### [Option 2] Whatsapp Center
![Captura de Pantalla 2019-09-28 a la(s) 10 27 25](https://user-images.githubusercontent.com/4945122/65817339-3e8d1980-e1dc-11e9-9d67-af0ec24923c9.jpg)

<hr>

### Notification
![Captura de Pantalla 2019-09-28 a la(s) 10 31 58](https://user-images.githubusercontent.com/4945122/65817282-abec7a80-e1db-11e9-8330-d12e38170eea.jpg)

<hr>

### [Option 1] Send SMS -> Send Message (Name to be defined)
![Captura de Pantalla 2019-09-28 a la(s) 10 06 41](https://user-images.githubusercontent.com/4945122/65817289-c4f52b80-e1db-11e9-87b9-a148eab0f7d0.png)

### [Option 2] Send Whatsapp
![Captura de Pantalla 2019-09-28 a la(s) 09 53 45](https://user-images.githubusercontent.com/4945122/65817343-54024380-e1dc-11e9-83b6-0a6a5fb811fa.png)

### Send Message/Whatsapp dialog

[...]

PR #19251, opened 2019-10-04
feat: [WIP] Whatsapp integration

fproldan:
### Whatsapp integration

Issue: #19196 
Feedback is appreciated

### Whatsapp Settings
Doctype to configure Twilio account

Captura de Pantalla 2019-10-04 a la(s) 14 17 03

### SMS Center
- New field message type.
Captura de Pantalla 2019-10-04 a la(s) 14 18 33

### Send SMS
- New field message type (default SMS).
- Attach document Print (only visible if message type is Whatsapp)
Captura de Pantalla 2019-10-04 a la(s) 14 18 54

### TODO

- [ ] Define a new name for SMS Center (for example Message Center), and apply the required modifications.

- [ ] Define a new name the Send SMS action (for example Send Message).

- [ ] Test definitions


mhbu50:
@fproldan Grate job!
I think this PR should be in frappe
It will be more useful for custome app which will not use Erpnext

sammish (dismissed):

fproldan:
We are planning to create a new notification channel to send notifications via WhatsApp. The idea is to move the WhatsApp settings (like they are in this pr) to Frappe and let the changes in SMS Center and Send Message button in ERPNext (using the Frappe's WhatsApp settings).
But, before we start to do it, we want to have some feedback from the Frappe team about what have done and what we are going to do. Maybe from one of you: @rohitwaghchaure @nabinhait @deepeshgarg007
The purpose of the feedback, besides the suggestion of changes and the contribution of new ideas, is to know if this has the possibility to be merged. Like, if you know that this is not going to be merged because of reasons you can tell us so we don't waste effort following the Frappe's coding rules.

Mangesh-Khairnar:
This feature would be added in Notification channel with twilio as its base. Closing this PR in ERPNext as it is going to be part of Frappe.

PR #19252, opened 2019-10-05
feat: Enhancement in landed cost voucher

deepeshgarg007:
image


olamide2:
@deepeshgarg007 , can we add parties to the additional charges item list? eg a vendor when an accounts payable account is chosen.

nabinhait:
> @deepeshgarg007 , can we add parties to the additional charges item list? eg a vendor when an accounts payable account is chosen.

@olamide2 Can you please give an example when we will select accounts payable as additional charges

olamide2:
@nabinhait  let me give a real life example that we encounter in my company

We buy some goods from China from our Chinese supplier.

These goods are cleared by a clearing agent who gives us 60 days payment terms

The landing cost of the goods is the cost from China plus the clearing fees, we use the landing cost voucher for these.

We will like to be able to add the cost of clearing to the account of the clearing agent by choosing accounts payable and the agents name in the other charges section. This will automatically update the clearing agents account and we can then make payment as contracted.

Regards

nabinhait:
> @nabinhait let me give a real life example that we encounter in my company
> 
> We buy some goods from China from our Chinese supplier.
> 
> These goods are cleared by a clearing agent who gives us 60 days payment terms
> 
> The landing cost of the goods is the cost from China plus the clearing fees, we use the landing cost voucher for these.
> 
> We will like to be able to add the cost of clearing to the account of the clearing agent by choosing accounts payable and the agents name in the other charges section. This will automatically update the clearing agents account and we can then make payment as contracted.
> 
> Regards

For the above example, you should add "Clearing charges" account in additional charges. And at the same time, you should create a Purchase Invoice where supplier is clearing agent and expense account is "Clearing Charges".

olamide2:
Thanks @nabinhait 

That is a viable work around, but it adds one more step to the process, this extra step of raising the PI could be avoided if the party option is available, just like we have in the journal entries.

nabinhait:
> Thanks @nabinhait
> 
> That is a viable work around, but it adds one more step to the process, this extra step of raising the PI could be avoided if the party option is available, just like we have in the journal entries.

I believe this is not a work-around, but appropriate way to do it. Because we will be able to track of invoices sent by clearing agents with other details like supplier-invoice-no etc.

erpnext/stock/doctype/landed_cost_voucher/landed_cost_voucher.js
```diff
@@ -30,6 +30,16 @@ erpnext.stock.LandedCostVoucher = erpnext.stock.StockController.extend({
 		this.frm.add_fetch("receipt_document", "posting_date", "posting_date");
 		this.frm.add_fetch("receipt_document", "base_grand_total", "grand_total");
 
+		this.frm.set_query("expense_account", "taxes", function() {
```
nabinhait, line 33: Use following for filtering accounts with account type = ["Tax", "Chargeable", "Income Account", "Expenses Included In Valuation", "Expenses Included In Asset Valuation"]

https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/queries.py#L131

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -68,6 +68,16 @@ frappe.ui.form.on('Stock Entry', {
 			}
 		});
 
+		frm.set_query("expense_account", "additional_costs", function() {
+			return {
+				filters: {
+					"root_type": "Expense",
+					"account_type": "Expenses Included In Valuation",
```
nabinhait: Use following for filtering accounts with account type = ["Tax", "Chargeable", "Income Account", "Expenses Included In Valuation", ""Expenses Included In Asset Valuation""]

https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/queries.py#L131

nabinhait (changes requested):

nabinhait (approved):

PR #19271, opened 2019-10-09
feat: duplicate linked task in project

Mangesh-Khairnar:
- Copies all the linked task in the project by adding a separate button for duplicate
Screenshot 2019-10-12 at 8 12 55 PM


erpnext/projects/doctype/project/project.js
```diff
@@ -80,6 +80,15 @@ frappe.ui.form.on("Project", {
 		frm.events.set_buttons(frm);
 	},
 
+	onload_post_render: function(frm) {
+		$(".menu-item-label[data-label='Duplicate']")
+			.parent('a')
+			.unbind("click")
+			.bind("click", function() {
+				frm.trigger('create_duplicate_project');
```
nabinhait: Function definition is `create_duplicate`?????

nabinhait (changes requested):
@Mangesh-Khairnar I don't think it should copy all tasks **implicitly** on click of **standard duplicate** button. If we want to copy tasks along with project then we should add another custom button "Duplicate Project with Tasks" button.

erpnext/projects/doctype/project/project.js
Mangesh-Khairnar:
```suggestion
				frm.trigger('create_duplicate');
```

Mangesh-Khairnar: Added changes

Mangesh-Khairnar:
@nabinhait Changed to a separate button "Duplicate Project with Tasks"

nabinhait (approved):

PR #19276, opened 2019-10-10
fix: Margin and Discount percentage set correctly via 'Update Items'

marination:
- Added appropriate check comparing price list rate with item rate 
- It will set margin and discount percentage correctly 

_SO


nabinhait (approved):

nabinhait (commented):
@marination Please send another PR for version-11-hotfix

PR #19294, opened 2019-10-13
fix: Typo in attendance date filter

### LINKED ISSUE #19290: Duplicate Attendance created with Leave Application on same date as Attendance
Steps:
1. Create attendance with status = Absent for 2019-09-03 and submit
2. Create Leave Application for 2019-09-03 and submit

Expected:
Status in Attendance created in Step 1 is updated to On Leave
Actual:
A new Attendance is created for 2019-09-03 with On Leave status

1. Attendance
image

2. Leave Application
image

3. Duplicate attendance record
image

Reason:
https://github.com/frappe/erpnext/blob/3f2c5c231972d350c2f4a7eee6440eed4557cc19/erpnext/hr/doctype/leave_application/leave_application.py#L128

The **typo** causes existing Attendance for the same date to be ignored and a new Attendance is created. The validation for new Attendance is ignored in line 147 below
https://github.com/frappe/erpnext/blob/3f2c5c231972d350c2f4a7eee6440eed4557cc19/erpnext/hr/doctype/leave_application/leave_application.py#L147

vijaywm:
fixes typo in filter parameter which creates duplicate Attendance

Closes #19290

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.py
+++ b/erpnext/hr/doctype/leave_application/leave_application.py
@@ -125,7 +125,7 @@ class LeaveApplication(Document):
 				status = "Half Day" if date == self.half_day_date else "On Leave"
 
 				attendance_name = frappe.db.exists('Attendance', dict(employee = self.employee,
-					attenance_date = date, docstatus = ('!=', 2)))
+					attendance_date = date, docstatus = ('!=', 2)))
 
 				if attendance_name:
 					# update existing attendance, change absent to on leave
--- a/erpnext/hr/doctype/leave_application/leave_application_dashboard.py
+++ b/erpnext/hr/doctype/leave_application/leave_application_dashboard.py
@@ -5,6 +5,12 @@ from frappe import _
 
 def get_data():
 	return {
+		'fieldname': 'leave_application',
+		'transactions': [
+			{
+				'items': ['Attendance']
+			}
+		],
         'reports': [
 			{
                 'label': _('Reports'),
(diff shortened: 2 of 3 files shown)


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -125,7 +125,7 @@ def update_attendance(self):
 				status = "Half Day" if date == self.half_day_date else "On Leave"
 
 				attendance_name = frappe.db.exists('Attendance', dict(employee = self.employee,
```
deepeshgarg007, line 127: @vijaywm Use frappe.db.get_value instead of frappe.db.exists since it just returns `True` or `False` and not the name of the record. `attendance_name` is used below to get the `Attendance` Doc and update it. Test case is also failing because of it. Please make the required changes

deepeshgarg007 (changes requested):

erpnext/hr/doctype/leave_application/leave_application.py
vijaywm: the name initially threw me off as well, so I checked. :) 
image
exists is returning None if not found else name of document  or list of document names
https://github.com/frappe/frappe/blob/df78b3647870261da86ad95bdf8a155204cf2e4d/frappe/database/database.py#L777

I checked Travis had failed because of some item_group test related error

deepeshgarg007: @vijaywm Following test case is also failing. Can you please check and fix that as well
image

deepeshgarg007 (approved):

PR #19301, opened 2019-10-14
feat(Stock Ledger): Batch Wise Incoming/Valuation Rate

SaiFi0102:
This spreadsheet should explain how Batch Wise Valuation rate would work.
The automated test case is also based on this sheet.

https://docs.google.com/spreadsheets/d/10GOG70ZuKPm4pjF8ZpvjniNHmo05i6Gs6jJedhaXKso/edit?usp=sharing

### Problem:
Valuation/Incoming Rate is stored for the whole (Item, Warehouse) Bin rather than (Item, Warehouse, Batch) combination.

Stock Value Difference / Cost of Goods Sold / Stock Adjustment amount is NOT calculated with respect to the selected Batchs in a Stock Out / Delivery Note

**Incoming (Same for all valuation methods)**
1. Stock In (Batch 1): Stock Value Difference: 10 * $100
2. Stock In (Batch 2): Stock Value Difference: 10 * $200

**Outgoing (FIFO)**
1. Stock Out (Batch **_2_**): Stock Value Difference (FIFO): -10 * $100
1. Stock Out (Batch **_1_**): Stock Value Difference (FIFO): -10 * $200

**Outgoing (Moving Average)**
1. Stock Out (Batch **_2_**): Stock Value Difference (M/A): -10 * $150
1. Stock Out (Batch **_1_**): Stock Value Difference (M/A): -10 * $150

**Outgoing (Batch Wise Valuation)**
1. Stock Out (Batch **_2_**): Stock Value Difference (BATCH WISE VAL): -10 * $200
1. Stock Out (Batch **_1_**): Stock Value Difference (BATCH WISE VAL): -10 * $100

### Note
Only Moving Average within (Item, Warehouse, Batch) combination is implemented and I don't see much use for FIFO within (Item, Warehouse, Batch) since in a Batched item you'd most probably create a new batch for each new Purchase/Stock In.

### Backwards Compatibility:
The user MUST enable Batch Wise Valuation in either Item Master or Stock Setting so that stock ledger stores batch-wise valuation rate separately. Therefore any old item and their stock ledger will have **0**/**no** effect since their Batch Wise Valuation will not be enabled and it will follow the same validation that if there is any stock ledger entry then you cannot enable/disable batch wise valuation.

### Use Cases
* Calculating Consignment Sale Gross Profitability
* Calculating accurate Gross Profit of a single Batch
* Calculating accurate Gross Profit of a single Purchase (sum of GP of all Batches in Purchase)
* Quenching your thirst for accuracy

### Screenshot of Batch Wise Valuation Explanation Spreadsheet
image


erpnext/stock/stock_ledger.py
```diff
@@ -241,46 +258,64 @@ def get_serialized_values(self, sle):
...
 				if not allow_zero_valuation_rate:
-					self.valuation_rate = get_valuation_rate(sle.item_code, sle.warehouse,
-						sle.voucher_type, sle.voucher_no, self.allow_zero_rate,
-						currency=erpnext.get_company_currency(sle.company))
+					new_valuation_rate = get_valuation_rate(sle.item_code, sle.warehouse,
+						sle.voucher_type, sle.voucher_no, sle.batch_no, self.allow_zero_rate,
+						currency=erpnext.get_company_currency(sle.company), batch_wise_valuation=self.batch_wise_valuation)
+
+		self.qty_after_transaction += flt(sle.actual_qty)
+		if self.batch_wise_valuation:
```
rmehta, line 307: move this to a smaller function?

erpnext/stock/stock_ledger.py
```diff
@@ -453,20 +516,52 @@ def get_stock_ledger_entries(previous_sle, operator=None,
...
-			AND warehouse = %s
-			AND valuation_rate >= 0
-			AND NOT (voucher_no = %s AND voucher_type = %s)
-		order by posting_date desc, posting_time desc, name desc limit 1""", (item_code, warehouse, voucher_no, voucher_type))
+	last_valuation_rate = None
+
+	if batch_wise_valuation is None:
+		valuation_method, batch_wise_valuation = get_valuation_method(item_code)
+
+	if batch_no and batch_wise_valuation:
```
rmehta, line 530: move this to a smaller function

rmehta (changes requested):
Please update the documentation too!

erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -73,6 +73,174 @@ def test_fifo(self):
...
+		in1_1 = make_stock_entry(item_code=item_code, target=warehouse, qty=10, basic_rate=100)
+		batch1 = in1_1.items[0].batch_no
+		assert_sle(get_sle(item_code=item_code, warehouse=warehouse, batch_no=batch1,
+			voucher_type=in1_1.doctype, voucher_no=in1_1.name)[0], {
+			'stock_value_difference': 1000, 'batch_stock_value': 1000, 'stock_value': 1000,
+			'batch_valuation_rate': 100, 'valuation_rate': 100,
+			'batch_qty_after_transaction': 10, 'qty_after_transaction': 10
+		})
+
+		in1_2 = make_stock_entry(item_code=item_code, target=warehouse, qty=10, basic_rate=80)
```
rmehta, line 99: break these into separate tests? also please add comments for each of these tests. So that reviewing is easy.

rmehta (changes requested):

erpnext/stock/doctype/stock_entry/test_stock_entry.py
SaiFi0102: Can't be broken, they're all dependent on each other. Comments I can do

erpnext/stock/stock_ledger.py
SaiFi0102: Are you talking about this condition: `if batch_no and batch_wise_valuation` or the query?

erpnext/stock/stock_ledger.py
SaiFi0102: The calculation? The calculation is very specific to this function. It can be done, but I feel it'll make it more confusing since I'm already using `self.batch_data` in the beginning of the method.

erpnext/stock/stock_ledger.py
SaiFi0102: Alright

erpnext/stock/utils.py
```diff
@@ -188,10 +191,23 @@ def get_avg_purchase_rate(serial_nos):
...
-	val_method = frappe.db.get_value('Item', item_code, 'valuation_method')
+	val_method, batch_wise_valuation, has_batch_no, has_serial_no = frappe.db.get_value('Item', item_code,
+		['valuation_method', 'batch_wise_valuation', 'has_batch_no', 'has_serial_no'])
+
 	if not val_method:
 		val_method = frappe.db.get_value("Stock Settings", None, "valuation_method") or "FIFO"
-	return val_method
+
+	if not has_batch_no or has_serial_no:
+		batch_wise_valuation = 0
```
marination, line 201: if it does have batch no and not serial no, batch_wise_valuation must  be 1 right?

SaiFi0102: If it does have batch no and does not have serial number than it can either use batch wise valuation (if applied for that item or globally) or the normal valuation method (for backwards compatibility)

If there is a serial no then it shouldn't use batch wise valuation but rather serial no wise valuation

SaiFi0102: Use batch wise valuation only when has batch no, does not have serial no and has batch wise valuation setting

erpnext/stock/doctype/stock_entry/test_stock_entry.py
SaiFi0102: The test is quite self explanatory. For a detailed understanding only the spreadsheet at https://docs.google.com/spreadsheets/d/10GOG70ZuKPm4pjF8ZpvjniNHmo05i6Gs6jJedhaXKso/edit?usp=sharing will help. Added a link in comments

erpnext/stock/stock_ledger.py
SaiFi0102: You will have to elaborate on this. This is not as simple as: move to new function. This would require some minor refactoring.

guilmori:
I need this feature, why it was closed without being merged ?

guilmori:
@SaiFi0102 Hello Saif, thanks for this great work. Are you using it in production ?

SaiFi0102:
> @SaiFi0102 Hello Saif, thanks for this great work. Are you using it in production ?

Yes, in 2 companies where traceability and costing of individual batches are very important

guilmori:
> > @SaiFi0102 Hello Saif, thanks for this great work. Are you using it in production ?
> 
> Yes, in 2 companies where traceability and costing of individual batches are very important

Awesome, I have the same requirement.

I guess this is pretty core change, so we cannot do this in a custom app right ?
You had to fork the erpnext branch, and handle all sort of merge conflicts when upgrading erpnext version ?

It is a shame this isn't part of the product, seems to me to be quite mandatory when working with batch items...

SaiFi0102:
> > > @SaiFi0102 Hello Saif, thanks for this great work. Are you using it in production ?
> > 
> > 
> > Yes, in 2 companies where traceability and costing of individual batches are very important
> 
> Awesome, I have the same requirement.
> 
> I guess this is pretty core change, so we cannot do this in a custom app right ?
> You had to fork the erpnext branch, and handle all sort of merge conflicts when upgrading erpnext version ?
> 
> It is a shame this isn't part of the product, seems to me to be quite mandatory when working with batch items...

No a custom app is not possible since it requires changes in the deeper parts of the core

Yes I have to deal with merge conflicts whenever I pull from a new version

TurkerTunali:
This is merged in v14. https://github.com/frappe/erpnext/pull/29804

Issue #8605, opened 2017-04-27
Avatar next to login name missing


lasalesi:
Affects ERPNExt v8.0.19 (master)
Sometimes (not always), the user avatar left of the login is not shown.
Affects Firefox but also Edge and IE.
The avatar is visible in personal settings and the file is public. 
Bench restart, reload, new upload of the file do not help.

netchampfaris:
Screenshot?

https://github.com/frappe/erpnext/wiki/Issue-Guidelines

lasalesi:
![grafik](https://cloud.githubusercontent.com/assets/26505250/26028839/5b4203fc-3828-11e7-820c-3016c50b3b7d.png)

MCD-50:
Unable to replicate the issue
Tried on Firefox, Edge and Chrome.

![screenshot from 2017-05-15 17-24-31](https://cloud.githubusercontent.com/assets/22857645/26056732/38192bd6-3994-11e7-85f1-0a7b183f4675.png)

MCD-50:
Please open again if any you face this problem again.

lasalesi:
Steps to reproduce: 
* use an avatar with whitespace characters.

Analysis:
* background-image: url(/files/ninja minion 48px.png)
* Invalid file name
* Should be like this: background-image: url('/files/ninja minion 48px.png')
* If this is manually corrected in the browser, it works

Location in html:
<span class="avatar-frame" style="background-image: url('/files/ninja minion 48px.png');" title="Max Power"></span>

MCD-50:
Replicating.
Thanks for giving steps.
Sending a PR

manassolanki:
Pushed a fix via 
https://github.com/frappe/frappe/pull/3386

PR #19314, opened 2019-10-15
Erpnext calendar map

Anurag810:
Don't merge without 
#8605


Anurag810:
will open when it is completed.

PR #19326, opened 2019-10-16
refactor: Expense Claim Cost Center

jiloysss:
* There is a cost center field added to the table (Expense Claim Detail)
* Changed Script in Expense Claim (python) where the cost center in GL Entry is based on the cost center field on each row (Expense Claim Detail)
* Filtered Cost Center in Expense Claim Detail where it will only fetch child account
* Hid Cost Center in Expense Claim
* Added Deferred Expense Accounts checkbox in Expense Claim Type


jiloysss:
EXPENSE CLAIM
Added field Cost Center in table

expense_claim

EXPENSE CLAIM TYPE
Added check box Deferred Expense Accounts
If checked, Default Account in table will be filtered with Asset Accounts. Expense Accounts if otherwise

Expense_claim_type

sammish (approved):

erpnext/hr/doctype/expense_claim/expense_claim.js
```diff
@@ -402,3 +403,10 @@ cur_frm.fields_dict['task'].get_query = function(doc) {
 		}
 	};
 };
+cur_frm.fields_dict["expenses"].grid.get_field("cost_center").get_query = function(){
+	return {
+		filters:{
+			"is_group": "child"
```
sammish, line 409: "is_group": 0

erpnext/hr/doctype/expense_claim/expense_claim.js
```diff
@@ -61,6 +61,7 @@ cur_frm.cscript.clear_sanctioned = function(doc) {
 };
 
 cur_frm.cscript.refresh = function(doc) {
+	cur_frm.set_df_property("cost_center", "hidden", true);
```
nabinhait, line 64: - Don't hide cost_center field, remove it entirely. And write a patch to update child cost centers based on parent.

- Copy cost center from first row on addition of new row. And on entering cost_center in any row, it should get copied to all rows with blank cost_center.

Refer following code for example:
https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order/purchase_order.js#L458
https://github.com/frappe/erpnext/blob/develop/erpnext/buying/doctype/purchase_order/purchase_order.js#L556

nabinhait (changes requested):

erpnext/hr/doctype/expense_claim/expense_claim.js
sammish: @nabinhait It has updated kindly check  #19663

sammish: @nabinhait It has updated again #19663 

Please check and verify.

PR #19339, opened 2019-10-17
fix: Currency Exchange for_selling and for_buying on the same day

toofun666:

In this fix:
* You can now add a separate exchange_rate in date for_selling and for_buying. You could not before because the unique field name was only calculated to allow a single name for a date
* tests did not account for for_selling and for_buying fields and thier uniqueness


toofun666:
Dears,
this update will help the vanilla systems to function better in Currency Exchange. The test has also been updated to cover the full scenario where for_selling and for_buying are required fields but were not covered in the test.

toofun666:
Let us get this going to put the design back into place with proper testing on Currency Exchange.
The problem was:
* for_selling or for_buying was a must
* the naming was based on parameters excluding these fields
* thus you could not enter a buying and selling currency exchange rate for the same day because of unique naming construction

nabinhait (approved):

PR #19361, opened 2019-10-20
refactor: loan repayment Schedule

sammish:
This feature help HR team to change Loan repayment Schedule if required.
I am requesting to merge both bench (Bench-11 and Bench-12 )
Peek 2019-10-20 09-03


santoshbb (approved):

erpnext/hr/doctype/leave_type/leave_type.json
```diff
@@ -102,19 +100,6 @@
    "fieldtype": "Check",
    "label": "Is Compensatory"
   },
-  {
-   "collapsible": 1,
-   "depends_on": "eval: doc.is_carry_forward == 1",
```
deepeshgarg007, line 107: @sammish can you explain why these fields are removed?

erpnext/hr/doctype/repayment_schedule/repayment_schedule.json
```diff
@@ -1,231 +1,82 @@
...
-   "read_only": 0, 
-   "remember_last_selected_value": 0, 
-   "report_hide": 0, 
-   "reqd": 0, 
-   "search_index": 0, 
-   "set_only_once": 0, 
-   "translatable": 0, 
-   "unique": 0
-  }, 
+   "allow_on_submit": 1,
```
deepeshgarg007, line 16: Can you give a use case why do you need to change the payment date after submit?

deepeshgarg007 (changes requested):

erpnext/hr/doctype/repayment_schedule/repayment_schedule.json
sammish: If the employee is in danger and cannot work for one or two months and the loan is pending that month, the HR manager can extend the loan for two months.
I created new PR https://github.com/frappe/erpnext/pull/19470

PR #19388, opened 2019-10-22
fix: Sales and Purchase Invoice Status and Payment Reconciliation

marination:
1) **Unpaid Sales Invoice** Status showed **'Paid'** after Credit note was made against it 
     - Fix: Status is now **'Credit Note Issued'**

2) **Payment Reconciliation** will only show standalone **Debit/Credit Notes** in 'UNRECONCILED PAYMENT DETAILS' section or **Payments** Table

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -1231,7 +1231,8 @@ class SalesInvoice(SellingController):
 					self.status = "Unpaid and Discounted"
 				elif flt(self.outstanding_amount) > 0 and getdate(self.due_date) >= getdate(nowdate()):
 					self.status = "Unpaid"
-				elif flt(self.outstanding_amount) < 0 and self.is_return==0 and frappe.db.get_value('Sales Invoice', {'is_return': 1, 'return_against': self.name, 'docstatus': 1}):
+				#Check if outstanding amount is 0 due to credit note issued against invoice
+				elif flt(self.outstanding_amount) <= 0 and self.is_return == 0 and frappe.db.get_value('Sales Invoice', {'is_return': 1, 'return_against': self.name, 'docstatus': 1}):
 					self.status = "Credit Note Issued"
 				elif self.is_return == 1:
 					self.status = "Return"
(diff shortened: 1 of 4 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, *args, **kwargs):
 
 	def set_indicator(self):
 		"""Set indicator for portal"""
-		if self.outstanding_amount < 0:
+		if self.outstanding_amount < 0 or (self.paid_amount == 0 and self.outstanding_amount == 0):
```
Anurag810: please write comment for this check. ``` (self.paid_amount == 0 and self.outstanding_amount == 0) ```

Anurag810 (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
marination: theres more changes required, on it

erpnext/controllers/status_updater.py
```diff
@@ -49,7 +49,8 @@ def validate_status(status, options):
 		["Submitted", "eval:self.docstatus==1"],
 		["Paid", "eval:self.outstanding_amount==0 and self.docstatus==1"],
 		["Return", "eval:self.is_return==1 and self.docstatus==1"],
-		["Debit Note Issued", "eval:self.outstanding_amount < 0 and self.docstatus==1"],
+		["Debit Note Issued",
+		 "eval:self.outstanding_amount <= 0 and self.docstatus==1 and self.is_return==0 and get_value('Purchase Invoice', {'is_return': 1, 'return_against': self.name, 'docstatus': 1})"],
```
nabinhait, line 53: frappe.db.get_value ?

nabinhait (changes requested):

erpnext/controllers/status_updater.py
marination: Screenshot 2019-11-01 at 12 00 08 PM

nabinhait (approved):

PR #19390, opened 2019-10-23
fix: setting incorrect field for party bank account

0Pranav:
Previously
wrong

Corrected
corrected

In payment entry if the customer had a default bank account set it would be fetched into company bank account rather than party bank account.


deepeshgarg007:
@0Pranav Please raise a PR for v11 as well

deepeshgarg007 (approved):

PR #19391, opened 2019-10-23
fix: setting incorrect field for party bank account

0Pranav:
Same as https://github.com/frappe/erpnext/pull/19390 for v11

### DIFF
--- a/erpnext/__init__.py
+++ b/erpnext/__init__.py
@@ -5,7 +5,7 @@ import frappe
 from erpnext.hooks import regional_overrides
 from frappe.utils import getdate
 
-__version__ = '11.1.66'
+__version__ = '11.1.67'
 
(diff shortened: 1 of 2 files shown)


deepeshgarg007:
@0Pranav target branch should be version-11-hotfix. From there the changes will be ported to version-11. Please change

erpnext/__init__.py
```diff
@@ -5,7 +5,7 @@
 from erpnext.hooks import regional_overrides
 from frappe.utils import getdate
 
-__version__ = '11.1.66'
+__version__ = '11.1.67'
```
deepeshgarg007, line 8: @0Pranav Please remove this change

deepeshgarg007 (changes requested):

PR #19393, opened 2019-10-23
Translate Korean language for sample by native speaker

yunjy-github:
Hi, I want to contribute to erpnext by translating Korean language.
The original version of ko.csv is also... good but... weird to native Korean users. 
Here is my sample translation.
Please let me know your opinion, before the translation work starts.
Thanks!


nabinhait:
@yunjy-github This is a great initiative. But as we don't know Korean language, it will be hard to review the translations. Can you please make a post on discuss.erpnext.com making a proposal and to find any other Korean user who can review the translations?

Closing the PR for now.

PR #19396, opened 2019-10-23
Enable method/function overload as general functionality for all functions

jau2017:
There should be an additional parameter in the User-Doctype „include overload“ which will be show and can be overwritten during login-phase. This way is a chance to activate/deactivate all overloads at once to go back to a „standard ERPNext“ to let any user do any functional test of new ERPNext-functions.
When the option „function overload“ will be activated a general modification module will be loaded and to all included functions the additional parameter (overload) will be passed. By this way the modification module will NEVER be changed by ERPNext and can stil remain  unchanged during every update. Inside the included functions there will be checked if the „include overload switch“ is set acting depending on this.
By this way all modifications can be encapsulated in an separate shell – so no modification inside the ERPNext-standard-programs will be required.
The case of credit limit check is only an example - the same logic will be for any interface for manufactoring or any other interface or modification.

Please imagine the following situation : you are using another financial system than ERPNext. You have to check the credit limit in case of an new order or invoice via this system.
Right now the appropiate function or hook has to be modified within ERPNext. This modification has to be located in the path /home/frappe/frappe-bench/erpnext/erpnext/hooks.py or/home/frappe/frappe-bench/erpnext/erpnext/MODULE/PROGRAM.py.  
1.  If there is any Bug in the result you will not be able (driven by the user) to different if this bug is inside ERPnext or in the modification
2.  When an update will appear you have to modify the new Version of the Hook and/or the module/function – even if there is no change in the customers business logic or in the used functions.

This Situation can be optimized by this function overload.


adityahase (changes requested):
You seem to be creating a PR from frappe:develop to frappe:version-12

Issue #19275, opened 2019-10-09
Shopping cart allows orders with quantities higher than in stock


conncampbell:
V12

Steps to reproduce:

1. Ensure "Allow items not in stock to be added to cart" is disabled in shopping cart settings.
2. Ensure item has at least 1 item in stock
3. add item to cart
4. change quantity to a value higher than what is actuall in stock
5. place order without error.

issues is caused by the following in cart.place_order:

This code gets the "Allow items not in stock to be added to cart" value, but returns it as a string.
cart_settings = frappe.db.get_value("Shopping Cart Settings", None,
		["company", "allow_items_not_in_stock"], as_dict=1)

later, this code tries to test for it:
if not cart_settings.allow_items_not_in_stock:

it should be:
if not int(cart_settings.allow_items_not_in_stock):

clarkejj:
Notice this commit, good call thanks https://github.com/frappe/erpnext/commit/e1f95c0529f0109f7cc70674b4bab75971436550#diff-6c2bea20a04f484ee53ab2c243eef0c7

PR #19398, opened 2019-10-24
fix: allow_items_not_in_stock should be evaluated

ashish-greycube:
**Technical:**
allow_items_not_in_stock is a checked field type and to evaluate it in IF condition **cint** should be used.

**Functional:**
Actual : In Shopping cart settings, 'Allow items not in stock to be added to cart' is checked.
Select items in cart having stock zero. From cart, place order. No validation is fired

Expected: When order is placed system should give below mesage for items not in stock.
throw(_("Only {0} in stock for item {1}")


nabinhait (approved):

clarkejj:
Thanks notice conncampbell's report https://github.com/frappe/erpnext/issues/19275

PR #19405, opened 2019-10-24
fix: Overwrite default cost center if item or item group has default cost center set.

surajshetty3416:
While creating a sales invoice from the sales order default cost center was getting fetched (in sales order item) instead of fetching it from item default. 

This fixes that issue.


Anurag810 (approved):

dhav-d:
the same issue occurs when invoices generated via POS

dhav-d:
when will this update be available on hosted accounts or which version build would/should have the fix?

PR #19407, opened 2019-10-24
fix(bom): maintain a default bom for an item

Mangesh-Khairnar:
**Issue:** In work order items to be manufactured need a default bom associated with it, currently it is possible to have multiple active boms but no default bom. Items with no default bom are not fetched in the work order. This allows an easy way to identify the items which have a bom associated with it.

**Fix:** This PR proposes a fix to maintain a default bom at all times for any item with active bom. This ensures that the items with active boms are fetched in work order

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -290,7 +290,8 @@ class BOM(WebsiteGenerator):
 		return valuation_rate
 
 	def manage_default_bom(self):
-		""" Uncheck others if current one is selected as default,
+		""" Uncheck others if current one is selected as default or
+			check the current one as default if it the only bom for the selected item,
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

djpalshikar:
@Mangesh-Khairnar this is causing us serious issues, request you to find a solution to the problem:
Why we need it:
We have two production depts. Electronics and Mechanical.
Electronics produces as per Stock requirements. So in this case say ITEM A is available in stock and need not be manufactured.

Where as ITEM A is a raw material in several thousand BOMs for our products.
All those final products are exploded completely as they are made to order.
Here there are certain common sub-assemblies which are also exploded, which is why we explode.

So in conclusion we have a situation where we need a BOM which has some items exploded and others which are not exploded.

PR #19422, opened 2019-10-28
fix: Search field entries included in Item Link field query

marination:
- Entries in **'Search Fields'** field were not considered while querying **Item Link field** in child tables
- Included Search field entries in search condition as well as in fields to fetch and display in **Item** Link field suggestion list.


dufani1:
Hi... @marination 
please note that this issue in v11 and 12

marination:
@dufani1  The fix has been sent on v11 as well

deepeshgarg007 (approved):

dufani1:
@marination thanks

PR #19423, opened 2019-10-28
fix(patch): Missing Patch for item Tax template

Anurag810:


erpnext/patches.txt
```diff
@@ -641,3 +641,4 @@ erpnext.patches.v12_0.set_produced_qty_field_in_sales_order_for_work_order
 erpnext.patches.v12_0.generate_leave_ledger_entries
 erpnext.patches.v12_0.set_default_shopify_app_type
 erpnext.patches.v12_0.replace_accounting_with_accounts_in_home_settings
+execute:frappe.db.set_value("Accounts Settings", None, "add_taxes_from_item_tax_template", 1) #setting default tax template enable
```
nabinhait, line 644: As you are using `set_value`, it will not call `validate` of Accounts Settings and will not set default. Also do `frappe.db.set_default("add_taxes_from_item_tax_template", 1)` via patch.

nabinhait (changes requested):

nabinhait (commented):
Merged via local, closing this PR.

PR #19427, opened 2019-10-28
fix: Expense claim paid through employee advance getting fetched as outstanding in Payment Entry

deepeshgarg007:
Expense Claim


erpnext/accounts/utils.py
```diff
@@ -688,12 +688,34 @@ def get_outstanding_invoices(party_type, party, account, condition=None, filters
 		"account": account
 	}, as_dict=True)
 
+	# Get GL entries for expense claim that are paid or allocated by employee advance
+	if party_type == "Employee":
+		advance_allocated = frappe.db.sql("""
```
nabinhait: This is not correct solution. The actual solution should be when we submit Expense Claim with Employee Advance entries, system should make 2 GL entries to update `against_voucher = "Expense Claim"` against payment entries.

nabinhait (changes requested):

nabinhait (approved):

PR #19431, opened 2019-10-29
refactor: report BOM Stock calculated 

Anurag810:
Refactored report BOM Stock calculated, also handled use case for multiple manufacturers and part nos in the item master.

### DIFF
--- a/erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
+++ b/erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
@@ -4,6 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 from frappe import _
+from frappe.utils.data import comma_and
 
 def execute(filters=None):
 #	if not filters: filters = {}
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
```diff
@@ -15,24 +15,39 @@ def execute(filters=None):
 
 	for row in data:
 		item_map = get_item_details(row.item_code)
+		manufacture_details = get_manufacturer_records(row.item_code)
```
nabinhait: it should be called outside for loop. Even `get_item_details` should be called outside.

erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
```diff
@@ -15,24 +15,39 @@ def execute(filters=None):
...
+	diff_qty = to_build - reqd_qty
+	return [row.item_code, row.description, manufacturers, part_no, row.actual_qty, str(to_build), reqd_qty, diff_qty, last_pur_price]
+
+
+
+def get_manufacturers_details(manufacture_details):
+	manufacturers = ""
+	part_no = ""
+	for details in manufacture_details:
+		manufacturers += details.get('manufacturer')+ ', '
```
nabinhait: it will always add "," at the end.

nabinhait (changes requested):

erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
Anurag810: done

erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
Anurag810: done

Anurag810:
@nabinhait please merge, I have done changes requested by you.

erpnext/manufacturing/report/bom_stock_calculated/bom_stock_calculated.py
```diff
@@ -13,26 +14,32 @@ def execute(filters=None):
...
-			diff_qty = 0 - reqd_qty
-			summ_data.append([row.item_code, row.description, item_map[row.item_code]["manufacturer"], item_map[row.item_code]["manufacturer_part_no"], row.actual_qty, "0.000", reqd_qty, diff_qty, last_pur_price])
 
+		summ_data.append(get_report_data(last_pur_price, reqd_qty, row, manufacture_details))
 	return columns, summ_data
 
+def get_report_data(last_pur_price, reqd_qty, row, manufacture_details):
+	manufacturer_arr = []
+	manufacturer_part_no_arr = []
+	for detail in manufacture_details:
```
nabinhait: Make a dict like `{item1: {'manufacturer_list': ['m1', 'm2'], 'manufacturer_part_numbers': ['12', '232']}}` in `get_manufacturer_records` function, so that we don't need iterate over this for loop for every single item.

nabinhait (changes requested):

nabinhait (approved):

PR #19432, opened 2019-10-29
fix: patch, item tax template showing 'Untitled' in the name

rohitwaghchaure:
**Issue**

Screenshot 2019-10-29 at 2 56 20 PM

**After Fix**

Screenshot 2019-10-29 at 3 39 02 PM


harshit-30:
@rohitwaghchaure how is it going to appear now, attach the screenshot na!

PR #19442, opened 2019-10-30
fix: maintenance schedule ux

0Pranav:
In maintenance schedule form

- Hide schedule section if the document is not saved.

- More descriptive error if sales person is not selected for an item
Screenshot 2019-10-30 at 10 58 18 AM


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -150,7 +150,7 @@ def validate_maintenance_detail(self):
 			elif not d.no_of_visits:
 				throw(_("Please mention no of visits required"))
 			elif not d.sales_person:
-				throw(_("Please select Incharge Person's name"))
+				throw(_("Please select Incharge Person's name for item: {0}".format(d.item_name)))
```
Thunderbottom: I think a better wording would be:

```suggestion
				throw(_("Please select an Incharge Person for item: {0}".format(d.item_name)))
```

Thunderbottom (changes requested):

Thunderbottom (approved):

gavindsouza (approved):

gavindsouza (approved):

PR #19459, opened 2019-10-30
fix(patch): update status and set auto repeat as not submittable

ruchamahabal:
Depends on: https://github.com/frappe/frappe/pull/8721

- **Problem**: Auto Repeat Permission and docstatus issue after upgrading to v12
auto_repeat_submit_perm_issue
**Fix**: Update DocPerm and docstatus for older auto repeat as Auto Repeat is not submittable in v12.

- **Problem**: Old Auto Repeat not working after upgrade to v12 (due to different statuses)
Statuses in v11: Active, Stopped, Cancelled, Draft, Expired, Disabled
Statuses in v12: Active, Disabled, Completed
**Fix**: Set Draft, Cancelled and Stopped as Disabled, Expired set to Completed on save and remaining set to Active

**In v11:**
auto-repeat-v12-issues

**After upgrading to v12:**
auto_repeat_after_upgrade_to_v12

**After patch:**
after_patch

### DIFF
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -641,3 +641,4 @@ erpnext.patches.v12_0.set_produced_qty_field_in_sales_order_for_work_order
 erpnext.patches.v12_0.generate_leave_ledger_entries
 erpnext.patches.v12_0.set_default_shopify_app_type
 erpnext.patches.v12_0.replace_accounting_with_accounts_in_home_settings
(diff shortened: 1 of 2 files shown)


netchampfaris:
@ruchamahabal Why is this patch in `erpnext`? Shouldn't it be in `frappe`?

PR #19461, opened 2019-10-30
fix: Bank Reconciliation Allows to Over Reconcile

Mitchy25:
Branch to Merge: Version -12

When using Bank Reconciliation, the system does not stop the user from 'Over-reconciling' a Payment Entry against a Bank Transaction.

This fix will throw an error when the user tries to reconcile a Payment Entry which has an unallocated_amount greater than the Bank Transaction's unallocated_amount.


nextchamp-saqib (approved):

erpnext/accounts/page/bank_reconciliation/bank_reconciliation.py
```diff
@@ -18,6 +18,10 @@ def reconcile(bank_transaction, payment_doctype, payment_name):
 	account = frappe.db.get_value("Bank Account", transaction.bank_account, "account")
 	gl_entry = frappe.get_doc("GL Entry", dict(account=account, voucher_type=payment_doctype, voucher_no=payment_name))
 
+	if payment_entry.unallocated_amount > transaction.unallocated_amount:
```
nextchamp-saqib: only check this condition if `payment_doctype ` is "Payment Entry" as other doctypes won't have `unallocated_amount`

nextchamp-saqib (changes requested):

PR #19464, opened 2019-10-31
minor: added description, uom fields in material request plan item table

rohitwaghchaure:
Screenshot 2019-10-31 at 2 12 28 PM


nabinhait (changes requested):
UOM field should be placed after Quantity field

rohitwaghchaure:
@nabinhait  Fixed 

Screenshot 2019-11-08 at 3 16 28 PM

nabinhait (approved):

PR #19479, opened 2019-11-01
fix: incorrect balance qty in stock ledger if batch filter set

rohitwaghchaure:
**Issue**
Screenshot 2019-11-01 at 5 22 39 PM

**After Fix**

image

### DIFF
--- a/erpnext/stock/report/stock_ledger/stock_ledger.py
+++ b/erpnext/stock/report/stock_ledger/stock_ledger.py
@@ -19,10 +19,26 @@ def execute(filters=None):
 	if opening_row:
 		data.append(opening_row)
 
+	actual_qty = stock_value = 0
+
 	for sle in sl_entries:
 		item_detail = item_details[sle.item_code]
 
 		sle.update(item_detail)
+
+		if filters.get("batch_no"):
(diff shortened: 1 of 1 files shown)


erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -19,10 +19,22 @@ def execute(filters=None):
...
+
 	for sle in sl_entries:
 		item_detail = item_details[sle.item_code]
 
 		sle.update(item_detail)
+
+		if filters.get("batch_no"):
+			actual_qty += sle.actual_qty
+			stock_value += sle.stock_value_difference
+
```
nabinhait, line 32:
```
if sle.voucher_type == 'Stock Reconciliation':
     actual_qty = sle.qty_after_transaction
     stock_value = sle.stock_value
```

nabinhait (changes requested):

erpnext/stock/report/stock_ledger/stock_ledger.py
rohitwaghchaure: Done

nabinhait (approved):

PR #19482, opened 2019-11-01
fix: Wrong datafield type in Task

MaxMorais:
The field `depends_on_tasks` is set as `Data`, but it stores a list of strings joined by comma, so with more than 5 tasks, we reach the field size limit of 144 chars and data get trunked, what make us lose data!

> Screenshots/GIFs


MaxMorais:
An more detail can be provided here

deepeshgarg007 (approved):

PR #19492, opened 2019-11-04
fix: invoice creation tool of invoices with no paid and outstanding amounts

gavindsouza:
### Error:
imageb9fe68

**Environment:** Python 3, ERPNext v12

---
### Code

```python
max([1, 34, None, 23])
```
---
### Output

**py2:** `34`

**py3:** `TypeError: '>' not supported between instances of 'NoneType' and 'int'`

### DIFF
--- a/erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
+++ b/erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
@@ -32,8 +32,10 @@ class OpeningInvoiceCreationTool(Document):
 				})
 				invoices_summary.update({company: _summary})
 
-				paid_amount.append(invoice.paid_amount)
-				outstanding_amount.append(invoice.outstanding_amount)
+				if invoice.paid_amount:
+					paid_amount.append(invoice.paid_amount)
+				if invoice.outstanding_amount:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
```diff
@@ -32,8 +32,10 @@ def prepare_invoice_summary(doctype, invoices):
 				})
 				invoices_summary.update({company: _summary})
 
-				paid_amount.append(invoice.paid_amount)
-				outstanding_amount.append(invoice.outstanding_amount)
+				if invoice.paid_amount:
```
deepeshgarg007, line 35: passing invoice.paid_amount and invoice.outstanding_amount inside flt while appending should solve this issue.

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
gavindsouza: Multiple ways to solve a problem! This one just seems a much more cheaper way to do it though

deepeshgarg007 (approved):

PR #19497, opened 2019-11-04
fix: allow Pending on review date

RJPvT:
In earlier build review date was added but not a check. It is not logical to overdue a task which is pending without checking the review date.

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -201,9 +201,9 @@ def set_multiple_status(names, status):
 def set_tasks_as_overdue():
 	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
 	for task in tasks:
-                if frappe.db.get_value("Task", task.name, "status") in 'Pending Approval' and not (getdate(frappe.db.get_value("Task", task.name, "review_date")) < getdate(today())):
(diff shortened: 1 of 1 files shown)


RJPvT:
bump?

erpnext/projects/doctype/task/task.py
```diff
@@ -201,6 +201,9 @@ def set_multiple_status(names, status):
 def set_tasks_as_overdue():
 	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
 	for task in tasks:
+		if frappe.db.get_value("Task", task.name, "status") in 'Pending Approval':
```
nabinhait, line 204: 'Pending Approval' is not a valid status of task

nabinhait (commented):
And even if the project is not reviewed, if we did not meet expected end date, it should be overdue. If it's not reviewed, then reset expected end date.

RJPvT:
> And even if the project is not reviewed, if we did not meet expected end date, it should be overdue. If it's not reviewed, then reset expected end date.

ADDITION: This makes no sense. is expected end date is including review then review date (and closing date) make no sense. a task should be done bij exp_end_date for review, reviewed by review date and closing date should be set by the actual closing date. 

This is handeled because only reason setting overdue will be skipped (for loop will continue) is status is Pending Review AND date is within the review date ?

If i change the status will de PR be ok ?

erpnext/projects/doctype/task/task.py
RJPvT: I will change

PR #19498, opened 2019-11-04
feat: Provision to fetch items from multiple BOMs in Stock Entry

marination:
- Currently the **'From BOM'** checkbox in Stock Entry allows only one BOM per Stock Entry
- Added **'Bill of Materials'** option in **'Get Items From'**
- Provision to fetch items from multiple BOMs
- Fields in popup toggle according to Purpose
- Working:

- Accesibiity:
  - It will be visible in draft condition depending on the Purpose 

_stock_entry_bom

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -238,6 +238,8 @@ frappe.ui.form.on('Stock Entry', {
 			}, __("Get items from"));
 		}
 
+		frm.events.show_bom_custom_button(frm);
+
 		if (frm.doc.company) {
 			frm.trigger("toggle_display_account_head");
 		}
@@ -253,13 +255,7 @@ frappe.ui.form.on('Stock Entry', {
 
 	stock_entry_type: function(frm){
 		frm.remove_custom_button('Bill of Materials', "Get items from");
-
-		if (frm.doc.docstatus === 0 &&
-			['Material Issue', 'Material Receipt', 'Material Transfer', 'Send to Subcontractor'].includes(frm.doc.purpose)) {
-				frm.add_custom_button(__('Bill of Materials'), function() {
-					frm.events.get_items_from_bom(frm);
-				}, __("Get items from"));
-		}
+		frm.events.show_bom_custom_button(frm);
 	},
 
 	purpose: function(frm) {
@@ -398,6 +394,15 @@ frappe.ui.form.on('Stock Entry', {
 		}
 	},
 
+	show_bom_custom_button: function(frm){
+		if (frm.doc.docstatus === 0 &&
+			['Material Issue', 'Material Receipt', 'Material Transfer', 'Send to Subcontractor'].includes(frm.doc.purpose)) {
+				frm.add_custom_button(__('Bill of Materials'), function() {
+					frm.events.get_items_from_bom(frm);
+				}, __("Get items from"));
+		}
+	},
+
 	get_items_from_bom: function(frm) {
 		let filters = function(){
 			return {filters: { docstatus:1 }};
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -387,6 +393,73 @@ frappe.ui.form.on('Stock Entry', {
...
+			options:"Warehouse"},
+			{"fieldname":"target_warehouse", "fieldtype":"Link", "label":__("Target Warehouse"),
+			options:"Warehouse"},
+			{"fieldname":"qty", "fieldtype":"Float", "label":__("Quantity"),
+			reqd: 1, "default": 1},
+			{"fieldname":"fetch_exploded", "fieldtype":"Check",
+			"label":__("Fetch exploded BOM (including sub-assemblies)"), "default":1},
+			{"fieldname":"fetch", "label":__("Get Items from BOM"), "fieldtype":"Button"}
+		]
+		if (frm.doc.stock_entry_type == 'Material Issue'){
```
rohitwaghchaure: stock_entry_type can be anything, use frm.doc.purpose

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -251,6 +251,17 @@ frappe.ui.form.on('Stock Entry', {
 		frm.trigger("setup_quality_inspection");
 	},
 
+	stock_entry_type: function(frm){
+		frm.remove_custom_button('Bill of Materials', "Get items from");
+
+		if (frm.doc.docstatus === 0 && ['Material Issue','Material Receipt',
+		'Material Transfer','Send to Subcontractor'].includes(frm.doc.purpose)) {
```
rohitwaghchaure: use standard in_list method

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -387,6 +398,73 @@ frappe.ui.form.on('Stock Entry', {
...
+						$.each(r.message, function(i, item) {
+							let d = frappe.model.add_child(cur_frm.doc, "Stock Entry Detail", "items");
+							d.item_code = item.item_code;
+							d.item_name = item.item_name;
+							d.item_group = item.item_group;
+							d.s_warehouse = values.source_warehouse;
+							d.t_warehouse = values.target_warehouse;
+							d.uom = item.stock_uom;
+							d.stock_uom = item.stock_uom;
+							d.conversion_factor = 1;
```
rohitwaghchaure: Why 1? bom has conversion_factor field

erpnext/stock/doctype/stock_entry/stock_entry.js
marination: in_list is a wrapper around list.includes(item), isn't any difference. I formatted it better in the latest commit

erpnext/stock/doctype/stock_entry/stock_entry.js
marination: it returns conversion factor if fetch_exploded_items is unchecked. I added conditional check in the latest commit.

rohitwaghchaure (dismissed):

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -238,6 +238,13 @@ frappe.ui.form.on('Stock Entry', {
 			}, __("Get items from"));
 		}
 
+		if (frm.doc.docstatus === 0 &&
+			['Material Issue', 'Material Receipt', 'Material Transfer', 'Send to Subcontractor'].includes(frm.doc.purpose)) {
+				frm.add_custom_button(__('Bill of Materials'), function() {
+					frm.events.get_items_from_bom(frm);
+				}, __("Get items from"));
+		}
```
rohitwaghchaure: Code is repeated, create a separate method and call it from refresh and stock_entry_type trigger

marination: done

rohitwaghchaure (approved):

PR #19499, opened 2019-11-04
feat: update payment request total with payment entry amount

melBoh:
On payment entry submit, currently it checks if there is any payment request and set them as paid.

With this PR, in case the payment entry does not cover the requested grand total, we keep the request open but the grand total is updated with the deduction of the payment entry amount.

### DIFF
--- a/erpnext/accounts/doctype/payment_request/payment_request.py
+++ b/erpnext/accounts/doctype/payment_request/payment_request.py
@@ -6,7 +6,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _
 from frappe.model.document import Document
-from frappe.utils import flt, nowdate, get_url
+from frappe.utils import flt, nowdate, get_url, fmt_money
 from erpnext.accounts.party import get_party_account, get_party_bank_account
 from erpnext.accounts.utils import get_account_currency
 from erpnext.accounts.doctype.payment_entry.payment_entry import get_payment_entry, get_company_defaults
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -398,6 +398,32 @@ def make_status_as_paid(doc, method):
...
+				"docstatus": 1})
+
+		if payment_request_name:
+			pr = frappe.get_doc("Payment Request", payment_request_name)
+			pr_amount = pr.grand_total
+			if pr.grand_total <= paid_amount:
+				# entirely paid; close PR
+				if pr.status != "Paid":
+					pr.db_set('status', 'Paid')
+					pr.add_comment("Comment", _("Paid by Payment Entry {0}".format(pe.name)))
```
nabinhait, line 418: Instead of adding comment, we should add Dashboard in payment request to show linked payment entries.

erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -398,6 +398,32 @@ def make_status_as_paid(doc, method):
...
+	paid_amount = pe.paid_amount
+	for ref in pe.references:
+		payment_request_name = frappe.db.get_value("Payment Request",
+			{"reference_doctype": ref.reference_doctype, "reference_name": ref.reference_name,
+				"docstatus": 1})
+
+		if payment_request_name:
+			pr = frappe.get_doc("Payment Request", payment_request_name)
+			pr_amount = pr.grand_total
+			if pr.grand_total <= paid_amount:
```
nabinhait, line 414: paid_amount should not be considered based on `pe.paid_amount`. Instead it should be based on `ref.allocated_amount`.

nabinhait (changes requested):
I think, currently the system does not update payment amount at all in payment request.

If you want to make it happend, then don't update the grand_total itself. You should add a separate field "paid amount" and update there. Otherwise, the original requested amount (grand_total) will be lost.

nabinhait (commented):
Closing this for now.

PR #19500, opened 2019-11-04
enhancement: Item-wise sales history report

Mangesh-Khairnar:
- Added filters for item group, from date/to date based on transaction date
- refactor query report to script report


erpnext/hr/utils.py
```diff
@@ -157,6 +157,16 @@ def get_employee_field_property(employee, fieldname):
 	else:
 		return False
 
+def get_company_set(company):
+	return frappe.db.sql_list("""
+		SELECT
+			name
+		FROM `tabCompany`
+		WHERE
+			parent_company=%(company)s
```
nabinhait: It will only return immediate children companies. But if the company tree has multiple levels, then it will not return all companies. Use `lft` and `rgt` to fetch all children.

erpnext/selling/report/item_wise_sales_history/item_wise_sales_history.py
```diff
@@ -0,0 +1,212 @@
...
+			"company": record.company
+		}
+		data.append(row)
+
+	return data
+
+def get_conditions(filters):
+	conditions = ''
+	if filters.get('item_group'):
+		conditions += "AND so_item.item_group = '%s'" %filters.item_group
```
nabinhait: Escape filters.item_group

nabinhait (changes requested):

nabinhait (approved):

PR #19501, opened 2019-11-04
feat: multiple company pos profile

Mangesh-Khairnar:
**Issue:** Currently changing POS profile is allowed for only the company that is currently loaded by default in sales invoice
**Proposed Fix:** Adding a company field in model on changing the POS profile will allow user to change to a POS profile to one that belongs to a non default company.
multipos_profile

### DIFF
--- a/erpnext/selling/page/point_of_sale/point_of_sale.js
+++ b/erpnext/selling/page/point_of_sale/point_of_sale.js
@@ -471,7 +471,44 @@ erpnext.pos.PointOfSale = class PointOfSale {
 				}
 			}
 
-			frappe.prompt(this.get_prompt_fields(),
+
+			let me = this;
+
+			var dialog = frappe.prompt([{
+					fieldtype: 'Link',
+					label: __('Company'),
+					options: 'Company',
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@Mangesh-Khairnar Set the default pos profile in the popup on selection of the company

erpnext/selling/page/point_of_sale/point_of_sale.js
```diff
@@ -471,7 +471,44 @@ erpnext.pos.PointOfSale = class PointOfSale {
...
+					fieldtype: 'Link',
+					label: __('Company'),
+					options: 'Company',
+					fieldname: 'company',
+					default: me.frm.doc.company,
+					reqd: 1,
+					onchange: function(e) {
+							me.get_default_pos_profile(this.value).then((r) => {
+								if (r && r.name) {
+									dialog.set_value('pos_profile', r.name);
```
rohitwaghchaure: Make existing pos_profile blank if no pos profile found against the newly selected company

rohitwaghchaure (approved):

PR #19513, opened 2019-11-06
fix: Added 'status' field in Payment Entry

marination:
- Status wasn't accessible in Report View
**After Fix:**

Screenshot 2019-11-06 at 1 55 26 PM

Screenshot 2019-11-06 at 1 54 56 PM


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -274,6 +277,15 @@ def validate_journal_entry(self):
 						frappe.throw(_("Against Journal Entry {0} does not have any unmatched {1} entry")
 							.format(d.reference_name, dr_or_cr))
 
+	def set_status(self, status=None):
```
gavindsouza: `status` is unutilized. plus codacy

nabinhait (approved):

PR #19525, opened 2019-11-07
fix: Accounting Dimension custom fields should be admin owned

deepeshgarg007:
Custom fields created by accounting dimension should be owned by Administrator


nabinhait (dismissed):

nabinhait (commented):
Travis is failing

PR #19526, opened 2019-11-07
fix: Account Balance and Stock Value out of sync error message

marination:
**Before:**

Screenshot 2019-11-07 at 3 00 07 PM

**After:**

Screenshot 2019-11-11 at 12 12 36 PM

### DIFF
--- a/erpnext/accounts/general_ledger.py
+++ b/erpnext/accounts/general_ledger.py
@@ -163,9 +163,16 @@ def validate_account_for_perpetual_inventory(gl_map):
 						.format(account), StockAccountInvalidTransaction)
 
 			elif account_bal != stock_bal:
-				frappe.throw(_("Account Balance ({0}) and Stock Value ({1}) is out of sync for account {2} and linked warehouse ({3}). Please create adjustment Journal Entry for amount {4}.")
-					.format(account_bal, stock_bal, account, comma_and(warehouse_list), stock_bal - account_bal),
-					StockValueAndAccountBalanceOutOfSync)
+				frappe.throw(
+					_("""<b>Account Balance</b> and <b>Stock Value</b> is out of sync for account <b>{0}</b> and it's linked warehouses.
+						<br></br>
+						Please create adjustment Journal Entry for amount <b>{1}</b>
+						<br></br>
+						<div style="text-align:right;">
+						<a href="desk#Form/Journal Entry/New Journal Entry 1" class="btn btn-primary">Make Adjustment Entry</a>
+						</div>""")
+					.format(account, stock_bal - account_bal),StockValueAndAccountBalanceOutOfSync,
+					title=_('Account Balance Out Of Sync'))
 
 def validate_cwip_accounts(gl_map):
 	if not cint(frappe.db.get_value("Asset Settings", None, "disable_cwip_accounting")) \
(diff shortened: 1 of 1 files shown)


erpnext/accounts/general_ledger.py
```diff
@@ -163,9 +163,16 @@ def validate_account_for_perpetual_inventory(gl_map):
 						.format(account), StockAccountInvalidTransaction)
 
 			elif account_bal != stock_bal:
-				frappe.throw(_("Account Balance ({0}) and Stock Value ({1}) is out of sync for account {2} and linked warehouse ({3}). Please create adjustment Journal Entry for amount {4}.")
-					.format(account_bal, stock_bal, account, comma_and(warehouse_list), stock_bal - account_bal),
-					StockValueAndAccountBalanceOutOfSync)
+				frappe.throw(
+					_("""<b>Account Balance</b> and <b>Stock Value</b> is out of sync for account <b>{0}</b> and it's linked warehouses.
```
nabinhait: - You have to split this message to translate this properly, otherwise that html code will also be translated.
- Show account and stock value as well

nabinhait (changes requested):

erpnext/accounts/general_ledger.py
marination: done

erpnext/accounts/general_ledger.py
```diff
@@ -163,9 +163,16 @@ def validate_account_for_perpetual_inventory(gl_map):
...
+				error_reason = _("Account Balance ({0}) and Stock Value ({1}) is out of sync for account {2} and it's linked warehouses.").format(
+					account_bal, stock_bal, frappe.bold(account))
+				error_resolution = _("Please create adjustment Journal Entry for amount {0} ").format(frappe.bold(stock_bal - account_bal))
+				button_text = _("Make Adjustment Entry")
+
+				frappe.throw("""{0}<br></br>{1}<br></br>
+					<div style="text-align:right;">
+					<button class="btn btn-primary" onclick="frappe.new_doc('Journal Entry')">{2}</button>
+					</div>""".format(error_reason, error_resolution, button_text),
+					StockValueAndAccountBalanceOutOfSync, title=_('Account Balance Out Of Sync'))
```
nabinhait, line 175: Title should be "Stock and Account Balance Out Of Sync"

erpnext/accounts/general_ledger.py
```diff
@@ -163,9 +163,16 @@ def validate_account_for_perpetual_inventory(gl_map):
...
-					.format(account_bal, stock_bal, account, comma_and(warehouse_list), stock_bal - account_bal),
-					StockValueAndAccountBalanceOutOfSync)
+				error_reason = _("Account Balance ({0}) and Stock Value ({1}) is out of sync for account {2} and it's linked warehouses.").format(
+					account_bal, stock_bal, frappe.bold(account))
+				error_resolution = _("Please create adjustment Journal Entry for amount {0} ").format(frappe.bold(stock_bal - account_bal))
+				button_text = _("Make Adjustment Entry")
+
+				frappe.throw("""{0}<br></br>{1}<br></br>
+					<div style="text-align:right;">
+					<button class="btn btn-primary" onclick="frappe.new_doc('Journal Entry')">{2}</button>
```
nabinhait, line 173: Journal Entry should also have 2 rows added automatically.

Warehouse Account will be debited if difference is positive, otherwise credited.
Stock Adjustment will be credited if difference is positive, otherwise debited.

nabinhait (changes requested):

nabinhait (approved):

PR #19541, opened 2019-11-10
feat: [production_plan -> fetching item description] Fetch item description from Material Request or Sales Order

MorezMartin:
In production pan, when fetching items from Material Request or  from Sales Order, the description was fetched from item master and not from previous transaction.
It is necessary to fetch it because the sales person may have change description and it can be important for the manufacturing person to know it.
Same logic than https://github.com/frappe/erpnext/pull/18907
Example here https://nextcloud.mh-receptions.com/s/Z75nAsJGr53AjEg
Thanks in advance,
Sincerely yours,
Martin


erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -162,7 +162,7 @@ def add_items(self, items):
 				'include_exploded_items': 1,
 				'warehouse': data.warehouse,
 				'item_code': data.item_code,
-				'description': item_details and item_details.description or '',
+				'description': data.description,
```
rohitwaghchaure: data.description or item_details.description?

MorezMartin: For me data.description is better.
data.description will fetch description from sales order (items / packed items) and from material requests, item_details.description will fetch it from item_masters
If a vendor modify description in sales order or in packed items, or in material request, i think it is better for manufacture to see it over the production plan and work orders.
In the same logic than https://github.com/frappe/erpnext/pull/18907
Sincerely yours,
Martin

rohitwaghchaure: Agreed, but in case if description is blank(no in your case) then system should have to set the description from the item master.

```
'description': data.description or item_details.description
```

MorezMartin: done

nabinhait (approved):

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -726,6 +727,6 @@ def get_sub_assembly_items(bom_no, bom_data, qty):
 				})
 
 			bom_item = bom_data.get(key)
-			bom_item["stock_qty"] += ((d.stock_qty * qty) / d.parent_bom_qty)
```
rohitwaghchaure, line 729: @MorezMartin Why this code was removed?

PR #19552, opened 2019-11-12
fix: email digest showing incorrect upcoming events

rohitwaghchaure:
**Issue**
On Tuesday showing calendar events of the Monday 

Screenshot 2019-11-12 at 4 34 22 PM

**After Fix**

image

### DIFF
--- a/erpnext/setup/doctype/email_digest/email_digest.py
+++ b/erpnext/setup/doctype/email_digest/email_digest.py
@@ -4,8 +4,8 @@
 from __future__ import unicode_literals
 import frappe
 from frappe import _
-from frappe.utils import fmt_money, formatdate, format_time, now_datetime, \
-	get_url_to_form, get_url_to_list, flt, get_link_to_report
+from frappe.utils import (fmt_money, formatdate, format_time, now_datetime,
+	get_url_to_form, get_url_to_list, flt, get_link_to_report, add_to_date, today)
 from datetime import timedelta
 from dateutil.relativedelta import relativedelta
 from frappe.core.doctype.user.user import STANDARD_USERS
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/email_digest/email_digest.py
```diff
@@ -825,4 +826,14 @@ def get_count_for_period(account, fieldname, from_date, to_date):
 		last_year_closing_count = get_count_on(account, fieldname, fy_start_date - timedelta(days=1))
 		count = count_on_to_date + (last_year_closing_count - count_before_from_date)
 
-	return count
\ No newline at end of file
+	return count
+
+def get_future_date_for_calendaer_event(frequency):
+	from_date = to_date = now_datetime().date()
```
nabinhait: Use `today()` function directly, it will return only date

erpnext/setup/doctype/email_digest/email_digest.py
```diff
@@ -825,4 +826,14 @@ def get_count_for_period(account, fieldname, from_date, to_date):
...
+
+	if frequency == "Weekly":
+		to_date = now_datetime().date() - relativedelta(days=1) + relativedelta(weeks=1)
+	elif frequency == "Monthly":
+		to_date = now_datetime().date() - relativedelta(days=1) + relativedelta(months=1)
```
nabinhait: Use `add_days` and `add_months` functions

nabinhait (changes requested):

erpnext/setup/doctype/email_digest/email_digest.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #19556, opened 2019-11-12
fix: skip leave ledger entry creation for denied leaves

Mangesh-Khairnar:
**Issue:** leave ledger entries are getting created against leave application that are not approved which are  skewing the output leave balance
#### Fixes
- Only create leave ledger entry for leave application that are approved
- add patch that will remove the leave ledger entry that are created against the rejected and cancelled leave application


nabinhait (dismissed):

nabinhait (commented):
@Mangesh-Khairnar Fix travis

nabinhait (approved):

PR #19559, opened 2019-11-12
fix: fetch leave approver defined in employee in leave application

Mangesh-Khairnar:
**Issue:** Leave approver defined in the employee was not getting fetched in the leave approvers list in a leave application.

**Fix:** Add approver from employee master separately in the approvers list.

### DIFF
--- a/erpnext/hr/doctype/department_approver/department_approver.py
+++ b/erpnext/hr/doctype/department_approver/department_approver.py
@@ -19,14 +19,19 @@ def get_approvers(doctype, txt, searchfield, start, page_len, filters):
 	approvers = []
 	department_details = {}
 	department_list = []
-	employee_department = filters.get("department") or frappe.get_value("Employee", filters.get("employee"), "department")
-	if employee_department:
-		department_details = frappe.db.get_value("Department", {"name": employee_department}, ["lft", "rgt"], as_dict=True)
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/department_approver/department_approver.py
```diff
@@ -26,7 +26,12 @@ def get_approvers(doctype, txt, searchfield, start, page_len, filters):
 		department_list = frappe.db.sql("""select name from `tabDepartment` where lft <= %s
 			and rgt >= %s
 			and disabled=0
-			order by lft desc""", (department_details.lft, department_details.rgt), as_list = True)
+			order by lft desc""", (department_details.lft, department_details.rgt), as_list=True)
+
+	leave_approver = frappe.get_cached_value('Employee', filters.get('employee'), 'leave_approver')
+	if leave_approver:
+		approver = frappe.db.get_value("User", leave_approver, ['name', 'first_name', 'last_name'])
+		approvers.append(approver)
```
nabinhait: If leave approver defined on employee, then only return that approver. Otherwise, fetch from department

nabinhait (changes requested):

erpnext/hr/doctype/department_approver/department_approver.py
Mangesh-Khairnar: added change

nabinhait (approved):

PR #19571, opened 2019-11-13
chore: pinned requirements

gavindsouza:
Pinned requirements. Will be updated by [dependabot](https://dependabot.com/) further ahead


cclauss:
Please consider alphabetizing the list.  It is much easier to spot duplicate and missing requirements when the list is ordered.

PR #19582, opened 2019-11-14
fix: Enhanced Payment reconciliation Debit/Credit Note query

marination:
- Payment Reconciliation showed ALL debit/credit notes in Payments Table
- **Use Case:** A debit note is made against an **unpaid** Sales Invoice for adjustment purpose. There is no reconciliation required here and thus the debit note must not appear in Payment Reconcilation. 
On the other hand a **paid** invoice must have it's note available for future reconciliation.

- Payment Reconciliation will show Debit/Credit Notes in the following cases:
   1) It is a standalone note
   2) It is created against a **Paid** Invoice

### DIFF
--- a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
+++ b/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
@@ -90,7 +90,11 @@ class PaymentReconciliation(Document):
 			FROM `tab{doc}`, `tabGL Entry`
 			WHERE
 				(`tab{doc}`.name = `tabGL Entry`.against_voucher or `tab{doc}`.name = `tabGL Entry`.voucher_no)
-				and `tab{doc}`.is_return = 1 and `tab{doc}`.return_against IS NULL
+				and `tab{doc}`.is_return = 1
+				and (`tab{doc}`.return_against IS NULL or
+					(`tab{doc}`.return_against in
+						(select reference_name from `tabPayment Entry Reference`
+						where reference_name = `tab{doc}`.return_against and docstatus != 2)))
 				and `tabGL Entry`.against_voucher_type = %(voucher_type)s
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
```diff
@@ -90,7 +90,11 @@ def get_dr_or_cr_notes(self):
 			FROM `tab{doc}`, `tabGL Entry`
 			WHERE
 				(`tab{doc}`.name = `tabGL Entry`.against_voucher or `tab{doc}`.name = `tabGL Entry`.voucher_no)
-				and `tab{doc}`.is_return = 1 and `tab{doc}`.return_against IS NULL
+				and `tab{doc}`.is_return = 1
+				and (`tab{doc}`.return_against IS NULL or
+					(`tab{doc}`.return_against in
+						(select reference_name from `tabPayment Entry Reference`
+						where reference_name = `tab{doc}`.return_against and docstatus != 2)))
```
nabinhait, line 97: Couldn't understand the usecase. If credit note is created against a paid invoice and return_against is already mentioned, that means the credit note is already reconciled, why is should be fetched?

marination: if payment entry(linked to invoice) is made against invoice , debit/credit note is sent as there is no refund involved. The linked payment entry isn't reversed or deducted due to debit/credit note. So this note can be used in the future to reconcile other invoices right?

nabinhait: No, currently if we create a credit note against paid invoice, then outstanding amount is updated on original invoice and it get's into negative. That's why we can't link the credit note to any other future invoice/payment. We need to link original invoice to the new invoice/payment.
Though it's not a good design, currently it works in that way.

PR #19591, opened 2019-11-15
fix: add status "expired" to doctype quotation

0Pranav:
closes FR-ISS-152101

Adds the `Expired` to the `Status` field of DocType `Quotation`.  Also adds a daily task to set expired status to expired quotations. Previously this was done on client-side which causes inconstancy in filters.


erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -185,6 +185,14 @@ def update_item(obj, target, source_parent):
 
 	return doclist
 
+def set_expired_status():
```
deepeshgarg007, line 188: @0Pranav Why are you fetching all the quotations and their docs. Just fetch the names of quotations for which you want to update status and update them in one single query. No need to for loop and get doc of each quotation and make explicit commit.

deepeshgarg007 (changes requested):

deepeshgarg007:
@0Pranav Please add a good description. No one will understand `closes FR-ISS-152101`.

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -185,6 +185,12 @@ def update_item(obj, target, source_parent):
...
+	from datetime import date
+	DATE_FORMAT = "%Y%m%d" # For converting python date to SQL comparable date
+	today = date.today().strftime(DATE_FORMAT)
+	frappe.db.sql("UPDATE tabQuotation SET status = 'Expired' WHERE valid_till < " + today)
```
deepeshgarg007:
```suggestion
	frappe.db.sql("""UPDATE `tabQuotation` SET status = 'Expired' 
		WHERE status != 'Expired' AND 'valid_till < %s""" , (nowdate()))
```
This should do I guess. No need to import date and use DATE_FORMAT

deepeshgarg007 (changes requested):

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -185,6 +185,13 @@ def update_item(obj, target, source_parent):
 
 	return doclist
 
+def set_expired_status():
+	from datetime import date
```
deepeshgarg007: @0Pranav Please remove the import and other two lines which are not required anymore.

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #19594, opened 2019-11-16
fix: job_opining.py

amutaher:
## Description of the issue

when creating a job opening depends on  staffing plan get an error message "`Job Openings for designation `**Analaystic** `already open or hiring completed as per Staffing Plan` **Test**"

## UseCase
I create staffing plan Called **Test** for designation **Analaystic** with  `Vacancies=1` and already have one employee with designation **Analaystic** 

## Reason of Issue
https://github.com/frappe/erpnext/blob/642441688687286ba0fce0df1b2319529c723bdd/erpnext/hr/doctype/job_opening/job_opening.py#L45

## Error			

`if self.planned_vacancies <= current_count:`

## Correction 		

`if self.planned_vacancies < current_count:
`

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -39,9 +39,11 @@ class BOM(WebsiteGenerator):
 			names = [d[-1][1:] for d in filter(lambda x: len(x) > 1 and x[-1], names)]
 
 			# split by (-) if cancelled
-			names = [cint(name.split('-')[-1]) for name in names]
-
-			idx = max(names) + 1
+			if names:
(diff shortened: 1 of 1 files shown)


amutaher:
## Other **Bug** related to this issue 

After creating job opening I want to make status=close when clicking Save Button check the same function  `validate_current_vacancies` and show the error message 
"`Job Openings for designation` **Analaystic** `already open or hiring completed as per Staffing Plan `**Test**"

must no check that method on `Update`

deepeshgarg007:
@AkramMutaher There are no commits related to the issue mentioned in the description. Please correct and reopen. Also raise the PR against `develop` branch and not against `version-12`. From `develop` it will be backported to `version-12`

PR #19598, opened 2019-11-17
fix: performance issue of sales invoice while save/submit

rohitwaghchaure:
**Issue**

Time required to save invoice with 14 items
Screenshot 2019-11-19 at 1 19 32 PM

**After Fix**

Time required to save invoice with 14 items
Screenshot 2019-11-19 at 12 34 40 PM


erpnext/controllers/accounts_controller.py
```diff
@@ -100,8 +100,8 @@ def validate(self):
 				self.validate_qty()
 
 		validate_regional(self)
-		if self.doctype != 'Material Request':
-			validate_pricing_rules(self)
+		# if self.doctype != 'Material Request':
```
nabinhait: Should it be commented?

nabinhait (changes requested):

erpnext/controllers/accounts_controller.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #19603, opened 2019-11-18
fix: Prefilled JV via Account Balance and Stock Value mismatch error message

marination:
Requires: https://github.com/frappe/frappe/pull/8835
- Shortened title and added primary action button provision **(see https://github.com/frappe/frappe/pull/8835)**

Screenshot 2019-11-18 at 3 30 47 PM

- **'Make JV'** button will route to Journal Entry and add rows in child table

_out_of_sync


marination:
@nabinhait  Including both account balance and stock value in the title makes it bulky and it doesn't fit. Mentioning either one of them is an option OR using a generic title 'Values out of sync' is another workaround. Anything better?

PR #19610, opened 2019-11-19
feat(Contacts): select billing contact for sales invoice

ronelvcabrera:
Most of the time, point of contact are different from sale, invoicing and on different aspect/phase of the transaction. This feature is to have contacts that are identified for billing.

Therefore, as seen below, when creating a sales invoice, the priority to retrieve a contact for Sales invoice are tagged as `is Billing Contact` contacts.

In connection with this PR https://github.com/frappe/frappe/pull/8840

### Usage:

Billing contact will be retrieved during creation of Invoice as point of contact at this stage is different from sale or other aspect/stage of the transaction.

core is billing contact new contact


ronelvcabrera:
Build error is related to this PR https://github.com/frappe/frappe/pull/8840 where function has been aded a parameter.

ronelvcabrera:
In relation to the linked PR (https://github.com/frappe/frappe/pull/8840/files), changes here will be moved in ERPNext.

ronelvcabrera:
Been having unrelated error on Travis CI and is the same with the another PR https://github.com/frappe/erpnext/pull/19612. 

Checked out `develop` branch and ran `bench run-tests --app erpnext` and still having the same error.

erpnext/accounts/party.py
```diff
@@ -117,7 +117,10 @@ def get_regional_address_details(party_details, doctype, company):
 	pass
 
 def set_contact_details(party_details, party, party_type):
```
deepeshgarg007, line 119:
```suggestion
def set_contact_details(party_details, party, party_type, doctype=None):
```

deepeshgarg007: Travis is failing because of this. Fix and reopen closing this for now.

deepeshgarg007 (changes requested):

PR #19611, opened 2019-11-19
fix: fetch leave approvers from both department and employee master

Mangesh-Khairnar:
fetch leave approver from both the employee master and department approvers on creation of leave application.

> Screenshots/GIFs


nabinhait (changes requested):
@Mangesh-Khairnar If one approver is selected both on employee and department, it will show twice.

nabinhait (approved):

PR #19612, opened 2019-11-19
feat(Sales/Purchase Order): optional to reference a Blanket Order

ronelvcabrera:
This feature is to be able to have an option to reference or not a Blanket Order when creating a Sales/Purchase Order.

We encounter a customer that doesn't want a Sales Order be under a existing Blanket Order. Currently, when creating a Sales Order that has the same details on an existing Blanket Order, system will automatically set a Blanket Order for said SO.

We added a checkbox that will cater such scenario. Ticking the check box will automatically set the Blanket Order and if not ticked, system will leave it as blank.

Feature is applied to both Purchase Order and Sales Order.

We added test case for both PO and SO to validate that the feature is working as expected.

### Create Sales Order through Blanket Order
bo sales order

### Create Sales Order manually without Blank Order
bo sales order no bo

### Create Sales Order Manually with Blanket Order
bo sales order w bo


ronelvcabrera:
having unrelated error on Travis CI

ronelvcabrera:
Been having unrelated error on Travis CI and is the same with the another PR https://github.com/frappe/erpnext/pull/19610. 

Checked out `develop` branch and ran `bench run-tests --app erpnext` and still having the same error.

nabinhait (approved):

PR #19618, opened 2019-11-19
fix: stock qty not displayed in pos

nextchamp-saqib:
Problem:
- Stock ledger displays quantity in warehouse.
- POS displays zero stock qty

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/pos.py
+++ b/erpnext/accounts/doctype/sales_invoice/pos.py
@@ -357,14 +357,11 @@ def get_customer_wise_price_list():
 
 def get_bin_data(pos_profile):
 	itemwise_bin_data = {}
-	cond = "1=1"
+	filters = { 'actual_qty': ['>', 0] }
 	if pos_profile.get('warehouse'):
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -357,14 +357,11 @@ def get_customer_wise_price_list():
...
 	itemwise_bin_data = {}
-	cond = "1=1"
+	filters = { 'actual_qty': ['>', 0] }
 	if pos_profile.get('warehouse'):
-		cond = "warehouse = %(warehouse)s"
+		filters.update({ 'warehouse': pos_profile.get('warehouse') })
 
-	bin_data = frappe.db.sql(""" select item_code, warehouse, actual_qty from `tabBin`
-		where actual_qty > 0 and {cond}""".format(cond=cond), {
-			'warehouse': frappe.db.escape(pos_profile.get('warehouse'))
```
nextchamp-saqib, line 366: frappe.db.escape was the culprit

rohitwaghchaure (approved):

PR #19663, opened 2019-11-22
refactor: Expense Claim Cost Center

jiloysss:
EXPENSE CLAIM
Added field Cost Center in table
expense_claim

EXPENSE CLAIM TYPE
Added check box Deferred Expense Accounts
If checked, Default Account in table will be filtered with Asset Accounts. Expense Accounts if otherwise
Expense_claim_type

UPDATE:
Don't hide cost_center field, remove it entirely. And write a patch to update child cost centers based on parent.
- Removed parent cost center
- Added patch to update child cost centers based on parent.

Copy cost center from first row on addition of new row. And on entering cost_center in any row, it should get copied to all rows with blank cost_center.
- Added scrip to copy cost center from first row on addition of new row based on the given reference
- (DONE) on entering cost_center in any row, it should get copied to all rows with blank cost_center


sammish (approved):

sammish:
@nabinhait I think most valuable your support. Kindly review and advise us if you need more changes, If not please merge to production bench.because we are expecting those changes in production bench as soon as possible.

Nb: If you require more changes, we are ready to change please advise us.

nabinhait (commented):
@jiloysss @sammish 
This is very pathetic code. You didn't even test it once after adding the `copy_from_first_row` and `copy_value_in_all_rows` related code.

I have sent another PR after cleanup. https://github.com/frappe/erpnext/pull/19807

nabinhait:
@sammish As this is a breaking change, will merge this (#19807) only in develop branch.

PR #19664, opened 2019-11-22
fix: Validation for parent cost center

deepeshgarg007:
Since javascript filters doesn't work in quick entry while creating new cost centers from quick entry
users often create new cost center under another child cost center after which they don't appear in the cost center tree. Adde a validation for it

image


nabinhait (changes requested):
Please add tests

nabinhait (approved):

PR #19667, opened 2019-11-22
fix: primary address not being fetched for customer

0Pranav:
before:
Screenshot 2019-11-22 at 1 52 13 PM

after:
Screenshot 2019-11-22 at 2 04 37 PM

### DIFF
--- a/erpnext/selling/doctype/customer/customer.js
+++ b/erpnext/selling/doctype/customer/customer.js
@@ -49,9 +49,8 @@ frappe.ui.form.on("Customer", {
 		})
 		frm.set_query('customer_primary_address', function(doc) {
 			return {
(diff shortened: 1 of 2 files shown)


Anurag810 (approved):

erpnext/selling/doctype/customer/customer.js
```diff
@@ -49,9 +49,9 @@ frappe.ui.form.on("Customer", {
 		})
 		frm.set_query('customer_primary_address', function(doc) {
 			return {
-				query: "erpnext.selling.doctype.customer.customer.get_customer_primary_address",
```
Anurag810, line 52: Remove function if not used..

Anurag810 (changes requested):

PR #19675, opened 2019-11-25
fix: Account type in Handling Difference in Inventory account

deepeshgarg007:
Back Port Of https://github.com/frappe/erpnext/pull/19674

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/verified/ae_uae_chart_template_standard.json
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/verified/ae_uae_chart_template_standard.json
@@ -188,6 +188,9 @@
                 },
                 "Expenses Included In Valuation": {
                     "account_type": "Expenses Included In Valuation"
+                },
+                "Stock Adjustment": {
+                    "account_type": "Stock Adjustment"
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

erpnext/accounts/doctype/account/chart_of_accounts/verified/ae_uae_chart_template_standard.json
```diff
@@ -1,465 +1,466 @@
...
-                    "Credit Card Charges": {}, 
-                    "Credit Card Swipe Charges": {}, 
+                    "Air Miles Card Charges": {},
+                    "Amex Credit Cards Charges": {},
+                    "Bank Finance & Loan Charges": {},
+                    "Credit Card Charges": {},
+                    "Credit Card Swipe Charges": {},
                     "PayPal Charges": {}
+                },
+                "Stock Expenses": {
```
nabinhait: Rename it to Stock Adjustment and place it under `Cost of Sale`

nabinhait (changes requested):

nabinhait (approved):

PR #19701, opened 2019-11-27
feat: Added Exempted doctype list in hooks

vishdha:
Usecase: 
As a User, I want to be able to delete the linked document to the one I'm trying to delete, so that I don't have to manually delete the document(s) 

Problem: 
Anytime you try to cancel a document that is linked somewhere else, it doesn't allow you to do so. Which is obviously a good thing to maintain data integrity. However, we should allow the User to cancel both the documents on another button click, whilst warning them of the consequences.

Solution:

cancel

Pull Request Depends on : https://github.com/frappe/frappe/pull/8905


malikabusamha:
There is a problem in the case of cancelling an asset that is linked with asset movement, due to the fact that asset movement gets cancelled only based on cancelation of asset and asset is linked with asset movement so it makes it impossible to cancel, It is a closed loop.
image

Alchez:
@MAlik944 That may be an issue with the circular linking of Asset and Asset Movement, which should be handled separately. If it's not a circular linking issue, then it could be added to the cancel-exempt doctypes.

PR #19708, opened 2019-11-27
feat: drill down on batch transactions (develop)

Alchez:
**Changes:**

- Add an Item filter to the Batch-Wise Balance History report.
- Clicking on a batch in the Batch-Wise Balance History navigates to the Stock Ledger report with the batch applied as a filter.
- Show items manufactured from the batch in the Stock Ledger.
- Indicate in and out quantities to the Stock Ledger report for every stock transaction (and a color formatter for out quantities).

<hr>

**Screenshots / GIFs:**

**Linking to Stock Ledger with batch filters:**

batch-ledger

**Changes to Stock Ledger report view:**

image


szufisher:
is it possible to make a generic and configurable report drill down feature. i.e in report design the user can specify which field under which condition,  by passing which options(filter)  and drill down to which target report?

Alchez:
@szufisher 

Do you mean to make this configurable on the desk? Or do you mean improving the reporting engine? Or both?

I'm not sure about the challenges for modifying Report Builder and Query Report reports, but for Script Reports, we could try adding some field controls to change cell links.

rohitwaghchaure (approved):

PR #19711, opened 2019-11-27
feat: allow searching from meta fields

scmmishra:
Allow searching itemss from meta fields for product website


erpnext/portal/product_configurator/utils.py
```diff
@@ -313,13 +313,20 @@ def get_items(filters=None, search=None):
...
+
+		# Build or filters for query
 		search = '%{}%'.format(search)
-		or_filters = [
-			['name', 'like', search],
-			['item_name', 'like', search],
-			['description', 'like', search],
-			['item_group', 'like', search]
-		]
+		or_filters = [[field, 'like', search] for field in search_fields]
```
nabinhait, line 333: We should not run LIKE statement in description field if there are huge numbers of item, the query becomes very slow. We also can't index description. Please refer https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/queries.py#L174

nabinhait (changes requested):

erpnext/portal/product_configurator/utils.py
scmmishra: I'll add the condition mentioned in the link you mentioned

scmmishra: Updated in both PRs

nabinhait (approved):

PR #19719, opened 2019-11-28
fix: added bank reconciliation page in the accounting module

Anurag810:
added bank reconciliation page in the accounting module
image

### DIFF
--- a/erpnext/config/accounts.py
+++ b/erpnext/config/accounts.py
@@ -197,6 +197,12 @@ def get_data():
 					"name": "Bank Reconciliation Statement",
 					"is_query_report": True,
 					"doctype": "Journal Entry"
+				},{
+					"type": "page",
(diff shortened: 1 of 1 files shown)


erpnext/config/accounts.py
```diff
@@ -197,6 +197,12 @@ def get_data():
 					"name": "Bank Reconciliation Statement",
 					"is_query_report": True,
 					"doctype": "Journal Entry"
+				},{
+					"type": "page",
+					"name": "bank-reconciliation",
+					"label": _("Bank Reconciliation"),
+					"icon": "fa fa-bar-chart",
+					"onboard": 1,
```
marination: Why onboard 1?

nabinhait (approved):

nabinhait:
@Anurag810 please send another PR for version-12-hotfix

PR #19720, opened 2019-11-28
fix: remove permission for all users to read leads

0Pranav:
Issue : In Lead Details, everything is visible to every other User irrespective of User Permissions applied on it

Cause : Lead had permissions for all users to be able to lead

Fix : Remove read permissions from all users

> Screenshots/GIFs

Before user could view all leads
image

After user can only view leads in child company
image


nabinhait (commented):
READ permissions were given to All for permlevel 1, which should only be used for field level permissions. I think only permlevel "0" is used to check form level permissions.
May be issue is in the Core, please consult with @surajshetty3416

0Pranav:
Issue is caused by another known issue in the framework

PR #19722, opened 2019-11-28
fix: remove permission for all users to read leads

0Pranav:
backport of https://github.com/frappe/erpnext/pull/19720/


0Pranav:
The issue is caused by another issue in the framework.

PR #19730, opened 2019-11-28
fix: set opportunity as lost on quotation close

0Pranav:
Issue: The status of the opportunity was set to `None` if a quotation against that opportunity was closed

Cause: It was explicitly set to `None`

Fix: When the quotation is set to lost, if the opportunity it was made against had no other quotations, the opportunity is also set to lost.

Other possible solutions: would be to not change the status of the opportunity at all 

GIF:

image

### DIFF
--- a/erpnext/selling/doctype/quotation/quotation.py
+++ b/erpnext/selling/doctype/quotation/quotation.py
@@ -67,7 +67,6 @@ class Quotation(SellingController):
 			opportunity = self.opportunity
 
 		opp = frappe.get_doc("Opportunity", opportunity)
-		opp.status = None
 		opp.set_status(update=True)
 
 	def declare_enquiry_lost(self, lost_reasons_list, detailed_reason=None):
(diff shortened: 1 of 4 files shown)


erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -67,8 +67,16 @@ def update_opportunity_status(self, opportunity=None):
...
-		opp.set_status(update=True)
+		quotations_on_opportunity = frappe.db.sql("""
+			SELECT name
+			FROM tabQuotation
+			WHERE `opportunity` = %s AND `status` != "Lost"
+		""",(opportunity))
+
+		if(len(quotations_on_opportunity) <= 1):
+			print("Setting as lost")
+			opp.status = "Lost"
```
nabinhait: Why we need to set status here? It should be set based on conditions defined on status_updater.py?

PR #19732, opened 2019-11-28
fix: UOM was not fetching in purchase invoice

Anurag810:
While selecting the Item in Purchase Invoice, the UOM is not getting fetched:
image

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
@@ -330,24 +330,6 @@ erpnext.accounts.PurchaseInvoice = erpnext.buying.BuyingController.extend({
 			frm: cur_frm
 		})
 	},
-
-	// CAN I REMOVE THIS IT WAS OVERWRITING THE transaction.js ka item_code event.
-	//item_code: function(frm, cdt, cdn) {
-	// 	var row = locals[cdt][cdn];
-	// 	if(row.item_code) {
-	// 		frappe.call({
-	// 			method: "erpnext.assets.doctype.asset_category.asset_category.get_asset_category_account",
-	// 			args: {
-	// 				"item": row.item_code,
-	// 				"fieldname": "fixed_asset_account",
-	// 				"company": frm.doc.company
-	// 			},
-	// 			callback: function(r, rt) {
-	// 				frappe.model.set_value(cdt, cdn, "expense_account", r.message);
-	// 			}
-	// 		})
-	// 	}
-	// }
 });
 
 cur_frm.script_manager.make(erpnext.accounts.PurchaseInvoice);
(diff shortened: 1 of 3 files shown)


erpnext/assets/doctype/asset_category/asset_category.py
```diff
@@ -29,7 +29,8 @@ def get_asset_category_account(fieldname, item=None, asset=None, account=None, a
 				account=None
 
 		if not account:
-			asset_category, company = frappe.db.get_value("Asset", asset, ["asset_category", "company"])
+			#if it is not finding the value it is returning single 'None' which is not iterable to asset_category and company.
+			asset_category, company = frappe.db.get_value("Asset", asset, ["asset_category", "company"]) if frappe.db.get_value("Asset", asset, ["asset_category", "company"]) else None, None
```
nabinhait: This is very bad way to writing code, you are doing get_value twice. Instead you should set the result in a variable.

erpnext/stock/get_item_details.py
```diff
@@ -261,6 +261,11 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 		elif (args.get('doctype') in ['Purchase Order', 'Purchase Receipt', 'Purchase Invoice']) or \
 			(args.get('doctype') == 'Material Request' and args.get('material_request_type') == 'Purchase'):
 			args.uom = item.purchase_uom if item.purchase_uom else item.stock_uom
+			if args.get('doctype') == 'Purchase Invoice':
+				from erpnext.assets.doctype.asset_category.asset_category import get_asset_category_account
+				expense_account = get_asset_category_account(fieldname = "fixed_asset_account", item = args.item_code, company= args.company)
+				if not expense_account:
```
nabinhait: This is not the logic. fixed_asset_account should only be used if item is a fixed asset item. otherwise it should be based on get_default_expense_account

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -331,22 +331,23 @@ erpnext.accounts.PurchaseInvoice = erpnext.buying.BuyingController.extend({
...
-					"company": frm.doc.company
-				},
-				callback: function(r, rt) {
-					frappe.model.set_value(cdt, cdn, "expense_account", r.message);
-				}
-			})
-		}
-	}
+	// CAN I REMOVE THIS IT WAS OVERWRITING THE transaction.js ka item_code event.
+	//item_code: function(frm, cdt, cdn) {
```
nabinhait: remove commented code

nabinhait (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
Anurag810: done

erpnext/stock/get_item_details.py
Anurag810: done

erpnext/assets/doctype/asset_category/asset_category.py
Anurag810: done

nabinhait (approved):

PR #19733, opened 2019-11-28
fix: Division by zero error in Stock Entry

marination:
- On changing item_code in Stock Entry Details **Quantity** wasn't triggered and so Basic Amount as well as subsequent fields didn't change.
- 'total_basic_amount' was 0 for some instances.
```
File "/home/frappe/benches/bench-12-2019-11-25-1/apps/erpnext/erpnext/stock/doctype/stock_entry/stock_entry.py", line 90, in on_submit
    self.make_gl_entries()
  File "/home/frappe/benches/bench-12-2019-11-25-1/apps/erpnext/erpnext/controllers/stock_controller.py", line 33, in make_gl_entries
    gl_entries = self.get_gl_entries(warehouse_account)
  File "/home/frappe/benches/bench-12-2019-11-25-1/apps/erpnext/erpnext/stock/doctype/stock_entry/stock_entry.py", line 658, in get_gl_entries
    (t.amount * d.basic_amount) / total_basic_amount
```
- After Fix:
  - Trigger Quantity as well to recalculate Basic Amount.
  - In presence of additional costs and absence of total basic amount , distribute additional costs based on quantity:
 ``` additional_cost_for_row = row_quantity * additional_cost / total_quantity```
  - Test case to secure the conditions: total basic amount =0 , additional costs >0, gl entries created for the additional costs only

  - Screenshot 2019-12-12 at 4 45 04 PM

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -27,6 +27,7 @@ class IncorrectValuationRateError(frappe.ValidationError): pass
 class DuplicateEntryForWorkOrderError(frappe.ValidationError): pass
 class OperationsNotCompleteError(frappe.ValidationError): pass
 class MaxSampleAlreadyRetainedError(frappe.ValidationError): pass
+class TotalBasicAmountZeroError(frappe.ValidationError): pass
 
 from erpnext.controllers.stock_controller import StockController
 
@@ -649,6 +650,12 @@ class StockEntry(StockController):
 		gl_entries = super(StockEntry, self).get_gl_entries(warehouse_account)
 
 		total_basic_amount = sum([flt(t.basic_amount) for t in self.get("items") if t.t_warehouse])
+
+		if self.get("additional_costs") and not total_basic_amount:
+			#If additional costs table is populated and total basic amount is
+			#somehow 0, interrupt transaction.
+			frappe.throw(_("Total Basic Amount in Items Table cannot be 0"), TotalBasicAmountZeroError)
+
 		item_account_wise_additional_cost = {}
 
 		for t in self.get("additional_costs"):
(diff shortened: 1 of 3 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -657,7 +657,7 @@ def get_gl_entries(self, warehouse_account):
 					item_account_wise_additional_cost.setdefault((d.item_code, d.name), {})
 					item_account_wise_additional_cost[(d.item_code, d.name)].setdefault(t.expense_account, 0.0)
 					item_account_wise_additional_cost[(d.item_code, d.name)][t.expense_account] += \
-						(t.amount * d.basic_amount) / total_basic_amount
+						(t.amount * d.basic_amount) / total_basic_amount if total_basic_amount else 0
```
nabinhait: if `total_basic_amount = 0` then you should return outside the `for` loop, there is no point in doing this unnecessary processing.

nabinhait (changes requested):
Also add tests

erpnext/stock/doctype/stock_entry/stock_entry.py
marination: Done

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -649,6 +650,12 @@ def get_gl_entries(self, warehouse_account):
 		gl_entries = super(StockEntry, self).get_gl_entries(warehouse_account)
 
 		total_basic_amount = sum([flt(t.basic_amount) for t in self.get("items") if t.t_warehouse])
+
+		if self.get("additional_costs") and not total_basic_amount:
+			#If additional costs table is populated and total basic amount is
+			#somehow 0, interrupt transaction.
+			frappe.throw(_("Total Basic Amount in Items Table cannot be 0"), TotalBasicAmountZeroError)
```
nabinhait: Actually it can be zero, if "Allow zero valuation rate" is mentioned.

I think, in that case, we can "total_qty" for proportionately divide additional charges.

nabinhait (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
marination: In that case if total_basic_amount is zero , all the **basic_amount** fields are also 0, **(t.amount * d.basic_amount) / total_basic_amount** will always be zero (given that the denominator>0). No point in passing **total_qty** into a loop that is always setting the value as 0?

marination: done

deepeshgarg007 (approved):

PR #19743, opened 2019-11-29
fields added to Salary Structure Assignment

santoshbb:
New fields added to "Salary Structure Assignment" as shown in below picture:

image

- It helps to generate all salary slips in one salary structure
Depending on value mentioned in custom fields conditions can be applied to generate salary slips
Ex: If employee is not allowed for HRA, we can select the HRA field as "No" and if is allowed can put the formula as shown in picture
image


nabinhait:
You can add custom fields for this. There are no limit of such field's addition. Also some of the fields are regional.

PR #19753, opened 2019-11-29
fix: payroll UI/UX fixes

Anurag810:
Before:
image
After:
working_fine


erpnext/hr/doctype/payroll_entry/payroll_entry.js
```diff
@@ -31,7 +31,11 @@ frappe.ui.form.on('Payroll Entry', {
 			}
 			if ((frm.doc.employees || []).length) {
 				frm.page.set_primary_action(__('Create Salary Slips'), () => {
-					frm.save('Submit');
+					frm.save('Submit').then(()=>{
+						frm.page.clear_primary_action();
+						cur_frm.refresh();
```
nabinhait: Use `frm` instead of `cur_frm`

nabinhait (changes requested):
Send another PR for version-12-hotfix

nabinhait (approved):

PR #19757, opened 2019-11-30
feat: added illustrations for onboarding

scmmishra:
Depends on https://github.com/frappe/frappe/pull/8926


erpnext/selling/setup_wizard_slide/add_a_few_customers/add_a_few_customers.json
```diff
@@ -7,14 +7,14 @@
  "domains": [],
  "help_links": [
   {
-   "label": "Customers",
+   "label": "Learn More",
```
ruchamahabal, line 10: Change the help link label for Suppliers slide too. Make it `Learn More` from `Suppliers`

erpnext/setup/setup_wizard_slide/welcome_to_erpnext!/welcome_to_erpnext!.json
```diff
@@ -7,13 +7,13 @@
  "domains": [],
  "help_links": [],
  "idx": 0,
- "image_src": "/assets/erpnext/images/illustrations/onboard.png",
+ "image_src": "/assets/erpnext/images/illustrations/desk-onboard.png",
```
ruchamahabal, line 10: Can we make the illustrations a bit darker? They do blend into the product but too light to be seen. If we are allocating space for the illustrations, they can be a bit darker and noticeable.

ruchamahabal, line 10: Also, I feel, the first slide's image can be more welcoming rather than just showing the desk illustration?

ruchamahabal (changes requested):
wizard
Also, Can we have the modal-header back? It gave some context about the entire process.

erpnext/selling/setup_wizard_slide/add_a_few_customers/add_a_few_customers.json
scmmishra: +1

erpnext/setup/setup_wizard_slide/welcome_to_erpnext!/welcome_to_erpnext!.json
scmmishra: The color of the illustrations is deliberate for the following reasons.
1. The rest of the product has no illustrations what-so-ever, so anything bold will be inconsistent with the overall feel of the product
1. The extreme light nature makes them as minimal as possible. The intention is for it to not be noticeable, so that the user can instead focus on the information he or she is supposed to enter.
1. The choice of not using any other colors was deliberate, because the product uses color explicitly to convey status information.
1. The colors used in the illustrations lie between the standard colors used in the product.

However there is a good case for having better colors and more contrast for a few reasons.
1. Higher contrast has a accessibility value to it.
1. The current colors is not entirely from the standard palette anyway.. Might as well go all out with bolder illustrations and make it a fresher approach

@netchampfaris @surajshetty3416 @ruchamahabal I could use some feedback on the points above...

scmmishra: The desk illustration is more relatable than the handshake before. Besides the Desk UI is synonymous to ERPNext within the group that knows about Frappe and ERPNext. I don't see anything fitting other than this. Would love alternate recommendations for the illustrations.

The copy is welcoming enough on the other hand.

scmmishra:
> wizard
> Also, Can we have the modal-header back? It gave some context about the entire process.

Modal header conveyed no information and no actions, it was just a static text. No point of adding the header without clear intent.

An alternate argument can be of having the title to change based on the slide or show slide number... But that is already shown clearly in the body, adding them will be creating redundancy in design.

scmmishra:
Closing Pull Request as I expect more changes.. Will open it once done

@netchampfaris  @ruchamahabal

PR #19758, opened 2019-11-30
fix(Project): add project in child for items

hrwX:
- Project was not being set in Items when Purchase Invoice, Purchase Receipt, Purchase Order were created from Projects Dashboards.


erpnext/projects/doctype/project/project.js
```diff
@@ -16,6 +16,54 @@ frappe.ui.form.on("Project", {
 					time_log.parenttype = 'Timesheet';
 					new_doc.time_logs = [time_log];
 
+					frappe.ui.form.make_quick_entry(doctype, null, null, new_doc);
```
deepeshgarg007: @hrwX can you move this code to a generalised function and  use it for all the doctypes. Also raise a PR on version-12-hotfix as well

deepeshgarg007 (changes requested):

erpnext/projects/doctype/project/project.js
hrwX: okay will do that

deepeshgarg007 (approved):

Issue #19770, opened 2019-12-02
Pricing Rule error : AttributeError: 'str' object has no attribute 'get'


RickyHutanoto:
Upon saving a new pricing rule with "Apply on = Transaction"
System show this message : AttributeError: 'str' object has no attribute 'get'

ERPNext version : 12.2.0
Frappe version : 12.0.2

khushalti:
I am looking into it

khushalti:
#19774 I have raised but Travis is failed not able to open Travis details as well

khushalti:
please close this issue this fix has been merged

PR #19774, opened 2019-12-02
fix:Pricing Rule error AttributeError: 'str' object has no attribute 'get'

khushalti:
#19770 

image


khushalti:
@deepeshgarg007  can you help me .. Travis build got failed but details are not loading what went wrong

image

scmmishra:
I've restarted the build.

scmmishra (commented):
are you sure your fix works.. I'm pretty sure the attribute error is caused on `d` in the list comprehension and not on `self` 

`self` is never gonna be a string in this case, so it will have an attribute `get`

khushalti:
@scmmishra yes it works i have tested...

khushalti:
image
I got above error on Travis build

khushalti:
@scmmishra any further updates on this?

scmmishra:
> @scmmishra yes it works i have tested...

It may work for multiple reasons, your fix is not it.

Just print the type of `self` it's not `str`
`self` is an instance of the Pricing Rule Class, it'll always have an attribute `get`

nabinhait (approved):

RickyHutanoto:
Suppose this fix also applied to Promotional Scheme

PR #19792, opened 2019-12-03
feat: Receivable / payable summary based on payment terms

nabinhait:


erpnext/accounts/report/accounts_receivable_summary/accounts_receivable_summary.js
```diff
@@ -106,6 +106,11 @@ frappe.query_reports["Accounts Receivable Summary"] = {
 			"label": __("Sales Person"),
 			"fieldtype": "Link",
 			"options": "Sales Person"
+		},
+		{
```
deepeshgarg007, line 110: @nabinhait Is this PR complete? There is no effect of the filter on report

Filter Applied:
image

Filter Not Applied:
image

A Due date column should also be added for the report to make sense if the filter is ticked

nabinhait: Yes, the PR is complete.
We can't add due date column, as it is a summary report. Basically, if we check "Based on Payment Terms" filter on summary report, we get ageing data as per payment terms due date. Nothing else are changed.

PR #19799, opened 2019-12-03
feat: add date filter in the fixed asset register

Mangesh-Khairnar:
- Added date filter which would calculate the asset value till that date only
Screenshot 2019-12-04 at 12 33 57 AM

> Screenshots/GIFs

### DIFF
--- a/erpnext/assets/report/fixed_asset_register/fixed_asset_register.py
+++ b/erpnext/assets/report/fixed_asset_register/fixed_asset_register.py
@@ -114,7 +114,7 @@ def get_data(filters):
 	data = []
 
 	conditions = get_conditions(filters)
-	depreciation_amount_map = get_finance_book_value_map(filters.finance_book, filters.date)
+	depreciation_amount_map = get_finance_book_value_map(filters.date, filters.finance_book)
 	pr_supplier_map = get_purchase_receipt_supplier_map()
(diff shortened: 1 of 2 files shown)


erpnext/assets/report/fixed_asset_register/fixed_asset_register.py
```diff
@@ -138,19 +140,22 @@ def get_data(filters):
...
 				"purchase_date": asset.purchase_date,
-				"current_value": current_value_map.get(asset.name)
+				"asset_value": asset_value
 			}
 			data.append(row)
 
 	return data
 
-def get_finance_book_value_map(finance_book=''):
+def get_finance_book_value_map(finance_book='', date=today()):
```
nabinhait: Don't write function in argument declaration, it creates problem (can't remember exactly what, but faced a problem earlier). Instead initialise inside the function like `if not date: date = today()`.

nabinhait (changes requested):

erpnext/assets/report/fixed_asset_register/fixed_asset_register.py
Mangesh-Khairnar: added changes

nabinhait (approved):

PR #19815, opened 2019-12-04
feat(marketplace): allow user to un-publish item

Mangesh-Khairnar:
Screenshot 2019-12-05 at 12 06 04 AM

> Screenshots/GIFs


erpnext/hub_node/api.py
```diff
@@ -119,34 +119,59 @@ def prepare_item(item):
...
 			'image_list': item.get('image_list')
-		}).insert(ignore_if_duplicate=True)
-
-	items = map_fields(items_to_publish)
+		}
+		if frappe.db.exists('Hub Tracked Item', item_code):
+			items_to_update.append(item)
+			hub_tracked_item = frappe.get_doc('Hub Tracked Item', item_code)
+			hub_tracked_item.update(hub_dict)
+			hub_tracked_items.save()
```
netchampfaris: Looks like you haven't tested this code. Please test these workflows before sending a Pull Request.

```suggestion
			hub_tracked_item.save()
```

erpnext/public/js/hub/pages/Home.vue
```diff
@@ -58,6 +58,11 @@ export default {
 		this.search_value = '';
 		this.get_items();
 	},
+	mounted() {
+		frappe.route.on('change', () => {
+			this.get_items();
```
netchampfaris: This will be called on every route change. Shouldn't this be called only on the `Home` route? Add a check for it.

erpnext/public/js/hub/pages/Item.vue
```diff
@@ -203,33 +199,39 @@ export default {
...
-			.catch(e => {
-				console.error(e);
-			});
+			hub
+				.call('add_item_to_user_saved_items', {
+					hub_item_name: this.hub_item_name,
+					hub_user: frappe.session.user
+				})
+				.then(() => {
+					const saved_items_link = `<b><a href="#marketplace/saved-items">${__(
```
netchampfaris: Translated strings on multiple lines won't work. They must be on a single line.

erpnext/public/js/hub/pages/Item.vue
```diff
@@ -203,33 +199,39 @@ export default {
...
-			.catch(e => {
-				console.error(e);
-			});
+			hub
+				.call('add_item_to_seller_featured_items', {
+					hub_item_name: this.hub_item_name,
+					hub_user: frappe.session.user
+				})
+				.then(() => {
+					const featured_items_link = `<b><a href="#marketplace/featured-items">${__(
```
netchampfaris: Translated strings on multiple lines won't work. They must be on a single line.

erpnext/public/js/hub/pages/Item.vue
```diff
@@ -300,10 +309,20 @@ export default {
 		},
 
 		unpublish_item() {
-			frappe.msgprint(__('This feature is under development...'));
+			frappe.confirm(__(`Unpublish ${this.item.item_name}?`), () => {
```
netchampfaris: Interpolated Strings don't work with translations. https://frappe.io/docs/user/en/translations#2-variables

netchampfaris (changes requested):

erpnext/patches/v12_0/set_published_in_hub_tracked_item.py
```diff
@@ -0,0 +1,12 @@
...
+
+def execute():
+	frappe.reload_doc("Hub Node", "doctype", "Hub Tracked Item")
+	if frappe.db.a_row_exists("Hub Tracked Item"):
+		return
+
+	frappe.db.sql('''
+		Update `tabHub Tracked Item`
+		SET published = 1
+	''')
```
netchampfaris: If there is a row in Hub Tracked Item then the SQL statement won't run. Is this the intended behavior?

netchampfaris (changes requested):

erpnext/patches/v12_0/set_published_in_hub_tracked_item.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	frappe.reload_doc("Hub Node", "doctype", "Hub Tracked Item")
+	if frappe.db.a_row_exists("Hub Tracked Item"):
```
Mangesh-Khairnar:
```suggestion
	if not frappe.db.a_row_exists("Hub Tracked Item"):
```

netchampfaris (approved):

PR #19862, opened 2019-12-08
feat(regional): a central place for regional address templates

barredterra:
Current situation:

- Every country implements it's own version of `update_address_template()`.
- No version of `update_address_template()` is tested.
- An Indian company cannot use a German Address Template and vice versa.
- Each template lives in his `erpnext/regional/{country}`.

With this PR:

- One single implementation.
- Unit-tests.
- Everybody can use any template.
- All templates live in `erpnext/regional/address_template/templates`.


nabinhait (changes requested):
@barredterra Travis failing

barredterra:
image

nabinhait (approved):

PR #19870, opened 2019-12-09
fix: website showing disabled items in product list

nextchamp-saqib:


erpnext/portal/product_configurator/utils.py
```diff
@@ -339,6 +339,10 @@ def get_items(filters=None, search=None):
 	where_conditions = ' and '.join(
 		[condition for condition in [show_in_website_condition, search_condition, filter_condition] if condition]
 	)
+	if where_conditions:
+		where_conditions += ' and disabled = 0'
+	else:
+		where_conditions += 'disabled = 0'
```
rohitwaghchaure: Better add where_conditions = 'disabled = 0' before line number 339

rohitwaghchaure (approved):

PR #19871, opened 2019-12-09
Backdated leave application

ParthKharwar:
Changes made:
- Check for backdated leave application added
- Only HR Manager can make backdated leave applications

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.py
+++ b/erpnext/hr/doctype/leave_application/leave_application.py
@@ -77,6 +77,10 @@ class LeaveApplication(Document):
 						frappe.throw(_("{0} applicable after {1} working days").format(self.leave_type, leave_type.applicable_after))
 
 	def validate_dates(self):
+		if self.from_date < frappe.utils.today():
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -77,6 +77,10 @@ def validate_applicable_after(self):
 						frappe.throw(_("{0} applicable after {1} working days").format(self.leave_type, leave_type.applicable_after))
 
 	def validate_dates(self):
+		if self.from_date and self.from_date < frappe.utils.today():
+			if "HR Manager" not in frappe.get_roles():
```
nabinhait, line 81: Role should  not be hard-coded, anyone can rename it. Also it should optional.

Instead, add a field "Role Allowed to Create Back-dated Leave Application" in HR Settings and validate based on it.

nabinhait (approved):

nabinhait:
@ParthKharwar Closing it for now, please reopen after fixing as per suggestion.

PR #19884, opened 2019-12-10
fix(WIP): added extra condition in leave application for ledger.

Anurag810:


nabinhait (approved):

nabinhait:
@Anurag810 Do we also need to fix this in version-12-hotfix?

PR #19902, opened 2019-12-11
fix: customer price list being overidden by default price list

0Pranav:
Issue: When creating a quotation from Customer form. Correct Selling price list would not be fetched, instead the default price list would be applied. GIF below

quotation issued wrong

Cause: `party_name` trigger wasn't being triggered onload.

Fix: Add `party_name` trigger in `onload`. GIF Below

quotation issue fixed

Notes:

Also changed `this.frm.doc.__islocal` to `this.frm.is_new()` as it is more semantic.

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -169,7 +169,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	onload: function() {
 		var me = this;
 
-		if(this.frm.doc.__islocal) {
+		if(this.frm.is_new()) {
 			var currency = frappe.defaults.get_user_default("currency");
 
 			let set_value = (fieldname, value) => {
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -183,6 +183,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				() => set_value('price_list_currency', currency),
 				() => set_value('status', 'Draft'),
 				() => set_value('is_subcontracted', 'No'),
+				() => {
```
deepeshgarg007, line 186: Changes should not be made in the transaction controller just for one specific use case. This might trigger unwanted changes in other transactions as well. All these contextual requirements should be handled in a custom make method. One such method for creating quotation from customer dashboard already exists.

deepeshgarg007 (changes requested):

erpnext/public/js/controllers/transaction.js
0Pranav: Passing the data contextually will only solve this use case, but any customizations that depend on the `party_name` trigger will still fail.

0Pranav:
Found issue to be deeper. Will re-open PR with a better solution.

PR #19903, opened 2019-12-11
fix: customer price list being overidden by default price list

0Pranav:
backport of https://github.com/frappe/erpnext/pull/19902

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -169,7 +169,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	onload: function() {
 		var me = this;
 
-		if(this.frm.doc.__islocal) {
+		if(this.frm.is_new()) {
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -183,6 +183,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				() => set_value('price_list_currency', currency),
 				() => set_value('status', 'Draft'),
 				() => set_value('is_subcontracted', 'No'),
+				() => {
+					if(this.frm.doc.party_name) { return this.frm.trigger('party_name'); } // pricelist selection
```
rohitwaghchaure, line 187: party_name field only exists in the quotation form and not in any other sales/purchases doctypes. So instead of adding this code in transaction.js file you can add it in the quotation.js file.

0Pranav:
Found issue to be deeper. Will re-open PR with a better solution.

PR #19911, opened 2019-12-11
feat: allow adding & deleting of items in submitted SO & PO

nextchamp-saqib:
Depends on: https://github.com/frappe/erpnext/pull/19994
Above PR fixes deletion of rows from a table displayed in a dialog.


deepeshgarg007:
@nextchamp-saqib Add some basic test cases for it

erpnext/controllers/accounts_controller.py
```diff
@@ -1163,6 +1163,26 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
 	sales_doctypes = ['Sales Order', 'Sales Invoice', 'Delivery Note', 'Quotation']
 	parent = frappe.get_doc(parent_doctype, parent_doctype_name)
 
+	deleted_childs = []
+	if len(parent.items) > len(data):
```
nabinhait: - remove this condition
- move deletion functionality to a new function
- deleted_children

nabinhait (changes requested):

nabinhait (approved):

PR #19919, opened 2019-12-11
fix: task validation error when adding tasks to projects

benknowles:
When adding a task to a project, if the project didn't have an Expected End Date the validation would fail.

This is because passing a None value to getdate() returns today's date, rather than being optional as expected.

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -47,9 +47,10 @@ class Task(NestedSet):
 		if not self.project or frappe.flags.in_test:
 			return
 
-		expected_end_date = getdate(frappe.db.get_value("Project", self.project, "expected_end_date"))
+		expected_end_date = frappe.db.get_value("Project", self.project, "expected_end_date")
 
 		if expected_end_date:
+			expected_end_date = getdate(expected_end_date)
 			validate_project_dates(expected_end_date, self, "exp_start_date", "exp_end_date", "Expected")
 			validate_project_dates(expected_end_date, self, "act_start_date", "act_end_date", "Actual")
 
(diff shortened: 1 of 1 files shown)


erpnext/projects/doctype/task/task.py
```diff
@@ -47,9 +47,10 @@ def validate_parent_project_dates(self):
 		if not self.project or frappe.flags.in_test:
 			return
 
-		expected_end_date = getdate(frappe.db.get_value("Project", self.project, "expected_end_date"))
+		expected_end_date = frappe.db.get_value("Project", self.project, "expected_end_date")
 
 		if expected_end_date:
+			expected_end_date = getdate(expected_end_date)
```
hrwX: Pass the argument as below
```suggestion
			validate_project_dates(getdate(expected_end_date), self, "act_start_date", "act_end_date", "Actual")
```

erpnext/projects/doctype/task/task.py
```diff
@@ -47,9 +47,10 @@ def validate_parent_project_dates(self):
...
 			validate_project_dates(expected_end_date, self, "exp_start_date", "exp_end_date", "Expected")
```
hrwX: Pass the argument as below
```suggestion
			validate_project_dates(getdate(expected_end_date), self, "act_start_date", "act_end_date", "Actual")
```

hrwX (changes requested):

hrwX:
@benknowles Also open a new PR for the fix for version-12-hotfix branch

erpnext/projects/doctype/task/task.py
Thunderbottom: there's already a validation for actual task dates below this.

hrwX (approved):

benknowles:
@rohitwaghchaure Could you merge this PR please, related to #19936 in develop

PR #19932, opened 2019-12-12
fix: added validation for cost center

0Pranav:
Issue: No validation for cost center for item in item_list, to check if the cost center set for each item belongs to the company selected. If user changes company of sales invoice after saving the doc with items, the error will be thrown from GL entry which is less contextual.

Fix: add validation for cost centers in sales invoice itself.

Error message:

Screenshot 2019-12-18 at 11 04 01 AM

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -90,6 +90,7 @@ class SalesInvoice(SellingController):
 		self.validate_account_for_change_amount()
 		self.validate_fixed_asset()
 		self.set_income_account_for_fixed_assets()
+		self.validate_item_cost_centers()
 		validate_inter_company_party(self.doctype, self.customer, self.company, self.inter_company_invoice_reference)
 
 		if cint(self.is_pos):
@@ -146,6 +147,12 @@ class SalesInvoice(SellingController):
 
 					elif asset.status in ("Scrapped", "Cancelled", "Sold"):
 						frappe.throw(_("Row #{0}: Asset {1} cannot be submitted, it is already {2}").format(d.idx, d.asset, asset.status))
+
+	def validate_item_cost_centers(self):
+		for item in self.get("items",None):
+			cc = frappe.get_doc("Cost Center",item.cost_center)
+			if cc.company != self.company:
+				frappe.throw(_("Row #{0} : Cost Center <strong>{1}</strong> does not belong to company <strong>{2}</strong>").format(item.idx, item.cost_center, self.company))
 
 	def before_save(self):
 		set_account_for_mode_of_payment(self)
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -147,6 +148,12 @@ def validate_fixed_asset(self):
 					elif asset.status in ("Scrapped", "Cancelled", "Sold"):
 						frappe.throw(_("Row #{0}: Asset {1} cannot be submitted, it is already {2}").format(d.idx, d.asset, asset.status))
 
+	def validate_item_cost_centers(self):
```
deepeshgarg007, line 151: 1. "Row #{0} in Item list" is not required, just "Row #{0}" is necessary. This is the standard pattern we have used everywhere
2.  Cost center name and company name should be bold

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
0Pranav: @deepeshgarg007 fixed formatting. But I still think "in Item list" should remain? there's a couple of child tables in sales invoice, so it might be hard to understand which row we are talking about.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -147,6 +148,12 @@ def validate_fixed_asset(self):
...
+		for item in self.get("items",None):
```
nabinhait: It will give error `None is not iterable` is item table is blank. You can simply use `self.items`.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -147,6 +148,12 @@ def validate_fixed_asset(self):
...
+			cc = frappe.get_doc("Cost Center",item.cost_center)
```
nabinhait: get_doc is expensive and you are calling it inside for loop. Instead fetch company of cost_center using `frappe.get_cached_value`, as in most cases cost center will be same for all items.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -147,6 +148,12 @@ def validate_fixed_asset(self):
...
+			if cc.company != self.company:
+				frappe.throw(_("Row #{0} : Cost Center <strong>{1}</strong> does not belong to company <strong>{2}</strong>").format(item.idx, item.cost_center, self.company))
```
nabinhait: Don't use html tag inside message, it will create problem while translating. Instead use frappe.bold(item.cost_center)`.

nabinhait: no space before `:`

nabinhait (changes requested):

nabinhait (approved):

PR #19944, opened 2019-12-13
fix: gl entries doesn't filter based on debit precision

nextchamp-saqib:


erpnext/accounts/general_ledger.py
```diff
@@ -90,8 +90,12 @@ def merge_similar_entries(gl_map):
 		else:
 			merged_gl_map.append(entry)
 
+	company = gl_map[0].company if gl_map else erpnext.get_default_company()
+	company_currency = erpnext.get_company_currency(company)
+	precision = get_field_precision(frappe.get_meta("GL Entry").get_field("debit"), company_currency)
+
 	# filter zero debit and credit entries
-	merged_gl_map = filter(lambda x: flt(x.debit, 9)!=0 or flt(x.credit, 9)!=0, merged_gl_map)
```
rohitwaghchaure, line 94: I think this was intentionally done, what's the issue?

rohitwaghchaure (approved):

PR #19955, opened 2019-12-16
feat: Capture return amount against Employee Advance via Journal Entry

deepeshgarg007:
When some part of Employee Advance is unused and returned back and a Journal Entry is made for the same even then the return amount is not reflected in Employee Advance and the same gets fetched in Expense Claim. 

Screenshot 2019-12-16 at 1 46 33 PM


rushikesherp:
@deepeshgarg007 , Is there  any functionality to return the return amount against this advance? Does it automatically pass journal entry for return and mark the advance entry as returned so that it cannot be adjusted against another claim?

deepeshgarg007:
> @deepeshgarg007 , Is there any functionality to return the return amount against this advance? Does it automatically pass journal entry for return and mark the advance entry as returned so that it cannot be adjusted against another claim?

@rushikesherp Within Journal Entry itself you'll have to mention the reference type and reference name and the return amount will be updated in Employee Advance and  it won't be considered in Expense Claim

image

erpnext/hr/doctype/employee_advance/employee_advance.json
```diff
@@ -1,737 +1,213 @@
...
+  {
+   "fieldname": "mode_of_payment",
+   "fieldtype": "Link",
+   "label": "Mode of Payment",
+   "options": "Mode of Payment"
+  },
+  {
+   "fieldname": "return_amount",
+   "fieldtype": "Currency",
+   "label": "Return Amount",
```
nabinhait: Returned Amount?

nabinhait (changes requested):

PR #19958, opened 2019-12-16
fix: none type object updates a row in accounts receivable report

nextchamp-saqib:
Issue:

Screenshot 2019-12-20 at 2 59 25 PM

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
@@ -285,7 +285,7 @@ class ReceivablePayableReport(object):
 
 	def set_party_details(self, row):
 		# customer / supplier name
-		party_details = self.get_party_details(row.party)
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@nextchamp-saqib Can you update a screenshot in the description what issue the non type object was causing

erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -285,7 +285,7 @@ def get_invoice_details(self):
 
 	def set_party_details(self, row):
 		# customer / supplier name
-		party_details = self.get_party_details(row.party)
+		party_details = self.get_party_details(row.party) or {}
```
nextchamp-saqib, line 288: For some reason (probably due to customer's data) `get_party_details` was returning `None`

nabinhait (approved):

PR #19035, opened 2019-09-13
feat: Loan management and accounting

deepeshgarg007:
Issue for the feature request: https://github.com/frappe/erpnext/issues/18997

Documentation PR: https://github.com/frappe/erpnext_com/pull/573

PR #19978, opened 2019-12-17
fix: update loan on salary slip submission and cancellation.

Anurag810:
It was not updating the repayment schedule, total payment, and Total paid amount on submitting the salary slip. 
image


deepeshgarg007:
@Anurag810 move this PR to version-12-hotfix. After this https://github.com/frappe/erpnext/pull/19035 is merged all changes you have made will be obsolete

PR #19983, opened 2019-12-18
fix: use open_mapped_doc instead of create_new_doc

0Pranav:
Issue: `erpnext.utils.create_new_doc` method used in Customer dashboard would not trigger fetches in quotation form. Similar issue seems to have been there. Similar issue was fixed in past in Lead dashboard using `open_mapped_doc`

Fix: Use `open_mapped_doc` instead of instead of `erpnext.utils.create_new_doc`

### DIFF
--- a/erpnext/selling/doctype/customer/customer.py
+++ b/erpnext/selling/doctype/customer/customer.py
@@ -258,7 +258,10 @@ def make_quotation(source_name, target_doc=None):
 	target_doc.run_method("set_other_charges")
 	target_doc.run_method("calculate_taxes_and_totals")
 
-	target_doc.selling_price_list = frappe.get_doc("Customer",source_name).default_price_list
+	price_list = frappe.get_value("Customer",source_name, 'default_price_list')
+	if price_list:
+		target_doc.selling_price_list = price_list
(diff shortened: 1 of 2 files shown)


erpnext/selling/doctype/customer/customer.py
```diff
@@ -238,6 +239,64 @@ def create_contact(contact, party_type, party, email):
...
+				"name":"party_name"
+			}
+		}}, target_doc, set_missing_values)
+
+	target_doc.quotation_to = "Customer"
+	target_doc.run_method("set_missing_values")
+	target_doc.run_method("set_other_charges")
+	target_doc.run_method("calculate_taxes_and_totals")
+
+	target_doc.selling_price_list = frappe.get_doc("Customer",source_name).default_price_list
```
deepeshgarg007: Why get the entire doc just to get the price list use `get_value` instead. Also override the default selling price list only if default price list for customer is available otherwise let it fetch the default price list.

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #19993, opened 2019-12-18
enhancement: Added Set Reserve Warehouse field to Purchase Order

marination:
Port of https://github.com/frappe/erpnext/pull/19991
- **Supplied Items** table in **Purchase Order** requires manual entering of **Reserve Warehouse** if it is not set in the Item's Defaults or in the BOM.
- Added a field **Set Reserve Warehouse** to set all **Reserve Warehouse** fields of child table **Supplied Items** in one go.
- **Reserve Warehouse** will be visible only in presence of Supplied Items table

**Working**:

_reserve_warehouse


nabinhait (approved):

nabinhait:
Let's do not add new feature in v11, closing this.

PR #20009, opened 2019-12-19
fix: bad credit to filter in purchase invoice

nextchamp-saqib:


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -382,21 +382,11 @@ cur_frm.fields_dict['items'].grid.get_field("item_code").get_query = function(do
 
 cur_frm.fields_dict['credit_to'].get_query = function(doc) {
 	// filter on Account
-	if (doc.supplier) {
```
deepeshgarg007, line 385: Same filter is there in sales_invoice.js as well. Change it there as well for consistency.

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #20013, opened 2019-12-19
fix: no role has cancelling permission for share transfer doctype

nextchamp-saqib:


erpnext/accounts/doctype/share_transfer/share_transfer.json
```diff
@@ -188,14 +188,15 @@
...
- "modified": "2019-11-07 13:31:17.999744",
+ "modified": "2019-12-19 13:15:59.001301",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Share Transfer",
  "owner": "Administrator",
  "permissions": [
   {
    "amend": 1,
+   "cancel": 1,
```
deepeshgarg007, line 199: Give submission and cancel permission to accounts manager as well

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #20023, opened 2019-12-20
fix(refactor): loan in HRMS

Anurag810:

> Screenshots/GIFs


erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -815,6 +819,23 @@ def update_status(self, salary_slip=None):
 				timesheet.set_status()
 				timesheet.save()
 
+	def update_loans(self, status):
+		for loan in self.get_loan_details():
+			doc = frappe.get_doc("Loan", loan.name)
+			if status == "Submit":
```
nabinhait: Can be based on docstatus, no need to pass args

erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -815,6 +819,23 @@ def update_status(self, salary_slip=None):
...
+				frappe.db.set_value("Repayment Schedule", loan.repayment_name, "paid", 1)
+				doc.reload()
+				doc.total_amount_paid += loan.total_payment
```
nabinhait: We should define a function inside loan.py which will calculate total_payment based on repayment schedule. And that function should be called from here. In this way, it will be reusable function to use it for patching (also pull is better than push).

erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -815,6 +819,23 @@ def update_status(self, salary_slip=None):
...
+				total_pay = doc.total_payment - loan.total_payment
+			elif status == "Cancel":
+				frappe.db.set_value("Repayment Schedule", loan.repayment_name, "paid", 0)
+				doc.reload()
+				doc.total_amount_paid -= loan.total_payment
+				total_pay = doc.total_payment + loan.total_payment
+
+			doc.submit()
```
nabinhait: `submit()` is misleading as loan is already submitted. What's the need of this?

nabinhait (changes requested):

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: done

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: done

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: done

Anurag810:
@nabinhait, please merge

nabinhait (approved):

nabinhait (commented):
@Anurag810 Please send another PR for develop branch

PR #20058, opened 2019-12-23
fix: Allow creation of multiple landed cost voucher against a Purchase Document

### LINKED ISSUE #20031: V12.2.2 or V12.2.3 Unable to post second Landed Cost Voucher

**Output of `bench version`**
```
erpnext 12.2.2
frappe 12.0.20
```

If you use the newer version please perform the following steps and you will get to know what we are talking about here.

1) Create a purchase invoice
2) Create a landed cost voucher and allocate it to that purchase invoice
3) Create another landed cost voucher, you will not be able to do so, as the system will through this message and you will be stuck. The difference it shows is exactly the same amount on the landed cost voucher, that was entered first.

You cannot cancel any landed cost voucher neither can you are able to create another one.

All data entered validated through normal process without any loss of its integrity.

### Observed result
Debit and Credit not equal for Purchase Invoice **invoice number where first LCV was allocated**. Difference is **value of first LCV allocated.**

Screen Shot 2019-12-20 at 1 29 06 PM

MAT-LCV-2019-00001 was the first LCV posted in sum of 10000.0 to the PINV-2019-00001. Now we are trying to allocate the second LCV it is producing the message as appearing in the above image. Now we cannot create another LCV nor cancel the previous LCV.

### Expected result
Landed Cost Voucher (LCV) should be posted or cancelled. We should also be able to post more LCVs

```
none
```

## Additional information

Mac OSx, Windows 10 etc., `ERPNext` install method **1) Easy from github i.e. frappe/bench, 2) from Bitnami**.

Please see following discussion for more information:
https://discuss.erpnext.com/t/landed-cost-voucher-on-submit-debit-and-credit-not-equal-for-purchase-invoice-pinv-01267-1-difference-is-420-0/55759?u=majid_ejaz

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/20031

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -88,7 +88,7 @@ class PurchaseReceipt(BuyingController):
 
 		if getdate(self.posting_date) > getdate(nowdate()):
 			throw(_("Posting Date cannot be future date"))
-	
+
 	def validate_cwip_accounts(self):
 		for item in self.get('items'):
 			if item.is_fixed_asset and is_cwip_accounting_enabled(item.asset_category):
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

majidejaz:
how can I affect my system?

I ran the command **bench update --patch** but it does not affect anything.

i am running this command on google cloud VM

PR #20082, opened 2019-12-25
fix: unpaid invoices are created even if current invoice is paid

nextchamp-saqib:
Port of #20081


nabinhait (approved):

mujeerhashmi:
when will this be merged to version-12

PR #20088, opened 2019-12-26
fix: undefined dr_or_cr_notes in case of party type Employee

nextchamp-saqib:
Back Port of #20087


deepeshgarg007 (approved):

erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
```diff
@@ -23,6 +23,8 @@ def get_nonreconciled_payment_entries(self):
 
 		if self.party_type in ["Customer", "Supplier"]:
 			dr_or_cr_notes = self.get_dr_or_cr_notes()
+		else:
+			dr_cr_notes = []
```
deepeshgarg007:
```suggestion
			dr_or_cr_notes = []
```

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #20105, opened 2019-12-27
fix: service item's cost showing incorrect amount in stock entry

rohitwaghchaure:
**Issue**
Service items cost = 500
Qty to manufature = 5
Amount showing as = 500
Expected amount = 5 * 500 = 2500

Screenshot 2019-12-27 at 11 41 13 AM

**After Fix**

image

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -767,7 +767,7 @@ def add_additional_cost(stock_entry, work_order):
 
 	items = {}
 	for d in bom.get(table):
-		items.setdefault(d.item_code, d.rate)
+		items.setdefault(d.item_code, d.amount)
 
 	non_stock_items = frappe.get_all('Item',
 		fields="name", filters={'name': ('in', list(items.keys())), 'ifnull(is_stock_item, 0)': 0}, as_list=1)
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -776,7 +776,7 @@ def add_additional_cost(stock_entry, work_order):
 		stock_entry.append('additional_costs', {
 			'expense_account': expenses_included_in_valuation,
 			'description': name[0],
-			'amount': items.get(name[0])
+			'amount': flt(items.get(name[0])) * flt(stock_entry.fg_completed_qty)
```
nabinhait: The service item cost should be applied proportionately.

`service_item_cost * stock_entry.fg_completed_qty / bom.qty`

nabinhait (changes requested):

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -776,7 +776,7 @@ def add_additional_cost(stock_entry, work_order):
 		stock_entry.append('additional_costs', {
 			'expense_account': expenses_included_in_valuation,
 			'description': name[0],
-			'amount': items.get(name[0])
+			'amount': flt(items.get(name[0])) * flt(stock_entry.fg_completed_qty) / flt(bom.quantity)
```
nabinhait, line 779: `items.get(name[0])` gives only rate, not amount

nabinhait (dismissed):

nabinhait (approved):

PR #20106, opened 2019-12-27
fix: service item's cost showing incorrect amount in stock entry

rohitwaghchaure:
Port https://github.com/frappe/erpnext/pull/20105

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -768,7 +768,7 @@ def add_additional_cost(stock_entry, work_order):
 
 	items = {}
 	for d in bom.get(table):
-		items.setdefault(d.item_code, d.rate)
+		items.setdefault(d.item_code, d.amount)
 
 	non_stock_items = frappe.get_all('Item',
 		fields="name", filters={'name': ('in', list(items.keys())), 'ifnull(is_stock_item, 0)': 0}, as_list=1)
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -777,7 +777,7 @@ def add_additional_cost(stock_entry, work_order):
 		stock_entry.append('additional_costs', {
 			'expense_account': expenses_included_in_valuation,
 			'description': name[0],
-			'amount': items.get(name[0])
+			'amount': flt(items.get(name[0])) * flt(stock_entry.fg_completed_qty)
```
nabinhait: The service item cost should be applied proportionately.

`service_item_cost * stock_entry.fg_completed_qty / bom.qty`

nabinhait (changes requested):

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -777,7 +777,7 @@ def add_additional_cost(stock_entry, work_order):
 		stock_entry.append('additional_costs', {
 			'expense_account': expenses_included_in_valuation,
 			'description': name[0],
-			'amount': items.get(name[0])
+			'amount': flt(items.get(name[0])) * flt(stock_entry.fg_completed_qty) / flt(bom.quantity)
```
nabinhait, line 780: items.get(name[0]) gives only rate, not amount

nabinhait (changes requested):

nabinhait (approved):

PR #20119, opened 2019-12-27
fix: incorrect batch fetched for the serialized items

rohitwaghchaure:
**Issue**
Steps to replicate the issue

1. Create an Item and enabled has serial no and has batch no

1. Make 2 purchase receipts of quantity 10, so that system will create 20 serial nos and 2 batches

1. Now make stock entry with purchase as Material Transfer and select the serial nos of second batch and keep the batch no field blank. System will fetch the 1st batch which is not belongs to the selected serial no

**After Fix**

System will fetch the batch based on the serial nos. If system found more than one batch then user has to manually select the batch also system which check the selected batch is belong to selected serial nos or not

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -46,14 +46,13 @@ class StockController(AccountsController):
 	def validate_serialized_batch(self):
 		from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
 		for d in self.get("items"):
-			if hasattr(d, 'serial_no') and not (d.serial_no and d.batch_no): continue
-
-			serial_nos = get_serial_nos(d.serial_no)
-			for serial_no_data in frappe.get_all("Serial No",
-				filters={"name": ("in", serial_nos)}, fields=["batch_no", "name"]):
-				if serial_no_data.batch_no != d.batch_no:
-					frappe.throw(_("Row #{0}: Serial No {1} does not belong to Batch {2}")
-						.format(d.idx, serial_no_data.name, d.batch_no))
+			if hasattr(d, 'serial_no') and hasattr(d, 'batch_no') and d.serial_no and d.batch_no:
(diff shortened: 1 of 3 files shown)


erpnext/controllers/stock_controller.py
```diff
@@ -42,6 +43,18 @@ def make_gl_entries(self, gl_entries=None, repost_future_gle=True, from_repost=F
...
+	def validate_serialized_batch(self):
+		from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
+		for d in self.get("items"):
+			if not (d.serial_no and d.batch_no): continue
+
+			serial_nos = get_serial_nos(d.serial_no)
+			for serial_no_data in frappe.get_all("Serial No",
+				filters={"name": ("in", serial_nos)}, fields=["batch_no", "name"]):
+				if serial_no_data.batch_no != d.batch_no:
+					frappe.throw(_("The batch {0} does not belog to serial no {1}")
```
nabinhait: The reverse validation should be added, as batch is superset.
"Row #{0}: Serial No {1} does not belong to Batch {2}"

nabinhait (changes requested):

nabinhait (dismissed):

nabinhait (approved):

Alchez:
@rohitwaghchaure @nabinhait 

Can we port this to hotfix branches as well please?

The fix made at

https://github.com/frappe/erpnext/blob/92b933f4acd14520a35ca80c395b1c682278a14e/erpnext/controllers/stock_controller.py#L50

is required to avoid problems in the validation for Sales Order documents, which doesn't have the `batch_no` field.

For reference, the current v12 branch has the following code:

https://github.com/frappe/erpnext/blob/7ed7c3237a523e219f25d347033ee89d7e3a1613/erpnext/controllers/stock_controller.py#L50

The other changes made in this PR might be relevant as well.

PR #20145, opened 2019-12-31
feat: Payroll based on attendance or leave

Anurag810:
We can set the payroll based on **attendance/Leave** from **HR settings**. As shown below.

HR SETTINGS

Salary slip will consider:
 
- **Leave without pay** as non-working days, if payroll based on is set to **Leave**.
-  **Absent days** as non-working days, if payroll based on is set to Attendance and it will consider attendance record with status **Absent** and **Half day**. The **half-day** is considered **half-day absent** during the creation of a salary slip.

Please suggest any enhancement required for this feature.
And can we provide this feature for multiple companies?


nabinhait:
@Anurag810 please resolve conflict

erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -193,21 +197,37 @@ def get_leave_details(self, joining_date=None, relieving_date=None, lwp=None, fo
...
 
 		holidays = self.get_holidays_for_employee(self.start_date, self.end_date)
-		actual_lwp = self.calculate_lwp(holidays, working_days)
+		non_working_days = 0
+		if payroll_based_on == 'Leave':
+			actual_lwp = self.calculate_lwp(holidays, working_days)
+		elif payroll_based_on == 'Attendance':
+			actual_absent_days = self.calculate_absents()
+		else:
+			frappe.throw(_("Please set <b>Payroll based on</b> in <b>HR settings</b>"))
```
nabinhait: Don't use HTML tags inside message, it creates problem in translation

erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -272,6 +292,25 @@ def calculate_lwp(self, holidays, working_days):
 				lwp = cint(leave[0][1]) and (lwp + 0.5) or (lwp + 1)
 		return lwp
 
+	def calculate_absents(self):
+		absent_days = 0
+
+		attendances = frappe.db.sql('''SELECT status, count(*) as non_working_days
+			FROM `tabAttendance`
+			WHERE
+				status NOT IN ("Present","On leave")
```
nabinhait, line 308: "On leave" with leave_type as LWP should also be treated as absent.

erpnext/hr/doctype/salary_slip/salary_slip.py
```diff
@@ -272,6 +292,25 @@ def calculate_lwp(self, holidays, working_days):
...
+				AND employee = "{0}"
```
nabinhait: This will give error if employee value has '"', instead write like `AND employee = %s`

nabinhait (changes requested):
- Add docs
- Add tests
- Write a patch to set `payroll_based_on` as 'Leave' in existing account.

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: Done

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: Done

erpnext/hr/doctype/salary_slip/salary_slip.py
Anurag810: Done

Olawale1:
Hi @Anurag810 

Thanks a lot for taking this up. I think it's important to add a field in HR Settings where the user can specify the fraction to be used for Half Day attendance. This is simply because many companies use the Half Day attendance to mark lateness to work and may not necessarily want that to deduct a half day's wages. The value could however be set as 0.5 by default

Cheers

erpnext/hr/doctype/salary_slip/salary_slip.py
nabinhait: "On leave" with leave_type as LWP?

nabinhait (changes requested):

nabinhait:
> Hi @Anurag810
> 
> Thanks a lot for taking this up. I think it's important to add a field in HR Settings where the user can specify the fraction to be used for Half Day attendance. This is simply because many companies use the Half Day attendance to mark lateness to work and may not necessarily want that to deduct a half day's wages. The value could however be set as 0.5 by default
> 
> Cheers

To mark late attendance, you should use a custom status or a custom field altogether. If you mix it up, then it can cause error in many places.

Olawale1:
> > Hi @Anurag810
> > Thanks a lot for taking this up. I think it's important to add a field in HR Settings where the user can specify the fraction to be used for Half Day attendance. This is simply because many companies use the Half Day attendance to mark lateness to work and may not necessarily want that to deduct a half day's wages. The value could however be set as 0.5 by default
> > Cheers
> 
> To mark late attendance, you should use a custom status or a custom field altogether. If you mix it up, then it can cause error in many places.

Hi @nabinhait 

I don't see how this will cause any errors anywhere else. What I am suggesting is for this feature only and I think it's quite critical too. Maybe I should try to explain a little better:

With this feature in it's current state, let's assume there are 20 Working Days and the employee's salary is $2000. If he/she is marked as Present for 18 Days and Half Day for 2 Days, the system will calculate salary as ((18+0.5+0.5)/20) x 2000 = $1900

The issue however is that Half Day in Attendance doesn't always mean that the Employee worked for exactly half of the working hours. It is mostly generically used to show that the Employee did not work full hours (lateness) on that day and hence, some organizations do not penalize the Employee for exactly half of the day's wages

Using my suggestion above, If Payroll based on 'Attendance' is selected in HR Settings, then an additional field appears to confirm fraction (or percentage) of Daily Wages to be deducted for Half Day attendance (0.5 or 50% by default). If the user adjusts this value to let's say 25%, the salary of the Employee in the earlier example is then calculated as ((18+0.75+0.75)/20) x 2000 = $1950

It's just a simple tweak in the calculation (factoring in the percentage specified for Half Day penalty), that's all! Using custom fields on the other hand would be much more complex and will not work with the feature unless the core code is hacked!

Kindly consider this

Thanks

Anurag810:
@nabinhait please merge

michellealva:
@Anurag810 what do we do if we want payroll to be based on both attendance and leave?

Anurag810:
> @Anurag810 what do we do if we want payroll to be based on both attendance and leave?
@michellealva 
There is an option to set attendance/leave in HR settings.

Anurag810:
Added tests

fnrfarid:
Hi,

I would like to suggest to develop an option/automation that 5 [x] days late should be considered as an absent or leave without pay. Also, there should be some configurable slots like, for 5 days, 1 LWP, for 6-9 days, 2 days LWP, for 10-14 days, 3 days LWP etc. 

Late attendances should result as absent or leave without pay as many companies follow this policy.

fnrfarid:
> 
> 
> @Anurag810 what do we do if we want payroll to be based on both attendance and leave?

@Anurag810 this is actually a valid requirement. I would like to process the payroll based on attendance, as well as based on leave without pay too. I guess, if we select "based on attendance" then, other leave without pay could be managed manually! then again, it would be a hassle.

Olawale1:
> I would like to suggest to develop an option/automation that 5 [x] days late should be considered as an absent or leave without pay. Also, there should be some configurable slots like, for 5 days, 1 LWP, for 6-9 days, 2 days LWP, for 10-14 days, 3 days LWP etc.
> 
This just buttresses my point about not making the Half Day attendance a rigid 50% Daily Wage deduction. It needs to be a variable set by the user

Kind regards,

Muad96:
I'd suggest to make **Absent Days** as Salary Component [Deduction] than can be configured from HR Settings. This is to avoid confusion when auditing/trace back the **LWP** days when **LWP** calculate from both **Leave Application** & **Attendance** Doctype.

Anurag810:
@nabinhait please review and merge

fnrfarid:
@anurag810, could you kindly share the latest updates on this?

nabinhait:
Create another refactored PR, closing this.
https://github.com/frappe/erpnext/pull/21258

PR #20149, opened 2019-12-31
feat: added provision to disable rounded total for salary slips

nextchamp-saqib:
Added a check to disable rounded totals in Salary Slips in HR Settings.

Issue:
lDpXAxi

### DIFF
--- a/erpnext/hr/doctype/hr_settings/hr_settings.py
+++ b/erpnext/hr/doctype/hr_settings/hr_settings.py
@@ -7,6 +7,8 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _
 from frappe.model.document import Document
+from frappe.utils import cint
+from frappe.custom.doctype.property_setter.property_setter import make_property_setter
 
 class HRSettings(Document):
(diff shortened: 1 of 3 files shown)


erpnext/hr/doctype/hr_settings/hr_settings.py
```diff
@@ -22,3 +24,12 @@ def validate_password_policy(self):
 		if self.email_salary_slip_to_employee and self.encrypt_salary_slips_in_emails:
 			if not self.password_policy:
 				frappe.throw(_("Password policy for Salary Slips is not set"))
+	
+	def on_update(self):
+		self.toggle_rounded_total()
+		frappe.clear_cache()
+
+	def toggle_rounded_total(self):
```
deepeshgarg007, line 32: This will only hide the rounded total field. If rounded total is disabled then net pay should not be rounded off and amount in words should also be equal to net pay in this case

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #20152, opened 2019-12-31
fix: customer credit limit label

nextchamp-saqib:
https://github.com/frappe/erpnext/pull/19162#issuecomment-569630573


nabinhait (approved):

deepeshgarg007:
@nextchamp-saqib Also raise for version-12

PR #20174, opened 2020-01-03
fix: sales invoice status when credit note is created but not paid

0Pranav:
Add new status `Credit Note Issued and Paid`. When a credit note is created against a sales invoice and a payment entry for the return is also made this status will be applied to the Sales Invoice.

Screenshot 2020-01-03 at 11 42 53 AM


0Pranav:
Solution needed is different. will reopen with desired solution.

PR #20181, opened 2020-01-03
fix: Added description and title to supplier selection popup in Material Request

marination:
Port of https://github.com/frappe/erpnext/pull/20179
- On creating Purchase Order from Material Request, users entered random suppliers (which weren't default suppliers for any item in the MR), and often got a blank Supplier field in the new PO.
_POPup

- Added description to field in popup
Screenshot 2020-01-03 at 7 15 38 PM

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.js
+++ b/erpnext/stock/doctype/material_request/material_request.js
@@ -217,8 +217,13 @@ frappe.ui.form.on('Material Request', {
 
 	make_purchase_order: function(frm) {
 		frappe.prompt(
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@marination system should fetch the selected supplier in the purchase order, kindly remove the condition which checks the default supplier of the materials request items

marination:
> @marination system should fetch the selected supplier in the purchase order, kindly remove the condition which checks the default supplier of the materials request items

@rohitwaghchaure  That defeats the whole purpose of the popup. I think not redirecting them to PO if any random supplier is selected and showing some error message would be better?

nabinhait (approved):

PR #20201, opened 2020-01-06
feat: Backdated leave application 

vishdha:
Changes made:

Check for backdated leave application added
Only HR Manager can make backdated leave applications


vishdha:
https://github.com/DigiThinkIT/erpnext/pull/292

nabinhait (approved):

PR #20202, opened 2020-01-06
fix: escape % in customer name

0Pranav:
issue: authorization_control.py would raise `TypeError: not enough arguments for format string` if customer name contained a "%" sign

### DIFF
--- a/erpnext/setup/doctype/authorization_control/authorization_control.py
+++ b/erpnext/setup/doctype/authorization_control/authorization_control.py
@@ -76,7 +76,7 @@ class AuthorizationControl(TransactionBase):
 		add_cond = ''
 		auth_value = av_dis
 
-		if val == 1: add_cond += " and system_user = '"+session['user'].replace("'", "\\'")+"'"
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/authorization_control/authorization_control.py
```diff
@@ -85,7 +85,7 @@ def bifurcate_based_on_type(self, doctype_name, total, av_dis, based_on, doc_obj
 			if doc_obj:
 				if doc_obj.doctype == 'Sales Invoice': customer = doc_obj.customer
 				else: customer = doc_obj.customer_name
-				add_cond = " and master_name = '"+cstr(customer).replace("'", "\\'")+"'"
+				add_cond = " and master_name = '"+cstr(customer).replace("'", "\\'").replace("%","\\%")+"'"
```
nabinhait: Use `frappe.db.escape` instead

nabinhait (changes requested):

nabinhait (approved):

PR #20205, opened 2020-01-06
Update de.csv

wojosc:
Translations in modul HR


nabinhait (approved):

MarkWestside:
These (correct) changes aren’t in the latest build anymore.

Leaves has two meanings: tree leaves (German: Blätter) and leaves as used in ERPNext (which would be Urlaub in German).
Version 14 uses Blätter and not Urlaub.

It would be great to have Leaves translated with Urlaub again.

PR #20210, opened 2020-01-07
feat(Contacts): select billing contact for sales invoice

ronelvcabrera:
Reopening PR - https://github.com/frappe/erpnext/pull/19610

Most of the time, point of contact are different from sale, invoicing and on different aspect/phase of the transaction. This feature is to have contacts that are identified for billing.

Therefore, as seen below, when creating a sales invoice, the priority to retrieve a contact for Sales invoice are tagged as `is Billing Contact` contacts.

In connection with this PR https://github.com/frappe/frappe/pull/8840

### Usage:

Billing contact will be retrieved during creation of Invoice as point of contact at this stage is different from sale or other aspect/stage of the transaction.

core is billing contact new contact


erpnext/erpnext_integrations/custom/contact.json
```diff
@@ -0,0 +1,60 @@
+{
```
deepeshgarg007, line 1: Why is the `is_billing_contact` field exported as fixtures or a custom field. It should be in standard doctype right? Please fix and reopen

deepeshgarg007 (changes requested):

erpnext/erpnext_integrations/custom/contact.json
ronelvcabrera: @deepeshgarg007 I have that PR before and it is advised to have on ERPNext PR. As other changes on this does not cover frappe scope. It is more on ERPNext related scope.

Here is my PR before https://github.com/frappe/frappe/pull/8840

See comment here https://github.com/frappe/frappe/pull/8840#issuecomment-558469686

PR #20214, opened 2020-01-07
fix: Repay advance from salary

Anurag810:
Repay advance form salary.
Screenshot 2020-01-07 at 1 16 06 PM

feat-repay-from_salary-advance


Anurag810:
#20366 closing

PR #20216, opened 2020-01-07
fix: Parent Item of Product Bundle should not appear in dialog

0Pranav:
In dialog for creating work order or request for raw material from Sales Order.

Before:
OPB is the parent item of Product Bundle: Oneplus Bundle
Screenshot 2020-01-07 at 3 47 23 PM

after

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.py
+++ b/erpnext/selling/doctype/sales_order/sales_order.py
@@ -376,6 +376,9 @@ class SalesOrder(SellingController):
 	def get_work_order_items(self, for_raw_material_request=0):
 		'''Returns items with BOM that already do not have a linked work order'''
 		items = []
+		item_codes = [i.item_code for i in self.items]
+		product_bundle_parents = [pb.new_item_code for pb in frappe.get_all("Product Bundle", {"new_item_code": ["in", item_codes]}, ["new_item_code"])]
+
 		for table in [self.items, self.packed_items]:
 			for i in table:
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -387,7 +387,7 @@ def get_work_order_items(self, for_raw_material_request=0):
 				else:
 					pending_qty = stock_qty
 
-				if pending_qty:
+				if pending_qty and not frappe.get_value("Product Bundle", {"new_item_code": i.item_code}):
```
rohitwaghchaure: get_value in for loop will slow down the execution, better prepare the tuple of product bundle items based on self.items and then one by one check the item present in that tuple or not

nabinhait (approved):

franciscobuendiao:
What if Parent Item doesn’t have Product Bundle but has BOM? To request “raw materials” makes mandatory to use Manufacturing module.

https://discuss.erpnext.com/t/request-for-raw-materials-from-so-v12-4-2/57241

franciscobuendiao:
After the fix, it can only be requested raw materials for _items that maintain stock_ or have _product bundles_, what about _**service items**_?

BOMs aren’t useful not only to request raw material or manufacturing, but to also plan costs to do a "minor" service.

Sir, if I'm not considering something, any comment would be really helpful.
https://discuss.erpnext.com/t/request-for-raw-materials-from-so-v12-4-2/57241

PR #20251, opened 2020-01-12
feat: Group by item wise sales register

deepeshgarg007:
Depends On : https://github.com/frappe/frappe/pull/9236
Depends On: https://github.com/frappe/frappe/pull/9413
Docs: https://github.com/frappe/erpnext_com/pull/663

Screenshot 2020-01-12 at 9 51 16 PM

Screenshot 2020-02-07 at 11 05 04 AM


nabinhait (changes requested):
- Add "group by" for Customer group and Territory
- Refactor code using small functions
- Remove Item Code, Item Name column if grouped by Item (same with other group by)
- If grouped by customer and customer id is set by naming series, show "Customer + Customer Name" in Grouped total row.
- Show tax columns total
- "% of grand total" not showing value for each row and total is "Nan"
- Make common functions for both sales / purchase

nabinhait (approved):

Issue #19293, opened 2019-10-13
Multiple problems with Plaid Integration from Canada


jd4ever:
@chdecultot FYI there are multiple problems when trying to integrate with Plaid in Canada: 1) Canadian banks do not show up 2) Account selector does not show up and 3) python exception

**1. Canadian banks do not show up**
I have been following the guidelines on Plaid integration and everything works until this step: https://erpnext.com/docs/user/manual/en/erpnext_integration/plaid_integration#bank-account-creation

Canadian banks do not show up in the list and do not show up when searching for them.

Plaid does have support for Canadian banks but require the following to `access Canadian institutions, existing customers must include 'CA' within the countryCodes array during Link initialization`
https://blog.plaid.com/plaid-in-canada/

To work around this, I had to go into my Plaid account > Customize > Institution select and manually add my bank to the selections. Then I had to return to ERPnext to try again and select my Canadian bank on the main linking screen.

**2. Account selector does not show up**
ERPnext fails silently if user does not chose an account during Plaid linking step. For the Account selector to show up, I had to again go into my Plaid account. So, Plaid account > customize > Select Account. Then change radio button to "enabled for multiple accounts"

**3. Python exception**
After selecting account, I was taken back to ERPnext where I get to choose the organization to link it to. So far so good. But then as soon as I click continue, I get the following python error:

`Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/app.py", line 61, in application
    response = frappe.handler.handle()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 21, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/handler.py", line 56, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/__init__.py", line 1032, in call
    return fn(*args, **newargs)
  File "/home/frappe/frappe-bench/apps/erpnext/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py", line 59, in add_bank_accounts
    for account in response["accounts"]:
TypeError: string indices must be integers`

racitup:
I can confirm I get the same error as 3 when using Plaid sandbox in the UK with version:
ERPNext: v12.1.6 (version-12)
Frappe Framework: v12.0.16 (version-12)
[...]

PR #20254, opened 2020-01-12
fix: Plaid TypeError in add_bank_accounts #19293

racitup:
Fixes item 3 in #19293 

The problem appears to be a python2 vs 3 problem. This fix is a more pythonic way of solving the check. I developed on version-12 and tested manually using Plaid. Tried rebasing to version-12-hotfix but got conflicts on code I hadn't changed.


nabinhait (commented):
Thanks for the fix. 

Raised a duplicate PR [#20335](https://github.com/frappe/erpnext/pull/20335) because you branch is not properly rebased. Closing this.

PR #20260, opened 2020-01-13
feat: Added Exempted doctype list in hooks

vishdha:
Usecase: 
As a User, I want to be able to delete the linked document to the one I'm trying to delete, so that I don't have to manually delete the document(s) 

Problem: 
Anytime you try to cancel a document that is linked somewhere else, it doesn't allow you to do so. Which is obviously a good thing to maintain data integrity. However, we should allow the User to cancel both the documents on another button click, whilst warning them of the consequences.

Solution:

cancel

Pull Request Depends on : https://github.com/frappe/frappe/pull/8905

Documentation: https://github.com/frappe/frappe_io/pull/242


vishdha:
@surajshetty3416 Please merge this. This is important for cancel feature

surajshetty3416 (approved):

PR #20269, opened 2020-01-13
fix: allow on submit for the delivery date field

rohitwaghchaure:
Allow on submit property has enabled for the field Delivery Date in the sales order but not in the sales order item. Now when user change the delivery date on sales order after submission then system reset the delivery date based upon the date set in the child table


deepeshgarg007 (approved):

deepeshgarg007:
@rohitwaghchaure  PR for version-12-hotfix?

PR #20285, opened 2020-01-14
fix: incorrect required qty for subcontracting purchase receipt

rohitwaghchaure:
**Issue**

image

**After Fix**
image

Steps to replicate issue

1. Create 4 boms for different items and add same raw materials in all 4 boms

1. Create subcontracted PO against that 4 items

1. Do stock entry transfer

1. Create purchase receipt entries against PO partially 

1. First entry will calculate the correct required qty where as second will shows the incorrect qty

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -270,7 +270,7 @@ class BuyingController(StockController):
 
 			non_stock_items = get_non_stock_items(item.purchase_order, item.item_code)
 
-			item_key = '{}{}'.format(item.item_code, item.purchase_order)
+			item_key = '{}{}{}'.format(item.item_code, item.purchase_order, item.purchase_order_item)
 
 			fg_yet_to_be_received = qty_to_be_received_map.get(item_key)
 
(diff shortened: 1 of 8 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -880,122 +1064,72 @@ def validate_item_type(doc, fieldname, message):
...
-		backflushed_batch_qty_map.setdefault((batch_data.item_code, fg_item), {})
-		backflushed_batch_qty_map[(batch_data.item_code, fg_item)][batch_data.batch_no] = batch_data.qty
+def set_serial_nos(raw_material, consumed_serial_nos, qty):
+	transferred_serial_nos = set(get_serial_nos(raw_material.serial_nos))
+	consumed_serial_nos = set(get_serial_nos(consumed_serial_nos))
+	serial_nos = transferred_serial_nos - consumed_serial_nos
 
-	return backflushed_batch_qty_map
+	if serial_nos and qty <= len(serial_nos):
+		serial_nos = list(serial_nos)[0:frappe.utils.cint(qty)]
```
nabinhait: What's the need of `frappe.utils.cint`? qty should always have integer value for serialized items. 

if it gets null value for any reason, then `cint` is needed in the previous line as well.

nabinhait (changes requested):

nabinhait:
@rohitwaghchaure Closing this, let's redesign this as per our discussion.

PR #20298, opened 2020-01-15
fix: incorrect number of GL Entries error in stock entry

rohitwaghchaure:
Currency precision is 3 and stock difference value is 0.001
image

**Error**
image


nabinhait (approved):

wanderer434:
I have currency precision 3. After I updated to v12.4.0 I started getting message "Incorrect number of GL Entries" when I try to submit a stock entry of type Manufacture

PR #19709, opened 2019-11-27
feat(employee): make a configurable birthday email template

DeeMysterio:
**previous system:**
It sent a birthday reminder email to all the employees except the one celebrating their birthday.

Screenshot 2019-11-27 at 7 41 12 PM

**The new system:**

- It sends out an email to all the employees including the one celebrating their birthday.

- It sends out an email per birthday guy. 

- Added a field in "HR Settings" saying "Birthday Email Template" to select an Email Template which if set, is used by the system to send out an email and has all the context about the employee otherwise a hard coded email message and subject is used.
 
hr

- The Email Template that is used to be used by existing(patches) and new sites(install_fixtures):

Screenshot 2019-11-22 at 5 40 26 PM

- A birthday email sent using a template designed with the available employee context in the Email Template:

Screenshot 2019-11-22 at 3 33 04 PM

PR #20305, opened 2020-01-15
fix: retrieve proper value from HR Settings (v12)

Alchez:
Traceback:
```
	'HRSettings' has no attribute 'disable_rounded_total'
```

Migrate breaks at trying to retrieve value while running a v12 patch, but the column doesn't yet exist for existing sites

### DIFF
--- a/erpnext/hr/doctype/hr_settings/hr_settings.py
+++ b/erpnext/hr/doctype/hr_settings/hr_settings.py
@@ -24,12 +24,12 @@ class HRSettings(Document):
 		if self.email_salary_slip_to_employee and self.encrypt_salary_slips_in_emails:
 			if not self.password_policy:
 				frappe.throw(_("Password policy for Salary Slips is not set"))
-	
+
 	def on_update(self):
 		self.toggle_rounded_total()
 		frappe.clear_cache()
 
 	def toggle_rounded_total(self):
-		self.disable_rounded_total = cint(self.disable_rounded_total)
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/hr_settings/hr_settings.py
```diff
@@ -24,12 +24,12 @@ def validate_password_policy(self):
...
 				frappe.throw(_("Password policy for Salary Slips is not set"))
-	
+
 	def on_update(self):
 		self.toggle_rounded_total()
 		frappe.clear_cache()
 
 	def toggle_rounded_total(self):
-		self.disable_rounded_total = cint(self.disable_rounded_total)
+		self.disable_rounded_total = cint(self.get("disable_rounded_total"))
```
nabinhait, line 33: This is not the right fix. Ideally, we should write a patch to set disable_rounded_total based on global_defaults. And in that patch the field will be reloaded. We will add the patch.

By the way, which patch is breaking because of this?

nabinhait (commented):
Closing this.

erpnext/hr/doctype/hr_settings/hr_settings.py
Alchez: This is breaking for us on a fork, but the patch exists for review on a PR up at #19709.

More specifically, this file: https://github.com/frappe/erpnext/pull/19709/files#diff-4c931075e81c4c445c73fcf15dd55af1.

PR #20320, opened 2020-01-16
fix: Delivery note not able to create sales return

vishdha:
before:
before_sales_return

after: 
after_sales_return

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.py
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.py
@@ -436,7 +436,7 @@ def make_sales_invoice(source_name, target_doc=None):
 	def update_item(source_doc, target_doc, source_parent):
 		target_doc.qty = to_make_invoice_qty_map[source_doc.name]
 
-		if source_doc.serial_no and source_parent.per_billed > 0:
+		if source_doc.serial_no and source_parent.per_billed > 0 and not source_parent.is_return:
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/delivery_note/delivery_note.py
```diff
@@ -436,7 +436,7 @@ def set_missing_values(source, target):
 	def update_item(source_doc, target_doc, source_parent):
 		target_doc.qty = to_make_invoice_qty_map[source_doc.name]
 
-		if source_doc.serial_no and source_parent.per_billed > 0:
+		if source_doc.serial_no and source_parent.per_billed > 0 and not source_parent.is_return:
```
nabinhait, line 439: What's the use case for this change?

nabinhait (changes requested):

erpnext/stock/doctype/delivery_note/delivery_note.py
vishdha: When you do sales return from delivery note if item is serialized then it call for 			 get_delivery_note_serial_no where it try to set serial number but sales return qty is negative and why it should set serial number. so it sould not call this function. why would need to set new serial number again.

nabinhait: @vishdha This function is called when creating Sales Invoice from Delivery Note and it ensures that serial no selected in invoice is from DN. I think the same is also required when we create negative invoice (credit note) from return Delivery Note.

PR #20347, opened 2020-01-20
fix: deprecated fetching item price based on min_qty

nabinhait:


mimiza:
Hey why is this deprecated? I spent money for this fix.

nabinhait:
> Hey why is this deprecated? I spent money for this fix.

After merging the fix, it was re-fetching item price in Delivery Note / Sales Invoice on change of qty, even if we make those documents against Sales Order. Also, the logic of fetching price (based on qty) was not quite clear and buggy in case of different sales UOM.
Hence, we have decided to deprecate "Min Qty" field entirely from Item Price. For the purpose, we can use "Pricing Rule".

PR #20356, opened 2020-01-20
fix: Gives a 404 error when clicked on the Download Required Material…

### LINKED ISSUE #20355: Gives a 404 error when clicked on the Download Required Material button while the form is not saved.

1. Open Production Plan.
2. Click on the new button.
3. Click on the Download Required Material button without saving the document.

### Observed result
On clicking the button 404 error is showing.

### Expected result
A proper message should show like 

> "Please save the document before downloading the material required."

```
Status: 404 The resource you are looking for is not available
```

abhijitkumbharInd:
Gives a 404 error when clicked on the Download Required Material button while the form is not saved. 
Fixes: #20355

When we create a new Production Plan and without saving if we click on the Download Required Material button it is showing 404 error (The resource you are looking for is not available.)

ezgif-6-36540e8adc4a

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.js
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.js
@@ -218,6 +218,10 @@ frappe.ui.form.on('Production Plan', {
 	},
 
 	download_materials_required: function(frm) {
+		if(frm.is_new()) {
+			frappe.throw(__("Please save the document before downloading the material required."));
+			return;
+		}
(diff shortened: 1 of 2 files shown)


nabinhait (changes requested):
Instead hide the button before saving, using `depends_on` property.

`eval:!doc.__islocal`

abhijitkumbharInd:
@nabinhait as per your suggestion I updated the PR please check.

deepeshgarg007 (approved):

PR #20361, opened 2020-01-21
fix: account not added to child company

nextchamp-saqib:
Fixes: Adding Account to Parent Company doesn't adds it to its Child Companies.

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -95,29 +95,29 @@ class Account(NestedSet):
 		# ignore validation while creating new compnay or while syncing to child companies
 		if frappe.local.flags.ignore_root_company_validation or self.flags.ignore_root_company_validation:
 			return
-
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/account/account.py
```diff
@@ -95,29 +95,29 @@ def validate_root_company_and_sync_account_to_children(self):
...
 		else:
 			descendants = get_descendants_of('Company', self.company)
 			if not descendants: return
-
 			parent_acc_name_map = {}
 			parent_acc_name, parent_acc_number = frappe.db.get_value('Account', self.parent_account, \
 				["account_name", "account_number"])
-			for d in frappe.db.get_values('Account',
-				{ "company": ["in", descendants], "account_name": parent_acc_name, 
-					"account_number": parent_acc_number },
```
nextchamp-saqib, line 116: Problem: When no account number is found, the filter is set as "account_number" = None

nabinhait (approved):

PR #20364, opened 2020-01-21
fix: account not added to child company

nextchamp-saqib:
Back Port of #20361

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -95,29 +95,30 @@ class Account(NestedSet):
 		# ignore validation while creating new compnay or while syncing to child companies
 		if frappe.local.flags.ignore_root_company_validation or self.flags.ignore_root_company_validation:
 			return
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/account/account.py
```diff
@@ -95,29 +95,30 @@ def validate_root_company_and_sync_account_to_children(self):
...
-				["company", "name"], as_dict=True):
+			filters = { 
+				"company": ["in", descendants],
+				"account_name": parent_acc_name, 
+			}
+			if parent_acc_number:
+				filters["account_number"] = parent_acc_number
+
+			for d in frappe.db.get_values('Account', filters=filters, fieldname=["company", "name"], as_dict=True):
+				print(d["company"], d["name"])
```
nabinhait: remove print

nabinhait (changes requested):

nabinhait (approved):

PR #20366, opened 2020-01-21
feat: Repay advance from salary

Anurag810:
Repay advance form salary.
Screenshot 2020-01-07 at 1 16 06 PM

feat-repay-from_salary-advance


Anurag810:
Docs: https://github.com/frappe/erpnext_com/pull/657

erpnext/hr/doctype/employee_advance/employee_advance.js
```diff
@@ -47,19 +55,35 @@ frappe.ui.form.on('Employee Advance', {
...
+			}else if (frm.doc.repay_unclaimed_amount_from_salary == 1 && frappe.model.can_create("Additional Salary")){
+				frm.add_custom_button(__("Deduction from salary"),  function() {
+					frm.events.make_deduction_via_additional_salary(frm)
+				}, __('Create'));
+			}
 		}
 	},
 
+	make_deduction_via_additional_salary: function(frm){
+		console.log("Working");
```
nabinhait: remove console.log

nabinhait (changes requested):

Anurag810:
Please merge

nabinhait (changes requested):
- Don't add `payroll_date` and `salary_component` fields in Employee Advance. It will be difficult to enter accurate data in this field for the user while creating advance.
Instead, it should be decided while creating Additional Salary record. So, instead of submitting additional salary record automatically on click of button, it should open Additional Salary record after  mapping data. Then HR User will select payroll date and salary component and submit it.
- Also, add "Reference Document Type" and "Reference Document" fields in Additional Salary to track linked employee advance ID. Same should be implemented for "Leave encashment" and "Employee Incentive" and "Retention Bonus", currently those are reversed linked.

PR #20376, opened 2020-01-21
fix: Zero division error while making finished good entry against the…

rohitwaghchaure:
Backport https://github.com/frappe/erpnext/pull/20375


sorinpaul:
Hello!
Can you fix this problem also for ERPNext: v12.4.2 (version-12) - Frappe Framework: v12.2.1 (version-12)?

PR #20393, opened 2020-01-23
fix: unlink references from batch on cancellation of stock transactions

rohitwaghchaure:
Backport https://github.com/frappe/erpnext/pull/20392


deepeshgarg007 (commented):
@rohitwaghchaure The references were unliked on cancel but after amending the document the new doc was not linked with the batch

deepeshgarg007 (approved):

PR #20405, opened 2020-01-23
fix: able to close if depend task is cancelled

RJPvT:
this is not possible in base situation.


nabinhait (commented):
Closing this PR because of unwanted commits in it. Raised another PR to fix this.
https://github.com/frappe/erpnext/pull/20433

PR #20406, opened 2020-01-23
fix: cancelled is same as closed for dependent task

RJPvT:
in this situation we cannot close parent task because dependend task was cancelled. this makes no sense.


nabinhait (approved):

nabinhait (commented):
Closing this, as we don't accept PR directly on the version-11 branch.

PR #20410, opened 2020-01-24
fix: Finance book filtering in financial statements

deepeshgarg007:
**Problem:**
While applying a filter on finance book in financial statements the records with only financial book set are displayed

**Solution:**
After this fix on application of finance book filter records with no finance book will also be seen along with the filtered finance book records.

### DIFF
--- a/erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.py
+++ b/erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.py
@@ -389,9 +389,9 @@ def get_additional_conditions(from_date, ignore_closing_entries, filters):
 
 	if filters.get("finance_book"):
 		if filters.get("include_default_book_entries"):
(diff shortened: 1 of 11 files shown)


erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.py
```diff
@@ -389,9 +389,9 @@ def get_additional_conditions(from_date, ignore_closing_entries, filters):
 
 	if filters.get("finance_book"):
 		if filters.get("include_default_book_entries"):
-			additional_conditions.append("finance_book in (%(finance_book)s, %(company_fb)s)")
+			additional_conditions.append("ifnull(finance_book) in (%(finance_book)s, %(company_fb)s, '')")
```
nabinhait: Use "or finance_book is NULL"

nabinhait (changes requested):

nabinhait (approved):

PR #20413, opened 2020-01-24
feat: add tax category in pos profile

nextchamp-saqib:
Added Tax Category field in POS Profile form.
While loading POS the tax category in Sales Invoice will be set to it. however on Customer/Address selection it may get overwritten.

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -420,7 +420,9 @@ class SalesInvoice(SellingController):
 
 		if pos:
 			self.allow_print_before_pay = pos.allow_print_before_pay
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -420,6 +420,7 @@ def set_pos_fields(self, for_validate=False):
 
 		if pos:
 			self.allow_print_before_pay = pos.allow_print_before_pay
+			self.tax_category = pos.get("tax_category")
```
nabinhait: Should it be `if not for_validate`?

nabinhait (approved):

sammish:
@nabinhait I am waiting for this merge in Master, Please support.

PR #20415, opened 2020-01-24
fix(Customer): Change `primary_address` fieldtype to Text

hrwX:
- If the data is too long, `Read Only` field throw `data too long` error so changed it to `Text`


Don-Leopardo:
Is it the only place where this happens?
The advantage of Read Only field is that it is shown even if it is empty.

hrwX:
This is the place where I encountered the error, So went ahead and changed it to `Text with Read Only`.

deepeshgarg007 (approved):

PR #20422, opened 2020-01-25
fix: Parent company defaulting on creation of new company

deepeshgarg007:
On creation of a new company the parent company is fetched from the Global defaults irrespective if it is a group company or not


nabinhait (approved):

nabinhait (changes requested):
@deepeshgarg007 Travis is failing

PR #20430, opened 2020-01-27
feat: show numeric values in item configurator

scmmishra:
Item variants can be numeric or based on values specified in the table. The former case was not handled in the item configuration dialog. 

This PR fixes it.

Once approved, I'll send this to other branches too


netchampfaris (commented):
Neither frappe nor erpnext has `numpy` as dependency. Is it bundled with python? If not is it worth adding it as a dependency?

scmmishra:
@netchampfaris ERPNext has pandas as a dependency, which brings along numpy

netchampfaris (approved):

PR #20431, opened 2020-01-27
fix: discounting not applied on sales invoice created by subscriptions

0Pranav:
- fix option value in `Subscriptions`
- fix fetching `Apply Discount Based On` in Sales Invoice created by system using Subscriptions.

### DIFF
--- a/erpnext/accounts/doctype/subscription/subscription.py
+++ b/erpnext/accounts/doctype/subscription/subscription.py
@@ -280,7 +280,7 @@ class Subscription(Document):
 
 		if self.additional_discount_percentage or self.additional_discount_amount:
 			discount_on = self.apply_additional_discount
-			invoice.apply_additional_discount = discount_on if discount_on else 'Grand Total'
+			invoice.apply_discount_on = discount_on if discount_on else 'Grand Total'
 
 		# Subscription period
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -280,7 +280,7 @@ def create_invoice(self, prorate):
 
 		if self.additional_discount_percentage or self.additional_discount_amount:
 			discount_on = self.apply_additional_discount
-			invoice.apply_additional_discount = discount_on if discount_on else 'Grand Total'
+			invoice.apply_discount_on = discount_on if discount_on else 'Grand Total'
```
nabinhait, line 283: Why this change?

nabinhait (changes requested):

erpnext/accounts/doctype/subscription/subscription.py
0Pranav: there's no field called `apply_additional_discount` in Sales Invoice. The concerned field is named `apply_discount_on`. 

Screenshot 2020-01-27 at 3 48 33 PM

nabinhait (approved):

PR #20433, opened 2020-01-27
fix: cannot close task if dependent task are not closed / cancelled

nabinhait:


RJPvT (approved):
Approved by my count.

PR #20435, opened 2020-01-27
fix: don't allow territory creation via quick entry

0Pranav:
fix: don't allow territory creation via quick entry
fix: set "All Territories" as default value for parent_territory


erpnext/setup/doctype/territory/territory.json
```diff
@@ -1,487 +1,174 @@
...
-   "remember_last_selected_value": 0, 
-   "report_hide": 0, 
-   "reqd": 0, 
-   "search_index": 0, 
-   "set_only_once": 0, 
-   "translatable": 0, 
-   "unique": 0
-  }, 
+   "bold": 1,
+   "default": "All Territories",
```
nabinhait: You cannot hard-code "All Territories" here. It can be renamed or have different value based on translation.

nabinhait: To set it as default use `get_root` function

nabinhait (changes requested):

erpnext/setup/doctype/territory/territory.py
```diff
@@ -8,16 +8,21 @@
...
 	nsm_parent_field = 'parent_territory'
 
 	def validate(self):
+		self.set_parent()
 		for d in self.get('targets') or []:
 			if not flt(d.target_qty) and not flt(d.target_amount):
 				frappe.throw(_("Either target qty or target amount is mandatory"))
 
+	def set_parent(self):
+		if not self.parent_territory:
```
nabinhait, line 23: You cannot set parent territory forcefully, this way none will be able to create multiple root territories. You can set only default value.

nabinhait (changes requested):

0Pranav:
extending this change to other doctypes in https://github.com/frappe/erpnext/pull/20452

PR #20436, opened 2020-01-27
fix: don't allow territory creation via quick entry

0Pranav:
backport of https://github.com/frappe/erpnext/pull/20435


0Pranav:
extended to other doctypes in https://github.com/frappe/erpnext/pull/20453

PR #20438, opened 2020-01-27
fix: Column 'project' in where clause is ambiguous

marination:
On addition of custom field 'project' in Purchase Order in a client account the following occurred:
```py
pymysql.err.InternalError: (1052, "Column 'project' in where clause is ambiguous")
```

### DIFF
--- a/erpnext/accounts/doctype/budget/budget.py
+++ b/erpnext/accounts/doctype/budget/budget.py
@@ -210,10 +210,10 @@ def get_requested_amount(args, budget):
 	item_code = args.get('item_code')
 	condition = get_other_condition(args, budget, 'Material Request')
 
-	data = frappe.db.sql(""" select ifnull((sum(mri.stock_qty - mri.ordered_qty) * rate), 0) as amount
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/budget/budget.py
```diff
@@ -233,7 +233,8 @@ def get_other_condition(args, budget, for_doc):
 	budget_against_field = frappe.scrub(args.get("budget_against_field"))
 
 	if budget_against_field and args.get(budget_against_field):
-		condition += " and %s = '%s'" %(budget_against_field, args.get(budget_against_field))
+		alias = 'mri' if for_doc == 'Material Request' else 'poi'
```
nabinhait: This is not a good pattern. Better fix the alias in the query itself and use `parent`, `child` there.

nabinhait (changes requested):

erpnext/accounts/doctype/budget/budget.py
marination: Done

nabinhait (approved):

PR #20440, opened 2020-01-27
fix: permissions for gst settings & hsn

nextchamp-saqib:


nabinhait (commented):
Don't add permissions to regional doctypes by default. The permissions are added via https://github.com/frappe/erpnext/blob/develop/erpnext/regional/india/setup.py#L79.

Otherwise, these doctypes will be available for all countries.

Closing it.

PR #20441, opened 2020-01-27
fix: permissions for gst settings & hsn

nextchamp-saqib:
back port of #20440


erpnext/regional/doctype/gst_hsn_code/gst_hsn_code.json
```diff
@@ -32,13 +32,37 @@
...
  ],
- "links": [],
- "modified": "2019-12-29 11:32:06.849631",
+ "modified": "2020-01-27 16:20:09.277395",
  "modified_by": "Administrator",
  "module": "Regional",
  "name": "GST HSN Code",
  "owner": "Administrator",
- "permissions": [],
+ "permissions": [
```
deepeshgarg007, line 40: Stock Manager and Item Manager Should also have permission for GST HSN CODE since they might need it to while creating new items

deepeshgarg007 (changes requested):

erpnext/regional/doctype/gst_hsn_code/gst_hsn_code.json
```diff
@@ -32,13 +32,64 @@
...
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "Accounts Manager",
+   "share": 1,
+   "write": 1
+<<<<<<< HEAD
```
deepeshgarg007, line 64: @nextchamp-saqib Conflicts were not resolved properly

deepeshgarg007 (changes requested):

nabinhait (commented):
Don't add permissions to regional doctypes by default. The permissions are added via https://github.com/frappe/erpnext/blob/develop/erpnext/regional/india/setup.py#L79.

Otherwise, these doctypes will be available for all countries.

Closing it.

PR #20450, opened 2020-01-28
fix: odometer value was not syncing properly

Anurag810:
Once the Vehicle is created, the Odometer value lets say is x.
Once we submit a Vehicle Log, we enter the new Odometer Value y which gets updated in the Vehicle(y) as well.
Now if we delete the Vehicle log, the Odometer Value should also revert back from y to x - which is not happening, it stays on the new Odometer value y.
 #20451

PR #20451, opened 2020-01-28
fix: odometer value was not syncing properly

Anurag810:
#20450 
Once the Vehicle is created, the Odometer value lets say is x.
Once we submit a Vehicle Log, we enter the new Odometer Value y which gets updated in the Vehicle(y) as well.
Now if we delete the Vehicle log, the Odometer Value should also revert back from y to x - which is not happening, it stays on the new Odometer value y.


Anurag810:
@deepeshgarg007 please merge done changes requested by you.

deepeshgarg007 (approved):

PR #20457, opened 2020-01-28
feat: create appointment against customer

0Pranav:
- create appointment against customer in addition to lead
- handle query parameters, needed for success call implementation

### DIFF
--- a/erpnext/crm/doctype/appointment/appointment.py
+++ b/erpnext/crm/doctype/appointment/appointment.py
@@ -24,6 +24,14 @@ class Appointment(Document):
 			return lead_list[0].name
 		return None
 
+	def find_customer_by_email(self):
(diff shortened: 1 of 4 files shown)


erpnext/crm/doctype/appointment/appointment.py
```diff
@@ -24,6 +24,10 @@ def find_lead_by_email(self):
 			return lead_list[0].name
 		return None
 
+	def find_customer_by_email(self):
+		customer_list = frappe.get_list(
+			'Customer', filters={'email_id': self.customer_email}, ignore_permissions=True
+		)
```
Anurag810, line 30: missing return satement

Anurag810 (changes requested):

PR #20462, opened 2020-01-29
fix: delivery note creation from pick list if no sales orders found (v12)

Alchez:
**Problem:**

If there are no sales orders against a Pick List, the system throws a validation while trying to create a new Delivery Note, which breaks the flow.

**Solution:**

If no sales orders are found, default to an empty Delivery Note.


marination:
As mentioned in https://github.com/frappe/erpnext/pull/20463

PR #20465, opened 2020-01-29
fix: Label and UX fixes while creating payment entry against customer

deepeshgarg007:
- Label for bank account in customer master
- Default account label in bank account
- Validation for default bank account


nabinhait (approved):

nabinhait:
@deepeshgarg007 Send another PR for version-12-hotfix branch

PR #20466, opened 2020-01-29
fix(translations): Incorrect syntax

surajshetty3416:
`_('some text with constant {0}'.format('5'))` is not the right syntax.

`_('some text with constant {0}').format('5')` is the right syntax.

https://frappe.io/docs/user/en/translations#2-variables


surajshetty3416:
https://github.com/frappe/erpnext/pull/20471/

PR #20471, opened 2020-01-30
fix(translations): Incorrect syntax

surajshetty3416:
`_('some text with constant {0}'.format('5'))` is not the right syntax.

`_('some text with constant {0}').format('5')` is the right syntax.

https://frappe.io/docs/user/en/translations#2-variables


erpnext/accounts/doctype/bank_account/test_bank_account.py
```diff
@@ -39,11 +39,11 @@ def test_validate_iban(self):
 			try:
 				bank_account.validate_iban()
 			except ValidationError:
-				msg = _('BankAccount.validate_iban() failed for valid IBAN {}'.format(iban))
+				msg = _('BankAccount.validate_iban() failed for valid IBAN {}').format(iban)
```
netchampfaris: Remove translation method

netchampfaris (approved):

PR #20509, opened 2020-02-03
fix: Payment Schedule not mapped from Purchase Order to Invoice.

marination:
- **Before Fix:**
_payment_schedu_before

- **After Fix:**
_pay_sched_after

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -443,6 +443,9 @@ def get_mapped_purchase_invoice(source_name, target_doc=None, ignore_permissions
 			"doctype": "Purchase Taxes and Charges",
 			"add_if_empty": True
 		},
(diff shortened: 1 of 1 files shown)


erpnext/buying/doctype/purchase_order/purchase_order.py
```diff
@@ -443,6 +443,9 @@ def update_item(obj, target, source_parent):
 			"doctype": "Purchase Taxes and Charges",
 			"add_if_empty": True
 		},
+		"Payment Schedule": {
```
deepeshgarg007, line 446: @marination  Payment Terms are intentionally not copied from Purchase Order to Invoice. The reason behind this is Payment Terms vary from business to business. Some people take payment at the PO level and other at the Invoice level and since Payment Term's due date is are based on invoice date, while coping payment terms same date is copied but PO and PI might have different  posting date which makes the Payment terms Incorrect. So to summarise its intentional

PR #20511, opened 2020-02-03
fix: customer filter default values for customer group and territory

### LINKED ISSUE #18813: customers default filter
Default customers page filter 
All Customer Groups and All Territories
not listing the sub customers groups such as commercial group

0Pranav:
fixes: https://github.com/frappe/erpnext/issues/18813

### DIFF
--- a/erpnext/selling/doctype/customer/customer_list.js
+++ b/erpnext/selling/doctype/customer/customer_list.js
@@ -1,3 +1,10 @@
 frappe.listview_settings['Customer'] = {
 	add_fields: ["customer_name", "territory", "customer_group", "customer_type", "image"],
+
+	onload: function(listview) {
+		frappe.route_options = {
+			"customer_group": "",
+			"territory": ""
+		};
+	}
 };


erpnext/selling/doctype/customer/customer_list.js
```diff
@@ -1,3 +1,10 @@
 frappe.listview_settings['Customer'] = {
 	add_fields: ["customer_name", "territory", "customer_group", "customer_type", "image"],
+
+	onload: function(listview) {
+		frappe.route_options = {
+			"customer_group": "",
```
nabinhait, line 6: From where those default values are coming?

0Pranav: it's set in boot.py https://github.com/frappe/erpnext/blob/develop/erpnext/startup/boot.py#L19

0Pranav: from here it's set in `frappe.sys_defaults` so it is inserted in filter with global defaults.

nabinhait: I have removed default territory from Selling Settings, cleared cache and reloaded the system. Still default territory is set in Customer list view in my local.

0Pranav:
The best solution to this issue seems to be to remove the default value from `Selling Settings`

PR #20531, opened 2020-02-06
fix: populate pricing_rules table on submission

0Pranav:
for following doctypes:
- Purchase Invoice
- Purchase Order
- Sales Invoice
- Sales Order
- Quotation
- Supplier Quotation
- Purchase Reciept
- Delivery Note

Fixes inconsistency where pricing_rules table was not populated for some documents.


erpnext/controllers/accounts_controller.py
```diff
@@ -298,6 +298,8 @@ def set_missing_item_details(self, for_validate=False):
 
 					if ret.get("pricing_rules"):
 						self.apply_pricing_rule_on_items(item, ret)
+						if self.docstatus == 1:
+							self.set_pricing_rules_table()
```
rohitwaghchaure, line 302: Pass the item in the method don't check again in below code
```
for item in self.items:
	if item.pricing_rules:
```

erpnext/controllers/accounts_controller.py
```diff
@@ -594,6 +596,21 @@ def update_against_document_in_jv(self):
...
+		if self.doctype in ['Purchase Order', 'Sales Order', 'Sales Invoice', 'Purchase Invoice',
+							'Supplier Quotation', 'Purchase Receipt', 'Delivery Note', 'Quotation']:
+			# Clear the child table if already set
+			# Only pricing rules which are applied will show up here
+			self.set("pricing_rules", [])
+
+			for item in self.items:
+				if item.pricing_rules:
+					self.append("pricing_rules", {
+						"pricing_rule": item.pricing_rules.split(",")[0],
```
nabinhait, line 609: Why are we showing only one pricing rule? If multiple rules applied, we should row all rules in different rows.

nabinhait (changes requested):

PR #20535, opened 2020-02-06
fix(HR): skip earned leaves check for max leaves set to zero or less

Mangesh-Khairnar:
**Issue:** Earn leave were not getting created when the max leaves allowed were set to 0 or less. Added a check for the same


Anurag810 (dismissed):

Anurag810:
@Mangesh-Khairnar please add test

Mangesh-Khairnar:
@Anurag810 This already has a test in the [code](https://github.com/frappe/erpnext/blob/5632712d00b949d31b23285888d152d0c2033368/erpnext/hr/doctype/leave_application/test_leave_application.py#L412)

nabinhait (approved):

PR #20536, opened 2020-02-06
fix: skip earned leaves check for max leaves set to zero or less

Mangesh-Khairnar:
backport of #20535


Anurag810:
@Mangesh-Khairnar add test

nabinhait (approved):

PR #20543, opened 2020-02-07
fix: Update paid amount for pos return

deepeshgarg007:
Paid amount is not updated automatically in for partial returns

For Eg:
- Let's say invoice at 10 items. Total invoice value is 1000.
- User is returning say 2-3 items only. Total return value is calculated as 200
- When saving Credit Note/Return Entry, the value in the Paid Amount field remains 1000. It throws validation message. User has to manually change the value in the Paid Amount field.


erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -634,23 +639,46 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
...
+				- this.frm.doc.write_off_amount), precision("grand_total"));
+		} else {
+			var total_amount_to_pay = flt(
+				(flt(grand_total*this.frm.doc.conversion_rate, precision("grand_total"))
+					- this.frm.doc.total_advance - this.frm.doc.base_write_off_amount),
+				precision("base_grand_total")
+			);
+		}
+
+		$.each(this.frm.doc['payments'] || [], function(index, data) {
```
nabinhait: There can be multiple payment rows based on mode of payments. We don't know by which mode-of-payment they are returning the payment or even returning the amount or not. I think we should add a field in "POS Settings" for "Default mode-of-payment for return" and allocate the return amount against it.
Otherwise, we can also blank the entire payment table and remove "is_pos" check. In that case, we are assuming that we are issuing Credit note to customer, not refunding any cash.

nabinhait (changes requested):

nabinhait (approved):

muchai:
Shouldn't this go into the hotfix please? One has to incorporate the changes in version 12 for now.

gbm001:
This seems to break POS SINV returns for me (in V12) in a non-default currency, because the party_account_currency is often different to the SINV currency. total_amount_to_pay seems to be calculated in the party_account_currency, rather than the SINV currency, and hence is the wrong amount to apply here.
https://discuss.erpnext.com/t/issues-with-pos-returns-in-multi-currency-v12/70412

gbm001:
Changing the last few lines of calculate_outstanding_amount to: 


			if self.doc.doctype == 'Sales Invoice' and self.doc.get('is_pos') and self.doc.get('is_return'):
				grand_total = self.doc.rounded_total or self.doc.grand_total
				amount_to_pay = flt(grand_total - self.doc.total_advance
					- flt(self.doc.write_off_amount), self.doc.precision("grand_total"))
				self.update_paid_amount_for_return(amount_to_pay)

seems to fix it for me.

PR #20562, opened 2020-02-10
fix: get_students not respecting the program

govindsmenokee:
The get_students function is not taking into account the program for which the fee needs to be created. If a student is enrolled for multiple programs then the fee schedule pull up the wrong count of students.

> Screenshots/GIFs


deepeshgarg007 (commented):
@govindsmenokee Please raise a PR  on `develop` and `version-12-hotfix` and not directly on `version-12`. Please rebase and reopen.

Issue #20526, opened 2020-02-05
Clinical Procedure Template | Auto fetching of item code from template name gives item not found error.


abhijitkumbharInd:
## Introduction
While creating a clinical procedure template if the user enters the template name item code automatically fetched into item code and once the user saves the form it always throwing error.
"Could not find Item Code". It is not understandable to the user that this item needs to create first hence we have to keep this as a blank.

**## Steps to reproduce the issue**

1. Create Clinical Procedure Template
2. Enter Template Name
3. Select Item group
4. Save the doc
5. Trowing error: Could not find Item Code: ***

### Expected result
To avoid confusion to a user no need to fetch item code from template name. User has to select item code manually

akurungadam:
Clinical Procedure Template by default creates the Item (with the template name) on saving. The user is not expected to create the Item first and then supply it but I agree this not not quite obvious to the user. Ideally, we should default the Item code field (to template name, as it's now) and then allow user to change the item name if required. If the item is not present, saving the template should create a new item (or use the existing one if exists) with the name provided, instead of throwing the error.

ruchamahabal:
> Clinical Procedure Template by default creates the Item (with the template name) on saving. The user is not expected to create the Item first and then supply it but I agree this not not quite obvious to the user. Ideally, we should default the Item code field (to template name, as it's now) and then allow user to change the item name if required. If the item is not present, saving the template should create a new item (or use the existing one if exists) with the name provided, instead of throwing the error.

Automatic creation of Items for Clinical Procedure Template is great but since Link fields are validated on save, copying the "Item Code" for an item in a Link field which is not created is anti-pattern and affects user-experience. If we want to copy the template name in the item code field it should be 'Data' field. Then on save the Item is created and a read_only Link field "Item" gets set in the Clinical Procedure Template. That is what is done in https://github.com/frappe/erpnext/pull/20811

Issue #20183, opened 2020-01-04
On version:-  v12.0.5 (version-12) , Payment entry gets created for holding invoice and without having reference of PI on PE


ShraddhaRanjane:
## Context information (for bug reports)

**Output of `bench version`**
```
(paste here)
```

## Steps to reproduce the issue

1. Create new Purchase Receipt (PR)
2. Through that PR ->Create PI
3. In Purchase Invoice ,If user enable flag Hold Invoice
4.It will ask to Set 'Release Date' and set release date
5.Now Submit that invoice
6. 'Create' Button  will appear
7.Create->Payment
8.If you click on payment then only soft message is displaying .like 'ACC-PINV-2020-00004 is on hold till 2020-01-06'.
But Payment Entry form gets open 
9.If you check on Payment entry form , no Reference gets append in child table 
10. Without having Reference of PI ,payment entry gets submitted
### Expected result
First of all, Payment Entry should not be created for Holding invoice. Should throw message while click on create->payment
OR
Otherwise, at least  Reference (of PI) should be shown on Payment Entry form. 

### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

Ravn10:
Also invoices on hold are being included in accounts payable report
Solution:- 
Would be better if we get a check box to include invoices on hold 
in the report

Issue #20101, opened 2019-12-26
On Journal Entry form, Negative value getting accepted inside Debit and Credit field


ShraddhaRanjane:
## Context information (for bug reports)

**Output of `bench version`**
```
(paste here)
```

## Steps to reproduce the issue

1. Login to Admin
2.Go to Journal Entry form
3.Enter Debit and Credit amount as negative
4. Save and Submit
5.Form get submit

### Expected result
Negative Value should not be accepted
### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

toofun666:
Can you tell us why it shouldn't? After all these years, you are the first one to complain from it? Did you consider the rounding adjustments for tax items that may be in the purchase/sales invoices?

PR #20585, opened 2020-02-14
fix:  Healthcare and journal entry fixes

radhikagattani183:
**Healthcare - Clinical Procedure Template** - Auto fetching of item code from template name gives item not found error. #20526
**Description :** While creating Clinical Procedure Template, when we mention the template name then it auto fetches in item code and description, whether the item of that template name is present or not.

**Journal Entry** - Negative values are accepted for debit and credit amount fields. #20101
**Description :** While creating Journal Entry, when in accounts child table if we enter negative values for debit and credit amount then journal entry gets created without showing any valid error message.

**Payment Entry** - Payment entry gets created without having references. #20183
**Description :**   Payment entry should not be created without any Payment References.

**Student Admission -** Student Admission accepting end date less than start date 
**Description:** End date of admission should not be accepted less than start date.


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -923,6 +923,12 @@ frappe.ui.form.on('Payment Entry', {
 				}
 			});
 		}
+	},
+
+	validate:function(frm) {
+		if (frm.doc.references.length === 0) {
```
nabinhait, line 929: Why so? Reference is not mandatory in payment entry for advance payment

erpnext/accounts/doctype/journal_entry/journal_entry.js
```diff
@@ -451,10 +451,20 @@ frappe.ui.form.on("Journal Entry Account", {
 	},
 
 	debit_in_account_currency: function(frm, cdt, cdn) {
+		var row = locals[cdt][cdn];
+		if(row.debit_in_account_currency < 0 || row.debit_in_account_currency === 0){
+			frappe.model.set_value(cdt, cdn, "debit_in_account_currency", '');
+			frappe.throw(__("Debit Amount cannot be zero or negative."));
```
nabinhait, line 457: Debit / credit amount can be zero

erpnext/accounts/doctype/journal_entry/journal_entry.js
```diff
@@ -451,10 +451,20 @@ frappe.ui.form.on("Journal Entry Account", {
 	},
 
 	debit_in_account_currency: function(frm, cdt, cdn) {
+		var row = locals[cdt][cdn];
+		if(row.debit_in_account_currency < 0 || row.debit_in_account_currency === 0){
+			frappe.model.set_value(cdt, cdn, "debit_in_account_currency", '');
```
nabinhait, line 456: why you are setting it as blank?

erpnext/education/doctype/student_admission/student_admission.js
```diff
@@ -11,5 +11,11 @@ frappe.ui.form.on('Student Admission', {
 
 	academic_year: function(frm) {
 		frm.trigger("program");
+	},
+	admission_end_date: function(frm) {
+		if(frm.doc.admission_end_date && frm.doc.admission_end_date <= frm.doc.admission_start_date){
+			frm.set_value("admission_end_date", "");
+			frappe.throw(__("Please select the valid <b>Admission End Date</b>"));
```
nabinhait:
```suggestion
			frappe.throw(__("Admission End Date should be after Admission Start Date"));
```

nabinhait (changes requested):

erpnext/accounts/doctype/journal_entry/journal_entry.js
radhikagattani183: On considering, Journal Entry can't be created with 0 credit/debit amount. Hence set it to blank.

erpnext/accounts/doctype/journal_entry/journal_entry.js
radhikagattani183: Then what is the use of creating JV with zero amount.?

erpnext/accounts/doctype/payment_entry/payment_entry.js
radhikagattani183: Then we have to handle the scenario where PO is created with on_hold flag and PE is created against that.

erpnext/accounts/doctype/journal_entry/journal_entry.js
nabinhait: In a specific row, either debit or credit will have value, other will be zero (or blank).

erpnext/accounts/doctype/payment_entry/payment_entry.js
nabinhait: Raise a separate PR for that scenario

nabinhait (commented):
@radhikagattani123 I think, accounts related changes does not make any sense. Can you make a separate PR only for healthcare chages? Closing this.

Issue #20587, opened 2020-02-14
Student form joining date accepting a future date


abhijitkumbharInd:
**## Description of the issue**
On student form by default, the date is selected as today's date but if the user changes the date to future date it is still acceptable.

**## Steps to reproduce the issue**

1. Open Student form.
2. Add the joining date as a future date.
3. Save the form. It is accepting the future date.

**### Observed result**
Joining date accepting a future date.

**### Expected result**
The joining date should not accept the future date.

lokesh-indictrans:
Working on it

PR #20588, opened 2020-02-14
fix: Student form joining date accepting a future date #20587

abhijitkumbharInd:
On student form by default, the date is selected as today's date but if the user changes the date to future date it was acceptable. Now added validation for joining date to not accept a future date.


deepeshgarg007 (commented):
@abhijitkumbharInd There are unrelated commits in this pull request. Please correct and reopen, closing this for now. Also I think there will be cases where student record will be created before joining of the student. Can you give a use case why this is required?

Issue #20598, opened 2020-02-15
fix: Employee field should be filtered based on department on doctype Instructor 


sharmatripti:
## Description of the issue
if department is selected there should be employees belongs to that department.

## Steps to reproduce the issue

1.Go to instructor list create new instructor
2.Select department and then select employee.

### Observed result
All employees are visible when one department is selected

### Expected result
Only those employees should be there who belongs to that department.
 
`
![AwesomeScreenshot-2020-2-15-1581754886036](https://user-images.githubusercontent.com/61048272/74584693-2b4be080-4ffb-11ea-8ed9-555e3dc0ce7d.gif)

OS` version / distribution, `ERPNext` install method, etc.

sharmatripti:
Work in Progress.

PR #20600, opened 2020-02-15
fix: Filtered Employee based on Department on Instructor doctype (#20…

sharmatripti:
Issue no #20598


erpnext/education/doctype/instructor/instructor.js
```diff
@@ -51,3 +50,8 @@ frappe.ui.form.on("Instructor", {
 		}
 	}
 });
+cur_frm.fields_dict['employee'].get_query = function(doc) {
```
deepeshgarg007, line 53: @sharmatripti Please use `frm.set_query` instead of the cur_frm which is the old way of filtering. Its a simple filter and can be easily implemented using `set_query`.Closing this PR , please fix and reopen

sharmatripti: ok

PR #20607, opened 2020-02-15
fix: validation on max group strength student group form

khushalti:

> Screenshots/GIFs


deepeshgarg007 (changes requested):
@khushalti it should not allow 0 students either, it doesn't makes any sense. Please make the required changes

deepeshgarg007 (approved):

Issue #20584, opened 2020-02-14
While selecting multiple items it adds one blank row on the item child table


abhijitkumbharInd:
**## Description of the issue**

When a user clicks on add multiple items button and selects multiple items. When he closes the popup window, child table shows one blank row followed by selected items.

**## Steps to reproduce the issue**
1. Go to Sales Order.
2. Click on Add multiple buttons.
3. The popup window appears. Select multiple items.
4. Click the 
![While selecting multiple items it adds one blank row on the item child table](https://user-images.githubusercontent.com/40858925/74508493-2ddc0680-4f25-11ea-8e82-c915641af236.gif)
close button to close the popup.
5. The blank row gets added

**### Observed result**
On item child table blank row gets added.

**### Expected result**
Item child table should contain only selected items.

sagargharge:
I'm looking into it.

PR #20612, opened 2020-02-15
fix: blank row added while adding multiple items on sales order(#20584)

sagargharge:
frappe/frappe#18532 
first row was added blank while adding multiple items on sales order form.


deepeshgarg007 (commented):
@sagargharge This is not the correct solution. This should be fixed in the framework itself so that the issue is resolved for all the transactions like Sales Invoice, Purchase Invoice, Purchase Order, etc. and not just Sales Order. Closing this PR for now. Please fix and reopen on appropriate github repo.

Issue #20606, opened 2020-02-15
Date Validation issue on Sales Order and Sales Invoice


namrataindictrans123:
Issue 
**Sales Order**
1. Create sales order 
2.Fill all details
3.In Auto repeat section select from date greater than to date 
4.Form will save without error
![sales_order_issue](https://user-images.githubusercontent.com/51775031/74586868-f303cc80-5011-11ea-9904-9d8857152a0f.gif)

**Sales Invoice**
1.Create New Sales Invoice
2.Fill all details
3.In subscription section select from date greater than to date
4.Form will save without error
![sales_invoice_issue](https://user-images.githubusercontent.com/51775031/74586871-fbf49e00-5011-11ea-8f51-0087864360dc.gif)

**Expected Result**
From date should be less than to date

PR #20618, opened 2020-02-15
fix: Date Validation issue on Sales Order and Sales Invoice  #20606

namrataindictrans123:
Issue No. #20606

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -91,6 +91,7 @@ class SalesInvoice(SellingController):
 		self.validate_fixed_asset()
 		self.set_income_account_for_fixed_assets()
 		self.validate_item_cost_centers()
+		self.validate_subscription_date()
 		validate_inter_company_party(self.doctype, self.customer, self.company, self.inter_company_invoice_reference)
 
 		if cint(self.is_pos):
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1271,6 +1272,12 @@ def set_status(self, update=False, status=None, update_modified=True):
 		if update:
 			self.db_set('status', self.status, update_modified = update_modified)
 
+	def validate_subscription_date(self):
+		if self.from_date > self.to_date:
+			frappe.throw("From Date cannot be greater than To Date in Subscription section.")
+		else:
```
deepeshgarg007, line 1278: `else` is not required since you are just passing

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -463,6 +464,12 @@ def validate_serial_no_based_delivery(self):
 				Item {0} is added with and without Ensure Delivery by \
 				Serial No.").format(item.item_code))
 
+	def validate_auto_repeat_date(self):
+		if self.from_date > self.to_date:
+			frappe.throw("From Date cannot be greater than To Date in Auto repeat section.")
+		else:
```
deepeshgarg007, line 470: Same here as well

deepeshgarg007 (changes requested):

deepeshgarg007:
@namrataindictrans123 Please update and reopen. Closing for now

PR #20623, opened 2020-02-15
fix: Future date of birth accepted at Sibling Detail section in Studen…

jprashant09:
Validation for sibling date of birth

### DIFF
--- a/erpnext/education/doctype/student/student.py
+++ b/erpnext/education/doctype/student/student.py
@@ -23,7 +23,7 @@ class Student(Document):
 
 	def validate_dates(self):
 		for sibling in self.siblings:
-			if sibling.date_of_birth and getdate(sibling.date_of_birth) >= getdate():
+			if sibling.date_of_birth and getdate(sibling.date_of_birth) > getdate():
 				frappe.throw(_("Row {0}:Sibling Date of Birth cannot be greater than today.").format(sibling.idx))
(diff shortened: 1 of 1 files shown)


erpnext/education/doctype/student/student.py
```diff
@@ -22,6 +22,10 @@ def validate(self):
 			self.update_student_name_in_linked_doctype()
 
 	def validate_dates(self):
+		for sibling in self.siblings:
+			if sibling.date_of_birth and getdate(sibling.date_of_birth) >= getdate(today()):
```
deepeshgarg007:
```suggestion
			if sibling.date_of_birth and getdate(sibling.date_of_birth) >= getdate():
```
getdate() by default gives today's date if nothing is passed

erpnext/education/doctype/student/student.py
```diff
@@ -22,6 +22,10 @@ def validate(self):
...
+				frappe.throw(_("Sibling Date of Birth cannot be greater than today at row #{0} ").format(sibling.idx))
```
deepeshgarg007:
```suggestion
				frappe.throw(_("Row {0}: Sibling Date of Birth cannot be greater than today").format(sibling.idx))
```

deepeshgarg007 (changes requested):

PR #20632, opened 2020-02-16
Fix: Wrong filter for supplier default bank account

P-Froggy:
The query for default bank account field in the supplier master only allowed company bank accounts. With this fix it's now possible to one of the supplier bank accounts as default bank account.


deepeshgarg007 (commented):
@P-Froggy  it is actually supposed to get the company bank account and not the Supplier Bank Account. Maybe the label can be changed from `Default Bank Account` to `Default Company Bank Account` as we have recently done in Customer Master as it was a bit misleading. Closing this PR for now and will raise a separate PR for the label fix.

PR #20635, opened 2020-02-17
fix: validated leave allocation

Anurag810:
Screenshot 2020-02-17 at 11 48 00 AM
After Fix:
Screenshot 2020-02-17 at 12 02 01 PM

### DIFF
--- a/erpnext/hr/doctype/leave_encashment/leave_encashment.py
+++ b/erpnext/hr/doctype/leave_encashment/leave_encashment.py
@@ -65,7 +65,7 @@ class LeaveEncashment(Document):
 		allocation = self.get_leave_allocation()
 
 		if not allocation:
-			frappe.throw(_("No Leaves Allocated to Employee: {0} for Leave Type: {1}".format(self.employee, self.leave_type)))
+			frappe.throw(_("No Leaves Allocated to Employee: {0} for Leave Type: {1}").format(self.employee, self.leave_type))
 
 		self.leave_balance = allocation.total_leaves_allocated - allocation.carry_forwarded_leaves_count\
 			- get_unused_leaves(self.employee, self.leave_type, allocation.from_date, self.encashment_date)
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/leave_encashment/leave_encashment.py
```diff
@@ -64,6 +64,9 @@ def get_leave_details_for_encashment(self):
 
 		allocation = self.get_leave_allocation()
 
+		if not allocation:
+			frappe.throw("No Leaves Allocated to Employee: {0} for Leave Type: {1}".format(self.employee, self.leave_type))
```
nabinhait: @Anurag810 Where is multi-lingual support for the message? This is bad :(

nabinhait (changes requested):

erpnext/hr/doctype/leave_encashment/leave_encashment.py
Anurag810: Done

erpnext/hr/doctype/leave_encashment/leave_encashment.py
```diff
@@ -64,6 +64,9 @@ def get_leave_details_for_encashment(self):
 
 		allocation = self.get_leave_allocation()
 
+		if not allocation:
+			frappe.throw(_("No Leaves Allocated to Employee: {0} for Leave Type: {1}".format(self.employee, self.leave_type)))
```
nabinhait:
```suggestion
			frappe.throw(_("No Leaves Allocated to Employee: {0} for Leave Type: {1}").format(self.employee, self.leave_type))
```

nabinhait (changes requested):

Anurag810:
@nabinhait please review and merge

nabinhait (approved):

PR #20643, opened 2020-02-17
fix: HSN/SAC Error while creating a new item and not selecting the fi…

abhijitkumbharInd:
While creating an item if the user tries to enter the HSN/SAC code. And if a user doesn't enter and clicked outside the HSN/SAC box then the error "The resource you are looking for is not available" occurs.

This issue is resolved by checking if the HSN/SAC is entered or not. If there is no HSN/SAC then further DB call `frappe.db.get_doc("GST HSN Code"...)` can't be made.

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -137,7 +137,7 @@ frappe.ui.form.on("Item", {
 	},
 
 	gst_hsn_code: function(frm) {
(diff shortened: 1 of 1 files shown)


deepeshgarg007 (commented):
@abhijitkumbharInd can you please update in the description what this PR is resolving

abhijitkumbharInd:
@deepeshgarg007 I just updated the description of PR please check.

deepeshgarg007 (approved):

PR #20658, opened 2020-02-18
fix: incorrect required qty for subcontracting purchase receipt

rohitwaghchaure:
Backport https://github.com/frappe/erpnext/pull/20285

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -261,7 +261,7 @@ class BuyingController(StockController):
 
 			non_stock_items = get_non_stock_items(item.purchase_order, item.item_code)
 
-			item_key = '{}{}'.format(item.item_code, item.purchase_order)
+			item_key = '{}{}{}'.format(item.item_code, item.purchase_order, item.purchase_order_item)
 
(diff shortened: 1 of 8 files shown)


erpnext/controllers/buying_controller.py
```diff
@@ -760,6 +747,66 @@ def validate_items(self):
...
+		'consumed_serial_nos': consumed_serial_nos
+	}
+
+def get_extra_materials_parent_item_qty(purchase_orders):
+	fg_yet_to_be_received = frappe._dict()
+
+	raw_materials = frappe.get_all("Purchase Order Item",
+		fields=["SUM(qty - received_qty) as qty", "item_code", "parent"],
+		filters = {"docstatus": 1, "parent": ('in', purchase_orders)},
+		group_by = "item_code")
```
nabinhait, line 801: `having qty > received_qty`

nabinhait (changes requested):

erpnext/controllers/buying_controller.py
rohitwaghchaure: Done

nabinhait:
@rohitwaghchaure Closing this, let's redesign this as per our discussion.

PR #20660, opened 2020-02-18
chore: SINV set_status remove redundant function calls

nextchamp-saqib:


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1243,25 +1243,28 @@ def set_status(self, update=False, status=None, update_modified=True):
 
 		precision = self.precision("outstanding_amount")
 		outstanding_amount = flt(self.outstanding_amount, precision)
+		due_date = getdate(self.due_date)
+		nowdate = getdate(nowdate())
```
deepeshgarg007:
```suggestion
		nowdate = getdate()
```
`getdate` by default gives nowdate if noting is passed, no need to pass `nowdate()`

deepeshgarg007 (changes requested):

nabinhait (approved):

PR #20662, opened 2020-02-18
fit: Added a Expense Receipt field for uploading pic or file

Dharmraj-48:
**Added a new Field as Expense Receipt in Expense Claim DocType under Expense Section which will allow to take image or file Receipt**

Screenshot from 2020-02-17 17-07-15


nabinhait (commented):
It is not a good UX, adding attachment inside child table. Closing this.

Alchez:
Hey @nabinhait, is there a specific reason why child tables shouldn't have any attachments? Seems like a valid use case to have the field there.

PR #20663, opened 2020-02-18
fix: Set Query on warehouse fields in Stock Settings

marination:
- Group Warehouses can't be used for transactions, shouldn't be allowed as input in the first place.

**Fixes:**
- Client side query on warehouse Link fields
- Server Side validation
Screenshot 2020-02-19 at 11 05 47 AM


deepeshgarg007 (changes requested):
@marination Please add server side validation for this as well. It will still accept Group Warehouse if entered manually or assigned from script on server side in test cases.

deepeshgarg007 (approved):

PR #20676, opened 2020-02-19
perf: improve gl entry submission

nextchamp-saqib:
Avoid `frappe.get_doc`, Avoid `db_set`, Refactored `set_status()` 
Added Indexes: GL Entry -> `against_voucher_type`, `voucher_type`


nabinhait (approved):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -125,6 +125,27 @@ def create_remarks(self):
 			else:
 				self.remarks = _("No Remarks")
 
+	def set_status(self):
```
nabinhait:
```suggestion
	def set_status(self, update=False, status=None, update_modified=True):
```

nabinhait: Remove "Purchase Invoice" entry from status updater.
https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/status_updater.py#L47~L57

nabinhait (changes requested):

nabinhait (approved):

PR #20687, opened 2020-02-20
fix: Additional salary can be created only for active employee

Anurag810:
Before users are also allowed to create an additional salary for an employee who left. 
Now, the only employees with active status are allowed in additional salary. 

#20686


Mangesh-Khairnar (approved):
LGTM

Issue #20228, opened 2020-01-09
'Last Purchase Rate' taking wrong on BOM.


ShraddhaRanjane:
## Description of the issue
'Last Purchase Rate' taking wrong on BOM form

## Context information (for bug reports)

**Output of `bench version`**
```
(paste here)
```

## Steps to reproduce the issue

1.Create new Item
2.Create new Purchase order(PO1) for 20 rs
3.Create Purchase receipt for (PR1) 20 rs
4.Item Price got created for 20 rs
5.Now, Create again create PO2 for same item with rs 25.
6.Create new PR2 for 25 rs.
7.Deleted PO2 and PR2  (which is of rs 25)
7.Create new BOM for same item
8.Select 'Rate of materials based on'= Last Purchase Rate
9.select item
10. on BOM,  rate is taking 25 for that item  (which has been deleted)

### Expected result
Last Purchase Rate  should be rs.20 on BOM

### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

PR #20689, opened 2020-02-20
fix: 'Last Purchase Rate' taking wrong on BOM

lokesh-indictrans:
#20228 'Last Purchase Rate' taking wrong on BOM.:
2nd Commit:
While fetching the last purchase rate, when the purchase order and purchase receipt both are none, the '>=' condition becomes true and it throws fatal error.  
Added condition when both purchase order and purchase receipt are none.

**Issue** 
1st Commit: I created one purchase order for item 'A' and enter rate manually as 20.
Again i created a new purchase order for same item and entered rate as 25.
Then, I go to BOM, selected rate of materials based on as  'Last Purchase Rate'.
Select item 'A' rate gets fetch as 25.
Then i cancelled and deleted that purchase order of rate 25. 
Again go to BOM select rate of materials based on as  'Last Purchase Rate'.
Selected item 'A'. Still it fetches rate as 25,It should fetch rate as 20 because the purchase order with rate 25 has been deleted.
bom_issue
After resolving this output come as:
BOM_issue


lokesh-indictrans:
Hi, As per your request i have send this PR again with only one commit and description.

nextchamp-saqib (approved):

erpnext/stock/doctype/item/item.py
```diff
@@ -1002,14 +1004,17 @@ def get_last_purchase_details(item_code, doc_name=None, conversion_rate=1.0):
...
 
+	
+	
 	purchase_order_date = getdate(last_purchase_order and last_purchase_order[0].transaction_date
 							   or "1900-01-01")
 	purchase_receipt_date = getdate(last_purchase_receipt and
 								 last_purchase_receipt[0].posting_date or "1900-01-01")
 
-	if (purchase_order_date > purchase_receipt_date) or \
+	if (purchase_order_date >= purchase_receipt_date) or \
```
nextchamp-saqib: When `last_purchase_order` and `last_purchase_receipt` are both `None`, `purchase_order_date ` and `purchase_receipt_date ` becomes equal and the `>=` messes up tests.

change condition to `if last_purchase_order and (purchase_order_date >= purchase_receipt_date) or \.....`

nextchamp-saqib (dismissed):

lokesh-indictrans:
@nextchamp-saqib As per your suggestion, i have added that condition. Please check.

nabinhait (approved):

marination:
@Mergyfio backport version-12-hotfix

marination:

PR #20693, opened 2020-02-21
fix: smaller then instead of bigger then :-( dumb mistake

RJPvT:
made a mistake last time. of course pending review is bigger when it is in the future instead of smaller

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -201,7 +201,7 @@ def set_multiple_status(names, status):
 def set_tasks_as_overdue():
 	tasks = frappe.get_all("Task", filters={"status": ["not in", ["Cancelled", "Closed"]]}, fields=["name", "status", "review_date"])
 	for task in tasks:
-		if task.status in 'Pending Review':
+		if task.status == "Pending Review":
 			if getdate(task.review_date) > getdate(today()):
 				continue
 		frappe.get_doc("Task", task.name).update_status()
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -202,7 +202,7 @@ def set_tasks_as_overdue():
 	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
 	for task in tasks:
 		if frappe.db.get_value("Task", task.name, "status") in 'Pending Review':
-			if getdate(frappe.db.get_value("Task", task.name, "review_date")) < getdate(today()):
+			if getdate(frappe.db.get_value("Task", task.name, "review_date")) > getdate(today()):
 				continue
 		frappe.get_doc("Task", task.name).update_status()
 
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -199,10 +199,10 @@ def set_multiple_status(names, status):
 		task.save()
 
 def set_tasks_as_overdue():
-	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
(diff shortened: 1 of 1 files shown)


erpnext/projects/doctype/task/task.py
```diff
@@ -202,7 +202,7 @@ def set_tasks_as_overdue():
 	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
```
hrwX:
```suggestion
	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]}, fields=["name", "review_date"])
```
It doesn't make sense to `frappe.db.get_value` when you can get all the values in `frappe.get_all`

hrwX (changes requested):

erpnext/projects/doctype/task/task.py
RJPvT: I could make this change but review date is not nessecary for most. In a test just now a standard query with this change is 1.2ms slower then the current setup. Should i make the change ?

erpnext/projects/doctype/task/task.py
```diff
@@ -202,7 +202,7 @@ def set_tasks_as_overdue():
...
 	for task in tasks:
 		if frappe.db.get_value("Task", task.name, "status") in 'Pending Review':
-			if getdate(frappe.db.get_value("Task", task.name, "review_date")) < getdate(today()):
+			if getdate(frappe.db.get_value("Task", task.name, "review_date")) > getdate(today()):
```
hrwX: If you get all the fields in `get_all`, it'll save a lot of `db.get_values` here, so its a no brainer change to make.

hrwX (changes requested):

erpnext/projects/doctype/task/task.py
RJPvT: Done

erpnext/projects/doctype/task/task.py
```diff
@@ -199,10 +199,10 @@ def set_multiple_status(names, status):
 		task.save()
 
 def set_tasks_as_overdue():
-	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]})
+	tasks = frappe.get_all("Task", filters={'status':['not in',['Cancelled', 'Closed']]}, fields=["name","status","review_date"])
```
hrwX:
```suggestion
	tasks = frappe.get_all("Task", filters={"status": ["not in", ["Cancelled", "Closed"]]}, fields=["name", "status", "review_date"])
```

hrwX:
```suggestion
	tasks = frappe.get_all("Task", filters={"status": ["not in", ["Cancelled", "Closed"]]}, fields=["name", "status", "review_date"])
```

erpnext/projects/doctype/task/task.py
```diff
@@ -199,10 +199,10 @@ def set_multiple_status(names, status):
...
 	for task in tasks:
-		if frappe.db.get_value("Task", task.name, "status") in 'Pending Review':
-			if getdate(frappe.db.get_value("Task", task.name, "review_date")) > getdate(today()):
+		if task.status in 'Pending Review':
```
hrwX:
```suggestion
		if task.status == "Pending Review":
```

hrwX (changes requested):

hrwX:
@RJPvT Can you open the same PR for `version-12-hotfix` as well as `develop`

RJPvT:
Yes, i will do that tonight.

RJPvT:
Also changed Closed to Completed in v12 en develop. This was introduced in v12 en was apparently not working anyway :-)

RJPvT:
@hrwX Are the PR's now up to code for implementation ? I have to manually reset 50+ tasks because of this.

nabinhait (approved):

PR #20694, opened 2020-02-21
Group Chart of Accounts & Consolidated Financials (Version 12 hotfix)

ghost:
GL accounts should be created in the chart of accounts of a child company (subsidiary) rather than in the parent (group) company. In this way, the group company only gets updated as accounts are added in the child companies.

Example:
- "1401 - GTBank" added in Subsidiary A.
- "1501 - Access Bank" added in Subsidiary B.
- Group company should be updated with both accounts.
Untitled

GL accounts should not be added in the parent company. Great feature for group of companies with different unrelated charts of acounts.


deepeshgarg007 (commented):
@Judesque  branch of from develop branch and open the PR again. Closing this for now.

PR #20710, opened 2020-02-24
perf: search for customer's return invoices then filter out gl_entries

nextchamp-saqib:
Back port of #20709


nabinhait (approved):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1520,6 +1520,9 @@ def get_loyalty_programs(customer):
 	else:
 		return lp_details
 
+def on_doctype_update():
+	frappe.db.add_index("Sales Invoice", ["customer", "is_return", "return_against"])
```
nabinhait, line 1524: @nextchamp-saqib Do same for purchase invoice

PR #20712, opened 2020-02-25
fix: same free item not working for pricing rule

rohitwaghchaure:
image

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/utils.py
+++ b/erpnext/accounts/doctype/pricing_rule/utils.py
@@ -435,7 +435,7 @@ def apply_pricing_rule_on_transaction(doc):
 				doc.calculate_taxes_and_totals()
 			elif d.price_or_product_discount == 'Product':
 				item_details = frappe._dict({'parenttype': doc.doctype})
-				get_product_discount_rule(d, item_details, doc)
+				get_product_discount_rule(d, item_details, doc=doc)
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -435,17 +435,17 @@ def apply_pricing_rule_on_transaction(doc):
...
 
 def get_applied_pricing_rules(item_row):
 	return (item_row.get("pricing_rules").split(',')
 		if item_row.get("pricing_rules") else [])
 
-def get_product_discount_rule(pricing_rule, item_details, doc=None):
+def get_product_discount_rule(pricing_rule, item_details, args=None, doc=None):
 	free_item = (pricing_rule.free_item
-		if not pricing_rule.same_item or pricing_rule.apply_on == 'Transaction' else item_details.item_code)
+		if not pricing_rule.same_item or pricing_rule.apply_on == 'Transaction' else item_details.item_code or args.item_code)
```
nabinhait: Whose `or` condition is this?
Is this like `(item_details.item_code or args.item_code)`?

nabinhait (changes requested):

erpnext/accounts/doctype/pricing_rule/utils.py
rohitwaghchaure: Yes I will make change

nabinhait (approved):

PR #20724, opened 2020-02-25
fix: Handle rows with same item code from Purchase Receipt to Invoice.

marination:
- Consider a Purchase Receipt with multiple rows in the items table all having **the same Item Code**.
- A Purchase Return exists against this receipt for the first row in the items table.
- While creating invoice via Purchase Receipt, it shows the following:
  Screenshot 2020-02-25 at 4 47 56 PM

- It fetches the return details of the item code. Since all rows have same item code it assumes all rows have been returned.

**After Fix:**
- Added 'purchase_receipt_item' field in Purchase Receipt Item. The return document will have exactly which row was returned.
- Rows that are returned will not be mapped to Purchase Invoice.

**Working:**
_PR_to_PI


nabinhait (changes requested):
Resolve conflicts

marination:
> Resolve conflicts

@nabinhait  done

marination:
Screenshot 2020-04-28 at 6 14 10 PM

nabinhait (approved):

PR #20727, opened 2020-02-25
fix: deductions calculation based on gross pay

Anurag810:


erpnext/hr/doctype/salary_structure/test_salary_structure.py
```diff
@@ -38,6 +37,29 @@ def make_holiday_list(self):
...
 			holiday_list.save()
 
+	def test_salary_structure_deduction_based_on_gross_pay(self):
+
+		emp = make_employee("test_employee_3@salary.com")
+
+		sal_struct = make_salary_structure("Salary Structure 2", "Monthly", dont_submit = True)
+
+		sal_struct.earnings = [sal_struct.earnings[0]]
+		sal_struct.deductions[0].amount_based_on_formula = 1
```
nabinhait:
```suggestion
		sal_struct.earnings[0].amount_based_on_formula = 1
```

erpnext/hr/doctype/salary_structure/test_salary_structure.py
```diff
@@ -38,6 +37,29 @@ def make_holiday_list(self):
...
+		sal_struct.deductions[0].formula =  "base"
```
nabinhait:
```suggestion
		sal_struct.earnings[0].formula =  "base"
```

nabinhait (changes requested):

nabinhait (approved):

Issue #20603, opened 2020-02-15
user is not able  to do 'Set as Lost' for Quotation


lokesh-indictrans:
"1. Login to Admin
2. Go to Quotation
3.Create new Quotation 
4.Enter All Details 
5.Save and Submit
6. 'Set as Lost'  button will appear
7.Click on that
8.Pop-up will appear.
9.'Lost Reason' field having link field to create lost for opptunity
10.Do not create Lost reason through link field
11. Enter any character  inside 'Lost Reason'
12.Click on 'Declare Lost' 
13.Message is displaying 'Could not find Row #1'"

![fatal_error](https://user-images.githubusercontent.com/59611773/75550486-04a89380-5a58-11ea-823f-587f20ad9cd3.png)

PR #20755, opened 2020-02-27
fix: user is not able to do 'Set as Lost' for Quotation (#20603)

lokesh-indictrans:
Issue:
Go to quotation form,
When user clicks on set as lost on quotation,
And enter any character in a dialog and submit, below fatal error appears.

fatal_error

Solved 
solution_qtn


lokesh-indictrans:
@nabinhait Check my pull request for issue #20603 . Let me know if any modifications required.

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -68,13 +68,19 @@ def update_opportunity_status(self, opportunity=None):
...
 
 			if detailed_reason:
 				frappe.db.set(self, 'order_lost_reason', detailed_reason)
 
 			for reason in lost_reasons_list:
-				self.append('lost_reasons', reason)
+				if reason.get('lost_reason') in lost_reasons_lst:
+					self.append('lost_reasons', reason)
+				else:
+					frappe.throw(_("Invalid lost reason <b>{0}</b>, please create a new lost reason".format(reason.get('lost_reason'))))
```
deepeshgarg007: @lokesh-indictrans This string won't be translated properly since it has html tags inside it. Use `frappe.bold()` function instead of HTML tags.

deepeshgarg007 (changes requested):

erpnext/selling/doctype/quotation/quotation.py
lokesh-indictrans: @deepeshgarg007  I have updated code for requested changes.

deepeshgarg007 (approved):

PR #20757, opened 2020-02-27
feat: auto set batch no on serial no selection

nextchamp-saqib:
If an item has serial nos as well as batch nos then on selection of serial no, the batch no it belongs will be auto fetched.


nabinhait (changes requested):
- "Selected Batch Numbers" should be "Selected Serial Numbers" if item has both serial and batch.
- Make warehouse mandatory to auto fetch serial nos
- Dialog does not appear if "Set Source Warehouse" selected in DN and then added item. Same issue with "Add serial no" once serial nos set using dialog

nabinhait (approved):

PR #20776, opened 2020-02-28
feat(selling): added delivery-window to customer

MyuddinKhatri:
added delivery-window to customer

some of customers have specific times when they take deliveries

screenshot-bloomstack_demo_8000-2020 02 28-12_47_26


nabinhait:
Thanks for first PR!

It is better to add these kind of fields via Custom Field, as these are just some information about customer, no business logic involved in this.

PR #20796, opened 2020-03-02
fix: Item alternative must have similar fields as original item

marination:
- **Maintain Stock** must be the same in both (obstructing transaction)
 Screenshot 2020-03-09 at 2 08 48 PM

- **Has Batch No, Has Serial No** and **Include Item in Manufacturing** can be different ( show alert)
 Screenshot 2020-03-09 at 2 08 25 PM


erpnext/stock/doctype/item_alternative/item_alternative.py
```diff
@@ -22,6 +22,23 @@ def validate_alternative_item(self):
 		if self.item_code == self.alternative_item_code:
 			frappe.throw(_("Alternative item must not be same as item code"))
 
+		fields = ["is_stock_item", "include_item_in_manufacturing","has_serial_no","has_batch_no"]
+		for field in fields:
+			if frappe.get_value("Item", self.item_code, field) != frappe.get_value("Item", self.alternative_item_code, field):
```
nabinhait: get_value should not be called inside for loop

erpnext/stock/doctype/item_alternative/item_alternative.py
```diff
@@ -22,6 +22,23 @@ def validate_alternative_item(self):
...
+				doc = frappe.get_doc("Item", self.item_code)
```
nabinhait: `get_doc` should be avoided. Use `get_meta` outside for loop.

nabinhait (changes requested):

nabinhait (approved):

PR #20798, opened 2020-03-02
fix: purchase return are allowed even when assets are not cancelled

nextchamp-saqib:
Back port of #20797


erpnext/controllers/buying_controller.py
```diff
@@ -99,6 +100,21 @@ def validate_stock_or_nonstock_items(self):
...
+		if self.doctype not in ['Purchase Receipt', 'Purchase Invoice'] or not self.is_return:
+			return
+
+		purchase_doc_field = 'purchase_receipt' if self.doctype == 'Purchase Receipt' else 'purchase_invoice'
+		not_cancelled_asset = [d.name for d in frappe.db.get_all("Asset", {
+			purchase_doc_field: self.return_against,
+			"docstatus": 1
+		})]
+		if self.is_return and len(not_cancelled_asset):
+			if len(not_cancelled_asset) < 4:
```
nabinhait: Why less than 4?

nabinhait (changes requested):

erpnext/controllers/buying_controller.py
nextchamp-saqib: Just for better message. If there are less than 4 assets not cancelled then display their names in the error message.

nabinhait (approved):

PR #20799, opened 2020-03-02
fix: Item alternative must have similar fields as orignal item

marination:
Port of https://github.com/frappe/erpnext/pull/20796


nabinhait (changes requested):
Added feedback on https://github.com/frappe/erpnext/pull/20796

nabinhait (approved):

PR #20805, opened 2020-03-02
fix: fetch sales person name

0925-pinka:
fetch_sales_person_name


nabinhait (approved):

erpnext/setup/doctype/sales_person/sales_person.js
```diff
@@ -19,6 +19,11 @@ frappe.ui.form.on('Sales Person', {
 				}
 			}
 		};
+	
+		frm.make_methods = {
+			'Sales Order': () => frappe.new_doc("Sales Order")
+				.then(() => frm.add_child("sales_team", {"sales_person": frm.doc.name}))
```
Alchez, line 25: @nabinhait, `frm` will not work here, since the context has now changed to the new form.

PR #20806, opened 2020-03-02
fix: Set batch no in Manufacturing Stock Entry as per transferred raw materials

nmimsnikhil:
Fetched batch number in stock_entry.py and applied to item

Before:
Batch_no_Task

After:
batch_number_fix


nmimsnikhil:
Frappe Pull Request Link:
https://github.com/frappe/erpnext/pull/20808

nabinhait (commented):
- What will happen if same item from multiple batches has been transferred?
- Similarly, multiple transfer entries may exists against single Work Order.
- Same is applicable for serial no as well.

Let us fix the issue at the core. Closing the PR.

PR #20808, opened 2020-03-02
fix: Set batch no in Manufacturing Stock Entry as per transferred raw materials

nmimsnikhil:
Fetched batch number in stock_entry.py and applied to item

Before:
Batch_no_Task

After:
batch_number_fix

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -398,16 +398,13 @@ class StockEntry(StockController):
 						frappe.bold(d.transfer_qty)),
 					NegativeStockError, title=_('Insufficient Stock'))
 
-	def set_serial_nos(self, work_order):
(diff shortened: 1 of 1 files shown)


pujanthakkar:
What if there are multiple batches of same item transferred ?

pujanthakkar:
> What if there are multiple batches of same item transferred ?

For Example when we transfer fabric, it comes in multiple batches & what we do is transfer multiple batches by adding multiples rows in stock entry for transfer to manufacture.

Currently in the Manufacture entry all those batches are not fetched and what we have to do is manually enter each batch again. Will this resolve that ?

rohitwaghchaure:
@nmimsnikhil Can you split the quantity in case of multiple batches has been transferred?

Alchez:
Hey @rohitwaghchaure, the frontported PR to [develop](https://github.com/frappe/erpnext/pull/20806) for this one was closed since @nabinhait  said you guys would fix it in the core. Should we still keep this open?

PR #20812, opened 2020-03-03
fix reason for return v12 hotfix

nmimsnikhil:
Added a field called "Reason for Return" in Delivery Note to capture reason for returns / rejections.

image


nmimsnikhil:
Adding a Reason for Return for each items

PR #20813, opened 2020-03-03
feat: added a field called 'reason for rejection' in delivery note to c…

nmimsnikhil:
Added a fields in "Delivery Note" DocType.

delivery_note1


nmimsnikhil:
Adding Fields reason for rejection for each items

PR #20814, opened 2020-03-03
fix: handle error handling (v12)

Alchez:
**Problem:**

When processing delivery routes, if there's bad inputs in the address, then the system fails trying to handle the error.

```python
TypeError: Expected a lat/lng dict or tuple, but got NoneType

During handling of the above exception, another exception occurred:

TypeError: Object of type 'TypeError' is not JSON serializable
```


nabinhait (approved):

Alchez:
@nabinhait, this is also made to develop at #20815

Issue #20650, opened 2020-02-17
Exchange Rate Revaluation bug v.12


Ashraf-Elsharqawy:
Any help pls !
I will explain an example to understand the problem
First, the base currency is EGP
1- A purchase invoice was made for a supplier in the dollar currency, the value of the bill is $ 1000, the Currency Exchange is 17, so the supplier balance in the base currency, 17000 after this transaction .
2- Then Exchange Rate Revaluation was done with a Currency Exchange 16
The balance of the supplier in general ledger is still 17,000 EGP
And the balance in trial balance for party is 16,000 EGP
3- A payment entry was made for the supplier with a value of $ 1,000 with Currency Exchange = 17
Here the balance has become zero in the general ledger
But in the trial balance for party became debit by 1000
Balance should be zero in all reports !!!
![2-4](https://user-images.githubusercontent.com/52603121/74671960-e79cd680-51b4-11ea-9a05-bebc42b05a87.jpg)
![2-5](https://user-images.githubusercontent.com/52603121/74671964-e8ce0380-51b4-11ea-8033-0e86a2046468.jpg)
![screenshot-localhost_4080-2020 02 14-23_06_48](https://user-images.githubusercontent.com/52603121/74671967-e9669a00-51b4-11ea-81be-b90bfb2f0652.jpg)
![screenshot-localhost_4080-2020 02 14-23_06_49](https://user-images.githubusercontent.com/52603121/74671968-e9ff3080-51b4-11ea-9d32-30961b4bfc2c.jpg)

[
![1](https://user-images.githubusercontent.com/52603121/74671949-e370b900-51b4-11ea-81b3-4274421f1c71.jpg)
![2](https://user-images.githubusercontent.com/52603121/74671952-e53a7c80-51b4-11ea-8d14-ef7fa81b32db.jpg)
![2-1](https://user-images.githubusercontent.com/52603121/74671953-e5d31300-51b4-11ea-8d11-dbc9b47ec8a8.jpg)
![2-2](https://user-images.githubusercontent.com/52603121/74671956-e66ba980-51b4-11ea-8fae-b749c87affab.jpg)
![2-3](https://user-images.githubusercontent.com/52603121/74671957-e7044000-51b4-11ea-80f2-939f960c42b2.jpg)

](url)

PR #20819, opened 2020-03-03
Exchange Rate Revaluation bug v.12

Ashraf-Elsharqawy:
Any help pls !
I will explain an example to understand the problem
First, the base currency is EGP
1- A purchase invoice was made for a supplier in the dollar currency, the value of the bill is $ 1000, the Currency Exchange is 17, so the supplier balance in the base currency, 17000 after this transaction .
2- Then Exchange Rate Revaluation was done with a Currency Exchange 16
The balance of the supplier in general ledger is still 17,000 EGP
And the balance in trial balance for party is 16,000 EGP
3- A payment entry was made for the supplier with a value of $ 1,000 with Currency Exchange = 17
Here the balance has become zero in the general ledger
But in the trial balance for party became debit by 1000
Balance should be zero in all reports.

https://github.com/frappe/erpnext/issues/20650


deepeshgarg007 (commented):
@Ashraf-Elsharqawy Please branch off from develop branch and open again. Closing this for now

Ashraf-Elsharqawy:
> @Ashraf-Elsharqawy Please branch off from develop branch and open again. Closing this for now

I could not understand   plz give steps   to do  that     thanks

PR #20826, opened 2020-03-04
fix: if mandatory fields are missing do not create address against lead

saurabh6790:
Screenshot 2020-03-04 at 1 01 19 PM

- added provision to create a new address from lead
Screenshot 2020-03-09 at 11 54 18 AM

Screenshot 2020-03-09 at 11 54 26 AM

### DIFF
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -135,10 +135,14 @@ class Lead(SellingController):
 
 		# do not create an address if no fields are available,
 		# skipping country since the system auto-sets it from system defaults
-		if not any([self.get(field) for field in address_fields if field != "country"]):
+		address = frappe.new_doc("Address")
+
+		mandatory_fields = get_mandatory_fields(address)
+
+		if not all([self.get(field) for field in mandatory_fields]):
+			frappe.msgprint(_('Missing mandatory fields in address.'), alert=True, indicator='yellow')
 			return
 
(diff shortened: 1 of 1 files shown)


erpnext/crm/doctype/lead/lead.py
```diff
@@ -370,3 +374,10 @@ def get_lead_with_phone_number(number):
 	lead = leads[0].name if leads else None
 
 	return lead
+
+def get_mandatory_fields(doc):
```
Don-Leopardo: I think that this method can be useful for other purposes. Maybe it should be in a more common file, like utils or something like that.

erpnext/crm/doctype/lead/lead.py
```diff
@@ -135,10 +135,17 @@ def create_address(self):
 
 		# do not create an address if no fields are available,
 		# skipping country since the system auto-sets it from system defaults
-		if not any([self.get(field) for field in address_fields if field != "country"]):
+		address = frappe.new_doc("Address")
+
+		mandatory_fields = [ df.fieldname for df in doc.meta.fields if df.reqd ]
+
+		if not all([self.get(field) for field in mandatory_fields]):
```
nabinhait, line 142: Why we are checking all mandatory fields?

Address related mandatory fields should be enough? There are possibilities, we skipped other mandatory fields for some other reason.

rohitwaghchaure (approved):

PR #20829, opened 2020-03-04
fix: added a Reason for Rejection field for each items

nmimsnikhil:
Added a field "Reason for Return" when creating Sales Return of Delivery Note for each items.

Before: its not showing Reason for Return field.
delivery_note_task

After:added Reason for Return field.
reason_for_return


nabinhait (approved):

nabinhait (changes requested):
The PR is good to go for Delivery Note, but should be implemented for there docs as well.

- Add reason field in Sales Invoice Item as well
- Show dialog in Sales Return via Sales Invoice, if Update Stock is checked
- Implement same feature for purchase return?
- Add documentation

nabinhait:
Closing it for now, please o the suggested changes and reopen the PR

PR #20860, opened 2020-03-06
feat: Nested set filtering for accounting dimension

deepeshgarg007:
Get child results if Document Type in Accounting dimension is a Nested Set

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.py
+++ b/erpnext/accounts/report/general_ledger/general_ledger.py
@@ -10,7 +10,7 @@ from frappe import _, _dict
 from erpnext.accounts.utils import get_account_currency
 from erpnext.accounts.report.financial_statements import get_cost_centers_with_children
 from six import iteritems
-from erpnext.accounts.doctype.accounting_dimension.accounting_dimension import get_accounting_dimensions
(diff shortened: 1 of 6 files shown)


nabinhait (dismissed):

nabinhait (approved):

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -194,12 +194,15 @@ def get_conditions(filters):
...
 
 	if accounting_dimensions:
 		for dimension in accounting_dimensions:
-			if filters.get(dimension):
-				conditions.append("{0} in (%({0})s)".format(dimension))
+			if filters.get(dimension.fieldname):
+				if frappe.get_cached_value('DocType', dimension.document_type, 'is_tree'):
+					filters[dimension.fieldname] = get_dimension_with_children(dimension.document_type,
+						filters.get(dimension.fieldname))
+				conditions.append("{0} in %({0})s".format(dimension.fieldname))
```
gavindsouza, line 205: @deepeshgarg007 invalid SQL statement?
```suggestion
				conditions.append("{0} in (%({0})s)".format(dimension.fieldname))
```

PR #20864, opened 2020-03-06
feat: Allow PI creation without PO

deepeshgarg007:
- Allow to create a Purchase Invoice against a supplier even though Purchase Order Required is set as "Yes" in Buying Settings 
- Checkboxes added in **Customer** and **Supplier**


nabinhait (changes requested):
Why is it only for allowing PI without PO?
The same thing is required for PR without PO, DN without SO and SI without SO.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -288,16 +288,26 @@ def set_against_expense_account(self):
...
+
+			if frappe.get_value('Supplier', self.supplier, 'allow_purchase_invoice_creation_without_purchase_order'):
+				return
+
 			for d in self.get('items'):
 				if not d.purchase_order:
-					throw(_("As per the Buying Settings if Purchase Order Required == 'YES', then for creating Purchase Invoice, user need to create Purchase Order first for item {0}").format(d.item_code))
+					throw(_("""Purchase Order Required for item {0}
+						To submit the invoice without purchase order please set
+						{1} as {2} in {3}""").format(frappe.bold(d.item_code), frappe.bold('Purchase Order Required'),
```
nabinhait:
```suggestion
						{1} as {2} in {3}""").format(frappe.bold(d.item_code), frappe.bold(_('Purchase Order Required')),
```

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -288,16 +288,26 @@ def set_against_expense_account(self):
...
 
 	def pr_required(self):
 		stock_items = self.get_stock_items()
 		if frappe.db.get_value("Buying Settings", None, "pr_required") == 'Yes':
 			for d in self.get('items'):
 				if not d.purchase_receipt and d.item_code in stock_items:
-					throw(_("As per the Buying Settings if Purchase Reciept Required == 'YES', then for creating Purchase Invoice, user need to create Purchase Receipt first for item {0}").format(d.item_code))
+					throw(_("""Purchase Receipt Required for item {0}
+						To submit the invoice without purchase receipt please set
+						{1} as {2} in {3}""").format(frappe.bold(d.item_code), frappe.bold('Purchase Receipt Required'),
```
nabinhait:
```suggestion
						{1} as {2} in {3}""").format(frappe.bold(d.item_code), frappe.bold(_('Purchase Receipt Required')),
```

nabinhait (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -267,16 +267,30 @@ def set_against_expense_account(self):
...
+			if frappe.get_value('Supplier', self.supplier, 'allow_purchase_invoice_creation_without_purchase_order'):
+				return
+
 			for d in self.get('items'):
 				if not d.purchase_order:
-					throw(_("As per the Buying Settings if Purchase Order Required == 'YES', then for creating Purchase Invoice, user need to create Purchase Order first for item {0}").format(d.item_code))
+					throw(_("""Purchase Order Required for item {0}
+						To submit the invoice without purchase order please set
+						{1} as {2} in {3}""").format(frappe.bold(d.item_code), frappe.bold(_('Purchase Order Required')),
+						frappe.bold('No'), get_link_to_form('Buying Settings', 'Buying Settings', 'Buying Settings')))
```
marination, line 279: `get_link_to_form ('Buying Settings', 'Buying Settings')` works too

marination (approved):

Ponyooooo:
Is it possible if we could also include this feature on the Items? Because sometimes there are suppliers with mixed items which means the item can be received or items are for billing only. Also, if possible to have a checkbox on the item record if it will be included in the printing of the transactions so that it will be automatically not included on the print but still be included or visible on the transaction?

PR #20893, opened 2020-03-09
fix: null fields being set in Integration Request

sunhoww:
In py3, `string.encode("utf-8")` results in strings of type **bytestrings** instead of the expected **unicode** as in py2. And py3 throws _TypeError_ when `json.dumps` is called on any object containing bytestrings.

Therefore the `json.dumps` called on the _kwargs_ during **Integration Request** creation fails. However this happens silently and some of the fields in the **Integration Request**._data_ json ends up with _null_.

This PR should prevent this from happening in py3 environments while leaving the behavior in py2 unchanged.


sunhoww:
incorrectly opened pr on branch _version-12_

PR #20900, opened 2020-03-11
fix: reserved qty for production issue for partial completion of work…

rohitwaghchaure:
**Issue**

Steps to replicate issue

1. Created work order for 5 qty and enabled skip material transfer

1. Created manufacturing entry for the 2 items

1. Reserved qty for production in the BIN has not updated.

1. Created manufacturing entry for the 3 items then Reserved qty for production has updated as Zero.

So for partial completion the Reserved qty for production has not updated if skip material transfer has enabled in the work order


nabinhait (changes requested):
Travis failing

rohitwaghchaure:
**Fixed**
image

nabinhait (commented):
Should we write a patch to recalculate reserved qty where skip_transfer was checked?

nabinhait (approved):

nabinhait (commented):
@rohitwaghchaure Please send another PR for develop branch

PR #20912, opened 2020-03-12
 fix(HR): Leave application company field made read-only and filter in Monthly attendance sheet

vishdha:
Made Leave Application company field Read only. Because, When We create leave application and someone change company field value it affect in attendance module. This simply affect Monthly Attendance report like this

cSXzfoV


nabinhait (approved):

marination:
@nabinhait leave_application.json line 199, extra comma is breaking migrate

PR #20913, opened 2020-03-12
fix: Party Name not visible in statement of accounts

deepeshgarg007:
Before:
image

After:
image


erpnext/accounts/report/general_ledger/general_ledger.html
```diff
@@ -1,8 +1,6 @@
 <h2 class="text-center">{%= __("Statement of Account") %}</h2>
 <h4 class="text-center">
-	{% if (filters.party_name) { %}
```
nabinhait, line 3: This has been done purposefully, in case there is naming series for customer, it should show customer_name value instead of ID.

nabinhait (changes requested):

erpnext/accounts/report/general_ledger/general_ledger.html
```diff
@@ -1,8 +1,6 @@
...
-		{%= filters.party_name %}
-	{% } else if (filters.party && filters.show_name) { %}
```
nabinhait, line 5: show_name can be the reason

nabinhait (changes requested):

PR #20902, opened 2020-03-11
fix: project link not set in accounts other than profilt and loss acc

nextchamp-saqib:
- Item wise projects are prioritised : Setting project in Sales Invoice Item will not be overwritten by Sales Invoice project.
- Projects & Cost Centres can be linked to all types of accounts while making GL Entry/Journal Entry

PR #20926, opened 2020-03-12
fix: project link not set in accounts other than profilt and loss acc 

nextchamp-saqib:
BackPort of #20902


deepeshgarg007:
@nextchamp-saqib please resolve merge conflicts

PR #20934, opened 2020-03-13
refactor: leave approval

Anurag810:
#20933


nabinhait (commented):
@Anurag810 We have removed `validate_leave_approver` method purposefully. So that user can create Workflow as per the use case. The Leave approver field is also non-mandatory for the same purpose.

If anyone don't want self-approval, the user can create Workflow around it.

Closing it

PR #20943, opened 2020-03-16
feat: added membership creation API

scmmishra:
This PR has the following changes
1. Set member email as unique
1. Created API for creation of membership

Version 12 PR: https://github.com/frappe/erpnext/pull/20944


erpnext/non_profit/doctype/membership/membership.py
```diff
@@ -50,4 +50,33 @@ def on_payment_authorized(self, status_changed_to=None):
...
+	member_list = frappe.get_all("Member", filters={'email': email})
+	if len(member_list):
+		return member_list[0]['name']
+	else:
+		new_member = frappe.new_doc('Member')
+		new_member.member_name = name
+		new_member.membership_type = membership_type
+		new_member.email = email
+		new_member.insert()
+		new_member.load_from_db()
```
nabinhait: why load_from_db?

erpnext/non_profit/doctype/membership/membership.py
```diff
@@ -50,4 +50,33 @@ def on_payment_authorized(self, status_changed_to=None):
 			self.load_from_db()
 			self.db_set('paid', 1)
 
+@frappe.whitelist()
+def create_membership(name, email, membership_type, from_date, to_date, currency, amount, paid):
```
nabinhait: So many mandatory args?

erpnext/non_profit/doctype/membership/membership.py
scmmishra: Removed this

erpnext/non_profit/doctype/membership/membership.py
scmmishra: Set a few to defaults, but most of these args are required

PR #20966, opened 2020-03-17
Check order as delivered

Dharmraj-48:
Added Condition to check if order is already delivered then prompt the error message

Screenshot from 2020-03-17 14-48-28

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order_list.js
+++ b/erpnext/selling/doctype/sales_order/sales_order_list.js
@@ -47,6 +47,46 @@ frappe.listview_settings['Sales Order'] = {
 		listview.page.add_menu_item(__("Re-open"), function() {
 			listview.call_for_selected_items(method, {"status": "Submitted"});
 		});
+	},
 
+	onload: function(doclist) {
+		const action = () => {
+			const selected_docs = doclist.get_checked_items();
+			const docnames = doclist.get_checked_items(true);
+
+			if (selected_docs.length > 0) {
+				for (let doc of selected_docs) {
+					debugger;
+					console.log(doc.docstatus);
+					if (!doc.docstatus) {
+						frappe.throw(__("Cannot create a Pick List from Draft documents."));
+					}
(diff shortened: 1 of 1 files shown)


gavindsouza:
@Dharmraj-48 how many commits are supposed to be a part of this PR?

Dharmraj-48:
commits are fixed for
https://github.com/frappe/erpnext/pull/20966

erpnext/selling/doctype/sales_order/sales_order_list.js
```diff
@@ -47,6 +47,46 @@ frappe.listview_settings['Sales Order'] = {
...
+	},
+
+	onload: function(doclist) {
+		const action = () => {
+			const selected_docs = doclist.get_checked_items();
+			const docnames = doclist.get_checked_items(true);
+
+			if (selected_docs.length > 0) {
+				for (let doc of selected_docs) {
+					debugger;
```
gavindsouza, line 59: cleanup required?

```suggestion
```

erpnext/selling/doctype/sales_order/sales_order_list.js
```diff
@@ -47,6 +47,46 @@ frappe.listview_settings['Sales Order'] = {
...
+					console.log(doc.docstatus);
```
gavindsouza, line 60:
```suggestion
```

gavindsouza (changes requested):

marination:
1) I think a better approach to avoiding mapping delivered sales orders would be hiding the **Pick List** button under **Create** in a Sales Order if the order is 100% delivered.
2) When it comes to fetching Sales Orders into a Pick List, it can be done from a new **Pick List** itself. A filtered list of Sales Orders is fetched to select so you don't have to check for delivery status again.
3) Adding only a creation provision for Pick List in the List View makes the other actions seem less important. Invoice, Delivery Note, Work Order creation from Sales Order are very frequent uses too. So i wouldn't handle this through List View.

PR #20980, opened 2020-03-18
Check order as delivered hotfix

Dharmraj-48:
Added condition so that  while picking list it should so error message for already delivered  orders
Screenshot from 2020-03-17 14-48-28


marination:
Left a comment on https://github.com/frappe/erpnext/pull/20966

PR #20992, opened 2020-03-19
fix: bom comparison issue

rohitwaghchaure:
**Issue**
If bom 1 and bom 2 are same then system throwing below error

Screenshot 2020-03-19 at 9 41 46 AM

**After Fix**
User not able to select the same BOM in BOM 1 and BOM 2
image


deepeshgarg007 (commented):
@rohitwaghchaure There should be a client side validation as well, same error still occurs if same BOM is selected.

image

deepeshgarg007 (approved):

PR #21006, opened 2020-03-19
ERPNext v12 Timesheet Wrongly Computes Hours Worked

ghost:
We are trying out ERPNext cloud for a prospect but had an issue with the Project module Timesheet. We realized hours computed are inaccurate.

When we entered the time from **8am to 5pm** same day (1st March), system computed 3 hours. This should be 9 hours. We tried to change the hours, but system would not allow the change.

See screenshot below:


deepeshgarg007 (commented):
@Judesque Please rebase from `develop` branch and reopen again

clarkejj:
No Project is specified here - please state exactly what steps you took to create the Timesheet, as somewhere you should get a message prompt validation to specify the Project?

PR #21011, opened 2020-03-20
fix: Add currency in options of Opportunity Amount

surajshetty3416:
- To show a correct symbol in list view

**Before:**
Screenshot 2020-03-20 at 12 53 50 PM

**After:**
Screenshot 2020-03-20 at 12 53 25 PM


deepeshgarg007 (approved):

deepeshgarg007:
@surajshetty3416 also send PR for `version-12-hotfix` and `version-11-hotfix`

PR #21013, opened 2020-03-20
fix: develop version as 12.0.0-dev instead of 12.2.0

sahil28297:
image


revant:
works

```python
In [1]: import semantic_version

In [2]: semantic_version.Version("12.0.0-dev")
Out[2]: Version('12.0.0-dev')

```

PR #21023, opened 2020-03-22
Abederp

aaabed-mis:

> Screenshots/GIFs


deepeshgarg007 (commented):
@aaabed-mis Please don't raise a PR directly against `version-12` branch. Since this is a new doctype I believe this should go in develop branch.  Also please explain what is the purpose of this new doctype. Is it required in core erpnext or specific only to your requirement and can be handled using a custom doctype

PR #21030, opened 2020-03-23
Changed label of payroll frequency hotfix

Dharmraj-48:
Replace label from
Fortnightly to Every two weeks
Bimonthly to Every two months

Screenshot from 2020-03-19 15-48-32


nabinhait (changes requested):
Fix as per https://github.com/frappe/erpnext/pull/21031#pullrequestreview-384596702

nabinhait (changes requested):
@Dharmraj-48 Please reopen the PR after fixing as per suggestion

PR #21052, opened 2020-03-25
feat(selling): added delivery window in sales order

MyuddinKhatri:
delivery_window
added delivery window in sales order where default delivery window time is fetched from customer if not set by customer then fetched from delivery settings and can be changed for particular delivery
trip in sales-order


nabinhait (commented):
Thanks for the PR.

But it does not look like a very generic requirement and can be managed via custom fields and Fetch From property very easily.

Closing this PR.

PR #21054, opened 2020-03-25
fix: Update Received Qty in Material Request as per Stock UOM

marination:
- **Received Qty** in Material Request Item must take value of **Qty as per Stock UOM** from Purchase Receipt Item (to handle multi uom cases)
- Material Requests show as Partially Received even though fully received due to having qty value from purchase receipt item (will differ if multi uom is used)
- Patch to update **received_qty** in Material Requests linked to Purchase Receipts


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, *args, **kwargs):
 			'target_parent_dt': 'Material Request',
 			'target_parent_field': 'per_received',
 			'target_ref_field': 'qty',
```
nabinhait:
```suggestion
			'target_ref_field': 'stock_qty',
```

erpnext/patches/v12_0/set_received_qty_in_material_request_as_per_stock_uom.py
```diff
@@ -0,0 +1,30 @@
...
+		doc = frappe.get_doc("Purchase Receipt", pr)
+		doc.status_updater = [
+			{
+			'source_dt': 'Purchase Receipt Item',
+			'target_dt': 'Material Request Item',
+			'join_field': 'material_request_item',
+			'target_field': 'received_qty',
+			'target_parent_dt': 'Material Request',
+			'target_parent_field': 'per_received',
+			'target_ref_field': 'qty',
```
nabinhait:
```suggestion
			'target_ref_field': 'stock_qty',
```

erpnext/patches/v12_0/set_received_qty_in_material_request_as_per_stock_uom.py
```diff
@@ -0,0 +1,30 @@
...
+			'source_field': 'stock_qty',
+			'percent_join_field': 'material_request'
+			}
+		]
+		doc.update_prevdoc_status()
```
nabinhait:
```suggestion
		doc.update_qty()
```

nabinhait (changes requested):

nabinhait (approved):

PR #21055, opened 2020-03-25
fix: Update Received Qty in Material Request as per Stock UOM

marination:
Port of https://github.com/frappe/erpnext/pull/21054


nabinhait (changes requested):
Check comments on https://github.com/frappe/erpnext/pull/21054#pullrequestreview-384587626

nabinhait (approved):

PR #21061, opened 2020-03-25
fix: Adding proper error message

anupamvs:
image

image


erpnext/accounts/party.py
```diff
@@ -303,7 +303,13 @@ def validate_party_accounts(doc):
 			company_default_currency = frappe.db.get_value('Company', account.company, "default_currency")
 
 		if existing_gle_currency and party_account_currency != existing_gle_currency:
-			frappe.throw(_("Accounting entries have already been made in currency {0} for company {1}. Please select a receivable or payable account with currency {0}.").format(existing_gle_currency, account.company))
+			if doc.doctype == 'Customer':
```
deepeshgarg007: @anupamvs There is a `validate_party_gle_currency` function just above this function which can be used here to validate party gle currency and avoid code duplication.

deepeshgarg007 (changes requested):

erpnext/accounts/party.py
anupamvs: @deepeshgarg007 Done!

deepeshgarg007 (dismissed):

deepeshgarg007 (approved):

PR #21062, opened 2020-03-26
fix: Add missing account settings link in accounts module

P-Froggy:
Add missing accounts settings link in accounts and settings module which was only accessible through search before

Anmerkung 2020-03-26 011256


deepeshgarg007 (commented):
@P-Froggy On develop branch Account Settings has already been added in the new desk. Closing this

PR #21063, opened 2020-03-26
fix: Add missing account settings link in accounts module

P-Froggy:
Add missing accounts settings link in accounts and settings module which was only accessible through search before

Anmerkung 2020-03-26 011256

### DIFF
--- a/erpnext/config/accounts.py
+++ b/erpnext/config/accounts.py
@@ -391,6 +391,11 @@ def get_data():
 			"items": [
 				{
 					"type": "doctype",
+					"name": "Accounts Settings",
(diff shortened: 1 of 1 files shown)


erpnext/config/accounts.py
```diff
@@ -389,6 +389,11 @@ def get_data():
 			"label": _("Settings"),
 			"icon": "fa fa-cog",
 			"items": [
+				{
```
deepeshgarg007, line 392: @P-Froggy Account Settings is already there under "Accounting Masters" section. Please check line no. 149. Closing this

PR #21079, opened 2020-03-27
fix: cannot set warehouse on deleting all so items and updating them

nextchamp-saqib:
BackPort of #21078


nabinhait (changes requested):
travis failing

marination (approved):

PR #21094, opened 2020-03-27
fix: Currency validation for default company accounts

deepeshgarg007:
Ideally the default accounts selected in Company Master should have the same currency as the company's default currently.

Assuming this the default accounts are fetched in multiple transactions but currently there is no validation as to which default accounts I can select in company master and this leads to incorrect calculations in many transactions.

This PR solves this issue


erpnext/setup/doctype/company/company.py
```diff
@@ -73,18 +74,22 @@ def create_default_tax_template(self):
 
 	def validate_default_accounts(self):
 		accounts = [
-			"default_bank_account", "default_cash_account",
-			"default_receivable_account", "default_payable_account",
-			"default_expense_account", "default_income_account",
-			"stock_received_but_not_billed", "stock_adjustment_account",
-			"expenses_included_in_valuation", "default_payroll_payable_account"
+			["Default Bank Account", "default_bank_account"], ["Default Cash  Account", "default_cash_account"],
```
nextchamp-saqib, line 77: You can make use of frappe.unscrub()

erpnext/setup/doctype/company/company.py
```diff
@@ -73,18 +74,22 @@ def create_default_tax_template(self):
...
-				for_company = frappe.db.get_value("Account", self.get(field), "company")
+		for account in accounts:
+			if self.get(account[1]):
+				for_company = frappe.db.get_value("Account", self.get(account[1]), "company")
 				if for_company != self.name:
-					frappe.throw(_("Account {0} does not belong to company: {1}").format(self.get(field), self.name))
+					frappe.throw(_("Account {0} does not belong to company: {1}").format(self.get(account[1]), self.name))
+
+				if get_account_currency(self.get(account[1])) != self.default_currency:
+					frappe.throw(_("""{0} currency must be same as company's default currency.
```
nextchamp-saqib, line 91: What if `account[0] = 'Stock Received But Not Billed'`?

deepeshgarg007: I guess the new change should solve both the issues, since unscrub can't  be used now

PR #21097, opened 2020-03-28
fix: item code showing as mandatory even if the 'Item Naming By' is set as Naming Series in stock settings

rohitwaghchaure:
**Issue**

Screenshot 2020-03-28 at 2 31 32 PM

**After Fix**

Screenshot 2020-03-28 at 2 54 29 PM


marination:
Why are we showing item code field at all in this case? Even if we set **item code** the **naming series** takes the value of `item_code` on save so item code isn't used here at all.

PR #21098, opened 2020-03-28
fix: item code showing as mandatory even if the 'Item Naming By' is set as Naming Series in stock settings

rohitwaghchaure:
Backport https://github.com/frappe/erpnext/pull/21097


marination:
Comment on https://github.com/frappe/erpnext/pull/21097

PR #21100, opened 2020-03-28
fix: payment request status fixes

nextchamp-saqib:
Problem: 
Creating a payment request and then fulfilling it with partial payment entry, the status of payment request changes to Paid.


deepeshgarg007 (changes requested):
@nextchamp-saqib Please resolve merge conflicts. Also logic for on_cancel is missing. If payment entry against an Order is cancelled the status in Payment Request remains the same.

deepeshgarg007 (approved):

PR #21102, opened 2020-03-28
fix: no server side validations for accounts in asset category

nextchamp-saqib:

PR #21105, opened 2020-03-28
fix: no server side validations for accounts in asset category

nextchamp-saqib:
BackPort of #21102


erpnext/assets/doctype/asset_category/asset_category.py
```diff
@@ -11,12 +11,29 @@
...
+			'accumulated_depreciation_account': { 'account_type': 'Accumulated Depreciation' },
+			'depreciation_expense_account': { 'account_type': 'Expense' },
+			'capital_work_in_progress_account': { 'account_type': 'Capital Work in Progress' }
+		}
+		for d in self.accounts:
+			for account in account_type_map.keys():
+				if d.get(account):
+					account_type = frappe.db.get_value('Account', d.get(account), 'account_type')
+					if account_type != account_type_map[account]['account_type']:
+						frappe.throw(_("Row {}: {} should be a {} account".format(d.idx, frappe.bold(frappe.unscrub(account)),
```
deepeshgarg007: The error message doesn't sounds right, For Eg: If account type of fixed asset account is incorrect then the error message will be shown as "Row 1: **Fixed Asset Account** should be a **Fixed Asset** account".  The user won't understand the mistake. Account Type should be explicitly mentioned like "Account type for {} should be {}"

deepeshgarg007 (changes requested):

erpnext/assets/doctype/asset_category/asset_category.py
nextchamp-saqib: Great Review! Thanks.

deepeshgarg007:
@nextchamp-saqib Test Cases are failing. Please fix

deepeshgarg007 (approved):

PR #21108, opened 2020-03-28
fix: auto created asset message

nextchamp-saqib:
Display up to 5 asset links in the message on auto asset creation


erpnext/controllers/buying_controller.py
```diff
@@ -672,19 +672,32 @@ def auto_make_assets(self, asset_items):
...
 						for qty in range(cint(d.qty)):
-							self.make_asset(d)
-						is_plural = 's' if cint(d.qty) != 1 else ''
-						messages.append(_('{0} Asset{2} Created for <b>{1}</b>').format(cint(d.qty), d.item_code, is_plural))
+							asset = self.make_asset(d)
+							created_assets.append(asset)
+						
+						if len(created_assets) > 5:
+							# dont show asset form links if more than 5 assets are created
+							messages.append(_('{} Assets created for {}').format(len(created_assets), is_plural, frappe.bold(d.item_code)))
```
nabinhait:
```suggestion
							messages.append(_('{} Asset{} created for {}').format(len(created_assets), is_plural, frappe.bold(d.item_code)))
```

nabinhait (changes requested):

nabinhait (approved):

PR #21117, opened 2020-03-29
fix: customer group price list not fetched in pos

nextchamp-saqib:
BackPort of #21096

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -429,13 +429,16 @@ class SalesInvoice(SellingController):
 					if (not for_validate) or (for_validate and not self.get(fieldname)):
 						self.set(fieldname, pos.get(fieldname))
 
-			customer_price_list = frappe.get_value("Customer", self.customer, 'default_price_list')
-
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -429,13 +429,16 @@ def set_pos_fields(self, for_validate=False):
...
 
-			if not customer_price_list:
-				self.set('selling_price_list', pos.get('selling_price_list'))
+			customer_price_list, customer_group = frappe.get_value("Customer", self.customer, ['default_price_list', 'customer_group'])
+
+			customer_group_price_list = frappe.get_value("Customer Group", customer_group, 'default_price_list')
+
+			selling_price_list = customer_price_list or customer_group_price_list or pos.get('selling_price_list')
+
+			self.set('selling_price_list', selling_price_list)
```
deepeshgarg007: Check if `selling_price_list` exists. What if none of the above price list is specified. It will set as 'None'

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #21122, opened 2020-03-30
feat: Link blanket order and quotation

deepeshgarg007:
Similar to Sales Order now a Quotation can also be created against a blanket order. Once a Sales Order is created against that Quotation, ordered quantity will be updated in Blanket Order

[Documentation Link](https://github.com/frappe/erpnext_documentation/pull/7)


marination:
Add validation for an item added multiple times

erpnext/manufacturing/doctype/blanket_order/blanket_order.py
```diff
@@ -17,7 +17,16 @@ def validate(self):
 
 	def validate_dates(self):
 		if getdate(self.from_date) > getdate(self.to_date):
-			frappe.throw(_("From date cannot be greater than To date")) 
+			frappe.throw(_("From date cannot be greater than To date"))
+
+		self.validate_duplicate_items()
```
marination: why is this inside `validate_dates` ?

marination (changes requested):

marination (approved):

marination:
https://github.com/frappe/erpnext_documentation/pull/7

PR #21123, opened 2020-03-31
fix: docfield of sales_order is not fetching route options for new doc

AndyOverLord:

I found that when clicking 'new doc' in the link field, the new doc form is not fetching the route options predefined.

The source code is using the wrong method to get `so` as docfield.
`frappe.meta.get_docfield` is not referring to the same docfield compared with `frm.get_docfield` in the doctype form.

I saw other places where using get_route_options_for_new_doc, docfield is using `frm.get_docfield` instead of `frappe.meta.get_docfield` otherwise the `get_route_options_for_new_doc` is not attached in that 'so' docfield.

[Reference: transaction.js](https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/transaction.js#L169)

### DIFF
--- a/erpnext/projects/doctype/project/project.js
+++ b/erpnext/projects/doctype/project/project.js
@@ -18,7 +18,7 @@ frappe.ui.form.on("Project", {
 		};
 	},
 	onload: function (frm) {
-		var so = frappe.meta.get_docfield("Project", "sales_order");
+		var so = frm.get_docfield("Project", "sales_order");
 		so.get_route_options_for_new_doc = function (field) {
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

deepeshgarg007:
@AndyOverLord `get_docfield` accepts only fieldnames and  is supposed to work only for child tables
Current fix is not correct

Screenshot 2020-06-09 at 5 29 07 PM

deepeshgarg007:
Reverting this

PR #21130, opened 2020-03-31
fix(patch): set_task_status

barredterra:
`for doctype in ('Task')` iterates over T, a, s, k. 
I think that's no how it was intended?


erpnext/patches/v12_0/set_task_status.py
```diff
@@ -4,7 +4,7 @@ def execute():
     frappe.reload_doctype('Task')
 
     # add "Completed" if customized
-    for doctype in ('Task'):
+    for doctype in ['Task']:
```
nabinhait: why for loop for a single item?

nabinhait (changes requested):

erpnext/patches/v12_0/set_task_status.py
barredterra: I dont' know :man_shrugging: . I'll remove the loop.

nabinhait (approved):

nabinhait (approved):

PR #21139, opened 2020-04-01
fix: list index out of range bug

scmmishra:
`net_profit[-1]` would throw an error if no values were present. This PR fixes that.


deepeshgarg007 (commented):
@prssanna has already made a fix for it in this [PR](https://github.com/frappe/erpnext/pull/20876). Closing this PR. If you feel any change is required, make them in that PR itself

PR #21144, opened 2020-04-01
Sum of years not needed.

anupamvs:
**Before:**
image

**Now:**
image


erpnext/selling/report/customer_acquisition_and_loyalty/customer_acquisition_and_loyalty.py
```diff
@@ -48,7 +48,7 @@ def execute(filters=None):
 			new = new_customers_in.get(key, [0,0.0])
 			repeat = repeat_customers_in.get(key, [0,0.0])
 
-			out.append([year, calendar.month_name[month],
+			out.append([str(year), calendar.month_name[month],
```
rohitwaghchaure: use cstr

PR #21167, opened 2020-04-06
fix: Validate Serial/Batch No naming series in Item itself

marination:
- Invalid Serial / Batch No naming series breaks elsewhere (while doing a stock transaction). The user has to navigate back to the Item to fix it.
image

- Validate In Item itself.
Screenshot 2020-04-06 at 1 58 41 PM

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -525,7 +525,8 @@ class Item(WebsiteGenerator):
 
 	def validate_naming_series(self):
 		for field in ["serial_no_series", "batch_number_series"]:
-			if self.get(field) and "." not in self.get(field):
+			series = self.get(field)
+			if series and "#" in series and "." not in series:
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item.py
```diff
@@ -522,6 +523,12 @@ def validate_item_type(self):
 		if self.has_serial_no == 0 and self.serial_no_series:
 			self.serial_no_series = None
 
+	def validate_naming_series(self):
+		for field in ["serial_no_series", "batch_number_series"]:
+			if self.get(field) and "." not in self.get(field):
```
nabinhait:
```suggestion
			series = self.get(field)
			if series and "#" in series and "." not in series:
```

nabinhait (changes requested):

erpnext/stock/doctype/item/item.py
marination: done

nabinhait (approved):

PR #21171, opened 2020-04-06
chore: rename bank reconciliation to bank clearance

nextchamp-saqib:


erpnext/patches/v12_0/rename_bank_reconciliation.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2018, Frappe and Contributors
```
rohitwaghchaure, line 1: Add this patch in patches.txt

rohitwaghchaure: Also need to delete Bank Reconciliation from the doctype
frappe.delete_doc("DocType", "Bank Reconciliation")

Doctype is still present even after running the patch on the local system 
Screenshot 2020-04-15 at 6 40 30 PM

rohitwaghchaure (approved):

PR #21179, opened 2020-04-06
fix: Rename bank account type doctypes

deepeshgarg007:
Renamed "Account Type" to "Bank Account Type"
Renamed "Account Subtype" to "Bank Account Subtype"


nextchamp-saqib (approved):

gavindsouza:
Renaming doesn't handle Custom DocPerm documents (and don't know if there are any other such hidden ones) as of now. We will need to add a patch to fix this!

It doesn't because there's no link field. Planning on changing that now...This means future migrations should work just fine. But this patch will require an amendment.

PR #21190, opened 2020-04-07
fix: cannot find accounting module while rendering breadcrumb

nextchamp-saqib:
Problem:
```
frappe.breadcrumbs.module_map = {
...
'Accounts': 'Accounting'
...
}
```

breadcrumb.js
```
// breadcrumbs.module = 'Accounts'

if (frappe.breadcrumbs.module_map[breadcrumbs.module]) {
	breadcrumbs.module = frappe.breadcrumbs.module_map[breadcrumbs.module];

        // Now breadcrumbs.module = 'Accounts'
}

if(frappe.get_module(breadcrumbs.module)) {    // cannot find module named 'Accounting'
....
}
```

### DIFF
--- a/erpnext/public/js/conf.js
+++ b/erpnext/public/js/conf.js
@@ -43,7 +43,6 @@ $.extend(frappe.breadcrumbs.preferred, {
 $.extend(frappe.breadcrumbs.module_map, {
 	'ERPNext Integrations': 'Integrations',
 	'Geo': 'Settings',
-	'Accounts': 'Accounting',
 	'Portal': 'Website',
 	'Utilities': 'Settings',
 	'Shopping Cart': 'Website',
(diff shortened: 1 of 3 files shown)


erpnext/public/js/conf.js
```diff
@@ -43,7 +43,6 @@ $.extend(frappe.breadcrumbs.preferred, {
 $.extend(frappe.breadcrumbs.module_map, {
 	'ERPNext Integrations': 'Integrations',
 	'Geo': 'Settings',
-	'Accounts': 'Accounting',
```
nabinhait, line 46: Is this intentional change?

nabinhait (changes requested):

erpnext/public/js/conf.js
nextchamp-saqib: Yes

nabinhait (approved):

PR #21197, opened 2020-04-07
feat: Provision to set Default Item Manufacturer

marination:
- **Is Default** checkbox added in Item Manufacturer.
Screenshot 2020-04-07 at 9 10 36 PM

- **Default Item Manufacturer** and **Default Manufacturer Part** No fields added to Item Master
Screenshot 2020-04-07 at 9 11 56 PM

- **Manufacturer Part No** field is now editable in all child tables with validation ( if invalid Part Nos are added)
ok

 
- Manufacturer and Part No auto fetched via `get_item_details` in child tables on setting item code.

**Minor Notes:**
- If is_default is unchecked and then saved , it will be cleared from Item Master if it was set there.
- If the default Item Manufacturer is deleted, it is again cleared from Item Master if it was set there.

### DIFF
--- a/erpnext/public/js/controllers/buying.js
+++ b/erpnext/public/js/controllers/buying.js
@@ -379,7 +379,31 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 				}
 			});
 		}
-	}
+	},
+
+	manufacturer_part_no: function(doc, cdt, cdn) {
+		const row = locals[cdt][cdn];
+
+		if (row.manufacturer_part_no) {
+			frappe.model.get_value('Item Manufacturer',
+				{
+					'item_code': row.item_code,
+					'manufacturer': row.manufacturer,
+					'manufacturer_part_no': row.manufacturer_part_no
+				},
+				'name',
+				function(data) {
+					if (!data) {
+						let msg = {
+							message: __("Manufacturer Part Number <b>{0}</b> is invalid", [row.manufacturer_part_no]),
+							title: __("Invalid Part Number")
+						}
+						frappe.throw(msg);
+					}
+				});
+
+			}
+		}
 });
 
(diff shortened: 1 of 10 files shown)


erpnext/public/js/controllers/buying.js
```diff
@@ -379,7 +379,31 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+				},
+				'name',
+				function(data) {
+					if (!data) {
+						let msg = {
+							message: __("Manufacturer Part Number <b>{0}</b> is invalid", [row.manufacturer_part_no]),
+							title: __("Invalid Part Number")
+						}
+						frappe.throw(msg);
+					}
```
rohitwaghchaure, line 402: Is this method just to check whether manufacturer_part_no is correct or not? If yes then better you change the fieldtype to select for manufacturer_part_no and add respective item's manufacture part no dynamically.

marination: done

erpnext/public/js/controllers/transaction.js
```diff
@@ -1952,3 +1956,31 @@ erpnext.show_serial_batch_selector = function (frm, d, callback, on_close, show_
 		}, show_dialog);
 	});
 }
+
+erpnext.set_manufacturer_part_no = function(frm, item, default_part_no) {
+	var df = frm.fields_dict["items"].grid.get_docfield("manufacturer_part_no");
+	var options  = default_part_no ? default_part_no : "";
```
rohitwaghchaure: User frm.set_df_property("manufacturer_part_no", "options", [default_part_no]);

rohitwaghchaure: No need of frappe.db.get_list("Item Manufacturer", {

rohitwaghchaure (approved):

PR #21200, opened 2020-04-08
fix: item price without any party not fetching

nextchamp-saqib:
Consider following problem:

- A customer CUS-001 has default price list Retail USD
- Item Price of ITM-001 has price list set as Retail USD
- Creating a quotation against CUS-001 for the item ITM-001 should fetch the item price
- The query responsible for fetching the item price has a condition `customer='CUS-001'`,  hence the item price is not fetched.

This PR helps finding the item prices against the customer or without customer. It orders the item price by customer so that item price having customer value set, will be prioritised.


marination (changes requested):
Please fix travis

PR #21203, opened 2020-04-08
fix: minor issues

rohitwaghchaure:
1. 
```
  File "/home/frappe/benches/bench-version-12-2020-03-19/apps/erpnext/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py", line 63, in get_coa
    forest = build_forest(generate_data_from_csv(file_name))
  File "/home/frappe/benches/bench-version-12-2020-03-19/apps/erpnext/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py", line 121, in build_forest
    path = return_parent(data, account_name)[::-1]
  File "/home/frappe/benches/bench-version-12-2020-03-19/apps/erpnext/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py", line 101, in return_parent
    frappe.throw(_("The parent account {0} does not exists")
TypeError: 'str' object is not callable
```

2. While creating a Stock Reconciliation, by default the option chosen for Purpose field is Stock Reconciliation. Since, a person might not even know there exists an option for Opening Entry. To fix this issue now onward, the purpose field will be blank and user has to select the purpose manually.

### DIFF
--- a/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
+++ b/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
@@ -97,7 +97,7 @@ def build_forest(data):
 				return [parent_account]
 			elif account_name == child:
 				parent_account_list = return_parent(data, parent_account)
-				if not parent_account_list:
+				if not parent_account_list and parent_account:
 					frappe.throw(_("The parent account {0} does not exists")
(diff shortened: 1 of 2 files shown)


vishdha:
TypeError: 'str' object is not callable frappe.throw cause problem

nabinhait (approved):

PR #21209, opened 2020-04-08
fix: Project Update Email Error

marination:
Port of https://github.com/frappe/erpnext/pull/21208

### DIFF
--- a/erpnext/projects/doctype/project/project.json
+++ b/erpnext/projects/doctype/project/project.json
@@ -1,5 +1,4 @@
 {
- "actions": [],
  "allow_import": 1,
  "allow_rename": 1,
  "autoname": "field:project_name",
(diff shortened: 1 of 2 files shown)


erpnext/projects/doctype/project/project.json
```diff
@@ -438,13 +439,15 @@
    "description": "Message will sent to users to get their status on the project",
    "fieldname": "message",
    "fieldtype": "Text",
-   "label": "Message"
+   "label": "Message",
+   "mandatory_depends_on": "collect_progress"
```
hrwX: @marination v12 doesn't have `mandatory_depends_on` feature. You'll have to use `set_df_property` and set `reqd` if `collect_progress` is checked.

hrwX (changes requested):

erpnext/projects/doctype/project/project.json
marination: @hrwX  Done

nabinhait (approved):

Issue #21210, opened 2020-04-08
Account (Child) Table in Journal Entry Form not updated on Entry Type change


abhishekbalam:
## Description of the issue

In Journal Entry Form, when Entry Type is changed more than once the child table fields are not updated. 

## Context information (for bug reports)

This could cause ledger entry errors if user is not aware of the 
right accounts to enter manually for each entry type.

## Steps to reproduce the issue

1. New Journal Entry Form
2. Change Entry Type for once (for eg to Bank Entry)
3. Now change it to something else (for eg to Opening Entry)
4. Old fields still exist and are not updated

### Observed result

The child table fields are relevent to Bank Entry.

### Expected result

Child table should have fields relevent to Opening Entry.

PR #21211, opened 2020-04-08
fix: Accounts Table not updating on changing Entry Type in Journal Entry Form.

abhishekbalam:
Issue: [21210](https://github.com/frappe/erpnext/issues/21210)

Child table (Journal Entry Account) values were not updating
when Entry Type is changed.

Now child table is cleared on Entry Type value change, and
new values are showing.

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -120,6 +120,10 @@ frappe.ui.form.on("Journal Entry", {
 				}
 			}
 		});
+	},
+	voucher_type: function(frm) {
+		frm.fields_dict.accounts.frm.clear_table('accounts');
+		frm.fields_dict.accounts.frm.refresh_fields();
 	}
 });
 


abhishekbalam:
I know that clear_table() should have been put in block starting from line  [378](https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/journal_entry/journal_entry.js#L378) but this function is going to be removed soon.

deepeshgarg007 (commented):
@abhishekbalam The details are correctly fetched if the `Journal Entry Account` table is empty on selection of Entry Type. This behaviour is intentional. Suppose you have entered many rows in child table manually and forget to select the appropriate `Entry Type`, then on selection of rows all you entries will be overridden. Not a bug. Closing this PR

deepeshgarg007:
> I know that clear_table() should have been put in block starting from line [378](https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/journal_entry/journal_entry.js#L378) but this function is going to be removed soon.

Why it will be removed soon?

abhishekbalam:
> @abhishekbalam The details are correctly fetched if the `Journal Entry Account` table is empty on selection of Entry Type. This behaviour is intentional. Suppose you have entered many rows in child table manually and forget to select the appropriate `Entry Type`, then on selection of rows all you entries will be overridden. Not a bug. Closing this PR

Okay, this makes sense. Thanks for explaining 👍

PR #21215, opened 2020-04-09
fix(UX): batch error message improved

vishdha:


erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -88,9 +88,9 @@ def validate_item(self):
 		if self.voucher_type != 'Stock Reconciliation':
 			if item_det.has_batch_no ==1:
 				if not self.batch_no:
```
marination, line 91:
```py
batch_item = self.item_code if self.item_code == item_det.item_name else self.item_code + ":" +  item_det.item_name
```
use `batch_item` in frappe.throw in `if` and `elif`

marination (changes requested):

marination (approved):

PR #21216, opened 2020-04-09
feat(Healthcare): Rehabilitation Module

ruchamahabal:
Documentation PR: https://github.com/frappe/erpnext_documentation/pull/20

Rehab / Physiotherapy sub-module in ERPNext Healthcare

rehab-module

- **Exercise Type**:

    The heart of rehab is mapping exercises.
    1. Template for prescribing exercises
    2. Set Healthcare Service Units (will be marked occupied when therapy sessions for this type are 
         booked)
    3. Attach Videos or PDF for Exercise Steps
    4. Add or edit steps in cards (images, title, desc)

    exercise_type

<details>
<summary>Therapy Type</summary>

1. Template for Therapies
2. Item created automatically for billing
3. Configure exercises, rates, and duration

rehab_therapy_type

</details>

<details>
<summary>Therapy Plan</summary>

1. Add Therapies
2. Track progress according to sessions
3. Automatically created on Patient Encounter submission

therapy_plan

</details>

<details>
<summary>Therapy Session</summary>

1. Therapy Type wise Sessions
2. Number of therapies updated in Therapy Plan on submit
3. See counts progress

therapy_session

</details>
<details>
<summary>Therapy Plan from Encounter</summary>

therapy_plan_from_encounter
</details>

</details>
<details>
<summary>Appointment Booking from Therapies</summary>

app_rehab

</details>

<details>
<summary>Invoicing</summary>

get_healthcare_services

</details>

<details>
<summary>Patient Assessment Template</summary>

Create templates for any Patient Assessment:

patient-assessment-template

</details>

<details>
<summary>Patient Assessment</summary>

Create Patient Assessments from Template:

assessment

</details>

</details>

- [x] Tests
- [x] Documentation

PR #21234, opened 2020-04-10
fix: Move Video Doctype from Education module to Core

ruchamahabal:
Move the Video DocType from Education to Core for using it across multiple modules.
One use case in Exercise Type to add a video for Exercise: frappe/erpnext#21216

Depends on: https://github.com/frappe/frappe/pull/9910


nabinhait (approved):

ruchamahabal:

PR #21242, opened 2020-04-11
fix: on item change UOM not updated

### LINKED ISSUE #19483: bug: on Item change UOM not updated
**Environment :**
ERPNext: v12.1.8 (version-12)
Frappe Framework: v12.0.18 (version-12) 

Item Table of the below doctypes affected : 

1. Sales Order
2. Sales Invoice
3. Delivery Note
4. Purchase Order
5. Purchase Invoice

#### Steps to Reproduce  : 
Step 1: Select Item (e.g. Burger having default_uom as "Nos")
Step 2 : Replace Item with another item having different default_uom ,  e.g. Coke having default_uom as "Litre")
Step 3:  Stock UOM of the item table changes, but UOM doesn't

Below is the gif explaining the same.

ItemUOM_Error

#### Actual Behavior : 
Item UOM **_not_** updated on selection of item

#### Expected Behavior : 
Item UOM should get updated on selection of item

PS : It works fine in Stock Entry doctype.

anupamvs:
**before**
b

**after**
a

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -259,13 +259,14 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 		expense_account = get_asset_category_account(fieldname = "fixed_asset_account", item = args.item_code, company= args.company)
 
 	#Set the UOM to the Default Sales UOM or Default Purchase UOM if configured in the Item Master
-	if args.get('doctype') in sales_doctypes:
-		args.uom = item.sales_uom if item.sales_uom else item.stock_uom
-	elif (args.get('doctype') in ['Purchase Order', 'Purchase Receipt', 'Purchase Invoice']) or \
-		(args.get('doctype') == 'Material Request' and args.get('material_request_type') == 'Purchase'):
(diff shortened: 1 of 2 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -259,14 +259,13 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 		expense_account = get_asset_category_account(fieldname = "fixed_asset_account", item = args.item_code, company= args.company)
 
 	#Set the UOM to the Default Sales UOM or Default Purchase UOM if configured in the Item Master
-	if not args.get('uom'):
```
nabinhait: You cannot remove this condition. It will lead to overwriting UOM on server side from https://github.com/frappe/erpnext/blob/version-12-hotfix/erpnext/controllers/accounts_controller.py#L279

I think we should not pass uom from js (https://github.com/frappe/erpnext/blob/version-12-hotfix/erpnext/public/js/controllers/transaction.js#L502). But we need to test it properly, it can lead to other consequences.

nabinhait (changes requested):

PR #21252, opened 2020-04-13
fix: show positive taxes in credit notes

nextchamp-saqib:
Credit Notes will have positive taxes shown in print format. However...

Caveat:
Tax Breakup table will show positive values and taxes table will show negative values
image


deepeshgarg007 (commented):
I don't think its correct to change the amounts in the docs. Closing this

PR #21262, opened 2020-04-14
feat: BOM template

rohitwaghchaure:
## **BOM Template**
Screenshot 2020-04-14 at 2 37 57 PM

## **Make Work Order**

Make work order against variant item using BOM Template and also provision to select variant item against the template item raw materials

Screenshot 2020-04-14 at 2 43 55 PM

## **Make BOM For Variant Item Using BOM Template**

Screenshot 2020-04-14 at 2 44 09 PM


erpnext/manufacturing/doctype/bom/bom.json
```diff
@@ -498,6 +499,17 @@
...
+   "fieldname": "has_variants",
+   "fieldtype": "Check",
+   "in_list_view": 1,
+   "label": "Has Variants",
+   "no_copy": 1,
+   "print_hide": 1,
+   "read_only": 1
   }
  ],
  "icon": "fa fa-sitemap",
```
marination, line 515: change modified timestamp, changes are not reflected

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -897,3 +901,84 @@ def get_bom_diff(bom1, bom2):
 					out.removed.append([df.fieldname, d.as_dict()])
 
 	return out
+
+def item_query(doctype, txt, searchfield, start, page_len, filters):
```
marination, line 905: include condition such that main item in BOM is not selectable in child table.

marination (changes requested):

erpnext/manufacturing/doctype/bom/bom.json
rohitwaghchaure: Done

marination (approved):

PR #21265, opened 2020-04-14
fix: Chart of account importer UX improved

vishdha:
Before:
Chart of Accounts Importer

After:
Chart of Accounts Importer (1)

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -89,7 +89,7 @@ class Account(NestedSet):
 				throw(_("Root cannot be edited."), RootNotEditable)
 
 		if not self.parent_account and not self.is_group:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/account/account.py
```diff
@@ -89,7 +89,7 @@ def validate_root_details(self):
 				throw(_("Root cannot be edited."), RootNotEditable)
 
 		if not self.parent_account and not self.is_group:
-			frappe.throw(_("Root Account must be a group"))
+			frappe.throw(_("Root Account must be a group for {0}".format(frappe.bold(self.name))))
```
rohitwaghchaure, line 92: frappe.throw(_("Root Account {0} must be a group".format(frappe.bold(self.name))))

deepeshgarg007 (approved):

PR #21274, opened 2020-04-14
Update project.json

wojosc:
unnecessary requirement for project creation as this is a field of the Type select which will be pre-filled. When creating a project of list view it will also bring up this field as it has "reqd": 1. A new projekt will have the status set to open in most cases.


deepeshgarg007 (commented):
@wojosc Even after removing reqd check the status field will still be shown in Quick Entry, any other specific reason for removing this?

wojosc:
Hi @deepeshgarg007 thank you for your response. It looks to me like bloat for the moment. I do not think that anyone would create a procekt which a status any than "open". As this is pre set anyway in the select field I came to this conclusion.

The feedback I get a lot is that many fields can be a little confusing because there need is unclear.

hrwX (approved):

PR #21342, opened 2020-04-20
fix: Re-order Item Error Email format

marination:
**Before:**
Screenshot 2020-04-20 at 6 30 39 PM

**After:**
Screenshot 2020-04-20 at 6 13 10 PM

_Multiple Errors_
Screenshot 2020-04-20 at 6 34 51 PM

- Also changed UOM's integer validation error message for clarity.
Screenshot 2020-04-20 at 6 37 45 PM

PR #21343, opened 2020-04-20
fix: Re-order Item Error Email format

marination:
Port of https://github.com/frappe/erpnext/pull/21342

### DIFF
--- a/erpnext/stock/reorder_item.py
+++ b/erpnext/stock/reorder_item.py
@@ -4,6 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 import erpnext
+import json
 from frappe.utils import flt, nowdate, add_days, cint
 from frappe import _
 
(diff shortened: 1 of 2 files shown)


erpnext/stock/reorder_item.py
```diff
@@ -199,18 +200,16 @@ def send_email_notification(mr_list):
...
-
-Please rectify these issues:
----
-<pre>
-%s
-</pre>
----
-Regards,
-Administrator""" % ("\n\n".join(exceptions_list),)
+	content = """Dear System Manager, <br> An error occured for certain Items while creating Material Requests based on Re-order level.
```
nabinhait: Email messages should also be translated?

nabinhait (changes requested):

erpnext/stock/reorder_item.py
marination: done

nabinhait (approved):

PR #21353, opened 2020-04-21
fix: Missing leading cost center for test data

tchalupnik:
Fix test data for leading cost center saying:
Screenshot 2020-04-21 at 11 48 14


tchalupnik:
It seems my code doesn't solve problem if field parent is required.
However I still face problem where I cannot run tests because of that.
Is there any solution rather than having parent required off and my snippet?

deepeshgarg007 (commented):
@mistillate Yeah, this won't solve the issue. This cost center (By Company Name )is created by itself from setup at the time of company creation. Please update your branch, reinstall the site and run-test again. The test should go through. Closing this PR

PR #21357, opened 2020-04-21
feat: (minor) Display more informative message when work order can't be created

TurkerTunali:
ERPNext was showing just Operation Name when work order can't be created. With this PR, it also shows item's name and bom name. (When creating more than 10 work orders via production planning, operation name is not sufficient)

Old Way:
image

When Updated:
image

> Screenshots/GIFs


marination (commented):
"Operation Time must be greater than 0 for Operation {0} in BOM {1}." would be a better suggestion. Adding the Item seems redundant since the operation time is set in the BOM.
Also please send this PR on branch `version-12-hotfix` and `develop`. We don't merge PRs on `version-12`.

marination:
closing this

PR #21359, opened 2020-04-21
fix: patch and validation message to fix target warehouse issue

rohitwaghchaure:
**Issue**

User has created user permission against warehouse
image

**In sales invoice**
User has enabled the update stock, due to user permissions issue target warehouse has set (even if the target warehouse has ignore user permissions)
image

**Incorrect Stock Ledger**

As target warehouse set system has created another stock ledger entry
Screenshot 2020-04-21 at 2 23 46 PM

**After Fix**

Screenshot 2020-04-22 at 2 17 45 AM

### DIFF
--- a/erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
+++ b/erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
@@ -1,0 +1,71 @@
+# Copyright (c) 2020, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	warehouse_perm = frappe.get_all("User Permission",
+		fields=["count(*) as p_count", "is_default", "user"], filters={"allow": "Warehouse"}, group_by="user")
+
+	if not warehouse_perm:
+		return
+
+	execute_patch = False
+	for perm_data in warehouse_perm:
+		if perm_data.p_count == 1 or (perm_data.p_count > 1 and frappe.get_all("User Permission",
+			filters = {"user": perm_data.user, "allow": "warehouse", "is_default": 1}, limit=1)):
+			execute_patch = True
+			break
+
+	if not execute_patch: return
+
+	for doctype in ["Sales Invoice", "Delivery Note"]:
+		if not frappe.get_meta(doctype + ' Item').get_field("target_warehouse").hidden: continue
+
+		cond = ""
+		if doctype == "Sales Invoice":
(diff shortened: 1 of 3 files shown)


rohitwaghchaure:
Dependency https://github.com/frappe/frappe/pull/10029

erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
```diff
@@ -0,0 +1,70 @@
...
+		if doctype == "Sales Invoice":
+			cond = " AND parent_doc.update_stock = 1"
+
+		data = frappe.db.sql(""" SELECT parent_doc.name as name, child_doc.name as child_name
+			FROM
+				`tab{doctype}` parent_doc, `tab{doctype} Item` child_doc
+			WHERE
+				parent_doc.name = child_doc.parent AND parent_doc.docstatus = 1
+				AND child_doc.target_warehouse is not null AND child_doc.target_warehouse != ''
+				AND child_doc.target_warehouse = child_doc.warehouse AND child_doc.creation > '2020-04-16' {cond}
```
nabinhait: why this condition`child_doc.target_warehouse = child_doc.warehouse`?

erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
```diff
@@ -0,0 +1,70 @@
...
+		for d in parent_names:
+			doc = frappe.get_doc(doctype, d)
+			doc.docstatus = 2
+			doc.update_stock_ledger()
+			doc.make_gl_entries_on_cancel(repost_future_gle=False)
+
+			# update stock & gl entries for submit state of PR
+			doc.docstatus = 1
+			doc.update_stock_ledger()
+			doc.make_gl_entries()
```
nabinhait: This block should only be executed if data.

nabinhait (changes requested):

erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
rohitwaghchaure: Because due to permission issue both warehouses has set same, in reality no one will set the same warehouse for delivery and customer warehouse

erpnext/patches/v12_0/repost_stock_ledger_entries_for_target_warehouse.py
rohitwaghchaure: Fixed

PR #21373, opened 2020-04-22
fix: Missing backlink in dashboard

wojosc:
Added Email Campaign Link to Dashboard. Placing should be discussed As there are no different sections for the moment.

### DIFF
--- a/erpnext/crm/doctype/lead/lead_dashboard.py
+++ b/erpnext/crm/doctype/lead/lead_dashboard.py
@@ -13,7 +13,7 @@ def get_data():
 		},
 		'transactions': [
 			{
-				'items': ['Opportunity', 'Quotation']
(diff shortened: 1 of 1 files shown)


deepeshgarg007 (changes requested):
Email Campaign does n't have any link for Lead, it has a Dynamic Link for Lead and Contact so you'll have to add non standard fieldname as well. Please refer other dashboards for this. Also you'll have to write a method to assign lead in the dynamic link when the (+) button is clicked from the dashboard.

wojosc:
HI @deepeshgarg007 thanks for the info. I will look into that. I would see to find a dashboard without the plus, as the lead should not me the source of creating this.

I'll get back to you!

deepeshgarg007:
@wojosc  I am closing this PR. Please reopen again once you have made the changes

PR #21375, opened 2020-04-22
chore: Commonify auto-filling warehouses in child tables

marination:
- All parent level warehouse fields used to set warehouses in child tables will work via `autofill_warehouse` in `TransactionController`
- Added functionality to bulk set Rejected Warehouse in Purchase Receipt and Purchase Invoice.

Impact on the following:
- `set_warehouse` fields in : **Purchase Invoice, Sales Invoice, Purchase Order, Sales Order, Delivery Note, Material Request** and **Purchase Receipt.**
- `set_reserve_warehouse` field in : **Purchase Order**
- `rejected_warehouse` field in:  **Purchase Invoice** and **Purchase Receipt**
- `source_warehouse` field in : **Work Order**
- `to_warehouse` fields in : **Delivery Note** and **Stock Entry**
- `source_warehouse` field in : **Stock Entry**

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -1889,21 +1889,16 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	},
 
 	set_reserve_warehouse: function() {
-		this.autofill_warehouse("reserve_warehouse");
+		this.autofill_warehouse(this.frm.doc.supplied_items, "reserve_warehouse", this.frm.doc.set_reserve_warehouse);
 	},
 
 	set_warehouse: function() {
-		this.autofill_warehouse("warehouse");
+		this.autofill_warehouse(this.frm.doc.items, "warehouse", this.frm.doc.set_warehouse);
 	},
 
-	autofill_warehouse : function (warehouse_field) {
-		// set warehouse in all child table rows
-		var me = this;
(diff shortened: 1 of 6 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -1889,21 +1889,16 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
-	autofill_warehouse : function (warehouse_field) {
-		// set warehouse in all child table rows
-		var me = this;
-		let warehouse = (warehouse_field === "warehouse") ? me.frm.doc.set_warehouse : me.frm.doc.set_reserve_warehouse;
-		let child_table = (warehouse_field === "warehouse") ? me.frm.doc.items : me.frm.doc.supplied_items;
-		let doctype = (warehouse_field === "warehouse") ? (me.frm.doctype + " Item") : (me.frm.doctype + " Item Supplied");
-
-		if(warehouse) {
+	autofill_warehouse : function (child_table, warehouse_field, warehouse) {
+		let doctype = child_table[0].doctype;
```
nextchamp-saqib: If child_table doesn't have rows it'll break. Just add a ? after [0].
`let doctype = child_table[0]?.doctype`

marination: done

deepeshgarg007 (approved):

PR #21381, opened 2020-04-23
fix: set end_date in Email Campaign

anupamvs:
**Issue-** end_date in Email Campaign is not set when we save the doc due to which status of Email Campaign is not updated over time.
image

**Fix-**
image


erpnext/patches.txt
```diff
@@ -658,3 +658,4 @@ erpnext.patches.v12_0.rename_mws_settings_fields
 erpnext.patches.v12_0.set_correct_status_for_expense_claim
 erpnext.patches.v12_0.set_updated_purpose_in_pick_list
 erpnext.patches.v12_0.repost_stock_ledger_entries_for_target_warehouse
+erpnext.patches.v12_0.update_end_date_and_status_in_email
```
ruchamahabal: Can we change the patch name from "update_end_date_and_status_in_email" to "update_end_date_and_status_in_email_campaign"?

erpnext/patches/v12_0/update_end_date_and_status_in_email.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import unicode_literals
+import frappe
+from frappe.utils import add_days, getdate, today
+
+def execute():
+    email_campaign = frappe.get_all('Email Campaign')
+    if not email_campaign:
+        return
```
ruchamahabal: No need of this. The for loop will take care of this as it will not enter the loop if there are no email campaigns found.

erpnext/patches/v12_0/update_end_date_and_status_in_email.py
```diff
@@ -0,0 +1,26 @@
...
+            end_date = add_days(getdate(doc.start_date), max(send_after_days))
+            doc.db_set("end_date", end_date)
+        today_date = getdate(today())
+        if doc.start_date > today_date:
+            doc.db_set("status", "Scheduled")
+        elif end_date >= today_date:
+            doc.db_set("status", "In Progress")
+        elif end_date < today_date:
+            doc.db_set("status", "Completed")
+        frappe.db.commit()
```
ruchamahabal: Commit is not required. It can lead to committing partially if there are exceptions after it is executed.

ruchamahabal (changes requested):

ruchamahabal:
@anupamvs Add a description for the PR. Also, please raise a PR for the develop branch as well.

erpnext/patches/v12_0/update_end_date_and_status_in_email.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import unicode_literals
+import frappe
+from frappe.utils import add_days, getdate, today
+
+def execute():
+    email_campaign = frappe.get_all('Email Campaign')
```
ruchamahabal: Add a check here to fix the failing Patch Test:

```
if frappe.db.exists('DocType', 'Email Campaign'):
        email_campaign = frappe.get_all('Email Campaign')
```
And then the rest of the code inside the if block

ruchamahabal (approved):

PR #21383, opened 2020-04-23
fix: cast nowdate string to datetime

gsnbng:
String to be converted to dateformat

To resolve the error message :
{'method_name': 'erpnext.assets.doctype.asset.asset.make_post_gl_entry', 
  File "/home/gsn/frappe-bench/apps/frappe/frappe/utils/background_jobs.py", line 99, in execute_job
    method(**kwargs)
  File "/home/gsn/frappe-bench/apps/erpnext/erpnext/assets/doctype/asset/asset.py", line 434, in make_post_gl_entry
    doc.make_gl_entries()
  File "/home/gsn/frappe-bench/apps/erpnext/erpnext/assets/doctype/asset/asset.py", line 361, in make_gl_entries
    and self.purchase_receipt_amount and self.available_for_use_date <= nowdate()):
TypeError: '<=' not supported between instances of 'datetime.date' and 'str'

### DIFF
--- a/erpnext/assets/doctype/asset/asset.py
+++ b/erpnext/assets/doctype/asset/asset.py
@@ -453,7 +453,7 @@ class Asset(AccountsController):
 
 		if ((self.purchase_receipt \
 			or (self.purchase_invoice and frappe.db.get_value('Purchase Invoice', self.purchase_invoice, 'update_stock')))
-			and self.purchase_receipt_amount and self.available_for_use_date <= nowdate()):
+			and self.purchase_receipt_amount and getdate(self.available_for_use_date) <= nowdate()):
 			fixed_asset_account = get_asset_category_account('fixed_asset_account', asset=self.name,
 					asset_category = self.asset_category, company = self.company)
 


erpnext/assets/doctype/asset/asset.py
```diff
@@ -453,7 +453,7 @@ def make_gl_entries(self):
 
 		if ((self.purchase_receipt \
 			or (self.purchase_invoice and frappe.db.get_value('Purchase Invoice', self.purchase_invoice, 'update_stock')))
-			and self.purchase_receipt_amount and self.available_for_use_date <= nowdate()):
+			and self.purchase_receipt_amount and getdate(self.available_for_use_date) <= nowdate()):
```
Thunderbottom, line 456: this won't fix it, the datatype for `available_for_use_date` is already set to `Date`.

`nowdate()` returns a string, which needs to be converted to datetime using `getdate()`. so the correct fix for this issue would be:

```suggestion
			and self.purchase_receipt_amount and self.available_for_use_date <= getdate(nowdate())):
```

Thunderbottom (changes requested):

erpnext/assets/doctype/asset/asset.py
gsnbng: Else where in the same code getdate is used for available_for_use_date, hence thought so. This is a background job, to be verified. Thank you

deepeshgarg007, line 456: also check if `available_for_use_date` is set. `getdate()` returns current date if `None` is passed. Please fix and reopen. Closing this for now

deepeshgarg007 (changes requested):

PR #21386, opened 2020-04-23
fix: quotation have expired status even if sales order exists

nextchamp-saqib:


erpnext/patches/v12_0/fix_quotation_expired_status.py
```diff
@@ -0,0 +1,37 @@
+import frappe
+
+
+def execute():
+    # fixes status of quotations which have status 'Expired' despite having valid sales order created
+
+	# filter out submitted expired quotations which has sales order created
+	cond = "qo.docstatus = 1 and qo.status = 'Expired'"
+	invalid_so_against_quo = """
```
rohitwaghchaure, line 8: invalid_so_against_quo has never used in below query, you have used so_against_quo

erpnext/patches/v12_0/fix_quotation_expired_status.py
```diff
@@ -0,0 +1,37 @@
...
+			and so_item.prevdoc_docname = qo.name
+            and qo.valid_till < so.transaction_date""" # check if SO was created after quotation expired
+
+	frappe.db.sql(
+		"""UPDATE `tabQuotation` qo SET qo.status = 'Expired' WHERE {cond} and not exists({so_against_quo})"""
+			.format(cond=cond, so_against_quo=so_against_quo),
+			(nowdate())
+		)
+    
+    valid_so_against_quo = """
```
rohitwaghchaure: valid_so_against_quo not used in the below query

rohitwaghchaure: Also check indentation, line no 18 & 24 not properly aligned

rohitwaghchaure (approved):

PR #21387, opened 2020-04-23
feat(Payroll Entry): Payroll Entry now lists itself as reference doc when creating journal…

ghost:
Payroll Entry aggregates a number of employee Salary Slips into a single Journal Entry. At present, the debit/credit accounts added do not list anything for their reference document. This means that the Payroll Entry does not track the Journal Entry it creates.

This PR slightly changes the behavior of the Payroll Entry's "Submit Salary Slips" button to cause the Payroll Entry to list itself as the reference document for all lines in its created Journal Entry.


theopen-institute:
Closing this pull request based on further work with broader functionality. New version should be incoming soon.

PR #21396, opened 2020-04-24
fix: rate gets overwritten when pricing rule is set

nextchamp-saqib:
### Problem: 

Promotional scheme configured for below case.

Case : Buy any of these 3 items and get additional coke free.
Coke
Aloo Tiki Burger
Veg Burger

min qty 3 to max qty 5 - > 1 Coke free
10+ qty - > 3 coke free

-------------------------
Please follow the exact steps below to reproduce the bug

Attached is the gif explaining the scenario.
Step 1 : Go to Sales Invoice.
Step 2 : Add Aloo Tiki Burger - 1 qty
Step 3 : Add Veg Burger - 3 qty , this will fetch free item coke. fine till now
Step 4: Add new Row , Select Coke , **the rate of all items will change to 20.0**

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -1407,12 +1407,12 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		for(var k in args) {
 			let data = args[k];
-
+			debugger;
 			if (data && data.apply_rule_on_other_items) {
 				me.frm.doc.items.forEach(d => {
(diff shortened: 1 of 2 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -1407,12 +1407,12 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 		for(var k in args) {
 			let data = args[k];
-
+			debugger;
```
rohitwaghchaure: Remove debugger

PR #21405, opened 2020-04-25
fix: fieldname update for 'Credit' and 'Debit'

AndyOverLord:
> Please provide enough information so that others can review your pull request:

You can skip this if you're fixing a typo

> Explain the **details** for making this change. What existing problem does the pull request solve?

'credit' updated to 'credit_in_account_currency'
'debit' updated to 'debit_in_account_currency'

> Screenshots/GIFs

PR #21416, opened 2020-04-26
fix: fieldname correct

AndyOverLord:

The field name need to be changed so that the update totals could get correct value.

> Screenshots/GIFs

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -347,8 +347,8 @@ cur_frm.cscript.update_totals = function(doc) {
 	var td=0.0; var tc =0.0;
 	var accounts = doc.accounts || [];
 	for(var i in accounts) {
-		td += flt(accounts[i].debit, precision("debit", accounts[i]));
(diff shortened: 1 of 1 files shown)


AndyOverLord:
Hi there,

Please review this pull request with a similar one together, [pull/21405 Link](https://github.com/frappe/erpnext/pull/21405).

erpnext/accounts/doctype/journal_entry/journal_entry.js
```diff
@@ -347,8 +347,8 @@ cur_frm.cscript.update_totals = function(doc) {
 	var td=0.0; var tc =0.0;
 	var accounts = doc.accounts || [];
 	for(var i in accounts) {
-		td += flt(accounts[i].debit, precision("debit", accounts[i]));
-		tc += flt(accounts[i].credit, precision("credit", accounts[i]));
+		td += flt(accounts[i].debit_in_account_currency, precision("debit", accounts[i]));
```
nabinhait, line 350: This is not correct. In the case of a multi-currency journal entry, if debit and credit in different currencies, total debit and credit will not match and lead to difference.

nabinhait (commented):
Closing the PR

PR #21420, opened 2020-04-26
fix: Update in field type

wojosc:
Warranty Period (in Days) changed from "Data" to "Int".


deepeshgarg007 (changes requested):
@wojosc Please update the modified time in json as well so that the changes take effect on migrate

wojosc (commented):
added time stamp

deepeshgarg007 (approved):

deepeshgarg007:
Reverting this, will require patch

PR #21428, opened 2020-04-26
fix: patch to delete appointment reminder scheduler job with wrong path

akurungadam:
Patch added to delete the erroneous scheduler job added earlier

Update error reported on [discuss](https://discuss.erpnext.com/t/error-while-update-1062-duplicate-entry-patient-appointment-send-appointment-reminder-for-key-primary/60830)


ruchamahabal (commented):
@akurungadam 
There is no "Schedule Job Type" DocType in version-12. It was added in version-13.
Can you please Raise a PR for the develop branch to update the existing Job Type with the correct path

ruchamahabal:
@akurungadam 
Fixed in https://github.com/frappe/erpnext/pull/21431
Closing this

PR #21437, opened 2020-04-27
feat(erpnext-integrations): added authorizenet payment gateway

MyuddinKhatri:
added authorizenet payment gateway to erpnext-integrations to make payment
authorizenet

### DIFF
--- a/erpnext/erpnext_integrations/doctype/authorizenet_settings/authorizenet_settings.py
+++ b/erpnext/erpnext_integrations/doctype/authorizenet_settings/authorizenet_settings.py
@@ -54,28 +54,31 @@ Return payment status after processing the payment
 
 """
 from __future__ import unicode_literals
-import frappe
+
 import imp
 import json
 import os
-import sys
 import re
-from six.moves.urllib.parse import urlencode
-from frappe.model.document import Document
-from frappe.integrations.utils import create_payment_gateway, create_request_log
-from frappe.utils import get_url, call_hook_method
-from frappe.utils.password import get_decrypted_password
+import sys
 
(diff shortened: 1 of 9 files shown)


Alchez:
Documentation in process at https://github.com/frappe/erpnext_documentation/pull/40.

MyuddinKhatri:
Some changes are requested in the documentation,
working on that 
[Documentation](https://github.com/frappe/erpnext_documentation/pull/40)

erpnext/erpnext_integrations/doctype/authorizenet_settings/authorizenet_settings.py
```diff
@@ -0,0 +1,200 @@
...
+	def validate(self):
+		create_payment_gateway('Authorizenet')
+		call_hook_method('payment_gateway_enabled', gateway="Authorizenet")
+
+	def get_payment_url(self, **kwargs):
+		return get_url("./integrations/authorizenet_checkout?{0}".format(urlencode(kwargs)))
+
+
+@frappe.whitelist()
+def charge_credit_card(data, card_number, expiration_date, card_code):
```
rohitwaghchaure, line 91: Passing card data in the whitelist method is not good, might get hack. Instead of capturing the card details on frappe page, can it be possible to do payment from the authorize net checkout page?

MyuddinKhatri: @rohitwaghchaure , we are looking into it.

parsimonyERP:
+! authorize.net is pretty big in the States.  It would be helpful to have this as a payment integration

PR #21442, opened 2020-04-27
fix: Blanket Order in SO/PO child tables

marination:
- Blanket order field in child table suggests any blanket order. You can link a Purchasing Blanket Order against a Sales Order.
Screenshot 2020-04-27 at 2 42 53 PM

**After Fix:**
- Commonified Blanket Order field `set_query` in `transaction.js`
- Filter out Blanket Orders by Type, Company, and also check if current row item is in the Blanket Order.
Screenshot 2020-04-27 at 2 46 35 PM

### DIFF
--- a/erpnext/controllers/queries.py
+++ b/erpnext/controllers/queries.py
@@ -371,6 +371,19 @@ def get_account_list(doctype, txt, searchfield, start, page_len, filters):
 		fields = ["name", "parent_account"],
 		limit_start=start, limit_page_length=page_len, as_list=True)
 
+def get_blanket_orders(doctype, txt, searchfield, start, page_len, filters):
+	return frappe.db.sql("""select distinct bo.name, bo.blanket_order_type, bo.to_date
+		from `tabBlanket Order` bo, `tabBlanket Order Item` boi
+		where
+			boi.parent = bo.name
+			and boi.item_code = {item_code}
+			and bo.blanket_order_type = '{blanket_order_type}'
+			and bo.company = {company}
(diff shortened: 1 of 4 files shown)


erpnext/controllers/queries.py
```diff
@@ -371,6 +371,19 @@ def get_account_list(doctype, txt, searchfield, start, page_len, filters):
 		fields = ["name", "parent_account"],
 		limit_start=start, limit_page_length=page_len, as_list=True)
 
+def get_blanket_orders(doctype, txt, searchfield, start, page_len, filters):
+	return frappe.db.sql("""select distinct bo.name, bo.blanket_order_type, bo.to_date
+		from `tabBlanket Order` bo, `tabBlanket Order Item` boi
+		where
+			boi.parent = bo.name
+			and boi.item_code = {item_code}
+			and bo.blanket_order_type = '{blanket_order_type}'
```
nabinhait, line 380:
```suggestion
			and bo.blanket_order_type = {blanket_order_type}
```

nabinhait (changes requested):

erpnext/controllers/queries.py
marination: @nabinhait  sql will process `blanket_order_type` as a column if there are no single quotes. I could either escape `blanket_order_type` (unnecessary) or  `repr(blanket_order_type)` before inserting into query. Or just let the single quotes be.

nabinhait (approved):

nabinhait (approved):

PR #21445, opened 2020-04-27
fix: only check for payment_account on bank entry

michellealva:
While clicking on Make Bank Entry, the system also checks for other mandatory fields which are already set (they become read-only).

image

Since all the fields (company, start and end date are mandatory before form submission, there is no need to check for them again after submission.


michellealva:

nabinhait (approved):

PR #21455, opened 2020-04-27
fix: added validation to not allow to select expired batch

rohitwaghchaure:
**Validation Message**

image

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -3,7 +3,7 @@
 
 from __future__ import unicode_literals
 import frappe, erpnext
-from frappe.utils import cint, flt, cstr
+from frappe.utils import cint, flt, cstr, get_link_to_form, today, getdate
 from frappe import _
 import frappe.defaults
 from erpnext.accounts.utils import get_fiscal_year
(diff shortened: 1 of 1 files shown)


erpnext/controllers/stock_controller.py
```diff
@@ -55,6 +55,12 @@ def validate_serialized_batch(self):
 						frappe.throw(_("Row #{0}: Serial No {1} does not belong to Batch {2}")
 							.format(d.idx, serial_no_data.name, d.batch_no))
 
+			if d.qty > 0 and d.get("batch_no") and self.docstatus < 2:
+				expiry_date = frappe.get_cached_value("Batch", d.get("batch_no"), "expiry_date")
+				if expiry_date and getdate(expiry_date) < getdate(today()):
```
nabinhait:
```suggestion
				if expiry_date and getdate(expiry_date) < getdate(self.posting_date):
```

nabinhait (changes requested):

erpnext/controllers/stock_controller.py
rohitwaghchaure: Fixed

nabinhait (approved):

PR #21467, opened 2020-04-28
fix: list index out of range error

nextchamp-saqib:


gavindsouza (changes requested):
https://github.com/frappe/erpnext/blob/1e4c8282a9894c8cde0fe864f85cbbe686c5282e/erpnext/accounts/report/gross_and_net_profit_report/gross_and_net_profit_report.py#L50 https://github.com/frappe/erpnext/blob/1e4c8282a9894c8cde0fe864f85cbbe686c5282e/erpnext/accounts/report/gross_and_net_profit_report/gross_and_net_profit_report.py#L53

`or []` can be removed since you've added this

nabinhait (approved):

PR #21468, opened 2020-04-28
fix: list index out of range error

nextchamp-saqib:
BackPort of #21467


erpnext/accounts/report/gross_and_net_profit_report/gross_and_net_profit_report.py
```diff
@@ -35,6 +35,11 @@ def execute(filters=None):
 		})
 		return columns, data
 
+
+	# to avoid error eg: gross_income[0] : list index out of range
+	if not len(gross_income): gross_income = [{}]
```
gavindsouza:
```suggestion
	if not gross_income:
	 	 gross_income = [{}]
```

empty lists also evaluate Falsy, so `len` isn't needed

PR #21469, opened 2020-04-28
fix: update lead if contact details are changed (v12)

Alchez:
**Problem:**

In the Lead DocType, we have fields to store its contact's phone information. After they're set, they become hidden, which can lead to outdated Lead reports.

**Solution:**

On saving a Contact linked with a Lead, update the info from Contact into the Lead.

### DIFF
--- a/erpnext/crm/utils.py
+++ b/erpnext/crm/utils.py
@@ -1,0 +1,24 @@
+import frappe
+
+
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:
@Alchez While testing your PR I got following Error

image

PR #21484, opened 2020-04-28
added backlink to MaintenanceVisit

wojosc:
added backlink. This needs to be represented in Maintenance Schedule in some way. Either as simple link in dashboard (which does not exist) or connected to a an entry in the Schedule.


rohitwaghchaure:
@wojosc Maintenance Schedule field is already available in the child table (Purposes) of the Maintenance Visit

Screenshot 2020-04-29 at 1 44 36 AM


Kindly create the Maintenance Visit from Maintenance Schedule, system will auto set the Maintenance Schedule name in the Maintenance Visit child table

Screenshot 2020-04-29 at 1 46 03 AM

wojosc:
Hi @rohitwaghchaure thank you for clarification.

Is there a known reason why this is in the table and not the DocType Level?

wojosc:
will continue in Issue https://github.com/frappe/erpnext/issues/21485. Not in PR.

PR #21488, opened 2020-04-29
fix: item price not fetching when customer is unset in item price

nextchamp-saqib:
Item price was not getting fetched if customer is not selected in Item Price due to `customer is null or customer = ''` condition in the item price fetching query.

This PR doesn't messes up previous code just adds a new check if price list doesn't exists then ignore party and try again


marination:
Left a review on https://github.com/frappe/erpnext/pull/21489

marination:
@nextchamp-saqib port the changes here too

marination (approved):

PR #21489, opened 2020-04-29
fix: item price not fetching when customer is unset in item price

nextchamp-saqib:
BackPort of #21488


erpnext/stock/get_item_details.py
```diff
@@ -669,6 +669,9 @@ def get_price_list_rate_for(args, item_code):
 			item_price_args["uom"] = args.get("stock_uom")
 			general_price_list_rate = get_item_price(item_price_args, item_code, ignore_party=args.get("ignore_party"))
 
+		if not general_price_list_rate:
+			general_price_list_rate = get_item_price(item_price_args, item_code, ignore_party=True)
+
```
marination: Due to line **662**, in the following `get_item_price` , the condition at line **626** is considered since there is no **supplier** or **customer**.  And it fetches the item price anyway. 
So it fails at line **656**  and tries at **665** where it is successful. 
If customer is not set it will be NULL in the db. 
Under what condition will it reach this attempt of fetching the Item Price ?

nextchamp-saqib: Before raising this PR, I was also unsure about the cause. After debugging the same issue thrice, i found that, customer was set to NULL but the same item price which is supposed to be fetch has supplier set as something. 

So the conditions which says that, 
`(customer = '' or customer is null) AND (supplier = '' or supplier is null)` fails the item price to be fetched.

Will change the `get_item_price()` query and remove the current changes.

erpnext/stock/get_item_details.py
```diff
@@ -624,7 +624,11 @@ def get_item_price(args, item_code, ignore_party=False):
 		elif args.get("supplier"):
 			conditions += " and supplier=%(supplier)s"
 		else:
-			conditions += " and (customer is null or customer = '') and (supplier is null or supplier = '')"
+			conditions += """and case when ifnull(selling, 0) = 1 then 
+								(customer is null or customer = '')	
+							 else
+								(supplier is null or supplier = '')
+							 end"""
```
marination:
```suggestion
			conditions += " and (customer is null or customer = '') or (supplier is null or supplier = '')"
```

marination: The current fix fails where Buying and Selling are both enabled.
Assume an Item Price is created for buying and selling both. For a buying transaction if Supplier is not set it should fetch it anyway. But, assume the Customer is set because we want it to fetch for a particular customer in selling, it fails. 
We dont know if the transaction is buying or selling in this area. So, checking for buying or selling Item Price is also incorrect. 'OR' should do the trick, if either aren't set it will fetch it (because it tried checking for customer/supplier once already).

marination (changes requested):

nextchamp-saqib:
line 627 `OR` filter won't work either. 
I have 2 solutions
 - Patch all the selling/buying item prices to remove supplier/customer if only one is enabled. 
 - Add another condition in the previous `if...else..` query, to handle the case for item prices having both selling and buying enabled.

marination:
> line 627 `OR` filter won't work either.
> I have 2 solutions
> 
> * Patch all the selling/buying item prices to remove supplier/customer if only one is enabled.
> * Add another condition in the previous `if...else..` query, to handle the case for item prices having both selling and buying enabled.

@nextchamp-saqib  thought about it too. Patch seems better, we shouldn't have hidden consequential data like that anyway.

marination (approved):

PR #21501, opened 2020-04-29
chore: validate and warn payment against paid invoices

nextchamp-saqib:
Problem:

Create an Invoice of x amount.
Create two Payment Entries and keep them in draft.
Now, you will be able to submit both of them one by one and without any notification the account's balance will get negative
First payment entry makes the invoice paid and second will affect in his accounts as '-x'

This PR warns payment made against invoices which have zero outstanding amount

Screenshot 2020-04-29 at 3 23 11 PM


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -265,6 +266,25 @@ def validate_reference_documents(self):
...
 
+	def validate_paid_invoices(self):
+		no_oustanding_refs = {}
+
+		for d in self.get("references"):
+			if not d.allocated_amount:
+				continue
+
+			if d.reference_doctype in ("Sales Invoice", "Purchase Invoice", "Fees"):
+				outstanding_amount = frappe.get_cached_value(d.reference_doctype, d.reference_name, "outstanding_amount")
```
rohitwaghchaure: On selection of the sales invoice, system fetch the outstanding amount so no need to fetch that value again

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -265,6 +266,25 @@ def validate_reference_documents(self):
...
+				if outstanding_amount <= 0:
```
rohitwaghchaure: If credit note has issued then invoice has the negative outstanding amount 

Screenshot 2020-04-29 at 9 41 32 PM

Need to handle the case of credit / debit note

erpnext/accounts/doctype/payment_entry/payment_entry.py
nextchamp-saqib: This is intentional. I have added a scenario to explain it.

rohitwaghchaure (approved):

PR #21519, opened 2020-04-30
chore: add total row in sales analytics report

nextchamp-saqib:


deepeshgarg007 (changes requested):
This won't work for this report since the sales analytics report contains tree reports as well like for Customer Group and Item Group so the value in total row is not always correct, it gets inflated. 

So either you'll have to calculate the total row in report itself or handle addition of total row for tree view reports in Query Report

deepeshgarg007 (changes requested):
Screenshot 2020-05-02 at 7 30 48 PM

The total row still shows inflated total. It shows 46000 instead of 23000. Also I think there is no need for total row for treeview reports since the total of child nodes is visible in parent node. So remove the total row logic from `get_rows_by_group`

erpnext/selling/report/sales_analytics/sales_analytics.py
```diff
@@ -232,8 +240,12 @@ def get_rows_by_group(self):
 					self.entity_periodic_data.setdefault(d.parent, frappe._dict()).setdefault(period, 0.0)
 					self.entity_periodic_data[d.parent][period] += amount
 				total += amount
+
+				if not total_row.get(scrub(period)): total_row[scrub(period)] = 0
```
deepeshgarg007: Remove this as well, it will cause errors.

deepeshgarg007 (changes requested):

deepeshgarg007:
@nextchamp-saqib Please fix test as well

deepeshgarg007 (approved):

PR #21524, opened 2020-04-30
chore: asset accounts should have company currency

nextchamp-saqib:
Asset accounts should only be in company currency


nabinhait (approved):

crescent28:
@nextchamp-saqib Reason for this fix? Can you elaborate to understand the context of the use case.

nextchamp-saqib:
Problem: Depreciation entries were not created if the asset account is based on different currency then company currency. Specifically creating asset value adjustment against an asset, throws error.

PR #21530, opened 2020-04-30
fix: handle make_gl_entry in case of cwip enable after puchasing

nextchamp-saqib:
BackPort of #21529


nabinhait (changes requested):
@nextchamp-saqib Please add a test for the use case

nabinhait (approved):

PR #21540, opened 2020-05-01
fix: outstanding amount validation fix

AndyOverLord:

When currency_precision is 2, then the outstanding amount as 0.01 should also be considered.

Hence the `abs(row.outstanding) >= 1.0/10 ** self.currency_precision` when self.currency_precision is 2.

There was a bug before that if the invoice has 0.01 amount (based on using 2 as currency precision), it will not show in the accounts receivable summary. This pull request will fix this issue which includes 0.01 invoice.

> Screenshots/GIFs


AndyOverLord:
Hi All, 

Any update on this PR. 

Thanks,

PR #21581, opened 2020-05-04
fix: cast nowdate string to datetime,validate available_for_use_date

gsnbng:
Fixed error due to nowdate in string - converted to datetime along with validation for available_for_use_date exist. Also, while calling in background at make_post_gl_entry, change made as available_for_use_date <=nowdate to ensure if the scheduled job has missed, can happen on subsequent days. This process has been verified.

### DIFF
--- a/erpnext/assets/doctype/asset/asset.py
+++ b/erpnext/assets/doctype/asset/asset.py
@@ -461,7 +461,8 @@ class Asset(AccountsController):
 
 		if ((self.purchase_receipt \
 			or (self.purchase_invoice and frappe.db.get_value('Purchase Invoice', self.purchase_invoice, 'update_stock')))
-			and self.purchase_receipt_amount and self.available_for_use_date <= nowdate()):
+			and self.purchase_receipt_amount and self.available_for_use_date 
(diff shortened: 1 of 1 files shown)


gsnbng:
This changes has been done in my system and found OK. Pls review

marination:
@gsnbng pleas resolve conflicts

erpnext/assets/doctype/asset/asset.py
```diff
@@ -461,7 +461,8 @@ def make_gl_entries(self):
 
 		if ((self.purchase_receipt \
 			or (self.purchase_invoice and frappe.db.get_value('Purchase Invoice', self.purchase_invoice, 'update_stock')))
-			and self.purchase_receipt_amount and self.available_for_use_date <= nowdate()):
+			and self.purchase_receipt_amount and self.available_for_use_date 
+			and self.available_for_use_date <= getdate(nowdate())):
```
marination, line 465: `self.available_for_use_date` is a string. This change will cause issues

marination:
@gsnbng Due to inactivity and effects of change on system, closing this PR.

PR #21586, opened 2020-05-04
fix: bom incorrect price list rate for raw material if price list currency is different from company currency

rohitwaghchaure:
backport https://github.com/frappe/erpnext/pull/21585

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -205,7 +205,7 @@ erpnext.bom.BomController = erpnext.TransactionController.extend({
 	item_code: function(doc, cdt, cdn){
 		var scrap_items = false;
 		var child = locals[cdt][cdn];
-		if(child.doctype == 'BOM Scrap Item') {
+		if (child.doctype == 'BOM Scrap Item') {
 			scrap_items = true;
 		}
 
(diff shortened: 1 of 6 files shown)


erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -215,8 +221,17 @@ erpnext.bom.BomController = erpnext.TransactionController.extend({
 
 		get_bom_material_detail(doc, cdt, cdn, scrap_items);
 	},
+
+	buying_price_list: function(doc) {
+		this.apply_price_list();
+	},
+
+	plc_conversion_rate: function(doc) {
+		this.apply_price_list();
```
nabinhait: if (!this.in_apply_price_list)

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -155,7 +157,7 @@ def get_bom_material_detail(self, args=None):
 			 'rate'			: rate,
 			 'qty'			: args.get("qty") or args.get("stock_qty") or 1,
 			 'stock_qty'	: args.get("qty") or args.get("stock_qty") or 1,
-			 'base_rate'	: rate,
+			 'base_rate'	: flt(rate) * (flt(self.conversion_factor) or 1),
```
nabinhait:
```suggestion
			 'base_rate'	: flt(rate) * (flt(self.conversion_rate) or 1),
```

nabinhait (changes requested):

nabinhait (approved):

PR #21597, opened 2020-05-05
fix: cannot make payment entry against shareholder

nextchamp-saqib:


nabinhait (approved):

nextchamp-saqib:

PR #21602, opened 2020-05-05
fix: adding Email and Phone in Contact child table

anupamvs:
**Issue-**
When converting Lead to the customer, in Primary Contact mobile and email is not saved.
image

**Fix-**
image

### DIFF
--- a/erpnext/selling/doctype/customer/customer.py
+++ b/erpnext/selling/doctype/customer/customer.py
@@ -164,9 +164,11 @@ class Customer(TransactionBase):
 				contact.phone = lead.phone
 				contact.mobile_no = lead.mobile_no
 				contact.is_primary_contact = 1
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/customer/customer.py
```diff
@@ -164,6 +164,8 @@ def create_lead_address_contact(self):
 				contact.phone = lead.phone
 				contact.mobile_no = lead.mobile_no
 				contact.is_primary_contact = 1
+				contact.append('email_ids', dict(email_id=lead.email_id, is_primary=1))
+				contact.append('phone_nos', dict(phone=lead.mobile_no, is_primary_mobile_no=1))
```
rohitwaghchaure: Check first lead.mobile_no available or not. If available then only append to phone_nos, same do with email_ids

anupamvs: done!

PR #21607, opened 2020-05-05
fix: Set Purchase Receipt and Delivery Note detail patch

deepeshgarg007:


erpnext/patches/v12_0/set_purchase_receipt_delivery_note_detail.py
```diff
@@ -3,6 +3,9 @@
 from collections import defaultdict
 
 def execute():
+
+	frappe.reload_doc('stock', 'doctype', 'delivery_note_item', force=True)
```
nabinhait, line 7: Also reload `purchase_receipt_item`

nabinhait (changes requested):

nabinhait (approved):

PR #21617, opened 2020-05-06
feat: Issue Metrics and SLA Enhancements

ruchamahabal:
Metrics Ideas via @harshit-30 

depends on: https://github.com/frappe/frappe/pull/10248
documentation: https://github.com/frappe/erpnext_documentation/pull/105

This PR adds the following metrics in Issues:
- **Average Response Time:** 
   The average time taken to respond to the Customer. This is calculated as the average of all time spans between Received and Sent Communications. Added in https://github.com/frappe/frappe/pull/10596

- **Resolution Time:** 
   Total time taken to close the ticket (from Issue creation to closing)

- **User Resolution Time:** 
   Many times a user has to wait for a customer's reply in order to solve some Issue. While measuring User's productivity this wait time shouldn't be taken into consideration. Hence, user resolution time is the total time taken by a user to close the ticket (Resolution time minus the total amount of time user had to wait for a customer's reply)

issue-metrics

- These metrics are set on "Close" and reset when the Issue reopens or splits except the Average Response Time which is set on every "Sent" communication.

<details>
        
<summary>SLA and SL Merged</summary>

- This PR also removes Service Level DocType and manages SLA only via Service Level Agreement DocType. Service Level was created for templatizing purposes but usually, users don't create so many SLAs so as to have templates for them.

</details>

<details>

<summary> New Control </summary>
- The SLA and Issues make use of a new control called "Duration". Since SLAs are usually set in Hours, the Days and Seconds are not shown for the SLA DocType:

   hide-days-seconds-in-sla

   SLA

</details>

<details>
<summary> Pause SLA on Status Feature </summary>

- Set the statuses on which you want to pause SLA in the SLA DocType:

sla-pause

- **Total Hold Time:** This just keeps adding all hold times together to generate a metric on How much time the issue was actually on Hold
     hold-time

- Also, when the status is changed to any of them set In the "Pause SLA On" child table, resolution and response fields are unset as they are going to be recalculated and the dashboard indicators change to:

sla-hold

</details>

- [x] Patch

- [x] Tests

- [x] Pause SLA Configuration


erpnext/patches/v13_0/update_sla_enhancements.py
```diff
@@ -0,0 +1,111 @@
...
+		sla_details = frappe.db.get_all('Service Level Agreement', fields=['name', 'service_level'])
+		priorities = frappe.db.get_all('Service Level Priority', fields=['*'], filters={
+			'parenttype': ('in', ['Service Level Agreement', 'Service Level'])
+		})
+
+		frappe.reload_doc('support', 'doctype', 'service_level_agreement')
+		frappe.reload_doc('support', 'doctype', 'service_level_priority')
+
+		for entry in sla_details:
+			values = frappe.db.get_value('Service Level', entry.service_level, ['holiday_list', 'employee_group'])
```
hrwX, line 23: Aren't we removing `Service Level` in favor of `Service Level Agreement` ? Then why not delete `Service Level` DocType ?

erpnext/patches/v13_0/update_sla_enhancements.py
```diff
@@ -0,0 +1,111 @@
...
+				employee_group = values[1]
+			frappe.db.set_value('Service Level Agreement', entry.name, {
+				'holiday_list': holiday_list,
+				'employee_group': employee_group
+			})
+
+		priority_dict = {}
+
+		for priority in priorities:
+			if priority.parenttype == 'Service Level Agreement':
```
hrwX, line 35: `Service Level Agreement` and `Service Level` both share the table `Service Level Priority`. So if you create a Service Level Agreement from Service Level, the values are stored in the table. So all you need to do is get the `Service Level Priorities` whose parent is `Service Level Agreement` and then db.set_value

erpnext/patches/v13_0/update_sla_enhancements.py
```diff
@@ -0,0 +1,111 @@
...
+					'resolution_time': resolution_time
+				})
+			if priority.parenttype == 'Service Level':
+				if not priority.parent in priority_dict:
+					priority_dict[priority.parent] = []
+				priority_dict[priority.parent].append(priority)
+
+
+		# copy Service Levels to Service Level Agreements
+		sl = [entry.service_level for entry in sla_details]
```
hrwX, line 49: Service Level Agreement has all the data a Service Level has. Afaik you do not need to copy it over

erpnext/patches/v13_0/update_sla_enhancements.py
```diff
@@ -0,0 +1,111 @@
...
+						'default_priority': priority.default_priority,
+						'response_time': convert_to_seconds(priority.response_time, priority.response_time_period),
+						'resolution_time': convert_to_seconds(priority.resolution_time, priority.resolution_time_period)
+					})
+					row.db_update()
+				sla.db_update()
+
+	# set issue status as Replied since Hold status is removed
+	if frappe.db.exists('DocType', 'Issue'):
+		issues_on_hold = frappe.db.sql("""
```
hrwX: Use ORM

erpnext/patches/v13_0/update_sla_enhancements.py
```diff
@@ -0,0 +1,111 @@
...
+
+		frappe.reload_doc('support', 'doctype', 'issue')
+		frappe.db.sql("""
+			UPDATE
+				`tabIssue`
+			SET
+				status='Replied'
+			WHERE
+				name in %(issues)s
+		""", {'issues': issues}, debug=1)
```
hrwX:
```suggestion
		""", {'issues': issues})
```

erpnext/support/doctype/issue/issue.js
```diff
@@ -41,7 +41,19 @@ frappe.ui.form.on("Issue", {
...
 		if (frm.doc.status !== "Closed" && frm.doc.agreement_fulfilled === "Ongoing") {
 			if (frm.doc.service_level_agreement) {
-				set_time_to_resolve_and_response(frm);
+				if (frm.doc.status == "Replied") {
+					frm.dashboard.clear_headline();
+					let message = {"indicator": "orange", "msg": __("Replied {0}", [moment(frm.doc.on_hold_since).fromNow()])};
+					frm.dashboard.set_headline_alert(
+						'<div class="row">' +
+							'<div class="col-xs-12">' +
+								'<span class="indicator whitespace-nowrap '+ message.indicator +'"><span class="hidden-xs">'+ message.msg +'</span></span> ' +
```
hrwX: remove `hidden-xs` class from `span` tag

erpnext/support/doctype/issue/issue.json
```diff
@@ -114,7 +120,7 @@
    "no_copy": 1,
    "oldfieldname": "status",
    "oldfieldtype": "Select",
-   "options": "Open\nReplied\nHold\nClosed",
+   "options": "Open\nReplied\nResolved\nClosed",
```
hrwX: - How is `Resolved` and `Closed` different from each other ??
- What is the user has to put an issue on Hold due to any reason ?? Would that break workflows or any customisations if any ?

erpnext/support/doctype/issue/issue.json
```diff
@@ -362,12 +370,57 @@
...
    "options": "Issue",
    "read_only": 1
+  },
+  {
+   "fieldname": "column_break_26",
+   "fieldtype": "Column Break"
+  },
+  {
+   "bold": 1,
+   "fieldname": "avg_response_time",
```
hrwX, line 380: Rather than showing an average response time per issue, why not create a report directly ??

erpnext/support/doctype/issue/issue.json
```diff
@@ -362,12 +370,57 @@
...
+   "fieldtype": "Duration",
+   "label": "User Resolution Time",
+   "read_only": 1,
+   "show_days": 1,
+   "show_seconds": 1
+  },
+  {
+   "fieldname": "on_hold_since",
+   "fieldtype": "Datetime",
+   "label": "On Hold Since",
```
hrwX, line 401: Above `Hold` status has been removed from statuses, then why are we calculating on hold since ?

erpnext/support/doctype/issue/issue.py
```diff
@@ -47,27 +47,72 @@ def set_lead_contact(self, email_id):
...
 					frappe.db.get_default("Company")
 
 	def update_status(self):
 		status = frappe.db.get_value("Issue", self.name, "status")
-		if self.status!="Open" and status =="Open" and not self.first_responded_on:
+		if self.status != "Open" and status == "Open" and not self.first_responded_on:
 			self.first_responded_on = frappe.flags.current_time or now_datetime()
 
-		if self.status=="Closed" and status !="Closed":
+		if self.status in ["Closed", "Resolved"] and status not in ["Resolved", "Closed"]:
```
hrwX, line 62: `Closed` and `Resolved` have same meaning ?

erpnext/support/doctype/issue/issue.py
```diff
@@ -311,27 +355,78 @@ def set_service_level_agreement_variance(issue=None):
 			if variance < 0:
 				frappe.db.set_value(dt="Issue", dn=doc.name, field="agreement_fulfilled", val="Failed", update_modified=False)
 
+def set_average_response_time(issue):
```
hrwX: Can't this be moved to reports ??

hrwX (changes requested):

erpnext/support/doctype/issue/issue.js
```diff
@@ -41,7 +41,19 @@ frappe.ui.form.on("Issue", {
 
 		if (frm.doc.status !== "Closed" && frm.doc.agreement_fulfilled === "Ongoing") {
 			if (frm.doc.service_level_agreement) {
-				set_time_to_resolve_and_response(frm);
+				if (frm.doc.status == "Replied") {
+					frm.dashboard.clear_headline();
+					let message = {"indicator": "orange", "msg": __("Replied {0}", [moment(frm.doc.on_hold_since).fromNow()])};
```
hrwX: why not use `frappe.datetime.comment_when` ?

hrwX (changes requested):

erpnext/patches/v13_0/update_sla_enhancements.py
ruchamahabal: There are two things happening in the first part of the patch:

1. Updating the Service Level Priority child table in SLA DocType to change the response and resolution time. For that, I need to first have the old data with time and time_period from Service Level Priority DocType. Then I convert it to seconds as the fieldtype is duration.

2. Also, if there are no SLA documents using particular Service Level documents, I have created SLA from those so that data is not lost after deleting Service Level DocType. Such Service Levels are getting filtered [here](https://github.com/frappe/erpnext/pull/21617/files#diff-fcaf5cae6f782fad8791e921e526e83aR48).

erpnext/patches/v13_0/update_sla_enhancements.py
ruchamahabal: > Service Level Agreement has all the data a Service Level has. Afaik you do not need to copy it over

If there are no SLA documents using particular Service Level documents, I have created SLA from those so that data is not lost after deleting Service Level DocType. Such Service Levels are getting filtered [here](https://github.com/frappe/erpnext/pull/21617/files#diff-fcaf5cae6f782fad8791e921e526e83aR48).

erpnext/support/doctype/issue/issue.json
ruchamahabal: **Resolved:** When users are reasonably sure that they have provided the customer with a solution to their problem, they can change the status of the issue to Resolved.

**Closed:** The Issue can be set as Closed if the customer acknowledges that the problem has been resolved.

These were introduced to have better clarity about the status of the issue.

> * What is the user has to put an issue on Hold due to any reason ?? Would that break workflows or any customisations if any ?

This is a matter of concern. Should we simulate "Hold" time on Replied as well as the Hold status and not remove Hold?

**Edit**: Not Removing Hold status

erpnext/support/doctype/issue/issue.json
ruchamahabal: Yes, we can have a report too, but the idea of having the metric inside the issue is to analyze how was the response time for a particular issue has been so that if there are complaints regarding slower responses from a customer, we can see it clearly in a ticket.

erpnext/support/doctype/issue/issue.json
ruchamahabal: Here "Hold" time is being simulated on the **Replied** status. Later on, we can give the user the option to decide what all status should contribute to total hold time.

erpnext/support/doctype/issue/issue.py
ruchamahabal: Resolved: When users are reasonably sure that they have provided the customer with a solution to their problem, they can change the status of the issue to Resolved.

Closed: The Issue can be set as Closed if the customer acknowledges that the problem has been resolved.

But the functionality remains the same

erpnext/support/doctype/service_level_priority/service_level_priority.json
```diff
@@ -9,14 +10,12 @@
...
-  "response_time_period",
   "cb_00",
-  "resolution_time",
-  "resolution_time_period"
+  "resolution_time"
  ],
  "fields": [
   {
-   "columns": 2,
+   "columns": 1,
```
hrwX:
```suggestion
   "columns": 2,
```
Screenshot 2020-06-11 at 3 06 40 PM

hrwX (changes requested):

hrwX (approved):

PR #21636, opened 2020-05-07
fix: add deduct functionality added for selling cycle

vishdha:
We have use case where we need to deduct tax in selling cycle

WheelHouse_Cannabis_Dispensary_ACC_SINV_2020_00897

we added add and deduct functionality for selling cycle
WheelHouse_Cannabis_Dispensary_ACC_SINV_2020_00897 (1)


nabinhait (commented):
@vishdha Please explain the use case

neilLasrado:
@nabinhait In the state of California for the cannabis industry,  

the cultivator (seller) is supposed to pay the cultivation tax on the sale of produce. However, this tax needs to be collected by the distributor (buyer) while making a purchase and remitted to the govt. Hence the cultivators need to raise Sales Invoices with the total amount - cultivation tax

Also if a distributor is reselling the produce to another distributor, the cultivation tax is 'passed along'. The final distributor who puts the produce in the retail market is responsible for paying the tax. So the distributors also need to raise Sales Invoices with the total amount - cultivation tax while selling to another distributer. 

You can read more about the same here -> https://www.cdtfa.ca.gov/industry/cannabis.htm#Cultivators

neilLasrado:
@nabinhait can you please look into this

vishdha:
@nabinhait  Please review this

erpnext/accounts/doctype/sales_taxes_and_charges/sales_taxes_and_charges.json
```diff
@@ -192,15 +184,31 @@
   {
    "fieldname": "dimension_col_break",
    "fieldtype": "Column Break"
+  },
+  {
+   "fieldname": "category",
+   "fieldtype": "Select",
+   "label": "Consider Tax or Charge for",
+   "options": "Valuation and Total\nValuation\nTotal",
```
nabinhait, line 192: What is the usecase for Valuation in sales cycle?

nabinhait (changes requested):
Seems like you have not fixed tax amount many places like 'item-wise sales register' and 'sales register' report. And gl entries for taxes on sales invoice.

Basically, search globally for all the instance of 'add_deduct_tax' which is implemented in purchase cycle and give it a thought whether it needs to replicated for sales or not.

nabinhait:
Add tests

sagarvora:
@vishdha Any updates?

PR #21637, opened 2020-05-07
fix: add deduct functionality added for selling cycle

vishdha:
We have use case where we need to deduct tax in selling cycle

WheelHouse_Cannabis_Dispensary_ACC_SINV_2020_00897

we added add and deduct functionality for selling cycle
WheelHouse_Cannabis_Dispensary_ACC_SINV_2020_00897 (1)


rohitwaghchaure:
Kindly don't send new feature on stable version (version-12-hotfix)

PR #21642, opened 2020-05-07
fix(accounting): fix incorrect output in Budget Variance Report

kevingdc:
This pull request fixes a bug in **Budget Variance Report** where actual expense amounts are added across different fiscal years. This was fixed by
updating the function and queries for computing the actual expense
amounts.

Expense entry:
journal_entry

Before:
incorrect

After:
correct

### DIFF
--- a/erpnext/accounts/report/budget_variance_report/budget_variance_report.py
+++ b/erpnext/accounts/report/budget_variance_report/budget_variance_report.py
@@ -2,187 +2,337 @@
 # License: GNU General Public License v3. See license.txt
 
 from __future__ import unicode_literals
+from six import iteritems
+
 import frappe
 from frappe import _
 from frappe.utils import flt
 from frappe.utils import formatdate
+
 from erpnext.controllers.trends import get_period_date_ranges, get_period_month_ranges
 
-from six import iteritems
-from pprint import pprint
+
 def execute(filters=None):
-	if not filters: filters = {}
+    if not filters:
+        filters = {}
 
-	columns = get_columns(filters)
-	if filters.get("budget_against_filter"):
-		dimensions = filters.get("budget_against_filter")
-	else:
-		dimensions = get_cost_centers(filters)
+    columns = get_columns(filters)
 
-	period_month_ranges = get_period_month_ranges(filters["period"], filters["from_fiscal_year"])
-	cam_map = get_dimension_account_month_map(filters)
+    if filters.get("budget_against_filter"):
+        dimensions = filters.get("budget_against_filter")
+    else:
+        dimensions = get_cost_centers(filters)
 
-	data = []
-	for dimension in dimensions:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/budget_variance_report/budget_variance_report.py
```diff
@@ -2,187 +2,419 @@
...
+                    account = account[0]
+                    cam_map.setdefault(ccd.name, {}).setdefault(account, {}).setdefault(
+                        year, {}
+                    ).setdefault(month, frappe._dict({"target": 0.0, "actual": 0.0}))
 
-			tav_dict.target = flt(ccd.budget_amount) * month_percentage / 100
+                    tav_dict = cam_map[ccd.name][account][year][month]
 
-			for ad in actual_details.get(ccd.account, []):
-				if ad.month_name == month:
```
deepeshgarg007, line 177: @kevingdc No need to make all these changes, a small change here would have fixed the issue. No need make extra queries to get budget_amount or monthly_distribution amount seperately.

`if ad.month == month and ad.fiscal_year == ccd.fiscal_year` this condition would have been enough. Also there are lot of inconsistencies in indentation spacing, somewhere spaces and somewhere tabs, we usually prefer tabs and that is what we have used in the entire project

deepeshgarg007 (changes requested):

erpnext/accounts/report/budget_variance_report/budget_variance_report.py
kevingdc: Hi @deepeshgarg007,

Thank you for the response. I appreciate the feedback, and you're right, adding the the condition `if ad.month == month and ad.fiscal_year == ccd.fiscal_year` would have fixed it. Although the reason why I added those functions was so I could also show the actual expense even though it doesn't have a budget allocated, an example is shown below:

Screen Shot 2020-05-08 at 4 55 26 PM

The old code, even though it would get fixed by adding the condition you mentioned wouldn't have shown it. The output would have been like this:

Screen Shot 2020-05-08 at 5 00 58 PM


I thought it would be more helpful to the users to see their actual expenses even though there wasn't a budget allocated for it. But if you feel that this is unnecessary, please tell me and I'll update my code to reflect your request. 

Thank you!

deepeshgarg007: Budget Variance Report is to check variation of Actual Expense from Budget. This doesn't seems the right place to show expense if budget doesn't exits, also it adds a lot of unnecessary query to get this done and fix the spaces for indentation as welll

kevingdc: @deepeshgarg007 got it. Updated the code to reflect this.

deepeshgarg007 (approved):

PR #21658, opened 2020-05-08
feat: apply item-wise pricing rules

Alchez:
**Changes:**

- Add "Ignore Pricing Rule" checkbox in all Buying and Selling transactions that apply pricing rules.
- Re-map pricing rule triggers to pick up item toggles instead of the parent toggle.

<hr>

**Screenshots / GIFs:**

item-pricing-rule

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -206,7 +206,7 @@ def get_pricing_rule_for_item(args, price_list_rate=0, doc=None, for_validate=Fa
 		"discount_amount_on_rate": []
 	})
 
-	if args.ignore_pricing_rule or not args.item_code:
+	if args.ignore_pricing_rule or args.item_ignore_pricing_rule or not args.item_code:
 		if frappe.db.exists(args.doctype, args.name) and args.get("pricing_rules"):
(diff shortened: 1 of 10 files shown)


rohitwaghchaure:
@Alchez I think ignore pricing rule at child item is not required, you can mange this using pricing rule too. Also this checkbox not display without expanding row, any new user this is difficult to understand.

About your use case if you don't want to apply particular pricing rule for particular customer then make pricing rule against customer / customer group.

Alchez:
> @Alchez I think ignore pricing rule at child item is not required, you can mange this using pricing rule too.

The main use-case here is that ignoring pricing rules can be important for applying manual discounts, and currently doing that on a per-item level is impossible without applying it to the whole order.

> Also this checkbox not display without expanding row, any new user this is difficult to understand.

I can add it to the grid view, but it's already getting crowded in there. I've placed it next to the rate though, so users can see it while changing the rate. Apart from that, the field will suffer from being out of view like any other field in a child table row.

> About your use case if you don't want to apply particular pricing rule for particular customer then make pricing rule against customer / customer group.

We want to ignore the pricing rule on one of the items for this particular instance only, which isn't easy to do without also ignoring the pricing rule on the rest of the items.

neilLasrado:
@rohitwaghchaure can you please check Rohan's comments?

PR #21682, opened 2020-05-11
fix: Adding status Overdue in Sales Order

anupamvs:
**Issue-**  In the Sales Order list, we can not apply **overdue** status filter. Unless click on "Overdue" status from the list then it will apply some filters and will show the result.
image

**After Fix-**
image


nabinhait:
closing this for now.

PR #21683, opened 2020-05-11
fix: Adding status Overdue in Sales Order

anupamvs:
**Issue-** In the Sales Order list, we can not apply overdue status filter. Unless click on "Overdue" status from the list then it will apply some filters and will show the result.
image

**Fix-**
image


erpnext/controllers/accounts_controller.py
```diff
@@ -1060,6 +1060,9 @@ def update_invoice_status():
 	frappe.db.sql(""" update `tabPurchase Invoice` set status = 'Overdue'
 		where due_date < CURDATE() and docstatus = 1 and outstanding_amount > 0""")
 
+	frappe.db.sql(""" update `tabSales Order` set status = 'Overdue'
```
nabinhait, line 1063: Don't set overdue status from background job, it will overwrite "To Deliver" and "To Bill" status, which is more important.

erpnext/selling/doctype/sales_order/sales_order.json
```diff
@@ -1016,7 +1016,7 @@
    "no_copy": 1,
    "oldfieldname": "status",
    "oldfieldtype": "Select",
-   "options": "\nDraft\nOn Hold\nTo Deliver and Bill\nTo Bill\nTo Deliver\nCompleted\nCancelled\nClosed",
+   "options": "\nDraft\nOn Hold\nTo Deliver and Bill\nTo Bill\nTo Deliver\nCompleted\nCancelled\nClosed\nOverdue",
```
nabinhait, line 1019: Do the same changes in purchase order also

erpnext/selling/doctype/sales_order/sales_order_list.js
```diff
@@ -9,13 +9,13 @@ frappe.listview_settings['Sales Order'] = {
 			// on hold
 			return [__("On Hold"), "orange", "status,=,On Hold"];
 		} else if (doc.status === "Completed") {
+			// Completed
 			return [__("Completed"), "green", "status,=,Completed"];
+		} else if (doc.status === "Overdue") {
```
nabinhait, line 14: no changes required as we are not setting from background jobs

nabinhait (changes requested):

rmehta:
I don't think this design is right. Overdue should be a separate status column than this that should be set by background job.

nabinhait:
> I don't think this design is right. Overdue should be a separate status column than this that should be set by background job.

Good suggestion 👍 
There are already 3 status fields in Sales Order, maybe we can add the "Overdue" status in the existing "Delivery Status" field. It has existing options like "Not Delivered", "Partially Delivered" and "Fully Delivered". And we can set overdue status from background job.

nabinhait:
closing this for now.

PR #21691, opened 2020-05-11
fix: Add misssing dimensions in GL entries

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/21689

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -821,7 +821,7 @@ class SalesInvoice(SellingController):
 
 					for gle in fixed_asset_gl_entries:
 						gle["against"] = self.customer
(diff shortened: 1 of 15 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -821,7 +821,7 @@ def make_item_gl_entries(self, gl_entries):
 
 					for gle in fixed_asset_gl_entries:
 						gle["against"] = self.customer
-						gl_entries.append(self.get_gl_dict(gle))
+						gl_entries.append(self.get_gl_dict(gle, item=asset))
```
nabinhait: item=item?

nabinhait (changes requested):

nabinhait (approved):

PR #21699, opened 2020-05-12
fix: purchase inv shows overdue for fraction of outstanding

nextchamp-saqib:
Fixes the status problem similar to sales invoice where if outstanding amount is 0.002 then it shows Overdue status while in the Form outstanding is zero


deepeshgarg007 (changes requested):
There are many test cases already on purchase invoice, where we don't check any status, Add `self.assertEqual` check for status in those test so this will get verified as well

deepeshgarg007 (approved):

deepeshgarg007 (approved):

PR #21710, opened 2020-05-13
fix: pound, gram to Ounce converion factor added

vishdha:
Added conversion gram , pound to ounce
and fix install_fixtures.py add_uom_data()
before: on call it create duplicate data
after: only missing data get updated


marination:
@vishdha please resolve conflicts

vishdha:
@marination Please look into this

marination (approved):

PR #21712, opened 2020-05-13
fix: error log title for failing bank transactions

Mangesh-Khairnar:
- Add a better error log title for failing bank transactions

### DIFF
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
@@ -209,7 +209,7 @@ def new_bank_transaction(transaction):
 			result.append(new_transaction.name)
 
 		except Exception:
-			frappe.throw(frappe.get_traceback())
+			frappe.throw(title=_('Bank transaction creation error'))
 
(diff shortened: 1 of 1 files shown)


erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
```diff
@@ -209,7 +209,7 @@ def new_bank_transaction(transaction):
 			result.append(new_transaction.name)
 
 		except Exception:
-			frappe.throw(frappe.get_traceback())
+			frappe.throw(title=_('Bank transaction creation error'))
```
deepeshgarg007, line 212: This is just title, what about error message?

Mangesh-Khairnar: If the message is not specified, traceback becomes the error message. https://github.com/frappe/frappe/blob/90b4af0662d5b125108514b7d78195189abcd083/frappe/__init__.py#L1567

deepeshgarg007 (approved):

Mangesh-Khairnar:

PR #21718, opened 2020-05-13
fix: Future date half day validation

Anurag810:
Allow half-Day for a future date only from leave application

### DIFF
--- a/erpnext/hr/doctype/attendance/attendance.py
+++ b/erpnext/hr/doctype/attendance/attendance.py
@@ -21,7 +21,7 @@ class Attendance(Document):
 		date_of_joining = frappe.db.get_value("Employee", self.employee, "date_of_joining")
 
 		# leaves can be marked for future dates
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/attendance/attendance.py
```diff
@@ -21,7 +21,7 @@ def validate_attendance_date(self):
 		date_of_joining = frappe.db.get_value("Employee", self.employee, "date_of_joining")
 
 		# leaves can be marked for future dates
-		if self.status not in ('On Leave', 'Half Day') and getdate(self.attendance_date) > getdate(nowdate()):
+		if ((self.status != 'On Leave') or (self.status != 'Half Day' and not self.leave_application)) and getdate(self.attendance_date) > getdate(nowdate()):
```
nabinhait:
```suggestion
		if self.status != 'On Leave' and not self.leave_application and getdate(self.attendance_date) > getdate(nowdate()):
```

nabinhait (changes requested):

nabinhait (approved):

PR #21721, opened 2020-05-13
fix: Patch to set status in old serial no data

marination:
Backport of https://github.com/frappe/erpnext/pull/21720


erpnext/patches/v12_0/set_serial_no_status.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+	frappe.reload_doc('stock', 'doctype', 'serial_no')
+
+	for serial_no in frappe.db.sql("""select name from `tabSerial No`""", as_dict = 1):
+		doc = frappe.get_doc("Serial No", serial_no.get("name"))
+		if not doc.status:
+			doc.set_status()
```
nabinhait: The number of serial nos can be huge and `get_doc` can take too much time.

Implement (copy) the logic of set_status here to avoid `get_doc` and calling `set_status`

nabinhait (changes requested):

erpnext/patches/v12_0/set_serial_no_status.py
marination: Done

erpnext/patches/v12_0/set_serial_no_status.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import unicode_literals
+import frappe
+from frappe.utils import getdate, nowdate
+
+def execute():
+	frappe.reload_doc('stock', 'doctype', 'serial_no')
+
+	for serial_no in frappe.db.sql("""select name from `tabSerial No`""", as_dict = 1):
+		sn_detail = frappe.db.get_value("Serial No", serial_no.get("name"), ["status", "delivery_document_type", "warranty_expiry_date"], as_dict=1)
```
nabinhait: No need for separate get_value, it can part of the select query

nabinhait: Also, select only those which does not have status

nabinhait (changes requested):

nabinhait (approved):

PR #21728, opened 2020-05-14
Hotfix products two column layout

nehasacher143:
**Task:** Change the layout of the Products listing page

**Description:** The layout is not a 2-column layout

**Screenshots:
Before:**
productsbefore

**After:**
productsafter

### DIFF
--- a/erpnext/public/css/products.css
+++ b/erpnext/public/css/products.css
@@ -1,0 +1,14 @@
+.products-list .card:nth-child(2n+1) {
+  margin-right: 10px;
+}
+
+.products-list .card {
+  max-width: 48%;
+}
+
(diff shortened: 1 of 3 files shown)


erpnext/public/css/products.css
```diff
@@ -0,0 +1,14 @@
...
+.products-list .card {
+  max-width: 49%;
+}
+
+@media screen and (max-width:767px) {
+  .products-list.row.no-gutters .card { 
+    max-width: 100%;
+    margin: 0;
+  }
+}
```
scmmishra: The convention is to use tabs and not spaces.

scmmishra (commented):
Any particular reason why this change in layout is suggested?

PR #21741, opened 2020-05-15
fix: Other taxes and charges check

govindsmenokee:
Currently other taxes and charges check is being done based on the tax_amount value which is incorrect. The correct check should be based on annual_taxable_earning value. The use case is when Surcharge needs to be levied on employees with annual_taxable_earning more than 50,000,000.

> Screenshots/GIFs


nabinhait (dismissed):

nabinhait (commented):
We don't accept direct PR to version-12 branch. Closing this, I have fixed the issue via https://github.com/frappe/erpnext/pull/22798

PR #21756, opened 2020-05-17
fix: Description for fields and patch for old loans

deepeshgarg007:
image

image


erpnext/patches/v13_0/update_old_loans.py
```diff
@@ -0,0 +1,83 @@
...
+	updated_loan_types = []
+
+	loans = frappe.get_all('Loan', fields=['name', 'loan_type', 'company', 'status', 'mode_of_payment',
+		'applicant_type', 'applicant', 'loan_account', 'payment_account', 'interest_income_account'])
+
+	for loan in loans:
+		# Update details in Loan Types and Loan
+		loan_type_company = frappe.db.get_value('Loan Type', loan.loan_type, 'company')
+
+		parent_income = frappe.get_value('Account', {'company': loan.company,
```
nabinhait: First try to get "Indirect Income" account as parent based on hard-coded account name `_("Indirect Account")`, if it is not found then return the root income account

nabinhait (changes requested):

PR #21779, opened 2020-05-19
fix: only auto-set serial nos and batches if allowed in Stock Settings (v12)

Alchez:
**Before:**

pick-list-auto-set-serial-batch-before

**After:**

pick-list-auto-set-serial-batch-after


marination:
Left a Review on https://github.com/frappe/erpnext/pull/21781

marination (approved):

PR #21780, opened 2020-05-19
fix: filter batches based on item and warehouse in Pick List (develop)

Alchez:
image


marination (approved):

crescent28:
@Alchez please provide steps to test the following fix.

Alchez:
@crescent28, if you create a Pick List from a Sales Order that has an item with an available batch number, you'll be able to see this field and be able to modify the batch.

I think @marination is working on a fix to have persistent display for both serial numbers and batches.

PR #21781, opened 2020-05-19
fix: only auto-set serial nos and batches if allowed in Stock Settings (develop)

Alchez:
**Before:**

pick-list-auto-set-serial-batch-before

**After:**

pick-list-auto-set-serial-batch-after


erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -116,12 +116,15 @@ def get_items_with_location_and_quantity(item_doc, item_location_map):
 		if item_location.serial_no:
 			serial_nos = '\n'.join(item_location.serial_no[0: cint(stock_qty)])
 
+		auto_set_serial_no = frappe.db.get_single_value("Stock Settings", "automatically_set_serial_nos_based_on_fifo")
+		auto_set_batch_no = frappe.db.get_single_value("Stock Settings", "automatically_set_batch_nos_based_on_fifo")
```
marination: `automatically_set_batch_nos_based_on_fifo` does not exist in Stock Settings

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -116,12 +116,15 @@ def get_items_with_location_and_quantity(item_doc, item_location_map):
...
+
 		locations.append(frappe._dict({
 			'qty': qty,
 			'stock_qty': stock_qty,
 			'warehouse': item_location.warehouse,
-			'serial_no': serial_nos,
-			'batch_no': item_location.batch_no
+			'serial_no': serial_nos if auto_set_serial_no else item_doc.serial_no,
+			'batch_no': item_location.batch_no if auto_set_batch_no else item_doc.batch_no
```
marination: The point of picking is that the  user **automatically** gets the oldest stock(Serial/Batch) recommended without having to explicitly set anything but item code and qty.
This change will keep the serial no blank when `set_item_locations` is first called, if the checkbox is not enabled in stock settings. The user can't fill the serial no either because the field is not visible until there's a value in it.
The user anyway shouldn't have to **explicitly** mention serial nos, because then it beats the whole purpose of Pick List.

erpnext/stock/doctype/pick_list/pick_list.py
Alchez: Ah, you're right, I'll remove this line

erpnext/stock/doctype/pick_list/pick_list.py
Alchez: I agree that having the system auto-pick stock based on FIFO is useful, and I'm not against that flow.

This PR is only intended to give some amount of freedom to the inventory manager as to what range of serial numbers or batches they can select before sending the list to the warehouse for picking.

Regarding the fields not showing up if they don't have values, should we consider always showing the fields too? Users can already change the serial and batch data right now.

What do you think?

marination: Yeah on second thought, we should let user select/change serial and batch data, there shouldn't be restrictions. Lets allow serial/batch no visibility. You will have to undo `'batch_no': item_location.batch_no if auto_set_batch_no else item_doc.batch_no` this line anyway.

Alchez: Alright, I've removed the batch variable, but I think removing the restriction from the serial and batch fields can maybe happen in separate PRs to all the branches, so I won't include it in this one.

marination (approved):

PR #21785, opened 2020-05-19
fix: tax id is not fetched when creating sales order from quotation

anupamvs:
Issue -
A0xVW0w

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.json
+++ b/erpnext/selling/doctype/sales_order/sales_order.json
@@ -282,6 +282,7 @@
    "width": "100px"
   },
   {
+   "fetch_from": "customer.tax_id",
    "fieldname": "tax_id",
    "fieldtype": "Data",
    "label": "Tax Id",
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/sales_order/sales_order.json
```diff
@@ -282,6 +282,8 @@
    "width": "100px"
   },
   {
+   "fetch_from": "customer.tax_id",
+   "fetch_if_empty": 1,
```
deepeshgarg007: fetch_if_empty should not be 1. Since this is a read_only field it should always be fetched from Customer master

deepeshgarg007 (changes requested):

erpnext/selling/doctype/sales_order/sales_order.json
anupamvs: @deepeshgarg007 Done!

deepeshgarg007 (approved):

PR #21797, opened 2020-05-20
fix: make tax and total debited on deduct in sales invoice on gl_ledger Creation

vishdha:
PR  Depends on https://github.com/frappe/erpnext/pull/21636

Problem : on Add_deduct_tax == "Deduct". It try to create gl_ledger. in round off function debit-credit should be zero. but when we deduct it give debit-credit give not zero result.

solve:
Added on Add we credit the amount and on deduct we debit the amount. that logic wrote in sales invoice.

before:
Peek 2020-05-19 18-49

after:
Peek 2020-05-19 18-47


vishdha:
upstream https://github.com/frappe/erpnext/pull/21797

PR #21801, opened 2020-05-20
feat: remove website settings from boot

scmmishra:
Adding Javascript tags to `head_html` would break desk boot completely as website settings was loaded in bootinfo during session creation. This is because the closing tag `</script>` in the `head_html` would get parsed in the DOM and it would break rest of the script. 

The way to solve this would be to actually remove the `head_html` from being included in the bootinfo, however the `website_settings` is not used anywhere across desk.

This PR removes website_settings from boot in ERPNext, since the data is not used anywhere across ERPNext or Frappe.


scmmishra:

hrwX (approved):

PR #21827, opened 2020-05-21
fix: website_list_for_contact, fix changed frappe.db.exist keyword argument

gbm001:
frappe.db.exists takes two arguments 'dt' and 'dn'. It no longer takes an argument 'filters'. This broken call in website_list_for_contact.py causes an exception which (somehow) causes authentication success instead of failure when viewing documents through the portal.

There are additional issues with this function, but those need additional thought and consideration and this part should be fixed ASAP.


nabinhait (approved):

nabinhait:
@Mergifyio backport develop

Thunderbottom:
@mergifyio frontport develop

nabinhait:
@Mergifyio frontport develop

gbm001:
Since the 'front-port' to develop didn't work, I have made a new pull request for develop here:
https://github.com/frappe/erpnext/pull/21958
rather than have it disappear into obscurity.

PR #21838, opened 2020-05-21
fix: Post Dated unallocated amount not considered in Advance Amount in AR/AP summary

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/21837

### DIFF
--- a/erpnext/accounts/party.py
+++ b/erpnext/accounts/party.py
@@ -609,7 +609,7 @@ def get_partywise_advanced_payment_amount(party_type, posting_date = None, futur
 	cond = "1=1"
 	if posting_date:
 		if future_payment:
-			cond = "DATE(creation) <= '{0}'".format(posting_date)
+			cond = "posting_date <= '{0}' OR DATE(creation) <= '{0}' """.format(posting_date)
 		else:
 			cond = "posting_date <= '{0}'".format(posting_date)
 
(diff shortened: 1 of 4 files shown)


erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -578,6 +578,26 @@ def get_gl_entries(self):
 				{1} {2}"""
 			.format(select_fields, conditions, order_by), values, as_dict=True)
 
+		if self.filters.show_future_payments:
+			values.insert(2, self.filters.report_date)
+
+			future_entries = frappe.db.sql("""
```
nabinhait: avoid a separate query to get future entries, it takes too much time in old databases.

erpnext/accounts/party.py
```diff
@@ -605,10 +605,14 @@ def get_party_shipping_address(doctype, name):
 	else:
 		return ''
 
-def get_partywise_advanced_payment_amount(party_type, posting_date = None, company=None):
+def get_partywise_advanced_payment_amount(party_type, posting_date = None, future_payment=0, company=None):
 	cond = "1=1"
 	if posting_date:
-		cond = "posting_date <= '{0}'".format(posting_date)
+		if future_payment:
+			cond = "DATE(creation) <= '{0}'".format(posting_date)
```
nabinhait:
```suggestion
			cond = "posting_date <= '{0}' or DATE(creation) <= '{0}'".format(posting_date, posting_date)
```

nabinhait (changes requested):

nabinhait (approved):

PR #21844, opened 2020-05-21
fix: convert goals point to flt

Anurag810:

> Screenshots/GIFs

### DIFF
--- a/erpnext/hr/doctype/appraisal_template/appraisal_template.py
+++ b/erpnext/hr/doctype/appraisal_template/appraisal_template.py
@@ -3,7 +3,7 @@
 
 from __future__ import unicode_literals
 import frappe
-from frappe.utils import cint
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/appraisal_template/appraisal_template.py
```diff
@@ -3,19 +3,19 @@
...
 	def validate(self):
 		self.check_total_points()
-		
-	def check_total_points(self):	
+
+	def check_total_points(self):
 		total_points = 0
 		for d in self.get("goals"):
-			total_points += int(d.per_weightage or 0)
+			total_points += flt(d.per_weightage or 0)
```
saurabh6790: if `per_weightage` is empty or null `flt` will anyhow return `0`. why or 0?

nabinhait (approved):

PR #21851, opened 2020-05-21
fix: expense account error message in DN

michellealva:
Changed error message if expense account not set for item in Delivery Note.
**Earlier:** Expense or Difference account is mandatory for Item IT - 6 as it impacts overall stock value
**After fix:** Expense Account not set for Item IT - 6. Please set an Expense Account for the item in the Items table

Note: Item - 6 is an example


michellealva:

nabinhait (approved):

nabinhait:

PR #21887, opened 2020-05-23
fix: Validation for dates

Anurag810:
#21886


erpnext/hr/doctype/upload_attendance/upload_attendance.py
```diff
@@ -20,6 +20,9 @@ def get_template():
 
 	args = frappe.local.form_dict
 
+	if getdate(args.from_date) > getdate(args.to_date):
+		frappe.throw("To Date should be greater than From Date")
```
mhbu50: Make the text translateable

mhbu50 (changes requested):

PR #21898, opened 2020-05-25
feat: Relational UOM Query

P-Froggy:
#### Description
Make the list of UOMs that can be selected during a transaction dependent on the respective item. In order to be selectable in the link field, a UOM either has to be the stock UOM of the item, be in the list of UOM conversion details or has a set conversion factor relative to the stock UOM of the item. Furthermore, the respective conversion factor is shown as preview in the link field.

#### Details
By default, all UOMs in the link field will be retrieved by transactions for selection, but a majority of the UOMs will never be used for the selected item. For example, it will never make sense to buy apples in degrees Celsius. 

Furthermore, an untrained user cannot know which UOMs have been created in the Item Master and therefore already have a predefined conversion factor.

To make the selection of UOMs more user-friendly, this pull request changes the following aspects:
- The UOM link field now only queries UOMs that are one of the following:
  - Stock UOM of the selected item
  - In the list of UOM conversion details in the item master
  - An UOM conversion factor of the stock UOM of the selected item. For example, if the stock uom is kg, then also the corresponding weight units gram, milligram, pounds, etc. will be available for selection
- UOMs are sorted in the same order as specified above, i.e., Standard Stock Unit of Measure is always at the top (unless a search text is entered)
- The conversion factors are rounded to the respective field precision. Rounded conversion factors of 0 are filtered out of the query.
- Below the respective UOM, the relative conversion factor to the stock UOM is indicated, except for the stock UOM itself (because it would always be 1) (see screenshot)

Anmerkung 2020-05-23 202514

Furthermore, the following minor aspects get changed:
- UOMs with conversion factors of 1 are no longer filtered out of the UOM conversion detail table in the item master because you need a way of specifying the allowed UOMs for the item, even if they have a conversion factor of 1
- Purchase and sales UOM also get automatically added to the UOM conversion detail table in the item master (like stock UOM)
- I removed the query-setter function for item code in the purchase invoice doctype because it was set twice (also in the buying controller, which purchase invoice extends)


P-Froggy:
Keep PR open

erpnext/controllers/queries.py
```diff
@@ -224,6 +224,64 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
 			}, as_dict=as_dict)
 
 
+def uom_query(doctype, txt, searchfield, start, page_len, filters, as_dict=False):
+	conditions = []
+	precision = get_field_precision(frappe.get_meta("UOM Conversion Detail").get_field("conversion_factor"))
```
sagarvora, line 264: Please import this function, check Codacy

sagarvora (changes requested):

erpnext/controllers/queries.py
P-Froggy: Added the import 👍

rohitwaghchaure:
@P-Froggy In some use cases the conversion factor is not fixed and it varies depending on the size. For example if I have metal rod which has two UOMS, Kg and Nos and KG is the stock uom. Also the metal rod has size and based on the size the kg conversion rate is different. So here I will not put the UOM Nos in the item's UOM Conversion Detail table or in the UOM Conversion Factor table because conversion between Nos to Kg is not fixed. So as per your changes I will not able to see the UOM Nos in the dropdown. I think is some cases conversion factor is fixed and is some cases it's depending on some other parameters.

PR #21903, opened 2020-05-25
fix: Item Price and Add to Cart not showing on Website

marination:
- Item info is fetched with `skip_quotation_creation=True`
- Price list (required to fetch item price) was attempted to be fetched from cart_quotation, which is blank if `skip_quotation_creation=True`
**Solution:** Fetch price list again via `_set_price_list` and make `_set_price_list` handle fetching price list without quotation data

**Before:**
Screenshot 2020-05-25 at 2 20 52 PM

**After:**
Screenshot 2020-05-25 at 2 20 28 PM

### DIFF
--- a/erpnext/shopping_cart/cart.py
+++ b/erpnext/shopping_cart/cart.py
@@ -319,7 +319,7 @@ def apply_cart_settings(party=None, quotation=None):
 def set_price_list_and_rate(quotation, cart_settings):
 	"""set price list based on billing territory"""
 
-	_set_price_list(quotation, cart_settings)
+	_set_price_list(cart_settings, quotation)
 
 	# reset values
 	quotation.price_list_currency = quotation.currency = \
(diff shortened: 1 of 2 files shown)


nabinhait (dismissed):

erpnext/shopping_cart/product_info.py
```diff
@@ -21,9 +21,11 @@ def get_product_info_for_website(item_code, skip_quotation_creation=False):
 	if not skip_quotation_creation:
 		cart_quotation = _get_cart_quotation()
 
+	selling_price_list = cart_quotation.get("selling_price_list") if cart_quotation else _set_price_list(cart_settings, None)
+
```
scmmishra, line 25:
```suggestion
	if not skip_quotation_creation:
		cart_quotation = _get_cart_quotation()
		selling_price_list = cart_quotation.get("selling_price_list", None)
	
	if not selling_price_list:
		selling_price_list = _set_price_list(cart_settings, None)

```

More verbose, but easier to read. Also instead of `if not selling_price_list` we can add the code in an else block depending on the functionality

erpnext/shopping_cart/cart.py
```diff
@@ -334,23 +334,28 @@ def set_price_list_and_rate(quotation, cart_settings):
 		# set it in cookies for using in product page
 		frappe.local.cookie_manager.set_cookie("selling_price_list", quotation.selling_price_list)
 
-def _set_price_list(quotation, cart_settings):
+def _set_price_list(cart_settings, quotation={}):
```
Alchez: Using mutable values as default arguments is discouraged

* https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/
* https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
* https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

doniafrizal:
version-13 also missing add to cart

doniafrizal:
@marination @Alchez @nabinhait can You help to Fix this same issue on version-13 ?

johnathon-walnut:
Is this missing in version 14 also?

PR #21920, opened 2020-05-25
chore: Added Italian Import Supplier Invoice to Menu

marination:
- Added Italian Import Supplier Invoice to Buying Module View if any Company is Italian
Screenshot 2020-05-25 at 9 34 48 PM

- Added Permissions similar to Purchase Invoice for Import Supplier Invoice

### DIFF
--- a/erpnext/regional/doctype/import_supplier_invoice/import_supplier_invoice.json
+++ b/erpnext/regional/doctype/import_supplier_invoice/import_supplier_invoice.json
@@ -1,5 +1,4 @@
 {
- "actions": [],
  "creation": "2019-10-15 12:33:21.845329",
  "doctype": "DocType",
  "editable_grid": 1,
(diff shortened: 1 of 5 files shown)


erpnext/regional/doctype/import_supplier_invoice/import_supplier_invoice.json
```diff
@@ -92,13 +91,54 @@
...
+   "email": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "Accounts User",
+   "share": 1,
+   "write": 1
+  },
+  {
+   "email": 1,
```
deepeshgarg007: Permission for regional specific doctypes should be added in the regional setup and add a patch to add permissions for existing users. Current fix will make the doctype visible for all countries.

Link for reference
https://github.com/frappe/erpnext/blob/develop/erpnext/regional/india/setup.py#L21

deepeshgarg007 (changes requested):

marination:
@deepeshgarg007  done

deepeshgarg007 (approved):

PR #21925, opened 2020-05-26
fix: check for warehouse in the woocommerce settings

Mangesh-Khairnar:
- **Fixes:** ``` frappe.exceptions.ValidationError: Please set Warehouse in Woocommerce Settings ```
- **Issue:** If user adds a warehouse into warehouse settings and deletes the stores warehouse from the warehouse master. Validation breaks, since, it expects that default warehouse to exists.


Mangesh-Khairnar:

nabinhait (approved):

PR #21974, opened 2020-05-27
fix: compare start and end time to prevent negative diff

scmmishra:


scmmishra:

nabinhait (approved):

PR #22000, opened 2020-05-28
fix: make transaction date of the oldest transaction as the last integration date

Mangesh-Khairnar:
**Issue** 
- Currently, last integration date is decided by the date/timezone of the server. However, when transactions are synced over a defined period, it fetches all the transactions based on the transaction date. 
- Transaction date is defined by a financial institution without any timezone information. [source](https://support.plaid.com/hc/en-us/articles/360008271754-Transaction-dates)
- This causes those transactions to be missed which are not in the same timezone as the server. Suppose servers are in the Region A with timezone that is ahead of the date that are recorded in plaid. Those transactions will not be synced, since the last integration date has changed in the bank account server.

**Proposed Fix**
- set the transaction date of the oldest transaction in the sync as the last integration date. This will make sure the transactions are in sync with the date  that the financial institution is following.

### DIFF
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
@@ -124,9 +124,8 @@ def add_account_subtype(account_subtype):
 
 @frappe.whitelist()
 def sync_transactions(bank, bank_account):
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

Mangesh-Khairnar:
@Mergifyio backport version-12-hotfix, version-11-hotfix

PR #22001, opened 2020-05-28
fix: transaction date not found in sales invoice

nextchamp-saqib:
Create a Pricing Rule with a specific item code, check is cumulative, set discount percentage and min max amount.

Adding the item in a sales invoice, throws unknown column transaction_date error.


rohitwaghchaure (approved):

nextchamp-saqib:

PR #22011, opened 2020-05-28
fix: Default period start date and period end date for financial statements

deepeshgarg007:
On selecting the "Date Range" filter for the first time user gets a "Date not in any active fiscal year" error if "To Date" is not in any active fiscal year


nabinhait (approved):

nabinhait:

PR #22027, opened 2020-05-28
fix: Validate Child Doctype Links

marination:
- During Data Import if a Purchase Order is mentioned in Purchase Receipt Item Table, user assumes PO and PR are linked.
- Import goes through, without checking for row id linking.
- The received qty does not reflect in the PO because the child table rows aren't actually linked. Hence, inconsistent data.

**Fix:**
- Check if `against doctype` is mentioned, check if the row detail of `against_doctype` is there.
- Eg. If `material_request` exists in a Stock Entry Detail row, but `material_request_item` does not, prompt error
-  On invalid linking:
    Screenshot 2020-05-28 at 10 45 14 PM

ToDo:

- [x] Cover Production Plan validation


marination:
depends on #22146

rmehta (commented):
@marination is this still required. Open for ~4 months now? Also tests missing.

marination:
@rmehta will close and reopen when i start working on it

PR #22036, opened 2020-05-29
fix: Migrate standard_tax_exemption_amount if field exists

nabinhait:


nabinhait:

brysonreece:
The original issue related to this pull request is still affecting v12 installations. Would it make sense to backport this to `version-12-hotfix` as well?

PR #22044, opened 2020-05-29
fix(Accounting): Prevent user creation of Payment Request

abhishekbalam:
User should not be able to create Payment Request from listview manually.

Payment request should only be created from Sales Order/ Purchase order of invoices.

From:
Screenshot 2020-05-29 at 5 59 41 PM

To:
Screenshot 2020-05-29 at 5 51 04 PM


nabinhait (commented):
What is the problem creating a payment request manually from list view?

abhishekbalam:
@nabinhait 

> What is the problem creating a payment request manually from list view?

You cant do it manually, the reference doctype and name cant be manually filled in the form, and they are mandatory.

Assumption here is it is made for creating Payment Requests from Sales/Purchase Order of invoices only

Empty Form:

image

Prefilled form for Purchase Invoice:

pp

nabinhait (approved):

PR #22048, opened 2020-05-29
fix: submitted sales order can be updated with no permission

nextchamp-saqib:
- Check permissions before updating items in sales order. If a user doesn't have create or write permissions for Sales Order they won't be able to update items.

- Conversion factor will now be modifiable from update items dialog (only if stock uom and uom differ)
   Screenshot 2020-05-29 at 10 00 12 PM


marination:
working on changes in this, will fix and merge

marination (dismissed):

rohitwaghchaure (approved):

PR #22049, opened 2020-05-29
fix: Project link not set in accounts other than profit and loss accounts

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/22051


erpnext/stock/doctype/purchase_receipt/purchase_receipt.json
```diff
@@ -107,6 +106,7 @@
   "amended_from",
   "range",
   "column_break4",
+  "project",
```
nabinhait: Project field is already there. And it is failing travis as well.

nabinhait (changes requested):

nabinhait (approved):

PR #22081, opened 2020-06-02
fix: Add total debit in Journal Entry list view

deepeshgarg007:
Replaced `Total Amount` from list view with `Total Debit` since total amount was only calculated for specific uses cases and it was just an empty space in list view for most of the entries.


deepeshgarg007:

rohitwaghchaure (approved):

PR #22086, opened 2020-06-03
fix(India): Reverse charge mechanism for GST

deepeshgarg007:
Currently system doesn't creates a tax liability against inward supplies that are liable to reverse charge
ITC can also be claimed on such reverse charges. So from now on the system will also create a Tax Liability for such inward supplies and net effect on such accounts will be zero.

Same will also be reflected  in GSTR3B report accordingly


rohitwaghchaure (approved):

rohitwaghchaure:
**GL Entries**

image

PR #22088, opened 2020-06-03
fix(India): Reverse charge mechanism for GST

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/22086

Fixes: #22141


erpnext/regional/doctype/gstr_3b_report/gstr_3b_report.py
```diff
@@ -158,7 +158,7 @@ def get_data(self):
 
 		self.prepare_data("Sales Invoice", outward_supply_tax_amounts, "sup_details", "osup_det", ["Registered Regular"])
 		self.prepare_data("Sales Invoice", outward_supply_tax_amounts, "sup_details", "osup_zero", ["SEZ", "Deemed Export", "Overseas"])
-		self.prepare_data("Purchase Invoice", inward_supply_tax_amounts, "sup_details", "isup_rev", ["Registered Regular"], reverse_charge="Y")
+		self.prepare_data("Purchase Invoice", inward_supply_tax_amounts, "sup_details", "isup_rev", ["Unregistered"], reverse_charge="Y")
```
mujeerhashmi: I think you should also consider "Overseas" gst category #22141

PR #22096, opened 2020-06-03
fix: '>=' not supported between instances of 'str' and 'int'

marination:
Protect our data types :(


Alchez:
If this is happening in multiple places, it might be better to debug why it's setting it as a string in the first place

marination:
@Alchez yeah we're on it

PR #22114, opened 2020-06-04
fix: pass date range instead of from date

prssanna:
Pass date range to leaderboard instead of from date

frappe: https://github.com/frappe/frappe/pull/10589


erpnext/startup/leaderboard.py
```diff
@@ -154,24 +163,36 @@ def get_all_sales_partner(from_date, company, field, limit = None):
...
 		limit %s
-	""", (from_date, company, cint(limit)), as_dict=1)
+	""".format(date_condition=date_condition), (company, cint(limit)), as_dict=1)
+
+def get_date_condition(date_range, field):
+	date_condition = ''
+	if date_range:
+		date_range = frappe.parse_json(date_range)
+		from_date, to_date = date_range
+		date_condition = "and {0} between '{1}' and '{2}'".format(field, from_date, to_date)
```
surajshetty3416: since this user input is directly passed to the query, the query becomes vulnerable to SQLi.

consider escaping the user inputs (using `frappe.db.escape`) or pass values as parameters to the query.

erpnext/startup/leaderboard.py
```diff
@@ -154,24 +163,36 @@ def get_all_sales_partner(from_date, company, field, limit = None):
...
 	return frappe.get_list('Sales Order', fields=[
 		'`sales_partner` as name',
 		'{} as value'.format(select_field),
 	], filters=filters, group_by='sales_partner', order_by='value DESC', limit=limit)
 
 @frappe.whitelist()
-def get_all_sales_person(from_date, company, field = None, limit = 0):
+def get_all_sales_person(date_range, company, field = None, limit = 0):
+	date_condition = get_date_condition(date_range, 'sales_order.transaction_date')
+	print('date condition', date_condition)
```
surajshetty3416:
```suggestion
```

surajshetty3416 (changes requested):

surajshetty3416 (approved):

nabinhait (approved):

centaur78:
hi @prssanna 

Getting the below warning while compiling 

```
Compiling Python Files...
../apps/erpnext/erpnext/startup/leaderboard.py:126: SyntaxWarning: str indices must be integers or slices, not tuple; perhaps you missed a comma?
  filters.append(['posting_date', 'between' [date_range[0], date_range[1]]])
```

PR #22115, opened 2020-06-04
fix(item_attribute): Handle whitespace issue in item attribute value and abbr.

KaviyaPeriyasamy:
This PR fixes the whitespace issues in the item attribute value and abbreviation.

## Issue
The item attribute value created mistakenly with white spaces at the beginning and end:

- In this case, if the user creates a new variant from the UI, it creates the item name with the whitespaces as in the item attribute value. 
- But in some other doctype, when the item attribute acts as a link field [stripping of values are handled], it displays the attribute values without whitespaces. Here, if we call the ```get_variant```  function of item_variant, it returns ```None``` but the variant with the same attribute value exists in the item list.

Eg: Attribute - "Colour", Attribute value -  " White"
      Item created - "T-Shirt- White"
      But if we look for "T-Shirt-White" in ```get_variant```  it returns None.
      
## Before Fix
Peek 2020-06-04 23-02
## After Fix
Peek 2020-06-05 10-21

### DIFF
--- a/erpnext/controllers/item_variant.py
+++ b/erpnext/controllers/item_variant.py
@@ -315,7 +315,7 @@ def make_variant_item_code(template_item_code, template_item_name, variant):
 			# 	exc=InvalidItemAttributeValueError)
 
 		abbr_or_value = cstr(attr.attribute_value) if item_attribute[0].numeric_values else item_attribute[0].abbr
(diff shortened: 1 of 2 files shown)


KaviyaPeriyasamy:
There may or may not be the use case to be valid. we will handle this case elsewhere. I will close this PR now.

Issue #21515, opened 2020-04-29
[Feature Request] H1 tag on every website page for SEO


nameduser0:
**Is your feature request related to a problem? Please describe.**
One of the vital aspects of any page for SEO is the main heading, or H1 tag, as explained [here](https://youtu.be/DvwS7cV9GmQ?t=420)
The homepage however only has H3 tags when using non-default Hero modes

**Describe the solution you'd like**
Change all website pages to use the H1 tag at least once

PR #22119, opened 2020-06-05
feat: Homepage refactor (develop)

### LINKED ISSUE #20744: Integration of auto-generated "Publications" section and home page section functionality on the HomePage
In Erpnext 12 under webpage the functionality is currently developed to make a modern good looking home page.
A new doc type homepage sections defines the content of sections. that can either be based on Cards or HTML.
However, there is also an section that is **hard-code and auto generated** named "Publications" that contains the last 3 Blogs.

I feel that these two rather similar functionalities that both generate a section should/could be integrated, so that the website manager has more control. As said, "Publications" has no controls and is fully hard code.

Update: 8 March 2020

Some experimenting with someone with hardly coding background.
At the home page defaults one can copy the generated html code from the webpage using show source code.

In the home page section doc i have created entries Hero, Products where I have copy pasted the code as described in paragraph above and than select Hero Section Based on "Home Page Section"

Using this approach there is almost full control over appearance of the Home Page.
Unfortunately, even when selecting Hero Section Based on "Home Page Section" the Publication Section is AutoCreated.   

In my view there are several ways to to allow maximum functionality>
1. Stick to current approach but add to home page also the now autogenerated "Publications"
**Furthermore add an enable/disable tickmark and !!! important !!! the "section order" number that is also defined in de homepage sections doc.**

2 An alternative would be to make of the Home page doc only a skeleton. Similar to print formats the home page section doc would contain a number of default sections code by frappe like hero, products, publications, and so on.  The user can use these " as they are" or copy and edit.

3. A kind of mix between 1 and 2........In the home page there are sections for Hero, Products, Publications, and so on....an a button "show code" that can be pasted in the home page section doc.
So...similar what I have done now, by copying the code form the generated home page.

In my view the amount of extra coding to implement these suggestions is very minimal...and the result is that the user has full freedom to design the homepage without the current restriction due to hardcoding some of the sections.

see experiment..may change over time  dogsloveit.erpnext.com/home

### LINKED ISSUE #21514: [V12 bug] Homepage meta description field hidden
## Description of the issue
By default, ERPNext has the homepage description field set to:  

> This is an example website auto-generated from ERPNext

This field is used to set the html head meta tag: description amongst others. This tag is used by search engines for the website content shown on the search page, so is _vital_ for SEO.

However, if you change the `Hero Section Based On` field in Homepage settings, the description field is hidden from view, but is still used to populate the head tags on the homepage, including a meta property tag which is not overridable by setting meta tags separately.

**Output of `bench version`**
```
ERPNext: v12.7.1 (version-12)
Frappe Framework: v12.5.1 (version-12)
```
https://github.com/frappe/erpnext/blob/e4fca91792fa3eef49a947697f4ea2af7c6a0ca9/erpnext/templates/pages/home.py#L49

Screen Shot 2020-04-29 at 17 59 26

1. Change `Hero Section Based On` field in Homepage settings to `Homepage Section`

### Observed result
Description field is hidden

### Expected result
Description field should **always** be visible

nameduser0:
Closes #20744
Closes #21514
Affects #21515

Refactored Homepage:
- Sections, Cards, Products & Blog Posts are now configured in Web Page Sections to allow reordering, mix 'n match and selection of specific objects (#20744)
- Homepage Description field remains visible because it is used in meta tags (#21514)
- Section titles can now be rendered with an H1 tag for SEO (#21515)
- Sections can be rendered as Bootstrap Grid, Card Group, Card Deck or Card Columns
- Cards can be rendered vertical or horizontal (the template macro should be reused for all cards, e.g. products, to give consistency)
- Card links can now be configured with anchor text and nofollow attributes for SEO
- Other Web Pages could also be parents to the new `Web Page Sections` (not just the Homepage)
- Other ERPNext docs that make sense as website published cards could easily be added to `Web Page Section Item`

New DocType hierarchy:
- Homepage (Single)
  - Web Page Sections (Child)
    - Web Page Section (Parent)
      - Web Page Section Item (Child; contains Dynamic Link to:)
        - Web Page Card (Parent)
        - Item (Parent)
        - Blog Post (Parent)

The functionality has been refactored with a view to being integrated with the frappe Web Page page builder, but this was too much work.
If acceptable after review, I will also submit to the version-12-hotfix branch.
I will now update the Homepage documentation to match.


erpnext/patches/v12_0/refactor_homepage_20744.py
```diff
@@ -0,0 +1,113 @@
...
+		frappe.db.delete('DocType', 'Homepage Featured Product')
+	except pymysql.err.ProgrammingError:
+		# Assume does not exist
+		homepage_products = []
+
+	frappe.reload_doc('Portal', 'doctype', 'Homepage')
+	new_homepage = frappe.get_doc('Homepage')
+
+	try:
+		cards = frappe.db.get_all('Homepage Section Card', fields='*', order_by='parent,idx')
```
Anurag810, line 39: why get all fields? list only those which are required.
 ``` fields = ['name', 'title']```

erpnext/patches/v12_0/refactor_homepage_20744.py
```diff
@@ -0,0 +1,113 @@
...
+		frappe.rename_doc('DocType', 'Homepage Section Card', 'Web Page Card', force=True)
+		frappe.reload_doc('Portal', 'doctype', 'Web Page Card')
+		frappe.desk.doctype.bulk_update.bulk_update.update('Web Page Card', 'route_follow', 1)
+	except pymysql.err.ProgrammingError:
+		cards = []
+
+	try:
+		sections = frappe.db.get_all('Homepage Section', fields='*', order_by='section_order')
```
Anurag810, line 47: why get all fields? list only those which are required.
```fields = ['name', 'title']```

File: erpnext/patches/v12_0/refactor_homepage_20744.py [hunk 3]
```diff
@@ -0,0 +1,113 @@
...
+		child = new_homepage.append('page_sections', {})
+		child.section_name = product_section.name
+		new_homepage.save()
+
+	frappe.db.sql_ddl("""DROP TABLE IF EXISTS `tabHomepage Featured Product`;""")
+	frappe.db.sql_ddl("""ALTER TABLE `tabWeb Page Section` DROP IF EXISTS `section_order`;""")
+
+	try:
+		blogs = frappe.db.get_all('Blog Post',
+			fields=['name'],
```
Anurag810, line 93: It already returns the name, if fields argument is not there.

Anurag810 (changes requested):

erpnext/patches/v12_0/refactor_homepage_20744.py
Anurag810: no need to specify fields = ["name"]

erpnext/patches/v12_0/refactor_homepage_20744.py
nameduser0: It's a patch, so gets run once only, so it makes little difference.

erpnext/patches/v12_0/refactor_homepage_20744.py
nameduser0: It's more verbose so people know exactly what is happening rather than needing to know the default behaviour.

erpnext/portal/doctype/homepage/test_homepage.py
```diff
@@ -1,19 +1,126 @@
...
 from __future__ import unicode_literals
 
 import frappe
 import unittest
 from frappe.utils import set_request
+from bs4 import BeautifulSoup
 from frappe.website.render import render
 
 class TestHomepage(unittest.TestCase):
+	def setup_test_data(self):
```
barredterra, line 13: Name this method "setUp" and it will get executed automatically before every test case. (Check the unittest docs for details.)

barredterra (changes requested):
Nice work. Btw, I don't think it should go into `version-12-hotfix`, as you mentioned, because it's not a fix.

erpnext/patches.txt
```diff
@@ -694,4 +694,5 @@ execute:frappe.delete_doc("Report", "Department Analytics")
 execute:frappe.rename_doc("Desk Page", "Loan Management", "Loan", force=True)
 erpnext.patches.v12_0.update_uom_conversion_factor
 erpnext.patches.v13_0.delete_old_purchase_reports
-erpnext.patches.v12_0.set_italian_import_supplier_invoice_permissions
\ No newline at end of file
+erpnext.patches.v12_0.set_italian_import_supplier_invoice_permissions
+erpnext.patches.v12_0.refactor_homepage_20744
```
nameduser0, line 698: Move patch to v13_0

erpnext/patches/v12_0/refactor_homepage_20744.py
```diff
@@ -0,0 +1,113 @@
+# Copyright (c) 2020, Frappe and Contributors
```
nameduser0, line 1: Move patch to v13_0

rmehta:
@casesolved-co-uk thanks for this PR, but we can deprecate this feature with the new Web Page Builder https://docs.erpnext.com/docs/user/manual/en/website/web-page-builder

Shall we remove the "Home Page" altogether?

@netchampfaris

netchampfaris:
@rmehta Yes, we can run a patch to create these pages using Page Builder.

nameduser0:
@rmehta & @netchampfaris so you can reference existing Item and Blog db content to create cards on other pages using the Web Page Builder?

If not it doesn't close #20744, removes a valuable feature and doesn't conform to SSOT/DRY principles

nameduser0:
Active

PR #22122, opened 2020-06-05
fix(HR): Fix half day error when only 1 day is selected

abhishekbalam:
When Half Day checkbox is selected for only one day, 
the Half Day Date is hidden but still mandatory.

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.js
+++ b/erpnext/hr/doctype/leave_application/leave_application.js
@@ -38,6 +38,9 @@ frappe.ui.form.on("Leave Application", {
 	},
 
 	validate: function(frm) {
+		if (frm.doc.from_date == frm.doc.to_date && frm.doc.half_day == 1){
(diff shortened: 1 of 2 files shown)


erpnext/hr/doctype/leave_application/leave_application.js
```diff
@@ -38,6 +38,9 @@ frappe.ui.form.on("Leave Application", {
 	},
 
 	validate: function(frm) {
+		if (frm.doc.from_date == frm.doc.to_date && frm.doc.half_day == 1){
```
Anurag810, line 41: Add this validation at python side as well

Anurag810 (changes requested):

erpnext/hr/doctype/leave_application/leave_application.js
abhishekbalam: Added in next commit 👍

Anurag810 (approved):

Anurag810 (approved):

PR #22123, opened 2020-06-05
fix(HR): Change Due Advance Amount to Pending Amount

abhishekbalam:
Fieldname change from `Due Advance Amount`
To `Pending Amount` as it is confusing.

Changed reference name in the code too for easy 
debugging in the future.


deepeshgarg007 (changes requested):
You'll also have to write a patch to copy data from old fieldname to new fieldname in existing records

abhishekbalam:
> You'll also have to write a patch to copy data from old fieldname to new fieldname in existing records

Sure, I'll figure it out and push 🖖

abhishekbalam:

Anurag810 (approved):

Anurag810:

PR #22129, opened 2020-06-06
fix employee leave balance report

ammarhararah:
I guess I don't need to explain alot.
**department_name** variable is supposed to be **department**
also append function does not return values so the assignment is not right!


rohitwaghchaure:
send PR on version-12-hotfix and not on version-12 branch

PR #22137, opened 2020-06-07
fix: shopping cart UI fixes

anupamvs:
Before-
image

After-
image

Before-
image

After-
image


erpnext/shopping_cart/cart.py
```diff
@@ -42,9 +42,9 @@ def get_cart_quotation(doc=None):
 
 	return {
 		"doc": decorate_quotation_doc(doc),
-		"shipping_addresses": [{"name": address.name, "display": address.display}
+		"shipping_addresses": [{"name": address.name, "title": address.address_title, "display": address.display}
```
marination, line 45: title isnt used anywhere as far i see. Don't think it is necessary to include it here.

anupamvs: @marination it is used in erpnext/templates/includes/cart/address_card.html

marination (dismissed):

anupamvs:

marination (approved):

PR #22155, opened 2020-06-08
feat(projects): bill all hours by default

barredterra:
### Problem
I frequently struggle with **Timesheet**. In the editable grid I can enter the time worked. For `billing_hours` I have to enter the **Timesheet Detail**, check `billable` and manually copy the working `hours` to `billing_hours`.

### Solution
With this PR, the `billing_hours` are equal to the `hours` worked by default. They can still be changed manually if required. This way it doesn't generate extra work if the values are different but saves some time when they are equal.

### DIFF
--- a/erpnext/projects/doctype/timesheet/timesheet.js
+++ b/erpnext/projects/doctype/timesheet/timesheet.js
@@ -206,9 +206,6 @@ frappe.ui.form.on("Timesheet Detail", {
 		update_billing_hours(frm, cdt, cdn);
 		update_time_rates(frm, cdt, cdn);
 		calculate_billing_costing_amount(frm, cdt, cdn);
-
-		// bill all `hours` by default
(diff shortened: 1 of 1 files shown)


erpnext/projects/doctype/timesheet/timesheet.js
```diff
@@ -206,6 +206,9 @@ frappe.ui.form.on("Timesheet Detail", {
 		update_billing_hours(frm, cdt, cdn);
 		update_time_rates(frm, cdt, cdn);
 		calculate_billing_costing_amount(frm, cdt, cdn);
+
+		// bill all `hours` by default
+		frappe.model.set_value(cdt, cdn, "billing_hours", locals[cdt][cdn].hours);
```
Alchez: I think this should happen in the `update_billing_hours` function

rohitwaghchaure (approved):

Issue #22157, opened 2020-06-09
[Bug] V12 - Pricing rule: Missing income accout


pipech:
## Description of the issue

Pricing Rule: When create Pricing rule which get free item, Income account of item will be missing.

![image](https://user-images.githubusercontent.com/24489363/84101706-ba608680-aa38-11ea-80ff-093b5ac2a827.png)

### Pricing Rule

**Get Free A item when buy A item.**
There are other pricing rules but they are all disabled except this one.

![image](https://user-images.githubusercontent.com/24489363/84101866-2ba03980-aa39-11ea-9cb2-0386801e547a.png)

**Output of `bench version`**
```
erpnext 12.2.2
frappe 12.6.1
```

ERPNext is version12-hotfix branch, so version might be off.

michellealva:
@pipech can you check if you have set the Income account in the item row? Expand the item row and check if this is set.

![image](https://user-images.githubusercontent.com/50285544/84499749-ccf4ed00-acd0-11ea-9a2f-ec51387cc316.png)

If you have set the default account in the company master, this should be automatically fetched.

PR #22158, opened 2020-06-09
fix(pricing_rule): missing income account when getting free product

pipech:
## Fix for > [Bug] V12 - Pricing rule: Missing income account #22157

Adding `income_account` value to `get_product_discount_rule` which is function getting value for free product.

Getting `income_account` value from existing get default function.


rohitwaghchaure (approved):

rohitwaghchaure:
@pipech send the PR on develop branch too

Issue #22159, opened 2020-06-09
[Bug] V12 - Pricing rule: Transaction rule doesn't got apply on frontend


pipech:
## Description of the issue

Pricing Rule: Transaction based rule doesn't got apply on frontend which will affect the most on POS page. Invoice total will be incorrect since POS page doesn't save before submit. (Transaction based rule will work correctly from backend.)

It doesn't work on both product and price discount.

#### Discount 10% on 1000+ THB Transaction

![image](https://user-images.githubusercontent.com/24489363/84103683-ffd38280-aa3d-11ea-9634-d26f72fd2311.png)

#### Get free tote bag at 10,000+ transaction

![image](https://user-images.githubusercontent.com/24489363/84103741-27c2e600-aa3e-11ea-8d69-93c58b69f5e1.png)

**Output of `bench version`**
```
erpnext 12.2.2
frappe 12.6.1
```

ERPNext is version12-hotfix branch, so version might be off.

PR #22160, opened 2020-06-09
fix(pricing_rule): transaction rule doesn't got apply on frontend

pipech:
## Fix for > [Bug] V12 - Pricing rule: Transaction rule doesn't got apply on frontend #22159

Apply transaction rule on frontend through `transaction.js` and `point_of_sale.js`.

This will work on both Sales Invoice page and Online POS page and both product discount and price discount.

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -175,6 +175,51 @@ def apply_pricing_rule(args, doc=None):
 
 	return out
 
+@frappe.whitelist()
+def apply_transaction_pricing_rule(doc):
+	from erpnext.accounts.doctype.pricing_rule.utils import apply_pricing_rule_on_transaction
+
+	# converting dict into doc
+	if isinstance(doc, string_types):
+		doc = json.loads(doc)
+	if doc:
+		doc = frappe.get_doc(doc)
+
+	apply_pricing_rule_on_transaction(doc)
+
+	discount = {}
+
+	if (
+		doc.apply_discount_on and (
+			doc.additional_discount_percentage or
+			doc.discount_amount
+		)
+	):
+		discount['price'] = {
+			'apply_discount_on': doc.apply_discount_on,
+			'additional_discount_percentage': doc.additional_discount_percentage,
+			'discount_amount': doc.discount_amount,
+		}
+
+	for item in doc.items:
+		if item.is_free_item:
+			discount['product'] = {
+				'item_code': item.item_code,
+				'qty': item.qty,
+				'rate': item.rate,
+				'price_list_rate': item.price_list_rate,
+				'is_free_item': item.is_free_item,
+				'item_name': item.item_name,
+				'description': item.description,
+				'stock_uom': item.stock_uom,
+				'uom': item.uom,
+				'conversion_factor': item.conversion_factor,
+				'amount': item.amount,
+				'income_account': item.income_account,
+			}
+
+	return discount
+
 def get_serial_no_for_item(args):
 	from erpnext.stock.get_item_details import get_serial_no
 
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -569,6 +569,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 										let key = item.name;
 										me.apply_rule_on_other_items({key: item});
 									}
+								},
+								() => {
+									me.frm.trigger('qty', cdt, cdn);
 								}
 							]);
 						}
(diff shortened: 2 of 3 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -569,6 +569,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 										let key = item.name;
 										me.apply_rule_on_other_items({key: item});
 									}
+								},
+								() => {
+									me.frm.trigger('qty', cdt, cdn);
```
rohitwaghchaure: Why trigger qty?

erpnext/public/js/controllers/transaction.js
```diff
@@ -1264,6 +1267,43 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 				}
 			}
 		});
+
+		const transaction_pricing_rule = this.frm.call({
```
rohitwaghchaure, line 1271: This should be call in the callback of apply_pricing_rule

erpnext/public/js/controllers/transaction.js
```diff
@@ -1264,6 +1267,43 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+			method: "erpnext.accounts.doctype.pricing_rule.pricing_rule.apply_transaction_pricing_rule",
+			args: {	doc: me.frm.doc },
+			callback: function(r) {
+				if (!r.exc && r.message) {
+					r = r.message;
+					if ('price' in r) {
```
rohitwaghchaure, line 1277: why not checking the value of the field price_or_product_discount

erpnext/public/js/controllers/transaction.js
pipech: To trigger apply_pricing_rule, actually i should change this to `this.apply_pricing_rule(item, true);`

erpnext/public/js/controllers/transaction.js
pipech: May i ask why?

erpnext/public/js/controllers/transaction.js
pipech: Return doesn't have `apply_transaction_pricing_rule` field.

erpnext/public/js/controllers/transaction.js
pipech: https://github.com/frappe/erpnext/blob/bc4737a1874b1c021c98fb50ae0c4b2dfecb5510/erpnext/public/js/controllers/transaction.js#L573-L575

pipech:
@rohitwaghchaure sorry to tag you but could you please review this ?

erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -175,6 +175,51 @@ def apply_pricing_rule(args, doc=None):
...
 	return out
 
+@frappe.whitelist()
+def apply_transaction_pricing_rule(doc):
+	from erpnext.accounts.doctype.pricing_rule.utils import apply_pricing_rule_on_transaction
+
+	# converting dict into doc
+	if isinstance(doc, string_types):
+		doc = json.loads(doc)
+	if doc:
```
sagarvora, line 185: I don't know if this condition is needed at all. And if it _is_ needed, what happens if it's falsy?

sagarvora (changes requested):

rohitwaghchaure:
@pipech I was doing the testing for the POS, and I have found one issue that while removing the main item from the cart the free items is still remained in the cart. Please check below GIF

pos_testing_issue

I am closing this PR, requesting you to kindly fix this issue and reopen the PR

federicocalvo:
@rohitwaghchaure @sagarvora Hi there, this PR is the only one that comes close to a correction in order to be able to use **pricing rules** with the **transaction** condition.

As of today, it still does not work and this PR was not reopened or revised.

Can it at least be reopened? We might be able to see about continuing it with our team.

Thanks

sagarvora:
👋🏼 @federicocalvo 

Can you please open another one on `develop`? Or do you intend to use this on the `v12` POS?

federicocalvo:
Hi @sagarvora, yes of course, but in our case it would be excluding the POS (I think it is very complex) but in the transactions as Sale Invoice (it should refer to Sales Order and Cuotation as well).

Then someone could separately try to solve it in point_of_sale.js.

sagarvora:
> Hi @sagarvora, yes of course, but in our case it would be excluding the POS (I think it is very complex) but in the transactions as Sale Invoice (it should refer to Sales Order and Cuotation as well).
> 
> Then someone could separately try to solve it in point_of_sale.js.

Great, tag / assign me. I'll review it.

Issue #22167, opened 2020-06-09
[Bug] V12 - Loyalty program: Loyalty Point Entry use wrong Tier


pipech:
## Description of the issue

Loyalty program: Loyalty Point Entry use wrong Tier.
It skipped the first tier and use second tier.

### Loyalty Program

![image](https://user-images.githubusercontent.com/24489363/84162823-03e1bd80-aa9b-11ea-9926-8f7207f72870.png)

### Loyalty Point Entry

The first entry, point entry used Silver Tier which is the second tier.

![image](https://user-images.githubusercontent.com/24489363/84162884-165bf700-aa9b-11ea-84a7-db73b448376e.png)

![image](https://user-images.githubusercontent.com/24489363/84163027-3ee3f100-aa9b-11ea-9573-6e8fe3913143.png)

**Output of `bench version`**
```
erpnext 12.2.2
frappe 12.6.1
```
ERPNext is version12-hotfix branch, so version might be off.

PR #22168, opened 2020-06-09
fix(loyalty): loyalty point entry use wrong tier

pipech:
## Fix for > [Bug] V12 - Loyalty program: Loyalty Point Entry use wrong Tier #22167

This is incorrect because it loop from highest and see if cumulative amount is less than min tier spend but it won't work if min tier spend is at 0 (Which it can be since min tier spend is not mandatory field)
that might be why he add i == 0 to skip this error.

The new method loop from lowest and check if cumulative spend is more than min spend.

### DIFF
--- a/erpnext/accounts/doctype/loyalty_program/loyalty_program.py
+++ b/erpnext/accounts/doctype/loyalty_program/loyalty_program.py
@@ -41,10 +41,16 @@ def get_loyalty_program_details_with_points(customer, loyalty_program=None, expi
 	loyalty_program = frappe.get_doc("Loyalty Program", loyalty_program)
 	lp_details.update(get_loyalty_details(customer, loyalty_program.name, expiry_date, company, include_expired_entry))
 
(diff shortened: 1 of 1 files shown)


olamide2:
Thank you. I have been complaining about this for months

rohitwaghchaure (approved):

rohitwaghchaure:
@pipech Kindly send this PR on develop branch too

marination:
@pipech Travis seems to be failing, please fix tests in `test_loyalty_program.py`
Screenshot 2020-06-10 at 7 08 19 PM

pipech:
I will but it will take time for me to setup testing properly.

I've tried to run it but in some version it will error on Running "before_test" or it will take too long just to run "before_test", I haven't got to unittest part.

Do you guys have any recommendation on setting up testing environment ?

But from initial look on unittest, I think the problem is customer doctype haven't got reload after create sales invoice so `loyalty_program_tier` will be none.

hoangdinhanphabe:
This issue still persists on v13 Beta 10

PR #22169, opened 2020-06-09
Leave balance report department and leave approvers fix

ammarhararah:
I guess I don't need to explain alot.
department_name variable is supposed to be department
also append function does not return values so the assignment is not right!


rohitwaghchaure:
@ammarhararah why changes in 37 files? Seems you have not rebase your local repo correctly, Kindly fix the issue and open the new PR

PR #22175, opened 2020-06-10
fix: set cost center in child table

Anurag810:
Set cost center in the child Table  from the cost center field in from outside the child table


Anurag810:

rohitwaghchaure (approved):

PR #20789, opened 2020-03-02
refactor: POS workflow

nextchamp-saqib:
## Goal
Point of Sale (online) will now create POS Invoices. These invoices are submitted faster than sales invoice as they don't hit Stock Ledger or General Ledger. These invoices will get merged into a consolidated Sales Invoices on POS Closing (or manually) based on customer.  Available Stock will be validated by recent POS Invoices and latest stock ledger entry.

## Roadmap
- [x] POS Invoices
- [x] POS Invoices can be merged into Sales Invoice
- [x] Auto merge on POS Closing
  - [x] Auto fetch transaction's details in POS Closing Voucher
- [x] Consolidated Credit Note for POS Returns
- [x] New UI
  - [x] Item Selector
     - [x] Fix item search and category selection
  - [x] Item Details
     - [x] Fix auto serial selection button for non serialised item
  - [x] Item Cart
  - [x] Payments
     - [x] Loyalty Points as a payment method
     - [x] Email Receipt / Print Receipt
  - [x] Past Orders Section
     - [x] Load past pos invoice in the UI. Can be submitted or draft
     - [x] Process Returns
- [x] POS Print Format
- [ ] Tests
- [ ] Docs

#### Item Selector

Screenshot 2020-04-12 at 3 20 08 PM

#### Item Details

Screenshot 2020-04-12 at 3 11 43 PM

#### Payments

Screenshot 2020-04-12 at 3 12 20 PM

#### Past Order List

Screenshot 2020-04-19 at 12 55 06 PM

PR #22190, opened 2020-06-10
Update POS.JS

Helkyd:

> Screenshots/GIFs


rohitwaghchaure:
@Helkyd we are removing the offline POS from the version 13.

For new pos check below links
https://github.com/frappe/erpnext/pull/20789
https://discuss.erpnext.com/t/feature-erpnext-v13-new-point-of-sale-beta-testing/61093

olamide2:
@rohit

Why not leave them to operate side by side ? The offline PoS is still a needed feature for some

Helkyd:
What i don't understand is why! As most of the places or countries using
POS will need the offline mainly African countries with slower links...

PR #22193, opened 2020-06-11
fix(UX): Notify user of expense account change in Purchase Invoice

deepeshgarg007:
User can enter the expense manually in Purchase Invoice if Perpetual Inventory is disabled.

But if user has perpetual inventory is enabled, expense is set by default based on certain conditions, this is too notify the user of the expense head change


deepeshgarg007:

nabinhait (approved):

PR #22195, opened 2020-06-11
fix: Handle unavailable Variants in Website

marination:
**Issue:(Web Portal)**
- When you click on **Configure** for an Template Item, it lets you add variant to cart even if out of stock
- On click of **Place Order** this happens:
`apps/erpnext/erpnext/shopping_cart/cart.py", line 81, in place_order`
      `if item.qty > item_stock.stock_qty[0][0]:`
`IndexError: tuple index out of range`

**Fix:**
- Check if `allow_items_not_in_stock` is enabled in Shopping Cart settings, if yes, Show Add to Cart, if no, show Not in Stock
Screenshot 2020-06-11 at 1 11 27 PM
- If Item is within cart somehow(older items), throw error saying Item is Not in Stock on clicking on **Place Order**


erpnext/portal/product_configurator/utils.py
```diff
@@ -243,6 +244,8 @@ def get_next_attribute_and_values(item_code, selected_attributes):
 	else:
 		product_info = None
 
+	product_info["allow_items_not_in_stock"] = cint(frappe.db.get_singles_value("Shopping Cart Settings", "allow_items_not_in_stock"))
```
rohitwaghchaure: From get_product_info_for_website you will get the info about the cart_settings
data.cart_settings.allow_items_not_in_stock?

PR #22199, opened 2020-06-11
fix: Don't prompt for Quality Inspection on Return Documents.

marination:
- Once QI is done on Receipt, don't prompt again on Return.

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -19,7 +19,7 @@ class QualityInspectionNotSubmittedError(frappe.ValidationError): pass
 class StockController(AccountsController):
 	def validate(self):
 		super(StockController, self).validate()
-		if not self.is_return:
+		if not self.get('is_return'):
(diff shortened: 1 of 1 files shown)


erpnext/controllers/stock_controller.py
```diff
@@ -19,7 +19,8 @@ class QualityInspectionNotSubmittedError(frappe.ValidationError): pass
 class StockController(AccountsController):
 	def validate(self):
 		super(StockController, self).validate()
-		self.validate_inspection()
+		if not self.is_return:
```
deepeshgarg007: image

Many doctypes are inherited from Stock Controller and some of them don't have the is return field. 
```suggestion
		if not self.get('is_return'):
```

deepeshgarg007 (changes requested):

marination:
@deepeshgarg007 done

deepeshgarg007 (approved):

PR #22210, opened 2020-06-12
fix: Revenue and Customer Count only in date range in Customer Acq. Report

abhishekbalam:
Improvement on this:
#22075

The revenue was being calculated from 
the 1st invoice ever created, and hence
is too high. 

Also territorywise customer counts were wrong,
for the same reason, as monthly compensates for 
report From Date by default. 

So limiting counting of customers and revenue 
within filter dates but considering all invoices
till To Date to determine new customers.


abhishekbalam:
Changes only done in the last function, had to 
convert spaces to tabs as I didn't have the last 
PR branch. From Line 162...

deepeshgarg007:
@abhishekbalam  can you resolve merge conflicts

abhishekbalam:
@deepeshgarg007 
Done, please check.
If possible pls run locally and see if the report looks fine 👍

erpnext/selling/report/customer_acquisition_and_loyalty/customer_acquisition_and_loyalty.py
```diff
@@ -174,14 +174,13 @@ def get_customer_stats(filters, tree_view=False):
 		filters, as_dict=1):
 
 		key = si.territory if tree_view else si.posting_date.strftime('%Y-%m')
+		new_or_repeat = 'new' if si.customer not in customers else 'repeat'
 		customers_in.setdefault(key, {'new': [0, 0.0], 'repeat': [0, 0.0]})
 
-		if not si.customer in customers:
-			customers_in[key]['new'][0] += 1
-			customers_in[key]['new'][1] += si.base_grand_total
+		if filters.from_date <= si.posting_date.strftime('%Y-%m-%d'):
```
deepeshgarg007:
```suggestion
		if getdate(filters.from_date) <= getdate(si.posting_date):
```

deepeshgarg007: Also don't forget to import getdate

deepeshgarg007 (changes requested):

erpnext/selling/report/customer_acquisition_and_loyalty/customer_acquisition_and_loyalty.py
abhishekbalam: Ok, done

deepeshgarg007 (approved):

PR #22211, opened 2020-06-12
feat: notify credit controller role users with credit limit extension requests

vishdha:
g3UnZrQ


rohitwaghchaure:
As this is a new feature we'll not add in the stable version

PR #22228, opened 2020-06-13
HR gross_pay

Helkyd:
Is adding the components with do_not_include_in_total so should be removed.//frm.set_value('gross_pay', gross_pay);

frappe.db.get_value('Salary Structure', {'name': frm.doc.salary_structure}, 'salary_component', (r) => {
		var gross_pay = 0.0;
		$.each(frm.doc["earnings"], function(i, earning) {
			if (earning.salary_component == r.salary_component) {
				earning.amount = wages_amount;
				frm.refresh_fields('earnings');
			}
			gross_pay += earning.amount;
		});
        //Is adding the components with do_not_include_in_total so should be removed.
		//frm.set_value('gross_pay', gross_pay);

		frm.doc.net_pay = flt(frm.doc.gross_pay) - flt(frm.doc.total_deduction);
		frm.doc.rounded_total = Math.round(frm.doc.net_pay);
		refresh_many(['net_pay', 'rounded_total']);
	});


erpnext/hr/doctype/salary_slip/salary_slip.js
```diff
@@ -156,7 +156,8 @@ var total_work_hours = function(frm, dt, dn) {
 			}
 			gross_pay += earning.amount;
 		});
-		frm.set_value('gross_pay', gross_pay);
```
rohitwaghchaure, line 159: This should not be removed. What if the do_not_include_in_total is unchecked? You should check the do_not_include_in_total condition at line no 153

Helkyd:
But this is already done on the py file... So here looks like a
duplication... Reason why I suggested to be removed.

rohitwaghchaure:
@Helkyd The JS code is not duplicate because py code calculate the gross pay amount on save of salary slip. But if user wants to see the gross pay amount while adding earning components then your changes will not help them. So just add the  if (! earning.do_not_include_in_total) in the if condition.

rohitwaghchaure:
@Helkyd Kindly fix the code and reopen the PR

PR #22233, opened 2020-06-13
fix: Billing address in for Purchase documents

deepeshgarg007:
Shipping Address and Billing Address in Purchase Invoice can be different similar to like Customer Address and Billing Address can be different in Sales Invoice


rohitwaghchaure (approved):

deepeshgarg007:

PR #22287, opened 2020-06-17
fix: Quality procedure fixes

marination:
- Check if any child procedure belongs to another parent. If yes:
  Screenshot 2020-06-17 at 5 58 26 PM

- On saving, parent is set in child processes. Set parent only for those children without a parent. Used `frappe.db.get_value` to reduce `get_doc` time and also to avoid triggering save for each child (which will run validations for all the children until it reaches a leaf node)  

- Filter child procedure field. 
  Screenshot 2020-06-17 at 5 34 35 PM

   Don't let users select :
   * The same process as the current document (Process cant be its own child)
   * The parent of the current document (Its parent cant be its child)

- Allow Renaming
- Disable New button in tree view (UX)
- Editable grid for Quality Procedure Process table (UX)


marination:

marination:
Codacy only failing due to json file

PR #21277, opened 2020-04-15
fix: asset maintenance fixes

nextchamp-saqib:
Problems fixed: 

Assign to gets auto fetched as maintenance team manager on save.
Assign to cannot be set before save.
Maintenance team with no manager doesn't lets asset maintenance to get saved.

PR #22290, opened 2020-06-17
fix: asset maintenance fixes

marination:
Port of https://github.com/frappe/erpnext/pull/21277


marination:

PR #22296, opened 2020-06-17
fix(patch): add patch for setting issue metrics

ruchamahabal:
Patch for setting the following metrics in old Issue documents:
- Average Response Time
- Resolution Time
- User Resolution Time

This feature was added in the PR: https://github.com/frappe/erpnext/pull/21617

### DIFF
--- a/erpnext/patches/v13_0/update_issue_metrics.py
+++ b/erpnext/patches/v13_0/update_issue_metrics.py
@@ -8,7 +8,8 @@ def execute():
 	if frappe.db.exists('DocType', 'Issue'):
 		frappe.reload_doctype('Issue')
 
-		for parent in frappe.get_all('Issue', order_by='creation desc', limit=500):
+		count = 0
+		for parent in frappe.get_all('Issue', order_by='creation desc'):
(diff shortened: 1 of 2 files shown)


erpnext/patches/v13_0/update_issue_metrics.py
```diff
@@ -0,0 +1,27 @@
...
+import frappe
+
+from frappe.core.doctype.communication.communication import set_avg_response_time
+from erpnext.support.doctype.issue.issue import set_resolution_time, set_user_resolution_time
+
+def execute():
+	if frappe.db.exists('DocType', 'Issue'):
+		frappe.reload_doctype('Issue')
+
+		for parent in frappe.get_all('Issue', order_by='creation desc', limit=500):
```
nabinhait: Why limit 500? Maybe you can limit records using a specific date.

nabinhait: If there are no limit, you should commit after every 100 records.

nabinhait (changes requested):

nabinhait (approved):

PR #22302, opened 2020-06-17
fix: Set Value of wrong Bank Account Field in Payment Entry

P-Froggy:
Company bank account was wrongly inserted into the field "Party Bank Account" in payment entry, instead of "Bank Account".
Also changes the label of "Default Bank Account" to "Default Company Bank Account", like suggested in PR #20632

#### Please Note
This is kind of a revert of PR #19390. However, it seems like the function "get_party_bank_account" in Bank Account doctype changed how it worked since it is clearly pulling the default company bank account.


marination:
Left a review on https://github.com/frappe/erpnext/pull/22303

deepeshgarg007 (approved):

PR #22303, opened 2020-06-18
fix: Set Value of wrong Bank Account Field in Payment Entry

P-Froggy:
Company bank account was wrongly inserted into the field "Party Bank Account" in payment entry, instead of "Bank Account".

Adds filters for "Party Bank Account" based on selected Party (I forgot to create a PR for version-12 back in February, see PR #20653).

Also changes the label of "Default Bank Account" to "Default Company Bank Account", like suggested in PR #20632

#### Please Note
This is kind of a revert of PR #19390. However, it seems like the function "get_party_bank_account" in Bank Account doctype changed how it worked since it is clearly pulling the default company bank account.

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.js
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.js
@@ -25,7 +25,7 @@ frappe.ui.form.on('Payment Entry', {
 		});
 		frm.set_query("party_type", function() {
 			return{
-				"filters": {
+				filters: {
 					"name": ["in", Object.keys(frappe.boot.party_account_types)],
 				}
 			}
@@ -33,14 +33,16 @@ frappe.ui.form.on('Payment Entry', {
 		frm.set_query("party_bank_account", function() {
 			return {
 				filters: {
-					"is_company_account":0
+					is_company_account: 0,
+					party_type: frm.doc.party_type,
+					party: frm.doc.party
 				}
 			}
 		});
 		frm.set_query("bank_account", function() {
 			return {
 				filters: {
-					"is_company_account":1
+					is_company_account: 1
 				}
 			}
 		});
@@ -326,7 +328,7 @@ frappe.ui.form.on('Payment Entry', {
 							() => {
 								frm.set_party_account_based_on_party = false;
 								if (r.message.bank_account) {
-									frm.set_value("party_bank_account", r.message.bank_account);
+									frm.set_value("bank_account", r.message.bank_account);
 								}
 							}
 						]);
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -326,7 +328,7 @@ frappe.ui.form.on('Payment Entry', {
 							() => {
 								frm.set_party_account_based_on_party = false;
 								if (r.message.bank_account) {
-									frm.set_value("party_bank_account", r.message.bank_account);
+									frm.set_value("bank_account", r.message.bank_account);
```
marination, line 331: The `get_party_details` method in Payment Entry calls `get_party_bank_account` from `bank_account.py`. There it simply fetches the **default bank account** from the Party mentioned in the Payment Entry. (Check `get_party_bank_account` in `bank_account.py`)
From what i see, the earlier version seems apt.

erpnext/buying/doctype/supplier/supplier.json
```diff
@@ -97,7 +97,7 @@
   {
    "fieldname": "default_bank_account",
    "fieldtype": "Link",
-   "label": "Default Bank Account",
+   "label": "Default Company Bank Account",
```
marination, line 100: Default Bank Account seems appropriate too. A party is not necessarily bound to a company. Also unnecessarily long label

erpnext/accounts/doctype/payment_entry/payment_entry.js
P-Froggy: The current implementation in the supplier doctype only allows to link a company bank account:
https://github.com/frappe/erpnext/blob/3c25308b27458413dafaf3094247ec2b1f326d95/erpnext/buying/doctype/supplier/supplier.js#L20-L26

So in other words, a company bank account must be linked in the supplier master, but in Payment Entry it fetches this same bank account for the "party bank account" field.

Maybe I'm missing something, but from what I see one of both has to be wrong (either the query filter in the supplier master or the add_fetch in payment entry). However, @deepeshgarg007 mentioned in https://github.com/frappe/erpnext/pull/20632#pullrequestreview-359493500 that it's supposed to be the company bank account.

erpnext/buying/doctype/supplier/supplier.json
P-Froggy: Based on the same review https://github.com/frappe/erpnext/pull/20632#pullrequestreview-359493500 this is the same label as for the "default bank account" field in the customer doctype, so it would be more consistent if both fields have the same label:
https://github.com/frappe/erpnext/blob/3c25308b27458413dafaf3094247ec2b1f326d95/erpnext/selling/doctype/customer/customer.json#L125-L130

deepeshgarg007 (approved):

PR #22314, opened 2020-06-18
fix: enable Lead create by email

anupamvs:
**Issue:** Lead is not showing in **Append to** section in Email account.

**Fix:**  Enabling Allow document creation via Email in Lead. So it lists out in the Append to section in Email account so that direct Lead is created from the Email account

image


anupamvs:

Anurag810 (approved):

PR #22328, opened 2020-06-18
fix(pricing_rules): pricing rule selector is wrong

pipech:
Pricing rule with different `apply_on` won't work properly.

Example:
- Rule A: Discount 10% on All Product Group [Priority 1]

image

- Rule B: Discount 20% on Brand A [Priority 20]

image

These rules won't work properly because `get_pricing_rule` function will stop getting rule as soon as it found matching rule but there are a case like these one that need to get rule across ['item_code', 'item_group', 'brand'].

So what we do here is we get all rules across ['item_code', 'item_group', 'brand'] then remove duplicate rules if there are one.


rohitwaghchaure:
@pipech travis is failing, kindly fix the failed test case

erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -30,13 +30,19 @@ class MultiplePricingRuleConflict(frappe.ValidationError): pass
...
-		pricing_rules.extend(_get_pricing_rules(apply_on, args, values))
-		if pricing_rules and not apply_multiple_pricing_rules(pricing_rules):
-			break
+		pricing_rules_all.extend(_get_pricing_rules(apply_on, args, values))
+
+	# removing duplicate pricing rule
+	pricing_rules_title = []
+	pricing_rules = []
+	for p in pricing_rules_all:
+		if p['title'] not in pricing_rules_title:
```
rohitwaghchaure: use "name"(primary key) instead of "title"

pipech: I've change `title` to `name`

https://github.com/frappe/erpnext/pull/22328/commits/303639db764672ba9347475419c347aad6d06e02

marination:
@pipech Please resolve conflicts

barredterra:
@pipech still want to get this merged?

@rohitwaghchaure can we remove the don't merge label?

rohitwaghchaure (approved):

PR #22330, opened 2020-06-18
fix(pricing_rule): apply_on logic dont get cleanup

pipech:
`apply_on` logic field doesn't got clean-up properly because apply_on child table has different name from field it self.

Solution is to call `apply_on_dict` before set field value to None to get correct field name.

Step to replicate:
 - Select Apply On `Item Code` then add item to `Apply Rule On Item Code` child table
 - Select Apply On `Brand` then add brand to `Apply Rule On Brand` child table
 - Save document
 - Select Apply On `Item Code`, item will still be on `Apply Rule On Item Code` child table

Expected behavior is:
 - After save, `Apply Rule On Item Code` child table will be clear

image


rohitwaghchaure (approved):

rohitwaghchaure:
@Mergifyio backport develop

PR #22345, opened 2020-06-19
fix(pricing_rule): clear discount amount

pipech:
Discount amount on previous selected item won't get clear when new item have 0% discount pricing rule.

**Step to replicate:**
(We'll use discount on Item and discount on brand since discount on all item and brand won't work according to #22328)
- Create pricing rule A and B
- Go to Sales Invoice
- Select Item A (Which have price discount 10%)
- Change Item B (Which have price discount 0%)
- Item B will get discount amount from Item A

**Expected result:**
When changing from item A to item B discount will get clear according to pricing rule of Item B

**Use case:**
Create promotion which have discount 10% on all item exclude brand A

Pricing rule A (Discount 10% on Item A):
image

Pricing rule B (Discount 10% on Brand 001):
image

### DIFF
--- a/erpnext/public/js/controllers/taxes_and_totals.js
+++ b/erpnext/public/js/controllers/taxes_and_totals.js
@@ -18,8 +18,12 @@ erpnext.taxes_and_totals = erpnext.payments.extend({
 
 		item.rate = flt(item.rate_with_margin , precision("rate", item));
 
-		if(item.discount_percentage){
+		if (item.discount_percentage) {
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@pipech 

image
If discount percentage is not set and user set discount amount then system apply the discount. With your PR the discount amount will be zero if discount percentage is zero.

pipech:
@rohitwaghchaure 

Discount amount will be zero if there is item pricing rule and discount percentage is zero.
Isn't that suppose to be the case ?

If item doesn't have matching pricing rule which set discount percentage to zero then discount amount won't get clear.

PR #22346, opened 2020-06-19
fix: Set default reference Id for "On Previous Row Amount" and "On Previous Row Total"

AfshanKhan:
- Added client side validation for "On Previous Row Amount" and "On Previous Row Total"
- On selecting "On Previous Row Amount" or "On Previous Row Total" in taxes table an error used to pop up in console which broke the script if reference row is not selected. And many times user is not aware that reference row field exists. So by default use previous row reference id 

image


deepeshgarg007 (approved):

deepeshgarg007:

PR #22349, opened 2020-06-19
fix: handle custom statuses for the pause SLA configuration

ruchamahabal:
Fixes for two issues:

- The response and resolution variables are unset on statuses configured for Hold and reset when the issue changes from a status configured for Hold to "Open" by adding the hold time. However, it should do so for any status which is not configured for hold otherwise the SLA fails for custom statuses as response and resolution variance fields are unset.
fix for code line: https://github.com/frappe/erpnext/blob/develop/erpnext/support/doctype/issue/issue.py#L104

- The total hold time field should only be set if on_hold_since is captured and hold configurations are enabled in the SLA document.


nabinhait (approved):

ruchamahabal:

PR #21685, opened 2020-05-11
fix: make queries show searchfields

hrwX:
- Standard and Custom queries did not show searchfields for Link fields.

version 12: #21760

PR #22357, opened 2020-06-20
fix: removing unnecessary else condition on customer form

khushalti:

> Screenshots/GIFs


erpnext/selling/doctype/customer/customer.py
```diff
@@ -339,16 +336,10 @@ def get_loyalty_programs(doc):
...
 def get_customer_list(doctype, txt, searchfield, start, page_len, filters=None):
-	from erpnext.controllers.queries import get_fields
-
+	fields = ["name", "customer_name", "customer_group", "territory"]
 	if frappe.db.get_default("cust_master_name") == "Customer Name":
 		fields = ["name", "customer_group", "territory"]
-	else:
-		fields = ["name", "customer_name", "customer_group", "territory"]
-
-	fields = get_fields("Customer", fields)
```
rohitwaghchaure, line 351: Why removed this? this was recently added https://github.com/frappe/erpnext/pull/21685

erpnext/selling/doctype/customer/customer.py
```diff
@@ -382,45 +370,10 @@ def check_credit_limit(customer, company, ignore_outstanding_sales_order=False,
 			.format(customer, customer_outstanding, credit_limit))
 
 		# If not authorized person raise exception
-		credit_controller_role = frappe.db.get_single_value('Accounts Settings', 'credit_controller')
```
rohitwaghchaure, line 385: Why removed send_emails code? email should be send to the concerned person?

PR #22366, opened 2020-06-21
fix(Immutable Ledger): Item wise backdated stock entry posting

deepeshgarg007:
Currently System won't allow backdated entry for any Item before the last stock transaction.

After this fix Item Wise Backdated will be checked

For Eg: If Stock Transaction is posted for Item A is made at 21 June 2020
Then Stock Transaction for Item B can still be made at 20 June 2020

### DIFF
--- a/erpnext/utilities/transaction_base.py
+++ b/erpnext/utilities/transaction_base.py
@@ -161,19 +161,22 @@ class TransactionBase(StatusUpdater):
 			if not (self.get("update_stock") or self.get("is_pos")):
 				return
 
-		fiscal_year = get_fiscal_year(self.get('posting_date'), as_dict=True).name
+		for item in self.get('items'):
+			fiscal_year = get_fiscal_year(self.get('posting_date'), as_dict=True).name
 
-		last_transaction_time = frappe.db.sql("""
-			select MAX(timestamp(posting_date, posting_time)) as posting_time
-			from `tabStock Ledger Entry`
-			where docstatus = 1""")[0][0]
+			last_transaction_time = frappe.db.sql("""
+				select MAX(timestamp(posting_date, posting_time)) as posting_time
+				from `tabStock Ledger Entry`
+				where docstatus = 1 and item_code = %s """, (item.item_code))[0][0]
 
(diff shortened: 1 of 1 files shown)


erpnext/utilities/transaction_base.py
```diff
@@ -161,19 +161,22 @@ def validate_with_last_transaction_posting_time(self):
...
-			where docstatus = 1""")[0][0]
-
-		cur_doc_posting_datetime = "%s %s" % (self.posting_date, self.get("posting_time") or "00:00:00")
-
-		if last_transaction_time and get_datetime(cur_doc_posting_datetime) < get_datetime(last_transaction_time):
-			frappe.throw(_("""Posting timestamp of current transaction
-				must be after last Stock transaction's timestamp which is {0}""").format(frappe.bold(last_transaction_time)),
-				title=_("Backdated Stock Entry"))
+		for item in self.get('items'):
+			fiscal_year = get_fiscal_year(self.get('posting_date'), as_dict=True).name
```
szufisher: this line can be before the `for item in self.get('items')`

erpnext/utilities/transaction_base.py
```diff
@@ -161,19 +161,20 @@ def validate_with_last_transaction_posting_time(self):
...
+		for item in self.get('items'):
+			last_transaction_time = frappe.db.sql("""
+				select MAX(timestamp(posting_date, posting_time)) as posting_time
+				from `tabStock Ledger Entry`
+				where docstatus = 1 and item_code = %s """, (item.item_code))[0][0]
+
+			cur_doc_posting_datetime = "%s %s" % (self.posting_date, self.get("posting_time") or "00:00:00")
+
+			if last_transaction_time and get_datetime(cur_doc_posting_datetime) < get_datetime(last_transaction_time):
+				frappe.throw(_("""Last Stock Transaction for item {0} was on {1}.
```
nabinhait: The multiline message creates problem in translations. Please use `<br>` tag to create a multiline message.

nabinhait (changes requested):

nabinhait (approved):

sammish:
@nabinhait @deepeshgarg007 when can we expect the release in version-13

Please update very urgent

PR #22388, opened 2020-06-22
fix: Login page doesn't load for new sites

akurungadam:
Login page broken as active domains are not set for new sites

### DIFF
--- a/erpnext/templates/includes/footer/footer_powered.html
+++ b/erpnext/templates/includes/footer/footer_powered.html
@@ -12,6 +12,7 @@
 } %}
 
 {% set link = '' %}
+{% set label = '' %}
 {% if domains %}
 	{% set label = domains[0].domain %}
 	{% set link = links[label] %}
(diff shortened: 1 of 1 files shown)


erpnext/templates/includes/footer/footer_powered.html
```diff
@@ -12,8 +12,8 @@
 } %}
 
 {% set link = '' %}
-{% set label = domains[0].domain %}
```
nabinhait, line 15: `let label = null;`

Otherwise, it will give error on line 20

nabinhait (approved):

erpnext/templates/includes/footer/footer_powered.html
nabinhait, line 15: let label = null;

Otherwise, it will give error on line 20

nabinhait (changes requested):

Anurag810:
@nabinhait #22394. both PR are for same issue.

nabinhait (approved):

revant:
https://travis-ci.com/github/frappe/frappe_docker/jobs/353812594#L827

Added a test to check index page of new site created using develop branch

PR #22403, opened 2020-06-23
fix: Insert Supplier Group via List View

marination:
**Issue:**
- While trying to Insert a Supplier Group via List View, the form was **Read Only** , even though a root node already existed. Every attempt to create an entry was taken as creation of root node.
 Screenshot 2020-06-23 at 4 53 38 PM

**Fix:**
- Since every Tree already has a Root Node instantiated, Check if current form has parent and if it is a new form.
- If it is not a new form(already saved without parent), it probably is the root node. Else, allow editing.
- Added filter in Parent Link field to avoid selecting Itself as parent
- Changes applied on Customer Group, Item group, Sales Person & Territory.


marination:

nabinhait (approved):

PR #22415, opened 2020-06-24
fix: Create salary slip button disappears

Anurag810:
Merge after:  https://github.com/frappe/frappe/pull/10801
Issue: Create salary Slip button disappear.


erpnext/payroll/doctype/payroll_entry/payroll_entry.js
```diff
@@ -49,13 +50,14 @@ frappe.ui.form.on('Payroll Entry', {
...
 			method: 'fill_employee_details',
-			callback: function(r) {
-				if (r.docs[0].employees){
-					frm.save();
-					frm.refresh();
-					if(r.docs[0].validate_attendance){
-						render_employee_attendance(frm, r.message);
-					}
+		}).then(r => {
+			if (r.docs[0].employees){
```
rohitwaghchaure: what if r.docs itself is blank?

rohitwaghchaure (approved):

PR #22420, opened 2020-06-24
fix: Taxes not getting fetched from Opportunity to Quotation

deepeshgarg007:
While creating a Quotation from Opportunity only the 'Tax Template' is fetched and the taxes in the "Taxes and Charges" table are not fetched and the user has to reselect the template to fetch the taxes.

Null object is returned if tax template is already assigned but taxes are not because of which only Tax template gets assigned and not the taxes.

Screenshot 2020-06-24 at 5 41 16 PM


deepeshgarg007:

erpnext/controllers/accounts_controller.py
```diff
@@ -883,7 +883,10 @@ def get_default_taxes_and_charges(master_doctype, tax_template=None, company=Non
 	if tax_template and company:
 		tax_template_company = frappe.db.get_value(master_doctype, tax_template, "company")
 		if tax_template_company == company:
-			return
+			return {
+				'taxes_and_charges': tax_template,
+				'taxes': get_taxes_and_charges(master_doctype, tax_template)
+			}
```
marination, line 889: the first time `get_default_taxes_and_charges` is called in `make_quotation` in opportunity.py. There due to no template it goes on line 891 directly and gets default template and taxes too.
The second time around it need not fetch it again, hence return works fine.
Perhaps check why it wasnt set or if it wasnt fetched due to another confuguration. Works fine for me

PR #22422, opened 2020-06-24
fix: fetch project-related info in Timesheet (v12)

Alchez:
**Changes:**

- On creating a new Timesheet from Task, the project and worked hours are pulled in the document as well
- Remove redundant "View" buttons for Timesheet and Expense Claim since the dashboard is already available
- Move `add_fetch` behaviour into JSON schema


marination (approved):

barredterra:
Did this go into develop as well? If not, we should port it!

marination:
@barredterra It did via https://github.com/frappe/erpnext/pull/22423

PR #22423, opened 2020-06-24
fix: fetch project-related info in Timesheet (develop)

Alchez:
**Changes:**

- On creating a new Timesheet from Task, the project and worked hours are pulled in the document as well
- Remove redundant "View" buttons for Timesheet and Expense Claim since the dashboard is already available
- Move `add_fetch` behaviour into JSON schema


erpnext/projects/doctype/task/task.py
```diff
@@ -219,6 +220,26 @@ def set_tasks_as_overdue():
...
+			"completed": source.status == "Completed",
+			"project": source.project,
+			"task": source.name
+		})
+
+	doclist = get_mapped_doc("Task", source_name, {
+			"Task": {
+				"doctype": "Timesheet"
+			}
+		}, target_doc, postprocess=set_missing_values, ignore_permissions=ignore_permissions)
```
marination, line 239: why ignore permissions ?

marination:
Please resolve conflicts

erpnext/projects/doctype/task/task.py
Alchez: I'm just passing whatever the function gets, this will be `False` by default

marination (dismissed):

marination (approved):

PR #22426, opened 2020-06-24
fix: Pick List empty table and Serial-Batch items handling

marination:
**Issues:**
1. On updating submitted Pick List it allowed for the locations table to be empty. Result, a bunch of empty Pick Lists.
2. No provision to pick items that are both serialised and batched
3. Bug due to https://github.com/frappe/erpnext/pull/21781 that set faulty serial nos in rows in case a single item was split into multiple rows (different warehouses). 

**Fixes:**
<details><summary>Empty Pick List</summary>

_Added indicators to show availability of stock. If no stock on updation after submission, table does not become empty. Instead Red indicators and a message feedback to the users will be shown._
 Screenshot 2020-06-24 at 6 35 03 PM
 
</details>

<details><summary>Serialised and Batched Items</summary>

_Items that are Serialised and Batched can be picked too._
batch-serial-pick-list

</details>

The last bug is reverted in this PR. Provision for user to set serial or batch nos manually will be worked upon later


marination:
Screenshot 2020-07-09 at 4 59 38 PM

nabinhait (approved):

PR #22428, opened 2020-06-24
feat(projects): added field toggle for removing task assignments (#413)

MyuddinKhatri:
added field to Projects Settings with check field "Remove Assignment on Task Completion"
and performs accordingly for removal of task assignment

We want to run reports on how many tasks a user completed


rohitwaghchaure:
@MyuddinKhatri Please don't send PR related to new Feature on the stable version (version-12-hotfix)

PR #22206, opened 2020-06-11
Stock Ageing TypeError: ‘<’ not supported between instances of ‘int’ …

clarkejj:
…and ‘str’

Re this user report https://discuss.erpnext.com/t/stock-ageing-report-typeerror-not-supported-between-instances-of-int-and-str/62605

Notes: 

ASSUMPTION:  This error arises in current version-12 production?

WARNING: This proposed change has not been tested

> Please provide enough information so that others can review your pull request:

> Explain the **details** for making this change. What existing problem does the pull request solve?

> Screenshots/GIFs

PR #22430, opened 2020-06-24
fix: stock_ageing.py report SyntaxError: can’t assign to function call

clarkejj:
Following this Stock Ageing report error reported in version-12
https://discuss.erpnext.com/t/stock-ageing-report-typeerror-not-supported-between-instances-of-int-and-str/62605/4

This PR below to fix that subsequently introduced this current **can’t assign to function SyntaxError**: https://github.com/frappe/erpnext/pull/22206/files/677978dd00e92533d7897dec041a901c0e3cbba9

This PR should fix the SyntaxError that @marination kindly flagged thanks!


marination (approved):

marination:

PR #22438, opened 2020-06-25
fix: Update route after rename and update doc

SDLyu:

After rename or update item group doc that route not change, it should change to correct route path.

This PR call `make_route` function in `after_rename` `on_update` event and make route change.

> Screenshots/GIFs

### Change `Node4` to `Node5`

Item_Group_Tree

#### Before

_MySQL_5_5_5-10_3_22-MariaDB-1_10_3_22_maria_bionic__docker_erpnext__a94a8fe5ccb19ba6_tabItem_Group

#### After

_MySQL_5_5_5-10_3_22-MariaDB-1_10_3_22_maria_bionic__docker_erpnext__a94a8fe5ccb19ba6_tabItem_Group


erpnext/setup/doctype/item_group/item_group.py
```diff
@@ -37,30 +37,34 @@ def validate(self):
...
+		return self.route
 
 	def on_trash(self):
 		NestedSet.on_trash(self)
 		WebsiteGenerator.on_trash(self)
 		self.delete_child_item_groups_key()
 
+	def after_rename(self, old_name, new_name, merge):
+		self.make_route()
+		self.save()
```
rohitwaghchaure: Instead of self.save() try below code
```
frappe.db.set_value("Item Group", self.item_group, "route", self.route)
```

erpnext/setup/doctype/item_group/item_group.py
```diff
@@ -37,30 +37,34 @@ def validate(self):
 	def on_update(self):
 		NestedSet.on_update(self)
 		invalidate_cache_for(self)
+		self.make_route()
```
rohitwaghchaure: why called make_route from on_update, it's already called from validate

SDLyu: Sorry, didn't notice that. Already remove `make_route` from `on_update`

erpnext/setup/doctype/item_group/item_group.py
SDLyu: Done. But no `item_group` attribute found, it should be `item_group_name`.

SDLyu:
Hi @rohitwaghchaure 
Is there any update?
Thanks.

erpnext/setup/doctype/item_group/item_group.py
```diff
@@ -43,24 +43,27 @@ def on_update(self):
 
 	def make_route(self):
 		'''Make website route'''
-		if not self.route:
-			self.route = ''
-			if self.parent_item_group:
-				parent_item_group = frappe.get_doc('Item Group', self.parent_item_group)
+		self.route = ''
+		if self.parent_item_group:
+			parent_item_group = frappe.get_doc('Item Group', self.parent_item_group)
```
rohitwaghchaure: Why get_doc? you have to get only 2 field use 
frappe.db.get_value("Item Group", self.parent_item_group, ["parent_item_group", "route"])

SDLyu: Fixed

PR #22440, opened 2020-06-25
fix: cannot cancel assets with repair pending

nextchamp-saqib:
Problem:
Assets having pending repair documents or pending maintenance documents shows inappropriate error message

Changes:
Error message now suggests, If an asset has 'In Maintenance' or 'Out of Order' statuses, then user must complete/cancel all the active maintenance or repairs before cancelling the asset.


nabinhait (approved):

nextchamp-saqib:

PR #22443, opened 2020-06-25
fix: show_dashboard + button depends on status

Anurag810:
In Expense Claim doctype when the status is Paid, payment Entry and Journal Entry links (In dashboard) should be disabled.

image


nabinhait (commented):
This is an anti-pattern, not used in any document.

Keep the + button, but add validation if expense claim is overpaid or payment made against rejected claim.

PR #22446, opened 2020-06-25
fix: removed fiscal year field from "Request for Quotation" doctype

AfshanKhan:
Removed"Fiscal Year" field from "Request for Quotation" doctype and checked for reports that would have ref Doctype as "Request for Quotation". There weren't any.


deepeshgarg007:
@AfshanKhan Please update proper description. Since our pull requests are public you should write description that makes sense to everyone, even outside our organization

deepeshgarg007 (dismissed):

deepeshgarg007:
@AfshanKhan Please check travis logs. There is an error in this

AfshanKhan:
@deepeshgarg007 fixed it. Travis is successful.

deepeshgarg007 (approved):

PR #22448, opened 2020-06-25
fix: job applicant fixes 

Anurag810:
fixes:
Don't allow to create multiple job offer for a single job applicant
Don't allow to create Employee onboarding without a job offer


erpnext/hr/doctype/employee_onboarding/employee_onboarding.py
```diff
@@ -13,6 +13,12 @@ class IncompleteTaskError(frappe.ValidationError): pass
 class EmployeeOnboarding(EmployeeBoardingController):
 	def validate(self):
 		super(EmployeeOnboarding, self).validate()
+		self.validate_duplicate_employee_onboarding()
+
+	def validate_duplicate_employee_onboarding(self):
+		emp_onboarding = frappe.db.exists("Employee Onboarding",{"job_applicant": self.job_applicant})
```
nabinhait, line 19: `name!=self.name`

erpnext/hr/doctype/job_offer/job_offer.py
```diff
@@ -15,6 +15,9 @@ def onload(self):
 
 	def validate(self):
 		self.validate_vacancies()
+		job_offer = frappe.db.exists("Job Offer",{"job_applicant": self.job_applicant})
```
nabinhait, line 18: `name!=self.name`

nabinhait (changes requested):

nabinhait (approved):

PR #22450, opened 2020-06-25
fix: Disable Renaming in Serial No

marination:
- Disabled Serial No Renaming as the serial no fields are just text fields in transactions. 
- Renaming does not rename the values in those fields, causing chaotic data.


marination:

rohitwaghchaure (approved):

PR #22452, opened 2020-06-25
fix: Add bold rounded total

AminovE99:
I'm working with Monogramm to develop a New design for print format and noticed that 
rounded total inside `Sales Invoice` is bold field, but inside `Purchase Invoice` - not a bold field.
It gives impact when you have custom `.important` selector inside Print Format:
Print format inside Sales Invoice:
```
...
<div class="row important data-field" data-fieldname="rounded_total" data-fieldtype="Currency">
...
```

Print format inside Purchase Invoice:
```
...
<div class="row data-field" data-fieldname="rounded_total" data-fieldtype="Currency">
...
```
image

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
@@ -3232,7 +3232,7 @@
    "allow_bulk_edit": 0,
    "allow_in_quick_entry": 0,
    "allow_on_submit": 0,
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@AminovE99 Enable bold property using customize form

AminovE99:
@rohitwaghchaure Anyway question is still open: Why inside `Sales Invoice` and `Purchase Invoice` we have different fields `rounded_total`?

PR #22454, opened 2020-06-25
refactor: update customs_tariff_number.json

wojosc:
Main information of tariff numbers are the number. A description is more or less double information as this is mostly the number. Thus a description should not be mandetory.


rohitwaghchaure:
@wojosc You need to change the modified date as well otherwise your changes will not applied.

wojosc:
> @wojosc You need to change the modified date as well otherwise your changes will not applied.

Thank you for the reminder. I added that change.

rohitwaghchaure (approved):

PR #22456, opened 2020-06-25
fix: Payment amount in payment terms showing negative amount

deepeshgarg007:
Payment Amount in Payment Terms is shown as negative if advance is allocated against a multicurrency Purchase Invoice

Befor:
Screenshot 2020-06-25 at 5 59 47 PM

After:
Screenshot 2020-06-25 at 6 10 59 PM


deepeshgarg007:

nabinhait (changes requested):
```
if party_account_currency == self.company_currency:
	grand_total = self.get("base_rounded_total") or self.base_grand_total
else:
	grand_total = self.get("rounded_total") or self.grand_total
```

Do the changes accordingly in all places

PR #22462, opened 2020-06-25
fix: Handling Empty tables in Production Plan

marination:
**Issues:**
- For an empty (without even an empty row) SO/MR table in Production Plan:
   `so_list = [d.sales_order for d in self.sales_orders if d.sales_order]`
    `AttributeError: 'ProductionPlan' object has no attribute 'sales_orders'`
- While Getting Items for Material Request Planning, it threw the following on **Items to Manufacture** table with **empty row**:
 Screenshot 2020-06-25 at 9 35 04 PM
- Also on old line 638, `doc.get('items')` does not exist

**Fixes:**
- For both issues check if table is empty (not instantiated) and also if table rows are empty
- Message for empty SO/MR table (with no rows at all)
  Screenshot 2020-06-25 at 9 38 42 PM

  Screenshot 2020-06-25 at 9 29 05 PM

- Message for empty **Items to Manufacture** table with empty row and no rows
  Screenshot 2020-06-25 at 9 29 20 PM

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -98,11 +98,17 @@ class ProductionPlan(Document):
 		elif self.get_items_from == "Material Request":
 			self.get_mr_items()
 
+	def get_so_mr_list(self, field, table):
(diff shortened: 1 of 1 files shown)


marination:

rohitwaghchaure (approved):

PR #22465, opened 2020-06-26
fix: add cost center filter in MR item

michellealva:
Added cost center filter in Material Request Item table.

**Before:**

All cost centers of all companies including root nodes were shown.

Screenshot 2020-06-26 at 8 59 24 AM

**After:**

Filter on selected company and child nodes.

 
Screenshot 2020-06-26 at 8 59 55 AM


michellealva:

PR #22467, opened 2020-06-26
fix: Send email to credit controllers to increase credit limit

nabinhait:
While sending emails to the users with role credit controller, system checks for Employee record which was not needed, removed it.


nabinhait:

rohitwaghchaure (approved):

PR #22468, opened 2020-06-26
fix: enable show_configure_button when shopping cart is enabled

anupamvs:
'Show Configure Button' should be auto-enabled when 'Shopping Cart' is enabled as it is not very obvious that this is important when it comes to Item Variants.
image


anupamvs:

rohitwaghchaure (approved):

PR #22469, opened 2020-06-26
fix: Handling Empty tables in Production Plan

marination:
Port of https://github.com/frappe/erpnext/pull/22462


marination:
Screenshot 2020-06-26 at 2 09 46 PM

PR #22470, opened 2020-06-26
fix: Sales Order to MR test failing

marination:
Fix from https://github.com/frappe/erpnext/pull/22469


marination:
test on develop
Screenshot 2020-06-26 at 2 13 59 PM

rohitwaghchaure (approved):

PR #22472, opened 2020-06-26
fix: apply filters in all transactions

michellealva:
Apply cost center and expense account filter in all transactions.

For eg, Added cost center filter and expense account filter in Material Request Item table which shows dropdown of accounts and cost centers belonging to the selected company.

**Before:**

All cost centers of all companies including root nodes were shown. Same for Expense account (accounts of all companies shown).

image

image

**After:**

Filter on selected company and child nodes.

image

image


michellealva:

PR #22481, opened 2020-06-27
fix: change button label in issue

michellealva:
Changed the label of the button from "Make" to "Create" so that it is consistent with other DocTypes.

**Before:**

image

**After:**

image


michellealva:

deepeshgarg007 (approved):

PR #22497, opened 2020-06-29
fix: Set Root as Parent if no parent in new tree view node

marination:
**Issue:**
- No parent even after save if no parent is set
  image

**Fix:**
- While creating a new Customer Group/ Territory / Sales Person from List View, if no Parent is filled, fetch root node as parent, just like Supplier Group


marination:

marination:

nabinhait (approved):

PR #22498, opened 2020-06-29
feat: Pending quantity report for Material Request

AfshanKhan:

> Screenshots/GIFs

Currently:
When a Material Request is raised, there is no option or field to check the pending quantity, % received and Quantity received is present but no direct column where the quantity pending is seen or any report.

Expected:
Screenshot 2020-06-29 at 5 11 51 PM


marination:
@AfshanKhan @nabinhait can't we include received qty information in **Requested Items to Order** Report and maybe rename that report too?
 We can get complete stock analysis of MR in one report.

PR #22502, opened 2020-06-29
fix: Added Missing Trends Reports to Buying/Selling Desk

marination:
- Added **Delivery Note** and **Sales Invoice Trends** to Selliing **Key Reports**
- Added **Purchase Receipt** and **Purchase Invoice Trends** to Buying **Key Reports**


marination:

nabinhait:
@marination Why invoice or stock trend reports should be key reports for the buying /selling module? The target user of those reports is stock/account users. But still, we can add those links under other reports, not under key reports.

marination:
@nabinhait done
- Buying
  Screenshot 2020-06-29 at 7 31 04 PM

- Selling
  Screenshot 2020-06-29 at 7 28 28 PM

nabinhait (approved):

PR #22511, opened 2020-06-29
fix: Do not ignore validate methods on GL entry submit

deepeshgarg007:
Do not ignore validations as no `before_save` methods written in Server Script or hooks are executed if validations are ignored


deepeshgarg007:

rohitwaghchaure (approved):

PR #22513, opened 2020-06-29
fix: Allow PR/DN without PO/SO if set in Supplier/Customer Master

P-Froggy:
#### Description
Allows creation of Purchase Receipt without Purchase Order if "Allow Purchase Invoice Creation Without Purchase Order" in the Supplier Master is ticked.
The same also applies for the corresponding option in Customer Master.

#### Details
PR #20864 added the option to overwrite the default behaviour of requiring a Purchase Order or a Sales Order for the creation of an Invoice based on the Supplier/Customer. However, this new logic has only been implemented in Purchase/Sales Invoice, but not in Purchase Receipt and Delivery Note.
Because the procurement process in ERPNext is _Purchase Order -> Purchase Receipt -> Purchase Invoice_ this logic should also apply for Purchase Receipt and Delivery Note.

#### Proposal
I know this is not a discussion forum, but I thought of doing a cleanup of the Buying and Selling Settings. The addition of a per supplier and per customer based requisition of PO, PR, SO and DN made the options in Buying and Selling Settings redundant. These option fields could easily be removed and instead, only the fields in Supplier and Customer Master could be used. This way you only need to check one field instead of two during creation of PR, PI, DN and SI.
Maybe the reviewer of this PR could share his/her opinion on this.

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.py
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.py
@@ -101,6 +101,9 @@ class DeliveryNote(SellingController):
 	def so_required(self):
 		"""check in manage account if sales order required or not"""
 		if frappe.db.get_value("Selling Settings", None, 'so_required') == 'Yes':
+			if frappe.get_value('Customer', self.customer, 'so_required'):
+				return
+
 			for d in self.get('items'):
 				if not d.against_sales_order:
 					frappe.throw(_("Sales Order required for Item {0}").format(d.item_code))
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -119,6 +119,9 @@ class PurchaseReceipt(BuyingController):
 
 	def po_required(self):
 		if frappe.db.get_value("Buying Settings", None, "po_required") == 'Yes':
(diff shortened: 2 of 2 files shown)


marination:
@P-Froggy Left a review on https://github.com/frappe/erpnext/pull/22514. Your point of concern is valid we will tweak the options to incorporate Purchase Receipt and Delivery Note too. Thanks !!!

> I know this is not a discussion forum, but I thought of doing a cleanup of the Buying and Selling Settings. The addition of a per supplier and per customer based requisition of PO, PR, SO and DN made the options in Buying and Selling Settings redundant. These option fields could easily be removed and instead, only the fields in Supplier and Customer Master could be used. This way you only need to check one field instead of two during creation of PR, PI, DN and SI.
Maybe the reviewer of this PR could share his/her opinion on this.

Regarding this, the use case as i currently see it is that, a majority of the users function via Buying/Selling settings itself and the options in it. The additional checkboxes in Customer / Supplier only affect that particular entity, but globally options in the Settings are applied.
It would be rather better for us to set the option once globally ("PO Required") and then based on certain suppliers or customers check "Allow Purchase Invoice Creation Without Purchase Order" only for that entity. So that I don't have to set the option for each and every supplier/customer.
However if your entire business process requires bypassing Purchase Order , you can again set `po_required` to **No** in Buying Settings.

PR #22514, opened 2020-06-29
fix: Allow PR/DN without PO/SO if set in Supplier/Customer Master

P-Froggy:
#### Description
Allows creation of Purchase Receipt without Purchase Order if "Allow Purchase Invoice Creation Without Purchase Order" in the Supplier Master is ticked.
The same also applies for the corresponding option in Customer Master.

#### Details
PR #20864 added the option to overwrite the default behaviour of requiring a Purchase Order or a Sales Order for the creation of an Invoice based on the Supplier/Customer. However, this new logic has only been implemented in Purchase/Sales Invoice, but not in Purchase Receipt and Delivery Note.
Because the procurement process in ERPNext is _Purchase Order -> Purchase Receipt -> Purchase Invoice_ this logic should also apply for Purchase Receipt and Delivery Note.

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -118,6 +118,9 @@ class PurchaseReceipt(BuyingController):
 
 	def po_required(self):
 		if frappe.db.get_value("Buying Settings", None, "po_required") == 'Yes':
+			if frappe.get_value('Supplier', self.supplier, 'allow_purchase_invoice_creation_without_purchase_order'):
+				return
+				
 			for d in self.get('items'):
 				if not d.purchase_order:
 					frappe.throw(_("Purchase Order number required for Item {0}").format(d.item_code))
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -118,6 +118,9 @@ def validate_with_previous_doc(self):
 
 	def po_required(self):
 		if frappe.db.get_value("Buying Settings", None, "po_required") == 'Yes':
+			if frappe.get_value('Supplier', self.supplier, 'allow_purchase_invoice_creation_without_purchase_order'):
```
marination, line 121: the checkbox clearly states that it affects Purchase Invoice, so, we only check this on Purchase Invoice or Sales Invoice validation. This just adds more unexplained behaviour which is unnecessary. Let's make the checkboxes function as per their labels.

marination:
@P-Froggy as mentioned in the review, the checkbox _Allow Purchase Invoice Creation Without Purchase Order_ should affect only Purchase Invoice as mentioned in it's label.
Maybe we can add additional settings for PR and DN without PO and SO respectively too. We will definitely get this done.
But using the currently available checkboxes to check PR or DN is not a good idea

marination:
Closing this

PR #22515, opened 2020-06-30
feat: Cost Center based Trend Reports

P-Froggy:
#### Description
Adds the option to base purchase and sales trend reports also on cost center.

image

(PS: Would be nice if you could do an automatic backport, since this isn't a big feature and for me personally it's quite useful)

### DIFF
--- a/erpnext/controllers/trends.py
+++ b/erpnext/controllers/trends.py
@@ -273,6 +273,12 @@ def based_wise_columns_query(based_on, trans):
 		else:
 			frappe.throw(_("Project-wise data is not available for Quotation"))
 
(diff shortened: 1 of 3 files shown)


rohitwaghchaure:
@P-Froggy cost center field is not available in sales order / purchase order, So while checking Sales Order Trends report getting below error

Screenshot 2020-07-06 at 4 06 26 PM

Also in sales invoice first row is showing as blank

Screenshot 2020-07-06 at 4 11 05 PM

Closing the PR, kindly check all this issues and fix it. After fixing all the issues reopen the PR again

PR #22522, opened 2020-06-30
fix: set half day date None if half day is unchecked

Anurag810:


erpnext/hr/doctype/leave_application/leave_application.js
```diff
@@ -40,6 +40,8 @@ frappe.ui.form.on("Leave Application", {
 	validate: function(frm) {
 		if (frm.doc.from_date == frm.doc.to_date && frm.doc.half_day == 1){
 			frm.doc.half_day_date = frm.doc.from_date;
+		}else if (frm.doc.half_day == 0){
+			frm.doc.half_day_date = undefined;
```
rohitwaghchaure: undefined? use ""

rohitwaghchaure (approved):

PR #22525, opened 2020-06-30
fix: only remove the selected Quotation item instead of all rows with the same item code

Alchez:
**Problem:**

While updating an item in the cart, it may be possible that multiple instances of that item might exist in the Quotation. If you remove one instance of this item (set quantity to 0), the system removes all instances of that item code instead.

**Fix:**

Pass the active item row to modify instead

### DIFF
--- a/erpnext/public/js/shopping_cart.js
+++ b/erpnext/public/js/shopping_cart.js
@@ -134,11 +134,12 @@ $.extend(shopping_cart, {
 		}
 	},
 
-	shopping_cart_update: function({item_code, qty, cart_dropdown, additional_notes}) {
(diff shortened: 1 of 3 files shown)


rohitwaghchaure:
@Alchez I'm trying to replicate the issue in my local account, but I'm not able to add same item multiple times in the shopping cart. Can you please share the GIF of the issue, it will be easy for me to replicate the issue in my local account

Alchez:
@rohitwaghchaure, I realized that is possible because of some modifications we did for our personal instance, so this might not be easily replicable for you.

I'll leave the decision for the PR up to you, but if you wanted to test this, you could add the same item multiple times in the connected Quotation, refresh the cart and try removing one of them.

PR #22526, opened 2020-06-30
fix: incorrect stock after merging the items

rohitwaghchaure:
### User has created stock reconciliation with Item A and Item B

Screenshot 2020-06-30 at 11 15 52 AM

### Then user has merged Item B with Item A, after that stock reconciliation record has used the second row and ignored the first

Screenshot 2020-06-30 at 2 47 13 PM

### Stock Balance after rename and merge
Ignored the 10 quantity
Screenshot 2020-06-30 at 2 50 05 PM

### Added validation that system will not merge the items, user has to first fix the stock reconciliation and then they can merge the items.

Screenshot 2020-06-30 at 2 42 04 PM


rohitwaghchaure:

erpnext/stock/doctype/item/item.py
```diff
@@ -656,6 +659,27 @@ def after_rename(self, old_name, new_name, merge):
...
+			FROM `tabStock Reconciliation Item`
+			WHERE item_code = %s and docstatus = 1
+			GROUP By item_code, warehouse, parent
+			HAVING records > 1
+		""", new_name, as_dict=1)
+
+		if not records: return
+		document = "stock reconciliation" if len(records) == 1 else "stock reconciliations"
+
+		msg = _("The item {0} and {1} are present in the following {2} : <br>"
```
marination:
```suggestion
		msg = _("The items {0} and {1} are present in the following {2} : <br>"
```

erpnext/stock/doctype/item/item.py
```diff
@@ -656,6 +659,27 @@ def after_rename(self, old_name, new_name, merge):
...
+			.format(frappe.bold(old_name), frappe.bold(new_name), document))
+
+		msg += ', '.join([get_link_to_form("Stock Reconciliation", d.parent) for d in records]) + "<br><br>"
+
+		msg += _("Note: To merge the items, create separate stock reconciliation for the old item {0}"
+			.format(frappe.bold(old_name)))
+
+		frappe.throw(_(msg), title=_("Can not merge {0} with {1}".format(old_name, new_name)))
```
marination:
```suggestion
		frappe.throw(_(msg), title=_("Merge not allowed".format(old_name, new_name)))
```
Since we are mentioning items in the message itself, lets keep the title to the point and short. Less effort to read

marination (approved):

PR #22533, opened 2020-07-01
Corrected Error in JSON output

finbyz:
JSON.dumps we were applying on Json data, which was putting additional "/" in the output. To avoid this we are using json.loads before using json.dumps.

> Screenshots/GIFs


deepeshgarg007:
Closing this, should be fixed via https://github.com/frappe/erpnext/pull/22656

PR #22536, opened 2020-07-01
fix: UI/UX Salary details and salary slip form cleanup

Anurag810:
UI/UX Salary detail:
Before in salary slip:
image
Now:
new-new
Before in salary Component:
new-new-old
Now:
new-new-2
 Moved some field to different section:
image


erpnext/payroll/doctype/salary_detail/salary_detail.json
```diff
@@ -181,23 +186,12 @@
...
-  {
-   "depends_on": "eval:doc.parenttype=='Salary Structure'",
-   "fieldname": "section_break_11",
-   "fieldtype": "Column Break"
-  },
-  {
-   "depends_on": "eval:doc.parenttype=='Salary Structure'",
-   "fieldname": "condition_and_formula_help",
-   "fieldtype": "HTML",
-   "label": "Condition and Formula Help",
```
rohitwaghchaure, line 193: Why removed Condition and Formula Help?

Anurag810: Because I have made the link for that which shows modal with every variable and formula examples.

ruchamahabal (approved):

PR #22541, opened 2020-07-01
fix: [v12] incorrect available_qty being set

sunhoww:
Duplicate of #22539 for v12.

### DIFF
--- a/erpnext/public/js/utils/serial_no_batch_selector.js
+++ b/erpnext/public/js/utils/serial_no_batch_selector.js
@@ -333,8 +333,8 @@ erpnext.SerialNoBatchSelector = Class.extend({
 							};
 						},
 						change: function () {
-							let val = this.get_value();
-							if (val.length === 0) {
+							const batch_no = this.get_value();
+							if (!batch_no) {
 								this.grid_row.on_grid_fields_dict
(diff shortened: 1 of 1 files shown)


erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -354,14 +354,11 @@ erpnext.SerialNoBatchSelector = Class.extend({
...
 
-							let batch_number = me.item.batch_no ||
-								this.grid_row.on_grid_fields_dict.batch_no.get_value();
-
 							if (me.warehouse_details.name) {
 								frappe.call({
 									method: 'erpnext.stock.doctype.batch.batch.get_batch_qty',
 									args: {
-										batch_no: batch_number,
+										batch_no: val,
```
Anurag810: please change the variable name. it should be self-descriptive.

Anurag810 (changes requested):

erpnext/public/js/utils/serial_no_batch_selector.js
sunhoww: I've updated the variable at its definition itself.

nabinhait (approved):

PR #22542, opened 2020-07-01
feat: set default project and customer if only one found in Timesheet

Alchez:
**Changes:**

- Add a new "Project" option in the "Create Sales Invoice" dialog box in Timesheet
- If only one customer or project needs to be billed, they're automatically fetched into the dialog box, otherwise none are selected

<hr>

**Screenshots / GIFs:**

timesheet-customer-project

### DIFF
--- a/erpnext/projects/doctype/timesheet/timesheet.js
+++ b/erpnext/projects/doctype/timesheet/timesheet.js
@@ -101,12 +101,60 @@ frappe.ui.form.on("Timesheet", {
 			});
 	},
 
-	make_invoice: function(frm) {
+	make_invoice: async function(frm) {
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:
@Alchez This looks specific case, better design maybe you can add Project and Customer field in the timesheet (parent level). For child table, by default system will pick up the project from the parent field if the field is blank.

Alchez:
@rohitwaghchaure I haven't added any fields to Timesheet or the child table, just in the dialog box. Project already exists per row in the child table, I'm just pulling the value.

And it's better to have the Project field in the child table since a user could potentially be working on multiple projects in a given day.

PR #22554, opened 2020-07-02
fix: changing Show Configure Button -> Enable Variants

anupamvs:
**Issue-** 'Show configure button' is not self- explanatory that it is used to enable variants
image

**Fix-**
image


anupamvs:

Anurag810 (changes requested):
image

I think we can split display settings into two columns there is lots of blank space.

erpnext/shopping_cart/doctype/shopping_cart_settings/shopping_cart_settings.js
```diff
@@ -16,8 +16,8 @@ $.extend(cur_frm.cscript, {
 	},
 	enabled: function() {
 		if (cur_frm.doc.enabled === 1) {
-			cur_frm.doc.show_configure_button = 1;
-			cur_frm.refresh_field('show_configure_button');
+			cur_frm.doc.enable_variants = 1;
```
Anurag810: use ```frm.doc``` instead of ```cur_frm```

Anurag810 (changes requested):

PR #22559, opened 2020-07-02
feat: Dunning

KanchanChauhan:
Related to https://discuss.erpnext.com/t/v13-feature-contributions-dunning-and-shipment/62503

**Dunning** is a reminder to the customer or vendor to make the payment on outstanding invoices that are past due dates. Optionally you can charge a fixed fee and/or interest.

**Dunning Type:**
There are Dunning Types or levels, For Example Dunning Type, First Notice will not have any extra amount, and Dunning Type Second Notice will have Rate of Interest and also Dunning fee.

Screen Shot 2020-07-14 at 6 03 56 PM

**Create Dunning:**
Screen Shot 2020-07-14 at 6 13 43 PM

**Dunning**
screencapture-kanchan-ch-8000-desk-2020-07-14-18_16_37

**Dunning Letter Print**
Screen Shot 2020-07-14 at 6 45 27 PM

**Payment Entry**
Screen Shot 2020-07-14 at 7 29 53 PM

Payment Entry creates 2 rows, 1 for Sales Invoice and another for Dunn collected on the Sales Invoice
Screen Shot 2020-07-14 at 7 29 27 PM


barredterra (commented):
Why is **Dunning Type Text Block** a table but `dunning.get_text_block` uses `frappe.get_value`, returning a single result, not the whole table?

The print format still has your company address in it. ;)

KanchanChauhan:
@barredterra We do not need whole table, we just need the text for the selected language in Print Setting. And we cannot just pop it in, it is Jinja hence needs to be rendered. Removed the address, thanks for that :)
Made lots of new changes, you can review again.

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1075,15 +1082,35 @@ def get_payment_entry(dt, dn, party_amount=None, bank_account=None, bank_amount=
...
+			else:	
+				pe.append("references", {
+				 'reference_doctype': dt,
+			 	'reference_name': dn,
+				 "bill_no": doc.get("bill_no"),
+				 "due_date": doc.get("due_date"),
+				 'total_amount': grand_total,
+				 'outstanding_amount': outstanding_amount,
+				 'allocated_amount': outstanding_amount
+				})
```
barredterra: Please consider to fix the indentation here.

barredterra (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.py
KanchanChauhan: Fixed

erpnext/accounts/doctype/dunning/dunning.js
```diff
@@ -0,0 +1,155 @@
...
+			frm.doc.__islocal ? 0 : 1
+		);
+		if (frm.doc.docstatus === 1 && frm.doc.status === "Unresolved") {
+			frm.add_custom_button(__("Resolve"), () => {
+				frm.set_value("status", "Resolved");
+			});
+		}
+		if (frm.doc.docstatus === 0) {
+			frm.trigger("get_dunning_letter_text");
+			frm.trigger("calculate_interest_and_amount");
```
nabinhait: Why is it called on refresh?

erpnext/accounts/doctype/dunning/dunning.js
```diff
@@ -0,0 +1,155 @@
...
+				frm.doc.due_date,
+				"days"
+			);
+			frm.set_value("overdue_days", overdue_days);
+		}
+	},
+	calculate_interest_and_amount: function (frm) {
+		frappe.call({
+			method:
+			"erpnext.accounts.doctype.dunning.dunning.calculate_interest_and_amount",
```
nabinhait: We can avoid this ajax call and do the calculation in js itself.

erpnext/accounts/doctype/dunning/dunning.py
```diff
@@ -0,0 +1,81 @@
...
+        doc = json.loads(doc)
+        if language:
+            filters = {'parent': dunning_type,
+                       'language': language}
+        else:
+            filters = {'parent': dunning_type,
+                       'is_default_language': 1}
+    letter_text = frappe.db.get_value('Dunning Letter Text',
+                                      filters,
+                                      ['body_text', 'closing_text', 'language'], as_dict=1)
```
nabinhait:
```suggestion
    letter_text = frappe.db.get_value('Dunning Letter Text',
        filters, ['body_text', 'closing_text', 'language'], as_dict=1)
```

We generally used this kind of indentation everywhere. Can you please fix this in all places?

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1075,15 +1082,35 @@ def get_payment_entry(dt, dn, party_amount=None, bank_account=None, bank_amount=
...
+				pe.append("references", {
+					'reference_doctype': 'Sales Invoice',
+					'reference_name': doc.get('sales_invoice'),
+					"bill_no": doc.get("bill_no"),
+					"due_date": doc.get("due_date"),
+					'total_amount': doc.get('outstanding_amount'),
+					'outstanding_amount': doc.get('outstanding_amount'),
+					'allocated_amount': doc.get('outstanding_amount')
+				})
+				pe.append("references", {
```
nabinhait, line 1095: As we are receiving payment against Dunning, GL Entries will be created for the dunning amount.
**Debtors** will be credited against Dunning and bank/cash will be debited, which is absolutely ok.

But we did not make GL entries on submission of Dunning, where **Debtors** should be debited, and **income account** should be credited.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1599,3 +1599,30 @@ def create_invoice_discounting(source_name, target_doc=None):
 	})
 
 	return invoice_discounting
+
+@frappe.whitelist()
+def create_dunning(source_name, target_doc=None):
+	invoice = frappe.get_doc("Sales Invoice", source_name)
+	dunning = frappe.new_doc("Dunning")
```
nabinhait: Instead of manually mapping each field, use `mapper` function. That way, all the fields with same name will be mapped automatically. And if a user adds some custom fields on both the document, those will also get mapped.

erpnext/accounts/doctype/dunning/dunning.py
```diff
@@ -0,0 +1,81 @@
...
+def resolve_dunning(doc, state):
+    for reference in doc.references:
+        if reference.reference_doctype == 'Sales Invoice' and reference.outstanding_amount <= 0:
+            dunnings = frappe.get_list('Dunning', filters={
+                                       'sales_invoice': reference.reference_name, 'status': ('!=', 'Resolved')})
+
+            for dunning in dunnings:
+                dunning_doc = frappe.get_doc('Dunning', dunning.name)
+                dunning_doc.status = 'Resolved'
+                dunning_doc.save()
```
nabinhait: We can use `frappe.db.set_value` here

nabinhait (changes requested):

nabinhait (changes requested):
Write a test case for payments against Dunning. Check GL entries and the outstanding amount on the invoice.

Another test to check GL entries on the submission of Dunning.

KanchanChauhan:
GL Entry on creating Dunning and deleting GL entry on Cancellation.

Screen Shot 2020-07-20 at 6 30 03 PM

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1599,3 +1599,37 @@ def create_invoice_discounting(source_name, target_doc=None):
...
+			letter_text = get_dunning_letter_text(dunning_type = dunning_type.name, doc = target.as_dict())
+			if letter_text:
+				target.body_text = letter_text.get('body_text')
+				target.closing_text = letter_text.get('closing_text')
+				target.language = letter_text.get('language')
+			amounts = calculate_interest_and_amount(target.posting_date, target.outstanding_amount,
+				target.rate_of_interest, target.dunning_fee, target.overdue_days)
+			target.interest_amount = amounts.get('interest_amount')
+			target.dunning_amount = amounts.get('dunning_amount')
+			target.grand_total = amounts.get('grand_total')
```
barredterra, line 1628: I don't think this code is at the right place in sales invoice. I would expect to find it somewhere in `class Dunning`, maybe in `validate`?

KanchanChauhan: This is for when we create the Dunning from Sales Invoice. We have these validations on Dunning as well, but we would like to see the calculated amounts right away when we create the Dunning from Sales Invoice.

erpnext/accounts/doctype/dunning/dunning.py
```diff
@@ -0,0 +1,121 @@
...
+				"against_voucher_type": "Dunning",
+				"cost_center": inv.cost_center,
+				"project": inv.project
+			}, inv.party_account_currency, item=inv)
+		)
+		gl_entries.append(
+			self.get_gl_dict({
+				"account": inv.against_income_account,
+				"against": self.customer,
+				"credit": self.dunning_amount,
```
nabinhait: Set `credit_in_account_currency, project`

erpnext/accounts/doctype/dunning/dunning.py
```diff
@@ -0,0 +1,121 @@
...
+				"cost_center": inv.cost_center
+			}, item=inv)
+		)
+		make_gl_entries(gl_entries, cancel=(self.docstatus == 2), update_outstanding="Yes", merge_entries=False)
```
nabinhait:
```suggestion
		make_gl_entries(gl_entries, cancel=(self.docstatus == 2), update_outstanding="No", merge_entries=False)
```

As we are not maintaining outstanding for Dunning, it should not be updated via GL Entry.

erpnext/accounts/doctype/dunning/dunning.py
```diff
@@ -0,0 +1,121 @@
...
+					if inv.party_account_currency==company_currency else self.dunning_amount,
+				"against_voucher": self.name,
+				"against_voucher_type": "Dunning",
+				"cost_center": inv.cost_center,
+				"project": inv.project
+			}, inv.party_account_currency, item=inv)
+		)
+		gl_entries.append(
+			self.get_gl_dict({
+				"account": inv.against_income_account,
```
nabinhait: Many users can prefer to book income in some other account than the invoice. It will good if we add a field to enter the income account in Dunning and book income against that account.

Also, `inv.against_income_account` generally hold multiple income accounts separated with a comma.

nabinhait (changes requested):

KanchanChauhan:
@nabinhait All the changes are done and tests are written. We will be doing documentation now but that will be separate PR.

erpnext/accounts/doctype/dunning/test_dunning.py
```diff
@@ -0,0 +1,100 @@
...
+    dunning_type.end_day = 20
+    dunning_type.dunning_fee = 20
+    dunning_type.rate_of_interest = 8
+    dunning_type.append(
+        "dunning_letter_text", {
+            'language': 'en',
+            'body_text': 'We have still not received payment for our invoice ',
+            'closing_text': 'We kindly request that you pay the outstanding amount immediately, including interest and late fees.'
+        })
+    dunning_type.save()
```
barredterra: Indentation is with spaces here ;)

nabinhait (dismissed):

erpnext/accounts/doctype/dunning/test_dunning.py
nabinhait: fixed

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -199,8 +199,8 @@ def validate_bank_accounts(self):
 
 	def validate_account_type(self, account, account_types):
 		account_type = frappe.db.get_value("Account", account, "account_type")
-		if account_type not in account_types:
-			frappe.throw(_("Account Type for {0} must be {1}").format(account, comma_or(account_types)))
+		# if account_type not in account_types:
+		# 	frappe.throw(_("Account Type for {0} must be {1}").format(account, comma_or(account_types)))
```
gavindsouza, line 203: Can we get rid of this method altogether given it just adds a DB call and does no validation anymore? [it's the same in v13, v14, develop]

PR #22567, opened 2020-07-03
Bug While fetching Conversation Factor

jigartarpara:
Bug:
In Sales Order's Item Table
we select abc item with conversation factor 20
so it will fetch correct data for conversation factor
now suppose you reselect another item in same row from abc to xyz with conversation factor 30
still it will fetch conversation factor to 20 only
because this api pass current conversation factor and backend will send it back without checking for new item

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -309,8 +309,7 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 	if item.stock_uom == args.uom:
 		out.conversion_factor = 1.0
 	else:
-		out.conversion_factor = args.conversion_factor or \
-			get_conversion_factor(item.name, args.uom).get("conversion_factor")
+		out.conversion_factor = get_conversion_factor(item.name, args.uom).get("conversion_factor")
 
 	args.conversion_factor = out.conversion_factor
 	out.stock_qty = out.qty * out.conversion_factor


erpnext/stock/get_item_details.py
```diff
@@ -309,8 +309,7 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 	if item.stock_uom == args.uom:
 		out.conversion_factor = 1.0
 	else:
-		out.conversion_factor = args.conversion_factor or \
-			get_conversion_factor(item.name, args.uom).get("conversion_factor")
+		out.conversion_factor = get_conversion_factor(item.name, args.uom).get("conversion_factor")
```
rohitwaghchaure, line 312: The fix is not correct, because this field is editable and user can add manually conversion factor in case conversion factor is not set against the selected UOM. With above changes system will set value again to zero even if user has manually set the conversion factor.

The correct fix can be, on selection of another item at JS side set the conversion factor as zero so that system will recalculate the conversion factor.

rohitwaghchaure:
@jigartarpara Don't send PR on version-12 branch, send fixes on version-12-hotfix

jigartarpara:
@rohitwaghchaure 
I don't think suggested solution can fixed this
Then what if selected another item hase already have conversation factor available in item master
then still it will fetch current conversation factor available which is for priviues item

jigartarpara:
I think if user change item then it should fetch cnversation factor available in new selected item master 
after that if user want s/he can change conversation factor
what you think?
@rohitwaghchaure

jigartarpara:
> @jigartarpara Don't send PR on version-12 branch, send fixes on version-12-hotfix

Sure I will do in future

PR #22568, opened 2020-07-03
fix: Item Tax Template permissions as per company

ahmadpak:
Currently item tax templates are not restricted to companies and all templates show up in the invoice neglected the selected company. This fix introduces a field for company in Item Tax Template and ensures only company related templates are shown in the Invoices.


deepeshgarg007:

deepeshgarg007 (approved):

PR #22570, opened 2020-07-03
fix: fetch "Employee Leave Policy" from "Employee Grade" if available

AfshanKhan:
Written "Fetch From" for leave_policy field in Employee doctype which fetches the data from Employee Grade doctype.


erpnext/hr/doctype/employee/employee.json
```diff
@@ -410,6 +410,7 @@
    "options": "Branch"
   },
   {
+   "fetch_from": "grade.default_leave_policy",
```
deepeshgarg007, line 413: `Fetch If Empty` should also be checked otherwise any other policy selected by user will get overridden by the default policy

deepeshgarg007 (changes requested):

erpnext/hr/doctype/employee/employee.json
AfshanKhan: oh okay cool

marination (approved):

marination:

PR #22573, opened 2020-07-03
feat: Added all companies option in employee tree to view employee ac…

deepeshgarg007:
In a Multi-Company setup there may be instances where employee belonging to Company 1 may report to another employee of  Company 2

Currently the tree view can only be viewed company wise and there is no option to view the hierarchy of all employees across all companies, so add an option for "All Companies" in the company filter.


marination (approved):

wojosc:
Hi @deepeshgarg007  I have a similar trouble on another subject: items

when setting a filter in Item List for Item Group I will only see that group but not the one below.

e.g. when filtering for vehicle I would like to also see car, lorry, truck etc. and the cars with 2 seats etc.

when filtering car I would like to see all cars with 2 seats, 4 seats etc.

Vehicles
|-car
||--2 seats
||--4 seats
||--5 seats
|--lorry
|--truck


Are we talking about the same thing here?

Kind Regards - viele Grüße
Wolfram

PR #22577, opened 2020-07-03
fix: Consider company fiscal year for getting balance

deepeshgarg007:
Suppose in a Multi-Company setup there are two companies one based in India and one based in USA.

Now fiscal years for both these companies will obviously be different, one will be from Apr-March and one will be from Jan-Dec

While getting balances it gets the default fiscal year based on current date for both the companies either in India or USA, so if you view the Profit and Loss Report, it will show you the correct amount based on the fiscal year selected but the if you view the balance in COA if will consider the default fiscal year and the balance visible there may be different

Solution: First give preference to Company Fiscal year and then default fiscal year

### DIFF
--- a/erpnext/accounts/utils.py
+++ b/erpnext/accounts/utils.py
@@ -133,7 +133,7 @@ def get_balance_on(account=None, date=None, party_type=None, party=None, company
 		acc = frappe.get_doc("Account", account)
 
 	try:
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

rohitwaghchaure (approved):

PR #22580, opened 2020-07-04
fix: grammatical fixes in module onboarding

michellealva:
grammatical fixes in module onboarding for Assets, Payroll, HR, Stock, Accounts, CRM, Manufacturing and Healthcare module


erpnext/stock/module_onboarding/stock/stock.json
```diff
@@ -31,15 +31,15 @@
...
-   "step": "Create a Supplier"
-  },
   {
    "step": "Introduction to Stock Entry"
   },
   {
    "step": "Create a Stock Entry"
   },
+  {
+   "step": "Create a Supplier"
```
marination: Let's keep the basic setups first so supplier aftrt product is okay, btw supplier was way down earlier. I got feedback to move it up :)

marination (changes requested):

marination (dismissed):

marination (approved):

PR #22585, opened 2020-07-06
feat(selling): added sales target to customer

MyuddinKhatri:
added sales target to the customer and displayed Monthly Sales Target and Total Monthly Sales as progress in Customer Dashboard

Sales Target Fields
sales_target

Sales Target Dashboard
sales_target_dashboard

### DIFF
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -707,3 +707,6 @@ erpnext.patches.v13_0.move_doctype_reports_and_notification_from_hr_to_payroll #
 erpnext.patches.v13_0.move_payroll_setting_separately_from_hr_settings #22-06-2020
 erpnext.patches.v13_0.check_is_income_tax_component #22-06-2020
 erpnext.patches.v12_0.add_taxjar_integration_field
+erpnext.patches.v12_0.update_sales_taxes_and_charges_template
+erpnext.patches.v13_0.update_customer_total_sales
+erpnext.patches.v13_0.setup_remove_assignment_on_task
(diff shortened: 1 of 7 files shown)


erpnext/patches/v13_0/update_customer_total_sales.py
```diff
@@ -0,0 +1,14 @@
...
+import frappe
+from erpnext.selling.doctype.customer.customer import update_customer_current_month_sales
+
+def execute():
+	'''Update customer monthly sales history based on sales invoices'''
+	frappe.reload_doctype("Customer")
+	customers = [d['name'] for d in frappe.get_list("Customer")]
+
+	for customer in customers:
+		update_customer_current_month_sales(customer)
```
rohitwaghchaure, line 14: This patch will take lot of time if the number of customers are more than 1,00,000

erpnext/patches.txt
```diff
@@ -707,3 +707,6 @@ erpnext.patches.v13_0.move_doctype_reports_and_notification_from_hr_to_payroll #
 erpnext.patches.v13_0.move_payroll_setting_separately_from_hr_settings #22-06-2020
 erpnext.patches.v13_0.check_is_income_tax_component #22-06-2020
 erpnext.patches.v12_0.add_taxjar_integration_field
+erpnext.patches.v12_0.update_sales_taxes_and_charges_template
```
rohitwaghchaure: What is the use of this patch and why the file of the patch not showing in this PR

erpnext/patches.txt
```diff
@@ -707,3 +707,6 @@ erpnext.patches.v13_0.move_doctype_reports_and_notification_from_hr_to_payroll #
...
+erpnext.patches.v13_0.update_customer_total_sales
+erpnext.patches.v13_0.setup_remove_assignment_on_task
```
rohitwaghchaure: What is the use of this patch and why the file of the patch not showing in this PR

erpnext/patches/v13_0/update_customer_total_sales.py
MyuddinKhatri: @rohitwaghchaure  Can you please suggest something for this?

rohitwaghchaure: get customer wise sales data from sales invoice using sql query(group by customer) and then update the value in the customer. If particular instance has 10,000 customer out of which only 100 customers has the sales invoices and not all then no need to check for each customer.

PR #22589, opened 2020-07-06
fix: Do not add GST tax amount in grand total for reverse charge invoices

deepeshgarg007:
Currently even though the net impact on GST account for reverse charge invoices is zero the GST amount still gets added in the invoice grand total which it shouldn't. This PR fixes that

Screenshot 2020-07-06 at 1 28 48 PM


deepeshgarg007:

rohitwaghchaure (approved):

PR #22601, opened 2020-07-07
fix: Payment reco error in multi-company setup

deepeshgarg007:
Screenshot 2020-07-07 at 2 10 10 PM

While doing payment reconciliation in a multi-company  setup for any other company than default company the above error is encountered since the company in JV created to reconcile is always the default.


deepeshgarg007:

rohitwaghchaure (approved):

PR #22605, opened 2020-07-07
fix: Changed error message in the Product Bundle

crescent28:
Changed the error message in the Product Bundle: 

When you create an item for the Product Bundle: 

1. Product Item [Parent] --> Uncheck Maintain stock 
2. Raw Material [Child] --> Stock item 

The product item should not be able to add in the child table, else the error should pop up.
Screenshot 2020-07-07 at 4 46 51 PM


marination (approved):

marination:

PR #22613, opened 2020-07-08
fix: Check if homepage is not setup-wizard

netchampfaris:
The new desk changes sets the default homepage to 'workspace'. To avoid this problem in the future, check if the value is not setup-wizard.

Also throw an exception if the check passes, so it exits with error.

```
Projects/benches/frappe-bench                                                                                                                                                                             
▶ bench --site f-test install-app erpnext

Installing erpnext...
  File "/Users/farisansari/.pyenv/versions/3.7.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/farisansari/.pyenv/versions/3.7.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 99, in <module>
    main()
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 18, in main
    click.Group(commands=commands)(prog_name='bench')
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/farisansari/Projects/benches/frappe-bench/env/lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/commands/__init__.py", line 26, in _func
    ret = f(frappe._dict(ctx.obj), *args, **kwargs)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/commands/site.py", line 209, in install_app
    _install_app(app, verbose=context.verbose)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/installer.py", line 76, in install_app
    out = frappe.get_attr(before_install)()
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/erpnext/erpnext/setup/install.py", line 37, in check_setup_wizard_not_completed
    frappe.throw(message, exc=frappe.IncompatibleApp)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/__init__.py", line 396, in throw
    msgprint(msg, raise_exception=exc, title=title, indicator='red', is_minimizable=is_minimizable)
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/__init__.py", line 375, in msgprint
    _raise_exception()
  File "/Users/farisansari/Projects/benches/frappe-bench/apps/frappe/frappe/__init__.py", line 326, in _raise_exception
    raise raise_exception(msg)
frappe.exceptions.IncompatibleApp: Cannot install ERPNext.

ERPNext can only be installed on a fresh site where the setup wizard is not completed.
You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall

```


scmmishra (dismissed):

netchampfaris:
@gavindsouza If any of the current site commands break, it will also break `--site all` command. To make sure `--site all` doesn't break if the command breaks for a single site, it has to be fixed separately (if that is the correct thing to do).

Also, this change is a breaking change even for a single site command, but this is what should have happened.

gavindsouza:
@netchampfaris I think https://github.com/frappe/frappe/pull/10946 helps make this happen

erpnext/setup/install.py
```diff
@@ -29,12 +29,12 @@ def after_install():
...
-		print("ERPNext can only be installed on a fresh site where the setup wizard is not completed")
-		print("You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall")
-		print()
-		return False
+	if frappe.db.get_default('desktop:home_page') != 'setup-wizard':
+		message = """Cannot install ERPNext.
+
+ERPNext can only be installed on a fresh site where the setup wizard is not completed.
+You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall"""
+		frappe.throw(message, exc=frappe.IncompatibleApp)
```
gavindsouza: This issue is related to the current site state and not app compatibility. Perhaps better if we don't raise `IncompatibleApp` here.

```suggestion
		frappe.throw(message)
```

erpnext/setup/install.py
```diff
@@ -29,12 +29,12 @@ def after_install():
...
-		print()
-		print("ERPNext can only be installed on a fresh site where the setup wizard is not completed")
-		print("You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall")
-		print()
-		return False
+	if frappe.db.get_default('desktop:home_page') != 'setup-wizard':
+		message = """Cannot install ERPNext.
+
+ERPNext can only be installed on a fresh site where the setup wizard is not completed.
+You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall"""
```
gavindsouza: The line "Cannot install ERPNext." should be handled by Frappe IMO.

```suggestion
		message = """ERPNext can only be installed on a fresh site where the setup wizard is not completed. 
You can reinstall this site (after saving your data) using: bench --site [sitename] reinstall"""

gavindsouza (changes requested):

erpnext/setup/install.py
netchampfaris: @gavindsouza The default exception is ValidationError, I think IncompatibleApp is better than that.

"If setup wizard is completed on a site, ERPNext is incompatible with the site"

surajshetty3416: `IncompatibleApp` seems okay in this case. 

Apart from that, since currently `IncompatibleApp` is used in the case of incompatible versions, we can add `IncompatibleVersion` error class for it instead?

gavindsouza (approved):

PR #22614, opened 2020-07-08
fix: Add default cost center in payment reconciliation JV

deepeshgarg007:
Add default cost center for Payment Reconciliation JV.

In a Multi-Company setup default cost center from Global defaults was used even if the company is different and while submitting, cost center validation was triggered


deepeshgarg007:

rohitwaghchaure (approved):

PR #22615, opened 2020-07-08
fix: log type mandatory error while exposing api call to employee checkin.

KaviyaPeriyasamy:
This PR helps to fix this [issue](https://discuss.erpnext.com/t/biometric-attendance-sync-tool/62122/12).

While exposing API calls from biometric attendance to employee check-in, the log type field seems mandatory. But in biometric-based attendance, most users will not mention the log type. So the log type will be ```None``` in most cases.


karthikeyan5 (dismissed):

erpnext/hr/doctype/employee_checkin/employee_checkin.js
```diff
@@ -6,5 +6,9 @@ frappe.ui.form.on('Employee Checkin', {
 		if(!frm.doc.time) {
 			frm.set_value("time", frappe.datetime.now_datetime());
 		}
+	},
+	refresh: (frm) => {
+		// make log type mandatory
+		frm.set_df_property('log_type', 'reqd', frm.doc.log_type ? 0 : 1);
```
deepeshgarg007: Why not make the `log_type` field an optional field itself? Why it needs to be done this way I am not able to understand

karthikeyan5: Yes, that's right... `log_type` was optional earlier. But in this PR https://github.com/frappe/erpnext/pull/20403 it was changed to mandatory.

But, there is a use case where `log_type` field can be left blank and this use case is only concerning the creation of employee checkin via API.

Assuming that there must be a reason why `log_type` was made mandatory in PR https://github.com/frappe/erpnext/pull/20403, the least possible change to make was to set this field to mandatory only when creating Employee Checkin from the UI.

If you find no use cases for marking this field mandatory in the UI also then we can remove this line.

Anurag810 (changes requested):
hello, @karthikeyan5 can we Automatically update the Last sync of checkin value based on the latest checkin/checkout time, on the creation of Employee Checkin / Checkout record. because it is not good that the user is setting it manually. if Employee is creating check-in records manually then he needs to set that field manually every time.

karthikeyan5:
> hello, @karthikeyan5 can we Automatically update the Last sync of checkin value based on the latest checkin/checkout time, on the creation of Employee Checkin / Checkout record. because it is not good that the user is setting it manually. if Employee is creating check-in records manually then he needs to set that field manually every time.

@Anurag810,  this PR is not at all related to "Last sync of checkin"... can we please take this up separately in an issue or something?

erpnext/hr/doctype/employee_checkin/employee_checkin.js
deepeshgarg007: > Yes, that's right... `log_type` was optional earlier. But in this PR #20403 it was changed to mandatory.
> 
> But, there is a use case where `log_type` field can be left blank and this use case is only concerning the creation of employee checkin via API.
> 
> Assuming that there must be a reason why `log_type` was made mandatory in PR #20403, the least possible change to make was to set this field to mandatory only when creating Employee Checkin from the UI.
> 
> If you find no use cases for marking this field mandatory in the UI also then we can remove this line.

Even from the UI if the `log_type` is entered then only it will make the field mandatory. If `log_type` is not entered, it will accept without it i guess, so it acts like a optional field, this statement doesn't makes any difference

Anurag810: @karthikeyan5 I think Deepesh is right just make it optional because if  log_type is entered then only it will make the field mandatory. anyway, it acts like an optional.

karthikeyan5: done... the js validations have been removed...

deepeshgarg007 (approved):

PR #22616, opened 2020-07-08
fix: buying price for non stock item in gross profit report

AfshanKhan:
The buying price of a non-stock item should only be calculated if the purchase order has a project or cost center against that item. Else it should be 0.


erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -253,38 +253,39 @@ def get_buying_amount(self, row, item_code):
...
+			average_buying_rate = get_incoming_rate(args)
+			self.average_buying_rate[item_code] =  flt(average_buying_rate)
 
 		return self.average_buying_rate[item_code]
 
-	def get_last_purchase_rate(self, item_code):
+	def get_last_purchase_rate(self, item_code, row):
+		condition = ''
+		if row.project:
+			condition += "AND a.project='%s'"%(row.project)
```
nabinhait:
```suggestion
			condition += " AND a.project='%s'" % (row.project)
```

erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -253,38 +253,39 @@ def get_buying_amount(self, row, item_code):
...
+		elif row.cost_center:
+			condition += "AND a.cost_center='%s'"%(row.cost_center)
```
nabinhait:
```suggestion
			condition += " AND a.cost_center='%s'" % (row.cost_center)
```

erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -253,38 +253,39 @@ def get_buying_amount(self, row, item_code):
...
 		if self.filters.to_date:
 			last_purchase_rate = frappe.db.sql("""
 			select (a.base_rate / a.conversion_factor)
 			from `tabPurchase Invoice Item` a
 			where a.item_code = %s and a.docstatus=1
+			{0}
 			and modified <= %s
```
nabinhait: Remove if...else condition and make a common query. Make `and modified <= %s` a part of `condition`.

nabinhait (changes requested):

nabinhait (approved):

PR #22617, opened 2020-07-08
refactor: shopping cart

anupamvs:
**Issue-**

- Cannot delete company because is linked with Shopping Cart Settings Even shopping cart is disabled.
- Cart is not hidden from the navigation bar if the Shopping Cart is disabled.
- 'Show configure button' in the Shopping Cart Settings is not self- explanatory.

**Fixes-**

- Making Company and User Group not mandatory.
- Making it mandatory depends on shopping cart enabled.
- Hiding cart from navigation bar if the shopping cart is disabled.
- Changing Show Configure Button -> Enable Variants.


Anurag810 (dismissed):
image
have you changed as we talked earlier to remove white space?

nabinhait (approved):

PR #22622, opened 2020-07-08
feat: added range for age in stock ageing

AfshanKhan:
Added filters and columns for "age range".
Showing the stock count of a particular stock in the appropriate range column. 
Screenshot 2020-07-08 at 7 25 47 PM


erpnext/stock/report/stock_ageing/stock_ageing.py
```diff
@@ -55,7 +56,25 @@ def get_average_age(fifo_queue, to_date):
 
 	return flt(age_qty / total_qty, 2) if total_qty else 0.0
 
+def get_range_age(filters, fifo_queue, to_date):
```
deepeshgarg007, line 59: @AfshanKhan this logic is not enough. If have items in multiple age ranges but it just displays age only for the first item in FIFO queue. I have available qty as 3 and 2 but in ageing columns only 1 and 1 is shown i.e only the first item in the FIFO Queue

image

deepeshgarg007 (changes requested):

erpnext/stock/report/stock_ageing/stock_ageing.py
AfshanKhan: fixed.

erpnext/stock/report/stock_ageing/stock_ageing.py
```diff
@@ -55,7 +56,25 @@ def get_average_age(fifo_queue, to_date):
...
+	range1 = range2 = range3 = above_range3 = 0.0
+	for item in fifo_queue:
```
deepeshgarg007, line 61: @AfshanKhan This is still not correct. The ranges should be added instead of just assigning the latest item count

deepeshgarg007 (changes requested):

erpnext/stock/report/stock_ageing/stock_ageing.py
```diff
@@ -263,3 +284,18 @@ def get_chart_data(data, filters):
 		},
 		"type" : "bar"
 	}
+
+def setup_ageing_columns(filters, range_columns):
+	for i, label in enumerate(["0-{range1}".format(range1=filters["range1"]),
+		"{range1}-{range2}".format(range1=cint(filters["range1"])+ 1, range2=filters["range2"]),
+		"{range2}-{range3}".format(range2=cint(filters["range2"])+ 1, range3=filters["range3"]),
+		"{range3}-{above}".format(range3=cint(filters["range3"])+ 1, above=_("Above"))]):
+			add_column(range_columns, label="Age in ("+ label +")", fieldname='range' + str(i+1))
```
deepeshgarg007:
```suggestion
			add_column(range_columns, label="Age ("+ label +")", fieldname='range' + str(i+1))
```

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #22627, opened 2020-07-09
fix: Remove rename related code from Serial No

marination:
- Serial No rename is disabled via https://github.com/frappe/erpnext/pull/22450
- Removed Rename related code, rename will be handled when Serial no refactor is worked upon


Anurag810 (approved):

marination:

sjain17:
What is the process to rename the serial number now?

marination:
@sjain17  this was done as the root cause wasnt known, as a temporary block. The same has now been found and this has been reverted via https://github.com/frappe/erpnext/pull/22746

PR #22629, opened 2020-07-09
fix: Due to decimal issue make purchase receipt button not showing in the purchase order

rohitwaghchaure:
**Issue**

User not able to view purchase receipt / purchase invoice button from the dropdown

Screenshot 2020-07-09 at 6 16 36 PM

After debug the issue came to know that system rounding the per_received to 100 and therefore the Purchase Receipt was not displaying 
Screenshot 2020-07-09 at 6 12 40 PM

**After Fix**

Screenshot 2020-07-09 at 6 12 27 PM


rohitwaghchaure:

Anurag810 (approved):

PR #22631, opened 2020-07-09
fix: incorrect delivered qty in Supplier-Wise Sales Analytics

rohitwaghchaure:
**Issue**
Sales Invoice with update stock enabled was not considered

Screenshot 2020-07-09 at 6 40 03 PM

**After Fix**
Screenshot 2020-07-09 at 6 39 37 PM


rohitwaghchaure:

Anurag810 (approved):

PR #22632, opened 2020-07-09
fix: update set query for warehouse after changing the company

AfshanKhan:
Previously after changing the company in material request form the options for warehouses were still showing warehouses of the company selected at first.
Screenshot 2020-07-09 at 6 36 08 PM
Screenshot 2020-07-09 at 6 36 22 PM

So made changes to update set query after changing the company
Screenshot 2020-07-09 at 6 42 21 PM
Screenshot 2020-07-09 at 6 42 33 PM


rohitwaghchaure:
@AfshanKhan the old code is working correctly in v12 (check [GIF](https://www.dropbox.com/s/vylc2wz09i34wid/material_request_warehouse.gif?dl=0)), it seems that the setup function is not using the current doc. You need to fix that.

AfshanKhan:
@rohitwaghchaure  Updated the code. Removed previous changes and replaced set_query with get_query.

marination (approved):

marination:

PR #22634, opened 2020-07-09
fix: default overwrite property and 'company' is not defined

Anurag810:


rohitwaghchaure:
@Anurag810 please share the information about the issue and fix

Anurag810:
> @Anurag810 please share the information about the issue and fix

@rohitwaghchaure so, additional salary ```overwrite_salary_structure_amount``` property overwrites the component in salary slip and only one additional salary is allowed at once (with ```overwrite_salary_structure_amount```). but an employee can have more than one incentive or additional salary, so disabled that property so the user can create more that one employee incentive(additional salary) in a single month.

rohitwaghchaure (approved):

PR #22645, opened 2020-07-10
fix: validation for additional salary

Anurag810:
Allowed if the payroll date in additional salary is before relieving date, even if employee status is left.
and some fixes regarding recurring additional salary


erpnext/payroll/doctype/additional_salary/additional_salary.py
```diff
@@ -33,12 +33,16 @@ def validate_dates(self):
 			frappe.throw(_("From Date can not be greater than To Date."))
 
 		if date_of_joining:
-			if getdate(self.payroll_date) < getdate(date_of_joining):
+			if self.payroll_date and getdate(self.payroll_date) < getdate(date_of_joining):
 				frappe.throw(_("Payroll date can not be less than employee's joining date."))
-			elif getdate(self.from_date) < getdate(date_of_joining):
+			elif self.to_date and getdate(self.from_date) < getdate(date_of_joining):
```
ruchamahabal:
```suggestion
			elif self.from_date and getdate(self.from_date) < getdate(date_of_joining):
```

ruchamahabal (changes requested):

ruchamahabal (approved):

PR #22646, opened 2020-07-10
fix: incorrect qty after transaction in stock ledger entry

rohitwaghchaure:
**Issue**

Screenshot 2020-07-10 at 1 38 13 PM

1. Make stock reco with serial nos, put serial numbers from other warehouse. For example 6 serial nos(S1, S2, S3, A1, A2, A3) present in warehouse A and you want to transfer 3(S1, S2, S3) of them to warehouse B. 

1. Add serial nos S1, S2, S3 in stock reco also add 3 more serial nos S4, S5, S6 in the same stock reconciliation and submit it

1. After that check the stock summary report for the warehouse A, it will not matched with the available serial nos in the warehouse A


rohitwaghchaure:
@Mergifyio backport develop

marination (approved):

PR #22648, opened 2020-07-10
fix: Unable to submit backdated stock transactions for different items

deepeshgarg007:
Always use current posting date to check Stock and Account Balance sync as there can be future transactions for other items


nabinhait (dismissed):

nabinhait (commented):
Travis is failing

nabinhait (approved):

PR #22649, opened 2020-07-10
fix: incorrect balance qty in stock ledger report

rohitwaghchaure:
**Issue**

in stock ledger, user has put the batch filter and balance quantity is not matching with the balance quantity in the report Batch wise Balance History

**Stock Ledger Report**
Screenshot 2020-07-10 at 6 00 48 PM

**Batch wise Balance History Report**
Screenshot 2020-07-10 at 6 01 39 PM

**After Fix**
Screenshot 2020-07-10 at 6 04 35 PM


rohitwaghchaure:
@Mergifyio backport develop

deepeshgarg007 (approved):

PR #22660, opened 2020-07-12
fix: parent and dependent task for duplicate projects

AfshanKhan:
When duplicating a project with its task, the parent task and dependent task used to be of the previous project. Fixed the linking of new project tasks.

Previous tasks status was being set for the new tasks after duplicating the project. Fixed that.

When removing a parent task from a task the linking was still maintained which prevented from deleting that task. Fixed that


rmehta (changes requested):
test missing, comment missing

PR #22665, opened 2020-07-13
fix: Block Invalid Serial No updates in Maintenance Schedule

marination:
**Issue:**
- There is an ability to assign any Serial No against any item in Maintenance Schedule's Items table
- On submit that Serial No gets updated too (amc expiry date is updated)

**Fix:**
- Validate whether Serial No. is valid or not to keep transaction consistent
- If not:
  Screenshot 2020-07-13 at 12 25 28 PM


marination:
@Mergifyio backport version-12-htofix

rohitwaghchaure (approved):

PR #22675, opened 2020-07-14
fix: Dont overwrite default warehouse in Material Request

marination:
**Issue:**
- On setting default warehouse at item level, `get_item_data` fetched default warehouse every single time it was triggered, overwriting what was already set
mr-warehouse-issue

**Fix:**
- overwrite warehouse only on item code change, otherwise don't fetch default warehouse on rate,qty,warehouse,etc trigger
mr-warehouse-fix


rohitwaghchaure (approved):

rohitwaghchaure:
@Mergifyio backport version-12-hotfix

PR #22676, opened 2020-07-14
fix: Unable to change link from new lead to existing customer

anupamvs:
**Issue-** In Opportunity, when 'Opportunity From' is set as 'Customer', the 'Customer' field still displays Leads.
image


Anurag810 (dismissed):

deepeshgarg007:
@anupamvs Works fine for me. Can you add a GIF here on how to replicate the issue

PR #22678, opened 2020-07-14
fix(report): fix messed up alignment in script report that extends financial_statements.js

DeeMysterio:
**Problem:**

I have a script report that is extending the file `apps/erpnext/erpnext/public/js/financial_statements.js`. So all the columns and everything in my report is quite messed up and **it doesn't get changed from the python file of my report**.

This is how the report looks:

Screenshot from 2020-07-09 13-17-22

I figured out that it was the `formatter` function in the file `financial_statements.js`.

In the case of my report, the `formatter` function on line 5 is also receiving `undefined` value for the parameter called `data` at some point. Hence it is giving javascript console errors at the line 7 at `data.account_name` and at the line 16 at `!data.parent_account`.

**Changes:**

Making those conditions a little bit robust by checking for the existence of `data` like so `data && column.fieldname=="account"` at line 6 for example and also on line 16 is solving the problem.

And the report looks good:

Screenshot from 2020-07-09 13-18-12

### DIFF
--- a/erpnext/public/js/financial_statements.js
+++ b/erpnext/public/js/financial_statements.js
@@ -3,7 +3,7 @@ frappe.provide("erpnext.financial_statements");
 erpnext.financial_statements = {
 	"filters": get_filters(),
 	"formatter": function(value, row, column, data, default_formatter) {
-		if (column.fieldname=="account") {
+		if (data && column.fieldname=="account") {
 			value = data.account_name || value;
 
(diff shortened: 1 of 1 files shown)


deepeshgarg007 (approved):

deepeshgarg007:
@DeeMysterio Please raise the PR on develop branch, any PR cannot be merged directly in verison-12 branch. Closing this

PR #22680, opened 2020-07-14
fix: while making Sales Invoice from Sales Order payment term was not…

AfshanKhan:
… getting fetched before saving the document. 

Now the payment schedule also gets fetch.

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.py
+++ b/erpnext/selling/doctype/sales_order/sales_order.py
@@ -697,6 +697,10 @@ def make_sales_invoice(source_name, target_doc=None, ignore_permissions=False):
 		"Sales Team": {
 			"doctype": "Sales Team",
 			"add_if_empty": True
+		},
+		"Payment Schedule": {
+			"doctype": "Payment Schedule",
+			"add_if_empty": True
 		}
 	}, target_doc, postprocess, ignore_permissions=ignore_permissions)
 


Anurag810 (approved):

deepeshgarg007 (commented):
@AfshanKhan @Anurag810 This is intentional I guess. The Invoice creation date and Sales Invoice creation dates are different in most case and it doesn't makes any sense to copy the payment terms then

AfshanKhan:
@deepeshgarg007 @Anurag810 If payment term is getting fetch from Sales Order then payment schedule should also be same right. They could change it later if they want to.

deepeshgarg007:
> @deepeshgarg007 @Anurag810 If payment term is getting fetch from Sales Order then payment schedule should also be same right. They could change it later if they want to.

Not necessarily, Refer my reply in this [PR](https://github.com/frappe/erpnext/pull/20509) for more info

AfshanKhan:
> > @deepeshgarg007 @Anurag810 If payment term is getting fetch from Sales Order then payment schedule should also be same right. They could change it later if they want to.
> 
> Not necessarily, Refer my reply in this [PR](https://github.com/frappe/erpnext/pull/20509) for more info

Ok got it

Anurag810:
I got your point @deepeshgarg007  that is a valid use case. It can vary from business to business payment can be taken at the Order level or Invoice level with different payment terms date.

PR #22681, opened 2020-07-14
fix: Quotation lost reason options fix

anupamvs:
**Issue-** When we making a Quotation as lost in the modal the Link field is Opportunity Lost Reason instead of Quotation Lost Reason.
quoatation_b

**Fix-**
quoatation_after


Anurag810 (commented):
Please update Description

erpnext/selling/sales_common.js
```diff
@@ -486,16 +486,22 @@ frappe.ui.form.on(cur_frm.doctype,"project", function(frm) {
 
 frappe.ui.form.on(cur_frm.doctype, {
 	set_as_lost_dialog: function(frm) {
+		console.log(frm.doc.lost_doctype);
```
deepeshgarg007, line 489: Remove console

erpnext/selling/sales_common.js
```diff
@@ -486,16 +486,22 @@ frappe.ui.form.on(cur_frm.doctype,"project", function(frm) {
...
-				"fieldname": "lost_reason",
-				"options": "Lost Reason Detail",
-				"reqd": 1},
-
-				{"fieldtype": "Text", "label": __("Detailed Reason"), "fieldname": "detailed_reason"},
+				{
+					"fieldtype": "Table MultiSelect",
+					"label": __("Lost Reasons"),
+					"fieldname": "lost_reason",
+					"options": frm.doctype=='Opportunity' ? 'Lost Reason Detail': 'Quotation Lost Reason Detail',
```
deepeshgarg007, line 497:
```suggestion
					"options": frm.doctype === 'Opportunity' ? 'Lost Reason Detail': 'Quotation Lost Reason Detail',
```

erpnext/setup/doctype/quotation_lost_reason_detail/quotation_lost_reason_detail.json
```diff
@@ -0,0 +1,31 @@
+{
```
deepeshgarg007, line 1: Add Patch to update data from `Lost Reason Detail` to `Quotation Lost Reason Detail` for existing records
Also since now we have a separate doctype for `Quotation` maybe `Lost Reason Detail` can be renamed to `Opportunity Lost Reason Detail` to reduce confusion

deepeshgarg007 (changes requested):

PR #22684, opened 2020-07-14
fix: Replace company abbr

nabinhait:
While replacing the abbr system renames all records of account, warehouse, etc where abbr added as a suffix. After renaming each document there was `commit`, which can cause partial update if it fails in a later stage. Removed commit.

Also in JS, removed `set_value` for the same reason. If it fails on the server side, it should not update abbr from js.


nabinhait:
@Mergifyio backport develop

rohitwaghchaure (approved):

PR #22686, opened 2020-07-14
 fix(GST): Do not add tax amount in grand total for reverse charge invoices

deepeshgarg007:


deepeshgarg007:

ishanloya:
Hi. I'm not sure if there was a discussion on the forum before this issue was merged.

Before this change, we were using two different accounts on the purchase invoice which cancelled each other out ("Reverse charge liability account" credited and "ITC reverse charge interim account" debited for the amount of the GST).
[more info on this here](https://icmai.in/TaxationPortal/upload/IDT/Article_GST/38.pdf)

This change requires a single account in place of them which gets both debited and credited. 

Since this changes the procedure for purchase invoice passing, I think there should have been some discussion/ heads-up about this change before it was merged. In case it was already discussed and I have missed it, please ignore. 

Thanks

ishanloya:
Also, from an accounts point of view, is it ok to debit and credit the same account in a single entry?

deepeshgarg007:
@ishanloya Please pull latest changes and check, the case where you are referring to use two different accounts where  one is debited and one is credited should also work. Are any other taxes are also applied along with GST? In that case you current configuration may fail. Will fix that as well shortly.

ishanloya:
Thanks for the quick reply @deepeshgarg007 . Appreciate it. I tried with two different accounts (one with the add option and other with the deduct option) but in this case both accounts are being credited and debited. This is happening even when there aren't any other taxes. If I remove these accounts from the GST settings page however, then the accounting works as expected. Is there any real downside to not having these accounts listed on the GST settings page?

PR #22691, opened 2020-07-15
fix: parent and dependent task for duplicate projects

AfshanKhan:
When duplicating a project with its task, the parent task and dependent task used to be of the previous project. Fixed the linking of new project tasks.


rohitwaghchaure:
@AfshanKhan Kindly add description about issue

AfshanKhan:
> @AfshanKhan Kindly add description about issue

Done

PR #22692, opened 2020-07-15
feat(payroll): Complete Onboarding with form tour

Anurag810:
### **Salary component:**
Salary-component-form-tour

### **Income Tax Slab:**
Income-tax-salb-tour

### **Salary Structure:**
tour for structure

### **Payroll Setting:**

payroll_settings_form_tour

### DIFF
--- a/erpnext/payroll/doctype/income_tax_slab/income_tax_slab.js
+++ b/erpnext/payroll/doctype/income_tax_slab/income_tax_slab.js
@@ -4,3 +4,26 @@
 frappe.ui.form.on('Income Tax Slab', {
 
 });
+
+frappe.tour['Income Tax Slab'] = [
+	{
+		fieldname: "__newname",
+		title: "Enter Name",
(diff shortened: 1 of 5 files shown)


Anurag810:
I think it is not needed for the salary slip(As everything id fetched) and structure assignment (Not much to do). please give you suggestions.

erpnext/payroll/doctype/income_tax_slab/income_tax_slab.js
```diff
@@ -4,3 +4,26 @@
 frappe.ui.form.on('Income Tax Slab', {
 
 });
+
+frappe.tour['Income Tax Slab'] = [
+	{
+		fieldname: "__newname",
+		title: "Enter Name",
+		description: __("Enter name for your Income Tax Slab"),
```
marination, line 12: Entering a name does not require explanation, i think this can be left out of the tour. Tour is just to show the users important fields or settings that may raise questions later.

PR #22696, opened 2020-07-15
feat: Add color field inside Issue doctype

AminovE99:
I thought that it can be good idea to add `color` field inside Issue. 
It will be useful if user wants to mark all `red` tasks as a very important(user can configure priority and color field manually(it is only one example from a lot of cases)).

Apart from that I saw that this field already implemented inside `ToDo` doctype.

image


rohitwaghchaure:
@AminovE99 send new feature on develop branch and not a stable branch

AminovE99:
@rohitwaghchaure Is it possible to have this in version 12?

PR #22698, opened 2020-07-15
feat(Human Resource): Complete Onboarding with form tour

Anurag810:
### leave Type: 

leave-type-tour

### **Holiday List:**
holiday-list-tour

### **HR Settings**
Hr-settings-tour

### **Employee**
employee-tour

### DIFF
--- a/erpnext/hr/doctype/employee/employee.js
+++ b/erpnext/hr/doctype/employee/employee.js
@@ -102,31 +102,25 @@ frappe.tour['Employee'] = [
 	{
 		fieldname: "first_name",
 		title: "Enter Name",
-		description: __("Enter Employee Name."),
 	},
 	{
 		fieldname: "company",
 		title: "Company",
-		description: __("Select Company.")
 	},
 	{
 		fieldname: "gender",
 		title: "Gender",
-		description: __("select Gender.")
 	},
 	{
 		fieldname: "date_of_birth",
 		title: "Date of Birth",
-		description: __("Select Date of Birth.")
 	},
 	{
 		fieldname: "date_of_joining",
 		title: "Date of Joining",
-		description: __("Select Date of Joining.")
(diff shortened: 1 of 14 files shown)


erpnext/hr/doctype/employee/employee.js
```diff
@@ -97,3 +97,36 @@ frappe.ui.form.on('Employee',{
...
+	{
+		fieldname: "date_of_joining",
+		title: "Date of Joining",
+		description: __("Select Date of Joining.")
+	},
+	{
+		fieldname: "holiday_list",
+		title: "Holiday List",
+		description: __("Select Holiday list.")
+	},
```
marination: These look like self explanatory fields. The purpose of form tour isn't to explain each and every field, but to explain those that are confusing or need some documentation to get clarity about. A simple data field such as Name or even Company selection doesn't require explanation.
Unnecessary tour of fields could annoy a user and his experience. So try to limit this to fields that really need explanation

erpnext/hr/doctype/hr_settings/hr_settings.js
```diff
@@ -5,4 +5,38 @@ frappe.ui.form.on('HR Settings', {
...
+		fieldname: "auto_leave_encashment",
+		title: "Auto Leave Encashment",
+		description: __("If checked, the system generates a draft Leave Encashment record on the expiry of the leave allocation for all encashable Leave Types")
+	},
+	{
+		fieldname: "restrict_backdated_leave_application",
+		title: "Restrict Backdated Leave Application",
+		description: __("If checked, the system will not allow making a backdated leave application.")
+	},
+];
```
marination, line 42: for this to reflect you have to set {"action": "Show Form Tour"} in hr_settings.json (onboarding step). Similarly for most other steps

PR #22711, opened 2020-07-16
fix: multiple pricing rules are not working on selling side

rohitwaghchaure:
## **Issue**

1. Created pricing rule for Territory "Ahemedabad" with Margin Type as "Percentage" and amount is 400. Also enabled "Apply Multiple Pricing Rules".
1. Created another pricing rule for Customer Group "Non Profit" with "Rate or Discount" as "Discount Percentage" and Discount Percentage as "5".  Also enabled "Apply Multiple Pricing Rules".
1. Created quotation, with Territory as "Ahemedabad" and Customer Group as "Non Profit", but system has applied only discount percentage and not margin percentage
Screenshot 2020-07-16 at 5 31 50 PM

## **After Fix**
Screenshot 2020-07-16 at 4 43 12 PM

### Also added details of applied pricing rules

Screenshot 2020-07-16 at 5 35 23 PM


rohitwaghchaure:

rohitwaghchaure:
Screenshot 2020-10-28 at 4 51 23 PM

PR #22716, opened 2020-07-16
Fix: Stock Reconciliation Invalid Quantity for Batched Item  

rohitwaghchaure:
**Issue**

For the past dated Stock Reconciliation, the batched item is showing the quantity as per current date instead of quantity as per posting date

Screenshot 2020-07-16 at 10 24 28 PM

**After Fix**
Screenshot 2020-07-16 at 10 14 01 PM


rohitwaghchaure:
@Mergifyio backport develop

PR #22717, opened 2020-07-16
fix: setting filter for project in kanban board

### LINKED ISSUE #22774: Missing filter on project in kanbanview + limit page length
Into a project there is a button to see a kanban view of the project, when pressed erpnext shows the kanban but without any filter and show all task of every projects.

**Step to reproduce:**
1. Create a project
2. Create some tasks releted to it
3. On the project page press "Kanban Board" (top left corner)

**Expected**:
See kanban with only the tasks releted to the current project
**Actual**:
See kanban with all tasks of all projects

Looking into the code i have saw that there are a feature in frappe that can resolve this problem, but this line of code not use it:
https://github.com/frappe/erpnext/blob/750069b6de64260ea2d20325186bc00360c1814c/erpnext/projects/doctype/project/project.py#L475

Should be (maybe):
`quick_kanban_board('Task', project, 'status', project) `

**Other request:**
Would it be possible to remove the limit of visible tasks in the kanban view or increase it by default?
There is no reference to the total number of tasks and if someone does not know that there are more than 20 tasks he may lose them. Not to mention that 20 tasks for a project are very few imho

anupamvs:
**Issue-** When we open kanban from project it doesn't set the filter for project in kanban board the result is it shows all tasks even it did not belong to that project.

image

**Fix-**
passed project parameter in quick_kanban_board which is used to add filter for project.

image


anupamvs:

nabinhait (approved):

PR #22718, opened 2020-07-17
fix: sql fetching wrong field for partywise advance payment amount

AndyOverLord:
This method is getting called on the report "Account Receivable Summary" and also "Account Payable Summary".

This method is getting called as below report python file.
 erpnext/accounts/report/accounts_receivable_summary/accounts_receivable_summary.py

I had a use case of writing opening journal entry with foreign currency, it will get advaced amount in the account currency. But the outstanding amount in "party_detail" is showing in company currency.

Hence it will have a difference between two values which cause paid amount is getting calculated wrongly.
As the code below:
image

By fetching the credit or debit value in account currency can solve this discrepancy. 

> Screenshots/GIFs

### DIFF
--- a/erpnext/accounts/party.py
+++ b/erpnext/accounts/party.py
@@ -618,7 +618,7 @@ def get_partywise_advanced_payment_amount(party_type, posting_date = None, futur
 		WHERE
 			party_type = %s and against_voucher is null
 			and {1} GROUP BY party"""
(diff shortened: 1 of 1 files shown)


deepeshgarg007 (commented):
@AndyOverLord can you update the screenshot with the currency in Journal Entry and the currency showing in report. It will be easier for us to understand the scenario

PR #22722, opened 2020-07-17
fix(maintenance-visit): change fieldtype of status to select

Mangesh-Khairnar:
- change field type of status in maintenance visit
- fixes field type data cannot have options


Mangesh-Khairnar:

ruchamahabal (approved):

PR #22723, opened 2020-07-17
fix: Not able to cancel application,allocation nad encashments

Anurag810:
Not able to cancel leave application, Leave allocation and encashments. due to cancel all feature.
Merge after : https://github.com/frappe/frappe/pull/11034 and https://github.com/frappe/frappe_docs/pull/25


ruchamahabal (changes requested):
@Anurag810 In order to prevent Leave Ledger Entry from getting auto-canceled, you have to add it to the `auto_cancel_exempted_doctypes` list in `hooks.py`. This flag in the .js file is not working

Check: https://github.com/frappe/erpnext/blob/dcef798be716b14ef932322dd9fc3bdc6bc947ad/erpnext/hooks.py#L270

Edit: Resolved. This is dependent on frappe - https://github.com/frappe/frappe/pull/11034

erpnext/hr/doctype/leave_allocation/leave_allocation.js
```diff
@@ -5,20 +5,23 @@ cur_frm.add_fetch('employee','employee_name','employee_name');
 
 frappe.ui.form.on("Leave Allocation", {
 	onload: function(frm) {
+		// Ignore cancellation of doctype on cancel all.
+		frm.ignored_doctypes_on_cancel_all = ["Leave Ledger Entry"];
```
netchampfaris: I think it should be `ignore` instead of `ignored` since we are declaring the behavior.

surajshetty3416 (approved):

PR #22735, opened 2020-07-18
fix: Expense claim outstanding while making payment entry

deepeshgarg007:
While manually selecting Expense Claim in Payment Entry references only sanctioned amount is fetched and taxes and charges are ignored

Expense Claim:
Screenshot 2020-07-18 at 9 07 21 PM

Before:
Screenshot 2020-07-18 at 9 21 22 PM

After:
Screenshot 2020-07-18 at 9 08 11 PM


deepeshgarg007:

nabinhait (approved):

PR #22738, opened 2020-07-19
fix: Multi currency payment reconciliation

deepeshgarg007:
- Default Company Currency symbol is visible on fetching invoices and payment for Parties having party account currency other than company currency

- Reconciling such invoices is not possible since Multi Currency check is not enabled while allocating amounts against such invoices via Journal Entry


deepeshgarg007:

nabinhait (approved):

PR #22741, opened 2020-07-19
fix: Error due to comma in Pricing rule name

deepeshgarg007:
image

Price Not Found error is thrown if Price Rule name has comma in it since price rules were saves as string splitted by commas


deepeshgarg007:

marination (approved):

nabinhait (approved):

rmehta:
Test case missing

rmehta (commented):
why no test case?

PR #22746, opened 2020-07-20
fix: Serial No Rename does not affect  Stock Ledger Entry

marination:
- Rename of a Serial No did not reflect in Stock Ledger Entry due to change in serial no field type from _Small Text_ to _Long Text_ in Stock Ledger Entry's Serial No field
- Added Long Text in `after_rename` condition , and reverted https://github.com/frappe/erpnext/pull/22627


nabinhait (approved):

marination:

PR #22755, opened 2020-07-21
fix: Tax amounts in HSN Wise Outward summary

deepeshgarg007:
If same item is added multiple times in an Invoice, Total Tax was added for each item which resulted in a mismatch between taxes in GST Itemised Sales Register and HSN wise Outward Summary


deepeshgarg007:
@Mergifyio backport develop

PR #22765, opened 2020-07-21
feat: Provision to make RFQ against Opportunity

anupamvs:
Added a button in to create Request For Quotation if the Opportunity is with items-
rfq

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.js
+++ b/erpnext/crm/doctype/opportunity/opportunity.js
@@ -76,6 +76,11 @@ frappe.ui.form.on("Opportunity", {
 					function() {
 						frm.trigger("make_supplier_quotation")
 					}, __('Create'));
+
+				frm.add_custom_button(__('Request For Quotation'),
+					function() {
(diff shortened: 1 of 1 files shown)


erpnext/crm/doctype/opportunity/opportunity.js
```diff
@@ -126,6 +131,13 @@ frappe.ui.form.on("Opportunity", {
 		})
 	},
 
+	make_request_for_quotation: function(frm) {
+		frappe.model.open_mapped_doc({
+			method: "erpnext.crm.doctype.opportunity.opportunity.make_request_for_quotation",
+			frm: cur_frm
```
deepeshgarg007: No need of `cur_frm`. You can use `frm` since you have it available. `cur_frm` was old pattern. Change above as well, wherever applicable

anupamvs: @deepeshgarg007 Done!

deepeshgarg007 (approved):

PR #22778, opened 2020-07-22
feat: Export CRM dashboard

anupamvs:
- Deleted dashboard_fixtures.py for CRM.
- Exported CRM Dashboard


nabinhait (dismissed):
In all the charts and number cards add dynamic filters for company

anupamvs:
@nabinhait Added!

nabinhait (approved):

nanoharsha:
@anupamvs could you share what you mean by exporting of dashboard . As we use the CRM functionality I also wanted to check on this.

anupamvs:
@nanoharsha These are the standard JSON generated for the module dashboard. More like this make it easy to customize the dashboard. Like we don't have to write any code for module dashboard.

PR #22788, opened 2020-07-23
Update transaction.js

aakvatech:
Added frappe.flags.hide_serial_batch_dialog = true that was missing in version-12

This causes the dialog to appear even for non-serial items as the default is left as undefined and falls under the criteria to show the dialog.

In version-13 the flag is correctly set


nabinhait (approved):

nabinhait:

PR #22790, opened 2020-07-23
fix: Sales Order Status Fix

KanchanChauhan:
**Problem:**
Sometimes the Sales Order has 0 value as well when it is a free service or FOC item. Such Sales orders always show as `Submitted`, no relevant status. 

**Solution:**
Added the condition to Status Updater and Sales Order List.

Screen Shot 2020-07-23 at 4 49 42 PM


nabinhait (commented):
If a Sales Order has zero amount, system updates `per_billed` based on qty invoiced.

It is updated based on https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/status_updater.py#L311

This fix is not a correct one, closing it.

Please investigate why per_billed is not updated based on invoiced qty in your case.

PR #22792, opened 2020-07-23
fix(Education): descriptions not copied while creating Fees from Fee Structure

ruchamahabal:
While creating Fees from Fees Structure the descriptions for fee components are not copied.

**Before:**

desc-not-copied

**After:**

desc-copied


ruchamahabal:

marination (approved):

PR #22800, opened 2020-07-24
fix(Education): course wise assessment report labels

ruchamahabal:
- In course wise assessment report, the chart does not have names for datasets. It's difficult to comprehend at first glance what every dataset is for. Added names for datasets
- There is a column Total Score which is actually showing the Final Grade. Changed the label. 

**Before:**

course-wise-assessment-no-labels

**After:**

course-wise-assessment-result


ruchamahabal:

deepeshgarg007 (approved):

PR #22814, opened 2020-07-26
fix: Quotation lost reason options fix

anupamvs:
**Issue-** When we making a Quotation as lost in the modal the Link field is Opportunity Lost Reason instead of Quotation Lost Reason.
q1

**Fix-**
q2


erpnext/patches/v13_0/rename_lost_reason_detail.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+    if frappe.db.exists("DocType", "Opportunity Lost Reason Detail") and frappe.db.exists("DocType", "Quotation Lost Reason Detail") and frappe.db.exists("DocType", "Lost Reason Detail"):
```
nabinhait: why checking db.exists for "Opportunity Lost Reason Detail" and "Quotation Lost Reason Detail"? Those are new table, should be reloaded with `frappe.reload_doc`

nabinhait (changes requested):

nabinhait (approved):

PR #22821, opened 2020-07-27
fix: get_items from product bundle for purchase order

AfshanKhan:
The item_details dict that was being returned before was sending doctype as 'Purchase Order' which was replacing 'Purchase Order Item to Purchase Order' for 'd[doctype]' in buying.js file at the place shown in below image because of which render row function was placing heading values into the first row
Screenshot 2020-07-27 at 6 01 49 PM

Before:
zTI0XmV

After:
Screenshot 2020-07-27 at 5 59 09 PM

**Edit: (by @marination )**
After client side fix:
prod-bundle


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -217,7 +217,7 @@ def get_pricing_rule_for_item(args, price_list_rate=0, doc=None, for_validate=Fa
 		doc = frappe.get_doc(doc)
 
 	if (args.get('is_free_item') or
-		args.get("parenttype") == "Material Request"): return {}
+		args.get("parenttype") in ["Material Request", "Purchase Order"]): return {}
```
marination: this is not the best way to fix it since it wont fetch pricing rule for Purchase Order, and it needs to. Material Request does not give priority to fiscal matters, but Purchase Order does.
The issue is that the context of the parent form is being passed due to which the header gets copied. 
get_pricing_rule_for_item is also called via get_item_details when an item is set in the child row, but it doesnt break there becuase the context of the child row is passed perhaps.
Will have to find a solution on client side for this

marination:

PR #22823, opened 2020-07-27
feat: Color field version 12

AminovE99:

Add a `color` field to ERPNext "_Issue_" DocType.

> Explain the details for making this change. What existing problem does the pull request solve?

The color can be used manually by the Support team to define a "_meaning_" to a color and thus quickly sort them.

Since the field is simply called `color`, the Kanban board GUI will automatically pick it up and display it.

Since only a field is added to the Doctype, there is no impact on ERPNext other features.

> Screenshots/GIFs

issue_2

See example Kanban board with some colors manually defined on Issues:

image


marination:
@AminovE99 don't you think this increases overhead ? People involved in support a lot of the times are loaded as it is, there is also the extra work of adding the customer sometimes or changing the subject or changing status or setting priority. This just increases the clicks. Not to forget actually solving the support ticket too.
Also the "meaning" behind the colors will have to be discussed and decided internally and remembered, the colors aren't mapped to any state in the system for reference.
Maybe mapping colors to **Priority** and auto fetching it on setting priority would be a better workaround ?
I'm not sure this will be beneficial. Just my opinion

AminovE99:
@marination I understand that are you talking about. In my expirience I understand how much time requiring to fill one form. But idea is that We can automatically fill this color if it necessary for us.

cromain:
@marination I agree mapping colors to priority is one of the best use of this feature.
It's a technical requirement which allows, if needed, such an automatic mapping. I don't think it adds extra work for people doing support. This feature is very useful to add visual information on Kanban, so I can display issues by status, and have a red flag automatically added if priority is high for example. (or whatever I configured) In fact this can allow user doing support listing issues faster.

AminovE99:
@nabinhait  can you assign for yourself this pull request?

AminovE99:
@nabinhait Can you review it?

Rolandinovich:
@nabinhait Good feature. It will help me in future

rohitwaghchaure:
Changing status never change the color then what is the use to add it in the core, you can add the custom field for your requirement.

PR #22829, opened 2020-07-28
fix: validate check out and check in time for inpatient occupancy

ruchamahabal:
Inpatient Record does not error check if Check out time is less than Check in time.

Before:

check-in

After:

checkout


erpnext/healthcare/doctype/inpatient_record/inpatient_record.py
```diff
@@ -30,6 +30,11 @@ def validate_dates(self):
 			(getdate(self.discharge_ordered_date) < getdate(self.scheduled_date)):
 			frappe.throw(_('Expected and Discharge dates cannot be less than Admission Schedule date'))
 
+		for entry in self.inpatient_occupancies:
+			if entry.check_in and entry.check_out and \
+				get_datetime(entry.check_in) > get_datetime(entry.check_out):
+				frappe.throw(_('Check Out datetime cannot be less than Check In datetime for entry #{0}').format(entry.idx))
```
marination:
```suggestion
				frappe.throw(_('Row #{0}: Check Out datetime cannot be less than Check In datetime').format(entry.idx))
```
Just to keep it standard throughout

marination (approved):

PR #22838, opened 2020-07-29
fix: cannot change customer fields if credit exhausted

nextchamp-saqib:
To replicate:
- Set a credit limit for a customer
- Set Administrator or any other user as Credit Controller in Accounts Setting
- Make sales invoice greater than the credit limit for the customer
- Now system won't allow changing Tax ID or any other field for that customer. 
 Screenshot 2020-08-20 at 11 27 17 AM

### DIFF
--- a/erpnext/selling/doctype/customer/customer.py
+++ b/erpnext/selling/doctype/customer/customer.py
@@ -184,6 +184,14 @@ class Customer(TransactionBase):
 	def validate_credit_limit_on_change(self):
 		if self.get("__islocal") or not self.credit_limits:
 			return
+		
+		last_credit_limits = [d.credit_limit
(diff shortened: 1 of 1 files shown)


nextchamp-saqib:

erpnext/selling/doctype/customer/customer.py
```diff
@@ -184,6 +184,14 @@ def validate_name_with_customer_group(self):
 	def validate_credit_limit_on_change(self):
 		if self.get("__islocal") or not self.credit_limits:
 			return
+		
+		last_credit_limits = [d.credit_limit
+			for d in frappe.db.get_all("Customer Credit Limit", filters={'parent': self.name}, fields=["credit_limit"], order_by="idx")]
```
nabinhait: Why is it based on idx? It should be checked based on company

nabinhait (changes requested):

nabinhait (approved):

PR #22843, opened 2020-07-29
fix: cannot submit value adjustment with custom dimension

nextchamp-saqib:
If a user has custom accounting dimension, then submitting asset value adjustment isn't possible.


nextchamp-saqib:

deepeshgarg007 (approved):

PR #22844, opened 2020-07-29
refactor: Crm reports cleanup

anupamvs:
**Lead Details**

image

**Lost Opportunity**

image


erpnext/crm/report/lead_details/lead_details.py
```diff
@@ -0,0 +1,164 @@
...
+			"fieldname": "phone",
+			"fieldtype": "Data",
+			"width": 120
+		},
+		{
+			"label": _("Notes"),
+			"fieldname": "notes",
+			"fieldtype": "Data",
+			"width": 120
+		},
```
marination: Not sure if adding Notes is useful. A report should have compact and to-the-point data. If its long form data, report is not the best way to show that

erpnext/crm/report/lead_details/lead_details.py
```diff
@@ -0,0 +1,164 @@
...
+def get_conditions(filters) :
+	conditions = []
+
+	if filters.get("territory"):
+		conditions.append("territory=%(territory)s")
+
+	if filters.get("status"):
+		conditions.append("status=%(status)s")
+
+	return " and {}".format(" and ".join(conditions)) if conditions else ""
```
marination:
```suggestion
		conditions.append("and territory=%(territory)s")

	if filters.get("status"):
		conditions.append("and status=%(status)s")

	return conditions if conditions else ""
```

erpnext/crm/report/lost_opportunity/lost_opportunity.py
```diff
@@ -0,0 +1,86 @@
...
+
+def get_data(filters):
+	return frappe.db.sql("""
+		SELECT
+			`tabOpportunity`.name,
+			`tabOpportunity`.opportunity_from,
+			`tabOpportunity`.party_name,
+			`tabOpportunity`.customer_name,
+			`tabOpportunity`.opportunity_type,
+			`tabOpportunity`.contact_by
```
marination: You can also fetch:
-  Lost Reason to analyse trends in why opportunities are lost
- Territory, to know where the most loss is happening
- Sales Stage to know where in the sales cycle improvement can happen
Just a suggestion

erpnext/crm/report/lost_opportunity/lost_opportunity.py
```diff
@@ -0,0 +1,86 @@
...
+	if filters.get("opportunity_from"):
+		conditions.append("opportunity_from=%(opportunity_from)s")
+
+	if filters.get("party_name"):
+		conditions.append("party_name=%(party_name)s")
+
+	if filters.get("contact_by"):
+		conditions.append("contact_by=%(contact_by)s")
+
+	return " and {}".format(" and ".join(conditions)) if conditions else ""
```
marination: here also, you prefix 'and' before each condition as parent query has conditions already. And `return conditions if conditions else "" `. Simpler the better

marination (changes requested):

erpnext/crm/report/lost_opportunity/lost_opportunity.py
anupamvs: @marination Done!

marination (dismissed):

anupamvs:
@marination docs https://github.com/frappe/erpnext_documentation/pull/133

marination (approved):

PR #22849, opened 2020-07-29
refactor: move datev-specific stuff to utils

barredterra:
There'a a report for germans, called "DATEV", that can be downloaded in a very special CSV format. All the CSV-specific code used to lived with the report.

Now the CSV-specific code gets his own place to live, in `regional.germany.utils.datev`. This way it can easily be reused for other reports (that will come in separate PRs).

Tests still pass, everything works like before and it still looks the same from the front-end.


deepeshgarg007 (approved):

barredterra:
Thanks, @deepeshgarg007

PR #22861, opened 2020-07-31
fix: show or hide section or attributes depending on other attributes…

AfshanKhan:
… for better UI

Before:
irfTxaD

After:
Screenshot 2020-07-31 at 9 27 08 PM


erpnext/stock/doctype/item/item.js
```diff
@@ -121,6 +121,10 @@ frappe.ui.form.on("Item", {
 			frm.set_df_property(fieldname, 'read_only', stock_exists);
 		});
 
+		['variants_section'].forEach((fieldname) => {
+			frm.set_df_property(fieldname, 'hidden', stock_exists);
+		});
```
marination:
```suggestion
		frm.set_df_property('variants_section', 'hidden', stock_exists);
```

erpnext/stock/doctype/item/item.js
```diff
@@ -121,6 +121,8 @@ frappe.ui.form.on("Item", {
 			frm.set_df_property(fieldname, 'read_only', stock_exists);
 		});
 
+		frm.set_df_property('variants_section', 'hidden', stock_exists);
```
marination: Missed in this in the last review. This will hide variants section in a variant too if stock exists, and its important for it to be visible there. I think just make the checkbox read only is best

AfshanKhan: ohh okay

marination (approved):

PR #22863, opened 2020-08-01
fix: Import pycountry and taxjar only if taxjar is enabled

deepeshgarg007:


sagarvora:
@deepeshgarg007 This is breaking the integration as the module `taxjar` is used in more than one function. Also, why not import it outside when it is already being installed in `requirements.txt`? Please revert / fix.

PR #22869, opened 2020-08-02
feat: uom in update items for SO & PO

nextchamp-saqib:
Update `uom` after submission of sales order and purchase order.

UOM field will be visible in expanded view

Screenshot 2020-07-29 at 7 14 48 PM


erpnext/controllers/accounts_controller.py
```diff
@@ -1290,6 +1292,9 @@ def validate_quantity(child_item, d):
 				child_item.conversion_factor = 1
 			else:
 				child_item.conversion_factor = flt(d.get('conversion_factor'))
+		
+		if d.get("uom"):
+			child_item.uom = d.get("uom")
```
marination, line 1320: - Fix: _Supplied Items_ table in Purchase order (subcontracting) does not get updated on changing UOM and conversion factor
- Add: Test for multi-uom in SO/PO (check if packed items is updated in SO and if supplied items is updated in PO)

erpnext/controllers/accounts_controller.py
```diff
@@ -1290,6 +1294,11 @@ def validate_quantity(child_item, d):
...
+			conversion_factor = flt(get_conversion_factor(child_item.item_code, child_item.uom).get("conversion_factor"))
+			child_item.conversion_factor = flt(d.get('conversion_factor')) or conversion_factor
```
nextchamp-saqib, line 1322: which conversion factor should have precedence?

erpnext/buying/doctype/purchase_order/test_purchase_order.py
```diff
@@ -202,6 +202,25 @@ def test_update_child_qty_rate_perm(self):
...
+			'uom': "_Test UOM 1",
+			'conversion_factor': 2,
+			'docname': po.get("items")[0].name
+		}])
+		update_child_qty_rate('Purchase Order', trans_item, po.name)
+		po.reload()
+
+		total_reqd_qty_after_change = sum([d.get("required_qty") for d in po.as_dict().get("supplied_items")])
+
+		self.assertEqual(total_reqd_qty_with_changed_uom, 2 * total_reqd_qty)
```
nextchamp-saqib:
```suggestion
		self.assertEqual(total_reqd_qty_after_change, 2 * total_reqd_qty)
```

erpnext/controllers/accounts_controller.py
marination: @nextchamp-saqib the one set in trans_items(set by user) should have precedence as conversion factor was added for that very reason.
Although on the ux end if you want to, we could fetch and set conversion factor **on changing uom only**, after that if they want to set a custom conversion factor they can. 
Because we know of only 1 use case where conversion factor changes all the time. 90% of the time it will change with uom mostly, so it will save such users extra data entry.
Just a suggestion

nextchamp-saqib: Just added the conversion factor fetch on UOM change

erpnext/public/js/utils.js
```diff
@@ -466,6 +466,33 @@ erpnext.utils.update_child_items = function(opts) {
...
+				callback: r => {
+					if(!r.exc) {
+						if (this.doc.conversion_factor == r.message.conversion_factor) return;
+						
+						const docname = this.doc.docname;
+						dialog.fields_dict.trans_items.df.data.some(doc => {
+							if (doc.docname == docname) {
+								doc.conversion_factor = r.message.conversion_factor;
+								dialog.fields_dict.trans_items.grid.refresh();
+								return true
```
marination: semi colon (codacy)

marination:
@nextchamp-saqib just resolve codacy, tested the rest, LGTM

marination (dismissed):

erpnext/public/js/utils.js
nextchamp-saqib:
```suggestion
								return true;
```

PR #22871, opened 2020-08-02
feat: Naming series for Sales/Purchase Return

anupamvs:
**Issue -** The naming series for Sales return is same as the Sales invoice. It should be distinguishing!

**Fix-**
Introduce a default naming series for return entries. So, in Sales Invoice, Delivery Note, Purchase Receipt, and Purchase Invoice, there will 2 series by default.

- Sales Invoice: ACC-SINV-RET-.YYYY.-
- Delivery Note: MAT-DN-RET-.YYYY.-
- Purchase Invoice:  ACC-PINV-RET-.YYYY.-
- Purchase Receipt: MAT-PR-RET-.YYYY.-

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
@@ -9,6 +9,7 @@
  "field_order": [
   "title",
   "naming_series",
+  "return_naming_series",
   "supplier",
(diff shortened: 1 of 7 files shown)


nabinhait (changes requested):
Just add a series for return entry in naming_series field itself, don't introduce a new field. The existing users must have added series for return on their own, it will break if we add another field.

PR #22911, opened 2020-08-04
feat: validate if removed item attributes exist in variants

scmmishra:
When an attribute is removed from Item Template. It still stays in the variant and the website listing of that item breaks.

This PR adds a check for this.

image

**Note to Reviewers**
- Please check if at all this validation is in the correct flow/order
- Check if the validation conditions are correct
- Depends on https://github.com/frappe/frappe/pull/11190


scmmishra:
@nabinhait @marination the frappe counterpart of this PR has been merged. I'd be great if you could look at this

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,72 @@ def validate_has_variants(self):
...
+		# get all item variants
+		items = [item["name"] for item in frappe.get_all("Item", {"variant_of": self.name})]
+		items_string = ', '.join(items)
+
+		# fetch all attributes of these items
+		item_attributes = frappe.get_all(
+			"Item Variant Attribute",
+			filters={"parent": ["in", "({})".format(items_string)]},
+			fields=["attribute", "parent"]
+		)
```
marination: Since we know some attribute has been deleted at this point from the template, we can just get the deleted attributes via `old_doc_attributes.difference(own_attributes)`. Then just get all item variant attribute details of only deleted attributes
(line 838) Looping through every Item Variant Attribute will have a lot of repetition.

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,72 @@ def validate_has_variants(self):
...
+				<td>{0}</td>
+				<td>{1}</td>
+			</tr>""".format(title, body)
+
+		rows = ''
+		for docname, attr_list in not_included.items():
+			link = "<a href='#Form/Item/{0}'>{0}</a>".format(frappe.bold(_(docname)))
+			rows += table_row(link, body(attr_list))
+
+		error_description = _('The following attributes exist in variants but not in the template. You can either delete the attribute from the items or add the attribute to template.')
```
marination:
```suggestion
		error_description = _('The following attributes exist in Variants but not in the Template. You can either delete the Variants or keep the attribute(s) in template.')
```
Can't delete attributes from already formed variants

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,72 @@ def validate_has_variants(self):
...
+		message = """
+			<div>{0}</div><br>
+			<table class="table">
+				<thead>
+					<td>{1}</td>
+					<td>{2}</td>
+				</thead>
+				{3}
+			</table>
+		""".format(error_description, _('Items'), _('Attributes'), rows)
```
marination:
```suggestion
		""".format(error_description, _('Variant Items'), _('Attributes'), rows)
```

erpnext/stock/doctype/item/item.py
scmmishra: @marination I've made the requested changes, can you re-review

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,78 @@ def validate_has_variants(self):
 			if frappe.db.exists("Item", {"variant_of": self.name}):
 				frappe.throw(_("Item has variants."))
 
+	def validate_attributes_in_variants(self):
+		if not self.has_variants:
```
marination:
```suggestion
		if not self.has_variants or self.get("__islocal"):
```
unsaved items won't have old doc.attributes

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,78 @@ def validate_has_variants(self):
...
+
+		# get all deleted attributes
+		deleted_attribute = list(old_doc_attributes.difference(set(own_attributes)))
+
+		# fetch all attributes of these items
+		item_attributes = frappe.get_all(
+			"Item Variant Attribute",
+			filters={
+				"parent": ["in", "({})".format(items_string)],
+				"attribute": ["in", "({})".format(', '.join(deleted_attribute))]
```
marination:
```suggestion
				"parent": ["in", items],
				"attribute": ["in", deleted_attribute]
```
**in** has to be checked with a **list or tuple** not a string.
Screenshot 2020-08-24 at 12 42 37 PM

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,78 @@ def validate_has_variants(self):
...
+				<td>{0}</td>
+				<td>{1}</td>
+			</tr>""".format(title, body)
+
+		rows = ''
+		for docname, attr_list in not_included.items():
+			link = "<a href='#Form/Item/{0}'>{0}</a>".format(frappe.bold(_(docname)))
+			rows += table_row(link, body(attr_list))
+
+		error_description = _('The following attributes exist in Variants but not in the Template. You can either delete the Variants or keep the attribute(s) in template.')
```
marination:
```suggestion
		error_description = _('The following deleted attributes exist in Variants but not in the Template. You can either delete the Variants or keep the attribute(s) in template.')
```

erpnext/stock/doctype/item/item.py
```diff
@@ -806,6 +807,78 @@ def validate_has_variants(self):
...
+		# Is old_attrs is a subset of new ones
+		# that means we need not check any changes
+		if old_doc_attributes.issubset(set(own_attributes)):
+			return
+
+		from collections import defaultdict
+
+		# get all item variants
+		items = [item["name"] for item in frappe.get_all("Item", {"variant_of": self.name})]
+		items_string = ', '.join(items)
```
marination: can be removed (items_string)

marination (approved):

scmmishra:

PR #22923, opened 2020-08-05
fix: Allocated advance amount for multi-currency payment

deepeshgarg007:
Incorrect Allocated amount for multi-currency advance allocation

Before:
Allocated amount in USD
Screenshot 2020-08-05 at 7 47 49 PM

After:
Allocated amount in INR
Screenshot 2020-08-05 at 7 39 20 PM


erpnext/controllers/accounts_controller.py
```diff
@@ -479,7 +479,7 @@ def set_advances(self):
 			if d.against_order:
 				allocated_amount = flt(d.amount)
 			else:
-				amount = self.rounded_total or self.grand_total
+				amount = self.base_rounded_total or self.base_grand_total
```
nabinhait, line 482:
```suggestion
				if self.party_account_currency == self.company_currency:
								amount = self.base_rounded_total or self.base_grand_total
				else:
								amount = self.rounded_total or self.grand_total
```

erpnext/controllers/accounts_controller.py
```diff
@@ -479,7 +479,19 @@ def set_advances(self):
...
+					party_type, party = self.get_party()
+
+					if party_type and party:
+						party_account_currency = get_party_account_currency(party_type, party, self.company)
+
+				if party_account_currency == self.company_currency:
+					amount = self.get('base_rounded_total') or self.base_grand_total
+				else:
+					amount = self.get('rounded_total') or self.grand_total
+				amount = self.base_rounded_total or self.base_grand_total
```
nabinhait: remove the line

erpnext/controllers/accounts_controller.py
```diff
@@ -479,7 +479,19 @@ def set_advances(self):
...
 				allocated_amount = flt(d.amount)
 			else:
-				amount = self.rounded_total or self.grand_total
+				party_account_currency = self.get('party_account_currency')
+
+				if not party_account_currency:
+					party_type, party = self.get_party()
+
+					if party_type and party:
+						party_account_currency = get_party_account_currency(party_type, party, self.company)
```
nabinhait: this is not required for set_advances method. Because the function is only for SI and PI and in both doctype party_account_currency field exists.

nabinhait (changes requested):

nabinhait (approved):

PR #22927, opened 2020-08-06
fix: Error due to comma in Pricing rule name

deepeshgarg007:
Backport Of : https://github.com/frappe/erpnext/pull/22741


marination:
@deepeshgarg007 missed this one https://github.com/frappe/erpnext/blob/97e96cec517a01ae2b46e16c6a90ee5c0af158bf/erpnext/controllers/taxes_and_totals.py#L596

marination (approved):

PR #22933, opened 2020-08-06
fix: show or hide section or attributes depending on other attributes…

AfshanKhan:
… for better UI

This is to backport https://github.com/frappe/erpnext/pull/22861 on to version-12-hotfix

### DIFF
--- a/erpnext/stock/doctype/item/item.json
+++ b/erpnext/stock/doctype/item/item.json
@@ -498,7 +498,7 @@
    "oldfieldtype": "Select"
   },
   {
-   "depends_on": "eval:doc.is_stock_item || doc.is_fixed_asset",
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/item/item.json
```diff
@@ -1059,7 +1059,7 @@
  "idx": 2,
  "image_field": "image",
  "max_attachments": 1,
- "modified": "2020-07-01 12:43:10.656530",
+ "modified": "2020-08-06 16:03:25.085360",
  "modified_by": "Administrator",
  "module": "Stock",
  "name": "Item",
```
marination, line 1066: depends on: batch_no for retain sample, missed it from the ported PR

marination (approved):

PR #22936, opened 2020-08-06
fix: Item configure

marination:
Port of https://github.com/frappe/erpnext/pull/22934 and https://github.com/frappe/erpnext/pull/22935

Another Fix: The product configurator would not show "Add to Cart" if product info is null. This PR overrides the effects of new changes in the configurator and shows the link anyway


scmmishra:
@Mergifyio backport version-12-hotfix version-13-beta version-13-beta-pre-release v12-pre-release

scmmishra (approved):

PR #22943, opened 2020-08-06
fix: Add help link in navbar settings

deepeshgarg007:
A new DocType "Navbar Settings"  has been introduced in Frappe using which user can update the app logo and add link under Settings and Help Dropdown

So removed hardcoded links under help section and added a patch to add those linked in Navbar Settings
On new site installation, standard help links will be added on setup
    
Depends On: https://github.com/frappe/frappe/pull/11213


erpnext/setup/install.py
```diff
@@ -103,3 +104,45 @@ def add_company_to_session_defaults():
...
+
+	for item in current_nabvar_items:
+		navbar_settings.append('help_dropdown', {
+			'item_label': item.item_label,
+			'item_type': item.item_type,
+			'route': item.route,
+			'action': item.action,
+			'is_standard': item.is_standard,
+			'hidden': item.hidden
+		})
```
barredterra, line 147:
```suggestion
	navbar_settings.extend('help_dropdown', erpnext_navbar_items)
	navbar_settings.extend('help_dropdown', current_nabvar_items)
```

deepeshgarg007: > 

I tried your suggestion and it seems to mess up the ordering of the items in the dropdown. I am not exactly sure why but my guess is if we simply extend, other doc properties like `idx` etc. are still there which may be the reason of mess up the ordering

barredterra: Weird... I think `extend` is implemented just like your for-loop.  🤔

erpnext/patches/v13_0/add_standard_navbar_items.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import unicode_literals
+import frappe
```
barredterra:
```suggestion
# import frappe
```

barredterra (changes requested):

erpnext/setup/install.py
deepeshgarg007: Yes it is, but in my for loop above I am not passing any other doc attributes like `idx` etc. But while extending with `current_navbar_items` all those are also passed and it messes up the ordering

barredterra (dismissed):

PR #22944, opened 2020-08-06
fix: Opportunity Status fix

anupamvs:
**Issue -** On various doc event (validate, on_submit, on_cancel) of Quotation, It always sets the opportunity status to **null**.

**Fix -** Opportunity status changes when quotation doc event/status changes.

| Quoatation Status/Event  | Opportunity Status |
| ------------- | ------------- |
| on_submit  | Quotation  |
| Lost  | Lost  | 
| on_cancel  | Open  |


anupamvs:
@rohitwaghchaure can you review this?

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -95,7 +94,7 @@ def on_submit(self):
 			self.company, self.base_grand_total, self)
 
 		#update enquiry status
-		self.update_opportunity()
+		self.update_opportunity('Quotation')
```
rmehta, line 96: what if Quotation is "Lost"? should this not be fixed via `status_updater.py`?

rmehta (changes requested):

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -50,20 +49,20 @@ def set_customer_name(self):
...
 
-	def update_opportunity_status(self, opportunity=None):
+	def update_opportunity_status(self, status, opportunity=None):
 		if not opportunity:
 			opportunity = self.opportunity
 
 		opp = frappe.get_doc("Opportunity", opportunity)
-		opp.status = None
+		opp.set_status(status=status)
 		opp.set_status(update=True)
```
rohitwaghchaure: why set_status called two times?

rohitwaghchaure (approved):

PR #22945, opened 2020-08-07
feat: allow Item image alt

scmmishra:
This PR adds a feature to add `alt` value to product image.

Field
image

Result
*Image alt shown as "Name"*
image


scmmishra:

erpnext/stock/doctype/item/item.json
```diff
@@ -1054,15 +1055,21 @@
    "fieldtype": "Data",
    "label": "Default Manufacturer Part No",
    "read_only": 1
+  },
+  {
+   "fieldname": "website_image_alt",
+   "fieldtype": "Data",
+   "label": "Image Alt Text"
```
marination:
```suggestion
   "label": "Image Alternative Text"
```
A layperson should also understand it

scmmishra: `Alt` is a generally accepted term. "Image Alternative Text" is not a clear label as well. Lets' make it "Image Description" ?

marination (approved):

PR #22956, opened 2020-08-08
fix: escape apostrophe in company name if exist

AfshanKhan:
If there was an apostrophe in the companies name the query for accounts payable summary used to break. So escaped it.


deepeshgarg007 (commented):
@AfshanKhan already a PR open for this https://github.com/frappe/erpnext/pull/22947

ruchamahabal:
Had missed the quotes

ruchamahabal (approved):

Issue #22957, opened 2020-08-08
Healthcare: InPatient transfer checkIn time validation


mujeerhashmi:
@ruchamahabal @akurungadam 
## Description of the issue
The InPatient transfer should not allow checkin time to be greater than `now`. This validation is missing.

ruchamahabal:
fixed by @akurungadam in https://github.com/frappe/erpnext/pull/22958
Closing

PR #22958, opened 2020-08-08
fix: Inpatient Record - Transfer check-in time validation

akurungadam:
Issue: https://github.com/frappe/erpnext/issues/22957

Added validation, client side only

Screen Shot 2020-08-08 at 20 21 06

### DIFF
--- a/erpnext/healthcare/doctype/inpatient_record/inpatient_record.js
+++ b/erpnext/healthcare/doctype/inpatient_record/inpatient_record.js
@@ -150,7 +150,7 @@ let transfer_patient_dialog = function(frm) {
 			if(check_in > frappe.datetime.now_datetime()){
 				frappe.msgprint({
 					title: __('Not Allowed'),
-					message: __('Check-in time should not be later than current time'),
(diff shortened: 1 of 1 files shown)


erpnext/healthcare/doctype/inpatient_record/inpatient_record.js
```diff
@@ -147,7 +147,12 @@ let transfer_patient_dialog = function(frm) {
 			if(dialog.get_value('service_unit')){
 				service_unit = dialog.get_value('service_unit');
 			}
-			if(!check_in){
+			if(check_in > frappe.datetime.now_datetime()){
+				frappe.msgprint({
+					title: __('Not Allowed'),
+					message: __('Check-in time should not be later than current time'),
```
ruchamahabal:
```suggestion
					message: __('Check-in time cannot be greater than the current time'),
```
To be consistent with the error messages

ruchamahabal (changes requested):

ruchamahabal (approved):

PR #22966, opened 2020-08-09
fix(membership): currency should be a link

barredterra:
Currency should be a Link field. If there's a valid reason memberships can only be in USD or INR, restricting the query might be a better approach.


ruchamahabal (approved):

erpnext/non_profit/doctype/membership/membership.json
```diff
@@ -90,9 +90,9 @@
   },
   {
    "fieldname": "currency",
-   "fieldtype": "Select",
+   "fieldtype": "Link",
    "label": "Currency",
-   "options": "USD\nINR"
+   "options": "Currency"
   },
```
ruchamahabal, line 96: @barredterra You will also need to change the modified timestamp in this case.

ruchamahabal (changes requested):

deepeshgarg007 (approved):

PR #22991, opened 2020-08-10
fix: HR desk page

Anurag810:
Removed payroll from HR desk page


erpnext/hr/desk_page/hr/hr.json
```diff
@@ -64,8 +64,8 @@
  "category": "Modules",
  "charts": [
   {
-   "chart_name": "Attendance Count",
-   "label": "Attendance Count"
+   "chart_name": "Outgoing Salary",
```
ruchamahabal: The Outgoing Salary chart is a part of the Payroll Module so it should be the default chart of Payroll and not HR Desk Page, right?

ruchamahabal (approved):

PR #22992, opened 2020-08-10
fix: handled condition if staffing isn't created

AfshanKhan:
When the staffing plan wasn't created DB returns following:
Screenshot 2020-08-10 at 11 05 12 PM

So added a check in the condition to check for None of the vacancies

Also added a condition to display error msg if there is no staffing exist. So if there is no staffing for a particular Designation this is how the error msg would look.
Screenshot 2020-08-10 at 11 04 14 PM


erpnext/hr/doctype/job_offer/job_offer.py
```diff
@@ -24,8 +24,8 @@ def validate_vacancies(self):
 		check_vacancies = frappe.get_single("HR Settings").check_vacancies
 		if staffing_plan and check_vacancies:
 			job_offers = self.get_job_offer(staffing_plan.from_date, staffing_plan.to_date)
-			if staffing_plan.vacancies - len(job_offers) <= 0:
-				frappe.throw(_("There are no vacancies under staffing plan {0}").format(frappe.bold(get_link_to_form("Staffing Plan", staffing_plan.parent))))
+			if staffing_plan.vacancies is None or staffing_plan.vacancies - len(job_offers) <= 0:
```
marination:
```suggestion
			if not staffing_plan.get("vacancies") or staffing_plan.vacancies - len(job_offers) <= 0:
```

erpnext/hr/doctype/job_offer/job_offer.py
```diff
@@ -24,8 +24,8 @@ def validate_vacancies(self):
...
+				frappe.throw(_("There are no vacancies under staffing plan {0}").format(frappe.bold(get_link_to_form("Staffing Plan", staffing_plan.parent) if staffing_plan.parent else 'for ' + self.designation)))
```
marination: Could be broken into two lines for readability. Define what to substitute in a variable and then use it in the next line

AfshanKhan:

marination (dismissed):

erpnext/hr/doctype/job_offer/job_offer.py
```diff
@@ -24,8 +24,13 @@ def validate_vacancies(self):
 		check_vacancies = frappe.get_single("HR Settings").check_vacancies
 		if staffing_plan and check_vacancies:
 			job_offers = self.get_job_offer(staffing_plan.from_date, staffing_plan.to_date)
-			if staffing_plan.vacancies - len(job_offers) <= 0:
-				frappe.throw(_("There are no vacancies under staffing plan {0}").format(frappe.bold(get_link_to_form("Staffing Plan", staffing_plan.parent))))
+
+			if not staffing_plan.get("vacancies") or staffing_plan.vacancies - len(job_offers) <= 0:
+				error_variable = 'for ' + frappe.bold(self.designation)
+				if staffing_plan.parent:
```
marination:
```suggestion
				if staffing_plan.get("parent"):
```
If no vacancies high chances are that parent is None too (going by your screenshot). Try to use get in most places where it is uncertain if the attribute exists

PR #23014, opened 2020-08-12
feat(Accounts): Add Condition Field In Pricing Rule

abhishekbalam:
Condition field to add conditional expressions like in [Assignment Rule](https://docs.erpnext.com/docs/user/manual/en/automation/assignment-rule):
`territory == 'India' and order_type != 'Maintenance'`

image

All fields available in the transaction (like Sales Invoice, Sales Order etc) are available
as their fieldnames directly for use in the conditions.


abhishekbalam:
@rohitwaghchaure @deepeshgarg007 

Docs Link:
https://github.com/frappe/erpnext_documentation/pull/130

erpnext/accounts/doctype/pricing_rule/pricing_rule.json
```diff
@@ -12,6 +13,7 @@
   "apply_on",
   "price_or_product_discount",
   "warehouse",
+  "condition",
```
rohitwaghchaure: should be in a different section

erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -51,6 +53,24 @@ def get_pricing_rules(args, doc=None):
...
+def filter_pricing_rule_based_on_condition(pricing_rules, doc=None):
+	filtered_pricing_rules = []
+	if doc:
+		for pricing_rule in pricing_rules:
+			if pricing_rule.condition:
+				try:
+					if frappe.safe_eval(pricing_rule.condition, None, doc.as_dict()):
+						filtered_pricing_rules.append(pricing_rule)
+				except Exception as e:
+					frappe.msgprint(_("Pricing Rule - " + pricing_rule.name + " - <b>condition</b> field error:<br>" + \
```
rohitwaghchaure: No need of msgprint, if condition is not satisfying then don't append

erpnext/accounts/doctype/pricing_rule/pricing_rule.json
abhishekbalam: Moved to "Advanced Settings" section.
Is that fine?

abhishekbalam:
Will write and push tests today itself.

deepeshgarg007:
@abhishekbalam any updates on this?

abhishekbalam:
@deepeshgarg007 
Will do this asap, I'm having bandwidth issues.

abhishekbalam:
@rohitwaghchaure @deepeshgarg007 

Test for condition field added.

image

abhishekbalam:
Hi, requesting review since this PR is open for too long:
@nabinhait @rohitwaghchaure @deepeshgarg007

deepeshgarg007 (approved):

PR #23019, opened 2020-08-13
feat: enable total row in Gross Profit Report

prssanna:
Screenshot 2020-08-13 at 11 34 49 AM


scmmishra:
Hey @prssanna 

In a recent [PR](https://github.com/frappe/frappe/pull/11259) we add this field to the ignore list. This means only new site creation will have this checkbox marked. We can write a patch to enable it across sites, not sure if it's necessary.

PR #23024, opened 2020-08-13
feat: validate workflow before so/po update items

nextchamp-saqib:
As of now, all users that have the "Submit" option for POs under role permissions manager are able to update the POs.

However, if we have a workflow process where we have different financial thresholds for the person's purchase authority.

I.e. Role 1 can approve up to 5,000. Role 2 can approve up to 10,000

When the user updates the PO, their role restrictions as set-up in work flow process need to apply as well.

This PR validates workflow conditions and allowed roles before updating items in submitted/approved sales orders & purchase orders


erpnext/controllers/accounts_controller.py
```diff
@@ -1224,15 +1225,38 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
...
+		workflow_doc = frappe.get_doc("Workflow", workflow)
+		current_state = doc.get(workflow_doc.workflow_state_field)
+		roles = frappe.get_roles()
+
+		transitions = []
+		for transition in workflow_doc.transitions:
+			if transition.next_state == current_state and transition.allowed in roles:
+				if not is_transition_condition_satisfied(transition, doc):
+					continue
+				transitions.append(transition.as_dict())
```
marination, line 1248: logic seems fine , but the transition condition is checked against an unsaved doc. 
So if user who can only approve upto **Rs.100**, uses Update Items and makes total **Rs.200**, `is_transition_condition_satisfied` is True as it considers grand total as 100 (previous version of doc as it is unsaved).
Think we need to check workflow after `save()`

Normally it wont give such a user the Approve/Reject Action button after user changes grand total to **Rs.200** and clicks on save. Hence user can't submit

PR #23027, opened 2020-08-13
fix: Item GL entries for purchase invoice

deepeshgarg007:
Currently for a Purchase Invoice with zero grand total no accounting entries are posted and the invoice is completely off the books since no GL entries are posted.

Consider the below scenario:

Screenshot 2020-08-13 at 8 45 09 PM

First row has the actual item and the second row has a service item which is used to allocate the downpayment, so the grand total of the invoice is zero.

Even though expense head for both the rows are different no accounting entry is passed.
Screenshot 2020-08-13 at 8 45 22 PM

Screenshot 2020-08-13 at 8 45 36 PM

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -400,8 +400,6 @@ class PurchaseInvoice(BuyingController):
 		update_linked_doc(self.doctype, self.name, self.inter_company_invoice_reference)
 
 	def make_gl_entries(self, gl_entries=None):
-		if not self.grand_total:
-			return
 		if not gl_entries:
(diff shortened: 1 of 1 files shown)


nabinhait:
@deepeshgarg007 Add tests

nabinhait (changes requested):
Check this line https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py#L506

If the item is sold in zero rate with update_stock checked, the system should still post GL entries related COGS and Stock-in-hand. So, `if flt(item.base_net_amount):` condition should be removed.

But if the same condition should be added for https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py#L587

nabinhait (approved):

PR #23029, opened 2020-08-13
feat: Add page builder to Item doctype for item specifications on website

deepeshgarg007:
## Added webpage template in Item DocType for designing Item Specification pages
image

## Item Page Template
image

- Full width option
- Provision to copy templates from Item Page Template


erpnext/public/js/utils.js
```diff
@@ -304,6 +304,43 @@ $.extend(erpnext.utils, {
 			}
 			frappe.ui.form.make_quick_entry(doctype, null, null, new_doc);
 		});
+	},
+
+	add_web_template: function(frm, cdt, cdn) {
+		let row = frm.selected_doc;
+		frappe.model.with_doc("Web Template", row.web_template).then((doc) => {
```
netchampfaris, line 311: This duplication is not right. There is an ongoing PR in frappe that deals with this duplication.

https://github.com/frappe/frappe/pull/11285/files#diff-a7f86b0d7ba77ad1dc83597f12ece957R53

We can use that approach once that is merged.

erpnext/stock/doctype/item/item.js
```diff
@@ -216,6 +216,28 @@ frappe.ui.form.on("Item", {
 
 	set_meta_tags(frm) {
 		frappe.utils.set_meta_tag(frm.doc.route);
+	},
+
+	copy_from_item_page_template: function(frm) {
```
netchampfaris: Some thoughts on templating Item Pages:

This Page Builder template implementation is not specific to Item so it doesn't make sense to make it exclusive for Item.
I don't think it makes sense to make "templates" out of templates. Page Builder templates are templates by themselves. If you are making similar item pages, you can make one item page and just duplicate it.
Also, Page Builder templates should allow dynamic values, and at least provide the Item document as the context.
Now that I think about it, it makes sense to make a Print Format Builder alternative for any type of WebsiteGenerator document but for it's web view. This needs more thought.

erpnext/stock/doctype/item/item.py
```diff
@@ -470,6 +472,13 @@ def set_shopping_cart_data(self, context):
 		from erpnext.shopping_cart.product_info import get_product_info_for_website
 		context.shopping_cart = get_product_info_for_website(self.name, skip_quotation_creation=True)
 
+	def set_page_blocks(self, context):
```
netchampfaris, line 475: I dont like this duplication at all. Atleast, we should extract the Page Builder parts of the code and put it in a class and share that class among "Page Builder"able doctypes.

netchampfaris (commented):
Let me work on the design for a generic Page Builder feature for every WebsiteGenerator doctype.

harshit-30:
@deepeshgarg007 when is this going to get merged.

PR #23034, opened 2020-08-14
fix(General Ledger): include Accounting Dimension columns in report

### LINKED ISSUE #23033: GL Entries with different accounting dimension values being conflated
## Description of the issue

Using ERPNext for a not-for-profit, we have defined an Accounting Dimension called "Fund" (based on a "Fund" DocType). In relatively frequent journal entries,  we have to transfer liabilities from one Fund to another, using a liability account called "Due To/From" (credited in the owing Fund, debited in the receiving Fund). However, when we click on the "View > Ledger" button for such an entry, the "Due To/From" lines have been collapsed into a single line, making it impossible to see at least one of the Funds involved in the transfer.

This is running the latest develop version of ERPNext (and Frappe), pulled as of today.

1. Create a custom Accounting Dimension with at least two values.
2. Create a journal entry with two lines that have the same account but different values for the accounting dimension.
3. Save and submit the journal entry.
4. Click on the "View > Ledger" button for the entry.

### Observed result

All of the lines with the same account have been combined into one, which (if you add the accounting dimension as a column) shows one of the values for the accounting dimension chosen apparently at random. The other value(s) of the accounting dimension in the journal entry are not visible in the ledger entry. This display is misleading, as it seems to impute all of the debits and credits in this account to just one accounting dimension, which may not be what the journal entry specified (and is not correct in our use case).

### Expected result

There should be separate lines for each value of the accounting dimension, so that all values of the accounting dimension involved in the journal entry can be read from the ledger entry.

N/A

## Additional information

The difficulty is connected with the fact that the accounting dimension columns are not displayed by default in the General Ledger report, so indeed that report has no basis by which the lines that differ only by the value of an Accounting Dimension could be distinguished. Corresponding to the absence of the accounting dimension columns, the values of the accounting dimensions are not included in the key used to consolidate lines of the report when the "Group by" value is "Group by Voucher (Consolidated)" -- which is the default in the view reached by the "View > Ledger" button. I will shortly submit a pull request which allows (and defaults to) the accounting dimensions columns to be included in the General Ledger report and be included in the consolidation key for this grouping.

gwhitney:
Prior to this commit, custom Accounting Dimensions are not
  (by default) included in a General Ledger report. Moreover,
  in part because they are not included, they are not used
  to distinguish GL Entry lines in the same voucher with the
  same Account when the "Group By" value is
  "Group by Voucher (Consolidated)". This situation leads to
  lines with different values for the accounting dimension
  to be ganged together in the General Ledger view of a Journal
  entry, making the entry difficult for an accountant to read.

  This commit alleviates the problem by adding a checkbox
  to control whether Accounting Dimension columns are included
  in the General Ledger report. When they are included, they
  are also considered as part of the key that distinguishes
  lines in the "Group by Voucher (Consolidated)" mode.

  The checkbox defaults to "checked" (true) because it seems
  to me that including the columns would generally be
  preferable to clients using custom Accounting Dimensions. I
  left it as a checkbox providing a very easy ability to
  return to the prior behavior because I wanted to be
  conservative as far as not eliminating/altering current
  behavior for users who prefer the current behavior.

  However, if the maintainers feel that the new behavior with
  the option checked is essentially universally preferable,
  I am happy to update the pull request to simply include
  custom Accounting Dimensions, and distinguish GL Entry Lines
  with different values for an Accounting Dimension, as
  the standard behavior. Please advise if you would prefer
  this.

  Resolves #23033.


gwhitney:
I looked at the Travis failure, but it is in frappe.patches.v13_0.rename_notification_fields in test_site, which is not part of my pull request (or touched by it in any way).

erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -146,6 +146,12 @@ frappe.query_reports["General Ledger"] = {
 				return frappe.db.get_link_options('Project', txt);
 			}
 		},
+		{
+			"fieldname": "include_dims",
+			"label": __("Include Accounting Dimensions in Ledger"),
```
deepeshgarg007:
```suggestion
			"label": __("Show Accounting Dimensions"),
```

erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -146,6 +146,12 @@ frappe.query_reports["General Ledger"] = {
 				return frappe.db.get_link_options('Project', txt);
 			}
 		},
+		{
+			"fieldname": "include_dims",
```
deepeshgarg007:
```suggestion
			"fieldname": "include_dimensions",
```
Update everywhere, try to avoid abbreviations and shorthand if possible for better code understanding

erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -146,6 +146,12 @@ frappe.query_reports["General Ledger"] = {
...
+			"label": __("Include Accounting Dimensions in Ledger"),
+			"fieldtype": "Check",
+			"default": 1
```
deepeshgarg007:
```suggestion
			"default": 0
```
Should be unchecked by default

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -128,6 +128,10 @@ def get_gl_entries(filters):
 		filters['company_fb'] = frappe.db.get_value("Company",
 			filters.get("company"), 'default_finance_book')
 
+	dim_fields = ""
```
deepeshgarg007:
```suggestion
	dimension_fields = "cost_center"
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -128,6 +128,10 @@ def get_gl_entries(filters):
...
+	if filters.get("include_dims"):
+		dim_fields = f", {', '.join(get_accounting_dimensions())}"
```
deepeshgarg007:
```suggestion
		dimension_fields = ', '.join([dimension_fields] + get_accounting_dimensions(as_list=1))
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -141,7 +145,7 @@ def get_gl_entries(filters):
 			party_type,
 			party,
 			voucher_type,
-			voucher_no,
+			voucher_no{dim_fields},
```
deepeshgarg007:
```suggestion
			voucher_no, {dimension_fields},
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -160,21 +164,21 @@ def get_gl_entries(filters):
 		{conditions}
 		AND posting_date <= %(to_date)s
 		AND cost_center = DCC_allocation.parent
-		""".format(select_fields_with_percentage=select_fields_with_percentage, conditions=get_conditions(filters).replace("and cost_center in %(cost_center)s ", ''))
+		""".format(dim_fields=dim_fields,select_fields_with_percentage=select_fields_with_percentage, conditions=get_conditions(filters).replace("and cost_center in %(cost_center)s ", ''))
```
deepeshgarg007:
```suggestion
		""".format(dimension_fields=dimension_fields, select_fields_with_percentage=select_fields_with_percentage, conditions=get_conditions(filters).replace("and cost_center in %(cost_center)s ", ''))
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -160,21 +164,21 @@ def get_gl_entries(filters):
...
 
 	gl_entries = frappe.db.sql(
 		"""
 		select
 			name as gl_entry, posting_date, account, party_type, party,
-			voucher_type, voucher_no, cost_center, project,
+			voucher_type, voucher_no{dim_fields}, cost_center, project,
```
deepeshgarg007:
```suggestion
			voucher_type, voucher_no, {dimension_fields}, cost_center, project,
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -160,21 +164,21 @@ def get_gl_entries(filters):
...
 			against_voucher_type, against_voucher, account_currency,
 			remarks, against, is_opening, creation {select_fields}
 		from `tabGL Entry`
 		where company=%(company)s {conditions}
 		{distributed_cost_center_query}
 		{order_by_statement}
 		""".format(
-			select_fields=select_fields, conditions=get_conditions(filters), distributed_cost_center_query=distributed_cost_center_query,
+			dim_fields=dim_fields, select_fields=select_fields, conditions=get_conditions(filters), distributed_cost_center_query=distributed_cost_center_query,
```
deepeshgarg007:
```suggestion
			dimension_fields=dimension_fields, select_fields=select_fields, conditions=get_conditions(filters), distributed_cost_center_query=distributed_cost_center_query,
```

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -350,8 +354,12 @@ def update_value_in_dict(data, key, gle):
 			if filters.get("group_by") != _('Group by Voucher (Consolidated)'):
 				gle_map[gle.get(group_by)].entries.append(gle)
 			elif filters.get("group_by") == _('Group by Voucher (Consolidated)'):
-				key = (gle.get("voucher_type"), gle.get("voucher_no"),
-					gle.get("account"), gle.get("cost_center"))
+				keylist = [gle.get("voucher_type"), gle.get("voucher_no"), gle.get("account")]
+				if filters.get("include_dims"):
+					for dim in get_accounting_dimensions():
+						keylist.append(gle.get(dim))
+				keylist.append(gle.get("cost_center"))
```
deepeshgarg007, line 366: Any specific reason for removing cost center from above and appending here?

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -350,8 +354,12 @@ def update_value_in_dict(data, key, gle):
 			if filters.get("group_by") != _('Group by Voucher (Consolidated)'):
 				gle_map[gle.get(group_by)].entries.append(gle)
 			elif filters.get("group_by") == _('Group by Voucher (Consolidated)'):
-				key = (gle.get("voucher_type"), gle.get("voucher_no"),
-					gle.get("account"), gle.get("cost_center"))
+				keylist = [gle.get("voucher_type"), gle.get("voucher_no"), gle.get("account")]
+				if filters.get("include_dims"):
+					for dim in get_accounting_dimensions():
```
deepeshgarg007: Get accounting dimensions as list once and store in a variable. Use the same variable to loop here and generate `dimensions_fields` for queries

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -478,7 +486,19 @@ def get_columns(filters):
 			"options": "Project",
 			"fieldname": "project",
 			"width": 100
-		},
+		}
+	])
+
+	if filters.get("include_dims"):
```
deepeshgarg007:
```suggestion
	if filters.get("include_dimensions"):
```

deepeshgarg007 (changes requested):

erpnext/accounts/report/general_ledger/general_ledger.js
gwhitney: Got it.

erpnext/accounts/report/general_ledger/general_ledger.js
gwhitney: I am changing to "Consider Accounting Dimensions," as it's not simply a matter of showing the dimensions or not; the actual general ledger lines produced by the report change depending on whether or not this box is checked.

erpnext/accounts/report/general_ledger/general_ledger.py
gwhitney: Renamed the variable, but I don't see how it can be initialized to "cost_center". The existing code handles cost_center separately and quite specially. I think it would be great if ERPNext was streamlined to have cost center just precisely be an accounting dimension that is present by default, but I feel the reality is that the code is far from that now and I don't see how this small PR can make any meaningful step toward true integration of cost center as an accounting dimension.

erpnext/accounts/report/general_ledger/general_ledger.py
gwhitney: the dimension_fields string must have the same number of commas in it as it has fields, otherwise there is no way to interpolate it into the query without possibly introducing a spurious empty field. I can see how "voucher_no{dim_fields}" did not look nice in the format below, so I am moving the additional comma to the end of the dimension_fields string. Hopefully that will make the format look more readable.

gwhitney: Also, as_list=1 is the default for get_accounting_dimensions, so it would seem slightly odd to specify it. (Since the usual use is occurring, why put in a notation that might suggest to the code reader that the parameter is being supplied for a particular reason?)

erpnext/accounts/report/general_ledger/general_ledger.py
gwhitney: OK, I have merged the two calls to get_accounting_dimensions to a single one up in get_results, and passed the resulting accounting_dimensions list down through the call chain to the two places it is used.

erpnext/accounts/report/general_ledger/general_ledger.py
gwhitney: Only so that the order of the fields in the key will be the same as the order of the fields in the SQL query. Not sure that it makes any difference, but having a single standard order for using/querying the fields seemed reasonable as the least likely to cause issues if that list is used for anything else in future revisions of the code. If you'd like me to consolidate the cost center back into the first line where the key is set, just let me know and I will go ahead and do so.

erpnext/accounts/report/general_ledger/general_ledger.py
gwhitney: OK, I think that covers all of your comments. I will test again and force-push the revised PR shortly. Thanks for your feedback! -- Glen

PR #23031, opened 2020-08-14
fix: AttributeError: 'ProgramFee' object has no attribute 'course'

aakvatech:
Backport: https://github.com/frappe/erpnext/pull/19695

Error occurs while program enrollment on new student.

This happens on version 12 and not on version 13.

Version 12 uses
course_list = [course.course for course in program.get_all_children()]

where as version 13 uses
course_list = [course.course for course in program.courses]

erpnext/erpnext/education/doctype/program_enrollment/program_enrollment.py

> Please provide enough information so that others can review your pull request:

> Explain the **details** for making this change. What existing problem does the pull request solve?

> Screenshots/GIFs

PR #23036, opened 2020-08-14
Revert "fix: AttributeError: 'ProgramFee' object has no attribute 'course'"

aakvatech:
Reverts frappe/erpnext#23031


ruchamahabal:
@aakvatech I have already reverted it: https://github.com/frappe/erpnext/pull/23032
Closing this

PR #23043, opened 2020-08-14
fix: Leave application status fix

anupamvs:
Issue - After canceling the leave the status is not updated.
image

Fix -
Making Enable Allow On Submit on Status field. 
image


Anurag810 (changes requested):
Write a patch for old data having docstatus = 2 and status = 'Aprroved' and set status = Cancelled.

Anurag810 (approved):

PR #22671, opened 2020-07-13
feat: Material Request and Stock Entry Enhancement

anupamvs:
- Added a field to set **Default In Transit Warehouse**(Warehouse Type = Transit) in Company master.
- Create Material Request With Purpose **Material Transfer**. And make Stock Entry against the Material Request.
- In Stock Entry provided a Checkbox **Add to Transit**  when the Stock Entry Type is '**Material Transfer**'.
- On Submitted Stock Entry, there is a custom button '**End Transit**' to create a reverse stock entry.

stockentry enhancement

PR #23050, opened 2020-08-17
fix: deleting unused Stock Entry Type

anupamvs:
Added code to delete **Send to Warehouse** and **Receive at Warehouse** from exiting site. (ref #22671)


barredterra:
I think you should create a new patch, not add to an existing one. The existing one was used already on other instances and will not be executed again. (Correct me if i'm wrong)

rohitwaghchaure (approved):

PR #23054, opened 2020-08-17
fix: General Ledger filter validation

marination:
- Hard to replicate but while routing back and forth between Balance Sheet and General Ledger, at a point, Balance Sheet's filters get passed on to General Ledger due to a failure of refreshing filters after routing
- Balance Sheet has no from date and to date , instead, from fiscal year and to fiscal year. As a result in GL Report:
 ```
     File ".../erpnext/erpnext/accounts/report/general_ledger/general_ledger.py", line 60, in validate_filters
        if filters.from_date > filters.to_date:
     TypeError: '>' not supported between instances of 'NoneType' and 'NoneType'
 ```
- Also on invoking GL Report from a server side call without from and to dates, it breaks(tuple index out of range) due to an empty 'to date' value

**Fix:**
- As a defensive fix , always validate the mandatory fields before proceeding


rohitwaghchaure:
@marination from date and to date are mandatory filters, so ideally code should not be execute if mandatory filters are not set.

marination:
@rohitwaghchaure you're right, but while going back and forth i think it tries to render the last generated report where at some points the wrong report filters get passed. From what i see it's making it to the server side somehow. 

I can't seem to replicate the JS issue, but i thought even if someone remotely calls the report from a remote endpoint, we should validate mandatory filters at the least

rohitwaghchaure (approved):

marination:

PR #23026, opened 2020-08-13
fix: recipients list for customer credit limit exceeded email

ruchamahabal:
**Problem:**

If a sales order which is exceeding the credit limit set for the customer is being submitted by a user who does not have a credit controller / sales manager role, on submit, system has to send emails to users with those roles. Following error shows up when Send Email button is clicked:

`TypeError: unhashable type: 'list'` error while sending customer credit limit exceeded email

customer-credit-limit

**Traceback:**

```python
Traceback (most recent call last):
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/app.py", line 64, in application
    response = frappe.api.handle()
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/api.py", line 58, in handle
    return frappe.handler.handle()
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/handler.py", line 30, in handle
    data = execute_cmd(cmd)
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/handler.py", line 69, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/__init__.py", line 1086, in call
    return fn(*args, **newargs)
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/erpnext/erpnext/selling/doctype/customer/customer.py", line 422, in send_emails
    frappe.sendmail(recipients=[args.get('credit_controller_users_list')], subject=subject, message=message)
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/__init__.py", line 507, in sendmail
    inline_images=inline_images, header=header, print_letterhead=print_letterhead)
  File "/home/frappe/benches/bench-version-13-2020-08-07/apps/frappe/frappe/email/queue.py", line 85, in send
    recipients = list(set(recipients))
TypeError: unhashable type: 'list'
```

**Fix:**

The `credit_controller_users_list` is already a list so it should not be passed as an argument inside another list. 

customer-credit-limit-works

PR #23059, opened 2020-08-17
fix: Credit Limit Email not working

marination:
**Issues:**
- `get_users_with_role` on line 396 returns a list of users. Same list was cast into yet another list on line 429. Thus recipients = [["abc@example.com"]]. This caused :
   ```
        File ".../apps/frappe/frappe/email/queue.py", line 85, in send
            recipients = list(set(recipients))
        TypeError: unhashable type: 'list'
    ```
- The credit controllers users list was formatted to ===> user (Fullname). This overwrote recipients list from ["abc@example.com"] to ["John Doe (abc@example.com)] which is invalid as a recipient address.
<br> Screenshot 2020-08-17 at 4 04 46 PM

**Fixes:**
- don't cast into list again while invoking frappe.sendmail
- Separate credit controller users list and the formatted list with better naming to reduce confusion. Send credit controller users list as argument in frappe.sendmail


deepeshgarg007:
@marination One more PR for the fix is also open https://github.com/frappe/erpnext/pull/23026

rohitwaghchaure (approved):

PR #23061, opened 2020-08-17
feat(queries): sort warehouses based on item quantity in descending manner

DeeMysterio:
**Existing system -**

The query set for warehouse dropdown sorts the warehouses by their names in descending order. This can sometimes get not easy to select a warehouse that has quantity in it for the chosen item as the warehouses that have the quantity can be at a lower position in the dropdown menu.

Screenshot from 2020-08-19 01-12-59

**New system -**

Sorted the warehouses based on the item quantity in them in descending order.

Screenshot from 2020-08-19 01-12-23

### DIFF
--- a/erpnext/controllers/queries.py
+++ b/erpnext/controllers/queries.py
@@ -468,24 +468,18 @@ def warehouse_query(doctype, txt, searchfield, start, page_len, filters):
 	conditions, bin_conditions = [], []
 	filter_dict = get_doctype_wise_filters(filters)
 
-	sub_query = """ select round(`tabBin`.actual_qty, 2) from `tabBin`
-		where `tabBin`.warehouse = `tabWarehouse`.name
-		{bin_conditions} """.format(
-		bin_conditions=get_filters_cond(doctype, filter_dict.get("Bin"),
-			bin_conditions, ignore_permissions=True))
-
 	query = """select `tabWarehouse`.name,
-		CONCAT_WS(" : ", "Actual Qty", ifnull( ({sub_query}), 0) ) as actual_qty
-		from `tabWarehouse`
+		CONCAT_WS(" : ", "Actual Qty", ifnull(round(`tabBin`.actual_qty, 2), 0 )) actual_qty
+		from `tabWarehouse` left join `tabBin`
+		on `tabBin`.warehouse = `tabWarehouse`.name {bin_conditions}
 		where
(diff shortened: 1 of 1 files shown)


erpnext/controllers/queries.py
```diff
@@ -468,26 +468,20 @@ def warehouse_query(doctype, txt, searchfield, start, page_len, filters):
...
+		CONCAT_WS(" : ", "Actual Qty", ifnull(round(tb.actual_qty, 2), 0 )) actual_qty
+		from `tabWarehouse` tw left join `tabBin` tb
+		on tb.warehouse = tw.name {bin_conditions}
 		where
-		   `tabWarehouse`.`{key}` like {txt}
+		   tw.`{key}` like {txt}
 			{fcond} {mcond}
-		order by
-			`tabWarehouse`.name desc
+		order by ifnull(round(tb.actual_qty, 2), 0) desc
```
rohitwaghchaure: why round?

erpnext/controllers/queries.py
```diff
@@ -468,26 +468,20 @@ def warehouse_query(doctype, txt, searchfield, start, page_len, filters):
...
-		bin_conditions=get_filters_cond(doctype, filter_dict.get("Bin"),
-			bin_conditions, ignore_permissions=True))
-
-	query = """select `tabWarehouse`.name,
-		CONCAT_WS(" : ", "Actual Qty", ifnull( ({sub_query}), 0) ) as actual_qty
-		from `tabWarehouse`
+	query = """select tw.name,
+		CONCAT_WS(" : ", "Actual Qty", ifnull(round(tb.actual_qty, 2), 0 )) actual_qty
+		from `tabWarehouse` tw left join `tabBin` tb
+		on tb.warehouse = tw.name {bin_conditions}
```
rohitwaghchaure: instead of tb, tw use proper table name `tabBin` and `tabWarehouse` so that you don't need to use replace method in line 482 and 484

erpnext/controllers/queries.py
DeeMysterio: fixed it

erpnext/controllers/queries.py
DeeMysterio: Okay, fixed that.

rohitwaghchaure (approved):

PR #23066, opened 2020-08-18
refactor: Issue form cleaned up and renamed Minutes to First Response field

ruchamahabal:
Depends on: https://github.com/frappe/frappe/pull/11302
- Issue doctype
   - Renamed Fields
      - Service Level Agreement Fulfilled (**agreement_fulfilled**) -> Service Level Agreement Status (**agreement_status**)
      - Minutes to First Response (**mins_to_first_response**) -> First Response Time (**first_response_time**)
   - Changed Fieldtypes
      - Response by Variance: Float -> Duration
      - Resolution by Variance: Float -> Duration
      - Minutes to First Response: Float -> Duration

   **Before:**

     issue-form-before

  **After:**

     issue-after

   - Fields re-ordered
   - Service Level Agreement Details section is only visible if Track Service Level Agreement is enabled in Support Settings
- Reports
   - Since "Minutes to First Response" has been renamed to "First Response Time", changed the related reports and updated desk pages
     - Minutes to First Response for Issues -> First Response Time for Issues

        **Before:**

        mins-issues

       **After:**

        first-response-time-issues

     - Minutes to First Response for Opportunity -> First Response Time for Opportunity


erpnext/crm/report/first_response_time_for_opportunity/first_response_time_for_opportunity.py
```diff
@@ -0,0 +1,39 @@
...
+import frappe
+
+def execute(filters=None):
+	columns, data = [], []
+	return columns, data
+
+def execute(filters=None):
+	columns = [
+		{
+			'fieldname': 'creation_date',
```
barredterra: execute defined twice

barredterra (commented):
Any reason why we need two reports on first response time? Could we instead have the doctype as an optional filter?

ruchamahabal:
> Any reason why we need two reports on first response time? Could we instead have the doctype as an optional filter?

Since both of them belong to different modules(CRM and Support), I think it is better to keep them separate.

barredterra:
I think we can have two reports but why duplicate the same 35 lines of code?

nabinhait (approved):

PR #23077, opened 2020-08-19
fix(stock_controller): precision based on `debit_in_account_currency`

iRaySpace:
**Overview**
This is a new PR to existing PR:
https://github.com/frappe/erpnext/pull/22860

This is a logic already added to the existing branch. However, `version-11` branch don't have this code yet.
You may refer it here: https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/stock_controller.py#L74


f-9t9it:
Also this one:
[#22852](https://github.com/frappe/erpnext/pull/22852)

Please push it to hotfix instead @iRaySpace

PR #23079, opened 2020-08-19
fix: Item Tax Updating via `Update Items` in SO/PO

marination:
### 1) Update Items
 **Update Items** in SO/PO did not set or fetch the Item Tax Template at all, resulting in faulty or no tax calculation on the new row

## **Fixes**:
- Fetch and set Item tax template and Item Tax template Map in new row via Update Items


marination:
For the reviewer: Unsure if there are cases where the tax calculation fix is wrong. Please let me know

deepeshgarg007 (approved):

PR #23090, opened 2020-08-19
fix: always use https for webhook callback urls

beeonthego:
https is expected for Webhook callbacks. for example, Shopify rejects webhook registration if the scheme is http.


rohitwaghchaure:
Don't send PR on version-12, send it on version-12-hotfix

PR #23097, opened 2020-08-20
fix: Create Opportunity without Default Company from Email

marination:
**Issue:**
- If no Default Company in Global Defaults, error on creating Opportunity from Communication (Email):
  image

**Fix:**
- Prompt to select company if no default found
  opp2

- If default exists prompt is auto-populated with default company
  default

- Feedback on Opportunity Creation
 Screenshot 2020-08-20 at 1 42 55 PM


asbasawaraj:
@marination Thanks for GIFs. Testing becomes easy when PRs are so descriptive.

PR #23126, opened 2020-08-24
fix: user created manual job card not linking job card operation with work order operation

rohitwaghchaure:
**Issue**

- User has created Job Card manually for the work order and not created job cards from the work order

- Also user has added same operation multiple times in work order

- Due to this system has not calculated Completed Quantity against the operation in the work order correctly.

**After Fix**
If one operation is added multiple times in the work order and user trying to make job card manually then system will ask to select the correct operation row number

Screenshot 2020-08-24 at 5 58 46 PM

Screenshot 2020-08-24 at 5 58 51 PM

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.py
+++ b/erpnext/manufacturing/doctype/job_card/job_card.py
@@ -9,10 +9,13 @@ from frappe.utils import flt, time_diff_in_hours, get_datetime, time_diff, get_l
 from frappe.model.mapper import get_mapped_doc
 from frappe.model.document import Document
 
+class OperationMismatchError(frappe.ValidationError): pass
+
 class JobCard(Document):
 	def validate(self):
 		self.validate_time_logs()
 		self.set_status()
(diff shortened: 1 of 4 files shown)


erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -9,10 +9,13 @@
 from frappe.model.mapper import get_mapped_doc
 from frappe.model.document import Document
 
+class OperatiobMismatchError(frappe.ValidationError): pass
```
rmehta: typo `OperationMismatchError`

rmehta (changes requested):

erpnext/manufacturing/doctype/job_card/job_card.py
rohitwaghchaure: Fixed

rohitwaghchaure:
@Mergifyio backport develop

PR #21273, opened 2020-04-14
refactor: Quoted Item Comparison Report

marination:
- Refactored Report View with entries grouped by supplier and RFQ reference too.
screencapture-ugh-in-8001-desk-2020-05-07-17_30_06

- Added Bar Graph and additional functional filters.
- Added Supplier Multi Select filter to compare between specific suppliers.

**Working:**
quoted-item-comp

- Added **'valid_till'** date in Supplier Quotation along with Expiry Status
Screenshot 2020-04-15 at 11 49 53 AM

- Added patch to set `valid_till` and daily job to set **expiry** **status**

PR #23127, opened 2020-08-24
feat: Quoted Item Comparison Report Enhancements v2

marination:
Enhancements over https://github.com/frappe/erpnext/pull/21273
Depends on https://github.com/frappe/frappe/pull/11386

### Columns /Report General
- Added **Valid Till**, **Lead Time (Days)** from Supplier Quotation Item, **Item**
- Ordered by transaction date. If same date , ordered alphabetically to preserve order of items for visual ease

### Filters
- Added Date Filters. Fetch supplier quotations for comparison within a given date range (mandatory)
- Added checkbox to include expired entries
- Made Supplier Quotation MultiSelect just like Supplier Filter

### Chart
- Only visible on adding Item Filter
- Added currency symbol in legend

### Multi Item Comparison
- The report only allowed single item comparisons, item code filter was mandatory. 
- Item filter is optional now. By default it will fetch Supplier-wise data for all Supplier Quotations and Show Items as well.
  Screenshot 2020-08-24 at 6 32 38 PM
- You can also add the Item filter. **The chart will only be visible on populating the Item Filter**
 Screenshot 2020-08-24 at 6 34 46 PM

### Expired Supplier Quotations
- Expiry of a Supplier Quotation depends upon it's **Valid Till** date
- By default the report only shows you data for Supplier Quotations that are valid or not expired
- If you want to still analyse expired Supplier Quotations you can enable **Include Expired**
- As per the color coding: Orange date are the ones a week or less away form expiry, Red dates are ones that are expiring the same day or already expired
 Screenshot 2020-08-24 at 6 37 00 PM


erpnext/buying/report/quoted_item_comparison/quoted_item_comparison.js
```diff
@@ -45,13 +60,12 @@ frappe.query_reports["Quoted Item Comparison"] = {
...
-			fieldtype: "Link",
+			fieldtype: "MultiSelectList",
 			label: __("Supplier Quotation"),
-			options: "Supplier Quotation",
 			fieldname: "supplier_quotation",
 			default: "",
-			get_query: () => {
-				return { filters: { "docstatus": ["<", 2] } }
+			get_data: function(txt) {
+				return frappe.db.get_link_options('Supplier Quotation', txt);
```
deepeshgarg007:
```suggestion
				return frappe.db.get_link_options('Supplier Quotation', txt, {'docstatus': ['<', 2]});
```

deepeshgarg007 (changes requested):
Also the status message didn't seemed to be working. I may be missing something. Please check once

deepeshgarg007 (approved):

PR #23131, opened 2020-08-25
fix: Form dashboard showing wrong balance

Anurag810:
Dashbord showing wrong data:
image


rohitwaghchaure (dismissed):

rohitwaghchaure:
@Anurag810 kindly add test case for this scenario

Anurag810:
It does not require test as these functions are already called during test

rohitwaghchaure (approved):

PR #23132, opened 2020-08-25
fix: [Portal] added filter show in website

anupamvs:
**Issue -** When the user sets Item Group as a filter in the Product Settings, it will show all the Item Group under the filter section on the Product page. Even if the "Show in Website" checkbox is unchecked in the Item Group Master.
image

**After Fix -**
image


anupamvs:

rohitwaghchaure (approved):

PR #23143, opened 2020-08-25
fix: Check if subscription is already created for current invoicing period

deepeshgarg007:
Multiple Invoices were being created for against a subscription on subscription update even though invoice for the current invoicing was already generated

- Add exception handling to not generate invoice in any case if invoice for current period is already generated
- Invoice for new month will be generated even though current invoices are unpaid
- Made invoices table read-only


nabinhait (approved):

nabinhait:
@deepeshgarg007 PR for develop branch?

PR #23144, opened 2020-08-25
fix: Homepage Featured Product image fetch issue

anupamvs:
**Issue -** While adding feature product it doesn't fetch item website image from the item master.
feat-product-b

**After Fix -**
feat-product-a


rohitwaghchaure (approved):

rohitwaghchaure:
@anupamvs send PR on version-12-hotfix if required

PR #23152, opened 2020-08-25
feat: Create Issue from Opportunity

michellealva:
This PR lets you create an Issue from the Opportunity via the create button.

issue

- The "Raised By" is auto-populated based on the contact set in the Opportunity
- The "Subject" field is the Customer Name


erpnext/crm/doctype/opportunity/opportunity.js
```diff
@@ -84,6 +84,14 @@ frappe.ui.form.on("Opportunity", {
 
 			frm.add_custom_button(__('Quotation'),
 				cur_frm.cscript.create_quotation, __('Create'));
+			
+			frm.add_custom_button(__('Issue'),
```
rohitwaghchaure: Check whether user has permission to create the Issue or not 
use below code
```
if (in_list(frappe.boot.user.can_create, "Issue")) {
   frm.add_custom_button(__('Issue'),
     function() {
                    frappe.new_doc('Issue', {
                              'raised_by': frm.doc.contact_email,
                              'subject': frm.doc.title
                    });
      }, __('Create'));
}
```

michellealva:
@rohitwaghchaure done, also, should the subject be copied from the customer's name or should it be left blank?

rmehta (changes requested):
👍 

- add opportunity field in Issue
- add "DocType Link" instead of button

anupamvs:
@rmehta should we add Dynamic Link in issue so that it can be linked to any doc?

rmehta:
@anupamvs not a bad thought. Issue is general purpose, the origin document can be anything.

PR #23165, opened 2020-08-25
fix: reverse journal entry for multi-currency

AfshanKhan:
Before when making a multi-currency reverse journal entry:

Before

After the changes:

After


AfshanKhan:

ruchamahabal (commented):
@AfshanKhan Can you add a test case for Reverse Journal Entry?

AfshanKhan:
@ruchamahabal added test case.

ruchamahabal (approved):

erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -1021,3 +1021,34 @@ def make_inter_company_journal_entry(name, voucher_type, company):
 	journal_entry.posting_date = nowdate()
 	journal_entry.inter_company_journal_entry_reference = name
 	return journal_entry.as_dict()
+
+@frappe.whitelist()
+def make_reverse_journal_entry(source_name, target_doc=None, ignore_permissions=False):
```
rmehta, line 1026: why do you need `ignore_permissions` in a whitelisted method? @AfshanKhan ? This beats the purpose

AfshanKhan: Understood. Will change this.

PR #23172, opened 2020-08-26
fix: query for contact person

AfshanKhan:
Before if you remove the contact person in quotation to change it the filter didn't use to works:
Screenshot 2020-08-26 at 4 25 48 PM

After the change, the values get fetch correctly:
Screenshot 2020-08-26 at 4 24 40 PM


AfshanKhan:

erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -163,6 +164,15 @@ erpnext.selling.QuotationController = erpnext.selling.SellingController.extend({
...
 
+	contact_query: function(doc) {
+		return {
+			filters: {
+				link_doctype: frappe.dynamic_link.doctype,
+				link_name: doc.party_name
+			}
+		};
+	},
+
```
marination, line 175: `sales_common.js` line 41, a query for this Link field already exists (hence the filters that were visible earlier). Maybe check why that doesn't work and fix that query

PR #23186, opened 2020-08-27
fix: Incorrect stock balance issue for batched item

rohitwaghchaure:
### **Issue** 

1. Make single stock reco and add same item and same warehouse with different batches in multiple rows and submit the stock reco.

1. After that check the stock balance report and you can see the system has not updated the balance quantity as per stock reco

Screenshot 2020-08-27 at 1 40 16 PM

### **After Fix**
Screenshot 2020-08-27 at 1 33 09 PM

### **Test case**
Screenshot 2020-08-27 at 2 42 23 PM

image


rohitwaghchaure:
@Mergifyio backport develop

PR #23192, opened 2020-08-27
fix: Unlink and delete batch created from stock reco on cancel

marination:
### Stock Reconciliation
- Create an opening Stock Reconciliation for a Batched Item. 
- A new batch should be created via this transaction. Source Doctype and Source name in batch will be, the Stock Reconciliation and it's name.
- Cancel the Stock Reconciliation. Batch still remains with Source details as cancelled Stock Reconciliation
- If you try to delete the Stock Reconciliation, it won't let you, as it is linked to this Batch.

### Batch Reference Deletion
- Make a Stock Entry 1(receipt) that creates a batch **B1** and Serial No **S1**
- Make another Stock Entry 2(receipt)adding to the same batch. So **S2** is added to **B1**
- So now Batch **B1** has balance qty 2
- Cancel Stock Entry 2, it calls `delete_auto_created_batches` which removes B1's reference from **ALL** serial nos that are linked to B1
- Hence instead of only **S2**, **S1 and S2** have no batch reference in them
- Now **S1** cannot be used in any transaction as it does not belong to any Batch, But **B1** has a balance qty of 1
 Screenshot 2020-08-27 at 9 35 33 PM

## **Fixes:**
- Call `delete_auto_created_batches` on cancelling Stock Reco. It should unlink serial nos in the transaction only and delete a batch if it **originated** from this transaction only
- Only delete Serial Nos pertaining to the cancelled transaction

 ## **Tests:** 

<details><summary> Stock Entry</summary>

**Simple Test for Serial and Batch Item Receipt**
1) Submit Stock Entry (Receipt) with Serial & Batched Item. Check if Batch is created with Reference in Serial No
2) Cancel same Stock Entry. Check if Batch is deleted and Serial No is Inactive

**Test to Check Batch Reference Deletion Issue**
1) Create 2 Stock Entries, both adding Serial Nos to **same** batch
2) Cancelling first Stock Entry (origin transaction of created batch) should throw a LinkExistsError
3) Cancelling second Stock Entry should make Serial Nos **that are, linked to mentioned batch and in that transaction only**, Inactive and Unlinked from the Batch.

Screenshot 2020-09-02 at 7 19 58 PM

</details>
<details><summary> Stock Reconciliation</summary>

**Simple Test for Serial and Batch Item Opening Reconciliation**
1) Submit Stock Reconciliation. Check if Serial No created is linked to Batch created
2) Cancel the same. Check if Serial No is Inactive and Batch is deleted

**Test for Stock Reconciliation with dependent Entry**
1) Create a Stock Reco, that creates a batch on submit. Hence, becoming the origin or source document of the batch
2) Create a Stock Entry that adds to the same batch
3) Cancel Stock Reco. It won't let you cancel as the batch it tries to delete on cancellation is linked with the Stock Entry (future transaction)
4) Cancel Stock Entry. Only Serial Nos in Stock Entry should be Inactive and unlinked from batch. Batch qty should decrease. Serial Nos generated via Stock Reconciliation should remain untouched.
Screenshot 2020-09-02 at 7 19 45 PM

</details> 

</details>

### DIFF
--- a/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
+++ b/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
@@ -45,6 +45,7 @@ class StockReconciliation(StockController):
 	def on_cancel(self):
 		self.delete_and_repost_sle()
 		self.make_gl_entries_on_cancel()
+		self.delete_auto_created_batches()
 
 	def remove_items_with_no_change(self):
 		"""Remove items if qty or rate is not changed"""
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -242,10 +242,11 @@ class StockController(AccountsController):
 					_(self.doctype), self.name, item.get("item_code")))
 
 	def delete_auto_created_batches(self):
+		from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
 		for d in self.items:
 			if not d.batch_no: continue
 
-			serial_nos = [sr.name for sr in frappe.get_all("Serial No", {'batch_no': d.batch_no})]
+			serial_nos = get_serial_nos(d.serial_no)
 			if serial_nos:
 				frappe.db.set_value("Serial No", { 'name': ['in', serial_nos] }, "batch_no", None)
 
--- a/erpnext/stock/doctype/item_alternative/test_item_alternative.py
+++ b/erpnext/stock/doctype/item_alternative/test_item_alternative.py
@@ -13,6 +13,7 @@ from erpnext.manufacturing.doctype.work_order.test_work_order import make_wo_ord
 from erpnext.buying.doctype.purchase_order.purchase_order import make_purchase_receipt, make_rm_stock_entry
 import unittest
 from erpnext.stock.doctype.purchase_receipt.test_purchase_receipt import set_perpetual_inventory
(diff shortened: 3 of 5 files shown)


rohitwaghchaure (dismissed):

erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
```diff
@@ -45,6 +45,7 @@ def on_submit(self):
 	def on_cancel(self):
 		self.delete_and_repost_sle()
 		self.make_gl_entries_on_cancel()
+		self.delete_auto_created_batches()
```
rohitwaghchaure, line 48: Instead of delete should we remove stock reco from reference? because there might be chances that batch has used in another transactions after stock reco

marination: @rohitwaghchaure i tested this scenario too, if the batch is in another transaction it doesnt let you cancel the **origin** transaction. It throws a "Cannot delete batch as it is linked to another SLE". So we have to delete or cancel all other dependent transactions before trying to cancel the origin transaction
I'll work on a test case for these now

marination:
Travis failing at `Could not find Item Group: _Test Item Group` :(
**Edit:** Travis passed

rohitwaghchaure (approved):

PR #23193, opened 2020-08-27
fix: can't multiply sequence by non-int of type 'float

bhavesh95863:


erpnext/selling/page/point_of_sale/pos_item_cart.js
```diff
@@ -356,7 +356,7 @@ erpnext.PointOfSale.ItemCart = class {
 				onchange: function() {
 					if (this.value || this.value == 0) {
 						const frm = me.events.get_frm();
-						frappe.model.set_value(frm.doc.doctype, frm.doc.name, 'additional_discount_percentage', this.value);
+						frappe.model.set_value(frm.doc.doctype, frm.doc.name, 'additional_discount_percentage', parseFloat(this.value));
```
rohitwaghchaure: use flt instead of parseFloat

bhavesh95863: changed

rohitwaghchaure (approved):

PR #23195, opened 2020-08-28
fix: Raise Error on over receipt/consumption for sub-contracted PR

marination:
**Issue:**
- Create a sub contracted Purchase Order
- Create two **Draft** Purchase Receipts that fully receive this very PO
- Submit the Receipts one by one. The first receipt should receive/consume the PO items/supplied items fully
- The second receipt will break as there is no `fg_yet_to_be_received` (Finish goods yet to be received) in the PO:
   
   ```
     File ".../apps/erpnext/erpnext/controllers/buying_controller.py", line 301, in 
     update_raw_materials_supplied_based_on_stock_entries
          qty = (rm_qty_to_be_consumed / fg_yet_to_be_received) * item.qty
    TypeError: unsupported operand type(s) for /: 'float' and 'NoneType'
   ```
- This happens on backflush based on materials transferred for subcontracting

**Fix:**
- Alert the user that an over receipt is taking place as this issue often occurs due to two different users creating PRs against the same PO
   Screenshot 2020-08-28 at 12 14 53 PM
- While looping if there's no PO against this item (accidentally or intentionally) don't proceed to create any qty maps as they will all be empty

**Note**
- There is still one problematic case not handled here i.e. Returning a sub-contracted Purchase Receipt. That will be done in a separate PR after more clarity in it's design

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -276,6 +276,9 @@ class BuyingController(StockController):
 		qty_to_be_received_map = get_qty_to_be_received(purchase_orders)
 
 		for item in self.get('items'):
+			if not item.purchase_order:
+				continue
+
 			# reset raw_material cost
 			item.rm_supp_cost = 0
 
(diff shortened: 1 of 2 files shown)


marination:
Screenshot 2020-08-28 at 2 22 42 PM

rohitwaghchaure (approved):

PR #23201, opened 2020-08-28
fix: style for website theme

barredterra:
- `erpnext/public/scss/website.scss` will be automatically included in Website Theme.
- The JavaScript file that manipulates the SCSS _on validate_ is no longer needed.

Depends on https://github.com/frappe/frappe/pull/11555.


barredterra:
@netchampfaris can we merge this?

PR #23212, opened 2020-08-30
fix: Better error feedback on creating SO from Quotation

marination:
**Issue:**
- Manually create a Lead (without Territory set)
- (Optional): Remove Default Customer Group form Selling Settings
- Create Quotation from this Lead
- Create Sales Order from this quotation, you will see:
 ss
- The problem here is that a Mandatory fields 'Territory' and 'Customer Group' were missing in the auto created Customer, and the message is misleading

**Fix:**
-  Added specifics, formatting and title to error message
 Screenshot 2020-08-30 at 9 00 30 PM


deepeshgarg007 (approved):

marination:

PR #22954, opened 2020-08-07
fix: events not deleted on cancelling maintenance schedule

scmmishra:
When cancelling a maintenance schedule, associated events would not be deleted. This was because they were not created properly in the first place. 

This PR fixes that

Depends On: https://github.com/frappe/frappe/pull/11221 [Merged]

PR #23219, opened 2020-08-31
fix: events not deleted on cancelling maintenance schedule

scmmishra:
Backport of https://github.com/frappe/erpnext/pull/22954

---

When cancelling a maintenance schedule, associated events would not be deleted. This was because they were not created properly in the first place. 

This PR fixes that

Depends On: https://github.com/frappe/frappe/pull/11221 [Merged]

### DIFF
--- a/erpnext/maintenance/doctype/maintenance_schedule/test_maintenance_schedule.py
+++ b/erpnext/maintenance/doctype/maintenance_schedule/test_maintenance_schedule.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 # See license.txt
 from __future__ import unicode_literals
+from frappe.utils.data import get_datetime, add_days
 
 import frappe
 import unittest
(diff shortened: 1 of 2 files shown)


erpnext/maintenance/doctype/maintenance_schedule/test_maintenance_schedule.py
```diff
@@ -2,11 +2,47 @@
...
+from frappe.utils.data import get_datetime, add_days
 
 import frappe
 import unittest
 
 # test_records = frappe.get_test_records('Maintenance Schedule')
 
 class TestMaintenanceSchedule(unittest.TestCase):
-	pass
+	def events_should_be_created_and_deleted(self):
```
marination:
```suggestion
	def test_events_should_be_created_and_deleted(self):
```

marination (approved):

Olawale1:
Hi @scmmishra 

Thanks for this patch. In our instance, the Events are not even getting created at all after submitting the Maintenance Schedule. Does this address the issue as well?

ERPNext: v12.10.1 (version-12)
Frappe Framework: v12.8.4 (version-12)

Kind regards,

Olawale1:
So, it turns out the Event actually gets created but it's totally invisible to the Creator of the Maintenance Schedule and System Manager! Only the 'Sales Person' designated in the schedule can view the Events... I think this is a rather poor design and should probably be replaced with something similar to the Asset Maintenance implementation

Thanks

Olawale1:
Created a Github Issue: https://github.com/frappe/erpnext/issues/23682

PR #23220, opened 2020-08-31
feat: Validate and Update Child Company Accounts on Rename/Adding Account

marination:
### **Rename**
- Assume parent company **Zoot**, with child companies **Child**, **Child2** and **Child3**
- Assume **Zoot** has a group account **Temporary Expenses** under parent account **Expenses**. The same get's added to all child company COAs as well
- If group account **Temporary Expenses** get's renamed or an account number get's added to it in the company **Zoot**, **_the same is not updated in Child Company Accounts_**. 

After renaming lets assume, it becomes **Temporary Expenditure**

### **Adding Account**
- Now If you add yet another account under account **Temporary Expenditure** in company **Zoot**, the same is account is not added in the child companies as account **Temporary Expenditure** does not exist in them, **Temporary Expenses** does

### **Rename**
- On renaming or changing account number in Parent Company Account, make sure the same is changed in Child Company corresponding accounts
 account-tree

- On renaming account name/number in child company, check if `allow_account_creation_against_child_company` is disabled. If yes, check if account being renamed exists in parent company. If yes:
  Screenshot 2020-09-23 at 2 40 16 PM


erpnext/accounts/doctype/account/test_account.py
```diff
@@ -111,6 +111,56 @@ def test_account_sync(self):
...
+		acc.is_group = 1
+		acc.company = "_Test Company 3"
+		acc.insert()
+
+		self.assertTrue(frappe.db.exists("Account", {'account_name': "Test Group Account", "company": "_Test Company 4"}))
+		self.assertTrue(frappe.db.exists("Account", {'account_name': "Test Group Account", "company": "_Test Company 5"}))
+
+		acc_tc_5 = frappe.db.get_value('Account', {'account_name': "Test Group Account", "company": "_Test Company 5"})
+		# Rename group account in one child company
+		update_account_number(acc_tc_5, "Test Modified Account")
```
deepeshgarg007: Shouldn't we check while renaming the account in child company itself instead of throwing an error on addition of account in parent company? If same account in parent company exits then allow rename only via parent company.

If the above is not possible then I don't think we should throw a validation error in case account of same name in child company doesn't exist. We can just `msgprint` to notify the user instead of throwing a validation error.

marination:
@deepeshgarg007 changes done, please check and let me know

deepeshgarg007 (approved):

PR #23225, opened 2020-09-01
fix: Update route after rename and update doc

SDLyu:

After rename or update item group doc that route not change, it should change to correct route path.

This PR call `make_route` function in `after_rename` `on_update` event and make route change.

> Screenshots/GIFs

### Change `Node4` to `Node5`

Item_Group_Tree

#### Before

_MySQL_5_5_5-10_3_22-MariaDB-1_10_3_22_maria_bionic__docker_erpnext__a94a8fe5ccb19ba6_tabItem_Group

#### After

_MySQL_5_5_5-10_3_22-MariaDB-1_10_3_22_maria_bionic__docker_erpnext__a94a8fe5ccb19ba6_tabItem_Group

### DIFF
--- a/erpnext/setup/doctype/item_group/item_group.py
+++ b/erpnext/setup/doctype/item_group/item_group.py
@@ -43,23 +43,26 @@ class ItemGroup(NestedSet, WebsiteGenerator):
 
 	def make_route(self):
 		'''Make website route'''
-		if not self.route:
-			self.route = ''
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@SDLyu If someone has renamed the parent item group then does your PR will update the route for node item group as well?
Also travis is failing, kindly fix the test cases and then reopen the PR

PR #23233, opened 2020-09-02
fix: set conversion factor while creating RFQ from Opportunity

ruchamahabal:
**Steps to replicate:**

1. Created Opportunity.
2. Select items in opportunity, with default UOM only
3. Create new RFQ
4. Select Opportunity in RFQ using **Get Items** option
5. Item is fetched in RFQ. Warehouse is selected
6. When saving RFQ, it throws the mandatory validation: conversion factor is 0.

Though it fetched default UoM correctly, but conversion factor was not set as 1

gif

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.py
+++ b/erpnext/crm/doctype/opportunity/opportunity.py
@@ -267,6 +267,12 @@ def make_quotation(source_name, target_doc=None):
 
 @frappe.whitelist()
 def make_request_for_quotation(source_name, target_doc=None):
+	def set_missing_values(source, target):
+		rfq = frappe.get_doc(target)
+		for item in rfq.items:
+			# opportunity item is not multi-uom
+			item.conversion_factor = 1.0
+
 	doclist = get_mapped_doc("Opportunity", source_name, {
 		"Opportunity": {
 			"doctype": "Request for Quotation"
@@ -279,7 +285,7 @@ def make_request_for_quotation(source_name, target_doc=None):
 				["uom", "uom"]
 			]
 		}
-	}, target_doc)
+	}, target_doc, set_missing_values)
 
 	return doclist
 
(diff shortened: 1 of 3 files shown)


erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -279,7 +285,7 @@ def make_request_for_quotation(source_name, target_doc=None):
 				["uom", "uom"]
 			]
 		}
-	}, target_doc)
+	}, target_doc, set_missing_values)
```
rohitwaghchaure: Instead of adding this code in set_missing_value, add key ``` "postprocess": update_item ``` after 286 line.
where as update_item will be the name of method. For reference check below code
```
def make_material_request(source_name, target_doc=None):
	requested_item_qty = get_requested_item_qty(source_name)

	def update_item(source, target, source_parent):
		# qty is for packed items, because packed items don't have stock_qty field
		qty = source.get("qty")
		target.project = source_parent.project
		target.qty = qty - requested_item_qty.get(source.name, 0)
		target.stock_qty = flt(target.qty) * flt(target.conversion_factor)

	doc = get_mapped_doc("Sales Order", source_name, {
		"Sales Order": {
			"doctype": "Material Request",
			"validation": {
				"docstatus": ["=", 1]
			}
		},
		"Packed Item": {
			"doctype": "Material Request Item",
			"field_map": {
				"parent": "sales_order",
				"uom": "stock_uom"
			},
			"postprocess": update_item
		},
		"Sales Order Item": {
			"doctype": "Material Request Item",
			"field_map": {
				"name": "sales_order_item",
				"parent": "sales_order"
			},
			"condition": lambda doc: not frappe.db.exists('Product Bundle', doc.item_code) and doc.stock_qty > requested_item_qty.get(doc.name, 0),
			"postprocess": update_item
		}
	}, target_doc)

	return doc
```

rohitwaghchaure (approved):

PR #23235, opened 2020-09-02
feat: added option to add custom remarks in payment entry

Anurag810:
Added option to add custom remarks in Payment entry.

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.json
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.json
@@ -1,4 +1,5 @@
 {
+ "actions": [],
  "allow_import": 1,
  "autoname": "naming_series:",
  "creation": "2016-06-01 14:38:51.012597",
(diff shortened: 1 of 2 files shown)


deepeshgarg007 (commented):
Why add this checkbox and make remarks ready only? I don't see any need for this. Any specific reason? If yes then please update in the description

Anurag810:
Please refer issue: ISS-20-21-04489. it was as suggested by Nabin.

deepeshgarg007 (changes requested):
@Anurag810 You didn't read the full suggestion. The main issue is still there. The standard system generated remarks are still not updated in case of reference or paid amount change

nabinhait (changes requested):
If custom remarks is checked, no need to update the remarks field based on reference.

https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/payment_entry/payment_entry.py#L456

Replace the line `if self.remarks: return` with `if self.custom_remarks: return`

deepeshgarg007 (approved):

PR #23238, opened 2020-09-02
feat: Option to print UOM after quantity

Semiranis:

In ERPNext printing, default is (small) UOM and (normal size) quantity, e.g. kg 10. Now user can activate alternative version with UOM after quantity, e.g. 10 kg, which is sometimes better.

Added alternative template file.
Added checkbox in Print Settings for activating this option.
Added code for using alternative template if option in settings is set.

I looked for such possibility and it was requested in forum:
https://discuss.erpnext.com/t/print-format-to-display-uom-after-the-quantity/11774

In Poland nobody uses UOM before quantity, so mentioned above "kg 10" looks strange for many people. 

I tried making custom print format, but it was too much work to customize all the documents in ERPNext. This option saved me a lot of time.

> Screenshots/GIFs


erpnext/setup/install.py
```diff
@@ -66,6 +67,16 @@ def create_compact_item_print_custom_field():
 	})
 
 
+def create_print_uom_after_qty_custom_field():
+	create_custom_field('Print Settings', {
```
rohitwaghchaure, line 71: Why custom field and not standard field?

erpnext/templates/print_formats/includes/item_table_qty_swapped.html
```diff
@@ -0,0 +1,5 @@
+{{ doc.get_formatted("qty", doc) }}
```
rohitwaghchaure, line 1: Instead of separate template you can use the if and else condition in the existing template

rohitwaghchaure:
@Semiranis closing this PR, kindly fix the code and reopen the PR

erpnext/setup/install.py
Semiranis: How can I do this? Print Settings are part of frappe/frappe, not erpnext. It's the only solution I found in erpnext code. I asked in forum, but nobody helped.

PR #23245, opened 2020-09-03
fix: Attendance calendar map fix

Anurag810:
#23247
Before:
image

Now:
image


erpnext/hr/doctype/shift_assignment/shift_assignment.py
```diff
@@ -103,7 +103,7 @@ def add_assignments(events, start, end, conditions=None):
 			"doctype": "Shift Assignment",
 			"start_date": d.start_date,
 			"end_date": d.end_date if d.end_date else nowdate(),
-			"title": cstr(d.employee_name) + \
+			"title": cstr(d.employee_name) + ":"+ \
```
rohitwaghchaure: Add space in between employee name and  ":"

rohitwaghchaure (approved):

PR #23257, opened 2020-09-04
fix: incorrect calculation for consumed qty for subcontract item

rohitwaghchaure:
## Issue

1. Set backflush Raw Materials based on as "Material Transferred for Subcontract" in buying settings

1. Created subcontracted purchase order and transferred two batches against same item in the stock entry

1. On submission of stock entry system not updated the correct Supplied Quantity in the purchase order
Screenshot 2020-09-04 at 12 19 43 PM

1. Another issue is when purchase receipt has created against PO, the system has considered the 1st batch quantity only and not the second batch quantity for consuming the raw materials from the supplier's warehouse.
Screenshot 2020-09-04 at 11 12 23 AM

### Purchase Order Supplied quantity
Screenshot 2020-09-04 at 12 55 53 PM

### Purchase Receipt Consumed Qty With two batches
Screenshot 2020-09-04 at 11 46 13 AM

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -322,7 +322,7 @@ class BuyingController(StockController):
 
 				if raw_material.batch_nos:
 					batches_qty = get_batches_with_qty(raw_material.rm_item_code, raw_material.main_item_code,
-						qty, transferred_batch_qty_map, backflushed_batch_qty_map)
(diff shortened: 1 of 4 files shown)


rohitwaghchaure:
Screenshot 2020-09-14 at 7 46 14 PM

nabinhait (approved):

PR #23262, opened 2020-09-04
fix: not able to make material request from SO

rohitwaghchaure:
**Issue**

1. Created sales order against BOM item and one of the raw materials has no description in the item master.

1. When trying to make materials request against the sales order (SO > Create > Request for Raw Materials), getting following error.
Screenshot 2020-09-04 at 3 13 22 PM

**After Fix**
If no description then set Item Name or item code


rohitwaghchaure:

rohitwaghchaure:
@Mergifyio backport develop

PR #23282, opened 2020-09-07
fix: Apply TDS on Purchase Invoice creation from Purchase Order and Purchase Receipt

deepeshgarg007:
**Steps to replicate:**
1. Add tax withholding category for a supplier
2. Create a PO or PR for that supplier
3. Create Invoice from the PO or PI and check if TDS category is visible or not

**Solution:**
Triggered supplier even on load if supplier is already available in doc

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
@@ -25,6 +25,12 @@ erpnext.accounts.PurchaseInvoice = erpnext.buying.BuyingController.extend({
 				this.frm.set_df_property("credit_to", "print_hide", 0);
 			}
 		}
+
+		// Trigger supplier event on load if supplier is available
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
```diff
@@ -25,6 +25,12 @@ erpnext.accounts.PurchaseInvoice = erpnext.buying.BuyingController.extend({
 				this.frm.set_df_property("credit_to", "print_hide", 0);
 			}
 		}
+
+		// Trigger supplier event on load if supplier is available
+		// The reason for this is PI can be created from PR or PO and supplier is pre populated
+		if (this.frm.doc.supplier) {
+			this.frm.trigger('supplier');
```
rohitwaghchaure, line 32: This is not working when you make purchase invoice using get items from purchase order

Screenshot 2020-09-08 at 10 08 54 AM

rohitwaghchaure (approved):

PR #23283, opened 2020-09-07
fix: Set student email id mandatory only if skip user creation is disabled

ruchamahabal:
The student email ID should not be mandatory if **Skip User creation for new Student** is enabled in Education Settings:

skip-user-creation

If the education module is being used by some organizations like primary school / pre-school / kindergarten, the students don't have their individual email IDs. Their parent's contact details can be added to the [Guardian](https://docs.erpnext.com/docs/user/manual/en/education/guardian) doctype and then a guardian can be created as a user in such cases.


erpnext/education/doctype/student/student.js
```diff
@@ -25,6 +25,12 @@ frappe.ui.form.on('Student', {
 					{party_type:'Student', party:frm.doc.name});
 			});
 		}
+
+		frappe.db.get_value('Education Settings', {name: 'Education Settings'}, 'user_creation_skip', (r) => {
+			if (r.user_creation_skip === "0") {
```
deepeshgarg007:
```suggestion
			if (cint(r.user_creation_skip) !== 1) {
```

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #23284, opened 2020-09-07
fix(selling): validate delivery date on sales order

DeeMysterio:
**Problem:**
It lets the user save and submit a Sales Order even if the expected delivery date is before the date of the Sales Order creation date.


rmehta:
This is by design, there can be cases where the shipment is made before the order is booked.

rohitwaghchaure:
@DeeMysterio Add custom script

PR #23285, opened 2020-09-07
fix(selling): validate delivery date on sales order

DeeMysterio:
**Problem:**
It lets the user save and submit a Sales Order even if the expected delivery date is before the date of the Sales Order creation date.


rohitwaghchaure:
@DeeMysterio Kindly fix travis

PR #23298, opened 2020-09-08
fix: asset movement date for backdated asset entry

nextchamp-saqib:
Problem: First asset movement was made based on the time of submitting asset. For an asset which was purchased earlier but the record is submitted today had its auto generated asset movement's transaction date as today.


nextchamp-saqib:

PR #23309, opened 2020-09-11
feat: added project in Sales Analytics report

anupamvs:
image


erpnext/selling/report/sales_analytics/sales_analytics.py
```diff
@@ -198,6 +202,28 @@ def get_sales_transactions_based_on_item_group(self):
...
+		entity = "project as entity"
+
+		self.entries = frappe.get_all(self.filters.doc_type,
+			fields=[entity, value_field, self.date_field],
+			filters={
+				"docstatus": 1,
+				"company": self.filters.company,
+				"project": ["!=", ""],
+				self.date_field: ('between', [self.filters.from_date, self.filters.to_date])
+			}, debug =True
```
deepeshgarg007: Remove debug

erpnext/selling/report/sales_analytics/sales_analytics.py
```diff
@@ -198,6 +202,28 @@ def get_sales_transactions_based_on_item_group(self):
...
+		)
+
+		self.entity_names = {}
+		for d in self.entries:
+			self.entity_names.setdefault(d.entity, d.entity_name)
```
deepeshgarg007: `entity_name` is never fetched in the query, how it is used here?

deepeshgarg007 (changes requested):

erpnext/selling/report/sales_analytics/sales_analytics.py
```diff
@@ -198,14 +202,32 @@ def get_sales_transactions_based_on_item_group(self):
...
+
 	def get_rows(self):
 		self.data = []
 		self.get_periodic_data()
 
 		for entity, period_data in iteritems(self.entity_periodic_data):
 			row = {
 				"entity": entity,
-				"entity_name": self.entity_names.get(entity)
+				"entity_name": self.entity_names.get(entity) if hasattr(self, 'entity_names') else None
```
deepeshgarg007, line 230: No need for this check since entity name won't be added for option Project, also the attribute is `entity_name` you have used `entity_names` here.

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #23314, opened 2020-09-12
fix: consumed qty logic for subcontracted raw materials

rohitwaghchaure:
**Issue**

Steps to replicate the issue

1. Set Backflush Raw Materials Based On as "Material Transferred for Manufacture"

1. Create subcontracted batched item SI-0001 and make BOM with one raw material(RM-0001) item and keep proportion as 1:1 for raw material and finished good in the BOM.

1. Make purchase order for subcontracted item SI-0001 with quantity as 500

1. Make material transfer entry from Stores to Supplier warehouse and transfer 552 raw material RM-0001

1. After create, partial purchase receipt for 200 quantity for item SI-0001, in the purchase receipt add multiple batches(4 batches) in the purchase receipt item with quantity as 30, 60, 70, 40.

1. After make purchase receipt for remaining 300 quantity, on save you can see the consumed quantity is showing as negative in the purchase receipt.

Consumed quantity for the raw materials in subcontracted purchase receipt showing in negative

Screenshot 2020-09-12 at 12 28 48 AM

**After Fix**
Screenshot 2020-09-12 at 12 29 48 AM

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _, msgprint
 from frappe.utils import flt,cint, cstr, getdate
-
+from six import iteritems
 from erpnext.accounts.party import get_party_details
 from erpnext.stock.get_item_details import get_conversion_factor
 from erpnext.buying.utils import validate_for_items, update_last_purchase_rate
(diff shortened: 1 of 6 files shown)


rohitwaghchaure:
image

erpnext/controllers/buying_controller.py
```diff
@@ -861,39 +862,45 @@ def get_subcontracted_raw_materials_from_se(purchase_order, fg_item):
...
-			batch_no_concat_syntax="STRING_AGG(prsi.batch_no, ',')"
-		)
-	}, (purchase_orders, ), as_dict=1)
+	distinct_purchase_receipts = {}
+	for pr in purchase_receipts:
+		key = (pr.purchase_order, pr.item_code, pr.parent)
+		if key not in distinct_purchase_receipts:
+			distinct_purchase_receipts.setdefault(key, [''])
+
+		distinct_purchase_receipts[key].append(pr.name)
```
nabinhait: These 3 lines can be written like 
`distinct_purchase_receipts.setdefault(key, []).append(pr.name)`

erpnext/controllers/buying_controller.py
```diff
@@ -861,39 +862,45 @@ def get_subcontracted_raw_materials_from_se(purchase_order, fg_item):
...
 
 	backflushed_raw_materials_map = frappe._dict()
-	for item in backflushed_raw_materials:
-		backflushed_raw_materials_map.setdefault(item.item_key, item)
+	for args, references in distinct_purchase_receipts.items():
```
nabinhait: Use `iteritems()` for python 3

nabinhait (changes requested):

nabinhait (approved):

PR #23316, opened 2020-09-12
fix: Make Reference fields mandatory in Quality Inspection (ux)

marination:
**Issue:**
- On creating a PR(or any doc that checks QI), if QI is required an error is thrown asking to set Quality Inspection for a certain item
- Ideally the user should go to the Item's row in the Items table and click on **Create New Quality Inspection** from the QI field dropdown and then create a QI. This way the reference name and type is set
- Although if a user goes to Quality Inspection List and creates a new entry, and doesn't set Reference name and type as it isn't mandatory, user wont know how to Link the QI and Purchase Receipt
- Since theres no reference the QI wont appear in the options in the Purchase Receipt

**Fix:**
- Make Reference Type and  Reference Name mandatory as a Quality Inspection against nothing makes no sense.
 Screenshot 2020-09-12 at 4 25 31 PM

### DIFF
--- a/erpnext/stock/doctype/quality_inspection/quality_inspection.json
+++ b/erpnext/stock/doctype/quality_inspection/quality_inspection.json
@@ -73,7 +73,8 @@
    "fieldname": "reference_type",
    "fieldtype": "Select",
    "label": "Reference Type",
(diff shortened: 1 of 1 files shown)


marination:
For the reviewer: Please correct me if i'm missing a scenario

rohitwaghchaure (approved):

PR #23318, opened 2020-09-12
fix: use Plaid's new API (develop)

Alchez:
**Reference:** https://discuss.erpnext.com/t/plaid-settings-missing-plaid-public-key/64635

<hr>

**Changes:**

- Bump `plaid-python` package from 3.4.0 to 6.0.0 to utilize new API
- Switch authentication from public key to token-based
- Remove public key from Plaid Settings

### DIFF
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
@@ -2,43 +2,59 @@
 # Copyright (c) 2018, Frappe Technologies Pvt. Ltd. and contributors
 # For license information, please see license.txt
 
-from __future__ import unicode_literals
-from frappe import _
-from frappe.utils.password import get_decrypted_password
-from plaid import Client
-from plaid.errors import APIError, ItemError
+import plaid
+import requests
+from plaid.errors import APIError, ItemError, InvalidRequestError
 
 import frappe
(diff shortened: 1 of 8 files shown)


erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
```diff
@@ -2,78 +2,87 @@
...
 
+	def get_link_token(self):
+		token_request = {
+			"client_name": self.client_name,
+			"client_id": self.settings.plaid_client_id,
+			"secret": self.settings.plaid_secret,
+			"products": self.products,
+			# only allow Plaid-supported languages (LAST: Sep-12-2020)
+			"language": frappe.local.lang if frappe.local.lang in ["en", "fr", "es", "nl"] else "en",
+			"country_codes": ["US"],
```
Mangesh-Khairnar:
```suggestion
			"country_codes": ["US", "CA", "FR", "IE", "NL", "ES", "GB"],
```

Mangesh-Khairnar (changes requested):

erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
Alchez: Added the codes

PR #23326, opened 2020-09-15
refactor(Job Application): New fields in Job Applicant and webform

KanchanChauhan:
**Issues:**
- Currently, the Job Applicant doctype is missing some basic fields like Phone Number and Country of Residence.
- The web form has Attach Resume but it never gets attached to the Job Application.
- One of the major issue which is that when you press Apply Now, it has a hard-coded href to Standard Job Application which makes it impossible to have a custom Job Application.

**Refactor and New features:**

JOB APPLICANT AND JOB APPLICATION WEBFORM

- Added the required fields and also added the rating filed so that the applicant can be rated internally and decision making becomes easy.
- Also added Salary Expectations fields

Screen Shot 2020-09-15 at 12 30 44 PM

- Made changes to the Web Form as well to include the Phone Number and Country of Residence field.
- Instead of Attach Resume, added the Link Resume field. 

Screen Shot 2020-09-15 at 12 33 35 PM

**JOB OPENING CHANGES**
- Added a field job_application_route which will route to a web form. If not defined will go to Standard Job Application.
- Job Opening Changes to accommodate the Salary Range, need it ad separate section because sometimes it is not published, it is for internal use. So added a separate Publish Salary Range option.

DOCTYPE
Screen Shot 2020-09-18 at 10 11 15 AM

JOBS WEBPAGE
Screen Shot 2020-09-18 at 10 20 52 AM


sagarvora:
:+1: 
Attachment is a bigger issue that needs to be addressed separately.

Maybe we could add validation for URL and Phone. For phone it's pretty easy: just add "Phone" as `options` for the Data field.

For URL, you could use `frappe.integrations.oauth2.validate_url`

Anurag810 (commented):
please apply the suggestion provided by sagar

KanchanChauhan:
@Anurag810 Phone Number check is already there but I could not find inbuilt validate_url on js. Could you please point me to it.

sagarvora:
@KanchanChauhan Sorry for the late response. If we don't need a protocol other than http(s), we could use `frappe.utils.is_url` in JS.

We could optionally [improve it](https://stackoverflow.com/questions/5717093/check-if-a-javascript-string-is-a-url).

KanchanChauhan:
@Anurag810 Changes done

Screen Shot 2020-10-07 at 7 26 30 PM

rmehta:
I think we should remove these from the product and let people make their own Web Forms and Web Pages for recruitment - same like Homepage, About Us - etc. These are CMS pages, should not have hard coded templates in the first place.

ruchamahabal (commented):
Can you fix the Translation syntax at line number 47 in `job_opening.py`?
 
Change

```python
frappe.throw(_("Job Openings for designation {0} already open \
					or hiring completed as per Staffing Plan {1}"
					.format(self.designation, self.staffing_plan)))
```

To

```python
frappe.throw(_("Job Openings for designation {0} already open or hiring completed as per Staffing Plan {1}").format(self.designation, self.staffing_plan))
```

ruchamahabal (approved):

PR #23327, opened 2020-09-15
fix(stock_controller): precision based on `debit_in_account_currency`

iRaySpace:
**Overview**
This is a logic already added to the existing branch. However, version-11 branch don't have this code yet.
You may refer it here: https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/stock_controller.py#L74

**Reference**
This is from #20299


sagarvora:
Commit for reference: https://github.com/frappe/erpnext/commit/61ad6292d6d3f64d8e2900130fd95deb2ddcb198

rohitwaghchaure (approved):

PR #23332, opened 2020-09-15
fix: made paid amount field read only

Anurag810:
Made 'paid amount' read-only because it not to be entered by the user. If, it 'is paid' is checked it means it is fully paid. and the paid amount is calculated in backend on save .


sagarvora:
Just something to take note of: branch name is misleading as `paid_amount` is actually a field in Purchase Invoice.

deepeshgarg007 (commented):
Users can check is paid and make partial payment as well which is a valid use case. How can that be done anymore?

PR #23333, opened 2020-09-15
fix: Program and Course Enrollment fixes

ruchamahabal:
- Program enrollment date not considered while creating course enrollments automatically.

   **Steps to replicate:**
    
   1. Create a Program Enrollment.
   2. On Submitting the Program Enrollment, Course Enrollments are automatically created for all courses under the program. But the Course Enrollment Date is set as the current date and not the enrollment date

- Validate Academic Term while checking for enrollment duplication

    **Steps to replicate:**
    
    Create 2 Program Enrollments for the same student and Academic Year and different Academic Term. The second one will throw duplicate validation because the academic term was not being considered while checking for duplicates. 
  
- Added Academic Term and Year date validations 
  
   **Before:** The enrollment date is not validated with the Academic Year / Term start and end dates

   **After:** Added validations
      
     academic-year

     validation

- Create Course Enrollment only for courses added in Program Enrollment

   **Before:**
   
   On submitting a Program Enrollment, course enrollments were getting created for all the courses in that program and not just the ones for which the student is enrolling (part of the enrolled courses child table).

   **After:** 

   Course Enrollments will only be created for the courses mentioned in the `enrolled courses` child table. For example, in the following program there are 2 courses:

   programs

  But student only wants to enroll for one course out of that. So create course enrollment automatically for only these courses for that student

  course-enroll


sagarvora:
Kindly fix Travis bug: https://travis-ci.com/github/frappe/erpnext/jobs/385064761#L1056

PR #23334, opened 2020-09-15
fix: mode of payment not fetched if no account is set

nextchamp-saqib:
While creating a pos invoice or paid sales invoice, fetching mode of payments from pos profile only fetched those which had accounts set. 
Now if mode of payment doesn't have account set, then before making pos invoice user will have to set the account.

### DIFF
--- a/erpnext/accounts/doctype/pos_profile/pos_profile.py
+++ b/erpnext/accounts/doctype/pos_profile/pos_profile.py
@@ -13,7 +13,7 @@ class POSProfile(Document):
 		self.validate_default_profile()
 		self.validate_all_link_fields()
 		self.validate_duplicate_groups()
-		self.check_default_payment()
+		self.validate_payment_methods()
 
 	def validate_default_profile(self):
 		for row in self.applicable_for_users:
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/pos_profile/pos_profile.py
```diff
@@ -52,14 +52,23 @@ def validate_duplicate_groups(self):
...
-		if self.payments:
-			default_mode_of_payment = [d.default for d in self.payments if d.default]
-			if not default_mode_of_payment:
-				frappe.throw(_("Set default mode of payment"))
-
-			if len(default_mode_of_payment) > 1:
-				frappe.throw(_("Multiple default mode of payment is not allowed"))
+	def validate_payment_methods(self):
+		if not self.payments:
+			frappe.throw(_("Payment methods are mandatory. Add atleast one payment methods."))
```
sagarvora: English:

```suggestion
			frappe.throw(_("Payment methods are mandatory. Please add at least one payment method."))
```

erpnext/accounts/doctype/pos_profile/pos_profile.py
```diff
@@ -52,14 +52,23 @@ def validate_duplicate_groups(self):
...
+
+		default_mode_of_payment = [d.default for d in self.payments if d.default]
+		if not default_mode_of_payment:
+			frappe.throw(_("Set default mode of payment"))
```
sagarvora: I know you didn't modify this, but _please_ :stuck_out_tongue: 
```suggestion
			frappe.throw(_("Please select a default mode of payment"))
```

erpnext/accounts/doctype/pos_profile/pos_profile.py
```diff
@@ -52,14 +52,23 @@ def validate_duplicate_groups(self):
...
+
+		if len(default_mode_of_payment) > 1:
+			frappe.throw(_("Multiple default mode of payment is not allowed"))
```
sagarvora:
```suggestion
			frappe.throw(_("You can only select one mode of payment as default"))
```

OR

```suggestion
			frappe.throw(_("Selecting multiple modes of payment as default is not allowed"))
```

sagarvora (changes requested):

PR #23341, opened 2020-09-16
feat: multiple purchase order against sales order

Anurag810:
Allowed multiple po against so.
suppose SO have 10 items and we want to create two PO for 5 item. this use case was not possible.


marination:
https://github.com/frappe/erpnext/pull/23357 already covers partial PO creation as well

PR #23347, opened 2020-09-17
fix: multiple pos issues

nextchamp-saqib:
Depends On: https://github.com/frappe/frappe/pull/11711
### Fixes
- Item Selector doesn't shows any item if first 40 items are not in stock
  - ~Fix: show items which are not in stock to avoid empty item selector~
  - Fix: query available items from bin table
- Auto fetch button fetches expired serial nos.
- Batch No. field in Item Details queries expired batches
- Creating Opening Entry from POS doesn't saves opening balances data
- POS doesn't refresh on route change

### Enhancement
- Keyboard Shortcuts can be viewed by clicking Shift+\ or by hovering over buttons / actions / fields
- Conversion Factor control from Item Details section


erpnext/accounts/doctype/pos_profile/pos_profile.json
```diff
@@ -306,10 +307,17 @@
...
+   "fieldname": "show_only_available_items",
+   "fieldtype": "Check",
+   "label": "Show only available items"
   }
  ],
  "icon": "icon-cog",
  "idx": 1,
+ "index_web_pages_for_search": 1,
  "links": [],
  "modified": "2020-10-01 17:29:27.759088",
```
rohitwaghchaure: As modified date is not updated getting below error while creating the POS Opening Entry

Screenshot 2020-10-13 at 11 59 58 PM

Screenshot 2020-10-13 at 11 59 52 PM

rohitwaghchaure:
Screenshot 2020-10-14 at 12 12 59 AM

Why difference is Zero?

rohitwaghchaure:
pos_testing

nextchamp-saqib:
> pos_testing

Fixed

PR #23357, opened 2020-09-18
fix: SO to PO flow improvement

marination:
## Enhancements on checkbox to create PO from SO
- Added checkbox for user to decide whether PO to be created for default supplier or a usual PO is to be created where user chooses supplier
  

-  On enabling checkbox,only rows with Default Supplier against them will be visible.
   
- Supplier field is added which will show Default Suppliers against Items
- Pending Qty field added which will show pending qty to order. It will consider already ordered qty

## Behaviour

<details><summary>Checkbox disabled, normal PO creation</summary>

All items will be mapped to a new PO and user will be redirected there.
  so-to-po

</details>

<details><summary>Checkbox enabled</summary>

POs created for each Supplier. So, if there were two item rows belonging to the same supplier, both will come under one PO for their supplier

def-supp-po

</details>

<details><summary>Partial PO creation</summary>

The Pending qty column will show exact amount pending, pending amount will be mapped

  partial-so-to-po-supplier

</details>


Anurag810 (commented):
please also take this use case under consideration.
https://github.com/frappe/erpnext/pull/23341

nabinhait (approved):

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -822,24 +816,21 @@ def set_missing_values(source, target):
...
-		suppliers.append(for_supplier)
-	else:
-		sales_order = frappe.get_doc("Sales Order", source_name)
-		for item in sales_order.items:
-			if item.supplier and item.supplier not in suppliers:
-				suppliers.append(item.supplier)
+	suppliers = [item.get('supplier') for item in selected_items if item.get('supplier') and item.get('supplier')]
+	suppliers = list(set(suppliers))
+
+	items_to_map = [item.get('item_code') for item in selected_items if item.get('item_code') and item.get('item_code')]
```
marination, line 826: 🫤

PR #23366, opened 2020-09-20
fix: stock reconciliation, incorrect serial nos fetched in the current serial no field

rohitwaghchaure:
### **Issue**

1. Create new item "Surface Disinfectant" with has serial no and set "Serial Number Series" as "SD.#####"

1. Create and submit the purchase receipt with item as "Surface Disinfectant" and quantity as 5, so system will create the serial no as SD00001, SD00002, SD00003, SD00004, SD00005

1. Create and submit delivery note for the item "Surface Disinfectant" with quantity as 3 and serial no as "SD00001", "SD00002", "SD00004"

1. The create the stock reconciliation for the item "Surface Disinfectant", system pulls the incorrect serial no in the current serial no field

Screenshot 2020-09-21 at 1 05 18 AM

### **After Fix**
Screenshot 2020-09-21 at 1 31 13 AM


rohitwaghchaure:

PR #23374, opened 2020-09-21
fix: use plaid's new API

Mangesh-Khairnar:
- backport of https://github.com/frappe/erpnext/pull/23318


Alchez:
@Mangesh-Khairnar, I've already made this at #23317

PR #23391, opened 2020-09-22
fix: leave application status fix

anupamvs:
making status field read only if application is submitted and canceled.

ref- #23043


Anurag810 (changes requested):
You can check both **Allow on submitt** and **Read only**  on doctype level. why writing this on js.

rohitwaghchaure:
This is not proper fix, why allow on submit enabled for the status field?

rohitwaghchaure (approved):

PR #23395, opened 2020-09-22
fix: quality procedure parent

anupamvs:
Issue - the user is not able to change the parent of quality procedure
image

After Fix - 
image


Anurag810 (changes requested):
you also need to set that procedure in the New Parent process and also remove it from the old parent process. 
You need to handle the child procedure in the process table

rohitwaghchaure:
@anupamvs kindly rebuild tree again if user changed the parent procedure

rohitwaghchaure (approved):

PR #23398, opened 2020-09-22
fix: quality procedure parent

anupamvs:
port of #23395


Anurag810 (commented):
you also need to set that procedure in the New Parent process and also remove it from the old parent process. 
You need to handle the child procedure in the process table

rohitwaghchaure:
@anupamvs rebuild the tree after change.

rohitwaghchaure (approved):

PR #23402, opened 2020-09-22
feat: Allow rich text edit for questions

madar2020:
The original text input for question does not allow rich text input 
p0
Therefore we have change Question doctype the type of question from small text to text editor so it becomes like this
p1
and we need to change the Quiz Question doctype also so we get proper out in the list view.
p3
the last thing we need to do is to allow the rich text to be correctly displayed when the quiz is displayed. we need to change how quiz.js in interpreting the question from innerText to innerHTML.
p4
will be changed to this
p5


erpnext/education/doctype/quiz_question/quiz_question.json
```diff
@@ -20,7 +20,7 @@
   {
    "fetch_from": "question_link.question",
    "fieldname": "question",
-   "fieldtype": "Data",
+   "fieldtype": "Text Editor",
```
ruchamahabal, line 23: This is the child table in Quiz doctype which fetches the question from **Question** doctype. So you will have to change the fieldtype to Text Editor over there too.

question

ruchamahabal (changes requested):
For the doctype changes to get synced, you will have to change the [modified](https://github.com/frappe/erpnext/pull/23402/files#diff-621ced780db0a7faa9530428d144ac8fR30) timestamp in JSON files too. The ideal way to edit a doctype is by making changes in the doctype view and saving it instead of manually changing JSON files. All the changes with the modified timestamp are exported to JSON if you are in developer mode.

erpnext/education/doctype/quiz_question/quiz_question.json
madar2020: What you have mentioned does make sense. It is supposed to be like that. however I did not notice any change in behavior in that

madar2020 (commented):
thats ok

erpnext/education/doctype/quiz_question/quiz_question.json
```diff
@@ -20,7 +20,7 @@
...
    "in_list_view": 1,
    "label": "Question",
    "read_only": 1
```
ruchamahabal, line 26: The field over here is `read_only`. Questions are created in the Question doctype so the need for fieldtype to be Text Editor is over there. This field just fetches the content from there. So both of them need to be of type Text Editor. Also as I mentioned before, please change the modified timestamp otherwise these changes won't get synced.

madar2020 (commented):
question doctype have been changed as requested and timestamp updated.

ruchamahabal:
@madar2020 Can you please update the description with screenshots and explanation?

madar2020:
> @madar2020 Can you please update the description with screenshots and explanation?

Done

madar2020 (commented):
done

ruchamahabal (approved):

PR #23407, opened 2020-09-23
fix: Conversion factor fixes in Stock Entry

deepeshgarg007:
Currenty in Stock Entry the UOM field is editable but the system doesn't allows you to change the conversion factor and always  overrides it with 1

There may be cases where the `stock_uom` and `uom` (transaction UOM) may be different and the conversion factor might be different

**Solution**:
1. Made conversion factor editable
2. Override conversion factor  with 1 only if transaction UOM and Stock UOM are same


nabinhait (dismissed):

rohitwaghchaure:
@deepeshgarg007 add test case that user will be able to change the default conversion factor

deepeshgarg007:
@rohitwaghchaure any updates on it?

rohitwaghchaure (approved):

PR #23408, opened 2020-09-23
fix: Item price duplicate checking

kevingdc:
This pull request fixes a bug in the **Item Price** DocType where some prices aren't being saved because they are said to be duplicates even though they're not since they don't have values in a specific fields. This happens because the `check_duplicates()` method doesn't check for null or empty values. 

An example of this is if I save a price that has a supplier, then duplicate that, remove the supplier, and then save it. The previous code will throw an error saying that it's a duplicate price since the system doesn't add an `and` condition for null or empty suppliers and finds the initial price with the supplier. 

Here are some screenshots to illustrate this:

The first item price:
Screen Shot 2020-09-23 at 3 48 41 PM

Trying to save the second price:
Screen Shot 2020-09-23 at 3 48 57 PM

This pull request fixes this issue by adding conditions to check for null or empty values if it finds that the field is empty.


erpnext/stock/doctype/item_price/item_price.py
```diff
@@ -49,27 +48,59 @@ def update_item_details(self):
...
+					`tabItem Price`
+				{conditions}
+			""".format(
+				conditions=conditions
+			),
+			self.as_dict(),
+		)
+
+		if price_list_rate:
+			frappe.throw(_("""
```
rmehta: translatable strings should be in a single line.

erpnext/stock/doctype/item_price/item_price.py
```diff
@@ -49,27 +48,59 @@ def update_item_details(self):
...
+				)
+
+		price_list_rate = frappe.db.sql(
+			"""
+				select
+					price_list_rate
+				from
+					`tabItem Price`
+				{conditions}
+			""".format(
```
rmehta: please collapse the whitespace here.

erpnext/stock/doctype/item_price/item_price.py
```diff
@@ -49,27 +48,59 @@ def update_item_details(self):
...
-			frappe.throw(_("Item Price appears multiple times based on Price List, Supplier/Customer, Currency, Item, UOM, Qty and Dates."), ItemPriceDuplicateItem)
+				conditions += " and {0} = %({0})s ".format(field)
+			else:
+				conditions += """
+						and (
+							isnull({0})
+							or {0} = ''
+						)
+					""".format(
+					field
```
rmehta: no whitespace for brackets.

rmehta (commented):
please add a relevant test case

rohitwaghchaure:
@kevingdc as there is conflicts in the item_price.py, kindly do rebase and resolve the conflicts

kevingdc:
> @kevingdc as there is conflicts in the item_price.py, kindly do rebase and resolve the conflicts

Hi @rohitwaghchaure! Merged the latest develop branch and merged conflicts :)

rohitwaghchaure (approved):

PR #23410, opened 2020-09-23
fix(Item): Change Button Labels in View dropdown

michellealva:
**Before:**

image

**After:**

image

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -26,19 +26,19 @@ frappe.ui.form.on("Item", {
 
 	refresh: function(frm) {
 		if (frm.doc.is_stock_item) {
-			frm.add_custom_button(__("Balance"), function() {
+			frm.add_custom_button(__("Stock Balance"), function() {
 				frappe.route_options = {
 					"item_code": frm.doc.name
 				}
 				frappe.set_route("query-report", "Stock Balance");
 			}, __("View"));
-			frm.add_custom_button(__("Ledger"), function() {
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item/item.js
```diff
@@ -26,19 +26,19 @@ frappe.ui.form.on("Item", {
...
 			}, __("View"));
-			frm.add_custom_button(__("Ledger"), function() {
+			frm.add_custom_button(__("Stock Ledger"), function() {
 				frappe.route_options = {
 					"item_code": frm.doc.name
 				}
 				frappe.set_route("query-report", "Stock Ledger");
 			}, __("View"));
-			frm.add_custom_button(__("Projected"), function() {
+			frm.add_custom_button(__("Stock Projected Qty"), function() {
```
abhishekbalam, line 41: Change the "Stock Projected Qty" to "Stock Projected Quantity", because everywhere else we don't use short forms.

abhishekbalam (changes requested):

erpnext/stock/doctype/item/item.js
michellealva: The reason I named it Stock Projected Qty is because the report itself is called "Stock Projected Qty". Should we change the report name to "Stock Projected Quantity"?

abhishekbalam: There are too many references to the report.
Let's keep it as is.

abhishekbalam (approved):

Anurag810 (approved):

PR #23412, opened 2020-09-23
refactor: book loss amount in the COGS instead of stock received but not billed

rohitwaghchaure:
**Issue**

1. Enable allow negative stock in stock settings

1. Create the test item with valuation rate as 200 and then create delivery note two times with 1 quantity

1. The create the purchase receipt with quantity as 2 and incoming rate as 500 (see below image)
Screenshot 2020-09-23 at 7 25 24 PM

Check GL entry and you can see the loss amount(600) has been booked under Stock Received But Not Billed account

Screenshot 2020-09-23 at 7 19 15 PM

As while making delivery note user has used the valuation rate as "200" therefore while making inward entry system has used the same rate and not the incoming rate 500. So the difference amount 1000 - 400 = 600 needs to be booked under COGS but as system booking it under Stock Received but not billed account, there is always balance remained in the Stock Received but not billed account which is incorrect

**After Fix**

Screenshot 2020-09-23 at 7 24 39 PM

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -207,6 +207,7 @@ class PurchaseReceipt(BuyingController):
 		from erpnext.accounts.general_ledger import process_gl_map
 
 		stock_rbnb = self.get_company_default("stock_received_but_not_billed")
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:

PR #23415, opened 2020-09-23
fix: Posting Date bug in load_defaults

josephalbaph:
this.frm.posting_date is always invalid and should be changed to this.frm.doc.posting_date

The effect of this bug fix is, a default Posting Date value may now be set in Custom Script's onload event, and the default value will be honored.
Example: (Assuming posting date has been included in standard filter)
```
frappe.ui.form.on('Journal Entry', {
	before_load(frm) {
	    var posting_date = $("input[data-fieldname='posting_date']")[0].value
	    posting_date = moment(posting_date)._d
	    frm.set_value('posting_date', posting_date )
	}
})
```

Without the fix, the posting date will always be today's date. With the bug fix, the default value for posting date which is taken from the posting date's Standard Filter vale is honored.

> Screenshots/GIFs


sagarvora (approved):

josephalbaph:
@netchampfaris @deepeshgarg007 
Can this pull request be reviewed? Thanks

sagarvora:
@Mergifyio refresh

nabinhait (approved):

PR #23416, opened 2020-09-23
feat: provision to draft quotation from portal

anupamvs:
Issue - 

- When users create a quotation form portal it is going to submitted directly. The current flow will not useful when product pricing is varying.

- UOM is not shown on the portal if we are not showing the item price.

Fixes-

- Added check in Shopping Cart Settings 'Draft Quotation', whether quotation has to save or submit.

image

image

image

- Showing item UOM even price is hidden.
image


erpnext/shopping_cart/doctype/shopping_cart_settings/shopping_cart_settings.json
```diff
@@ -166,13 +167,20 @@
    "fieldname": "enable_variants",
    "fieldtype": "Check",
    "label": "Enable Variants"
+  },
+  {
+   "default": "0",
+   "depends_on": "eval: doc.enable_checkout == 0",
+   "fieldname": "draft_quotation",
+   "fieldtype": "Check",
+   "label": "Draft Quotation"
```
deepeshgarg007: "Draft Quotation" is not a proper label. Write a more descriptive label to the user. Also add description what enabling and disabling this check will do. Since this is dependent on enable checkout it should be in the checkout section, display settings is not the appropriate place

erpnext/shopping_cart/cart.py
```diff
@@ -96,7 +96,7 @@ def place_order():
 def request_for_quotation():
 	quotation = _get_cart_quotation()
 	quotation.flags.ignore_permissions = True
-	quotation.submit()
+	quotation.save() if get_shopping_cart_settings().draft_quotation else quotation.submit()
```
deepeshgarg007: Break this into two lines. Save the quotation. If draft is not checked then submit as well. 
For eg:

```suggestion
	quotation.save() 
	if not get_shopping_cart_settings().draft_quotation: 
	    quotation.submit()
```

deepeshgarg007 (changes requested):

erpnext/templates/generators/item/item_add_to_cart.html
```diff
@@ -10,7 +10,10 @@ <h4>
 			{{ product_info.price.formatted_price_sales_uom }}
 			<small class="text-muted">({{ product_info.price.formatted_price }} / {{ product_info.uom }})</small>
 		</h4>
+		{% else %}
+			{{ _("UOM") }} : {{ product_info.uom }}
```
deepeshgarg007: Don't use abbreviation here, some may not understand what is UOM, use full form

PR #23429, opened 2020-09-24
feat: (RFQ/SQ) Link to Material Requests in Tools section

marination:
Docs: https://github.com/frappe/erpnext_documentation/pull/212

**Before:**
Screenshot 2020-10-16 at 6 02 19 PM

**After:**
- Removed **Link to Material Requests** button from under items table in Request for Quotation and Supplier Quotation
- Shifted the same to a Tools section next to **Get Items From**
  Screenshot 2020-09-24 at 3 42 40 PM 
 Screenshot 2020-09-24 at 3 43 26 PM

Works (as it used to):
link-to-mr


erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
```diff
@@ -326,6 +326,11 @@ erpnext.buying.RequestforQuotationController = erpnext.buying.BuyingController.e
 				d.show();
 			}, __("Get items from"));
 
+			//Link Material Requests
```
sagarvora:
```suggestion
			// Link Material Requests
```
Sorry, couldn't resist.

erpnext/buying/doctype/supplier_quotation/supplier_quotation.js
```diff
@@ -54,6 +54,12 @@ erpnext.buying.SupplierQuotationController = erpnext.buying.BuyingController.ext
 						}
 					})
 				}, __("Get items from"));
+
+			//Link Material Requests
```
sagarvora:
```suggestion
			// Link Material Requests
```

erpnext/public/js/controllers/buying.js
```diff
@@ -422,6 +359,62 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 
 cur_frm.add_fetch('project', 'cost_center', 'cost_center');
 
+erpnext.buying.link_to_mrs = function(frm) {
+	frappe.call({
+		method: "erpnext.buying.utils.get_linked_material_requests",
+		args:{
+			items: frm.doc.items.map((item) => {return item.item_code;})
```
sagarvora:
```suggestion
			items: frm.doc.items.map((item) => item.item_code);
```

erpnext/public/js/controllers/buying.js
```diff
@@ -422,6 +359,62 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+		},
+		callback: function(r) {
+			if(!r.message || r.message.length == 0) {
+				frappe.throw({message: __("No pending Material Requests found to link for the given items."), title: __("Note")});
```
sagarvora: Is this really an error? Shouldn't this be `msgprint`?
```suggestion
				frappe.msgprint( __("No pending Material Requests found to link for the given items."), __("Note"));
				return;
```

erpnext/public/js/controllers/buying.js
```diff
@@ -422,6 +359,62 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+			}
+			else {
```
sagarvora: We can remove the `else` block if we're returning from above `if` block. Not needed.

erpnext/public/js/controllers/buying.js
```diff
@@ -422,6 +359,62 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+				var i = 0;
+				var item_length = frm.doc.items.length;
+				while (i < item_length) {
+					var qty = frm.doc.items[i].qty;
```
sagarvora: We are using `frm.doc.items[i]` a lot! Should be a variable.

sagarvora (changes requested):
I know you aren't the original author, but please fix this if it isn't too much work.

PR #23431, opened 2020-09-24
fix: include item_code in items result to allow adding product info in custom templates

beeonthego:
In some cases, users/developers may want to display prices on product list page by creating custom page templates. The method erpnext.shopping_cart.product_info.set_product_info_for_website can add product info to items in the page context. 
set_product_info_for_website relies on item_code to fetch product info. Including item_code in the query result will make it possible.


sagarvora:
Please ensure correct base branch.

PR #23449, opened 2020-09-29
fix: Stock and account value out of sync on Stock Entry cancellation with serial nos

deepeshgarg007:
On any outgoing Stock Ledger Entry containing Serial Nos the outgoing values is decided based on the incoming value of the serial no 
https://github.com/frappe/erpnext/blob/develop/erpnext/stock/stock_ledger.py#L239

While cancelling a Stock Entry for Manufacture containing a serialized finished or semi finished good on submission of cancellation SLE following method is called and purchase rate in Serial No becomes 0 or None before the outgoing rate and stock value difference is updated in SLE and a stock and account value mismatch occurs.

https://github.com/frappe/erpnext/blob/develop/erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py#L44

**Solution**: Get incoming rate from original SLE on cancellation instead of Serial No master


rohitwaghchaure:
@deepeshgarg007 please put the description, also possible add the screenshots of the error

PR #23450, opened 2020-09-29
fix: Change Error Message in Work Order

michellealva:
Before:

image

After:

Changed message to: **Completed Qty cannot be greater than 'Qty to Manufacture'**

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.py
+++ b/erpnext/manufacturing/doctype/work_order/work_order.py
@@ -434,7 +434,7 @@ class WorkOrder(Document):
 			elif flt(d.completed_qty) <= max_allowed_qty_for_wo:
 				d.status = "Completed"
 			else:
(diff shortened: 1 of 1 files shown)


sagarvora (approved):

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -434,7 +434,7 @@ def update_operation_status(self):
 			elif flt(d.completed_qty) <= max_allowed_qty_for_wo:
 				d.status = "Completed"
 			else:
-				frappe.throw(_("Completed Qty can not be greater than 'Qty to Manufacture'"))
+				frappe.throw(_("Completed Qty cannot be greater than 'Qty to Manufacture'"))
```
marination, line 437: Add a fitting title too

PR #23453, opened 2020-09-29
feat: validate Sales Invoice for Germany

barredterra:
- Validate **Sales Invoice** for Germany. Generic approach that allows to add validation for all DocTypes based on AccountsController.
- Friendly approach: the user can still create the invoice however he likes, but the system reminds him of the correct way.
- Add required field "Company Tax ID" to **Sales Invoice**

Legal basis: https://www.gesetze-im-internet.de/ustg_1980/__14.html


rohitwaghchaure:
@barredterra conflicts sales_invoice.json, kindly rebase your PR

erpnext/regional/germany/accounts_controller.py
```diff
@@ -0,0 +1,57 @@
...
+		field_name = field.get("field_name")
+		regulation = field.get("regulation")
+		if field_name and not doc.get(field_name):
+			missing(field_map.get(field_name), regulation)
+
+
+def missing(field_label, regulation):
+	"""Notify the user that a required field is missing."""
+	context = 'Specific for Germany. Example: Remember to set Company Tax ID. It is required by § 14 Abs. 4 Nr. 2 UStG.'
+	msgprint(_('Remember to set {field_label}. It is required by {regulation}.', context=context).format(
```
rohitwaghchaure, line 53: msgprint will allow user to submit the sales invoice, just want to display the message and not wants to stop users to submit the sales invoice?

barredterra: Yes, the users can do whatever they like. It's just a reminder how things should be done to comply with local laws. This way we don't have to worry too much about edge cases where the laws might not apply.

rohitwaghchaure (approved):

PR #23454, opened 2020-09-29
fix(template): cards on second row overlaps the one before 

madar2020:
fix second row of cards overlaps with the one before

- before the fix the home page with section added as cards looks like this:

p1

- after applying the fix it will look like this

p2


madar2020 (commented):
fix(template): second row of cards overlaps with the one before

rohitwaghchaure:
@madar2020 screenshots are not properly visible, also in the screenshot of the fix the EDU is not completely visible.

madar2020:
ok

nabinhait (approved):

PR #23460, opened 2020-09-29
fix: Avoid validation of group cost center on repost

marination:
- On cancelling a backdated submitted entry, when dealing with old data, it throws:  **Cost Center <name> is a group cost center and group cost centers cannot be used in transactions**
- Here, since the system reposts GL entries of future transactions that already are submitted. If these have a group cost center too, user cant cancel this transaction.
- Neither can user change the cost center

**Fix:**
- While validating group cost center check if its via reposting or not. If via reposting, don't validate
- Essentially only validate this condition for newly created transactions


deepeshgarg007 (commented):
Can't we just ignore this validation while reposting? There is already a flag available.

deepeshgarg007 (approved):

PR #23463, opened 2020-09-29
fix: tds calculation, skip invoices with "Apply Tax Withholding Amount" has disabled

rohitwaghchaure:
**Issue**

Three invoices against supplier A is as follows -

1) 8,688 (Apply TDS - yes)
2) 48,244 (Apply TDS - no)
3) 822 (Apply TDS - yes) -> System has applied tds by considering the amount of second invoice

TDS invoices amount = 9510 and TDS should not be deduct from the 3rd invoice but system has deducted


deepeshgarg007:
@rohitwaghchaure Test case is failing

rohitwaghchaure:
Screenshot 2020-10-03 at 4 15 33 PM

rohitwaghchaure:
@Mergifyio backport develop

deepeshgarg007 (approved):

PR #23465, opened 2020-09-29
fix(Accounts): Add Currency filter in Trial Balance report

abhishekbalam:
image

image

### DIFF
--- a/erpnext/accounts/report/trial_balance/trial_balance.js
+++ b/erpnext/accounts/report/trial_balance/trial_balance.js
@@ -76,8 +76,7 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 				"fieldname": "presentation_currency",
 				"label": __("Currency"),
 				"fieldtype": "Select",
-				"options": erpnext.get_presentation_currency_list(),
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/trial_balance/trial_balance.js
```diff
@@ -72,6 +72,13 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 				"fieldtype": "Link",
 				"options": "Finance Book",
 			},
+			{
+				"fieldname": "presentation_currency",
+				"label": __("Currency"),
+				"fieldtype": "Select",
+				"options": erpnext.get_presentation_currency_list(),
+				"default": frappe.defaults.get_user_default("currency")
```
deepeshgarg007: By default it should show in Company currency and not in user currency. This can be removed

abhishekbalam: Ok done.

deepeshgarg007 (approved):

aymenit2008:
The opening  and closing balance  not change currency

PR #23471, opened 2020-09-30
fix: currency symbol in profit and loss report

rohitwaghchaure:
**Issue**

Screenshot 2020-09-30 at 4 38 10 PM

**After Fix**

Screenshot 2020-09-30 at 5 43 13 PM

### DIFF
--- a/erpnext/accounts/report/profit_and_loss_statement/profit_and_loss_statement.py
+++ b/erpnext/accounts/report/profit_and_loss_statement/profit_and_loss_statement.py
@@ -32,12 +32,12 @@ def execute(filters=None):
 
 	chart = get_chart_data(filters, columns, income, expense, net_profit_loss)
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/profit_and_loss_statement/profit_and_loss_statement.py
```diff
@@ -32,7 +32,7 @@ def execute(filters=None):
 
 	chart = get_chart_data(filters, columns, income, expense, net_profit_loss)
 
-	default_currency = frappe.get_cached_value('Company', filters.company, "default_currency")
+	default_currency = filters.presentation_currency or frappe.get_cached_value('Company', filters.company, "default_currency")
```
sagarvora: maybe rename the variable to `currency`:
```suggestion
	currency = filters.presentation_currency or frappe.get_cached_value('Company', filters.company, "default_currency")
```

PR #23476, opened 2020-10-01
fix: Payment Schedule not fetching

anupamvs:
**issue:** Payment Schedule is not fetched from Customer Master in Sales Invoice.
before

**After Fix-**
after


erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -345,13 +345,14 @@ def create_invoice(self, prorate):
...
 
 		# Due date
-		invoice.append(
-			'payment_schedule',
-			{
-				'due_date': add_days(invoice.posting_date, cint(self.days_until_due)),
-				'invoice_portion': 100
-			}
-		)
+		if not frappe.get_value(self.party_type, self.party, 'payment_terms'):
```
deepeshgarg007:
```suggestion
		if self.days_until_due:
```
Default payment terms template can be there in Company master, Customer and Supplier Group, Customer and Supplier, so this check won't work. If days_until_due is mentioned then that should be honoured otherwise the default template should be applied.

deepeshgarg007 (changes requested):
Make the requested changes

deepeshgarg007 (approved):

PR #23478, opened 2020-10-01
feat: Supplier Quotation UX fixes

marination:
**In Parent:**
- Removed **Subscription** option from **Create** dropdown
- Added **Date** and **Valid Till** to List View

**In Child Table:**
- Shifted  **Project** to section **Accounting Dimension**
- Added **Expected Delivery Date** which will be mapped into the same field in PO
- **Is Free Item** visible only if it has some value and is shifted to first section
- Shifted Manufacture section to below Warehouse and Reference
- Created separate section for **Price List Rate, Discount Amount**, etc
- Renamed **Lead Time in days** -> **Supplier Lead Time (days)**
- Removed Stock UOM from Grid View. Added **UOM** and **Amount** instead

**Dialog:**
- Added Get items from **Request for Quotation**. Dialog fetches RFQs that have current doc Supplier in it. If no Supplier user will be prompted to select the same first.

ToDo:

- [x] Custom Query Get Items from RFQ in Supplier Quotation


erpnext/buying/doctype/request_for_quotation/request_for_quotation.py
```diff
@@ -354,3 +354,32 @@ def get_supplier_tag():
...
+			`tabRequest for Quotation` rfq, `tabRequest for Quotation Supplier` rfq_supplier
+		where
+			rfq.name = rfq_supplier.parent
+			and rfq_supplier.supplier = '{0}'
+			and rfq.docstatus = 1
+			and rfq.company = '{1}'
+			{2}
+		order by rfq.transaction_date ASC
+		limit %(page_len)s offset %(start)s """ \
+		.format(filters.get("supplier"), filters.get("company"), conditions),
```
gavindsouza, line 382: - Pass as args instead of .format
- Use ORM whenever possible. ```frappe.db.get_all('DocType', fields= ['`tabDocField`.fieldname'], filters = [['DocField', 'label', 'like', '%doctype%']])```

PR #23480, opened 2020-10-01
fix: incorrect operation time calculation for batch size

rohitwaghchaure:
### **Issue**

1. Create bom with operations

1. Set operation time as 40 mins and batch size as 5

1. So for Two quantity operation time should be 16 mins but system has calculated incorrect time in work order
Screenshot 2020-10-01 at 11 56 34 PM

### **After Fix**
Operation time for 2 finished good quantity in the work order

Screenshot 2020-10-01 at 11 54 09 PM


rohitwaghchaure:
@mergifyio backport develop

rohitwaghchaure:
### Add test case
Screenshot 2020-10-05 at 4 57 06 PM

erpnext/manufacturing/doctype/work_order/test_work_order.py
```diff
@@ -371,6 +371,49 @@ def test_work_order_with_non_transfer_item(self):
 		ste1 = frappe.get_doc(make_stock_entry(wo.name, "Manufacture", 1))
 		self.assertEqual(len(ste1.items), 3)
 
+	def test_work_order_with_batch_size(self):
```
marination: lets rename this to `test_job_card_time_with_batch_size` ? The test should describe what we're testing. Rest looks good

marination (approved):

marination:
@rohitwaghchaure please port manually

govindsmenokee:
@rohitwaghchaure This is incorrect. If you specify batch size the the operation time will be considered for the batch as a whole. This was added to handle operations like Baking in manufacturing where the operation will take 1 hr even if there are 5 or 10 boards in the oven. To handle your case where only 2 items have to be produced please use batch size of 1(default).
With this change now the entire purpose is defeated.

PR #23482, opened 2020-10-02
fix: also update employee field when renamed

### LINKED ISSUE #16148: No after_rename in Employee doctype
Version:
> ERPNext: v10.1.73 (master)
> Frappe Framework: v10.1.66 (master)

Steps to Reproduce:
_Rename_ from menu items in **Employee** form.

Actual behavior:
`name` field in db gets renamed. `employee` field remains unaffected.

Expected behavior:
`employee` field should also be renamed to be the same as `name`.

Screenshots:
github_issue-employee_rename

Notes:
Found this while trying to set attendance in **Employee Attendance Tool** for a renamed **Employee**. Worst case scenario - another **Employee** with the old employee id exists and is marked for attendance.

sunhoww:
Fixes #16148. Although originally this issue was found in v10, subsequent versions continue to have this behaviour.

### DIFF
--- a/erpnext/hr/doctype/employee/employee.py
+++ b/erpnext/hr/doctype/employee/employee.py
@@ -56,6 +56,9 @@ class Employee(NestedSet):
 			if existing_user_id:
 				remove_user_permission(
 					"Employee", self.name, existing_user_id)
+	
+	def after_rename(self, old, new, merge):
+		self.db_set("employee", new)
 
 	def set_employee_name(self):
 		self.employee_name = ' '.join(filter(lambda x: x, [self.first_name, self.middle_name, self.last_name]))


sagarvora:
As far as I know, we have this field only for backwards compatibility.
@nabinhait Shouldn't this field go away in `v13`?

sunhoww:
This is the line in **Employee Attendance Tool** which still reference the `employee` field. Should I have updated this instead? 

https://github.com/frappe/erpnext/blob/262590b6726854ed0578666783a346aeaa33105c/erpnext/hr/doctype/employee_attendance_tool/employee_attendance_tool.py#L27

nabinhait (approved):

PR #23483, opened 2020-10-02
fix: Handle Blank from/to range in Numeric Item Attribute

marination:
- On leaving from range blank:
   ```
         elif self.from_range >= self.to_range:
      TypeError: '>=' not supported between instances of 'str' and 'int' 
   ```

**Fix:**
- Added default value as 0 from from range, to range and increment in Item Attribute doctype
- Convert to float before comparing on validate


sagarvora:
Fields of type `Float` already default to zero:

Screenshot from 2020-10-02 13-15-25

I don't think we need to set a default manually.

marination:
@sagarvora in the db after save , yes.
 It will still be a blank field on the front-end when you haven't yet saved your new Item Attribute.
Now atleast it pre-populates 0, so the user knows what kind of data is expected. Lesser cognitive load.

sagarvora:
@marination Thank you for the clarification.

rohitwaghchaure (approved):

PR #23487, opened 2020-10-02
fix: Check if show in website is enabled for a Variant

marination:
**Issue:**
- Via _Data Import_ users can enable **Show Website** and **Show Website (Variant)** both for a non variant Item
- So, even when **Show Website** is disabled Item is still visible in the Portal due to **Show Website (Variant)**
- Such an Item shouldn't be allowed to have **Show Website (Variant)** enabled as it isn't visible via the form view

**Fix:**
- Validate if the right 'show website' is enabled
Screenshot 2020-10-02 at 3 53 29 PM


erpnext/stock/doctype/item/item.py
```diff
@@ -204,6 +205,17 @@ def make_route(self):
 			return cstr(frappe.db.get_value('Item Group', self.item_group,
 					'route')) + '/' + self.scrub((self.item_name if self.item_name else self.item_code) + '-' + random_string(5))
 
+	def validate_show_website(self):
+		# check if the right 'show in website' is enabled
+		if self.variant_of and self.show_in_website:
+			show_in_website = frappe.bold(_("Show in Website"))
+			frappe.throw(_("Item {0} is a Variant. {1} cannot be enabled for an Item Variant.")
+				.format(frappe.bold(self.item_code), show_in_website))
```
rohitwaghchaure, line 213: Can we enabled show_variant_in_website and disabled show_in_website for variant item if show_in_website has enabled?

PR #23488, opened 2020-10-02
fix: Check if show in website is enabled for a Variant

marination:
Port of https://github.com/frappe/erpnext/pull/23487


rohitwaghchaure:
Added comment in the main PR

PR #23495, opened 2020-10-03
fix: check fiscal year

barredterra:
- "Start Date" and "End Date" have to be in the same **Fiscal Year**.
- Use the start date of the identified **Fiscal Year** in header.

### DIFF
--- a/erpnext/regional/report/datev/datev.py
+++ b/erpnext/regional/report/datev/datev.py
@@ -16,6 +16,7 @@ from csv import QUOTE_NONNUMERIC
 
 import frappe
 from frappe import _
+from erpnext.accounts.utils import get_fiscal_year
 import pandas as pd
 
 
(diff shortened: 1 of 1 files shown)


erpnext/regional/report/datev/datev.py
```diff
@@ -507,6 +521,9 @@ def download_datev_csv(filters=None):
 		filters = json.loads(filters)
 
 	validate(filters)
+	fiscal_year = get_fiscal_year(date=filters.get('from_date'), company=filters.get('company'))
+	filters['fiscal_year_start'] = fiscal_year[1]
```
rohitwaghchaure, line 525: @barredterra better design will be 

1. Add fiscal year filters

1. Based on the selected fiscal year, auto set the from date and to date in the filters

1. Add validation that from date, to date should have the same fiscal year

rohitwaghchaure:
@barredterra kindly fixed the code and reopen the PR

erpnext/regional/report/datev/datev.py
barredterra: Thanks for the review.

I don't share your conclusion. Export can be for an arbitrary time period (for example, a month or a week) within the same fiscal year. Setting the From Date and To Date based on fiscal year implies that the export is only for a whole year.

We can still use it as a filter so that the user knows exactly which fiscal year is used. But it doesn't help much, i think.

rohitwaghchaure (approved):

PR #23497, opened 2020-10-03
fix: fiscal year can be shorter than 12 months

barredterra:
A **Fiscal Year** can be shorter than 12 months, in some jurisdictions under certain circumstances.

For example, in the [USA](https://www.law.cornell.edu/uscode/text/26/443) and [Germany](https://de.wikipedia.org/wiki/Gesch%C3%A4ftsjahr#Rumpfgesch%C3%A4ftsjahr). A company may begin operations and start it's fiscal year in april. Now it can have a  _short fiscal year_ until december. Then it may switch to a "normal" fiscal year from january to december.

Currently, ERPNext does not allow a **Fiscal Years** with less than 12 months. This PR introduces a checkbox "Is Short Year" that bypasses this validation.


barredterra:
@Mergifyio backport develop

rohitwaghchaure (approved):

PR #23500, opened 2020-10-03
fix(HR): Appraisal form Enhancements

abhishekbalam:
The "Calculate Total Score" button does not look good and 
it feels obtrusive. An alternate solution without using the button.
The calculation is triggered by the child table field change events.

Before:
image

After:

[GIF Link](http://frappe-gifs.surge.sh/gifs/appraisal.gif)

Also made some edge case fixes where the Total Score was being set to NaN and improved a validation error message.


erpnext/hr/doctype/appraisal/appraisal.js
```diff
@@ -10,45 +10,75 @@ frappe.ui.form.on('Appraisal', {
...
-			d.score_earned = 0;
+		if (flt(d.score) > 5) {
+			frappe.msgprint(__("Score must be less than or equal to 5"));
+			d.score = 0;
+			refresh_field('score', d.name, 'goals');
+		}
+		else {
+			frm.trigger('set_score_earned');
 		}
+<<<<<<< Updated upstream
```
Anurag810: Please resolve conflicts

Anurag810 (changes requested):

erpnext/hr/doctype/appraisal/appraisal.js
abhishekbalam: done

erpnext/hr/doctype/appraisal/appraisal.js
```diff
@@ -10,45 +10,70 @@ frappe.ui.form.on('Appraisal', {
...
+			frm.set_value('total_score', 0);
+			return;
+		}
+		for (let i = 0; i<goals.length; i++) {
 			total = flt(total)+flt(goals[i].score_earned)
 		}
-		frm.set_value('total_score', total);
+		if (!isNaN(total)) {
+			frm.set_value('total_score', total);
+			frm.refresh_field('calculate_total');
```
Anurag810, line 41: ```frm.set_value``` already refresh field

Anurag810 (changes requested):

erpnext/hr/doctype/appraisal/appraisal.json
```diff
@@ -1,775 +1,254 @@
...
+  "start_date",
+  "end_date",
+  "department",
+  "section_break0",
+  "goals",
+  "total_score",
+  "section_break1",
+  "remarks",
+  "other_details",
+  "company",
```
Anurag810, line 25: Move the company field at the top.

erpnext/hr/doctype/appraisal/appraisal.json
```diff
@@ -1,775 +1,254 @@
...
+  "column_break_17",
+  "amended_from"
```
Anurag810, line 27: Move this field to top tooo.

erpnext/hr/doctype/appraisal/appraisal.json
Anurag810, line 25: The value of the company should be fetched from employee master. Keep the field read-only doesn't allow to change.
image

Anurag810 (changes requested):

abhishekbalam:
@Anurag810 
Changes made and gif updated.

dawoodjee:
Why cant this be automated on change of the score field?

PR #23506, opened 2020-10-04
fix: last purchase rate in item prices report

nextchamp-saqib:
**Issue:** 
- Below query fetches 466 results. On removing `group by` fetches 1900+ results. So rows were being grouped into one row per item resulting in loss of ordered data.
```
select * from (
	(select
		po_item.item_code,
		po.transaction_date as posting_date,
		po_item.base_rate
        from `tabPurchase Order` po, `tabPurchase Order Item` po_item
	where po.name = po_item.parent and po.docstatus = 1)
	union
	(select
		pr_item.item_code,
		pr.posting_date,
		pr_item.base_rate
	from `tabPurchase Receipt` pr, `tabPurchase Receipt Item` pr_item
        where pr.name = pr_item.parent and pr.docstatus = 1)
) result 
group by result.item_code
order by result.item_code asc, result.posting_date desc
```

- Last Purchase Rate of item were fetched from Purchase Orders and Receipts. Invoices with update stock were not considered.

Fix:
- Remove `group by` to get all the Purchase Order/Receipt/Invoice Item's rate

### DIFF
--- a/erpnext/stock/report/item_prices/item_prices.py
+++ b/erpnext/stock/report/item_prices/item_prices.py
@@ -80,35 +80,10 @@ def get_last_purchase_rate():
 
 	item_last_purchase_rate_map = {}
 
-	query = """select * from (select
-					result.item_code,
-					result.base_rate
-					from (
-						(select
-							po_item.item_code,
-							po_item.item_name,
-							po.transaction_date as posting_date,
-							po_item.base_price_list_rate,
-							po_item.discount_percentage,
-							po_item.base_rate
-						from `tabPurchase Order` po, `tabPurchase Order Item` po_item
-						where po.name = po_item.parent and po.docstatus = 1)
-						union
-						(select
-							pr_item.item_code,
-							pr_item.item_name,
-							pr.posting_date,
-							pr_item.base_price_list_rate,
-							pr_item.discount_percentage,
-							pr_item.base_rate
-						from `tabPurchase Receipt` pr, `tabPurchase Receipt Item` pr_item
-						where pr.name = pr_item.parent and pr.docstatus = 1)
-				) result
-				order by result.item_code asc, result.posting_date desc) result_wrapper
-				group by item_code"""
+	query = """select item_code, last_purchase_rate from `tabItem`"""
 
 	for d in frappe.db.sql(query, as_dict=1):
-		item_last_purchase_rate_map.setdefault(d.item_code, d.base_rate)
+		item_last_purchase_rate_map.setdefault(d.item_code, d.last_purchase_rate)
 
 	return item_last_purchase_rate_map
 
--- a/erpnext/stock/report/item_prices/item_prices.py
+++ b/erpnext/stock/report/item_prices/item_prices.py
@@ -77,39 +77,26 @@ def get_price_list():
 	return item_rate_map
 
 def get_last_purchase_rate():
-
 	item_last_purchase_rate_map = {}
 
-	query = """select * from (select
-					result.item_code,
-					result.base_net_rate
-					from (
-						(select
-							po_item.item_code,
-							po_item.base_net_rate
-						from `tabPurchase Order` po, `tabPurchase Order Item` po_item
-						where po.name = po_item.parent and po.docstatus = 1
-						order by po.transaction_date desc, po.name desc limit 1)
-						union
-						(select
-							pr_item.item_code,
-							pr_item.base_net_rate
-						from `tabPurchase Receipt` pr, `tabPurchase Receipt Item` pr_item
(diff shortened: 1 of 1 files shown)


erpnext/stock/report/item_prices/item_prices.py
```diff
@@ -82,33 +82,34 @@ def get_last_purchase_rate():
...
-							po_item.item_name,
-							po.transaction_date as posting_date,
-							po_item.base_price_list_rate,
-							po_item.discount_percentage,
-							po_item.base_rate
+							po_item.base_net_rate
 						from `tabPurchase Order` po, `tabPurchase Order Item` po_item
-						where po.name = po_item.parent and po.docstatus = 1)
+						where po.name = po_item.parent and po.docstatus = 1
+						order by po.transaction_date desc, po.name desc limit 1)
```
shreyanshj: Will this line really work?
The idea of this function is to create a `item_last_purchase_rate_map = {}`
This line would return only a single row from all the Purchase Order Items matching their Parent POs.

As my understanding of MySQL is limited.
I ran this in my environment - and it indeed churns up a single row because of limit 1. I remove that, all the item lists start coming up. Only when all Purchase Order Items are as output, will it match with their respective PR and PIs.
If that remains, the map will never be created.

erpnext/stock/report/item_prices/item_prices.py
```diff
@@ -82,33 +82,34 @@ def get_last_purchase_rate():
...
-							pr_item.item_name,
-							pr.posting_date,
-							pr_item.base_price_list_rate,
-							pr_item.discount_percentage,
-							pr_item.base_rate
+							pr_item.base_net_rate
 						from `tabPurchase Receipt` pr, `tabPurchase Receipt Item` pr_item
-						where pr.name = pr_item.parent and pr.docstatus = 1)
+						where pr.name = pr_item.parent and pr.docstatus = 1
+						order by pr.posting_date desc, pr.posting_time desc, pr.name desc limit 1)
```
shreyanshj: Same as above - this `limit 1` is going to dump only a single row from the output - then we will never be creating the *all item map* which is then consumed by calling function to dump the report.
I might be missing something really big here which is why I am wondering - did you get the report correct after this?

shreyanshj (changes requested):
I strongly feel that the change in query is not correct. But, another opinion would be better considering I have limited confidence in SQL queries. I have already tried to run this in my environment and it definitely returns a single row. Until this function is called for each item, the report won't be generated.

marination:
@nextchamp-saqib lets discuss this , I have some doubts whether PO should be involved here or not

shreyansh-ithead:
This bug is in relation to the "Item Prices" Report.
It ideally has no connection what so ever to the PO document. Ideally, it shows only 4 variations - Last Purchase Price, Valuation, Price List and BoM Rate. None of which have any connection to a PO. So, even I agree that PO is not really correct pivot.

But, my comment was not about the PO - it was about the query and use of `limit 1` - that is not correct. This function takes no argument and is expected to return a complete map of item-code vs last_purchase_rate - so that Last Purchase Rate column can be updated in the "Item Prices" Report.
But, with `limit 1`, it would return only a single row (irrespective of whether PO is considered or not).

shreyanshj:
This bug is in relation to the "Item Prices" Report.
It ideally has no connection what so ever to the PO document. Ideally, it shows only 4 variations - Last Purchase Price, Valuation, Price List and BoM Rate. None of which have any connection to a PO. So, even I agree that PO is not really correct pivot.

But, my comment was not about the PO - it was about the query and use of limit 1 - that is not correct. This function takes no argument and is expected to return a complete map of item-code vs last_purchase_rate - so that Last Purchase Rate column can be updated in the "Item Prices" Report.
But, with limit 1, it would return only a single row (irrespective of whether PO is considered or not).

marination:
@shreyanshj Hey Shreyansh, my comment was a general feedback, we'll definitely consider what you said and test this well. Feedback is always good. Thanks

shreyanshj:
@marination apologies - I somehow didn't realize that you were not the original author - I misread it as a cross comment from nextchamp-saqib.

erpnext/stock/report/item_prices/item_prices.py
nextchamp-saqib: Yep. You are correct. The set of data I tested on generated same result so I didn't catch the problem. Will modify the query to fetch one po/pr/pi transaction for every stock item.

nextchamp-saqib:
> @nextchamp-saqib lets discuss this , I have some doubts whether PO should be involved here or not

As per my understanding this report tries to show the Last purchase rate from item master which considers all buying documents (even non update stock purchase invoice). Not sure if PO should be omitted.

PR #23509, opened 2020-10-04
fix: Update Bin via Update Item on Purchase/Sales Order 

AndyOverLord:
**Issue:**
- On deletion of row Bin was not updated
- Furthermore, in PO, Material Request and requested wasn't updated either on row deletion

**Fix:**
- Explicitly update bin on row deletion in `validate_and_delete_children`
- First validate if row can be deleted, if yes proceed to delete
- Then update linked documents , like MR. Without these up to date Bin will get outdated values after recalculation. It recalculates from child table rows
- Finally update requested & ordered / reserved qty in Bin 
- Added Tests
- Re-ran patches to updates **requested**(MR), **reserved**(SO) and **ordered**(PO) qty
- Broke row deletion logic into smaller functions

**Miscellanous:**
- Made field **Reserved Qty for subcontracting** read only in **Bin**


erpnext/controllers/accounts_controller.py
```diff
@@ -1204,6 +1204,7 @@ def set_purchase_order_defaults(parent_doctype, parent_doctype_name, child_docna
 	child_item.description = item.description
 	child_item.schedule_date = trans_item.get('schedule_date') or p_doc.schedule_date
 	child_item.uom = trans_item.get("uom") or item.stock_uom
+	child_item.warehouse = p_doc.set_warehouse
```
marination: we'll have to do something similar to how the warehouse is set in `set_sales_order_defaults`, there's a high chance `set_warehouse` may not be set

marination (changes requested):

erpnext/controllers/accounts_controller.py
```diff
@@ -1235,6 +1236,12 @@ def validate_and_delete_children(parent, data):
 
 		d.cancel()
 		d.delete()
+		
+		from erpnext.stock.stock_balance import update_bin_qty, get_ordered_qty
+		frappe.errprint(f"Item Code: {d.item_code}, Warehouse: {d.warehouse}")
+		update_bin_qty(d.item_code, d.warehouse, {
+			"ordered_qty": get_ordered_qty(d.item_code, d.warehouse)
+		})
```
marination: - will have to include `reserved_qty` too as the same issue goes for Sales Order.
- Remove the `frappe.errprint`

erpnext/controllers/accounts_controller.py
AndyOverLord: Good point, have edited the code accoridng to your advice. Thank you.

erpnext/controllers/accounts_controller.py
AndyOverLord: Agree, have added the `reserved_qty`, removed "frappe.errprint".

marination:
@AndyOverLord can you add unit tests for the same ?

erpnext/controllers/accounts_controller.py
```diff
@@ -1235,6 +1236,12 @@ def validate_and_delete_children(parent, data):
 
 		d.cancel()
 		d.delete()
+		
+		from erpnext.stock.stock_balance import update_bin_qty, get_reserved_qty, get_ordered_qty
+		update_bin_qty(d.item_code, d.warehouse, {
+			"reserved_qty": get_reserved_qty(d.item_code, d.warehouse),
+			"ordered_qty": get_ordered_qty(d.item_code, d.warehouse)
+		})
```
marination: updating both will be time consuming, update it based on the doctype. reserved qty if sales order, otherwise ordered qty

AndyOverLord:
> @AndyOverLord can you add unit tests for the same ?

@marination Sorry, not sure how to add unit tests for this? Any instructions I can follow?

marination:
@AndyOverLord you can refer to `test_sales_order.py`, in that `test_update_child_removing_item` is a unit test for removing items. You can add a before and after `reserved_qty` check (assertEqual) there.
You can can add a before and after `reserved_qty` check in `test_update_child_adding_new_item` as well to increase test coverage.
The same in `test_purchase_order`.

nabinhait:
@marination @rohitwaghchaure Can you add test cases for this PR? Also need to write a patch for reposting reserved and ordered qty. This is actually critical, should be fixed asap.

marination:
> @marination @rohitwaghchaure Can you add test cases for this PR? Also need to write a patch for reposting reserved and ordered qty. This is actually critical, should be fixed asap.

@nabinhait done

marination:
Recent test added
Screenshot 2021-03-31 at 9 03 40 PM

Issue #23475, opened 2020-10-01
Supplier sourced items in Sub-contracting module


KanchanChauhan:
**Problem**

- In a workflow where a Supplier or Sub-Contractor assembles or produces the product, some of the raw material is provided by the user, and some are sourced by the Supplier himself.
- For the user, such items do not have any stock value or valuation rate.
- They should exist only in BOM and Purchase Order.
- They should also be visible in Purchase Order to Sub-Contractor so as to know that they have to source it themselves.
- Currently, there is no way to flag such items.

**Solution**

- Create a child table in BOM for Supplier Sourced Items.
- Also show these items in the Purchase Order at zero value in the separate child table of Sub-Contractor

asbasawaraj:
@KanchanChauhan The items sourced by the sub-contractor should be added in the BOM and they should be shown in PO so that sub-contractor knows he needs to sources those items on his own. This part I have clarity.

Work order doesn't come into the picture in the above situation. How does adding those details benefit in the work order help?

KanchanChauhan:
@asbasawaraj I understand, there should not be a work order for these kinds of BOMs. Just trying to cover up all the scenarios, it will just be a read-only table pulled from BOM if someone decides to have a Supplier Work order also for their reference.
Also to let you know, we have started the development and it will be ready soon.

asbasawaraj:
@KanchanChauhan Today, I, @rohitwaghchaure and @marination had a discussion on this. While we have to support more use cases, we should not also try to keep the product simple to use. It's hard to strike that balance.

We are trying to come up with alternative designs as well. 

Sure, It will be good to have PR.

creamdory:
Hi @asbasawaraj about this. As what @KanchanChauhan wrote, we suggest to have it in another child table. This table will be visible also in the PO for subcontracting. We decided to do this instead of the tag in the item (since an item is not always sourced by supplier so this is just BOM level) in order to exclude these item in the transfer for subcontracting. 

> @KanchanChauhan The items sourced by the sub-contractor should be added in the BOM and they should be shown in PO so that sub-contractor knows he needs to sources those items on his own. This part I have clarity.
> 
> Work order doesn't come into the picture in the above situation. How does adding those details benefit in the work order help?

rmehta:
Why not just attach the BOM with the Purchase Order? This way we don't need to copy the items in the PO? Instead of another child table, these can be just tagged in the BOM "Supplier bought", right?

[...]

PR #23518, opened 2020-10-06
feat: Supplier Sourced Items in BOM and Purchase Order

KanchanChauhan:
Minor and a basic feature required by Manufacturing companies.

While creating a BOM for Sub-Contracting, there might be raw material which Supplier will have to procure themselves, not always though hence we cannot put this flag in Item Master. 

Does not require any Stock Entries or Accounts-related stuff. Just the information which is very important to the user and they need to see on BOM and Purchase Order.

- Added a checkbox Has Supplier Sourced Items
- It will populate the Childtable with Item Code, Item Name, Description, Qty, UOM.
Screen Shot 2020-10-05 at 6 14 57 PM

Screen Shot 2020-10-05 at 6 15 52 PM

- In Purchase Order on validate if BOM exists, it will pull the Supplier Sourced Items for reference to the Supplier and populate in the read only table with no effect on totals.
- The qty is calculated based on item qty multiplied by the required raw material qty in BOM.
screencapture-kanchan-ch-8000-desk-2020-10-06-16_00_40

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.js
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.js
@@ -27,7 +27,6 @@ frappe.ui.form.on("Purchase Order", {
 				filters: {'company': frm.doc.company}
 			}
 		});
-
 	},
 
 	onload: function(frm) {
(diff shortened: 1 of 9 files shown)


marination:
Closing for reasons mentioned in https://github.com/frappe/erpnext/issues/23475

PR #23520, opened 2020-10-06
feat: balance serial nos in stock leder report

rohitwaghchaure:
**Issue**
Currently there is no any report which will give us the information about the serial nos available as per the selected date

**Solution**

Added new column. Balance Serial No in the stock ledger report which will give us the information about the available serial nos as on selected date

Screenshot 2020-10-06 at 12 49 47 PM


rohitwaghchaure:
@mergifyio backport develop

PR #23521, opened 2020-10-06
fix: Asset update maintenance status

kevingdc:
Fix a bug in the **Asset** DocType where their statuses are not being reverted to their original status after a repair or maintenance when being processed by the `update_maintenance_status()` function. This was fixed by adding a call to `set_status()` in the `update_maintenance_status()` function if there is no **Asset Maintenance Task** or **Asset Repair**.

### DIFF
--- a/erpnext/utilities/transaction_base.py
+++ b/erpnext/utilities/transaction_base.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import frappe
 import frappe.share
 from frappe import _
-from frappe.utils import cstr, now_datetime, cint, flt, get_time, get_datetime, get_link_to_form
+from frappe.utils import cstr, now_datetime, cint, flt, get_time, get_datetime, get_link_to_form, date_diff, nowdate
 from erpnext.controllers.status_updater import StatusUpdater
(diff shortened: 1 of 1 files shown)


erpnext/assets/doctype/asset/asset.py
```diff
@@ -40,21 +60,30 @@ def on_cancel(self):
...
+				"Purchase Invoice" if self.purchase_invoice else "Purchase Receipt"
+			)
 			reference_name = self.purchase_invoice or self.purchase_receipt
 			reference_doc = frappe.get_doc(reference_doc, reference_name)
-			if reference_doc.get('company') != self.company:
-				frappe.throw(_("Company of asset {0} and purchase document {1} doesn't matches.").format(self.name, reference_doc.get('name')))
-
+			if reference_doc.get("company") != self.company:
+				frappe.throw(
+					_(
```
rmehta: Not a good idea to introduce extra lines for brackets. I think the whole purpose of Python is to remove brackets and the cognitive load they bring.

rmehta (changes requested):
Please don't refactor existing code, it makes your contribution harder to review. Also needs relevant tests.

PR #23524, opened 2020-10-06
fix(Employee): Field level change in Exit section

michellealva:
**Before:**

image

**After:**

image

**Changes made:**
- Removed "Reason for Resignation" since it's similar to "Reason for Leaving"
- Changed fieldtype of "Reason for Leaving" to small text.
- Made "Encashment Date" dependent on "Leave Encashed?" field

### DIFF
--- a/erpnext/hr/doctype/employee/employee.json
+++ b/erpnext/hr/doctype/employee/employee.json
@@ -109,7 +109,6 @@
   "encashment_date",
   "exit_interview_details",
   "held_on",
-  "reason_for_resignation",
   "new_workplace",
   "feedback",
   "lft",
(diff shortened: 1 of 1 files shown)


Anurag810 (changes requested):
Have you removed the relieving date because it is not visible in the second screenshot?

michellealva:
@Anurag810 no, it's dependent on the status of the employee and is only shown when the status is left. In the second screenshot, the status was still Active, hence not visible.

rohitwaghchaure (approved):

rohitwaghchaure:
@michellealva we need to write a patch to shift data from Reason for Resignation to Reason for Leaving

PR #23526, opened 2020-10-07
chore: validate reference document

nextchamp-saqib:

PR #23527, opened 2020-10-07
chore: validate reference document

nextchamp-saqib:
BackPort of #23526


barredterra (approved):

barredterra (changes requested):
`from frappe import _`

PR #23530, opened 2020-10-07
fix: Chart of Accounts

barredterra:
- "Chart Of Accounts" was used only in treeview. "Chart of Accounts" (with lowercase "o") is used everywhere else.
- "Create > Company" instead of "Create > New Company". This is the standard. "New" is implied by "create".


marination (dismissed):

marination:
@barredterra A little global search for instances of "Chart Of Accounts" and fixing those up would help too. I think  a field label in Company still uses the incorrect one. There's not very many

barredterra:
@marination sorry, I trusted the translation tool which showed only one occurrence.

marination (approved):

nabinhait (approved):

Issue #23448, opened 2020-09-28
V12 HR, Leave Encashment BUGS. 


MahmoudRizk:
*  BUG 1:
https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L70

``allocation.from_date`` is calling an attribute that does not exist which always returns ``None``, where ``allocation = self.get_leave_allocation() ``. The query used to make this object does not contain ``from_date`` as shown below.

https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L81-L86
<br>
* BUG 2:
``self.leave_balance`` calculation does not make sense, because it subtracts the unused leave balance from the total leave balance allocation which is not correct. It needs to be the unused leave balance only. For example, If an employee has 21 days leave allocated and did not take any vacations, the total unused days will also be 21 and the total balance at the end will be zero!!  

https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L69-L70

PR #23532, opened 2020-10-07
fix: Wrong calculations for Leave Balance in Leave Encashment (#23448)

MahmoudRizk:
*  BUG 1:
https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L70

``allocation.from_date`` is calling an attribute that does not exist which always returns ``None``, where ``allocation = self.get_leave_allocation() ``. The query used to make this object does not contain ``from_date`` as shown below.

https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L81-L86
<br>
* BUG 2:
``self.leave_balance`` calculation does not make sense, because it subtracts the unused leave balance from the total leave balance allocation which is not correct. It needs to be the unused leave balance only. For example, If an employee has 21 days leave allocated and did not take any vacations, the total unused days will also be 21 and the total balance at the end will be zero!!  

https://github.com/frappe/erpnext/blob/24e5a6172aaba2974d3097ea07758ec79ed3a953/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L69-L70

### DIFF
--- a/erpnext/hr/doctype/leave_allocation/leave_allocation.py
+++ b/erpnext/hr/doctype/leave_allocation/leave_allocation.py
@@ -210,8 +210,8 @@ def get_unused_leaves(employee, leave_type, from_date, to_date):
 	leaves = frappe.get_all("Leave Ledger Entry", filters={
 		'employee': employee,
 		'leave_type': leave_type,
-		'from_date': ('>=', from_date),
-		'to_date': ('<=', to_date)
+		'from_date': ('<=', from_date),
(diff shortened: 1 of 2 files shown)


MahmoudRizk:
@Anurag810 
Can someone review this PR, please?

PR #23533, opened 2020-10-07
fix: use party name as title

barredterra:
**Purchase Invoice** and **Sales Invoice** show the party name as title.

This PR introduces the same to **Purchase Order** and **Sales Order**. Previously they were using the customer and supplier id (name) which can also be from a naming series. Having "Frappe Pvt Ltd" as the title is nicer than "SUPP-00001". Also it is more consistent like this.


erpnext/buying/doctype/purchase_order/purchase_order.json
```diff
@@ -1130,11 +1130,11 @@
...
   }
  ],
- "search_fields": "status, transaction_date, supplier,grand_total",
+ "search_fields": "status, transaction_date, supplier, grand_total",
  "show_name_in_global_search": 1,
  "sort_field": "modified",
  "sort_order": "DESC",
  "timeline_field": "supplier",
- "title_field": "supplier",
+ "title_field": "supplier_name",
```
marination, line 1138: from what i see `supplier_name` is only populated if the Supplier Code and Name differ. It is blank if they are the same. So keeping it `supplier` is best

barredterra: Let's change that as well. Can you point me to the code you're talking about?

marination: Hey really sorry, it just toggles supplier_name's display, the value still remains. This works 👍

barredterra: I cannot reproduce what you're saying on my instance. `supplier_name` get's set even when it is identical to `supplier`. I guess it is just hidden to not show duplicate information.

marination (approved):

PR #23534, opened 2020-10-07
fix: error log while creating company

anupamvs:
Issue- 
Error While creating an error log during company creation. (error log is not created.)
image

After Fix-
image


marination:
Please put the right screenshot or no screenshot. This screenshot needs a fix in frappe
Edit: Add a proper description so someone with no context knows what led to this screenshot

marination (approved):

PR #23549, opened 2020-10-08
fix: project value is missing from procurement-tracker

anupamvs:
**Issue-** the project value is missing from Procurement Tracker for Material Request
image

**After Fix-**
image


erpnext/buying/report/procurement_tracker/procurement_tracker.py
```diff
@@ -207,7 +207,8 @@ def get_mapped_mr_details(conditions):
 			child.qty,
 			child.item_code,
 			child.uom,
-			par.status
+			par.status,
+			child.project
```
marination: change the `par` to `parent` please
Also, correct me if im wrong, project fetching is required for PO,PR,PI too right ?

nabinhait (approved):

PR #23552, opened 2020-10-08
fix: Taxable value in GSTR 3B report

deepeshgarg007:
- Used `base_net_total` field to show taxable value instead of `net_total` so that taxable value is shown in INR and not in any other currency for multi-currency invoices

- Consideration of Registered Regular customers in Reverse Charge calculation


rohitwaghchaure:
@deepeshgarg007 put the description about issue as well fix

deepeshgarg007:
@Mergifyio backport develop

PR #23557, opened 2020-10-08
feat: Supplier Sourced Items in BOM

KanchanChauhan:
@marination @asbasawaraj Following from what @rmehta suggested, here is a different implementation. Less invasive and no child tables.

**Defining the problem again:** 
While creating a BOM for Sub-Contracting, there might be raw material which Supplier will have to procure themselves like nuts and bolts. While creating a Transfer Stock entry from PO, these can be excluded one by one but it is impossible to do so if you have more than 100 items. The same issue was raised in today's webinar as well by someone.

**Solution:**
- Added a flag in BOM Item and Exploded BOM. On selection, it will set rate and amount to 0, and even on validate it will remain 0
Screen Shot 2020-10-08 at 4 39 32 PM

- In Purchase Order, when it is a Sub-contract, these items will not show up in the Supplied item.
For example here, BOM has 2 raw materials where 1 is Supplier Sourced and it does not show up in PO.

Screen Shot 2020-10-08 at 4 43 10 PM

Screen Shot 2020-10-08 at 4 43 20 PM

- It will have no impact on Stock Entries, it will not show up.
Screen Shot 2020-10-08 at 5 24 05 PM

@marination @rohitwaghchaure Please let us know your thoughts on this.


KanchanChauhan:
Test cases and Documentation is WIP

KanchanChauhan:
Test cases passed on local on new site, travis is failing for unrelated test cases:

Screen Shot 2020-10-09 at 10 06 16 PM

Documentation PR:
https://github.com/frappe/erpnext_documentation/pull/168

rohitwaghchaure (approved):

PR #23558, opened 2020-10-08
feat: Therapy Plan Template

ruchamahabal:
**Documentation PR**: https://github.com/frappe/erpnext_documentation/pull/182

- Two ways in which therapies are billed
   - Some hospitals use pre-configured therapy plans which are prescribed as packages. No. of sessions to be conducted and the total cost is fixed. A therapy plan created using the Therapy Plan Template will not be billed for individual sessions but for the package as a whole.
   - In other cases, Therapy Plans are created only to keep a track of the number of sessions prescribed and total sessions completed. Sessions are billed individually.

- Therapy Plan Template
   - An item will be created for the therapy plan template as a whole. The linked item details section is for specifying details about the linked item. On updating `item_name`, `item_group`, `description` fields the linked item will be updated automatically.
   - In the Therapy Types section, the therapy types can be selected. Their rate is fetched from the Therapy Type document. The `amount` is calculated as a product of `no_of_sessions` and `rate`. Item Price is created and updated for the Linked Item accordingly.

   therapy-plan-template

- Therapy Plan

  - The Therapy Types are populated from the Therapy Plan Template.
     therapy-plan

- Billing

  - While fetching healthcare services for invoicing, the individual Therapy Sessions of the Therapy Plan are only available for billing if the corresponding plan is created using a Therapy Plan Template. Instead, the whole Therapy Plan is available for as an item.

     therapy-plan-template-biiling


rohitwaghchaure:
@ruchamahabal 

Some observations while testing your PR

1. In therapy session, I can use same date, time and patient for the different sessions. (should have validation for overlap) 

1. Add option to make Sales Invoice from the Therapy Plan, also add sales invoice in the dashboard 

1. Therapy Plan Template, in the item code if I've put the existing item code then system not throws any error and also not save the template.

ruchamahabal:
@rohitwaghchaure 

Validating duplication now, added Sales Invoice creation option. Docs added.

Also, if I add an existing item code:

item_code_already_exists

nabinhait (commented):
Test failing on travis

nabinhait (approved):

PR #23564, opened 2020-10-09
fix: negative stock error while submitting stock reco for batch item

rohitwaghchaure:
**Issue**

1. Create batched item and make purchase receipt with 20 quantity and valuation rate as 100 on 7th October

1. Create delivery note against the same batched item with quantity as 20 and valuation rate as 100 on 9th October
 
1. Create stock reconciliation for the same batched item and same batch on 8th October, keep the quantity same and just change the valuation rate from 100 to 120. Once you try to submit the stock reco you will get below error

Screenshot 2020-10-09 at 11 59 55 AM


rohitwaghchaure:
Screenshot 2020-10-09 at 6 34 03 PM

PR #23567, opened 2020-10-09
chore: do not allow future transactions

nextchamp-saqib:
Problem:
- If a future transaction is posted then user is unable to create transactions on until that particular future date.
- Only option is to delete the stock ledger entry associated with that future transaction after cancelling.

Screenshot 2020-10-09 at 2 31 55 PM

Fix:
- Not allow future posting until immutable ledger is stable
Screenshot 2020-10-09 at 3 34 27 PM

### DIFF
--- a/erpnext/utilities/transaction_base.py
+++ b/erpnext/utilities/transaction_base.py
@@ -5,7 +5,7 @@ from __future__ import unicode_literals
 import frappe
 import frappe.share
 from frappe import _
-from frappe.utils import cstr, now_datetime, cint, flt, get_time, get_datetime, get_link_to_form
+from frappe.utils import cstr, now_datetime, cint, flt, get_time, get_datetime, get_link_to_form, date_diff, nowdate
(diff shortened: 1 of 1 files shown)


erpnext/utilities/transaction_base.py
```diff
@@ -29,7 +29,13 @@ def validate_posting_time(self):
 			except ValueError:
 				frappe.throw(_('Invalid Posting Time'))
 
+		self.validate_future_posting()
 		self.validate_with_last_transaction_posting_time()
+	
+	def validate_future_posting(self):
```
deepeshgarg007, line 46: This will also prevent future posting of pure accounting transactions like Sales Invoice with no update stock. Future posting for pure accounting transaction should be allowed

deepeshgarg007 (changes requested):

hazemtolba:
i want to allow user to make transactions on past days to entering the backlog ,  where this screen  displayed
image

PR #23569, opened 2020-10-09
fix: Remove Production Order reference from Item Validation

marination:
- Production Order has been removed almost 2 years or so back, and Production Plan is in action as of now
- Removed check on Production Order in item's `check_if_linked_document_exists` function


marination:
For the reviewer:
Please let me know if another check instead of the one removed should be added or if im missing some case

rohitwaghchaure (approved):

PR #23583, opened 2020-10-11
fix: can't save item price after adding child table 

anupamvs:
**Issue-** After adding child table in item price, we can't save form if child table has data.
item_price


rohitwaghchaure:
@anupamvs kindly add the screenshots of the error

anupamvs:
@rohitwaghchaure updated

rohitwaghchaure (approved):

PR #23586, opened 2020-10-11
fix: sales order payment terms date is same as quotation

nextchamp-saqib:
Problem Fixed:
- Create Sales Order from Quotation
- The due dates in payment terms are as per the quotation date. 
- It should be calculated as x days after Sales Order posting date. X is Credit Days set in the Payment terms master.

### DIFF
--- a/erpnext/selling/doctype/quotation/quotation.py
+++ b/erpnext/selling/doctype/quotation/quotation.py
@@ -183,10 +183,6 @@ def _make_sales_order(source_name, target_doc=None, ignore_permissions=False):
 			"Sales Team": {
 				"doctype": "Sales Team",
 				"add_if_empty": True
-			},
-			"Payment Schedule": {
-				"doctype": "Payment Schedule",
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -183,10 +183,6 @@ def update_item(obj, target, source_parent):
 			"Sales Team": {
 				"doctype": "Sales Team",
 				"add_if_empty": True
-			},
-			"Payment Schedule": {
-				"doctype": "Payment Schedule",
-				"add_if_empty": True
```
rohitwaghchaure, line 189: Check the PR https://github.com/frappe/erpnext/pull/17390

rohitwaghchaure:
@nextchamp-saqib this fix is not proper, there might be chances that user wants the same dates in the sales order which are there in the quotation. Also please check the below PR
https://github.com/frappe/erpnext/pull/17390

PR #23593, opened 2020-10-12
fix: cannot merge pos invoice if validate selling price is checked

nextchamp-saqib:
Problem:
- While validating selling price, `base_rate` should be >= `last_purchase_rate_in_sales_uom` 
- Here there are 2 problems:
   - `last_purchase_rate` = `base_net_rate` / `conversion_factor` -- So comparison should be with `base_net_rate` while validating selling price
   - `last_purchase_rate_in_sales_uom` is calculated as `last_purchase_rate` / `conversion_factor` -- it should be muliplies instead of divide
- Same goes for `last_valuation_rate`

Fix:
- compare `base_net_rate` with `last_valuation_rate_sales_uom` and `last_purchase_rate_in_sales_uom`
- Error message:
Screenshot 2020-10-12 at 10 31 47 AM  &nbsp; Screenshot 2020-10-12 at 12 58 42 PM

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/test_pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/test_pos_invoice.py
@@ -376,14 +376,9 @@ class TestPOSInvoice(unittest.TestCase):
 			"rate": 14,
 			'included_in_print_rate': 1
 		})
-		pos_inv.submit()
-
-		merge_pos_invoices()
-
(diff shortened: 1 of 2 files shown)


nabinhait (changes requested):
I think existing validation is wrong. The system should validate **net_rate** with last purchase rate and valuation_rate.

nabinhait (approved):

PR #23595, opened 2020-10-12
feat: opening invoice creation tool with background jobs

nextchamp-saqib:
Solves timeout problem while creating large number of opening invoices.

Screenshot 2020-10-12 at 12 03 33 PM

Docs:  frappe/erpnext_documentation#181


erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
```diff
@@ -161,40 +143,83 @@ def get_item_dict():
...
-		if self.invoice_type == "Sales":
-			args["is_pos"] = 0
+		return invoice
 
-		return args
+	def make_invoices(self):
+		self.validate_company()
+		invoices = self.get_invoices()
+		if len(invoices) < 50:
+			self.start_import(invoices)
```
nextchamp-saqib:
```suggestion
			start_import(invoices)
```

nabinhait (approved):

Issue #23613, opened 2020-10-13
Bank Statement Reconciliation report does not filter by company and show even disabled items


aakvatech:
## Description of the issue
Bank Reconciliation Statement report does not filter bank accounts by company and show even disabled accounts

## Context information (for bug reports)
If you have multi company scenario and many bank accounts that are no longer active, then it becomes difficult in bank reconciliation statement report to filter the account to reconcile and it also shows disabled accounts so futher confusion is created.

**Output of `bench version`**
erpnext 12.12.0
frappe 12.10.2

## Steps to reproduce the issue

1. Create 2 companies with 5 bank accounts each
2. Disable 3 bank accounts in each

### Observed result
Open Bank Reconciliation Statement Report and you will see all 10 bank accounts listed.

### Expected result
Open Bank Reconciliation Statement Report and you should see 2 bank accounts for a selected company only.

### Stacktrace / full error message
N/A

## Additional information
FC hosted

aakvatech:
PR raised for v12

https://github.com/frappe/erpnext/pull/23614

PR #23615, opened 2020-10-13
feat: Add company and correct filter in bank reconciliation statement

aakvatech:
… report filters

If you have multi company scenario and many bank accounts that are no longer active, then it becomes difficult in bank reconciliation statement report to filter the account to reconcile and it also shows disabled accounts so futher confusion is created.

https://github.com/frappe/erpnext/issues/23613

image


rohitwaghchaure:
@aakvatech don't send PR on stable branch version-12, instead of this send PR on version-12-hotfix

aakvatech:
Well noted

PR #23622, opened 2020-10-13
fix: setting user defined precision instead of 2

AfshanKhan:
If a user had set the precision of more than 2 they would get the wrong value in the Project Billing Summary report.


AfshanKhan:

rohitwaghchaure (approved):

PR #23630, opened 2020-10-14
fix: setting user precision instead of 2

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/23622


erpnext/projects/report/billing_summary.py
```diff
@@ -6,6 +6,7 @@
 import frappe
 from frappe import _
 from frappe.utils import time_diff_in_hours, flt
+from frappe.model.meta import get_field_precision
```
rohitwaghchaure: get_field_precision imported but not used

AfshanKhan: removed

nabinhait (approved):

PR #23632, opened 2020-10-14
fix: fuel expense amount of vehicle log

nextchamp-saqib:
Problem:
- Make a vehicle log with fuel qty and fuel price.
- Create expense claim against it
- The expense amount is set as fuel price. It should have been `qty * price`


erpnext/hr/doctype/vehicle_log/test_vehicle_log.py
```diff
@@ -6,18 +6,28 @@
...
+from erpnext.hr.doctype.employee.test_employee import make_employee
+from erpnext.hr.doctype.vehicle_log.vehicle_log import make_expense_claim
 
 class TestVehicleLog(unittest.TestCase):
-	def test_make_vehicle_log_and_syncing_of_odometer_value(self):
-		employee_id = frappe.db.sql("""select name from `tabEmployee` where status='Active' order by modified desc limit 1""")
-		employee_id = employee_id[0][0] if employee_id else None
+	def setUp(self):
+		self.employee_id = frappe.db.sql("""select name from `tabEmployee` where name='testdriver@example.com'""")
+		self.employee_id = employee_id[0][0] if self.employee_id else None
```
nabinhait:
```suggestion
		employee_id = frappe.db.sql("""select name from `tabEmployee` where name='testdriver@example.com'""")
		self.employee_id = employee_id[0][0] if employee_id else None
```

erpnext/hr/doctype/vehicle_log/test_vehicle_log.py
```diff
@@ -40,6 +50,27 @@ def test_make_vehicle_log_and_syncing_of_odometer_value(self):
...
+			"price": frappe.utils.flt(500)
+		})
+		vehicle_log.save()
+		vehicle_log.submit()
+
+		expense_claim = make_expense_claim(vehicle_log.name)
+		fuel_expense = expense_claim.expenses[0].amount
+		self.assertEqual(fuel_expense, 50*500)
+
+		frappe.db.sql("delete from `tabExpense Claim` where name=%s", expense_claim.name)
```
nabinhait: use `delete_doc` to delete child table data as well

nabinhait (changes requested):

nabinhait (approved):

PR #23639, opened 2020-10-14
fix: Check for backdated validation only for transaction company

deepeshgarg007:
Stock Queues for Different warehouses should be independent. Currently for eg: If post a Stock Transaction for Item A on say
'14-10-2020' system won't allow you to post a Stock Transaction  for the same item on '13-10-2020' even for a different warehouse.

This should not be the case and entry on '13-10-2020' should be allowed for a different warehouse if any other stock entry for Item A in other company doesn't exists after '13-10-2020'

Moved backdated entry validation from transaction base to stock ledger entry which allows to validate more complex scenarios where both source and target warehouses in like Stock Transfer Entry has to be validated


nabinhait (approved):

nabinhait (commented):
- Check linter/translation error
- check travis

PR #23640, opened 2020-10-14
fix(Address): Set default user Company in links table when "Is Your Company Address" is set

abhishekbalam:
The "Is Your Company Address" checkbox does not do anything on selection.

In this PR, when the checkbox is selected it does 2 things:
- Gets user's default company and adds it to the links list
- Adds a filter to the "link_doctype" field in child table to only allow "Company" doctype to be set
And on disabling the checkbox:
- Removes the filter in "link_doctype" in child table
- Clears child table so if there are any existing empty rows, the filter won't remain.

Gif:
gif

Also removed the field from frappe and moved to ERPNext as Address doctype customisation. 
PR Link: https://github.com/frappe/frappe/pull/11712

There was already a customised field called tax_category in accounts module, so I exported there itself.

Ideally it should have a neutral location like maybe "Utilities" module ?


marination:
@abhishekbalam how will the supplier /customer/ whatever else be linked since this whole table is dedicated only to company? 
Earlier it was a generic document dynamic linking table, and it was actively used to link Suppliers/Customers

Maybe the whole company address and dealing with company against an address can be moved to another section? Unsure, but this is a breaking change from what i see. Please Correct me If Im wrong.

abhishekbalam:
@marination 
All that functionality is still available, and this PR restricts the doctype linking only when the "Is Your Company Address" checkbox is selected. 

This PR has customisations to check if the checkbox is selected and that:
1. Company entry is available in links table (serverside validation)
2. No other doctypes should be added to this address other than Company using a filter in the child table field. We need this in case of multiple companies at one address. Also adding supplier/customer should not be allowed with the option selected as it would defeat the purpose of having a flag to mark an address as "company address".

If linking is needed with other doctypes along with Company, you can still do that if the checkbox isn't selected.

abhishekbalam:
@netchampfaris 

I have moved get_shipping_address() method from address.py in Frappe as it was using the flag and the method was used only in party.js in ERPNext.
It would have given error otherwise and it's not called anywhere in Frappe.

Similarly, for the validation functions in address.py in the doctype which use the flag, I have overridden them here but removed the flag in the conditions in Frappe, hope that is fine.

erpnext/public/js/address.js
```diff
@@ -0,0 +1,26 @@
...
+// For license information, please see license.txt
+
+frappe.ui.form.on("Address", {
+	is_your_company_address: function(frm) {
+		if(frm.doc.is_your_company_address){
+			frm.add_child('links', {
+				link_doctype: 'Company',
+				link_name: frappe.defaults.get_user_default('Company')
+			});
+			frm.fields_dict.links.grid.get_field('link_doctype').get_query = function() {
```
netchampfaris: A better way to avoid this duplication is to write this method in `onload` event so it will run only once, but it will get the filters dynamically based on form values.

```js
onload(frm) {
    frm.set_query('link_doctype', 'links', () => {
		return {
			filters: frm.doc.is_your_company_address ? {
				name: 'Company'
			} : null
		}
    })
}

```

abhishekbalam: I tried using the onload event but there are [filters being set in address.js](https://github.com/frappe/frappe/blob/develop/frappe/contacts/doctype/address/address.js#L17) in frappe on refresh event which are replacing the one we set with onload, unfortunately.

On taking a closer look I realised that these filters are to list only doctypes in the link_doctype field which have an address field which is actually important. I don't think we can override the refresh event as it has a lot of other code too.

image

Instead I have replaced the part in address.js in ERPNext where I was setting the field filter to null and triggered a form refresh so that the default filters are reset. Also updated the old query code as you showed.

I have triggered the refresh event using frm.trigger() instead of calling frm.refresh() because the page scrolls up and it's slightly disorienting.

Are these changes fine? If it is, I will update the gif.

netchampfaris (approved):

abhishekbalam:
@netchampfaris 
All tests passed except Travis, which seems to be broken.

sagarvora:
I've restarted Travis.

PR #23654, opened 2020-10-15
Use integer to store GST State Number

sunnyakaxd:

> Screenshots/GIFs


sagarvora:
I think you'll also need to ["reactivate" the patch](https://github.com/frappe/erpnext/blob/develop/erpnext/patches.txt#L572) to sync this field.

sagarvora:
Please follow [contribution guidelines](https://github.com/frappe/erpnext/wiki/Contribution-Guidelines).

sunnyakaxd:
Seems like a pretty small fix. See if your team can add this. Thanks a lot.

PR #23658, opened 2020-10-15
fix: overproduction, not allowed to transfer extra materials

rohitwaghchaure:
backport https://github.com/frappe/erpnext/pull/23647


sagarvora:
Shouldn't this go to `version-12-hotfix` instead?

PR #23659, opened 2020-10-15
fix: duplicate validation for Course Enrollment

ruchamahabal:
Duplication Validation method in place but was not called on validate:

duplicate


sagarvora (approved):

sagarvora (changes requested):
Please fix [broken test](https://travis-ci.com/github/frappe/erpnext/jobs/400324804#L1070).

nabinhait (approved):

PR #23663, opened 2020-10-16
fix: removed extra space from label "Rate"

AfshanKhan:
Due to this extra space in the label "Rate", when someone used to export the "Supplier Quotation" template from our Data Import the template used to have this extra space in the column name.
Screenshot 2020-10-16 at 11 32 27 AM

Which used to give an error when they used this template to import data.
Screenshot 2020-10-16 at 11 35 44 AM

### DIFF
--- a/erpnext/buying/doctype/supplier_quotation_item/supplier_quotation_item.json
+++ b/erpnext/buying/doctype/supplier_quotation_item/supplier_quotation_item.json
@@ -12,8 +12,6 @@
   "item_name",
   "column_break_3",
   "lead_time_days",
-  "expected_delivery_date",
-  "is_free_item",
   "section_break_5",
   "description",
   "item_group",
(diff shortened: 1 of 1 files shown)


AfshanKhan:

nabinhait (commented):
Why so many changes in json? Seems like a lot of field rearrangements. Are those intentional?

AfshanKhan:
> Why so many changes in json? Seems like a lot of field rearrangements. Are those intentional?

@nabinhait I just removed the extra space and saved normally. These changes occurred by themselves. I could just make a change in the JSON file and modify the timestamp if that works?

nabinhait:
> > Why so many changes in json? Seems like a lot of field rearrangements. Are those intentional?
> 
> @nabinhait I just removed the extra space and saved normally. These changes occurred by themselves. I could just make a change in the JSON file and modify the timestamp if that works?

It happens when your local is not migrated and synced with file system. Always migrate before saving a doctype.

AfshanKhan:
> > > Why so many changes in json? Seems like a lot of field rearrangements. Are those intentional?
> > 
> > @nabinhait I just removed the extra space and saved normally. These changes occurred by themselves. I could just make a change in the JSON file and modify the timestamp if that works?
> 
> It happens when your local is not migrated and synced with file system. Always migrate before saving a doctype.

@nabinhait got it. Will make a new pr with updated changes.

PR #23664, opened 2020-10-16
fix: None type error if the Pricing Rule applicable_for is None

adityaduggal:
… applicable_for

There are many cases when you want a company wide pricing rule without the applicable_for field and hence for such rules it was giving a none type error.

> Screenshots/GIFs


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -49,9 +49,10 @@ def validate_mandatory(self):
 			if self.apply_on == apply_on and len(self.get(field) or []) < 1:
 				throw(_("{0} is not added in the table").format(apply_on), frappe.MandatoryError)
 
-		tocheck = frappe.scrub(self.get("applicable_for", ""))
-		if tocheck and not self.get(tocheck):
-			throw(_("{0} is required").format(self.meta.get_label(tocheck)), frappe.MandatoryError)
+		if self.get("applicable_for", "") != None:
```
nextchamp-saqib:
```suggestion
		if self.get("applicable_for", "") is not None:
```

nextchamp-saqib (dismissed):

PR #23666, opened 2020-10-16
fix: Payment Reconciliation client side validations

marination:
**Issue:**
- If **Company** was left blank and user tried to select a **Party** :
  ```
      TypeError: get_party_account() missing 1 required positional argument: 'company'
   ```

**Fix:**
- Added a common method `check_mandatory` to check for Company and/or Party Type which is critical to most fields
- Added the check before selecting a **party** too
- Since `get_party_account` handles None company, make sure it can accept the same as an argument

Screenshot 2020-10-16 at 12 20 00 PM

Screenshot 2020-10-16 at 12 19 28 PM


marination:
travis passed but still shows pending , you can go to the job log and see

PR #23667, opened 2020-10-16
feat: show only available items in point of sale

nextchamp-saqib:
- Added a checkbox in POS Profile to hide unavailable items from offline & online POS item list


erpnext/accounts/doctype/sales_invoice/pos.py
```diff
@@ -175,6 +175,13 @@ def get_items_list(pos_profile, company):
 		if args_list:
 			cond = "and i.item_group in (%s)" % (', '.join(['%s'] * len(args_list)))
 
+	bin_join = bin_cond = ""
+	if pos_profile.get('hide_unavailable_items'):
+		bin_join = ",`tabBin` b"
+		bin_cond = "and i.item_code = b.item_code and ifnull(b.actual_qty, 0) > 0 "
```
sagarvora, line 181: One could argue that _available_ items are better represented by `projected_qty`.

nabinhait: In case of POS, projected_qty does not make that much sense.

nabinhait (approved):

michaeleino:
this is a great addition, but now do the same behaviour as "Display Items In Stock" .. it hides the service items :(

PR #23675, opened 2020-10-16
 feat: balance serial nos in stock ledger report

rohitwaghchaure:
backport https://github.com/frappe/erpnext/pull/23520


erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -55,6 +60,26 @@ def execute(filters=None):
...
+def update_available_serial_nos(available_serial_nos, sle):
+	serial_nos = get_serial_nos(sle.serial_no)
+	key = (sle.item_code, sle.warehouse)
+	if key not in available_serial_nos:
+		available_serial_nos.setdefault(key, [])
+
+	existing_serial_no = available_serial_nos[key]
+	for sn in serial_nos:
+		if sle.actual_qty > 0:
+			if sn in existing_serial_no:
```
nabinhait, line 72: If exists why should I remove it?

erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -55,6 +60,26 @@ def execute(filters=None):
...
+				existing_serial_no.remove(sn)
+			else:
+				existing_serial_no.append(sn)
+		else:
+			if sn in existing_serial_no:
+				existing_serial_no.remove(sn)
+			else:
+				existing_serial_no.append(sn)
```
nabinhait, line 80: Why I will append it in case of negative qty? Is this a canceled state? In that case, specifically, add that condition.

nabinhait (changes requested):
This functionality will be used rarely, can we do it optional using a filter?

erpnext/stock/report/stock_ledger/stock_ledger.py
rohitwaghchaure: I have seen some cases where delivery was made first and then receive of the serial nos, to handle that case I wrote this code.

erpnext/stock/report/stock_ledger/stock_ledger.py
rohitwaghchaure: I have seen some cases where delivery was made first and then receive of the serial nos, to handle that case I wrote this code.

PR #23678, opened 2020-10-17
fix: Change wrong format_date to formatdate

josephalbaph:
Installing erpnext...
Updating DocTypes for erpnext       : [===                                     ]An error occurred while installing erpnext:
Module import failed for Process Statement Of Accounts (erpnext.accounts.doctype.process_statement_of_accounts.process_statement_of_accounts Error: cannot import name 'format_date' from 'frappe.utils' (/home/erpfrappe/frappe-bench/apps/frappe/frappe/utils/__init__.py))

This is caused by wrong format_date name. The correct name is formatdate. (Author is probably confused with format_datetime)

> Screenshots/GIFs

### DIFF
--- a/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.py
+++ b/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.py
@@ -11,7 +11,7 @@ from frappe.core.doctype.communication.email import make
 
 from frappe.utils.print_format import report_to_pdf
 from frappe.utils.pdf import get_pdf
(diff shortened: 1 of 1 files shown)


sagarvora:
Hi, please look at following lines:
https://github.com/frappe/frappe/blob/develop/frappe/utils/data.py#L269-L296

As you can see, the function was renamed to `format_date` and `formatdate` still works for compatibility. So ideally, both should be working.

sagarvora:
I think your `frappe` app might be pointing to an older branch when you were running `migrate`. Closing PR for now, feel free to re-open.

PR #23687, opened 2020-10-19
feat: allow setting currency in the CoA importer

Alchez:
For multi-currency setups, accounts may need to be setup with different currencies. This PR allows optionally adding a currency in the Chart of Accounts Importer.

<hr>

**Changes:**

- Allow setting a new "Currency" field in the CoA Importer
- [BREAKING] Require setting the "Is Group" flag for group accounts


erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
```diff
@@ -75,15 +81,6 @@ def add_suffix_if_duplicate(account_name, account_number, accounts):
 
 	return account_name, account_name_in_db
 
-def identify_is_group(child):
```
nextchamp-saqib, line 78: We could keep this and still be able to set currency right? Any particular reason to remove auto-setting of `is_group`?

Since `create_charts()` is used in tally migrator where we just pass the nested accounts dictionary without setting `is_group` and the above function sets it correctly.

Alchez: The main reason for removing group is that it tends to assume the group of an account based on the number of passed keys, which seems both inaccurate and an inefficient way of assigning groups. Instead, I'm suggesting a manual but sure way of setting it.

For Tally, however, we could look into restoring this function only, and then try and phase that into a more accurate flow as well.

What do you think?

nextchamp-saqib:
@Alchez

Please resolve the merge conflicts.

Also, can you make it so that if the template doesn't have `is_group` column defined then it determines the value as before?
This is to ensure that existing behavior for other than standard fixtures doesn't get disturbed.

Alchez:
> @Alchez
> 
> Please resolve the merge conflicts.
> 
> Also, can you make it so that if the template doesn't have `is_group` column defined then it determines the value as before?
> This is to ensure that existing behavior for other than standard fixtures doesn't get disturbed. 

I can put it back, but that logic was flawed and was setting a couple of accounts to is_group, when it shouldn't be. I can't pinpoint examples right now, but I remember observing them when I was working on this change.

We can instead come up with proper and concrete rules for auto-setting group accounts, and apply those.

nextchamp-saqib:
@Alchez  We can get the currency setting part merged for now. Or if you can find the example which had an incorrect `is_group` set, then we can look into fixing it.

erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
```diff
@@ -75,15 +81,6 @@ def add_suffix_if_duplicate(account_name, account_number, accounts):
...
-	if child.get("is_group"):
-		is_group = child.get("is_group")
-	elif len(set(child.keys()) - set(["account_type", "root_type", "is_group", "tax_rate", "account_number"])):
-		is_group = 1
```
Alchez, line 82: @nextchamp-saqib, these are the lines I was talking about that misbehave in setting the `is_group` flag. With the new optional currency column addition, this may or may not match, and it will erratically set the group flag.

PR #23695, opened 2020-10-21
fix: incorrect backflush qty in manufacture entry

rohitwaghchaure:
1. Create work order for 100 quantity and set overproduction percentage as 30

1. Create "Material Transfer for Manufacture" entry against the work order for the 60 Quantity and then finish for 60 quantity.

1. Create "Material Transfer for Manufacture" entry against the same work order for the 60 Quantity.

1. Create finish good entry only for the 50 quantity and check the raw material quantity (it's not showing correct value) 

Screenshot 2020-10-21 at 11 07 42 AM

**After Fix**

Screenshot 2020-10-21 at 11 01 56 AM

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -1321,8 +1321,10 @@ class StockEntry(StockController):
 				for sr in get_serial_nos(item.serial_no):
 					sales_order = frappe.db.get_value("Serial No", sr, "sales_order")
 					if sales_order:
-						frappe.throw(_("Item {0} (Serial No: {1}) cannot be consumed as is reserverd\
-						 to fullfill Sales Order {2}.").format(item.item_code, sr, sales_order))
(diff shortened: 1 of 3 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1320,8 +1323,10 @@ def validate_reserved_serial_no_consumption(self):
 				for sr in get_serial_nos(item.serial_no):
 					sales_order = frappe.db.get_value("Serial No", sr, "sales_order")
 					if sales_order:
-						frappe.throw(_("Item {0} (Serial No: {1}) cannot be consumed as is reserverd\
-						 to fullfill Sales Order {2}.").format(item.item_code, sr, sales_order))
+						msg = (("(Serial No: {0}) cannot be consumed as it's reserverd to fullfill Sales Order {1}.")
```
marination:
```suggestion
						msg = (_("(Serial No: {0}) cannot be consumed as it's reserverd to fullfill Sales Order {1}.")
```
```suggestion
						msg = (_("(Serial No: {0}) cannot be consumed as it's reserverd to fullfill Sales Order {1}.")
```

PR #23697, opened 2020-10-21
feat: Unlink Sales Order from Project and minor Project UX

marination:
## **Issue:**
- Create a Sales Order, lets say SAL-ORD-001
- Create a Project from this Sales Order, the Project gets the name SAL-ORD-001. The Project gets linked in the Sales Order (read only field), and the Sales Order gets linked in the Project
- Great we already have a circular dependency
- Try deleting the Project, because you don't want it anymore:
  Screenshot 2020-10-21 at 4 19 07 PM
- There is also no way to unlink the Sales order from the project, removing Sales Order from the project doesn't help. There is a link in the Sales Order too

## **Fix:**
- Added an **Unlink Sales Order** tool, that will unlink the Sales Order that this Project was created for and from
   Screenshot 2020-10-21 at 4 22 30 PM
- It will remove the project link from the Sales Order and vice versa.
- It will also recalculate the Sales Costs in the Project
  project-so-link
- The Estimated Cost is left untouched as it could have been manually set too and it is just an estimate
- If there's nothing to unlink, feedback is given:
 Screenshot 2020-10-21 at 4 32 23 PM-
- Users can unlink and keep the project if used against a Sales Invoice or any other document

## **Minor Enhancement:**
**Before:**
Screenshot 2020-10-21 at 4 38 32 PM

**After:**
Screenshot 2020-10-21 at 4 39 27 PM

**Duplicate Project with Tasks** and **Unlink Sales Order** is Shifted under Tools
  Screenshot 2020-10-21 at 4 36 00 PM
**Kanban Board** and **Gantt Chart** are shifted under **Views**
  Screenshot 2020-10-21 at 4 36 07 PM
**Set Status** is the primary button group

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -493,3 +493,18 @@ def set_project_status(project, status):
 
 	project.status = status
 	project.save()
+
+@frappe.whitelist()
+def unlink_sales_order(project_name):
+	# If project belongs to / is created from a specfic Sales Order
+	# unlink SO
+	project = frappe.get_doc("Project", project_name)
+
+	if frappe.db.exists("Sales Order", project_name) and frappe.db.get_value("Sales Order", project_name, "project"):
+		frappe.db.set_value("Sales Order", project_name, "project", '')
+
+		project.sales_order, project.customer = '', ''
+		project.save()
+		return project_name
+
+	return
--- a/erpnext/projects/doctype/project/project.js
+++ b/erpnext/projects/doctype/project/project.js
@@ -64,10 +64,6 @@ frappe.ui.form.on("Project", {
 
 	set_buttons: function(frm) {
 		if (!frm.is_new()) {
-			frm.add_custom_button(__('Duplicate Project with Tasks'), () => {
-				frm.events.create_duplicate(frm);
-			});
-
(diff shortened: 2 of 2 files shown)


marination:
need to rethink this whole flow, just unlinking is not enough. Sales orders to Project linking needs some redesign
Problems:
- Cant link a submitted SO to a pre-existing project
- This PR only solves unlinking if a Project is linked to only one SO or created from SO. It doesn't handle the case where the Project is renamed, or if multiple SOs are linked a Project.

erpnext/projects/doctype/project/project.py
```diff
@@ -493,3 +493,18 @@ def set_project_status(project, status):
...
 	project.status = status
 	project.save()
+
+@frappe.whitelist()
+def unlink_sales_order(project_name):
+	# If project belongs to / is created from a specfic Sales Order
+	# unlink SO
+	project = frappe.get_doc("Project", project_name)
+
+	if frappe.db.exists("Sales Order", project_name) and frappe.db.get_value("Sales Order", project_name, "project"):
```
marination, line 503: Project can be renamed too

marination:
Closing this as it doesnt help improve the flow

PR #23708, opened 2020-10-22
fix: expected amount in pos closing entry detail

### LINKED ISSUE #23125: Change amount included in "Pos Closing Entry" on hand Cash balance - V13 beta 4
Version 13 beta 4

At end of day when the cashier is closing out for the day, the amount of change giving to customers is still added to the expected Cash closing figure

Let me give an example

If a cashier makes sales of $400 and is giving $500 cash. She will give out $100 change to the customer. 

At closing the cash on hand should be $400.

But the Pos closing is showing $500.

We need to add a column for total change and deduct this from the on hand expected cash.

For now the end of day figure is wrong

mcgun182:
expected amount in POS closing entry was equal to total cash received, not total amount in invoices as it should be
closes #23125
image

### DIFF
--- a/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
+++ b/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
@@ -128,13 +128,13 @@ function refresh_payments(d, frm, remove) {
 	d.payments.forEach(p => {
 		const payment = frm.doc.payment_reconciliation.find(pay => pay.mode_of_payment === p.mode_of_payment);
 		if (payment) {
-			if (!remove) payment.expected_amount += flt(d.grand_total);
+			if (!remove) payment.expected_amount += flt(p.amount-d.change_amount);
 			else payment.expected_amount -= flt(p.amount);
 		} else {
 			frm.add_child("payment_reconciliation", {
 				mode_of_payment: p.mode_of_payment,
 				opening_amount: 0,
-				expected_amount: p.amount
+				expected_amount: p.amount-d.change_amount
 			})
 		}
 	})
--- a/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
+++ b/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
@@ -128,7 +128,7 @@ function refresh_payments(d, frm, remove) {
 	d.payments.forEach(p => {
 		const payment = frm.doc.payment_reconciliation.find(pay => pay.mode_of_payment === p.mode_of_payment);
 		if (payment) {
-			if (!remove) payment.expected_amount += flt(p.amount);
+			if (!remove) payment.expected_amount += flt(d.grand_total);
 			else payment.expected_amount -= flt(p.amount);
 		} else {
 			frm.add_child("payment_reconciliation", {
--- a/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.py
+++ b/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.py
@@ -113,12 +113,12 @@ def make_closing_entry_from_opening(opening_entry):
 		for p in d.payments:
 			existing_pay = [pay for pay in payments if pay.mode_of_payment == p.mode_of_payment]
 			if existing_pay:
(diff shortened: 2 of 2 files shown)


erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
```diff
@@ -128,7 +128,7 @@ function refresh_payments(d, frm, remove) {
 	d.payments.forEach(p => {
 		const payment = frm.doc.payment_reconciliation.find(pay => pay.mode_of_payment === p.mode_of_payment);
 		if (payment) {
-			if (!remove) payment.expected_amount += flt(p.amount);
+			if (!remove) payment.expected_amount += flt(d.grand_total);
```
nextchamp-saqib: Hey @mcgun182. This won't work for invoices paid by multiple modes. For example, if an invoice of grand total 1500 is paid via 1000 Cash and 500 Credit Card, then expected amount of Cash & Credit Card both will be set as 1500.

nextchamp-saqib (changes requested):

erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
mcgun182: I've made an adjustment, will that be a better solution?

erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
```diff
@@ -128,13 +128,13 @@ function refresh_payments(d, frm, remove) {
 	d.payments.forEach(p => {
 		const payment = frm.doc.payment_reconciliation.find(pay => pay.mode_of_payment === p.mode_of_payment);
 		if (payment) {
-			if (!remove) payment.expected_amount += flt(p.amount);
+			if (!remove) payment.expected_amount += flt(p.amount-d.change_amount);
```
nextchamp-saqib, line 131: This won't work either. Since change amount will be subtracted from all mode of payments. 
If change is given via cash then only cash mode of payment should be subtracted by change amount. 

The problem is valid. However to solve it correctly we might have to check which mode of payment has same account as change amount's account. Then subtract the change amount from that one matching mode of payment.

nextchamp-saqib (changes requested):

mcgun182:
Well, I thank that's a rare case. Usually multiple mode of payments are made when one of them is limited / not sufficient. But I'm aware that my code is not a correct solution, as I'm not good at coding but I need the PoS system to work at my store. Could you please make a necessary changes at next update? Thankyou very much for your hard work in refactoring the PoS.

Sent from Yahoo Mail on Android 
 
  On Sat, Oct 24, 2020 at 12:44, Saqib<notifications@github.com> wrote:   


@nextchamp-saqib requested changes on this pull request.

In erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js:
> @@ -128,13 +128,13 @@ function refresh_payments(d, frm, remove) {
 	d.payments.forEach(p => {
 		const payment = frm.doc.payment_reconciliation.find(pay => pay.mode_of_payment === p.mode_of_payment);
 		if (payment) {
-			if (!remove) payment.expected_amount += flt(p.amount);
+			if (!remove) payment.expected_amount += flt(p.amount-d.change_amount);

This won't work either. Since change amount will be subtracted from all mode of payments.
If change is given via cash then only cash mode of payment should be subtracted by change amount.

The problem is valid. However to solve it correctly we might have to check which mode of payment has same account as change amount's account. Then subtract the change amount from that one matching mode of payment.

erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
mcgun182: Well, I thank that's a rare case. Usually multiple mode of payments are made when one of them is limited / not sufficient. But I'm aware that my code is not a correct solution, as I'm not good at coding but I need the PoS system to work at my store. Could you please make a necessary changes at next update? Thankyou very much for your hard work in refactoring the PoS.

nextchamp-saqib:
Will be fixed with https://github.com/frappe/erpnext/pull/24227

PR #23714, opened 2020-10-23
fix: Education Module fixes

ruchamahabal:
- Added filters for Course (according to program) and Academy Term (according to the academic year)

  academic-term

  course-filter
  
  Similar filters added in Assessment Plan and Assessment Result.

- Validation for attendance marked for future dates in Student Attendance and Student Attendance Tool

   attendance-future-dates

   attendance-tool

- Validation for Attendance marked outside of Academic Year dates

   attendance_outside_academic_year


fderyckel:
Hey @ruchamahabal, 
I think the rows in instructor log should be fetched automatically.  
Once an instructor has been scheduled for a course (or has been assigned to a student group), then it will be added to the instructor log type.  
This is the same model as on the guardian doctype where children are automatically added to the guardian.

Issue #23715, opened 2020-10-24
Forced to set two cost centers in Expense Claim


wojosc:
Hello!

when submitting Expense Claims you are forced to define a cost. When looking into Expense Claim you will find that cost_center ist filled.
![image](https://user-images.githubusercontent.com/22279621/97077968-830a0500-15e8-11eb-9f49-e9c7d1446dca.png)

In Table Expense Claim Detail you will find a further field cost_center.
![image](https://user-images.githubusercontent.com/22279621/97077986-b77dc100-15e8-11eb-81d3-a0325106f521.png)
This is not auto populated and leads to confusion on which is to be filled.

Cost centers should be on the Expense Claim Detail as on a single Expense Claim could be several Claims like food, travel and hospitality.
The field cost_center on DocType Expense Claim should probably be removed?!

Further a Cost Center should not be a mandatory field in general. Accounting regulations do not force you to use cost centers.

Thank you.

nabinhait:
We are refactoring the expense claim form, will take care the issue in that PR, closing it for now

PR #23716, opened 2020-10-24
Update expense_claim.json

wojosc:
https://github.com/frappe/erpnext/issues/23715


nabinhait:
We are refactoring the expense claim form, will take care the issue in that PR, closing it for now

PR #23721, opened 2020-10-25
fix(Bank Reconciliation): Clear Bank Transactions against Journal Entry Account

### LINKED ISSUE #23660: Can't reconcile both sides of a bank transfer
## Description of the issue
I create a Bank Entry journal entry to record a transfer from one Bank Account (owned by the company) to another Bank Account (also owned by the company). Say it is from "Savings" to "Checking". Then I get the "Checking" statement (say) and upload it into ERPNext, and do a Bank Reconciliation. When I get to the Bank Transaction for receiving the transfer, I can click on it and the correct Bank Entry transaction comes up and I click Reconcile and voila, the Bank Transaction goes to status Reconciled and all is well and when I am done, the Checking account reconciles perfectly (let's suppose). But now when I get the "Savings" statement and upload it into ERPNext and do the Bank Reconciliation, when I come to the Bank Transaction for sending the transfer, I click on it and there is nothing for it to match. So I cannot complete the Reconciliation. So the Bank Transaction just sits there, confusing the issue for future Bank Reconciliations for the "Savings" account (and potentially offering an erroneous reconciliation if I do another transfer for the same amount and this time the Savings statement comes in first).

ERPNext needs to allow both sides, the Savings account and the Checking account, to reconcile.

**Output of `bench version`**
```
erpnext 13.0.0-dev
frappe 13.0.0-dev
```

## Additional information

OS version / distribution, `ERPNext` install method, etc:

Linux Debian 10 (buster); installed by an on-line guide specifically for Debian

## Analysis

Clearly the underlying issue is as follows: ERPNext marks whether an entry has been reconciled by setting a Clearance Date field on the overall entry. But of course for Journal Entry documents in particular, Clearance is actually a property of a single Journal Entry Account record _within_ a Journal Entry rather than a property of the Journal Entry as a whole. Specifically, it's clear that in the case of a transfer from one institution to another, the Clearance Date of the outgoing funds from one does not necessarily have to be the same as the Clearance Date of the incoming funds in the other, depending on the mode of transfer.

So moving the Clearance Date field from being a field in the Journal Entry to being a field in the Journal Entry Account records would reflect reality better, and would clearly solve the problem: when the Checking account is reconciled, it would only reconcile with one of the two Journal Entry Account lines in the Bank Entry. And so when the Savings statement comes in, the transaction there can reconcile with the other line in that same Bank Entry.

I am willing to make an effort to relocate Clearance Date field along these lines, but it would be a fair amount of work and would not want to devote that time without a sense that this is an approach the developers of ERPNext would support, so that the resulting changes, which will be potentially larger than a typical Pull Request from an outside contributor, would stand a reasonable chance of being accepted. Please can you provide some clarity on the views of the ERPNext maintainers regarding the approach of changing the reconciliation system so that Bank Transactions are reconciled against individual lines in Journal Entry documents rather than against an entire Journal Entry wholesale?

Thank you very much for your thoughts and consideration on this. It's currently very problematic that you can't smoothly reconcile both sides of a bank transfer, which is why I filed this as a bug (rather than say as a feature request).

gwhitney:
Prior to this change, a Bank Transaction would reconcile against an entire
  Journal Entry. But a single Journal Entry (for example a transfer) could
  have individual Journal Entry Accounts lines for multiple different bank
  accounts. Thus, if the Bank Transactions from the statement of one Bank
  Account had already been reconciled, the corresponding Journal Entry would
  be marked as entirely cleared. Then when one later attempts to reconcile the
  Bank Transactions from the other Bank Account on the other side of the
  transfer, there is nothing "left" for the other side of the transfer to
  reconcile with.

  This commit improves the situation by moving the clearance_date field from
  Journal Entry to Journal Entry Account, and reconciling Bank Transactions
  against individual Journal Entry Account lines. This is a more realistic
  representation of what is going on, allows for different clearance_date
  values for the two different sides of non-same-day transfers, and alleviates
  the problem of only being able to reconcile one side of a transfer.

  Along the way, it also fixes several other issues: with a small parallel
  change to frappe.ui.Page in a separate PR in that project, the accumulation
  of "extra" field boxes on the top bar of a Bank Reconciliation if you click
  on "Reconcile This Account" multiple times is eliminated; also with that
  same change to frappe.ui.Page, this commit immediately filters the
  Bank Transactions in a Bank Reconciliation to just those matching the
  Bank Account of the Reconciliation; makes the payment amount information
  in the Reconciliation Tool more compact so that it can be displayed even
  on narrow screens; properly objects if the amount of the payment is larger
  than the Bank Transaction being reconciled; when searching for a payment,
  produces all associated Payment Entries for any specified Purchase Invoice
  or Expense Claim; correctly unclears payment entries when a payment row of a
  Bank Transaction is deleted; and allows for easy traversal from a reconciled
  Bank Transaction to the Journal Entry documents it is reconciled against,
  and vice versa.

Resolves #23660

Note that in my analysis of that issue, I inquired whether this (relatively minor) architectural change, of reconciling bank transactions with Journal Entry Accounts rather than with entire Journal Entry documents, would meet with the approval of ERPNext developers. Having no response and needing to proceed with our institution's bookkeeping, I thought I would go ahead and make the changes and offer them as a pull request. I think this PR offers a number of advantages over the previous state of Bank Reconciliation, and I hope that the developers find it to be helpful. Thank you for all of the effort you have put into ERPNext.

### DIFF
--- a/erpnext/accounts/doctype/bank_clearance/bank_clearance.py
+++ b/erpnext/accounts/doctype/bank_clearance/bank_clearance.py
@@ -23,19 +23,18 @@ class BankClearance(Document):
 		if not self.include_reconciled_entries:
 			condition = "and (clearance_date IS NULL or clearance_date='0000-00-00')"
 
-		journal_entries = frappe.db.sql("""
+		journal_entry_accounts = frappe.db.sql("""
 			select
-				"Journal Entry" as payment_document, t1.name as payment_entry,
+				"Journal Entry Account" as payment_document, t2.name as payment_entry,
 				t1.cheque_no as cheque_number, t1.cheque_date,
-				sum(t2.debit_in_account_currency) as debit, sum(t2.credit_in_account_currency) as credit,
-				t1.posting_date, t2.against_account, t1.clearance_date, t2.account_currency
+				t2.debit_in_account_currency as debit, t2.credit_in_account_currency as credit,
+				t1.posting_date, t2.against_account, t2.clearance_date, t2.account_currency
 			from
 				`tabJournal Entry` t1, `tabJournal Entry Account` t2
 			where
 				t2.parent = t1.name and t2.account = %(account)s and t1.docstatus=1
 				and t1.posting_date >= %(from)s and t1.posting_date <= %(to)s
 				and ifnull(t1.is_opening, 'No') = 'No' {condition}
-			group by t2.account, t1.name
 			order by t1.posting_date ASC, t1.name DESC
 		""".format(condition=condition), {"account": self.account, "from": self.from_date, "to": self.to_date}, as_dict=1)
 
@@ -47,7 +46,7 @@ class BankClearance(Document):
 				"Payment Entry" as payment_document, name as payment_entry,
 				reference_no as cheque_number, reference_date as cheque_date,
 				if(paid_from=%(account)s, paid_amount, 0) as credit,
-				if(paid_from=%(account)s, 0, received_amount) as debit,
+				if(paid_to=%(account)s, 0, received_amount) as debit,
 				posting_date, ifnull(party,if(paid_from=%(account)s,paid_to,paid_from)) as against_account, clearance_date,
 				if(paid_to=%(account)s, paid_to_account_currency, paid_from_account_currency) as account_currency
 			from `tabPayment Entry`
@@ -88,7 +87,7 @@ class BankClearance(Document):
 					pi.posting_date ASC, pi.name DESC
 			""", {"account": self.account, "from": self.from_date, "to": self.to_date}, as_dict=1)
 
-		entries = sorted(list(payment_entries) + list(journal_entries + list(pos_sales_invoices) + list(pos_purchase_invoices)),
+		entries = sorted(payment_entries + journal_entry_accounts + pos_sales_invoices + pos_purchase_invoices,
 			key=lambda k: k['posting_date'] or getdate(nowdate()))
 
 		self.set('payment_entries', [])
--- a/erpnext/accounts/doctype/bank_transaction/bank_transaction.js
+++ b/erpnext/accounts/doctype/bank_transaction/bank_transaction.js
@@ -2,14 +2,42 @@
 // For license information, please see license.txt
 
 frappe.ui.form.on('Bank Transaction', {
-	onload(frm) {
+	onload: function(frm) {
 		frm.set_query('payment_document', 'payment_entries', function() {
 			return {
 				"filters": {
-					"name": ["in", ["Payment Entry", "Journal Entry", "Sales Invoice", "Purchase Invoice", "Expense Claim"]]
+					"name": ["in", ["Payment Entry", "Journal Entry Account", "Sales Invoice Payment", "Purchase Invoice"]]
 				}
 			};
 		});
+	},
+	refresh: function(frm) {
+		if (frm.doc.payment_entries.length > 0) {
+			frm.doc.payment_entries.forEach(entry => {
+				let parent_type = "";
+				if (entry.payment_document === "Journal Entry Account") {
+					parent_type = "Journal Entry";
+				}
+				if (entry.payment_document === "Sales Invoice Payment") {
+					parent_type = "Sales Invoice";
+				}
+				if (parent_type) {
+					frappe.db.get_value(entry.payment_document, {'name': entry.payment_entry}, 'parent', msg => {
+						frm.add_custom_button(entry.payment_document + __(" for ") + entry.allocated_amount,
+							function() {
+								frappe.set_route("Form", parent_type, msg.parent);
+							}, __('View Reconciled Payment...')
+						);
+					}, parent_type);
+				} else {
+					frm.add_custom_button(entry.payment_document + __(" for ") + String(entry.allocated_amount),
+						function() {
+							frappe.set_route("Form", entry.payment_document, entry.payment_entry);
+						}, __('View Reconciled Payment...')
+					);
+				}
+			});
+		}
 	}
 });
 
@@ -24,8 +52,11 @@ const update_clearance_date = (frm, cdt, cdn) => {
 		frappe.xcall('erpnext.accounts.doctype.bank_transaction.bank_transaction.unclear_reference_payment',
 			{doctype: cdt, docname: cdn})
 			.then(e => {
-				if (e == "success") {
-					frappe.show_alert({message:__("Document {0} successfully uncleared", [e]), indicator:'green'});
+				if (e.status == "success") {
+					frappe.show_alert({
+						message: __("Document {0} successfully uncleared", [e.entry]),
+						indicator: 'green'
+					});
 				}
 			});
 	}
--- a/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
+++ b/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
@@ -45,24 +45,115 @@ class BankTransaction(StatusUpdater):
 	def clear_linked_payment_entries(self):
 		for payment_entry in self.payment_entries:
 			allocated_amount = get_total_allocated_amount(payment_entry)
-			paid_amount = get_paid_amount(payment_entry, self.currency)
+			paid_amount = self.get_paid_amount(payment_entry.payment_document,
+				payment_entry.payment_entry, self.currency)
 
 			if paid_amount and allocated_amount:
 				if  flt(allocated_amount[0]["allocated_amount"]) > flt(paid_amount):
-					frappe.throw(_("The total allocated amount ({0}) is greated than the paid amount ({1}).").format(flt(allocated_amount[0]["allocated_amount"]), flt(paid_amount)))
+					frappe.throw(_("The total allocated amount ({0}) is greater than the paid amount ({1}).").format(flt(allocated_amount[0]["allocated_amount"]), flt(paid_amount)))
 				else:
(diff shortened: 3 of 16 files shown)


gwhitney:
The associated PR to Frappe which would enhance this PR (in particular, prevent accumulating fields in a Bank Reconciliation screen and ensure that only unreconciled Bank Transactions of the Bank Account being reconciled are shown) is https://github.com/frappe/frappe/pull/11776.

gwhitney:
Checked in mostly whitespace changes for the sake of Frappe Linter, Codacy, and Sider checks.

gwhitney:
UPDATE: received word from the Frappe framework maintainers that changes are desired with the associated pull request frappe/frappe#11776. Therefore, this PR for ERPNext should be considered Work-in-Progress (and so it does **not** make sense to merge it into develop) until a mutually agreeable change is settled on for the necessary Frappe changes, and I can update this PR correspondingly.

In addition, now with more experience running with this code change locally, I see that there is a need to add back in the ability to reconcile all of the Journal Entry Account lines in a single Journal Entry simultaneously, as sometimes they match the Bank Transaction as a whole, and it is extra work to reconcile them one at a time. (But I need to preserve the new ability to also reconcile them individually, as sometimes it is a single Journal Entry Account line that matches a Bank Transaction, even though there are other lines in the same Journal Entry.)

As I don't see a way to mark this as work-in-progress, please let me know if you prefer that I close it for now and file a new PR when the above amendments are ready, or leave this open and simply update it when ready. Thanks!

gwhitney:
UPDATE: I have received feedback on the Frappe framework change https://github.com/frappe/frappe/pull/11776 enhancing the standard fields feature of the frappe.ui.Page, and updated this pull request accordingly. I once again am recommending it for merging into the main development line of ERPNext as an improvement to the bank reconciliation process.

 In addition, I have taken the time while waiting for the Frappe issue to resolve to further organize the improvements to the bank reconciliation system in this pull request. The core principles are now laid out in a significant comment block at the top of `erpnext/accounts/page/bank_reconciliation/bank_reconciliation.py` (let me know if you would like them moved to somewhere in the developer documentation; if so, just point out where). But in short, the basic idea is that Bank Transactions can only reconcile against payment documents that correspond to a single General Ledger Entry against the associated Account of the Bank Account of the Bank Transaction. Other document types, like Expense Claim or Journal Entry which might correspond to multiple individual items like Payment Entry or Journal Entry Account, are handled just as convenient shortcuts in the reconciliation process. This coherent view of the reconciliation process still (of course) allows the two sides of a transfer between banks to reconcile against two different Journal Entry Account lines of a Journal Entry recording the transfer. It also leads to numerous other improvements, detailed in the commit message. Looking forward to your feedback.

gwhitney:
Checked in mostly whitespace changes to satisfy the Sider/Codacy checks.

gwhitney:
I see that there is an SQL error in a query generated by the test suite. My apologies -- as noted in a separate issue, I am unable to run the test suite locally. I will track down the error and amend the pull request within a day. Thanks very much for your patience.

gwhitney:
OK, I just amended the pull request to (hopefully) fix the SQL syntax errors. Also, I changed the special custom_filter_configs value that will trigger no adding of fields in the (Base) List view on the Bank Reconciliation page to one that is innocuous for existing Frappe code, so that this change can be merged into develop without breaking anything, and then the field non-duplication will take effect when the related Frappe pull request as linked above is merged.

gwhitney:
Two more small fixes uncovered by the Travis run.

gwhitney:
[NOTE: This comment appears to be moot, as it appears the Codacy test has been dropped.]

I don't understand the Codacy issue:

BankTransaction.get_paid_amount is too complex (19) (MC0001)
def get_paid_amount(self, entry_type, entry_name, currency):

What is the difficulty here and how should I fix it?

Other than that, all of the checks pass. I had to slightly update the Bank Transaction test suite to be sure that the Company values of the inserted transactions were correct. All now appears to be well, and I look forward to your feedback.

gwhitney:
Just corrected two small typos in the SQL query for Purchase Invoices in the Bank Reconciliation Statement. I remain confident in this pull request and look forward to your feedback. Thanks!

gwhitney:
Sider parameters appear to have changed, in that the latest run produced new failures that were not flagged before. In any case, I have checked in fixes for everything that was alerted. Mostly it was whitespace and  punctuation, but I will say that Sider also caught one bona fide bug (that pre-existed this PR, but is now fixed) -- namely, the Account was not being filled in correctly in any new Payment Entry created from the Reconciliation Tool.

gwhitney:
The Travis error is in the procurement tracker, so does not seem to have to do with this PR. Let me know if there's anything else you'd like done before this PR is reviewed.

gwhitney:
Rebased on develop to avoid merge conflicts.

nameduser0:
@gwhitney Hi Glen, I'm also currently working on some bank transaction/journal entry stuff and found this so thought I'd add my thoughts. I think the reason it doesn't already do this is that normally there would be two different journal entries.
E.g.
Journal1: From Bank1 to Limbo Account
Journal2: From Limbo to Bank2
Consider the scenario where a transfer takes multiple days to clear that bridge a month end for example.

It's fairly unusual to have journals going directly from one bank to another I would think (although it makes perfect sense to me), just like a customer invoice:
Journal1: Sales and debtors (Sale)
Journal2: Debtors and bank (Payment)

So you need to be careful that all the other use cases where there are at least two journals for a transaction aren't messed up. Having said all that, I'm no accountant, so the validity of this should really be checked by someone who is.
For all I know you know this already and have made sure everything's okay, so feel free to ignore this!

gwhitney:
In re casesolved-co-uk's comment, I must say that I too was not originally trained as an accountant. That said, I have years of experience with not-for-profit accounting in practice, and none of the institutions I've ever worked with record transfers as two journal entries, nor have we ever had an auditor recommend doing it that way. Moreover, here's the top hit I got by websearching for "bookkeeping journal entry for bank transfer": https://support.accountingseed.com/hc/en-us/articles/217569887-Record-Bank-Transfers and it clearly instructs the reader to make one entry, crediting the "From" account and debiting the "To" account. So I stand by my position that there really isn't any technically correct way to reconcile Bank Transactions other than against individual Journal Entry Account rows rather than against entire Journal Entry records. Hence, the need for this PR, which unfortunately appears will have to be re-done from scratch once #24273 lands.

nameduser0:
@gwhitney No problem, as long as you've done the research it's all good. I agree with you, I think this is a use-case that should be catered for, even if it means changing the core.

Having looked at the sheer volume of the changes though, if you do end up resubmitting it, I'd suggest breaking it down into multiple meaningful independent PR's. I know this may not be easy or even possible. This affects the very core of the product, so as you can imagine, ERPNext will be very cautious of huge changes, possibly to the point of ignoring them completely.

gwhitney:
Yes, well it seems pretty clear this will have to be completely redone, since it received no attention for three months and now there are a bunch of incompatible changes coming down the pipeline. I guess my plan is to wait for them to land, and then re-do this. I like your suggestion of trying to do it in clearer steps. I doubt that I can necessarily do it in multiple PRs -- maybe I can split it a bit that way -- but at least I can present it as a series of smaller, clearly-explained commits within the PR. Thanks!

MichaelPinkowski:
@gwhitney Glen, as the original sponsor of the Plaid integration (and related Bank Reconciliation code) I just want to take a minute to thank you for the incredible effort that you have put in here.  You are clearly bringing a bunch of expertise in code/ accounting, and UI that is really valuable.  Thanks so much for this contribution.

nameduser0:
@gwhitney 
> I doubt that I can necessarily do it in multiple PRs -- maybe I can split it a bit that way -- but at least I can present it as a series of smaller, clearly-explained commits within the PR

My only thoughts on splitting it up is:

- Make your changes so that you can do what you need, even if it means replicating fields (two different clearance dates, one at each level)
- Once that is proven, migrate existing db entries and code to the JEA level, maybe a document at a time?
- Once that is proven, remove the old fields and extinct code

I know, a lot of work and sustained involvement, but it might be the only way to actually get it in and is probably safer. I agree more attention from the core devs wouldn't go amiss, but I guess they're busy with other agendas

gwhitney:
> Make your changes so that you can do what you need, even if it means replicating fields (two different clearance dates, one at each level)

That sounds to me like a recipe for disaster; redundant, potentially conflicting information in the database? In the intermediate stage, could a Bank Transaction reconcile against either a Journal Entry or a Journal Entry Account? How would one choose? I am very afraid that I would just make things much worse if I tried to enable reconciliation against Journal Entry Account records first and then afterwards disable reconciliation against Journal Entry records. But I do very much appreciate your efforts to make suggestions that could help this get in -- thank you! I will think about your ideas when things settle with respect to Bank Reconciliation and I make another crack at this.

gwhitney:
@MichaelPinkowski thank you for the encouraging words!

nameduser0:
@gwhitney Because you're catering for one very specific niche use-case, as an example: if JE already reconciled, reconcile to JEA. I only briefly looked at the code so not sure what's involved, so these are just thoughts. But yeah, just trying to help!

nameduser0:
> one very specific niche use-case

Actually I take that back, this will be quite common where the company has credit cards linked to their main bank account. They will pay the credit card with a transfer, and both accounts are likely to have their own GL account

gwhitney:
Right, we certainly have to reconcile both sides of a single Journal Entry fairly often. Thanks for that supportive comment, casesolved-co-uk!

PR #23728, opened 2020-10-26
feat: Close Production Plan

marination:
**Documentation:** https://github.com/frappe/erpnext_documentation/pull/185

**Case:**
- User creates a Sales Order with 3 items
- They then create a Production Plan for the Same 
- One of the three items (item A) is produced via an Independent Work Order, not linked with Prod Plan
- The other two are produced via Production Plan (Item B and C)
- Now the Production Plan can't be completed since it will throw an overproduction against Sales Order Error if you try to produce Item A, it just lays incomplete

**Solution:**
- Allow users to **Close** a Production Plan. This essentially tells users that no more Work orders to be created against this Production Plan
- The same can be **Re-opened**
prod-plan

**Miscellanous Fix**
- **Production Plan Material Request Warehouse** is a table multiselect field in Production Plan (hidden)
- It must be a child table at the doctype level ([source](https://docs.erpnext.com/docs/user/manual/en/customize-erpnext/articles/table-multiselect-field)). Error while saving doctype saying the same


marination:
Translation Syntax fix is here https://github.com/frappe/erpnext/pull/23753

rohitwaghchaure (approved):

PR #23729, opened 2020-10-26
fix: copying po no when mapping doc

AfshanKhan:
Usually, the flow goes like, a Purchase Order is created from which a Sales Order is created which has a field "Customer's Purchase Order" which gets updated with Purchase Order Number. Now a "Delivery Note" or a "Sales Invoice" can be created from that "Sales Order". After creating either "Delivery Note" or "Sales Invoice" when we create the other one i.e. Sales Invoice from Delivery Note or Delivery Note from Sales Invoice, the "Customer's Purchase Order" field didn't use to get copied from either of both. Fixed that.
Screenshot 2020-10-26 at 5 02 29 PM

Screenshot 2020-10-26 at 5 02 47 PM


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/23730

erpnext/controllers/selling_controller.py
```diff
@@ -371,13 +371,39 @@ def update_stock_ledger(self):
...
+		sales_orders = list(set([d.get(ref_fieldname1) for d in self.items if d.get(ref_fieldname1)]))
+		if sales_orders:
+			so_po_nos = frappe.get_all('Sales Order', 'po_no', filters = {'name': ('in', sales_orders)})
+			if so_po_nos and so_po_nos[0].get('po_no'):
+				self.po_no += ', '.join(list(set([d.po_no for d in so_po_nos if d.po_no])))
+		delivery_notes = list(set([d.get(ref_fieldname2) for d in self.items if d.get(ref_fieldname2)]))
+		if delivery_notes:
+			dn_po_nos = frappe.get_all('Delivery Note', 'po_no', filters = {'name': ('in', delivery_notes)})
+			if dn_po_nos and dn_po_nos[0].get('po_no'):
+				self.po_no += ', '.join(list(set([d.po_no for d in dn_po_nos if d.po_no])))
```
nabinhait: The current code can lead to repititive po_no if same po_no is present in Sales Order and Delivery Note.

```suggestion
		po_nos = []
		sales_orders = list(set([d.get(ref_fieldname1) for d in self.items if d.get(ref_fieldname1)]))
		if sales_orders:
			so_po_nos = frappe.get_all('Sales Order', 'po_no', filters = {'name': ('in', sales_orders)})
			if so_po_nos and so_po_nos[0].get('po_no'):
				po_nos = [d.po_no for d in so_po_nos if d.po_no]

		delivery_notes = list(set([d.get(ref_fieldname2) for d in self.items if d.get(ref_fieldname2)]))
		if delivery_notes:
			dn_po_nos = frappe.get_all('Delivery Note', 'po_no', filters = {'name': ('in', delivery_notes)})
			if dn_po_nos and dn_po_nos[0].get('po_no'):
				po_nos += [d.po_no for d in dn_po_nos if d.po_no]

		self.po_no = ', '.join(list(set(po_nos)))
```

nabinhait: Also you can write a small function `get_po_nos(ref_doctype, ref_fieldname)` which will return po_nos from the ref doc.

nabinhait (changes requested):

erpnext/controllers/selling_controller.py
```diff
@@ -371,13 +371,29 @@ def update_stock_ledger(self):
...
+		po_nos = []
+		self.get_po_nos('Sales Order', 'against_sales_order', po_nos)
+		self.get_po_nos('Sales Invoice', 'against_sales_invoice', po_nos)
+		self.po_no = ', '.join(list(set(po_nos)))
+
+	def get_po_nos(self, ref_doctype, ref_fieldname, po_nos):
+		doc_list = list(set([d.get(ref_fieldname) for d in self.items if d.get(ref_fieldname)]))
+		if doc_list:
+			po_no_list = frappe.get_all(ref_doctype, 'po_no', filters = {'name': ('in', doc_list)})
+			if po_no_list and po_no_list[0].get('po_no'):
```
marination: we could have a dict such as `[{'po_no': None}, {'po_no': 'PO-1234'}]`, in that case it just skips the whole `po_no_list `

marination:
@AfshanKhan   
**Case:** 
Create two Sales Orders. One with Po no 'PO-1234' and one with 'PO-5678'.
Fetch these in one Sales Invoice (get Items from) and submit. Now Sales Invoice Po no is `PO-1234, PO-5678`.
From this Sales Invoice create a Delivery Note . The Po no in the Delivery Note is:
Screenshot 2020-11-09 at 2 41 07 PM

When the list looks like this => ['**PO-5678**', '**PO-1234**', '**PO-1234, PO-5678**'], `set` wont work since the last element is a comma joined string.

marination (approved):

PR #23735, opened 2020-10-27
fix: Payment Terms not fetched in Purchase Invoice from Purchase Receipt

marination:
**Issue:**
- Create PO and PR against it
- Now create a Purchase Invoice against this Purchase Receipt, the Payment Terms Template is fetched but the Terms are not fetched from the Template
- This occurs due to the incorrect order of events. 
- From PR to PI, PI's `onload` is called via the doc mapper, at this time it gets no Payment Terms Template. Thus, it adds a row for Payment terms with invoice portion 100%
- After this, Template is fetched and set from client side, but since table is populated, it does nothing
ptt_issue

**Fix:**
- Set the Template while mapping the doc so that during `onload` it will fetch the correct Terms
- Renamed `get_pyt_term_template` => `get_payment_terms_template`
ptt


marination:
Screenshot 2020-10-27 at 2 50 21 PM

marination:

PR #22197, opened 2020-06-11
fix: set_query in leave application

Anurag810:
allowed LWPS in set_query

PR #23748, opened 2020-10-28
fix(HR) : Filter Leave Type based on allocation for a particular employee (#22050)

marination:
Port of https://github.com/frappe/erpnext/pull/22050 and https://github.com/frappe/erpnext/pull/22197


abhishekbalam (approved):
Seems ok 👍

marination:
Travis failed on "Could not find _Test Item Group", unrelated, 2.7 py tests passed

PR #23749, opened 2020-10-28
fix: Do not allow Company as accounting dimension

deepeshgarg007:
Company doctype should not be allowed as an accounting dimension as Company is already linked with every GL Entry so it doesn't make sense making it as an accounting dimension.

Also making company as a dimension breaks some accounting reports as company is used as keys in many queries for filtering

For Eg:
`
frappe.db.sql("""
		select
			name as gl_entry, posting_date, account, party_type, party,
			voucher_type, voucher_no, {dimension_fields}
			cost_center, project,
			against_voucher_type, against_voucher, account_currency,
			remarks, against, is_opening, creation {select_fields}
		from tabGL Entry
		where company=%(company)s {conditions}
		{distributed_cost_center_query}
		{order_by_statement}
		""".format(dimension_fields=dimension_fields, select_fields=select_fields, conditions=get_conditions(filters), distributed_cost_center_query=distributed_cost_center_query,order_by_statement=order_by_statement),ilters, as_dict=1)
` 

image


deepeshgarg007:

PR #23757, opened 2020-10-29
fix: Ignore cancelled entries in stock balance report

deepeshgarg007:
Cancelled stock entries should not be considered in Stock Balance report as it inflates the In and Out quantity values because  of cancelled transactions

Before:
Screenshot 2020-10-29 at 12 23 10 PM

After:
Screenshot 2020-10-29 at 12 09 46 PM


deepeshgarg007:

nabinhait (approved):

PR #23764, opened 2020-10-29
fix: added project link in timesheet form

anupamvs:
Added project link in the timesheet parent form it will auto add the project in the child table (time logs)
timesheet


rohitwaghchaure (approved):

wojosc:
Nice move! Why not also do this for Task while your at it? :)

(If you ask me I can't see the necessity for Time Sheet Detail. This could be easily be shown in List or Reportview using filters)

PR #23773, opened 2020-10-29
fix: po_detail field has no value for subcontracted stock entry

rohitwaghchaure:
**Issue**

1. po_detail field is hidden which used to link the Purchase Order Supplied Items with the subcontracted Stock Entry.

1. If user has manually added row in the subcontracted Stock Entry then po_detail field has set as blank

**After Fix**
If user has manually added row in the subcontracted Stock Entry then automatically system will set the value for the field po_detail.


rohitwaghchaure:
@Mergifyio backport develop

PR #23806, opened 2020-11-04
Fix leave ledger patch

AfshanKhan:
While running the patch if there was any existing entry in the Leave Ledger Entry List, the patch wasn't executed further. Hence not creating missing Leave Ledger entries.


Anurag810 (commented):
reviewed

PR #23809, opened 2020-11-04
fix: Added link of bank reconciliation and clearance in accounting desk page

nabinhait:
Screenshot 2020-11-04 at 1 01 15 PM


marination:
@nabinhait https://github.com/frappe/erpnext/pull/23850

PR #23836, opened 2020-11-05
fix: correcting description field in taxes and charges for accounts with account number + account name

pateljannat:
**Issue:**
1. Wrong Description was getting set in Taxes and Charges for Accounts having account number + account name.

wrong description

2. This was also visible on Print Formats.

affecting print format

**Fix:**
Screenshot 2020-11-05 at 5 07 16 PM
Screenshot 2020-11-05 at 5 13 53 PM


erpnext/public/js/controllers/accounts.js
```diff
@@ -146,18 +146,21 @@ cur_frm.cscript.account_head = function(doc, cdt, cdn) {
...
 			callback: function(r) {
-				frappe.model.set_value(cdt, cdn, "rate", r.message.tax_rate || 0);
-				frappe.model.set_value(cdt, cdn, "description", r.message.account_name);
+				if(d.charge_type!=="Actual"){
+					frappe.model.set_value(cdt, cdn, "rate", r.message.tax_rate || 0);
+					frappe.model.set_value(cdt, cdn, "description", r.message.account_name);
+				}
+				else if(d.charge_type == 'Actual'){
+					frappe.model.set_value(cdt, cdn, "description", r.message.account_name);
+				}
```
surajshetty3416: code can be simplified by
```suggestion
				if (d.charge_type !== "Actual") {
					frappe.model.set_value(cdt, cdn, "rate", cint(r.message.tax_rate));
				}
				frappe.model.set_value(cdt, cdn, "description", r.message.account_name);
```
|

Also, please make sure if we are not introducing [this bug again](https://github.com/frappe/erpnext/pull/13843)

surajshetty3416 (changes requested):

deepeshgarg007 (approved):

PR #23837, opened 2020-11-05
feat: added column cost_center to receivable reports

Anurag810:
https://github.com/frappe/erpnext/pull/23835


erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -160,6 +160,8 @@ def update_voucher_balance(self, gle):
 			else:
 				# advance / unlinked payment or other adjustment
 				row.paid -= gle_balance
+		if gle.cost_center:
+			row.cost_center =  str(gle.cost_center)
```
rohitwaghchaure: Why str?

nabinhait (approved):

PR #23879, opened 2020-11-10
fix: customer po no not copy from SI to DI and DI to SI

rohitwaghchaure:
1. Make Sales Invoice without sales order and put customer PO number.

1. Make delivery note against the above sales invoice and check customer PO number field in the sales invoice

1. The customer PO number field has no value


marination:
@rohitwaghchaure  duplicate https://github.com/frappe/erpnext/pull/23729

Issue #22411, opened 2020-06-23
Notify customers n days before subscription is due


laydros:
### Current behavior:
Currently the Subscription DocType sends an invoice to a customer at the end of the subscription period. The module supports a "grace period" allowing the customer a certain number of days after the end of the subscription period to send payment before ending the plan.

### Desired change:
I would like the option to instead send an invoice to the customer a certain number of days before the end of the subscription period. This way the plan will still end on the end of the subscription period, but the customer will be notified ahead of time and can send payment before the period is over.

### Alternative considered:
In practice, the current behavior and my requested behavior are essentially the same. However the way we communicate subscriptions to our customers is a prepaid model. They have a subscription for a given period, and they must pay ahead of time for the next period. 

The UI for this could be very similar to the current UI for grace period. Perhaps there could be a dropdown to choose between grace period and invoice ahead (since it would be odd to use both) and the number of days currently for grace period could be be instead used for the invoice ahead option.

laydros:
@deepeshgarg007 @nextchamp-saqib @PawanMeh I've set up code to implement this, so I wanted some input on how I could format it for submission. The [contribution guidelines](https://github.com/frappe/erpnext/wiki/Preparing-a-Contribution-Proposal) say to contact the module maintainers before sending a pull request. This is the first time I have contributed to an open source project, so I appreciate your patience and guidance. 

Basically I added a checkbox to enable early invoice and an int for days before end to the subscription doctype, and added a check to process_for_active that will trigger generate_invoice if it is checked and within the number of days before.

I wondered if instead this function should be wrapped into the "Generate invoice at beginning of period" feature.

![Screenshot_2020-10-21 ACC-SUB-2020-00004](https://user-images.githubusercontent.com/246713/96787746-46f45b80-13c0-11eb-8aae-4ebe407e476e.png)
Thanks for any advice on this.

PR #23881, opened 2020-11-10
feat: trigger subscription invoice n days before period end

laydros:
Add ability to trigger invoice *n* days before end of period. 

- Allows generation of subscription invoice a configurable number of days before the end of the period.
- This gives the customer time to make arrangements to pay before the end of the subscription period.
- Adds new checkbox "Generate Invoice Early," and a field to specify how many days early.

For issue #22411

**Documentation PR:** [frappe/erpnext_documentation#192](https://github.com/frappe/erpnext_documentation/pull/192)

erpn-subscription_early_invoice


laydros:
@deepeshgarg007 @nextchamp-saqib do you guys have any feedback on this?

deepeshgarg007:
@laydros the Days until due here can be kept as 10, right? Why do we need a separate field for this?

laydros:
> @laydros the Days until due here can be kept as 10, right? Why do we need a separate field for this?

@deepeshgarg007 the idea is that one could adjust how long before the due date to generate the invoice. For example my employer typically sends the invoice to the customer 60 days before due.

deepeshgarg007:
> > @laydros the Days until due here can be kept as 10, right? Why do we need a separate field for this?
> 
> @deepeshgarg007 the idea is that one could adjust how long before the due date to generate the invoice. For example my employer typically sends the invoice to the customer 60 days before due.

Yes, so in this case the days until due can be kept as 60 and the subscription start date can be put accordingly, why introduce one more field for it?

laydros:
> > > @laydros the Days until due here can be kept as 10, right? Why do we need a separate field for this?
> > 
> > 
> > @deepeshgarg007 the idea is that one could adjust how long before the due date to generate the invoice. For example my employer typically sends the invoice to the customer 60 days before due.
> 
> Yes, so in this case the days until due can be kept as 60 and the subscription start date can be put accordingly, why introduce one more field for it?

I added the additional field in case another user might want to configure the feature for a different time period, and be more useful in general to ERPNext. For my companies use it will be fine to just have a single checkbox to generate the invoice 60 days before due date.

Would you like me to update the code to work in that manner? Should I instead add the field for number of days early to subscription settings?

laydros:
@deepeshgarg007 Hi, I just wanted to follow up if you had seen my last comment. Do you want me to update the code with no date configuration and just hard-code it to 60 days? Or is it something I should try to add to a settings page instead of being on the subscription page?

deepeshgarg007:
@laydros closing this for now. Doesn't seem to be requested by many users. If many people support this then we can reopen.

asharamseervi:
Hi Deepesh, It's not fair to close PR if it not have raised any feature request earlier! Instead you should have discussed, what's actually wrong or how this feature should have contributed as per guidelines, etc.

Subscription module is too generic at ERPNext, and this feature will allow subscribers (customers) to renew or pay early fixed price subscription! Potential prospect was looking for this feature early 2021, dropped due to less features in subscription management!

PR #23896, opened 2020-11-12
fix: duplicate items validation for POS Invoice when allow multiple items is disabled

pateljannat:
**Issue:** 

1.  In **Selling Settings** disable **Allow Item to Be Added Multiple Times in a Transaction**.
2. Fill a POS Invoice and save it.
3. An error used to appear.

Screenshot 2020-11-12 at 4 50 32 PM

4. This was because the function that validates duplicate items had no handling for POS Invoice.


nabinhait (approved):

Papitom101:
My client is experiencing this right now. I will appreciate a solution as soon as possible. I have checked add multiple items, still no solution

PR #23903, opened 2020-11-13
fix: Don't copy terms and discount from SO to PO

marination:
**Issue:**
While creating PO from SO:
- Terms were copied. Terms for PO and SO are usually not the same.
- Discount was copied along with pricing rules in the line item.

**Fix:**
- Terms, Discount and Pricing Rule are not copied with our without drop shipping (against default supplier or not)

<details><summary>Usual Case:</summary>

Consider Selling Price set as 120 and selling rule sets a 10% discount. Buying Price is 100 with no rule (no discount)

mapping-fix-final

</details>

<details><summary>Drop Shipping:</summary>
Consider Selling Price set as 120 and selling rule sets a 10% discount. Buying Price is 100 with no rule (no discount)

mapping-fix-dropship

</details>


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -930,14 +934,16 @@ def update_item(source, target, source_parent):
 				["parent", "sales_order"],
 				["stock_uom", "stock_uom"],
 				["uom", "uom"],
-				["conversion_factor", "conversion_factor"],
-				["delivery_date", "schedule_date"]
```
nabinhait: Seems it was correct.
"Expected delivery date" in SO can be set as "Required By Date" in PO

nabinhait (changes requested):

nabinhait (approved):

PR #23914, opened 2020-11-16
fix: pos item search includes non stock items

nextchamp-saqib:
- POS search query doesn't excludes non stock items
- POS Invoice has update stock checked by default so adding a non stock item from Point of Sale will throw error while merging such invoice.


muchai:
What happens if you want to include an item like transport charges, which are not stock items, while doing a POS invoice?

PR #23935, opened 2020-11-17
fix: duplicate rows in earning tables if not overewrite

Anurag810:
the Additional salary components with overwrite property applied against an employee are appearing more than once in a Salary Slip.
Screenshot 2020-11-17 at 6 18 26 PM


Anurag810:
closing for now.

PR #23940, opened 2020-11-18
fix: only set to none auto created batches

aliriocastro:

In Stock Reconciliation, only auto-created batches should be set to None, if we set to None all of them, it becomes difficult when trying only to amend the Stock Reconciliation, last behavior was that we have to set all (batch, qty and valuation rate)

> Screenshots/GIFs


rohitwaghchaure:
@aliriocastro you have sent PR on stable version, kindly send PR on version-12-hotfix. Also travis is failing kindly fix the travis as well

PR #23941, opened 2020-11-18
fix: place of supply change on address change

pateljannat:
**Issue:**
1. In Doctypes Sales Order, Purchase Order, Purchase Invoice and Purchase Receipt, place of supply used to not change on change of address.
2. On Saving however the right place of supply was set. So the change was happening on the server-side, but the user had no feedback before saving the document.

Screenshot 2020-11-18 at 10 16 31 AM


deepeshgarg007 (changes requested):
@pateljannat place of supply is a regional field for India. All the handling for this should be done only in the regional controller files and not in standard controller files

erpnext/regional/india/utils.py
```diff
@@ -236,7 +236,7 @@ def get_tax_template(master_doctype, company, is_inter_state, state_code):
 		if tax_category.gst_state == number_state_mapping[state_code] or \
 	 		(not default_tax and not tax_category.gst_state):
 			default_tax = frappe.db.get_value(master_doctype,
-				{'company': company, 'disabled': 0, 'tax_category': tax_category.name}, 'name')
```
deepeshgarg007: No need to remove this company filter

deepeshgarg007 (changes requested):

deepeshgarg007 (changes requested):
Remove return_taxes flag from line no 153 and from taxes.js as well

deepeshgarg007 (approved):

Issue #21337, opened 2020-04-20
BUG: User None not Found in Email Digest


dawoodjee:
I’m running the latest version of Frappe and ERPNext in production. No customisation or modifications have been done.

I’ve deleted and added users before, this problem started when I setup email digest and added all the users.

![image](https://user-images.githubusercontent.com/7912989/79724531-04a27280-82e8-11ea-8380-8b51dc2ae6e7.png)

I cant login or do anything else. Other sites are fine.

Issue is only fixed by deleting Digest entry from the database

Orriginally posted on the forum: https://discuss.erpnext.com/t/cant-use-erpnext-due-to-user-not-found-error/57337

pateljannat:
This will be fixed with #23949

PR #23949, opened 2020-11-18
fix: email digest user not found

pateljannat:
**Issue:**
1. In an Email Digest, after clicking Send Now, the whole site used to become inaccessible and an error used to appear.

user not found

2. This used to happen because on clicking Send Now, the code was setting the recipient of email as the session user, send the email, and then reset the original user as session user.
3. For all this frappe.set_user function was used and this function clears the session after the user gets set. Hence the error.

**Fix:** Not changing the session user before sending the email.


pateljannat:
closes #21337

rohitwaghchaure (approved):

dawoodjee:
If we cannot use `frappe.set_user()` what's the right way to set permissions for a guest user's initiated action that inserts a record like Payment Entry for example?

PR #23963, opened 2020-11-19
fix: add remarks to sales invoice

vorasmit:
This is in line with purchase invoice remarks - sample remarks as shown in report
Sample purchase reports

Default for Sales Invoice was 'No Remarks'. Added remarks to make reports more meaningful.

**Before:**
Report-Before

Patch to update existing Sales Invoice and GL entry working

Patch Working

GL Patch Working

**After:**
Meaningful Reports


erpnext/patches.txt
```diff
@@ -677,4 +677,5 @@ erpnext.patches.v12_0.set_multi_uom_in_rfq
 erpnext.patches.v12_0.update_state_code_for_daman_and_diu
 erpnext.patches.v12_0.rename_lost_reason_detail
 erpnext.patches.v12_0.update_leave_application_status
-erpnext.patches.v12_0.update_payment_entry_status
\ No newline at end of file
+erpnext.patches.v12_0.update_payment_entry_status
+erpnext.patches.v12_0.update_sales_invoice_remarks
```
deepeshgarg007: Not sure if patch is needed for all instances. Some instances might have years old invoices with "No Remarks" which users might not even look again and this may take a lot of time for instances with high volume of data. This can be done only on specific instances where necessary in my opinion.

vorasmit: True.

Can add a filter for years, say update from 01-01-2019 or 01-01-2020. 
Or I can simply remove this patch.

What do you suggest?

deepeshgarg007 (changes requested):
@vorasmit I think the patch can be removed. People who want to update can do on their specific instances based on their requirement

deepeshgarg007 (dismissed):

PR #23969, opened 2020-11-20
feat: Putaway

marination:
Documentation: https://github.com/frappe/erpnext_documentation/pull/203
Depends on https://github.com/frappe/frappe/pull/11976

## Putaway Rule

- You can create a Putaway Rule that will specify what Item this rule is applicable on. It will also specify the capacity in a Warehouse for this Item. You can set a priority to determine which rule should be applicable first.
saved-putaway-rule

- This rule basically says Item **Rice (kg)** can be held in Warehouse **Zone 2: Rack 2 - HUL** which has a maximum capacity of 5000 kg **for this Item**.

## Working
- The Putaway Rules are applied on Purchase Receipts and Stock Entries (only of type **Material Receipt** and **Material Transfer**) 
- They are applied on the save of a Purchase Receipt/Stock Entry and also on enabling **Apply Putaway Rule** checkbox
   stock-entry-transfer-putaway

- If all the Rules are exhausted and the Warehouses are at capacity, feedback is given. 
 Screenshot 2020-12-08 at 7 35 11 PM

## Warehouse Capacity Summary

You can also see how much of your warehouses are occupied and can change the Capacity too via Warehouse Capacity Summary.
Screenshot 2020-11-26 at 10 16 06 AM

ToDo:

- [x] Improve feedback formatting for unallocated Items

- [x] Add UOM in Rule and Update as per UOM

- [x] Tests (Update tests with more use cases)

- [x] Documentation(update docs with new design)

- [x] Create a Capacity vs Current Stock Level Summary

- [x]  Change Rule Application Design(apply on save in form) and Block Over Receipt via Applied Rule

- [x] Apply Rule on checkbox trigger


marination:
Screenshot 2020-11-24 at 4 51 48 PM

rohitwaghchaure (approved):

PR #23973, opened 2020-11-20
fix: repositioned accounting details to relevant position

AfshanKhan:
In Sales Invoice and POS Invoice, both the Accounting detail section should be at the top of the form rather than at the end.
Before:

After changes:

Screenshot 2020-11-20 at 5 29 12 PM

Screenshot 2020-11-20 at 5 57 06 PM

Also In Purchase Invoice, a section called "More Information" actually contains payment information, changed it to "Payment Information"
Before:

After the change:
Screenshot 2020-11-20 at 5 57 51 PM


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/23974

nabinhait:
@AfshanKhan I think this change is not needed. This accounting section is rarely entered by the user, the system automatically set Debit/Credit To account on selection of party. Closing the PR.

PR #23974, opened 2020-11-20
fix: repositioned accounting details to relevant position

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/23973


nabinhait:
@AfshanKhan I think this change is not needed. This accounting section is rarely entered by the user, the system automatically set Debit/Credit To account on selection of party. Closing the PR.

PR #23975, opened 2020-11-21
fix: Validation for journal entry with 0 debit and credit values

deepeshgarg007:
Journal Entry with absolutely 0 debit and credit values was allowed to created.

Screenshot 2020-11-21 at 12 07 47 PM

This PR adds a validation for it
Screenshot 2020-11-21 at 12 07 39 PM


nabinhait (approved):

nabinhait (changes requested):
Travis failing, fix test cases

PR #23986, opened 2020-11-23
fix: Update .pylintrc

laydros:
pylint fails without headers. This adds the needed headers and allows tabs instead of spaces


ankush:
pylint config is removed now. Please use flake8 instead, both have pretty much the same functionality. 

flake8 config is already added in repo.

PR #23991, opened 2020-11-24
fix: job card error handling for operations field

pateljannat:
**Issue:**

1. In a Job Card, the Operation field has a filter based on the Work Order selected.
2. So when the user clicks on the Operation field without selecting a Work Order, an error appears.

Screenshot 2020-11-24 at 3 44 19 PM

**Fix:**

1. Error handling for the scenario where the Operation field is clicked before selecting Work Order.
2. Message is displayed to the user asking them to select a Work Order in this scenario.

Screenshot 2020-11-24 at 3 40 58 PM


rohitwaghchaure (approved):

rohitwaghchaure:
@pateljannat do backport on version-12-hotfix

PR #23998, opened 2020-11-24
Filters for tax templates

hasnain2808:
This PR elevates the application of "company filter" on "taxes and charges templates" from SellingController to TransactionController and also on POS profile

Previously only on doctypes that extended SellingController like SO used to get the filter now all doctypes that extend TranactionController(PO, PI, etc) will get it.

This solves the issue when the user selects the wrong template (usually of another company) and then gets the error on save(very common)

Improves usability since there can be lots of templates defined.

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -213,8 +213,8 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			this.frm.set_query("taxes_and_charges", function() {
 				return {
 					filters: [
-						['Sales Taxes and Charges Template', 'company', '=', me.frm.doc.company],
-						['Sales Taxes and Charges Template', 'docstatus', '!=', 2]
+						['company', '=', me.frm.doc.company],
+						['docstatus', '!=', 2]
 					]
 				};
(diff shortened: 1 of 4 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -209,6 +209,17 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			});
 		}
 
+		if (this.frm.fields_dict.taxes_and_charges) {
+			this.frm.set_query("taxes_and_charges", function() {
```
nabinhait, line 213: `taxes_and_charges` field is there in purchase cycle also, but it is linked to `Purchase Taxes and Charges Template`.

After this change, seems like purchase cycle doctype will also be linked to "Sales Taxes and Charges Template".

nabinhait (changes requested):

erpnext/public/js/controllers/transaction.js
hasnain2808: Fixed:
Removed the doctype name since it gets added according to the doctype in link.js

nabinhait (approved):

PR #24001, opened 2020-11-24
fix: incorrect stock quantity if 'Allow Multiple Material Consumption…

rohitwaghchaure:
1. Enable "Allow Multiple Material Consumption" in manufacturing settings

1. Make work order for quantity 4 and complete finished goods for 2 quantity

1. After, make material consumption for 2 quantity, system will show incorrect quantity in the item line
stock_qty_issue

**After Fix** 

stock_qty_after_fix


nabinhait (changes requested):
Add a test case

PR #24009, opened 2020-11-26
fix: shipping charges not sync from shopify

rohitwaghchaure:
If shipping charges added in the Shopify then while syncing orders with erpnext the rates are not added

### DIFF
--- a/erpnext/erpnext_integrations/connectors/shopify_connection.py
+++ b/erpnext/erpnext_integrations/connectors/shopify_connection.py
@@ -244,6 +244,15 @@ def update_taxes_with_shipping_lines(taxes, shipping_lines, shopify_settings):
 	"""Shipping lines represents the shipping details,
 		each such shipping detail consists of a list of tax_lines"""
 	for shipping_charge in shipping_lines:
+		if shipping_charge.get("price"):
+			taxes.append({
+				"charge_type": _("Actual"),
+				"account_head": get_tax_account_head(shipping_charge),
+				"description": shipping_charge["title"],
+				"tax_amount": shipping_charge["price"],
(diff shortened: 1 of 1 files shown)


leela:
Hi @rohitwaghchaure , I guess this will fix the issue with upcoming syncs. But what about already synced orders, How do we fix them?

nabinhait (approved):

erpnext/erpnext_integrations/connectors/shopify_connection.py
```diff
@@ -244,6 +244,15 @@ def update_taxes_with_shipping_lines(taxes, shipping_lines, shopify_settings):
...
+		if shipping_charge.get("price"):
+			taxes.append({
+				"charge_type": _("Actual"),
+				"account_head": get_tax_account_head(shipping_charge),
+				"description": shipping_charge["title"],
+				"tax_amount": shipping_charge["price"],
+				"cost_center": shopify_settings.cost_center
+			})
+
 		for tax in shipping_charge.get("tax_lines"):
```
Alchez, line 256: The `shipping_charge` price contains the cumulative value from all the tax lines. I don't think we should add both.

```suggestion
	for shipping_charge in shipping_lines:
		shipping_tax_lines = shipping_charge.get('tax_lines') or [shipping_charge]
		for tax in shipping_tax_lines:
```

Alchez (changes requested):

erpnext/erpnext_integrations/connectors/shopify_connection.py
rohitwaghchaure: We have a customer who apply the tax on shipping, so in his case we have to consider both shipping price as well tax applied on it.

PR #24019, opened 2020-11-27
Material request wrong status issue

pateljannat:
Depends on:  frappe/frappe#12010

**Issue:**

1. In a Material Request, when the qty of unordered items was very insignificant compared to the qty of ordered items, the document used to reflect wrong status.

Screenshot 2020-11-27 at 7 27 24 PM

2. In the above example even though the second item is not ordered, the status was reflecting as Ordered. The **% Ordered** field shows 100%. 

3. This issue was because of rounding. In the database, the proper value of the field was set, the rounding was on the frontend.


pateljannat:
Related Pr: https://github.com/frappe/frappe/pull/12010

erpnext/stock/doctype/material_request/material_request_list.js
```diff
@@ -1,24 +1,25 @@
...
+		} else if (doc.transfer_status && doc.docstatus != 2) {
 			if (doc.transfer_status == "Not Started") {
 				return [__("Not Started"), "orange"];
 			} else if (doc.transfer_status == "In Transit") {
 				return [__("In Transit"), "yellow"];
 			} else if (doc.transfer_status == "Completed") {
 				return [__("Completed"), "green"];
 			}
-		} else if(doc.docstatus==1 && flt(doc.per_ordered, 2) == 0) {
+		} else if (doc.docstatus==1 && flt(doc.per_ordered, precision) == 0) {
```
marination, line 15: problems here: 
1) if my precision is set to 2 in System Settings, we will still face this problem. `flt(99.9975, 2) = 100`
2) Let assume default precision is 3 , this works, status is Partially Ordered. But the per_ordered field still shows **100%**. This one is a frappe problem in Percent field formatter. 
Screenshot 2020-12-10 at 9 37 09 PM

This fix seems okay, but maybe fixing it on frappe too would perhaps complete the fix. Right now it is incomplete. You can check out the `Percent` formatter in `formatters.js` in Frappe. It rounds to two places only, maybe we can use precision there.

pateljannat: @marination I have already fixed the Percent field precision in frappe, https://github.com/frappe/frappe/pull/12010

nabinhait (approved):

PR #24022, opened 2020-11-28
fix: overriding po

AfshanKhan:
Issue : While Making the Delivery Note, User is entering the Customer PO Detail but While Saving the Delivery Note is Disappear the PO Details.
1) PO Details Entered
image

2) PO Details Disappears After saving.
image

To fix this we are now checking if the user has entered a PO and if yes we don't override it.


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24023

erpnext/controllers/selling_controller.py
```diff
@@ -371,9 +371,9 @@ def update_stock_ledger(self):
 		self.make_sl_entries(sl_entries)
 
 	def set_po_nos(self):
-		if self.doctype == 'Sales Invoice' and hasattr(self, "items"):
+		if self.doctype == 'Sales Invoice' and hasattr(self, "items") and not self.po_no:
```
nabinhait: This will break when user will create one sales invoice against multiple delivery notes associated with multiple po nos. It will only set po no for first DN.

The solution might be always append the po nos with existing and at the end, take unique pos.

nabinhait (changes requested):

erpnext/controllers/selling_controller.py
AfshanKhan: @nabinhait So here the case is like if a user enters any PO number we should not edit it. Like we should not either remove it or add something to it. If the user hasn't entered any PO number then only this will run and will bring all PO from all doctype i.e delivery notes and add it for the user.
If what you mean is whether or not the user has entered any PO we should anyway add the PO we get. Then that could be implemented. Just let me know.

AfshanKhan:
@nabinhait this can be merge now?

nabinhait (approved):

PR #24033, opened 2020-11-30
feat: Patient History Enhancements

ruchamahabal:
## Patient History Settings

Configurable Standard and Custom doctypes which will be a part of the Patient Medical Record
- **Standard**: Doctype and date field are `read_only`. These are set up with fixtures + patch and the standard fields are also setup. You can select what other fields (including custom fields) you want to see in the Patient Medical Record and thereby the Patient History.
- **Custom**: Similar configuration available for Custom doctypes.

patient-history-settings-1

## Patient Medical Record

- Hooks have been set up for creating, updating, and deleting Patient Medical Records for doctypes configured in Patient History Settings.
- Code for maintaining medical records has been removed from all Standard doctypes
- The fields are formatted using the `format_value` function in `frappe.utils.formatters`. A separate formatter has been written to take care of Table fields.

## Patient History

- Basic code clean-up. Will make it a separate class in Phase 2.
- Filters for Doctype (MultiSelectList) and DateRange have been added.
 
   filters

- Link for the individual document is now shown so that you can navigate to the reference document from the Patient History view.

patient-history

- [x] Documentation

- [x] Tests

**Documentation PR**: https://github.com/frappe/erpnext_documentation/pull/206


rohitwaghchaure:
On selection of patient, Select Document Type and Date Range field added. But after changing the patient the same fields added again.
Screenshot 2021-01-21 at 11 19 24 PM

Date fieldname should be select field with options as fields of respective custom doctype which has field type as 'Date'
Screenshot 2021-01-21 at 11 33 20 PM

ruchamahabal:
@rohitwaghchaure 

> On selection of patient, Select Document Type and Date Range field added. But after changing the patient the same fields added again.
> Screenshot 2021-01-21 at 11 19 24 PM

Fixed:

patient-history-fix

 
> Date fieldname should be select field with options as fields of respective custom doctype which has field type as 'Date'
> Screenshot 2021-01-21 at 11 33 20 PM

Actually even I thought so, but the problem is that since it is in a child table, I cannot have different options for each row. If I set options for a row it gets reset for every row.

rohitwaghchaure (approved):

PR #24034, opened 2020-11-30
fix: don't cancel job card if manufacturing entry has made

rohitwaghchaure:
Screenshot 2020-11-30 at 3 09 20 PM

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.py
+++ b/erpnext/manufacturing/doctype/job_card/job_card.py
@@ -10,6 +10,7 @@ from frappe.model.mapper import get_mapped_doc
 from frappe.model.document import Document
 
 class OperationMismatchError(frappe.ValidationError): pass
+class JobCardCancelError(frappe.ValidationError): pass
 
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -110,39 +111,54 @@ def update_work_order(self):
...
+		wo = frappe.get_doc('Work Order', self.work_order)
+		if self.operation_id:
+			self.validate_produced_quantity(for_quantity, wo)
+			self.update_work_order_data(for_quantity, time_in_mins, wo)
+
+	def validate_produced_quantity(self, for_quantity, wo):
+		if self.docstatus < 2: return
+
+		if wo.produced_qty > for_quantity:
+			first_part_msg = (_("The {0} {1} has used to calculate the valuation cost for the finished good {2}.")
```
marination:
```suggestion
			first_part_msg = (_("The {0} {1} is used to calculate the valuation cost for the finished good {2}.")
```

PR #24039, opened 2020-12-01
fix: GSTR B2C report

rohitwaghchaure:
**Issue**
Two invoices should be show in the report, but only one invoice is showing

Screenshot 2020-11-30 at 10 10 38 PM

**After Fix**
Screenshot 2020-11-30 at 10 11 32 PM


deepeshgarg007 (approved):

deepeshgarg007:
@rohitwaghchaure please front port this to develop as well

PR #24043, opened 2020-12-01
feat: pos invoice merging with background jobs

nextchamp-saqib:
When creating a POS Closing Entry, if a large number invoices are to be merged into sales invoice it can lead to request timeout. Merging will be queued into background jobs if number of invoices are greater than 10


nextchamp-saqib:
Ported along with https://github.com/frappe/erpnext/pull/24351

PR #24046, opened 2020-12-02
feat: add jinja templating in Contract Template

kennethsequeira:
Working:

contract_jinja

- adds jinja support to contract terms template
- this way contract texts can be edited better in the editor itself instead of relying solely on print formats to define text. Currently, there is no other way to define fields dynamically. 
- Also fixed fetching the fulfilment terms table from template to contract

EDIT: docs: https://github.com/frappe/erpnext_documentation/pull/215


erpnext/crm/doctype/contract_template/contract_template.py
```diff
@@ -5,6 +5,29 @@
 from __future__ import unicode_literals
 import frappe
 from frappe.model.document import Document
+from frappe.utils.jinja import validate_template
+from six import string_types
+import json
 
 class ContractTemplate(Document):
 	pass
```
marination: remove the pass

erpnext/crm/doctype/contract_template/contract_template.json
```diff
@@ -41,10 +43,20 @@
...
+  },
+  {
+   "fieldname": "section_break_6",
+   "fieldtype": "Section Break"
+  },
+  {
+   "fieldname": "contract_template_help",
+   "fieldtype": "HTML",
+   "label": "Contract Template Help",
+   "options": "<h4>Contract Template Example</h4>\n\n<pre>Contract for Customer {{ party_name }}\n\n-Valid From : {{ start_date }} \n-Valid To : {{ end_date }}\n</pre>\n\n<h4>How to get fieldnames</h4>\n\n<p>The fieldnames you can use in your email template are the fields in the document from which you are sending the email. You can find out the fields of any documents via Setup &gt; Customize Form View and selecting the document type (e.g. Contract)</p>\n\n<h4>Templating</h4>\n\n<p>Templates are compiled using the Jinja Templating Language. To learn more about Jinja, <a class=\"strong\" href=\"http://jinja.pocoo.org/docs/dev/templates/\">read this documentation.</a></p>"
```
marination:
```suggestion
   "options": "<h4>Contract Template Example</h4>\n\n<pre>Contract for Customer {{ party_name }}\n\n-Valid From : {{ start_date }} \n-Valid To : {{ end_date }}\n</pre>\n\n<h4>How to get fieldnames</h4>\n\n<p>The fieldnames you can use in your Contract Template are the fields in the Contract document for which you are creating the template. You can find out the fields of any documents via Setup &gt; Customize Form View and selecting the document type (e.g. Contract)</p>\n\n<h4>Templating</h4>\n\n<p>Templates are compiled using the Jinja Templating Language. To learn more about Jinja, <a class=\"strong\" href=\"http://jinja.pocoo.org/docs/dev/templates/\">read this documentation.</a></p>"
```

erpnext/crm/doctype/contract_template/contract_template.py
```diff
@@ -5,6 +5,29 @@
...
+		doc = json.loads(doc)
+
+	contract_template = frappe.get_doc("Contract Template", template_name)
+	contract_terms = None
+
+	if contract_template.contract_terms:
+		contract_terms = frappe.render_template(contract_template.contract_terms, doc)
+	
+	return {
+		'contract_template': contract_template,
```
marination, line 29: rather than passing the whole doc, check if fulfilment is required. If yes then return fulfilment terms as well. I could have fulfilment terms but it may not be required, why should we loop over it for nothing on the client side.

erpnext/crm/doctype/contract/contract.js
```diff
@@ -1,23 +1,27 @@
...
+				},
+				callback: function(r) {
+					if (r && r.message) {
+						frm.set_value("contract_terms", r.message.contract_terms);
+						
+						// Populate the fulfilment terms table from a contract template, if any
+						r.message.contract_template.fulfilment_terms.forEach(element => {
+							let d = frm.add_child("fulfilment_terms");
+							d.requirement = element.requirement;
+						});
```
marination: This populates the table. But the checkbox is not enabled until i save, so i cannot see the table until i hit Save. All of that should be dynamic. Maybe set the value of the checkbox here followed by the table ? fetch from doesn't seem to work here.

kennethsequeira:
@marination docs PR:

https://github.com/frappe/erpnext_documentation/pull/215

marination (approved):

PR #23838, opened 2020-11-05
fix: fiscal year can be shorter than 12 months

barredterra:
Frontport of #23497.

Additional changes:

- Replace unnecessary SQL query in `fiscal_year.get_from_and_to_date()`
- JS used to set `year_end_date` to 12 months after `year_start_date`. Disable this for short years.
- `year_start_date` and `year_end_date` used to be editable and made read_only via JS. Remove the related js code and enable `set_only_once` for these fields.

Bildschirmfoto 2020-11-05 um 12 55 59

PR #24050, opened 2020-12-03
refactor(Fiscal Year): to avoid using SQL query to get fiscal year dates

Anuja-pawar:
Changed frappe.db.sql() to frappe.db.get_value() to get 'from_date' and 'to_date' values from Fiscal Year


abhishekbalam (approved):

barredterra:
Wasn't this already fixed with https://github.com/frappe/erpnext/pull/23838/files#diff-4f3b15d8c6e2a2a1157d699974233ede1808d2d81d36e3185ed8dd60984b9d7d ?

PR #24051, opened 2020-12-03
feat(Issue): Added filter for customer field

Anuja-pawar:
Added filter for customer field to select only enabled customers.

Before:
Screenshot 2020-12-03 at 5 51 51 PM

After:
Screenshot 2020-12-03 at 5 50 04 PM

Docs PR:
https://github.com/frappe/erpnext_documentation/pull/210


erpnext/support/doctype/issue/issue.js
```diff
@@ -1,6 +1,13 @@
 frappe.ui.form.on("Issue", {
 	onload: function(frm) {
 		frm.email_field = "raised_by";
+		frm.set_query('customer', function () {
```
marination:
```suggestion
		frm.set_query("customer", function () {
```
sorry but consistency

marination (approved):

PR #24054, opened 2020-12-03
feat(Accounting Dimension): accounting dimension doctypes for other apps

gwhitney:
Prior to this change, there was simply a single hardcoded list of doctypes
  with accounting dimensions in ERPNext. However, when extending the
  capabilities of ERPNext using an add-on app rather than by modifying
  ERPNext itself, it may well be necessary to create new doctypes in other
  apps that will have accounting dimensions as well.

  So, this PR moves the list of doctypes with accounting dimensions into
  hooks.py so it can be accumulated from multiple apps. It also allows the
  list of doctypes in make_dimension_in_accounting_doctypes be specifiable,
  for more convenient use in the after_install hooks of other apps.


gwhitney:
Not sure what the Documentation Required error is/means. If there is something I need to do for documentation, please just let me know.

deepeshgarg007:
@gwhitney Please update the Pull Request and resolve conflicts

deepeshgarg007 (approved):

PR #24057, opened 2020-12-04
fix: paid amount in Sales Invoice POS return resets to 0

ruchamahabal:
**Steps to replicate:**

1. Create a Sales Invoice for POS
2. Create a Return / Credit Note against it and Save.
3. On submit, the amount in the payments table gets reset to 0 because of this code.

https://github.com/frappe/erpnext/blob/a31731e73d3faf53b319c9744ef8d9997fd865ce/erpnext/public/js/controllers/taxes_and_totals.js#L676-L680

The default flag in the record gets reset here and is not set in this function:

https://github.com/frappe/erpnext/blob/a31731e73d3faf53b319c9744ef8d9997fd865ce/erpnext/controllers/taxes_and_totals.py#L639

si-pos-invoice-before

**Fix**: Set default flag for payment method during return

si-pos-invoice-after


marination (approved):

marination:
@ruchamahabal one small test for POS Invoice Credit Note ?

ruchamahabal:
@marination There is already a test for pos returns

https://github.com/frappe/erpnext/blob/0ebace5da2dbb93928d633e739778d4e7a0b6f73/erpnext/accounts/doctype/sales_invoice/test_sales_invoice.py#L732-L763

Looks like the test was not failing because this gets reset on the client side:

https://github.com/frappe/erpnext/blob/a31731e73d3faf53b319c9744ef8d9997fd865ce/erpnext/public/js/controllers/taxes_and_totals.js#L676-L680

marination:
@ruchamahabal okay cool we can skip tests here

PR #24062, opened 2020-12-04
[WIP] fix: improve translation pattern

barredterra:
I noticed that the following snippet is flagged as invalid, which made me tune the regex a bit:

```js
__("{0} edited this {1}", [
	frappe.user.full_name(this.frm.doc.modified_by).bold(),
	"<br>" + comment_when(this.frm.doc.modified),
])
```

### Before
Bildschirmfoto 2020-12-10 um 11 52 36

### After
Bildschirmfoto 2020-12-10 um 11 50 33
Bildschirmfoto 2020-12-10 um 11 49 47

Try it yourself: regexr.com/5i7as


netchampfaris:
@barredterra Multiline translated strings are not supported. That is why they are not matched in the regex pattern.

@surajshetty3416 Am I missing something here?

surajshetty3416:
https://frappeframework.com/docs/user/en/translations#3-blocks
According to docs, multiline translatable strings are not supported probably because there are some gotchas in translating multiline strings.

We can allow variables in a different line though (if it's not supported already).

barredterra:
> We can allow variables in a different line though (if it's not supported already).

Yes, currently it's not supported. Do you know what the gotchas are with multiline strings?

barredterra:
@surajshetty3416 I changed it so that multiline strings are not allowed.

surajshetty3416 (dismissed):

PR #24065, opened 2020-12-05
fix: Added TDS Payable account number and an error message

Anuja-pawar:
One of the customers reported this issue they faced while creating a new company with the Chart of Accounts template _standard with numbers_.

Screenshot 2020-12-06 at 3 41 20 PM

It was thrown because _TDS Payable_ had no account number assigned in the standard template with numbers.
So, we added the _TDS Payable_ account number in a template _Standard Chart of Accounts with numbers_.
Also, on digging into it further, we found out this error message which appears when you don't have any active fiscal year for that specific company (or a fiscal year with no company), so we updated that error message to make it more precise.

Before:
Screenshot 2020-12-06 at 4 57 36 PM

After:
Screenshot 2020-12-07 at 11 09 18 AM


erpnext/accounts/utils.py
```diff
@@ -78,7 +78,10 @@ def get_fiscal_years(transaction_date=None, fiscal_year=None, label="Date", verb
 			else:
 				return ((fy.name, fy.year_start_date, fy.year_end_date),)
 
-	error_msg = _("""{0} {1} not in any active Fiscal Year.""").format(label, formatdate(transaction_date))
+		error_msg = _("""{0} {1} not in any active Fiscal Year""").format(label, formatdate(transaction_date))
```
marination: why is this in the for loop

erpnext/accounts/utils.py
```diff
@@ -78,7 +78,10 @@ def get_fiscal_years(transaction_date=None, fiscal_year=None, label="Date", verb
 			else:
 				return ((fy.name, fy.year_start_date, fy.year_end_date),)
 
-	error_msg = _("""{0} {1} not in any active Fiscal Year.""").format(label, formatdate(transaction_date))
+	error_msg = _("""{0} {1} not in any active Fiscal Year""").format(label, formatdate(transaction_date))
```
marination:
```suggestion
	error_msg = _("""{0} {1} is not in any active Fiscal Year""").format(label, formatdate(transaction_date))
```

erpnext/accounts/utils.py
```diff
@@ -78,7 +78,10 @@ def get_fiscal_years(transaction_date=None, fiscal_year=None, label="Date", verb
...
+	if company:
+		error_msg = _("""{0} for {1}""").format(error_msg, company)
```
marination:
```suggestion
		error_msg = _("""{0} for {1}""").format(error_msg, frappe.bold(company))
```

marination (approved):

PR #24070, opened 2020-12-06
feat: new tally migrator

nextchamp-saqib:
**Masters**
- Fetch Item Groups from tally masters
- Fetch city and state from pincode
- Fix chart of accounts import

**Transactions**
- Create opening entry from Tally Trial Balance report of previous FY
- Import all transactions of current FY 
- Fix multi-currency transactions import
---
- [x] Docs - https://github.com/frappe/erpnext_documentation/pull/214
- [ ] Tests


Anurag810 (changes requested):
Fix Sider

PR #24072, opened 2020-12-07
fix: drop ship partial order fixed

pateljannat:
**Issue:**

1. When the user creates a Sales Order with drop ship items and orders partially, they were able to create only one Purchase Order.
2. While trying to create the second Purchase Order they got a validation **Purchase Order already created for all Sales Order items** even when some items were not ordered.

Screenshot 2020-12-07 at 10 31 15 AM

**What to test:**
- While creating PO from SO, check "Against Default Supplier" in the dialog and partially create PO for a drop ship item.
- Submit PO.
- Repeat the same for remaining qty. It should smoothly allow partial POs for drop ship case.


marination:
@pateljannat add a test where a drop ship item is partially ordered using `make_purchase_order_for_default_supplier` and then the pending qty is ordered again.

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -831,7 +831,7 @@ def update_item(source, target, source_parent):
 
 	for supplier in suppliers:
 		po = frappe.get_list("Purchase Order", filters={"sales_order":source_name, "supplier":supplier, "docstatus": ("<", "2")})
-		if len(po) == 0:
+		if len(po) == 0 or any( item.get("delivered_by_supplier") == 1 for item in selected_items):
```
marination: selected_items is fetched from the table in the dialog (client side) right ? From what i know `delivered_by_supplier` is not sent in the selected items. So `item.get("delivered_by_supplier")` will always be `None`.

Anyway we know that `selected_items` here contains drop ship items only, this method is specifically for that.
Partial Drop ship wasnt considered before. 
Lets remove line 833 and 834 (if condition) altogether.
Remove `else: suppliers =[]` as well and bring all the logic outside the if else. Lets just map whatever we get to the Purchase Order.
Logic to set `qty` as pending qty is there in  `update_item`. So our mapping will be correct in case of partial.

marination (approved):

PR #24074, opened 2020-12-07
fix(Production Planning Report): incorrect required_qty

hasnain2808:
Production Planning report required_qty column displaying multiplied values.
the product requires 2 qty of raw material.
In sales order 1 the qty is 2 so 4 (2*2) reqd qty is correct.
but In sales order 2 the qty is 3 so 12 reqd qty is incorrect it should be  6 (3*2).

image

After Fix
image


rohitwaghchaure (approved):

rohitwaghchaure:
@hasnain2808 send PR on the version-12-hotfix branch

hasnain2808:
This report is not present in v12, is it?
I think it was added in v13

rohitwaghchaure:
@hasnain2808 yeah right sorry

hasnain2808:
Np @rohitwaghchaure

PR #24078, opened 2020-12-07
fix: double exception in payroll

AfshanKhan:
Steps:

- Keep salary component account and payroll payable account same. The user used to get the below error.
Screen Recording 2020-12-07 at 7 09 44 PM

The issue was that it was throwing the error while creating submitting journal entry and the catch exception block in payroll entry was catching the Exception of type Exception which was creating the issue.
Now it does not throw the 2nd error if the type is of error. 
Screen Recording 2020-12-07 at 7 14 06 PM


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24080

nabinhait (approved):

PR #24086, opened 2020-12-08
feat: temporary against account

barredterra:
In DATEV, the against account has a meaning. It tries to fit everything into one record:

| Account | Against Account | Amount
|---------|------------------|---------:
| Revenue (19% VAT) | Customer | € 119.00

VAT gets deducted and booked to different accounts automatically. Receivables will be tracked, etc. This approach comes to its limitations when there are multiple tax rates, advance payments, etc. on one invoice.

ERPNext however, books like this:

| Debit | Credit | Account
|---------:|------------------:|---------
| € 119.00 | | Debtors
| | € 19.00 | VAT 19%
| | € 100.00 | Revenue

This is much more flexible, but the against account has no meaning. We just debit or credit individual accounts.

Previously we tried to guess the against account for the DATEV Export. This was achieved by using one payable/receivable account per party, merging all those tables and coalescing through the values.

From [discussion on the forum](https://discuss.erpnext.com/t/erpnext-datev-kompatibel-machen/44127/84?u=rmeyer) (german) it turned out that a better strategy would be to use a temporary against account for the whole export. In the external system, all GL Entries will be made against this temporary account and it's balance will always be zero. The column "Belegfeld 1" contains the voucher number, tying multiple rows together into one transaction.

Pros:

- Entries are identical to ERPNext (because in ERPNext the against account has no meaning).
- Less code, less errors.
- No more dedicated accounts for parties, meaning less work in daily business.

Cons:

- Automatic processing in DATEV will not be possible, meaning more work for the tax consultant.

### DATEV Settings

datev_settings

### Full Export

full_export

### Payment Entry only

payment_entry

### Stock Entry only

stock_entry

Docs: https://github.com/frappe/erpnext_documentation/pull/221


barredterra:
@nabinhait anything I can do to get this merged?

nabinhait (approved):

PR #24087, opened 2020-12-08
feat: Batch wise item selling pricing

deepeshgarg007:
By default the latest selling item price will be added to the batch, this can be changed by the user later

Based on the item price selected in batch the rate will be applied in transactions 

Screenshot 2020-12-10 at 10 17 50 PM


mujeerhashmi:
@deepeshgarg007 
I guess it would be more appropriate to maintain the price in the batch document rather than creating the item price for every batch. Imagine this scenario in a pharmacy store. The drugs will be procured multiple times and having to create item price documents for every batch will be a huge task.

@akurungadam @ruchamahabal Your opinions/suggestions will add more value to this.

deepeshgarg007:
> @deepeshgarg007
> I guess it would be more appropriate to maintain the price in the batch document rather than creating the item price for every batch. Imagine this scenario in a pharmacy store. The drugs will be procured multiple times and having to create item price documents for every batch will be a huge task.

@mujeerhashmi yes that makes more sense. Will make the changes

rohitwaghchaure (approved):

PR #24090, opened 2020-12-09
fix: scan barcode does not update barcode item field in sales order

hasnain2808:
transaction.js adds an item to the child table when a barcode is input in the scan_barcode field.
It used a bool variable 'from_barcode' to track the status 
As soon as a barcode is added in the scan barcode field there is a race condition between the onchange triggers on the barcode field(of the newly added row) and item_code field(of the newly added row)  to update the value of 'from_barcode'

If the item_code field trigger runs before the barcode field trigger: the barcode is properly displayed
Reason: The item code_field makes the 'from_barcode' False and when the barcode field trigger runs it sets the 'from_barcode' field to true before making it false

If the barcode field trigger runs before the item_code field trigger the item_code field actually removes the barcode 
Reason: The update by barcode field trigger on  'from_barcode' is lost because item_code has not yet set it to false. Now when the item code trigger checks 'from_barcode' it sees False since the barcode field set it to false on exit

Fix: used counters to remember how many times the update was done


erpnext/public/js/controllers/transaction.js
```diff
@@ -408,7 +408,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 				show_description(row_to_modify.idx, row_to_modify.item_code);
 
-				this.frm.from_barcode = true;
+				this.frm.from_barcode = this.frm.from_barcode ? this.frm.from_barcode + 1: 1;
```
marination:
```suggestion
				this.frm.from_barcode = this.frm.from_barcode ? this.frm.from_barcode + 1 : 1;
```

erpnext/public/js/controllers/transaction.js
```diff
@@ -492,7 +492,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			d.item_code = "";
 		}
 
-		this.frm.from_barcode = true;
+		this.frm.from_barcode = this.frm.from_barcode ? this.frm.from_barcode + 1: 1;
```
marination:
```suggestion
		this.frm.from_barcode = this.frm.from_barcode ? this.frm.from_barcode + 1 : 1;
```

marination (approved):

PR #24092, opened 2020-12-09
feat: Project template with dependent tasks

pateljannat:
1. In Project, we have the functionality to create Sub-Tasks and Dependent Tasks, however the same doesn't exist while creating a Project Template.

2. Having tasks in the Project template will help users reuse the Project Template to create multiple Projects.

3. Tasks will now have a checkbox **Is Template** which will indicate that the task is meant to be used by a Project Template.

project-task

4. Such tasks can be added to Project Template Doctype.

projects-project-template

5. The Begin and Duration fields that were previously a part of the **Project Template Task** child Table, will now be in Task doctype itself.

projects-task-timeline

**Documentation PR**: https://github.com/frappe/erpnext_documentation/pull/222


erpnext/projects/doctype/project/project.py
```diff
@@ -54,17 +54,57 @@ def copy_from_template(self):
...
+				issue = task_details.issue,
+				is_group = task_details.is_group,
+				start = task_details.start,
+				duration = task_details.duration
+			)).insert()
+
+	def dependency_mapping(self, template_tasks, project_tasks):
+		for tmp_task in template_tasks:
+			prj_task = list(filter(lambda x: x.subject == tmp_task.subject, project_tasks))[0]
+			prj_task = frappe.get_doc("Task", prj_task.name)
```
deepeshgarg007: project_tasks and template_tasks already has task doc, no need to get_doc here again

erpnext/projects/doctype/project/project.py
```diff
@@ -54,17 +54,57 @@ def copy_from_template(self):
...
+			self.check_depends_on_value(tmp_task, prj_task, project_tasks)
+			self.check_for_parent_tasks(tmp_task, prj_task, project_tasks)
+
+	def check_depends_on_value(self, tmp_task, prj_task, project_tasks):
```
deepeshgarg007: Rename `tmp_task` to `template_task` and `prj_task` to `project_task`

erpnext/projects/doctype/task/task.json
```diff
@@ -360,14 +363,30 @@
...
+  {
+   "default": "0",
+   "fieldname": "is_template",
+   "fieldtype": "Check",
+   "label": "Is Template"
+  },
+  {
+   "fieldname": "start",
+   "fieldtype": "Int",
+   "label": "Begin On (Days)"
```
deepeshgarg007, line 377: Show Begin On and Duration field only for Template Tasks, you can use depends_on property for this

erpnext/projects/doctype/project_template/project_template.py
```diff
@@ -3,8 +3,27 @@
...
+	def validate(self):
+		self.validate_dependencies()
+
+	def validate_dependencies(self):
+		for task in self.tasks:
+			task_details = frappe.get_doc("Task", task.task)
+			if task_details.depends_on:
+				for dependency_task in task_details.depends_on:
+					if not self.check_dependent_task_presence(dependency_task.task):
+						task_details_format = """<a href="#Form/Task/{0}">{0}</a>""".format(task_details.name)
```
deepeshgarg007: No need for this. We already have a utility for this. Search for `get_link_to_form`

erpnext/patches/v13_0/update_project_template_tasks.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2019, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+import frappe
+
+def execute():
+    frappe.reload_doc("projects", "doctype", "project_template")
+    for template_name in frappe.db.sql(""" select name from `tabProject Template` """, as_dict=1):
```
deepeshgarg007: Format Queries, a good example

```
frappe.db.sql("""SELECT 
        item_name, description, default_warehouse 
    FROM
        tabItem 
    WHERE
        disabled = 0""") 
```

erpnext/projects/doctype/project/test_project.py
```diff
@@ -7,60 +7,127 @@
...
+		if not task1:
+			task1 = create_task(subject="Test Temp Task for dependency", is_template=1, begin=3, duration=1)
+
+		task2 = task_exists("Test Temp Task with dependency")
+		if not task2:
+			task2 = create_task(subject="Test Temp Task with dependency", depends_on=task1.name, is_template=1, begin=2, duration=2)
+		
+		template = make_project_template("Test Project with Templ - dependent tasks", [task1, task2])
+		project = get_project("Test Project with Templ - dependent tasks", template)
+		tasks = frappe.get_all('Task', '*', dict(project=project.name), order_by='creation asc')
```
deepeshgarg007: Just fetch the required fields instead of all fields

erpnext/projects/doctype/project/test_project.py
```diff
@@ -7,60 +7,127 @@
...
 
 class TestProject(unittest.TestCase):
-	def test_project_with_template(self):
-		frappe.db.sql('delete from tabTask where project = "Test Project with Template"')
-		frappe.delete_doc('Project', 'Test Project with Template')
+	def test_project_with_template_having_no_parent_and_depend_tasks(self):
+		""" 
+		Test Action: Basic Test of a Project created from template. The template has a single task.
+		"""
+		frappe.db.sql('delete from tabTask where project = "Test Project with Templ - no parent and dependend tasks"')
```
deepeshgarg007: Naming consistency with Task and Project naming. Use title case for the entire project name. `Test Project With Template - No Parent and Dependant Tasks`

deepeshgarg007 (changes requested):

deepeshgarg007 (changes requested):
- Reload Project Template Task and Task doctype in patch
- Status of Template task should be "Template"
- We can also add description field (Read Only) in Project Template Tasks so that its easier for user to identify which tasks are added in template

erpnext/patches/v13_0/update_project_template_tasks.py
```diff
@@ -0,0 +1,45 @@
...
+                new_task = frappe.get_doc(dict(
+                    doctype = "Task",
+                    subject = task.subject,
+                    start = task.start,
+                    duration = task.duration,
+                    task_weight = task.task_weight,
+                    description = task.description,
+                    is_template = 1
+                )).insert()
+                print(new_task)
```
deepeshgarg007: Remove print statements

deepeshgarg007 (changes requested):
Add indicator for Template task in project_list.js

deepeshgarg007 (approved):

PR #24101, opened 2020-12-10
fix: rebuild depends on task on validate

scmmishra:
Dependencies are stored in a code field in Task Doctype, this means that every time a task dependency was updated, the field was updated. The function `update_depends_on` did this, however the previous logic would iterate over the dependent task and just add those which are not already in the list. However, it did not manage the tasks that were removed from the list.

The simple solution to this is to rebuild the field data entirely.

### ToDo

- [x] Fix
- [x] Patch
- [x] Tests


scmmishra:
@nextchamp-saqib can you review this?

PR #22859, opened 2020-07-31
feat: Return tracking in PR/DN

marination:
Docs: https://github.com/frappe/erpnext_documentation/pull/154

## Transaction
<details><summary>Purchase Receipt</summary>
  
  - Returned Qty fields in Items Table
    Screenshot 2020-07-31 at 4 01 17 PM

  - % Returned in parent doc
   Screenshot 2020-07-31 at 4 02 02 PM

  - Status 'Return Issued' added on original PR if 100% returned. If partially returned , priority given to 'To Bill'
   Screenshot 2020-07-31 at 4 03 48 PM

</details>

<details><summary>Delivery Note</summary>

  - Returned Qty fields in Items Table
    Screenshot 2020-07-31 at 8 14 34 PM

  - % Returned in parent doc
    Screenshot 2020-07-31 at 8 15 02 PM

  - Status 'Return Issued' added on original DN if 100% returned. If partially returned , priority given to 'To Bill'
    Screenshot 2020-07-31 at 8 15 33 PM

</details>

<details><summary>Reports Changes</summary>

- Added Returned Amount to **Delivered** and **Purchased** Items to be Billed Reports.
- **Pending Amount = Total Amount - Billed Amount  - Returned Amount**. Essentially only non-returned items should be counted as pending
- If **Pending Amount** > 0, only then will the item be visible in the Report. So if a PR is 50 % Billed and 50% Returned, it won't be visible in the Report.

Screenshot 2020-09-10 at 6 48 42 PM

</details>

ToDo:
- [x]  Return for DN
- [x]  Tests for PR/DN
- [x]  Report Fixes (Received Items to be Billed, Delivered Items to be Billed)
- [x] Patch to set qty and % 
- [x]  Test to check Correct Mapping and Billing % Updation (bill after return and bill before return)

PR #24104, opened 2020-12-10
chore: Return tracking in PR/DN

marination:
Port of https://github.com/frappe/erpnext/pull/22859


erpnext/controllers/sales_and_purchase_return.py
```diff
@@ -257,20 +284,25 @@ def set_missing_values(source, target):
 		doc.run_method("calculate_taxes_and_totals")
 
 	def update_item(source_doc, target_doc, source_parent):
```
Anurag810, line 286: I think we need to split this into smaller methods/functions.

PR #24120, opened 2020-12-14
fix: adding additional salary to leave encashment

AfshanKhan:
Setting Additional Salary made while submitting Leave Encashment in the document


nabinhait (commented):
Why 2-way linking is needed?

AfshanKhan:
> Why 2-way linking is needed?

@nabinhait Doctype of Leave Encashment has a field for Additional Salary, plus on cancellation of Leave Encashment we are setting that field for Additional Salary as empty in `on_cancel()` function. The code looked incomplete, so added this linking.

PR #24124, opened 2020-12-14
fix: Payment Entry created against a Payroll Entry should only consider Salary Slips in that Payroll Entry

ruchamahabal:
**Problem**: On clicking **Make Bank Entry** in Payroll Entry, the system is calculating the amount for salary slips for the whole period instead of calculating the salary slips related to that particular payroll entry.

**Fix**: Filter Salary Slips linked to the Payroll Entry while fetching Salary Slips for creating a Payment Entry.

### DIFF
--- a/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
+++ b/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
@@ -312,8 +312,8 @@ class PayrollEntry(Document):
 
 		cond = self.get_filter_condition()
 		salary_slip_name_list = frappe.db.sql(""" select t1.name from `tabSalary Slip` t1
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

nabinhait (changes requested):
If the salary slip is submitted via Payroll Entry, it should be considered for payment entry, because accrual entry is made for this.
But currently, if a user creates salary slip manually, then submit it via payroll entry, then it is not linked to payroll entry but accrual entry is made against it. Hence, if payment entry does not made against it later, it's a problem.

Solution: If we accept the changes in PR, then we need to update payroll entry link in salary slip while submitting salary slip via PE.

Connect me on Telegram of clarification needed.

nabinhait:
Fixed via #24540

PR #24126, opened 2020-12-14
feat: Voice Call Settings doctype added

leela:
Added a new doctype `Voice Call Settings`, This manages user level voice settings including Twilio number, call receiving device etc.

Along with Twilio app `Voice Call Settings` doctype is required to make Twilio work sanely.

no-docs (Docs are part of Twilio Integration app)


Ma7m0d:
Does it support asterisk ??

PR #24130, opened 2020-12-15
fix: fix for not having fiscal year while creating new company

Anuja-pawar:
Issue: If you have multiple fiscal years with multiple companies, when you create a new company it may throw this error even if you have a fiscal year for a current date since it is applicable to a particular company.

Screenshot 2020-12-15 at 2 16 36 PM

 The fiscal year is required to get TDS rates while setting a chart of accounts for a new company. Now you can not create another fiscal year with dates overlapping with the existing fiscal year, or to create a fiscal year specific to this new company, you need to create a company first.

Solution: This fix handles this error and lets the user create a new company irrespective of that. The user will have to add TDS rates later in this case.


Anurag810 (dismissed):

Anurag810 (commented):
Travis failing

erpnext/regional/india/setup.py
```diff
@@ -629,15 +629,20 @@ def set_salary_components(docs):
...
 		accounts = [dict(company=company, account=tds_account)]
 
-	fiscal_year = get_fiscal_year(today(), company=company)[0]
-	docs = get_tds_details(accounts, fiscal_year)
+	try:
+		fiscal_year = get_fiscal_year(today(), verbose=0, company=company)[0]
+	except FiscalYearError:
+		pass
 
+	docs = get_tds_details(accounts, fiscal_year=None)
```
nabinhait: It should be like this:
```suggestion
	fiscal_year = None
	try:
		fiscal_year = get_fiscal_year(today(), verbose=0, company=company)[0]
	except FiscalYearError:
		pass

	docs = get_tds_details(accounts, fiscal_year)
```

nabinhait (changes requested):

erpnext/regional/india/setup.py
Anuja-pawar: I have declared fiscal_year = None above, removing it from the argument though, since it's unnecessary.

nabinhait (approved):

sagarvora:
This fix never worked! See https://github.com/resilient-tech/india-compliance/issues/566

Issue #24151, opened 2020-12-16
[version-12] Payroll: When a submitted Salary Slip gets cancelled, the corresponding Repayment schedule line of Loan remains paid


pm-at:
## Description of the issue
As mentioned in Subject

## Context information (for bug reports)

**Output of `bench version`**
```
erpnext 12.6.0
frappe 12.4.1
```

## Steps to reproduce the issue

1. Create a loan to be repaid from Salary and submit it
2. Run Payroll so that a line from Repayment Schedule gets paid
3. Cancel the Salary Slip which marked Repayment Schedule as Paid

### Observed result
Repayment Schedule entry remains paid even though its corresponding Salary Slip is cancelled

### Expected result
Paid flag should be removed from Repayment Schedule entry corresponding to the cancelled Salary Slip

Issue #24150, opened 2020-12-16
[version-12] Payroll: Loans with repay from salary not getting closed when all repayment schedules are paid


pm-at:
## Description of the issue
Suppose a loan is given to an employee with Repay from Salary marked. There is only one repayment schedule left to be paid. If Payroll Process is run for the Employee corresponding to the Period of Last Repayment Schedule, the Loan is not marked as closed even though the Salary Slips are Submitted. This issue is already fixed in my local environment so raising PR for it shortly.

## Context information (for bug reports)

**Output of `bench version`**
```
erpnext 12.6.0
frappe 12.4.1
```

## Steps to reproduce the issue

1. Create a loan with multiple repayment schedules and Repay from Salary marked
2. Run Payroll for all but the last period corresponding to repayment schedule so that only one installment should be left to be paid.
3. Run Payroll for period corresponding to the last pending repayment schedule

### Observed result
1. Repayment Schedule is marked as Paid
2. `Status` of Loan remains `Disbursed`

### Expected result
`Status` of Loan should become `Closed`

PR #24153, opened 2020-12-16
fix: Loan Doc update from Salary Slip #24150 #24151

pm-at:
Fixes below issues:
1. #24150 [version-12] Payroll: Loans with repay from salary not getting closed when all repayment schedules are paid
2. #24151 [version-12] Payroll: When a submitted Salary Slip gets cancelled, the corresponding Repayment schedule line of Loan remains paid


deepeshgarg007 (changes requested):
@pm-at Please remove the issue no comments. We don't use such comments anywhere in our code. Also please add unit tests as well for both the scenarios.

For more info refer this https://frappeframework.com/docs/user/en/guides/automated-testing/unit-testing

PR #24157, opened 2020-12-16
fix: Auto Repeat Import

ruchamahabal:
`get_next_schedule_date` method is now moved to the Auto-Repeat class in the framework. Removing the import for the same.


nabinhait (approved):

nabinhait:
@ruchamahabal please also fix in https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v9_0/fix_subscription_next_date.py

ruchamahabal:
Fixed in: https://github.com/frappe/erpnext/pull/24158

PR #24169, opened 2020-12-17
fix: allow addition and removal of employee in payroll Entry

Anurag810:
Allow Addition and removal of the employee during payroll. Delete and button is visible now.
Screenshot 2020-12-17 at 3 15 05 PM


abhishekbalam (dismissed):

deepeshgarg007 (changes requested):
@Anurag810 Its not this simple. Just removing read-only fields won't do. You'll have to filter the employees based on the parameters selected in Payroll Entry. Check for employees whose salary has been already precessed for the current month. Also some employee accruals are also needed to be considered

dufani1:
This is very good

dufani1:
@Anurag810  Please add filter of selecting employees based on salary currency

Anurag810:
> @Anurag810 Its not this simple. Just removing read-only fields won't do. You'll have to filter the employees based on the parameters selected in Payroll Entry. Check for employees whose salary has been already precessed for the current month. Also some employee accruals are also needed to be considered

done

rohitwaghchaure (approved):

PR #24177, opened 2020-12-18
fix: error popup for submitted doc

AfshanKhan:
We have added calculate_totals in salary_slip.js. This method also triggers on_refresh event.

When a user, with employee role access the submitted salary slip, the system throws error Insufficient permissions for Salary Slip.

Reason:
Employee role does not have a submit permission on salary slip.
calculate_total run server-side method
As salary slip is already submitted, `runserverobj` check validate submit permissions

Now if the document is submitted it doesn't call the server-side method.

### DIFF
--- a/erpnext/payroll/doctype/salary_slip/salary_slip.js
+++ b/erpnext/payroll/doctype/salary_slip/salary_slip.js
@@ -214,14 +214,16 @@ frappe.ui.form.on('Salary Slip Timesheet', {
 });
 
 var calculate_totals = function(frm) {
-	if (frm.doc.earnings || frm.doc.deductions) {
-		frappe.call({
-			method: "set_totals",
-			doc: frm.doc,
-			callback: function() {
-				frm.refresh_fields();
-			}
-		});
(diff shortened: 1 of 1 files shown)


erpnext/payroll/doctype/salary_slip/salary_slip.js
```diff
@@ -214,14 +214,16 @@ frappe.ui.form.on('Salary Slip Timesheet', {
...
+	if (frm.doc.docstatus === 0) {
+		if (frm.doc.earnings || frm.doc.deductions) {
+			frappe.call({
+				method: "set_totals",
+				doc: frm.doc,
+				callback: function() {
+					frm.refresh_fields();
+				}
+			});
+		}
```
marination, line 226: Looks fine. 
It also made me look at the refresh function. Unrelated but since this file is touched , the function `toggle_fields` in `refresh` can easily be moved inside the Doctype specification **Display Depends On**.

Anurag810:
please do suggested changes.

Anurag810 (approved):

PR #24179, opened 2020-12-18
fix: to update Sales Person's incentives On Save

Anuja-pawar:
Issue: On creating Partial Sales Invoice against Sales Order, A Sales Person's _Contribution to Net Total_ gets updated On Save but _Incentives_ remain the same. 

Basically, on any alteration to Sales Items in the Doc, the Total gets changed. So, sales person's contribution to total and incentives should get updated.

Before:
Sales Order-
Screenshot 2020-12-18 at 1 17 50 PM

Partial Sales Invoice-
Screenshot 2020-12-18 at 1 18 46 PM

After:
Partial Sales Invoice-
Screenshot 2020-12-18 at 1 22 29 PM


erpnext/controllers/selling_controller.py
```diff
@@ -153,6 +153,10 @@ def calculate_contribution(self):
 				self.base_net_total * sales_person.allocated_percentage / 100.0,
 				self.precision("allocated_amount", sales_person))
 
+			sales_person.incentives = flt(
```
nabinhait: I think commission_rate is not mandatory. So, if commission_rate is there then only it should calculate incentives. Otherwise, if the user enters incentives manually it will become zero.

nabinhait (changes requested):

PR #24182, opened 2020-12-21
fix(ux): don't set primary action if workflow is set

scmmishra:
In `refresh` for Payroll Entry, the client script would add primary action based on certain conditions. In case a workflow was set for Payroll Entry, this would be overridden by the framework, however the `refresh` in the client script would still run showing the default primary action for a second before its removed. 

### Broken

https://user-images.githubusercontent.com/18097732/102754312-cb4c6600-4392-11eb-8704-68343954b765.mov

### Fixed

https://user-images.githubusercontent.com/18097732/102754338-d69f9180-4392-11eb-8dbb-8b4047e5ff63.mov

This PR fixed this


scmmishra:

PR #24215, opened 2020-12-26
Fix (Regional - UAE): Unknown fieldnames

rtdany10:
Updated wrong field names. Please update at the earliest.
Throws error when creating important documents like PO, PI etc.


rtdany10:
Seems like it was some version mismatch error. Seems to work fine in the latest update.

Closing this PR.

PR #24225, opened 2020-12-28
feat: Batch wise item pricing

deepeshgarg007:
By default the latest selling item price will be added to the batch, this can be changed by the user later

Based on the item price selected in batch the rate will be applied in transactions 

Screenshot 2020-12-10 at 10 17 50 PM

### DIFF
--- a/erpnext/public/js/controllers/buying.js
+++ b/erpnext/public/js/controllers/buying.js
@@ -195,10 +195,6 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 		this._super(doc, cdt, cdn);
 	},
 
-	batch_no: function(doc, cdt, cdn) {
-		this._super(doc, cdt, cdn);
-	},
-
 	received_qty: function(doc, cdt, cdn) {
 		this.calculate_accepted_qty(doc, cdt, cdn)
 	},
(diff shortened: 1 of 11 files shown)


erpnext/public/js/controllers/buying.js
```diff
@@ -195,6 +195,10 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
 		this._super(doc, cdt, cdn);
 	},
 
+	batch_no: function(doc, cdt, cdn) {
+		this._super(doc, cdt, cdn);
+	},
```
rohitwaghchaure: As we are using this feature for selling side, no need to call this from the buying.js file

erpnext/public/js/controllers/transaction.js
```diff
@@ -1104,6 +1104,11 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 		}
 	},
 
+	batch_no: function(doc, cdt, cdn) {
+		let item = frappe.get_doc(cdt, cdn);
+		this.apply_price_list(item, true);
+	},
```
rohitwaghchaure: Put this code in sales_common.js file

rohitwaghchaure:
@deepeshgarg007 

Found one issue
1. Rate has not fetched correctly using Batch Selector

Concern Points
1. Currently user can change the rate after pulling it from the batch, should we allow to change the rate?
2. There is no provision to select the batch on the sales order, so there might be chances that rate will be different at the time of delivery note. In this case we should allow use to submit the delivery note even if the sales order and delivery note rate is different.

PR #24227, opened 2020-12-28
fix(pos): multiple pos fixes and additions

nextchamp-saqib:
Depends on https://github.com/frappe/frappe/pull/12176

- Added checkbox to allow/prevent rate & discount change
- Prevent user from selecting any pos profile while creating pos profile
- Fix: Cannot remove item from Draft invoices
- Fix: Sales UOM from Item master not fetching for POS Invoices


mcgun182 (commented):
nvm

PR #24273, opened 2021-01-01
feat(Bank Reconciliation): Redesign

### LINKED ISSUE #24491: bug: V12 Bank Reconciliation multiple attempts creates multiple page filters
## Description of the issue
Running `Reconcile this account` multiple times causes multiple filters
May be fixed by #24273 

1. Open the Bank Reconciliation page
2. Complete the filters
3. Run `Reconcile this account`
4. Run `Reconcile this account` again

### Observed result
Bank Account, Status and Company filters are added again

### Expected result
No additional filters

## Additional information
Caused by `setup_filter_area()` from `base_list.js` being run each time
ERPNext: v12.17.0
Frappe Framework: v12.14.0

hasnain2808:
**Redesign of Bank Reconciliation**  
Bank Statement Import
image
Bank Reconciliation Tool
image
Matching Vouchers
image
Creating New Voucher
image
Updating Bank Transaction
image

Docs Link: https://github.com/frappe/erpnext_documentation/pull/230


gwhitney:
Correct me if I am wrong, but this appears to be primarily a user-interface redesign. Is there any chance of incorporating improvements to the underlying significant issue with bank reconciliation as addressed in #23721 that I filed three months ago? The basic problem is illustrated by a transfer between two different bank accounts. Such a transfer shows up in the statements from both accounts. If you reconcile the Bank Transaction from the first account against the (single) Journal Entry for the transfer, then there is no way to reconcile the Bank Transaction from the second account, since the matching Journal Entry is already "used up". The core difficulty is that a single Journal Entry can represent multiple actions against different Bank Accounts. That's why a Bank Transaction has to be reconciled against a single Journal Entry Account row, not against an entire Journal Entry. I would be happy to work on top of this branch and incorporate the sort of changes from #23721 if it would help get this very important improvement into ERPNext more quickly. Please just let me know how I can lend my effort to getting this improved. Much obliged.

hasnain2808:
> Correct me if I am wrong, but this appears to be primarily a user-interface redesign. Is there any chance of incorporating improvements to the underlying significant issue with bank reconciliation as addressed in #23721 that I filed three months ago? The basic problem is illustrated by a transfer between two different bank accounts. Such a transfer shows up in the statements from both accounts. If you reconcile the Bank Transaction from the first account against the (single) Journal Entry for the transfer, then there is no way to reconcile the Bank Transaction from the second account, since the matching Journal Entry is already "used up". The core difficulty is that a single Journal Entry can represent multiple actions against different Bank Accounts. That's why a Bank Transaction has to be reconciled against a single Journal Entry Account row, not against an entire Journal Entry. I would be happy to work on top of this branch and incorporate the sort of changes from #23721 if it would help get this very important improvement into ERPNext more quickly. Please just let me know how I can lend my effort to getting this improved. Much obliged.


Hey,
This PR solves a lot of UX issues primarily. Although it does add better ways to create vouchers from Bank Transactions.

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
```diff
@@ -0,0 +1,430 @@
...
+			'unallocated_amount', 'reference_number',
+			 'party_type', 'party'],
+	)[0]
+
+
+@frappe.whitelist()
+def create_journal_entry_bts( bank_transaction, reference_number, reference_date, posting_date, entry_type,
+	# Create a new journal entry based on the bank transaction
+	second_account, mode_of_payment=None, party_type=None, party=None):
+	bank_transaction = frappe.get_doc("Bank Transaction", bank_transaction)
```
nameduser0: Why are you passing in arguments that are already in the bank transaction? See #24493

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
```diff
@@ -0,0 +1,430 @@
...
+
+	journal_entry_dict = {
+		"doctype": 'Journal Entry',
+		"voucher_type" : entry_type,
+		"company" : company,
+		"posting_date" : posting_date,
+		"cheque_date" : reference_date,
+		"cheque_no" : reference_number,
+		"mode_of_payment" : mode_of_payment
+	}
```
nameduser0: Why aren't you using `get_mapped_doc`? See #24493

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
```diff
@@ -0,0 +1,430 @@
...
+		paid_amount = bank_transaction.deposit
+	else:
+		paid_amount = bank_transaction.withdrawal
+
+	vouchers = json.dumps([{
+		"payment_doctype":"Journal Entry",
+		"payment_name":journal_entry.name,
+		"amount":paid_amount}])
+
+	return reconcile_vouchers(bank_transaction.name, vouchers)
```
nameduser0, line 157: Why are you returning the same bank transaction and not the journal entry?

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
```diff
@@ -0,0 +1,430 @@
...
+		"company" : company,
+		"posting_date" : posting_date,
+		"cheque_date" : reference_date,
+		"cheque_no" : reference_number,
+		"mode_of_payment" : mode_of_payment
+	}
+	journal_entry = frappe.get_doc(journal_entry_dict)
+	journal_entry.set("accounts", accounts)
+	journal_entry.insert()
+	journal_entry.submit()
```
nameduser0, line 145: Wouldn't it be better to allow this to be used without being submitted to the database for user intervention? Like through `frappe.model.open_mapped_doc` ? See #24493

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
```diff
@@ -0,0 +1,430 @@
...
+		},
+		fields=['date', 'deposit', 'withdrawal', 'currency',
+			'description', 'name', 'bank_account', 'company',
+			'unallocated_amount', 'reference_number',
+			 'party_type', 'party'],
+	)[0]
+
+
+@frappe.whitelist()
+def create_journal_entry_bts( bank_transaction, reference_number, reference_date, posting_date, entry_type,
```
nameduser0: Why is this specifically on the bank_reconciliation_tool and not the Bank Transaction doctype itself? See #24493

nameduser0 (changes requested):
Consider all the different use-cases for a function like `create_journal_entry_bts` compared to #24493.
Please allow #24492 into V12.

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
gwhitney: I just want to second casesolved-co-uk's request that if there's a way to create journal entries from bank transactions, which is a good thing, there needs to be a mode in which it only creates drafts, and does not automatically submit the journal entries.

nameduser0:
With regards the UX changes:

- I think you're assuming users will generally create documents (e.g. payment entry) to match against before a bank transaction comes in. This may not be the case. The user should be able to create a document straight away, and they may need to create multiple similar documents to the same Supplier/Customer/Employee/GL account. e.g. filter and select all phone bills across multiple months and create payments and journal entries for them all in one step, instead of individually
- It seems like you're creating brand new forms for existing documents, which takes a different UX approach to the Sales Order for example. You can create a number of different documents from a Sales Order using the action drop-down, which pops up a mapped document in the familiar format. We should be replicating existing patterns to help with user familiarity
- I don't really see why this functionality can't be done from the Bank Transaction list view

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
hasnain2808: those are updated values from the dialog

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
hasnain2808: It is not needed to use mapped doc since I am not mapping rather creating a new doc

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
hasnain2808: It is used to update the bank reconciliation data table on the frontend. Optimizes Network calls.

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
hasnain2808: You can click on `edit in full page` to edit all the fields as per need

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
hasnain2808: It uses fields passed from the dialog that is used in the Bank Reconciliation tool only

hasnain2808:
> With regards the UX changes:
> 
> * I think you're assuming users will generally create documents (e.g. payment entry) to match against before a bank transaction comes in. This may not be the case. The user should be able to create a document straight away, and they may need to create multiple similar documents to the same Supplier/Customer/Employee/GL account. e.g. filter and select all phone bills across multiple months and create payments and journal entries for them all in one step, instead of individually

This is not assumed you can either create a document and then match or create it directly

> * It seems like you're creating brand new forms for existing documents, which takes a different UX approach to the Sales Order for example. You can create a number of different documents from a Sales Order using the action drop-down, which pops up a mapped document in the familiar format. We should be replicating existing patterns to help with user familiarity

This new tool makes the Bank reconciliation process faster with no routes. In the approach, you are suggesting it is necessary to go to each bank transaction and match it with a voucher. This is not at all an upgrade over just updating the clearance dates
 
> * I don't really see why this functionality can't be done from the Bank Transaction list view
1. The tool gives you a dashboard to view the balances and feedback when you are reconciling.
2. The state is stored
1. It basically makes the bank reconciliation process much faster

erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.py
gwhitney: I am confused; once a journal entry is submitted, it's permanently in the ledger and can only be reversed, not deleted. Some folks who don't even have the permissions to submit journal entries, only create them as drafts for a supervisor to review and submit, might well use this feature. Why the insistence on immediately submitting, when it might not actually be the desired Journal Entry after all? We strive to avoid erroneous journal entries to keep from cluttering our ledger with reversals, and careful and deliberate review is a key to that. Inserting and immediately submitting seems to be an obstacle to that, and against the spirit of why the draft and submitted stages were created. It seems to me that submitting should always be a manual operation, or at least ERPNext should support that mode for users who wish to operate that way. Thanks for your consideration!

hasnain2808: When edit in full page is clicked it does not insert or submit the entry
It opens the form view of the draft which can be edited and submitted.
There is a condition before these two lines which controls the behaviour

hasnain2808:
depends on https://github.com/frappe/frappe/pull/12362

nextchamp-saqib (approved):

nameduser0:
This new tool breaks the idea of continuous bank import with the Plaid integration. It mandates that bank reconciliation be done on "statements" with an opening and closing balance. With plaid syncing bank accounts every hour, this doesn't make any sense.

It removes the original Bank Reconciliation tool which was implemented by the person who contributed the Plaid integration, so also removes the manual Plaid "Synchronize this account" button, for those not syncing every hour.

A new feature that makes no attempt to build on existing functionality, reinvents the existing wheel instead of building on it, and only takes into account one use-case so other existing businesses get a nasty shock when they upgrade.

@MichaelPinkowski @nabinhait @hasnain2808

hasnain2808:
Hi @casesolved-co-uk ,
Thanks for your valuable feedback

> This new tool breaks the idea of continuous bank import with the Plaid integration. It mandates that bank reconciliation be done on "statements" with an opening and closing balance. With plaid syncing bank accounts every hour, this doesn't make any sense.

I do not think it breaks that idea. You can add and match any transaction from the tool. If in your use case you do not want to use the balance mismatch functionality ignore the balances. You can put any value in the closing balance field. I think we can make the closing balance field non-mandatory so it is not necessary to add any value.


> It removes the original Bank Reconciliation tool which was implemented by the person who contributed the Plaid integration, so also removes the manual Plaid "Synchronize this account" button, for those not syncing every hour.

The new tool is better in terms of usability. Manually syncing the plaid transactions is available in `Plaid Settings`.
image
 
Bank Reconciliation is a more generic tool and can be used for Bank Transactions from any source.


> A new feature that makes no attempt to build on existing functionality, reinvents the existing wheel instead of building on it, and only takes into account one use-case so other existing businesses get a nasty shock when they upgrade.

The new tool does not remove any of the existing functionalities rather adds many more. We had done an extensive survey of the issues being faced by the users in the old tool and then tried solving as many as possible. Most of the changes were UI/UX. Please do tell which use cases were available in the older tool but are not available in the new tool.

nameduser0:
> We had done an extensive survey ... which use cases were available in the older tool but are not available in the new tool.

Clearly your survey did not include any Plaid users, as evidenced by:

> **_I think_** we can make the closing balance field non-mandatory so it is not necessary to add any value.

In other words you have no idea and didn't design it to support the use-case

> I started going through our older implementation and talking with mentors to understand what is expected from this project. This is the list we came up with after 2 hrs of brainstorming.

initital plan

NOTE: no mention of Plaid at all in your planning or in the entire article. Honesty is always the best policy.

[Your article](https://erpnext.com/blog/development/redesigning-bank-reconciliation)

hasnain2808:
The blog does not include everything that was done in the redesign. The blog includes only some parts. This is mentioned at the end of the blog.

We redesigned the bank reconciliation tool and the bank statement import. Plaid integration was not redesigned hence no mention of it

Moreover we did make sure that the redesign does not break the integration and it is working perfectly fine. Tested it multiple times.

You can add the plaid credentials in plaid settings. You can link new accounts and sync transactions by using the buttons.

The only thing that didn't work was that the tool didn't show the transactions without putting a non zero value in closing balance because depends-on is considering 0 as a falsy value. For this issue I have already given a workaround to put any value. The fix for it is also raised which would make sure it works with zero/ null values too.

If there are still use cases in your mind that are not available in the redesign please do tell.

PR #24276, opened 2021-01-03
fix: add item taxes at the same times as sales and purchase taxes

FHenry:
fix: add item taxes at the same times as sales and purchase taxes


deepeshgarg007:
@FHenry We are already making tax template templates for those rates. What the use case of making Item Tax templates as well? Please elaborate and reopen

FHenry:
@deepeshgarg007 In france, in domestic sells, there is no use of Sales Taxes as they are implemented into ERPNExt (v12, it was ok in v11), we use now Item Tax.
The only tax on selling goods and service is VAT. It depending of what your selling, not depending of customer :
VAT 5.5% (soft drink, books),VAT  10%(food),VAT  20% (all other product and services)
Let's take an exemple : 
When you add an item to quote/order or invoice. 
Product A : Price (excl. tax) is 100 € / VAT 20 %
When added to a domestic invoice 
Line 1 : Product A (qty 1) : Price (excl. tax) is 100 €, VAT 20% :  20 €, Price (incl. tax) 120 €
At the bottom of invoice Total Price (excl. tax) 100 %, Total VAT 20 % 20€ , Total price (incl. tax) 120 €
VAT is calculated on each line on sum up at the end into a total

This cannot be done with Sales Taxes in v12,  the item tax depends on Item Taxes not Sales Taxes.

At every installation of ERP Next V12, I have to recreate manually Item Tax

Do you want to me suggest another way to do it, like another button, and another item_tax_item.josn files that match this case ?

I'm not allow to reopen the PR, should I make a new one according your next answer ?

PR #24294, opened 2021-01-04
fix: Don't validate warehouse values between Material Request and Stock Entry

marination:
- Remove validation that checks if _**warehouse** in Stock Entry is the same as that in the Material Request that it was pulled from_.
  Screenshot 2021-01-04 at 6 51 19 PM
- The warehouse could change from a Material Request to Stock Entry often as different user roles raise Material Requests and different user roles stock up warehouses.
- Also at the time of creating MR, it is possible that a purchase user wants to issue/receive an Item but does not know from /to which warehouse. Later the stock user will select a warehouse according to stock levels. In such cases also the warehouse will change.
- Material Requests are usually subject to a lot of changes in the real world, so keeping it flexible would be better.


nabinhait (changes requested):
Fix the test case (check travis)

PR #24298, opened 2021-01-05
feat: deleting account & stock entries on deletion of transaction

nextchamp-saqib:
Added a checkbox in Accounts Settings to allow deletion of linked GL and SL entries on deletion of invoice / receipt
Screenshot 2021-01-05 at 1 11 00 PM

This is done for user who doesn't wants to lose the document id after cancelling the document. They can now cancel and delete the document to get the same document id again.

### DIFF
--- a/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
+++ b/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
@@ -21,7 +21,7 @@
   "book_asset_depreciation_entry_automatically",
   "add_taxes_from_item_tax_template",
   "automatically_fetch_payment_terms",
-  "delete_linked_entries",
+  "delete_linked_ledger_entries",
   "deferred_accounting_settings_section",
   "automatically_process_deferred_accounting_entry",
   "book_deferred_entries_based_on",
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -219,14 +220,20 @@
    "fieldtype": "Select",
    "label": "Book Deferred Entries Based On",
    "options": "Days\nMonths"
+  },
+  {
+   "default": "0",
+   "fieldname": "delete_linked_entries",
+   "fieldtype": "Check",
+   "label": "Delete Accounting and Stock Entries on deletion of Transaction"
```
nabinhait:
```suggestion
   "label": "Delete Accounting and Stock Ledger Entries on deletion of Transaction"
```

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -219,14 +220,20 @@
    "fieldtype": "Select",
    "label": "Book Deferred Entries Based On",
    "options": "Days\nMonths"
+  },
+  {
+   "default": "0",
+   "fieldname": "delete_linked_entries",
```
nabinhait:
```suggestion
   "fieldname": "delete_linked_ledger_entries",
```

nabinhait (changes requested):

nabinhait (approved):

PR #24311, opened 2021-01-05
fix: Update item_tax_template_dashboard.py

wojosc:
added missing backlink and label. 

In Item Groups I can set Item Tax Temple in Table taxes.

image

For example in Item I can also set Item Tax Temple inside a Table called taxes:
image


deepeshgarg007 (changes requested):
@wojosc can you please add screenshots of the dashboards. I fell these changes might not work properly because there is no field with name `item_tax_template` in any of these two doctypes

wojosc:
This would be the dashboard of Item Tax Template with the Item highlighted in green:
image

wojosc:
@deepeshgarg007 yes, you are right. the fields are not found in the doctypes item or item group directly but in the subtable with the according information.

deepeshgarg007 (approved):

PR #24315, opened 2021-01-06
feat: capture Rate of stock UOM in purchase

anupamvs:
**Usecase-** Consider stock UOM of Pen is NOS and Conversion UOM is Box, Currently, the system shows rate as per Conversion UOM.

**Solution-** Added a field on item level to show Rate Field also as per stock UOM.

**Calculation-** rate/conversion_factor

image

Added to Doctypes:
- Purchase Invoice
- Purchase Order
- Purchase Receipt

ToDo:

- [x] Add patch to populate field in existing doctypes

`no-docs`


erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -779,16 +781,23 @@
    "no_copy": 1,
    "print_hide": 1,
    "read_only": 1
+  },
+  {
+   "fieldname": "stock_uom_rate",
+   "fieldtype": "Currency",
+   "label": "Rate of Stock UOM",
```
marination, line 791:
```suggestion
   "label": "Rate (Stock UOM)",
```

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -40,6 +41,7 @@
   "base_amount",
   "pricing_rules",
   "is_free_item",
```
marination: move the checkbox to the bottom of the column. Alignment is messy as it is at the top

deepeshgarg007 (changes requested):
Server Side code seems to be missing. Add that as well

anupamvs:
@deepeshgarg007 @marination Done the changes.

erpnext/public/js/controllers/buying.js
```diff
@@ -213,6 +212,22 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+	},
+
+	conversion_factor: function(doc,cdt,cdn) {
+		this.calculate_stock_uom_rate(doc, cdt, cdn);
+		this._super(doc, cdt, cdn);
+	},
+
+	calculate_stock_uom_rate: function(doc, cdt, cdn) {
+		let item = frappe.get_doc(cdt, cdn);
+		item.stock_uom_rate = item.rate/item.conversion_factor;
```
marination:
```suggestion
		item.stock_uom_rate = flt(item.rate) / (flt(item.conversion_factor) || 1);
```
Better to be cautious

erpnext/controllers/buying_controller.py
```diff
@@ -491,6 +492,11 @@ def set_qty_as_per_stock_uom(self):
 					# Set Received Qty in Stock UOM
 					d.received_stock_qty = flt(d.received_qty) * flt(d.conversion_factor, d.precision("conversion_factor"))
 
+	def set_rate_of_stock_uom(self):
+		if self.doctype in ["Purchase Receipt", "Purchase Invoice", "Purchase Order"]:
+			for d in self.get("items"):
+				d.stock_uom_rate = d.rate / d.conversion_factor
```
marination: So the field `stock_uom_rate` is takes company currency (INR). `rate` field takes the currency in the transaction , lets say USD.
You are then dividing price in USD by conversion factor and displaying it in INR
Screenshot 2021-01-19 at 8 49 36 PM
`Rate (Stock UOM)` should have been **1 USD or 73.46 INR**
1 USD would be better, lets show rate in stock uom in transaction currency

erpnext/buying/doctype/purchase_order_item/purchase_order_item.json
```diff
@@ -726,13 +727,19 @@
    "fieldname": "more_info_section_break",
    "fieldtype": "Section Break",
    "label": "More Information"
+  },
+  {
+   "fieldname": "stock_uom_rate",
+   "fieldtype": "Currency",
+   "label": "Rate (Stock UOM)",
+   "read_only": 1
```
marination: should this be visible if uom = stock uom. Then it will just show the same value. Hide if both uom fields are the same maybe ? (display depends on)

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -779,16 +781,23 @@
    "no_copy": 1,
    "print_hide": 1,
    "read_only": 1
+  },
+  {
+   "fieldname": "stock_uom_rate",
+   "fieldtype": "Currency",
+   "label": "Rate (Stock UOM)",
+   "read_only": 1
```
marination: Hide if both uom fields are the same (display depends on)

erpnext/stock/doctype/purchase_receipt_item/purchase_receipt_item.json
```diff
@@ -861,12 +862,18 @@
    "fieldtype": "Float",
    "label": "Received Qty in Stock UOM",
    "print_hide": 1
+  },
+  {
+   "fieldname": "stock_uom_rate",
+   "fieldtype": "Currency",
+   "label": "Rate (Stock UOM)",
+   "read_only": 1
```
marination: Hide if both uom fields are the same (display depends on)

erpnext/public/js/controllers/buying.js
```diff
@@ -213,6 +212,22 @@ erpnext.buying.BuyingController = erpnext.TransactionController.extend({
...
+		refresh_field("stock_uom_rate", item.name, item.parentfield);
```
marination: Should change on changing UOM too. Does not happen correctly as seen in Purchase Order:
1200/15 = 80, not 5.33. It seems to show 80/15, which is not the right rate.
Screenshot 2021-01-19 at 9 11 12 PM

marination (changes requested):
Also via **Update Items** in Purchase Order, it does not get updated on changing rate (check for uom change and conversion factor change too). Please check that as well.

marination:
@anupamvs Now that we have discussed to uniformly add it on sales side as well, you can try to move the js controller logic in transaction controller itself, instead of buying controller.
Also add a patch to set this field in existing records.

nabinhait (approved):

PR #24318, opened 2021-01-07
fix: validating party currency with doc currency

AfshanKhan:
In validate_currency: Consider party_account_currency = company_currency then, self.currency can be anything the validation error is not thrown. But ideally, it should throw an error if self.currency is not equal to party_account_currency.

In set_payment_schedule and validate_payment_schedule_amount: 
With current condition:
Screenshot 2021-01-07 at 11 39 55 AM

After fix:
Screenshot 2021-01-07 at 11 40 34 AM


nextchamp-saqib (dismissed):

nextchamp-saqib:
@AfshanKhan  Travis failing due to these changes

erpnext/controllers/accounts_controller.py
```diff
@@ -829,8 +829,8 @@ def validate_currency(self):
 				party_account_currency = get_party_account_currency(party_type, party, self.company)
 
 				if (party_account_currency
-						and party_account_currency != self.company_currency
-						and self.currency != party_account_currency):
+						and (self.currency != party_account_currency
+						or self.currency != self.company_currency)):
```
deepeshgarg007: @AfshanKhan I don' think this is correct. A company having Default currency as INR can make invoices in UDS. This will block those. Let's discuss the issue again and reopen this.

deepeshgarg007 (changes requested):

erpnext/controllers/accounts_controller.py
```diff
@@ -829,10 +829,10 @@ def validate_currency(self):
 				party_account_currency = get_party_account_currency(party_type, party, self.company)
 
 				if (party_account_currency
-						and party_account_currency != self.company_currency
-						and self.currency != party_account_currency):
-					frappe.throw(_("Accounting Entry for {0}: {1} can only be made in currency: {2}")
-								 .format(party_type, party, party_account_currency), InvalidCurrency)
+						and (self.currency != party_account_currency
+						and self.currency != self.company_currency)):
+					frappe.throw(_("Accounting Entry for {0}: {1} can only be made in currency: {2} or {3}")
```
deepeshgarg007: The earlier message was correct, a party cannot use two currencies at a time, this is only possible for the first transaction.

erpnext/controllers/accounts_controller.py
```diff
@@ -898,7 +898,7 @@ def set_payment_schedule(self):
 		date = self.get("due_date")
 		due_date = date or posting_date
 
-		if party_account_currency == self.company_currency:
+		if party_account_currency == self.company_currency and party_account_currency == self.currency:
```
deepeshgarg007:
```suggestion
		if self.currency == self.company_currency:
```

erpnext/controllers/accounts_controller.py
```diff
@@ -959,7 +959,7 @@ def validate_payment_schedule_amount(self):
 			for d in self.get("payment_schedule"):
 				total += flt(d.payment_amount)
 
-			if party_account_currency == self.company_currency:
+			if party_account_currency == self.company_currency and party_account_currency == self.currency:
```
deepeshgarg007:
```suggestion
			if self.currency == self.company_currency:
```

deepeshgarg007 (changes requested):

PR #24319, opened 2021-01-07
fix: calculation of remaining_sub_periods if relieving date before mo…

AfshanKhan:
photo_2021-01-01_14-59-25

The above error would occur when suppose the employee was relieved in December but we are trying to submit the document in later months let's say January.

This would occur because as the employee was relieved in December the end date is considered that but as we are trying to save or submit the document in January the start date is considered to be of January and hence the difference is 0

Now we decrease the start date to the month they were relieved.


nabinhait (approved):

nabinhait:
@AfshanKhan Should be fixed in version-12?

PR #24320, opened 2021-01-07
fix: validation for ldc against supplier and section code

pateljannat:
1. Validation message added that will prevent the creation of more than one Lower Deduction Certificate against a particular supplier and section code.

Screenshot 2021-01-07 at 12 43 02 PM


erpnext/regional/doctype/lower_deduction_certificate/lower_deduction_certificate.py
```diff
@@ -24,3 +28,10 @@ def validate(self):
 			<= fiscal_year.year_end_date):
 			frappe.throw(_("Valid Upto date not in Fiscal Year {0}").format(frappe.bold(self.fiscal_year)))
 
+	def validate_supplier_against_section_code(self):
```
deepeshgarg007, line 31: Also check for dates in this. There can be two LDC certificate for a supplier and section code for non overlapping dates

erpnext/regional/doctype/lower_deduction_certificate/lower_deduction_certificate.py
```diff
@@ -24,3 +28,20 @@ def validate(self):
...
+		duplicate_certificate = frappe.db.get_value('Lower Deduction Certificate', {'supplier': self.supplier, 'section_code': self.section_code}, ['name', 'valid_from', 'valid_upto'], as_dict=True)
+		if duplicate_certificate and self.are_dates_overlapping(duplicate_certificate):
+			certificate_link = get_link_to_form('Lower Deduction Certificate', duplicate_certificate.name)
+			frappe.throw(_("There is already a valid Lower Deduction Certificate {0} for Supplier {1} against Section Code {2} for this time period.")
+				.format(certificate_link, frappe.bold(self.supplier), frappe.bold(self.section_code)))
+
+	def are_dates_overlapping(self,duplicate_certificate):
+		valid_from = duplicate_certificate.valid_from
+		valid_upto = duplicate_certificate.valid_upto
+		if valid_from <= getdate(self.valid_from) and getdate(self.valid_from) <= valid_upto:
```
deepeshgarg007:
```suggestion
		if valid_from <= getdate(self.valid_from) <= valid_upto:
```

erpnext/regional/doctype/lower_deduction_certificate/lower_deduction_certificate.py
```diff
@@ -24,3 +28,20 @@ def validate(self):
...
+			return True
+		elif valid_from <= getdate(self.valid_upto) and getdate(self.valid_upto) <= valid_upto:
```
deepeshgarg007:
```suggestion
		elif valid_from <= getdate(self.valid_upto) <= valid_upto:
```

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #24324, opened 2021-01-07
fix: Company Wise Valuation Rate for RM in BOM

marination:
**<h2>Issue:</h2>**
- When **Rate Of Materials Based On** is selected as **Valuation Rate** in BOM, it picks the weighted average valuation rate for a particular raw material entered.
- This valuation rate was calculated from among **all** Bins, irrespective of the company, which was incorrect.
- Here the first four rows are Bins that belong to **company A**. The last belongs to **company B**.
Screenshot 2021-01-07 at 3 29 51 PM
- The weighted average valuation rate here will be calculated as such: `(sum of stock values) / (sum of actual qty)`
- Since company wasn't considered the valuation picked would be **14500 (sum of ALL stock values) / 144 = 100.69**
 Screenshot 2021-01-07 at 3 35 42 PM

**<h2>Fix:</h2>**
- Pass company as a criteria while filtering Bins to calculate valuation rate from in a BOM.
- For company A, the valuation rate must be 100 **(14300 / 143 = 100)**
- For company B, the valuation rate must be 200 **(200 / 1 = 200)**
  bom-rm-rate-company

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -411,7 +411,7 @@ cur_frm.cscript.hour_rate = function(doc) {
 
 cur_frm.cscript.time_in_mins = cur_frm.cscript.hour_rate;
 
-cur_frm.cscript.bom_no	= function(doc, cdt, cdn) {
+cur_frm.cscript.bom_no = function(doc, cdt, cdn) {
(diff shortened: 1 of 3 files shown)


marination:
For the reviewer: Should it also be warehouse specific if a source warehouse is mentioned in BOM ?

rohitwaghchaure (approved):

PR #24325, opened 2021-01-07
feat: Voice Call Settings doctype added (#24126)

leela:
(cherry picked from commit 29778e2fba4b1f073fdfc048f784f755c57a1eeb)

Added a new doctype Voice Call Settings, This manages user level voice settings including Twilio number, call receiving device etc.

Along with Twilio app, Voice Call Settings doctype is required to make Twilio work sanely.

no-docs (Docs are part of Twilio Integration app)


prasadarr:
@leela  "no-docs (Docs are part of Twilio Integration app)" ??
How are users supposed to know what this feature even is and how to set it up?!! There are no docs in the feature, regardless, should be a part of docs.erpnext.com. The "Voice Call Settings" doesn't even have breadcrumbs to navigate to. You're a senior member, come on!

@nabinhait

PR #24332, opened 2021-01-08
fix: payment entry multi-currency issue

rohitwaghchaure:
Steps to replicate issue

1. Company currency set as INR
2. Make supplier and set default payable account in the USD
3. Make Purchase invoice against the above supplier with currency as USD
4. Make payment entry against the purchase invoice and set the mode of payment for the USD currency
5. System will change the Paid Amount from INR to USD but amount will remain same
6. Due to this extra unallocated amount set in the payment entry which is wrong.

**After Fix**
System set the Paid Amount in the USD and not same as INR


Anurag810 (changes requested):
Translation Failing

PR #24339, opened 2021-01-11
fix: to use Stock UoM in BOM Stock Report

Anuja-pawar:
Issue: 
On checking _show exploded view_, BOM Stock Report shows BOM Qty in Stock UoM instead of Item UoM, and not having a UoM column in the report creates a confusion as to why Qty suddenly changes. Also, considering Item UoM causes wrong results in _Enough Parts to Build_ calculation, as it is checked against Stock Qty (Stock UoM).

Solution: 
Added UoM column and modified UoM to Stock UoM irrespective of the status of _show exploded view_ checkbox.


nabinhait (approved):

nabinhait:
@Anuja-pawar Should be fixed in version-12 as well?

Anuja-pawar:
> @Anuja-pawar Should be fixed in version-12 as well?

Yes, will backport this

PR #24348, opened 2021-01-12
fix: check for tax_rate

pateljannat:
**Issue:**

1. Create a Sales Invoice with multiple rows in Taxes and Charges, make sure to have a row with rate set and one with rate as 0.

Screenshot 2021-01-12 at 1 53 30 PM

2. GSTR-1 report shows double entries for such invoices.
2. Because of this, the total value of the report was also getting affected.
3. The report was considering tax rows with the rate 0 as well.

Screenshot 2021-01-12 at 1 50 01 PM


nabinhait (approved):

nabinhait:
@pateljannat Send another PR to develop branch

pateljannat:
@nabinhait #24376

PR #24356, opened 2021-01-13
fix: Subscription prepaid date validation

pateljannat:
Issue:

1. Subscription Invoice got generated on the wrong date.
2. The subscription start date is 15-01-2021.

image

3. The invoice got created before 15-01-2021

image

4. The subscription had Generate Invoice At Beginning Of Period checked, this didn't behave as expected.


nabinhait (approved):

nabinhait:
@pateljannat Should it be fixed in version-12 as well?

pateljannat:
@nabinhait  #24373

PR #24369, opened 2021-01-14
fix: calculation of remaining_sub_periods if relieving date before month start date

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/24319/commits/56540d4a2f8473a0048ff043a5ff100d7c587afc


Anurag810 (commented):
Approved

PR #24372, opened 2021-01-14
fix: Ignore group cost center validation for period closing voucher

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/24371

### DIFF
--- a/erpnext/accounts/doctype/gl_entry/gl_entry.py
+++ b/erpnext/accounts/doctype/gl_entry/gl_entry.py
@@ -138,9 +138,10 @@ class GLEntry(Document):
 			frappe.throw(_("{0} {1}: Cost Center {2} does not belong to Company {3}")
 				.format(self.voucher_type, self.voucher_no, self.cost_center, self.company))
 
-		if not self.flags.from_repost and self.cost_center and _check_is_group():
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/gl_entry/gl_entry.py
```diff
@@ -138,9 +138,10 @@ def _check_is_group():
 			frappe.throw(_("{0} {1}: Cost Center {2} does not belong to Company {3}")
 				.format(self.voucher_type, self.voucher_no, self.cost_center, self.company))
 
-		if not self.flags.from_repost and self.cost_center and _check_is_group():
-			frappe.throw(_("""{0} {1}: Cost Center {2} is a group cost center and group cost centers cannot
-				be used in transactions""").format(self.voucher_type, self.voucher_no, frappe.bold(self.cost_center)))
+		if not self.flags.from_repost and not self.flags.from_period_closing_voucher \
```
nabinhait, line 141: Why do we need to use this flag everywhere? We can check if `self.voucher_type=='Period Closing Voucher'`

nabinhait (changes requested):

PR #24392, opened 2021-01-18
feat: re-linking bank accounts with plaid

nextchamp-saqib:
BackPort of https://github.com/frappe/erpnext/pull/23913


erpnext/accounts/doctype/bank/bank.js
```diff
@@ -21,6 +22,12 @@ frappe.ui.form.on('Bank', {
 			frappe.contacts.render_address_and_contact(frm);
 		}
 	},
```
nextchamp-saqib:
```suggestion
```

PR #24399, opened 2021-01-18
fix: Item wise purchase registry item_name error

pateljannat:
**Issue:**

Item-wise Purchase Register Report throws an error for invoices not having an item_code

**Steps to replicate:**

1. Create a Purchase Invoice without selecting an item by directly entering the item name.

Screenshot 2021-01-18 at 3 30 48 PM

2. Make sure the item has a stock quantity. Save the Invoice.

3. Go to Item-wise Purchase Register Report. Filter by date. The report throws an error.

Screenshot 2021-01-18 at 3 35 13 PM


erpnext/accounts/report/item_wise_purchase_register/item_wise_purchase_register.py
```diff
@@ -53,8 +53,8 @@ def _execute(filters=None, additional_table_columns=None, additional_query_colum
 
 		row = {
 			'item_code': d.item_code,
-			'item_name': item_record.item_name,
-			'item_group': item_record.item_group,
+			'item_name': item_record.item_name if item_record else d.item_name,
+			'item_group': item_record.item_group if item_record else d.item_group,
```
marination, line 57: @pateljannat where do such cases occur where the user may not enter Item Code? (curious)
This field (item code in Purchase Invoice) is not kept mandatory because it could be blank only in the case if it was created from **Opening Invoice Creation Tool**.
In which case you will see **Is Opening** set to yes in the Invoice:
Screenshot 2021-01-20 at 5 16 44 PM

The fix looks alright, but should a case like this occur manually at all ? And opening invoices aren't item purchases so they shouldn't even be fetched in this report, right ?

pateljannat: @marination I encountered such a scenario through a support issue. The customer had 3 such invoices where there was no item code. Also, they had **Is Opening** set as No. The item names mentioned were Labour outsource or Freight Charges. So I think the customer is using this for scenarios wherein its not actually an item but rather a service. 

As you mentioned that this field was not kept mandatory because it could be blank only in the case if it was created from Opening Invoice Creation Tool, then maybe there should be a validation that triggers at the submission of the invoice itself? I had discussed this issue with Rohit before making the fix and he mentioned that creation of invoices without item code is allowed. Please let me know your thoughts.

deepeshgarg007:
@marination @pateljannat This is not just limited to opening invoices. Even for petty expenses like stationery users just use the item name to book expenses and don't create a separate item master. Seems like a valid scenario to me.

marination (approved):

PR #24402, opened 2021-01-18
fix: validate tax template for tax category

AfshanKhan:
Only one template should be available in "Purchase Taxes And Charges Templates" and "Sales Taxes And Charges Templates" for each Tax Category.

Steps for testing:
1. Create a Tax Category
2. Now make either a "Purchase Taxes And Charges Template" or a "Sales Taxes And Charges Template" and select the created Tax Category and save it.
3. Now if you again try to create a "Purchase Taxes And Charges Template" if you had created that in step 2 or a "Sales Taxes And Charges Template" if you had created that in step 2 and select the same Tax Category then "A template with tax category {said tax category} already exists. Only one template is allowed with each tax category" is displayed to the user.


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24403

nabinhait (approved):

PR #24412, opened 2021-01-19
Fix einvoice: UX issues

KaviyaPeriyasamy:
- Prevent the generation of IRN for un-saved local changes invoice

einvoice

- Cancel EWB URL changed                  
-  Added QR image to the file creation itself. Fixed the issue of overlapping QR images between invoices.

### DIFF
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -17,19 +17,26 @@ erpnext.setup_einvoice_actions = (doctype) => {
 			};
 
 			if (!irn && !__unsaved) {
-				const action = () => {
-					frappe.call({
-						method: 'erpnext.regional.india.e_invoice.utils.get_einvoice',
-						args: { doctype, docname: name },
-						freeze: true,
-						callback: (res) => {
-							const einvoice = res.message;
-							show_einvoice_preview(frm, einvoice);
-						}
-					});
-				};
+				frappe.db.get_single_value("E Invoice Settings", "generate_irn_only_on_submission").then(val => {
+					if ((val === 1  && frm.doc.docstatus == 1) || (!val)){
+						const action = () => {
+							if(frm.doc.__unsaved) {
+								frappe.throw(__('Please save the document to generate IRN.'));
+							}
+							frappe.call({
+								method: 'erpnext.regional.india.e_invoice.utils.get_einvoice',
+								args: { doctype, docname: name },
+								freeze: true,
+								callback: (res) => {
+									const einvoice = res.message;
+									show_einvoice_preview(frm, einvoice);
+								}
+							});
+						};
 
-				add_custom_button(__("Generate IRN"), action);
+						add_custom_button(__("Generate IRN"), action);
+					}
+				});
 			}
 
 			if (irn && !irn_cancelled && !ewaybill) {
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -17,26 +17,22 @@ erpnext.setup_einvoice_actions = (doctype) => {
 			};
 
 			if (!irn && !__unsaved) {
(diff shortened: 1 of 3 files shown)


erpnext/regional/india/e_invoice/utils.py
```diff
@@ -518,9 +519,13 @@ def fetch_gstin_details(self, gstin):
 			self.raise_error()
 
 		except Exception:
-			self.log_error()
-			self.raise_error(True)
-	
+			if '401' in str(sys.exc_info()[1]):
```
karthikeyan5: match '401 Client Error' string. just 401 might match other errors.

erpnext/regional/india/e_invoice/einvoice.js
```diff
@@ -17,19 +17,26 @@ erpnext.setup_einvoice_actions = (doctype) => {
...
-						method: 'erpnext.regional.india.e_invoice.utils.get_einvoice',
-						args: { doctype, docname: name },
-						freeze: true,
-						callback: (res) => {
-							const einvoice = res.message;
-							show_einvoice_preview(frm, einvoice);
-						}
-					});
-				};
+				frappe.db.get_single_value("E Invoice Settings", "generate_irn_only_on_submission").then(val => {
```
karthikeyan5: fetch this along with `'einvoicing_enabled'`

karthikeyan5 (changes requested):
- [ ] also, add the fix for QR image file

nextchamp-saqib:
> * Add generate IRN only on bill submission checkbox

If a sales invoice is submitted and then while generating IRN, the portal finds some validation errors, then the user will have to cancel and amend the document in v13 to edit it again. There's no option to delete for now.
This was the reason to not make it mandatory after submission.

If there could have been an endpoint to validate e-invoice JSON then this could have been solved.

---

> * Handle unauthorized access issue due to token expired before the expiry time

Auto fetching of the token is already handled ([line:451](https://github.com/frappe/erpnext/pull/24412/files#diff-3472506ff6985dc1b4dd97365eecb6d3c13cbc6d4b91416b621bb806f8642083L451)) but was not working as there was a permission issue while saving E-Invoice Settings in [line:489](https://github.com/frappe/erpnext/pull/24412/files#diff-3472506ff6985dc1b4dd97365eecb6d3c13cbc6d4b91416b621bb806f8642083L489)

However we might need to add `self.e_invoice_settings.reload()` after `save` to get the latest `auth_token`

KaviyaPeriyasamy:
> * [X]  also, add the fix for QR image file
Fixed via this commit https://github.com/frappe/erpnext/pull/24412/commits/45cdc88e626add6567415c6c1592bb8c7647b39a . Please check

erpnext/regional/india/e_invoice/utils.py
KaviyaPeriyasamy: Replaced via this commit https://github.com/frappe/erpnext/pull/24412/commits/2f1782019773e4ab7206b21ff002bc999b926cfc

erpnext/regional/india/e_invoice/einvoice.js
KaviyaPeriyasamy: Feature removed due to some validation issues

KaviyaPeriyasamy:
> > * Add generate IRN only on bill submission checkbox
> 
> If a sales invoice is submitted and then while generating IRN, the portal finds some validation errors, then the user will have to cancel and amend the document in v13 to edit it again. There's no option to delete for now.
> This was the reason to not make it mandatory after submission.
> 
> If there could have been an endpoint to validate e-invoice JSON then this could have been solved.

This is an acceptable point. Removing ```generate_irn_only_on_submission``` feature via https://github.com/frappe/erpnext/pull/24412/commits/978dc0f00fa5eaff714e21b5b27952952102b0c5

KaviyaPeriyasamy:
@nextchamp-saqib  Removed the fixes of 401 - unauthorized access issue.  Can you review again this PR?

karthikeyan5 (approved):

nextchamp-saqib (approved):

PR #24413, opened 2021-01-19
fix: cant update address in sales order

Anurag810:
not able to change the address. on update it reset sales charges and taxes. 

image

### DIFF
--- a/erpnext/regional/india/taxes.js
+++ b/erpnext/regional/india/taxes.js
@@ -38,8 +38,10 @@ erpnext.setup_auto_gst_taxation = (doctype) => {
 					company: frm.doc.company
 				},
 				callback: function(r) {
-					if(r.message) {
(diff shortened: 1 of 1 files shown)


erpnext/regional/india/taxes.js
```diff
@@ -38,8 +38,10 @@ erpnext.setup_auto_gst_taxation = (doctype) => {
 					company: frm.doc.company
 				},
 				callback: function(r) {
-					if(r.message) {
-						frm.set_value('taxes_and_charges', r.message.taxes_and_charges);
+					if (r.message) {
```
deepeshgarg007, line 41: Just check the docstatus. Don't make the call itself if its a submitted doc

deepeshgarg007:
@Anurag810 closing this, Reopen after updating

PR #24414, opened 2021-01-19
fix: use supplied year for IRS 1099 forms

Alchez:
IRS 1099 form downloads would hardcode the year to 2018. This PR makes the forms pull the fiscal year from the filters instead.


agritheory:
Given that these forms are supposed to be sent out by 2/1 I would recommend that this PR be backported into v-12 (hotfix?) as well. US regional module did not exist in version 11.

Alchez:
Backported the changes to [v12-hotfix](https://github.com/frappe/erpnext/pull/24425).

deepeshgarg007 (approved):

Issue #24411, opened 2021-01-19
Gross pay and Total deductions displayed on the salary slip include components with do not include in total ticked - v13.0.0-beta.9 (version-13-beta)


muchai:
This subsequently affects net pay.

This is what is displayed on the salary slip. Because the figures are ok on the salary register report.

https://github.com/frappe/erpnext/pull/23519

aakvatech:
```
    def set_totals(self):
        self.gross_pay = 0
        if self.salary_slip_based_on_timesheet == 1:
            self.calculate_total_for_salary_slip_based_on_timesheet()
        else:
            self.total_deduction = 0
            if self.earnings:
                for earning in self.earnings:
                    if not earning.do_not_include_in_total:
                        self.gross_pay += flt(earning.amount)
            if self.deductions:
                for deduction in self.deductions:
                    if not deduction.do_not_include_in_total:
                        self.total_deduction += flt(deduction.amount)
            self.net_pay = flt(
                self.gross_pay) - flt(self.total_deduction) - flt(self.total_loan_repayment)
        self.set_base_totals()
```

This resolves it. I think the check about do not include in total was skipped.

aakvatech:
https://github.com/frappe/erpnext/blob/6fb66f94289a3a627bb6689418c8512a319101ea/erpnext/payroll/doctype/salary_slip/salary_slip.py#L1100

muchai:
@AfshanKhan This problem arises again in the latest v13.0.0-beta.14 release.

muchai:
It was fixed in v13.0.0-beta.13

muchai:
A slightly different problem, it doesn't set do_not_include_in_total

PR #24415, opened 2021-01-19
fix: select sal comp while making sal struct

AfshanKhan:
While making salary structure salary components didn't use to fetch correctly. Changed the query for the same.

### DIFF
--- a/erpnext/payroll/doctype/salary_structure/salary_structure.js
+++ b/erpnext/payroll/doctype/salary_structure/salary_structure.js
@@ -70,6 +70,9 @@ frappe.ui.form.on('Salary Structure', {
 		});
 	},
 
+	company: function(frm) {
+		frm.trigger('set_earning_deduction_component');
+	},
 
 	currency: function(frm) {
 		calculate_totals(frm.doc);
@@ -117,6 +120,7 @@ frappe.ui.form.on('Salary Structure', {
 		fields_read_only.forEach(function(field) {
 			frappe.meta.get_docfield("Salary Detail", field, frm.doc.name).read_only = 1;
 		});
+		frm.trigger('set_earning_deduction_component');
 	},
 
 	assign_to_employees:function (frm) {
--- a/erpnext/payroll/doctype/salary_structure/salary_structure.py
+++ b/erpnext/payroll/doctype/salary_structure/salary_structure.py
@@ -216,7 +216,7 @@ def get_earning_deduction_components(doctype, txt, searchfield, start, page_len,
 	return frappe.db.sql("""
 		select t1.salary_component
 		from `tabSalary Component` t1, `tabSalary Component Account` t2
-		where t1.salary_component = t2.parent
+		where t1.name = t2.parent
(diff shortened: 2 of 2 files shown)


Anurag810 (approved):

muchai:
Can I chip on this please. Not sure why the filter for salary component on salary structure, filters for company. However, I see a small problem with this.

On Salary component, the accounts section is only visible, based on this condition: eval:doc.statistical_component != 1

It means, if you want to use statistical salary components on Salary Structure, it's won't be possible to select them.

I could be missing more information.

AfshanKhan:
Hi @muchai yeah, you are right. This will solve that issue https://github.com/frappe/erpnext/pull/24424. Thanks for your input.

muchai:
Hi @AfshanKhan seen the changes, tested but still one cannot select statistical components. I think the filter should return for company, or all if company not set on the salary component.

AfshanKhan:
> Hi @AfshanKhan seen the changes, tested but still one cannot select statistical components. I think the filter should return for company, or all if company not set on the salary component.

Hi @muchai this is intentional. The company should be added to the component. Just as a precaution. And this is something a user can add if not present.

muchai:
> > Hi @AfshanKhan seen the changes, tested but still one cannot select statistical components. I think the filter should return for company, or all if company not set on the salary component.
> 
> Hi @muchai this is intentional. The company should be added to the component. Just as a precaution. And this is something a user can add if not present.

Ok. Then the query doesn't still return statistical components. Even the most recent update.

AfshanKhan:
> > > Hi @AfshanKhan seen the changes, tested but still one cannot select statistical components. I think the filter should return for company, or all if company not set on the salary component.
> > 
> > 
> > Hi @muchai this is intentional. The company should be added to the component. Just as a precaution. And this is something a user can add if not present.
> 
> Ok. Then the query doesn't still return statistical components. Even the most recent update.

Check this one https://github.com/frappe/erpnext/pull/24430 If still your statistical components are not returned then maybe they aren't configured correctly

muchai:
> > > > Hi @AfshanKhan seen the changes, tested but still one cannot select statistical components. I think the filter should return for company, or all if company not set on the salary component.
> > > 
> > > 
> > > Hi @muchai this is intentional. The company should be added to the component. Just as a precaution. And this is something a user can add if not present.
> > 
> > 
> > Ok. Then the query doesn't still return statistical components. Even the most recent update.
> 
> Check this one #24430 If still your statistical components are not returned then maybe they aren't configured correctly

Had tried that one. Will check the configurations.

muchai:
Please check on this too: https://github.com/frappe/erpnext/issues/24411

Ponyooooo:
is it correct that this is still not solved? the company filter on the salary structure is the cause of this since when a salary component is a statistical component the user should not provide or encode the default account which ofcourse you will also not need to encode the company. I think the statistical component feature is not doing its job to be added to the salary structure and will not be encoded to a journal entry.

PR #24425, opened 2021-01-20
fix: use supplied year for IRS 1099 forms

Alchez:
Backport of #24414.


deepeshgarg007 (dismissed):

deepeshgarg007:
@Alchez Travis seems to be failing for python 2.7. Can you fix that?

Alchez:
> @Alchez Travis seems to be failing for python 2.7. Can you fix that?

Oh, that's from the f-strings in Python 3.6+

I thought Frappe had dropped Python 2 support, is that not the case anymore?

@deepeshgarg007

AfshanKhan:
> > @Alchez Travis seems to be failing for python 2.7. Can you fix that?
> 
> Oh, that's from the f-strings in Python 3.6+
> 
> I thought Frappe had dropped Python 2 support, is that not the case anymore?
> 
> @deepeshgarg007

@Alchez No, Frappe still supports Python 2 for version 12. It would be great if you can please make the appropriate changes and push them on priority. Thanks

Alchez:
@deepeshgarg007 @AfshanKhan I've changed the format method, should pass for 2.7 now

AfshanKhan (approved):

PR #24440, opened 2021-01-21
fix(HR): Remove payroll card from HR workspace

abhishekbalam:
Since Payroll has a dedicated workspace, this is not required in HR workspace.

image


abhishekbalam:
@Anurag810 

Travis failing due to some other reason.

Anurag810 (commented):
Approved

PR #24445, opened 2021-01-22
feat: shopping portal changes

anupamvs:
- Capturing Lead source(**Product Inquiry**) when a lead is created from the product detail page's 'Contact Us' button.
- Added **Continue Shopping** button on cart page with redirects to the all-products page.
sc-changes

### DIFF
--- a/erpnext/shopping_cart/cart.py
+++ b/erpnext/shopping_cart/cart.py
@@ -180,6 +180,13 @@ def create_lead_for_item_inquiry(lead, subject, message):
 	lead_doc.update(lead)
 	lead_doc.set('lead_owner', '')
 
+	if not frappe.db.exists('Lead Source', 'Product Inquiry'):
+		frappe.get_doc({
+			'doctype': 'Lead Source',
+			'source_name' : 'Product Inquiry'
+		}).insert(ignore_permissions=True)
+	lead_doc.set('source', 'Product Inquiry')
+
 	try:
 		lead_doc.save(ignore_permissions=True)
 	except frappe.exceptions.DuplicateEntryError:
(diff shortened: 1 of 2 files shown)


erpnext/templates/pages/cart.html
```diff
@@ -47,6 +47,9 @@
 
 	{% if doc.items %}
 		<div class="place-order-container">
+			<a class="btn btn-primary-light mr-2" href="/all-products">
+				{{ _("Add More") }}
+			</a>
```
marination: As dicussed let's make this **Continue Shopping**

erpnext/shopping_cart/cart.py
```diff
@@ -180,6 +180,13 @@ def create_lead_for_item_inquiry(lead, subject, message):
 	lead_doc.update(lead)
 	lead_doc.set('lead_owner', '')
 
+	if not frappe.db.exists('Lead Source', 'Product Inquiry'):
+		frappe.get_doc({
+			'doctype': 'Lead Source',
+			'source_name' : 'Product Inquiry'
+		}).insert(ignore_permissions=True)
+	lead_doc.set('source', 'Product Inquiry')
+
```
marination, line 189: _**Just a suggestion**_: you can maybe add **Product Enquiry** to `default_lead_sources` in `install_fixtures.py` for new instances.
And also add a small patch to create this Lead Source.
It will be a one time thing and will reduce clutter in this function.
```suggestion
	lead_doc.set('source', 'Product Inquiry')
```

anupamvs: @marination what if the user deletes that Lead Source doc?

marination: Almost anything can be deleted that way but since this via portal good to be defensive. fair enough @anupamvs

Issue #24390, opened 2021-01-18
[Refactor] Use new API in client scripts


rmehta:
There are ~615 instances where client scripts are using old API based on the global `cur_frm` object. Convert this to new api.

Example:

```
cur_frm.cscript.refresh = function(doc, cdt, cdn) {
	cur_frm.cscript.set_root_readonly(doc);
}
```

Should be refactored to

```
frappe.ui.form.on('Sales Person', {
  refresh: (frm) => frm.events.set_root_readonly(frm.doc);
});
```

Along with this, also break down long functions into smaller ones.

develop084:
@rmehta  Please assign this issue to me I will make it resolve

rmehta:
Nice! Please feel free to connect with @deepeshgarg007 for any queries.

PR #24448, opened 2021-01-22
Update student_group.js

fderyckel:
reference to [here](https://github.com/frappe/erpnext/issues/24390) about  refactoring  add_fetch using new  api.

> Screenshots/GIFs

### DIFF
--- a/erpnext/education/doctype/student_group/student_group.js
+++ b/erpnext/education/doctype/student_group/student_group.js
@@ -1,6 +1,9 @@
-cur_frm.add_fetch('student', 'title', 'student_name');
 
-frappe.ui.form.on('Student Group', {
+frappe.ui.form.on('Student Group', { 
+	setup: function(frm) { 
+		frm.add_fetch('student', 'title', 'student_name');
+	}, 
+	
 	onload: function(frm) {
 		frm.set_query('academic_term', function() {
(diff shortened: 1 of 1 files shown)


erpnext/education/doctype/student_group/student_group.js
```diff
@@ -1,6 +1,9 @@
-cur_frm.add_fetch('student', 'title', 'student_name');
 
-frappe.ui.form.on('Student Group', {
+frappe.ui.form.on('Student Group', { 
+	setup: function(frm) { 
+		frm.add_fetch('student', 'title', 'student_name');
+	},
```
marination, line 5: This can be moved to **Fetch From** in the DocType specification of **Student Group** DocType.
Essentially no js code is required to make this work. You can remove `add_fetch` from here and drive as mentioned above.

Note: We do not encourage using `add_fetch` in the develop branch (version 13 branch) since it can be driven via the concerned DocType. It is actively used only on the previous versions (v12, v11, etc.)

fderyckel: I see.  Thanks a lot for that.  Not sure if it has already been updated on the json of the doctype.  Otherwise, i'll try to resubmit another PR.  
I will then close this.

PR #24460, opened 2021-01-25
fix: duplication of additional salary component in salary slip

Anurag810:
duplicates row from additional salary in salary details component rows.
image

and overwrite function was not working.


erpnext/payroll/doctype/salary_detail/salary_detail.json
```diff
@@ -9,9 +9,9 @@
   "abbr",
   "column_break_3",
   "amount",
-  "year_to_date",
```
ruchamahabal, line 12: Do not remove the Year to Date field

erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -60,6 +60,8 @@ def validate(self):
 				frappe.msgprint(_("Total working hours should not be greater than max working hours {0}").
 								format(max_working_hours), alert=True)
 
+		# frappe.throw("Hello")
```
ruchamahabal:
```suggestion
```

Remove this.

erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -604,6 +635,9 @@ def update_component_row(self, struct_row, amount, key, overwrite=1, additional_
 			component_row.amount = amount
 			component_row.deduct_full_tax_on_selected_payroll_date = struct_row.deduct_full_tax_on_selected_payroll_date
 
+	def get_overwritten_components(self, key):
+		return [ele.salary_component for ele in self.get(key) if ele.overwritten == 1]
```
ruchamahabal:
```suggestion
		return [entry.salary_component for entry in self.get(key) if entry.overwritten]
```
Use `entry` instead of `ele`. No need to explicitly check whether `ele.overwritten == 1`. `entry.overwritten` will evaluate to True.

erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -566,12 +568,39 @@ def add_tax_components(self, payroll_period):
...
+
+				if struct_row.get("is_additional_component") and d.additional_salary == additional_salary and not overwrite:
+					additional_salary_row = d
+					component_found = True
+				elif struct_row.get("is_additional_component") and not d.additional_salary and overwrite:
+					overwrite_component = d
+					overwritten = overwrite
+					component_found = True
+				else:
+					component_row = d
```
ruchamahabal: Can you add tests for these cases?

erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -566,12 +568,39 @@ def add_tax_components(self, payroll_period):
...
+
+		overwritten_components = self.get_overwritten_components(key)
+		if struct_row.salary_component in overwritten_components and (is_salary_structure_component or (struct_row.get("is_additional_component") and overwrite)):
+			return
+
 		for d in self.get(key):
 			if d.salary_component == struct_row.salary_component:
-				component_row = d
-		if not component_row or (struct_row.get("is_additional_component") and not overwrite):
+				if additional_salary == d.additional_salary and d.overwritten == 1:
```
ruchamahabal:
```suggestion
				if additional_salary == d.additional_salary and d.overwritten:
```

ruchamahabal (changes requested):

Anurag810:
@ruchamahabal, @nabinhait. actually, this PR requires a little bit of refactoring. as many edge cases were missing after the fix. working with @sagarvora on this PR: #24824 will merge.
@ruchamahabal please also do one round of review(if possible) from your side.

PR #24470, opened 2021-01-26
feat: Batch wise item pricing

deepeshgarg007:
Screenshot 2021-01-26 at 5 33 47 PM


nabinhait (approved):

sagun-01:
I am facing an issue with this fix. If i have say a total of 10 qty of item 1, with batch "A" having 5 and Batch "B" having 5, and i want to sell 10 of these items, it only picks the rate of the batch with the earlier expiry, doesnt fetch rate of the second batch. Please advise.

nabinhait:
@deepeshgarg007

manan-ft:
Not able to make this feature work. Filed an issue for it 

https://github.com/frappe/erpnext/issues/26377

PR #24479, opened 2021-01-27
fix: adding UOM in update items

Anuja-pawar:
Backport of https://github.com/frappe/erpnext/pull/22869

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.json
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.json
@@ -1056,7 +1056,7 @@
  "idx": 105,
  "is_submittable": 1,
  "links": [],
- "modified": "2020-09-14 14:36:12.418690",
+ "modified": "2021-01-22 20:27:11.418690",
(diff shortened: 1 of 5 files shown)


erpnext/buying/doctype/purchase_order/purchase_order.json
```diff
@@ -1056,7 +1056,7 @@
  "idx": 105,
  "is_submittable": 1,
  "links": [],
- "modified": "2020-09-14 14:36:12.418690",
+ "modified": "2021-01-22 20:27:11.418690",
```
nextchamp-saqib, line 1059: enable `track_changes`

nextchamp-saqib (approved):

erpnext/buying/doctype/purchase_order/purchase_order.json
Anuja-pawar: Yep, it's enabled

nabinhait (approved):

PR #24492, opened 2021-01-28
feat: Create Journal Entry from Bank Transaction, requires frappe bugfix - V12

nameduser0:
Add to the list of options given when reconciling bank transactions.
Hopefully one of many PRs on the way to automating bank transaction journals using Plaid.

Related to frappe/frappe#12280

PR #24493, opened 2021-01-28
feat: Create Journal Entry from Bank Transaction, requires frappe bugfix - V13

nameduser0:
V13 port of #24492

### DIFF
--- a/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
+++ b/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
@@ -8,6 +8,7 @@ from erpnext.controllers.status_updater import StatusUpdater
 from frappe.utils import flt
 from six.moves import reduce
 from frappe import _
+from frappe.model.mapper import get_mapped_doc
 
 class BankTransaction(StatusUpdater):
 	def after_insert(self):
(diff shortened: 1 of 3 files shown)


hasnain2808:
Please Check https://github.com/frappe/erpnext/pull/24273
Creating Journal Entry from Bank Transaction

nameduser0:
> Please Check #24273

I've left a review. I think you need to consider other use-cases for your code.
For example, automatic bank reconciliation via Plaid and creation of missing JE's based on previously manually created vouchers, reconciled against matching bank transactions. Also JE creation with manual intervention.

hasnain2808:
> > Please Check #24273
> 
> I've left a review. I think you need to consider other use-cases for your code.
> For example, automatic bank reconciliation via Plaid and creation of missing JE's based on previously manually created vouchers, reconciled against matching bank transactions. Also JE creation with manual intervention.

All of these were considered and possible

nameduser0:
Moving from #24492:
- [ ] fix: payment document
- [ ] fix: autoreconcile on payment/JE creation
- [ ] fix: move Bank Transaction reference to JEA and change to `Journal Entry` not `Bank Entry`
- [ ] feat: multi-entry JE to same account/party from Bank Transaction list view

PR #24510, opened 2021-02-01
fix: calculated discount percentage

AfshanKhan:
Changing the discount amount wouldn't change the discount percentage, fixed that.

Before:
b4

After:
after


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24511

nabinhait (approved):

PR #24514, opened 2021-02-01
fix: consider select perm while setting party details

saurabh6790:
Before 

photo_2021-02-01_17-54-55


nabinhait (approved):

prasadarr:
no documentation sent

PR #24532, opened 2021-02-02
fix: plaid client version to support latest API - V12

nameduser0:
https://github.com/plaid/plaid-python/blob/master/CHANGELOG.md


nabinhait (approved):

nameduser0:
@nabinhait Thanks! You also need to do #24531

PR #24534, opened 2021-02-03
fix: item_group on update items

pateljannat:
**Issue:**

1. While adding Items to the **Submitted Sales Order** with **Update Items**, Item Group in the Sales Order Item table was not fetched.


erpnext/controllers/accounts_controller.py
```diff
@@ -1319,6 +1319,7 @@ def set_sales_order_defaults(parent_doctype, parent_doctype_name, child_docname,
 	child_item.item_code = item.item_code
 	child_item.item_name = item.item_name
 	child_item.description = item.description
+	child_item.item_group = item.item_group
```
marination:
```suggestion
	for field in ("item_code", "item_name", "description", "item_group"):
	      child_item.update({field: item.get(field)})
```

marination:
@pateljannat given that `set_sales_order_defaults` and `set_purchase_order_defaults` was created approx 2 years ago, i see a lot of redundant lines of code. They could easily be commonified.
Update Items' code is very bulky.

erpnext/controllers/accounts_controller.py
```diff
@@ -1309,45 +1309,28 @@ def add_taxes_from_tax_template(child_item, parent_doc):
...
+	child_item = frappe.new_doc(child_doctype, p_doc, child_docname)
 	item = frappe.get_doc("Item", trans_item.get('item_code'))
-	child_item.item_code = item.item_code
-	child_item.item_name = item.item_name
-	child_item.description = item.description
-	child_item.delivery_date = trans_item.get('delivery_date') or p_doc.delivery_date
+	for field in ("item_code", "item_name", "description", "item_group"):
+	    child_item.update({field: item.get(field)})
+	date_fieldname = "delivery_date" if child_doctype == "Sales Order Item" else "schedule_date"
+	child_item.date_fieldname = trans_item.get(date_fieldname) or p_doc.date_fieldname
```
marination: Cant access attributes that do not exist by dot notation. More importantly you cannot use dot notation to chain a string value. travis is  breaking due to this.
`date_fieldname` does not get substituted here.
```suggestion
	child_item.update({date_fieldname: trans_item.get(date_fieldname) or p_doc.date_fieldname})
```

marination (approved):

PR #24539, opened 2021-02-03
fix: validation for disabled warehouse

pateljannat:
**Issue:**

1. Disable a warehouse.
2. Add that warehouse to any transaction.
3. Submit the transaction.
4. The document submits successfully even though the warehouse entered is disabled.

**Fix:**

1. Validation added to detect disabled warehoused.

Screenshot 2021-02-03 at 2 32 15 PM

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -313,7 +313,7 @@ class StockController(AccountsController):
 		return serialized_items
 
 	def validate_warehouse(self):
-		from erpnext.stock.utils import validate_warehouse_company, is_disabled_warehouse
+		from erpnext.stock.utils import validate_warehouse_company, validate_disabled_warehouse
 
 		warehouses = list(set([d.warehouse for d in
 			self.get("items") if getattr(d, "warehouse", None)]))
@@ -329,7 +329,7 @@ class StockController(AccountsController):
 		warehouses.extend(from_warehouse)
 
 		for w in warehouses:
-			is_disabled_warehouse(w)
+			validate_disabled_warehouse(w)
 			validate_warehouse_company(w, self.company)
 
 	def update_billing_percentage(self, update_modified=True):
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -313,7 +313,7 @@ class StockController(AccountsController):
 		return serialized_items
 
 	def validate_warehouse(self):
-		from erpnext.stock.utils import validate_warehouse_company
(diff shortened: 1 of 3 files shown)


erpnext/controllers/stock_controller.py
```diff
@@ -329,6 +329,7 @@ def validate_warehouse(self):
 		warehouses.extend(from_warehouse)
 
 		for w in warehouses:
+			is_disabled_warehouse(w)
```
marination:
```suggestion
			validate_disabled_warehouse(w)
```
`is_disabled_warehouse` (is the warehouse disabled?) kind of implies that it will return a true/false value back like `is_perpetual_inventory_enabled`.
If its a function that **primarily** carries out a task within itself then it's best to describe what that function is doing.

erpnext/controllers/stock_controller.py
```diff
@@ -313,7 +313,7 @@ def get_serialized_items(self):
 		return serialized_items
 
 	def validate_warehouse(self):
-		from erpnext.stock.utils import validate_warehouse_company
+		from erpnext.stock.utils import validate_warehouse_company, is_disabled_warehouse
```
marination:
```suggestion
		from erpnext.stock.utils import validate_warehouse_company, validate_disabled_warehouse
```

erpnext/stock/utils.py
```diff
@@ -284,6 +284,10 @@ def is_group_warehouse(warehouse):
 	if frappe.db.get_value("Warehouse", warehouse, "is_group"):
 		frappe.throw(_("Group node warehouse is not allowed to select for transactions"))
 
+def is_disabled_warehouse(warehouse):
```
marination:
```suggestion
def validate_disabled_warehouse(warehouse):
```

erpnext/stock/utils.py
```diff
@@ -284,6 +284,10 @@ def is_group_warehouse(warehouse):
 	if frappe.db.get_value("Warehouse", warehouse, "is_group"):
 		frappe.throw(_("Group node warehouse is not allowed to select for transactions"))
 
+def validate_disabled_warehouse(warehouse):
```
marination, line 287: should apply to stock reconciliation as well right ? Stock Reconciliation's validate doesn't trigger Stock Controller's validate.
So you can manually call this function there too or maybe add in the stock ledger entry validate. Whatever fits you best.

nabinhait: Not needed, it will be validated via Stock Ledger Entry

nabinhait (approved):

PR #24546, opened 2021-02-03
Validation for disabled warehouse v12

pateljannat:
Backport of https://github.com/frappe/erpnext/pull/24539


marination:
Left comments on https://github.com/frappe/erpnext/pull/24539

marination (approved):

PR #24551, opened 2021-02-04
fix: Assignment Rule Unassign Condition doesn't work

shariquerik:
The customer has created an Assignment Rule for Lead Doctype which basically removes the assigned person when the status of the Lead changes to 'Converted'. 

This Assignment Rule is working if we manually change the status to 'Converted', but if a user creates a 'Customer' from Lead which basically changes the status of Lead to 'Converted' in the background then Assignment Rule is not getting triggered.


nabinhait (approved):

shariquerik:

shariquerik:
Backport PR: https://github.com/frappe/erpnext/pull/24890

Issue #23977, opened 2020-11-21
Material Request item Plan for Warehouse parent group not calculate actual quantity


yasser1248:
## Description of the issue
**Module:** Manufacture Production Planning

we have multiple warehouses in which we store different parts used for production:
when make material request plan item for BOM
erpnext didn’t subtract available actual quantity from Warehouse parent group 
when chose a child warehouse 1 only . erpnext calculate and subtract actual quantity

how to solve this problem to fetch actual quantity on parent all warehouse group

### Observed result
erpnext didn’t subtract available actual quantity from Warehouse parent group 
### Expected result
erpnext subtract available actual quantity from Warehouse parent group as it when chose a Childe warehouse

![Manufacture Production Planning](https://user-images.githubusercontent.com/74818740/99881273-65e54800-2c21-11eb-9672-af88f38ab06d.png)

### OS version
Ubuntu 18.04
ERPNext: v12.13.0 (version-12)
Frappe Framework: v12.11.0 (version-12)

MaxMorais:
Related to #15781

MaxMorais:
@yasser1248 we do have faced the same issue for one of our customers

I'll prepare the pull request on top of this discussion

sanjay141986:
Please check if the issue resolved in the Latest Release

Issue #15781, opened 2018-10-24
Enhancement in Material Request Planning


nagarajannxweb:
Before creating the material request from production plan, should consider sub-assemblies stock.  

Example : 

Sales order of Item A is 100
Stock Balance of Item A is 10

Item A has 4 Multi-Level Sub Assembly and also have stock in hand like Item A1 - 10, Item A2 - 5 , Item A3 - 3. So the material request should create for Item A - 90, Item A1 - 80, Item A2 - 75 and Item A3 - 72.

- [ ] Add Warehouse group in the material request table

- [ ] Add Material Request Type in the child table 

- [ ] Fetch Items in material request table based on material request type

nagarajannxweb:
- [ ] Add Warehouse group in the material request table

- [ ] Add Material Request Type in the child table 

- [ ] Fetch Items in material request table based on material request type

PR #24560, opened 2021-02-06
Update production_plan.py

MaxMorais:
As commented in #23977 and #15781

Production planning doesn't take in consideration the availability of raw materials in all warehouses
This code changes addresses these problems

> Screenshots/GIFs


erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -625,6 +639,26 @@ def get_bin_details(row, company, for_warehouse=None, all_warehouse=False):
...
+	'''
+	Run thorught all warehouses and verify the projected_qty for an item_code, to
+	determine, if is needed to include the item_code in MR
+	'''
+	bin_details = get_bin_details({'item_code': item_code}, company, all_warehouse=True)
+	qty_needed = sum(map(lambda d: d['projected_qty'], bin_details)) - qty
+	return qty_needed < 0
+
+
+def qty_needed_in_mr(item_code, qty, company):
```
rmehta, line 652: both these functions are similar. why duplicate code?

rmehta (changes requested):

PR #24570, opened 2021-02-08
fix: reposting issue with same posting date and posting time

rohitwaghchaure:
**Issue**

Incorrect balance qty in the stock ledger report, if posting date and posting time is same for two different stock transactions.

Screenshot 2021-02-08 at 11 44 00 AM

**After Fix**

Screenshot 2021-02-08 at 11 43 34 AM


erpnext/stock/stock_ledger.py
```diff
@@ -202,8 +202,7 @@ def get_sle_against_current_voucher(self):
 			where
 				item_code = %(item_code)s
 				and warehouse = %(warehouse)s
-				and voucher_type = %(voucher_type)s
-				and voucher_no = %(voucher_no)s
+				and timestamp(posting_date, posting_time) = timestamp(%(posting_date)s, %(posting_time)s)
```
nabinhait: we need to avoid milliseconds
```suggestion
				and timestamp(posting_date, time_format(posting_time, '%H:%i:%s')) = timestamp(%(posting_date)s, time_format(%(posting_time)s, '%H:%i:%s'))
```

nabinhait (changes requested):

nabinhait (approved):

PR #24572, opened 2021-02-08
feat: Department Wise Appointment Type charges

ruchamahabal:
### Appointment Type Set up

- You can now set up department wise appointment/consultation charges using Appointment Type

   image

- If you want to set a generic default charge irrespective of any department, you can add a record with no department selected:

  image

### Billing

Here, appointment type charge overrides the practitioner charge:

- If appointment type is selected:
   - If department wise charge and billing item is set up, use
   - else, use the generic setup (not specific to any department)
   - If no generic set up is done, use practitioner billing item and charges
- else, use practitioner billing item and charges
- Validation thrown if none of the above has been set up

If **Automate Appointment Invoicing** is enabled in the Healthcare Settings, then, while creating the Appointment, the appropriate billing item, and the paid amount is fetched as per the above logic.

billing

Documentation PR: https://github.com/frappe/erpnext_documentation/pull/246

- [x] Tests
- [x] Documentation


mujeerhashmi:
I know this is not the correct place to make a request. But it is inline with this development. There are hospitals who do not take appointments but do direct consultations on daily basis for walk-in patients. We need automated patient encounter invoicing and a flag to differentiate if it was outpatient encounter or inpatient encounter.

ruchamahabal:
> There are hospitals who do not take appointments but do direct consultations on daily basis for walk-in patients. We need automated patient encounter invoicing.

This can be taken up separately as a feature request.

> A flag to differentiate if it was outpatient encounter or inpatient encounter.

However, we already have a differentiator here. If an Inpatient Record is linked to the Encounter then it is an IP encounter else an OP encounter.

mujeerhashmi:
> However, we already have a differentiator here. If an Inpatient Record is linked to the Encounter then it is an IP encounter else an OP encounter.

I also had thought so but when you schedule an admission from an outpatient encounter the `Inpatient Status` and `Inpatient Record` are populated in it.

ruchamahabal:
> I also had thought so but when you schedule an admission from an outpatient encounter the `Inpatient Status` and `Inpatient Record` are populated in it.

Oh Right! Will think of a way to handle this.

mujeerhashmi:
> > There are hospitals who do not take appointments but do direct consultations on daily basis for walk-in patients. We need automated patient encounter invoicing.
> 
> This can be taken up separately as a feature request.

Raised feature request #24582

nabinhait (approved):

PR #24580, opened 2021-02-08
fix: Consolidated Financial Statement report not works if child compa…

rohitwaghchaure:
**Issue**

If child company account is not present in the parent company account then while checking the Consolidated Financial Statement report system throws the validation that "Account name Test 123 is not present in the parent company"

Screenshot 2021-02-08 at 7 04 24 PM

**After Fix**

System will display the records even if the child company account not present in the parent account


nabinhait (approved):

nabinhait:
@rohitwaghchaure Please send PR for develop branch

PR #24606, opened 2021-02-11
fix: Issues with packing items

ruchamahabal:
**Problems**: 

In Delivery Note created from a Sales Order

1. Serial Number from “Add Serial No” on packed item detail doesn’t appear unless you close and reopen the row dialog.
2. The “save” button doesn’t always appear since the form is not dirty after adding serial numbers via the **Add Serial No** button.

No gif since Licecap is not working


nabinhait (approved):

nabinhait:
@ruchamahabal Is this fix also needed in develop branch?

ruchamahabal:
@nabinhait Fixed in https://github.com/frappe/erpnext/pull/24615

PR #24610, opened 2021-02-11
refactor: set missing values of pos sales invoices

nextchamp-saqib:
Commonify pos invoice & sales invoice `set_missing_values`


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -453,70 +447,65 @@ def validate_time_sheets_are_submitted(self):
...
-
-		if pos:
-			if not for_validate:
-				self.tax_category = pos.get("tax_category")
+		return frappe.get_cached_doc('POS Profile', self.pos_profile) if self.pos_profile else {}
 
-			if not for_validate and not self.customer:
-				self.customer = pos.customer
+	def get_customer_specific_info(self):
+		customer_price_list, customer_group_price_list = None, None
```
```suggestion
		customer_price_list, customer_group_price_list, currency = None, None, None
```

PR #24611, opened 2021-02-11
fix: Added parent task expected end date validation

shariquerik:
Currently, a Task's Expected Date can be set as a later date and the system accepts it without validating the Expected End of the Parent Task, that shouldn't be the case, because we don't allow Parent Task to get completed before the Child Task

### DIFF
--- a/erpnext/projects/doctype/task/task.py
+++ b/erpnext/projects/doctype/task/task.py
@@ -30,7 +30,7 @@ class Task(NestedSet):
 
     def validate(self):
         self.validate_dates()
-		self.validate_parent_expected_end_date()
+        self.validate_parent_expected_end_date()
         self.validate_parent_project_dates()
         self.validate_progress()
         self.validate_status()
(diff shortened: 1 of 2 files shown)


erpnext/projects/doctype/task/task.py
```diff
@@ -30,6 +30,7 @@ def get_customer_details(self):
 
     def validate(self):
         self.validate_dates()
+		self.validate_parent_expected_end_date()
```
ruchamahabal: Check the indentation

erpnext/projects/doctype/task/task.py
```diff
@@ -45,6 +46,12 @@ def validate_dates(self):
             frappe.throw(_("{0} can not be greater than {1}").format(frappe.bold("Actual Start Date"), \
                 frappe.bold("Actual End Date")))
 
+	def validate_parent_expected_end_date(self):
+		if self.parent_task:
+			parent_exp_end_date = frappe.db.get_value("Task", self.parent_task, "exp_end_date")
+			if getdate(self.get('exp_end_date')) > getdate(parent_exp_end_date):
+				frappe.throw(_("Expected End Date should be less than or equal to parents task's Expected End Date {0}.").format(getdate(parent_exp_end_date)))
```
ruchamahabal:
```suggestion
			if parent_exp_end_date and getdate(self.get("exp_end_date")) > getdate(parent_exp_end_date):
				frappe.throw(_("Expected End Date should be less than or equal to parent task's Expected End Date {0}.").format(getdate(parent_exp_end_date)))
```

1. Check if `parent_exp_end_date` is set since the system should allow setting any date for the child task if the parent's end date is missing.
2. Make the quote-style uniform.
3. Minor change in the message.

ruchamahabal (changes requested):

erpnext/projects/doctype/task/task.py
shariquerik: Done

erpnext/projects/doctype/task/task.py
shariquerik: Done

ruchamahabal:
Check the failing test `test_project_template_having_parent_child_tasks`. You will have to change the dates there.

ruchamahabal (approved):

shariquerik:
Backport PR: https://github.com/frappe/erpnext/pull/24889

PR #24618, opened 2021-02-12
fix: incorrect incoming rate for the sales return

rohitwaghchaure:
**Issue**

Make standalone sales invoice return entry with update stock and submit it. After submit check the stock ledger entry, you will found that the incoming rate has set as zero in the stock ledger entry. Due to zero incoming rate system has calculated incorrect valuation rate for the future entries.

Screenshot 2021-02-12 at 1 18 34 PM

**After Fix**

Screenshot 2021-02-12 at 1 20 25 PM


akshay83:
@rohitwaghchaure is this fix included in version 13 or current develop branch also?
as during **testing** of the newer version I came across a similar situation. 

ERPNext: v13.x.x-develop () (develop)

image

PR #24622, opened 2021-02-12
fix: supplier was not able to Submit RFQ due to insufficient permission

Anurag810:
On submission of "Request for Quotation" from supplier portal. Supplier Quotation not got created due to insufficient permission.


erpnext/setup/doctype/item_group/item_group.json
```diff
@@ -277,7 +277,7 @@
    "export": 1,
    "print": 1,
    "report": 1,
-   "role": "Customer",
+   "role": "All",
```
marination, line 280: This will revert https://github.com/frappe/erpnext/commit/e816ed89e6923bc5d378ccc0440e68aa3b61d414
Maybe add a role, don't remove an existing role

erpnext/stock/doctype/item/item.json
```diff
@@ -1120,6 +1120,15 @@
   {
    "read": 1,
    "role": "Manufacturing User"
+  },
+  {
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "report": 1,
+   "role": "All",
```
marination, line 1127: why is **All** required here ? The issue occurs on a user with **Supplier** role right ? How is that handled ?

marination:
@Anurag810 please make the description more descriptive, mention the issue but also explain the fix and how it impacts the current flow.

erpnext/setup/doctype/item_group/item_group.json
Anurag810: It was removed by mistake.

erpnext/stock/doctype/item/item.json
Anurag810: I can explain better on call. Actually, this was paired with Rohit, and he just told me to give select permission to all. it will be fixed.

erpnext/setup/doctype/item_group/item_group.json
Anurag810: Yeah but that was not the correct fix. We have to provide select permission to all in the case to fix this issue.

PR #24631, opened 2021-02-15
fix: check if customer provided item while setting opening stock

AfshanKhan:
If customer provided item was created with opening stock following error was throw:

image

Added a condition to skip if the item was customer provided item


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24633

marination (approved):

PR #20135, opened 2019-12-29
feat: Validity for Item taxes

deepeshgarg007:
image

Docs: https://github.com/frappe/erpnext_com/pull/637

PR #24643, opened 2021-02-16
fix: Don't throw exception on invoice lines when there is no item_code

### LINKED ISSUE #24640: bug: Purchase and Sales Invoice lines should never need an item code for anything, to support one-off manually typed lines
Purchase and Sales Invoice lines should never need an item code for anything, to support one-off manually typed lines

Manual selection of item tax templates should be possible

_Originally posted by @casesolved-co-uk in https://github.com/frappe/erpnext/pull/20135#r576528638_

@deepeshgarg007

nameduser0:
fixes #24640

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -1885,7 +1885,6 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			frappe.throw(__("Please enter Item Code to get batch no"));
 		} else if (doc.doctype == "Purchase Receipt" ||
 			(doc.doctype == "Purchase Invoice" && doc.update_stock)) {
-
 			return {
 				filters: {'item': item.item_code}
 			}
@@ -1911,9 +1910,8 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	set_query_for_item_tax_template: function(doc, cdt, cdn) {
 		var item = frappe.get_doc(cdt, cdn);
 		if(!item.item_code) {
-			frappe.throw(__("Please enter Item Code to get item taxes"));
+			return doc.company ? {filters: {company: doc.company}} : {};
 		} else {
-
 			let filters = {
 				'item_code': item.item_code,
 				'valid_from': ["<=", doc.transaction_date || doc.bill_date || doc.posting_date],


erpnext/public/js/controllers/transaction.js
```diff
@@ -1911,7 +1911,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 	set_query_for_item_tax_template: function(doc, cdt, cdn) {
 		var item = frappe.get_doc(cdt, cdn);
 		if(!item.item_code) {
-			frappe.throw(__("Please enter Item Code to get item taxes"));
+			// Do nothing
 		} else {
```
marination: In the case of not having an item code for manual entries, `set_query_for_item_tax_template` is only triggered if a user triggers the field.
The **Item** Tax Template is tied to an **Item**. And so if there is no item code, just an item name for quick invoicing, it should rightfully throw this error.
In majority of the cases Item codes are used, for which this validation is important.

I'd question why the user is trying to select an Item Tax Template without a valid item ? Don't see anything to fix here. Just my views.

nameduser0: > In the case of not having an item code for manual entries, set_query_for_item_tax_template is only triggered if a user triggers the field.

For clarity: if the user triggers the `item_tax_template` field

> The Item Tax Template is tied to an Item. And so if there is no item code, just an item name for quick invoicing, it should rightfully throw this error.

I'm afraid you're incorrect. The Item Tax Template is not tied to an item, the template content is saved in the `item_tax_rate` field on the `* Invoice Item` record as json. There is no reason for there to be an `item_code` at all. This storage is vitally important so that Invoices are fixed when submitted, otherwise changes to the template will affect past records. The Sales Invoice page shows the `Item Name` in the table instead of the `Item Code`, unlike the Sales Order page to allow just this functionality.

I have tested this change and it works well, but if left as it is the line prevents arbitrary line items on a Sales Invoice.

> In majority of the cases Item codes are used, for which this validation is important.

But not in *all* cases. There are numerous perfectly valid use-cases that should *never* require an item_code:

- Purchase invoices for one-off items, like lunch bought by the company
- Sales of bespoke items

In any case, this line just overrides the doctype mandatory field validation.

nameduser0:
What's the holdup please? This is a serious regression which cannot be released in v13

Introduced in #20135

@nabinhait @deepeshgarg007

erpnext/public/js/controllers/transaction.js
marination: @casesolved-co-uk Apologies, the use case was not as clear to me. This seems very valid, i checked if it breaks anything. Server side validations are in place, so all good.

**A change I'd suggest:**
In the `if(!item.item_code)` section , we should at least filter the templates by **company** in absence of item code.

marination:
@casesolved-co-uk Please also update the branch after that, we can get this merged

erpnext/public/js/controllers/transaction.js
```diff
@@ -1910,10 +1910,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 
 	set_query_for_item_tax_template: function(doc, cdt, cdn) {
 		var item = frappe.get_doc(cdt, cdn);
-		if(!item.item_code) {
-			frappe.throw(__("Please enter Item Code to get item taxes"));
-		} else {
-
+		if(item.item_code) {
```
deepeshgarg007: @casesolved-co-uk please filter item tax templates by company if item code is not entered. Currently, it won't apply any filter in case item code is not entered.

erpnext/public/js/controllers/transaction.js
nameduser0: Now done

erpnext/public/js/controllers/transaction.js
nameduser0: Now done and tested

erpnext/public/js/controllers/transaction.js
```diff
@@ -1911,9 +1910,14 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
 		var item = frappe.get_doc(cdt, cdn);
 		if(!item.item_code) {
-			frappe.throw(__("Please enter Item Code to get item taxes"));
+			if (doc.company) {
+				return {
+					filters: {company: doc.company}
+				};
+			} else {
+				return {};
+			}
```
marination:
```suggestion
                      let company_filter = {filters: {company: doc.company}};
                      return doc.company ? company_filter : {};
```

casesolved-co-uk: Now you're just being pedantic. Has your change been tested? Because mine has and I'm not testing yours.
You should be grateful that I have found, fixed and tested a serious regression and bug in your product.

marination (approved):

PR #24645, opened 2021-02-16
fix: broken image style

prssanna:
Fixes broken css after https://github.com/frappe/frappe/pull/12370

### DIFF
--- a/erpnext/public/less/hub.less
+++ b/erpnext/public/less/hub.less
@@ -32,7 +32,13 @@ body[data-route*="marketplace"] {
 	}
 
 	.hub-image-loading, .hub-image-broken {
-		.img-background();
+		content: " ";
+		display: block;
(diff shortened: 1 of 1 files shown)


erpnext/public/less/hub.less
```diff
@@ -32,7 +32,13 @@ body[data-route*="marketplace"] {
...
 	.hub-image-loading, .hub-image-broken {
-		.img-background();
+		content: " ";
+		display: block;
+		position: absolute;
+		left: 0;
+		height: 100%;
+		width: 100%;
+		background-color: var(--bg-light-gray);
 		display: flex;
```
sagarvora, line 42: `display` property being repeated.

prssanna: Fixed, thanks!

PR #24651, opened 2021-02-16
fix(selling): cancel sales order with cancelled PE

ankush:
- Allow cancelling sales order with cancelled payment entry. Ignoring GL entries while cancelling the document is required to cancel it, reverse entries are created by accounts controller.
- Add test for checking this scenario

Related issue: ISS-20-21-09586


deepeshgarg007 (approved):

ankush:

PR #24652, opened 2021-02-16
fix: In-Transit Feature enhancements

anupamvs:
- Change the label in the company (**Default In Transit Warehouse** -> **Default In-Transit Warehouse**)
- Added Default In-transit Warehouse for warehouse.
- Auto fetch Default In-transit Warehouse based in source warehouse and company in Stock Entry.


deepeshgarg007 (commented):
@anupamvs what is the use case of adding the default in-transit warehouse field in the warehouse master?

nabinhait (approved):

PR #24653, opened 2021-02-16
fix: rounding error in finished goods qty

Anuja-pawar:
Issue: Rounding error while calculating the sum for finished goods qty.

Screenshot 2021-02-16 at 5 59 29 PM

Steps to replicate:

1.  Create a BOM
2. Create a Work Order against it
3. Create a Stock Entry (Manufacture) against it
4. Split the Finished Goods in the target warehouse into 10 line (or any) items (Refer to the Screenshot) (Also, don't forget to check `Is Finished Item` for each of them)

Note: Qty should be in decimal, not the whole number in order to replicate the issue.


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -276,9 +276,10 @@ def validate_fg_completed_qty(self):
 					item_wise_qty.setdefault(d.item_code, []).append(d.qty)
 
 		for item_code, qty_list in iteritems(item_wise_qty):
-			if self.fg_completed_qty != sum(qty_list):
+			total = round(sum(qty_list), frappe.get_precision("Stock Entry Detail", "qty"))
```
rohitwaghchaure: Use flt instead of round

nextchamp-saqib (approved):

PR #24658, opened 2021-02-17
fix(selling): cancel sales order with cancelled PE

ankush:
Backport of #24651


deepeshgarg007:
@ankush In version-12 we delete the GL Entry itself on cancellation of the Voucher. I think this issue might not happen in version-12

Issue #10604, opened 2017-08-30
Recurring email message


mnatalia:
According to the topic https://discuss.erpnext.com/t/recurring-email-message/27686.
Could you please with new subscription module add possibility to change email message in recurring documents.

umairsy:
For suggestion on this feature, please add a comment on the #6313

PR #24660, opened 2021-02-17
fix: Add homepage to sitemap

nameduser0:
Fixes the puzzling omission of the homepage from the sitemap
Crawlers should resolve the canonical path of the page if there are duplicates:
`['', 'home']`
`<link rel="canonical" href="http://v13:8002">`
[https://moz.com/learn/seo/canonicalization](https://moz.com/learn/seo/canonicalization)

Related frappe/frappe#10613

@netchampfaris 

### DIFF
--- a/erpnext/templates/pages/home.py
+++ b/erpnext/templates/pages/home.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 import frappe
 
 no_cache = 1
+sitemap = 1
 
 def get_context(context):
 	homepage = frappe.get_doc('Homepage')


nameduser0:
Any luck with this?
Considering Rushabh is always keen for contributions, a lot of them are ignored!

ossplus:
Your canonical link is wrong. #10604
yes, add sitemap = 1 can include home in the sitemap,
but the home URL in the sitemap is "domain/home", the right URL should be "domain".

nameduser0:
@gmplab that's not true. Nothing in the canonicalisation docs says the sitemap and tag need to be the same, only that they resolve to the same page so that they can be crawled, which they do.

PR #24663, opened 2021-02-18
fix: permission error for website customer

AfshanKhan:
To replicate:
Login with the Website User
Go to Product Page
Select any Product and add it to the Cart.

image

This happens because the customer doesn't have permission to the DocType "Territory" and `get_list()` doesn't ignore permissions. Changing it to `get_all()' solves the permission issue.


deepeshgarg007:
@AfshanKhan Should be fixed via this PR as well https://github.com/frappe/erpnext/pull/24577

AfshanKhan:
> @AfshanKhan Should be fixed via this PR as well #24577

Yup, it will. Closing this.

PR #24665, opened 2021-02-18
fix: error message when making reverse journal entry

AfshanKhan:
Previous error message: `For {0}, only credit accounts can be linked against another debit entry` which does not give a clear message the new message `"Row #{0}: For {1}, you can select reference document only if account gets credited` tries to be more clear


marination:
@AfshanKhan please add a description so the reviewer, or really anyone, has context to why the changes were made and why the pre-existing message was misleading.

AfshanKhan:
> @AfshanKhan please add a description so the reviewer, or really anyone, has context to why the changes were made and why the pre-existing message was misleading.

@marination Updated. Got the point.

nabinhait (approved):

PR #24666, opened 2021-02-18
fix: error message when making reverse journal entry

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/24665


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24666

nabinhait (approved):

PR #24680, opened 2021-02-19
fix: Add phone icon on read only phone fields

shariquerik:
Phone Icon is missing from read_only phone fields.
Before:
image

After:
image

### DIFF
--- a/erpnext/public/js/telephony.js
+++ b/erpnext/public/js/telephony.js
@@ -1,13 +1,16 @@
 frappe.ui.form.ControlData = frappe.ui.form.ControlData.extend( {
 	make_input() {
-		this._super();
+		if(!this.df.read_only) {
+			this._super();
+		}
 		if (this.df.options == 'Phone') {
 			this.setup_phone();
 		}
 	},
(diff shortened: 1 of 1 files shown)


shariquerik:
Dependent Link PR: https://github.com/frappe/frappe/pull/12442

erpnext/public/js/telephony.js
```diff
@@ -1,13 +1,16 @@
 frappe.ui.form.ControlData = frappe.ui.form.ControlData.extend( {
 	make_input() {
-		this._super();
+		if(!this.df.read_only) {
```
leela: Why is this check needed?

shariquerik: > Why is this check needed?

If the field is `read_only` we do not have to make the input. There is two div with class `control-input` and `control-value` if the field is `read_only` the content inside `control-input` should not be created only div with `control-value` should be created and vice versa.

leela (dismissed):
Looks good. Please check sider issues before merging.

nabinhait (approved):

PR #24695, opened 2021-02-22
fix(india): escape for special characters in JSON

ankush:
JSON does not accept special whitespace characters like tab, carriage return, line feed.

JSON.dumps escapes json specific characters.

Ref: https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf

Related issue: ISS-20-21-09811


nabinhait (approved):

nabinhait:
@ankush Please send another PR to `version-12-hotfix` branch

PR #24701, opened 2021-02-22
fix: item attributes not editable until refresh

shariquerik:
The item attributes under Item master at first are not editable, but after refresh they are editable

Steps to reproduce:

- Go to item list
- Search for a product either through a filter or manually which is a variant
- Open that variant item
- On the top, most of the page the Item template product will be mentioned- Open that
- Go to system attributes, it will not allow editing
- Refresh the page and check the item attributes will be now editable.


marination (approved):

marination:
@shariquerik backport ?

shariquerik:

shariquerik:
Backport PR: https://github.com/frappe/erpnext/pull/24887

PR #24707, opened 2021-02-23
fix: add item taxes at the same times as sales and purchase taxes

FHenry:
Next of  #24276

In france, in domestic sells, there is no use of Sales Taxes as they are implemented into ERPNExt (v12, it was ok in v11), we use now Item Tax.
The only tax on selling goods and service is VAT. It depending of what your selling, not depending of customer :
VAT 5.5% (soft drink, books),VAT 10%(food),VAT 20% (all other product and services)
Let's take an exemple :
When you add an item to quote/order or invoice.
Product A : Price (excl. tax) is 100 € / VAT 20 %
When added to a domestic invoice
Line 1 : Product A (qty 1) : Price (excl. tax) is 100 €, VAT 20% : 20 €, Price (incl. tax) 120 €
At the bottom of invoice Total Price (excl. tax) 100 %, Total VAT 20 % 20€ , Total price (incl. tax) 120 €
VAT is calculated on each line on sum up at the end into a total

This cannot be done with sales taxes, the item tax depends on Item Taxes not Sales Taxes.

At every installation of ERP Next V12, I have to recreate manually Item Tax

Do you want to me suggest another way to do it, like another button, and another item_tax_item.josn files that match this case ?


deepeshgarg007:
@FHenry Please raise the PR on develop branch

PR #24725, opened 2021-02-24
feat: (get_items_from) filter material request item in purchase order

nextchamp-saqib:
Depends on: https://github.com/frappe/frappe/pull/12471

While fetching Material Request in Purchase Order:

Screenshot 2021-05-10 at 5 05 43 PM

Checking "Select Individual Items" will switch enable users to select/filter Material Request Item:

Screenshot 2021-05-10 at 5 05 55 PM


leela:
Hi @nextchamp-saqib ,

`Fetch Material Request Items` and `Get Items` buttons convey similar meaning irrespective of their purpose. Please rename them appropriately!

And also What happens when I select single item from `material request1` and whole another `material request2`? What I see is only one item from `material request1` is got selected. Is that expected?

erpnext/stock/doctype/material_request/material_request.py
```diff
@@ -269,7 +272,10 @@ def update_status(name, status):
 	material_request.update_status(status)
 
 @frappe.whitelist()
-def make_purchase_order(source_name, target_doc=None):
+def make_purchase_order(source_name, target_doc=None, args={}):
```
ankush, line 275: mutable args will cause problems down the line. 

---

Also make_purchase_receipt (and many similar functions) expects only one positional argument and one optional argument. This is breaking all of them. Reference: https://github.com/frappe/erpnext/issues/27557

ChillarAnand: @ankush Linter should have caught this. 

Added [CI check for the same here](https://github.com/ChillarAnand/erpnext/commit/af03a3606dd9a1f002bb5f5e892fa992c7fc32f5). Will raise a PR once [this PR gets merged](https://github.com/frappe/erpnext/pull/27362).

ankush: Semgrep catches it only when it's actually dangerous. ;)

ankush: https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/common-mistakes/default-mutable-list.yaml 
https://github.com/returntocorp/semgrep-rules/blob/develop/python/lang/correctness/common-mistakes/default-mutable-dict.yaml

PR #24726, opened 2021-02-24
fix: currency symbol in update items

AfshanKhan:
When you choose update items on a submitted purchase order the currency symbol is set to company default. it should be set to supplier/ purchase order default.

The numbers look correct, it is just the symbol. Very confusing for operators.

Before:
Screenshot 2021-02-24 at 2 33 48 PM

After fix:
Screenshot 2021-02-24 at 2 35 15 PM


AfshanKhan:
Backport at https://github.com/frappe/erpnext/pull/24727

PR #24749, opened 2021-02-25

FHenry:


marination (approved):

marination:
@FHenry please send a PR on `version-12-hotfix` and `develop`. `version-12` is a production branch

FHenry:
@marination Thank,'s to take time to review this PR. 
I just create two new PR #24853 AND #24855

As I understand there is no merge by Frappe Team from version-12 into 13 or develop ? When we fix something into a version we must also create another PR into upper version, I'm I right ? Should I do a PR on version-13 also ?

PR #24756, opened 2021-02-26
fix: fetch Material Requests of type Customer Provided  in Stock Entry using Get Items From

sagarvora:
- Added setter for **Material Request Type** with options **Material Transfer**, **Material Issue**, and **Customer Provided**. It needs to be chosen as two types cannot be clubbed together while creating Stock Entry.
- Added a mandatory setter for **Customer** depending on Material Request Type **Customer Provided**, since material requests to two different customers cannot be logically clubbed into the same stock entry. The setter's input is focused automatically when the Material Request Type **Customer Provided** is chosen.
- `erpnext.utils.map_current_doc` now returns the `MultiSelectDialog` object for further use.

**Tested Locally**

---

### Before

image

### After

image

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -231,15 +231,37 @@ frappe.ui.form.on('Stock Entry', {
 			}, __("Get Items From"));
 
 			frm.add_custom_button(__('Material Request'), function() {
-				erpnext.utils.map_current_doc({
+				const allowed_request_types = ["Material Transfer", "Material Issue", "Customer Provided"];
+				const depends_on_condition = "eval:doc.material_request_type==='Customer Provided'";
+				const d = erpnext.utils.map_current_doc({
 					method: "erpnext.stock.doctype.material_request.material_request.make_stock_entry",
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -231,15 +231,37 @@ frappe.ui.form.on('Stock Entry', {
...
 					target: frm,
 					date_field: "schedule_date",
-					setters: {},
+					setters: [{
+						fieldtype: 'Select',
+						label: __('Purpose'),
+						options: allowed_request_types.join("\n"),
+						fieldname: 'material_request_type',
+						default: "Material Transfer",
+						mandatory: 1,
```
marination, line 247: The use case mentioned makes sense, but let's not make this mandatory due to the volume of edge cases we see often.
Usually the user raising the material request will differ from the one raising the stock entry.
MR user could want to **Issue** some stock, but SE user may decide to **transfer** that _Issued_ stock elsewhere. So its hard to tell what MR purpose, maps to what Stock Entry type.

The rest looks good.

marination (dismissed):

marination (approved):

PR #24765, opened 2021-02-28
Update requirements.txt

wstates:
pip install frappe
An installation error will occur

pip install git+https://github.com/frappe/frappe.git
Installing directly with the git address works fine

> Screenshots/GIFs


ankush:
Use Frappe bench to install frappe apps.

https://github.com/frappe/bench

PR #24768, opened 2021-03-01
chore: improve error message for missing fields.

ankush:
Screenshot 2021-03-01 at 10 47 37 AM


erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -504,7 +504,7 @@ def eval_condition_and_formula(self, d, data):
 			return amount
 
 		except NameError as err:
-			frappe.throw(_("Name error: {0}").format(err))
+			frappe.throw(_("Name error: {0} <br> This error can be due to missing or deleted field.").format(err))
```
marination: Wherever possible try, add titles to error messages. The `Message` seems very non human
```suggestion
			frappe.throw(_("{0} <br> This error can be due to missing or deleted field.").format(err), 
			    title=_("Name error"))
```

marination (approved):

PR #24773, opened 2021-03-01
fix: GL Entries for AP/AR Summary

### LINKED ISSUE #22846: Accounts Receivable Summary mismatch with General ledger and Customer Ledger Summary-v13
- General Ledger Debit Amount and Customer Ledger Summary Invoiced Amount agree.
- General Ledger Credit Amount and Customer Ledger Summary Paid Amount do not agree, because Paid Amount in CLS includes cancelled entries.
- 
- Invoiced Amount and Paid Amount in Accounts Receivable Summary, do not agree with what's on the GL and Customer Ledger Summary. Could be a reason for this, but it's not clear where the figures in the Accounts Receivable Summary come from.

UrvashiKishnani:
SQL query modified to fetch only those GL Entries for Accounts Payable Summary and Accounts Receivable Summary reports where the corresponding payment entry is in submitted state.

Issue:
Currently the AP and AR Summary report calculates the advance amount from payment entries. However, the filter to check whether the payment entry is in submitted state is not present.

Steps to reproduce:
1. Create a payment entry for a certain supplier which is not linked to any payment invoice. This will add the amount paid in the payment entry to the advance payment column in the accounts payable report.
2. Cancel the above payment entry. The advance payment column in the accounts payable report still reflects this amount and the paid amount column is adjusted to tally the outstanding amount. However, these numbers do not make sense against the Accounts Payable report and the calculated paid amount is incorrect.

Closes #22846

### DIFF
--- a/erpnext/accounts/party.py
+++ b/erpnext/accounts/party.py
@@ -606,20 +606,19 @@ def get_partywise_advanced_payment_amount(party_type, posting_date = None, futur
 	cond = "1=1"
 	if posting_date:
 		if future_payment:
-			cond = "gle.posting_date <= '{0}' OR DATE(creation) <= '{0}' """.format(posting_date)
+			cond = "posting_date <= '{0}' OR DATE(creation) <= '{0}' """.format(posting_date)
 		else:
-			cond = "gle.posting_date <= '{0}'".format(posting_date)
+			cond = "posting_date <= '{0}'".format(posting_date)
 
 	if company:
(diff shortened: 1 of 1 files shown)


deepeshgarg007 (changes requested):
@UrvashiKishnani No need for joins. Just add a filter for is_cancelled = 0 like shown below
image

deepeshgarg007 (approved):

PR #24776, opened 2021-03-01
fix: Add warning for invalid GST invoice numbers

ankush:
GST Invoice numbers should be 16 characters alphanumeric with dash(/) or slash(-) only. Add check for doc.name before saving and warn about naming series.


ankush:
@leela  moved all regex patterns to global variables. This way `re.compile` is actually beneficial too.

Added note about why arbitrary date was chosen.

deepeshgarg007 (approved):

shreyanshj:
I don't think this is a right patch. This is enforcing 16 digit for Purchase Invoices as well - GST law doesn't ask you to file your Purchase Invoices - it only requires to push data for Sales.
Any particular reason why a blanket control is required? Am I mistaken?

PR #24793, opened 2021-03-03
fix: make quick entry for Tax Category work

barredterra:
### Problem
Quick Entry was already enabled for Tax Category. However, the only field "Title" did not have "Mandatory" or "Allow in Quick Entry" checked. Therefore the Quick Entry was skipped and the user redirected to the form view.

### Solution
Because Title is the only field and the name depends on it, the field should indeed be mandatory. With this setting in place, Quick Entry works as expected. This results in a much smoother UX.


barredterra:

deepeshgarg007 (approved):

PR #24805, opened 2021-03-04
refactor: Setup Taxes and Charges

barredterra:
ERPNext tries to create appropriate Taxes and Charges Templates (and corresponding VAT accounts) for every country.

This process uses very basic input data. For example, the data for the country of Finland looks like this:

https://github.com/frappe/erpnext/blob/99c8b8efcfabf83b267bf6c106e5bba70675f75f/erpnext/setup/setup_wizard/data/country_wise_tax.json#L439-L444

When creating a Company based in Finland, ERPNext will go ahead and

1. create a liability **Account** named "VAT" with a tax rate of 24 %, 
2. create a **Sales Taxes and Charges Template** named "Finland Tax", using the **Account** from (1)
3. create a **Purchase Taxes and Charges Template** named "Finland Tax", using the **Account** from (1)
4. create an **Item Tax Template** named "Finland Tax", using the **Account** from (1)

The problem with this approach is that it's far too coarse. For example, we might use ...

- ... tax assets, not just liabilities
- ... multiple accounts per template
- ... different accounts for sales and purchase taxes.
- ... different accounts depending on the Chart of Accounts used.

This PR introduces a more complex data structure that can be used optionally. This way the above cases can be solved. At the same time we can continue to use the simplistic data for countries where we don't have better information.

For example, the data for Germany would look like this:

https://github.com/frappe/erpnext/blob/99c8b8efcfabf83b267bf6c106e5bba70675f75f/erpnext/setup/setup_wizard/data/country_wise_tax.json#L483-L737

You can see that it first distinguishes between the CoA used. Here we have different templates for the CoAs "SKR04 mit Kontonummern", "SKR03 mit Kontonummern", "Standard with Numbers" and all others ("*").

In the second level we can define different templates for sales, purchase and item taxes. The fields are named exactly like in the corresponding doctype. But there is one important difference: the link to **Account** (`account_head`, or `tax_type` in the case of Item Tax) is not  a string (because we don't know it) but a `dict` containing the necessary data to find or create the **Account**. Again, the keys for the **Account** are just the same as in the doctype.

The code has been refactored to use the new structure and reuse existing accounts.


barredterra:
Please merge #24819 first, as it solves some "Duplicate Entry" errors.

barredterra:
Regarding the failing test: `tax_rule.get_tax_template` returns the correct value, but I guess it gets overwritten by indian localization?

```
FAIL: test_taxes_fetch_via_tax_rule (erpnext.accounts.doctype.tax_rule.test_tax_rule.TestTaxRule)
  File "/home/runner/frappe-bench/apps/erpnext/erpnext/accounts/doctype/tax_rule/test_tax_rule.py", line 165, in test_taxes_fetch_via_tax_rule
    self.assertEqual(quotation.taxes_and_charges, "_Test Sales Taxes and Charges Template - _TC")
AssertionError: 'In State GST - _TC' != '_Test Sales Taxes and Charges Template - _TC'
- In State GST - _TC
+ _Test Sales Taxes and Charges Template - _TC
```

deepeshgarg007:
@barredterra I am unable to update the branch from my end, please update your branch once and push

barredterra:
Done, @deepeshgarg007

deepeshgarg007:
> Regarding the failing test: `tax_rule.get_tax_template` returns the correct value, but I guess it gets overwritten by indian localization?
> 
> ```
> FAIL: test_taxes_fetch_via_tax_rule (erpnext.accounts.doctype.tax_rule.test_tax_rule.TestTaxRule)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/frappe-bench/apps/erpnext/erpnext/accounts/doctype/tax_rule/test_tax_rule.py", line 165, in test_taxes_fetch_via_tax_rule
>     self.assertEqual(quotation.taxes_and_charges, "_Test Sales Taxes and Charges Template - _TC")
> AssertionError: 'In State GST - _TC' != '_Test Sales Taxes and Charges Template - _TC'
> - In State GST - _TC
> + _Test Sales Taxes and Charges Template - _TC
> ```

I'll look into this

erpnext/setup/setup_wizard/data/country_wise_tax.json
```diff
@@ -580,26 +824,137 @@
...
+									"tax_rate": 14.00
+								}
+							}
+						]
+					}
+				],
+				"*": [
+					{
+						"title": "In State GST",
+						"is_default": 1,
```
deepeshgarg007: @barredterra I guess "allow edits from maintainers" is not enabled for this PR hence I am not able to make any changes. Remove the `is_default` check from here to fix the test case.

erpnext/setup/setup_wizard/data/country_wise_tax.json
```diff
@@ -580,26 +824,137 @@
...
-		},
-		"VAT 14%": {
-			"account_name": "VAT 14%",
-			"tax_rate": 14.00
+		"chart_of_accounts": {
+			"*": {
+				"item_tax_templates": [
+					{
+						"title": "In State GST",
+						"is_default": 1,
```
deepeshgarg007: Item Tax Templates don't have any `is_default` property so this can be removed as well

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #24812, opened 2021-03-07
feat: Sales / Purchase Transaction Type based Item Tax

zerodiscount:
Issue: 
- ERPNext currently does not allow  separate account heads for taxes paid/collected under same category - for example - VAT Paid and VAT Collected are under same account and thus reporting totals separately requires sorting / filtering the ledger.
- Additionally, an item may require a different tax rate for purchases than for sales. Scenario - an item being imported from an overseas supplier may not have VAT on supplier invoice - VAT being charged by customs in another document (not on purchase invoice). Same item when sold to domestic customer may require to charge VAT on sales invoice. Thus requiring different item Purchase and Sales tax rates. Conversely exported items may have zero VAT taxes but have Purchase VAT,
- Having multiple items with different VAT / tax rates on a single invoice - requires that tax be applied to item total and NOT to Net Total as is the capability via Tax Rule (Sales / Purchase taxes templates only apply to Net total - same tax on all items).
-  Additionally most electronic invoice / EDI submissions require that any applicable item tax is included under item_tax_rate for that line item in the json - and not as a subsequent separate calculated tax table as is the case when using tax rules and linked sales / purchase tax templates.  These templates are not intended to / do not update the item_tax_rate field for the line item.

Solution:
- This PR modifies the item tax rate table  - allowing to add one more column - "transaction type" to the row. 
- image
- By default - transaction type is left at ALL - meaning current functionality of ERPNext is not altered - the same account head and item_tax_rate are applied to all transactions. 
- Additionally a choice is now offered - where transaction type can be set to Sales or Purchase (in addition to default ALL) - in which case that item_tax rate and account head are only applied for the selected Sales or Purchase transaction type. 
- Multiple taxes may be applied to item (additional rows) - this functionality is retained / not altered.
- Multiple tax categories continue to apply to item (additional rows in item doctype under tax). This functionality is NOT altered. This PR now allows for different rates and account heads to be applied for sales / purchases taxes for each tax category.

Use case COA setup: 
- By default - ERPNext functionality is left unchanged.
- By using separate account heads and separate rates for  Sales or Purchase Transactions - taxes collected or paid can be separated for each tax category - if needed. A Group account head for each tax category would still allow to show the net VAT to be settled at end of month  to Govt (for example). Monthly payable / totals report can be generated easily with separated listing of each tax paid and collected. A third account head under same Group account showing settlement payments to govt would allow for journal entry transactions to be separated from tax collected / paid from customers / suppliers. Thus via the ledgers of each account head - separate Registers for VAT paid, VAT collected and VAT settled can be maintained. Significant ledger clarity and rate customization can be achieved with this small change.


zerodiscount:
@deepeshgarg007 @rohitwaghchaure @nabinhait, can I get a review on this

deepeshgarg007 (commented):
I see various issues with this. This contains hardcoded transaction types. I feel this setting will conflict in cases there are multiple scenarios in one type of transaction. This adds one more level of complexity to the current tax configuration. We already have "Tax Category" tagging feature at various doc level. I feel those use cases you have mentioned can be handled using Tax Category in current implementation.

zerodiscount:
> I see various issues with this. This contains hardcoded transaction types. I feel this setting will conflict in cases there are multiple scenarios in one type of transaction. This adds one more level of complexity to the current tax configuration. We already have "Tax Category" tagging feature at various doc level. I feel those use cases you have mentioned can be handled using Tax Category in current implementation.

@deepeshgarg007 - thank you for the review. The current item tax setup even with different category does NOT allow you to apply a different tax rate for purchase and sales taxes - singularly without manual edit. Please check this. When you create an item tax - say 2 rows with 2 different categories using 2 different item tax templates - ALL categories of tax would apply for ALL transactions of that item. A purchase tax category would apply to a sales transaction in addition to the sales tax category. Please confirm if my understanding is incorrect.  Additionally we have maintained default ERP setup so most users need not change anything. Lastly - yes sales and purchase are hard coded - but for item transactions with external entities (where taxes usually apply) - these are generally the 2 transaction types. Additional transaction types may be added later. In any case since Default ERP setup  is maintained - having additional options hardcoded or not is not harmful.  Sincerely request that we reconsider and advise - especially if our understanding of tax category use is incorrect. Thanks.

zerodiscount:
> I see various issues with this. This contains hardcoded transaction types. I feel this setting will conflict in cases there are multiple scenarios in one type of transaction. This adds one more level of complexity to the current tax configuration. We already have "Tax Category" tagging feature at various doc level. I feel those use cases you have mentioned can be handled using Tax Category in current implementation.

@deepeshgarg007 - I have also been thinking about your suggestion of using Tax Category. First and foremost - you cannot use it as there is no filter on it to apply it to a transaction type. Additionally, philosophically - my understanding - Tax Category is meant to apply to different types of taxes such as VAT, Sales, Excise, Octroi, State Tax, City Tax, etc... Under these tax categories - an item may have a different rate for a different transaction type. So to use a different Tax Category for each item tax rate - would NOT be consistent.  In any case - once again - default is ALL - so ERPNext remains unchanged. This PR permits customization only if you seek it. Hope you agree - and please confirm if our philosophical understanding of Tax Category is misplaced.

deepeshgarg007:
> > I see various issues with this. This contains hardcoded transaction types. I feel this setting will conflict in cases there are multiple scenarios in one type of transaction. This adds one more level of complexity to the current tax configuration. We already have "Tax Category" tagging feature at various doc level. I feel those use cases you have mentioned can be handled using Tax Category in current implementation.
> 
> @deepeshgarg007 - I have also been thinking about your suggestion of using Tax Category. First and foremost - you cannot use it as there is no filter on it to apply it to a transaction type. Additionally, philosophically - my understanding - Tax Category is meant to apply to different types of taxes such as VAT, Sales, Excise, Octroi, State Tax, City Tax, etc... Under these tax categories - an item may have a different rate for a different transaction type. So to use a different Tax Category for each item tax rate - would NOT be consistent. In any case - once again - default is ALL - so ERPNext remains unchanged. This PR permits customization only if you seek it. Hope you agree - and please confirm if our philosophical understanding of Tax Category is misplaced.

You can use Item Tax Template in combination with Sales/Purchase Taxes and Charges template in case Sales and Purchase tax rates are different. Just disbale this settings in Accounts Settings and only accounts in Sales/Purchase Taxes and charges template will applied and other accounts in Item Tax Template will be ignored

Screenshot 2021-03-14 at 7 45 27 PM

zerodiscount:
> 
> You can use Item Tax Template in combination with Sales/Purchase Taxes and Charges template in case Sales and Purchase tax rates are different. Just disbale this settings in Accounts Settings and only accounts in Sales/Purchase Taxes and charges template will applied and other accounts in Item Tax Template will be ignored
>

Thank you for the suggestion. I tried using those templates as suggested. I am on the latest V13 beta. For our version of ERP,  Sales and Purchase Tax templates seem to only apply to Net Total or Previous row Total/Amount. They do not have option to apply to same row (item total). Consequently the same rate applies to all items on invoice and shipping charges (if used as an item). We tried this and can share a sample invoice if needed. Are we missing something - How to issue invoice for multiple items with different tax rates using these templates? Different classes of goods / shipping services, etc. commonly have different VAT rates worldwide. Sincerely will appreciate a  clarification in this regard. 

Many thanks for your continued support. Best regards.

Screenshot 2021-03-14 at 4 16 44 PM

liyakatalilal:
> Just disbale this settings in Accounts Settings and only accounts in Sales/Purchase Taxes and charges template will applied and other accounts in Item Tax Template will be ignored

Disabling Item Tax template to preceded over Sales/Purchase tax template will not work in the above scenario, because different items have a different rate of taxes in the Item Table of PI or SI.

@zerodiscount: I haven't tried using the Tax category along with the Tax template in the Item Tax table. Capturing a Tax category as 'Sales' & 'Purchase' at Item level and also using same categories in Sales/Purchase tax template. Create a validation check for matching the Tax Category then apply the corresponding rate from the Item Tax table in the transaction.

I haven't tested this but appears to be a simpler solution than your proposed solution.

Screenshot 2021-04-01 at 7 05 54 PM

nameduser0:
I agree there is a use-case for this, but you can use the Tax Category as a work-around. Here's my use-case:

In the UK (& EU I believe) we need to have separate Purchase and Sales VAT Accounts to cater for some rules called Reverse Charge. This basically means the same VAT is applied to both accounts to cancel each other out on the same transaction.
This means, at the moment, you need to define 2 sets (because we have 0%, 5% and 20% rates) of Item Tax Templates: Purchase and Sales.
If this feature were implemented, we could just define 1 set of templates and the Item Tax Template lines that got applied would depend on the transaction type.

As a work-around, you can define two sets of Tax Categories: Supplier and Customer, and ERPNext will pick the correct Item Tax Template from the Item/Item Group. It would also be nice if the Item/Item Group had a `Valid for` column to allow `Sale`, `Purchase` or `<blank>` to remove the Supplier/Customer Category sets, because VAT rules also change depending on Import/Export.

As a review comment, instead of using `All` I would just leave that option blank.

zerodiscount:
> > Just disbale this settings in Accounts Settings and only accounts in Sales/Purchase Taxes and charges template will applied and other accounts in Item Tax Template will be ignored
> 
> Disabling Item Tax template to preceded over Sales/Purchase tax template will not work in the above scenario, because different items have a different rate of taxes in the Item Table of PI or SI.

Hope it is settled that we cannot use sales and purchase tax templates - as they only apply to Net total for all items.

> 
> @zerodiscount: I haven't tried using the Tax category along with the Tax template in the Item Tax table. Capturing a Tax category as 'Sales' & 'Purchase' at Item level and also using same categories in Sales/Purchase tax template. Create a validation check for matching the Tax Category then apply the corresponding rate from the Item Tax table in the transaction.
> 
> I haven't tested this but appears to be a simpler solution than your proposed solution.

@LiyakatAliLal  - Agreed - we can use tax category to separate sales and purchase tax rates. Once an item tax rate is linked to tax category in item master - that item tax rate only applies on sales invoice if that tax category is selected in the invoice. Plus the invoice restricts you to 1 category. It means you have to create a category for each scenario - Sales+VAT+State vs. Purchase+ZeroVAT+ OtherImport..... Tax categories apply automatically when they are linked to a supplier or customer - which means now we have to go back and manually link a tax category for each customer/supplier in the database depending on scenario. If a tax changes - then we have to update the category and once again go through this process..... So the net result is that we have shifted complexity from Item Tax to Tax Category - as you will need twice the number of categories when there is a difference in item tax rates.

@casesolved-co-uk - agreed this PR is an easier way and perhaps more intuitive to separate tax account heads. Unsure if just creating tax categories would work by iteself - as you would need to also link those categories to customers and suppliers. In some instances like online shopping - where new customers are created on the fly - manual linking may not work. As regards keeping the default at ALL - it was chosen as it would imply that item tax rate  applies to ALL transaction types. Leaving it "Blank" would not be clear as to whether the item tax rate applies to any transaction type. ERPNext currently separates taxes into 2 categories Sales and Purchase (with tax templates). We could choose "BOTH" if the community wants... "ALL" is broader - and allows for future possibility of other transaction types (I cannot think of one right now).

@deepeshgarg007 - this PR does not add to complexity - it keeps default behavior at ALL - it allows for flexibility. Seems there are many use cases out there that may need this or find this as an easier and more intuitive way to fix their tax setup requirements.

leonardamsler:
Hello, 
I find myself facing the very same issue. Prior to stumbling upon this issue, I shared the belief that @zerodiscount 's solution was the most suitable one, and I was about to initiate a new issue. Could you please inform me if there have been any updates concerning this matter?

leonardamsler:
I have tried using what @zerodiscount has coded on my instance and indeed the apparence of the item tax templates is as expected but everything utilizing that field "Transaction Type" remains to be done, for example the tax logic for Sales/Purchase Orders.

Also, the creation of Sales Order is also completely bugged now (see screen shot).

Have you done the whole functionality implementation?

Item tax template
Bugged Sales Orders

PR #24815, opened 2021-03-08
fix: breadcrumbs for Company

walstanb:
### Issue
Breadcrumbs for Company were using an older route.

### Fix
Added Breadcrumbs to company of module Accounts

**Before**

image
image

**After**

image


icnbing:
image

This is my case. I have the same problem. Newly installed.

PR #24819, opened 2021-03-08
fix: autoname for Item Tax Template

barredterra:
Previously, autoname of **Item Tax Template** was set to `field:title`. Therefore the title had to be unique across companies. This conflicts, for example, with `taxes_setup.py` which assumes that title is only unique within a company.

This problem is already solved in **Sales Taxes and Charges Template** and **Purchase Taxes and Charges Template**. They use the `doc.autoname()` method to add the company abbreviation:

https://github.com/frappe/erpnext/blob/98cc7b175196d623ab17dcf805755dbc5554aee3/erpnext/accounts/doctype/sales_taxes_and_charges_template/sales_taxes_and_charges_template.py#L15-L18

https://github.com/frappe/erpnext/blob/98cc7b175196d623ab17dcf805755dbc5554aee3/erpnext/accounts/doctype/purchase_taxes_and_charges_template/purchase_taxes_and_charges_template.py#L15-L18

This PR adds the same `autoname` method to **Item Tax Template**. A patch is not technically needed, only if we want all **Item Tax Templates** to be named consistently.

Test records of **Item** are modified to add the abbreviation of the test Company (`_TC`).


deepeshgarg007 (commented):
@barredterra tests are failing. Can you fix those?

deepeshgarg007 (approved):

PR #24785, opened 2021-03-02
fix: duplication of earnings row on saving Salary Slip

walstanb:
**Issue:** 
Duplicate rows were being added to the Earnings table on saving Salary Slip if the additional salary component had the field **Overwrite Salary Structure Amount** unchecked. See linked issue for more details.

Resolves #24684 

Peek 2021-03-02 14-04

**Fix:**
Removed condition that caused a new row to be added incorrectly to Salary Slip every time it was re-saved. Due to this condition, following code in `else` block was always unreachable:

```python
if overwrite:
# ---
else:
	component_row.additional_amount = amount

if not overwrite and component_row.default_amount:
	amount += component_row.default_amount
```

PR #24824, opened 2021-03-09
refactor(payroll): simplified logic for additional salary

### LINKED ISSUE #24684: Duplicate Additional Salary when saving Salary Slip
## Description of the issue
Duplicate additional salaries on saving the salary slip where additional salary is without "Overwrite Salary Structure Amount"

**Output of `bench version`**
```
ERPNext: v13.0.0-beta.11 (HEAD)
Frappe Framework: v13.0.0-beta.10 (HEAD)
```

1. Create Holiday List for the current period
2. Assign default holiday list to the company
3. Create Salary Structure 
4. Create Employee
5. Assign a salary structure to the employee
6. Create Salary component "Incentive" as earnings
7. Create additional salary "Incentive" and uncheck "Overwrite Salary Structure Amount"
8. Create a salary slip for the employee and save it

### Observed result
In earnings, you will see 2 records for Incentive

Incentive		TZS 600.00
Incentive		TZS 600.00

### Expected result
Only 1 record should appear

Incentive		TZS 600.00

walstanb:
### Issue Fixed
Duplicate rows were being added to the `earnings` table on saving Salary Slip if the additional salary had the field **Overwrite Salary Structure Amount** unchecked. Resolves #24684.

### New Behavior
Additional salaries are now treated separately from _default_ salaries, although they may refer to the same salary component. 
For example: if there is a fixed transit allowance and an additional allowance for transit is applicable for the current payroll period, the same will reflect as an additional row in Salary Slip `earnings` table. This is useful to differentiate properly. 

It is also a standard practice to sometimes have multiple additional salaries referring to the same component (e.g. deadline incentive). The new change allows for each of these incentives to be shown separately. The new change also allows proper linking of each earnings row with their `Additional Salary` document. 

In a future release, the fields `default_amount` and `additional_amount` can be deprecated as they will now always equal `amount`. 

An exception to this behaviour is Additional Salaries with  **Overwrite Salary Structure Amount** checked. If these exist, all _default_ rows of the same component will be removed.

Note that existing Salary Slips will remain as-is.

#### additional_salary.py

`get_additional_salaries` function: 
- Renamed from earlier name: `get_additional_salary_component`
- No longer whitelisted, as it wasn't being accessed from frontend.
- No longer returns `salary_components_details`, this is now being fetched in `salary_slip.py`.

**salary_slip.py**
- Replaced spaces with tabs.
- `add_additional_salary_components` updated for changes made in other functions.
- `get_salary_slip_row` renamed to `get_salary_component_data`, optimised to use `frappe.get_value` instead of `frappe.get_doc`.

`update component_row` function:
- Changed parameter names to be more logical.
- Removed overwrite parameter as `additional_salary` parameter now gets a `dict` object containing relevant data.
- Updated logic to find `component_row`.
- If an Additional Salary with `overwrite` checked is being updated, _default_ rows of the same component get removed.
- Removed condition that caused a new row to be added incorrectly to Salary Slip every time it was re-saved.
- Cleaner updation / creation of component row.

### TL;DR
- Additional salary components now have their own row.
- Incorrect duplication of rows on <kbd>Save</kbd> fixed.
- Code refactored to make it simpler to understand.

---

**Before**

Peek 2021-03-02 14-04

**After**

salary slip after

Also tested for other cases mentioned here: https://github.com/frappe/erpnext/pull/24785#issuecomment-792621048

### DIFF
--- a/erpnext/payroll/doctype/additional_salary/additional_salary.py
+++ b/erpnext/payroll/doctype/additional_salary/additional_salary.py
@@ -89,10 +89,11 @@ class AdditionalSalary(Document):
 		no_of_days = date_diff(getdate(end_date), getdate(start_date)) + 1
 		return amount_per_day * no_of_days
 
-@frappe.whitelist()
-def get_additional_salary_component(employee, start_date, end_date, component_type):
-	additional_salaries = frappe.db.sql("""
-		select name, salary_component, type, amount, overwrite_salary_structure_amount, deduct_full_tax_on_selected_payroll_date
+def get_additional_salaries(employee, start_date, end_date, component_type):
+	additional_salary_list = frappe.db.sql("""
+		select name, salary_component as component, type, amount,
+		overwrite_salary_structure_amount as overwrite,
+		deduct_full_tax_on_selected_payroll_date
 		from `tabAdditional Salary`
 		where employee=%(employee)s
 			and docstatus = 1
@@ -102,7 +103,7 @@ def get_additional_salary_component(employee, start_date, end_date, component_ty
 					from_date <= %(to_date)s and to_date >= %(to_date)s
 				)
 		and type = %(component_type)s
-		order by salary_component, overwrite_salary_structure_amount DESC
+		order by salary_component, overwrite ASC
 	""", {
 		'employee': employee,
 		'from_date': start_date,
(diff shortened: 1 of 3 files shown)


sagarvora:
Please note that the **Frappe Linter / Translation Syntax Check** fail is a false-negative, as the regex being tested fails for strings broken into multiple lines.

sagarvora:
Similar to another open PR: https://github.com/frappe/erpnext/pull/24460

Anurag810:
awesome work @sagargharge and @walstanb, will test and merge...

Anurag810:
@sagarvora and @walstanb 

I was testing your PR locally and found one edge case missing which was before there.

Excepted behavior(Which is still not covered):

1. "Overwrite Salary Structure Amount" property checked will overwrite salary structure amount.
2. "Overwrite Salary Structure Amount" property is not checked will create a new row with additional salary reference.
3. recurring additional salary will have the same behavior as mentioned in the second point.

i have created three additional salary(refer screen shoot):
image

After that, I have created a salary slip for that:
it shows me only one row(refer to the screenshot).
image


Please check and revert back

sagarvora:
<s>
My understanding was the if an additional salary exists with **overwrite** checked, other Additional Salaries of the same component are to be ignored. Will fix it so that only the Salary Structure amount gets overwritten.
</s>
<br><br>
Fixed.

Anurag810:
@sagarvora  Tested it again and found that the non-overwrite component is still not present, please refer to the screenshot.

image

we need to put both recurring as well as a non-overwrite component in the components row.

image

sagarvora:
Anurag, the changes I made were force-pushed. Are you sure you force-pulled and checked if the branch is correct? It's working perfectly for me:

Salary Structure:

Screenshot-2021-03-15-112538

Additional Salary:

Screenshot-2021-03-15-112749

Salary Slip:

Screenshot-2021-03-15-112229

Anurag810:
Sorry, my bad. I checked it once again and it is working fine now. waiting for the green tick now.

sagarvora:
-  Translation Syntax Check is a false-negative as mentioned above.
- Travis fails are unrelated.

Anurag810 (approved):

Anurag810:
@sagarvora please also create one for V13

sagarvora:

OneiricIT:
@walstanb we have come across an issue where if the Salary Structure is using components in the formula, and there are multiple additional salary added (set to not overwrite) for the same component, the formula only uses the last additional salary row in the salary slip earning or deduction table (does not accumulate all the additional salary with same components into a singular value).

For example:

- Salary Structure Earning component has a formula being '((base / 52) * 2) - (PL)'. 
- ((base / 52) * 2) = $2000.
- 2x Additional Salary are added for component PL for the payroll period - each is $250, so total of PL combined is $500.
- We see the three rows within the Earning table, however the Salary value is $1750 as opposed to $1500 per below.

image

We have tried a few variations within the formula to combine the PL component without success. Any comment on this?

Thanks

PR #24831, opened 2021-03-10
fix: fix Payment references from disappearing on adding Cost Center in Payment Entry

Anuja-pawar:
Issue: When you create a Payment Entry manually, the Payment reference gets cleared on adding Cost Center in the Payment Entry.

Before:

before_PEfix

After:

after_PEfix

-> Fixed the issue and also added a cost center filter to the 'Get Outstanding Invoice' pop-up.


AfshanKhan (dismissed):

AfshanKhan:
@Anuja-pawar also make this PR for the `version-13-beta-pre-release` branch

PR #24836, opened 2021-03-10
fix: Unequal debit and credit issue on RCM Invoice

deepeshgarg007:


nabinhait (dismissed):

nabinhait:
@deepeshgarg007 Travis failed

erpnext/regional/india/utils.py
```diff
@@ -799,23 +791,20 @@ def make_regional_gl_entries(gl_entries, doc):
 
 	return gl_entries
 
-@frappe.whitelist()
-def get_regional_round_off_accounts(company, account_list):
```
marination, line 803: remove from `hooks.py` too

PR #24842, opened 2021-03-10
fix(payroll): allow duplicate additional salaries

sagarvora:
As discussed in #24824, it can be useful to have multiple additional salaries created separately:

> Take **Deadline Incentive** for example. It may be desirable to show the same component with possibly the same amount multiple times in the Salary Slip. One for each deadline achieved.

---
In any case, this logic shouldn't be in the `before_insert` method as one can easily circumvent it by saving with a different amount initially and then re-saving with the desired amount.


Anurag810 (commented):
approved

Issue #20791, opened 2020-03-02
Pricing Rule of Adding Margin isn't working


raymondliew1992:
Kindly refer to the link below:
https://discuss.erpnext.com/t/pricing-rule-add-margin-not-working/58664

ERPNext: v12.4.2 (version-12)
Frappe Framework: v12.2.1 (version-12)

dj12djdjs:
It seems this is still a problem when it comes to applying multiple price rules to a margin.
https://discuss.erpnext.com/t/pricing-rule-w-apply-multiple-pricing-rules-margin/72118

PR #24844, opened 2021-03-10
fix: multiple price rules margin.

dj12djdjs:
Multiple pricing rules are only applied to discount amount or percentage and disregards margin amount and percentage noted in the this issue #20791 
I have some further information at this [discussion.](https://discuss.erpnext.com/t/pricing-rule-w-apply-multiple-pricing-rules-margin/72118)

This PR applies the same logic to margin as already existed for discount rate.


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -330,13 +330,16 @@ def get_pricing_rule_details(args, pricing_rule):
...
-		item_details.margin_rate_or_amount = 0.0
+	for apply_on in ['Percentage', 'Amount']:
+		if pricing_rule.margin_type != apply_on: continue
+
+		field = 'margin_rate_or_amount'
+		if field not in item_details:
+                        item_details.setdefault(field, 0)
+                        item_details.setdefault('margin_type', apply_on)
+
+		item_details[field] += (pricing_rule.get(field, 0)
```
barredterra, line 342: How can I understand the `+=`? Margins from multiple price rules get added together? This changes the previous behaviour, right?

barredterra (changes requested):

erpnext/accounts/doctype/pricing_rule/pricing_rule.py
dj12djdjs: Yes, this does alter the previous behavior of pricing rules.

So, the new behavior is just copying what is being applied to discount from multiple price rules.
https://github.com/frappe/erpnext/blob/c6a1e3b450e3ce585da374fe41d041bae68cfcc5/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L356-L368

Previously, when multiple pricing rules are applied only the last pricing rules margin would be applied.
https://github.com/frappe/erpnext/blob/c6a1e3b450e3ce585da374fe41d041bae68cfcc5/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L333-L339

[Here](https://discuss.erpnext.com/t/pricing-rule-w-apply-multiple-pricing-rules-margin/72118) you can see the behavior previous to this patch in action vs what I was expecting.

barredterra: Okay, makes sense. Thanks for clarifying.

dj12djdjs:
Are we switching coding style to use spaces instead of tabs? I keep getting Sider problems that indentation is using tabs.

barredterra (commented):
Sorry to request changes again, but maybe consider adding a test to ensure your desired behavior works and will continue to work in the future. As an example, see this test case: https://github.com/frappe/erpnext/blob/87dddbe868e0aa57bee141ae94986b6bbdd7abaf/erpnext/accounts/doctype/pricing_rule/test_pricing_rule.py#L103-L148

nabinhait (approved):

nabinhait:
@dj12djdjs Can you please send another PR to the develop branch?

dj12djdjs:
@nabinhait @barredterra 

Yes I will redo on develop with some tests.

PR #24852, opened 2021-03-11
feat: configurable action if the same purchase/selling rate is not maintained

ruchamahabal:
There is a checkbox in Buying and Selling Settings to ensure that the same rate is maintained in the Purchase and Sales transactions respectively. However, sometimes user wants to allow different rates. In such cases, they disable this checkbox and continue the transactions but they might forget to enable it again.

Now the action to be taken if this rate is not maintained is configurable (Stop / Warn):

maintain-same-rate

**Stop**:

`throw` - Stops the transaction

image

**Warn**:

`msgprint` - Let's you save

image

A similar configuration has been implemented in Selling Settings too.

Documentation PR: https://github.com/frappe/erpnext_documentation/pull/258


nabinhait (approved):

ruchamahabal:
Closing this since the commit has been cherry-picked in https://github.com/frappe/erpnext/pull/25193

PR #24861, opened 2021-03-12
feat: provision to make auto journal entry if there is difference between stock and account value

rohitwaghchaure:
**Issue**

Screenshot 2021-03-12 at 12 04 01 PM

If the stock and account value has the difference then system throws the above validation and force user to make the journal entry. Once user create the Journal Entry then system allow user to complete the transaction. With the new changes we have added provision to make the JV automatically if there is difference between stock and account value.

Screenshot 2021-03-12 at 12 08 52 PM

**Report**

Added report to know which are the incorrect stock value transactions

Screenshot 2021-03-12 at 12 10 25 PM


olamide2:
This is a pretty good idea

PR #24862, opened 2021-03-12
fix: added correct path in hooks

AfshanKhan:
For loan processing added correct paths in hooks.py


deepeshgarg007:
@AfshanKhan send this on v13-pre-release as well

AfshanKhan:
> @AfshanKhan send this on v13-pre-release as well

Backport at https://github.com/frappe/erpnext/pull/24865

PR #24863, opened 2021-03-12
fix: PO not created against all selected suppliers (drop shipping)

marination:
**Issue:**
- Create 3 Items with drop shipping checkbox checked.
- Create Sales order with this the above 3 items created above. Now for each item row select different supplier for drop shipping and submit SO
- Now create Purchase order and in the Pop-up select "Against Default Supplier" > Then it will create PO only for the 1st item row. However, it should create 3 POs as per the different suppliers selected in the SO.
drop-ship-bug

**Fix:**
- Don't return doc while looping over suppliers and creating POs, 
- Return list of created PO doc objects after loop
- Added test case for coverage. **This case** shouldn't break again.
drop-ship-fix


marination:
Screenshot 2021-03-31 at 3 02 12 PM

PR #24878, opened 2021-03-15
feat: add total available stock field in PO

anupamvs:
Add current total available inventory/stock field in PO Item table

image


rohitwaghchaure (approved):

Olawale1:
Great! Could you pls add this in Material Request and Stock Entry too ?

Kind regards,

PR #24903, opened 2021-03-16
test: set correct qty when creating subcontract transfer entry

sagarvora:
Since only 10 units are received, but 20 are used, it was failing.


rohitwaghchaure:
@sagarvora Can you change the qty from 10 to 20 at line no 725, this is not a transferred entry. This is a material receipt entry.

PR #24904, opened 2021-03-16
ci: use ubuntu 18 specifically

hasnain2808:
Use ubuntu 18 instead of latest to ensure we get the exact same environment each time


sagarvora:
Why not `20.04`?

sagarvora (approved):

PR #24910, opened 2021-03-17
fix: copy_parent_value_in_all_row function error

AndyOverLord:

This utility function is not being used in any where. But I found it's useful for a custom feature I would like to achive, which is to create a custom button to update all the expected_delivery_date in purchase order items table. 

I replaced fieldname to parent_fieldname as fieldname is not necessary exist for my use case.
By fixing the issue, I can use schedule_date as parent field to update all the child table items' expected_delivery_date.

> Screenshots/GIFs


erpnext/public/js/utils.js
```diff
@@ -293,7 +293,7 @@ $.extend(erpnext.utils, {
 	},
 	copy_parent_value_in_all_row: function(doc, dt, dn, table_fieldname, fieldname, parent_fieldname) {
 		var d = locals[dt][dn];
-		if(d[fieldname]){
+		if(d[parent_fieldname]){
```
marination, line 296: - Since this is a utility function, all logic needs to be generic (that serves everyone). 
- `parent_fieldname` and `fieldname` are both critical to this function.
- The check is to make sure that someone does not accidentally pass a non-existent fieldname, mistyped fieldname or no fieldname at all. So you could check for both these fields since they are required.
```suggestion
		if(d[parent_fieldname] && d[fieldname]){
```

marination: - `parent_fieldname` must pertain to a field that is not in the table, and is in the outer form. 
- `d` here contains the child row, which will only have `fieldname`.

Your fix is checking for the parent field in the child row, which is incorrect. Am I missing something ?

AndyOverLord: Hi @marination , Thanks for your reply.

I'm not very sure what's the original developer's purpose by writing the `copy_parent_value_in_all_row`. But if we take a look of its sibling function `copy_value_in_all_rows` on line 153 same file:

image

It assumes that `field_name` exists both on outer form and child table. And it only pastes the value in if `d[field_name]` has value.

What I want to achive by using `copy_parent_value_in_all_row` is that it can take outer form's field value and overrides to every rows in child table.

My use case in custom script:
image

In my case, since all purchase orders submitted with schedule_date in outer form will use `copy_value__in_all_rows` to add schedule_date in items table. We don't use schedule date field that much, but users would like to set `expected_delivery_date` on items table, which is a field not existing in outer form.

Hence I have modified this change to achieve my goal which is to prompt user to update an ETA which overrides `schedule_date` in outer form and all child table rows plus set the value in "expected_delivery_date" in all child table rows as well.

I would suggest to rename this function as **overrides_parent_value_in_all_rows** and then remove the if statement validation, just loop through each item and overrides the `parent_fieldname` to `d[field_name]`.

Sorry I realized my reply might be too long and probably doesn't make sense.
Please leave any your questions or feedbacks. Thank you! 😊

AndyOverLord (commented):
Hi there, please review my new commit changes. Thanks

erpnext/public/js/utils.js
marination: @AndyOverLord Makes sense. Just some poorly defined utilities. Will try to clean up the rest later. Your changes look fine.

marination (dismissed):

marination:
@AndyOverLord please  fix sider issues and send this to `version-12-hotfix` as well

erpnext/public/js/utils.js
```diff
@@ -291,17 +291,16 @@ $.extend(erpnext.utils, {
...
-	copy_parent_value_in_all_row: function(doc, dt, dn, table_fieldname, fieldname, parent_fieldname) {
-		var d = locals[dt][dn];
-		if(d[fieldname]){
-			var cl = doc[table_fieldname] || [];
-			for(var i = 0; i < cl.length; i++) {
+	overrides_parent_value_in_all_rows: function(doc, dt, dn, table_fieldname, fieldname, parent_fieldname) {
+		let d = locals[dt][dn];
+		if(doc[parent_fieldname]){
+			let cl = doc[table_fieldname] || [];
+			for(let i = 0; i < cl.length; i++) {
```
marination:
```suggestion
		if (doc[parent_fieldname]) {
			let cl = doc[table_fieldname] || [];
			for (let i = 0; i < cl.length; i++) {
```

marination (approved):

PR #24913, opened 2021-03-17
ci: get branch name from GITHUB_REF if GITHUB_BASE_REF not available

sagarvora:
The environment variable GITHUB_BASE_REF is only available for `pull_request` events. Consequently, `workflow_dispatch` and `push` events [fail during install](https://github.com/frappe/erpnext/actions/runs/659940318).

This PR gets the branch name from GITHUB_REF if GITHUB_BASE_REF is unavailable. This can work, as GITHUB_REF refers to the correct branches for `workflow_dispatch` and `push` events.

Further reading:
- https://docs.github.com/en/actions/reference/events-that-trigger-workflows
- https://docs.github.com/en/actions/reference/environment-variables
- https://stackoverflow.com/a/39296572
- https://stackoverflow.com/a/58034787

---

**Local Testing** (ASD available, ASDF not)

Screenshot-2021-03-17-111816

### DIFF
--- a/.github/helper/install.sh
+++ b/.github/helper/install.sh
@@ -12,7 +12,7 @@ sudo apt install npm
 
 pip install frappe-bench
 
-git clone https://github.com/frappe/frappe --branch "${GITHUB_BASE_REF}" --depth 1
+git clone https://github.com/frappe/frappe --branch "${GITHUB_BASE_REF:-${GITHUB_REF##*/}}" --depth 1
 bench init --skip-assets --frappe-path ~/frappe --python "$(which python)" frappe-bench
 
 mkdir ~/frappe-bench/sites/test_site


hasnain2808:
@sagarvora 
that reference is used to get the corresponding frappe branch to run the tests

image
The reference in GITHUB_REF may not be present in frappe, which would break the workflow

sagarvora:
@hasnain2808 You're right. Which is why, I haven't use `GITHUB_REF` as-is. I am getting the branch from this value.

hasnain2808:
> @hasnain2808 You're right. Which is why, I haven't use `GITHUB_REF` as-is. I am getting the branch from this value.

Yes I got that
The real issue is that the branch name you will get from it is the branch from which pr was raised. It may or may not be present in the frappe repository.

sagarvora:
> > @hasnain2808 You're right. Which is why, I haven't use `GITHUB_REF` as-is. I am getting the branch from this value.
> 
> Yes I got that
> The real issue is that the branch name you will get from it is the branch from which pr was raised. It may or may not be present in the frappe repository.

Yes, but this is the assumption on which it has been working previously as well. This is what `GITHUB_BASE_REF` gives you as well.

hasnain2808:
> > > @hasnain2808 You're right. Which is why, I haven't use `GITHUB_REF` as-is. I am getting the branch from this value.
> > 
> > Yes I got that
> > The real issue is that the branch name you will get from it is the branch from which pr was raised. It may or may not be present in the frappe repository.
> 
> Yes, but this is the assumption on which it has been working previously as well. This is what `GITHUB_BASE_REF` gives you as well.

GitHub base ref gives the branch against which the pr was raised. The branch in which the pr will be merged
This is how even Travis worked

sagarvora:
> GitHub base ref gives the branch on which the pr was raised.

Okay, now I understood what you mean. But it is not the case. Here's how it works:

- `pull_request` event: GITHUB_REF is wrong (PR branch), but GITHUB_BASE_REF is present so all is well.
- [push](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#push) event: GITHUB_REF is the "updated ref", meaning the branch that was updated.
- [workflow_dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch) event: GITHUB_REF is the "Branch that received dispatch".

sagarvora:
There's an alternate option as well. We can get the branch by running a command in the cloned repo and store it as an env var for further use.

But my current solution works without needing to run any additional command.

PR #24921, opened 2021-03-17
fix: precision and formatted document for stock level in item dashboard.

Anurag810:
image


erpnext/stock/dashboard/item_dashboard.py
```diff
@@ -42,11 +44,20 @@ def get_data(item_code=None, warehouse=None, item_group=None,
 		limit_start=start,
 		limit_page_length='21')
 
+	precision = get_field_precision(frappe.get_meta("Stock Ledger Entry").get_field("stock_value"))
```
marination: Lets use the system precision only ? Even the Stock Ledger Report uses system precision itself, not many people care about or set precision at the Stock Ledger Entry level, since in other places this field isnt formatted with field precision. For consistency.
```suggestion
	precision = cint(frappe.db.get_single_value("System Settings", "float_precision"))
```

Anurag810: @Marination Working...

erpnext/stock/dashboard/item_dashboard.py
```diff
@@ -2,6 +2,8 @@
 
 import frappe
 from frappe.model.db_query import DatabaseQuery
+from frappe.model.meta import get_field_precision
+from frappe.utils import flt
```
marination:
```suggestion
from frappe.utils import flt, cint
```

marination (approved):

PR #24927, opened 2021-03-17
feat: add options to disable sending welcome email and allow enrolmen…

madar2020:
By default, Welcome email is sent for every new Student. There no way currently to disable this. Thus I have provided an option in the applicant doctype, if disabled, no welcome email will be send when a new Student is created.
The other feature added is that enrolment for existed user through applicant throws an error as it tries to create a new user. This option is now added as it automatically.

no-docs


hasnain2808:
It's preferable that such settings are editable at the document/ transaction level. It's easier to verify and debug.

https://github.com/frappe/erpnext/wiki/Model-Design-Guidelines#make-contextual-settings

madar2020:
I have converted the option of send email from local to global

erpnext/education/doctype/student_applicant/student_applicant.json
```diff
@@ -260,16 +260,17 @@
...
   {
-   "fieldname": "student_category",
-   "fieldtype": "Link",
-   "label": "Student Category",
-   "options": "Student Category"
+   "default": "1",
+   "description": "By default, Welcome email is sent for every new Student. If disabled, no welcome email will be send when a new Student is created.",
+   "fieldname": "send_welcome_email",
+   "fieldtype": "Check",
+   "label": "Send welcome email for new Applicants"
```
hasnain2808:
```suggestion
   "label": "Send welcome email"
```

erpnext/education/doctype/student_applicant/student_applicant.json
```diff
@@ -260,16 +260,17 @@
    "read_only": 1
   },
   {
-   "fieldname": "student_category",
-   "fieldtype": "Link",
-   "label": "Student Category",
-   "options": "Student Category"
+   "default": "1",
+   "description": "By default, Welcome email is sent for every new Student. If disabled, no welcome email will be send when a new Student is created.",
```
hasnain2808: Label looks self-explanatory to me
```suggestion
   
```

erpnext/education/doctype/student/student.py
```diff
@@ -52,7 +52,7 @@ def update_student_name_in_linked_doctype(self):
 	def check_unique(self):
 		"""Validates if the Student Applicant is Unique"""
 		student = frappe.db.sql("select name from `tabStudent` where student_applicant=%s and name!=%s", (self.student_applicant, self.name))
-		if student:
+		if student and student.student_email_id != frappe.session.user.email:
```
hasnain2808, line 55: Can you please add some comments to explain the reason for adding the condition

erpnext/education/doctype/education_settings/education_settings.json
```diff
@@ -1,149 +1,149 @@
...
-  }
- ],
- "quick_entry": 1,
- "restrict_to_domain": "Education",
- "sort_field": "modified",
- "sort_order": "DESC",
- "track_changes": 1
-}
\ No newline at end of file
+    "actions": [],
```
hasnain2808: I don't think this file should be committed since it does not have any relevant changes.

hasnain2808 (changes requested):

erpnext/education/doctype/student/student.py
madar2020: this is to avoid throwing the error id the user is already registered and logined. user should be able to apply for different courses.

madar2020 (commented):
changes done

erpnext/education/doctype/education_settings/education_settings.json
madar2020: yep

erpnext/education/doctype/student/student.py
```diff
@@ -68,7 +68,7 @@ def create_student_user(self):
 				'last_name': self.last_name,
 				'email': self.student_email_id,
 				'gender': self.gender,
-				'send_welcome_email': 1,
+				'send_welcome_email': self.send_welcome_email,
```
hasnain2808, line 71: `send_welcome_email` is property of student applicant so unaccesible here
Tests are failing too
https://github.com/frappe/erpnext/runs/2211410243?check_suite_focus=true#step:11:68

erpnext/education/doctype/student_applicant/student_applicant.json
```diff
@@ -11,7 +11,7 @@
   "middle_name",
   "last_name",
   "program",
-  "student_category",
+  "send_welcome_email",
```
hasnain2808, line 14: should this go in the student doctype?

hasnain2808 (changes requested):

PR #24944, opened 2021-03-18
feat: Project Profitability Report

pateljannat:
The profitability report shows the Profitability and Utilization of each employee based on the Timesheets generated.

This report shows the Profitability with the following data:

- Timesheet
- Salary Slip generated using the Timesheet
- Sales Invoice generated using the Timesheet

profitability-report

**Documentation PR:** https://github.com/frappe/erpnext_documentation/pull/266


Anurag810 (changes requested):
Update description with use-cases and also add some screenshots which will help...

erpnext/hr/doctype/hr_settings/hr_settings.json
```diff
@@ -143,13 +144,19 @@
    "fieldname": "send_leave_notification",
    "fieldtype": "Check",
    "label": "Send Leave Notification"
+  },
+  {
+   "default": "8",
+   "fieldname": "default_working_hours",
```
Anurag810: I think **Standard Working Hours** will sound Better.

Add a checkbox **Fetch Standard Working Hours From Shift Type**.
if unchecked, Show this field to capture **Standard Working Hours**.

erpnext/projects/report/profitability/profitability.json
```diff
@@ -0,0 +1,39 @@
...
+ "prepared_report": 0,
+ "ref_doctype": "Timesheet",
+ "report_name": "Profitability",
+ "report_type": "Script Report",
+ "roles": [
+  {
+   "role": "Projects User"
+  },
+  {
+   "role": "HR User"
```
Anurag810: Add HR Manager too...

erpnext/hr/doctype/hr_settings/hr_settings.json
ruchamahabal: +1 for Standard Working Hours. 

However, IMO we should not have this checkbox. If timesheets having different **Start** and **End Date**  are filled up, calculating the exact shift hours would become very complex for such reports, not just from the development perspective, but even for the user to understand the calculations. Moreover, shifts are for defining the start and end times but the total hours are more or less constant. Hence, I think calculating the metrics based on Standard Working Hours seems more feasible so we should just keep **Standard Working Hours** here.

erpnext/projects/report/profitability/profitability.json
```diff
@@ -0,0 +1,39 @@
...
+ "is_standard": "Yes",
+ "json": "{}",
+ "modified": "2021-03-18 10:20:15.559305",
+ "modified_by": "Administrator",
+ "module": "Projects",
+ "name": "Profitability",
+ "owner": "Administrator",
+ "prepared_report": 0,
+ "ref_doctype": "Timesheet",
+ "report_name": "Profitability",
```
ruchamahabal:
```suggestion
 "report_name": "Project Profitability",
```

1. Rename report to Project Profitability since it is with respect to customer-wise timesheets, etc.
Update wherever necessary.
2. Also add a Project column and filter.

erpnext/projects/report/profitability/profitability.py
```diff
@@ -0,0 +1,162 @@
...
+		return None
+
+	labels = []
+	utilization = []
+
+	for entry in data:
+		labels.append(entry.get("title") + " - " + str(entry.get("end_date")))
+		utilization.append(entry.get("utilization"))
+	charts = {
+		'data': {
```
ruchamahabal: Try to keep uniform quote styles throughout a file

erpnext/projects/report/profitability/profitability.py
```diff
@@ -0,0 +1,162 @@
...
+		},
+		{
+			"fieldname": "profit",
+			"label": _("Profit"),
+			"fieldtype": "Currency",
+			"options": "currency",
+			"width": 120
+		},
+		{
+			"fieldname": "end_date",
```
ruchamahabal: Add Start Date for the Timesheet.

Suggesting a different order for the columns:

- Customer
- Employee
- Employee Name
- Sales Invoice
- Timesheet
- Bill Amount
- Cost
- Profit
- Utilization
- Fractional Cost
- Total Billed Hours
- Start Date
- End Date

erpnext/projects/report/profitability/profitability.py
```diff
@@ -0,0 +1,162 @@
...
+	return columns, data, None, charts
+
+def get_data(filters):
+	conditions = get_conditions(filters)
+	default_working_hours = frappe.db.get_single_value("HR Settings", "default_working_hours")
+	sql = """ 
+			select 
+				*, 
+				t.gross_pay * t.utilization as fractional_cost, 
+				t.grand_total - t.gross_pay * t.utilization as profit
```
ruchamahabal: Let's separate out the calculations in a separate method. Always better to do so when the calculation can change based on some conditions. 

For example:
- If the employee currency is not the same as the company currency, you will have to take into consideration `base_gross_pay` instead of `gross_pay`.
- Similarly, if sales invoice currency is not the same as company currency, you will have to consider `base_grand_total` instead of `grand_total`

So, just fetch the data in this function and then calculate the metrics by handling these cases in another function. Also, write multiline SQL with better formatting. Having conditions on separate lines, writing keywords in all caps makes it easier to read.

image

image

erpnext/projects/report/profitability/profitability.py
```diff
@@ -0,0 +1,162 @@
...
+			"fieldtype": "Data",
+			"width": 120
+		},
+		{
+			"fieldname": "employee",
+			"label": _("Employee"),
+			"fieldtype": "Link",
+			"options": "Employee",
+			"width": 150
+		},
```
ruchamahabal:
```suggestion
		{
			"fieldname": "employee",
			"label": _("Employee"),
			"fieldtype": "Link",
			"options": "Employee",
			"width": 120
		},
		{
			"fieldname": "employee_name",
			"label": _("Employee Name"),
			"fieldtype": "Data",
			"width": 120
		},
```

Employee column should be shown first then Employee Name

ruchamahabal (changes requested):

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,239 @@
...
+					tabTimesheet.title as employee_name,tabTimesheet.parent_project as project,
+					tabTimesheet.start_date,tabTimesheet.end_date,
+					tabTimesheet.total_billed_hours,tabTimesheet.name as timesheet,
+					ss.base_gross_pay,ss.total_working_days,
+					tabTimesheet.total_billed_hours/(ss.total_working_days * {0}) as utilization
+					FROM
+						`tabSalary Slip Timesheet` as sst join `tabTimesheet` on tabTimesheet.name = sst.time_sheet
+						join `tabSales Invoice Timesheet` as sit on sit.time_sheet = tabTimesheet.name
+						join `tabSales Invoice` as si on si.name = sit.parent and si.status != "Cancelled"
+						join `tabSalary Slip` as ss on ss.name = sst.parent and ss.status != "Cancelled" """.format(standard_working_hours)
```
ruchamahabal, line 46: If `standard_working_hours` are not set, then the report breaks.

image

I think instead of setting Standard Working Hours to some default value, we should add a message in this report, before calculating and returning the data, like:

The metrics for this report are calculated based on the Standard Working Hours. Please set **Standard Working Hours** in HR Settings.

Provide a link to the Settings page.

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,239 @@
...
+	for row in data:
+		row.fractional_cost = row.base_gross_pay * row.utilization
+		row.profit = row.base_grand_total - row.base_gross_pay * row.utilization
+	return data
+
+def get_conditions(filters):
+	conditions = []
+
+	if filters.get("company"):
+		conditions.append("tabTimesheet.company='{0}'".format(filters.get("company")))
```
ruchamahabal:
```suggestion
	if filters.get("company"):
		conditions.append("tabTimesheet.company='{0}'".format(frappe.db.escape(filters.get("company"))))
```

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,239 @@
...
+
+	if filters.get("start_date"):
+		conditions.append("tabTimesheet.start_date>='{0}'".format(filters.get("start_date")))
+
+	if filters.get("end_date"):
+		conditions.append("tabTimesheet.end_date<='{0}'".format(filters.get("end_date")))
+
+	if filters.get("customer_name"):
+		conditions.append("si.customer_name='{0}'".format(filters.get("customer_name")))
```
ruchamahabal:
```suggestion
	if filters.get("customer_name"):
		conditions.append("si.customer_name='{0}'".format(frappe.db.escape(filters.get("customer_name"))))
```

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,239 @@
...
+
+	if filters.get("employee"):
+		conditions.append("tabTimesheet.employee='{0}'".format(filters.get("employee")))
```
ruchamahabal:
```suggestion
	if filters.get("employee"):
		conditions.append("tabTimesheet.employee='{0}'".format(frappe.db.escape(filters.get("employee"))))
```

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,239 @@
...
+
+	if filters.get("project"):
+		conditions.append("tabTimesheet.parent_project='{0}'".format(filters.get("project")))
```
ruchamahabal:
```suggestion
	if filters.get("project"):
		conditions.append("tabTimesheet.parent_project='{0}'".format(frappe.db.escape(filters.get("project"))))
```

ruchamahabal (changes requested):

erpnext/projects/report/project_profitability/project_profitability.py
```diff
@@ -0,0 +1,210 @@
...
+					"values": utilization
+				}
+			]
+		},
+		"type": "bar",
+		"colors": ["#84BDD5"]
+	}
+	return charts
+
+def get_columns():
```
rmehta, line 109: columns can now be configured in the Report doc

ruchamahabal: Is this meant for Standard reports too? Doesn't work maybe due to this:

https://github.com/frappe/frappe/blob/8de131dab29b69b33729240a66bc3483513dafc4/frappe/core/doctype/report/report.py#L123-L127

In the case of standard reports, it directly calls the execute function from the module path.

ruchamahabal (approved):

erpnext/projects/report/project_profitability/project_profitability.py
rmehta: @ruchamahabal fix it !

PR #24949, opened 2021-03-19
fix: serial no trim issue

pateljannat:
**Issue:**

1. Serial No field for transaction doctypes was causing issues with manual entry as well as entry through a barcode reader.
2. The unnecessary trim code was causing the cursor to go back to the previous line.

Screen Recording 2021-03-18 at 6 29 58 PM (1)

**Steps to replicate:**

1. Create any new transaction document (e.g. Purchase Receipt)
2.  Enter an item and open the item row.
3. In the Serial no field, enter the serial number. 
4. Then press enter to go to the next line.
5. The cursor should stay on the new line and should not go back to the previous line.


dufani1:
This is a serious issue and it is appearing on SI, DN, PI, and PR too.

PR #24950, opened 2021-03-19
fix: serial no trim issue

pateljannat:
Backport of https://github.com/frappe/erpnext/pull/24949


dufani1:
@nabinhait please help to merge this fix as hotfix
This is very critical issue

erpnext/public/js/controllers/transaction.js
```diff
@@ -635,21 +635,15 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 			}
 			else {
 				var valid_serial_nos = [];
-
+				var serialnos = [];
```
anupamvs, line 638: serialnos -> serial_nos

PR #24979, opened 2021-03-23
Sets "Maintain Stock" option defaults to unchecked

hsrai:
- In Item doctype, changes default value of "Maintain Stock" option to
  unchecked

> Screenshots/GIFs


hsrai:
Was not intended to upstream branch.

PR #24980, opened 2021-03-23
fix: repost not completed backdated transactions

rohitwaghchaure:
1. Repost the entries which are in state Queued / In Progress since long time
2. Don't check Values out of sync, if repost item valuation entries are exists

### DIFF
--- a/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
+++ b/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
@@ -5,7 +5,7 @@
 from __future__ import unicode_literals
 import frappe, erpnext
 from frappe.model.document import Document
-from frappe.utils import cint, get_link_to_form
+from frappe.utils import cint, get_link_to_form, add_to_date, today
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
```diff
@@ -112,4 +111,24 @@ def notify_error_to_stock_managers(doc, traceback):
...
+		return
+
+	for d in frappe.get_all('Company', filters= {'enable_perpetual_inventory': 1}):
+		check_if_stock_and_account_balance_synced(today(), d.company)
+
+def get_repost_item_valuation_entries():
+	date = add_to_date(today(), hours=-12)
 
+	return frappe.db.sql(""" SELECT name from `tabRepost Item Valuation`
+		WHERE status != 'Completed' and creation <= %s and docstatus = 1
```
nabinhait, line 132: why creation less than 12 hours ago?

nabinhait (changes requested):

nabinhait (approved):

PR #24888, opened 2021-03-15
fix: Allow zero valuation in stock reconciliation

ankush:
Stock reconciliation can not be done for customer provided item as they have zero valuation. This change adds a checkbox in item table to allow such items. The checkbox is automatically checked if item is "Customer provided item".

UI screenshot (checkbox is newly added):
image

Related issue: ISS-20-21-10248

PR #24985, opened 2021-03-23
fix: Allow zero valuation in stock reconciliation

ankush:
backport of #24888


marination (approved):

marination:
Merging as two unrelated tests (inpatient medication and gtr3b) are failing

PR #24994, opened 2021-03-24
refactor(HR): Add missing reports and doctypes to the Workspace

ruchamahabal:
screencapture-erp-healthcare-8001-app-hr-2021-03-24-13_04_21


Anurag810 (changes requested):
Employee tax and Benefits are now Part of Payroll...

Anurag810 (changes requested):
I think the Key report and others report doesn't sound good.
we can recategorize and make label them something else...
Just a thought...

ruchamahabal:
> I think the Key report and others report doesn't sound good.
> we can recategorize and make label them something else...
> Just a thought...

All the important reports have been grouped under Key Reports and unimportant ones under Other Reports. Thought of keeping it uniform with other modules.

Buying:

image

Selling:

image

Stock:

image

PR #25000, opened 2021-03-24
fix: GST on freight charge in e-invoicing

deepeshgarg007:
In many scenarios where delivery of the goods is involved freight and forwarding charges are also applied on the Goods delivered and this freight and forwarding is also taxable which means GST has to be applied on the freight and forwarding charges as well

Companies that are not pure transportation companies have to divide these extra charges between items and report total taxable value as Item Cost + Freight and Forwarding charges

**Solution:**
Add a custom field for taxable to maintain total taxable values inclusive of freight and forwarding charges which can be used for e-invoice generation and other GST reports


erpnext/hooks.py
```diff
@@ -279,7 +279,7 @@
 		'validate': ['erpnext.regional.india.utils.set_place_of_supply']
 	},
 	('Sales Invoice', 'Purchase Invoice'): {
-		'validate': ['erpnext.regional.india.utils.validate_document_name']
+		'validate': ['erpnext.regional.india.utils.validate_document_name', 'erpnext.regional.india.utils.update_taxable_values']
```
nextchamp-saqib, line 282: The `taxable_value` field will need to be added to the Purchase Invoice.

nextchamp-saqib: Is this supposed to be calculated for Purchase Invoices? If yes, then we need to add the `taxable_value` field for the Purchase Invoice Item too.

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -171,10 +171,15 @@ def get_item_list(invoice):
...
-		item.gross_amount = abs(item.base_net_amount)
-		item.taxable_value = abs(item.base_net_amount)
+
+		if invoice.apply_discount_on == 'Net Total' and invoice.discount_amount:
+			item.discount_amount = abs(item.base_amount - item.base_net_amount)
+		else:
+			item.discount_amount = 0
+
+		item.unit_rate = abs((item.taxable_value - item.discount_amount)/ item.qty)
+		item.gross_amount = abs(item.taxable_value + item.discount_amount)
```
```suggestion
		item.gross_amount = abs(item.taxable_value) + item.discount_amount
```

Consider `item.taxable_value = -200` & calculated `item.discount_amount = 20` then `gross_amount = 180` which is wrong.

Tested locally, this creates a mismatch.

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -256,20 +265,38 @@ def update_invoice_taxes(invoice, invoice_value_details):
...
 		if t.account_head in gst_accounts_list:
 			if t.account_head in gst_accounts.cess_account:
 				# using after discount amt since item also uses after discount amt for cess calc
 				invoice_value_details.total_cess_amt += abs(t.base_tax_amount_after_discount_amount)
 
 			for tax_type in ['igst', 'cgst', 'sgst']:
 				if t.account_head in gst_accounts[f'{tax_type}_account']:
-					invoice_value_details[f'total_{tax_type}_amt'] += abs(t.base_tax_amount_after_discount_amount)
+
+					invoice_value_details[f'total_{tax_type}_amt'] += abs(t.tax_amount)
```
```suggestion
					invoice_value_details[f'total_{tax_type}_amt'] += abs(tax_amount)
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -256,20 +265,38 @@ def update_invoice_taxes(invoice, invoice_value_details):
...
+				update_other_charges(t, invoice_value_details, gst_accounts_list, invoice, considered_rows)
 		else:
-			invoice_value_details.total_other_charges += abs(t.base_tax_amount_after_discount_amount)
+			invoice_value_details.total_other_charges += abs(t.tax_amount)
```
```suggestion
			invoice_value_details.total_other_charges += abs(tax_amount)
```

PR #25007, opened 2021-03-25

FHenry:
From #24749


FHenry:
Should I do this PR on version-13 branch as now the branch exists ? It semmes that branch version-13 was not already merge into version-13-beta, that's why there is so much diff in ths PR, there is only on file change normally (erpnext/setup/doctype/global_defaults/global_defaults.js).

FHenry:
I close this PR because of #25176 is better

PR #25008, opened 2021-03-25
fix: Add french chart accounts with account number

FHenry:
Add french chart accounts with account number and fix chart of account without number


FHenry:
Ic lose this PR as this one  #25178  is branchbetter directly on version-13

PR #25022, opened 2021-03-26
fix: Employee profile pic upload access for erpnext user

shariquerik:
Issue 1:
When an `Employee` document is connected with `User` (i.e. user_id is set as `User` in `Employee` document) and if the User's profile pic is already set, the user cannot change the Employee's profile pic.

Issue 2:
Cannot Upload Employee's Profile Picture when the user has `Employee Role` (without write access to Employee document) provided that `Employee` record is linked with `User`.


shariquerik:
Dependent PR: https://github.com/frappe/frappe/pull/12694

Anurag810 (approved):

PR #25024, opened 2021-03-26
feat: Delayed Tasks Summary

pateljannat:
1. The Delayed Deliverables Summary report helps identify the tasks which have exceeded their Expected End Dates.

2. The report is generated based on the Task Doctype.

delayed-deliverables-summary

### Calculation of Delay

3. Tasks have a date field called **Completed On**, which becomes visible when a Tasks status is changed to Completed.

##### Scenario 1

4. If a task is marked as Completed and the Completed On field is set, then the delay is calculated as the difference between the Completed On and the Expected End Date.

```
Delay = Completed On - Expected End Date
```
##### Scenario 2

5. If the Completed On field is not set, then the delay is calculated as the difference between the current date and the Expected End Date.

```
Delay = Current Date - Expected End Date
```

**Documentation PR:** https://github.com/frappe/erpnext_documentation/pull/269


ruchamahabal:
@pateljannat I think we should rename it to Delayed Tasks Summary instead of deliverables since all tasks might not be "deliverables"

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.json
```diff
@@ -0,0 +1,26 @@
...
+ "module": "Projects",
+ "name": "Delayed Tasks Summary",
+ "owner": "Administrator",
+ "prepared_report": 0,
+ "ref_doctype": "Task",
+ "report_name": "Delayed Tasks Summary",
+ "report_type": "Script Report",
+ "roles": [
+  {
+   "role": "Projects User"
```
ruchamahabal, line 23: Add Projects Manager role too

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.js
```diff
@@ -0,0 +1,48 @@
...
+			"options": ["Low", "Medium", "High", "Urgent"],
+			"default": "Low",
+			"reqd": 1
+		},
+		{
+			"fieldname": "status",
+			"label": __("Status"),
+			"fieldtype": "Select",
+			"options": ["Open", "Working","Pending Review","Overdue","Completed"],
+			"default": "Overdue"
```
ruchamahabal: Add a blank option for both Status and Priority filters as they are optional filters.

image

erpnext/projects/doctype/task/task.json
```diff
@@ -153,8 +154,10 @@
    "fieldname": "exp_start_date",
    "fieldtype": "Date",
    "label": "Expected Start Date",
+   "mandatory_depends_on": "eval: doc.is_template == 0;",
    "oldfieldname": "exp_start_date",
-   "oldfieldtype": "Date"
+   "oldfieldtype": "Date",
+   "reqd": 1
```
ruchamahabal: Let's not keep the Date fields mandatory. In the report, we can either filter out such tasks, or else we can keep the delay as 0 for the tasks which don't have the end date set.

erpnext/projects/doctype/task/task.json
```diff
@@ -381,14 +387,20 @@
    "fieldname": "duration",
    "fieldtype": "Int",
    "label": "Duration (Days)"
+  },
+  {
+   "depends_on": "eval: doc.status == \"Completed\"",
+   "fieldname": "completed_on",
+   "fieldtype": "Date",
+   "label": "Completed On"
```
ruchamahabal: Let's keep the "Completed On" field mandatory on changing status to Completed. Also, add a validation to check "Completed On" does not exceed the current date.

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.py
```diff
@@ -0,0 +1,120 @@
...
+def get_data(filters):
+	conditions = get_conditions(filters)
+	tasks = frappe.get_all("Task", conditions, ["name", "subject", "exp_start_date", "exp_end_date", "status", "priority", "completed_on", "progress"], order_by="creation")
+	for task in tasks:
+		if task.exp_end_date:
+			if task.completed_on:
+				task.delay = date_diff(task.completed_on, task.exp_end_date)
+			else:
+				task.delay = date_diff(nowdate(), task.exp_end_date)
+	return tasks
```
ruchamahabal:
```suggestion

def get_data(filters):
	conditions = get_conditions(filters)
	tasks = frappe.get_all("Task",
		filters=conditions,
		fields=["name", "subject", "exp_start_date", "exp_end_date",
			"status", "priority", "completed_on", "progress"],
		order_by="creation"
	)

	for task in tasks:
		if task.exp_end_date:
			if task.completed_on:
				task.delay = date_diff(task.completed_on, task.exp_end_date)
			elif task.status == "Completed":
				# task is completed but completed on is not set (for older tasks)
				task.delay = 0
			else:
				# task not completed
				task.delay = date_diff(nowdate(), task.exp_end_date)
		else:
			# task has no end date, hence no delay
			task.delay = 0

	return tasks
```

1. Space out your functions better to improve readability.
2. Cover more cases
     - Task is completed but Completed On is not set (for older tasks). This will always show the task as Delayed but we can't be sure of this since the "Completed On" field was just introduced. Let's set the delay for such tasks to be 0.
     - If Task has no End Date, there is no delay
3. Check for uniform quote styles throughout the file.

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.py
```diff
@@ -0,0 +1,120 @@
...
+
+def get_chart_data(data):
+	on_track = []
+	delay = []
+
+	for entry in data:
+		if entry.get("delay") > 0:
+			delay.append(entry.get("delay"))
+		else:
+			on_track.append(entry.get("delay"))
```
ruchamahabal: Since you just want to show the counts of delayed and on-track tasks, declare them as integers instead of lists and increment count as per conditions.

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.py
```diff
@@ -0,0 +1,120 @@
...
+		{
+			"fieldname": "completed_on",
+			"fieldtype": "Date",
+			"label": "Actual End Date",
+			"width": 130
+		},
+		{
+			"fieldname": "delay",
+			"fieldtype": "Data",
+			"label": "Delay",
```
ruchamahabal:
```suggestion
			"label": "Delay (In Days)",
```

erpnext/projects/report/delayed_tasks_summary/test_delayed_tasks_summary.py
```diff
@@ -0,0 +1,54 @@
...
+				"subject": "_Test Task 98",
+				"status": "Completed",
+				"priority": "Low",
+				"delay": -1
+			}
+		]
+		report = execute(filters)
+		data = list(filter(lambda x: x.subject == "_Test Task 99", report[1]))[0]
+		
+		for key in ["subjet", "status", "priority", "delay"]:
```
ruchamahabal:
```suggestion
		for key in ["subject", "status", "priority", "delay"]:
```
I guess this is the reason the tests are failing.

erpnext/projects/report/delayed_tasks_summary/test_delayed_tasks_summary.py
```diff
@@ -0,0 +1,54 @@
...
+			self.assertEqual(expected_data[0].get(key), data.get(key))
+
+		filters.status = "Completed"
+		report = execute(filters)
+		data = list(filter(lambda x: x.subject == "_Test Task 98", report[1]))[0]
+
+		for key in ["subjet", "status", "priority", "delay"]:
```
ruchamahabal:
```suggestion
		for key in ["subject", "status", "priority", "delay"]:
```

ruchamahabal (changes requested):

erpnext/projects/doctype/task/task.json
pateljannat: As we are not keeping the end date as mandatory, I have also made the start and end date filters in the report non-mandatory. If we keep the filters as mandatory then tasks with no date get filtered out and won't appear in the report at all.

erpnext/projects/report/delayed_tasks_summary/delayed_tasks_summary.py
```diff
@@ -0,0 +1,131 @@
...
+			elif task.status == "Completed":
+				# task is completed but completed on is not set (for older tasks)
+				task.delay = 0
+			else:
+				# task not completed
+				task.delay = date_diff(nowdate(), task.exp_end_date)
+		else:
+			# task has no end date, hence no delay
+			task.delay = 0
+
```
ruchamahabal, line 36: We should sort the tasks according to delay (descending)

ruchamahabal (approved):

PR #25025, opened 2021-03-26
feat(Production Plan): Enhancements in Material Request Plan Item

ruchamahabal:
### Consider Safety Stock in Required Qty Calculation

While creating a Production Plan, sometimes we need to include safety stock to calculate Required Qty. There is a Safety Stock field in the Item Master which indicates the minimum amount of stock that should be present.

Added a field to show Safety Stock in the Material Request Item table in Production Plan.
If **_Include Safety Stock in Required Qty Calculation_** is checked, then, the safety stock configured in the Item master will be added to the respective item's required qty.

safety_stock

### New fields in Material Request Plan Item table

- Ordered Qty: Bin
- Reserved Qty for Production: Bin

new-fields

- Required Qty as per BOM: Currently, we directly show the calculated Required Qty. However, the Required Qty as per BOM is not visible i.e. `planned_qty * bom_qty`. Added a field to show that so that calculations become clearer.

image

### Added more fields to the **Required Materials Download**
   - Required Qty as per BOM
   - Ordered Qty
   - Safety Stock
   - Required Qty for Production
   
  download


TurkerTunali:
It is a great feature. We may need to add that feature to docs.

PR #25063, opened 2021-03-30
fix: do not fetch stopped MR in production plan

ankush:
- Ignore stopped Material Requests (MR) while using fetch button from production plan
- Add filter on MR field inside child table.

Related issue: ISS-20-21-10757

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.js
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.js
@@ -25,6 +25,16 @@ frappe.ui.form.on('Production Plan', {
 			}
 		});
 
+		frm.set_query('material_request', 'material_requests', function() {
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/production_plan/production_plan.js
```diff
@@ -25,6 +25,16 @@ frappe.ui.form.on('Production Plan', {
...
 		});
 
+		frm.set_query('material_request', 'material_requests', function() {
+			return {
+				filters: {
+					material_request_type: "Manufacture",
+					docstatus: 1,
+					status: ["!=", "Stopped"],
+				}
+			}
```
marination: Sider
```suggestion
			};
```

marination (approved):

PR #25070, opened 2021-03-30
fix: Update Bin via Update Item on Purchase/Sales Order

marination:
Port of https://github.com/frappe/erpnext/pull/23509

**Issue:**
- On deletion of row Bin was not updated via Update Items
- Furthermore, in PO, Material Request and requested wasn't updated either on row deletion

**Fix:**
- Explicitly update bin on row deletion in `validate_and_delete_children`
- First validate if row can be deleted, if yes proceed to delete
- Then update linked documents , like MR. Without these up to date Bin will get outdated values after recalculation. It recalculates from child table rows
- Finally update requested & ordered / reserved qty in Bin 
- Added Tests
- Broke row deletion logic into smaller functions
- Re-ran patches to updates **requested**(MR), **reserved**(SO) and **ordered**(PO) qty

**Miscellanous:**
- Made field **Reserved Qty for subcontracting** read only in **Bin**

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -252,6 +252,7 @@ class PurchaseOrder(BuyingController):
 		self.update_prevdoc_status()
 
 		# Must be called after updating ordered qty in Material Request
+		# bin uses Material Request Items to recalculate & update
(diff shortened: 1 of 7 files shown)


marination:
none of the tests are failing due to this change. Can confirm

PR #25079, opened 2021-03-31
fix: PO not created against all selected suppliers (drop shipping)

marination:
Port of https://github.com/frappe/erpnext/pull/24863


marination:
4 unrelated tests failing

PR #25105, opened 2021-03-31
fix: Picked Qty conversion from Stock Qty to Qty while creating DN from Pick List

### LINKED ISSUE #24970: Error in Qty fetched in Delivery Note from Pick List (Multiple UOM)
## Description of the issue
When you create a Delivery Note from Pick List Qty fetched in the child table is Stock Qty where as UOM fetched is UOM (not Stock UOM).
This causes a problem when you have multiple UOMs

This bug represents an issue only when the item has Multiple UOMs

**Output of `bench version`**
```
ERPNext: v13.0.0-beta.11 (version-13-beta)

Frappe Framework: v13.0.0-beta.10 (version-13-beta)
```

1. Create an Item with Multiple UOMs and Has Batch No
2. Create a Sales Order (and Submit it)
3. Create Pick List from Sales Order and Submit it (ERPNext will fetch Qty based on available batches)
Screenshot 2021-03-22 at 4 12 04 PM

4. Create Delivery Note from Sales Order
Screenshot 2021-03-22 at 4 13 00 PM

### Observed result
Here the Qty fetched is Stock Qty (should be Qty) and UOM fetched is UOM (which is correct)

### Expected result
Qty fetched in Delivery Note should be Qty and not Stock Qty

Not Applicable

## Additional information

NA

Anuja-pawar:
Fixes #24970

**Issue occurs while creating a Delivery Note from a Pick List


erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -346,7 +346,7 @@ def create_delivery_note(source_name, target_doc=None):
 
 		if dn_item:
 			dn_item.warehouse = location.warehouse
-			dn_item.qty = location.picked_qty
+			dn_item.qty = location.picked_qty / location.conversion_factor
```
marination:
```suggestion
			dn_item.qty = flt(location.picked_qty) / (flt(location.conversion_factor) or 1)
```

marination (approved):

PR #25117, opened 2021-04-01
fix(Payroll): component amount calculation based on formula with abbr not working

### LINKED ISSUE #25036: bug: Deductions not coming through onto Salary Slip
Could be related to #24465

Salary Structure:

Screen Shot 2021-03-26 at 21 58 20

Salary Slip:

Screen Shot 2021-03-26 at 22 04 50

**Output of `bench version`**
```
ERPNext: v13.0.0-dev
Frappe Framework: v13.0.0-dev
```

1. Create Employee, Salary Structure, Assignment
2. Create Payroll Entry & submit

### Observed result
No deduction on salary slip

### Expected result
Deduction

ruchamahabal:
Fixes: https://github.com/frappe/erpnext/issues/25036

### Problem

Component amount calculation based on a formula with abbreviation not working 

### Steps to replicate

Create the following documents:

**Salary Structure:**

image

**Salary Structure Assignment:**

image

**Salary Slip created from Payroll Entry:**

salary-comp-abbr-before

### After Fix

salary-comp-abbr-after

### DIFF
--- a/erpnext/payroll/doctype/salary_slip/salary_slip.py
+++ b/erpnext/payroll/doctype/salary_slip/salary_slip.py
@@ -618,12 +618,15 @@ class SalarySlip(TransactionBase):
 
 			component_row = self.append(component_type)
 			for attr in (
-				'depends_on_payment_days', 'salary_component', 'abbr'
+				'depends_on_payment_days', 'salary_component',
 				'do_not_include_in_total', 'is_tax_applicable',
 				'is_flexible_benefit', 'variable_based_on_taxable_salary',
 				'exempted_from_income_tax'
 			):
 				component_row.set(attr, component_data.get(attr))
+
+			abbr = component_data.get('abbr') or component_data.get('salary_component_abbr')
+			component_row.set('abbr', abbr)
 
 		if additional_salary:
 			component_row.default_amount = 0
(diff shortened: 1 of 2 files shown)


ruchamahabal:
Fix not needed for `version-13-pre-release` branch.

erpnext/payroll/doctype/salary_slip/salary_slip.py
```diff
@@ -618,13 +618,16 @@ def update_component_row(self, component_data, amount, component_type, additiona
 
 			component_row = self.append(component_type)
 			for attr in (
-				'depends_on_payment_days', 'salary_component', 'abbr'
```
sagarvora, line 621: Okay, turns out that setting the `abbr` separately was never the issue. Python did not find a comma after `abbr` so it assumed `abbrdo_not_include_in_total` to be one string.

sagarvora: It's my fault :sweat_smile: 
https://github.com/frappe/erpnext/pull/24824/files#diff-a6a219b6ecafbe24da6ed9f0453f790941824d8728616368a238a74e4a110a14R620

image

sagarvora: image

ankush: wait, implicit string concatenation this should've been caught by semgrep 🤔

ruchamahabal: Was aware of the missing comma while raising this fix but the deduction component was still not getting computed in some cases. I checked all occurrences of abbr. The `salary_component_abbr` field from Salary Component gets fetched with the alias `abbr` everywhere. So the deductions are coming through just fine now.

PR #25153, opened 2021-04-02
fix(regional): remove shipping address GSTIN validation for e-invoice

bhavesh95863:


sagarvora:
There's a stack overflow in the failing patch tests. Any idea why?

PR #25156, opened 2021-04-02
feat: manage Python 3 compatiblity with dependencies

Alchez:
A handful of integrations are dropping / have already dropped Py2 support. And their dependency changes are causing conflicts with Frappe's dependency versions.

Related Frappe PR at https://github.com/frappe/frappe/pull/12784.

`no-docs`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,15 +1,13 @@
-braintree==3.57.1
 frappe
-gocardless-pro==1.11.0
-googlemaps==3.1.1
-pandas>=1.0.5
-plaid-python>=7.0.0
-pycountry==19.8.18
-PyGithub==1.44.1
-python-stdnum==1.12
-python-youtube==0.6.0
-taxjar==1.9.0
-tweepy==3.8.0
-Unidecode==1.1.1
-WooCommerce==2.1.1
-pycryptodome==3.9.8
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,15 +1,11 @@
-braintree==3.57.1
-frappe
```
gavindsouza, line 2: Probably leave `frappe` as a dependency still? At least for the sake of correctness

Alchez: Alright, re-added it

requirements.txt
```diff
@@ -1,15 +1,12 @@
-braintree==3.57.1
 frappe
-gocardless-pro==1.11.0
-googlemaps==3.1.1
```
gavindsouza, line 4: `googlemaps` is being used in the delivery trip controller of the Stock module. Although `frappe` installs it because of its dependency, it's best to track it here as well. We may not need to pin it here though.

Couldn't find usages of `braintree` or `pycryptodome` so I think dropping them is ok.

gavindsouza (changes requested):

requirements.txt
Alchez: Alright, I added googlemaps back in

gavindsouza (approved):

PR #25159, opened 2021-04-02
fix: object referencing the same address issue

KaviyaPeriyasamy:
Scenario:

While generating IRN, the validation error is not refreshing on every click. It gets append to the previous log due to the error log object referring to the same memory address.

Before fix:

error

After fix:

fix

### DIFF
--- a/erpnext/regional/india/e_invoice/utils.py
+++ b/erpnext/regional/india/e_invoice/utils.py
@@ -397,7 +397,7 @@ def safe_json_load(json_string):
 		frappe.throw(_("Error in input data. Please check for any special characters near following input: <br> {}").format(snippet))
 
 def validate_einvoice(validations, einvoice, errors=None):
-	if not errors:
+	if errors is None:
 		errors = []
 	for fieldname, field_validation in validations.items():
 		value = einvoice.get(fieldname, None)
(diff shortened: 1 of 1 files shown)


karthikeyan5 (approved):

karthikeyan5 (changes requested):
Please use the pythonic way to do this as mentioned here:


https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

KaviyaPeriyasamy:
> Please use the pythonic way to do this as mentioned here:
> 
> https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

Can you please review it again? Changed as per the suggestion https://github.com/frappe/erpnext/pull/25159/commits/9c832f3a9062c86aae4398004b0a37c0d098045f

karthikeyan5 (approved):

nextchamp-saqib (approved):

nextchamp-saqib:
@KaviyaPeriyasamy Thanks for taking this up. Can you also raise one PR for version-12-hotfix branch?

KaviyaPeriyasamy:
> @KaviyaPeriyasamy Thanks for taking this up. Can you also raise one PR for the version-12-hotfix branch?

For version-12-hotfix https://github.com/frappe/erpnext/pull/25165

PR #25166, opened 2021-04-03
fix(india): critical e-invoicing fixes

nextchamp-saqib:
Port of #25164

### DIFF
--- a/erpnext/regional/india/e_invoice/utils.py
+++ b/erpnext/regional/india/e_invoice/utils.py
@@ -86,10 +86,10 @@ def get_doc_details(invoice):
 		invoice_date=invoice_date
 	))
 
-def get_party_details(address_name):
+def get_party_details(address_name, company_address=None, billing_address=None, shipping_address=None):
 	d = frappe.get_all('Address', filters={'name': address_name}, fields=['*'])[0]
 
-	if (not d.gstin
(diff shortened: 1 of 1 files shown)


erpnext/regional/india/e_invoice/utils.py
```diff
@@ -86,10 +86,10 @@ def get_doc_details(invoice):
 		invoice_date=invoice_date
 	))
 
-def get_party_details(address_name):
+def get_party_details(address_name, company_address=None, billing_address=None, shipping_address=None):
```
sagarvora, line 89: - Let's remove unused parameters here. Only `shipping_address` should suffice.
- Will need to fix the call to this function as well.
- Might need changes in other branches as well.

sagarvora (dismissed):

erpnext/regional/india/e_invoice/utils.py
nextchamp-saqib: I am working on re-writing the validations with bulk e-invoice generation PR. So, we could keep this as it is? This will get overwritten in a few weeks. 

Or should I just remove these for now?

sagarvora: Your choice. As long as you'll fix it in a few weeks.

PR #25178, opened 2021-04-03
fix: Add french chart accounts with account number

FHenry:
Add french chart accounts with account number and fix chart of account without number

#25008 #25009

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/verified/fr_plan_comptable_general.json
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/verified/fr_plan_comptable_general.json
@@ -1,1609 +1,1609 @@
 {
-    "country_code": "fr", 
-    "name": "France - Plan Comptable General", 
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:
@FHenry I have tested your PR on local machine, chart of accounts has been made but for the chart of account "France - Plan Comptable General" as you have not mentioned the account type therefore in the company default accounts are not set. For more detail please check below image

Screenshot 2021-04-15 at 4 43 45 PM

Also kindly open PR in develop branch and not directly in the version-13 branch

FHenry:
@rohitwaghchaure Could you remind me where the defaut account is set (in the code) ? I will fix it.
Actually the defaut code selected for french account plan is not correct at all. 
I prefer have empty defaut code rather than bad ones.

Here on my Fresh install 
image

That's already the case with the french account plan ...
Default customer code must be 411 bank should be 512, and so on,

PR #25190, opened 2021-04-04
feat(HR): share doc with employee approvers if they don't have access

ruchamahabal:
Depends on: https://github.com/frappe/frappe/pull/12799
Documentation PR: https://github.com/frappe/erpnext_documentation/pull/279
### Problem

- When an employee is created, a user linked to that employee gets created with restricted User Permissions for Company and Employee link fields if **Create User Permissions** is checked. (mostly followed by all)
- Now, whenever a user sets the employee as a Leave Approver for anyone, they have to explicitly create new user permissions (as shown below) for them since they are restricted to their own records because of the reason mentioned in the first point.  The same applies to expense and shift request approvers.

   image

- If they set the approvers at the Department level, they will have to create user permissions for giving them access to all employee records in that department.
- Ideally, when employees are set as Leave Approvers they should have access to the documents.

### Enhancement

- Whenever a new Leave Application is created, check if the leave approver has permission to access the application. If not, share the document with the approver with "submit" permission.

   share-approver

- When the approver changes in the record, revoke the shared document for the approver that was set earlier.
- The same functionality is implemented for Shift Request Approver and Expense Claim


erpnext/hr/doctype/expense_claim/test_expense_claim.py
```diff
@@ -110,6 +110,34 @@ def test_rejected_expense_claim(self):
...
+
+		expense_claim.reload()
+		expense_claim.expense_approver = user
+		expense_claim.save()
+
+		frappe.set_user(user)
+		expense_claim.reload()
+		expense_claim.status = "Approved"
+		expense_claim.submit()
+		frappe.set_user("Administrator")
```
sagarvora, line 138: This should probably come in a `finally` block. That way, the user gets reset even though above test errors/fails.

PR #25196, opened 2021-04-05
fix: Company field in Warehouse

noahjacob:
Issue:
Company field wasn't visible before and used to take default values, but when the defaults were removed, on saving it used to throw an error since company field is mandatory and wasn't visible or editable in the form.
Solution:
Company field is now editable and is set to read only after saving it once.
Peek 2021-04-09 19-57


erpnext/stock/doctype/warehouse/warehouse.json
```diff
@@ -70,9 +71,9 @@
    "oldfieldname": "company",
    "oldfieldtype": "Link",
    "options": "Company",
-   "remember_last_selected_value": 1,
    "reqd": 1,
-   "search_index": 1
+   "search_index": 1,
+   "set_only_once": 1
```
marination: I believe `set_only_once` just makes sure the value of this field never changes after the first insert. it does not impact the field visibility.

What you should additionally do is:
-  make this field mandatory as it is an important field
- make **Read Only Depends on** as `eval: !doc.__islocal`. This means while the document is new, unsaved and un-inserted for the first time, the field should be editable. After the first save it should be read only

marination:
@noahjacob please also add images or a gif of the impact after the fix.

marination (approved):

PR #25128, opened 2021-04-01
fix(India): create property setters for shorter naming series on setup

sagarvora:
- This is in compliance with GST Rules and is being done to avoid ValidationError.
- FIxed relevant tests as well.

PR #25204, opened 2021-04-05
fix(India): create property setters for shorter naming series (#25134)

ankush:
backport of https://github.com/frappe/erpnext/pull/25128

### DIFF
--- a/erpnext/regional/india/setup.py
+++ b/erpnext/regional/india/setup.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals
 
 import frappe, os, json
 from frappe.custom.doctype.custom_field.custom_field import create_custom_fields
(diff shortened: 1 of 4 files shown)


erpnext/regional/india/setup.py
```diff
@@ -19,6 +20,7 @@ def setup(company=None, patch=True):
 # TODO: for all countries
 def setup_company_independent_fixtures():
 	make_custom_fields()
+	make_property_setters()
```
nabinhait, line 23: @ankush Property setter should not be created if called from the patch. Because, it will replace customer's customized series which might be within permitted length.
I have fixed this in develop and version-13 via https://github.com/frappe/erpnext/commit/aa809ceffef679fa8c12472f5dc51fb734c7438c. Can you please make that changes in v12 as well?

PR #25210, opened 2021-04-05
fix: list lookup with undefined variable

sunhoww:
Fixes an issue in **SerialNoBatchSelector** that prevented the `available_qty` from being set.


ankush (approved):
Variable name was changed in https://github.com/frappe/erpnext/pull/22539

PR #25221, opened 2021-04-06
feat: create Quality Inspections from account and stock documents

Alchez:
If a stock item requires inspection before purchase or delivery, the system alerts the user that a Quality Inspection document is required to move ahead. But there is no quick way to create these documents.

<hr>

**Screenshots / GIFs:**

quality-inspection

`no-docs`


erpnext/controllers/accounts_controller.py
```diff
@@ -1412,6 +1412,38 @@ def validate_and_delete_children(parent, data):
 	for d in deleted_children:
 		update_bin_on_delete(d, parent.doctype)
 
+
+@frappe.whitelist()
+def make_quality_inspections(doctype, docname, items):
```
marination: Accounts controller does not seem like the best location. Quality Inspection is carried out on stock/inventory and has no accounting impact. We can place this in `quality_inspection.py` itself perhaps.

Alchez: Ah, you're right. I can move this to the stock controller or quality inspection. Either way works.

Alchez:
Hello, any update on this?

erpnext/public/js/controllers/transaction.js
```diff
@@ -261,11 +261,19 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
-		var me = this;
-		var inspection_type = in_list(["Purchase Receipt", "Purchase Invoice"], this.frm.doc.doctype)
+
+		const me = this;
+		if (!this.frm.is_new() && this.frm.doc.docstatus === 0) {
+			this.frm.add_custom_button(__("Quality Inspection(s)"), () => {
+				me.make_quality_inspection();
+			}, __("Create"));
+			this.frm.page.set_inner_btn_group_as_primary(__('Create'));
+		}
```
marination, line 271: Add this to Stock Entry too, if `inspection_required` is True. It has it's own `setup_quality_inspection`

erpnext/controllers/stock_controller.py
```diff
@@ -500,6 +503,38 @@ def repost_future_sle_and_gle(self):
...
 				self.company, self.doctype, self.name)
 
+
+@frappe.whitelist()
+def make_quality_inspections(doctype, docname, items):
+	items = json.loads(items).get('items')
+	inspections = []
+
+	for item in items:
+		if item.get("sample_size") > item.get("qty"):
```
marination: qi-ux

erpnext/public/js/controllers/transaction.js
```diff
@@ -1915,6 +1923,125 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
...
+						items: data
+					},
+					freeze: true,
+					callback: function (r) {
+						if (r.message) {
+							frappe.msgprint({
+								message: __("Quality Inspections Created: {0}", [r.message.join(", ")]),
+								indicator: "green"
+							});
+						}
```
marination: A better callback would be to redirect to Quality Inspection List with the filter as the current transaction name, so users can directly see the list of QIs created.

If it's one QI, redirect to that saved QI form. The `msgprint` with links doesn't help a lot since the next action is to complete the Inspections.

marination (changes requested):
Please also add one basic test for this.

Alchez:
@marination 

Thanks for the review! I've added the changes you suggested, but I wasn't able to open any transaction documents to test them out.

<details>
<summary>Traceback</summary>

```js
TypeError: Cannot read property 'extend' of undefined
    at eval (sales_order__js:326)
    at ScriptManager.setup (script_manager.js:179)
    at FrappeForm.setup (form.js:92)
    at FrappeForm.refresh (form.js:274)
    at FormFactory.render (formview.js:107)
    at formview.js:92
    at Object.callback (model.js:216)
    at Object.callback [as success_callback] (request.js:78)
    at 200 (request.js:122)
    at Object.<anonymous> (request.js:247)
```

</details>

I'll also need help on the test, should I just check for QIs created if the server function is called? Or do you want a visual test as well?

erpnext/controllers/accounts_controller.py
```diff
@@ -1450,6 +1450,7 @@ def validate_and_delete_children(parent, data):
 	for d in deleted_children:
 		update_bin_on_delete(d, parent.doctype)
 
+
```
ankush, line 1452: if you drop this commit, semgrep won't crib about issues in that file.


Anyway, the problem is this line:

https://github.com/frappe/erpnext/blob/b0e77cbd89394a609108768466fef97cb00902bf/erpnext/controllers/accounts_controller.py#L228

This should be 
`data_field = None`

Alchez: Thanks for the suggestion, I added a fix for it anyway

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -115,6 +115,13 @@ frappe.ui.form.on('Stock Entry', {
 			return;
 		}
 
+		if (!frm.is_new() && frm.doc.docstatus === 0) {
+			frm.add_custom_button(__("Quality Inspection(s)"), () => {
+				frm.trigger("make_quality_inspection");
```
marination: reuse the method in `transaction.js`
```suggestion
				let transaction_controller = new erpnext.TransactionController({frm: frm});
				transaction_controller.make_quality_inspection();
```

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -142,6 +149,129 @@ frappe.ui.form.on('Stock Entry', {
 		});
 	},
 
+	make_quality_inspection: function (frm) {
```
marination: This can be removed, reuse the method in `transaction.js`

marination (changes requested):
- Consolidate these buttons on **Stock Entry**
  Screenshot 2021-05-26 at 3 28 31 PM

- Yes, just check for QIs created if the server function is called.
- I've tested the rest, works as intended 👍🏻

marination:
@Alchez can you add a test for this, we can then merge this

Alchez:
@marination, I added a test for the new function

image

marination (approved):

PR #25232, opened 2021-04-07
fix: keyerror while validating coa file

nextchamp-saqib:
If a file with no parent account column is used for importing COA, then validating accounts throws `KeyError: 'parent_account'`

Note: This happens only if a new column for eg. an account number column is added to the template & the parent account column is removed. Such that total columns remain the same. 

Ref Issue. ISS-21-22-00193

### DIFF
--- a/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
+++ b/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
@@ -293,6 +293,11 @@ def validate_accounts(file_name):
 	accounts_dict = {}
 	for account in accounts:
 		accounts_dict.setdefault(account["account_name"], account)
+		if not hasattr(account, "parent_account"):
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
```diff
@@ -293,6 +293,11 @@ def validate_accounts(file_name):
 	accounts_dict = {}
 	for account in accounts:
 		accounts_dict.setdefault(account["account_name"], account)
+		if not hasattr(account, "parent_account"):
+			msg = _("Please make sure the file you are using has 'Parent Account' column present in the header.")
+			msg += "<br><br>"
+			msg += _("Alternatively, you can download the template and fill your data in.")
+			frappe.throw(msg)
```
marination:
```suggestion
			frappe.throw(msg, title=_("Parent Account Missing"))
```

marination (approved):

PR #25238, opened 2021-04-07
feat: Improve DATEV export

barredterra:
### Custom fields

We need to keep track of a party's external ID for external bookkeeping. In the past, dedicated payable/receivable accounts per party were used to achieve this. The account number was the ID for external bookkeeping. This PR makes the following changes:

- Modify the **Party Account** table to make the link to **Account** non-mandatory.

    This way we can add other properties that apply per party, per company.

- Add a custom field "Debtor/Creditor Number" to **Party Account** for german users.
- Add a patch that removes the link to **Account** and adds the account number as Debtor/Creditor Number.

    - The DATEV report continues to work as expected.
    - Payables/receivables will be booked to the default account as specified in **Company** again. They can be tracked using the reports "Accounts payable" and "Accounts receivable".

If present, the Debtor/Creditor Number will be exported as the party's account number.

### Transactions

The report was lacking info about the involved parties. For example, you would see 100,00 € invoiced but not to whom.

- Add columns with Customer/Supplier Name
- Add columns with Debtor/Creditor Number

datev_export

### Master data

Export was using data from primary **Contact** for email and phone, and address data from primary/billing **Address**.

- Use phone and email from primary/billing **Address**. Less code, easier to understand for the user.

Export was using data from primary **Contact** for an individual party's first and last name.

- Parse party's full name to obtain first and last name.

### Todo

- [x] Patch to add custom field to existing instances
- [x] Patch to copy account number to debtor/creditor number
- [x] add docs

Docs: https://github.com/frappe/erpnext_documentation/pull/290


barredterra:
Sider seems to be wrong about "indentation contains mixed spaces and tabs". `datev.py` contains only tab indentation.

barredterra:
This PR has been posted on Discuss and in the "ERPNext German Chapter" Telegram group on April 8, 2021. So far there have been no objections or change requests regarding this PR.

barredterra:

rmehta (approved):

barredterra:
Thanks, @rmehta !

PR #25260, opened 2021-04-09
feat: Dimension-wise Accounts Balance Report

AfshanKhan:
A report which shows the balance `(debit-credit)` for the `Accounts` depending on `Accounting Dimension`

image

`no-docs`


deepeshgarg007 (changes requested):
@AfshanKhan please add a total column after the dimension columns as well showing the total of all dimensions for an account

AfshanKhan:
> @AfshanKhan please add a total column after the dimension columns as well showing the total of all dimensions for an account

@deepeshgarg007 Done.

PR #25300, opened 2021-04-12
fix: Let Administrator delete company transactions

GangaManoj:
_Issue:_ Only the user who created the company is able to use the 'Delete Company Transactions' option.

_Fix:_ Lets the Administrator use the option as well.

image_2021-04-12_18-26-47


abhishekbalam:

PR #25310, opened 2021-04-13
fix(minor): fix dynamically changing grid properties

rmehta:
Fixes dynamically changing grid properties using `frm.fields_dict.[grid_fieldname].grid.update_docfield_property(fieldname, property, value)`

Depends on: https://github.com/frappe/frappe/pull/12839


erpnext/accounts/doctype/journal_entry/journal_entry.js
```diff
@@ -431,15 +429,6 @@ cur_frm.cscript.validate = function(doc,cdt,cdn) {
 	cur_frm.cscript.update_totals(doc);
 }
 
-cur_frm.cscript.select_print_heading = function(doc,cdt,cdn){
-	if(doc.select_print_heading){
-		// print heading
-		cur_frm.pformat.print_heading = doc.select_print_heading;
```
rmehta, line 437: pformat is not used anymore

surajshetty3416 (approved):

PR #25312, opened 2021-04-13
fix: invalid comparison between NoneType & float

nextchamp-saqib:
fix: TypeError: '<' not supported between instances of 'NoneType' and 'float'


nextchamp-saqib:
Closing as already fixed by https://github.com/frappe/erpnext/pull/25222

PR #25314, opened 2021-04-13
fix(e-invoicing): validations & tax calculation fixes

nextchamp-saqib:
Fixes:
- Error handling if no credentials added for the company
- Tax calculation for GST on Freight Charges
- Invoice eligibility for e-invoicing
- Cancel e-way bill dialog
- Address validations for shipping
- Validations on e-invoice json (Total Item Value = Grand Total)
- Place of supply is incorrectly set for some addresses

Feat:
- Bulk e-invoice generation from list view


erpnext/patches/v12_0/add_einvoice_status_field.py
```diff
@@ -0,0 +1,69 @@
...
+from frappe.custom.doctype.custom_field.custom_field import create_custom_fields
+
+def execute():
+	company = frappe.get_all('Company', filters = {'country': 'India'})
+	if not company:
+		return
+
+	# move hidden einvoice fields to a different section
+	custom_fields = {
+		'Sales Invoice': [
```
rmehta, line 13: add e-invoice as a separate doctype?

erpnext/regional/report/e_invoice_summary/e_invoice_summary.py
```diff
@@ -0,0 +1,106 @@
...
+	data = frappe.get_all(
+		'Sales Invoice',
+		filters=query_filters,
+		fields=[d.get('fieldname') for d in get_columns()]
+	)
+
+	return data
+
+def get_columns():
+	return [
```
rmehta, line 49: columns are configurable in the report doctype

PR #25316, opened 2021-04-13
fix: shipment pickup_to, pickup_from functionality.

18alantom:
### Description
- Removes auto entry of **Pickup to** and **Pickup from** fields in the **Shipment** DocType.
- Both fields have been made mandatory to force manual entry of values.

### Behaviour After PR
**Pickup to** has to be manually set.
after

### Behaviour Before PR
**Pickup to** is auto set.
before

Setting **Pickup from** auto sets **Pickup to** 5 hours ahead, this leads to invalid values which breaks validation in controller.
Screen Shot 2021-04-13 at 18 03 34

### DIFF
--- a/erpnext/stock/doctype/shipment/shipment.js
+++ b/erpnext/stock/doctype/shipment/shipment.js
@@ -363,11 +363,6 @@ frappe.ui.form.on('Shipment', {
 		if (frm.doc.pickup_date < frappe.datetime.get_today()) {
 			frappe.throw(__("Pickup Date cannot be before this day"));
 		}
-		if (frm.doc.pickup_date == frappe.datetime.get_today()) {
-			var pickup_time = frm.events.get_pickup_time(frm);
-			frm.set_value("pickup_from", pickup_time);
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/shipment/shipment.js
```diff
@@ -381,7 +376,18 @@ frappe.ui.form.on('Shipment', {
 				frappe.throw(__("Pickup Time cannot be in the past"));
 			}
 		}
-		frm.trigger('set_pickup_to_time');
+	},
+	pickup_to: function(frm) {
```
ankush, line 380: I think it would be better to just remove `pickup_to`, `pickup_from` and `get_pickup_time`. 

There's a server-side validation for this: https://github.com/frappe/erpnext/blob/39b1cd827a5e3ba04c29189426e5f85f7fe77daf/erpnext/stock/doctype/shipment/shipment.py#L36-L38

ankush (changes requested):

PR #25344, opened 2021-04-15
test: Introduced cypress tests in ERPNext

nabinhait:


package.json
```diff
@@ -15,6 +15,8 @@
     "snyk": "^1.290.1"
   },
   "dependencies": {
+    "cypress": "^6",
+    "cypress-file-upload": "^5",
```
ankush: These should be dev dependency (or we can rely on FF, it dynamically installs Cypress)

ankush:
- removed dependency from package.json
- Fixed some linter/sider errors
- Added CI config (mostly same as frappe/frappe). Keeping it non-parallel for now, until test time exceeds setup time. @surajshetty3416 can comment on this.

ankush (approved):
Only projectId needs to be updated.

cypress/support/commands.js
```diff
@@ -0,0 +1,326 @@
+import 'cypress-file-upload';
```
surajshetty3416, line 1:
```suggestion
import '../../apps/frappe/cypress/support/commands.js'
```
@ankush can you try to extend commands from frappe somewhat like this instead of copying?

PR #25352, opened 2021-04-15
fix: exclude spurious Stock Entry Types from 'consumed' calculation

18alantom:
### Description
Exclude the following **Stock Entry Type**s 
- Material Transfer for Manufacture
- Material Transfer
- Send to Subcontractor 
 
from being used for **Consumed** column calculation.

### Why
Using the above types leads to incorrect calculation.


erpnext/stock/report/itemwise_recommended_reorder_level/itemwise_recommended_reorder_level.py
```diff
@@ -55,14 +55,28 @@ def get_item_info(filters):
...
 def get_consumed_items(condition):
+	stock_entry_types_to_exclude = ', '.join([f"'{se}'" for se in [
+		"Material Transfer for Manufacture",
+		"Material Transfer",
+		"Send to Subcontractor"
+	]])
+	condition += f"""
+		and (
+			stock_entry_type is NULL
+			or stock_entry_type not in ({stock_entry_types_to_exclude})
```
ankush: Use `purpose` instead. Users are allowed to define more stock entry types with the standard set of purposes.

https://docs.erpnext.com/docs/user/manual/en/stock/stock-entry#22-stock-entry-type

erpnext/stock/report/itemwise_recommended_reorder_level/itemwise_recommended_reorder_level.py
```diff
@@ -55,14 +55,28 @@ def get_item_info(filters):
...
+			on `tabStock Ledger Entry`.voucher_no = `tabStock Entry`.name
+		where
+			actual_qty < 0
 			and voucher_type not in ('Delivery Note', 'Sales Invoice')
 			%s
-		group by item_code
-	""" % condition, as_dict=1)
+		group by item_code""" % condition, as_dict=1)
 
 	consumed_items_map = {}
```
ankush: Not related to your changes but this can probably be converted to single-line dict comprehension, which is easier to read and faster.

Initializing and updating a dict with one-liner for-loop is generally anti-pattern. Same with lists.

erpnext/stock/report/itemwise_recommended_reorder_level/itemwise_recommended_reorder_level.py
```diff
@@ -55,14 +55,28 @@ def get_item_info(filters):
...
+			or stock_entry_type not in ({stock_entry_types_to_exclude})
+		)
+	"""
+	condition = condition.replace("posting_date",
+					"`tabStock Ledger Entry`.posting_date")
 	consumed_items = frappe.db.sql("""
 		select item_code, abs(sum(actual_qty)) as consumed_qty
-		from `tabStock Ledger Entry`
-		where actual_qty < 0
+		from `tabStock Ledger Entry` left join `tabStock Entry`
```
ankush: Consider using aliases for table names.

erpnext/stock/report/itemwise_recommended_reorder_level/itemwise_recommended_reorder_level.py
```diff
@@ -55,14 +55,28 @@ def get_item_info(filters):
 
 
 def get_consumed_items(condition):
+	stock_entry_types_to_exclude = ', '.join([f"'{se}'" for se in [
+		"Material Transfer for Manufacture",
+		"Material Transfer",
+		"Send to Subcontractor"
+	]])
```
ankush: Consider separating the list of purpose as a separate variable from query construction.

ankush (changes requested):

ankush (approved):

ankush:
@18alantom please send backport for `version-12-hotfix` branch too.

Issue #25293, opened 2021-04-12
POS window do not support special Characters on Scan.


rakshithrddy:
When a bar-code with a special characters like "-", "$" "_" etc is scanned in POS search field, the special characters are filtered out. 

Actual output: 
![Screenshot from 2021-04-12 15-59-16](https://user-images.githubusercontent.com/36509967/114380749-0c69bf80-9ba8-11eb-99c5-57e708b01c64.png)

Required Output: 
![Screenshot from 2021-04-12 15-55-44](https://user-images.githubusercontent.com/36509967/114380768-11c70a00-9ba8-11eb-8ea8-736d608a57fd.png)

PR #25353, opened 2021-04-15
fix(pos): special character scanning in point of sale

rakshithrddy:
Fix for issue number #25293
"REQUIRES DEBUGGING for Develop Branch"

Over ride decodeKeyEvent function to support special Characters on Bar-code scan in POS Window.


nextchamp-saqib:
@rakshithrddy Tested after merging it with develop branch and the scanning is working as expected. Can you confirm if it works for you too?

rakshithrddy:
> @rakshithrddy Tested after merging it with develop branch and the scanning is working as expected. Can you confirm if it works for you too?

Yes, It's working as expected.

PR #25354, opened 2021-04-15
feat: Transaction Deletion Record

GangaManoj:
**_What it does:_** 
This feature allows you to delete all the docs associated with the specified company, except for the ones belonging to the DocTypes listed in the _Excluded DocTypes_ table. This feature is also linked with the _Delete Company Transactions_ button in the Company DocType- clicking on that would submit a Transaction Deletion Record doc.

**_Fields:_**

- Company : To specify the company whose associated docs should be deleted
- Summary: Table that will only be displayed after submission that shows the DocTypes whose docs were deleted, along with the number of docs that were deleted for each.
- Excluded DocTypes: Table that lists the DocTypes whose docs wouldn't be deleted. A user can choose to remove any of the DocTypes from the list, but they can't add any.

Screenshot 2021-05-09 at 12 07 22 AM

Docs: https://github.com/frappe/erpnext_documentation/pull/316


erpnext/setup/doctype/transaction_deletion_log/transaction_deletion_log.py
```diff
@@ -0,0 +1,134 @@
...
+	def validate(self):
+		doctypes_to_be_ignored_list = get_doctypes_to_be_ignored()
+		for doctype in self.doctypes_to_be_ignored:
+			if doctype.doctype_name not in doctypes_to_be_ignored_list:
+				frappe.throw(_("DocTypes should not be added manually to the 'DocTypes That Won't Be Affected' table."))
+
+	def before_submit(self):
+		frappe.only_for('System Manager')
+		company_obj = frappe.get_doc('Company', self.company)
+		if frappe.session.user != company_obj.owner and frappe.session.user != 'Administrator':
```
nextchamp-saqib: This can be moved to the validation method.

erpnext/setup/doctype/transaction_deletion_log/transaction_deletion_log.json
```diff
@@ -0,0 +1,79 @@
+{
```
rmehta: Should be named "Transaction Deletion Tool" not log. A log is a static entity. This tool has actions linked.

rmehta (changes requested):

erpnext/setup/doctype/transaction_deletion_log/transaction_deletion_log.json
GangaManoj: Done.

erpnext/setup/doctype/transaction_deletion_log/transaction_deletion_log.py
GangaManoj: Done

erpnext/setup/doctype/transaction_deletion_tool/test_transaction_deletion_tool.py
```diff
@@ -0,0 +1,68 @@
...
+	})		
+	company.insert(ignore_if_duplicate = True)
+
+def create_transaction_deletion_request(company):
+	tdr = frappe.get_doc({
+		'doctype': 'Transaction Deletion Tool',
+		'company': company
+	})
+	tdr.insert()
+	tdr.before_submit()
```
nextchamp-saqib:
```suggestion
	tdr.submit()
```

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
```diff
@@ -0,0 +1,134 @@
...
+		company_obj.total_monthly_sales = 0
+		company_obj.sales_monthly_history = None
+		company_obj.save()
+		# Clear notification counts
+		clear_notifications()
+
+		singles = frappe.get_all('DocType', filters = {'issingle': 1}, pluck = 'name')
+		tables = frappe.get_all('DocType', filters = {'istable': 1}, pluck = 'name')
+		doctypes_to_be_ignored_list = singles + get_doctypes_to_be_ignored()
+		docfields = frappe.get_all('Docfield',
```
nextchamp-saqib:
```suggestion
		docfields = frappe.get_all('DocField', 
```

nextchamp-saqib (changes requested):

asbasawaraj:
@GangaManoj I think we should get rid of the 'Delete Company Transactions' button. 

1. It is an anti-pattern. The pattern is to provide the buttons on the top right. 
2. Leaving the button there provides two ways of doing one thing from the user's perspective but in the backend, it creates the 'Transaction Deletion Tool' record anyway.

We can just add a button on the top right, where usually buttons are, with the 'Delete Transactions' label.

What do you think?

GangaManoj:
@asbasawaraj https://github.com/frappe/erpnext/pull/25617
How does this look?

I think we should keep the button in some form though, for simplicity's sake. The tool is mostly for people who'd like to add to the list of DocTypes being deleted(and to maintain a record of all such deletion operations).

erpnext/setup/doctype/company/company.py
```diff
@@ -613,4 +613,14 @@ def get_default_company_address(name, sort_key='is_primary_address', existing_ad
...
+
+@frappe.whitelist()
+def create_transaction_deletion_request(company):
+	tdr = frappe.get_doc({
+		'doctype': 'Transaction Deletion Tool',
+		'company': company
+	})
+	tdr.insert()
+	tdr.docstatus = 1
+	tdr.save()
```
nextchamp-saqib:
```suggestion
	tdr.submit()
```

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.js
```diff
@@ -0,0 +1,32 @@
...
+				populate_doctypes_to_be_ignored(doctypes_to_be_ignored_array, frm);
+				frm.fields_dict['doctypes_to_be_ignored'].grid.set_column_disp('no_of_docs', false);
+				frm.refresh_field('doctypes_to_be_ignored');
+			}
+		});
+
+		frm.get_field('doctypes_to_be_ignored').grid.cannot_add_rows = true;
+		frm.fields_dict['doctypes_to_be_ignored'].grid.set_column_disp('no_of_docs', false);
+		frm.refresh_field('doctypes_to_be_ignored');
+	}
```
nextchamp-saqib:
```suggestion
	},

	refresh: function(frm) {
		frm.fields_dict['doctypes_to_be_ignored'].grid.set_column_disp('no_of_docs', false);
		frm.refresh_field('doctypes_to_be_ignored');
	}
	
```

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.js
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2021, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Transaction Deletion Tool', {
+	onload: function(frm) {
+		let doctypes_to_be_ignored_array;	
+		frappe.call({
```
nextchamp-saqib: Only populate if the `docstatus == 0`

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
```diff
@@ -0,0 +1,144 @@
...
+	def validate(self):
+		frappe.only_for('System Manager')
+		company_obj = frappe.get_doc('Company', self.company)
+		if frappe.session.user != company_obj.owner and frappe.session.user != 'Administrator':
+			frappe.throw(_('Transactions can only be deleted by the creator of the Company or the administrator.'), 
+				frappe.PermissionError)
+		doctypes_to_be_ignored_list = get_doctypes_to_be_ignored()
+		for doctype in self.doctypes_to_be_ignored:
+			if doctype.doctype_name not in doctypes_to_be_ignored_list:
+				frappe.throw(_("DocTypes should not be added manually to the 'DocTypes That Won't Be Affected' table."))
```
nextchamp-saqib:
```suggestion
				frappe.throw(_("DocTypes should not be added manually to the 'Excluded DocTypes' table. You are only allowed to remove entries from it. "), title=_("Not Allowed"))
```

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
```diff
@@ -0,0 +1,144 @@
...
+from frappe.model.document import Document
+from frappe import _
+from frappe.desk.notifications import clear_notifications
+
+class TransactionDeletionTool(Document):
+	def validate(self):
+		frappe.only_for('System Manager')
+		company_obj = frappe.get_doc('Company', self.company)
+		if frappe.session.user != company_obj.owner and frappe.session.user != 'Administrator':
+			frappe.throw(_('Transactions can only be deleted by the creator of the Company or the administrator.'),
```
nextchamp-saqib:
```suggestion
			frappe.throw(_('Transactions can only be deleted by the creator of the Company or the Administrator.'), 
```

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
```diff
@@ -0,0 +1,144 @@
...
+		# reset company values
+		company_obj.total_monthly_sales = 0
+		company_obj.sales_monthly_history = None
+		company_obj.save()
+		# Clear notification counts
+		clear_notifications()
+
+		singles = frappe.get_all('DocType', filters = {'issingle': 1}, pluck = 'name')
+		tables = frappe.get_all('DocType', filters = {'istable': 1}, pluck = 'name')
+		doctypes_to_be_ignored_list = singles + get_doctypes_to_be_ignored()
```
nextchamp-saqib: `get_doctypes_to_be_ignored` returns all the hard-coded doctypes. If a user removes a doctype from that hard-coded list, it will not be accounted for as all the doctypes are fetched here.

You should filter out the doctype names from `self.doctypes_to_be_ignored` and append to `doctypes_to_be_ignored_list`

nextchamp-saqib (changes requested):

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
GangaManoj: but isn't `self.doctypes_to_be_ignored` populated using `get_doctypes_to_be_ignored()` as well? So any changes to the hard-coded list would reflect on both, right? Unless the hard-coded list is changed after the table gets populated

nextchamp-saqib: It's other way around, the populated table may get changed by user. 

Consider if user removed 'Customer' from the excluded doctypes table after saving the document, then it won't be ignored as the ignored list is not derived from "self.doctypes_to_be_ignored"

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.js
GangaManoj: Done

erpnext/setup/doctype/transaction_deletion_tool/transaction_deletion_tool.py
GangaManoj: Ohh right, that makes sense.

nextchamp-saqib (approved):

PR #25374, opened 2021-04-17
fix: commit  changes to shipment status in database

ankush:
Problem: The status change is not committed to the database. 
Steps to reproduce: 
- Create any shipment. 
- Submit it.
- Refresh page (important)
- Problem 1: Staus will be Draft but the indicator will show correctly as shown in the screenshot.
- Cancel shipment.
- Refresh page (important)
- Problem 2: status will still be in draft but the indicator will show cancelled.

Solution:
- use `db_set` to commit the change.
- Add patch to fix existing records
- add CI rule to catch similar errors for `on_cancel`

image

### DIFF
--- a/erpnext/patches/v13_0/update_shipment_status.py
+++ b/erpnext/patches/v13_0/update_shipment_status.py
@@ -1,0 +1,14 @@
+import frappe
+
+def execute():
+	frappe.reload_doc("stock", "doctype", "shipment")
+
+	# update submitted status
+	frappe.db.sql("""UPDATE `tabShipment`
+					SET status = "Submitted"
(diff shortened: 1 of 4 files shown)


erpnext/patches/v13_0/update_shipment_status.py
```diff
@@ -0,0 +1,12 @@
+import frappe
+
+def execute():
+	# update submitted status
```
rohitwaghchaure, line 6: To fix travis add frappe.reload_doc('stock', 'doctype', 'shipment')

ankush: Done

rohitwaghchaure (approved):

PR #25375, opened 2021-04-17
fix: remove duplicate keys from dictionaries

ankush:
Problem: While just duplicate keys in dicts are not harmful, differing duplicate keys can be confusing or buggy.

solution: Kept the last one or the newest one based on expected behaviour.


ankush:
- Translation check is False positive. 
- Semgrep check is fixed in https://github.com/frappe/erpnext/pull/25374

rohitwaghchaure (approved):

PR #25380, opened 2021-04-18
fix: hiding Rounding Adjustment field

anupamvs:
Hiding **Rounding Adjustment** field if **Disable Rounded Total** is checked.


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -883,6 +884,7 @@
    "read_only": 1
   },
   {
+   "depends_on": "eval:!doc.disable_rounded_total",
```
ankush, line 887: `modified` is not changed in this file, so this change won't migrate.

ankush (approved):

PR #25407, opened 2021-04-20
fix: add payment term in sales invoice

AfshanKhan:
Steps to replicate:
1. Make a Sales Order with any Payment Terms Template.
2. Make a Delivery Note from that Sales Order.
3. Make a Sales Invoice from that Delivery Note.

The Payment Terms Template selected in Sales Order does not get fetched in the Sales Invoice.


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -419,6 +419,12 @@ def set_missing_values(self, for_validate=False):
 
 		super(SalesInvoice, self).set_missing_values(for_validate)
 
+		if not self.get('payment_terms_template'):
```
nextchamp-saqib: I think we should move this to `set_missing_values` of `make_sales_invoice` under `delivery_note.py`.

As it is being followed in purchase cycle

AfshanKhan: Made the changes

deepeshgarg007:
@AfshanKhan closing this, new PR will be opened after making this configurable

PR #25413, opened 2021-04-20
feat!: add pick batch button

18alantom:
### Old Functionality
- **Add Serial No** was used for both Serial and Batch number entries.  
- The dialog box triggered by this button is ambiguous.
before

### New Functionality
- Depending on whether item has serial number or batch number, **Add Serial No** or **Pick Batch No** is shown.
- If both are present **Add Serial No** is shown.
after

_Note : In the examples, Item ACTCK has only Batch No enabled, Item INFSTW has both enabled._

'no-docs'


Alchez:
@18alantom, shouldn't both the buttons show up if both serial numbers and batches are enabled?

Would it be better to split the Selector dialog, one each for serial and batch? And then let all the controls handle it accordingly?

18alantom:
@Alchez yeah ideally both should show, but the code responsible for the dialog box gives **Serial No**  higher precedence so both would lead to the serial number dialog showing. 

So this is just a tentative cosmetic PR to avoid confusion as to which dialog box is gonna pop up; incorporating both into a single dialog requires UX changes and much more refactoring.

erpnext/public/js/utils.js
```diff
@@ -48,31 +48,41 @@ $.extend(erpnext, {
...
+			['has_serial_no', 'has_batch_no'],
+			({has_serial_no, has_batch_no}) => {
+				grid_row.doc.has_serial_no = has_serial_no;
+				grid_row.doc.has_batch_no = has_batch_no;
+
+				if (has_serial_no) {
+					attachSelectorButton(__("Add Serial No"), grid_row.grid_form.fields_dict.serial_no.$wrapper);
+				} else if (has_batch_no) {
+					attachSelectorButton(__("Pick Batch No"), grid_row.grid_form.fields_dict.batch_no.$wrapper);
+				} else {
```
rohitwaghchaure: Why else?

erpnext/public/js/utils.js
```diff
@@ -48,31 +48,41 @@ $.extend(erpnext, {
...
+			let $btnDiv = $("<div>").css({"margin-bottom": "10px", "margin-top": "10px"})
+				.appendTo(appendLoction);
+			let $btn = $(`<button class="btn btn-sm btn-default">${innerText}</button>`)
+				.appendTo($btnDiv);
+
+			$btn.on("click", function() {
+				let callback = '';
+				let on_close = '';
+				me.show_serial_batch_selector(grid_row.frm, grid_row.doc,
+					callback, on_close, true);
```
rohitwaghchaure: just put single quotes instead of callback and on_close and remove the row 64 and 65

erpnext/public/js/utils.js
```diff
@@ -743,3 +736,14 @@ $(document).on('app_ready', function() {
 		});
 	}
 });
+
+function attachSelectorButton(innerText, appendLoction, context, grid_row) {
```
nabinhait: don't use came casing

nabinhait (changes requested):

nabinhait (approved):

PR #25415, opened 2021-04-20
feat(India): Bootstrapped GST Setup 

deepeshgarg007:
- Add separate GST accounts for Sales and Purchases on Company Setup
   - **Sales Accounts**:
      - Output Tax SGST
      - Output Tax CGST
      - Output Tax IGST
    - **Purchase Accounts**:
      - Input Tax SGST
      - Input Tax CGST
      - Input Tax IGST
 - Auto create item tax templates for all GST Tax Rates
 - Auto create Sales/Purchase Taxes and  Charges Template and add  tax category for In-State and Out-State to the templates
 - Made setup faster by removing redundant checks and validations
 - Automatically add default GST Accounts in GST Settings

docs: https://github.com/frappe/erpnext_documentation/pull/362


erpnext/setup/setup_wizard/operations/taxes_setup.py
```diff
@@ -160,12 +193,14 @@ def get_or_create_account(company_name, account):
...
+			'account_name': account.get('account_name'),
+			'account_number': account.get('account_number', ''),
+			'company': company_name
 		},
 		or_filters={
-			'account_name': account.get('account_name'),
-			'account_number': account.get('account_number')
+			'company': company_name,
+			'root_type': root_type,
+			'is_group': 0
```
barredterra: Company is both in `filters` and `or_filters`.

The intention for having `account_name` and `account_number` in `or_filters` was that each should be unique and one is enough to identify the account. Having both in filters, especially `'account_number' == ''` might not work.

erpnext/setup/setup_wizard/data/country_wise_tax.json
```diff
@@ -818,31 +818,266 @@
 	"India": {
 		"chart_of_accounts": {
 			"*": {
+				"tax_categories": [
```
barredterra: This approach for creating Tax Categories looks fine. I will change #25858 to use the same approach.

erpnext/setup/setup_wizard/operations/taxes_setup.py
```diff
@@ -147,8 +167,21 @@ def make_item_tax_template(company_name, template):
 			if 'tax_rate' not in tax_row:
 				tax_row['tax_rate'] = account_data.get('tax_rate')
 
-	return frappe.get_doc(template).insert(ignore_permissions=True)
-
+	doc = frappe.get_doc(template)
+	doc.flags.ignore_links = True
+	doc.flags.ignore_validate = True
+	doc.insert(ignore_permissions=True)
```
barredterra, line 183: Is all this ignoring really necessary? I don't remember getting any errors, at least for the german setup, and if there are any they might be worth fixing...

erpnext/setup/setup_wizard/data/country_wise_tax.json
barredterra: Are the tax categories specific to the chart of accounts? At least in Germany they are not. If this is the case generally, we could move tax categories two levels up, to be specific to country only.

deepeshgarg007: No they are not, will move them 2 levels up

erpnext/setup/setup_wizard/operations/taxes_setup.py
deepeshgarg007: This is not to ignore any errors but to reduce the setup time. India has a lot of fixtures that are created during setup which takes a lot of time. Have not done any profiting yet so cannot say if this helps or not. Will remove if there is no significant change

barredterra: I see. Could you please add a comment mentioning this?

erpnext/public/js/setup_wizard.js
```diff
@@ -139,36 +139,10 @@ erpnext.setup.slides_settings = [
...
-
 			if (!this.validate_fy_dates()) {
 				return false;
 			}
 
-			// Validate bank name
-			if(me.values.bank_account){
-				frappe.call({
-					async: false,
-					method: "erpnext.accounts.doctype.account.chart_of_accounts.chart_of_accounts.validate_bank_account",
```
nabinhait, line 153: Why this is removed? The code was there because many people enter just "Bank Accounts" in Bank field in Setup Wizard and it raises duplication error.

erpnext/setup/setup_wizard/operations/taxes_setup.py
nabinhait: @deepeshgarg007 I agree with Raffael, `account_number` and `account_name` should be in `or_filters`

nabinhait (changes requested):

nabinhait (approved):

PR #25417, opened 2021-04-21
fix: display reconcile tool when closing balance 0

hasnain2808:


erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.js
```diff
@@ -79,7 +79,6 @@ frappe.ui.form.on("Bank Reconciliation Tool", {
 					frm.doc.bank_account &&
 					frm.doc.bank_statement_from_date &&
 					frm.doc.bank_statement_to_date &&
```
nextchamp-saqib:
```suggestion
					frm.doc.bank_statement_to_date
```

nextchamp-saqib (changes requested):

nextchamp-saqib (approved):

PR #25421, opened 2021-04-21
feat: Enable custom field search on POS

AfshanKhan:
### Use-case:
Consider Item has custom fields that have values in a language different than system language.

This feature enables users to search for those items with the values in those custom or standard fields in Point of Sale.

Added a table in POS Setting called **"POS Search Fields"**
image

The options for **"Field"** are the fields present in the Item master.
image

After selection of any field that field's value will also be considered in the search for the item in Point of Sale.

### For example:
We have selected **"Description"**.
image

We added **"هواء"** in one of our Items description.
image

### Now if we go to Point of Sale.
Without searching anything this is the screen.
image

After we search with keyword **"هواء"**
image


erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -139,8 +139,16 @@ def get_conditions(item_code, serial_no, batch_no, barcode):
 	if serial_no or batch_no or barcode:
 		return "item.name = {0}".format(frappe.db.escape(item_code))
 
-	return """(item.name like {item_code}
-		or item.item_name like {item_code})""".format(item_code = frappe.db.escape('%' + item_code + '%'))
+	cond = "("
+	cond += """item.name like {item_code}
+		or item.item_name like {item_code}""".format(item_code = frappe.db.escape('%' + item_code + '%'))
+	search_fields = frappe.get_all('POS Search Fields', fields = ['fieldname'])
```
nextchamp-saqib, line 155: It would be better if we moved this to a seperate function.

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -139,8 +139,16 @@ def get_conditions(item_code, serial_no, batch_no, barcode):
...
+	if search_fields:
+		for field in search_fields:
```
nextchamp-saqib, line 157: We should probably verify if the field is not already used in the default condition. For eg. item_name

If such is the case, do not add the custom search condition. 

Or we could stop them from getting added to the child table itself by a python validation on POS settings.

AfshanKhan: The better approach in my opinion is to not show those in the drop-down itself. Hence it won't be available for selection and solve our issue. What do you say?

erpnext/selling/page/point_of_sale/point_of_sale.py
AfshanKhan: Okay will do that

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -81,7 +81,7 @@ def get_items(start, page_length, price_list, item_group, pos_profile, search_va
 			condition=condition,
 			bin_join_selection=bin_join_selection,
 			bin_join_condition=bin_join_condition
-		), {'warehouse': warehouse}, as_dict=1)
+		), {'warehouse': warehouse}, as_dict=1, debug=True)
```
nextchamp-saqib:
```suggestion
		), {'warehouse': warehouse}, as_dict=1)
```

nextchamp-saqib (approved):

erpnext/accounts/doctype/pos_settings/pos_settings.js
```diff
@@ -21,6 +29,39 @@ frappe.ui.form.on('POS Settings', {
...
+				debugger;
+				if (search_fields_datatypes.includes(d.fieldtype) && !(do_not_include_fields.includes(d.fieldname))) {
+					return [d.label];
+				} else {
+					return null;
+				}
+			});
+
+			fields.unshift('');
+			frm.set_df_property('pos_search_fields', 'options', fields, cur_frm.docname, 'field');
```
nextchamp-saqib:
```suggestion
			frm.fields_dict.pos_search_fields.grid.update_docfield_property('field', 'options', fields);
```

erpnext/accounts/doctype/pos_settings/pos_settings.js
```diff
@@ -21,6 +29,39 @@ frappe.ui.form.on('POS Settings', {
 			);
 		});
 
+	},
+
+	add_search_options: function(frm) {
+		frappe.model.with_doctype("Item", () => {
+			var fields = $.map(frappe.get_doc("DocType", "Item").fields, function(d) {
+				debugger;
```
nextchamp-saqib:
```suggestion

```

erpnext/accounts/doctype/pos_settings/pos_settings.js
nextchamp-saqib: Tried locally, the `update_docfield_property` works

nextchamp-saqib (changes requested):

PR #25422, opened 2021-04-21
fix: state code for Other Territory

finbyz:
Other Territory state code should be 97.

It's already corrected in version-13 but missed in version-12

> Screenshots/GIFs


ankush:
Original PR: https://github.com/frappe/erpnext/pull/24993

deepeshgarg007 (approved):

PR #25425, opened 2021-04-21
fix: change subcontracted item display

18alantom:
### Before Fix
Screen Shot 2021-04-21 at 18 47 21

### After Fix
The Subcontracted Item field now shows only the Subcontracted Item Code
Screen Shot 2021-04-21 at 18 48 24


ankush (approved):
LGTM. Required in link fields.

PR #25454, opened 2021-04-23
fix: circular loop in project task

shadrak98:
Removed circular loop of assigning parent task by adding filter in parent task.

image


erpnext/projects/doctype/task/task.js
```diff
@@ -32,7 +32,8 @@ frappe.ui.form.on("Task", {
 
 		frm.set_query("parent_task", function () {
 			let filters = {
-				"is_group": 1
+				"is_group": 1,
+				name: ["!=", frm.doc.name]
```
ankush:
```suggestion
				"name": ["!=", frm.doc.name]
```
for consistency

ankush (approved):

Issue #25338, opened 2021-04-14
Change of paid amount does not update allocated amount in payment entry. Both pay and receive type. - v13.0.1


muchai:
Create payment entry.

Change paid amount to a figure e.g. less than the invoice amount.

System doesn't update allocated amount.

ukwuezeobinna:
I noticed this also

PR #25460, opened 2021-04-23
fix: to update Allocated amount after Paid Amount is changed in PE

Anuja-pawar:
**Issue:** Allocated Amount doesn't get updated after the Paid amount is changed in Payment Entry.

**Steps to Replicate:**
1. Create a Payment Entry.
2. Change Paid Amount After the Outstanding Invoices are fetched.

**Before:**
PE_before

**After:**
PE_after

**Fixes** https://github.com/frappe/erpnext/issues/25338

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.js
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.js
@@ -582,7 +582,7 @@ frappe.ui.form.on('Payment Entry', {
 		}
 
 		if(frm.doc.payment_type == "Receive")
-			frm.events.allocate_party_amount_against_ref_docs(frm, frm.doc.paid_amount);
+			frm.events.allocate_party_amount_against_ref_docs(frm, frm.doc.paid_amount, 1);
 		else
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -800,22 +800,18 @@ frappe.ui.form.on('Payment Entry', {
 				//If allocate payment amount checkbox is unchecked, set zero to allocate amount
 				row.allocated_amount = 0;
 
-			} else if (frappe.flags.allocate_payment_amount != 0 && !row.allocated_amount) {
+			} else if (frappe.flags.allocate_payment_amount != 0 && (!row.allocated_amount || paid_amount_change)) {
 				if (row.outstanding_amount > 0 && allocated_positive_outstanding > 0) {
```
deepeshgarg007:
```suggestion
				if (row.outstanding_amount > 0 && allocated_positive_outstanding >= 0) {
```
You won't need the second else if condition then

deepeshgarg007 (approved):

PR #25430, opened 2021-04-21
feat: stock entry by warehouse

dj12djdjs:
From feature request: #25429

PR #25462, opened 2021-04-23
feat: stock entry by source warehouse

### LINKED ISSUE #25429: Create Stock Entry moving all items from warehouse A to warehouse B.
I've found this use case is a common request and most people seem to be doing it manually. 
https://discuss.erpnext.com/t/how-to-move-items-form-warehouse-to-another/26570/3
https://discuss.erpnext.com/t/moving-all-stock-item-quantities-from-warehouse-a-to-b/31643/4
https://discuss.erpnext.com/t/bulk-transfer-of-material-from-one-warehouse-to-another-warehouse/59444/2
https://discuss.erpnext.com/t/is-there-a-way-to-transfer-entire-contents-of-a-one-warehouse-to-another/42763/10

As part of my specific work flow I transfer about 300 items from one warehouse to another about 3 times.
1st stock transfer is going to a outgoing warehouse where the goods are prepped for shipping.
2nd stock transfer is going from the outgoing warehouse to the other warehouse "receiving warehouse". During this time shipping is added the expenses.
3rd stock transfer is going from the receiving warehouse to their respective bins.

Adding these items one by one or by data import is still very slow.

I've added a feature that works for my use case. Maybe this works for other use cases as well.
stock-entry-from-warehouse

dj12djdjs:
port of: #25430 

closes #25429


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1734,3 +1734,32 @@ def validate_sample_quantity(item_code, sample_quantity, qty, batch_no = None):
...
+		and i.has_variants = 0 and i.has_serial_no = 0 and i.has_batch_no = 0
+		and exists(select name from `tabWarehouse` where lft >= %s and rgt <= %s and name=bin.warehouse)
+	""", (lft, rgt))
+
+	res = []
+	for d in set(items):
+		stock_bal = get_stock_balance(d[0], d[2], posting_date, posting_time, with_valuation_rate=True)
+
+		res.append({
+			"item_code": d[0],
```
ankush: Use `as_dict` in `frappe.db.sql` so you can use column name instead of index.

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1734,3 +1734,32 @@ def validate_sample_quantity(item_code, sample_quantity, qty, batch_no = None):
...
+
+@frappe.whitelist()
+def get_items_from_warehouse(warehouse, posting_date, posting_time, company):
+	lft, rgt = frappe.db.get_value("Warehouse", warehouse, ["lft", "rgt"])
+	items = frappe.db.sql("""
+		select i.name, i.item_name, bin.warehouse, i.item_group, i.description, i.stock_uom
+		from tabBin bin, tabItem i
+		where i.name=bin.item_code and i.disabled=0 and i.is_stock_item = 1
+		and i.has_variants = 0 and i.has_serial_no = 0 and i.has_batch_no = 0
+		and exists(select name from `tabWarehouse` where lft >= %s and rgt <= %s and name=bin.warehouse)
```
ankush: This is a correlated subquery. Not sure why it's required. Why not just join the tables?

erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1734,3 +1734,32 @@ def validate_sample_quantity(item_code, sample_quantity, qty, batch_no = None):
...
 		sample_quantity = qty_diff
 	return sample_quantity
+
+@frappe.whitelist()
+def get_items_from_warehouse(warehouse, posting_date, posting_time, company):
+	lft, rgt = frappe.db.get_value("Warehouse", warehouse, ["lft", "rgt"])
+	items = frappe.db.sql("""
+		select i.name, i.item_name, bin.warehouse, i.item_group, i.description, i.stock_uom
+		from tabBin bin, tabItem i
+		where i.name=bin.item_code and i.disabled=0 and i.is_stock_item = 1
```
ankush:
```suggestion
		on i.name=bin.item_code where i.disabled=0 and i.is_stock_item = 1
```

Prefer explicit joins over implicit joins.

erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -273,6 +273,31 @@ frappe.ui.form.on('Stock Entry', {
 					}
 				})
 			}, __("Get Items From"));
+
+			frm.add_custom_button(__("Source Warehouse"), function() {
+				if(!frm.doc.from_warehouse) frappe.throw(__("Please select source warehouse!"));
```
ankush: For consistency, it would be better to open a modal dialogue that asks for selecting a warehouse? 🤔 

Take a look at how it's done in Stock Reconciliation.

ankush (changes requested):

erpnext/stock/doctype/stock_entry/stock_entry.py
dj12djdjs: I was lazy. Just copied from here: https://github.com/frappe/erpnext/blob/59769c7a8685fb22ba18cccd9283e2e912784b59/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py#L477

I'll re-write the query joining the tables.

muchai:
Let me see about including inventory dimensions. Our use case is stock is stored in numbered crates. Which are moved too. We manage this has inventory dimensions, and as well as serialized inventory.

PR #25486, opened 2021-04-27
fix: stock move dialog duplicate submit actions (V13)

dj12djdjs:
Port #24859 to V13


ankush (approved):
LGTM. We should remove this once we have better safeguards.

PR #25492, opened 2021-04-27
fix: unauthorized issue due to token expiry

KaviyaPeriyasamy:
This PR fixes the below issue, 

```Traceback (most recent call last):
  File "/home/frappe/frappe-bench/apps/frappe/frappe/integrations/utils.py", line 42, in make_post_request
    frappe.flags.integration_request.raise_for_status()
  File "/home/frappe/frappe-bench/env/lib/python3.6/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error:  for url: https://gsp.adaequare.com/enriched/ei/api/invoice
```

** After this [condition](https://github.com/frappe/erpnext/blob/version-12/erpnext/regional/india/e_invoice/utils.py#L596) only, updating of new token happens. But in our scenario, the token expires before the expiry time.

### DIFF
--- a/erpnext/regional/india/e_invoice/utils.py
+++ b/erpnext/regional/india/e_invoice/utils.py
@@ -664,8 +664,12 @@ class GSPConnector():
 			self.raise_error()
 
 		except Exception:
-			log_error()
-			self.raise_error(True)
+			if '401 Client Error' in str(sys.exc_info()[1]):
+				self.fetch_auth_token()
+				self.fetch_gstin_details(gstin)
+			else:
+				log_error()
+				self.raise_error(True)
 	
 	@staticmethod
 	def get_gstin_details(gstin):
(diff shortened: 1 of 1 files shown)


karthikeyan5:
can be temporarily fixed (i.e. you will face the problem once a month) by setting "API Access Token Validity" to One month in the https://gsp.adaequare.com/ portal.

image

karthikeyan5:
```
Maybe we should reduce the duration for which it tries to regenerate the token. 

Since the expiry is of one month maybe we can reduce it to 15 days? Or 10?

I don't like the idea of redirecting every 401 requests to generate token

```
@nextchamp-saqib in a private conversation

KaviyaPeriyasamy:
Also, 401 client error fixed by the following, (note: this is more of a workaround)
1. Go to E Invoice Settings and open the browser console.
2. Check the doc's token_expiry by ```cur_frm.doc```
3. Then execute this ```cur_frm.doc.token_expiry = null;```
4. Finally save and reload the document.

Closing this PR for now as it is fixed by the above method.

Issue #25496, opened 2021-04-28
issue with installing the requirements.txt


Kevin-Justin:
![image](https://user-images.githubusercontent.com/75354764/116331676-d9a80400-a802-11eb-9f7b-ece29aeb6231.png)

ankush:
Hi @Kevin-Justin 

If you were trying to install using `pip install -r requirements.txt` then it will not work. Please follow the installation guide here: https://frappeframework.com/docs/user/en/installation or consider using docker: https://github.com/frappe/erpnext#containerized-installation

PR #25501, opened 2021-04-28
chore: remove frappe from requirements.txt

ankush:
Due to recent changes in pip dependency resolver, in some random cases, pip thinks frappe is not installed and tries to fetch it from pypi, which results in errors like #25496

Until a better solution is available, frappe should not be part of requirements.txt

`frappe` in requirements.txt does nothing other than indicating dependency to the reader.

Alternatives:
- use relative path and editable install `-e '../frappe'`. Con: hardcoding path, might cause reinstall. Pros: will never fetch from pypi.
- use `git:` url in requirements.txt. Con: This will cause unnecessary re-fetching of the package, will have to deal with versioning separately. 
- use old dependency resolver and older pip (this is removed in latest pip) `--use-deprecated=legacy-resolver`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-frappe
+# frappe   # https://github.com/frappe/frappe is installed during bench-init
 gocardless-pro~=1.22.0
 googlemaps  # used in ERPNext, but dependency is defined in Frappe
 pandas~=1.1.5


Alchez:
If we want to retain readability, we could keep it on as a comment with a note. Not using the alternatives makes sense to me, outside of not publishing the Frappe app on pypi.

gavindsouza (approved):
Apparently, if you have an editable install, and requirements.txt says a non-editable install, this will force a reinstall in non-editable mode. ref: https://github.com/pypa/pip/blob/7a77484a492c8f1e1f5ef24eaf71a43df9ea47eb/src/pip/_internal/resolution/resolvelib/resolver.py#L156

Removing `frappe` from `requirements.txt` seems for the best. 

Although I tried adding some initial checks to ensure Frappe was installed via `setup.cmdclass`, it doesn't seem to cover "all" scenarios. The check is supposed to be a guard against bad installation methods.

```python
def check_frappe():
	import sys, json, subprocess
	class MissingDependencyError(LookupError):
		pass

	data = (
		subprocess.check_output(
			f"{sys.executable} -m pip list --format json", shell=True, encoding="utf8", env={},
		)
		or "[]"
	)
	available_modules = json.loads(data)

	if not filter(lambda x: x["name"] == "frappe", available_modules):
		raise MissingDependencyError(
			f"You must have the Frappe Framework installed before ERPNext!"
		)

class DevelopSetupWrapper(develop):
	def __init__(self, *args, **kwargs):
		self.announce("Running develop install")
		atexit.register(check_frappe)
		super().__init__(*args, **kwargs)

	def run(self):
		check_frappe()
		develop.run(self)

setup(
	name="erpnext",
	version=version,
	description="Open Source ERP",
	author="Frappe Technologies",
	author_email="info@erpnext.com",
	cmdclass={"develop": DevelopSetupWrapper},
	packages=find_packages(),
	zip_safe=False,
	include_package_data=True,
	install_requires=install_requires,
)
```

The alternatives proposed don't work with the current Frappe app setup pipelines either, unless we expend more engineering efforts. 

Another method we tried was adding `install_requires.append(f"frappe @ file://{os.getcwd()}/../frappe/" )` to the `setup.py` but that makes the frappe install non-editable.

PR #25506, opened 2021-04-28
fix: item stock levels displaying inconsistently

noahjacob:
**Issue :**

- Opening Item the first time.
image
- Opening it the second time without refreshing.
image

**After FIx**
dash_fix

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -477,11 +477,10 @@ $.extend(erpnext.item, {
 
 								let no_of_combinations = lengths.reduce((a, b) => a * b, 1);
 								let msg;
-								if (no_of_combinations === 1){
-									msg = __("Make {0} Variant").format(no_of_combinations);
-								}
-								else {
-									msg = __("Make {0} Variants").format(no_of_combinations);
+								if (no_of_combinations === 1) {
+									msg = __("Make {0} Variant", [no_of_combinations]);
+								} else {
+									msg = __("Make {0} Variants", [no_of_combinations]);
 								}
 								me.multiple_variant_dialog.get_primary_btn().html(msg);
 								me.multiple_variant_dialog.enable_primary_action();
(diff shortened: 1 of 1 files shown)


ankush:
@noahjacob semgrep failures are unrelated to your change but can you add a separate commit fixing those too?

ankush:
A longer explanation of this bug:

item-dashboard is loaded using async call in `item.js`:
https://github.com/frappe/erpnext/blob/7f932cfb30f32224cbf500efa86ea7d87462b543/erpnext/stock/doctype/item/item.js#L382

This makes a network request the first time it is called, and then it's stored in localStorage.

The next request uses a cached version. 

Also in `item.js` following line is called after updating item attribute details, this causes race condition:

https://github.com/frappe/erpnext/blob/7f932cfb30f32224cbf500efa86ea7d87462b543/erpnext/stock/doctype/item/item.js#L771

When this line is executed in...

1. First request - item_dashboard.js is still loading so nothing is done there.
2. Subsequent requests - item_dashboard.js is already loaded and it's making a network request to get item data.  During this time the section is empty hence `refresh_sections()` removes it.

Solution: By reordering execution, this race condition can be avoided.

erpnext/stock/doctype/item/item.js
```diff
@@ -473,11 +474,16 @@ $.extend(erpnext.item, {
...
-									.html(__(
-										`Make ${no_of_combinations} Variant${no_of_combinations === 1 ? '' : 's'}`
-									));
+								let msg;
+								if (no_of_combinations === 1){
+									msg = __("Make {0} Variant").format(no_of_combinations);
+								}
+								else {
+									msg = __("Make {0} Variants").format(no_of_combinations);
+								}
```
surajshetty3416:
```suggestion
								if (no_of_combinations === 1) {
									msg = __("Make {0} Variant", [no_of_combinations]);
								} else {
									msg = __("Make {0} Variants", [no_of_combinations]);
								}
```

surajshetty3416 (changes requested):

ankush (approved):

sagarvora:
@ankush @noahjacob, has this been backported to `v13`?

**EDIT:** This code is present, but does not seem to work as intended (stock levels go away if you refresh any item page), we're working on a better fix.

ankush:
@sagarvora it was backported but it broke after https://github.com/frappe/frappe/pull/13355

check comments.

PR #25509, opened 2021-04-28
feat: Exchange rate difference handling in procurement cycle

GangaManoj:
**_Issue:_**
When a Purchase Invoice is created from a Purchase Receipt(or vice versa), if the exchange rate is different at the time of creation of the latter, the GL entries created for the _Stock Received But Not Billed_ account will not be balanced.

Screenshot 2021-04-29 at 4 15 18 AM

**_Fix:_**
The amount for the _Stock Received But Not Billed_ account for the second transaction will be made equal to that of the first and the difference will be assigned to the _Exchange Gain/Loss_ account.

For the Purchase Receipt:

Screenshot 2021-04-29 at 4 32 30 AM

For the corresponding Purchase Invoice(created after the exchange rate was changed):

Screenshot 2021-04-29 at 4 35 09 AM

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -568,7 +568,7 @@ class PurchaseInvoice(BuyingController):
 							print("*" * 100)
 							print("2")
 
-					else:			# delete comment later - add condition and change GL entries here? $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+					else:
 						if not self.is_internal_transfer():
 							gl_entries.append(
 								self.get_gl_dict({
@@ -637,8 +637,7 @@ class PurchaseInvoice(BuyingController):
 								expense_account = service_received_but_not_billed_account
 
 					if not self.is_internal_transfer():
-						# check exchange rate difference
-						# purchase_receipt_name = self.as_dict()['items'][0]['purchase_receipt']
+						# check if the exchange rate has changed
 						purchase_receipt_name = item.purchase_receipt
 						purchase_receipt = frappe.get_doc('Purchase Receipt', purchase_receipt_name)
 						if self.conversion_rate == purchase_receipt.conversion_rate:
@@ -651,12 +650,10 @@ class PurchaseInvoice(BuyingController):
 								}, account_currency, item=item))
 		
 						else:
-							# Purchase Invoice: item.expense_account - Stock Received But Not Billed 
 							gle = frappe.get_all('GL Entry')
 							for gl_entry_name in gle:
 								gl_entry = frappe.get_doc('GL Entry', gl_entry_name)
 								if gl_entry.voucher_no == purchase_receipt.name:
-									# if gl_entry.account == self.as_dict()['items'][0]['expense_account']:
 									if gl_entry.account == item.expense_account:
 										debit_at_old_exchange_rate = gl_entry.credit
 										break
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -627,13 +635,47 @@ def make_item_gl_entries(self, gl_entries):
...
-						gl_entries.append(self.get_gl_dict({
-								"account": expense_account,
-								"against": self.supplier,
-								"debit": amount,
-								"cost_center": item.cost_center,
-								"project": item.project or self.project
-							}, account_currency, item=item))
+						# check if the exchange rate has changed
+						purchase_receipt_name = item.purchase_receipt
+						purchase_receipt = frappe.get_doc('Purchase Receipt', purchase_receipt_name)
```
deepeshgarg007: No need to get the entire doc. Use frappe.db.get_value('Purchase Receipt', {'name': item.purchase_receipt}, ['conversion_factor'])

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -627,13 +635,47 @@ def make_item_gl_entries(self, gl_entries):
...
+							gl_entries.append(
+								self.get_gl_dict({
+									"account": expense_account,
+									"against": self.supplier,
+									"debit": debit_at_old_exchange_rate,
+									"cost_center": item.cost_center,
+									"project": item.project or self.project
+								}, account_currency, item=item)
+							)
+							discrepancy_caused_by_exchange_rate_difference = amount - debit_at_old_exchange_rate
```
deepeshgarg007: No need to query GL Entries for difference amount. This you can calculate using (item.qty * item.rate) * (purchase_receipt_converison_factor - purchase_invoice_conversion_factor)

Also no need for if else block. Only if condition should be sufficient. If purchase_receipt_converison_factor != purchase_invoice_conversion_factor then just add 2 additional entries for difference amount

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
GangaManoj: Done

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
GangaManoj: Done

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -634,6 +634,29 @@ def make_item_gl_entries(self, gl_entries):
...
 							}, account_currency, item=item))
+		
+						# check if the exchange rate has changed
+						purchase_receipt_conversion_rate = frappe.db.get_value('Purchase Receipt', {'name': item.purchase_receipt}, ['conversion_rate'])
+						if self.conversion_rate != purchase_receipt_conversion_rate:
+							gl_entries.append(
+								self.get_gl_dict({
+									"account": expense_account,
+									"against": self.supplier,
+									"debit": item.qty * item.rate * purchase_receipt_conversion_rate,
```
deepeshgarg007: @GangaManoj expense amount entry is already passed as per exchange rate in Purchase Invoice. Only difference amount should be handled here

deepeshgarg007 (changes requested):
- Similar entries should be passed when Purchase Receipt from Purchase Invoice is created
- Maybe this can be moved to accounts_controller.py and commodified for both scenarios
- Add a test case for this

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
GangaManoj: Done

PR #25517, opened 2021-04-29
fix: fieldname when updating docfield property

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/25516

### DIFF
--- a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.js
+++ b/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.js
@@ -234,7 +234,7 @@ erpnext.accounts.PaymentReconciliationController = frappe.ui.form.Controller.ext
 		});
 
 		if (invoices) {
-			this.frm.fields_dict.payment.grid.update_docfield_property(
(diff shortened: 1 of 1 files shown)


muchai:
Could be missing something, but something's not really ok with the fix.

It now displays payments, and allows selection of an invoice.

But it doesn't display invoices, until at least an invoice is selected in the payments section.

AfshanKhan:
> Could be missing something, but something's not really ok with the fix.
> 
> It now displays payments, and allows selection of an invoice.
> 
> But it doesn't display invoices, until at least an invoice is selected in the payments section.

@muchai can you please share a video or gif of the problem?

muchai:
After clicking get unreconciled entries, doesn't display open invoices
image

After selecting at least one invoice
image

When I leave the invoice number field blank, and click on reconcile.
image

In short, this only allows reconciling a payment one invoice, at a time, as is you can't use one payment to reconcile multiple invoices.

AfshanKhan:
@muchai Thanks for the detailed explanation. I get your point. I'll get back to you on this.

PR #25519, opened 2021-04-29
feat: add pending qty section to batch/serial selector dialog

18alantom:
### Description
Adds 3 new fields to the _Batch/ Serial No_ selector dialog if _BOM No_ and _For Quantity_ have been entered.
1. _Required Qty_ : Qty of an item required to manufacture  given _For Quantity_
2. _Pending Qty_ : Qty required to fulfill _Required Qty_, calculated across multiple rows.
3. _Total Selected Qty_ : Total qty of an item selected across multiple rows.

Along with the above changes, _Total Qty_ has been renamed to _Selected Qty_.

### Before
No indication of how much quantity has to be selected.
before

### After
after

_Note : the dual dialogs is an annoying, unrelated issue which needs fixing._

'no-docs'


erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -313,7 +315,21 @@ erpnext.SerialNoBatchSelector = Class.extend({
 
 		qty_field.set_input(total_qty);
 	},
+	update_pending_qtys: function() {
+		const pendingQtyField = this.dialog.fields_dict.pending_qty;
```
nabinhait: We don't use camel casing in variable names, please rename all variables using "_" like `pending_qty_field`

erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -511,3 +528,60 @@ erpnext.SerialNoBatchSelector = Class.extend({
 		];
 	}
 });
+
+function getPendingQtyFields(me) {
```
nabinhait: don't use camel case in function name either.

nabinhait (changes requested):

nabinhait (approved):

PR #25526, opened 2021-04-29
feat: Add Create Expense Claim button in Delivery Trip

GangaManoj:
_Proposed Changes:_

- **Delivery Note:**
1.  Add a button that allows you to create an Expense Claim record from a Delivery Trip record, if the driver is an employee.
2. Add a new field called _Employee_ that displays the Employee Code for the driver.

Screenshot 2021-04-30 at 3 57 10 AM

- **Expense Claim:**
1. Add a new field called _Delivery Trip_ whose value is fetched from the Delivery Trip record from which the expense claim is created. This field will only be displayed if it's non-empty.
2. Fetch the Employee details from the Delivery Trip doc from which the Expense Claim record is created.

photo_2021-05-04 05 40 13

Docs: https://github.com/frappe/erpnext_documentation/pull/312


erpnext/stock/doctype/delivery_trip/delivery_trip.js
```diff
@@ -40,6 +40,17 @@ frappe.ui.form.on('Delivery Trip', {
 		})
 	},
 
+	onload: function (frm) {
+		// if (frm.doc.docstatus == 1) {
```
nextchamp-saqib: Uncomment this?

nextchamp-saqib (changes requested):

erpnext/hr/doctype/expense_claim/expense_claim.json
```diff
@@ -365,13 +366,19 @@
    "label": "Total Taxes and Charges",
    "options": "Company:company:default_currency",
    "read_only": 1
+  },
+  {
+   "fieldname": "delivery_trip",
```
nextchamp-saqib, line 372: Only show this field if it is set, keep it hidden if not set.

erpnext/stock/doctype/delivery_trip/delivery_trip.js
GangaManoj: Done

erpnext/hr/doctype/expense_claim/expense_claim.json
GangaManoj: Done

nextchamp-saqib:
@GangaManoj  Just realized, the Employee field is not mandatory in Driver doctype, so there might be cases where Employee won't be fetched from Driver. In such cases, the Expense Claim button should be hidden.

erpnext/stock/doctype/delivery_trip/delivery_trip.py
```diff
@@ -394,3 +395,15 @@ def get_driver_email(driver):
...
 	email = frappe.db.get_value("Employee", employee, "prefered_email")
 	return {"email": email}
+
+@frappe.whitelist()
+def make_expense_claim(source_name, target_doc=None):
+	doc = get_mapped_doc("Delivery Trip", source_name,
+		{"Delivery Trip": {
+			"doctype": "Expense Claim",
+			"field_map": {
+				"delivery_trip": "name"
```
nabinhait:
```suggestion
				"name": "delivery_trip"
```

nabinhait (changes requested):

erpnext/stock/doctype/delivery_trip/delivery_trip.py
GangaManoj: Done

erpnext/stock/doctype/delivery_trip/test_delivery_trip.py
```diff
@@ -7,7 +7,8 @@
 
 import erpnext
 import frappe
-from erpnext.stock.doctype.delivery_trip.delivery_trip import get_contact_and_address, notify_customers
+from erpnext.stock.doctype.delivery_trip.delivery_trip import get_contact_and_address, notify_customers, make_expense_claim
+from erpnext.hr.doctype.employee.test_employee import make_employee
```
nextchamp-saqib:
```suggestion
```

nabinhait:
@GangaManoj Why 16 commits are required for this feature?

GangaManoj:
@nabinhait It started out with 3 and additional commits were made to correspond to each suggestion that was made during the review process + to fix Sider issues. I thought I should make separate ones for each, since they all serve different purposes. Is that a problem? Is there a convention I should be following?

PR #25559, opened 2021-05-03
fix: prevent spurious defaults for items when making prec from dnote

18alantom:
### Description
Adds conditional that prevents spurious values from **Purchase Order** to be added to **Sales Order Item** map in the absence of sales order detail in the map.

This was preventing **Internal Purchase Receipt**s from being created from **Delivery Note**s.

(fix by @rohitwaghchaure)

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -1761,7 +1761,8 @@ def get_delivery_note_details(internal_reference):
 		filters={'parent': internal_reference})
 
 	for d in si_item_details:
-		so_item_map.setdefault(d.name, d.so_detail)
+		if d.so_detail:
+			so_item_map.setdefault(d.name, d.so_detail)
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1761,7 +1761,8 @@ def get_delivery_note_details(internal_reference):
 		filters={'parent': internal_reference})
 
 	for d in si_item_details:
-		so_item_map.setdefault(d.name, d.so_detail)
+		if d.so_detail:
+			so_item_map.setdefault(d.name, d.so_detail)
```
ankush: `d.name` is supposed to be unique so `setdefault` is not required. If you just want to exclude `d.so_details = None` then it can be done in single dict comprehension too.



```python
so_item_map = {d.name: d.so_detail for d in si_item_details if d.so_detail}
```

ankush (approved):

PR #25561, opened 2021-05-03
fix: prevent spurious defaults for items when making prec from dnote

18alantom:
Backported from [#25559 ](https://github.com/frappe/erpnext/pull/25559)


18alantom:
Wrong branch for some reason.

PR #25564, opened 2021-05-03
fix: Updating Standard Notification's channel field

shariquerik:
Some Standard Notification does not have a `channel` field, since its a mandatory field we cannot edit those Notification
channel


shariquerik:
Patch: https://github.com/frappe/frappe/pull/13101

nabinhait (approved):

PR #25568, opened 2021-05-04
fix: use prec qty to update porder received qty

18alantom:
### Description
- **Purchase Order**'s `received_qty` is calculated using **Purchase Receipt**'s`qty` which is the Accepted Quantity.
- This allows for additional **Purchase Receipt**'s to be made against a **Purchase Order** in case there is rejected quantity in the **Purchase Receipt**.

### Example
PR with Rejected Quantity
Screen Shot 2021-05-04 at 11 39 06

#### Before
Screen Shot 2021-05-04 at 11 37 32
Screen Shot 2021-05-04 at 11 37 53

#### After
Screen Shot 2021-05-04 at 11 43 52
Screen Shot 2021-05-04 at 11 43 41

(fix by @rohitwaghchaure)


nabinhait:
@18alantom @ankush "Received %" in PO should be updated based on received_qty in PR only, including both accepted and rejected qty. If in case, the user returns those rejected items, they should make a return entry, which will reverse the effect of PO updation. In some cases, the supplier does not provide replacement for rejected items, then PO should be fully received.

nabinhait:
@18alantom @ankush "Received %" in PO should be updated based on received_qty in PR only, including both accepted and rejected qty. If in case, the user returns those rejected items, they should make a return entry, which will reverse the effect of PO updation. In some cases, the supplier does not provide replacement for rejected items, then PO should be fully received.

PR #25569, opened 2021-05-04
fix: use prec qty to update porder received qty

18alantom:
Backported from [#25568](https://github.com/frappe/erpnext/pull/25568).


nabinhait:
@18alantom @ankush "Received %" in PO should be updated based on received_qty in PR only, including both accepted and rejected qty. If in case, the user returns those rejected items, they should make a return entry, which will reverse the effect of PO updation. In some cases, the supplier does not provide replacement for rejected items, then PO should be fully received.

PR #25590, opened 2021-05-05
fix: use get_serial_nos for splitting

18alantom:
### Description
- Uses `get_serial_nos`.
- Different functions for setting and updating causes issues.
- Example : when there are trailing whitespaces.


rohitwaghchaure (approved):

manasan3010:
@18alantom, Why not clean the User's Input on the client side by modifying the JS function `erpnext.TransactionController.prototype.serial_no`

Issue #25579, opened 2021-05-04
Outdated web-hooks version for Shopify connection.


rakshithrddy:
The webhooks version currently used in Erpnext develop for Shopify extension is 2020-04, but in Shopify, the version 2020-04 is not currently supported/outdated

![Screenshot from 2021-05-04 19-32-50](https://user-images.githubusercontent.com/36509967/117015714-9de3d180-ad0f-11eb-96ad-6115761cc25c.png)

PR #25600, opened 2021-05-06
fix: update shopify api version

ankush:
Bump Shopify API version from 2020-04 to 2021-04 (the latest stable version.) 

closes https://github.com/frappe/erpnext/issues/25579

The current connector uses orders, products and webhooks endpoints. No breaking change in all 3. 

Reference: 

- https://shopify.dev/concepts/about-apis/versioning/release-notes/2020-10
- https://shopify.dev/concepts/about-apis/versioning/release-notes/2021-01
- https://shopify.dev/concepts/about-apis/versioning/release-notes/2021-04

### DIFF
--- a/erpnext/erpnext_integrations/connectors/shopify_connection.py
+++ b/erpnext/erpnext_integrations/connectors/shopify_connection.py
@@ -335,13 +335,13 @@ def get_url(shopify_settings):
 
 	if not last_order_id:
 		if shopify_settings.sync_based_on == 'Date':
-			url = get_shopify_url("admin/api/2020-10/orders.json?limit=250&created_at_min={0}&since_id=0".format(
+			url = get_shopify_url("admin/api/2021-04/orders.json?limit=250&created_at_min={0}&since_id=0".format(
 				get_datetime(shopify_settings.from_date)), shopify_settings)
 		else:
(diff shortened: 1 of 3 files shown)


erpnext/erpnext_integrations/connectors/shopify_connection.py
```diff
@@ -335,13 +335,13 @@ def get_url(shopify_settings):
 
 	if not last_order_id:
 		if shopify_settings.sync_based_on == 'Date':
-			url = get_shopify_url("admin/api/2020-10/orders.json?limit=250&created_at_min={0}&since_id=0".format(
+			url = get_shopify_url("admin/api/2021-04/orders.json?limit=250&created_at_min={0}&since_id=0".format(
```
Alchez, line 338: Might be useful to store this in a global constant in the Shopify module and use that everywhere. It'll be easier to update in the future.

ankush: Planned to be changed in future update. 

Right now options are to create a new file or require circular imports, so it will create more mess.

Alchez: Ah, I see. No problem then.

PR #25613, opened 2021-05-06
chore: replace assertEquals with alias assertEqual

ankush:
`assertEquals` has been deprecated, while not fully removed it will still keep giving warnings in tests / CI.

ref: https://docs.python.org/3/library/unittest.html#deprecated-aliases


ankush:
9/10 sider warnings were false positive. Spaces were used in a separate section, so it wouldn't cause failure. Fixed anyway.

nabinhait (approved):

PR #25635, opened 2021-05-08
ci: Enable Parallel tests

surajshetty3416:
### Run unit tests parallelly
across multiple instances to reduce wait time for green tick. 
(After this the wait time will be down to around **~20 mins** from **~42 mins**)

- [x] Fix failing tests
- [x] Enable coverage for parallel test
- [x] Fix deprecations warnings for assertEquals

---
- Cleaner test reports
		**Before:**
		Screenshot 2021-05-08 at 1 46 38 PM
		**Now:**
		Screenshot 2021-05-08 at 1 47 35 PM

Dependent on https://github.com/frappe/frappe/pull/13165
**Note:** This is dependant on [test orchestrator](https://github.com/frappe/test-orchestrator) to balance test time across instances ([to avoid this](https://github.com/frappe/erpnext/pull/25635#issuecomment-835662892)).


surajshetty3416:
Build time across instances is not balanced and is highly uneven.

Screenshot 2021-05-08 at 8 47 38 PM

Alchez:
@surajshetty3416, can we not salvage anything from this? The cleaner reports are a good change.

surajshetty3416:
@Alchez I'll reopen this with a better solution... Stay Tuned!

erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -152,7 +152,7 @@ def validate_dates(self):
 			frappe.throw(_("Valid from date must be less than valid upto date"))
 
 	def validate_condition(self):
-		if self.condition and ("=" in self.condition) and re.match("""[\w\.:_]+\s*={1}\s*[\w\.@'"]+""", self.condition):
+		if self.condition and ("=" in self.condition) and re.match(r'[\w\.:_]+\s*={1}\s*[\w\.@\'"]+', self.condition):
```
Alchez, line 155: I think you should keep the multi-string, since the regexp has both the quotes

Alchez (changes requested):

erpnext/accounts/doctype/pricing_rule/pricing_rule.py
surajshetty3416: @Alchez single quote inside regex has been escaped so it should not be an issue.

Alchez: Ah, no worries then

PR #25645, opened 2021-05-09
perf(minor): remove unnecessary comprehensions

ankush:
- remove several unnecessary comprehensions from functions that accept a generator.
- Using `[x for x in iter]` causes a list to be built first then passed to the outer function.
- `any` and `all` can take generator instead. This makes memory usage O(1) and actually makes these functions short-circuiting. E.g. if the first condition fails then `all` will immediately return false instead of evaluating all the entries.
- `sum`, `min`, `max` => memory usage become O(1)
- `list`, `set`, `.join()` => roughly halves memory usage, as list is not required to be built.
- lastly, it's two fewer characters to read/think about 😁

### DIFF
--- a/erpnext/accounts/doctype/c_form/c_form.py
+++ b/erpnext/accounts/doctype/c_form/c_form.py
@@ -54,7 +54,7 @@ class CForm(Document):
 			frappe.throw(_("Please enter atleast 1 invoice in the table"))
 
 	def set_total_invoiced_amount(self):
-		total = sum([flt(d.grand_total) for d in self.get('invoices')])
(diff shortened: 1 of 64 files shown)


Alchez:
@ankush, should we add this to a style guide, if one doesn't exist already? Seems like it could help for future contributions.

ankush:
@Alchez I'll just add them in automated checks.

ankush:
- 1 failure is unrelated to changes and appearing in other PRs too. 
- All changes are minor and only remove comprehension where it's allowed. Hence, merging without a review.

PR #25649, opened 2021-05-10
feat: replaced product id with sku code when creating item from shopify'

rakshithrddy:
Replaced product id with SKU code when creating an item from Shopify.

### DIFF
--- a/erpnext/erpnext_integrations/doctype/shopify_settings/sync_product.py
+++ b/erpnext/erpnext_integrations/doctype/shopify_settings/sync_product.py
@@ -95,7 +95,7 @@ def create_item(shopify_item, warehouse, has_variant=0, attributes=None,variant_
 		"variant_of": variant_of,
 		"sync_with_shopify": 1,
 		"is_stock_item": 1,
-		"item_code": cstr(shopify_item.get("item_code")) or cstr(shopify_item.get("id")),
+		"item_code": shopify_item.get("sku") or get_sku(shopify_item) or cstr(shopify_item.get("item_code")),
 		"item_name": shopify_item.get("title", '').strip(),
 		"description": shopify_item.get("body_html") or shopify_item.get("title"),
 		"shopify_description": shopify_item.get("body_html") or shopify_item.get("title"),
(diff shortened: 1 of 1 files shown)


erpnext/erpnext_integrations/doctype/shopify_settings/sync_product.py
```diff
@@ -95,7 +95,7 @@ def create_item(shopify_item, warehouse, has_variant=0, attributes=None,variant_
 		"variant_of": variant_of,
 		"sync_with_shopify": 1,
 		"is_stock_item": 1,
-		"item_code": cstr(shopify_item.get("item_code")) or cstr(shopify_item.get("id")),
+		"item_code": shopify_item.get("sku") or get_sku(shopify_item) or cstr(shopify_item.get("item_code")),
```
ankush, line 98: This would break consistency for existing users.

If you instead want incoming Shopify item to be mapped to an existing item if Shopify SKU == ERPNext ITEM code. Then better way would be to match them before creating the item. 

You could instead try and modify this function to match with an existing item and update it instead of creating new item.

https://github.com/frappe/erpnext/blob/b0e77cbd89394a609108768466fef97cb00902bf/erpnext/erpnext_integrations/doctype/shopify_settings/sync_product.py#L258

ankush (changes requested):

ankush:
@rakshithrddy 

What if we add this check inside `is_item_exists` function? That would add additional heuristics but won't affect any existing users. 

```python
def is_item_exists(shopify_item, attributes=None, variant_of=None):
	if variant_of:
		name = variant_of
	else:
		name = frappe.db.get_value("Item", {"item_name": shopify_item.get("item_name")})
		# try and match based on SKU
		if not name:
			name = frappe.db.get_value("Item", {"item_code": shopify_item.get("stock_keeping_unit")})
```

I know this complicates an already complicated function, but we can think of refactoring if the solution itself is correct.

ankush:
We have refactored Shopify integration and moved it to a separate app. It now considers matching item_code <=> Shopify SKU.  

https://github.com/frappe/ecommerce_integrations

Closing due to inactivity. Reopen if required.

PR #25654, opened 2021-05-10
fix: Wrong amount on CR side in general ledger report for customer when different account currencies are involved

AfshanKhan:
### Problem:
General Ledger shows incorrect `Debit/Credit` values for some currency exchanges. for eg. INR to USD, INR is base currency & USD is party currency

image

image

### Steps to Reproduce:
Step 1: Create and submit an invoice in USD
Step 2: Try to see the general ledger from the invoice itself
Step 3: change the currency in the General Ledger report to USD we can see the difference in debit and credit amount, but cannot in company currency the values are equal.

### Solution:
Refactor `convert_to_presentation_currency` function

1. If all the accounts in GL Entries have the same account currency & that is equal to presentation currency. Then use the "credit_in_account_currency" field to display the appropriate values
2. If all the accounts in GL Entries have the same account currency & that is not equal to presentation currency. Then convert all the debit/credit values from company currency into the presentation currency @ report date
3. If accounts in GL Entries have different account currency. Then convert all the values from company currency into presentation currency @ report date

image

### DIFF
--- a/erpnext/accounts/report/utils.py
+++ b/erpnext/accounts/report/utils.py
@@ -81,8 +81,7 @@ def convert_to_presentation_currency(gl_entries, currency_info, company):
 	presentation_currency = currency_info['presentation_currency']
 	company_currency = currency_info['company_currency']
 
-	pl_accounts = [d.name for d in frappe.get_list('Account',
-		filters={'report_type': 'Profit and Loss', 'company': company})]
+	account_currencies = list(set(entry['account_currency'] for entry in gl_entries))
 
 	for entry in gl_entries:
 		account = entry['account']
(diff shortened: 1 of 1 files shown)


nextchamp-saqib:
@nabinhait Can you review if the following conditions look correct? Specifically the exchange rate fetching @ report date

> - If all the accounts in GL Entries have the same account currency & that is equal to presentation currency. Then use the "credit_in_account_currency" field to display the appropriate values
> - If all the accounts in GL Entries have the same account currency & that is not equal to presentation currency. Then convert all the debit/credit values from company currency into the presentation currency @ report date
> - If accounts in GL Entries have different account currency. Then convert all the values from company currency into presentation currency @ report date

PR #25655, opened 2021-05-10
fix: Outstanding Amount calculation error in Payment Entry from Sales Order and Sales Invoice

hrwX:
- **Sales Order**
    - While creating a  Payment Entry against Sales Order, the outstanding amount is not set to zero if a full payment is made against it.
    - The `d.set` in `set_missing_ref_details` in Payment Entry only updates the values in document not in the database, which let to the outstanding amount not updating in Payment Entry for Sales Order.
- **Sales Invoice**
   - After creating a Payment Entry against Sales Order, when you create a Sales Invoice and check `update_advances_automatically`, the outstanding amount is not set to zero in Payment Entry.
   - This is due to the fact that in `on_submit` the changes are never commited to db and in Payment Entry, in function `get_reference_details`, the frappe.get_doc gets stale value from the database and the outstanding amount in never set to zero.

**Steps to Reproduce**:
- Create a Sales Order and then create Payment Entry against it.
- After submission, create a Sales Invoice and check `allocate_advances_automatically` and submit it.

**After Fix**:
- https://user-images.githubusercontent.com/7310479/117761902-cb67d800-b245-11eb-92ae-483a6d5bc1e3.mp4
- Peek 2021-05-11 10-39


kennethsequeira:
The fix will need to be backported as it's affecting all versions

hrwX:
@deepeshgarg007 @rohitwaghchaure  Can you guys review this ?? Seems like a legitimate issue and has been reported in the community.

- https://discuss.erpnext.com/t/payment-entry-updating-the-outstanding-possible-bug/34887
- https://discuss.erpnext.com/t/payment-entry-allocation-incorrect/74987

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -225,6 +225,7 @@ def on_submit(self):
 		self.update_serial_no()
 
 		if not cint(self.is_pos) == 1 and not self.is_return:
+			self.db_update()
```
deepeshgarg007: Why a `db_update` is needed here?
I am assuming this should be taken care of by this function
https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/utils.py#L341

Even if it doesn't it should be placed somewhere in this function instead of directly in the sales invoice `on_submit` trigger

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
hrwX: Makes sense, will push the change

hrwX:
@deepeshgarg007 make the necessary changes

deepeshgarg007:
@hrwX Issue fixed via https://github.com/frappe/erpnext/pull/25799

image

Closing this

PR #25675, opened 2021-05-12
fix: use batch, serial in future sle check for neg qty

18alantom:
### Description
1. Uses batch and serial numbers filters to check future Stock Ledger Entries for negative values.
2. Checks future SLE for negative values for matching warehouse and batch.


erpnext/stock/stock_ledger.py
```diff
@@ -867,3 +870,8 @@ def _round_off_if_near_zero(number: float, precision: int = 6) -> float:
 		return 0
 
 	return flt(number)
+
+def _switch_values(target, keys, value_from=None, value_to=""):
```
nabinhait: @18alantom Very bad choice of function naming, variable naming and implementation, the name does not tell what it does.

It should be like: 
```suggestion
def sanitize_args(args, keys):
	for k in keys:
		if args[k] == None:
			args[k] == ''
	return args
```

nabinhait (changes requested):

erpnext/stock/stock_ledger.py
18alantom: @nabinhait I agree; should `sanitize_args` be in stock_ledger.py, or moved elsewhere?

nabinhait: keep it here only

marination:
@18alantom add tests for this ?

erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -168,6 +168,35 @@ def validate_with_last_transaction_posting_time(self):
...
+
+		if len(future_negative_entries) > 0:
+			voucher_type, voucher_no, item_code, warehouse, posting_date = future_negative_entries[0]
+			frappe.throw(_('Quantity not available for') + ' ' +
+				frappe.bold(item_code) + ' ' + _('in warehouse') + ' ' +
+				frappe.bold(warehouse) + ' ' + _('for future') + ' ' +
+				frappe.bold(voucher_type) + ' ' + frappe.bold(voucher_no) +
+				' ' + _('at the date of posting') + ' ' +
+				'(' + posting_date.isoformat() + ')')
+
```
rohitwaghchaure, line 201: Translation check failing can you check Semgrep

erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -168,6 +168,35 @@ def validate_with_last_transaction_posting_time(self):
...
+	def validate_future_sle_by_batch_for_outgoing_entries(self):
+		allow_neg = frappe.db.get_single_value('Stock Settings', 'allow_negative_stock')
+		if allow_neg or self.actual_qty is None or self.actual_qty >= 0 or self.batch_no is None:
+			return
+
+		future_negative_entries = frappe.db.sql("""select
+					voucher_type, voucher_no, item_code, warehouse, posting_date
+				from `tabStock Ledger Entry`
+				where
+					warehouse = %{warehouse}s
```
rohitwaghchaure: Travis failing

18alantom:
This PR was spuriously closed in the great PR purge of August 2021.

Many a test has been patched to get CI to pass but to no avail, new patches fail!
Fault lies not with the PR, but with the tests that do not account for future negative quantities.

erpnext/accounts/doctype/pos_closing_entry/test_pos_closing_entry.py
```diff
@@ -26,6 +26,8 @@ def tearDown(self):
 		frappe.db.sql("delete from `tabPOS Profile`")
 
 	def test_pos_closing_entry(self):
+		is_allow_neg = frappe.db.get_single_value('Stock Settings', 'allow_negative_stock')
+		frappe.db.set_value('Stock Settings', 'Stock Settings', 'allow_negative_stock', 1)
```
ankush, line 30: Way too much duplicate code, should be a decorator instead, or if it applies class-wide then setupclass/teardownclass.

erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -837,6 +837,50 @@ def test_conversion_factor_change(self):
...
+			dict(item_code=item_code,
+				qty=2,
+				from_warehouse=warehouse_names[0],
+				to_warehouse=warehouse_names[1],
+				batch_no=batch_no,
+				posting_date='2021-07-02',          # Illegal SE
+				purpose='Material Transfer')
+		]
+		illegal_sequence_executor = create_stock_entry_sequence_executor(illegal_sequence)
+		self.assertRaises(frappe.ValidationError, illegal_sequence_executor)
```
ankush, line 881:
```suggestion
		self.assertRaises(frappe.ValidationError, create_stock_entries, illegal_sequence)
```

args, kwargs can be passed to it directly, no need to create a partially applied funciton here.

erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -907,3 +951,33 @@ def get_multiple_items():
...
+		warehouse=warehouse_names[0],
+		valuation_rate=100,
+		qty=opening_qty,
+		batch_no=batch_no,
+	)
+	return warehouse_names
+
+def create_stock_entry_sequence_executor(sequence_of_entries):
+	return lambda: create_stock_entries(sequence_of_entries)
+
```
ankush, line 980:
```suggestion
```

Not required.

ankush (changes requested):

erpnext/accounts/doctype/pos_closing_entry/test_pos_closing_entry.py
ankush: @18alantom can you remove as many of these as possible? Let's try to get this merged without toggling of such flags 😅 

If required change the test case to ERPNextTestCase to stop committing in other tests.

18alantom: Yup for sure, which is what I wanted to do almost as soon as I flipped the flag.

18alantom:
Closed cause #27945

PR #25676, opened 2021-05-12
fix: bank statement import via google sheet

### LINKED ISSUE #25523: Bank Reconciliation Tool Error
Hi,
We’re having some inconvenients with bank reconciliation tool.
First, it doesn’t work if you’re importing with goggle sheet. It recognizes the field, but doesn’t import the file.

image

image

And this is all.

Error log:

File “/home/diamo/Frappe/apps/frappe/frappe/utils/background_jobs.py”, line 100, in execute_job
method(**kwargs)
File “/home/diamo/Frappe/apps/erpnext/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py”, line 100, in start_import
import_file = ImportFile(“Bank Transaction”, file = import_file_path, import_type=“Insert New Records”)
File “/home/diamo/Frappe/apps/frappe/frappe/core/doctype/data_import/importer.py”, line 331, in init
frappe.throw(_(“Invalid template file for import”))
File “/home/diamo/Frappe/apps/frappe/frappe/init.py”, line 433, in throw
msgprint(msg, raise_exception=exc, title=title, indicator=‘red’, is_minimizable=is_minimizable, wide=wide, as_list=as_list)
File “/home/diamo/Frappe/apps/frappe/frappe/init.py”, line 412, in msgprint
_raise_exception()
File “/home/diamo/Frappe/apps/frappe/frappe/init.py”, line 366, in _raise_exception
raise raise_exception(msg)
frappe.exceptions.ValidationError: Invalid template file for import

It’s strange because model template have been downloaded from the ERP. The only way that it works is importing files from the device. Both images are exactly the same, but the difference is that from my device it works and from google sheet doesn’t.
After that,

image

The statement from my device has been successfully imported, but I couldn’t find the way to do the reconciliation. The only option I have is “go to transactions list” button, that shows this

image

So, I’ve imported the statement, but I can’t find the way to reconcile automatically those. I can do it individually, but it it’s practical if you have a lot of transactions.

Is anybody using this? Could you explain me how they do, and if I’m doing something wrong or if it’s a bug?

Installed Apps
ERPNext: v13.1.0 
Frappe Framework: v13.1.0 

Thanks in advance!

hasnain2808:
fixes https://github.com/frappe/erpnext/issues/25523

### DIFF
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
@@ -239,6 +239,7 @@ frappe.ui.form.on("Bank Statement Import", {
 					"withdrawal",
 					"description",
 					"reference_number",
+					'bank_account'
 				],
 			},
 		});
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
@@ -239,7 +239,7 @@ frappe.ui.form.on("Bank Statement Import", {
 					"withdrawal",
 					"description",
 					"reference_number",
-					'bank_account'
+					"bank_account"
 				],
 			},
 		});
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.json
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.json
@@ -146,7 +146,7 @@
   },
   {
    "depends_on": "eval:!doc.__islocal && !doc.import_file\n",
-   "description": "Must be a publicly accessible Google Sheets URL",
+   "description": "Must be a publicly accessible Google Sheets URL and Adding Bank Account column is necessary for importing via Google Sheets",
    "fieldname": "google_sheets_url",
    "fieldtype": "Data",
    "label": "Import from Google Sheets"
@@ -202,7 +202,7 @@
  ],
  "hide_toolbar": 1,
  "links": [],
- "modified": "2021-02-10 19:29:59.027325",
+ "modified": "2021-05-12 14:17:37.777246",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Bank Statement Import",
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.json
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.json
@@ -146,7 +146,7 @@
   },
   {
    "depends_on": "eval:!doc.__islocal && !doc.import_file\n",
-   "description": "Must be a publicly accessible Google Sheets URL and Adding Bank Account column is necessary for importing via Google Sheets",
+   "description": "Must be a publicly accessible Google Sheets URL and adding Bank Account column is necessary for importing via Google Sheets",
    "fieldname": "google_sheets_url",
    "fieldtype": "Data",
    "label": "Import from Google Sheets"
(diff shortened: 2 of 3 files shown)


erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py
```diff
@@ -67,9 +74,10 @@ def start_import(self):
 				data_import=self.name,
 				bank_account=self.bank_account,
 				import_file_path=self.import_file,
+				google_sheets_url=self.google_sheets_url,
 				bank=self.bank,
 				template_options=self.template_options,
-				now=frappe.conf.developer_mode or frappe.flags.in_test,
+				now=1#frappe.conf.developer_mode or frappe.flags.in_test,
```
ankush: intentional? If yes, remove the trailing comment.

erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
```diff
@@ -239,6 +239,7 @@ frappe.ui.form.on("Bank Statement Import", {
 					"withdrawal",
 					"description",
 					"reference_number",
+					'bank_account'
```
ankush:
```suggestion
					"bank_account"
```
consistency :)

erpnext/accounts/doctype/bank_statement_import/bank_statement_import.json
```diff
@@ -146,7 +146,7 @@
   },
   {
    "depends_on": "eval:!doc.__islocal && !doc.import_file\n",
-   "description": "Must be a publicly accessible Google Sheets URL",
+   "description": "Must be a publicly accessible Google Sheets URL and Adding Bank Account column is necessary for importing via Google Sheets",
```
ankush:
```suggestion
   "description": "Must be a publicly accessible Google Sheets URL and adding Bank Account column is necessary for importing via Google Sheets",
```

erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py
```diff
@@ -47,6 +47,13 @@ def validate(self):
 
 	def start_import(self):
 
+		preview = frappe.get_doc("Bank Statement Import", self.name).get_preview_from_template(
+			self.import_file, self.google_sheets_url
+		)
+
+		if 'Bank Account' not in json.dumps(preview):
+			frappe.throw('Please Add the Bank Account Column')
```
ankush:
```suggestion
			frappe.throw(_("Please add the Bank Account column"))
```

nextchamp-saqib (approved):

barredterra:
@hasnain2808 nice work on the bank import, thanks!

We get our bank data as XML files according to the [CAMT standard](https://www.sepaforcorporates.com/swift-for-corporates/a-practical-guide-to-the-bank-statement-camt-053-format/) ([example parser code](https://programtalk.com/vs2/?source=python%2F9466%2Fbank-statement-import%2Faccount_bank_statement_import_camt%2Fcamt.py#)). Do you think we could refactor bank import to use different parsers for different file formats? Think `class CsvParser`, `class XMLParser`, etc., and specific parsers inheriting from them: `class CamtParser(XmlParser)`. During the import, the user can choose which parser to use. Or we can try to auto-detect it with standard methods like `is_csv()` / `is_xml()`.

There might also be multiple CSV / XML files zipped together (say, one for each day). It would be much more comfortable to upload the zip file instead of the individual files.

hasnain2808:
Hi @barredterra ,

This should be possible

Achieving this would involve the following steps

1. Adding the relevant format utils/ parser in the framework
2. Calling the parser in data import(framework) so that we can import using that format
3. Adding a utility in bank statement import to manage the bank field 

Create an issue with feature request tag so that we may track it

PR #25677, opened 2021-05-12
fix: bank statement import via google sheet

### LINKED ISSUE #25523: Bank Reconciliation Tool Error

hasnain2808:
fixes https://github.com/frappe/erpnext/issues/25523

### DIFF
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
@@ -239,6 +239,7 @@ frappe.ui.form.on("Bank Statement Import", {
 					"withdrawal",
 					"description",
 					"reference_number",
+					'bank_account'
 				],
 			},
 		});
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
@@ -239,7 +239,7 @@ frappe.ui.form.on("Bank Statement Import", {
 					"withdrawal",
 					"description",
 					"reference_number",
-					'bank_account'
+					"bank_account"
 				],
 			},
 		});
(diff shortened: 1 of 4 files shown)


nextchamp-saqib (approved):
Works!

PR #25683, opened 2021-05-12
feat: linking supplier with an item group for filtering items

noahjacob:
**Changes:**

- Created a doctype to create links between supplier and item_groups.

image

- Updated filters for buying module if any item groups are linked to the supplier.
item_group
(If no item groups are linked to the supplier it does not apply the filter)

-  Added Supplier Item Group in Supplier dashboard.
image

- **[Docs](https://github.com/frappe/erpnext_documentation/pull/333)**

### DIFF
--- a/erpnext/buying/doctype/supplier_item_group/supplier_item_group.json
+++ b/erpnext/buying/doctype/supplier_item_group/supplier_item_group.json
@@ -1,0 +1,51 @@
+{
+ "actions": [],
+ "creation": "2021-05-07 18:16:40.621421",
+ "doctype": "DocType",
+ "editable_grid": 1,
+ "engine": "InnoDB",
+ "field_order": [
+  "supplier",
+  "item_group"
+ ],
+ "fields": [
+  {
+   "fieldname": "supplier",
+   "fieldtype": "Link",
+   "in_list_view": 1,
+   "label": "Supplier",
+   "options": "Supplier"
+  },
+  {
+   "fieldname": "item_group",
+   "fieldtype": "Link",
+   "in_list_view": 1,
+   "label": "Item Group",
+   "options": "Item Group"
+  }
+ ],
+ "index_web_pages_for_search": 1,
+ "links": [],
+ "modified": "2021-05-07 18:16:40.621421",
+ "modified_by": "Administrator",
+ "module": "Buying",
+ "name": "Supplier Item Group",
+ "owner": "Administrator",
+ "permissions": [
+  {
+   "create": 1,
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "System Manager",
+   "share": 1,
+   "write": 1
+  }
+ ],
+ "sort_field": "modified",
+ "sort_order": "DESC",
+ "track_changes": 1
+}
--- a/erpnext/buying/doctype/supplier_item_group/supplier_item_group.json
+++ b/erpnext/buying/doctype/supplier_item_group/supplier_item_group.json
@@ -14,19 +14,21 @@
    "fieldtype": "Link",
    "in_list_view": 1,
    "label": "Supplier",
-   "options": "Supplier"
+   "options": "Supplier",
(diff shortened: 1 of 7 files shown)


erpnext/controllers/queries.py
```diff
@@ -216,11 +216,20 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
...
 
+	if filters.get('supplier'):
+		item_group_list = frappe.get_list('Supplier Item Group',filters = {'supplier': filters.get('supplier')}, fields = ['item_group'])
+		item_groups = []
+		for i in item_group_list:
+			item_groups.append(i.item_group)
+		del filters['supplier']
+		if item_groups:
+			filters['item_group'] = ['in',item_groups]
+			print(item_groups)
```
rohitwaghchaure: remove print statement

rohitwaghchaure: remove print statement

erpnext/controllers/queries.py
```diff
@@ -216,11 +216,20 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
 		if not field in searchfields]
 	searchfields = " or ".join([field + " like %(txt)s" for field in searchfields])
 
+	if filters.get('supplier'):
+		item_group_list = frappe.get_list('Supplier Item Group',filters = {'supplier': filters.get('supplier')}, fields = ['item_group'])
+		item_groups = []
+		for i in item_group_list:
+			item_groups.append(i.item_group)
+		del filters['supplier']
```
rohitwaghchaure, line 226: Add space in the code which improves the readability

rohitwaghchaure:
@noahjacob add supplier item group doctype in the dashboard of supplier

image

rohitwaghchaure (approved):

rohitwaghchaure:
frappe.get_list checks whether the user has permission of Supplier Item Group or not, if the person who is creating the purchase order has not permission of the doctype Supplier Item Group then system will not allowed user to make the purchase order and throw below error. 

Screenshot 2021-05-19 at 1 04 38 PM


You can use frappe.get_all instead of frappe.get_list because in this case we don't need to check permission for the doctype 'Supplier Item Group'

erpnext/buying/doctype/supplier_item_group/supplier_item_group.json
```diff
@@ -0,0 +1,51 @@
...
+ "permissions": [
+  {
+   "create": 1,
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
+   "read": 1,
+   "report": 1,
+   "role": "System Manager",
```
rohitwaghchaure, line 45: Why just system manager? give permission to purchase user, purchase manager

erpnext/controllers/queries.py
```diff
@@ -216,11 +216,22 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
 		if not field in searchfields]
 	searchfields = " or ".join([field + " like %(txt)s" for field in searchfields])
 
+	if filters.get('supplier'):
```
barredterra, line 219: @noahjacob, @rohitwaghchaure filters argument is a string containing json, when called from the frontend. The following code breaks with this PR:

```js
frappe.call({
    method: 'erpnext.controllers.queries.item_query',
    args: {
        doctype: 'Item',
        txt: '',
        searchfield: 'name',
        start: 0,
        page_len: 99,
        filters: {
            supplier: 'Unico Plastics Inc.',
            is_sales_item: 1
        },
        as_dict: true
    },
    callback: (r) => console.log(r.message)
});
```

Bildschirmfoto 2021-06-02 um 17 36 03


filters string needs to be parsed to json first.

Please fix or revert this PR.

nabinhait: already fixed via #25915 , will release in sometime

barredterra: @nabinhait while it doesn't produce an error anymore, filters will still be ignored in my example.

PR #25694, opened 2021-05-13
fix: timeout error while loading warehouse tree

rohitwaghchaure:
backport https://github.com/frappe/erpnext/pull/25693
**Issue**
Customer has 2500 warehouses and they wanted to see all the warehouses in the tree view. But system was not able to load all the warehouses within 3mins and thrown an error that "Request Time Out"

Screenshot 2021-05-13 at 1 39 47 AM

**Code Debug**
Using bench console did profiling and found that system taking more than 30 mins to load all the warehouses and the method  get_stock_value_from_bin which get the stock value from bin against the warehouse executes every time against each warehouse (even though warehouse is parent warehouse). So if customer has 2500 warehouses then this method was executing 2500 times.

Even check mysql processlist and found that the bottleneck is at the get_stock_value_from_bin query

Screenshot 2021-05-13 at 1 08 37 AM

**Solution**

Optimized the code and no of calls to database by adding new method. With few number of db call get all the child warehouses stock value data and using recursion method updated stock value in the parent warehouses.

**After Fix**

After the fix system has took 30 seconds to load all the warehouses

Screenshot 2021-05-13 at 1 33 22 AM

### DIFF
--- a/erpnext/stock/doctype/warehouse/warehouse.py
+++ b/erpnext/stock/doctype/warehouse/warehouse.py
@@ -3,8 +3,9 @@
 
 from __future__ import unicode_literals
 import frappe, erpnext
-from frappe.utils import cint, nowdate
+from frappe.utils import cint, flt
 from frappe import throw, _
+from collections import defaultdict
 from frappe.utils.nestedset import NestedSet
 from erpnext.stock import get_warehouse_account
 from frappe.contacts.address_and_contact import load_address_and_contact
@@ -139,8 +140,6 @@ class Warehouse(NestedSet):
 
 @frappe.whitelist()
 def get_children(doctype, parent=None, company=None, is_root=False):
-	from erpnext.stock.utils import get_stock_value_from_bin
-
 	if is_root:
 		parent = ""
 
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -153,13 +151,52 @@ def get_children(doctype, parent=None, company=None, is_root=False):
...
+def get_warehouse_wise_stock_value(company):
+	warehouses = frappe.get_all('Warehouse',
+		fields = ['name', 'parent_warehouse'], filters = {'company': company})
+
+	parent_warehouse = {d.name : d.parent_warehouse for d in warehouses}
+	filters = {'warehouse': ('in', [data.name for data in warehouses])}
+	bin_data = frappe.get_all('Bin', fields = ['sum(stock_value) as stock_value', 'warehouse'],
+		filters = filters, group_by = 'warehouse')
+
+	warehouse_wise_stock_value = {}
```
ankush:
```suggestion
	warehouse_wise_stock_value = defaultdict(float)
```

erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -153,13 +151,52 @@ def get_children(doctype, parent=None, company=None, is_root=False):
...
+
+def update_value_in_parent_warehouse(warehouse_wise_stock_value, parent_warehouse_dict, warehouse, stock_value):
+	parent_warehouse = parent_warehouse_dict.get(warehouse)
+	if not parent_warehouse:
+		return
+
+	if parent_warehouse in warehouse_wise_stock_value:
+		warehouse_wise_stock_value[parent_warehouse] += flt(stock_value)
+	else:
+		warehouse_wise_stock_value[parent_warehouse] = flt(stock_value)
```
ankush:
```suggestion
	warehouse_wise_stock_value[parent_warehouse] += flt(stock_value)
```

ankush (commented):
using `defaultdict` would simplify some code. (need to add `from collections import defaultdict` though)

erpnext/stock/doctype/warehouse/warehouse.py
rohitwaghchaure: Done

erpnext/stock/doctype/warehouse/warehouse.py
rohitwaghchaure: Done

PR #25699, opened 2021-05-13
fix: Process Statement of Accounts formatting

Anuja-pawar:
Process Statement of Accounts enhancement:

- added field to print with letter head (header, footer)
- added field to print with Terms and Conditions
- renamed third column from Party to Remarks
- formatted headline 'Statement of Accounts' according to the proposed design

ezgif com-gif-maker


deepeshgarg007 (approved):

deepeshgarg007:
@Anuja-pawar also backport this to `version-13-hotfix` branch

PR #25712, opened 2021-05-14
fix: total time calc for job card

18alantom:
### Description
Adds to total time only if completed qty is more than 0.

### Before
Screen Shot 2021-05-14 at 17 03 55

### After
Screen Shot 2021-05-14 at 17 02 00


rohitwaghchaure:
@18alantom this is not correct fix, there might be chances that to make 1 qty requires more than 1 day

PR #25714, opened 2021-05-14
feat: Show net values in Party Accounts

deepeshgarg007:
Consider the following scenario where both payment and credit note is allocated against invoices.

image

The statement of account for that customer in this scenario is shown as follows:
image

Many times this credit/debit note allocation is internal to the user of the system and want to show only net values to external entities like Auditors, Customers and Suppliers.

For this added a check in General Ledger to show net values for party accounts when viewing consolidated view grouped by voucher

Screenshot 2021-05-14 at 7 35 29 PM

SOA post enabling net values
image

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.js
+++ b/erpnext/accounts/report/general_ledger/general_ledger.js
@@ -166,6 +166,11 @@ frappe.query_reports["General Ledger"] = {
 			"fieldname": "show_cancelled_entries",
 			"label": __("Show Cancelled Entries"),
 			"fieldtype": "Check"
+		},
+		{
+			"fieldname": "show_net_values_in_party_account",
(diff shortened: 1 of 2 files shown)


nabinhait (approved):

erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -166,6 +166,11 @@ frappe.query_reports["General Ledger"] = {
 			"fieldname": "show_cancelled_entries",
 			"label": __("Show Cancelled Entries"),
 			"fieldtype": "Check"
+		},
+		{
+			"fieldname": "show_net_values_in_party_account",
+			"label": __("Show Net Values in Party Account"),
```
nabinhait, line 172: Show Net Values for Party Account?

PR #25722, opened 2021-05-15
feat: Increase number of supported currency exchanges

### LINKED ISSUE #25603: Many unsupported exchange rates: Switch from frankfurter.app to exchangerate.host or allow custom API
**Problem:**
Exchange rates are currently fetched from https://www.frankfurter.app/. Unfortunately, this API only supports about 30-40 currencies with many world currencies still being not supported. This makes using a multi-currency setup very cumbersome to impossible to maintain if exchange rates have to be added manually.

**Proposed solution:**
There are other free APIs available that support more exchange rates. One promising API seems to be this one: https://exchangerate.host/ 
It is also free and open source (https://github.com/Formicka/exchangerate.host) and apparently supports 170 currencies (+ 6000 cryptocurrencies - not sure but might be also interesting for some ERPnext users?!).

It doesn't seem to be too complicated to adjust the current code to support. Unfortunately, I am hosting my instance on frappe.cloud, so I don't have the option to work on the code.
In [erpnext/setup/utils.py (L101)](https://github.com/frappe/erpnext/blob/5ce0569f80c622d6d7ff480015e8d79b82bf93b5/erpnext/setup/utils.py#L101) the frankfurter.app API is currently called. It seems one would only need to adjust the API call and link to exchangerate.host and this would directly add >100 additional currencies supported by ERPnext (+many cryptocurrencies).
It would be amazing if someone from the core team could test this. It doesn't seem to be a hard fix.

**Potential alternatives**
Potentially, ERPnext could allow the user to specify the API (and its settings) similar to how it is currently available for SMS APIs (https://docs.erpnext.com/docs/user/manual/en/setting-up/sms-setting). This might be a little bit more work, but would make multi-currency support even more flexible. If the core team wishes to have frankfurter.app still as default, this could be left as a default value. However this way, the user could change to the API they wish to use and could also use a commercial API (such as fixer.io) if they need/want. 

**Additional context**
This issue has been raised already quite some time ago in this thread in the Forum:
https://discuss.erpnext.com/t/is-there-any-plan-for-the-missing-exchange-rates-on-www-frankfurter-app/48782
I believe many companies especially in developing countries with lesser known currencies, that are currently unsupported by the frankfurter.app, would benefit from this.

rtdany10:
Switch from frankfurter.app to exchangerate.host to accomodate more currency usage.
Closes #25603


moisesbenzan:
Hey! Check out this version of the change I made. IMO seems cleaner and gives an accurate result. I've tested this on my ERPNext installation.
```
diff --git a/erpnext/setup/utils.py b/erpnext/setup/utils.py
index 13269a828237c36086555a69adcee607fed4bcdf..c551dd266c480255a75499c8b1bb58612c66565c 100644
--- a/erpnext/setup/utils.py
+++ b/erpnext/setup/utils.py
@@ -98,14 +98,15 @@ def get_exchange_rate(from_currency, to_currency, transaction_date=None, args=No
 
 		if not value:
 			import requests
-			api_url = "https://frankfurter.app/{0}".format(transaction_date)
+			api_url = "https://api.exchangerate.host/convert"
 			response = requests.get(api_url, params={
-				"base": from_currency,
-				"symbols": to_currency
+				"date": transaction_date,
+				"to": to_currency,
+				"from": from_currency,
 			})
 			# expire in 6 hours
 			response.raise_for_status()
-			value = response.json()["rates"][to_currency]
+			value = response.json()["result"]
 
 			cache.set_value(key, value, expires_in_sec=6 * 60 * 60)
 		return flt(value)

```

rtdany10:
Thank you @moisesbenzan 
Your code accounts for the transaction date as well, which I forgot to take into account. Will update the code.

moisesbenzan:
@rtdany10 You're welcome! This issue affects us all so it's on our own best interest to get it solved.

For the maintainers: I could make an MR for version-13 if needed as this will greatly help a lot of the users.

rtdany10:
Tests failing due to different hard coded values. Will rewrite the test and change the values in a while.

rtdany10:
@gavindsouza The current test suit has the value fetched from fixer.io. There is a mismatch issue in historical data. Is it okay to replace those values to make the test pass?

aschl:
@rtdany10 Yes, I guess we need to adjust the tests in order to pass the PR. I assume the different internal calculation of daily averages lead to slightly different rates between fixer.io and exchangerate.host. It might make sense to relax the test and allow a range of possible values (e.g.+- 0.5%) to pass the test?!
Btw: What is the reason not directly pushing to version-13-hotfix?

rtdany10:
> @rtdany10 Yes, I guess we need to adjust the tests in order to pass the PR. I assume the different internal calculation of daily averages lead to slightly different rates between fixer.io and exchangerate.host. It might make sense to relax the test and allow a range of possible values (e.g.+- 0.5%) to pass the test?!

I have updated the values so that the test passes.

> Btw: What is the reason not directly pushing to version-13-hotfix?

Unsure as to why I chose that branch. Should've chose develop and let the frappe team decide on the branch. My bad.

aschl:
Any progress on this?
@rtdany10 @moisesbenzan: If I understand the check results correctly, the issue is with python 2. According to https://discuss.erpnext.com/t/dropping-python-2-support-for-frappe-framework/61159 , V13 does not support python 2 anymore. In order to get the PR passed (and push it to the recent V13 version), can we drop the PR for version-12 and instead create a new one for version-13? 
(Let me know if I misunderstood the checks. I hope it can get merged by the Frappe team soon.)

rtdany10:
> Any progress on this?

The work from our side is over. Frappe team has to decide whether or not to merge this PR.

@nabinhait @gavindsouza

aschl:
Ok. Thanks for clarifying :)

aschl:
Any progress on this?
@rtdany10 : Is it possible that this PR is not getting reviewed because it was submitted to the V12 branch? Can we maybe move it to V13?

rtdany10:
> Any progress on this?
> @rtdany10 : Is it possible that this PR is not getting reviewed because it was submitted to the V12 branch? Can we maybe move it to V13?

Not sure, I'll create a pull request for v13 too then.

rtdany10:
> Any progress on this?
> @rtdany10 : Is it possible that this PR is not getting reviewed because it was submitted to the V12 branch? Can we maybe move it to V13?

Pull request on develop branch:  #26237

aschl:
Any updates? @Frappe Team: Any feedback in terms of review?

rtdany10:

Issue #25638, opened 2021-05-08
POS closing Entry Expected Amount


michaeleino:
## Description of the issue
POS closing entry is showing incorrect value numbers
 
## Context information (for bug reports)

**Output of `bench version`**
```
erpnext 13.2.1
frappe 13.2.1
```
**Also V12 affected**
```
erpnext 12.20.0
frappe 12.16.3
```

## Steps to reproduce the issue

1. create `POS opening entry`
2. sell some products from POS
3. create `POS closing entry`

### Observed result
Expected amount is totally incorrect, it appears by minus value and includes the "Change Amount"
**below is example for:**
```
Sale for:  75
Client paid: 100
Cashier change amount: 25
```
**so the expected amount should be 100-25 = 75** but unfortunately the expected amount is reversed by a "minus - " and value is not correct..
**V13 screenshot**
![Screenshot from 2021-05-08 15-43-33](https://user-images.githubusercontent.com/6287237/117541480-53ff3200-b014-11eb-94f9-1bbe7f3f03a0.png)

**V12 screenshot**
![Screenshot from 2021-05-09 21-27-17](https://user-images.githubusercontent.com/6287237/117584486-96586a00-b10d-11eb-8d7f-767956e0b201.png)

### Expected result
Expected amount should show the correct collected amount (Paid amount - Change amount), for the cash payment method (i think there is change amount only for cash payment method).

## Additional information

OS version / distribution, `ERPNext` install method, etc.
Ubuntu 18.04, ERPNext installed by easyinstall

Labyrinth-00:
I am also having similar issues, I cant make sense of the pos closing vouchers.

Komal-Saraf0609:
Fixed with PR: https://github.com/frappe/erpnext/pull/25868, so closing the issue.

michaeleino:
this is not fixed in version 12
@Komal-Saraf0609  why closed!?

```
ERPNext: v12.25.0 (version-12)
Frappe Framework: v12.22.2 (version-12)
```

PR #25737, opened 2021-05-17
fix: expected amount in pos closing payments table

nextchamp-saqib:
To replicate Issue:
- Create a pos closing with 2 invoices
- Save the closing entry
- Remove one of the invoices
- Save the closing entry
- Expected amount should be incorrect


michaeleino:
this fixes part of this #25638 (the negative value) .. but the amount is still not correct when there is a change amount

PR #25740, opened 2021-05-18
fix: Accumulated depreciation

anushka19:
In asset category in accounts table in accumulated depreciation column it is fetching depreciation and in depreciation column, it is fetching nothing and later if we try submitting it will throw an error to change inside accumulated depreciation column to accumulated depreciation. So now both can be auto fetched.
Before fix:
assets

After Fix:
assets_after


ankush (approved):
LGTM. 👍

PR #25753, opened 2021-05-19
fix: super call syntax error

netchampfaris:


centaur78:
hi  @netchampfaris 

I have updated this.... but though the empty quotation page is rectified. it seems now if you start a new quote and try to select any party.... nothing gets filled... Address is empty and the currency price list defaults to the company currency ??

Strange there are no errors in the console as well

PR #25759, opened 2021-05-19
fix: cannot bypass e-invoicing for non gst item invoices

nextchamp-saqib:
Bypass IRN generation for sales invoices with non-gst item

### DIFF
--- a/erpnext/regional/india/e_invoice/utils.py
+++ b/erpnext/regional/india/e_invoice/utils.py
@@ -43,7 +43,7 @@ def validate_eligibility(doc):
 	invalid_supply_type = doc.get('gst_category') not in ['Registered Regular', 'SEZ', 'Overseas', 'Deemed Export']
 	company_transaction = doc.get('billing_address_gstin') == doc.get('company_gstin')
 	no_taxes_applied = not doc.get('taxes')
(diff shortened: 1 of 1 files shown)


erpnext/regional/india/e_invoice/utils.py
```diff
@@ -43,8 +43,9 @@ def validate_eligibility(doc):
 	invalid_supply_type = doc.get('gst_category') not in ['Registered Regular', 'SEZ', 'Overseas', 'Deemed Export']
 	company_transaction = doc.get('billing_address_gstin') == doc.get('company_gstin')
 	no_taxes_applied = not doc.get('taxes')
+	has_non_gst_item = any(d for d in doc.get('items') if frappe.db.get_value('Item', d.get('item_code'), 'is_non_gst'))
```
deepeshgarg007: `is_non_gst` check is also maintained at the Item child tables which is fetched from item master, you can directly check in the child table instead of querying Item Master

Also what if any invoice has both GST and Non GST items, not sure if this is a valid scenario

nextchamp-saqib: Yeah, will fetch it from child table.

deepeshgarg007 (approved):

PR #25770, opened 2021-05-20
feat: charts

pateljannat:
Fixes in Leave Balance Report

- [ ] Set Leave Period from_date and to_date as a default filter for dates.
- [ ] Removed unused lines of code
- [ ] Added Charts

Screenshot 2021-05-20 at 3 55 44 PM


pateljannat:
This report shows the number of Leaves Allocated, Leaves Takes and Leaves Expired by each employee over a period.

PR #25774, opened 2021-05-21
fix: to update title on party change

Anuja-pawar:
Issue: Title is set by default to Party Name. Hence, it doesn't get updated once the document is saved in case the Party is updated later.

Steps to Replicate:

1. Create any of the invoices and save the draft
2. Change the party (supplier/customer). The title won't be updated.

Solution: Instead of setting default, Title is fetched from Party Name. This would help to update the title if the Party is changed after saving the document.


deepeshgarg007:
@Anuja-pawar closing this for now. We can reopen once we have a better solution

PR #25775, opened 2021-05-21
fix: wrong round off gl entry posted in case of purchase invoice

nextchamp-saqib:
Problem:
- Purchase Invoice with
  - Grand Total = $ 55,107.45
  - Round Off = $ -0.45
  - Rounded Total =  $ 55,107
  - Base Grand Total = 225,940.55
  - Base Round Off = -1.84
  - Base Rounded Total = 225,938.70
- Purchase Invoice creates round off gl entry with `purchase_invoice.py -> make_gle_for_rounding_adjustment()`
  - With this the debit/credit difference = 225,940.55 - 1.84 - 225938.70 = 0.01
- While posting the gl entry
  - In `general_ledger.py -> round_off_debit_credit()`
  - if `precision = 3`, then `allowance = (1.0 / (10**3)) = 0.001`
  - Hence round off entry is again modified with `make_round_off_gle`
  - `make_round_off_gle`
    - Now, the debit/credit is added/substracted by `debit_in_account_currency` which leads to invalid `round_off` value
    - Screenshot 2021-05-21 at 11 14 11 AM
    - For the above eg., `debit_credit_diff = debit_credit_diff - round_off = 0.01 - (-0.45) = 0.46`
    - This now makes `debit_credit_diff` even larger resulting in invalid gl entries and with no validation after this, it is posted successfully

### DIFF
--- a/erpnext/accounts/general_ledger.py
+++ b/erpnext/accounts/general_ledger.py
@@ -185,10 +185,10 @@ def make_round_off_gle(gl_map, debit_credit_diff, precision):
 	for d in gl_map:
 		if d.account == round_off_account:
 			round_off_gle = d
-			if d.debit_in_account_currency:
-				debit_credit_diff -= flt(d.debit_in_account_currency)
+			if d.debit:
+				debit_credit_diff -= flt(d.debit)
 			else:
-				debit_credit_diff += flt(d.credit_in_account_currency)
+				debit_credit_diff += flt(d.credit)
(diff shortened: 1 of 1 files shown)


Muzzy73:
@nextchamp-saqib I believe the same issue is in Sales Invoice as well.

nextchamp-saqib:
@Muzzy73 This should fix for both sales & purchases

PR #25776, opened 2021-05-21
fix: warehouse not found in stock entry

rohitwaghchaure:
**Issue**

1. Created new stock entry type as Outward with purpose as 'Material Transfer'
2. In stock entry selected stock entry type as Outward but system has changed it to Material Transfer and throw below error

Screenshot 2021-05-21 at 11 22 55 AM

### New changes

**Added checkbox add to Transit in the stock entry type to pull automatically in the stock entry**


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -615,7 +614,8 @@ frappe.ui.form.on('Stock Entry', {
 	},
 
 	set_tansit_warehouse: function(frm) {
```
anupamvs: set_tansit_warehouse -> set_transit_warehouse

anupamvs (approved):

PR #25781, opened 2021-05-21
fix(e-invoicing): 'NoneType' object is not iterable

nextchamp-saqib:


erpnext/regional/india/e_invoice/utils.py
```diff
@@ -43,7 +43,7 @@ def validate_eligibility(doc):
 	invalid_supply_type = doc.get('gst_category') not in ['Registered Regular', 'SEZ', 'Overseas', 'Deemed Export']
 	company_transaction = doc.get('billing_address_gstin') == doc.get('company_gstin')
 	no_taxes_applied = not doc.get('taxes')
-	has_non_gst_item = any(d for d in doc.get('items') if d.get('is_non_gst'))
+	has_non_gst_item = any([d for d in doc.get('items') if d.get('is_non_gst')])
```
ankush: This would ideally behave the same way. `any(generator)` = `any(list(generator))`

Please leave an explanation if that is not the case.

nextchamp-saqib: Looks like `doc.get('items')` was the problem. Thanks!

ankush: Cool. You can remove the list comprehension now.

PR #25785, opened 2021-05-21
refactor: Additional Salary form clean up

Anurag810:
TODO:

- [ ] Update screenshot in docs

**Before:**
Screenshot 2021-05-21 at 1 25 18 PM

**Now:**
image

image

### DIFF
--- a/erpnext/payroll/doctype/additional_salary/additional_salary.json
+++ b/erpnext/payroll/doctype/additional_salary/additional_salary.json
@@ -86,7 +86,7 @@
   },
   {
    "depends_on": "eval:(doc.is_recurring==0)",
-   "description": "Date on which this component is applied",
+   "description": "The date on which Salary Component with Amount will contribute for Earnings/Deduction in Salary Slip. ",
    "fieldname": "payroll_date",
    "fieldtype": "Date",
    "in_list_view": 1,
@@ -186,7 +186,7 @@
   {
    "fieldname": "salary_component_and_amount_section",
    "fieldtype": "Section Break",
-   "label": "Salary Component and Amount"
+   "label": "Salary Component And Amount"
   },
   {
    "fieldname": "column_break_13",
@@ -204,7 +204,7 @@
  ],
  "is_submittable": 1,
  "links": [],
- "modified": "2021-05-21 13:33:56.945865",
+ "modified": "2021-05-21 13:57:23.670851",
  "modified_by": "Administrator",
  "module": "Payroll",
  "name": "Additional Salary",
--- a/erpnext/payroll/doctype/additional_salary/additional_salary.json
+++ b/erpnext/payroll/doctype/additional_salary/additional_salary.json
@@ -7,25 +7,30 @@
  "editable_grid": 1,
  "engine": "InnoDB",
  "field_order": [
+  "employee_details_section",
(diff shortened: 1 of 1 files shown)


erpnext/payroll/doctype/additional_salary/additional_salary.json
```diff
@@ -7,25 +7,30 @@
...
   "employee_name",
-  "salary_component",
-  "type",
-  "amount",
-  "ref_doctype",
-  "ref_docname",
-  "amended_from",
   "column_break_5",
   "company",
   "department",
```
ruchamahabal, line 16: `department` field should be below `employee_name` since it is fetched from the `employee` field.

erpnext/payroll/doctype/additional_salary/additional_salary.json
```diff
@@ -7,25 +7,30 @@
...
+  "salary_component_and_amount_section",
+  "salary_component",
   "currency",
+  "amount",
+  "column_break_13",
+  "type",
```
ruchamahabal: `type` (salary component type) field should be below `salary_component` since it is fetched from it.

erpnext/payroll/doctype/additional_salary/additional_salary.json
```diff
@@ -171,11 +177,34 @@
...
+  },
+  {
+   "fieldname": "employee_details_section",
+   "fieldtype": "Section Break",
+   "label": "Employee Details"
+  },
+  {
+   "fieldname": "salary_component_and_amount_section",
+   "fieldtype": "Section Break",
+   "label": "Salary Component And Amount"
```
ruchamahabal:
```suggestion
   "fieldname": "salary_details_section",
   "fieldtype": "Section Break",
   "label": "Salary Details"
```
Since this section also covers dates and recurring property, we should label it Salary Details instead.

erpnext/payroll/doctype/additional_salary/additional_salary.json
```diff
@@ -171,11 +177,34 @@
...
+  },
+  {
+   "fieldname": "column_break_13",
+   "fieldtype": "Column Break"
+  },
+  {
+   "fieldname": "additional_salary_properties_and_references_section",
+   "fieldtype": "Section Break",
+   "label": "Additional Salary Properties And References"
```
ruchamahabal:
```suggestion
  {
   "fieldname": "properties_and_references_section",
   "fieldtype": "Section Break",
   "label": "Properties and References"
```
Since the doctype name is Additional Salary, we don't need to write it again in the Section title. Also, words like "and" should not be title cased.

ruchamahabal (changes requested):

erpnext/payroll/doctype/additional_salary/additional_salary.json
Anurag810: @ruchamahabal 
Yeah, that was in my mind but. There will be three fields in a single column having lots of (employee_name, department, and employee), which will leave only one field for second column. causing lots of white space in the second column. like three fields in the first column and only single fields in the second column, but anyway **they are in the same section** (**EmPloyee details**).

ruchamahabal (approved):

PR #25812, opened 2021-05-24
fix: profitability test

pateljannat:
Fix failing test

```
ERROR: test_project_profitability (erpnext.projects.report.project_profitability.test_project_profitability.TestProjectProfitability)
1000.0 1000.0
5000.0 5000.0
  File "/home/runner/frappe-bench/apps/erpnext/erpnext/projects/report/project_profitability/test_project_profitability.py", line 46, in test_project_profitability
5000.0 5000.0
    utilization = timesheet.total_billed_hours/(self.salary_slip.total_working_days * standard_working_hours)
1000.0 1000.0
```

### DIFF
--- a/erpnext/projects/report/project_profitability/test_project_profitability.py
+++ b/erpnext/projects/report/project_profitability/test_project_profitability.py
@@ -43,7 +43,6 @@ class TestProjectProfitability(unittest.TestCase):
 		self.assertEqual(self.salary_slip.total_working_days, row.total_working_days)
 
 		standard_working_hours = frappe.db.get_single_value("HR Settings", "standard_working_hours")
-		print("standard_working_hours", standard_working_hours)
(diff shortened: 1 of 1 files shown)


erpnext/projects/report/project_profitability/test_project_profitability.py
```diff
@@ -9,7 +9,7 @@
 
 class TestProjectProfitability(unittest.TestCase):
 	@classmethod
-	def setUp(self):
+	def setUpClass(self):
```
ankush:
```suggestion
	def setUpClass(cls):
```

`self` should be changed to `cls` everywhere in classmethods. Your rest of the functions will still work as expected.

pateljannat: keeping setUp and tearDown instead of class methods for now

ankush (approved):

PR #25822, opened 2021-05-25
feat: enable/disable gl entry posting for change given in pos

nextchamp-saqib:
For a Point of Sale Transaction, if a change amount is given then Debtors posting is undesirable for some users.
(For Rs. 10 change given)

If GL posting of change amount enabled: 

Screenshot 2021-05-25 at 1 23 37 PM

If GL posting of change amount disabled:

Screenshot 2021-05-25 at 1 23 35 PM

Docs - https://github.com/frappe/erpnext_documentation/pull/349


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -253,14 +254,20 @@
   {
    "fieldname": "column_break_19",
    "fieldtype": "Column Break"
+  },
+  {
+   "default": "1",
+   "fieldname": "post_change_gl_entries",
+   "fieldtype": "Check",
+   "label": "Post Ledger Entries for Given Change"
```
nabinhait, line 262: @nextchamp-saqib Please add field description, no context for Change in Accounts Settings. The label can be "Create Ledger Entries for Change Amount"

PR #25828, opened 2021-05-25
fix: Add Hold status column in the Issue Summary Report

Anuja-pawar:
On Hold (status) column is added in the Issue Summary report

Screenshot 2021-05-26 at 11 22 01 AM

### DIFF
--- a/erpnext/support/report/issue_summary/issue_summary.js
+++ b/erpnext/support/report/issue_summary/issue_summary.js
@@ -42,6 +42,7 @@ frappe.query_reports["Issue Summary"] = {
 				"",
 				{label: __('Open'), value: 'Open'},
 				{label: __('Replied'), value: 'Replied'},
+				{label: __('Hold'), value: 'Hold'},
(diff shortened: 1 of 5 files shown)


erpnext/support/report/issue_summary/issue_summary.js
```diff
@@ -42,6 +42,7 @@ frappe.query_reports["Issue Summary"] = {
 				"",
 				{label: __('Open'), value: 'Open'},
 				{label: __('Replied'), value: 'Replied'},
+				{label: __('Hold'), value: 'Hold'},
```
nextchamp-saqib: Can we rename this as "On Hold"? "On Hold Issues" makes more sense than "Hold Issues"

anupamvs: 'On Hold' seems good.

ruchamahabal:
This will need a patch to update existing issues to **On Hold** from **Hold**

deepeshgarg007 (approved):

PR #25829, opened 2021-05-25
fix: add validation for 'for_qty' else throws errors

18alantom:
### Description
- Add validation for **Pick List**'s `for_qty`, because setting this value to 0 throws errors, and `for_qty = 0` doesn't have a use case.
- Removed `read_only` because the value is set in a dialog which can't be retriggered unless the **Work Order** is set again.

### DIFF
--- a/erpnext/stock/doctype/pick_list/pick_list.json
+++ b/erpnext/stock/doctype/pick_list/pick_list.json
@@ -71,8 +71,7 @@
    "description": "Qty of raw materials will be decided based on the qty of the Finished Goods Item",
    "fieldname": "for_qty",
    "fieldtype": "Float",
-   "label": "Qty of Finished Goods Item",
-   "read_only": 1
+   "label": "Qty of Finished Goods Item"
   },
   {
    "fieldname": "amended_from",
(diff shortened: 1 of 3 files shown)


ankush:
@18alantom can you have a look at the failing tests?

erpnext/stock/doctype/pick_list/pick_list.json
```diff
@@ -71,8 +71,7 @@
    "description": "Qty of raw materials will be decided based on the qty of the Finished Goods Item",
    "fieldname": "for_qty",
    "fieldtype": "Float",
-   "label": "Qty of Finished Goods Item",
-   "read_only": 1
+   "label": "Qty of Finished Goods Item"
```
ankush:
```suggestion
   "label": "Qty of Finished Goods Item",
   "read_only": 1
```
What is the rationale for making this editable?

ankush (changes requested):

erpnext/stock/doctype/pick_list/pick_list.json
```diff
@@ -114,7 +114,7 @@
  ],
  "is_submittable": 1,
  "links": [],
- "modified": "2020-03-17 11:38:41.932875",
+ "modified": "2021-05-25 16:51:37.210932",
```
ankush:
```suggestion
 "modified": "2020-03-17 11:38:41.932875",
```

schema not modified. So this change can be reverted.

ankush (approved):

ankush:
please port this to `develop` @18alantom

PR #25830, opened 2021-05-25
fix: Broken links in help

anushka19:
In Item and Asset Category doctype the documentation links  help were broken.

### DIFF
--- a/erpnext/public/js/help_links.js
+++ b/erpnext/public/js/help_links.js
@@ -644,14 +644,14 @@ frappe.help.help_links["List/Payment Request"] = [
 frappe.help.help_links["List/Asset"] = [
 	{
 		label: "Managing Fixed Assets",
-		url: docsUrl + "user/manual/en/accounts/managing-fixed-assets",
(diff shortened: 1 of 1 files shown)


erpnext/assets/doctype/asset_category/asset_category.js
```diff
@@ -4,7 +4,7 @@
 frappe.ui.form.on('Asset Category', {
 	onload: function(frm) {
 		frm.add_fetch('company_name', 'accumulated_depreciation_account', 'accumulated_depreciation_account');
-		frm.add_fetch('company_name', 'depreciation_expense_account', 'accumulated_depreciation_account');
+		frm.add_fetch('company_name', 'depreciation_expense_account', 'depreciation_expense_account');
```
ankush: I think these two commits are not required. That PR is already merged.

While git will handle this in most cases, it's better to avoid this. Do `git rebase -i HEAD~3` and drop them from your branch and force push.

anushka19: okay done

ankush (approved):

PR #25844, opened 2021-05-26
refactor(asset maintenance log): streamline asset maintenance log fields

jmlescain:
Rationale: Streamline the Asset Maintenance Log.

- Separate Description to another section and make it collapsible (Default is collapsed.) This is beneficial for longer descriptions. Old behavior had the description on a column break.
Old:
Screenshot from 2021-05-26 18-50-49
New:

- Prefill "Actions Performed" with the Asset Maintenance Task Description. Actions to be taken might be described in the Task Description anyway. (Only fetch if the field is empty, to prevent overwriting.)

- Set "Actions Performed" as read only after submit.

This changes doesn't need updates to the docs. Will also not need new tests.

### DIFF
--- a/erpnext/assets/doctype/asset_maintenance_log/asset_maintenance_log.json
+++ b/erpnext/assets/doctype/asset_maintenance_log/asset_maintenance_log.json
@@ -25,6 +25,7 @@
   "assign_to_name",
   "due_date",
   "completion_date",
+  "section_break_7",
   "description",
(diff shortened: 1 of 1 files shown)


jeremythegreat01:
Hi Team Frappe ( @deepeshgarg007 and @nextchamp-saqib),

Good day!

Have you had a chance to look into this?

Thank you.

PR #25847, opened 2021-05-26
fix: Create POS Invoice for Product Bundles

GangaManoj:
_Issue:_
Users are unable to create POS Invoices for Product Bundles, as their item_codes correspond to non-stock Items.

_Fix:_
- Check if non-stock items are Product Bundles.
- If so, check stock availability of included Items.

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -139,7 +139,12 @@ class POSInvoice(SalesInvoice):
 				if allow_negative_stock:
 					return
 
-				available_stock = get_stock_availability(d.item_code, d.warehouse)
+				if frappe.db.get_value('Item', d.item_code, 'is_stock_item'):
+					available_stock = get_stock_availability(d.item_code, d.warehouse)
+				else:
+					if frappe.db.exists('Product Bundle', d.item_code):
+						available_stock = get_bundle_availability(d.item_code, d.warehouse)
+
 				item_code, warehouse, qty = frappe.bold(d.item_code), frappe.bold(d.warehouse), frappe.bold(d.qty)
 				if flt(available_stock) <= 0:
 					frappe.throw(_('Row #{}: Item Code: {} is not available under warehouse {}.')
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -139,7 +139,12 @@ def validate_stock_availablility(self):
 				if allow_negative_stock:
 					return
 
-				available_stock = get_stock_availability(d.item_code, d.warehouse)
+				if frappe.db.get_value('Item', d.item_code, 'is_stock_item'):
```
nextchamp-saqib: Let's handle these check inside the `get_stock_availablity()` since the same method is being used in `point_of_sale.py`

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -455,15 +461,32 @@ def get_existing_payment_request(self, pay):
...
+	product_bundle = frappe.get_doc('Product Bundle', bundle_item_code)
+
+	bundle_bin_qty = 1000000
+	for item in product_bundle.items:
+		item_bin_qty = get_bin_qty(item.item_code, warehouse)
+
+		max_num_of_bundles = item_bin_qty / item.qty
+		if bundle_bin_qty > max_num_of_bundles:
+			bundle_bin_qty = max_num_of_bundles
+
```
nextchamp-saqib:
```suggestion
	bundle_bin_qty = 1000000
	for item in product_bundle.items:
		sub_item_bin_qty = get_bin_qty(item.item_code, warehouse)
		sub_item_pos_reserved_qty = get_pos_reserved_qty(item.item_code, warehouse)
		
		available_qty = sub_item_bin_qty - sub_item_pos_reserved_qty
		
		max_available_bundles = available_qty / item.qty
		if bundle_bin_qty > max_available_bundles:
			bundle_bin_qty = max_available_bundles

```

nextchamp-saqib: Can you check if this works? I just wanted to add pos reserved qty of sub-items since sub-items could be sold individually, but ended up renaming variables.

nextchamp-saqib (changes requested):

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
GangaManoj: yep, it works.

nextchamp-saqib (approved):

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -213,8 +214,9 @@ def validate_non_stock_items(self):
 		for d in self.get("items"):
 			is_stock_item = frappe.get_cached_value("Item", d.get("item_code"), "is_stock_item")
 			if not is_stock_item:
-				frappe.throw(_("Row #{}: Item {} is a non stock item. You can only include stock items in a POS Invoice. ")
-					.format(d.idx, frappe.bold(d.item_code)), title=_("Invalid Item"))
+				if not frappe.db.exists('Product Bundle', d.item_code):
+					frappe.throw(_("Row #{}: Item {} is a non stock item. You can only include stock items in a POS Invoice. ")
```
nextchamp-saqib:
```suggestion
					frappe.throw(_("Row #{}: Item {} is a non stock item. You can only include stock items in a POS Invoice.")
```

PR #25861, opened 2021-05-28
fix(Payment Entry): Incorrect allocated amount

GangaManoj:
_Issues:_
- Amount is positive for return Sales Invoices when rate is negative.

Screenshot 2021-05-28 at 5 33 15 AM

- Total Allocated Amount is positive for Payment Entries that contain only return Invoices(which have negative values for allocated_amount).

_Fix:_
- Throw an error when negative rate is entered for return Sales Invoices.
- Set Total Allocated Amount as zero if sum of allocated_amount for all invoices is negative.

Screenshot 2021-05-28 at 5 39 49 AM

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -413,8 +413,12 @@ class PaymentEntry(AccountsController):
 				base_total_allocated_amount += flt(flt(d.allocated_amount) * flt(d.exchange_rate),
 					self.precision("base_paid_amount"))
 
-		self.total_allocated_amount = abs(total_allocated_amount)
-		self.base_total_allocated_amount = abs(base_total_allocated_amount)
+		if total_allocated_amount > 0:
+			self.total_allocated_amount = total_allocated_amount
+			self.base_total_allocated_amount = base_total_allocated_amount
+		else:
+			self.total_allocated_amount = 0
+			self.base_total_allocated_amount = 0
 
 	def set_unallocated_amount(self):
 		self.unallocated_amount = 0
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -413,8 +413,12 @@ def set_total_allocated_amount(self):
 				base_total_allocated_amount += flt(flt(d.allocated_amount) * flt(d.exchange_rate),
 					self.precision("base_paid_amount"))
 
-		self.total_allocated_amount = abs(total_allocated_amount)
-		self.base_total_allocated_amount = abs(base_total_allocated_amount)
+		if total_allocated_amount > 0:
```
deepeshgarg007, line 416: I guess this was intentional, only the allocated amount in references should be negative and the total allocated is intentionally kept positive. Just add validation for this
in the `validate_allocated_amount` method. But -300 can still be allocated against -500.
Also please elaborate on what issue allocation like this caused that is still a bit unclear to me

erpnext/controllers/status_updater.py
```diff
@@ -118,6 +118,7 @@ class StatusUpdater(Document):
 	def update_prevdoc_status(self):
 		self.update_qty()
 		self.validate_qty()
+		self.validate_rate()
```
deepeshgarg007, line 121: Is this related to this fix?

deepeshgarg007 (changes requested):

erpnext/controllers/status_updater.py
GangaManoj: Yeah, it's for the first issue I'd mentioned in the PR description. Vrinda Oza had pointed out that the amount was positive in her return invoice, even though it was supposed to be negative. It was because she'd entered the rate(and not just the quantity) as negative.

PR #25876, opened 2021-05-28
fix: choose correct Salary Structure Assignment when getting data for formula eval

### LINKED ISSUE #21307: [Bug-V12] HR: Salary Structure Assignments not respected by Salary Slips
## Description of the issue
Multiple Salary Structure Assignments can be created for the same employee with different `From Date`s. The documentation states:
> Select the From Date from which this particular Salary Structure will be applicable.

When creating a salary slip, I would expect the `From Date` to be respected (the Salary Slip Start/End Dates should be used to select the relevant Assignment), and also any Cancelled Salary Structure Assignments should be ignored. Neither of these are true.

**Output of `bench version`**
```
erpnext 12.6.0
frappe 12.4.1
```

1. Create an employee
2. Create a basic Salary Structure that picks up the base amount
3. (You may need to create other HR objects)
4. Create two Salary Structure Assignments for an Employee: one with `From Date` in the past (2019-04-01) with Base of 100 and one in the future (2020-05-01) with Base of 200
5. Cancel the future Assignment
6. Create a Salary Slip for the employee

### Observed result
The Salary Slip for the current month will be for 200 using the future, cancelled Assignment

### Expected result
The Salary Slip should be for 100 using the old Active Assignment

### LINKED ISSUE #24601: Salary slip picks latest Salary Structure Assignment even if the salary structure is cancelled.-v13 beta 11
TO replicate

Create a salary structure assignment, say on date 1-Jan-2021, and submit it

Now, create another salary structure assignment, on 5-Jan-2021, and cancel it

The salary slip will pick the base on the second salary structure assignment created on 5-Jan-2021

sagarvora:
Resolves #24601 
Resolves #21307

## Changes Made
- Improved validation for Payroll Period to see if employee hasn't joined or has already left
- Choosing correct salary structure with following additional filters:
    - Salary Structure Assignment's **From Date** should be less than or equal to (Payroll Period Start Date or Employee Joining Date whichever is later)
    - Salary Structure Assignment should be submitted
- Gets the Salary Structure Assignment with From Date closest to Payroll Start Date / Employee Joining Date instead of choosing the one which was last modified (`order_by`)
- Error message if Salary Structure Assignment with above filters not found

### DIFF
--- a/erpnext/payroll/doctype/salary_structure/test_salary_structure.py
+++ b/erpnext/payroll/doctype/salary_structure/test_salary_structure.py
@@ -6,7 +6,7 @@ import frappe
 import unittest
 import erpnext
 from frappe.utils.make_random import get_random
-from frappe.utils import nowdate, add_days, add_years, getdate, add_months
+from frappe.utils import nowdate, add_days, add_years, getdate, add_months, get_first_day, date_diff
 from erpnext.payroll.doctype.salary_structure.salary_structure import make_salary_slip
 from erpnext.payroll.doctype.salary_slip.test_salary_slip import make_earning_salary_component,\
 	make_deduction_salary_component, make_employee_salary_slip, create_tax_slab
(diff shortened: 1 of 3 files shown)


sagarvora:
Changes have been tested locally.

sagarvora:
Breaking tests are unrelated. Directly linked tests have been fixed.

erpnext/payroll/doctype/salary_structure/test_salary_structure.py
```diff
@@ -139,10 +140,23 @@ def make_salary_structure(salary_structure, payroll_frequency, employee=None, do
 	else:
 		salary_structure_doc = frappe.get_doc("Salary Structure", salary_structure)
 
+	filters = {'employee':employee, 'docstatus': 1}
+	if not from_date and payroll_period:
+		from_date = payroll_period.start_date
```
sagarvora, line 145: - This code has been added because after newly introduced validation, a Salary Structure Assignment must be applicable on or before the Salary Slip date.
 - This code assumes that employee has joined before Payroll Period. Alternatively, `from_date` can be manually passed to `make_salary_structure` in these cases.

ruchamahabal:
LGTM

ruchamahabal (approved):

ruchamahabal:
@sagarvora Can you also raise a fix for `version-13-hotfix` branch?

PR #25877, opened 2021-05-28
fix(Payroll): no need to set `abbr` specially in Salary Slip

sagarvora:
This PR reverts #25117 and fixes the main issue: namely missing comma.
Refer https://github.com/frappe/erpnext/pull/25117/files#r641717041 for more info.

### DIFF
--- a/erpnext/payroll/doctype/salary_slip/salary_slip.py
+++ b/erpnext/payroll/doctype/salary_slip/salary_slip.py
@@ -621,15 +621,12 @@ class SalarySlip(TransactionBase):
 
 			component_row = self.append(component_type)
 			for attr in (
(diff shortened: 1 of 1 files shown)


ruchamahabal:
I have to check one case before merging this.

ruchamahabal:
Was aware of the missing comma while raising the earlier fix but the deduction component was still not getting computed in some cases. I checked all occurrences of `abbr`. The `salary_component_abbr` field from Salary Component gets fetched with the alias `abbr` everywhere. So the deductions are coming through just fine now. Merging.

ruchamahabal (approved):

sagarvora:
Thanks, @ruchamahabal!

PR #25886, opened 2021-05-30
fix: order and time of operations in multilevel BOM work order

ankush:
**Problem**: When multilevel BOM is ticked, work order tries to fetch all operations in single query and sorts by `idx` which gives meaningless ordering.

**Solution**: order by level (reversed) i.e. in a tree representation bottom-most BOM operations are listed first.

Before (meaningless ordering, wrong index, operations time not considering exploded qty):
Screenshot 2021-06-11 at 2 49 04 PM

After (Order of operations = reverse level order traversal, operations time considering exploded qty):
Screenshot 2021-06-11 at 2 49 51 PM

Also added `__repr__` for BOMTree that shows complete BOM like this:
Screenshot 2021-06-11 at 2 25 28 PM

TODO:

- [x] Get BOM as easily parsable tree structure. (recursive dict)
- [x] Fix orders of operations in Work Order
- [x] bug with item level ops not getting added to list
- [x] Better tree repr of BOM
- [x] Correct operation time for multilevel BOM 
- [x] tests (redo for changes)

WARNING: Ideally, this ordering should not be relied upon and a separate work order should be made for each subassembly. You lose the benefit of parallel traceability and there could be duplication if there's a BOM that's part of two different nodes in the BOM tree. This change is for sake of correctness.

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -1,7 +1,8 @@
 # Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 # License: GNU General Public License v3. See license.txt
 
-from __future__ import unicode_literals
+from typing import List
+from collections import deque
 import frappe, erpnext
 from frappe.utils import cint, cstr, flt, today
 from frappe import _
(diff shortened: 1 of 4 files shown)


marination:
@ankush failing test `test_pending_and_transferred_qty` fixed https://github.com/frappe/erpnext/pull/26043

ankush:
@marination rebased branch 👍

marination (approved):

PR #25891, opened 2021-05-31
refactor: update cost updates operation time and hour rates in BOM

noahjacob:
**Changes:**
**Workstation**
- Hour rates are no longer updated directly in the BOM when rates are changed in the workstations.

**Update Cost (BOM)**
- Update cost now fetches latest rates from workstation and routing(if linked).

_Operation Time_
routing

_Hour Rates_
routing_work

- Added test cases.
- **[Docs](https://github.com/frappe/erpnext_documentation/pull/345)**

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -81,7 +81,7 @@ class BOM(WebsiteGenerator):
 		self.validate_operations()
 		self.calculate_cost()
 		self.update_stock_qty()
-		self.update_cost(update_parent=False, from_child_bom=True, save=False)
+		self.update_cost(update_parent=False, from_child_bom=True, update_hour_rate = False, save=False)
 
 	def get_context(self, context):
 		context.parents = [{'name': 'boms', 'title': _('All BOMs') }]
@@ -213,7 +213,7 @@ class BOM(WebsiteGenerator):
 		return flt(rate) * flt(self.plc_conversion_rate or 1) / (self.conversion_rate or 1)
 
 	@frappe.whitelist()
-	def update_cost(self, update_parent=True, from_child_bom=False, save=True):
+	def update_cost(self, update_parent=True, from_child_bom=False, update_hour_rate = True, save=True):
 		if self.docstatus == 2:
 			return
 
@@ -242,7 +242,7 @@ class BOM(WebsiteGenerator):
 
 		if self.docstatus == 1:
 			self.flags.ignore_validate_update_after_submit = True
-			self.calculate_cost()
+			self.calculate_cost(update_hour_rate)
 		if save:
 			self.db_update()
 
@@ -403,31 +403,46 @@ class BOM(WebsiteGenerator):
 		bom_list.reverse()
 		return bom_list
 
-	def calculate_cost(self):
+	def calculate_cost(self, update_hour_rate = False):
 		"""Calculate bom totals"""
-		self.calculate_op_cost()
+		self.calculate_op_cost(update_hour_rate)
 		self.calculate_rm_cost()
 		self.calculate_sm_cost()
 		self.total_cost = self.operating_cost + self.raw_material_cost - self.scrap_material_cost
 		self.base_total_cost = self.base_operating_cost + self.base_raw_material_cost - self.base_scrap_material_cost
 
-	def calculate_op_cost(self):
+	def calculate_op_cost(self, update_hour_rate = False):
 		"""Update workstation rate and calculates totals"""
 		self.operating_cost = 0
 		self.base_operating_cost = 0
 		for d in self.get('operations'):
 			if d.workstation:
-				if not d.hour_rate:
-					hour_rate = flt(frappe.db.get_value("Workstation", d.workstation, "hour_rate"))
-					d.hour_rate = hour_rate / flt(self.conversion_rate) if self.conversion_rate else hour_rate
-
-			if d.hour_rate and d.time_in_mins:
-				d.base_hour_rate = flt(d.hour_rate) * flt(self.conversion_rate)
-				d.operating_cost = flt(d.hour_rate) * flt(d.time_in_mins) / 60.0
-				d.base_operating_cost = flt(d.operating_cost) * flt(self.conversion_rate)
+				self.update_rate_and_time(d, update_hour_rate)
 
 			self.operating_cost += flt(d.operating_cost)
 			self.base_operating_cost += flt(d.base_operating_cost)
+
+	def update_rate_and_time(self, row, update_hour_rate = False):
+		if not row.hour_rate or update_hour_rate:
+			hour_rate = flt(frappe.get_cached_value("Workstation", row.workstation, "hour_rate"))
+			row.hour_rate = (hour_rate / flt(self.conversion_rate)
+				if self.conversion_rate and hour_rate else hour_rate)
+
+			if self.routing:
+				row.time_in_mins = flt(frappe.db.get_value("BOM Operation", {
+						"workstation": row.workstation,
+						"operation": row.operation,
+						"sequence_id": row.sequence_id,
+						"parent": self.routing
+				}, ["time_in_mins"]))
+
+		if row.hour_rate and row.time_in_mins:
+			row.base_hour_rate = flt(row.hour_rate) * flt(self.conversion_rate)
+			row.operating_cost = flt(row.hour_rate) * flt(row.time_in_mins) / 60.0
+			row.base_operating_cost = flt(row.operating_cost) * flt(self.conversion_rate)
+
+		if update_hour_rate:
+			row.db_update()
 
 	def calculate_rm_cost(self):
 		"""Fetch RM rate as per today's valuation rate and calculate totals"""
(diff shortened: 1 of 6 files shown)


erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,15 @@
 class Routing(Document):
 	def validate(self):
 		self.set_routing_id()
+		self.update_op_time()
```
ankush: This should probably done after-saving i.e. in `on_update` method?

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,15 @@
...
+
+	def update_op_time(self):
```
ankush:
```suggestion
	def update_operation_time_in_linked_BOMs(self):
```

This or something shorter that conveys what function does.

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,15 @@
...
+		bom_list = frappe.db.sql("""select name, routing from `tabBOM`
```
ankush: No need to select `routing` field. Also you can replace this with ORM function, something like:

```python
frappe.get_list("BOM", filters={"routing": self.name}, pluck="name")
```

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,15 @@
...
+			where routing = %s""", (self.name))
+		for bom_no in bom_list:
+			frappe.db.sql("""update `tabBOM Operation` set time_in_mins = %s
+				where parent = %s """,
+				(self.operations[0].time_in_mins, bom_no[0]))
```
ankush: I am not understanding this part well. All operations are getting set to the same value? `self.operations[0].time_in_mins`

ankush (changes requested):
Add 1-2 test cases for this and also a note in docs.

erpnext/manufacturing/doctype/routing/routing.py
noahjacob: Updated it, previous one was incorrect.

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,18 @@
 class Routing(Document):
 	def validate(self):
 		self.set_routing_id()
+	
+	def on_update(self):
+		self.update_operation_time_in_linked_BOMs()
```
marination: contrary to what ankush suggested, this is very non standard naming for our codebase
```suggestion
		self.update_operation_time_in_linked_boms()
```

erpnext/manufacturing/doctype/routing/test_routing.py
```diff
@@ -21,6 +21,20 @@ def setUpClass(cls):
...
+		setup_operations(operations)
+		routing_doc = create_routing(routing_name="Testing Route", operations=operations)
+		bom_doc = setup_bom(item_code=self.item_code, routing=routing_doc.name)
+		self.assertEqual(routing_doc.operations[0].time_in_mins, 30)
+		routing_doc.operations[0].time_in_mins = 90
+		routing_doc.save()
+		self.assertEqual(routing_doc.operations[0].time_in_mins, 90)
+		bom_doc.reload()
+		self.assertEqual(bom_doc.operations[0].time_in_mins, 90)
+
```
ankush:
```suggestion
		routing_doc.operations[0].time_in_mins = 90
		routing_doc.operations[1].time_in_mins = 42.2
		routing_doc.save()
		self.assertEqual(routing_doc.operations[0].time_in_mins, 90)
		self.assertEqual(routing_doc.operations[1].time_in_mins, 42.2)
		bom_doc.reload()
		self.assertEqual(bom_doc.operations[0].time_in_mins, 90)
		self.assertEqual(bom_doc.operations[1].time_in_mins, 42.2)
		
```

Might be better to add check rows other than just the first one. Also, the operations cost update should be checked.

erpnext/manufacturing/doctype/routing/test_routing.py
```diff
@@ -21,6 +21,20 @@ def setUpClass(cls):
 	def tearDownClass(cls):
 		frappe.db.sql('delete from tabBOM where item=%s', cls.item_code)
 
+	def test_update_operation_time_in_linked_boms(self):
+		operations = [{"operation": "Test Operation A", "workstation": "Test Workstation A", "time_in_mins": 30},
+			{"operation": "Test Operation B", "workstation": "Test Workstation A", "time_in_mins": 20}]
+
```
ankush:
```suggestion

```

Enable stripping of trailing whitespace in your editor.

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -11,6 +11,18 @@
...
+	def on_update(self):
+		self.update_operation_time_in_linked_boms()
+
+	def update_operation_time_in_linked_boms(self):
+		bom_list = frappe.get_list("BOM", filters={"routing": self.name}, pluck="name")
+		for bom_no in bom_list:
+			for operation in self.operations:
+				frappe.db.sql("""
+					update `tabBOM Operation` set time_in_mins = %s
+					where operation = %s and parent = %s """,
```
ankush: This updates the row-level time_in_mins but dependent values such as row-level operations cost and total operations cost are not updated.

erpnext/manufacturing/doctype/routing/test_routing.py
```diff
@@ -61,6 +75,7 @@ def create_routing(**args):
 
 	if not args.do_not_save:
 		try:
+			doc.set('operations', {})
```
ankush:
```suggestion
			doc.set('operations', [])
```
Child table fields are a list.

ankush (changes requested):

erpnext/manufacturing/doctype/workstation/test_workstation.py
```diff
@@ -34,9 +98,12 @@ def make_workstation(*args, **kwargs):
...
+		doc.hour_rate_labour = args.hour_rate_labour
 		doc.insert()
 
 		return doc
 	except frappe.DuplicateEntryError:
-		return frappe.get_doc("Workstation", workstation_name)
\ No newline at end of file
+		return frappe.get_doc("Workstation", workstation_name)
+
+
```
ankush:
```suggestion
```

erpnext/manufacturing/doctype/workstation/test_workstation.py
```diff
@@ -34,9 +98,12 @@ def make_workstation(*args, **kwargs):
 			"doctype": "Workstation",
 			"workstation_name": workstation_name
 		})
-
+		doc.hour_rate_rent = args.hour_rate_rent
+		doc.hour_rate_labour = args.hour_rate_labour
```
ankush:
```suggestion
		doc.hour_rate_rent = args.get("hour_rate_rent")
		doc.hour_rate_labour = args.get("hour_rate_labour")
```

Not mandatory fields might cause unnecessary test failures.

erpnext/manufacturing/doctype/routing/routing.py
```diff
@@ -4,14 +4,27 @@
...
 from __future__ import unicode_literals
 import frappe
-from frappe.utils import cint
+from frappe.utils import cint, flt
 from frappe import _
 from frappe.model.document import Document
 
 class Routing(Document):
+	def onload(self):
+		self.calculate_operating_cost()
```
ankush:
```suggestion
```

This will be very confusing down the line. It will show one value when loaded from desk and another in other computations.

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -403,32 +403,45 @@ def _get_children(bom_no):
...
+		if not row.hour_rate or update_hour_rate:
+			hour_rate = flt(frappe.get_cached_value("Workstation", row.workstation, "hour_rate"))
+			row.hour_rate = (hour_rate / flt(self.conversion_rate)
+				if self.conversion_rate and hour_rate else hour_rate)
+
+			if self.routing:
+				row.time_in_mins = flt(frappe.db.get_value("BOM Operation", {
+						"workstation": row.workstation,
+						"parent": self.routing
+				}, ["time_in_mins"]))
```
ankush: There is no filter for which row time needs to be fetched. This seems to be buggy. Can add additional filter for operation and sequence_id to be sure.

erpnext/manufacturing/doctype/routing/test_routing.py
```diff
@@ -48,7 +46,40 @@ def test_sequence_id(self):
...
+			{
+				"operation": "Test Operation A",
+				"workstation": "_Test Workstation A",
+				"hour_rate_rent": 300,
+				"hour_rate_labour": 750 ,
+				"time_in_mins": 30
+			},
+			{
+				"operation": "Test Operation B",
+				"workstation": "_Test Workstation B",
```
ankush, line 60: This test will likely fail if both operations are performed on same workstation. Refer comment above ^

ankush (changes requested):

ankush (approved):

ankush:
@noahjacob port this to `develop`

PR #25899, opened 2021-05-31
fix: wrap dates in getdate for leave application

pateljannat:
### Issue

1. An error used to appear on saving the leave application that was backdated.
2. It was because of the inconsistencies in the date formats.

Screenshot 2021-05-31 at 4 25 42 PM

### Steps to Replicate

1. Go to HR Settings doctype. You will find a section called Leave Settings.
2. It has a field called **Restrict Backdated Leave Application**. Check it.
3. Create a backdated Leave Application. Save.
4. It should save without throwing the error mentioned above.

### DIFF
--- a/erpnext/hr/doctype/leave_application/leave_application.py
+++ b/erpnext/hr/doctype/leave_application/leave_application.py
@@ -4,7 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 from frappe import _
-from frappe.utils import cint, cstr, date_diff, flt, formatdate, getdate, today, get_fullname, add_days, nowdate
+from frappe.utils import cint, cstr, date_diff, flt, formatdate, getdate, get_link_to_form, today, get_fullname, add_days, nowdate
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -85,7 +84,7 @@ def validate_applicable_after(self):
 
 	def validate_dates(self):
 		if frappe.db.get_single_value("HR Settings", "restrict_backdated_leave_application"):
-			if self.from_date and self.from_date < frappe.utils.today():
+			if self.from_date and getdate(self.from_date) < getdate(today()):
```
ruchamahabal:
```suggestion
			if self.from_date and getdate(self.from_date) < getdate():
```
If you are using `getdate()` you don't need `today()` since getdate by default returns current date.

ruchamahabal (approved):

PR #25900, opened 2021-05-31
fix: added separate table to track scheduled time logs in job card

rohitwaghchaure:
**Issue**

System auto set the time logs in the job card even though job has not started
Screenshot 2021-05-31 at 5 42 30 PM

**After Fix**
Added separate tables to track scheduled time logs and actual time logs

Screenshot 2021-05-31 at 5 41 51 PM

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.py
+++ b/erpnext/manufacturing/doctype/job_card/job_card.py
@@ -25,6 +25,7 @@ class JobCard(Document):
 		self.set_status()
 		self.validate_operation_id()
 		self.validate_sequence_id()
+		self.set_expected_and_actual_time()
 
 	def validate_time_logs(self):
 		self.total_completed_qty = 0.0
@@ -49,8 +50,11 @@ class JobCard(Document):
 
 			self.total_completed_qty = flt(self.total_completed_qty, self.precision("total_completed_qty"))
 
-	def get_overlap_for(self, args, check_next_available_slot=False):
+	def get_overlap_for(self, args, check_next_available_slot=False, doctype=None):
 		production_capacity = 1
+
+		if not doctype:
+			doctype = 'Job Card Time Log'
 
 		if self.workstation:
 			production_capacity = frappe.get_cached_value("Workstation",
@@ -67,14 +71,14 @@ class JobCard(Document):
 			extra_cond = " or (%(from_time)s <= jctl.from_time and %(to_time)s <= jctl.to_time)"
 
 		existing = frappe.db.sql("""select jc.name as name, jctl.to_time from
-			`tabJob Card Time Log` jctl, `tabJob Card` jc where jctl.parent = jc.name and
+			`tab{2}` jctl, `tabJob Card` jc where jctl.parent = jc.name and
 			(
 				(%(from_time)s > jctl.from_time and %(from_time)s < jctl.to_time) or
 				(%(to_time)s > jctl.from_time and %(to_time)s < jctl.to_time) or
 				(%(from_time)s <= jctl.from_time and %(to_time)s >= jctl.to_time) {0}
 			)
 			and jctl.name != %(name)s and jc.name != %(parent)s and jc.docstatus < 2 {1}
-			order by jctl.to_time desc limit 1""".format(extra_cond, validate_overlap_for),
+			order by jctl.to_time desc limit 1""".format(extra_cond, validate_overlap_for, doctype),
 			{
 				"from_time": args.from_time,
 				"to_time": args.to_time,
@@ -86,6 +90,10 @@ class JobCard(Document):
 
 		if existing and production_capacity > len(existing):
 			return
+
+		if not existing and doctype == 'Job Card Time Log':
+			self.get_overlap_for(args, doctype='Job Card Schedule Time',
+				check_next_available_slot=check_next_available_slot)
 
 		return existing[0] if existing else None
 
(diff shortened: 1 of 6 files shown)


erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -49,9 +50,12 @@ def validate_time_logs(self):
 
 			self.total_completed_qty = flt(self.total_completed_qty, self.precision("total_completed_qty"))
 
-	def get_overlap_for(self, args, check_next_available_slot=False):
+	def get_overlap_for(self, args, doctype='', check_next_available_slot=False):
```
ankush:
```suggestion
	def get_overlap_for(self, args, check_next_available_slot=False, doctype=None):
```
New keyword arguments should be preferably added to the end to avoid breaking non-kwarg function calls.  Also `None` is a better default than an empty string.

ankush (changes requested):

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -67,14 +71,14 @@ def get_overlap_for(self, args, check_next_available_slot=False):
 			extra_cond = " or (%(from_time)s <= jctl.from_time and %(to_time)s <= jctl.to_time)"
 
 		existing = frappe.db.sql("""select jc.name as name, jctl.to_time from
```
ankush, line 73:
```suggestion
		existing = frappe.db.sql(f"""select jc.name as name, jctl.to_time from
```

It's quite difficult to follow `.format()` when replacement values are not in order. We can use f-strings now.

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -67,14 +71,14 @@ def get_overlap_for(self, args, check_next_available_slot=False):
...
-			`tabJob Card Time Log` jctl, `tabJob Card` jc where jctl.parent = jc.name and
+			`tab{2}` jctl, `tabJob Card` jc where jctl.parent = jc.name and
```
ankush, line 74:
```suggestion
			`tab{doctype}` jctl, `tabJob Card` jc where jctl.parent = jc.name and
```

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -67,14 +71,14 @@ def get_overlap_for(self, args, check_next_available_slot=False):
...
 			(
 				(%(from_time)s > jctl.from_time and %(from_time)s < jctl.to_time) or
 				(%(to_time)s > jctl.from_time and %(to_time)s < jctl.to_time) or
 				(%(from_time)s <= jctl.from_time and %(to_time)s >= jctl.to_time) {0}
```
ankush, line 78:
```suggestion
				(%(from_time)s <= jctl.from_time and %(to_time)s >= jctl.to_time) {extra_cond}
```

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -67,14 +71,14 @@ def get_overlap_for(self, args, check_next_available_slot=False):
...
 			)
 			and jctl.name != %(name)s and jc.name != %(parent)s and jc.docstatus < 2 {1}
```
ankush, line 80:
```suggestion
			and jctl.name != %(name)s and jc.name != %(parent)s and jc.docstatus < 2 {validate_overlap_for}
```

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -67,14 +71,14 @@ def get_overlap_for(self, args, check_next_available_slot=False):
...
-			order by jctl.to_time desc limit 1""".format(extra_cond, validate_overlap_for),
+			order by jctl.to_time desc limit 1""".format(extra_cond, validate_overlap_for, doctype),
```
ankush, line 81:
```suggestion
			order by jctl.to_time desc limit 1""",
```

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -100,6 +108,27 @@ def schedule_time_logs(self, row):
...
+					if row.get(field):
+						time_list.append(get_datetime(row.get(field)))
+
+			if time_list:
+				self.set(time_fields[0], min(time_list))
+				if time_fields[1] == 'actual_end_time' and not self.time_logs[-1].to_time:
+					self.set(time_fields[1], '')
+					return
+
+				self.set(time_fields[1], max(time_list))
```
ankush, line 130:
```suggestion
		for child_table, start_field, end_field in [('scheduled_time_logs', 'expected_start_time', 'expected_end_time'),
			('time_logs', 'actual_start_time', 'actual_end_time')]:
			if not self.get(child_table):
				continue

			time_list = []
			for row in self.get(child_table):
				for field in ['from_time', 'to_time']:
					if row.get(field):
						time_list.append(get_datetime(row.get(field)))

			if time_list:
				self.set(start_field, min(time_list))
				if end_field == 'actual_end_time' and not self.time_logs[-1].to_time:
					self.set(end_field, '')
					return

				self.set(end_field, max(time_list))
```

fields[0] and fields[1] can be reaplced with better variable names. Same with doctype -> child_table.

ankush (commented):
Functionally works as expected. Some changes can be done for improving readability.

guilmori:
I question the use case of having a child table for Scheduled Times.
It would make more sense to me having these fields right in the Job Card doctype: 

> Expected Start Time

> Expected End Time

> Duration

rohitwaghchaure:
@guilmori Yeah but when to complete one job required more than one day then we need to consider working hours. Also we couldn't consider the lunch time while doing the scheduling.

guilmori:
Will this get merged eventually ? 
The current behavior where the single list is treated as both Scheduled and Actual time makes no sense and is simply unusable.

guilmori:
Any updates ?

PR #25901, opened 2021-05-31
fix: Removed set_status after cancel from Expense Claim and Employee Advance

pateljannat:
### Issue
1. From Expense Claim and Employee Advance, the set_status function is called explicitly.
2. This is not needed as the Status gets updated to Cancelled even without calling these functions on cancel.

### DIFF
--- a/erpnext/hr/doctype/expense_claim/expense_claim.py
+++ b/erpnext/hr/doctype/expense_claim/expense_claim.py
@@ -44,10 +44,9 @@ class ExpenseClaim(AccountsController):
 
 		paid_amount = flt(self.total_amount_reimbursed) + flt(self.total_advance_amount)
 		precision = self.precision("grand_total")
-		if (self.is_paid or (flt(self.total_sanctioned_amount) > 0
-			and flt(self.grand_total, precision) ==  flt(paid_amount, precision))) \
-			and self.docstatus == 1 and self.approval_status == 'Approved':
-				status = "Paid"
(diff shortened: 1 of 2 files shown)


ankush:
Semgrep failure is not false positive, though it may seem so because `set_status` is also called from `validate`.

If it's required in `on_submit` use `self.db_set`.

erpnext/hr/doctype/expense_claim/expense_claim.py
```diff
@@ -36,22 +36,22 @@ def validate(self):
...
+		if (self.is_paid or (flt(self.total_sanctioned_amount) > 0 and self.docstatus == 1
+			and flt(self.grand_total, precision) == flt(paid_amount, precision))) and self.approval_status == 'Approved':
+			status = "Paid"
 		elif flt(self.total_sanctioned_amount) > 0 and self.docstatus == 1 and self.approval_status == 'Approved':
-			self.status = "Unpaid"
+			status = "Unpaid"
 		elif self.docstatus == 1 and self.approval_status == 'Rejected':
-			self.status = 'Rejected'
+			status = 'Rejected'
+		self.db_set("status", status)
```
ruchamahabal: Since `set_status` is also called on validate when changes are not committed to the db, we can use a flag `update=False` in the method to understand whether to use `db_set` or not. Pass True to `set_status` when called from `on_submit`. Keep it false by default.

ruchamahabal (approved):

ruchamahabal:
@pateljannat port this to `develop` too?

PR #25915, opened 2021-06-02
fix: not able to select the item code in work order

rohitwaghchaure:
**Issue**

While selecting Item To Manufacture in the work order, system throwing below error

Screenshot 2021-06-02 at 2 02 33 PM

### DIFF
--- a/erpnext/controllers/queries.py
+++ b/erpnext/controllers/queries.py
@@ -204,7 +204,7 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
 
 	if "description" in searchfields:
 		searchfields.remove("description")
-	
+
 	columns = ''
 	extra_searchfields = [field for field in searchfields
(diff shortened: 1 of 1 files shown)


nabinhait (approved):

erpnext/controllers/queries.py
```diff
@@ -216,9 +216,10 @@ def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=Fals
 		if not field in searchfields]
 	searchfields = " or ".join([field + " like %(txt)s" for field in searchfields])
 
-	if filters.get('supplier'):
-		item_group_list = frappe.get_all('Supplier Item Group', filters = {'supplier': filters.get('supplier')}, fields = ['item_group'])
-		
+	if filters and isinstance(filters, dict) and filters.get('supplier'):
```
barredterra, line 219: https://github.com/frappe/erpnext/pull/25683#discussion_r644081791

PR #25927, opened 2021-06-03
fix: to fetch the correct field in Tax Rule

Anuja-pawar:
Issue: Wrong field name `supplier_group_name` is fetched in the Tax Rule (instead of `supplier_group`)

ezgif com-gif-maker (1)


nextchamp-saqib (changes requested):
You can use Fetch From field to replace add_fetch and Mandatory Depends On to replace toggle_reqd

nextchamp-saqib (approved):

PR #25936, opened 2021-06-03
fix: staffing plan vacancies data type issue

pateljannat:
### Issue and steps to replicate.

1. Create a Staffing plan for a designation that has at least one employee already present.
2. In the Staffing Plan Details table enter that designation. Observe the Number of Positions auto-populate. It should be greater than 0 for this scenario.
3. Don't enter any vacancy. Leave the field blank.
4. Save. 

### Actual Result
1. An Error used to appear due to inconsistencies in data types.

Screenshot 2021-06-03 at 4 02 21 PM

###  Expected Result
1. It should save without any error.


erpnext/hr/doctype/staffing_plan/staffing_plan.py
```diff
@@ -41,7 +41,7 @@ def set_total_estimated_budget(self):
 
 			detail.total_estimated_cost = 0
 			if detail.number_of_positions > 0:
-				if detail.vacancies > 0 and detail.estimated_cost_per_position:
+				if detail.vacancies and detail.vacancies > 0 and detail.estimated_cost_per_position:
```
ruchamahabal:
```suggestion
				if detail.vacancies and detail.estimated_cost_per_position:
```

No need to check if it's explicitly greater than 0 since the first condition will only return `True` if it's not `None` and greater than 0.

ruchamahabal (dismissed):

erpnext/hr/doctype/staffing_plan/staffing_plan.js
```diff
@@ -103,4 +103,4 @@ var set_total_estimated_budget = function(frm) {
 		})
 		frm.set_value('total_estimated_budget', estimated_budget);
 	}
-}
\ No newline at end of file
+}
```
ruchamahabal:
```suggestion
};
```

ruchamahabal:
Unrelated test case failing. Merging

PR #25942, opened 2021-06-03
fix: filter type for item query

rohitwaghchaure:
**Issue**
Screenshot 2021-06-03 at 5 57 37 PM

### DIFF
--- a/erpnext/controllers/queries.py
+++ b/erpnext/controllers/queries.py
@@ -4,6 +4,7 @@
 from __future__ import unicode_literals
 import frappe
 import erpnext
+import json
 from frappe.desk.reportview import get_match_cond, get_filters_cond
 from frappe.utils import nowdate, getdate
 from collections import defaultdict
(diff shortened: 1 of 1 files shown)


barredterra:
Thanks for fixing, @rohitwaghchaure !

erpnext/controllers/queries.py
```diff
@@ -4,6 +4,8 @@
 from __future__ import unicode_literals
 import frappe
 import erpnext
+import json
+from six import string_types
```
ankush:
```suggestion
```

erpnext/controllers/queries.py
```diff
@@ -198,6 +200,9 @@ def get_accounts(with_account_type_filter):
 def item_query(doctype, txt, searchfield, start, page_len, filters, as_dict=False):
 	conditions = []
 
+	if isinstance(filters, string_types):
```
ankush:
```suggestion
	if isinstance(filters, str):
```

ankush (commented):
We can ignore py2 support now.

barredterra (approved):

PR #25944, opened 2021-06-03
fix: boarding status in Employee Onboarding and Separation

ruchamahabal:
**Steps to replicate:**

1. Create an Employee Onboarding.
2. Submit it. A project gets created automatically.
3. Change the status for Employee Onboarding manually to "Completed". Save.
4. Open the doc again. Status changes to Pending.

This is because Employee Onboarding fetches the status from the linked Project on load and overrides manually set status. The document is set as dirty so even though you did not make any changes on opening it you have to save it with the updated status.

The same thing happens with the Employee Separation doctype too.

https://user-images.githubusercontent.com/24353136/120643093-5c238380-c493-11eb-97bc-3cc4cc59e545.mp4

**Fix**:

1. Since boarding status is dependent on a project for both the doctypes, made it read-only.
2. Since it is dependent on the project, it should not be updated on load but as soon as the project gets updated. Added a function `update_employee_boarding_status` to update boarding status on updating the project. 
3. Update tests to check if boarding status is getting updated properly on project update.

https://user-images.githubusercontent.com/24353136/120645659-4a8fab00-c496-11eb-83ec-7ea621be5017.mp4


erpnext/hr/utils.py
```diff
@@ -114,16 +114,23 @@ def get_onboarding_details(parent, parenttype):
 		filters={"parent": parent, "parenttype": parenttype},
 		order_by= "idx")
 
-@frappe.whitelist()
-def get_boarding_status(project):
+def update_employee_boarding_status(project):
```
nabinhait: Why employee onboarding related functions and common class in hr/utils? Creating a separate file or putting in employee_onboarding.py will make more sense.

nabinhait (changes requested):

nabinhait (approved):

PR #25947, opened 2021-06-03
test(perf): eliminate repeat creation of HSN codes

ankush:
In tests, we have 5 Indian companies and all of them trigger the creation of HSN code. 

Adding a flag in tests to ignore repeat creation of HSN codes saves ~1.5-2 minutes. (in all test runners)

### DIFF
--- a/erpnext/regional/india/setup.py
+++ b/erpnext/regional/india/setup.py
@@ -27,6 +27,9 @@ def setup_company_independent_fixtures(patch=False):
 	add_print_formats()
 
 def add_hsn_sac_codes():
+	if frappe.flags.in_test and frappe.flags.created_hsn_codes:
+		return
+
 	# HSN codes
 	with open(os.path.join(os.path.dirname(__file__), 'hsn_code_data.json'), 'r') as f:
 		hsn_codes = json.loads(f.read())
@@ -37,6 +40,9 @@ def add_hsn_sac_codes():
 	with open(os.path.join(os.path.dirname(__file__), 'sac_code_data.json'), 'r') as f:
 		sac_codes = json.loads(f.read())
 	create_hsn_codes(sac_codes, code_field="sac_code")
+
+	if frappe.flags.in_test:
+		frappe.flags.created_hsn_codes = True
 
 def create_hsn_codes(data, code_field):
 	for d in data:


erpnext/regional/india/setup.py
```diff
@@ -27,6 +27,9 @@ def setup_company_independent_fixtures(patch=False):
 	add_print_formats()
 
 def add_hsn_sac_codes():
+	if frappe.flags.in_test and frappe.flags.created_hsn_codes:
```
sagarvora, line 30:
```suggestion
	if frappe.flags.created_hsn_codes:
```

ankush: I was just being extra-cautious, not touching that flag unless in_test.  😄  But I guess this is fine. It's only set in_test.

ankush: How about removing the `in_test` on L46 instead?

erpnext/regional/india/setup.py
```diff
@@ -38,6 +41,9 @@ def add_hsn_sac_codes():
 		sac_codes = json.loads(f.read())
 	create_hsn_codes(sac_codes, code_field="sac_code")
 
+	if frappe.flags.in_test:
+		frappe.flags.created_hsn_codes = True
```
ankush, line 45:
```suggestion
	frappe.flags.created_hsn_codes = True
```

sagarvora:
Actually, no. Why set the `created_hsn_codes` flag, when it's not being used? That is unless we want to skip creating HSN codes if already created in production environments as well.

ankush:
@sagarvora In that case, it's better to leave the `in_test` flag in both conditions, this behaviour (reading or writing the flag) is not intended in production.

ankush:
Patch test failure is unrelated, likely due to patch changes in https://github.com/frappe/frappe/pull/13419

ankush:
A better solution for production would be to first fetch all available HSN codes in a set using DB/ORM query and then only insert them if they are not in a set. `frappe.db.exists` isn't a good option because it will require ~14K additional SQL queries for first run too. 

We can also benefit from ORM function for bulk creation like this. E.g. Django's `bulk_create` handles bulk operations by chunking and reduces queries drastically (often just 1 massive query 😄 ) 
Ref: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#bulk-create

ankush:
Failure is unrelated to this (and appearing in another recent PR too)

PR #25967, opened 2021-06-06
fix: payroll entry employee detail issue

anupamvs:
**Issue-** While selecting employee  in Payroll Entry getting this error
image


erpnext/payroll/doctype/payroll_entry/payroll_entry.py
```diff
@@ -673,7 +673,7 @@ def employee_query(doctype, txt, searchfield, start, page_len, filters):
 	include_employees = []
 	emp_cond = ''
 	if filters.start_date and filters.end_date:
-		employee_list = get_employee_list(filters)
+		employee_list = get_employee_list(filters) or []
```
gavindsouza: Shouldn't `get_employee_list` always return a list given its name? Add a fallback to return an empty list there instead of here.

gavindsouza (changes requested):

gavindsouza:
@anupamvs Use more descriptive, less ambiguous PR titles, commit messages and descriptions 😅

gavindsouza (approved):

PR #25970, opened 2021-06-07
feat: Exchange rate difference handling in procurement cycle

GangaManoj:
### _Issue:_
When a Purchase Invoice is created from a Purchase Receipt(or vice versa), if the exchange rate is different at the time of creation of the latter, the GL entries created for the Stock Received But Not Billed account will not be balanced.

Screenshot 2021-04-29 at 4 15 18 AM

### _Fix:_
The amount for the _Stock Received But Not Billed_ account for the second transaction will be made equal to that of the first and the difference will be assigned to the _Exchange Gain/Loss_ account.

_Case 1:_ The Purchase Receipt is created first at a particular exchange rate, but the exchange rate at the time of creation of the Purchase Invoice is different.

<details>
<summary>Images</summary>

For the Purchase Receipt:

Screenshot 2021-04-29 at 4 32 30 AM

For the corresponding Purchase Invoice(created after the exchange rate was changed):

Screenshot 2021-04-29 at 4 35 09 AM

</details>

_Case 2:_ The Purchase Invoice is created first at a particular exchange rate, but the exchange rate at the time of creation of the Purchase Receipt is different.

<details>
<summary>Images</summary>

For the Purchase Invoice:

Screenshot 2021-06-07 at 7 51 08 AM

For the corresponding Purchase Receipt(created after the exchange rate was changed):

Screenshot 2021-06-07 at 7 47 50 AM

</details>

`no-docs`

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -517,7 +517,7 @@ class PurchaseInvoice(BuyingController):
 			if d.category in ('Valuation', 'Total and Valuation')
 			and flt(d.base_tax_amount_after_discount_amount)]
 
-		exchange_rate_map, net_rate_map = self.get_purchase_receipt_details()
+		exchange_rate_map, net_rate_map = get_pr_or_pi_details(self)
 
 		for item in self.get("items"):
 			if flt(item.base_net_amount):
@@ -716,24 +716,6 @@ class PurchaseInvoice(BuyingController):
 
 							self.negative_expense_to_be_booked += flt(item.item_tax_amount, \
 								item.precision("item_tax_amount"))
-
-	def get_purchase_receipt_details(self):
-		purchase_receipts = []
-		pr_items = []
-
-		for item in self.get('items'):
-			if item.get('purchase_receipt'):
-				purchase_receipts.append(item.purchase_receipt)
-			if item.get('pr_detail'):
-				pr_items.append(item.pr_detail)
-			
-		exchange_rate_map = frappe._dict(frappe.get_all('Purchase Receipt', filters={'name': ('in',
-			purchase_receipts)}, fields=['name', 'conversion_rate'], as_list=1))
-
-		net_rate_map = frappe._dict(frappe.get_all('Purchase Receipt Item', filters={'name': ('in',
-			pr_items)}, fields=['item_code', 'net_rate'], as_list=1))
-
-		return exchange_rate_map, net_rate_map
 
 	def get_asset_gl_entry(self, gl_entries):
 		arbnb_account = self.get_company_default("asset_received_but_not_billed")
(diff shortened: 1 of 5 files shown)


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -633,6 +633,29 @@ def make_item_gl_entries(self, gl_entries):
 								"project": item.project or self.project
 							}, account_currency, item=item))
 
+						# check if the exchange rate has changed
+						purchase_receipt_conversion_rate = frappe.db.get_value('Purchase Receipt', {'name': item.purchase_receipt}, ['conversion_rate'])
```
deepeshgarg007: Let's move this query out of the for loop and make a  function that returns a map will all the relevant details which we can use here

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -633,6 +633,29 @@ def make_item_gl_entries(self, gl_entries):
...
+						if purchase_receipt_conversion_rate and self.conversion_rate != purchase_receipt_conversion_rate:
```
deepeshgarg007: Along with conversion rate diff we will also have to check if the net rate is same for both the item in both the docs

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -633,6 +633,29 @@ def make_item_gl_entries(self, gl_entries):
...
+							discrepancy_caused_by_exchange_rate_difference = (item.qty * item.rate) * (purchase_receipt_conversion_rate - self.conversion_rate)
```
deepeshgarg007:
```suggestion
							discrepancy_caused_by_exchange_rate_difference = (item.qty * item.net_rate) * (purchase_receipt_conversion_rate - self.conversion_rate)
```

erpnext/accounts/doctype/purchase_invoice/test_purchase_invoice.py
```diff
@@ -230,6 +230,23 @@ def check_gle_for_pi(self, pi):
 			self.assertEqual(expected_values[gle.account][1], gle.debit)
 			self.assertEqual(expected_values[gle.account][2], gle.credit)
 
+	def test_purchase_invoice_with_exchange_rate_difference(self):
```
deepeshgarg007, line 233: Let's also test the GL Entry Amount too in the test

erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -286,6 +286,10 @@ def make_item_gl_entries(self, gl_entries, warehouse_account=None):
 
 					self.add_gl_entry(gl_entries, warehouse_account_name, d.cost_center, stock_value_diff, 0.0, remarks,
 						stock_rbnb, account_currency=warehouse_account_currency, item=d)
+					print("*"* 30)
```
deepeshgarg007: Remove print statements

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
GangaManoj: Done.

erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
GangaManoj: Done.

erpnext/accounts/doctype/purchase_invoice/test_purchase_invoice.py
GangaManoj: Done.

deepeshgarg007 (approved):

PR #25975, opened 2021-06-07
fix: Update leave allocation after submit

pateljannat:
1. Previously,  users could not modify the number of leaves once a Leave Allocation was submitted.
2. In scenarios where the user had made a mistake in the leaves allocated, there was no easy way to rectify it.
3. The only way was to cancel and amend the Leave Allocation, but this was not feasible if there were approved Leave applications for this allocation.
4. Now, the users can update the New Leaves Allocated field after submit as well.

Screenshot 2021-06-08 at 6 57 14 PM

5. Leave Ledger Entries will be created accordingly for the modified leaves.
6. If the user makes the number of new leaves allocated lesser than the leaves applications approved for that period, a validation message will be shown.

Screenshot 2021-06-08 at 6 56 58 PM


erpnext/hr/doctype/leave_allocation/leave_allocation.py
```diff
@@ -55,6 +56,40 @@ def on_cancel(self):
...
+		if self.has_value_changed("new_leaves_allocated"):
+			ledger_entries = frappe.get_all("Leave Ledger Entry",
+								filters={
+									"transaction_type": "Leave Allocation",
+									"transaction_name": self.name,
+									"employee": self.employee,
+									"company": self.company,
+									"leave_type": self.leave_type
+								},
+								fields=["leaves"])
```
gavindsouza: 1. You can use 'pluck' instead. 
2. Running `validate_against_leave_applications` before this could mean one less db call

erpnext/hr/doctype/leave_allocation/test_leave_allocation.py
```diff
@@ -164,6 +164,41 @@ def test_creation_of_leave_ledger_entry_on_submit(self):
...
+		leave_allocation = create_leave_allocation()
+		leave_allocation.submit()
+		self.assertTrue(leave_allocation.total_leaves_allocated, 15)
+		leave_allocation.new_leaves_allocated = 40
+		leave_allocation.update()
+		self.assertTrue(leave_allocation.total_leaves_allocated, 40)
+
+	def test_leave_subtraction_after_submit(self):
+		frappe.db.sql("delete from `tabLeave Allocation`")
+		frappe.db.sql("delete from `tabLeave Ledger Entry`")
```
gavindsouza, line 181: There's `frappe.db.delete`

erpnext/hr/doctype/leave_allocation/leave_allocation.py
```diff
@@ -55,6 +56,40 @@ def on_cancel(self):
...
+			for entry in ledger_entries:
+				total_existing_leaves += entry.leaves
+
+			leaves_to_be_added = self.new_leaves_allocated - total_existing_leaves
+			args = dict(
+				leaves=leaves_to_be_added,
+				from_date=self.from_date,
+				to_date= self.to_date,
+				is_carry_forward=0
+			)
```
gavindsouza: Inconsistent pattern of using literals and constructors. Maybe just stick to literals since they save you the effort of looking up if `dict` exists in locals() etc

erpnext/hr/doctype/leave_allocation/leave_allocation.py
```diff
@@ -55,6 +56,40 @@ def on_cancel(self):
 		if self.carry_forward:
 			self.set_carry_forwarded_leaves_in_previous_allocation(on_cancel=True)
 
+	def on_update_after_submit(self):
+		if self.has_value_changed("new_leaves_allocated"):
+			ledger_entries = frappe.get_all("Leave Ledger Entry",
```
gavindsouza: This whole block should be a separate method on its own. Something like `update_leave_ledger` perhaps?

gavindsouza (changes requested):

nabinhait (approved):

PR #25977, opened 2021-06-08
fix: validate Product Bundle for existing transactions before deletion

AfshanKhan:
### Problem
The user can delete **Product Bundles** even if a transaction exists against them.

### Steps to replicate
1. Create a **Product Bundle**.
2. Create a **Sales Invoice** with the **Item** as the **Product Bundle**.
3. Now delete the **Product Bundle**.

### Enhancement
Now there is a validation to check if any transaction exists for the product bundle and the following message is displayed to the user when they try deleting any **Product Bundle**.
image


erpnext/selling/doctype/product_bundle/product_bundle.py
```diff
@@ -18,6 +20,14 @@ def validate(self):
 		from erpnext.utilities.transaction_base import validate_uom_is_integer
 		validate_uom_is_integer(self, "uom", "qty")
 
+	def on_trash(self):
+		invoices = frappe.db.get_all("Sales Invoice Item", {"item_code": self.new_item_code, "docstatus": ["!=", 2]}, ["parent"])
```
nextchamp-saqib: The product bundle can be linked with any transactions, stock entry, purchase invoice. 

Check the `check_if_linked_document_exists` method from `item.py`

erpnext/selling/doctype/product_bundle/product_bundle.py
```diff
@@ -18,6 +20,14 @@ def validate(self):
...
+		invoice_links = []
+		for invoice in invoices:
+			invoice_links.append(get_link_to_form('Sales Invoice', invoice['parent']))
+		if len(invoice_links):
+			frappe.throw("This Product Bundle is linked with Sales Invoice: {0}. You will have to cancel these invoices in order to cancel this Product Bundle".format(", ".join(invoice_links)))
```
nextchamp-saqib:
```suggestion
			frappe.throw("This Product Bundle is linked with Sales Invoice: {0}. You will have to cancel these invoices in order to delete this Product Bundle".format(", ".join(invoice_links)), title=_("Not Allowed"))
```

nextchamp-saqib (changes requested):

deepeshgarg007 (commented):
We should also provide a disable checkbox similar to item master to disable any specific product bundle

erpnext/selling/doctype/product_bundle/product_bundle.py
```diff
@@ -18,6 +20,27 @@ def validate(self):
...
+			"Stock Entry", "Stock Reconciliation", "Sales Order", "Purchase Order", "Material Request"]
+
+		invoice_links = []
+		for doctype in linked_doctypes:
+			item_doctype = doctype + " Item"
+
+			if doctype == "Stock Entry":
+				item_doctype = doctype + " Detail"
+
+			invoices = frappe.db.get_all(item_doctype, {"item_code": self.new_item_code, "docstatus": ["!=", 2]}, ["parent"])
```
nextchamp-saqib:
```suggestion
			invoices = frappe.db.get_all(item_doctype, {"item_code": self.new_item_code, "docstatus": 1}, ["parent"])
```

nextchamp-saqib (changes requested):

PR #25978, opened 2021-06-08
fix: validate Product Bundle for existing transactions before deletion

AfshanKhan:
Backport of https://github.com/frappe/erpnext/pull/25977

### DIFF
--- a/erpnext/selling/doctype/product_bundle/product_bundle.py
+++ b/erpnext/selling/doctype/product_bundle/product_bundle.py
@@ -4,6 +4,8 @@
 from __future__ import unicode_literals
 import frappe
 
+from frappe.utils import get_link_to_form
+
 from frappe import _
 
 from frappe.model.document import Document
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/product_bundle/product_bundle.py
```diff
@@ -18,6 +20,27 @@ def validate(self):
...
+			"Stock Entry", "Stock Reconciliation", "Sales Order", "Purchase Order", "Material Request"]
+
+		invoice_links = []
+		for doctype in linked_doctypes:
+			item_doctype = doctype + " Item"
+
+			if doctype == "Stock Entry":
+				item_doctype = doctype + " Detail"
+
+			invoices = frappe.db.get_all(item_doctype, {"item_code": self.new_item_code, "docstatus": ["!=", 2]}, ["parent"])
```
AfshanKhan, line 34: invoices = frappe.db.get_all(item_doctype, {"item_code": self.new_item_code, "docstatus": 1}, ["parent"])

PR #25993, opened 2021-06-09
fix: timesheet amount issue

anupamvs:
**Issues**-
- Timesheet amount is not changed when a user changes currency in the Sales Invoice.
- Project reference(Project Name) is missing from the Sales Invoice Timesheet child table.
- The exchange rate is set to 0 by default in Timesheet.


erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -826,7 +835,8 @@ frappe.ui.form.on('Sales Invoice', {
 			'time_sheet': row.parent,
 			'billing_hours': row.billing_hours,
 			'billing_amount': flt(row.billing_amount) * flt(exchange_rate),
-			'timesheet_detail': row.name
+			'timesheet_detail': row.name,
+			'project_name': row.project_name.name
```
ruchamahabal:
```suggestion
			'project_name': row.project_name
```

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -447,6 +447,15 @@ erpnext.accounts.SalesInvoiceController = erpnext.selling.SellingController.exte
...
 		this.frm.refresh_field("base_paid_amount");
+	},
+
+	conversion_rate() {
+		this._super();
+		$.each(cur_frm.doc.timesheets, function(i, d) {
+			let row = frappe.get_doc(d.doctype, d.name)
+			frappe.model.set_value(d.doctype, d.name, 'billing_amount', flt(d.billing_amount)*flt(cur_frm.doc.conversion_rate))
+		});
+		calculate_total_billing_amount(cur_frm)
```
ruchamahabal: On changing currency, the amount item table and timesheets are different. Am I missing something?


https://user-images.githubusercontent.com/24353136/123066130-6818ba80-d42d-11eb-8584-4f92d65f241a.mp4

ruchamahabal (dismissed):

pateljannat (approved):

PR #25995, opened 2021-06-09
fix: do not translate issue priority

hasnain2808:
while auto-creating issues via email, the priority comes in English

If the priorities are translated and saved it is not able to find the priority and throws a link validation error


pateljannat:
@hasnain2808 have you raised a PR for v13 as well?

hasnain2808:
No,
Should backport after this gets merged na?

pateljannat:
Doesn't happen automatically on erpnext. You'll have to raise a separate PR

hasnain2808:
> Doesn't happen automatically on erpnext. You'll have to raise a separate PR

Yes once this gets merged

pateljannat (approved):

PR #25997, opened 2021-06-09
fix(Issue): Remove response_by and resolution_by if sla is removed

GangaManoj:
**_Issue:_**
The _Response By_ and _Resolution By_ fields continue to be visible even after SLA is removed.

_**Fix:**_
Make the display for the two fields depend on the _Service Level Agreement_ field.

<details>
<summary> Images </summary>

Before Removing SLA:

Screenshot 2021-06-10 at 3 29 55 AM

After Removing SLA:

Screenshot 2021-06-10 at 3 31 12 AM

</details>

### DIFF
--- a/erpnext/support/doctype/issue/issue.py
+++ b/erpnext/support/doctype/issue/issue.py
@@ -29,6 +29,9 @@ class Issue(Document):
 		self.update_status()
 		self.set_lead_contact(self.raised_by)
 
+		if not self.service_level_agreement:
+			self.reset_associated_fields()
+
 	def on_update(self):
 		# Add a communication in the issue timeline
 		if self.flags.create_communication and self.via_customer_portal:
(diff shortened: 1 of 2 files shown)


harshit-30:
@GangaManoj is it hiding or emptying the values set?

GangaManoj:
@harshit-30 I've reset the values as well now, thanks for pointing it out.

erpnext/support/doctype/issue/issue.py
```diff
@@ -29,6 +29,9 @@ def validate(self):
 		self.update_status()
 		self.set_lead_contact(self.raised_by)
 
+		if not self.service_level_agreement:
+			self.reset_associated_fields()
```
nextchamp-saqib:
```suggestion
			self.reset_sla_fields()
```

erpnext/support/doctype/issue/issue.py
```diff
@@ -54,6 +57,13 @@ def set_lead_contact(self, email_id):
 				self.company = frappe.db.get_value("Lead", self.lead, "company") or \
 					frappe.db.get_default("Company")
 
+	def reset_associated_fields(self):
```
nextchamp-saqib:
```suggestion
	def reset_sla_fields(self):
```

nextchamp-saqib (changes requested):

nextchamp-saqib:
@GangaManoj  Don't forget to raise one for `develop` too

ankush:
@GangaManoj @nextchamp-saqib ref: https://github.com/frappe/erpnext/pull/26291

PR #25998, opened 2021-06-10
fix(Opening Invoice Creation Tool): Display message on successful creation of opening invoices

GangaManoj:
_Issue:_
No message is displayed if creation of opening invoices is successful, if there are less than five invoices, causing people to press the button multiple times, thereby creating additional invoices.

_Fix:_
Display appropriate message on successful creation of invoices.

Screenshot 2021-06-10 at 5 58 58 AM


erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
```diff
@@ -213,6 +213,8 @@ def start_import(invoices):
 	if errors:
 		frappe.msgprint(_("You had {} errors while creating opening invoices. Check {} for more details")
 			.format(errors, "<a href='/app/List/Error Log' class='variant-click'>Error Log</a>"), indicator="red", title=_("Error Occured"))
+	if len(invoices) < 5:
```
deepeshgarg007, line 216: I guess the screen is already frozen while the invoices are being created. If the invoices are not created via a background job then the invoice names are returned to the client-side so this message can be printed there if invoice names are found

deepeshgarg007: I guess the screen is already frozen while the invoices are being created. If the invoices are not created via a background job then the invoice names are returned to the client-side so this message can be printed there if invoice names are found

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
GangaManoj: Done.

erpnext/accounts/doctype/opening_invoice_creation_tool/opening_invoice_creation_tool.py
```diff
@@ -216,7 +216,7 @@ def start_import(invoices):
 	return names
 
 def publish(index, total, doctype):
-	if total < 5: return
+	if total < 50: return
```
nextchamp-saqib:
```suggestion
	if total < 50:
		return
```

nextchamp-saqib (changes requested):

nextchamp-saqib:
@GangaManoj Port this to `version-13-hotfix`

PR #25999, opened 2021-06-10
fix: check for duplicate payment terms in Payment Term Template

AfshanKhan:
Included Payment Terms for comparison in a function that checks for a duplicate with other Payment Terms in Payment Term Template.


AfshanKhan:
@nextchamp-saqib this can be merged as the same was merged on version-13-hotfix?

PR #26000, opened 2021-06-10
fix: Only display GST card in Accounting Workspace if it's in India

GangaManoj:
_Issue:_
The GST card is visible in the Accounting Workspace even for sites that aren't based in India.

Screenshot 2021-06-10 at 12 39 03 PM

_Fix:_

- Set only_for to India for the GST card.
- Filter cards by the only_for field (https://github.com/frappe/frappe/pull/13463)


deepeshgarg007 (approved):

GangaManoj:
@Mergifyio backport develop

PR #26005, opened 2021-06-10
fix: syntax error in Stock Entry

walstanb:
**Resolves:**
https://discuss.erpnext.com/t/new-stock-entry-page-wont-load/76694
https://discuss.erpnext.com/t/some-pages-show-blank-form-fields/76711

image

PR #26008, opened 2021-06-10
fix: syntax error in Stock Entry

walstanb:
Backport of #26005


ankush:
I guess this was closed because the issue only affects `develop` (native classes)?

walstanb:
Yes.

> I guess this was closed because the issue only affects `develop` (native classes)?

PR #25878, opened 2021-05-29
feat: subcontract code refactor and enhancement

rohitwaghchaure:
Done code refactoring and introduced new class Subcontracting in new file subcontracting.py

Feature Enhancement

1. Provision to edit consumed qty in the Purchase Receipt
Screenshot 2021-05-29 at 3 44 29 PM

2. Provision to get non consumed components from the Supplier
Screenshot 2021-05-24 at 6 42 48 PM

**Added Subcontract Order Summary Report** 

image

Docs https://github.com/frappe/erpnext_documentation/pull/343

PR #26020, opened 2021-06-11
fix: Report Subcontracted Raw Materials to be Transferred

marination:
**Before:**
- Over complicated logic (since there was no proper linking between PO and Stock Entry back then)
- Shows pending qty as negative on over transferring
- Also showed all Purchase Orders even if they were fully received. (nothing left to transfer)
- If multiple rows of the same raw material exist in the PO, it counts transferred qty as sum of all the rows. The key here was "Raw Material Item Code" which is incorrect in case of multiple rows with the same item code but belonging to different Finished Goods.
Screenshot 2021-06-10 at 9 27 28 PM

**After:**
- Simplified logic
- Only show information from POs that are yet to be received and have pending qty to transfer
- Inaccurate pending qty in case of same Raw Material Item Code fixed
Screenshot 2021-06-10 at 9 23 24 PM

>Did not enhance this report too much. When https://github.com/frappe/erpnext/pull/25878 gets merged, it can deprecate this report easily.


marination:
PDF generation tests are failing in unrelated module tests. Can confirm no impact from this PR.

PR #26027, opened 2021-06-12
feat: add Inactive status to Employee

ruchamahabal:
Inactive status for Employee can be set when they are:
1. Seasonal Workers
2. Daily Wage Workers
3. Contract-based Employees

In these cases, the employees come and go. They are expected to join back on a periodic basis and they don't actually "Leave". In this period they don't get paid and are not applicable for promotion, transfer, or retention bonus(since they are not leaving).

**Documentation PR:** https://github.com/frappe/erpnext_documentation/pull/351


nabinhait (approved):

nirajregmi:
Even if I make employee "Inactive", the employee's data is reflecting in attendance as absent. If employee is inactive then its attendance report should not be shown in monthly attendance report but it is marked as absent in attendance report.

PR #26040, opened 2021-06-14
fix(item): Update custom button label to avoid conflict

deepeshgarg007:
image


ankush:
This should probably resolve it: https://github.com/frappe/frappe/pull/13420

PR #26044, opened 2021-06-14
feat(General Ledger): Implement multi-account selection

GangaManoj:
- Allow the user to select more than one Account at a time while generating a General Ledger report.

Screenshot 2021-06-14 at 1 25 23 PM

<details>
<summary> Testing Info </summary>

_Steps to re-create:_

- Create and submit invoices/journal entries. This will generate GL Entries associated with multiple accounts.
- Open General Ledger report.
- Try setting more than one value in the Account filter as shown in the image.

</details>

`no-docs`

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.js
+++ b/erpnext/accounts/report/general_ledger/general_ledger.js
@@ -39,7 +39,6 @@ frappe.query_reports["General Ledger"] = {
 			"fieldtype": "MultiSelectList",
 			"options": "Account",
 			get_data: function(txt) {
-				console.log("txt = ", txt)
 				return frappe.db.get_link_options('Account', txt, {
 					company: frappe.query_report.get_filter_value("company")
 				});
--- a/erpnext/accounts/report/general_ledger/general_ledger.js
+++ b/erpnext/accounts/report/general_ledger/general_ledger.js
@@ -36,16 +36,13 @@ frappe.query_reports["General Ledger"] = {
 		{
 			"fieldname":"account",
 			"label": __("Account"),
-			"fieldtype": "Link",
+			"fieldtype": "MultiSelectList",
 			"options": "Account",
-			"get_query": function() {
-				var company = frappe.query_report.get_filter_value('company');
-				return {
-					"doctype": "Account",
-					"filters": {
-						"company": company,
-					}
-				}
+			get_data: function(txt) {
+				console.log("txt = ", txt)
+				return frappe.db.get_link_options('Account', txt, {
+					company: frappe.query_report.get_filter_value("company")
+				});
 			}
 		},
 		{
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/general_ledger/general_ledger.js
```diff
@@ -36,16 +36,13 @@ frappe.query_reports["General Ledger"] = {
...
-			"get_query": function() {
-				var company = frappe.query_report.get_filter_value('company');
-				return {
-					"doctype": "Account",
-					"filters": {
-						"company": company,
-					}
-				}
+			get_data: function(txt) {
+				console.log("txt = ", txt)
```
deepeshgarg007: Remove console

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -87,7 +91,7 @@ def set_account_currency(filters):
 		account_currency = None
 
 		if filters.get("account"):
-			account_currency = get_account_currency(filters.account)
+			account_currency = get_account_currency(filters.account[0])
```
deepeshgarg007: Instead of simply taking the first account it should check if only one account is selected or all selected accounts have the same currency then apply this

erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -205,10 +209,18 @@ def get_gl_entries(filters, accounting_dimensions):
 
 def get_conditions(filters):
 	conditions = []
+
 	if filters.get("account") and not filters.get("include_dimensions"):
-		lft, rgt = frappe.db.get_value("Account", filters["account"], ["lft", "rgt"])
-		conditions.append("""account in (select name from tabAccount
-			where lft>=%s and rgt<=%s and docstatus<2)""" % (lft, rgt))
+		account_conditions = ""
```
deepeshgarg007: Instead of adding multiple "OR" conditions one SET of accounts and can be made and just one single condition for accounts should be passed

deepeshgarg007 (changes requested):

deepeshgarg007:
Also, this should be a `feat:` not a `fix:` as this PR adds a new functionality

erpnext/accounts/report/general_ledger/general_ledger.js
GangaManoj: Right, sorry. Done.

erpnext/accounts/report/general_ledger/general_ledger.py
GangaManoj: Okay, but what should we do if neither of those condition are met?

erpnext/accounts/report/general_ledger/general_ledger.py
GangaManoj: Could you give it a look now? It's not exactly what you're asking for, but I've reduced it to a single condition.

erpnext/accounts/report/general_ledger/general_ledger.py
deepeshgarg007: Nothing, just let the filters be same, no change

GangaManoj: Done.

PR #26046, opened 2021-06-14
fix: precision rate for packed items in internal transfers

noahjacob:
**Issue:**
image
- While trying to create an internal transfer in a delivery note with packed items, the above error was thrown.

**Reason:**
- There is no "rate" field for packed items.
- No condition to check if doctype is a _Packed Item_ or a _Delivery Note Item_.

**Fix:**
- Checks if the doctype is a packed item and tries to fetch the respective field's precision, i.e _"incoming_rate"_ for Packed Items.

### DIFF
--- a/erpnext/controllers/selling_controller.py
+++ b/erpnext/controllers/selling_controller.py
@@ -330,9 +330,15 @@ class SellingController(StockController):
 
 				# For internal transfers use incoming rate as the valuation rate
 				if self.is_internal_transfer():
-					rate = flt(d.incoming_rate * d.conversion_factor, d.precision('rate'))
-					if d.rate != rate:
-						d.rate = rate
+					if d.doctype == "Packed Item":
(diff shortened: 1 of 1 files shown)


erpnext/controllers/selling_controller.py
```diff
@@ -330,9 +330,15 @@ def set_incoming_rate(self):
...
-						d.rate = rate
+					rate_field = "rate"
+					if d.doctype == "Packed Item":
+						rate_field = "incoming_rate"
+						row_rate_field = d.incoming_rate
+					else:
+						row_rate_field = d.rate
+					rate = flt(d.incoming_rate * d.conversion_factor, d.precision(rate_field))
+					if row_rate_field != rate:
+						row_rate_field = rate
```
ankush: This is likely a bug. `d.rate` won't change like it used to before.

ankush (changes requested):

ankush (approved):

PR #26049, opened 2021-06-14
fix: escaped warehouse value for sql query

noahjacob:
**Issue:**
- Unescaped warehouse name values would throw a query syntax error if contained special characters like " **'** " while trying to create a stock ledger entry.(Eg. SOMEONE **'** S SUPERMARKET )

**Fix:**
- Escaped warehouse name fields for query.

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -558,11 +558,8 @@ def future_sle_exists(args):
 	or_conditions = []
 	for warehouse, items in warehouse_items_map.items():
 		or_conditions.append(
-			"warehouse = '{}' and item_code in ({})".format(
-				warehouse,
-				", ".join(frappe.db.escape(item) for item in items)
-			)
-		)
+			f"""warehouse = {frappe.db.escape(warehouse)}
+				and item_code in ({', '.join(frappe.db.escape(item) for item in items)})""")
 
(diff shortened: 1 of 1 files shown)


erpnext/controllers/stock_controller.py
```diff
@@ -558,8 +558,8 @@ def future_sle_exists(args):
 	or_conditions = []
 	for warehouse, items in warehouse_items_map.items():
 		or_conditions.append(
-			"warehouse = '{}' and item_code in ({})".format(
-				warehouse,
+			"warehouse = {} and item_code in ({})".format(
```
ankush, line 561: Replace this with f-string for readability?

erpnext/controllers/stock_controller.py
```diff
@@ -558,8 +558,8 @@ def future_sle_exists(args):
...
+				frappe.db.escape(warehouse),
 				", ".join(frappe.db.escape(item) for item in items)
 			)
```
ankush:
```suggestion
			f"""warehouse = {frappe.db.escape(warehouse)}
				and item_code in ({', '.join(frappe.db.escape(item) for item in items)})"""
```


F-strings would be better for readability here 🤔

ankush (approved):

ankush:
@noahjacob can you backport it to `develop` as well?

PR #26058, opened 2021-06-14
fix:Sales order qty update fails in "Update Items" button

TurkerTunali:
ERPNext generates "Cannot set quantity less than delivered quantity" error even the delivered qty is zero when user clicks "Update Items".
"date_unchanged" variable gets false value because of new_date is string. 

"getdate(new_date)" corrects the date comparison.

ERPNext_PR

> Screenshots/GIFs

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -1312,7 +1312,7 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
 			qty_unchanged = prev_qty == new_qty
 			uom_unchanged = prev_uom == new_uom
 			conversion_factor_unchanged = prev_con_fac == new_con_fac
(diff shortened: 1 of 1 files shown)


TurkerTunali:
Can we re-open this?

ankush:
Since it looks like you figured out a solution, can you open a PR?

TurkerTunali (commented):
Looks good.

TurkerTunali:
> Since it looks like you figured out a solution, can you open a PR?

Done.

ankush:
I am sorry I didn't notice this before. But we don't accept changes to stable branches directly (like `version-12`, `version-13`)

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#which-branch

TurkerTunali:
Branch corrected https://github.com/frappe/erpnext/pull/26992 .

PR #26073, opened 2021-06-16
fix: auto unlink warehouse from item on delete

18alantom:
### Description
On deletion of a warehouse, unlink warehouse from Item Default automatically, by deleting Item Default row.

Ref: FR-ISS-256317

### DIFF
--- a/erpnext/stock/doctype/warehouse/test_warehouse.py
+++ b/erpnext/stock/doctype/warehouse/test_warehouse.py
@@ -2,6 +2,7 @@
 # License: GNU General Public License v3. See license.txt
 from __future__ import unicode_literals
 
+import random
 import unittest
 
 import frappe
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -138,6 +139,15 @@ def convert_to_group(self):
...
 			return 1
 
+	def unlink_from_items(self):
+		item_default_names = frappe.db.get_list(
+			'Item Default',
+			pluck='name',
+			filters={'default_warehouse':self.name}
+		)
+		for name in item_default_names:
+			frappe.get_doc("Item Default", name).delete()
```
ankush: Use `frappe.delete_doc` instead of get_doc().delete()? `get_doc()` makes additional db calls to fetch data (that's about to be nuked anyway). If you look at implementation of `delete` method, it just calls `frappe.delete_doc`. 

https://frappeframework.com/docs/user/en/api/document#frappedelete_doc

ankush (changes requested):

erpnext/stock/doctype/warehouse/test_warehouse.py
```diff
@@ -146,4 +181,25 @@ def get_group_stock_account(company, company_abbr=None):
...
+	item_default = frappe.new_doc("Item Default")
+	item_default.default_warehouse = warehouse
+	item_default.parent = item_code
+	item_default.company = company
+
+	item_doc = frappe.get_doc('Item', item_code)
+	item_doc.item_defaults = [item_default]
+	item_doc.save()
+
+def create_item(item_code):
```
ankush: There are *two* similar functions in `test_item.py`, maybe you can just reuse them.

erpnext/stock/doctype/warehouse/test_warehouse.py
```diff
@@ -146,4 +181,25 @@ def get_group_stock_account(company, company_abbr=None):
 		if not company_abbr:
 			company_abbr = frappe.get_cached_value("Company", company, 'abbr')
 		group_stock_account = "Current Assets - " + company_abbr
-	return group_stock_account
\ No newline at end of file
+	return group_stock_account
+
+def assign_default_warehouses_to_item(item_code, default_warehouses, company):
+	warehouse = random.choice(default_warehouses)
```
ankush: Better to avoid random behaviour in tests wherever possible.

erpnext/stock/doctype/warehouse/test_warehouse.py
```diff
@@ -92,6 +93,40 @@ def test_warehouse_merging(self):
...
+			assign_default_warehouses_to_item(item, warehouse_ids, company)
+
+		# Delete warehouses
+		for warehouse in warehouse_ids:
+			frappe.delete_doc("Warehouse", warehouse)
+
+		# Check Item existance
+		for item in item_names:
+			item_exists = frappe.db.exists("Item", item)
+			self.assertTrue(item_exists, item)
```
ankush: This will become somewhat flaky because of how frappe.db.exists is defined. Maybe cast it to `bool` first? Also second parameter is a message that will be shown on failure. Just `item` isn't that useful.

erpnext/stock/doctype/warehouse/test_warehouse.py
```diff
@@ -92,6 +93,40 @@ def test_warehouse_merging(self):
...
+			if not item_exists:
+				continue
```
ankush:
```suggestion
```
This code is no-op. unittest will fail on first assertion failure.

ankush (changes requested):

ankush (approved):

ankush:
@18alantom please send a port to `develop` branch too.

PR #26077, opened 2021-06-16
fix: removed extra space from label rate

nemesis189:
There was an added space in the label "Rate" of Purchase Invoice Item which was causing problem with Data Import.


deepeshgarg007:
@nemesis189 Please send this PR to `version-13-hotfix` branch and  back port it to `version-12-hotfix` branch as well

PR #26081, opened 2021-06-16
fix: time out while submit / cancel the stock transactions with more than 50 Items

rohitwaghchaure:
**Issue**

Profiling output before fix

Screenshot 2021-06-16 at 5 51 52 PM

**After Fix**

Screenshot 2021-06-16 at 8 00 25 PM


rohitwaghchaure:
image

PR #26083, opened 2021-06-16
fix: Accounting Dimensions for payroll entry accrual Journal Entry

deepeshgarg007:
Accrual Journal Entry was not getting submitted from Payroll Entry due to missing mandatory accounting dimension.

image

image

This PR adds accounting dimensions to the accrual journal entry


this-gavagai:
I believe that this patch has lead to a regression. In my case, I am no longer able to submit Payroll Entries. I believe the issue is that I have Department listed as an accounting dimension, but the department field already exists as an employee filter. When I create a new payroll entry, Department isn't being listed as a Payroll Dimension.

PR #26088, opened 2021-06-17
refactor: Replace usage of render with get_response

surajshetty3416:
- Replace `render` imports with `get_response`
- Import `clear_cache` from `utils.py`

changes required for https://github.com/frappe/frappe/pull/12334


surajshetty3416:
All tests are passing with dependent code in Frappe.
Screenshot 2021-06-17 at 8 21 17 PM

PR #26092, opened 2021-06-17
fix: insufficient permission for Dunning error

Anuja-pawar:
**Issue**: 
When a user with no permission for Dunning tries to submit a Payment Entry,
**Insufficient Permission for Dunning** error is raised (even though that payment entry has nothing to do with dunning),
because `resolve_dunning()` is called from `Hooks.py` on submitting the Payment Entry.

Screenshot 2021-06-17 at 7 10 19 PM

Setting `ignore_permissions = True` for this.


nextchamp-saqib:
@Anuja-pawar Also raised one on `develop`

PR #25254, opened 2021-04-08
feat: Tax Detail Report

nameduser0:
Provides a generic Tax Detail report that expands Sales and Purchase Invoice details for detailed Sales and Goods Tax reports. But operates from GL Entries, so should be useful for generating any form of Tax report including those that need to include Journal Entry amendments. Has been tested as okay for Payroll reporting for Income Tax returns.

The report can be split into filtered sections whose total appears as a section subtotal and a summary total in v13. The report can then be saved as a Custom report that can be run later using any date filters required.

Requires frappe/frappe#12534
Part-addresses #21233
Documented at https://github.com/frappe/erpnext_documentation/pull/322  .
https://docs.erpnext.com/docs/user/manual/en/accounts/accounting-reports#3-taxes

TaxDetailGif

@surajshetty3416 @barredterra I would appreciate your reviews!

PR #26094, opened 2021-06-17
feat: Tax Detail Report

nameduser0:
Tax Detail Report: Backport of #25254
See also #25884
https://github.com/frappe/erpnext_documentation/pull/322


nameduser0:
@nabinhait @deepeshgarg007 Test failing due to #25884 which also needs to be backported

deepeshgarg007:
@casesolved-co-uk Please cherry-pick that commit and add to this PR. Editing from maintainers is not allowed

Andriesvn:
Nothing ever happened to getting this on v13?

nameduser0:
No, frankly it suits me that they can't be bothered. I've included it in our UK module available on our hosting service.

Andriesvn:
Sad to hear. 
Question though, How are you handling 0 value Tax detail such as 0% VAT or VAT Exempt values? Since your querying GL entries with this report and ERPNext removes 0 value GL Entries before posting them otherwise validation fails them as you can see here?
https://github.com/frappe/erpnext/blob/86ef9329797f558a4e8ec7f35a6527f86090482a/erpnext/accounts/doctype/gl_entry/gl_entry.py#L79

PR #26104, opened 2021-06-18
fix: disabled cancellation for sales order if linked to drafted sales inv

nemesis189:
Sales Order cannot be *cancelled if it is linked to a Sales Invoice in draft state


nemesis189:
added a test in the 3rd commit, tab/space error popped up so fixed it in the last commit

nemesis189:
added test and tab/space error fixes

PR #26107, opened 2021-06-18
fix: use Stripe's Price API for plan-price information

Alchez:
Stripe has reworked their API to create and manage subscriptions. Initially, we had to connect to the [Plans API](https://stripe.com/docs/api/plans) in order to manage plans (a combination of products and prices), but they've since reworked it to use the new [Price API](https://stripe.com/docs/api/prices).

The Plans API is still backwards-compatible, but using the new API is a trivial change. I've also changed the fieldname in Subscription Plan to reflect the terms used in the Price API instead.


Alchez:
I'm not sure why the patch is failing to fetch the old column, since I haven't reloaded the doctype fields in the patch. Any help?

PR #26119, opened 2021-06-20
feat: subcontract code refactor and enhancement

rohitwaghchaure:
Done code refactoring and introduced new class Subcontracting in new file subcontracting.py

Feature Enhancement

1. Provision to edit consumed qty in the Purchase Receipt
Screenshot 2021-05-29 at 3 44 29 PM

2. Provision to get non consumed components from the Supplier
Screenshot 2021-05-24 at 6 42 48 PM

**Added Subcontract Order Summary Report** 

image

Docs https://github.com/frappe/erpnext_documentation/pull/343

Backport https://github.com/frappe/erpnext/pull/25878


shashankshirke:
Hi guys,

When trying to create a Purchase Order for a service item it throws an error message saying:

"Row #1: Finished Good Item is not specified for service item [item-name]"

Screenshot 2022-07-26 at 7 08 43 PM

We already have a finished good created and it has an active BOM (with all the sub-assemblies and raw material items added).

My question here is, how does the system know that a particular service item is linked to a finished good item?

PR #26126, opened 2021-06-21
fix: allow to select group warehouse while downloading materials from production plan

rohitwaghchaure:
Screenshot 2021-06-21 at 2 22 26 PM


Ravn10:
Hi,
This does not work as expected ,
it downloads only stock of 'for_warehouse'  and not of the selected warehouse.
can it be fixed?

PR #26131, opened 2021-06-21
feat: fetching details from supplier/customer groups

noahjacob:
**Feature:**
- Added a button to fetch details from supplier/customer groups in their respective masters.

**Supplier**
supplier_group

**Customer**
customer_group
 
[**Docs**](https://github.com/frappe/erpnext_documentation/pull/367)


erpnext/buying/doctype/supplier/supplier.py
```diff
@@ -51,6 +51,23 @@ def validate(self):
...
+			for account in doc.accounts:
+				child = self.append('accounts')
+				child.company = account.company
+				child.account = account.account
+			self.save()
+
+		if not self.payment_terms and doc.payment_terms:
+			self.payment_terms = doc.payment_terms
+
+
```
marination: You're already resetting `self.payment_terms` and `self.accounts`, no need to check again if it's empty. Save once at the end.
```suggestion
		if doc.accounts:
			for account in doc.accounts:
				child = self.append('accounts')
				child.company = account.company
				child.account = account.account

		if doc.payment_terms:
			self.payment_terms = doc.payment_terms
        
        self.save()
```

erpnext/selling/doctype/customer/customer.py
```diff
@@ -78,6 +78,32 @@ def validate(self):
...
+				child.company = credit.company
+				child.credit_limit = credit.credit_limit
+			self.save()
+
+		if not self.payment_terms and doc.payment_terms:
+			self.payment_terms = doc.payment_terms
+
+		if not self.default_price_list and doc.default_price_list:
+			self.default_price_list = doc.default_price_list
+
```
marination: If there's repetition try to commonify it. Here again you dont have to check the fields that you reset.
```py

    tables = [["accounts", "account"], ["credit_limits", "credit_limit"]]
    fields = ["payment_terms", "default_price_list"]

    for row in tables:
        table, field = row[0], row[1]
        if not doc.get(table): continue

        for entry in doc.get(table):
            child = self.append(table)
            child.update({"company": entry.company, field: entry.get("field")})

    for field in fields:
        if not doc.get(field): continue
        self.update({field: doc.get(field)})

    self.save()
```

marination (approved):

PR #26132, opened 2021-06-21
fix: fetch preferred shipping address

noahjacob:
**Issue:**
- Preferred shipping address was not fetched when creating a purchase order.
- Only billing address was getting fetched properly.

image

image

address_fix

**After Fix:**
after_fix

### DIFF
--- a/erpnext/accounts/custom/address.py
+++ b/erpnext/accounts/custom/address.py
@@ -33,6 +33,8 @@ def get_shipping_address(company, address = None):
 	if address and frappe.db.get_value('Dynamic Link',
 		{'parent': address, 'link_name': company}):
 		filters.append(["Address", "name", "=", address])
+	if not address:
+		filters.append(["Address", "is_shipping_address", "=", 1])
 
 	address = frappe.get_all("Address", filters=filters, fields=fields) or {}
 
--- a/erpnext/public/js/utils/party.js
+++ b/erpnext/public/js/utils/party.js
@@ -274,9 +274,9 @@ erpnext.utils.validate_mandatory = function(frm, label, value, trigger_on) {
 	return true;
 }
 
-erpnext.utils.get_shipping_address = function(frm, callback){
+erpnext.utils.get_shipping_address = function(frm, callback) {
 	if (frm.doc.company) {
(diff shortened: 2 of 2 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -909,6 +909,18 @@ erpnext.TransactionController = class TransactionController extends erpnext.taxe
 				}
 			});
 
+			frappe.call({
```
deepeshgarg007: @noahjacob a method to fetch shipping address is already called on line no 896, can you check why that is not working and fix that instead of fetching shipping address again

Made a similar fix for purchase invoice. The company trigger itself was not getting called. Maybe this might help
https://github.com/frappe/erpnext/pull/26100

deepeshgarg007 (changes requested):

erpnext/public/js/controllers/transaction.js
noahjacob: image

@deepeshgarg007 it's not fetching because of line 279 (party.js). The condition can be moved elsewhere? Maybe in the callback on line 297?

deepeshgarg007: If this condition is the reason then it means that the Purchase Document is an internal transfer document. The address setting in this case is done on the server-side. Check this function in `sales_invoice.py`. Please check first why the address is not getting updated from here first

image

noahjacob: @deepeshgarg007  this is a not condition 
`if (!(frm.doc.inter_com_order_reference || frm.doc.internal_invoice_reference ||
frm.doc.internal_order_reference)) {`
So this checks if it is not an internal transfer right. I tested it with a non-internal transfer and it returns the callback.

deepeshgarg007: @noahjacob please remove the `!` then and let it return the callback only in case of internal docs.
Don't exactly remember the reason for adding the not condition.

Removing that will solve the issue, right?

noahjacob: Yes, that should solve the issue. I'll also rename inter_com_order_reference. The fieldname is inter_company_order_reference in the doctype.

deepeshgarg007 (approved):

PR #26133, opened 2021-06-21
feat: Optionally allow rejected quality inspection on submission

marination:
**Issue:**
 - On creating a QI against a Purchase Receipt, it is possible that this QI is rejected.
 - In this case, the submission of the Purchase Receipt is blocked due to a rejected QI.
 - It could be possible that after rejection the rejected qty is updated in the Purchase Receipt and the user wants to submit it.

**Solution:**
- Added **Action If Quality Inspection Is Rejected** in Stock Settings
   Screenshot 2021-06-21 at 4 23 28 PM
- If it is set to **Warn** it will show a toast message on submission, if not it will block the submission
   qi-reject-warn
- Cleaned up code to make it more readable

Docs: https://github.com/frappe/erpnext_documentation/pull/365


erpnext/controllers/stock_controller.py
```diff
@@ -356,42 +356,68 @@ def update_billing_percentage(self, update_modified=True):
...
+		action = frappe.get_doc('Stock Settings').action_if_quality_inspection_is_not_submitted or "Stop"
+		qa_docstatus = frappe.db.get_value("Quality Inspection", row.quality_inspection, "docstatus")
+
+		if not qa_docstatus == 1:
+			link = frappe.utils.get_link_to_form('Quality Inspection', row.quality_inspection)
+			msg = f"Row #{row.idx}: Quality Inspection {link} is not submitted for the item: {row.item_code}"
+			if action == "Stop":
+				frappe.throw(_(msg), title=_("Inspection Submission"), exc=QualityInspectionNotSubmittedError)
+			else:
+				frappe.msgprint(_(msg), alert=True)
```
marination:
```suggestion
				frappe.msgprint(_(msg), alert=True, indicator="orange")
```

PR #26142, opened 2021-06-22
fix: too many writes while renaming company abbreviation

nextchamp-saqib:
If a company has 20k+ warehouses or accounts, then replacing the abbreviation is never completed and throws error while being executed in a background job

### DIFF
--- a/erpnext/setup/doctype/company/company.py
+++ b/erpnext/setup/doctype/company/company.py
@@ -407,8 +407,6 @@ def replace_abbr(company, old, new):
 
 	frappe.only_for("System Manager")
 
-	frappe.db.set_value("Company", company, "abbr", new)
-
 	def _rename_record(doc):
 		parts = doc[0].rsplit(" - ", 1)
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/company/company.py
```diff
@@ -420,10 +420,12 @@ def _rename_records(dt):
 		for d in doc:
 			_rename_record(d)
 
+	frappe.db.auto_commit_on_many_writes = 1
 	for dt in ["Warehouse", "Account", "Cost Center", "Department",
 			"Sales Taxes and Charges Template", "Purchase Taxes and Charges Template"]:
 		_rename_records(dt)
 		frappe.db.commit()
+	frappe.db.auto_commit_on_many_writes = 0
```
ankush: Better to put this in `try:finally` block.

nextchamp-saqib: Yes, ofcourse!

PR #26145, opened 2021-06-22
fix: Fixed sql syntax error in get_project_name method

nemesis189:
Trailing 'AND' in the {{match_cond}} was causing a syntax error, so just changed its position


nextchamp-saqib:
@nemesis189 Also raise one for `version-13-hotfix`

PR #26150, opened 2021-06-22
fix: Changed profitability analysis report width

nemesis189:
Adjusted the width of Profitability Analysis Report to fit the screen and reduce white space

**Before**
image

**After**
image

### DIFF
--- a/erpnext/accounts/report/profitability_analysis/profitability_analysis.py
+++ b/erpnext/accounts/report/profitability_analysis/profitability_analysis.py
@@ -168,21 +168,24 @@ def get_columns(filters):
 			"label": _("Income"),
 			"fieldtype": "Currency",
 			"options": "currency",
-			"width": 120
+			"width": 305
+
 		},
 		{
 			"fieldname": "expense",
 			"label": _("Expense"),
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/profitability_analysis/profitability_analysis.py
```diff
@@ -168,21 +168,24 @@ def get_columns(filters):
 			"label": _("Income"),
 			"fieldtype": "Currency",
 			"options": "currency",
-			"width": 120
+			"width": 305
```
deepeshgarg007, line 171: @nemesis189 Can you add some screenshots of the report in description. Want to see how it looks now. Also why 305 and 307 and not some rounded whole no like 320 or 300?

nemesis189: adjusted the values for the table to fit the whole display, lesser values would leave a noticeable amount of space and looks quite asymmetric

deepeshgarg007 (approved):

deepeshgarg007:
@nemesis189 backport this to version-13-hotfix branch too

PR #26159, opened 2021-06-22
feat: Handle Asset on Issuing Credit Note

GangaManoj:
### Problems Addressed:

- Unable to create Credit Notes for Items that are Fixed Assets.
- The Asset isn't returned to its original status after issuing the Credit Note.
- The Asset linked with an Item isn't copied on creating a Credit Note- which could potentially raise an error.
- Fix GL Entry creation for Return Invoices linked with Assets.

<details>
<summary> Images</summary>

_Before the fix:_
- For Invoice:

Screenshot 2021-06-23 at 11 37 54 AM

- For Return Invoice:

Screenshot 2021-06-23 at 11 40 20 AM

_After the fix:_
- For Invoice:

Screenshot 2021-06-23 at 11 37 54 AM

- For Return Invoice:

Screenshot 2021-06-23 at 11 42 32 AM

</details>

- The message displayed when the Asset linked with an Item isn't specified in the Items table of Sales Invoice is too vague.

<details>
<summary> Images</summary>

_Old Message:_ 

Screenshot 2021-06-22 at 11 45 25 PM

_New Message:_

Screenshot 2021-06-23 at 11 52 39 AM

</details>


erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -917,7 +917,11 @@ def make_item_gl_entries(self, gl_entries):
 		for item in self.get("items"):
 			if flt(item.base_net_amount, item.precision("base_net_amount")):
 				if item.is_fixed_asset:
-					asset = frappe.get_doc("Asset", item.asset)
+					if item.get('asset'):
+						asset = frappe.get_doc("Asset", item.asset)
+					else:
+						frappe.throw(_("Enter Asset linked with Item {0}: {1} in row {2}.")
```
```suggestion
						frappe.throw(_("Row #{0}: You must select an Asset for Item {1}.", title=_("Missing Asset"))
```

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -917,7 +917,11 @@ def make_item_gl_entries(self, gl_entries):
...
+							.format(item.item_code, item.item_name, item.idx))
```
```suggestion
							.format(item.idx, item.item_name))
```

nextchamp-saqib (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -149,7 +149,7 @@ def validate_fixed_asset(self):
 					if self.update_stock:
 						frappe.throw(_("'Update Stock' cannot be checked for fixed asset sale"))
 
-					elif asset.status in ("Scrapped", "Cancelled", "Sold"):
+					elif asset.status in ("Scrapped", "Cancelled") or asset.status == "Sold" and not self.is_return:
```
```suggestion
					elif asset.status in ("Scrapped", "Cancelled") or (asset.status == "Sold" and not self.is_return):
```

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -917,22 +917,32 @@ def make_item_gl_entries(self, gl_entries):
...
+							item.base_net_amount, item.finance_book, True)
+					else:
+						fixed_asset_gl_entries = get_gl_entries_on_asset_movement(asset,
+							item.base_net_amount, item.finance_book)
 
 					for gle in fixed_asset_gl_entries:
 						gle["against"] = self.customer
 						gl_entries.append(self.get_gl_dict(gle, item=item))
 
 					asset.db_set("disposal_date", self.posting_date)
```
nextchamp-saqib: Disposal date needs to be reset, on credit note creation

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -13,7 +13,7 @@
 from erpnext.stock.doctype.delivery_note.delivery_note import update_billed_amount_based_on_so
 from erpnext.projects.doctype.timesheet.timesheet import get_projectwise_timesheet_data
 from erpnext.assets.doctype.asset.depreciation \
-	import get_disposal_account_and_cost_center, get_gl_entries_on_asset_disposal
+	import get_disposal_account_and_cost_center, get_gl_entries_on_asset_movement
```
nextchamp-saqib: `get_gl_entries_on_asset_movement` can be named `get_gl_entries_on_asset_sale`?

The problem with `get_gl_entries_on_asset_movement` is that, `asset_movement` is another part of Assets that don't involve `gl_entries`, so it can cause confusion.

You could rename the method, or use the same but introduce an argument like `is_return=True`

nextchamp-saqib (changes requested):

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
GangaManoj: But `get_gl_entries_on_asset_sale` doesn't make sense here and neither does the old name, because we're now considering two different cases here- asset disposal and asset regain. `get_gl_entries_on_asset_sale` would imply that only the former is being handled there. But I agree, `asset_movement` can be confusing, we should think of a better name.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
GangaManoj: Done.

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
nextchamp-saqib: `asset disposal and asset regain` sounds good

GangaManoj: Done.

PR #25782, opened 2021-05-21
feat: enhancements in Training Event

Anurag810:
**Documentation PR:** https://github.com/frappe/erpnext_documentation/pull/326

- [x] Docs
- [x] Add Multiple(set_query) no duplicate employee 
- [x] Attendance: remove from Mandatory / optional => Present / Absent
- [x] Is mandatory  checkbox

image

PR #26164, opened 2021-06-23
fix: Training Event

Anurag810:
front-port of following pull-request: #26071


ruchamahabal:
Since the commits from https://github.com/frappe/erpnext/pull/25782 were already front-ported during the last release, removed those commits from this PR. Only front porting missing commit.

PR #26166, opened 2021-06-23
fix: reposting  taking more than 45 mins to complete one record

rohitwaghchaure:
**Issue**
Time took to complete reposting for one record (2757 seconds = 45.95 mins)

Screenshot 2021-06-23 at 10 41 47 AM

**After Fix**
Screenshot 2021-06-23 at 11 29 06 AM


nabinhait (commented):
This seems overkill to me, can you try adding index on voucher_detail_no?

PR #26170, opened 2021-06-23
fix: Website item group logic for product listing in Item Group pages

marination:
**Issue:**
- The Website Item Groups table was not considered while fetching items for the Item Group Page
- So if Item Iphone 11 belongs to Item Group Products but has **Services** in the Website Item Groups table, it should be viisble in the **Services** web page and the **Products** web page
- Filters werent considering  Website Item Groups either
- Filters that were of fieldtype **Table Multiselect** broke the product query

**Fix:**
- Passed an argument to query engine to know when query is for item group page
- If for item group page, get data with regards to website item group table
- This query is fast since there's one filter and that shortens the table beforehand
- This data is merged with the results from the Item master (results only considering item attributes and field filters)
- The combined data is then sorted as per weightage
- **For filters:** added condition to consider Website Item Group, handled **Table Multiselect** fieldtype filters too

**Consider the foll.g items:**
Screenshot 2021-06-23 at 2 32 42 PM

**On the Item Group Page:**
- Items that have _Item Group as Services_ **OR** Items with _Website Item Group as Services_ must be visible
  Screenshot 2021-06-23 at 4 04 38 PM
- Similarly, DTH plan that belogns to _Services_ has Website Item Group as _Products_ and is visible on **Products**
  Screenshot 2021-06-23 at 4 06 57 PM

### DIFF
--- a/erpnext/portal/product_configurator/test_product_configurator.py
+++ b/erpnext/portal/product_configurator/test_product_configurator.py
@@ -43,6 +43,30 @@ class TestProductConfigurator(unittest.TestCase):
 				"show_variant_in_website": 1
 			}).insert()
 
+	def create_regular_web_item(self, name, item_group=None):
+		if not frappe.db.exists('Item', name):
+			doc = frappe.get_doc({
+				"description": name,
+				"item_code": name,
+				"item_name": name,
+				"doctype": "Item",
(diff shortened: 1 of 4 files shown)


marination:

gavindsouza (approved):

PR #26173, opened 2021-06-23
fix: incorrect response by variance & resolution by variance

nextchamp-saqib:
Problem: 

For DocTypes having SLA defined, the `reponse_by_variance` was re-calculated on `validate` from `hooks.py` even after the first response is sent.
The calculation made the `reponse_by_variance` negative which meant that SLA is failed even though the response was sent on time.

The same has been fixed for `resolution_by_variance` too.


erpnext/patches/v13_0/update_response_by_variance.py
```diff
@@ -0,0 +1,31 @@
...
+
+def execute():
+	if frappe.db.exists('DocType', 'Issue') and frappe.db.count('Issue'):
+		invalid_issues = frappe.get_all('Issue', {
+			'first_responded_on': ['is', 'set'],
+			'response_by_variance': ['<', 0]
+		}, ["name", "response_by_variance", "timestampdiff(Second, `first_responded_on`, `response_by`) as variance"])
+
+		# issues which has response_by_variance set as -ve
+		# but didn't diff between first_responded_on & response_by is +ve i.e SLA isn't failed
```
nextchamp-saqib:
```suggestion
		# but diff between first_responded_on & response_by is +ve i.e SLA isn't failed
```

erpnext/patches/v13_0/update_response_by_variance.py
```diff
@@ -0,0 +1,31 @@
...
+		for issue in invalid_issues:
+			frappe.db.set_value('Issue', issue.get('name'), 'response_by_variance', issue.get('variance'), update_modified=False)
+
+		invalid_issues = frappe.get_all('Issue', {
+			'resolution_date': ['is', 'set'],
+			'resolution_by_variance': ['<', 0]
+		}, ["name", "resolution_by_variance", "timestampdiff(Second, `resolution_date`, `resolution_by`) as variance"])
+
+		# issues which has resolution_by_variance set as -ve
+		# but didn't diff between resolution_date & resolution_by is +ve i.e SLA isn't failed
```
nextchamp-saqib:
```suggestion
		# but diff between resolution_date & resolution_by is +ve i.e SLA isn't failed
```

PR #26176, opened 2021-06-23
fix: User is not able to change item tax template

deepeshgarg007:
Due to introduction of net rate based item taxes in the following PR users were unable to change Item Tax Templates manually
https://github.com/frappe/erpnext/pull/25961

This PR fixes that issue

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -436,15 +436,15 @@ def get_barcode_data(items_list):
 	return itemwise_barcode
 
 @frappe.whitelist()
-def get_item_tax_info(company, tax_category, item_codes, item_tax_templates=None, item_rates=None):
+def get_item_tax_info(company, tax_category, item_codes, item_rates=None, item_tax_templates=None):
 	out = {}
-	if isinstance(item_codes, string_types):
+	if isinstance(item_codes, (str,)):
 		item_codes = json.loads(item_codes)
 
-	if isinstance(item_rates, string_types):
+	if isinstance(item_rates, (str,)):
 		item_rates = json.loads(item_rates)
 
-	if isinstance(item_tax_templates, string_types):
+	if isinstance(item_tax_templates, (str,)):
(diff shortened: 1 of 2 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -436,20 +436,28 @@ def get_barcode_data(items_list):
 	return itemwise_barcode
 
 @frappe.whitelist()
-def get_item_tax_info(company, tax_category, item_codes, item_rates=None):
+def get_item_tax_info(company, tax_category, item_codes, item_tax_templates=None, item_rates=None):
```
gavindsouza: Refrain from adding new keyword arguments in the middle. This could break apps that use this API with positional args instead.

erpnext/stock/get_item_details.py
```diff
@@ -436,20 +436,28 @@ def get_barcode_data(items_list):
...
 	out = {}
 	if isinstance(item_codes, string_types):
 		item_codes = json.loads(item_codes)
 
 	if isinstance(item_rates, string_types):
 		item_rates = json.loads(item_rates)
 
+	if isinstance(item_tax_templates, string_types):
```
gavindsouza: `string_types` for `PY3` is defined as `(str, )`. Now that we don't support `PY2`, we can get rid of `six` completely. Don't add more `six` code now; it'll make maintainability hard for us.

erpnext/stock/get_item_details.py
```diff
@@ -508,7 +514,8 @@ def _get_item_tax_template(args, taxes, out=None, for_validate=False):
 		return None
 
 	# do not change if already a valid template
-	if args.get('item_tax_template') in taxes:
+	if args.get('item_tax_template') in [t.item_tax_template for t in taxes]:
```
gavindsouza: Checks for memberships are better done over sets instead. 
```suggestion
	if args.get('item_tax_template') in {t.item_tax_template for t in taxes}:
```

ref: https://stackoverflow.com/a/2831242/10309266

PR #26187, opened 2021-06-24
fix: Address Card issues in e-commerce

marination:
**Issue:**
- Add Address modal in Shopping Cart, has a country field that kept auto clearing anything added to it. The styles are also missing for the same
- Hover styles on **Add Address** and **Change** buttons in address card
  hover-btn-add-address

**Fix:**
- Country field retains data. Its style will be fixed in frappe.
  country-field-modal
- Hover state on address and change button
  Screenshot 2021-06-24 at 11 58 50 AM


marination:

PR #26191, opened 2021-06-24
fix: update leave allocation after submit

pateljannat:
Backport of #25975


ruchamahabal:
https://github.com/frappe/erpnext/pull/26191/checks?check_run_id=2902692376

Leave Allocation test case failing

PR #26192, opened 2021-06-24
fix(Work Order): added freeze when trying to stop work order

noahjacob:
**Issue:**
- Status of the _Work Order_ would sometimes not get updated and throw an error while trying to do a db_set.(Lock Wait Timeout)

**Potential Fix:**
- Added freeze parameter to prevent multiple calls.

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.js
+++ b/erpnext/manufacturing/doctype/work_order/work_order.js
@@ -705,6 +705,7 @@ erpnext.work_order = {
 		frappe.call({
 			method: "erpnext.manufacturing.doctype.work_order.work_order.stop_unstop",
 			freeze: true,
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -704,6 +704,7 @@ erpnext.work_order = {
 	stop_work_order: function(frm, status) {
 		frappe.call({
 			method: "erpnext.manufacturing.doctype.work_order.work_order.stop_unstop",
+			freeze: true,
```
ankush, line 707:
```suggestion
			freeze: true,
			freeze_message: __("Updating Work Order status"),
```

ankush (approved):

PR #26194, opened 2021-06-24
fix: Website item group logic for product listing in Item Group pages

marination:
Port of https://github.com/frappe/erpnext/pull/26170


marination:

PR #26219, opened 2021-06-25
fix: list index out of range on using include UOM in Stock Projected Qty report

noahjacob:
**Issue:**
- Throws index error when trying to generate a report after setting the _Include UOM_ filter in the **Stock Projected Qty** Report.
image

**Fix:**
- Sets the default conversion factor to 1 if None or not present.


ankush:
Note: This also fixes another error where conversion factor was getting picked from previous row for some reason.

Example:
First 3 conversion factors should be same, instead last 3 are same  because of the bug.
Screenshot 2021-06-25 at 2 33 12 PM

ankush:
@noahjacob please backport to `version-13-hotfix` and `version-12-hotfix`

PR #26225, opened 2021-06-26
perf: reduce number of queries to validate selling price

pruthvi145:
**~58% reduction in `SellingController.validate_selling_price`**

## Changes Made
 - Optimize the query call from linear to constant.
 - Use table `tabBin` instead of `tabStock Ledger Entry` for fetching `last_valuation_rate`.
 - Format the error message.

### Before

Screenshot-2021-06-28-114423

### After

Screenshot-2021-06-28-114519


erpnext/controllers/selling_controller.py
```diff
@@ -169,39 +168,96 @@ def set_qty_as_per_stock_uom(self):
...
+				idx,
+				bold(item_name),
+				bold(ref_rate_field),
+				bold("net rate"),
+				bold(rate),
+				get_link_to_form("Selling Settings", "Selling Settings"),
+			), title=_("Invalid Selling Price"))
+
+		if (
+			not frappe.db.get_single_value("Selling Settings", "validate_selling_price")
```
ankush: Re-order conditional so DB call is last. Because of its short-circuiting nature, it won't get executed if previous checks can conclude.

erpnext/controllers/selling_controller.py
```diff
@@ -169,39 +168,96 @@ def set_qty_as_per_stock_uom(self):
...
+					item.idx,
+					item.item_name,
+					last_purchase_rate_in_sales_uom,
+					"last purchase rate"
+				)
+
+			if is_internal_customer or not is_stock_item:
+				continue
+
+			valuation_rate_map[(item.item_code, item.warehouse)] = None
```
ankush, line 215: This "initialization" can probably be avoided using `defaultdict`, making it more readable.

erpnext/controllers/selling_controller.py
```diff
@@ -169,39 +168,96 @@ def set_qty_as_per_stock_uom(self):
...
+				item_code, warehouse, valuation_rate
+			from
+				`tabBin`
+			where
+				({" or ".join(or_conditions)})
+				and valuation_rate > 0
+		""")
+
+		for rate in valuation_rates:
+			valuation_rate_map[rate[:2]] = rate[2]
```
ankush: Probably better to use `as_dict` instead of arbitrary indexing in results.

erpnext/controllers/selling_controller.py
sagarvora: `defaultdict` isn't suitable for this purpose as it just gives a default value if a key is not found. But here we are setting the key, using the dictionary without values to generate the query, and then populating the value after the query runs.

The alternative is creating a set for generating the query and using the output of the query to generate another object as a dict.

PR #26231, opened 2021-06-28
fix: Do not consider cancelled entries in party dashboard

deepeshgarg007:
Customer Dashboard was showing incorrect values due to consideration of cancelled entries


manasan3010:
@deepeshgarg007 Given that reverse entries are created for cancelled documents, party's unpaid amount should have balanced itself correctly.

Is this just a performance fix or bug fix?

PR #26232, opened 2021-06-28
fix: Webform Permission for custom doctype

shariquerik:
Problem Statement
Supplier portal webforms for custom doctype are visible to all suppliers instead of specific to each supplier.

Example:
Custom Doctype `Order Assignment` is linked to `Purchase Order` via a Link field.
image

A Webform `SO Schedule` is created which displays the `Order Assignment`, but a supplier can see all the `Order Assignments`, whereas he/she should only be able to see `Order Assignment` for which he/she has access to. 
Before:
image

After:
image

Docs: https://github.com/frappe/frappe_docs/pull/173


shariquerik:
Related PR: https://github.com/frappe/frappe/pull/13594

surajshetty3416:
@shariquerik Please fix warnings raised by linter.

erpnext/tests/test_webform.py
```diff
@@ -0,0 +1,137 @@
...
+		new_contact.append('links', {
+			"link_doctype": "Supplier",
+			"link_name": supplier.name
+		})
+		new_contact.append('email_ids', {
+			"email_id": contact_email,
+			"is_primary": 1
+		})
+
+		new_contact.insert(ignore_mandatory=True)
```
18alantom, line 75: Maybe this can be moved into a separate function, or rename the function perhaps?

shariquerik: Renamed it

18alantom:
PR is good to go; failing Cypress is unrelated.

shariquerik:

Issue #25965, opened 2021-06-05
Appointment Booking link broken


chrisdent:
Logged in as a Supplier, on the menu on the left "Appointment Booking" takes to  /book_appointment it appears a 404 Error.

PR #26233, opened 2021-06-28
fix: Added a message to enable appointment booking if disabled

mohammedyusufshaikh:
Initially If the appointment scheduling was disabled it use to route to 404 page but after the fix it shows a message to the user to enable the appointment booking scheduling from appointment booking settings 
Screenshot 2021-06-28 at 2 26 48 PM

Resolves Issue #25965


erpnext/www/book_appointment/index.py
```diff
@@ -14,7 +14,8 @@ def get_context(context):
 	if is_enabled:
 		return context
 	else:
-		frappe.local.flags.redirect_location = '/404'
+		frappe.redirect_to_message("Appointment Scheduling Disabled","Enable appointment booking from appointment booking settings"
+									,http_status_code=302,indicator_color="red")
```
ruchamahabal:
```suggestion
		frappe.redirect_to_message(_("Appointment Scheduling Disabled"), _("Appointment Scheduling has been disabled for this site")
									,http_status_code=302,indicator_color="red")
```

1. This message is meant for a user who is booking an appointment from the portal. They aren't aware that there is a doctype named **Appointment Booking Settings** in the system. This action is not expected from them so we just need to tell them that it's disabled.
2. Always make messages and title translatable.

erpnext/www/book_appointment/index.py
```diff
@@ -14,7 +14,8 @@ def get_context(context):
 	if is_enabled:
 		return context
 	else:
-		frappe.local.flags.redirect_location = '/404'
+		frappe.redirect_to_message(_("Appointment Scheduling Disabled"), _("Appointment Scheduling has been disabled for this site")
```
ruchamahabal: Sider points to undefined `_`

You will have to add an import for the translation string
`from frappe import _`

ruchamahabal (approved):

ruchamahabal:
Congratulations on your first contribution 🥳

ruchamahabal:
@mohammedyusufshaikh please backport the same for `version-13-hotfix` branch too

mohammedyusufshaikh:
ok

chrisdent:
If appointment scheduling is disabled the link for appointment booking in the portal for customer/supplier shouldn't appear? Just as a better experience for customer/supplier?

ruchamahabal:
@chrisdent Yes, that's always better but currently the standard portal menu items are hardcoded in `hooks.py` hence we cannot hide/show it programmatically. 

https://github.com/frappe/erpnext/blob/2af6ea68ae94fed5f363750b837eaa8a1e877e49/erpnext/hooks.py#L189

The framework allows you to create your own website sidebar to override this.

Issue #4087, opened 2015-09-29
(Suggestion) Organization Chart


Ferohers:
My suggestion is putting an organization chart under HR module. HR manually draws organization chart and updates it all the time. Time to time managers (plus consultants such as, KPMG) require it to look at bigger picture. I actually made a research and checked other open hr systems if I could bring something, after my research found out that it is above my skills.

Thank You!

Ref another open source project which has a basic org. chart system: http://sourceforge.net/projects/sentrifugo/

jwrober:
This would be a great enhancement to HR module.

Issue #18819, opened 2019-08-22
Organizational Chart Page


Tropicalrambler:
Create a page style view where users can visualize an automatically-generated organizational chart, with a better visualization than the standard tree view.

This organizational chart can allow changing positions in the chart.
Each employee box allows connection to a card-like pop-up or dialog with further information as per ERPNext held-data. 

User can download information as vcard file, to import directly into Android, iPhone, or other contact management system.  

Links take users to their social media profiles for work or personal (if listed) (Twitter, LinkedIn, Etc.)

idolaru:
Hands up here. A very good idea.

kennethsequeira:
@Tropicalrambler can you post any visual references we can look at since there are no standardized ways of making an org chart.

xickomesquita:
If @Tropicalrambler don't mind:

![image](https://user-images.githubusercontent.com/38265198/74953286-4003ec00-53e0-11ea-8c14-961a1d991b95.png)

![image](https://user-images.githubusercontent.com/38265198/74953315-47c39080-53e0-11ea-812c-b6bd9fdb7d41.png)

![image](https://user-images.githubusercontent.com/38265198/74953359-57db7000-53e0-11ea-86b1-3954cd011ac6.png)

![image](https://user-images.githubusercontent.com/38265198/74953410-63c73200-53e0-11ea-9ef2-ea4eb614b3d4.png)

saarthakgupta:
Hello All,

Just wanted to check if this is something which is available now ? this is a very useful and exciting feature which will help small orgs immensely.

Thank you in advance for your time!

PR #26261, opened 2021-06-29
feat: Organizational Chart

ruchamahabal:
## Organizational Charts

An organizational chart shows the organizational structure by depicting connections between different employees with their designation, image, and name. In large organizations where it's difficult to remember names, people can interact with the org chart and know the hierarchy through faces and other info. The **Edit** button navigates to the employee master.

### Interactions:

https://user-images.githubusercontent.com/24353136/124869278-975b2a00-dfde-11eb-8c91-d067e6ce94b9.mp4

### Expand All and Export

1. Option to expand all the nodes. Interaction is disabled until you collapse all.
2. Export to PNG

https://user-images.githubusercontent.com/24353136/126934843-9d1b100d-054c-4a3e-8175-752d832be0cd.mp4

## Mobile View

image

Untitled 2

After clicking on sibling group:

node-group

Documentation PR: https://github.com/frappe/erpnext_documentation/pull/364

ToDo

- [x] Optimise rendering
- [x] Handle multiple root nodes / orphan nodes
- [x] Expand all nodes
- [x] Export Chart
- [x] Tests
- [x] Documentation


olamide2:
This is Great

Will we be able to change relationship by using  drag and drop from the chart ?

ruchamahabal:
> This is Great
> 
> Will we be able to change relationship by using drag and drop from the chart ?

Thanks. No this is just for visualising the hierarchy based on the "reports to" field and not building the employee tree.

aschl:
Closes https://github.com/frappe/erpnext/issues/4087 and https://github.com/frappe/erpnext/issues/18819

.github/helper/install.sh
```diff
@@ -42,5 +42,5 @@ sed -i 's/socketio:/# socketio:/g' Procfile
 sed -i 's/redis_socketio:/# redis_socketio:/g' Procfile
 
 bench get-app erpnext "${GITHUB_WORKSPACE}"
-bench start &
+bench start &> bench_run_logs.txt &
```
nabinhait, line 45: Should it be `bench start > bench_run_logs.txt &`

nabinhait (changes requested):

.github/helper/install.sh
ankush: I added this to log the bench console which can be shown in case of some failure. 

`&>` is for capturing both stdout and stderr.

nabinhait (approved):

PR #26272, opened 2021-06-30
fix: moving campaign from selling to CRM

anupamvs:
Changing Campaign module Selling -> CRM


ruchamahabal:
For the reviewer:

1. Check related imports
2. Check migration and installation on a new site
3. Check if a patch is required
4. Check if Desk Page updated

erpnext/crm/doctype/campaign/campaign.py
```diff
@@ -1,11 +1,8 @@
-# Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
-# License: GNU General Public License v3. See license.txt
+# Copyright (c) 2021, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
 
-from __future__ import unicode_literals
 import frappe
-
 from frappe.model.document import Document
-from frappe.model.naming import set_name_by_naming_series
```
pateljannat, line 8: Import for set_name_by_naming_series has been removed but its still being called from autoname function

erpnext/crm/doctype/campaign/campaign.js
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2021, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Campaign', {
+	refresh: function(frm) {
+		erpnext.toggle_naming_series();
+
+		if(frm.doc.__islocal) {
+			frm.toggle_display("naming_series", frappe.boot.sysdefaults.campaign_naming_by=="Naming Series");
```
pateljannat, line 9: The field campaign_naming_by is still in the Selling settings. Should we create a CRM setting and move it there?

pateljannat (changes requested):

erpnext/crm/doctype/campaign/campaign.js
anupamvs: Yeah! CRM setting is not introduced yet. In my roadmap maybe in next month.

pateljannat (approved):

Issue #13355, opened 2018-03-21
Asset mandatory for fixed Asset Item in Purchase Invoice


PawanMeh:
1. Currently if an item is marked as an Asset Item, the purchase invoice requires you to create an asset and submit it also. This is not practical as "in use date" of asset and many other attributes of an asset are not known when the Purchase Invoice is received but are known later on once the asset arrives in the premises.

2. As a interim solution, I suggest removing mandatory submission of Asset for purchase invoice for asset item and allowing to make asset in draft and writing validations accordingly.

Of course, the ideal flow of this should be from purchase receipt but that would involve a lot of design changes.

PR #26273, opened 2021-06-30
fix: item stock levels section diappears on refresh

walstanb:
- Moved `make_dashboard` to end of `refresh` since `refresh_sections` is done while refreshing fields in [#13355](https://github.com/frappe/frappe/pull/13355).
- Fixes #25506.

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -93,10 +93,6 @@ frappe.ui.form.on("Item", {
 
 		erpnext.item.edit_prices_button(frm);
 		erpnext.item.toggle_attributes(frm);
-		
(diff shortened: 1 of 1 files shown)


ankush:
I don't think this will fix it. 

refer this comment: https://github.com/frappe/frappe/pull/13355#issuecomment-867432287

I'd suggest disabling developer mode while testing.

sagarvora:
@ankush  I agree

sagarvora:
The root issue causing the stock levels section to go away when `refresh_sections` is called needs fixing.

ankush:
@sagarvora @walstanb can you verify this one? https://github.com/frappe/erpnext/pull/26305

PR #26274, opened 2021-06-30
fix(ux): hide irrelevant fields for asset items

ankush:
Before:
image

After: 

image


erpnext/stock/doctype/item/item.js
```diff
@@ -142,8 +142,26 @@ frappe.ui.form.on("Item", {
...
+				'inspection_criteria',
+				'sales_details',
+				'defaults',
+				'over_delivery_receipt_allowance',
+				'over_billing_allowance',
+				'supplier_details',
+				'taxes',
+			],
+			!frm.doc.is_fixed_asset
+		);
```
marination: can be driven by display depends on

nabinhait (approved):

PR #26283, opened 2021-07-01
fix: to fetch supplier group in Tax Rule 

Anuja-pawar:
Backport of https://github.com/frappe/erpnext/pull/25927


AfshanKhan:
@Anuja-pawar this PR doesn't have any changed files.

PR #26289, opened 2021-07-01
fix: Fixed Budget Variance Graph color from all black to default

nemesis189:
Budget Variance report had an all black graph due to an issue, fixed back to its default colors(pink,blue,..)

**Before:**

**After:**


deepeshgarg007 (commented):
@nemesis189 Please add a screenshot of the graph in the description

deepeshgarg007 (approved):

PR #26302, opened 2021-07-02
fix: set query for training events

Anurag810:
Fixed Query for a training event


erpnext/hr/doctype/training_event/training_event.js
```diff
@@ -20,11 +20,10 @@ frappe.ui.form.on('Training Event', {
...
 			});
 		}
-	}
-});
+		frm.events.set_employee_query(frm);
+	},
 
-frappe.ui.form.on("Training Event Employee", {
-	employee: function (frm) {
+	set_employee_query: function (frm) {
```
ruchamahabal:
```suggestion
	set_employee_query: function(frm) {
```

PR #26305, opened 2021-07-02
fix: stock level dashboard disappearing after refresh

### LINKED ISSUE #26260: Stock Levels keep disappearing from item detailed page
## stock levels disappeared from item detailed page, the panner that shows the quantities of the item and the related warehouses

**Output of `bench version`**
```
(paste here)
```

1.go to stock
2.click item from the quick access
3.choose any item and open the item detailed page
4.check stock levels 

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` ERPNext: v13.5.2 (HEAD)
Frappe Framework: v13.5.2 (HEAD)

ankush:
Problem: When `refresh_section` is called, all `.custom` class dashboards are removed, this makes the stock dashboard go away. Earlier this was fixed by reordering execution in https://github.com/frappe/erpnext/pull/25506 but now refresh section gets called after each field change due to https://github.com/frappe/frappe/pull/13355

Solution: use `stock-levels` as a class instead of the default `custom` that's removed. This also means we need to manually clear it before creating a new one to avoid duplicates. 

Also added UI test, so this doesn't break again. 

PS: To test you need to disable developer mode (as that disables caching.)

closes: https://github.com/frappe/erpnext/issues/26260

### DIFF
--- a/.github/helper/install.sh
+++ b/.github/helper/install.sh
@@ -42,6 +42,6 @@ sed -i 's/socketio:/# socketio:/g' Procfile
 sed -i 's/redis_socketio:/# redis_socketio:/g' Procfile
 
 bench get-app erpnext "${GITHUB_WORKSPACE}"
-bench start &
+bench start &> bench_run_logs.txt &
 bench --site test_site reinstall --yes
 bench build --app frappe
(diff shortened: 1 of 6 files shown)


ankush:
UI test failing because of this.

Sidenote: One shouldn't need to SSH just to see bench logs... need to find a way to make these more visible in CI.

Screenshot 2021-07-02 at 10 52 05 PM

ankush:
potentially better fix (applies to all dashboards): https://github.com/frappe/frappe/pull/13650

kossanah:
@marination @rohitwaghchaure  pls help to merge this fix. I really need this fix

ankush:
@kossanah we will ensure that fix is present in next release (~15-16th July)

ankush:
Merging this so it becomes usable for now... will undo it once a better fix is merged in Frappe framework.

PR #26309, opened 2021-07-03
fix: Update Ewaybill Tool Version

vorasmit:
As govt makes changes to excel tool, it changes version and old versions are withdrawn.

Changes to excel tool is not affecting our JSON file, and it's working with just change in version.

image


deepeshgarg007 (changes requested):
@vorasmit Please update the version check in tests too

PR #26316, opened 2021-07-03
feat: add french address template

FHenry:
add french address template

`no-docs`


erpnext/regional/address_template/templates/france.html
```diff
@@ -0,0 +1,6 @@
+{% if address_line1 %}{{ address_line1 }}{% endif -%}
+{% if address_line2 %}<br>{{ address_line2 }}{% endif -%}
+{% if pincode %}<br>{{ pincode }}{% endif -%}
+{% if pincode %} {{ city }}{% endif -%}
```
ankush: This looks like it should be `if city` and also `<br` maybe missing?
```suggestion
{% if city %} {{ city }}{% endif -%}
```

Can you cite some references for this template so it's easy for reviewers to merge?

FHenry: In france adress are 
"Digit street number" "name of the street"
"ZipCode" (ak pincode in ERPNext, number id of the city in french postal codification) "city name" (on the same line)
"Country"

For exemple :
8 rue des arbres
26000 VALENCE
France

There region or county but they are never display on adress.
Same as Email or phone number, the postman don't need to know this personal information

FHenry (commented):
In france adress are
"Digit street number" "name of the street"
"ZipCode" (ak pincode in ERPNext, number id of the city in french postal codification) "city name" (on the same line)
"Country"

For exemple :
8 rue des arbres
26000 VALENCE
France

There region or county but they are never display on adress.
Same as Email or phone number, the postman don't need to know this personal information

ankush (approved):

PR #26335, opened 2021-07-05
fix: optimized reposting code to improve performance

rohitwaghchaure:
**Issue**

While doing reposting for Repost Item Valuation system took more than 2 hours and even the reposting was not completed after 2 hours. After debug the codeI found that for the Manufacture type of stock entry, system does the reposting multiple times (as per number of raw materials used in the stock entry). For example in the below Manufacture stock entry, 10 raw materials were used to produce finished goods. 

Screenshot 2021-07-05 at 4 43 10 PM

During repost if out of 10 the 5 materials has been reposted then system was reposting finished good sle 5 times and due to which system was taking a lot of time to repost the transactions and remain in the In Progress state.

**After Fix**
System will do reposting of finished goods only once after completion of raw materials reposting. After change in the code system took 2 mins to complete reposting. Also added index on the column voucher_detail_no field in the table `tabStock Ledger Entry`

image

### DIFF
--- a/erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.json
+++ b/erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.json
@@ -142,6 +142,7 @@
    "oldfieldtype": "Data",
    "print_width": "150px",
    "read_only": 1,
+   "search_index": 1,
    "width": "150px"
   },
   {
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:
Fixed https://github.com/frappe/erpnext/pull/26432

PR #26357, opened 2021-07-06
fix: error popup for COA errors

AfshanKhan:
Previously:
problem

This used to happen because after upload `save_doc` function was called and it used to save the doc and refresh the page.

Now:
fix

Adding the function in `validate` fixes the problem.


erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
```diff
@@ -322,7 +324,7 @@ def validate_root(accounts):
 			error_messages.append("Root Type for {0} must be one of the Asset, Liability, Income, Expense and Equity".format(account.get("account_name")))
 
 	if error_messages:
-		return "<br>".join(error_messages)
+		frappe.throw(_("<br>".join(error_messages)))
```
nabinhait: It should not be translated here, append translated messages into error_messages.

Also, translate this string - https://github.com/frappe/erpnext/pull/26357/files#diff-5b59f3a16e0b23d23b478ff4a9e6fd1157018e8cadf4a385974e39508f13c558R29

nabinhait (changes requested):

nextchamp-saqib (approved):

PR #26370, opened 2021-07-07
fix: validation check for batch for stock reconciliation type in stock entry

noahjacob:
**Issue:**
- There were no filters while selecting batch in stock reconciliation.
image

- The check for valid batch no. was skipped if Voucher_Type = "Stock Reconciliation".

**Changes**
- Added Filters for batch no.
image

- Validation check for batch also validates if Voucher_type is "Stock Reconciliation"

### DIFF
--- a/erpnext/stock/doctype/stock_reconciliation/test_stock_reconciliation.py
+++ b/erpnext/stock/doctype/stock_reconciliation/test_stock_reconciliation.py
@@ -16,6 +16,7 @@ from erpnext.stock.utils import get_incoming_rate, get_stock_value_on, get_valua
 from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
 from erpnext.stock.doctype.purchase_receipt.test_purchase_receipt import make_purchase_receipt
 
+
 class TestStockReconciliation(unittest.TestCase):
 	@classmethod
 	def setUpClass(self):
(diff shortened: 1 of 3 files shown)


erpnext/stock/doctype/stock_reconciliation/test_stock_reconciliation.py
```diff
@@ -316,6 +317,26 @@ def test_backdated_stock_reco_future_negative_stock(self):
 		dn2.cancel()
 		pr1.cancel()
 
+	def test_valid_batch(self):
+		create_batch_item_with_batch("Testing Batch Item 1", "001")
+		create_batch_item_with_batch("Testing Batch Item 2", "002")
+		sr = create_stock_reconciliation(item_code="Testing Batch Item 1", qty=1, rate=100, batch_no="002"
+			, do_not_submit=True)
+		self.assertRaises(frappe.ValidationError, sr.submit)
```
ankush, line 325: nice!

ankush (approved):

erpnext/stock/doctype/stock_reconciliation/test_stock_reconciliation.py
noahjacob: :)

PR #26374, opened 2021-07-07
fix: change print_format_type from Server to Jinja

mhbu50:


mhbu50:
UP

ChillarAnand (approved):

ankush:
reference to original change: https://github.com/frappe/frappe/pull/7284 

`Server` was renamed to `Jinja`, it works regardless of this change because all the checks for print format use `!= 'JS'`.

PR #26383, opened 2021-07-08
feat: GST naming series - inform users about two extra chars on amending

rtdany10:
GST rules mandates 16 as the upper limit of number of characters. Users therefore might unknowingly set 15/16 chars in the naming series which will cause issue upon amending(extra 2 chars). The mess it can create in the middle of a financial year is huge considering the fact that we have immutable ledgers.

This PR informs users to limit it to 14 in case of submittable documents.
image

### DIFF
--- a/erpnext/regional/india/utils.py
+++ b/erpnext/regional/india/utils.py
@@ -163,7 +163,8 @@ def validate_document_name(doc, method=None):
 		return
 
 	if len(doc.name) > 16:
-		frappe.throw(_("Maximum length of document number should be 16 characters as per GST rules. Please change the naming series."))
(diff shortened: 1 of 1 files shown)


ankush:
FYI https://github.com/frappe/frappe/pull/13697

rtdany10:
> FYI [frappe/frappe#13697](https://github.com/frappe/frappe/pull/13697)

Woah! That's a nice feature!
This pull request isn't required then. Shall I close it?

rtdany10:
> FYI [frappe/frappe#13697](https://github.com/frappe/frappe/pull/13697)

@ankush Can we have this PR for v13 since the above mentioned PR was merged only on develop?

ankush:
Yeah, that feature could take some time.

PR #26384, opened 2021-07-08
fix: bad item name definition

FHenry:
in work order and job card, Item name is not correctly set in v12, fetch from is correct

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.json
+++ b/erpnext/manufacturing/doctype/job_card/job_card.json
@@ -275,7 +275,7 @@
    "read_only": 1
   },
   {
-   "fetch_from": "work_order.item_name",
+   "fetch_from": "work_order.production_item.item_name",
    "fieldname": "item_name",
    "fieldtype": "Read Only",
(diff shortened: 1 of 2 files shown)


ankush:
will backport from `develop` PR once it's final. Closing for now.

ref: https://github.com/frappe/erpnext/pull/26386#discussion_r685045376

erpnext/manufacturing/doctype/job_card/job_card.json
```diff
@@ -275,7 +275,7 @@
    "read_only": 1
   },
   {
-   "fetch_from": "work_order.item_name",
+   "fetch_from": "work_order.production_item.item_name",
```
ankush, line 278: This should just be `production_item.item_name`?

fetch_from doesn't work in nested manner. 


Restore branch and reopen PR, so we can get this fix in v12.

FHenry: That's the tricks, I know it should not work, but it is.  So that
production_item.item_name do not work.

PR #26385, opened 2021-07-08
fix: item name fetch from in Job Card

FHenry:
fix: item name fetch from in Job Card


FHenry:
No PR on version 13 I know. But please follow up upsteam your branch, version-13-hotfix cannot be merge in version-13 !!!

PR #26386, opened 2021-07-08
fix: item name displayed in Job Card

FHenry:
fix: item name displayed in Job Card


erpnext/manufacturing/doctype/job_card/job_card.json
```diff
@@ -280,7 +280,7 @@
    "read_only": 1
   },
   {
-   "fetch_from": "production_item.item_name",
+   "fetch_from": "work_order.production_item.item_name",
```
ankush, line 283: fetch_from doesn't work like this. It only works with `field_in_this_doc.link_field` format. Reference implementation: https://github.com/frappe/frappe/blob/develop/frappe/core/doctype/doctype/doctype.py#L199 

Can you describe the problem this PR solves? some steps to reproduce your problem would be great.

ankush (changes requested):

erpnext/manufacturing/doctype/job_card/job_card.json
FHenry: Bug is not in develop but in 12 only....
So only this PR https://github.com/frappe/erpnext/pull/26384 should be considerate not the one on devlop branch,. We already discuss this on the forum

PR #26421, opened 2021-07-10
feat: Optionally allow rejected quality inspection on submission

marination:
**Issue:**
 - On creating a QI against a Purchase Receipt, it is possible that this QI is rejected.
 - In this case, the submission of the Purchase Receipt is blocked due to a rejected QI.
 - It could be possible that after rejection the rejected qty is updated in the Purchase Receipt and the user wants to submit it.

**Solution:**
- Added **Action If Quality Inspection Is Rejected** in Stock Settings
   Screenshot 2021-06-21 at 4 23 28 PM
- If it is set to **Warn** it will show a toast message on submission, if not it will block the submission
   qi-reject-warn
- Cleaned up code to make it more readable

Docs: https://github.com/frappe/erpnext_documentation/pull/365

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -356,42 +356,68 @@ class StockController(AccountsController):
 		}, update_modified)
 
 	def validate_inspection(self):
-		'''Checks if quality inspection is set for Items that require inspection.
-		On submit, throw an exception'''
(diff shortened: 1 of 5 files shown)


marination:

marination:
already ported to v13 via branch backport

PR #26427, opened 2021-07-11
feat(Non Profit): API Endpoint to update halted Razorpay subscriptions

ruchamahabal:
When all payment retries have been exhausted, Razorpay moves the subscription to the halted state. The customer has to manually retry the charge or change the card linked to the subscription for the subscription to move back to the active state.

Sample payload for halted subscription: https://razorpay.com/docs/webhooks/payloads/#subscription-halted

- Changed Subscription Activated checkbox to Subscription Status field. Added a patch for the same.
- Added an API endpoint to update Razorpay subscription as halted. Razorpay webhook can be used to update this:
 
https://user-images.githubusercontent.com/24353136/125204114-27c49380-e299-11eb-9a03-d56a9cc89bb0.mp4

Documentation PR: https://github.com/frappe/erpnext_documentation/pull/366


erpnext/non_profit/doctype/membership/membership.py
```diff
@@ -294,9 +288,61 @@ def trigger_razorpay_subscription(*args, **kwargs):
...
+	if not data.event == "subscription.halted":
+		return
+
+	subscription = data.payload.get("subscription", {}).get("entity", {})
+	subscription = frappe._dict(subscription)
+
+	try:
+		member = get_member_based_on_subscription(subscription.id, customer_id=subscription.customer_id)
+		if not member:
+			frappe.throw(_("Member {0} not found"))
```
ruchamahabal:
```suggestion
			frappe.throw(_("Member with Razorpay Subscription ID {0} not found").format(subscription.id))
```

ruchamahabal:
Unrelated test case failing. Merging.

PR #26440, opened 2021-07-12
fix: don't get links if serial no creation details are None

18alantom:
### Description
Serial No creation details being None is causing `urllib.parse.quote` to throw an [error](https://stackoverflow.com/questions/55078654/troubelshooting-raise-typeerrorquote-from-bytes-expected-bytes).


ankush:
@18alantom can you update the description and add the case where document type/name are None?

18alantom:
> @18alantom can you update the description and add the case where document type/name are None?

Seems like an anomaly (ref: ISS-21-22-03916); deliberate removal maybe, but these fields are read-only.

Issue #26447, opened 2021-07-12
Item Group is hard-linked to Material Request


wojosc:
Hello everyone,

Item Group is linked to Material Request. Once Material Request ist submitted the Item Group can not be deleted as it is a linke field inside the table Material Request Item.

I suggest the field item_group in Material Request Item is changed from type Link to Data and only pulls the information as a string. This way there is no link between the two.

The field can though still be used for reports etc. but Item groups can then be deleted once they are retired.

A further option would be to Deactive Item Groups.

Also Item Groups can be merged. This will also chance the link inside Material Request which will make them useless on long term. Right?

PR will follow.

wojosc:
I see this is carried forward through all transactions sales order, sales invoice etc.

Please advise on if there is a strategy for this. Thank you.

PR #26448, opened 2021-07-12
Update material_request_item.json

wojosc:
https://github.com/frappe/erpnext/issues/26447

> Screenshots/GIFs

### DIFF
--- a/erpnext/stock/doctype/material_request_item/material_request_item.json
+++ b/erpnext/stock/doctype/material_request_item/material_request_item.json
@@ -217,11 +217,10 @@
   },
   {
    "fieldname": "item_group",
-   "fieldtype": "Link",
+   "fieldtype": "Data",
    "label": "Item Group",
    "oldfieldname": "item_group",
    "oldfieldtype": "Link",
-   "options": "Item Group",
    "print_hide": 1,
    "read_only": 1,
    "search_index": 1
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/material_request_item/material_request_item.json
```diff
@@ -217,11 +217,10 @@
   },
   {
    "fieldname": "item_group",
-   "fieldtype": "Link",
+   "fieldtype": "Data",
```
ankush, line 220: https://github.com/frappe/erpnext/issues/26447 

Referring to your original issue a better solution could be to "disable" an item group so no new items can be added to it. 

I don't think throwing away references or leaving stale references like this is a good idea. 

In MR it _might_ be okay.  In invoices? A big no. It's used by many reports. 

---

aside: please follow contribution guidelines regarding PR title and semantic pr/commits.

ankush (changes requested):

erpnext/stock/doctype/material_request_item/material_request_item.json
wojosc: Thanks for your feedback. As this structure is used all over the system I will take that into account and now know this for future reference.

I'm trying to thing of any reason why I would need any more than a text reference in all the transactions. I see that it makes navigation through the system easier as a link is clickabkle and the data is not. But beyond that? Beeing able to see all the transaction linked to the item group rather than finding that information at a customer/suppliert or item level.

Best

wojosc:
closed due to unclarities of design. It might clutter the system at some point but not in a major matter if aware.

PR #26464, opened 2021-07-13
feat: Added dispatch address fields in Sales Ord, Sales Inv, Delivery Note for Eway Bill

nemesis189:
Added Dispatch Address field to specify from where the delivery is actually dispatched  since it need not be same as the company address, modified the actualFromStatePin for Eway bill generation accordingly.

**Sales order:**
Screenshot 2021-07-14 at 12 42 33 PM

**Delivery Note:**
Screenshot 2021-07-14 at 12 44 19 PM

**Sales Invoice:**
Screenshot 2021-07-14 at 12 45 41 PM

`no-docs`

### DIFF
--- a/erpnext/regional/india/utils.py
+++ b/erpnext/regional/india/utils.py
@@ -432,7 +432,7 @@ def get_ewb_data(dt, dn):
 		billing_address = frappe.get_doc('Address', doc.customer_address)
 
 		#added dispatch address
-		dispatch_address = frappe.get_doc('Address', doc.dispatch_address_name)
+		dispatch_address = frappe.get_doc('Address', doc.dispatch_address_name) if doc.dispatch_address_name else company_address
 		shipping_address = frappe.get_doc('Address', doc.shipping_address_name)
 
 		data = get_address_details(data, doc, company_address, billing_address, dispatch_address)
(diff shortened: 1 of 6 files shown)


erpnext/regional/india/utils.py
```diff
@@ -519,10 +521,10 @@ def get_gstins_for_company(company):
 			`tabDynamic Link`.link_name = %(company)s""", {"company": company})
 	return company_gstins
 
-def get_address_details(data, doc, company_address, billing_address):
+def get_address_details(data, doc, company_address, billing_address, dispatch_address):
 	data.fromPincode = validate_pincode(company_address.pincode, 'Company Address')
-	data.fromStateCode = data.actualFromStateCode = validate_state_code(
-		company_address.gst_state_number, 'Company Address')
+	data.fromStateCode = validate_state_code(company_address.gst_state_number, 'Company Address')
+	data.actualFromStateCode = validate_state_code(dispatch_address.gst_state_number, 'Company Address')
```
deepeshgarg007: Dispatch Address is not a mandatory field. This should only be called if dispatch address is set else `actualFromStateCode` will be same as company billing state

deepeshgarg007 (changes requested):

deepeshgarg007 (commented):
Try generating E-Way bill once without the dispatch address and check there should not be any other errors because of this. Also update `test_eway_bill_json` in `test_sales_invoice.py` to check these fields as well.

deepeshgarg007 (approved):

PR #26472, opened 2021-07-13
fix: Unallocated amount in Payment Entry after taxes

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/26412

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -514,6 +514,7 @@ class PaymentEntry(AccountsController):
 		self.base_total_allocated_amount = abs(base_total_allocated_amount)
 
 	def set_unallocated_amount(self):
+		self.unallocated_amount = 0
 		if self.party:
 			total_deductions = sum(flt(d.amount) for d in self.get("deductions"))
 			if self.payment_type == "Receive" \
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -509,18 +514,17 @@ def set_total_allocated_amount(self):
...
-		self.unallocated_amount = 0
 		if self.party:
 			total_deductions = sum(flt(d.amount) for d in self.get("deductions"))
 			if self.payment_type == "Receive" \
-				and self.base_total_allocated_amount < self.base_received_amount_after_tax + total_deductions \
-				and self.total_allocated_amount < self.paid_amount_after_tax + (total_deductions / self.source_exchange_rate):
-				self.unallocated_amount = (self.received_amount_after_tax + total_deductions -
+				and self.base_total_allocated_amount < self.base_received_amount + total_deductions \
+				and self.total_allocated_amount < self.paid_amount + (total_deductions / self.source_exchange_rate):
+				self.unallocated_amount = (self.received_amount + total_deductions -
```
clefincode, line 525: Ther is error here please fix it line 522 must be:
self.unallocated_amount = (self.base_received_amount + total_deductions -

**base_received_amount**   not **received_amount** 
and it case error when Receive multi currency payment

clefincode (commented):
i have fix it on my client server please fix it and incloud it in next update here is the needed update
image

PR #26473, opened 2021-07-13
feat(India): Bootstrapped GST Setup

deepeshgarg007:
Back Port Of: https://github.com/frappe/erpnext/pull/25415

`no-docs`


umairsy:
- Set default tax rate to zero
- In Sales Taxes and Charges Master, set rate to zero
1. The name of Item Tax Template is GST-9%, It should be GST-18%
2. The chart of account creates tax account heads with a percentage 9 added. It should be 0
3. The sales tax template and purchase tax template also have a percentage set to 9 or 18 percent. It should be 0.
4. Further to point 3, the name (description) should also not have 9% in it.  This gets printed in Sales Invoice even when the percentage is not 9%
5. 2. account head where tax percentage should be 0 by default

@deepeshgarg007 can we get these records reviewed and assigned to someone please?

Issue #26474, opened 2021-07-13
Pricing Rule Naming


wojosc:
When creating new Pricing Rule it is mandetory to set title. Also Title is a unique field. This makes creating Pricing Rules very cumbersome.
![grafik](https://user-images.githubusercontent.com/22279621/125444922-16b93bea-51dd-4107-ac2a-d48f4e4490fb.png)

![grafik](https://user-images.githubusercontent.com/22279621/125444950-7b70e19e-849c-470e-81a1-8a7a436f6f1d.png)

We tried changing the naming series in the docType (possible since v13) but this does not change anything...
![grafik](https://user-images.githubusercontent.com/22279621/125445214-5687931c-2640-4468-8892-4e68e7303f66.png)

I suggest to remove the mandetory and uniqueness in the field "title" to be able to make use of naming_series in Naming.

wojosc:
@reggaetune @canlann

PR #26475, opened 2021-07-13
Making Naming Series on Pricing Rule usable

wojosc:
https://github.com/frappe/erpnext/issues/26474

> Screenshots/GIFs

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.json
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.json
@@ -94,9 +94,7 @@
    "fieldname": "title",
    "fieldtype": "Data",
    "label": "Title",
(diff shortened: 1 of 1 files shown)


ankush:
https://github.com/frappe/erpnext/blob/13fffeb97b8fcc984fc598710121b886b3bf69d9/erpnext/accounts/doctype/pricing_rule/pricing_rule.json#L5

This will anyways not allow you to have the same title twice. So removing `unique` alone isn't enough.



---

aside: please add a proper title and short description to PR and make commit messages semantic.

wojosc:
@ankush thank you for your reply. We will be able to have the uniqueness of the docType trough using Naming Series for this doctype. At the moment we are not able to do so though.

Does anyone know of any reason why the title field is mandatory and unique?

PR #26476, opened 2021-07-13
fix: item group portal issues

18alantom:
### Description
PR does 2 things
- Show child _Item Group_ _Items_ if child _Item Group_ and _Item_ have _Show in Website_ selected
- Maintain _Item Group_ as a filter for an _Item Group_ portal page when field filters are set. 

**Before**: On setting a filter all items were shown while ignoring the implicit item group filter.

(ref issues: ISS-21-22-03005, ISS-21-22-03661)


18alantom:
_Note: `persistent_filters` set as a global using jinja, bit of a hack fix._

marination:
@18alantom Maybe this could be cleaner, what do you think:
- We know there are two sources of querying items here: `ProductQuery` in `get_context` and `get_products_html_for_website` that runs on filter change
- `get_products_html_for_website` is the issue here that only handles filters passed to it. Now these filters are reset because `index.js` is not differentiating between an Item group page and an All Products page.

**Approach:**
- Lets add `data-item-group` on the page for identification. Get name from the item group context getter
   ```html
        <div class="item-group-content" itemscope itemtype="http://schema.org/Product" data-item-group="{{ name }}">
    ```
- Check for `data-item-group` in `get_items_with_filters` in `index.js`
   ```js
        let item_group = $(".item-group-content").data("item-group");
    ```
- Pass item_group to `get_products_html_for_website` in args if item_group exists. And in this method if item_group exists in args, add item_group filter to field filters. `get_products_for_website` should filter correctly

Essentially lets keep the item group filter bit as seamless as we can on the front end, because implicit filter

marination (approved):

marination:
@18alantom backport this to `version-13-hotfix` and `version-13-pre-release` also

PR #26498, opened 2021-07-14
fix: Paging buttons not working on item group portal page

marination:
Port of https://github.com/frappe/erpnext/pull/26496


marination:
unrelated `test_subcontracting_over_receipt` failing

PR #26499, opened 2021-07-14
fix: WIP needs to be set before submit on skip_transfer

noahjacob:
**Issue:**
- System throws error on submit if WIP Warehouse field is empty even if "Skip Material Transfer to WIP Warehouse" is checked.
image

**Expected:**
- User should be able to leave the WIP Warehouse field empty if "Skip Material Transfer to WIP Warehouse" is enabled, since WIP field is marked non-mandatory on enabling the option and should fetch source warehouses in stock entry as per item source warehouses

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.py
+++ b/erpnext/manufacturing/doctype/work_order/work_order.py
@@ -239,7 +239,7 @@ class WorkOrder(Document):
 		self.create_serial_no_batch_no()
 
 	def on_submit(self):
-		if not self.wip_warehouse:
+		if not self.wip_warehouse and not self.skip_transfer:
 			frappe.throw(_("Work-in-Progress Warehouse is required before Submit"))
 		if not self.fg_warehouse:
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -239,8 +239,9 @@ def before_submit(self):
 		self.create_serial_no_batch_no()
 
 	def on_submit(self):
-		if not self.wip_warehouse:
-			frappe.throw(_("Work-in-Progress Warehouse is required before Submit"))
+		if not self.skip_transfer:
+			if not self.wip_warehouse:
+				frappe.throw(_("Work-in-Progress Warehouse is required before Submit"))
```
ankush:
```suggestion
		if not self.skip_transfer and not self.wip_warehouse:
			frappe.throw(_("Work-in-Progress Warehouse is required before Submit"))
```

Merge conditions instead of nested?

ankush (dismissed):

PR #26503, opened 2021-07-14
feat: total billing hours for timesheet in Sales Invoice

anupamvs:
As we add the timesheet to the invoice we can see the total billing amount of the timesheet, but there is no provision to see the "Total billing hours of timesheets".
image


ruchamahabal:
@anupamvs add a small test for this

PR #26506, opened 2021-07-15
fix: check if field_filters is None

18alantom:
Backport from PR #26504

### DIFF
--- a/erpnext/portal/product_configurator/utils.py
+++ b/erpnext/portal/product_configurator/utils.py
@@ -101,7 +101,7 @@ def get_products_html_for_website(field_filters=None, attribute_filters=None):
 	return html
 
 def set_item_group_filters(field_filters):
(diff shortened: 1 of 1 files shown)


erpnext/portal/product_configurator/utils.py
```diff
@@ -101,7 +101,7 @@ def get_products_html_for_website(field_filters=None, attribute_filters=None):
 	return html
 
 def set_item_group_filters(field_filters):
-	if 'item_group' in field_filters:
+	if field_filters is not None and 'item_group' in field_filters:
```
AfshanKhan, line 104: `if field_filters and 'item_group' in field_filters:`

18alantom: This is correct but increases ambiguity hence the `None` check.

PR #26510, opened 2021-07-15
fix(ux): set default operation time to 0 

noahjacob:
**Before:**
- Threw an error if operation time was left empty.
- No default value.
image

**Changes:**
- Default value for sub operation time is set to 0.

### DIFF
--- a/erpnext/manufacturing/doctype/sub_operation/sub_operation.json
+++ b/erpnext/manufacturing/doctype/sub_operation/sub_operation.json
@@ -19,6 +19,7 @@
    "options": "Operation"
   },
   {
+   "default": "0",
    "description": "Time in mins",
    "fieldname": "time_in_mins",
    "fieldtype": "Float",
@@ -38,7 +39,7 @@
  "index_web_pages_for_search": 1,
  "istable": 1,
  "links": [],
- "modified": "2020-12-07 18:09:18.005578",
+ "modified": "2021-07-15 16:39:41.635362",
  "modified_by": "Administrator",
  "module": "Manufacturing",
  "name": "Sub Operation",


erpnext/manufacturing/doctype/operation/operation.js
```diff
@@ -10,5 +10,12 @@ frappe.ui.form.on('Operation', {
 				}
 			};
 		});
+	},
+});
+frappe.ui.form.on('Sub Operation', {
+	operation: function(frm, cdt, cdn) {
+		let row = locals[cdt][cdn]
+		row.time_in_mins = 0
+		frm.refresh_field('sub_operations')
```
ankush, line 19: Why this much code for setting default? Just edit child table doctype and put `0.0` as default value. 

To fix existing records, write a patch that changes all `NULL` value to 0.

ankush (approved):

erpnext/manufacturing/doctype/operation/operation.js
```diff
@@ -10,5 +10,12 @@ frappe.ui.form.on('Operation', {
...
 	}
```
ankush: Why this much code for setting default? Just edit child table doctype and put `0.0` as default value. 

To fix existing records, write a patch that changes all `NULL` value to 0. Though I doubt you'll need that too, because no records would be saved if this error occured, right?

ankush (changes requested):

erpnext/manufacturing/doctype/operation/operation.js
noahjacob: yea I'll just edit doctype default

ankush (approved):

PR #26511, opened 2021-07-15
fix(ux): set operation time to 0 if None (bp #26510)

noahjacob:
Backport of #26510


ankush:
https://github.com/frappe/erpnext/pull/26510#discussion_r670362066

PR #25991, opened 2021-06-09
fix(Issue): Calculate first_response_time based on working hours

GangaManoj:
_Problem:_
First Response Time calculation does not take working hours into consideration- it just takes the difference between first_responded_on and time of creation of the Issue.

_Fix:_
Checks if first_responded_on and time of creation of the Issue are:

- on the same day
- during working hours
- on working days

PR #26520, opened 2021-07-15
fix: Calculate first_response_time based on working hours

GangaManoj:
Backport of https://github.com/frappe/erpnext/pull/25991


erpnext/support/doctype/issue/issue.py
```diff
@@ -25,13 +25,8 @@ def validate(self):
 		if not self.raised_by:
 			self.raised_by = frappe.session.user
```
nextchamp-saqib:
```suggestion
		self.change_service_level_agreement_and_priority()
		self.update_status()
```

erpnext/support/doctype/issue/issue.py
```diff
@@ -25,13 +25,8 @@ def validate(self):
...
 
-		self.change_service_level_agreement_and_priority()
-		self.update_status()
 		self.set_lead_contact(self.raised_by)
```
nextchamp-saqib, line 32:
```suggestion
		if not self.service_level_agreement:
			self.reset_sla_fields()

```

nextchamp-saqib (approved):

PR #26546, opened 2021-07-19
fix: assigning values to rows in sales register reports

AfshanKhan:
### Problem:

Suppose user permission are set on account. The value for unrealized profit and loss account is getting updated when making a Sales Invoice.
This shows wrong values in Sales Register Report and Item-Wise Sales Register Report.

### Solution:

Adding `_unrealized` to the fieldname for Unrealized Profit Loss Account in Sale Register Report so the accounts are distinguishable.
Add ignore user permission check for unrelized profit and loss account in Sales Invoice
Don't add that column for Unrealized Profit Loss Account if its not an internal transaction


erpnext/accounts/report/sales_register/sales_register.py
```diff
@@ -84,7 +84,7 @@ def _execute(filters, additional_table_columns=None, additional_query_columns=No
 		# Add amount in unrealized account
 		for account in unrealized_profit_loss_accounts:
 			row.update({
-				frappe.scrub(account): flt(internal_invoice_map.get((inv.name, account)))
+				frappe.scrub(account+"_unrealized"): flt(internal_invoice_map.get((inv.name, account)))
```
deepeshgarg007, line 87: If unrealized profit and loss account is used then it means its an internal transaction and no income is being booked so doesn't matter even if income account is being overridden

deepeshgarg007 (changes requested):

erpnext/accounts/report/sales_register/sales_register.py
AfshanKhan: @deepeshgarg007  Exactly. So suppose there is an account let's say ` income_account` which for a transaction has value 100. Now this account was also part of internal transaction and that is why it is in unrealized profit and loss ka list too. With a value 0. Now according to the code there are 2 columns with the same account. One for income and another for unrealized profit and loss. First the value will be assigned to income_account 100. Then later on as the fieldname is same so instead of assigning 0 to that 2nd column it is being overwritten to the 1st column which had 100. Now it has value 0. Similarly for every row the value for income_account will be 0.

deepeshgarg007: > @deepeshgarg007 Exactly. So suppose there is an account let's say ` income_account` which for a transaction has value 100. Now this account was also part of internal transaction and that is why it is in unrealized profit and loss ka list too. With a value 0. Now according to the code there are 2 columns with the same account. One for income and another for unrealized profit and loss. First the value will be assigned to income_account 100. Then later on as the fieldname is same so instead of assigning 0 to that 2nd column it is being overwritten to the 1st column which had 100. Now it has value 0. Similarly for every row the value for income_account will be 0.

For an internal transaction there won't be any income booked, so doesn't matter, right? Or is it affecting any other transactions? Please add screenshots before/after for this so that there is clear understanding

AfshanKhan:
@deepeshgarg007 Added the changes

PR #26549, opened 2021-07-19
fix:Ignore mandatory fields while creating payment reconciliation Journal entry

nemesis189:
Ignore mandatory fields which prevents payment reconciliations from being created


erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
```diff
@@ -306,5 +306,5 @@ def reconcile_dr_cr_note(dr_cr_notes, company):
 				}
 			]
 		})
-
+		jv.flags.ignore_mandatory = True
```
AfshanKhan, line 309: @nemesis189 for which fields are we adding this? All mandatory fields are getting assigned some values right?

deepeshgarg007 (approved):

Issue #26444, opened 2021-07-12
GSTR3B: The report is not correct when we have invoices with discount on grand total.


mujeerhashmi:
## Description of the issue
Issue #21231 is again evident in v13.

## Context information (for bug reports)

**Output of `bench version`**
```
erpnext 13.6.0
frappe 13.6.0
```

mujeerhashmi:
@deepeshgarg007 please have a look at it.

mujeerhashmi:
Ping

PR #26553, opened 2021-07-19
fix: Additional discount calculations in Invoices

deepeshgarg007:
Reverting client side handling of dynamic item tax templates which caused issue while calculations on total after application of additional discount

These changes were introduced in this PR
https://github.com/frappe/erpnext/pull/25961


mujeerhashmi:
Is this a fix for #26444

deepeshgarg007:
> Is this a fix for #26444

No this for a separate issue

PR #26560, opened 2021-07-20
fix: Price list rate not fetched for return sales invoice fixed

nemesis189:
When a sales invoice is created with 'Is Return' checked and a price list is selected(other than standard), the items added to the child table wouldn't auto fetch the price list rate. Fixed it by removing the condition that prevented it.

Steps:
1. create sales invoice with Is Return checked
2. select a custom price list
3. add items and check its rate

port for : https://github.com/frappe/erpnext/pull/26559

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -74,9 +74,8 @@ def get_item_details(args, doc=None, for_validate=False, overwrite_warehouse=Tru
 
 	update_party_blanket_order(args, out)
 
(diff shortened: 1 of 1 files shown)


AfshanKhan (approved):

wikiodisha1:
Same issue in Version-12

AfshanKhan:
@nemesis189 Verify and port this on `version-12-hotfix` branch as well.

AfshanKhan:
https://github.com/frappe/erpnext/pull/26593 fix for `version-12-hotfix`

PR #26562, opened 2021-07-20
fix: Production Plan: traverse all BOM levels looking for manufactured items not in stock

nameduser0:
Currently (v13.6) the production plan can either deal with 2-level BOMs (non-exploded option) or BOMs that are always made directly from the base materials (exploded items imported from the BOM).

So it does not support, for example, 3-level BOMs where there may be stock of intermediate manufactured parts.

This is a fix that enables any level of nested BOMs and checks stock at each level so you only need to manufacture the correct number of parts at each BOM level.

fixes: #26563

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -530,14 +530,24 @@ class ProductionPlan(Document):
 		bom_data = sorted(bom_data, key = lambda i: i.bom_level)
 
 		for data in bom_data:
-			data.qty = data.stock_qty
-			data.production_plan_item = row.name
-			data.fg_warehouse = row.warehouse
-			data.schedule_date = row.planned_start_date
-			data.type_of_manufacturing = manufacturing_type or ("Subcontract" if data.is_sub_contracted_item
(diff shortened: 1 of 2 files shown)


nameduser0:
Test failure unrelated

nameduser0:
bug2: When `Include Exploded Items` is not checked, and `Make Work Order For Sub Assembly Items` is checked, the PP always makes Work Orders for all BOM levels for all parts, despite the PP knowing from the material planning that only some parts may be required.

Workaround:
The process is meant to lead you to make Work Orders with `Make Work Order For Sub Assembly Items` unchecked, then a Manufacture Material Request, then separate Work Orders for missing parts, but this behaviour is not intuitive. You're already making a manufacturing plan to make some goods, why would all the necessary Work Orders not be made directly against that?

nameduser0:
Fixes bug2 above but has been ported from v3.6 and not tested on this branch so may not work, but I had to fix it so the tests pass.

Incidentally I am extremely sceptical about the recent changes; a `bom_level` saved in the BOM sends shivers down my spine. Consider having a machined bolt with a BOM used in assemblies at all "levels".

nameduser0:

Issue #26465, opened 2021-07-13
feat: Production Plan: populate Planned Start Date with Sales Order line Delivery Date


nameduser0:
**Is your feature request related to a problem? Please describe.**
The production plan planned start date for lines is populated with the current date & time. There are better sources of information than this

**Describe the solution you'd like**
Copy the Delivery Date from the Sales Order line
Nice-to-have: a Manufacturing Setting for `Default Production Lead Time` in days which is subtracted from the SO line delivery date to give the planned start date.

Related: #25800

nameduser0:
😄

PR #26568, opened 2021-07-21
feat: Production Plan: add a production lead time setting and use it to set planned start dates on the plan from delivery dates

nameduser0:
fixes: #26465

no-docs

### DIFF
--- a/erpnext/manufacturing/doctype/manufacturing_settings/manufacturing_settings.json
+++ b/erpnext/manufacturing/doctype/manufacturing_settings/manufacturing_settings.json
@@ -29,7 +29,8 @@
   "update_bom_costs_automatically",
   "add_corrective_operation_cost_in_finished_good_valuation",
   "column_break_23",
-  "make_serial_no_batch_from_work_order"
+  "make_serial_no_batch_from_work_order",
+  "production_lead_time_days"
  ],
  "fields": [
   {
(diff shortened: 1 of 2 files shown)


nameduser0:
Unrelated test failure

nameduser0:
@rmehta @umairsy As usual, valuable contributions ignored and marked stale

nameduser0:
😆 [https://erpnext.org/contribute](https://erpnext.org/contribute)

erpnext/manufacturing/doctype/manufacturing_settings/manufacturing_settings.json
```diff
@@ -175,13 +176,20 @@
    "fieldname": "add_corrective_operation_cost_in_finished_good_valuation",
    "fieldtype": "Check",
    "label": "Add Corrective Operation Cost in Finished Good Valuation"
+  },
+  {
+   "default": "10",
+   "description": "Subtracted from Sales Order and Material Request delivery dates for Production Plan start date",
+   "fieldname": "production_lead_time_days",
```
ankush, line 183: A configuration like this on the BOM document would be better? Single lead time value can't apply to all items for most businesses.  

Also the linked issue you mentioned is fixed now, so Work Orders now respect set dates from ProdPlan.

https://github.com/frappe/erpnext/pull/27031

nameduser0: I would suggest the BOM should have enough `Operation` information to work out a lead time if you wanted to. I'd suggest using this as a default. I don't have time to rework or improve this.

PR #25845, opened 2021-05-26
fix: production plan not fetching sales order of a variant

noahjacob:
**Current Issue:**
- Created a Sales Order for an item variant.(The parent item has a BOM)
so_1

- On creating a Production Plan and trying to fetch in this Sales Order using Item Code, it doesn't get fetched.

so_2

**Reason:**
- No condition in SQL query when item variant BOM does not exist.

**Solution:**
- The query must fetch the parent item's BOM if the item variant does not have a BOM.

**After Fix**:
- Updated Query.
- Parent item BOM.

image

- When item variant has a BOM.

variant_with

- When item variant doesn't have a BOM. (Fetches parent BOM)

variant_without

PR #26569, opened 2021-07-21
feat: Production Plan: make the item_code filter into a 'like' search rather than a link field

nameduser0:
Allows you to add multiple items with similar item_codes to the same production plan rather than the strict single item according to a link field

Consider a realistic scenario: a clothes printing shop is much more likely to want to make all t-shirts, or all medium t-shirts, or all white t-shirts, or all t-shirt sizes with a single design (new behaviour) than they are to make a single t-shirt design, in a single size, in a single colour (existing behaviour).

no-docs

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.json
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.json
@@ -194,6 +194,7 @@
   {
    "fieldname": "po_items",
    "fieldtype": "Table",
+   "label": "Work Order Items",
    "no_copy": 1,
    "options": "Production Plan Item",
    "reqd": 1
@@ -332,12 +333,12 @@
   },
   {
    "fieldname": "section_break_25",
-   "fieldtype": "Section Break"
+   "fieldtype": "Section Break",
+   "hidden": 1
   },
   {
    "fieldname": "prod_plan_references",
    "fieldtype": "Table",
-   "hidden": 1,
    "label": "Production Plan Item Reference",
    "options": "Production Plan Item Reference"
   },
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.json
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.json
@@ -98,11 +98,10 @@
    "label": "Filters"
   },
   {
+   "description": "Searches for item codes like this string",
    "fieldname": "item_code",
-   "fieldtype": "Link",
-   "in_list_view": 1,
-   "label": "Item Code",
-   "options": "Item"
+   "fieldtype": "Data",
+   "label": "Item Code"
   },
   {
    "depends_on": "eval: doc.get_items_from == \"Sales Order\"",
@@ -364,7 +363,7 @@
  "index_web_pages_for_search": 1,
  "is_submittable": 1,
  "links": [],
- "modified": "2021-06-28 20:00:33.905114",
+ "modified": "2021-07-20 17:55:37.164051",
  "modified_by": "Administrator",
  "module": "Manufacturing",
  "name": "Production Plan",
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.json
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.json
@@ -93,12 +93,13 @@
    "collapsible": 1,
    "collapsible_depends_on": "eval: doc.__islocal",
    "depends_on": "eval: doc.get_items_from",
+   "description": "These filters are used in searches triggered by the buttons below",
    "fieldname": "filters",
    "fieldtype": "Section Break",
    "label": "Filters"
   },
   {
-   "description": "Searches for item codes like this string",
+   "description": "Look for item codes like this text",
    "fieldname": "item_code",
    "fieldtype": "Data",
    "label": "Item Code"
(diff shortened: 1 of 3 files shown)


nameduser0:
Won't-fix old format SQL

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -62,11 +62,10 @@ def get_pending_material_requests(self):
 			mr_filter += " and mr.transaction_date <= %(to_date)s"
 		if self.warehouse:
 			mr_filter += " and mr_item.warehouse = %(warehouse)s"
-
 		if self.item_code:
-			item_filter += " and mr_item.item_code = %(item)s"
+			item_filter += " and mr_item.item_code like %(item)s"
 
-		pending_mr = frappe.db.sql("""
+		pending_mr = frappe.db.sql("""   # nosemgrep
```
```suggestion
		pending_mr = frappe.db.sql(""" 
```

This is becoming part of the query string. I've removed the rule that's causing unnecessary warnings. So you can undo this change.

ankush:
This looks neat 👍  Especially because planning is usually not done on item level it makes sense to use "LIKE" filtering... will do some manual testing before merging.

nameduser0:
@ankush Thanks. It's not the best solution. I would have liked to change the buttons to be additive instead of clear table and fill from scratch, but then I would need to implement checks to see if the lines were already added. This quick fix is the next best thing.

erpnext/manufacturing/doctype/production_plan/production_plan.json
```diff
@@ -98,11 +98,10 @@
    "label": "Filters"
   },
   {
+   "description": "Searches for item codes like this string",
```
ankush: Replace "string" with text, string is programmer-talk. Or remove the full description. We usually don't like writing descriptions, it should be obvious wherever possible.

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -79,7 +78,7 @@ def get_pending_material_requests(self):
 				"from_date": self.from_date,
 				"to_date": self.to_date,
 				"warehouse": self.warehouse,
-				"item": self.item_code,
+				"item": '%' + self.item_code + '%',
```
ankush: when `self.item_code` is not entered by the user, this will give a string concatenation error. (Same with SO query down below)

ankush (changes requested):

ankush:
btw we were thinking of removing all these filter fields and adding modal dialog (already present in many other documents) Might take some time for this though. 

this is from picklist:
Screenshot 2021-07-30 at 3 30 20 PM

erpnext/manufacturing/doctype/production_plan/production_plan.json
nameduser0: I'll change the word but I think this is the wrong policy on descriptions.

This is an extremely powerful and easy self-documentation opportunity. ERP systems are very much **not** straightforward to pick up, each field is rarely obvious and often has knock-on implications in many related documents and fields. This one field for example doesn't just get used to filter Sales Orders, but also the Items that are added to the PP. This is not obvious.

My opinion (who has spent several years with ERP system users) is that **every** field or section should have a description, unless it truly is obvious like `Date of Birth` kind of stuff. A user shouldn't need to refer to the documentation to make sense of the form, which incidentally rarely fully describes the features anyway.

Another very much not obvious field is `Include Exploded Items`. I had to read the code before I understood what that did. Who wants a production plan that doesn't `Include Exploded Items` ? As it turns out, it's just a broken, badly named feature that I had to fix which I renamed `Make From Scratch`.

nameduser0:
> btw we were thinking of removing all these filter fields and adding modal dialog (already present in many other documents) Might take some time for this though.

Makes sense if it adds functionality, like a full set of custom field filters. But in general I don't agree many dialogs are the answer, they break the user experience, like the Warehouse dialog on `Get Raw Materials For Production`. This should be on-page.

Dialogs are also a bad idea if they force the user to enter the same thing twice, like filtering for Sales Orders containing an Item, then filtering again for Items. Or if they lose work that takes time. Adding to the existing fields is easy because it's data on the page that is saved, so easy to edit. On a dialog the filters would be lost.

Another example of pointless feature creep is the `awesomplete` js that moves the page around every time you complete a field. This is extremely annoying. Users like things to be where they left them, not constantly searching for them because they moved or the section was auto-collapsed. Imagine someone moving all your stuff around your room every day without asking and with no way to stop it. You would quickly move.

In general, your time would be better spent making existing functionality 100% bug-free and usable than continuous feature-creep and refactoring. ERPNext is losing customers daily.

nameduser0:

erpnext/manufacturing/doctype/production_plan/production_plan.py
ankush: I don't think the change you did in last commit will solve this. 

F-strings will convert `None` to string literal `'None'` and then search for item code like `%None%`

nameduser0: In all instances the additional filter conditions are only added if item_code is truthy

ankush:
LGTM. Just resolve conflicts or allow maintainer edits.

nameduser0:
> LGTM. Just resolve conflicts or allow maintainer edits.

@ankush See comments on #25845. conflicts are unresolvable due to that breaking change.

nameduser0:
@ankush Your last comment on the other thread is also incorrect. A BOM captures both the raw materials (subitems) and process (operations). So as I stated, if items have the same BOM, they are by definition the same item. And if this is how the system is being used, it is simply abuse.

ankush:
Not everyone captures all the processes in BOM. We are software developers, we don't enforce good systems to end-users.  🤷 

If you want this strict check then add a setting in manufacturing settings to disallow selling/fetching BOM template... even make it the default behaviour.

PR #26570, opened 2021-07-21
fix: Production Plan: respects delivered quantity from Sales Orders

### LINKED ISSUE #26534: bug: Production Plan - Get Items For Work Order from Sales Order - does not respect delivered quantity
## Description of the issue
The delivered quantity from a sales order is not respected when creating a production plan leading to over-production

**Output of `bench version`**
```
13.6
```

1. Create a product and insert stock of qty=10
2. Create a BOM for the Item
3. Create a Sales Order for the Item qty=50
4. Create a delivery note for the SO, and deliver qty=10 from stock
5. Create a production plan for the Sales Order

### Observed result
Planned qty = 50

### Expected result
Planned qty = 40

nameduser0:
fixes #26534

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -121,9 +121,9 @@ class ProductionPlan(Document):
 			item_condition = ' and so_item.item_code = {0}'.format(frappe.db.escape(self.item_code))
 
 		items = frappe.db.sql("""select distinct parent, item_code, warehouse,
-			(qty - work_order_qty) * conversion_factor as pending_qty, description, name
+			(qty - work_order_qty - delivered_qty) * conversion_factor as pending_qty, description, name
 			from `tabSales Order Item` so_item
-			where parent in (%s) and docstatus = 1 and qty > work_order_qty
+			where parent in (%s) and docstatus = 1 and qty > work_order_qty + delivered_qty
 			and exists (select name from `tabBOM` bom where bom.item=so_item.item_code
 					and bom.is_active = 1) %s""" % \
 			(", ".join(["%s"] * len(so_list)), item_condition), tuple(so_list), as_dict=1)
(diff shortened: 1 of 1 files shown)


nameduser0:
Unrelated test failures

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -121,9 +121,9 @@ def get_so_items(self):
 			item_condition = ' and so_item.item_code = {0}'.format(frappe.db.escape(self.item_code))
 
 		items = frappe.db.sql("""select distinct parent, item_code, warehouse,
-			(qty - work_order_qty) * conversion_factor as pending_qty, description, name
+			(qty - work_order_qty - delivered_qty) * conversion_factor as pending_qty, description, name
```
ankush, line 124: This becomes problematic when `delivered_qty` is also manufactured via a work order.

E.g. assume everything is for a single item

SO - 100 Qty
DN - 10 Qty from stock
WO - 50 Qty
DN - 50 Qty recently manufactured

2nd WO = 100 - 50 - 60 = -10 qty (which doesn't get added in table)
Actually pending qty = 100 - 60 = 40

So in the case of multi-stage production plans. This will cause double subtraction of manufactured quantity.

ankush: Can you also change this and subtract produced quantity?

ankush: This _might_ work:

Pending qty = required qty - (WO qty - Produced qty) - delivered qty

---
Nope. This will again show higher qty between produced and delivery stage.

nameduser0: It seems to me this is a symptom of multiple accounting on the Sales Order Item.

Are you saying that all of these quantities are additive only, so nothing is subtracted (unless cancelled)? Which means they will always add up to more than the ordered quantity?

Seems to me this needs a re-think. If a work order is complete, why is it still a work order quantity and not transferred to Produced Quantity?
Shouldn't a plan take into account the Planned Quantity?
No idea what Projected, Actual and Ordered Quantities are.

Screen Shot 2021-07-31 at 01 44 36

nameduser0: Logically this is what I'd expect:
pending_qty = qty - delivered_qty - actual_qty - ordered_qty - planned_qty - work_order_qty - produced_qty + returned_qty
(I'd also expect that to be a property of the doc, whether calculated or stored, but...)

Expected transfer of quantities (debit one, credit another):
actual -> delivered
ordered -> actual -> delivered
planned -> work_order -> produced -> delivered
returned is effectively an addition to the original qty that must be re-delivered

actual_qty = available_stock_qty ?
ordered_qty = purchase_ordered_qty ?

Whether attempting to account for everything against the sales order item is a good idea or not is another question.
It's certainly brave.

nameduser0:

ankush:
As pointed out by you, this will require some refactoring in the way we are tracking the progress of various processes on sales order before change can be done for respecting delivered qty. Closing for now.

PR #26582, opened 2021-07-21
fix: COA importer account type validation error fix

nemesis189:
COA importer would always throw error during accounts type validation because of the if condition used in creating the account types list. It never got satisfied because string '1'  was compared with int '1'.

Steps:
1. upload a file in coa importer tool
2. check for any errors being thrown (in this case : Please identify/create Account (Group) for type - Bank, Stock, Cash)

### DIFF
--- a/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
+++ b/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
@@ -300,7 +300,7 @@ def validate_accounts(file_name):
 			msg += _("Alternatively, you can download the template and fill your data in.")
 			frappe.throw(msg, title=_("Parent Account Missing"))
 		if account["parent_account"] and accounts_dict.get(account["parent_account"]):
-			accounts_dict[account["parent_account"]]["is_group"] = 1
+			accounts_dict[account["parent_account"]]["is_group"] = '1'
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
```diff
@@ -352,15 +352,15 @@ def get_mandatory_account_types():
 
 def validate_account_types(accounts):
 	account_types_for_ledger = ["Cost of Goods Sold", "Depreciation", "Fixed Asset", "Payable", "Receivable", "Stock Adjustment"]
-	account_types = [accounts[d]["account_type"] for d in accounts if not accounts[d]['is_group'] == 1]
+	account_types = [accounts[d]["account_type"] for d in accounts if not accounts[d]['is_group'] == '1']
```
ankush, line 355: cast user input to integer using `cint` or `int` instead? We don't use string for checkbox fields anywhere else.

nemesis189: Yeah have to discuss more on this issue with Deepesh, can this PR be put on hold?

ankush: You can make it "Draft" 😄

PR #26584, opened 2021-07-21
fix: Production Plan: load document defaults for plan items & remove name column from listview

### LINKED ISSUE #26566: bug: Production Plan Item does not respect default values for 'Include Exploded Items' and 'Make Work Order for Sub Assembly Items'
## Description of the issue
Generated Production Plan Item child table items do not respect default values for 'Include Exploded Items' and 'Make Work Order for Sub Assembly Items' when using the buttons to populate

**Output of `bench version`**
```
13.6
```

1. Set your custom defaults to the opposite to below
2. Create a Production plan with items using the buttons

### Observed result
Always:
'Include Exploded Items' = 1
'Make Work Order for Sub Assembly Items' = 0

### Expected result
Set defaults are applied

nameduser0:
fixes #26566

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -194,7 +194,6 @@ class ProductionPlan(Document):
 					})
 
 			pi = self.append('po_items', {
-				'include_exploded_items': 1,
 				'warehouse': data.warehouse,
 				'item_code': data.item_code,
 				'description': data.description or item_details.description,
@@ -205,6 +204,7 @@ class ProductionPlan(Document):
 				'planned_start_date': now_datetime(),
 				'product_bundle_item': data.parent_item
 			})
+			pi._set_defaults()
 
 			if self.get_items_from == "Sales Order":
 				pi.sales_order = data.parent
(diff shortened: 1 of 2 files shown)


nameduser0:
@rmehta @umairsy As usual, valuable contributions ignored and marked stale

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -205,6 +204,7 @@ def add_items(self, items):
 				'planned_start_date': now_datetime(),
 				'product_bundle_item': data.parent_item
 			})
+			pi._set_defaults()
```
ankush, line 207: The change is acceptable, I am just wondering why this isn't handled at the framework level already. I never got time to investigate fully, hence unmerged. 😅

ankush (approved):

erpnext/manufacturing/doctype/production_plan/production_plan.py
ankush: Also `_set_default` is not used anywhere in ERPNext because as per convention `_method` is supposed to be internal and can be changed without any warning.

nameduser0:

ankush:
Merging this for now.... will revisit once we have a better solution for setting defaults in these cases.

PR #26585, opened 2021-07-21
fix: incorrect amount in work order required items table. 

ankush:
Problem: function used in BOM wasn't computing amount as per `qty` parameter. 
Steps to reproduce:
- Create a BOM
- Create Work Order with more than 1 qty
- Amount won't be correct in Work Order required items table. 

Screenshot 2021-07-21 at 4 33 41 PM

This has no impact on stock/accounting. It's just shown there for reference.


ankush:
Test failures are unrelated. Mostly due to doc rename PR on Frappe.

PR #26589, opened 2021-07-21
feat: add voucher-specific data to DATEV export

barredterra:
In ERPNext there is a report for german accountants, called "DATEV", that can be downloaded in a very special CSV format.

From discussion with our accountant it turned out that he needs more data to make sense of our transactions.

This PR adds some columns to the DATEV export, with additional information such as number, date and due date of the corresponding receipt.

The `get_transactions` method was refactored to allow joining additional tables for some voucher types.


nabinhait (approved):

nabinhait:
@barredterra Please send another PR to develop branch

Issue #26321, opened 2021-07-03
BUG: Migration Error from v12 to v13 (opportunity Unknown column 'first_response_time')


FHenry:
File "/workspace/development/frappe-bench/apps/erpnext/erpnext/patches/v13_0/rename_issue_doctype_fields.py", line 47, in execute
    frappe.db.set_value('Opportunity', entry.name, 'first_response_time', mins_to_first_response, update_modified=False)
  File "/workspace/development/frappe-bench/apps/frappe/frappe/database/database.py", line 667, in set_value
    values, debug=debug)
  File "/workspace/development/frappe-bench/apps/frappe/frappe/database/database.py", line 146, in sql
    self._cursor.execute(query, values)
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/cursors.py", line 148, in execute
    result = self._query(query)
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/cursors.py", line 310, in _query
    conn.query(q)
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/connections.py", line 548, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/connections.py", line 775, in _read_query_result
    result.read()
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/connections.py", line 1156, in read
    first_packet = self.connection._read_packet()
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/connections.py", line 725, in _read_packet
    packet.raise_for_error()
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/protocol.py", line 221, in raise_for_error
    err.raise_mysql_exception(self._data)
  File "/workspace/development/frappe-bench/env/lib/python3.7/site-packages/pymysql/err.py", line 143, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.OperationalError: (1054, "Unknown column 'first_response_time' in 'field list'")

FHenry:
New ERPNext v12.22.0 fresh installation without any customisation in french (don't know of it's the root cause)
login, complete installation, go to create opportunity, save it (change default type and stage because translated to do works here it's another problem)

set logging : 2 into common_site_confing.js

[...]

PR #26591, opened 2021-07-21
fix: migration v12 to13 problem opportunity

FHenry:
fix: #26321
into erpnext/crm/doctype/opportunity/opportunity.json the attribut modified is the same in v12.22.0 as in v13.7.0
The result is on apps/erpnext/erpnext/patches/v13_0/rename_issue_doctype_fields.py on
frappe.reload_doc('crm', 'doctype', 'opportunity')
it return false, and the tabOpporunity is not altered with ALTER TABLE xxx MODIFY

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.json
+++ b/erpnext/crm/doctype/opportunity/opportunity.json
@@ -430,7 +430,7 @@
  "icon": "fa fa-info-sign",
  "idx": 195,
  "links": [],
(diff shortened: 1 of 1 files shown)


FHenry:
just a ping to remove stale bot inactive. And the bug still occurs, and this fix is a good one

ankush:
Fixed by https://github.com/frappe/erpnext/pull/26668

It will be part of next release.

PR #26593, opened 2021-07-22
 fix: Price list rate not fetched for return sales invoice fixed

wikiodisha1:
When a sales invoice is created with 'Is Return' checked and a price list is selected(other than standard), the items added to the child table wouldn't auto fetch the price list rate. Fixed it by removing the condition that prevented it.

Steps:

   create sales invoice with Is Return checked
   select a custom price list
   add items and check its rate

port for : #26559


AfshanKhan (dismissed):

AfshanKhan:
@wikiodisha1 Please fix slider

PR #26600, opened 2021-07-22
fix: incorrect bom name

rohitwaghchaure:
**Issue**

Screenshot 2021-07-22 at 3 38 34 PM

Steps to replicate issue

1. Make new bom > select item > change conversion factor manually in the BOM Item (child table)
2. Save the BOM, you will see the incorrect name

Resolves: https://frappe.io/app/issue/ISS-21-22-03745

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -83,7 +83,7 @@ frappe.ui.form.on("BOM", {
 
 		if (!frm.doc.__islocal && frm.doc.docstatus<2) {
 			frm.add_custom_button(__("Update Cost"), function() {
-				frm.events.update_cost(frm);
+				frm.events.update_cost(frm, true);
 			});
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -325,6 +325,7 @@ frappe.ui.form.on("BOM", {
 			freeze: true,
 			args: {
 				update_parent: true,
+				save:false,
```
ankush: This will break the "Update cost' after submit as it won't commit the change. 

https://github.com/frappe/erpnext/blob/4f1f039145f401f6b7f601dda7855730d445379c/erpnext/manufacturing/doctype/bom/bom.py#L313-L314

rohitwaghchaure: Fixed

ankush (approved):
Can reproduce and fix also works as expected.

PR #26604, opened 2021-07-23
fix: Salary component account filter

pateljannat:
1. When creating a salary component, we select an account in the Accounts table.
2. For salary component type Earning, only Liability accounts are needed, and for salary component deduction only Expense accounts are needed.
3. Based on these conditions, added filters when selecting an account.

Screenshot 2021-07-23 at 12 43 24 PM


ruchamahabal:
Unrelated test case failing. Hence merged

PR #26605, opened 2021-07-23
fix: Salary component account filter

pateljannat:
Backport of https://github.com/frappe/erpnext/pull/26604

### DIFF
--- a/erpnext/payroll/doctype/salary_component/salary_component.js
+++ b/erpnext/payroll/doctype/salary_component/salary_component.js
@@ -4,9 +4,9 @@
 frappe.ui.form.on('Salary Component', {
 	setup: function(frm) {
 		frm.set_query("account", "accounts", function(doc, cdt, cdn) {
-			var d = locals[cdt][cdn];
+			let d = frappe.get_doc(cdt, cdn);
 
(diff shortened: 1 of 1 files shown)


erpnext/payroll/doctype/salary_component/salary_component.js
```diff
@@ -5,10 +5,17 @@ frappe.ui.form.on('Salary Component', {
 	setup: function(frm) {
 		frm.set_query("account", "accounts", function(doc, cdt, cdn) {
 			var d = locals[cdt][cdn];
+
+			var root_type = "Liability";
```
anupamvs: use var->let

erpnext/payroll/doctype/salary_component/salary_component.js
```diff
@@ -5,10 +5,17 @@ frappe.ui.form.on('Salary Component', {
 	setup: function(frm) {
 		frm.set_query("account", "accounts", function(doc, cdt, cdn) {
 			var d = locals[cdt][cdn];
```
anupamvs: let d = frappe.get_doc(cdt, cdn)

ruchamahabal (approved):

PR #26607, opened 2021-07-23
fix:  updating lead status while customer creation

Anurag810:
ignore permission while updating lead status while customer creation.
FhpzWyn

### DIFF
--- a/erpnext/selling/doctype/customer/customer.py
+++ b/erpnext/selling/doctype/customer/customer.py
@@ -136,7 +136,7 @@ class Customer(TransactionBase):
 		if self.lead_name:
 			lead = frappe.get_doc('Lead', self.lead_name)
 			lead.status = 'Converted'
-			lead.save()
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/customer/customer.py
```diff
@@ -136,7 +136,7 @@ def update_lead_status(self):
 		if self.lead_name:
 			lead = frappe.get_doc('Lead', self.lead_name)
 			lead.status = 'Converted'
-			lead.save()
+			lead.save(ignore_permissions=True)
```
anupamvs: instead of getting doc and saving. You can use frappe.db.set_value

Anurag810: Done

Anurag810:
@pateljannat please merge

pateljannat (approved):

PR #26609, opened 2021-07-23
fix: syntax error

ankush:


ankush:
closing 
https://github.com/frappe/erpnext/pull/26610

PR #26613, opened 2021-07-23
fix: wrong operation time in Work Order

rohitwaghchaure:
**Issue**
Created BOM 100 Qty with operation time as 100
Screenshot 2021-07-23 at 4 44 45 PM

Then created work order against the BOM and system has calculated incorrect operation time
Screenshot 2021-07-23 at 4 44 12 PM

**After Fix**

Screenshot 2021-07-23 at 4 43 38 PM

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.py
+++ b/erpnext/manufacturing/doctype/work_order/work_order.py
@@ -487,21 +487,20 @@ class WorkOrder(Document):
 			return
 
 		operations = []
-		if not self.use_multi_level_bom:
-			bom_qty = frappe.db.get_value("BOM", self.bom_no, "quantity")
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -497,7 +497,7 @@ def _get_operations(bom_no, qty=1):
 
 			for d in bom_traversal:
 				if d.is_bom:
-					operations.extend(_get_operations(d.name, qty=d.exploded_qty))
+					operations.extend(_get_operations(d.name, qty=d.exploded_qty/bom_qty))
```
ankush: this would've broken nested operations as `exploded_qty` already considers BOM_qty.

Changed the logic to:
- If multi-level BOM => Fetch and add all child operation
- Add all top-level operations regardless of multi-level BOM.

PR #26615, opened 2021-07-23
fix: employee status server-side validation

ruchamahabal:
Employees with the "Inactive" status are not available for selection in transactions, not fetched in payroll or bulk attendance tool, etc. but if a record is auto-created or created from the employee link dashboard, the document can be saved. Added a server-side validation which is triggered on relevant transactions having Employee link field.

Allowing suspended employees in transactions because some organizations still process payroll, create additional salary, attendance records, etc. for a Suspended employee. So restricting creation for suspended employees can be left to their choice.


ruchamahabal:
PR already opened for semgrep fixes: #25901

PR #26619, opened 2021-07-23
fix: Reset weight_per_unit on replacing Item

GangaManoj:
### Issue:

On changing an Item in a table, its weight_per_unit doesn't get reset.

<details>
<summary>Steps to replicate issue</summary>

- Create a new Sales Order.
- Enter an Item with a default weight_per_unit.
- Replace this Item with another Item.
- Check the weight_per_unit now. You'll find that it hasn't changed.
</details>

### Fix:

On replacing an Item, the weight_per_unit field gets reset to the new Item's default weight_per_unit(or zero, if it doesn't have one).

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -312,7 +312,7 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 		"transaction_date": args.get("transaction_date"),
 		"against_blanket_order": args.get("against_blanket_order"),
 		"bom_no": item.get("default_bom"),
-		"weight_per_unit": args.get("weight_per_unit") or item.get("weight_per_unit"),
+		"weight_per_unit": item.get("weight_per_unit"),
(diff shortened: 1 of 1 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -312,7 +312,7 @@ def get_basic_details(args, item, overwrite_warehouse=True):
 		"transaction_date": args.get("transaction_date"),
 		"against_blanket_order": args.get("against_blanket_order"),
 		"bom_no": item.get("default_bom"),
-		"weight_per_unit": args.get("weight_per_unit") or item.get("weight_per_unit"),
+		"weight_per_unit": item.get("weight_per_unit"),
 		"weight_uom": args.get("weight_uom") or item.get("weight_uom")
```
ankush:
```suggestion
		"weight_uom": item.get("weight_uom")
```

Looks like weight UOM has same problem.

ankush (changes requested):

ankush (approved):

PR #26624, opened 2021-07-25
fix: JWT decoding error fix

harshpwctech:
This is a fix for the PyJWT package for python 3.9 where the decoding function expects the options of "verify_signature" as a dict.


ankush (approved):
LGTM. 

Ref:

- frappe/frappe updated package version in this commit: https://github.com/frappe/frappe/commit/165ff8e1bfe1a8e94b9995626ea5dd188b2e23f2 
- Changelog: https://pyjwt.readthedocs.io/en/latest/changelog.html

ankush:
This doesn't require backporting, as only `develop` version was upgraded. Right? @harshpwctech

PR #26638, opened 2021-07-26
fix(e-invoicing): cannot cancel invoice if IRN cancelled on portal

nextchamp-saqib:
E-Invoice Cancellation Scenario
- A user cancelled the e-invoice on the portal and didn't cancel in ERPNext
- After 24 hours he tried cancelling the Sales Invoice which throws the error "Cancel IRN before cancelling the Sales Invoice"
- Now, if he tries to Cancel IRN using ERPNext, the GST Portal validation doesn't allow the cancellation and throws the error
- Ultimately, the user cannot cancel the Sales Invoice now

Fix:
- Allow user to check IRN Cancelled checkbox manually in such cases.


erpnext/patches/v12_0/show_einvoice_irn_cancelled_field.py
```diff
@@ -0,0 +1,12 @@
...
+import frappe
+
+def execute():
+	company = frappe.get_all('Company', filters = {'country': 'India'})
+	if not company:
+		return
+
+	irn_cancelled_field = frappe.db.exists('Custom Field', {'dt': 'Sales Invoice', 'fieldname': 'irn_cancelled'})
+	if irn_cancelled_field:
+		frappe.db.set_value('Custom Field', irn_cancelled_field, 'depends_on', '')
```
nextchamp-saqib:
```suggestion
		frappe.db.set_value('Custom Field', irn_cancelled_field, 'depends_on', 'eval: doc.irn')
```

erpnext/regional/india/setup.py
```diff
@@ -457,7 +457,7 @@ def make_custom_fields(update=True):
 			depends_on='eval:in_list(["Registered Regular", "SEZ", "Overseas", "Deemed Export"], doc.gst_category) && doc.irn_cancelled === 0'),
 
 		dict(fieldname='irn_cancelled', label='IRN Cancelled', fieldtype='Check', no_copy=1, print_hide=1,
-			depends_on='eval:(doc.irn_cancelled === 1)', read_only=1, allow_on_submit=1, insert_after='customer'),
+			allow_on_submit=1, insert_after='customer'),
```
nextchamp-saqib:
```suggestion
			depends_on='eval: doc.irn', allow_on_submit=1, insert_after='customer'),
```

PR #26661, opened 2021-07-27
feat: Enhancements in Tax Withholding Category

deepeshgarg007:
This PR adds some checks in the tax withholding category to make the Tax withholding process more flexible and manageable.

image

For Eg for TDS 194Q as per Indian Tax law which has come into effect from 1st July 2021 if the aggregate value of goods bought from a supplier exceeds 50 Lakhs (Cumulative Threshold) then a 10% Tax has to be withheld for the amount exceeding the cumulative threshold. Following checks have been added to provision for this

1. **Consider Entire Party Ledger Amount Check**
Enabling this check will also consider Purchase Invoices not having the "Apply Tax Withholding" checkbox checked. Since this provision was introduced on 1st July old invoices in this fiscal year might not have the Apply Tax Withholding check enabled for all invoices

2.  **Only Deduct Tax On Excess Amount**
Currently, tax is withheld for the entire party ledger amount. In some cases, tax has to be withheld only for the amount exceeding the cumulative threshold and not on the entire party ledger amount. Enabling this check will deduct tax only on the amount exceeding the cumulative threshold

3.  **Round Off Tax Amount**
In many countries, the withheld tax amount has to be rounded off to the nearest integer. Enabling this check will round off the deducted tax amount to the nearest integer

#no-docs


nabinhait (approved):

erpnext/accounts/doctype/tax_withholding_category/tax_withholding_category.py
```diff
@@ -235,10 +238,15 @@ def get_deducted_tax(taxable_vouchers, fiscal_year, tax_details):
...
+		'docstatus': 1
+	}
 
-	supp_credit_amt = frappe.db.get_value('Purchase Invoice', {
-		'name': ('in', vouchers), 'docstatus': 1, 'apply_tds': 1
-	}, 'sum(net_total)') or 0.0
+	if not cint(tax_details.consider_party_ledger_amount):
+		invoice_filters.update({'apply_tds': 1})
+
+	supp_credit_amt = frappe.db.get_value('Purchase Invoice', invoice_filters, 'sum(net_total)') or 0.0
```
surajshetty3416, line 249: `supplier_credit_amt`?

PR #26665, opened 2021-07-27
feat: don't recompute taxes

ankush:
Use case:

- When using API / integrations there will be cases when computed tax details are available but there's no simple way to add them. 
- Adding this flag allows "Actual" tax type but also lets integrations define how to distribute taxes using the item_wise_tax_details json.

`no-docs` not intended for users.


ankush:
failing tests are from doc rename PR on frappe:develop

deepeshgarg007 (approved):

PR #26669, opened 2021-07-27
fix: reload Manufacturing Settings in patch

barredterra:
### Reproduce

Restore a v12 backup into a v13 bench and run `bench migrate`.

### Error

```
InvalidColumnName: Invalid Fieldname: <b>make_serial_no_batch_from_work_order</b>
```

### Fix

Reload **Manufacturing Settings** doctype


ankush:
duplicate of https://github.com/frappe/erpnext/pull/26641 😄

barredterra:
@ankush I should really check the existing patches before writing my own 😆

PR #26680, opened 2021-07-27
fix: added progress bar in repost item valuation

rohitwaghchaure:
backport https://github.com/frappe/erpnext/pull/26630

### DIFF
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -127,30 +127,26 @@ def make_entry(args, allow_negative_stock=False, via_landed_cost_voucher=False):
 	sle.submit()
 	return sle
 
-def repost_future_sle(args=None, voucher_type=None, voucher_no=None, allow_negative_stock=None, via_landed_cost_voucher=False):
+def repost_future_sle(args=None, voucher_type=None, voucher_no=None, allow_negative_stock=None, via_landed_cost_voucher=False, doc=None):
 	if not args and voucher_type and voucher_no:
-		args = get_args_for_voucher(voucher_type, voucher_no)
+		args = get_items_to_be_repost(voucher_type, voucher_no, doc)
 
-	distinct_item_warehouses = {}
-	for i, d in enumerate(args):
-		distinct_item_warehouses.setdefault((d.item_code, d.warehouse), frappe._dict({
-			"reposting_status": False,
-			"sle": d,
-			"args_idx": i
-		}))
+	distinct_item_warehouses = get_distinct_item_warehouse(args, doc)
 
-	i = 0
+	i = get_current_index(doc) or 0
 	while i < len(args):
+		validate_item_warehouse(args[i])
+
 		obj = update_entries_after({
-			"item_code": args[i].item_code,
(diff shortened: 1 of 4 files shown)


erpnext/stock/stock_ledger.py
```diff
@@ -127,30 +127,24 @@ def make_entry(args, allow_negative_stock=False, via_landed_cost_voucher=False):
 	sle.submit()
 	return sle
 
-def repost_future_sle(args=None, voucher_type=None, voucher_no=None, allow_negative_stock=None, via_landed_cost_voucher=False):
+def repost_future_sle(args=None, doc=None, voucher_type=None, voucher_no=None, allow_negative_stock=None, via_landed_cost_voucher=False):
```
```suggestion
def repost_future_sle(args=None, voucher_type=None, voucher_no=None, allow_negative_stock=None, via_landed_cost_voucher=False, doc=None):
```

New kwarg should be last. Possibility of breaking someone's function call that's using positional args.

erpnext/stock/stock_ledger.py
```diff
@@ -127,30 +127,24 @@ def make_entry(args, allow_negative_stock=False, via_landed_cost_voucher=False):
...
 	while i < len(args):
 		obj = update_entries_after({
-			"item_code": args[i].item_code,
-			"warehouse": args[i].warehouse,
-			"posting_date": args[i].posting_date,
-			"posting_time": args[i].posting_time,
+			"item_code": args[i].get('item_code'),
+			"warehouse": args[i].get('warehouse'),
+			"posting_date": args[i].get('posting_date'),
+			"posting_time": args[i].get('posting_time'),
```
ankush: Any particular need for this change? 
IF any of these are None, `.get` will propagate `None` which would eventually fail, instead of failing early.

erpnext/stock/stock_ledger.py
```diff
@@ -159,18 +153,61 @@ def repost_future_sle(args=None, voucher_type=None, voucher_no=None, allow_negat
 					args.append(data.sle)
 				elif data.sle_changed and not data.reposting_status:
 					args[data.args_idx] = data.sle
-				
+
 				data.sle_changed = False
 		i += 1
 
-def get_args_for_voucher(voucher_type, voucher_no):
```
ankush, line 166: There might be someone using this as its public functionality. 

how about adding it back like this for backward compatibility :

```python
def get_args_for_voucher(voucher_type, voucher_no):
	return get_items_to_be_repost(voucher_type, voucher_no)
```

ankush (changes requested):

erpnext/stock/stock_ledger.py
rohitwaghchaure: Fixed

erpnext/stock/stock_ledger.py
rohitwaghchaure: Added validation

erpnext/stock/stock_ledger.py
rohitwaghchaure: get_args_for_voucher is not used anywhere in the code and it's normal get_all earlier

ankush (approved):

PR #26700, opened 2021-07-29
refactor!: drop shopify integration from ERPNext

ankush:
The integration is moved to a separate app with additional enhancements.

The app is still FOSS and licensed under the same license as ERPNext. Any
data migration required is taken care of after installing the app and enabling it.

New app: https://github.com/frappe/ecommerce_integrations

This is a **BREAKING CHANGE**. Do not backport this to v13.


Alchez:
@ankush, question on the new app: is that dependent on a particular version of ERPNext? We're using v12, and we're wondering if we could use it.

ankush:
@Alchez  only major requirement I can imagine in Python 3.6+. 

I haven't tested on v12, but ideally, it should work barring few changes done to accommodate correct item wise taxation (which is only present on v13 right now, I guess I should get it backported)

Let me know if you find any issue.

Alchez:
@ankush, ah gotcha.

We'll spin it up on a dev instance and test it out. We're currently using our own fork over at https://github.com/ParsimonyGit/shopify_integration, but if the official one works for v12, we might think about archiving ours.

Thanks!

PR #26725, opened 2021-07-29
fix: Multiple fixes in payment entry

deepeshgarg007:
- Added validation for Received Amount if Paid Account and Received Account Currency are same

- Validation while unlinking references on Invoice cancellation

- Total Allocated and Unallocated amount fixes when allocating standalone credit note in a Payment

For example, let's say a Payment of 1000 has been made against and a standalone credit note of 300 is also allocated in the same payment entry against some other invoice 

For Example:
These are the references
Sales Invoice 1: 1000
Sales Invoice 2: 300
Sales Invoice 3: -300 (Credit)

If first Invoice 2 is cancelled things are fine, but if Invoice 1 is cancelled after it then things go wrong as the payment type of the entry is "Receive" but the only remaining invoice is credit note, due to which and Allocated and Unallocated amount goes wrong , added a validation to prevent this

image


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -118,6 +121,11 @@ def set_bank_account_data(self):
 			if not self.get(field):
 				self.set(field, bank_data.account)
 
+	def validate_payment_type_with_outstanding(self):
+		total_outstanding = sum(d.allocated_amount for d in self.get('references'))
+		if total_outstanding < 0 and self.party_type == 'Customer' and self.payment_type == 'Receive':
+			frappe.throw(_("Cannot receive from customer against negative outstanding"), title=_("Incorrect Payment Type"))
```
nabinhait, line 127: We should allow receiving payments, but in that payment entry earlier credit notes should not be adjusted

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -468,13 +478,22 @@ def apply_taxes(self):
...
 		self.set_unallocated_amount()
 		self.set_difference_amount()
 
+	def validate_amounts(self):
+		self.validate_received_amount()
+	
+	def validate_received_amount(self):
+		if self.paid_from_account_currency == self.paid_to_account_currency:
+			if self.paid_amount != self.received_amount:
+				frappe.throw(_("Received Amount cannot be greater than Paid Amount"))
```
nabinhait:
```suggestion
				frappe.throw(_("Received Amount should be same as Paid Amount"))
```

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -468,13 +478,22 @@ def apply_taxes(self):
 	def set_amounts(self):
 		self.set_received_amount()
 		self.set_amounts_in_company_currency()
-		self.set_amounts_after_tax()
```
nabinhait, line 471: `set_amounts` is also called from `get_payment_entry` function. `set_amounts_after_tax ` function should also be called separately there?

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/test_payment_entry.py
```diff
@@ -107,7 +107,7 @@ def test_payment_entry_against_si_usd_to_usd(self):
 		pe = get_payment_entry("Sales Invoice", si.name, bank_account="_Test Bank USD - _TC")
 		pe.reference_no = "1"
 		pe.reference_date = "2016-01-01"
-		pe.target_exchange_rate = 50
+		pe.source_exchange_rate = 50
```
nabinhait, line 110: Ideally, it should work irrespective of we set source_exchange_rate or target_exchange_rate.

nabinhait (changes requested):

erpnext/accounts/doctype/payment_entry/payment_entry.py
deepeshgarg007: I have not blocked receiving payments, I didn't completely understand this. Can you please elaborate

erpnext/accounts/doctype/payment_entry/payment_entry.py
deepeshgarg007: set_amounts is called from `validate`. Why call from `get_payment_entry`. Can't think of any reason

erpnext/accounts/doctype/payment_entry/test_payment_entry.py
deepeshgarg007: There should be a base or reference for this, right? User may change the exchange rate manually and the visible option to the user in the form is source hence I used that as a reference. For test cases, we have to set one so that the values we are checking in test match the expected value.

ushashmi:
@deepeshgarg007 since you are working on payment entry. I would like to draw your attention on below multicurrency bug; 

https://discuss.erpnext.com/t/payment-entry-error-other-currency-payment/79123

Appreciate if you can also consider this in your fix

deepeshgarg007:
> @deepeshgarg007 since you are working on payment entry. I would like to draw your attention on below multicurrency bug;
> 
> https://discuss.erpnext.com/t/payment-entry-error-other-currency-payment/79123
> 
> Appreciate if you can also consider this in your fix

@ushashmi this bug has been fixed and also release via PR https://github.com/frappe/erpnext/pull/26980

nabinhait (approved):

PR #26729, opened 2021-07-29
perf: Optimize get_attribute_filters

netchampfaris:
`get_attribute_fitlers` did stupid computations just to fetch a list of attribute values. A simple query is faster. Tested on real site with lots of items.

### Before

image

### After
image

Paired with @ankush and @shadrak98 

@marination Please make sure to get this in E-commerce refactor.


erpnext/shopping_cart/filters.py
```diff
@@ -55,37 +55,27 @@ def get_field_filters(self):
...
-
-		return valid_attributes
+		result = frappe.db.sql(
+			"""
+			select
+				distinct attribute, attribute_value
+			from
+				`tabItem Variant Attribute`
+			where
+				attribute in %(attributes)s
```
ankush, line 72: Test failing here because `attributes` is empty which results in a malformed sql query. `in ()`

Screenshot 2021-08-11 at 8 23 39 PM

ankush (changes requested):

nabinhait (approved):

PR #26730, opened 2021-07-29
perf: speed up validate_account_for_perpetual_inventory

ahmadpak:
In the file general_ledger.py method **validate_account_for_perpetual_inventory** should only run if the doctype is **Journal Entry**

This issue causing unwanted calculations resulting in delayed submission of related documents

### DIFF
--- a/erpnext/accounts/general_ledger.py
+++ b/erpnext/accounts/general_ledger.py
@@ -147,7 +147,7 @@ def make_entry(args, adv_adj, update_outstanding, from_repost=False):
 	gle.submit()
 
 def validate_account_for_perpetual_inventory(gl_map):
-	if cint(erpnext.is_perpetual_inventory_enabled(gl_map[0].company)):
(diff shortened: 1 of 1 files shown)


erpnext/accounts/general_ledger.py
```diff
@@ -147,7 +147,7 @@ def make_entry(args, adv_adj, update_outstanding, from_repost=False):
 	gle.submit()
 
 def validate_account_for_perpetual_inventory(gl_map):
-	if cint(erpnext.is_perpetual_inventory_enabled(gl_map[0].company)):
+	if cint(erpnext.is_perpetual_inventory_enabled(gl_map[0].company)) and gl_map[0].voucher_type=="Journal Entry":
```
ankush, line 150:
```suggestion
	if gl_map[0].voucher_type=="Journal Entry" and cint(erpnext.is_perpetual_inventory_enabled(gl_map[0].company)):
```


Reduce one more function call by re-ordering conditional.

ankush (approved):
LGTM, this is already done on develop branch, no harm in doing on v12 too.

PR #26735, opened 2021-07-30
fix: COGS account in purchase receipt

rohitwaghchaure:
Due to negative transaction we book COGS in purchase receipt, if default cogs account is not set then system throw the error. This PR Will fix the validation message and book difference amount in Stock Received but Not Billed account.


deepeshgarg007:

deepeshgarg007:

PR #26737, opened 2021-07-30
fix: Payment Reconciliation party validation issue

nemesis189:
**Problem :**
During Payment Reconciliation, while creating new journal entries in the process the party validation caused problems for customers that were once active but now disabled.
So this means all the parties in the journal entry associated with the reconciliation needed to be active and not disabled, which is not ideal.

**Solution :**
Added a flag to ignore the `validate_party_frozen_disabled` check during the validation.

**Steps:**
1. create a journal entries with 2 different customers( any party)
    

2. submit the entry
3. disable one of the parties(in this case i disabled cust2)
4. Go to Payment Reconciliation, set the mandatory fields and click 'Get Unreconciled Entries', you should get the journal entry you made recently
    

5. Try and reconcile payment against that Journal entry, you should not get any errors now


AfshanKhan:
@nemesis189 the backports have failed. You will need to create them manually.

PR #26743, opened 2021-07-30
patch: Delete all orphaned tables

GangaManoj:
Patch for https://github.com/frappe/erpnext/pull/26239.


erpnext/patches/v13_0/delete_orphaned_tables.py
```diff
@@ -0,0 +1,68 @@
...
+        Since Transaction Deletion Record resets the naming series after deletion,
+        it allows the creation of new docs with the same names as the deleted ones.
+    """
+
+    parent_creation_time = frappe.db.get_value(doc['parenttype'], doc['parent'], 'creation')
+    child_creation_time = frappe.db.get_value(doctype, doc, 'creation')
+
+    if getdate(parent_creation_time) > getdate(child_creation_time):
+        return True
+    return False
```
ankush:
```suggestion
    return getdate(parent_creation_time) > getdate(child_creation_time)
```

erpnext/patches/v13_0/delete_orphaned_tables.py
```diff
@@ -0,0 +1,68 @@
...
+        for doctype in tdr_doc.doctypes:
+            if (doctype.doctype_name not in affected_doctypes) and is_not_child_table(doctype.doctype_name):
+                affected_doctypes.append(doctype.doctype_name)
+
+    return affected_doctypes
+
+def is_not_child_table(doctype):
+    if frappe.get_value('DocType', doctype, 'istable'):
+        return False
+    return True
```
ankush:
```suggestion
    return not bool(frappe.get_value('DocType', doctype, 'istable'))
```

erpnext/patches/v13_0/delete_orphaned_tables.py
```diff
@@ -0,0 +1,68 @@
...
+def get_affected_doctypes():
+    affected_doctypes = []
+
+    tdr_docs = frappe.get_all('Transaction Deletion Record', pluck="name")
+    
+    for tdr in tdr_docs:
+        tdr_doc = frappe.get_doc("Transaction Deletion Record", tdr)
+
+        for doctype in tdr_doc.doctypes:
+            if (doctype.doctype_name not in affected_doctypes) and is_not_child_table(doctype.doctype_name):
```
ankush: Instead of doing this just remove duplicates at the end or use `set`.  `not in <list>` gets expensive as list grows larger.

erpnext/patches/v13_0/delete_orphaned_tables.py
```diff
@@ -0,0 +1,68 @@
...
+
+def check_for_new_doc_with_same_name_as_deleted_parent(doc, doctype):
+    """
+        Compares creation times of parent and child docs.
+        Since Transaction Deletion Record resets the naming series after deletion,
+        it allows the creation of new docs with the same names as the deleted ones.
+    """
+
+    parent_creation_time = frappe.db.get_value(doc['parenttype'], doc['parent'], 'creation')
+    child_creation_time = frappe.db.get_value(doctype, doc, 'creation')
```
ankush: These could probably be queried along with other data from the loop where this function gets called. There could be 1000s of orphaned docs, so these additional queries in loop could be expensive.

erpnext/patches/v13_0/delete_orphaned_tables.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) 2019, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
+
+import frappe
+from frappe.utils import getdate
+
+def execute():
+    if has_deleted_company_transactions():
```
ankush, line 12: You need to do `reload_doc` on TDR, so v12->v13 updates don't fail. Check patch test results.

ankush (changes requested):
Looks correct. Some minor changes are required.

erpnext/patches/v13_0/delete_orphaned_tables.py
GangaManoj: fix: Make code more concise

erpnext/patches/v13_0/delete_orphaned_tables.py
GangaManoj: Done.

erpnext/patches/v13_0/delete_orphaned_tables.py
GangaManoj: Done.

erpnext/patches/v13_0/delete_orphaned_tables.py
GangaManoj: I did the second, but I don't think `parent_creation_time` can be fetched anywhere else.

nextchamp-saqib (approved):

PR #26237, opened 2021-06-28
feat: Increase number of supported currency exchanges

rtdany10:
Switch from frankfurter.app to exchangerate.host to accommodate more currency usage.
Closes #25603

#no-docs

PR #26763, opened 2021-08-01
feat: Increase number of supported currency exchanges

rtdany10:
Switch from frankfurter.app to exchangerate.host to accommodate more currency usage.
Backport of #26237 

#no-docs


rtdany10:
@nabinhait

rtdany10:
Unrelated test failing.
@nabinhait can we merge this PR?

nabinhait (approved):

barredterra:
@rtdany10 maybe we should make this configurable. Currently it's difficult for a user to know where the exchange rates are coming from, and the API is changed frequently. It is a legal requirement in Europe to make sure that exchange rates are trusted and accurate, and that the source is known.

rtdany10:
> @rtdany10 maybe we should make this configurable. Currently it's difficult for a user to know where the exchange rates are coming from, and the API is changed frequently. It is a legal requirement in Europe to make sure that exchange rates are trusted and accurate, and that the source is known.

I agree to the fact that it should be configurable. Although, I don't believe APIs were changed frequently. Even though we switched the API endpoint by this pull request, the source of both the service remain the same, European Central Bank.

sagarvora:
**exchangerate.host** requires an access token now. perhaps we should revert to frankfurter.app and backport configurability.

PR #24206, opened 2020-12-25
fix: Tax calculation for Recurring additional salary

Anurag810:
Before: Tax calculation of Recurring Additional Salary was calculated on monthly basis.
Suppose, There is a Recurring Additional salary for three months, then every month the amount(monthly) is contributed to yearly taxable earning. then according to tax slabs, it calculate TDS/Income tax

Now: The whole Amount for three months is taken at once and contributed to yearly Taxables earning and tax is calculated accordingly. As taxes are on CTC.

PR #26765, opened 2021-08-01
Revert "fix: Tax calculation for Recurring additional salary"

nabinhait:
Reverts frappe/erpnext#24206

Reverting this because it does not calculate taxable income properly in case of overwritten additional salary

### DIFF
--- a/erpnext/payroll/doctype/additional_salary/additional_salary.py
+++ b/erpnext/payroll/doctype/additional_salary/additional_salary.py
@@ -112,11 +112,11 @@ class AdditionalSalary(Document):
 		no_of_days = date_diff(getdate(end_date), getdate(start_date)) + 1
 		return amount_per_day * no_of_days
 
-@frappe.whitelist()
 def get_additional_salaries(employee, start_date, end_date, component_type):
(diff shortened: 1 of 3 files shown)


erpnext/payroll/doctype/additional_salary/additional_salary.py
```diff
@@ -112,11 +112,11 @@ def get_amount(self, sal_start_date, sal_end_date):
 		no_of_days = date_diff(getdate(end_date), getdate(start_date)) + 1
 		return amount_per_day * no_of_days
 
-@frappe.whitelist()
 def get_additional_salaries(employee, start_date, end_date, component_type):
 	additional_salary_list = frappe.db.sql("""
-		select name, salary_component as component, type, amount, overwrite_salary_structure_amount as overwrite,
-		deduct_full_tax_on_selected_payroll_date, is_recurring
+		select name, salary_component as component, type, amount,
```
anupamvs, line 117: Query Formatting https://github.com/frappe/erpnext/wiki/Coding-Standards#use-simple-indenting

PR #26775, opened 2021-08-02
perf: various minor perf fixes for ledger postings

ankush:
- Ledger posting related code is on the "critical path" for most stock/accounting transactions as it generates ledger entries. This usually takes time roughly on the order of size of child tables.
- Changed several methods to use `db.value_cache` for fetching details that won't change during a db transaction, this cache is dropped frequently when `db` instance dies, so it's safe to use. 
- reordered conditionals to avoid queries. 

depends on https://github.com/frappe/frappe/pull/13880 

Outcome: Along with that ^ PR, ~20-25% saving in queries and time while submitting large documents with repeated validations. More stats in linked PR. 

Note to reviewers: review commit by commit, would be easier that way.

### DIFF
--- a/erpnext/accounts/utils.py
+++ b/erpnext/accounts/utils.py
@@ -960,6 +960,9 @@ def get_voucherwise_gl_entries(future_stock_vouchers, posting_date):
 	return gl_entries
 
 def compare_existing_and_expected_gle(existing_gle, expected_gle, precision):
+	if len(existing_gle) != len(expected_gle):
+		return False
+
 	matched = True
 	for entry in expected_gle:
 		account_existed = False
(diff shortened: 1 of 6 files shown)


erpnext/accounts/utils.py
```diff
@@ -959,23 +959,27 @@ def get_voucherwise_gl_entries(future_stock_vouchers, posting_date):
...
-			break
-	return matched
+def compare_existing_and_expected_gle(existing_gle, expected_gle, precision: int) -> bool:
+	""" Compare existing GLE with expected GLE, if they are same return True else False"""
+	if len(existing_gle) != len(expected_gle):
+		return False
+
+	# Sort GLEs to line them up next to one another for comparing.
+	# Assuming both are same they should be lined up next to each other.
+	comparator = lambda gle: (gle.account, gle.cost_center, gle.credit, gle.debit)
```
ankush: `cost_center` can be None, need to `cstr`

ankush:
Dropped a refactor that would require additional tests for now, will add in separate PR (in next release)

marination (approved):

marination:
Asset, POS and Attendance tests failing on other PRs as well. Not caused by this PR

PR #26652, opened 2021-07-26
feat: add `total_billing_hours` to Sales Invoice

barredterra:
### Problem

Timesheets get fetched into a sales invoice. Currently we calculate only the total billing amount. This makes it hard to add a position "Consulting, 30 hours @ 50 USD". We have to divide the total by our rate or sum the hours manually.

### Solution

Add a field `total_billing_hours`, next to `total_billing_amount`.

### Demo

https://user-images.githubusercontent.com/14891507/127000697-109ce1d9-b579-4631-9478-b30f8ac7f0b9.mov

PR #26783, opened 2021-08-02
feat: add `total_billing_hours` to Sales Invoice

barredterra:
Port of  #26652

### Additional changes

- refactor to keep only one way of adding timesheets to the sales invoice.

    (Previously one function was fetching timesheets from the backend and adding them to the table in the frontend. Another one was calling a backend method that adds timesheets to the table and then refreshes the frontend. Now we use the first approach only.)

- add `from_time` and `to_time` to **Sales Invoice Timesheet**. This is useful for printing timesheets on the **Sales Invoice** for the customer to check.
- order timesheets by `start_time`. Useful for the same reason.
- add some section and column breaks to **Sales Invoice Timesheet**

    Sales Invoice Timesheet

`no-docs`


ruchamahabal:
@barredterra Can you fix the conflict and sider issues? I don't have edit access to this branch. Or you can give Edit Access to maintainers.

barredterra:
Hi @nabinhait, @ruchamahabal mentioned some remaining issues that need to be addressed. Did you fix them yourself?

nabinhait:
> Hi @nabinhait, @ruchamahabal mentioned some remaining issues that need to be addressed. Did you fix them yourself?

Yes

PR #26799, opened 2021-08-03
fix: Customers 'primary_address' not updated automatically (#26798)

### LINKED ISSUE #26798: Customer 'primary_address' not updated when Address updated
## Steps to Reproduce

1. On a Customer, find or create an address:
image

2. Assign this as the Customer's primary address.

image

**Note** : See how the Read-Only field `primary_address` was automatically populated?

3.  Next, edit the Address for that Customer.  Modify the values for Street, City, State, etc.

image

4.  Scroll down on Customer, and re-examine the Primary Address.  See how the values here did **not automatically update.**?

image

## Workaround
* Backspace over the 'Customer Primary Address' field, and blank it.
* Save the customer record.
* Re-select the 'Customer Primary Address' field from the drop-down list.
* Save the customer record again.

## Proposed Solution

I will submit a PR for this Issue.

brian-pond:
Closes #26798

TLDR:  When an Address is updated, and that Address is a Customer's Primary Address, update the Read Only field `customer.primary_address`

### DIFF
--- a/erpnext/accounts/custom/address.py
+++ b/erpnext/accounts/custom/address.py
@@ -1,7 +1,7 @@
 import frappe
 from frappe import _
 from frappe.contacts.doctype.address.address import Address
-from frappe.contacts.doctype.address.address import get_address_templates
+from frappe.contacts.doctype.address.address import get_address_templates, get_address_display
 
 class ERPNextAddress(Address):
 	def validate(self):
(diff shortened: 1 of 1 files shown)


erpnext/accounts/custom/address.py
```diff
@@ -22,6 +22,18 @@ def validate_reference(self):
...
+		"""
+		After Address is updated, update the related 'Primary Address' on Customer.
+		"""
+		address_display = get_address_display(self.as_dict())
+		filters = { "customer_primary_address": self.name }
+		customers = frappe.db.get_all("Customer", filters=filters, as_list=True)
+		for customer_name in customers:
+			doc_customer = frappe.get_doc("Customer", customer_name[0])
+			doc_customer.primary_address = address_display
+			doc_customer.save()
```
anupamvs: use `frappe.db.set_value` or `frappe.db.update`

brian-pond: Why do you believe either of those would be an improvement?

ankush: @PMojito Performance reasons. 

`get_doc` fetches entire document, its child tables etc. Then `save()` goes through all the validations. 

`db.set_value` bypasses all that and updates in a single SQL query.

brian-pond: That's good feedback; change complete!

deepeshgarg007 (approved):

nabinhait (approved):

PR #26803, opened 2021-08-04
fix: Faulty Gl Entry for Asset LCVs

marination:
**Issue:**
- Introduced via https://github.com/frappe/erpnext/pull/25379, both the GL entries created via `add_lcv_gl_entries` in `get_asset_gl_entry` in Purchase Receipt, were crediting into their respective accounts (asset account and expense account) due to which:
  Screenshot 2021-08-04 at 10 27 31 AM

- 80 were the added charges via LCV, 80+80 = 160 was credited.

**Fix:**
- Debit into asset account and credit from expense account

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -501,7 +501,7 @@ class PurchaseReceipt(BuyingController):
 		self.add_gl_entry(gl_entries, expenses_included_in_asset_valuation, item.cost_center, 0.0, flt(item.landed_cost_voucher_amount),
 			remarks, asset_account, project=item.project, item=item)
 
-		self.add_gl_entry(gl_entries, asset_account, item.cost_center, 0.0, flt(item.landed_cost_voucher_amount),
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -501,7 +501,7 @@ def add_lcv_gl_entries(self, item, gl_entries):
 		self.add_gl_entry(gl_entries, expenses_included_in_asset_valuation, item.cost_center, 0.0, flt(item.landed_cost_voucher_amount),
 			remarks, asset_account, project=item.project, item=item)
 
-		self.add_gl_entry(gl_entries, asset_account, item.cost_center, 0.0, flt(item.landed_cost_voucher_amount),
+		self.add_gl_entry(gl_entries, asset_account, item.cost_center, flt(item.landed_cost_voucher_amount), 0.0,
```
ankush: Using kwargs (named arguments) would avoid a similar issue in future.

marination:
Screenshot 2021-08-04 at 2 05 15 PM

ankush (approved):

PR #26829, opened 2021-08-05
fix: bad price list rate on PO from MR from SO

FHenry:
fix #26828

related to :
#26830 version 13 hotfix
#26831 develop

PR #26831, opened 2021-08-05
fix: bad price list rate on PO from MR from SO

### LINKED ISSUE #26828: Wrong buying price on Purchase Order from Material Request from Selling Order
Standard selling price list 
    Item 1  / Rate : 500 €

Standard buying price list 
     Item 1  /  Rate : 0.50 €
     Item 1  / Supplier 1 / Rate : 0.30 €

Create Purchase Order PO0001 for Supplier 1 with Item 1, Qty 1 => Item Rate 0.30 € **OK**

Create Sales Order SAL0001 for Customer 1 with Item 1, Qty 1 => Item Rate 500 € **OK**
Create Material Request MAT0001 from SAL0001
Create Purchase Order PO0002 from MAT0001 (select Supplier 1 on popup) => Item Rate 500 € **KO**, should be 0.30

Create Purchase Order PO0003 for Supplier 1
Use "Get Item from Open Material request", Check MAT0001  => Item Rate 500 € **KO**, should be 0.30

Item 1 rate is 0.50 instead of 1000€

I've track it until what I can, erpnext/stock/get_item_details.py method get_item_details received the Customer 1 in args and get_price_list_rate make bad sql request
select name, price_list_rate, uom
from `tabItem Price` where item_code=%(item_code)s
and price_list=%(price_list)s
and coalesce(uom, '') in ('', %(uom)s) and **customer=%(customer)s** and %(transaction_date)s between
coalesce(valid_from, '2000-01-01') and coalesce(valid_upto, '2500-12-31')
order by valid_from desc, uom desc 
{'item_code': 'Item 1', 'price_list': 'Buying Standard', 'customer': 'Customer 1', 'supplier': 'Supplier 1', 'uom': 'Unit', 'transaction_date': '2021-08-05'}

Test customer instead of supplier

Related to #2964 close in 2015 ?

image

image

https://discuss.erpnext.com/t/sales-order-material-request-purchase-order-no-price-rate-picked-up-from-price-list/72109/2

FHenry:
fix #26828

related to 
#26829 version-12-hotfix
#26830 version-13-hotfix


erpnext/stock/get_item_details.py
```diff
@@ -74,9 +74,17 @@ def get_item_details(args, doc=None, for_validate=False, overwrite_warehouse=Tru
 
 	update_party_blanket_order(args, out)
 
-	
+	# we do not send customer args for price list rate in case of purchase order but we save it for latter use
+	if doc and doc.get('doctype') == 'Purchase Order':
```
deepeshgarg007: This is a generic query and hardcoding doctype here doesn't make much sense. We can pass two more args as `is_buying` and `is_selling` based on the doctype it is called from and ignore parties based on the `is_buying` and `is_selling` flag and use the same in item price query as well. This sounds like a more generic approach to me

deepeshgarg007 (changes requested):

erpnext/stock/get_item_details.py
FHenry: @deepeshgarg007 ok, why not, but there is test on docytype just few lines before (line 60 for exemple). I apply the same method that are existing into existing code.
But anyway I try your way, but in this case shouldn't be better to have  `price_type` parameter pass to get_price_list_rate. I witch case we'll have is_buying and is_selling true/false at the same time ?
There is already  a transaction_type args use line 671. Wo you mind If I use it ?

FHenry: Update PR is on her way

FHenry:
@deepeshgarg007 Could have a look on this PR ? I'll be happy to change the way I do it if you think it must be rewrite. The bug is in version-12 and version-13, I would like to deliver a fix to my customer not too long after he found it.

FHenry:
@deepeshgarg007 could have a look please ?

FHenry:
Hello Sir,
 @deepeshgarg007 @ChillarAnand @pateljannat @ruchamahabal @nextchamp-saqib 

Could somebody some one have a look, this is an actual bug in version-12 version-13 and develop, The fix recommended is here, just please let me know what to do to make this bug fixed into ERPNext core branch

FHenry:
Hello Sir,
@deepeshgarg007 @ChillarAnand @pateljannat @ruchamahabal @nextchamp-saqib just tell what to do. I have actually an specific app to fix this bug that override 3 methods of frappe frappe.model.mapper.map_docs, frappe.model.mapper.make_mapped_doc, erpnext.stock.get_item_details.apply_price_list, I would like to get rid of them.

PR #26847, opened 2021-08-09
fix: update get_exchange_rate for test mode

rtdany10:
Patching `get_exchange_rate` to not make API calls on test mode.
https://github.com/frappe/erpnext/pull/26237#issuecomment-894381780

What other changes can we make to be sure its fail proof? @ankush 
Maybe return 1.00 instead of 0.00 by default?

#no-docs


rtdany10:
Not running API call and returning 0.00 causes failure of lot more test than the expected.
The api returns `None` for `_Test Currency` and `flt(value)` makes it 0.00.

Maybe instead of not making api call on test mode, can we return 1.00 if the api returns None?

erpnext/setup/utils.py
```diff
@@ -107,6 +107,8 @@ def get_exchange_rate(from_currency, to_currency, transaction_date=None, args=No
 			# expire in 6 hours
 			response.raise_for_status()
 			value = response.json()["result"]
+			if not value and frappe.flags.in_test:
+				return 1.00
```
ankush: Ideally, I would use a library like https://pypi.org/project/responses/ for mocking HTTP response from API, but adding another dependency just for one test doesn't feel justified. 

In tests we need to validate two things:
- Code is creating appropriate requests to the API. 
- Code is acting properly on valid/invalid responses given by API.

You could patch out the `request` library's `get` call and validate both requirements mentioned above. Essentially creating a fake API that validates input and gives output.

rtdany10:
Any way we can avoid implicit calls? I forgot to account that the exchange rate tests itself would fail if we were to return a fixed value when test flag is on.

ankush:
@rtdany10 seems like it would be difficult to patch implicit calls, let's merge this for now and I'll revisit the implicit calls later. 

We've to avoid using "test currency", that's all. More context here: https://github.com/Formicka/exchangerate.host/issues/71

erpnext/setup/doctype/currency_exchange/test_currency_exchange.py
```diff
@@ -37,18 +37,45 @@ def save_new_records(test_records):
...
+			if self.status_code != 200:
+				raise frappe.DoesNotExistError
+
+		def json(self):
+			return self.json_data
 
+	if args[0] == "https://api.exchangerate.host/convert" and kwargs.get('params'):
+		if kwargs['params'].get('date') and kwargs['params'].get('from') and kwargs['params'].get('to'):
+			if test_exchange_values.get(kwargs['params']['date']):
+				return PatchResponse({ 'result': test_exchange_values[kwargs['params']['date']] }, 200)
```
ankush:
```suggestion
				return PatchResponse({'result': test_exchange_values[kwargs['params']['date']]}, 200)
```

erpnext/setup/doctype/currency_exchange/test_currency_exchange.py
```diff
@@ -37,18 +37,45 @@ def save_new_records(test_records):
...
+
+	return PatchResponse({ 'result': None }, 404)
```
ankush:
```suggestion
	return PatchResponse({'result': None}, 404)
```

ankush (dismissed):
some minor style changes required, else LGTM.

PR #26854, opened 2021-08-09
fix: Faulty Gl Entry for Asset LCVs (#26803)

marination:
Backport of https://github.com/frappe/erpnext/pull/26803

**Issue:**
- Introduced via https://github.com/frappe/erpnext/pull/25379, both the GL entries created via `add_lcv_gl_entries` in `get_asset_gl_entry` in Purchase Receipt, were crediting into their respective accounts (asset account and expense account) due to which:
  Screenshot 2021-08-04 at 10 27 31 AM

- 80 were the added charges via LCV, 80+80 = 160 was credited.

**Fix:**
- Debit into asset account and credit from expense account

**To test:**
Create an Fixed Asset Item, enable auto creation of asset
Create Purchase Receipt for this Asset, asset will be auto created and in Draft state.
Now create LCV against this Purchase Receipt, while asset is in Draft state.
Check if LCV updates Asset gross purchase amount and check GL Entry against PR.

### DIFF
--- a/erpnext/stock/doctype/landed_cost_voucher/test_landed_cost_voucher.py
+++ b/erpnext/stock/doctype/landed_cost_voucher/test_landed_cost_voucher.py
@@ -253,6 +253,8 @@ class TestLandedCostVoucher(unittest.TestCase):
 
 	def test_asset_lcv(self):
 		"Check if LCV for an Asset updates the Assets Gross Purchase Amount correctly."
+		frappe.db.set_value("Company", "_Test Company", "capital_work_in_progress_account", "CWIP Account - _TC")
+
 		if not frappe.db.exists("Asset Category", "Computers"):
 			create_asset_category()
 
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/landed_cost_voucher/test_landed_cost_voucher.py
```diff
@@ -250,6 +251,38 @@ def test_multi_currency_lcv(self):
...
+		if not frappe.db.exists("Item", "Macbook Pro"):
+			create_fixed_asset_item()
+
+		pr = make_purchase_receipt(item_code="Macbook Pro", qty=1, rate=50000)
+
+		# check if draft asset was created
+		assets = frappe.db.get_all('Asset', filters={'purchase_receipt': pr.name})
+		self.assertEqual(len(assets), 1)
+
+		frappe.db.set_value("Company", pr.company, "capital_work_in_progress_account", "CWIP Account - _TC")
```
ankush: Move the setting of CWIP account before creating PR? 

Test becomes flaky depending on which tests ran. Example: https://github.com/frappe/erpnext/runs/3277810413

marination: @ankush https://github.com/frappe/erpnext/pull/26873

PR #26872, opened 2021-08-10
fix: Pos return payment mode issue

nemesis189:
**Problem :**
On making a return POS Invoice and selecting a payment mode which is not default, eg. we choose Credit Card instead of Cash(default), on submitting the payment mode gets reset to the default one.

**Solution :**
Split the function the sets the default mop for Sales anad POS Invoices and it triggers only for Sales Invoice.

Steps:
1. Create a Return against a POS Invoice
2. change the mode of payment to a non-default one
3. check the mode of payment after submission


erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -47,7 +47,10 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 
 		if (in_list(["Sales Invoice", "POS Invoice"], this.frm.doc.doctype) && this.frm.doc.is_pos &&
 			this.frm.doc.is_return) {
-			this.update_paid_amount_for_return();
+			if (this.frm.doc.doctype == "Sales Invoice"){
```
nextchamp-saqib:
```suggestion
			if (this.frm.doc.doctype == "Sales Invoice") {
```

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -747,17 +750,15 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 				precision("base_grand_total")
 			);
 		}
-
+
```
nextchamp-saqib:
```suggestion

```

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -747,17 +750,15 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
...
 		}
-
+		
 		this.frm.doc.payments.find(pay => {
 			if (pay.default) {
 				pay.amount = total_amount_to_pay;
 			} else {
 				pay.amount = 0.0
-			}
+			}
```
nextchamp-saqib:
```suggestion
			}
```

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -747,17 +750,15 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 				precision("base_grand_total")
 			);
 		}
-
+		
 		this.frm.doc.payments.find(pay => {
 			if (pay.default) {
 				pay.amount = total_amount_to_pay;
 			} else {
 				pay.amount = 0.0
```
nextchamp-saqib:
```suggestion
				pay.amount = 0.0;
```

AfshanKhan:
@nemesis189 also check these.
https://github.com/frappe/erpnext/pull/26853
https://github.com/frappe/erpnext/pull/26884

erpnext/public/js/controllers/taxes_and_totals.js
```diff
@@ -734,7 +737,7 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 		}
 	}
 
-	update_paid_amount_for_return() {
+	set_total_amount_to_default_mop(){
```
AfshanKhan: In `taxes_and_total.py` there is a function `update_paid_amount_for_return()` which does the same thing. So maybe don't rename the function for consistency.

erpnext/controllers/taxes_and_totals.py
```diff
@@ -595,7 +595,8 @@ def calculate_outstanding_amount(self):
 				self.doc.precision("outstanding_amount"))
 
 			if self.doc.doctype == 'Sales Invoice' and self.doc.get('is_pos') and self.doc.get('is_return'):
-			 	self.update_paid_amount_for_return(total_amount_to_pay)
+			 	self.set_total_amount_to_default_mop(total_amount_to_pay)
+				self.calculate_paid_amount()
```
nextchamp-saqib:
```suggestion
				self.set_total_amount_to_default_mop(total_amount_to_pay)
				self.calculate_paid_amount()
```

manan-ft:
Hi, is this merged in v13.11?

PR #26873, opened 2021-08-10
fix: Set CWIP Account in company at the start to avoid flaky test

marination:
Due to https://github.com/frappe/erpnext/pull/26803
Move setting CWIP account at the start.
Flaky Test Example: https://github.com/frappe/erpnext/runs/3277810413

> Already ported to v13-hotfix


marination:
POS, Attendance and SO tests, all unrelated, failing.

ankush:
Failing tests are due to doc rename/amend PR 😐

PR #26877, opened 2021-08-10
refactor: shows opening balance from filtered from_date

noahjacob:
### Stock Balance
- While calculating opening balances for the stock balance report, opening balances were only calculated for stock recos created before the filtered from_date and not the set from_date.
- If a "**Stock Reconciliation**" was created for opening stock and the stock balance report was generated with from_date set to the date when the opening stock was created, the **"Opening Balance "** would not reflect.

<details>
<summary>Stock Balances</summary>

#### Stock Balance with same date as the Stock Reconciliation

image

#### Stock Balance with a date after the Stock Reconciliation

image

</details>

### Stock Ledger

- If a stock entry was created on filtered from_date at 00:00:00, the entry was considered as the Opening Balance.

<details>
<summary>Stock Ledgers</summary>

#### Stock Ledger with a date before the Stock Entry

image

#### Stock Ledger with same date as the Stock Entry created at 00:00:00

image

</details>

### Stock Balance
- Now also calculates opening balances if the date is the same as **from_date** in the filters.
<details>
<summary>Stock Reconciliation</summary>

image

</details>

image

### Stock Ledger
- Initially only considers sles before the `from_date`.
- Now checks if any "Stock Reconciliation" Entry is present with the same filtered `from_date`. If any, the entry is removed from the Stock Ledger and added as the Opening Balance.

image

`no_docs`


erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -198,7 +198,8 @@ def get_item_warehouse_map(filters, sle):
 
 		value_diff = flt(d.stock_value_difference)
 
-		if d.posting_date < from_date:
+		if d.posting_date < from_date or (d.posting_date == from_date
+			and d.voucher_type == 'Stock Reconciliation'):
```
ankush: Should we check if `purpose` is `"Opening Stock"` here?

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -198,7 +198,9 @@ def get_item_warehouse_map(filters, sle):
 
 		value_diff = flt(d.stock_value_difference)
 
-		if d.posting_date < from_date:
+		if d.posting_date < from_date or (d.posting_date == from_date
+			and d.voucher_type == 'Stock Reconciliation' and
+			frappe.db.get_value('Stock Reconciliation', d.voucher_no, ['purpose']) == 'Opening Stock'):
```
ankush: This seems like correct behaviour now 👍 

As discussed, better to do this in the Stock Ledger report too. 

Closing other PRs, as it would be easier to port once this is final and merged.

ankush (approved):

PR #26886, opened 2021-08-10
refactor: shows opening balance from filtered from_date (bp #26877)

noahjacob:
Backport of #26877


ankush:
https://github.com/frappe/erpnext/pull/26877#discussion_r686699882

PR #26914, opened 2021-08-12
feat: batch-wise item valuation

18alantom:
### Description
This PR switches the legacy valuation for batched items to Batch-wise Item Valuation. Legacy valuation method for
batched items did not use to consider batches.

To understand the difference, consider the following incoming transactions, assume the same item *I*:
|# |Type   |Qty|Rate|Warehouse|Batch No|
|--|-------|---|----|---------|--------|
|1 |Inward |1  |100 |A        |X       |
|2 |Inward |1  |150 |A        |Y       |

If an outward entry is to be made:
|# |Type   |Qty|Warehouse|Batch No|
|--|-------|---|---------|--------|
|3 |Outward| -1|A        |Y       |

what will the item valuation be if the item uses *FIFO Valuation*, *Moving
Average Valuation*?

<details>
<summary>
<b>Before :: FIFO: 100, MA: 125</b>
</summary>
Valuation is considered on an item and warehouse basis so after transactions 1 and
2 this is how the queues will look:

|# |*Key* | Stock Queue       |
|--|------|-------------------|
|1 |(I, A)|[(1,100)]         |
|2 |(I, A)|[(1,100), (1,150)] |

Since outward entry (*# 3*) is for item *I* and warehouse *A* the last 
queue fetched for valuation is keyed by I and A. Batch is ignored here.

*Note: A key isn't implicitly used*
</details>

<details>
<summary>
<b>After :: FIFO: 150, MA: 150</b>
</summary>
Valuation is considered on an item, batch, and warehouse basis, so after
transactions 1 and 2 this is how the queues will look:

|# |*Key*    | Stock Queue |
|--|---------|-------------|
|1 |(I, X, A)|[(1,100)]    |
|2 |(I, Y, A)|[(1,150)]    |

Since outward entry (*# 3*) is for item *I*, batch *Y*, and warehouse *A* the last 
queue fetched for valuation is keyed by I, Y, and A. Batch is not ignored here.

*Note: A key isn't implicitly used*
</details>

### Additional Notes
- *Batchwise Balance History* has been refactored to account for the changes in this PR.
- Existing batches will continue to use the legacy valuation method.
- Batchwise-item valuation will apply only to batches created after this PR has been rolled out.
- The following situation is possible (by using the `use_batchwise_valuation` flag on _Batch_ and _Stock Ledger Entry_):
  1. Item A has two batches X, Y, and Z.
  2. Y uses batches-wise valuation.
  3. X and Z use legacy valuation.
  4. The valuation of Y is completely isolated from that of X and Z.

### Steps to test (QA Team)
1. Create a batched item.
2. Make _inward_ and _outward_ transactions using the created item 
    - _Inward_: Purchase Receipts, Material Receipt Stock Entries, etc.
    - _Outward_: Material Issue, Delivery Notes, etc.
    - Vary the Batch No across transactions (check the example above).
    - Vary the Rate across the transaction.
3. Check the stock ledger entry values
    - The stock queue, valuation should not change between two transactions if the batch no doesn't match.
    - Example:
      1. _Inward_ :: batch A, rate 100, qty 1: stock queue will be `[[1, 100]]`.
      2. _Inward_ :: batch B, rate 120, qty 2: stock queue will be `[[2, 120]]` not `[[1,100], [2, 120]]`.
4. Valuation Rate is used on outward transactions, this too you will get from the stock ledger entries; transactions on batch B shouldn't affect the valuation rate of batch A.

Stuff that may help:
- https://frappe.io/kb/stock/calculation-of-valuation-rate-in-fifo-and-moving-average
- https://frappe.io/kb/stock/how-valuation-rate-is-calculated-for-an-item

'no-docs'


18alantom:
### Note

The development of this PR has been on hold for a while now because the decision/implementation of how to fix prior records hasn't been elucidated yet.

Initially the plan was to use **Repost Item Valuation** in a patch to fix the valuation rate of batched items, but this wouldn't scale for hefty instances (100,000s of records).

So the underlying issue really is: _What is the best way to fix records (Stock Ledger, General Ledger and various vouchers)?_ Ignoring batch-wise item valuation the other reason to do this is [_backdated entries_](https://en.wikipedia.org/wiki/Grandfather_paradox).

The use of **Repost Item Valuation** (the way it is now) is like using a nuke when what you really need is a sniper.

Once the problem of fixing these records is solved this PR can be completed, either with a patch or something else.

Refactoring of **Repost Item Valuation** is in the roadmap, so the completion of this may be put off at least until that has been done.

18alantom:
### Continuation Note

Difficulties circumvented. The decision of what to do when this PR is deployed has been made.

Batches created before this PR is merged will use the old valuation method. Batches created after this PR will use batch-wise valuation.

18alantom:
Screen Shot 2021-10-13 at 16 51 59

_**Note**: Adding this comment as a reminder that there was a time when all tests passed on this PR because they will inevitably start failing again._ ✌️

asoral:
+1 for this PR

asoral:
Any update on this?

erpnext/stock/stock_ledger.py
```diff
@@ -936,45 +970,100 @@ def get_sle_by_voucher_detail_no(voucher_detail_no, excluded_sle=None):
...
+		warehouse=warehouse,
+		voucher_no=voucher_no,
+		voucher_type=voucher_type
+	)
+
+	last_valuation_rate = None
+	if batch_no and warehouse:
+		# 1. Get valuation_rate from last SLE with matching batch_no, warehouse
+		last_valuation_rate = frappe.db.sql("""SELECT valuation_rate
+			FROM `tabStock Ledger Entry` FORCE INDEX (batch_no_item_code_warehouse_index)
```
ankush, line 978: This index isn't explicitly named, if forcing index it should be explicitly named.

ankush: IMO this is kinda pointless. Batch numbers are fully unique right now, so batch index alone should be enough 😄

ankush:
manual test: "qty_after_transaction" seems to be broken. Updates current batch qty instead of total.

ankush:
We discussed a potentially simpler solution requiring less state in SLE:

- Whenever a batch is getting consumed, construct the batch's ledger dynamically to fetch the average valuation rate at that posting time. 

This appears to be expensive but batches are usually short-lived so we are probably fetching 10-1000 SLEs (that too from index) and this whole computation can be a single query. This means batches won't follow FIFO internally for each batch (only matters in case of multiple entries with a different valuation for a single batch) but overall valuation will respect batch nos. 


This will still require feature a toggle and branch-off valuation functionality to add a new method totaling to 4 methods: serial no, batch, regular(fifo), regular(moving average)

ankush:
continued here: https://github.com/frappe/erpnext/pull/29804

PR #26916, opened 2021-08-12
fix: Account currency validation for first transaction

deepeshgarg007:
At the moment, the party account currency and the document currency can be different. This provision was kept for legacy users who already had transactions before multi-currency was introduced and cannot be forced to keep party account currency in the document currency.

But due to this provision, many new users end up using incorrect party accounts having the wrong currency and then don't get values with the proper exchange rate in the general ledger. Have added a validation such that parties (customers and suppliers) already having transactions won't be affected but only the newer ones will be validated. This will prevent errors in future for new users

`no-docs`


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -342,6 +343,20 @@ def validate_party(self):
...
+	def validate_party_account_currency(self):
+		for d in self.get("accounts"):
+			if d.party_type in ("Customer", "Supplier"):
+				party_gle_currency = get_party_gle_currency(d.party_type, d.party, self.company)
+				party_account_currency = get_account_currency(d.account)
+				party_currency = frappe.db.get_value(d.party_type, d.party, "default_currency")
+
+				if not party_gle_currency and (party_account_currency != party_currency):
+					frappe.throw(
+						_("Party Account {0} currency and default party currency should be same").format(
```
marination:
```suggestion
						_("Party Account {0}'s currency and party's default currency should be the same").format(
```

deepeshgarg007:

kingsleyuk2003:
image

PR #26923, opened 2021-08-12
fix: Cascade deletion for Company

deepeshgarg007:
An attempt to delete a company with no transactions leads to link validation errors as during setup it gets linked to a few doctypes

Added cascade deletion of those doctypes on the deletion of a company. The doctypes are as follows

1. Process Deferred Accounting
2. GST Settings

`no-docs`


erpnext/regional/india/utils.py
```diff
@@ -870,4 +870,17 @@ def set_item_tax_from_hsn_code(item):
...
+
+def delete_gst_settings_for_company(doc, method):
+	if doc.country != 'India':
+		return
+
+	gst_settings = frappe.get_doc("GST Settings")
+
+	for d in reversed(gst_settings.get('gst_accounts')):
+		if d.company == doc.name:
+			gst_settings.remove(d)
```
ankush: Removing items from the list while iterating over it can lead to RuntimeError.

PR #26924, opened 2021-08-12
fix: fetch from more than one sales order in Maintenance Visit

### LINKED ISSUE #26336: Get Items From in Maintenance Visit
Hi everyone!

In Maintenance Visit you are not able to fetch from more than one sales order
fetch_from_maintenance

PawanMeh:
Fixes #26336

### DIFF
--- a/erpnext/maintenance/doctype/maintenance_visit/maintenance_visit.js
+++ b/erpnext/maintenance/doctype/maintenance_visit/maintenance_visit.js
@@ -73,12 +73,16 @@ erpnext.maintenance.MaintenanceVisit = class MaintenanceVisit extends frappe.ui.
 		if (this.frm.doc.docstatus === 0) {
 			this.frm.add_custom_button(__('Maintenance Schedule'),
 				function () {
+					if (!me.frm.doc.customer) {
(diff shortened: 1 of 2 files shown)


ankush:
This will also allow SOs from different customers to be pulled in a single Schedule / Visit, right? That will be kinda confusing since the document only has a single field for it. 

Maybe a reasonable restriction for this change could be that multiple SOs should be from the same customer? (Like how it's done in Pick List right now)

Screenshot 2021-08-12 at 7 29 05 PM

PawanMeh:
You select a customer on the schedule and the search results in the dialogue are filtered by the customer selected already. Also, that is not part of this fix anyways. (Based on the customer the SOs are filtered already)

ankush:
> You select a customer on the schedule

This is not mandatory right now, so you can do "Get items from" and select multiple SOs with different customers.

PawanMeh:
Ok maybe someone can pick it up from here then as this fix was intended to just fix the issue of not being able to select multiple items even if it is the same customer. 

Also being able to select other items which do not belong to the same customer is across the application as well, you can do it on delivery note as well and possibly other documents.

delivery_note

ankush (approved):

PR #26926, opened 2021-08-12
fix(Salary Slip): wrong formula calculation

anupamvs:
If the earnings component (`HRA`) with property depends on payment day and amount of deduction component depends upon one of the earning components with a formula like (`gross_pay - HRA`). Then Deduction amount of the component is wrongly calculated.


nabinhait:
@anupamvs This seems problematic, we can discuss the issue

anupamvs:
@nabinhait Sure

anupamvs:
Closing for now.

PR #26931, opened 2021-08-12
fix: item name display in job card

FHenry:
Item name into job card is not correctly fetch


erpnext/manufacturing/doctype/job_card/job_card.json
```diff
@@ -280,7 +280,7 @@
    "read_only": 1
   },
   {
-   "fetch_from": "production_item.item_name",
+   "fetch_from": "work_order.production_item.item_name",
```
ankush, line 283: please respond to this first https://github.com/frappe/erpnext/pull/26386#discussion_r685045376

ankush (changes requested):

PR #26936, opened 2021-08-13
fix: filtering of items in Sales and Purchase Orders

AfshanKhan:
Steps to replicate the problem:

1. Create an Item and uncheck the checkbox `Is Sales Item` under Sales Details Section and `Is Purchase Item` under Purchase, Replenishment Details Section.
2. Now create a Sales order or a Purchase Order and submit it.
3. Use `Update Item` button on the top to update the items of this order.
4. Add a new row and search for the newly created Item whose `Is Sales Item` and `Is Purchase Item` checkbox are unchecked.
5. Add the item fill the details and click on `Update`

The Item gets added to the Order. This behaviour is wrong.

This PR aims to fix this error.


erpnext/public/js/utils.js
```diff
@@ -465,7 +465,23 @@ erpnext.utils.update_child_items = function(opts) {
...
+				filters = {"is_sales_item": 1};
+			} else if (frm.doc.doctype == 'Purchase Order') {
+				if (frm.doc.is_subcontracted == "Yes") {
+					filters = {"is_sub_contracted_item": 1};
+				} else {
+					filters = {"is_purchase_item": 1};
+				}
+			}
+			return {
+				query : "erpnext.controllers.queries.item_query",
```
AfshanKhan:
```suggestion
				query: "erpnext.controllers.queries.item_query",
```

erpnext/public/js/utils.js
```diff
@@ -465,7 +465,23 @@ erpnext.utils.update_child_items = function(opts) {
...
+				filters: filters
+			}
```
AfshanKhan:
```suggestion
			};
```

erpnext/public/js/utils.js
```diff
@@ -465,7 +465,23 @@ erpnext.utils.update_child_items = function(opts) {
...
+		}
```
AfshanKhan, line 484:
```suggestion
		};
```

erpnext/public/js/utils.js
```diff
@@ -465,7 +465,23 @@ erpnext.utils.update_child_items = function(opts) {
...
+					filters = {"is_sub_contracted_item": 1};
+				} else {
+					filters = {"is_purchase_item": 1};
+				}
+			}
+			return {
+				query: "erpnext.controllers.queries.item_query",
+				filters: filters
+			};
+		};
```
AfshanKhan:
```suggestion
		}
```

PR #26940, opened 2021-08-13
refactor!: drop e-invoicing integration from erpnext

nextchamp-saqib:
v14 onwards, the integration will be moved to a separate app https://github.com/frappe/erpnext_gst_compliance
Data migration is taken care of after installing the app.


erpnext/patches/v13_0/einvoicing_deprecation_warning.py
```diff
@@ -0,0 +1,9 @@
+import click
+
+
+def execute():
+	click.secho(
+		"E-Invoicing Integration is moved to a separate app and will be removed from ERPNext in version-14.\n"
```
nextchamp-saqib:
```suggestion
		"Indian E-Invoicing integration is moved to a separate app and will be removed from ERPNext in version-14.\n"
```

PR #26942, opened 2021-08-13
fix: (Stock Analytics) Copy previous balance dict object instead of assigning

marination:
Fix for https://github.com/frappe/erpnext/pull/26673/ (for release)

Due to plain assignment, dict mutation gave wrong monthly values

### DIFF
--- a/erpnext/stock/report/stock_analytics/stock_analytics.py
+++ b/erpnext/stock/report/stock_analytics/stock_analytics.py
@@ -144,7 +144,8 @@ def get_periodic_data(entry, filters):
 		# if period against item does not exist yet, instantiate it
 		# insert existing balance dict against period, and add/subtract to it
 		if periodic_data.get(d.item_code) and not periodic_data.get(d.item_code).get(period):
-			periodic_data[d.item_code][period] = periodic_data[d.item_code]['balance']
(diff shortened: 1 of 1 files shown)


erpnext/stock/report/stock_analytics/stock_analytics.py
```diff
@@ -144,7 +144,8 @@ def get_periodic_data(entry, filters):
 		# if period against item does not exist yet, instantiate it
 		# insert existing balance dict against period, and add/subtract to it
 		if periodic_data.get(d.item_code) and not periodic_data.get(d.item_code).get(period):
-			periodic_data[d.item_code][period] = periodic_data[d.item_code]['balance']
+			previous_balance = periodic_data[d.item_code]['balance'].copy()
```
ankush, line 147: Should this be a deepcopy? `balance` also seems to be a nested dict.

ankush: Ignore. It's just a WH -> number mapping.

ankush (approved):

erpnext/stock/report/stock_analytics/stock_analytics.py
marination: @ankush its not 
Screenshot 2021-08-13 at 3 38 24 PM

PR #26949, opened 2021-08-13
fix: Pricing Rule on Transaction Based on Coupon

### LINKED ISSUE #26948: Transaction Pricing Rule get applied even if it a based-on coupon
## Description of the issue

- Pricing Rule created for coupon gets applied to all transactions without coupon is applied

1. Create Pricing Rule for Coupon Based select Apply on Transaction, Price or Product Discount - Price
2. In Any Document [Quotation,..] which applies the pricing rule, by default discount is applied to all without the Coupon.

lovinmaxwell:
> Pricing Rule creates for coupon-based get applied for all transition 

Fixes #26948


erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -475,6 +475,10 @@ def apply_pricing_rule_on_transaction(doc):
 						frappe.msgprint(_("User has not applied rule on the invoice {0}")
 							.format(doc.name))
 					else:
+						if d.coupon_code_based:
+							if not doc.get('coupon_code') or doc.get('coupon_code') is None:
+								doc.set(field, 0)
+								continue
 						doc.set(field, d.get(pr_field))
```
nextchamp-saqib:
```suggestion
						if not d.coupon_code_based:
							doc.set(field, d.get(pr_field))
						elif doc.get('coupon_code'):
							# coupon code based pricing rule
							coupon_code_pricing_rule = frappe.db.get_value('Coupon Code', doc.get('coupon_code'), 'pricing_rule')
							if coupon_code_pricing_rule == d.name:
								# if selected coupon code is linked with pricing rule
								doc.set(field, d.get(pr_field))
							else:
								# reset discount if not linked
								doc.set(field, 0)
						else:
							# if coupon code based but no coupon code selected
							doc.set(field, 0)
```

nextchamp-saqib (dismissed):

nextchamp-saqib (approved):

PR #26956, opened 2021-08-16
fix: item name is missing into job card

### LINKED ISSUE #26954: Version-12 Item name missing in Job Card
Bug only exist in version 12

Create a BOM for the Item :  "Item 1"

BOM use one Operation

Create an OF for this BOM,
Create Job card for this OF

image

FHenry:
fix #26954

This fix must not be merge into upper version because in version-13 and develop the field production_item is a typefield link and readonly:1

Could be related to #26386

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.json
+++ b/erpnext/manufacturing/doctype/work_order/work_order.json
@@ -472,7 +472,7 @@
  "image_field": "image",
  "is_submittable": 1,
  "links": [],
- "modified": "2019-12-04 11:20:04.695123",
+ "modified": "2021-08-16 11:20:04.695123",
  "modified_by": "Administrator",
  "module": "Manufacturing",
  "name": "Work Order",
--- a/erpnext/manufacturing/doctype/work_order/work_order.json
+++ b/erpnext/manufacturing/doctype/work_order/work_order.json
@@ -107,6 +107,8 @@
   },
   {
    "depends_on": "eval:doc.production_item",
+   "fetch_from": "production_item.item_name",
+   "fetch_if_empty": 1,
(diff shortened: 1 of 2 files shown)


FHenry:
@ankush There is the fix for the version-12 only

ankush (dismissed):

erpnext/manufacturing/doctype/work_order/work_order.json
```diff
@@ -107,6 +107,8 @@
   },
   {
    "depends_on": "eval:doc.production_item",
+   "fetch_from": "production_item.item_name",
```
ankush, line 110: This looks good. You need to change the `modified` key of JSON file so this change is migrated.

ankush (dismissed):

erpnext/manufacturing/doctype/work_order/work_order.json
FHenry: Thank's, I try again and I can confirm that 2 changes are needed, job_card.json with the strange fetch from "work_order.production_item.item_name" (that should not work but it does..., may be as field production_item is type ReadOnly and not Link  like as in future version) populated by  and not link in this version) and and work_order.json. 

This is the type of bug fix in 13 (and develop) but not in 12, It's really strange for me that fix in develop first, but I'll manage to play as your rule

FHenry:
I try again and I can confirm that 2 changes are needed, job_card.json with the strange fetch from "work_order.production_item.item_name" (that should not work but it does..., may be because field job_card.production_item is type ReadOnly and not Link like as in future version) and the change in work_order.json.

ankush (approved):

PR #26992, opened 2021-08-17
fix: Sales order qty update fails in "Update Items" button

TurkerTunali:
Branch corrected https://github.com/frappe/erpnext/pull/26058

  ERPNext generates "Cannot set quantity less than delivered quantity" error even the delivered qty is zero when user clicks "Update Items".
  "date_unchanged" variable gets false value because of new_date is string. 
  
  "getdate(new_date)" corrects the date comparison.

https://user-images.githubusercontent.com/710051/129911124-9798ef95-1d5e-4d3d-b766-eff91c50c837.mp4
  
  
  
  > Screenshots/GIFs
  
  

> Screenshots/GIFs

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -1312,7 +1312,7 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
 			qty_unchanged = prev_qty == new_qty
 			uom_unchanged = prev_uom == new_uom
 			conversion_factor_unchanged = prev_con_fac == new_con_fac
-			date_unchanged = prev_date == new_date if prev_date and new_date else False # in case of delivery note etc
+			date_unchanged = prev_date == getdate(new_date) if prev_date and new_date else False # in case of delivery note etc
 			if rate_unchanged and qty_unchanged and conversion_factor_unchanged and uom_unchanged and date_unchanged:
 				continue
 


erpnext/controllers/accounts_controller.py
```diff
@@ -1312,7 +1312,7 @@ def validate_quantity(child_item, d):
 			qty_unchanged = prev_qty == new_qty
 			uom_unchanged = prev_uom == new_uom
 			conversion_factor_unchanged = prev_con_fac == new_con_fac
-			date_unchanged = prev_date == new_date if prev_date and new_date else False # in case of delivery note etc
+			date_unchanged = prev_date == getdate(new_date) if prev_date and new_date else False # in case of delivery note etc
```
ankush, line 1315: While this does fix `date_unchanged` behaviour; are you sure this is the reason behind the error you're seeing? 

Because if you change qty then `qty_unchanged` should become False and this check will be irrelevant. 

The error you saw comes from here: https://github.com/frappe/erpnext/blob/fbc163cea66b52494595435200c43ea581fe9bd2/erpnext/controllers/accounts_controller.py#L1827-L1829

TurkerTunali: Code works as expected. Let me explain:

There are 10 lines in the sales order. Other lines have problem with quantities. So throwing the "Cannot set quantity less than delivered quantity" error is OK for them. But, the edited line doesn't have any problem with the date or quantity.

ERPNext was checking the other lines because of date error. This code prevents it.

ankush: Ok. Merging as this fixes date_unchanged behaviour and prevents unnecessary validations.

ankush (approved):

erpnext/controllers/accounts_controller.py
ankush: Need to tinker with some configs to get this ported to develop / v13 branches.

PR #26995, opened 2021-08-18
fix: replacing $.each -> forEach in SI

anupamvs:


erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -451,11 +451,14 @@ erpnext.accounts.SalesInvoiceController = class SalesInvoiceController extends e
 
 	currency() {
 		super.currency();
-		$.each(cur_frm.doc.timesheets, function(i, d) {
-			let row = frappe.get_doc(d.doctype, d.name)
-			set_timesheet_detail_rate(row.doctype, row.name, cur_frm.doc.currency, row.timesheet_detail)
-		});
-		calculate_total_billing_amount(cur_frm)
+		console.log(this.frm);
```
rmehta: console?

rmehta (changes requested):

PR #26997, opened 2021-08-18
fix: add child item groups into the filters

AfshanKhan:
### Problem:
If for a Sales Person if the target which is set for them has value for Item Group selected as any Root/Grouped Item Group and the Sales Orders from those Sales Person contains the Item with Item Groups as child of the Item Group that was selected in their target. The report was only considering the Items with the Parent Item Group that was set in the target and not the Child Item Group of that Item Group.

With this PR the Child Item Groups are also considered.


erpnext/selling/report/sales_partner_target_variance_based_on_item_group/item_group_wise_sales_target_variance.py
```diff
@@ -44,6 +44,13 @@ def get_data(filters, period_list, partner_doctype):
 		if d.item_group not in item_groups:
 			item_groups.append(d.item_group)
 
+	if item_groups:
+		for item_group in item_groups:
+			if frappe.db.get_value("Item Group", {"name":item_group}, "is_group"):
+				for child_item_group in frappe.get_all("Item Group", {"parent_item_group":item_group}):
+					if child_item_group['name'] not in item_groups:
+						item_group.append(child_item_group['name'])
```
AfshanKhan:
```suggestion
						item_groups.append(child_item_group['name'])
```

PR #27014, opened 2021-08-18
fix: stock analytics report date range issues and add company filter

### LINKED ISSUE #4812: Modification in Stock analytics report
1. Item Name should be displayed (either in place of Item Code OR in separate column)
2. search/filter by Item Name
3. Option to hide zero value rows
4. From report heading, it's not clear that the analysis is of what (Inword, Outword, balance or anything else?)
5. Option to filter by Company

image

ankush:
Do transactions like these:
Screenshot 2021-08-18 at 7 51 42 PM

Open stock analytics report and compute with Monthly frequency. 30th December entry is missing. This is because while calculating periods "end date" is considered, which for the first period is 30th Dec + 1 month. 

Screenshot 2021-08-18 at 7 51 14 PM

Solution: "round down" dates according to the selected frequency. 

Closes: FR-ISS-302008

Also added company filters on the report (similar to other stock reports)

Screenshot 2021-08-18 at 8 41 05 PM


ankush:
Will also add company as filter to close https://github.com/frappe/erpnext/issues/4812

rohitwaghchaure (approved):

PR #26480, opened 2021-07-14
feat(regional): South Africa VAT Audit Report

Anuja-pawar:
- [x] Add two DocTypes- South Africa VAT Settings & South Africa VAT Accounts, 
        To fetch accounts for which VAT report should be generated

Screenshot 2021-07-14 at 12 24 30 PM

- [x] Create VAT Audit Report

Screenshot from 2021-08-09 11-22-58

- [x] Handle 0% tax rate (added custom field `Is Zero Rated` in Item)
- [x] Documentation

PR #27017, opened 2021-08-19
feat(regional): South Africa VAT Audit Report

Anuja-pawar:
Backport of b478449 from https://github.com/frappe/erpnext/pull/26480 and 08ae49c from https://github.com/frappe/erpnext/pull/26851

---
- [x] Add two DocTypes- South Africa VAT Settings & South Africa VAT Accounts, 
        To fetch accounts for which VAT report should be generated

Screenshot 2021-07-14 at 12 24 30 PM

- [x] Create VAT Audit Report

Screenshot from 2021-08-09 11-22-58

- [x] Handle 0% tax rate (added custom field `Is Zero Rated` in Item)
- [x] Documentation


Ponyooooo:
Is there a way to enable this feature even though the base country of the company is not South Africa? Since this is also applicable for some countries.

erpnext/regional/report/vat_audit_report/vat_audit_report.py
```diff
@@ -0,0 +1,269 @@
...
+				{select_columns}
+			FROM
+				`tab{doctype}`
+			WHERE
+				docstatus = 1 {where_conditions}
+				and is_opening = "No"
+			ORDER BY
+				posting_date DESC
+			""".format(select_columns=self.select_columns, doctype=doctype,
+				where_conditions=conditions), self.filters, as_dict=1)
```
surajshetty3416, line 62: @Anuja-pawar There are a couple of instances like this where ORM could have been used instead of raw SQL.

(**Note:** Always use ORM if possible as it makes the code more readable and it minimizes SQLi risk)

Andriesvn:
I'm not so sure that this would be considered a VAT report. 

I think there might be some misunderstanding as to what VAT Report is in this case. VAT is not just Item Based. VAT is charged on almost anything and everything not just "Purchase Invoice" and "Sales Invoice". Bank charges include VAT, Rental Agreements include VAT, Returns include VAT, The list goes on. This is what confuses me in regards to how ERP handles VAT. I mean its a Tax included in all payments and receipts and already handled by Sales Taxes and Purchases Taxes in ERPNext. (Just not correctly when it comes to VAT Exempt | Is Zero Rated as the GL does not store 0 value entries). 

So my big question when looking at this is why its being duplicated? Why are you adding an additional Doctype for VAT when Sales Taxes and Purchases Taxes already exists? Then only linking it to "Purchase Invoice" and "Sales Invoice" instead of everything from Returns, Payments, Receipts, etc.? And worst of all, its locked to a specific region | country.

Would it not have been easier to simply setup taxes and charges correctly then Union query the following tables
- Purchase Taxes and Charges
- Sales Taxes and Charges
- Advance Taxes and Charges
- Landed Cost Taxes and Charges
- Expense Taxes and Charges

for an Audit report? 
Can someone please explain this to me?

Anuja-pawar:
@Andriesvn This report is particularly made for South Africa (Delivery Request), and has scope changes. The requirements were specific about making different sections for sales and purchases, and for tax rates 15% as well as for 0% (hence zero-rated) thus it can not be generic to all countries.
The new doctype is just to identify the VAT Accounts as Tax and Charges may have other charges that are not to be included in VAT report

Andriesvn:
That's the point, Sales and Purchases are not the only things that make up a VAT report. The purpose of a VAT Report is to submit your VAT correctly at the end of the VAT period. This report does not take into account anything other than Sales and Purchases.  Submitting this report to the Receiver of Revenue would trigger an audit at the end of the day as its missing everything else including basic things like Credit Notes (Sales or Purchases wise). 

And Yes, I agree that you are correct as not all countries have VAT specifically at exactly the same rates, All countries do have Taxes, Specifically Sales taxes which is handled in all countries by ERPNext. Which is the reason for my confusion as to why ERPNext tends to treat VAT as anything other than Taxes. My point is simply instead of having a report that's linked to Each and every country, Why not just create a general "Taxes Report" for all countries that would show How much taxes per category was charged vs paid per Tax Category for a certain period. Taxes that are again already recorded in ERPNext. This would solve 90% of VAT Reports and not need a Per Country specification.  In specific cases such as United Arab Estates I would understand the need for a specific report | a specific routine but 90% of other cases a general "Taxes Report" solve most VAT related issues.

Yes the new doctype is to link the Vat Account to the report, I understand that. but you see my point? Now we need to link a tax to a report again just for one specific country for something that (if im not mistaken) everybody wants to see, "How much of a specific Tax Category did i sell or pay for in a specific period"

Just to prove my point, Will you be creating a separate VAT report for each and every country here?
https://www.uscib.org/value-added-tax-rates-vat-by-country/

Andriesvn:
Here`s another VAT Report to prove my point again https://github.com/frappe/erpnext/pull/27154

The only mayor difference in this case between this report and that one is mostly the naming used in the report itself.  But essentially both reports do the same thing which is comparing the "Tax you sold" to the "Tax you Purchased" per category over a certain period. Just incorrectly again as it did not again take into account any credit notes you might have issued, any tax you might have occurred in landed costs or any bank charges that you might have been charged tax on in that period, just to name a few. But both in this case again create new settings to link the VAT Account To the Vat Report. Both Query the exact same thing and just display it under different names. SA report has 0 rated vat included however so does most other VAT Regime Based countries. 

I'm not trying to start and argument in this case as allot of work did go into preparing both of these reports however I feel the way ERPNext is handling the VAT issue as a whole is very inefficient. It might be time to rethink the issue of "Creating a Separate VAT Report Per country" issue. ERPNext has Tax Categories already separating the deferent "Tax Type Transactions". 

One possible Solution: Allow 0 Value GL Entries if the Account is a Tax Account then you would not even need to query all the tables listed in https://github.com/frappe/erpnext/pull/27017#issuecomment-912547262 .

Another more accurate but more costly method could be to create a Tax ledger. Whenever a Tax Transaction occurs, Post the Tax category, Total amount  for the specific tax category, the document which the transaction belongs to, and the total amount which the was tax calculated on into the Tax ledger. This would allow you to easily query any tax totals per category(ies) with the totals on which the tax was calculated on (an Obvious requirement for any VAT | TAX Submissions) and would already include any additions or deductions during a period and would even allow Validations on any other document submissions as you would be able to hook onto a Tax ledger entry`s Validation hook that you could later use for example VAT Period Locking and | or Recording VAT Submissions just to name a few. 

These are just examples of how this Tax issue could be handled differently instead of Creating a New Setting and a New Report Per Country.

PR #27021, opened 2021-08-19
fix: discard empty rows from update items

ankush:
image

If you leave empty rows while using "Update items", you'll see a weird error like "Item None not found" or

image

Solution: discard rows that don't have item code.


ankush:
btw unit tests won't run and show as "stuck" because of a GitHub bug. They are skipped for JS only changes, but since they are "required" for merging they are shown as "expected to start".

Working on fixing that.

rohitwaghchaure (approved):

Issue #26822, opened 2021-08-05
Production Plan Status Showing "in Progress" even all the work orders are completed


casakthiveln:
## Description of the issue
Production Plan Status Showing "in Progress" even all the work orders are completed

## Context information (for bug reports)

**Output of `bench version`**
```
V 13.8.0
```

## Steps to reproduce the issue

1.
2.
3.

### Observed result
![ezgif com-gif-maker](https://user-images.githubusercontent.com/63579207/128329768-47854f7c-db72-4024-8dbf-af9fea4ba912.gif)

### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

dvdl16:
[This line](https://github.com/frappe/erpnext/blob/version-13/erpnext/manufacturing/doctype/production_plan/production_plan.py#L300) should be changed:

```python
		if self.total_produced_qty > 0:
			self.status = "In Process"
			#if self.total_produced_qty == self.total_planned_qty:
			if self.total_produced_qty >= self.total_planned_qty:
				self.status = "Completed"
```

marination:
@dvdl16 can you raise a PR for this?

casakthiveln:
Thank you all.....

PR #27027, opened 2021-08-19
fix: set production plan to completed even on over production

18alantom:
fix for issue [#26822](https://github.com/frappe/erpnext/issues/26822)

Fixes production plan showing incorrect status (on overproduction).


ankush:
@18alantom please add a patch for this too 😄

PR #27036, opened 2021-08-19
fix: allow to change incoming rate manually in case of stand-alone credit note

rohitwaghchaure:
Use case:

There is an issue reported by customer that the Price List Rate defined in the sales invoice is not reflecting in the Stock Asset account when they create the stand-alone credit note with update stock enabled.
Users think that the Price List Rate and Inventory Costing is the same. 
We have incoming rate field already in the sales invoice item table, but that field is not editable. 

**Fix:**
Now user able to edit the rate in case of stand-alone credit note


erpnext/controllers/selling_controller.py
```diff
@@ -371,18 +371,19 @@ def set_incoming_rate(self):
...
-					"posting_date": self.get('posting_date') or self.get('transaction_date'),
-					"posting_time": self.get('posting_time') or nowtime(),
-					"qty": qty if cint(self.get("is_return")) else (-1 * qty),
-					"serial_no": d.get('serial_no'),
-					"company": self.company,
-					"voucher_type": self.doctype,
-					"voucher_no": self.name,
-					"allow_zero_valuation": d.get("allow_zero_valuation")
-				}, raise_error_if_no_rate=False)
+				if not (d.incoming_rate or self.get('is_return')):
```
marination: We want to set the incoming rate if it is a return case and user has not set anything already
```suggestion
				if not d.incoming_rate and self.get('is_return'):
```

marination (dismissed):

marination:
@rohitwaghchaure travis failing

PR #27039, opened 2021-08-19
feat: common party accounting

nextchamp-saqib:
The PR incorporates the solution of [this](https://github.com/ahmadpak/doubleledger) app for handling scenarios where sales invoices/purchase invoices are to be made against suppliers/customers

The solution proposed by the app is to:
- Create a journal entry on the submission of Sales Invoices against a customer which has the primary role of a supplier.
- The journal entry will reconcile the newly created sales invoice and converts the invoice balance into an advance against the supplier
- This advance can be used to reconcile purchase invoices against the supplier

A new doctype Party Link is introduced to link a customer with a supplier.

CleanShot 2021-08-19 at 18 05 15@2x

A new toggle is added to accounts settings to enable/disable the automatic creation of the journal entry.

---

Thanks @ahmadpak for the solution!

---

Docs: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/articles/common_party_accounting


ahmadpak:
My pleasure @nextchamp-saqib

erpnext/accounts/doctype/party_link/party_link.py
```diff
@@ -0,0 +1,12 @@
...
+# For license information, please see license.txt
+
+import frappe
+from frappe import _
+from frappe.model.document import Document
+
+class PartyLink(Document):
+	def validate(self):
+		if self.primary_role not in ['Customer', 'Supplier']:
+			frappe.throw(_("Allowed primary roles are 'Customer' and 'Supplier'. Please select one of these roles only."),
```
deepeshgarg007, line 11: Add validation to ensure that one Supplier is not linked to multiple customers and vice versa as I think we have not handled that scenario yet

erpnext/accounts/doctype/party_link/party_link.json
```diff
@@ -0,0 +1,78 @@
...
+ "module": "Accounts",
+ "name": "Party Link",
+ "owner": "Administrator",
+ "permissions": [
+  {
+   "create": 1,
+   "delete": 1,
+   "email": 1,
+   "export": 1,
+   "print": 1,
```
deepeshgarg007, line 66: Add roles for accounts manager and other accounts user

erpnext/controllers/accounts_controller.py
```diff
@@ -1362,6 +1362,66 @@ def is_internal_transfer(self):
...
+		if frappe.db.get_single_value('Accounts Settings', 'enable_common_party_accounting'):
+			party_link = self.get_common_party_link()
+			if party_link and self.outstanding_amount:
+				self.create_advance_and_reconcile(party_link)
+
+	def get_common_party_link(self):
+		party_type, party = self.get_party()
+		party_link = frappe.db.exists('Party Link', {'secondary_role': party_type, 'secondary_party': party})
+		if party_link:
+			return frappe.db.get_value('Party Link', party_link, ['primary_role', 'primary_party'], as_dict=True)
```
deepeshgarg007: Can be checked in just one query? `get_value` should be enough to check and even get values for common party link

erpnext/controllers/accounts_controller.py
```diff
@@ -1362,6 +1362,66 @@ def is_internal_transfer(self):
...
+	def create_advance_and_reconcile(self, party_link):
+		secondary_party_type, secondary_party = self.get_party()
+		primary_party_type, primary_party = party_link.primary_role, party_link.primary_party
+
+		primary_account = get_party_account(primary_party_type, primary_party, self.company)
+		secondary_account = get_party_account(secondary_party_type, secondary_party, self.company)
+
+		jv = frappe.new_doc('Journal Entry')
+		jv.voucher_type = 'Journal Entry'
+		jv.naming_series = 'ACC-JV-.YYYY.-'
```
deepeshgarg007: Let's not set any naming series? The user might have deleted the default one or may have set some other naming series as default

erpnext/controllers/accounts_controller.py
```diff
@@ -1362,6 +1362,66 @@ def is_internal_transfer(self):
...
+		jv.posting_date = self.posting_date
+		jv.company = self.company
+		jv.remark = 'Adjustment for {} {}'.format(self.doctype, self.name)
+
+		reconcilation_entry = frappe._dict()
+		advance_entry = frappe._dict()
+		cost_center = erpnext.get_default_cost_center(self.company)
```
deepeshgarg007: Can we use the Sales/Purchase Invoice parent or child table cost center instead of the default one? Will be consistent if the user has not used the default one in Sales/Purchase Invoice

deepeshgarg007 (changes requested):

deepeshgarg007 (approved):

PR #27040, opened 2021-08-19
fix: Shopping cart Exchange rate validation

marination:
**Issue:**
- Create a Company **ABC** with currency **INR** and have a Selling Price List with currency **INR**
- In Shopping Cart Settings, enable it, set company as **ABC** and Price list as INR selling price list. Save.
- Go to this Selling Price List, change it's currency to USD/any other currency
- This appears:
  Screenshot 2021-08-19 at 5 52 44 PM
- This same issue occurs on changing company currency too. If it is changed to AUD (with country Australia) from INR, while creating sales taxes and charges template for Australia, the same error occurs

**Actual Issue:**
- This validation is called from `hooks.py` and on saving Shopping Cart Settings
- The validation is in place to make sure that "missing currency" error does not appear while someone is using the shopping cart.
- **How it validates this? By checking if a currency exchange record exists**. In almost all cases people rely on the currency exchange API to fetch exchange rate
- Currency Exchange record is made for manual cases.
- In this case it almost always throws this error because `Currency Exchange` record does not exist (and is not mandatory)

**Fix:**
- Use `get_exchange_rate` to check for price list exchange rate. It returns value from Currency Exchange` record if a fresh on exists else it hits the exchange rate API
- Move cart exchange rate validation for Price List from hooks to doc event.
- Call cart exchange rate validation on PL update only if PL is in cart and currency is changed

### DIFF
--- a/erpnext/hooks.py
+++ b/erpnext/hooks.py
@@ -251,7 +251,7 @@ doc_events = {
 			"erpnext.support.doctype.issue.issue.set_first_response_time"
 		]
 	},
-	("Sales Taxes and Charges Template", 'Price List'): {
+	"Sales Taxes and Charges Template": {
 		"on_update": "erpnext.shopping_cart.doctype.shopping_cart_settings.shopping_cart_settings.validate_cart_settings"
 	},
 	"Website Settings": {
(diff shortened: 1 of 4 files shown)


nabinhait (approved):

marination:
tests that are failing are in POS (due to framework) and in SLA

PR #27044, opened 2021-08-19
fix: Eway bill version changed to 1.0.0421

nemesis189:
**Problem:**
As govt makes changes to excel tool, it changes version and old versions are withdrawn.
Changes to excel tool is not affecting our JSON file, and it's working with just change in version.
image

**Solution:**
Changed the version in the method to latest one


rohitwaghchaure:
@nemesis189 can you fix the test case
Screenshot 2021-08-23 at 2 17 42 PM

nemesis189:
> @nemesis189 can you fix the test case
> Screenshot 2021-08-23 at 2 17 42 PM

oh sorry, my bad! Raised a PR just now

PR #27050, opened 2021-08-20
fix: Shopping cart Exchange rate validation

marination:
Backport of https://github.com/frappe/erpnext/pull/27040

Shopping Cart Settings -> E Commerce Settings


nabinhait:
@marination We have changed the exchange rate API from `frankfurter` to `exchangerate.host` in version-13 as well. Do you need any changes due to that?

marination:
> @marination We have changed the exchange rate API from `frankfurter` to `exchangerate.host` in version-13 as well. Do you need any changes due to that?

@nabinhait this is in draft because v13 was still using frankfurter, this PR is fit for exchangerate.host

marination:
Unrelated `test_appointment_cancel` failing

PR #27052, opened 2021-08-20
fix: update scrap table item details; typo

### LINKED ISSUE #26207: BOM Scrap Child Table does not complete like materials table on import
When importing BOMs and only completing the following columns in the scrap table, the child table does not update the rate, name, amount, etc.

The material table works as expected.

scrap_items		
item_code	stock_qty	stock_uom

Workaround, add scrap to the material table to track costs

ERPNext: v13.5.0 (version-13)
Frappe Framework: v13.5.1 (version-13)

18alantom:
Updates scrap item details on save.

Fixes [#26207](https://github.com/frappe/erpnext/issues/26207)


marination:
unrelated test_closing_entry failing, even on other PRs.

PR #27063, opened 2021-08-20
feat: Manufacturing : in BOM Operation time can be fix #27062

### LINKED ISSUE #27062: Operation time can be fix (not depending on quantity to produce)
Into BOM, an operation time can be not depending on quantity to produce during an OF
For example, start and prepare the oven always take 15 mins, what ever who want to make 1 cake or 5000 cakes

FHenry:
Issue #27062
Into BOM, an operation time can be not depending on quantity to produce during an OF
For example, start and prepare the oven always take 15 mins, what ever who want to make 1 cake or 5000 cakes


erpnext/manufacturing/doctype/bom_operation/bom_operation.json
```diff
@@ -117,13 +118,21 @@
    "fieldname": "sequence_id",
    "fieldtype": "Int",
    "label": "Sequence ID"
+  },
+  {
+   "default": "0",
+   "description": "If checked, on Work Order time will be fixed (do not depends on quantity to produce)",
+   "fieldname": "fixed_time",
+   "fieldtype": "Check",
+   "in_list_view": 1,
```
ankush: Have you considered alternatives like batch sizes and "BOM Qty"? That would solve the problem in most cases.


Also this behaviour has to be changed in other places where BOM operations table is used... like the BOM doctype.

erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -470,17 +470,35 @@ def set_work_order_operations(self):
...
 					from
 						`tabBOM Operation`
 					where
-						parent = %s order by idx
+						parent = %s
+						and fixed_time = 0
+						order by idx
 					""", bom_no, as_dict=1)
 
+			for o in frappe.db.sql(
```
ankush: This doesn't seem to be correct. Why the same query is running twice and added to output?

ankush (changes requested):

erpnext/manufacturing/doctype/work_order/work_order.py
FHenry: It's not the same query, one time with "and fixed_time = 0", inthis case time_in_minutes is * {qty}  and the other time with and "fixed_time = 1" without the * {qty} to get all the case

erpnext/manufacturing/doctype/bom_operation/bom_operation.json
FHenry: The batch size cannot be use is this use case, to make 
BOM for 1 cherry cake, I need article (flour, cherry, butter,etc...) and the first operation "warm'up the oven" (at 150 C°)  it take 15 mins. 
On the OF I can produce 1 to 150 (or more) cherry cake this time (and cost) will not change.
I don't want to make BOM for 1 cherry cake , BOM for 2 cherry cake, BOM for 3 cherry cake, etc...

On anoter BOM for blueberry cake, I need article (flou, blue berry, butter,etc...) and the first operation will be "warm'up the oven," (at 180 °C) it take 18 mins. 
On the OF I can produce 1 or 150 blueberry cake this time (and cost) will not change.
Same I don't want to make BOM for 1 blueberry cake, bom for 2 blue berry cake ,etc..

That why there is no other doctype to update, only the table child Doctype  bom_operation use into BOM 
image

erpnext/manufacturing/doctype/work_order/work_order.py
ankush: That's a lot of duplication of code TBH. Anyway, need to evaluate first whether this feature should exist and in this form.... will take care of duplication of code later.

guilmori:
I need this feature also.
Use case is a fermenter tank, the operation duration should remain fixed, no matter the planned volume inside the tank.

erpnext/manufacturing/doctype/work_order/work_order.py
FHenry: I can tell you it a needed feature not just for me but because I work with other ERP that manage production process (open source and proprietary) and it a feature they've got so ERPNext, to my humble point of view need it.As you can see another user ask for it and I can swear I don'tt know him, I was even surprise to see a comment on my PR. 
Would you like me to ask on forum to see what user tell ?

olamide2:
This seems like a logical feature.

szufisher:
in SAP term, this is a kind of setup operation.

erpnext/manufacturing/doctype/work_order/work_order.py
FHenry: @ankush I've review the code and remove the duplication part, I let you check.

ankush:
added `hotfix` label to avoid auto-closing PR... we will add this feature... Just need some time to cleanup the UX related to this and review the PR.

ankush (approved):
LGTM.

Also checked Job card behavior for ops, since it uses WO as reference for timing it also works as expected. 

pending: update of docs.

PR #27065, opened 2021-08-20
fix: partial rename while replacing abbreviation

nextchamp-saqib:
Problem:
- A user renames company abbreviation from ABC to XYZ
- The process starts in a background job and before renaming any warehouse or account and other records, the company's `abbr` is set as XYZ (new value)
- If the background process fails, the company's abbr is changed but cost-center/account/warehouse some of these records are not renamed completely
- If the user tries to rename the abbreviation again, the renaming process now starts assuming the old abbreviation as XYZ

Proposed Change:
- Set company doc's `abbr` to the new value only when all the records have been renamed.

### DIFF
--- a/erpnext/setup/doctype/company/company.py
+++ b/erpnext/setup/doctype/company/company.py
@@ -419,11 +419,11 @@ def replace_abbr(company, old, new):
 			_rename_record(d)
 	try:
 		frappe.db.auto_commit_on_many_writes = 1
-		frappe.db.set_value("Company", company, "abbr", new)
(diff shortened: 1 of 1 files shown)


nabinhait:
@nextchamp-saqib I merged this PR without noticing `dont-merge` label. Is there any work pending?

nextchamp-saqib:
> @nextchamp-saqib I merged this PR without noticing `dont-merge` label. Is there any work pending?

No, I was just going to add another fix to it. It won't cause any problem.

PR #27066, opened 2021-08-21
Bug Related Setting Batch Number in Item Table

tailorraj:
Bug While Scanning Batch number in **Scan Barcode Field**

Issue: When we scan batch numbers in the scan barcode field the item gets fetched and also the batch number, this fetched batch number is overwritten by the set_batch_number function.

Resolution:
Whenever we scan batch numbers in the scan barcode table, the set_batch_number should check whether the batch number is there or not, if not then only get the batch numbers via FIFO allocation.

### DIFF
--- a/erpnext/selling/sales_common.js
+++ b/erpnext/selling/sales_common.js
@@ -388,7 +388,7 @@ erpnext.selling.SellingController = class SellingController extends erpnext.Tran
 	*/
 	set_batch_number(cdt, cdn) {
 		const doc = frappe.get_doc(cdt, cdn);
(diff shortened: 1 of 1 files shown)


marination:
@tailorraj I believe the same is fixed here > https://github.com/frappe/erpnext/pull/27081
Can you confirm if https://github.com/frappe/erpnext/pull/27081 resolves your issue? We can close this pull request

rohitwaghchaure:
Already fixed https://github.com/frappe/erpnext/pull/27081

PR #27086, opened 2021-08-23
fix: Don't create inward SLE against SI unless is internal customer enabled

marination:
**Issue:**
- Due to duplication/data import **Target Warehouse** could be copied onto multiple Sales Invoices from one Sales Invoice
- If target warehouse exists accidentally, and the customer is not actually an internal customer, inward SLEs were created against such Sales Invoices

**Fix:**
- Check for internal customer along with target warehouse before making SLEs
- Test to check if inward SLE is made if target warehouse is accidentally set but customer is not internal

> In this case a user messed around with property setter and disabled no copy on this hidden target warehouse field in Sales Invoice. `no_copy` is enabled otherwise. This is a defensive fix. Either way the check should be stricter before making something as irreversible as SLEs.

> **Another Note:** It seems like before the introduction of internal customer, target warehouse inward entries were the norm if target warehouse exists (a test exists for this behaviour). Now since target warehouse is hidden if not internal customer this behaviour is discontinued.


marination:
`test_closing_entry` failing

PR #27124, opened 2021-08-24
feat: Improve Product Bundle handling

GangaManoj:
## 1. Gross Profit Report

_Problem:_
- When grouped by Invoice, the Gross Profit Report isn't very readable, with several rows being displayed for each Invoice, representing each Item in the Invoice.
- In the case of Product Bundles, only the Parent Item is displayed, and not the Child Items.

_Proposed Solution:_
- Display the data in a tree format, with the Sales Invoice at the first level, Product Bundles and non-Bundle Items at the second level, and Child Items at the third.

<details>
<summary>Images</summary>

_Before:_

Screenshot 2021-08-25 at 3 10 01 AM

_After:_

Screenshot 2021-08-25 at 12 35 18 AM
</details>

## 2. Selling Settings

_Proposed Changes:_

- Add a new checkbox: _Calculate Product Bundle Price based on Child Items' Rates_

Screenshot 2021-08-24 at 11 46 10 PM

On enabling _Calculate Product Bundle Price based on Child Items' Rates_: 
- The Rate column of Packed Items will be made editable.
- The rate and price of Product Bundles in the Items table will be updated based on the rates of its child Items.

_Note_:
If the Rate of the Product Bundle is changed now, it will get reset to the sum based on the rates of its child items on saving the doc.

<details>
<summary>Example</summary>

_Bundle Item Rates:_

Screenshot 2021-08-24 at 11 54 33 PM

_Parent Item's Rate and Amount:_

Screenshot 2021-08-24 at 11 52 23 PM

</details>

## 3. Quotation

_Proposed Change:_

- Add a new table called _Bundle Items_ to display the child Items of all Parent Bundles listed in the Items table.

Screenshot 2021-08-27 at 2 25 16 AM

## 4. Purchase Order

- When creating a Purchase Order from a Sales Order, if there are any Product Bundles in the Items table, they'll be replaced with their child Items.

<details>
<summary>Images</summary>

_Sales Order:_

Screenshot 2021-08-30 at 9 18 27 PM

_Purchase Order:_

Screenshot 2021-08-31 at 8 35 47 PM

</details>

- Add a new read-only field called Product Bundle in the Items table to store the name of the Product Bundle when the mapping specified above takes place.

Screenshot 2021-08-31 at 8 38 30 PM

- Make Sales Order field editable so POs can be linked with them later.

Editable SO Link

## 5. Purchase Receipt

- Add a new read-only field called Product Bundle in the Items table to store the names of Product Bundles when a Purchase Receipt is created from a Purchase Order that was created from a Sales Order containing Product Bundles.

Screenshot 2021-09-01 at 4 15 09 PM

## 6. Purchase Invoice

- Add a new read-only field called Product Bundle in the Items table to store the names of Product Bundles when a Purchase Invoice is created directly from a Purchase Order that was created from a Sales Order containing Product Bundles, or from a Purchase Receipt that was created from such a Purchase Order.

Screenshot 2021-09-01 at 4 14 17 PM

`no-docs`

<details>

_For Gross Profit Report:_
1. Open the report.
2. Set 'Group By' to 'Invoice'.
3. Verify if the items are displayed in a tree format with the Invoice at the top, followed by Product Bundles/Non-Bundle Items, and then the Child Items(if any).
(If confused, refer image under Gross Profit Report > Images > After of this PR description)

_For Calculate Product Bundle Price based on Child Items' Rates_:
1. Open Selling Settings.
2. Check _Calculate Product Bundle Price based on Child Items' Rates_.
3. Create a Product Bundle.
4. Create a new Quotation.
5. Add the Product Bundle to the Items table of the Quotation.
6. Save.
7. Check the Bundle Items table. The sum of (Qty x Rate) of all the Bundle Items with parent as the Product Bundle in question should be equal to the new Amount of the Product Bundle in the Items table, and its rate should be this new Amount/Qty.
8. Also check if the Rate column of the Bundle Items table is now editable.
9. Open Selling Settings again and uncheck _Calculate Product Bundle Price based on Child Items' Rates_.
10. Now check if the Rate column of the Bundle Items table is read-only.
11. If you change the rate of the Product Bundle in the Items table now, it will not be reset to the Amount based on the Bundle Items' rates on saving anymore.

_For Quotation:_
1. Create a Product Bundle.
2. Create a new Quotation.
3. Add the Product Bundle to the Items table of the Quotation.
4. Save.
5. Check if the Bundle Items table was created.

_For Purchase Order:_
1. Create a new Purchase Order.
2. Add an Item.
3. Expand the Item row.
4. Check if the Sales Order field under the 'Warehouses and References' section is editable.

1. Create a new Sales Order.
2. Now try adding a Product Bundle, save and submit.
3. Create > Purchase Order.
4. Check if the Packed Items in the SO had replaced the Product Bundle in the Items table of the PO.

</details>


erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -354,6 +422,107 @@ def load_invoice_items(self):
...
+					'item_group': None,
+					'brand': None, 
+					'dn_detail': None, 
+					'delivery_note': None, 
+					'qty': None, 
+					'item_row': None, 
+					'is_return': row.is_return, 
+					'cost_center': row.cost_center,
+					'base_net_amount': frappe.db.get_value('Sales Invoice', row.parent, 'base_net_total')
+				})
```
nextchamp-saqib: Can be extracted into a method

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -103,6 +106,45 @@ def cleanup_packing_list(doc, parent_items):
 		if d not in delete_list:
 			doc.append("packed_items", d)
 
+def update_product_bundle_price(doc, parent_items):
+	"""Updates the prices of Product Bundles based on the rates of the Items in the bundle."""
+
+	parent_items_index = 0
+	bundle_price = 0
+	parent_items_doctype = doc.items[0].doctype
```
nextchamp-saqib: Will this break if `doc.items` is empty?

nextchamp-saqib (changes requested):

erpnext/stock/doctype/packed_item/packed_item.py
GangaManoj: Theoretically, yes, but why would anyone save an Invoice/Order that doesn't have any Items?

nextchamp-saqib: Hmm, but better return if there are no items

erpnext/stock/doctype/packed_item/packed_item.json
```diff
@@ -215,13 +216,20 @@
    "fieldname": "conversion_factor",
    "fieldtype": "Float",
    "label": "Conversion Factor"
+  },
+  {
+   "fetch_from": "item_code.valuation_rate",
+   "fieldname": "rate",
+   "fieldtype": "Currency",
+   "in_list_view": 1,
+   "label": "Rate"
```
nextchamp-saqib:
```suggestion
   "label": "Rate",
   "read_only": 1
```

erpnext/accounts/report/gross_profit/gross_profit.py
GangaManoj: Done.

erpnext/stock/doctype/packed_item/packed_item.json
GangaManoj: Is this necessary, considering we're already using Property Setters to achieve this?

erpnext/stock/doctype/packed_item/packed_item.py
GangaManoj: Done.

erpnext/stock/doctype/packed_item/packed_item.json
nextchamp-saqib: Initially, the rate column will be editable and the check would be disabled which shouldn't be the case

GangaManoj: Ah, that makes sense.

nextchamp-saqib (approved):

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -39,8 +39,10 @@ def update_packing_list_item(doc, packing_item_code, qty, main_item_row, descrip
 	# check if exists
 	exists = 0
 	for d in doc.get("packed_items"):
-		if d.parent_item == main_item_row.item_code and d.item_code == packing_item_code and\
-				d.parent_detail_docname == main_item_row.name:
+		if d.parent_item == main_item_row.item_code and d.item_code == packing_item_code:
+			if d.parent_detail_docname != main_item_row.name:
+				d.parent_detail_docname = main_item_row.name
```
marination, line 44: This could use a comment, the logic is very confusing

erpnext/stock/doctype/packed_item/packed_item.py
```diff
@@ -103,6 +108,40 @@ def cleanup_packing_list(doc, parent_items):
...
+		
+def update_parent_item_price(doc, parent_item_code, bundle_price):
+	parent_item_doc = doc.get('items', {'item_code': parent_item_code})[0]
+
+	current_parent_item_price = parent_item_doc.amount
+	if current_parent_item_price != bundle_price:
+		parent_item_doc.amount = bundle_price
+		update_parent_item_rate(parent_item_doc, bundle_price)
+
+def update_parent_item_rate(parent_item_doc, bundle_price):
```
marination, line 142: `update_parent_item_price` vs `update_parent_item_rate`: without context, hard to tell the difference.

PR #27141, opened 2021-08-25
feat: provision to create customer from opportunity

anupamvs:
opp-to-customer

`no-docs`

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.py
+++ b/erpnext/crm/doctype/opportunity/opportunity.py
@@ -288,6 +288,24 @@ def make_request_for_quotation(source_name, target_doc=None):
 	return doclist
 
 @frappe.whitelist()
+def make_customer(source_name, target_doc=None):
+	def set_missing_values(source, target):
+		if source.opportunity_from == "Lead":
+			target.lead_name = source.party_name
+
+	doclist = get_mapped_doc("Opportunity", source_name, {
+		"Opportunity": {
+			"doctype": "Customer",
+			"field_map": {
+				"currency": "default_currency",
+				"customer_name": "customer_name"
+			}
+		}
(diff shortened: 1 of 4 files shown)


erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -287,6 +287,24 @@ def update_item(obj, target, source_parent):
...
+	def set_missing_values(source, target):
+		if source.opportunity_from == "Lead":
+			target.lead_name = source.party_name
+
+	doclist = get_mapped_doc("Opportunity", source_name, {
+		"Opportunity": {
+			"doctype": "Customer",
+			"field_map": {
+				"currency": "default_currency",
+				"customer_name": "customer_name"
```
ruchamahabal, line 301: 1. Also, add a "**From Opportunity**" field in Customer so that users can pull in relevant opportunity custom fields using fetch from while creating the customer from opportunity.
2. We need to carry forward address and contact linking to the Customer document too. So after adding this opportunity field in the customer on the `after_insert` event:
    - Add reference link of the **customer** in all the addresses and contacts linked to that opportunity
    - We can also do the same thing for the "**From Lead**" field so that lead addresses also get carry forwarded

ruchamahabal (changes requested):

erpnext/crm/doctype/opportunity/opportunity.py
ruchamahabal: While creating an address from opportunity, it's linked to lead only. So no need to link opportunity-related address contacts separately with the customer.


https://user-images.githubusercontent.com/24353136/130798041-c54b3cbc-24b4-40c7-ab26-660c271f1740.mp4

ruchamahabal (approved):

PR #27169, opened 2021-08-26
fix: Remove duplicate `create_internal_customer` utility

marination:
Introduced via https://github.com/frappe/erpnext/pull/27086

Already on version-13-hotfix via https://github.com/frappe/erpnext/pull/27166


marination:
Unrelated `test_closing_entry` failing

PR #27170, opened 2021-08-26
fix: negative qty validation on stock reco cancellation

ankush:
Steps to reproduce:

1. Use any item with 0 stock or make stock zero with SR (stock reconciliation). 
2. Make qty 10 using SR
3. Create a delivery note with 10 qty. 
4. Cancel the stock reco that added 10 qty.

Observed: SR cancels.
Expected: SR cancellation gets blocked due to negative qty in future DN.

Root Cause:
In stock reconciliation cancellation negative stock setting is ignored as `db.get_value` is returning string `'0'` which is not cast to int/bool for further logic. 

This causes negative qty, which eventually gets caught by reposting but by design this should stop cancellation.

Solution: cast setting value correctly, added test for regression.


marination:
Unrelated `test_closing_entry` failing

marination (approved):

PR #27179, opened 2021-08-26
chore: remove deprecated and empty QUnit tests

ankush:
Remove all empty QUnit tests. They've been deprecated for some time now. 

Non-empty tests are kept for reference for migration in future.


ankush:
Bulk deleted using:
`ls **/doctype/**/test*.js | xargs wc -l | grep 23 | awk '{print $2}' | xargs rm`

nabinhait (approved):

ankush:
UI test failures are irrelevant... will get it fixed separately.

PR #27197, opened 2021-08-27
feat: (consistency) Add Primary Address and Contact section in Supplier

marination:
- The same is present in customer and is inconsistent with supplier
   Screenshot 2021-08-27 at 3 00 02 PM
    Screenshot 2021-08-27 at 6 07 47 PM

- Users can have multiple **primary** addresses against a supplier, they can choose one with the highest priority among them here

`no-docs`


erpnext/selling/doctype/customer/customer.py
```diff
@@ -246,9 +252,15 @@ def validate_credit_limit_on_change(self):
...
-				where name=%s""", self.name)
+			frappe.db.sql(f"""
+				UPDATE `tabCustomer`
+				SET
+					customer_primary_contact=null,
+					customer_primary_address=null,
+					mobile_no=null,
+					email_id=null,
+					primary_address=null
+				WHERE name='{self.name}'""")
```
ankush: Why f-string here?

ankush (approved):

ankush:
Coveralls died for some reason. Not test failure.

PR #27195, opened 2021-08-27
fix: patches were breaking while migrating

shadrak98:
Fixes for the following errors occurred while migrating:

- frappe.exceptions.ValidationError: Fieldtype cannot be changed from Data to Currency
- frappe.exceptions.LinkValidationError: Could not find Report: E-Invoice Summary
- pymysql.err.OperationalError: (1054, "Unknown column 'first_response_time' in 'field list'")
- TypeError: unsupported operand type(s) for *: 'NoneType' and 'int'
- frappe.exceptions.LinkExistsError: Cannot delete or cancel because Report "A" is linked with Desktop Icon "B" 
- frappe.database.database.InvalidColumnName: Invalid field name

PR #27200, opened 2021-08-27
fix: patches were breaking during migration

shadrak98:
Fixes for the error:

- Only Options allowed for Data field are: Email, Name, Phone
- frappe.exceptions.MandatoryError: [Item Tax Template, *********]: company (for multiple companies)


ankush:
This also needs to be backported after ports of https://github.com/frappe/erpnext/pull/27195 are merged. (avoiding now to avoid conflicts) @shadrak98

PR #27214, opened 2021-08-28
fix: cannot create task with *Apply Strict User Permissions*

ahmadpak:
If user permissions is set for company and apply strict user permissions is set. The system does not allow to create tasks in the tree view


erpnext/projects/doctype/task/task.py
```diff
@@ -318,6 +318,14 @@ def add_node():
 
 	if args.parent_task == 'All Tasks' or args.parent_task == args.project:
 		args.parent_task = None
+		if args.parent_task == args.project:
+			args.company == frappe.get_value('Project', args.project, 'company')
+		else:
+			from frappe.defaults import get_user_default
```
ankush: Any specific reason for not making this global import at top?

ankush: Also @ahmadpak can you please refresh the PR by rebasing/merging `develop`.

ahmadpak: No specific reason. I will do import to the top also rebase

PR #27228, opened 2021-08-30
fix: Stock Ageing report issues for serialized items

### LINKED ISSUE #27225: Stock Ageing report showing incorrect data
WhatsApp Image 2021-08-30 at 11 57 22 AM

I believe non-serialized items' data are correct. But for serialized items, the report shows erroneous data.
ERPNext v13

rtdany10:
Stock ageing report shows incorrect data for serialized items because:
- get_range_age() assumes item[0] to be int/float while it is serial number(str) for serialized items.
- fifo_queue.remove(serial_no) causes shift in index of the list and thereby not looping through every object in the list.

Closes #27225
#no-docs


rtdany10:
@ankush @marination @rohitwaghchaure 
Backport to version-13 as well, please.

ankush (approved):

PR #27230, opened 2021-08-30
ci: concurrency control for CI jobs

ankush:
similar to https://github.com/frappe/frappe/pull/14061

This will ensure that only one instance of the following tests run per PR and cancel previous running/queued jobs when new commits are pushed. 

- Server / unit tests
- UI tests
- Patch test


ankush:
Backports are not required. Active PRs are supposed to be on `develop` only.

PR #27240, opened 2021-08-30
ci: switch to latest LTS Ubuntu version

ankush:
This change was deferred earlier as it was causing some failures.
However, a huge userbase is already on latest LTS, so it makes sense to bump OS version in CI too. 

---

Also dropped translation linter as it is migrated to semgrep.


ankush:
ignore failing UI test, it's a _known failure_

.github/workflows/patch.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
```
gavindsouza: Should we use `ubuntu-latest` instead?

ankush: I thought it will auto-bump to 22.04 _immediately_.  But it seems Github takes some time after the official release, so it should be okay.

gavindsouza (approved):

PR #27252, opened 2021-08-31
fix: pass labels through translation function

18alantom:
### Description
Some noob forgot to call `_` on the labels in the Process Loss Report. Hence added them.


ankush (approved):

ankush:
@18alantom can you take a stab at formulating semgrep rule for this? So all new reports / new fields will have translate calls. 

You'll have to duplicate this rule to apply specifically on report labels. 

https://github.com/frappe/erpnext/blob/88aa7b824e3e34fa7f67c1e3ebbc2787cb87c53d/.github/helper/semgrep_rules/ux.yml#L2-L13

PR #27278, opened 2021-09-01
fix: production plan UX and validation message

rohitwaghchaure:
Changed validation message

Screenshot 2021-09-01 at 8 33 33 AM

While adding the items, the Material Request Warehouse will be use as a Target Warehouse
Screenshot 2021-09-01 at 8 40 53 AM

While downloading the raw materials if there is no conversion factor set between Purchase UOM and Stock UOM then system was throwing the below error

Screenshot 2021-09-01 at 8 52 00 AM

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -331,7 +331,7 @@ class ProductionPlan(Document):
 	def get_production_items(self):
 		item_dict = {}
 		for d in self.po_items:
-			item_details= {
+			item_details = {
 				"production_item"		: d.item_code,
 				"use_multi_level_bom"   : d.include_exploded_items,
 				"sales_order"			: d.sales_order,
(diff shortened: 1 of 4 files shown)


erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -468,6 +468,7 @@ def create_work_order(self, item):
 			wo.fg_warehouse = warehouse.get('fg_warehouse')
 		try:
 			wo.flags.ignore_mandatory = True
+			wo.flags.ignore_validate = True
```
ankush: Why ignore validate? This will also skip other methods:

https://github.com/frappe/erpnext/blob/f1b77360ee3b5cc418f074aefa83a1871964da99/erpnext/manufacturing/doctype/work_order/work_order.py#L41-L55

rohitwaghchaure: Fixed, added to test some cases and then forgot to remove it.

ankush (approved):

PR #27280, opened 2021-09-01
ci(semgrep): add translation checks for report labels

18alantom:
Add translation checks for report labels

I know this 👇  looks ugly
```
      - pattern: |
          {..., "label": "...", ...}
```
but `yaml` asks me to wrap the pattern in quotes when writing it this way:
```
      - pattern: {..., "label": "...", ...}
```

### DIFF
--- a/.github/helper/semgrep_rules/report.yml
+++ b/.github/helper/semgrep_rules/report.yml
@@ -3,6 +3,8 @@ rules:
   paths:
     include:
     - "**/report"
+    exclude:
+    - "**/regional"
   pattern-either:
   - patterns:
       - pattern: |
--- a/.github/helper/semgrep_rules/report.yml
+++ b/.github/helper/semgrep_rules/report.yml
@@ -1,0 +1,19 @@
+rules:
+- id: frappe-missing-translate-function-in-report-python
+  paths:
+    include:
+    - "**/report"
+  pattern-either:
+  - patterns:
+      - pattern: |
+          {..., "label": "...", ...}
+      - pattern-not: |
+          {..., "label": _("..."), ...}
+  - patterns:
+      - pattern: dict(..., label="...", ...)
+      - pattern-not: dict(..., label=_("..."), ...)
+  message: |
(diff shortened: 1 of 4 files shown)


.github/helper/semgrep_rules/ux.yml
```diff
@@ -7,6 +7,14 @@ rules:
...
       - pattern: frappe.throw("...", ...)
       - pattern-not: frappe.throw(_("..."), ...)
+  - patterns:
+      - pattern: |
+          {..., "label": "...", ...}
+      - pattern-not: |
+          {..., "label": _("..."), ...}
+  - patterns:
+      - pattern: dict(..., label="...", ...)
+      - pattern-not: dict(..., label=_("..."), ...)
```
ankush: One minor change to avoid false +ves: Target this to `**/report/` glob pattern only, you'll have to duplicate the rule for this.
example:
https://github.com/frappe/erpnext/blob/a60a3d1c8b6a3e3f2eeb0c495ebb39a403382441/.github/helper/semgrep_rules/frappe_correctness.yml#L100-L102

18alantom:
Screen Shot 2021-09-01 at 13 47 23

@ankush 2 FNs in the tests, this doesn't seem to be happening in `erpnext`; I ran the rule and checked the output. Any idea why this is happening?

ankush:
@18alantom that's probably because test runner is respecting the glob pattern. It's a bug IMO. Ignore it for now.

.github/helper/semgrep_rules/report.yml
```diff
@@ -0,0 +1,19 @@
+rules:
+- id: frappe-missing-translate-function-in-report-python
+  paths:
+    include:
+    - "**/report"
```
ankush, line 5:
```suggestion
  paths:
    include:
    - "**/report"
    exclude:
    - "**/regional"
```

Checked current failures, more than 1/2 are regional reports. They won't need translations.

ankush (approved):

PR #27281, opened 2021-09-01
feat: link items to supplier / customer

DeeMysterio:
**Changes:**

- Replaced the DocType "Supplier Item Group" with a new one called "Party Specific Items" to link specific items based on item code/item brand/item group to a specific Customer / Supplier while creating purchase/sales transactions.

**Screenshots/Gifs:**

psi

**Documentation:**
https://docs.erpnext.com/docs/v13/user/manual/en/party_specific_item


erpnext/selling/doctype/party_specific_items/party_specific_items.py
```diff
@@ -0,0 +1,18 @@
...
+	def validate(self):
+		exists = frappe.db.exists({
+			'doctype': 'Party Specific Items',
+			'party_type': self.party_type,
+			'party': self.party,
+			'restrict_based_on': self.restrict_based_on,
+			'based_on': self.based_on,
+		})
+		if exists:
+			frappe.throw(_(f"This item filter has already been applied for the {self.party_type}"))
```
ankush:
```suggestion
			frappe.throw(_(f"This item filter has already been applied for the {0}".format(self.party_type)))
```

https://frappeframework.com/docs/user/en/translations

erpnext/patches/v13_0/replace_supplier_item_group_with_party_specific_items.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2019, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+from __future__ import unicode_literals
```
ankush:
```suggestion
```

Not required anymore. We don't support py2.

erpnext/patches/v13_0/replace_supplier_item_group_with_party_specific_items.py
```diff
@@ -0,0 +1,25 @@
...
+
+import frappe
+
+def execute():
+	sig = frappe.db.get_all("Supplier Item Group", fields=["name", "supplier", "item_group"])
```
ankush: This doctype was recently added so patches will fail unless you do `reload_doc` on all the relevant doctypes.

erpnext/patches/v13_0/replace_supplier_item_group_with_party_specific_items.py
```diff
@@ -0,0 +1,25 @@
...
+		psi.based_on = item.item_group
+		psi.insert()
+
+	supplier_dt = frappe.get_doc("DocType", "Supplier")
+	for link in supplier_dt.links:
+		if link.link_doctype == "Supplier Item Group":
+			supplier_dt.links.remove(link)
+			frappe.delete_doc("Supplier Item Group", link.name, force=1)
+			supplier_dt.append('links', {'link_doctype': 'Party Specific Items', 'link_fieldname':'party'})
+			supplier_dt.save()
```
ankush:
```suggestion
```

This can/should be done by editing doctype?

erpnext/selling/doctype/party_specific_items/party_specific_items.py
DeeMysterio: yeah, saw Semgrep failing at there, writing test!

erpnext/selling/doctype/party_specific_item/party_specific_item.py
```diff
@@ -0,0 +1,19 @@
...
+
+import frappe
+from frappe import _
+from frappe.model.document import Document
+
+
+class PartySpecificItem(Document):
+	def validate(self):
+		exists = frappe.db.exists({
+			'doctype': 'Party Specific Items',
```
rohitwaghchaure: Party Specific Items has been changed to Party Specific Item

erpnext/patches/v13_0/replace_supplier_item_group_with_party_specific_item.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2019, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+import frappe
+
+
+def execute():
+	if frappe.db.table_exists('Supplier Item Group'):
+		frappe.reload_doc("buying", "doctype", "supplier_item_group")
```
rohitwaghchaure: Instead of supplier_item_group, reload party_specific_item table

rohitwaghchaure:
Set title field as Party
Screenshot 2021-09-13 at 5 59 16 PM

rohitwaghchaure (approved):

chechani:
Thanks for introducing this Party Specific items , Its a co-incidence that yesterday we discussed with one client the same requirement and we found introduced the same day. We learnt that this setting applied to Sales Invoice and Sales Order but not to Quotation. We suggest that it should be applied to Quotation and option may be given on which doc type it should be applied for more flexibility.

PR #27293, opened 2021-09-01
fix: Zero division error while fetching unconsumed materials

marination:
**Issue:**
```py
 File ".../apps/erpnext/erpnext/stock/doctype/stock_entry/stock_entry.py", line 1003, in get_items
    self.get_unconsumed_raw_materials()
  File ".../apps/erpnext/erpnext/stock/doctype/stock_entry/stock_entry.py", line 1200, in get_unconsumed_raw_materials
    (flt(work_order_qty) - flt(wo.produced_qty))
```

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -1200,10 +1200,10 @@ class StockEntry(StockController):
 
 			wo_item_qty = item.transferred_qty or item.required_qty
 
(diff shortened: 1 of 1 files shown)


marination:
For the reviewer: Wouldnt defaulting to 1 mean that 1 qty is yet to produce ? Should there be a validation if the qty to produce is 0 instead of a fallback denominator?

rohitwaghchaure (approved):

PR #27298, opened 2021-09-01
fix: General Ledger translation issues

### LINKED ISSUE #27277: General ledger error when choose language not english
due to this commit 
https://github.com/frappe/erpnext/commit/ab76ff0836cf658e68a7509f64f325fa8814c70e

the contrbuter added a translation to groub by 
			"options": ["", __("Group by Voucher"), __("Group by Voucher (Consolidated)"),
				__("Group by Account"), __("Group by Party")],
as you can see if this values is translated ,, it will be used in the prepared report in Arabic [ the language I used] 
that will make the report missed up . 

the fix for these option to be English ,

rtdany10:
Options of group_by filter are translated. Assertions were made against English sentences and hence would always return false in cases where English isn't the system language.

Solution: don't translate options server side. Only displayed labels are translated client side. 

closes #27277

### DIFF
--- a/.github/helper/semgrep_rules/report.yml
+++ b/.github/helper/semgrep_rules/report.yml
@@ -19,3 +19,16 @@ rules:
   languages: [python]
   severity: ERROR
 
+- id: frappe-translated-values-in-business-logic
+  paths:
+    include:
+    - "**/report"
+  patterns:
+    - pattern-inside: |
+        {..., filters: [...], ...}
+    - pattern: |
+        {..., options: [..., __("..."), ...], ...}
+  message: |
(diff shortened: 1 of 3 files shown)


rtdany10:
@ankush unrelated test failing. Can we merge this?

rtdany10:
@ankush

ankush (changes requested):
A better solution would be to remove translations from business logic altogether.

The options are anyways translated, so it's not required in the code. 

Example: `value` attribute contains actual value while display label can be whatever translate function returns. 

Screenshot 2021-09-07 at 5 22 32 PM

ankush:
Oh well, removing all translation calls also removes it from `_messages` 🤦

ankush (approved):

PR #27300, opened 2021-09-01
fix: Braintree payment processed twice

nemesis189:
**Problem:**
Paid twice by clicking on the same payment link, creates a second integration request for the exact same payment request as before and no payment entry is created for the second payment

**Solution:**
Added a validation method to check if the Payment Request is already paid

**Ref Issue:** ISS-21-22-05023


erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -542,3 +542,13 @@ def set_missing_values(source, target):
...
+
+def validate_payment(doc, method=""):
+	if not frappe.db.has_column(doc.reference_doctype, 'status'):
+		return
+
+	status = frappe.db.get_value(doc.reference_doctype, doc.reference_docname, 'status')
+	if status == 'Paid':
+		frappe.throw(_("The Payment Request {0} is already paid, cannot process payment twice").format(doc.reference_docname))
+	else:
+		return
```
nextchamp-saqib:
```suggestion
```

nextchamp-saqib (approved):

PR #27318, opened 2021-09-02
feat: currency exchange settings

### LINKED ISSUE #27317: Configurable currency exchange API
Currently, the exchange rate is fetched using an hardcoded API. There are plenty of exchange rate service providers and each user might have different preferences. So, it will be better to make it configurable.

> @rtdany10 maybe we should make this configurable. Currently it's difficult for a user to know where the exchange rates are coming from, and the API is changed frequently. It is a legal requirement in Europe to make sure that exchange rates are trusted and accurate, and that the source is known.

_Originally posted by @barredterra in https://github.com/frappe/erpnext/issues/26763#issuecomment-910505590_

A Currency Exchange Settings page to configure API, parameters.

image

rtdany10:
Currency Exchange Settings will enable users to use exchange rate provider of their choice.
CES Change API

Closes #27317
#no-docs (temp)


erpnext/setup/doctype/currency_exchange_settings/currency_exchange_settings.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) 2021, Wahni Green Technologies Pvt. Ltd. and contributors
```
ankush: This doctype/settings belong to accounting. It was there in the setup folder mostly because of legacy code. There's no reason to keep it that way now 😅

ankush (changes requested):
@marination @deepeshgarg007 @nextchamp-saqib @barredterra 

Your comments? From UX perspective most users won't understand what endpoint/parameters etc even means. It might be a better idea to have a list of providers from which users can choose the one they want.

marination:
@ankush I agree with your solution but have a different reasoning.

@rtdany10 
My concern is error handling. Most APIs have their quirks and will need special error handling. Making something as generic (the API field) for currency exchange rates doesn't sound like the best idea stability wise. Setup may go through, but what about after that ? Things like unsupported currencies and the service being down need to be handled separately.

I would also suggest giving a list of services, that we support, to choose from so that we know that these are well handled and compatible with ERPNext. We can incrementally build on these services as and when there is demand for them.

rtdany10:
> Things like unsupported currencies and the service being down need to be handled separately.

I believe that the function to get exchange rate from the provider does handle this. It returns 0.00 and logs an error. 

But again, as pointed out, having some select-able pre-configured options will make it easier for the end user, like in the case of Email Accounts where certain service providers can be selected without configuring Email Domain.

I can add some service providers; would be nice if someone shares a preferred list of service providers. @barredterra and other end users.

@ankush @marination

barredterra:
@rtdany10 let's go with exchangerate.host and frankfurter.app for now. The official API of the European Central Bank would also be interesting: https://sdw-wsrest.ecb.europa.eu/help/

rtdany10:
https://user-images.githubusercontent.com/52111700/140075928-bf584750-0c05-435b-96e2-5d7d6051b619.mp4

I hope this is fine. Will make the fields read only.

erpnext/accounts/doctype/currency_exchange_settings/currency_exchange_settings.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2021, Wahni Green Technologies Pvt. Ltd. and contributors
```
rtdany10: ToDo: Avoid code repetition.

deepeshgarg007:

PR #27328, opened 2021-09-03
fix: add lab test rate to default selling price list

rtdany10:
Billable lab test's rates weren't added to the default selling price list in cases where multiple selling price lists were present.
This PR adds the rates to the default one.


rtdany10:
@pateljannat please add backport labels as well.

ChillarAnand:
@rtdany10 Thanks for submitting the patch. 

We are moving healthcare module into a seperate app and it is available at https://github.com/frappe/healthcare. All new features/fixes will be developed here.  

The above patch will be available in v13 hotfix as well. Tracking it at https://github.com/frappe/erpnext/pull/27353

PR #27330, opened 2021-09-03
fix: manually added weight per unit reset to zero after save

rohitwaghchaure:
Due to recent [changes](https://github.com/frappe/erpnext/pull/26619), manually set weight per unit reset to zero.

When user make the delivery from the sales order, they manually set the Weight Per Unit against the item but after saving of the delivery note the rate reset to zero.

weight_per_unit_issue


marination (approved):

marination:
@rohitwaghchaure please fix linter errors (whitespace)

PR #27348, opened 2021-09-06
fix (refactor): Tax Withholding for Advances using Payment Entry against suppliers

deepeshgarg007:
Currently, deducting TDS was only possible via Purchase Invoice. But in many scenarios, TDS had to be deducted against Payment for PO or directly in Payment Entry. This PR adds functionality to make that process simpler for users

Steps to test are mentioned in the below docs. Please check both the steps using PO and without PO

https://docs.erpnext.com/docs/v13/user/manual/en/accounts/tax-withholding-category#32-deducting-tax-at-source-on-advances

Also, verify the GL Entries for the vouchers as well which should be as follows:

Here a PO was made for the amount 5000 with no taxes applied

1. Payment Entry Ledger
image

2. Purchase Invoice Ledger
image


deepeshgarg007:

Issue #19413, opened 2019-10-25
Depends on Payment Days is not working in formula calculation


cospin:
_Discuss link: https://discuss.erpnext.com/t/depends-on-payment-days-is-not-working-in-formula-calculation/54270_

## Description of the issue
When using a variable whose value is affected by "Depends on Payment Days" in the formula of any other component, it takes the total value and not the value with the "Depends on Payment Days".

## Context information

**Output of `bench version`**
```
erpnext 12.1.7
frappe 12.0.17
```

## Steps to reproduce the issue

1. Create a new _Salary Structure_
2. Add a new _Salary Component_ for example Basic Salary (`BS`) with a formula value as: `base`, and set _Depends on Payment Days (DPD)_ as **enabled**.
3. Add a new _Salary Component_ for example Basic Salary Two (`BS2`) with a formula value as: `BS`, and with _Depends on Payment Days (DPD)_ **disabled**.
4. Test the _Salary Structure_ with an employee that has less _Payment Days_ than _Working Days_

#### Example configuration

Working Days: 30
Payment Days: 15
`base`: 1.000

Basic Salary (`BS`) = `base`
Basic Salary Two (`BS2`) = `BS`

### Observed result
`BS` = 500 _(Correct as it's affected by DPD)_
`BS2` = 1.000 _(Incorrect as `BS` has a value of 500, it should be also 500)_

### Expected result
`BS` = 500
`BS2` = 500

**My scenario is as follows:**

I need to calculate other components, taking into account the `BS` (affected by _DPD_), but adding another variable that should not be affected by _DPD_, example:

`BS` = `base` = 1.000 = 500 (with _DPD_)
`X` = 5.000 _(Fixed value that should not be affected by DPD)_
`BS2`= `BS` + `X` = 5.500 (But right know it displays 6.000 incorrectly)

clarkejj:
Thanks for reporting this issue - two requests

Please also post this to discuss.erpnext.com,. That way all the community is informed and benefits, and can expedite responses and possible fixes, thanks

Add a link to this post here, and on discuss to link to here

cospin:
Thanks! Done!

clarkejj:
Thanks https://discuss.erpnext.com/t/depends-on-payment-days-is-not-working-in-formula-calculation/54270

onlinebizsoft:
I have a few questions too
1. What counted in "Payment days" ?
[...]

PR #27349, opened 2021-09-06
fix(Payroll): incorrect component amount calculation if dependent on another payment days based component

ruchamahabal:
## Before:

Incorrect component amount calculation if dependent on another payment days based component

**Steps to replicate:** 

1. Create a salary structure. Add a component based on payment days. (eg: HRA)

    based_on_payment_days

2. Add another component that is dependent on this previous component by formula. (eg: EPF)
     `1800 if ((gross_pay - HRA) >21000) else (gross_pay - HRA) * 0.12`
     
    epf

3. Mark the employee Absent for a few days as this works fine when payment days are equal to working days
4. Create a Salary Slip.
     - Base is 29,029
     - So, HRA should be `base * 0.20` =  5805.8.
     - HRA depends on payment days so updated value should be (5805.8 * 20) / 22 = 5278. This is correct
     - Now, EPF is `1800 if ((gross_pay - HRA) >21000) else (gross_pay - HRA) * 0.12`. So it should be `(gross_pay - HRA) * 0.12` = `(15,834.00 - 5278) * 0.12` = 1266.72. But the EPF is amount is 1203 which is incorrect.

      epf-calc

## After:

The method `set_component_amounts_based_on_payment_days` is being called after the earnings and deductions have been calculated and it only updates the components which are based on payment days, not the ones which are dependent on the components which are based on payment days. So EPF (deduction component) gets the older value of HRA (not based on payment days)

**Solution**:  Update component amount based on payment days during component calculation itself so that every other component dependent on it gets the updated value.

correct-pf

Fixes Issue: https://github.com/frappe/erpnext/issues/19413
Support Issue: ISS-21-22-04520
Discuss Thread: https://discuss.erpnext.com/t/depends-on-payment-days-is-not-working-in-formula-calculation/54270


nabinhait (approved):

jigneshpshah:
This is an important update , if someone is using variables like payment_days/total_working_days in formula before this fix , they need to update their formula else will lead to erros in salary calculation

jigneshpshah:
My test results for  the above fix

Update component amount based on payment days during component calculation itself so that every other component  on DEDUCTION Side dependent on it gets the updated value.


#### Employer PF (Expense) _**doesn't**_ get updated value of HRA , but  Employee PF & Employer PF Payable  gets updated value of HRA

Earning 

|  Component |  
|---|
|  Basic |   Yes |  
|   HRA|   Yes  |  
|  Employer PF (Expense) |  

Deduction 

|  Component |  
|---|
|  Employee PF | 
|  Employer PF Payable |  

This will be a critical issue as for  Employer PF - the expense and payable amount would not match

dexterhussain:
The issue is back in V14


I am facing a problem in v14 , If I am making a salary structure whose frequency is monthly, and I want to pay him for 10 days from whole month, I select days from 1-10 march , if the gross salary is 1500 then system is even giving 1500 for whole month instead of giving 150. Anyone who have encounter this issue? depends on payment days is checked in Salary components.
the salary slip even shows working days as 10 but still salary is calculated for whole month.
if I make a salary structure for 1 day and select from 1-30 , even then system is calculating salary for as 50 for whole month.


image

image

dexterhussain:
image

dexterhussain:
no matter how many working days there are, the salary is already calculated for whole month

PR #27360, opened 2021-09-06
fix: missing link in dashboard missing on SI from DN

### LINKED ISSUE #27359: Create Sales Invoices from Delivery Note : Missing dashboard link to Delivery note
Create a Sales Order
Create a Delivery Note from the SO
Create an Sales Invoice from DN

image

after the fix

image

FHenry:
fix #27359

The field `tabDelivery Note Item`.against_sales_invoice is always empty on the process SO=>DN=>SI


FHenry:
bug in version-12, version-13 and develop

noahjacob (dismissed):

FHenry:
What is the other workflow? 
This is the only users use.

nextchamp-saqib:
@FHenry Have you tried using the Document Links section for this?
Recently, https://github.com/frappe/frappe/pull/13823 has provided a way to add internal links in the dashboard.

FHenry:
@nextchamp-saqib 

>  Have you tried using the Document Links section for this?
> Recently, [frappe/frappe#13823](https://github.com/frappe/frappe/pull/13823) has provided a way to add internal links in the dashboard.

Nice in develop branch, what about version-12 and version-13 ???? This is a bug in actual production version that must be fix

marination (approved):

marination:
I think dashboard links in framework right now are problematic for child table linking. So this is the best way to go

marination:

marination:
@FHenry reverting this due to https://github.com/frappe/erpnext/pull/23570 (this fix has been attempted before and breaks, reverting until theres a better solution)

FHenry:
Ok, I understand, a better fix can be set against_sales_invoice when create SI from DN ?

anandbaburajan:
> @FHenry reverting this due to #23570 (this fix has been attempted before and breaks, reverting until theres a better solution)

@marination I've tried to fix it here: https://github.com/frappe/frappe/pull/21722. It works AFAIK, but do you have any ideas on how it might break?

marination:
@anandbaburajan I'm sorry I have forgotten any context I had with this issue since 2021 :(

PR #27374, opened 2021-09-07
fix: added checkbox show remarks in the filters for AR & AP reports

Anuja-pawar:
Added checkbox 'Show Remarks' in the filters to add column 'Remarks' whenever required in the reports Accounts Receivable & Accounts Payable.

ezgif com-gif-maker (4)

### DIFF
--- a/erpnext/accounts/report/accounts_payable/accounts_payable.js
+++ b/erpnext/accounts/report/accounts_payable/accounts_payable.js
@@ -4,7 +4,7 @@
 frappe.query_reports["Accounts Payable"] = {
 	"filters": [
 		{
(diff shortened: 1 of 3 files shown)


deepeshgarg007:
@Anuja-pawar This was recently removed by @rohitwaghchaure via https://github.com/frappe/erpnext/pull/26679

nextchamp-saqib:
I remember it was removed due to performance reasons

Anuja-pawar:
@deepeshgarg007 Is there any particular reason to do that? A customer have raised an issue, 'cause they need this column to be in the reports

nextchamp-saqib (approved):

PR #27379, opened 2021-09-07
feat: POS write off feature

nemesis189:
Added the **Write Off** feature in POS system. Any outstanding amount is considered as the write off amount and is written off.

**Updates:**
- Created **Write Off** section and added checkbox for the same in POS Profile
- Displaying 'Write Off Amount' instead of 'Change Amount' or 'To Be Paid' 
- Updated `pos_invoice_merge_log.py` to create entries for write off amounts on merging invoices

   
   

**Sample of GL Entries after creating POS Merge Log:**
image


erpnext/accounts/doctype/pos_profile/pos_profile.json
```diff
@@ -350,6 +353,22 @@
...
+   "fieldtype": "Section Break",
+   "label": "Write Off"
+  },
+  {
+   "fieldname": "column_break_45",
+   "fieldtype": "Column Break"
+  },
+  {
+   "default": "0",
+   "description": "Each time, any outstanding amount left will be considered as write off amount automatically",
```
nextchamp-saqib:
```suggestion
```

nemesis189:
Both the tests passed on my instance:

erpnext/controllers/taxes_and_totals.py
```diff
@@ -643,7 +643,7 @@ def calculate_change_amount(self):
 				self.doc.base_write_off_amount, self.doc.precision("base_change_amount"))
 
 	def calculate_write_off_amount(self):
-		if flt(self.doc.change_amount) > 0:
+		if flt(self.doc.write_off_outstanding_amount_automatically) == 1:
```
nextchamp-saqib, line 646: `test_pos_change_amount` is failing due to this. `write_off_outstanding_amount_automatically` needs to be enabled to calculate `write_off_amount`

PR #27383, opened 2021-09-07
fix: Move related fields together in Selling Settings

GangaManoj:
Backport for https://github.com/frappe/erpnext/pull/27306.


erpnext/selling/doctype/selling_settings/selling_settings.json
```diff
@@ -25,7 +25,7 @@
   "editable_price_list_rate",
   "validate_selling_price",
   "editable_bundle_item_rates",
-  "transaction_settings_section",
+  "sales_transactions_settings_section",
```
nextchamp-saqib, line 28: The section break fieldname is not changed.

nextchamp-saqib: Also, the modified timestamp should also change, else these changes won't reflect in any site instance

nextchamp-saqib (changes requested):

erpnext/selling/doctype/selling_settings/selling_settings.json
GangaManoj: Done.

nextchamp-saqib (approved):

PR #27393, opened 2021-09-08
fix: no validation on item defaults

nextchamp-saqib:
Validate company for item defaults. 

Throw proper message in case of invalid account, warehouse, or cost center selected.

CleanShot 2021-09-08 at 16 40 03@2x


erpnext/stock/doctype/item/item.py
```diff
@@ -782,6 +782,26 @@ def validate_item_defaults(self):
...
+		for item_default in self.item_defaults:
+			for doctype, field in [
+				['Warehouse', 'default_warehouse'],
+				['Cost Center', 'buying_cost_center'],
+				['Cost Center', 'selling_cost_center'],
+				['Account', 'expense_account'],
+				['Account', 'income_account']
+			]:
+				if item_default.get(field):
+					company = frappe.db.get_value(doctype, item_default.get(field), 'company')
```
ankush:
```suggestion
					company = frappe.db.get_value(doctype, item_default.get(field), 'company', cache=True)
```

Won't help much in a single item but when multiple items are created in the same transaction this will reduce queries.

ankush (approved):

PR #27398, opened 2021-09-08
fix: Autoname for customer and supplier

deepeshgarg007:
Introduced a new option "Auto Name" in Selling and Buying Settings

Screenshot 2021-09-15 at 10 18 18 AM

There may be instances when the user wants to name the customer based on custom fields or using some other auto name logic

This PR allows users to use standard auto naming functionality for naming customers

### DIFF
--- a/erpnext/buying/doctype/buying_settings/buying_settings.json
+++ b/erpnext/buying/doctype/buying_settings/buying_settings.json
@@ -28,7 +28,7 @@
    "fieldname": "supp_master_name",
    "fieldtype": "Select",
    "label": "Supplier Naming By",
-   "options": "Supplier Name\nNaming Series"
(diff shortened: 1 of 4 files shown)


chechani:
This type of setting should be for Patient also and I started discussion thread here https://discuss.erpnext.com/t/upgrade-safe-turunaround-to-use-auto-name-for-patient-naming-instead-of-naming-series/74522

chechani:
In Healthcare setting , patient name should be facilitated in the similar manner

ChillarAnand:
Thanks, @chechani.

Would you like to send a PR for the same?

bhavesh95863:
@ChillarAnand 
I sent PR for that.
https://github.com/frappe/erpnext/pull/27532

PR #27412, opened 2021-09-09
fix(ux): clean invalid fields from variant setting

ankush:
- Remove fields that are already existing from options
- Add a message for case when no additional fields can be added.

before:
ALL fields are shown, even the ones that are already added. 
Screenshot 2021-09-09 at 6 40 52 PM

after:
Screenshot 2021-09-09 at 6 40 04 PM

Screenshot 2021-09-09 at 6 38 16 PM

### DIFF
--- a/erpnext/stock/doctype/item_variant_settings/item_variant_settings.js
+++ b/erpnext/stock/doctype/item_variant_settings/item_variant_settings.js
@@ -2,18 +2,31 @@
 // For license information, please see license.txt
 
 frappe.ui.form.on('Item Variant Settings', {
-	setup: function(frm) {
+	refresh: function(frm) {
 		const allow_fields = [];
-		const exclude_fields = ["naming_series", "item_code", "item_name", "show_in_website",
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item_variant_settings/item_variant_settings.js
```diff
@@ -4,17 +4,30 @@
 frappe.ui.form.on('Item Variant Settings', {
 	setup: function(frm) {
```
nextchamp-saqib: Do the options refresh by just being in the `setup` trigger? I have seen instances where options are bugged due to not being refreshed.

ankush: I've also received random reports of options not getting set. Should we move this to `refresh`?

nextchamp-saqib: Yes, moving to `refresh` does fix the issue.

Issue #27335, opened 2021-09-03
Duplicated Contact Name on patient creation using already existing customer


NoxOdonell:
## Description of the issue
When creating a new patient with an already created customers it shows the following error
![image](https://user-images.githubusercontent.com/90055445/132029543-91005d49-8154-46cf-8d32-f172f8f300d0.png)

## Context information (for bug reports)
when you create a customer an then you fill up all fields from patient information, when saving it says "duplicated contact name", i changed the name of contact and saved and it worked, customer creation saves a contact and patient creation tries to save a new contact with the same name
**Output of `bench version`**
```

bench_manager 0.0.1
erpnext 13.10.0
frappe 13.10.0

```

ChillarAnand:
Continued at https://discuss.erpnext.com/t/duplicated-contact-name-on-patient-creation-using-already-existing-customer/79840

PR #27427, opened 2021-09-10
fix(healthcare): Duplicate Contact error on add Patient

akurungadam:
- Duplicate contact error if existing Customer is selected while creating Patient. Reported on [discuss](https://discuss.erpnext.com/t/duplicated-contact-name-on-patient-creation-using-already-existing-customer/79840)
  - Does not create new contact, instead link the Contact linked to selected Customer
- Validate existing User (email and mobile) before creating user on Patient update

Fixes https://github.com/frappe/erpnext/issues/27335


erpnext/healthcare/doctype/patient/patient.py
```diff
@@ -93,23 +93,27 @@ def set_missing_customer_details(self):
...
+			'user_type': 'Website User',
+			'gender': self.sex,
+			'phone': self.phone,
+			'mobile_no': self.mobile,
+			'birth_date': self.dob
+		})
+		user.flags.ignore_permissions = True
+		user.enabled = True
+		user.send_welcome_email = True
+		user.add_roles('Patient')
```
mujeerhashmi, line 115: A patient is a customer as well when you select link patient to customer.

mujeerhashmi:
I had reported this earlier, when you add a user the contact creation task is enqueued [Check Here](https://github.com/frappe/frappe/blob/22f143898bbd690cf17324b4f4eae9abbce9c214/frappe/core/doctype/user/user.py#L102).
So when a patient is created, In `on_update` the sequence should be
1. self.create_website_user()
2. create_customer(self)
3. self.set_contact()

So the contact creation enqueued in `create_website_user()` is finished when the execution reaches `set_contact()`. Else we may end up with two contacts, One linked to website-user and the other to the patient.

akurungadam:
Thanks, User creation is initiated only after Contact is inserted and linked to Patient (and Customer) in order to avoid duplication - as you can see [`create_contact`](https://github.com/frappe/frappe/blob/22f143898bbd690cf17324b4f4eae9abbce9c214/frappe/core/doctype/user/user.py#L969) only proceeds to create Contact if a Contact with same email doesn't already exist.

But I feel if we change the sequence and initiate `create_website_user` first, then this scenario can possibly happen - as we can't quite be sure the enqueued job is finished by the time `set_contact` inserts a new Contact. So keeping the enqueue towards the end will be safe right?

erpnext/healthcare/doctype/patient/patient.py
akurungadam: As it's for the portal access, I feel it's best not to add Customer Role considering that not all privileges for Customer will be relevant to Patient.

mujeerhashmi: OK. But when you have e-commerce enabled for patients like online healthcare packages this is required. Maybe a whitelist will help people like me to override this function?

mujeerhashmi:
> - as you can see [`create_contact`](https://github.com/frappe/frappe/blob/22f143898bbd690cf17324b4f4eae9abbce9c214/frappe/core/doctype/user/user.py#L969) only proceeds to create Contact if a Contact with same email doesn't already exist.

Ok, there is a check here for duplication. Great !!

erpnext/healthcare/doctype/patient/patient.py
akurungadam: IMO, it would be easier to configure the Patient role without having to override. If it's about having additional permissions for the Patient User. Would that be ok?

mujeerhashmi: We set the default role at sign-up as Patient. But the user registered will not be able to place the order online for not being a customer is what I feel. We wanted to override for adding a customer role to a patient registered online.

If needed we will do this in a separate PR.

ChillarAnand (approved):

erpnext/healthcare/doctype/patient/patient.py
akurungadam: yes, that would help as this is merely a fix to the reported issue, thanks

ChillarAnand:
Closes https://github.com/frappe/erpnext/issues/27335

erpnext/healthcare/doctype/patient/patient.py
```diff
@@ -93,23 +93,27 @@ def set_missing_customer_details(self):
...
+		frappe.db.set_value(self.doctype, self.name, 'user_id', user.name)
```
ruchamahabal:
```suggestion
		self.db_set('user_id', user.name)
```

ruchamahabal (approved):

PR #27430, opened 2021-09-10
fix: fail migration due to None type during v13_0.update_returned_qty_in_pr_dn

dj12djdjs:
If there is a Delivery Note or Purchase Receipt is_return =1 and does not contain an original document the patch fails.
I think most case when this fails is from Data Import when migrating from an older system.
[https://discuss.erpnext.com/t/issue-while-migrating-update/79336/2](https://discuss.erpnext.com/t/issue-while-migrating-update/79336/2)

### DIFF
--- a/erpnext/patches/v13_0/update_returned_qty_in_pr_dn.py
+++ b/erpnext/patches/v13_0/update_returned_qty_in_pr_dn.py
@@ -13,7 +13,7 @@ def execute():
 	frappe.reload_doc('stock', 'doctype', 'stock_settings')
 
 	def update_from_return_docs(doctype):
-		for return_doc in frappe.get_all(doctype, filters={'is_return' : 1, 'docstatus' : 1, 'return_against': 'IS NOT NULL'}):
+		for return_doc in frappe.get_all(doctype, filters={'is_return' : 1, 'docstatus' : 1, 'return_against': ('!=', '')}):
 			# Update original receipt/delivery document from return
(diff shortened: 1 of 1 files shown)


erpnext/patches/v13_0/update_returned_qty_in_pr_dn.py
```diff
@@ -13,7 +13,7 @@ def execute():
 	frappe.reload_doc('stock', 'doctype', 'stock_settings')
 
 	def update_from_return_docs(doctype):
-		for return_doc in frappe.get_all(doctype, filters={'is_return' : 1, 'docstatus' : 1}):
+		for return_doc in frappe.get_all(doctype, filters={'is_return' : 1, 'docstatus' : 1, 'return_against': 'IS NOT NULL'}):
```
ankush:
```suggestion
		for return_doc in frappe.get_all(doctype, filters={'is_return' : 1, 'docstatus' : 1, 'return_against': ('!=', '')}):
```

In filters `{key: value}` means `key = value` in SQL. 

You can confirm by adding `debug=True` to inspect generated query.

ankush (changes requested):

erpnext/patches/v13_0/update_returned_qty_in_pr_dn.py
dj12djdjs: Good to know! Thanks!

ankush (approved):

PR #27433, opened 2021-09-11
fix: Template Error due to use of single quote

sagarvora:
image


sagarvora:
@mergifyIO backport version-13-hotfix

PR #27441, opened 2021-09-11
fix: PO/PINV -  Check if doctype has company_address field before setting the value

### LINKED ISSUE #27419: "Field company_address not found." Error when creating a new Purchase Order or Purchase Invoice
## Description of the issue
Upon creating a new Purchase Order, or a Purchase Invoice, an error pops up indicating that the field "company_address" is not found. This happened after upgrading to the latest release.

 I do suspect this new problem as come from changes made to fix issue #27217: https://github.com/frappe/erpnext/pull/27217

1. Create a new Purchase Order, or Create a new Purchase Invoice

### Observed result
Error modal with the content: "Field company_address not found."

### Expected result
No error, and the page loads as usual.

ERPNext: v13.10.2 (version-13)
Frappe Framework: v13.2.1 (version-13)

vamagithub:
closes #27419

Upon creating a new Purchase Order, or a Purchase Invoice, an error pops up indicating that the field "company_address" is not found

This is happening because doctypes like Purchase Order/ PINV doesn't have field "company_address".
After adding a "if" condition before setting the value will solve the issue.

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -864,7 +864,8 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 						if (r.message) {
 							me.frm.set_value("billing_address", r.message);
 						} else {
-							me.frm.set_value("company_address", "");
+							if(frappe.meta.get_docfield(me.frm.doctype, 'company_address'))				
+								me.frm.set_value("company_address", "");						
 						}
 					}
 				});
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -864,7 +864,7 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 						if (r.message) {
 							me.frm.set_value("billing_address", r.message);
 						} else {
-							if (frappe.meta.get_docfield(me.frm.doctype, 'company_address')){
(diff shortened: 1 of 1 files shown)


erpnext/public/js/controllers/transaction.js
```diff
@@ -864,7 +864,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 						if (r.message) {
 							me.frm.set_value("billing_address", r.message);
 						} else {
-							me.frm.set_value("company_address", "");
+							if(frappe.meta.get_docfield(me.frm.doctype, 'company_address')){			
+								me.frm.set_value("company_address", "");
+							}
```
ankush:
```suggestion
							if (frappe.meta.get_docfield(me.frm.doctype, 'company_address')){
								me.frm.set_value("company_address", "");
							}
```

formatting changes

erpnext/public/js/controllers/transaction.js
```diff
@@ -864,7 +864,9 @@ erpnext.TransactionController = erpnext.taxes_and_totals.extend({
 						if (r.message) {
 							me.frm.set_value("billing_address", r.message);
 						} else {
-							me.frm.set_value("company_address", "");
+							if (frappe.meta.get_docfield(me.frm.doctype, 'company_address')){
```
ankush:
```suggestion
							if (frappe.meta.get_docfield(me.frm.doctype, 'company_address')) {
```

ankush (approved):

PR #27457, opened 2021-09-13
Removed read only from title to allow title modification

asamawi:

> Screenshots/GIFs

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.json
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.json
@@ -552,8 +552,7 @@
    "fieldtype": "Data",
    "hidden": 1,
    "label": "Title",
-   "print_hide": 1,
-   "read_only": 1
+   "print_hide": 1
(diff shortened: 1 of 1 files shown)


ankush:
We don't accept changes to stable branches. : https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#which-branch 

---

Also, this change won't work. Renaming the title is deliberately not allowed.. .

https://github.com/frappe/erpnext/blob/bab644a249de4355d6700f53a7bfbf0114ebb30c/erpnext/accounts/doctype/payment_entry/payment_entry.py#L632-L640

asamawi:
Client has naming series for customers and suppliers and wants to rename the payment entry to have name of the party.

ankush:
Seems to be a valid use case, please open a new issue. Or if you have a solution that can take care of both then send a PR :) 

FYI you can also try making a server script and set `doc.title` in `before_submit` hook. That won't get overwritten.

PR #27467, opened 2021-09-13
feat: Grant commission on certain items only

### LINKED ISSUE #27276: Commission on certain Items only
### User Story

For our main services we pay a commission to our sales partners. But on the same invoice there will be other items, like travel expenses, which are not applicable for commission. Currently, in ERPNext the commission is calculated based on the net total. We need to calculate our own base total for commission ("Commission Base"). This should be the sum of net amounts of all items which are applicable for commission. The Commission Base multiplied by Commission Rate gives the Commission Total.

### Possible solutions

a) Add a field "Grant Commission" in Item and Transaction Item tables. Only line items where "Grant Commission" is active will be included in the calculation.

b) ~Add a link to Item Group in Sales Partner and Transaction Item tables. Only line items where "Item Group" is identical or a child of Item Group in Sales Partner will be included in the calculation.~

barredterra:
Close #27276

- "Amount Eligible for Commission" is calculated as the sum of net amounts of line items where "Grant Commission" is active.
- Total Commission = Amount Eligible for Commission * Commission Rate

https://user-images.githubusercontent.com/14891507/133156499-f885e847-99ea-4ae1-a75d-87b29a8dc6b2.mov

Docs:

- Item: https://docs.erpnext.com/docs/v13/user/manual/en/stock/item/edit?wiki_page_patch=cf7d5e4852
- Sales Invoice Item: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/sales-invoice/edit?wiki_page_patch=801ef0a039

Thanks to @aisenyi and @sagarvora for their help with this feature.


mujeerhashmi:
Oh, I totally missed out on this feature. It's great. One suggestion, the commission rate % may not be the same for all the items in a sales invoice. So defining the commission rate % in item master would be more feature-rich.  Then calculate the commission per line item & totaling them would be more extensive.

barredterra:
@mujeerhashmi currently the commission is defined per **Sales Partner**. So to make your suggestion work we would need a mapping between **Item** and **Sales Partner**. This makes things much more complicated, so I would suggest waiting until someone really has this requirement.

erpnext/stock/doctype/item/item.json
```diff
@@ -24,6 +24,7 @@
   "allow_alternative_item",
   "is_stock_item",
   "include_item_in_manufacturing",
+  "grant_commission",
```
marination: Can this be moved to the `Sales Details` section ? If required users can Customize the position later.

barredterra: Sure :)

barredterra: Done
Bildschirmfoto 2021-09-23 um 12 41 32

barredterra:
@marination any other changes required?

sagarvora:
Additional changes made:
- Make **Grant Commission** in Transaction Item Read Only
- Rename **Commission Base** to **Amount Eligible for Commission**
- Calculate **Amount Eligible for Commission**  in backend as well
- Ignore **Quotation**, but include **POS Invoice** when calculating commission and contribution in the backend
- Add Test Case
- Semantic / Code Quality fixes

@barredterra Please update documentation if needed.

barredterra:
@marination I'd appreciate your review on this PR :)

barredterra:

barredterra:
@marination any further changes required here? Thanks in advance! :)

hrwX:
@barredterra This looks good, works as expected 👌

sagarvora:
@marination @deepeshgarg007 @nextchamp-saqib 
It's been almost 2 months since this PR was last changed. Is it safe to assume that no further changes are required?

nextchamp-saqib (dismissed):

nextchamp-saqib:
Looks good to me!

erpnext/stock/doctype/item/item.json
```diff
@@ -1020,6 +1021,12 @@
    "fieldname": "website_image_alt",
    "fieldtype": "Data",
    "label": "Image Description"
+  },
+  {
+   "default": "1",
```
nextchamp-saqib, line 1026: This shouldn't be enabled by default right? @sagarvora @barredterra

sagarvora: Previously, this was applicable for all items. This keeps it backwards compatible. Alternative is patch with default as 0. But I think that there are generally lesser number of items where commission isn't granted.

sagarvora: If you think it is getting undue attention because the checkbox defaults to 1, we can change the field to be **Don't Grant Commission** instead.

bhavesh95863:
Hey @barredterra ,
Why do we not have a patch to update the old document item table? I have updated one of my server and this feature blocks commission calculation.
I show there is a grant_commission checkbox in the item table but that value is 0 and because of that amount_eligible_for_commission is 0.

PR #27471, opened 2021-09-14
feat: Merge POS invoices based on customer group

deepeshgarg007:
Customers can now be merged based on customer group. 

1. Select the "Merge Invoice Based On" customer group. If this option is selected the customer invoice validation will be ignored
2. Select the Group Customer based on which the invoices have to be merged.  

Screenshot 2021-09-16 at 7 27 57 PM

`no-docs`


erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.js
```diff
@@ -4,13 +4,18 @@
...
 				filters: {
 					'docstatus': 1,
 					'customer': doc.customer,
 					'consolidated_invoice': ''
 				}
 			}
 		});
+	},
+
+	merger_invoices_based_on: function(frm) {
```
nextchamp-saqib:
```suggestion
	merge_invoices_based_on: function(frm) {
```

PR #27475, opened 2021-09-14
fix: Handle Excess/Multiple Item Transfer against Job Card

marination:
[Documentation](https://docs.erpnext.com/docs/v13/user/manual/en/manufacturing/manufacturing-settings)

This PR fixes multiple issues that occur via Work order (with operations) where Material Transfer Against is set to **Job Card**

> PS: The fixes are really small. I have over explained so that the flow is understood and also for reference

**Case:**
- BOM with operations
- Work Order has Material Transfer Against is set to **Job Card**
- Submit Work order. Job cards are auto created

Following are the issues in the individual Job Cards (while transferring/completing):
----------------

### Excess Transfer against Job Card
- In Manufacturing settings , added a new **Job card** section
- Added a new checkbox **Allow Excess Material Transfer**. This will show the Material transfer button even after required qty is satisfied so that users can transfer extra RM

### Material Transfer / Material Request Buttons
- Condition for Material Transfer / Material Request Buttons to be visible was that the `FG Qty from Transferred Raw Materials` field should not be equal to the `Qty to Manufacture` in the JC
- This means that if under transferred or over transferred these buttons are visible.
- If just the right amount is transferred these buttons are not, which is absurd UX
- Also they were visible after cancellation of JC

After the **fix**:
- **Material Request** button is only visible if  `FG Qty from Transferred Raw Materials` is lesser than the `Qty to Manufacture` field (under transfer) OR if excess transfer is allowed in Manufacturing Settings
- **Material Transfer** button is only visible if any row in the items table has more material left to transfer OR if excess transfer is allowed in Manufacturing Settings

### Negative `For Quantity` in Stock Entry from JC
- Since **Material Transfer** Button was visible in over transfer case, on excess transfer `For Quantity` would become negative since required qty is already fulfilled
- This **fix** makes sure it is set to 0 in such cases

### Status in JC
- Assume 2 RMs in JC to manufacture 1 qty of FG
- They are transferred in two stock entries. Both stock entries have `For Quantity` set to 1 (Fg qty). This causes `FG Qty from Transferred Raw Materials` to become 2 in the Job card
- The status in Job card was set to **Work in Progress**, in over transfer cases
- This **fix** sets it to **Completed** if over transferred or just the right amount is transferred

### Miscellanous
- Renamed ‘Transferred Qty’ to ‘FG Qty from Transferred Raw Materials’ in JC (fieldname is the same)
- Renamed ’Allow Multiple Material Consumption’ to ‘Allow Continuous Material Consumption’ (fiedname is same)
- Secured denominator variables in `get_transfered_raw_materials` to avoid ZeroDivisionError


marination:
Manually backported

PR #27485, opened 2021-09-14
Enhance Point Of Sales Code

zaqoutabed:


nextchamp-saqib:
Hi, @zaqoutabed ,

The `get_item_details` doesn't consider items that are already sold via a POS Invoice. Hence we need to make a separate call for each item and on every click of the item.

zaqoutabed:
> Hi, @zaqoutabed ,
> 
> The `get_item_details` doesn't consider items that are already sold via a POS Invoice. Hence we need to make a separate call for each item and on every click of the item.

`check_stock_availability` it used to check if the item has stock qty
and `trigger_new_item_events` it already return the `availability_qty` 

I use the return value of `availability_qty` to check item stock availability

zaqoutabed:
image

zaqoutabed:
also in the method `trigger_new_item_events`
I think no need to fire qty trigger  because the default value of `qty = 1` for new item
and the system use that qty to make totals calculations in the POS Form
and when trigger is fired again it will consume time that we need !

image

nextchamp-saqib:
> > Hi, @zaqoutabed ,
> > The `get_item_details` doesn't consider items that are already sold via a POS Invoice. Hence we need to make a separate call for each item and on every click of the item.
> 
> `check_stock_availability` it used to check if the item has stock qty
> and `trigger_new_item_events` it already return the `availability_qty`
> 
> I use the return value of `availability_qty` to check item stock availability

`trigger_new_item_events` calls `get_item_details` to fetch `available_qty`
And `get_item_details` don't consider the stock which has already been sold with a POS Invoice.

To better understand the problem,
- Create a new item and make a stock entry for it with 10 qty
- Create an invoice from POS, add the item and set the qty as 10, Then checkout and submit the invoice
- Now the stock quantity for that item should be zero.
- Now, create another invoice and try to add that item
- Since the item is completely sold, `available_qty` should be zero and the system should throw an error saying, stock quantity not available.

But if you remove `check_stock_availability`, the `available_qty` remains 10, and hence the error is not thrown.

nextchamp-saqib:
> I think no need to fire qty trigger because the default value of `qty = 1` for new item

Yes, the `qty` trigger can be omitted for a new item.

PR #27488, opened 2021-09-14
fix: Args missing error on changing Price List currency with cart enabled

marination:
After E-commerce Refactor due to merge conflicts, keyword args introduced via https://github.com/frappe/erpnext/pull/27050 became positional args.
This caused the error below on calling `validate_cart_settings()` within function `check_impact_on_shopping_cart` in `price_list.py` :

image


marination:
Merged again for Linter to pass, the rest is all passing

PR #27491, opened 2021-09-14
refactor: finance books

GangaManoj:
## Company

- Add a new checkbox 'Enable Finance Books'.
- Make 'Default Finance Book' mandatory if 'Enable Finance Books' is checked.
- Move 'Default Finance Book' to the 'Fixed Asset Defaults' section.

Screenshot 2021-09-14 at 7 10 37 PM

## Asset

- If 'Enable Finance Books' is checked:
   - The 'Finance Book' column in the Finance Books table will be visible.
    
     Screenshot 2021-09-14 at 7 19 24 PM

   - Add Row and Delete Row will be enabled for the Finance Books table.

     Screenshot 2021-09-14 at 7 29 41 PM

     Screenshot 2021-09-14 at 7 30 30 PM

   - The 'Finance Book' column in the Depreciation Schedule table will be visible.

     Screenshot 2021-09-14 at 7 31 40 PM

- If 'Enable Finance Books' is unchecked:
   - The 'Finance Book' column in the Finance Books table will not be visible.
   - Add Row and Delete Row will not be enabled for the Finance Books table.
   - The 'Finance Book' column in the Depreciation Schedule table will not be visible.
   
   Screenshot 2021-09-14 at 7 38 12 PM

## Journal Entry

- The 'Finance Book' field will only be visible if 'Enable Finance Books' is checked.

_When checked:_

Screenshot 2021-09-14 at 7 43 04 PM

_When unchecked:_

Screenshot 2021-09-14 at 7 44 21 PM

<details>
<summary>Updated Docs</summary>

- Finance Book: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/finance-book
- Company > Fixed Asset Defaults: https://docs.erpnext.com/docs/v13/user/manual/en/setting-up/company-setup#24-fixed-asset-defaults
- Journal Entry: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/journal-entry#1-how-to-create-a-journal-entry
- Asset: https://docs.erpnext.com/docs/v13/user/manual/en/asset/asset#42-depreciation-schedule

</details>


erpnext/setup/doctype/company/company.json
```diff
@@ -139,6 +140,7 @@
    "fieldname": "default_finance_book",
    "fieldtype": "Link",
    "label": "Default Finance Book",
+   "mandatory_depends_on": "enable_finance_books",
```
nextchamp-saqib, line 144: `default_finance_book` should be hidden if `enable_finance_books` is unchecked

erpnext/setup/doctype/company/company.py
```diff
@@ -211,6 +212,15 @@ def validate_parent_company(self):
 			if not is_group:
 				frappe.throw(_("Parent Company must be a group company"))
 
+	def check_if_finance_books_are_enabled(self):
+		from frappe.custom.doctype.property_setter.property_setter import make_property_setter
+
+		enable_finance_books = cint(self.enable_finance_books)
+
+		make_property_setter("Asset Finance Book", "finance_book", "hidden", not(enable_finance_books), "Check", validate_fields_for_doctype=False)
```
nextchamp-saqib, line 220: This also needs to be executed with a patch for existing users who are not using finance books

We can maybe check if the default finance book is set OR if a user has more than one finance book, then enable finance books.
If not, make these property setters to hide the finance book field.

nextchamp-saqib (changes requested):

erpnext/assets/doctype/asset/asset.js
```diff
@@ -71,6 +71,13 @@ frappe.ui.form.on('Asset', {
 				filters: { item_code: doc.item_code }
 			}
 		});
+
+		frappe.db.get_value("Company", frm.doc.company, "enable_finance_books", (r) => {
```
nextchamp-saqib, line 75: In Asset Category too.

erpnext/setup/doctype/company/company.json
GangaManoj: Done.

erpnext/assets/doctype/asset/asset.js
GangaManoj: But there's a problem there- Asset Categories aren't linked with a single Company. What if one of the companies listed has Finance Books enabled, but the rest don't?

erpnext/setup/doctype/company/company.py
GangaManoj: I just wrote the patch, but then I realised that this is all wrong- I shouldn't be using property setters here, because that would affect the visibility of that field for docs linked with ALL companies. 

I should make js changes instead- make those fields hidden by default, but after entering a Company, check its Enable Finance Books value and decide its visibility accordingly.

erpnext/assets/doctype/asset/asset.js
nextchamp-saqib: Make it visible if any one of the companies has it enabled. 

But also add a validation that, if for a company, FB is enabled, then it is mandatory to select a FB just for that company's row

GangaManoj: But the list of companies and finance books are in two separate tables, so if any of the Companies listed has Finance Books enabled, the Finance Books entered will be linked with all of them.

PR #27496, opened 2021-09-14
test: add custom TestCase class and use in stock module

ankush:
Problem: Tests leave too much state in db which makes next tests flaky and dependent on state. 

Solution: 
- Create a custom test case class and stop committing in inheriting classes. 
- Expand the reach of this class slowly till all major tests are covered. Starting with stock related tests for now because it's a pressing requirement to merge 2 other PRs.


ankush:
will do it the "proper way" and reopen later. 😄

PR #27499, opened 2021-09-14
fix: Tags getting fetched correctly in Get Supplier in RFQ (Request For Quotation)

### LINKED ISSUE #26343: Tags not fetching
When I select tools in RFQ and get supplier from tag, data couldn’t fetch, system is show only that tag which I add first time in tag list.

Ex: let’s say when I add tag abc & xyz

And assign that tag to supplier, after that I add 1 more tag aaa, in RFQ I select tools and get supplier from tag, so the system is show on first two tag, not showing another tag.

vamagithub:
closes  #26343

### Output - Solution

https://user-images.githubusercontent.com/16986940/133311742-c1e5201d-d657-48ad-baee-8025086b68b1.mp4

### DIFF
--- a/erpnext/buying/doctype/request_for_quotation/request_for_quotation.py
+++ b/erpnext/buying/doctype/request_for_quotation/request_for_quotation.py
@@ -396,12 +396,10 @@ def get_item_from_material_requests_based_on_supplier(source_name, target_doc = 
 def get_supplier_tag():
 	filters = {"document_type": "Supplier"}
 	tags = list(set(tag.tag for tag in frappe.get_all("Tag Link", filters=filters, fields=["tag"]) if tag))
-    
+
 	if tags and tags != frappe.cache().hget("Supplier", "Tags"):
 		frappe.cache().hset("Supplier", "Tags", tags)
-
 	return frappe.cache().hget("Supplier", "Tags")
-
 
 @frappe.whitelist()
 @frappe.validate_and_sanitize_search_inputs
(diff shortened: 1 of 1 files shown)


erpnext/buying/doctype/request_for_quotation/request_for_quotation.py
```diff
@@ -394,11 +394,11 @@ def get_item_from_material_requests_based_on_supplier(source_name, target_doc =
...
-	if not frappe.cache().hget("Supplier", "Tags"):
-		filters = {"document_type": "Supplier"}
-		tags = list(set(tag.tag for tag in frappe.get_all("Tag Link", filters=filters, fields=["tag"]) if tag))
-		frappe.cache().hset("Supplier", "Tags", tags)
+	filters = {"document_type": "Supplier"}
+	tags = list(set(tag.tag for tag in frappe.get_all("Tag Link", filters=filters, fields=["tag"]) if tag))
 
+	if tags and tags != frappe.cache().hget("Supplier", "Tags"):
+		frappe.cache().hset("Supplier", "Tags", tags)
 	return frappe.cache().hget("Supplier", "Tags")
```
ankush:
```suggestion
	return tags
```

This cache is kinda useless now. Why not just return the output directly?
The root cause from what I can tell is lack of cache invalidation when new tags are added on Supplier documents. 

This is a simple select query and is not used that often.  IMO it's better to get rid of caching completely here.

ankush (changes requested):

ankush (approved):

PR #27508, opened 2021-09-15
fix: Shopping Cart and Variant Selection

marination:
- Cart Icon badge (navbar) now appears on adding first item
- Payment summary section now updates on increase/decrease in cart qty
- Incorrect No of items in payment summary is fixed
- Empty Cart better empty cart message
- Screen freeze on placing order until redirection (no need to hide sections)
- Consider unpublished variants while building item variant cache (used to throw error because certain variants did not have website items against them)


marination:
Linter failing due to cache conflict (when it overlaps over two commits)
UI test failing (unrelated)

TurkerTunali:
A screenshot would be great here.

PR #27513, opened 2021-09-15
fix: table data deleted on submitted maintenance schedule

noahjacob:
### Issue
- Schedule data gets erased if "Generate Schedule" is triggered on a submitted schedule and refreshed.

schedule_bug

### Fix
- Disabled "Generate Schedule " button when document is in submitted state.
- Removed unnecessary query to delete rows.

### DIFF
--- a/erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
+++ b/erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
@@ -16,9 +16,9 @@ from erpnext.utilities.transaction_base import TransactionBase, delete_events
 class MaintenanceSchedule(TransactionBase):
 	@frappe.whitelist()
 	def generate_schedule(self):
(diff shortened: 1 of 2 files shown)


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -16,26 +16,25 @@
...
-				child.item_name = d.item_name
-				child.scheduled_date = s_list[i].strftime('%Y-%m-%d')
-				if d.serial_no:
-					child.serial_no = d.serial_no
-				child.idx = count
-				count = count + 1
-				child.sales_person = d.sales_person
-				child.completion_status = "Pending"
-				child.item_reference = d.name
+		if not self.docstatus:
```
```suggestion
		if self.docstatus != 0:
			return	
```

Always use [guard conditions](https://en.wikipedia.org/wiki/Guard_(computer_science)) to avoid over indenting.

ankush (approved):

PR #27524, opened 2021-09-15
fix: Tax breakup based on items, missing GST fields

nemesis189:
**Issues:**
- Tax breakup table would combine the items with same HSN/SAC code into one row.
   

- GST fields missing in POS Invoice
- Warehouse validation missing in Sales invoice missing from POS invoice

**Solution:**
- Tax breakup based on `item_code` instead of `gst_hsn_code`
   

- Added GST fields to custom fields in setup, added a patch for the same.
- Added the warehouse validation


deepeshgarg007 (dismissed):

deepeshgarg007 (approved):

deepeshgarg007:

PR #27538, opened 2021-09-16
fix: Validate if item exists on uploading items in stock reco

marination:
**Issue:**
Uploading (via Upload button) **non existent item** in stock reco and then changing warehouse or batch  (trigger JS action) gave an error:
```py
  File ".../stock_reconciliation/stock_reconciliation.py", line 590, in get_stock_balance_for
    with_serial_no = True if item_dict.get("has_serial_no") else False

  AttributeError: 'NoneType' object has no attribute 'get'
```

>PS: Maybe some basic checks can be added during data upload itself to not manually check these. Will check framework

### DIFF
--- a/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
+++ b/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
@@ -621,8 +621,8 @@ def get_stock_balance_for(item_code, warehouse,
 
 	if not item_dict:
 		# In cases of data upload to Items table
-		msg = f"Item {item_code} does not exist."
-		frappe.throw(msg=_(msg), title=_("Missing"))
+		msg = _("Item {} does not exist.").format(item_code)
+		frappe.throw(msg, title=_("Missing"))
 
 	serial_nos = ""
 	with_serial_no = True if item_dict.get("has_serial_no") else False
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
```diff
@@ -619,6 +619,11 @@ def get_stock_balance_for(item_code, warehouse,
 	item_dict = frappe.db.get_value("Item", item_code,
 		["has_serial_no", "has_batch_no"], as_dict=1)
 
+	if not item_dict:
+		# In cases of data upload to Items table
+		msg = f"Item {item_code} does not exist."
```
szufisher: translate _() does not support  f'' formatted string.

ankush:
```suggestion
		msg = _("Item {} does not exist.").format(item_code)
```

erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.py
```diff
@@ -619,6 +619,11 @@ def get_stock_balance_for(item_code, warehouse,
...
+		frappe.throw(msg=_(msg), title=_("Missing"))
```
ankush:
```suggestion
		frappe.throw(msg, title=_("Missing"))
```

ankush (approved):
LGTM.

We can think of adding link field validations during/post CSV imports?

marination:
@ankush yes that will be good and also maybe a hook that can enable us to write custom checks too ? Because unlike data import this import doesn't give that error traceback bit (if i'm not wrong),and causes a lot of type issues too (historically) like comparison between `str` and `int`.

ankush:

PR #27545, opened 2021-09-16
fix: cannot add deductions in internal transfer payment entry

nextchamp-saqib:
Due to recent changes from 7141860c04fd1541f8d346a48e3e058f4ad443c9, the `received_amount` was being set as `paid_amount` even if there were deductions such as transfer fees was added to the payment entry. 

To replicate the issue:
- Create a Payment Entry of type Internal Transfer
- Set From & To account
- Set Paid Amount as 1000 and Received Amount as 900
- Add a new row to the Deductions table and set Account, Cost Center & Amount as 100
- Now save the payment entry
- You should see that Received Amount has been set as Paid Amount and Difference Amount is -100

Expected Behaviour:
- The Received Amount should have been 900 since there has been a transfer fee deduction added manually.

---

- [ ] Test Case


deepeshgarg007:

deepeshgarg007 (approved):

PR #27551, opened 2021-09-16
fix: Deferred revenue entries post account freezing

deepeshgarg007:
image

Administrator user has all the roles so if accounting entries are frozen up to date and any Role allowed to post entries is mentioned then admin is able to bypass this entry.

All the automated entries posted by the system as made by the user Administrator, so prevent Admin user from posting any entries if books are frozen 

`no-docs`


deepeshgarg007:
@Mergifyio backport develop

PR #25358, opened 2021-04-15
refactor: maintenance schedule and visit enhancements.

noahjacob:
**Improvements**
- Schedule is generated on save.
- End date is now automatically set based on periodicity and no. of visits.
- Improved form.

maintena

- Dialog Box for creating a new visit.
- Status and Dates and updated after submitting scheduled visits.
- Serial_no filters only serial nos entered.

dialog

- Created links between Maintenance Visit and Maintenance Schedule child tables.

PR #27554, opened 2021-09-16
fix: Maintenance Schedule child table status for legacy data

marination:
**Issue:**
- Pre-existing `schedules` table in **Maintenance Schedule** has blank `completion_status` after [this PR](https://github.com/frappe/erpnext/pull/25358) introduced linking between **Maintenance Visit** and **Maintenance Schedule Detail Row**
- Due to this the **Create** button is not visible (visible if any one of the rows has status **Pending**)

**Fix:**
- Patch to set Completion status to `Pending` in all schedules table rows. Currently its hard to identify which rows are fully completed because of the absence of linking in legacy data.
- Made completion status field editable on submit (in case users want to rectify) and set a default value of `Pending` in it


erpnext/patches/v13_0/set_status_in_maintenance_schedule_table.py
```diff
@@ -0,0 +1,8 @@
+import frappe
+
+def execute():
+	frappe.db.sql("""
+		UPDATE `tabMaintenance Schedule Detail`
+		SET completion_status = 'Pending'
+		WHERE docstatus = 1
```
marination: Include legacy draft schedules too (that were created just before this refactor was available)
```suggestion
		WHERE docstatus < 2
```

PR #27560, opened 2021-09-17
fix(ux): added exception of template item in filters

noahjacob:
### Changes

- Including template items in item filters while creating a BOM.


erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -31,7 +31,8 @@ frappe.ui.form.on("BOM", {
 			return {
 				query: "erpnext.manufacturing.doctype.bom.bom.item_query",
 				filters: {
-					"is_stock_item": 1
+					"is_stock_item": 1,
+					"is_template_item": 1
```
ankush:
```suggestion
```

This seems confusing. Should we just implicitly `OR` this intead in query code? Because template items are only an exception to it.

PR #27561, opened 2021-09-17
fix: add (uom, brand) Item details in an Item Price

18alantom:
### Description
The UOM and Brand don't get added to an Item Price when a new Item is saved. This PR adds those details to a newly created Item Price on new Item save.


erpnext/stock/doctype/item/item.py
```diff
@@ -635,8 +637,8 @@ def validate_name_with_item_group(self):
 
 	def update_item_price(self):
 		frappe.db.sql("""update `tabItem Price` set item_name=%s,
-			item_description=%s, brand=%s where item_code=%s""",
-					(self.item_name, self.description, self.brand, self.name))
+			item_description=%s, brand=%s, uom=%s where item_code=%s""",
+					(self.item_name, self.description, self.brand, self.name, self.stock_uom))
```
marination: Positional values must be in the right order
```suggestion
					(self.item_name, self.description, self.brand, self.stock_uom, self.name))
```

ankush: Just use named arguments instead? keeping track of 5 arguments in order is bound to be error prone. @18alantom

erpnext/stock/doctype/item/item.py
```diff
@@ -634,9 +636,23 @@ def validate_name_with_item_group(self):
...
-		frappe.db.sql("""update `tabItem Price` set item_name=%s,
-			item_description=%s, brand=%s where item_code=%s""",
-					(self.item_name, self.description, self.brand, self.name))
+		frappe.db.sql("""
+				UPDATE `tabItem Price`
+				SET
+					item_name=%(item_name)s,
+					item_description=%(item_description)s,
+					brand=%(brand)s,
+					uom=%(uom)s
```
ankush:
```suggestion
```

Item prices can have different UOMs, also they don't change after the first transaction. No need to update them.

erpnext/stock/doctype/item/item.py
```diff
@@ -634,9 +636,23 @@ def validate_name_with_item_group(self):
...
+				WHERE item_code=%(item_code)s
+			""",
+			dict(
+				item_name=self.item_name,
+				item_description=self.description,
+				brand=self.brand,
+				uom=self.stock_uom,
```
ankush:
```suggestion
```

ankush (changes requested):

ankush (dismissed):

erpnext/stock/doctype/item/item.py
```diff
@@ -634,9 +636,21 @@ def validate_name_with_item_group(self):
...
 	def update_item_price(self):
-		frappe.db.sql("""update `tabItem Price` set item_name=%s,
-			item_description=%s, brand=%s where item_code=%s""",
-					(self.item_name, self.description, self.brand, self.name))
+		frappe.db.sql("""
+				UPDATE `tabItem Price`
+				SET
+					item_name=%(item_name)s,
+					item_description=%(item_description)s,
+					brand=%(brand)s,
```
ankush:
```suggestion
					brand=%(brand)s
```

minor sql error.

ankush:
unrelated failures. Probably related to https://github.com/frappe/frappe/pull/14102

ankush (approved):

Issue #27480, opened 2021-09-14
Production Plan Status Showing "Completed" but Work orders are Still not yet completed


casakthiveln:
## Description of the issue
Production Plan Status Showing "Completed" but Work orders are not yet Started and completed. When a single work order is completed, the production plan updating its status as "Completed".
## Context information (for bug reports)

**Output of `bench version`**
```
ERPNext: v13.11.0 (version-13)

Frappe Framework: v13.11.0 (version-13)


```

## Steps to reproduce the issue

1.
2.
3.

### Observed result
![Plan](https://user-images.githubusercontent.com/63579207/133237167-62d1b746-9e21-4abf-a7da-ecba9c302ee1.gif)

### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

casakthiveln:
Please see this thread. After this change the above issue raised.
https://github.com/frappe/erpnext/issues/26822#issue-961643072

ankush:
@casakthiveln 

Can you please let us know whether some work orders over-produced and hence total produced qty was more than total required qty?

If that is the case then we can work out a better solution.

casakthiveln:
@ankush 
First i thank you for allocating your time for this issue.
My Company is dealing with Dip Tea Bag Production like Lipton Tea,
One box of Finished goods will require 100 Tea bags and we are preparing a production plan.
This FG will have two sub-assembly
1. Convertion of Tea into Tea bags (in Machine)
2. Packing of 100 Tea bags per box (Manual Labour)

But while running the machine we will have process wastage of 2%. So, my machine counter will show 102 tea bags but our actual good production is 100.

So, after submitting the job card for 102 tea bags, in work order we are transferring material for 102 tea bags and finishing the work order with actual planned production which will add the cost of process loss to good production. please see below

![Untitled](https://user-images.githubusercontent.com/63579207/133764194-709532d6-47f7-4fb5-a6b9-45a5e34481ab.png)

Now what happening, After competition of one work order (Sub-Assembly) with the said over production, the production plan status change to "Completed" but one work order is still yet to complete to finish the plan. it is not considering whether all work orders under the plan are completed or not.

Before the fix for #26822 , The production plan status is showing "in Process" even all work orders are completed.

Pls help in this regard. Thanks in advance.

casakthiveln:
@ankush 
[...]

PR #27567, opened 2021-09-17
fix: update production plan status

18alantom:
Patch to update Production Plan status which may incorrectly not be set to _Completed_.

Related to PR #27027


ankush:
Maybe we should refactor this a bit and improve the condition:

`if all WO are finished => PP is finished` 😄 

ref: https://github.com/frappe/erpnext/issues/27480

erpnext/patches/v13_0/update_production_plan_status.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (c) 2021, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+import frappe
+
+
+def execute():
+	frappe.db.sql("""
+		UPDATE `tabProduction Plan`
```
ankush: `reload_doc` this. It didn't exist in v10.

casakthiveln:
@rohitwaghchaure sir, please review this and add in the update...I'm facing huge problem with production plan status update

ankush:
@18alantom this patch just fixes existing records, future Plan-WO combinations should also follow this right?

18alantom:
> @18alantom this patch just fixes existing records, future Plan-WO combinations should also follow this right?

@ankush I didn't exactly understand what you meant.

Are you saying [this](https://github.com/frappe/erpnext/pull/27027) check should be rewritten to check WO completions instead?

ankush:
> Are you saying this check should be rewritten to check WO completions instead?



Yes

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -575,6 +575,15 @@ def set_sub_assembly_items_based_on_level(self, row, bom_data, manufacturing_typ
...
 			self.append("sub_assembly_items", data)
 
+	def check_have_work_orders_completed(self):
+		wo_status = frappe.db.get_list(
+			"Work Order",
+			filters={"production_plan": self.name},
+			fields="status",
+			pluck="status"
+		)
+		return all(map(lambda x: x == "Completed", wo_status))
```
```suggestion
		return all(x == "Completed" for x in wo_status))
```

map/lambda unnecessary.

casakthiveln:
@marination pls look into this

ankush (approved):

ankush:
Ignoring raw query for patches. When/if postgres support comes, patch queries will be irrelevant anyway.

casakthiveln:
Thank you everyone

PR #27570, opened 2021-09-17
fix: View Stock / Accounting Ledger button not showing in the Stock Entry

rohitwaghchaure:
**Issue**

Screenshot 2021-09-17 at 6 05 58 PM

**After Fix**
Screenshot 2021-09-17 at 6 04 13 PM


ankush:
note: only affects `develop`

PR #27580, opened 2021-09-18
Remove customer mandate in pl

KrithiRamani:
1. Removed the mandate, to be able to select several SO from various customers in PL.
2. Not setting any value to picked_qty while pulling Sales Order Details. Value will be set to stock_qty, on submit only.


KrithiRamani:
mentioned wrong base. closing PR

PR #27584, opened 2021-09-19
fix: Improvements in COA Importer

deepeshgarg007:
- Removed unnecessary validations
- Validation for the no of columns
- Fixes for parsing group accounts

### DIFF
--- a/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
+++ b/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
@@ -25,8 +25,16 @@ from erpnext.accounts.doctype.account.chart_of_accounts.chart_of_accounts import
 
 
 class ChartofAccountsImporter(Document):
-	def validate(self):
-		validate_accounts(self.import_file)
+	pass
+
+def validate_columns(data):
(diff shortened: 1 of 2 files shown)


erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py
```diff
@@ -364,20 +371,12 @@ def get_mandatory_account_types():
 
 def validate_account_types(accounts):
 	account_types_for_ledger = ["Cost of Goods Sold", "Depreciation", "Fixed Asset", "Payable", "Receivable", "Stock Adjustment"]
-	account_types = [accounts[d]["account_type"] for d in accounts if not accounts[d]['is_group'] == 1]
+	account_types = [accounts[d]["account_type"] for d in accounts if not cint(accounts[d]['is_group']) == 1]
```
8633brown, line 374: I dont think this cast is needed. As long as the 'parent_account' properties tally up correctly it should be set to `1`  [here](https://github.com/frappe/erpnext/blob/c800ff50689fe4778360430a7783e7848f81705f/erpnext/accounts/doctype/chart_of_accounts_importer/chart_of_accounts_importer.py#L315)

Issue #27431, opened 2021-09-10
Creating existing, partially depreciated assets with correct depreciation is impossible


blewsky:
## Description of the issue
Creating/uploading existing, partially depreciated assets is broken. When saving/submitting the asset, 1. the "“Depreciation Posting Date”  is dropped, 2. “Expected Value After Useful Life” is changed to a wrong number. Therefore, the depreciation schedule ends up being calculated wrong. This way it is impossible to create existing, partially depreciated assets with a correct depreciation schedule.

## Steps to reproduce the issue

1. Importing the Assets via Data Import with “Calculate Depreciation” = 0 and “Is Existing Asset” = 1, and the respective amount for “Opening Accumulated Depreciation”
2. Entering the Gross Purchase Amount for past Available-for-use and Purchase Dates.
3. Selecting “Calculate Depreciation” and “Allow Monthly Depreciation”, adding “Number of Depreciations Booked”
4. Adding Finance book details (Depreciation method etc.), setting “Expected Value After Useful Life” = 0
5. Saving the Draft. Then ERPnext computes the depreciation schedule gets automatically. The schedule looks fine. The first bug is, that the “Depreciation Posting Date” is empty again and the “Value After Depreciation” is a wrong number (should be 0 in my understanding).
6.. I add the missing “Depreciation Posting Date” and save the Draft. After saving, the “Expected Value After Useful Life” is changed (in my case to exactly the amount of the “Opening Accumulated Depreciation” ) and thus the depreciation schedule is somehow changed to meet the new “Expected Value After Useful Life”. I can fix the “Expected Value After Useful Life” again and set it to 0. If I save the Draft again it will add another time the “Opening Accumulated Depreciation” to it and so on. I don’t see any way to keep the right depreciation schedule for future depreciations.

### Expected result
After saving/submitting, field values should not be changed, dropped by ERPnext. Creating/uploading existing, partially depreciated assets is broken. Otherwise it's impossible to start with a correct opening balance when porting to ERPnext.

## Additional information
More information on this problem and more detailed discussion in the forum:
https://discuss.erpnext.com/t/how-to-add-existing-partially-depreciated-assets/79989

ERPNext: v13.10.2 (HEAD) (also tested on v.13.9.2)
Frappe Framework: v13.10.0 (HEAD)

hans-permana:
hi @nextchamp-saqib , I saw that this issue is being fixed in v13, is there any chance that this fix is also ported back to v12? We have encountered the same issue in v12.

PR #27585, opened 2021-09-19
fix: Calculate depreciation_amount accurately

GangaManoj:
## Proposed Changes

- While modifying the _Depreciation Schedule_ of an Asset after its sale, the _Depreciation Amount_ is not calculated correctly if _Depreciation Entries_ have been posted for it already. This PR ensures that the _Depreciation Amount_ calculated will be accurate.

<details>
<summary>Example</summary>

Initial Asset Value = Rs 100,000
Expected Value After Useful Life = Rs 10,000
Number of Depreciation = 3
Frequency of Depreciation = 12

Screenshot 2021-09-20 at 4 50 24 AM

_Before:_ 

On selling the Asset on 31/12/2021, the Depreciation Amount for the second row should have been 30,000, but instead it's being calculated incorrectly as 20,000.

Screenshot 2021-09-20 at 4 52 22 AM

_After:_

Screenshot 2021-09-20 at 4 53 26 AM

</details>

- If the asset is then returned, the Depreciation Schedule should be reset, but the Depreciation Amount for subsequent rows are not being calculated accurately. This PR fixes that as well.

<details>
<summary>Example</summary>

_Before:_ 

On returning the Asset, the Depreciation Amount for the third row should have been 30,000, but instead it's being calculated incorrectly as -30,000.

Screenshot 2021-09-27 at 10 24 42 PM

_After:_

Screenshot 2021-09-20 at 4 50 24 AM

</details>

- If an Asset was set to be sold in the future on the day it was scheduled to be depreciated, once the Asset gets returned, the Depreciation Entry posted on its sale would not be reversed, even if the posting date for the Credit Note is in the future. This PR adds an extra condition to let the Depreciation Entry be cancelled in cases like this.

<details>
<summary>Explanation</summary>

Screenshot 2021-09-20 at 4 50 24 AM

If this Asset is set to be sold on 31-12-2021(which is in the future right now), an additional Depreciation Entry would be made for that date. On returning this Asset, the Depreciation Schedule should be reset and the Depreciation Entry posted on its sale should be reversed, but because the sale was made on the scheduled date for depreciation, the Depreciation Entry does not get reversed. 

With the new condition, the Depreciation Entry made on sale will be reversed if the sale was set to happen in the future(i.e. if the Posting Date for the Sales Invoice is in the future) but gets returned, even if the date of sale = scheduled date for depreciation.

</details>

- Add more tests for Depreciation Schedule.
- Categorise tests into test suites.


blewsky:
Will that also fix https://github.com/frappe/erpnext/issues/27431 ?

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1041,11 +1040,18 @@ def delete_depreciation_entry_made_after_sale(self, asset):
...
-				if not self.sale_was_made_on_original_schedule_date(asset, schedule, row, posting_date_of_original_invoice):
+				if not self.sale_was_made_on_original_schedule_date(asset, schedule, row, posting_date_of_original_invoice) \
+					or self.sale_happens_in_the_future(posting_date_of_original_invoice):
+
 					reverse_journal_entry = make_reverse_journal_entry(schedule.journal_entry)
 					reverse_journal_entry.posting_date = nowdate()
+					frappe.flags.is_reverse_depr_entry = True
 					reverse_journal_entry.submit()
 
+					asset.flags.ignore_validate_update_after_submit = True
```
nextchamp-saqib, line 1052:
```suggestion
					frappe.flags.is_reverse_depr_entry = False
					asset.flags.ignore_validate_update_after_submit = True
```

erpnext/assets/doctype/asset/test_asset.py
```diff
@@ -724,32 +1012,36 @@ def create_asset(**args):
...
+		"asset_owner": args.asset_owner or "Company",
+		"is_existing_asset": args.is_existing_asset or 1
 	})
 
 	if asset.calculate_depreciation:
 		asset.append("finance_books", {
-			"depreciation_method": "Straight Line",
-			"frequency_of_depreciation": 12,
-			"total_number_of_depreciations": 5
+			"depreciation_method": args.depreciation_method or "Straight Line",
```
nextchamp-saqib, line 1034:
```suggestion
			"finance_book": args.finance_book,
			"depreciation_method": args.depreciation_method or "Straight Line",
```

nextchamp-saqib:

PR #27596, opened 2021-09-20
fix: Tax Breakup table headers fix

nemesis189:
Headers updated for Tax Breakup table since only item wise breakup is displayed


deepeshgarg007:
@Mergifyio backport enterprise-hotfix

PR #27597, opened 2021-09-20
feat: Sales Commission Payout

AfshanKhan:
Introduced 2 new DocTypes: Sales Commission and Process Sales Commission

Sales Commission Payout is a feature for paying out the commission earned by Sales Person. Payout can happen via Additional Salary or Payment Entry

More informations can be found in the documentations.

Documentations [[Sales Commission Payout](https://docs.erpnext.com/docs.erpnext.com/docs/v14/user/manual/en/human-resources/sales-commission-payout), [Process Sales Commission](https://docs.erpnext.com/docs.erpnext.com/docs.erpnext.com/docs/v14/user/manual/en/human-resources/process-sales-commission)]

`no-docs`


erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+			doc = doc = frappe.new_doc("Sales Commission")
+			doc.sales_person = record
+			doc.from_date = self.from_date
+			doc.to_date = self.to_date
+			doc.pay_via_salary = self.pay_via_salary
+			doc.process_sales_commission_reference = self.name
+			doc.add_contributions()
+			doc.insert()
+			if not frappe.db.get_single_value("Selling Settings", "approval_required_for_sales_commission_payout"):
+				doc.reload()
```
deepeshgarg007, line 95: Why doc.reload is needed here?

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1397,6 +1400,8 @@ def get_reference_details(reference_doctype, reference_name, party_account_curre
 					exchange_rate = 1
 		elif reference_doctype == "Gratuity":
 			outstanding_amount = ref_doc.amount - flt(ref_doc.paid_amount)
+		elif reference_doctype == "Sales Commission":
+			outstanding_amount = 0
```
deepeshgarg007, line 1404: What if partial commission is only paid out, are we tracking partial payments anywhere?

erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+		self.validate_from_to_dates()
+		self.validate_salary_component()
+
+	def validate_from_to_dates(self):
+		return super().validate_from_to_dates("from_date", "to_date")
+
+	def validate_salary_component(self):
+		if self.pay_via_salary:
+			if not frappe.db.get_single_value("Payroll Settings", "salary_component_for_sales_commission"):
+				frappe.throw(_("Please set {0} in {1}").format(frappe.bold("Salary Component for Sales Commission"), get_link_to_form("Payroll Settings", "Payroll Settings")))
```
deepeshgarg007: Break this into two lines for better readability

erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+
+	def on_submit(self):
+		self.process_sales_commission()
+
+	def process_sales_commission(self):
+		filter_date = "transaction_date" if self.commission_based_on=="Sales Order" else "posting_date"
+		records = [entry.name for entry in frappe.db.get_all(self.commission_based_on, filters={"company": self.company, "docstatus":1, filter_date: ('between', [self.from_date, self.to_date])})]
+		sales_persons_details = frappe.get_all("Sales Team", filters={"parent": ['in', records]}, fields=["sales_person", "commission_rate", "incentives", "allocated_percentage", "allocated_amount", "parent"])
```
deepeshgarg007: Break into two lines

erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+		if len(sales_persons_details):
```
deepeshgarg007: No need to check `len`, just `sales_persons_details` should be fine

erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+			sales_persons_details_map = self.map_sales_persons_details(sales_persons_list, sales_persons_details)
+			self.make_sales_commission_document(sales_persons_details_map, filter_date)
+
+	def get_sales_persons_list(self, sales_persons):
+		sales_persons_list = sales_persons
+		if self.department or self.designation or self.branch:
+			for person in sales_persons:
+				emp = frappe.db.get_value("Sales Person", filters={"name":person}, fieldname="employee", as_dict=True)['employee']
+				if emp:
+					employee_details = frappe.db.get_value("Employee", filters={"name":emp}, fieldname=["company", "department", "designation", "branch"], as_dict=True)
```
deepeshgarg007: Two queries in for loop. These can easily be moved out of the loop.
Also remember to add a space after `:`

erpnext/payroll/doctype/process_sales_commission/process_sales_commission.py
```diff
@@ -0,0 +1,83 @@
...
+					if self.designation and self.designation != employee_details["designation"]:
+						sales_persons_list.remove(person)
+						continue
+					if self.branch and self.branch != employee_details["branch"]:
+						sales_persons_list.remove(person)
+						continue
+
+		return sales_persons_list
+
+	def map_sales_persons_details(self, sales_persons, sales_persons_details):
```
deepeshgarg007: This function will become redundant if the queries are moved out of the for loop

erpnext/payroll/doctype/sales_commission/sales_commission.js
```diff
@@ -0,0 +1,133 @@
...
+				'label': __('Mode of Payment'),
+				'options': 'Mode of Payment',
+				"get_query": function () {
+					return {
+						filters: {
+							type: ["in", ["Bank", "Cash"]]
+						}
+					};
+				},
+				'reqd': 1
```
deepeshgarg007, line 71: This should not be mandatory, user should be able to make a payment entry even without the mode of mayment

erpnext/payroll/doctype/sales_commission/sales_commission.py
```diff
@@ -0,0 +1,138 @@
...
+		self.validate_from_to_dates()
+
+		self.validate_salary_component()
+		self.calculate_total_contribution_and_total_commission_amount()
+
+	def validate_from_to_dates(self):
+		return super().validate_from_to_dates("from_date", "to_date")
+
+	def validate_amount(self):
+		if self.total_commission_amount <= 0:
```
deepeshgarg007, line 21: We have a docfield check for this in framework itself, you can use that

erpnext/payroll/doctype/sales_commission/sales_commission.py
```diff
@@ -0,0 +1,138 @@
...
+			'reference_doctype': 'Sales Commission',
+			'reference_name': self.name,
+			'due_date': self.to_date,
+			'total_amount': self.total_commission_amount,
+			'outstanding_amount': self.total_commission_amount,
+			'allocated_amount': self.total_commission_amount,
+		}
+		doc.append("references", reference)
+
+def add(record, sales_person):
```
deepeshgarg007: Add a better function name and some comments to explain what this function does
Why contributions with docstatus 0 are being considered?

erpnext/payroll/doctype/sales_commission/sales_commission.py
```diff
@@ -0,0 +1,138 @@
...
+		doc.received_amount = self.total_commission_amount
+		doc.source_exchange_rate = 1
+		doc.target_exchange_rate = 1
+		doc.set("references", [])
+		self.add_references(doc)
+		doc.submit()
+
+		self.db_set("reference_doctype", "Payment Entry")
+		self.db_set("reference_name", doc.name)
+		self.db_set("status", "Paid")
```
deepeshgarg007, line 128: What if there is only a partial payout, we should track this in an outstanding amount field just like in sales invoices and then updated the status accordingly.

deepeshgarg007 (changes requested):

erpnext/payroll/doctype/sales_commission/sales_commission.py
AfshanKhan: That only checks for less than 0 not for equal to 0

erpnext/payroll/doctype/sales_commission/sales_commission.js
AfshanKhan: This is required to get the paid from account for payment entry. Hence mandatory

pushkarjsh88:
@deepeshgarg007 What is the current status on this?  Can this be reopened since it will be required in the Healthcare in some or there other way.

@ChillarAnand

federicocalvo:
Hi @AfshanKhan ! this pr is very useful and a big missing feature in ERPNext. They should add people from Frappe in the control so that it does not die.

PR #27600, opened 2021-09-20
feat(regional): toggle for reduced depreciation rate as per IT Act

nextchamp-saqib:
The rate of depreciation should only be reduced if a company maintains a depreciation schedule as per the income tax act.

The rate shouldn't be reduced for the depreciation schedules which the company maintains in their books.

Added a toggle in the Finance Book to enable/disable the automatic reduction in the depreciation rate.

CleanShot 2021-09-20 at 16 26 49@2x


nextchamp-saqib:

PR #27601, opened 2021-09-20
fix(ux): warn when overbilling allowance was bypassed due to role

ankush:
Change:
- Show message when user has overbilled but system ignored it due to their role. 
- Refactor code (excessive indentation)
- minor perf changes - extract loop invariants like global settings and user's roles. 

Screenshot 2021-09-20 at 4 30 47 PM

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -986,46 +986,49 @@ class AccountsController(TransactionBase):
 		total_overbilled_amt = 0.0
 
 		for item in self.get("items"):
-			if item.get(item_ref_dn):
-				ref_amt = flt(frappe.db.get_value(ref_dt + " Item",
-					item.get(item_ref_dn), based_on), self.precision(based_on, item))
-				if not ref_amt:
-					frappe.msgprint(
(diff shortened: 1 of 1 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -980,42 +980,55 @@ def validate_multiple_billing(self, ref_dt, item_ref_dn, based_on, parentfield):
...
-					   (item.get(item_ref_dn), self.name))[0][0]
+			if not item.get(item_ref_dn):
+				continue
+
+			ref_amt = flt(frappe.db.get_value(ref_dt + " Item",
+				item.get(item_ref_dn), based_on), self.precision(based_on, item))
+			if not ref_amt:
+				frappe.msgprint(
+					_("Warning: System will not check overbilling since amount for Item {0} in {1} is zero")
+						.format(item.item_code, ref_dt))
```
nextchamp-saqib:
```suggestion
				frappe.msgprint(
					_("System will not check overbilling since amount for Item {0} in {1} is zero")
						.format(item.item_code, ref_dt), title=_("Warning"), indicator="orange")
```

ankush:
@mergifyio backport version-12-hotfix version-13-hotfix

PR #27602, opened 2021-09-20
fix(Payment Reconciliation): reference row added in allocation table

Anuja-pawar:
Added Reference Row field in Payment Reconciliation Allocation table.
Fixes this [issue](https://discuss.erpnext.com/t/bug-in-payment-reconciliation-v13-11-0/80407).

Try reconciling a sales invoice with journal entry to replicate the issue.


nextchamp-saqib:

PR #27606, opened 2021-09-20
fix: Dynamic fiscal years for IRS 1099 print format

marination:
**Issue:**
- The IRS 1099 print format for Supplier, was broken as {{ fiscal_year }} is not present in Supplier doctype but was present in the print format.
Screenshot 2021-09-21 at 11 20 52 AM
- You check this format in your local and search for `fiscal_year`

**Fix:**
- Dynamically pull fiscal year in print format
- Patch to set changed HTML in print format as it is now a custom print format. JSON will not change on migrate.

### DIFF
--- a/erpnext/patches/v14_0/dynamic_fiscal_years_for_irs_1099.py
+++ b/erpnext/patches/v14_0/dynamic_fiscal_years_for_irs_1099.py
@@ -1,0 +1,268 @@
+import frappe
+
+def execute():
+	# pull from_year and to_year dynamically in irs_1099 print format
+	irs_1099_print_format = frappe.get_doc("Print Format", "IRS 1099 Form")
+	irs_1099_print_format.html = """
+		{% set from_year, to_year = doc.get_fiscal_years_for_irs_1099() %}
+		<div id="copy_a" style="position: relative; top:0cm; width:17cm;height:28.0cm;">
+		<table>
+		<tbody>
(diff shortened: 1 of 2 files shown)


erpnext/patches/v14_0/dynamic_fiscal_years_for_irs_1099.py
```diff
@@ -0,0 +1,270 @@
+import frappe
+
+
+def execute():
+	# pull from_year and to_year dynamically in irs_1099 print format
+	irs_1099_print_format = frappe.get_doc("Print Format", "IRS 1099 Form")
+	irs_1099_print_format.html = """
```
ankush, line 7: this won't update in new installations, right? So we should change the `.json` for this too. 

Instead re-adding whole format here, we can read the json file and pick the corrected format from there.

ankush (changes requested):

erpnext/patches/v14_0/dynamic_fiscal_years_for_irs_1099.py
ankush: ignore this

ankush: `frappe.reload_doc("regional", "print_format", "irs_1099_form", force=True)`

works with custom formats too.

PR #27609, opened 2021-09-20
fix: Unlink PO on cancelling SO

GangaManoj:
_Problem:_

- On cancelling a Sales Order, Purchase Orders linked with it also get cancelled.

_Fix:_

- Unlink associated Purchase Orders on cancelling Sales Orders.


erpnext/controllers/accounts_controller.py
```diff
@@ -815,6 +816,57 @@ def on_cancel(self):
 			if frappe.db.get_single_value('Accounts Settings', 'unlink_advance_payment_on_cancelation_of_order'):
 				unlink_ref_doc_from_payment_entries(self)
 
+			if self.doctype == "Sales Order":
```
deepeshgarg007, line 818: We can also create Purchase Order from Sales Order as well. Shouldn't this be other way round as well?

erpnext/controllers/accounts_controller.py
```diff
@@ -815,6 +816,57 @@ def on_cancel(self):
...
+	def unlink_ref_doc_from_po(self):
+		print("\n"*5, "*"*50, "\n"*5)
+
+		so_items = []
+		for item in self.items:
+			so_items.append(item.name)
+
+		print("SO Items: ", so_items)
+
+		linked_po = frappe.get_all(
```
deepeshgarg007: As far as I know, `set_value` only updates one record at a time, can you try with multiple items as well?

deepeshgarg007 (changes requested):

erpnext/controllers/accounts_controller.py
GangaManoj: yeah, it worked perfectly when I tried it on the console, which is why this is so baffling.

erpnext/controllers/accounts_controller.py
GangaManoj: There doesn't seem to be an option to do that.

Screenshot 2021-10-28 at 7 20 19 PM

deepeshgarg007:

PR #27611, opened 2021-09-20
fix: Hero Slider Control & Alignment fixes

shariquerik:
**Issue:**
1. When a user tries to change an image using the slider left-right controls, it doesn't work.
2. The Title, Subtitle, Action Button Alignment is below the Slider Controls.

**Cause:**
In the current hero slider HTML, the slider has an `id` attribute which is set as `slider_name`.
Now if the slider name doesn't have any space (e.g. SliderName) in it then the left-right controls work, but if the slider name has space (e.g Slider Name) it doesn't work.

**Solution:**
Creating a unique hash for slider and setting it as `id` instead of `slider_name`

**CONTROLS**
**Before:**
SliderBug

**After:**
SliderFixed

**ALIGNMENT**
**Before:**
image

**After:**
image

### DIFF
--- a/erpnext/public/scss/shopping_cart.scss
+++ b/erpnext/public/scss/shopping_cart.scss
@@ -31,6 +31,14 @@ body.product-page {
 .carousel-control-prev,
 .carousel-control-next {
 	opacity: 1;
+	width: 8%;
+
+	@media (max-width: 1200px) {
+		width: 10%;
+	}
+	@media (max-width: 768px) {
+		width: 15%;
+	}
 }
 
 .carousel-body {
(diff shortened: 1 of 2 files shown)


harshit-30:
@shariquerik @marination please review this at your earliest, we need to release this in our next release. 

@shariquerik please do check, if the buttons **(Primary Button)** work properly after the fix.

shariquerik:
> @shariquerik please do check, if the buttons **(Primary Button)** work properly after the fix.

Yes, primary button works fine after the fix, I have tested.

marination (approved):

marination:

marination:

PR #27623, opened 2021-09-21
fix: set item uom as stock_uom if it isn't set

18alantom:
- **UOM** doesn't get set on _Upload_ if it isn't present in the `.csv` but **Stock UOM** gets set.
- This will copy the **Stock UOM** value to **UOM** if it isn't set.


erpnext/stock/doctype/stock_entry/stock_entry.js
```diff
@@ -323,6 +323,12 @@ frappe.ui.form.on('Stock Entry', {
 		attach_bom_items(frm.doc.bom_no)
 	},
 
+	before_save: function(frm) {
+		frm.doc.items.forEach((item) => {
+			item.uom = item.uom??item.stock_uom;
```
ankush, line 328: better to avoid "too recent" js features.

ankush (dismissed):

erpnext/stock/doctype/stock_entry/stock_entry.js
ankush:
```suggestion
			item.uom = item.uom || item.stock_uom;
```

Both are equivalent in this case. Switching to or for better compatibility.

PR #27624, opened 2021-09-21
fix: previous_sle() fetched incorrect sle for batched items

noahjacob:
### For Example
Let's consider Item **"Batch Item 1"** with batch **"NSJ0001"** and "Allow Negative Stock" is disabled.
image

Now if we try to make a back-dated stock entry for Batch **"NSJ0001"**, it does not throw a negative stock error.
sle

## Reason
- `previous_sle()` did not consider batch_no while fetching the last SLE. It used to fetch only on t he basis of item_code and warehouse.
- This would fetch wrong actual_qty for batch items since there could be other batch's stock entries before the original batch date. 

## Fix

- `previous_sle()` now fetches last SLE based on batch_no for batched items.


ankush:
Duplicate of https://github.com/frappe/erpnext/pull/26914

PR #27626, opened 2021-09-21
fix: empty row on new maintenance visit.

noahjacob:
### Changes
- Clearing empty table row on creating a new visit.

table

### After Changes
table_fix


erpnext/maintenance/doctype/maintenance_visit/maintenance_visit.js
```diff
@@ -27,6 +27,9 @@ frappe.ui.form.on('Maintenance Visit', {
 		if (frm.doc.__islocal) {
 			frm.set_value({mntc_date: frappe.datetime.get_today()});
 		}
+		if (frm.doc.purposes[0].item_name == undefined) {
```
ankush:
```suggestion
		if (frm.doc.purposes.length && frm.doc.purposes[0].item_name == undefined) {
```

PR #27629, opened 2021-09-21
fix: distribution of additional costs in mfg stock entry

ankush:
Problem: allocation of additional cost for scrap item is broken. 

Root cause: allocation logic was changed to avoid adding it to scrap item but it doesn't allocate correctly causing excess allocation. Ref:  https://github.com/nabinhait/erpnext/commit/e10f0a5a6cd344b0e2ea2b65a992b37e9a4594d6

Solution:

For Mfg/repack stock entries: Allocate all additional cost on FG items only.
For other stock entries: allocate all additional costs on all incoming items.

(refer added test case for example)


ankush:

ankush:
needs a test for GL entries.

erpnext/patches/v13_0/fix_additional_cost_in_mfg_stock_entry.py
```diff
@@ -0,0 +1,70 @@
...
+	for item in stock_entry.items:
+		item.db_update()
+	return stock_entry
+
+
+def create_repost_item_valuation(stock_entry):
+	from erpnext.controllers.stock_controller import create_repost_item_valuation_entry
+
+	# turn on recalculate flag so reposting corrects the incoming/outgoing rates.
+	frappe.db.set_value("Stock Ledger Entry", {"voucher_no": stock_entry.name}, "recalculate_rate", 1)
```
rohitwaghchaure: Should only for FG and not for the raw materials, you could add actual_qty > 0 condition as well

ankush: Added filter for actual_qty > 0

PR #27660, opened 2021-09-26
fix: tax rate being overridden in case of 0.00

rtdany10:
In expense claim, tax rate could be different for different expenses.
Therefore, rate is kept as 0.00 and tax amount entered manually in the tax table.
But fetch functionality overrides the tax rate (upon saving) and mess up the amount.

This PR removes the fetch functionality of the rate column.
#no-docs Please backport to v13 and v12 as well.

### DIFF
--- a/erpnext/hr/doctype/expense_taxes_and_charges/expense_taxes_and_charges.json
+++ b/erpnext/hr/doctype/expense_taxes_and_charges/expense_taxes_and_charges.json
@@ -56,8 +56,6 @@
   },
   {
    "columns": 2,
-   "fetch_from": "account_head.tax_rate",
-   "fetch_if_empty": 1,
    "fieldname": "rate",
    "fieldtype": "Float",
    "in_list_view": 1,
--- a/erpnext/hr/doctype/expense_claim/test_expense_claim.py
+++ b/erpnext/hr/doctype/expense_claim/test_expense_claim.py
@@ -176,7 +176,7 @@ def generate_taxes():
 	account = create_account(company=company_name, account_name="Output Tax CGST", account_type="Tax", parent_account=parent_account)
 	return {'taxes':[{
 		"account_head": account,
-		"rate": 0,
+		"rate": 9,
 		"description": "CGST",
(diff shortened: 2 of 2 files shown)


rtdany10:
Introducing an expense tax template (like in case of items in invoices) would be a better way to handle taxes.
Thoughts?

erpnext/hr/doctype/expense_taxes_and_charges/expense_taxes_and_charges.json
```diff
@@ -56,8 +56,6 @@
   },
   {
    "columns": 2,
-   "fetch_from": "account_head.tax_rate",
-   "fetch_if_empty": 1,
```
deepeshgarg007, line 60: @rtdany10 fetch_if_empty was enabled so it won't override if the user has updated any value manually. Can you please give a more elaborate example in the description? I am unable to understand the use case for which it is causing an issue

rtdany10: Tax rate
Sorry I didn't include it earlier.

deepeshgarg007: @rtdany10 I get the issue now. One test case seems to be failing because of this change. Can you fix that too?

rtdany10: Yes, I'll fix the test case.
Also, can you give your opinion on tax template against each expense?

rtdany10:
@Mergifyio backport version-12-hotfix version-13-hotfix

deepeshgarg007:
@rtdany10 In this the modified timestamp wasn't changed so the changes didn't apply on migrate, can raise a PR to fix this

rtdany10:
> @rtdany10 In this the modified timestamp wasn't changed so the changes didn't apply on migrate, can raise a PR to fix this

My bad, will raise a PR. Thanks!

PR #27661, opened 2021-09-27
refactor: fetching of account balance in chart of accounts

nextchamp-saqib:
### Before: 

Took ~40 secs

CleanShot 2021-09-27 at 11 10 34

### After:

Took 18 secs

CleanShot 2021-09-27 at 11 08 32

Changes:
- Display Nodes and then Asynchronously fetch balance
- Only fetch balance if the user has permissions for GL Entry

Depends On: https://github.com/frappe/frappe/pull/14310


nextchamp-saqib:

PR #27664, opened 2021-09-27
fix: cannot set custom label for 'total' field in print format

nextchamp-saqib:
The 'Total' field is rendered with a custom template in Sales Invoice print formats.

The template fetches the label from meta, which required users to change the label from the Customize Form to change the label in print format

To test the fix:
- Open any custom print format of Sales Invoice
- Using Print Format Builder, change the label of Total field
- Check if the label is changed in any document's print preview.


nextchamp-saqib:
@mergifyio backport enterprise-hotfix

nextchamp-saqib:

PR #27665, opened 2021-09-27
fix: trim sales invoice custom field lengths

nextchamp-saqib:
Applies #27136 just on custom field


erpnext/regional/india/setup.py
```diff
@@ -426,7 +434,8 @@ def make_custom_fields(update=True):
 			'fieldtype': 'Date',
 			'insert_after': 'driver_name',
 			'default': 'Today',
-			'print_hide': 1
+			'print_hide': 1,
+			'length': 10
```
ankush: Length for `Date` type?

deepeshgarg007 (approved):

deepeshgarg007:

PR #27668, opened 2021-09-27
fix: Batch scans get overwritten on the same row

marination:
**Issue:**
- Scan a batch no e.g. BAT-001 , it adds a new row with batch BAT-001
- Scan a second batch no belonging to the same item (BAT-002), it overwrites the previous row and changes BAT-001 to BAT-002, and increments qty by 1

**Fix:**
- Everytime a **new batch is scanned** add a **new row** since we maintain one batch per item row
- If same batch is scanned twice, increment qty in pre-existing batch row
- Broke `scan_barcode` event into smaller functions for readability

**To test:**
- Test barcode, serial no and batch scan in stock transactions (stock entry , purchase receipt, delivery note, sales/purchase invoice)
- Test barcode scan in PO and SO (they don't have batch serial fields and ideally users would not scan a batch there but you can test it , it will add a row)


ankush:
>If the same batch is scanned twice, show toast saying that batch is already added

Increment the quantity instead of showing the message? :thinking:

marination:
@ankush yeah sounds better

marination:

PR #27673, opened 2021-09-27
feat: Create PL for several SOs and Create multiple DN from PL, grouped by customer.

KrithiRamani:
1. To enable creation of PL for SOs of several customers, removed validation for Customer field.

Screenshot 2021-09-20 at 11 57 24 AM

2. Set Picked Qty to zero, on select of Items. On submit, the Picked Qty will be set to Stock Qty.
Screenshot 2021-09-20 at 12 01 30 PM

3. From a PL created from several SOs, on click of Create -> Delivery Note
       - Separate DN for sales orders(grouped by customer) will be created
       - Separate DN for all items that do not have any SO linked.
Screenshot 2021-09-27 at 9 45 18 PM
Screenshot 2021-09-27 at 9 45 38 PM

4. Added new field per_picked in SO, set this on submit of PL. Used this field to filter SO in Get Items of PL.
Screenshot 2021-09-27 at 9 47 53 PM


erpnext/stock/doctype/pick_list/pick_list.js
```diff
@@ -143,13 +143,10 @@ frappe.ui.form.on('Pick List', {
 			docstatus: 1,
 			per_delivered: ['<', 100],
 			status: ['!=', ''],
+			per_picked: ['=',0],
```
rohitwaghchaure: per_picked < 100, could be partial

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -41,6 +51,24 @@ def before_submit(self):
...
 				.format(frappe.bold(item.item_code), frappe.bold(item.idx)), title=_("Quantity Mismatch"))
 
+	def before_cancel(self):
+		#update picked_qty in SO Item on cancel of PL
+		self.update_so()
+
+	def update_so(self):
+		for item in self.locations:
+			if item.sales_order_item:
+				frappe.db.set_value("Sales Order Item",item.sales_order_item,"picked_qty",item.picked_qty)
```
rohitwaghchaure: What if same item picked multiple times in the different pick lists? For example in the sales order item's qty is 5 and use has made 2 pick list with 4 qty in the first pick list and 1 qty in the second pick list

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -41,6 +51,24 @@ def before_submit(self):
...
+				so_doc = frappe.get_doc("Sales Order",item.sales_order)
+				if self.docstatus == 2:
+					per_picked = 0
```
rohitwaghchaure: What if there is one more pick list exists

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -27,9 +27,19 @@ def validate(self):
...
 		self.set_item_locations()
+		# set percentage picked in SO
+		for i in self.get('locations'):
+			if i.sales_order and frappe.db.get_value("Sales Order",i.sales_order,"per_picked") > 0:
+				frappe.throw("Row " + str(i.idx) + " has been picked already!")
 
 	def before_submit(self):
 		for item in self.locations:
+			# set picked_qty to stock_qty, before submit
+			item.picked_qty = item.stock_qty
```
rohitwaghchaure: if item.picked_qty is set by the user manually then shouldn't be override

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -314,64 +344,88 @@ def get_available_item_locations_for_other_item(item_code, from_warehouses, requ
...
+				'rate': 'rate',
+				'name': 'name',
+				'parent': '',
+			}
+		}
+		map_pl_locations(pick_list.locations,None)
+		delivery_note.insert(ignore_mandatory = True)
 
+	frappe.msgprint('Delivery Note(s) created for the Pick List!')
 	return delivery_note
```
rohitwaghchaure, line 423: create_delivery_note looks a big function so difficult to read , could you split it into smaller functions

erpnext/stock/doctype/pick_list/pick_list.py
KrithiRamani: Added validation on submit of PL.

erpnext/stock/doctype/pick_list/pick_list.py
KrithiRamani: Handled.

erpnext/stock/doctype/pick_list/pick_list.py
KrithiRamani: Done

KrithiRamani:
Made changes as suggested and reopening PR

KrithiRamani:
shows too many commits . will rebase and create new PR

PR #27690, opened 2021-09-28
fix: cannot delete a project if linked with sales order

nextchamp-saqib:
Backport of #27689


erpnext/projects/doctype/project/test_project.py
```diff
@@ -32,7 +33,22 @@ def test_project_with_template(self):
...
+
+		so.reload()
+		self.assertEqual(so.project, project.name)
+
+		project.delete()
+
+		so.reload()
+		self.assertFalse(so.project)
+
+def get_project(name, template):
```
nextchamp-saqib:
```suggestion
def get_project(name):
```

PR #27694, opened 2021-09-29
fix: restore inadvertently removed function `calculate_total_billing_amount`

sagarvora:
`calculate_total_billing_amount` was inadvertently removed [here](https://github.com/frappe/erpnext/commit/8547cb5448f5a38e59f5f6800615957a2081b1e7#diff-bb9310f10e3db773b4664e5f5fd681670d41beca380c1a967fcfc44acea8b2cdL972), although it is still being used whenever the `currency` event is triggered.

---

image

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
@@ -1004,6 +1004,18 @@ frappe.ui.form.on("Sales Invoice Timesheet", {
 	}
 });
 
+var calculate_total_billing_amount =  function(frm) {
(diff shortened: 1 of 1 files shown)


ankush:
Looks like it was removed intentionally. The commit message is misleading :sweat_smile: 

https://github.com/frappe/erpnext/pull/26783/commits/b57521a33719258ee3d9b41bc99ec50e4982ccaa

sagarvora:
@ankush,

Thanks for the feedback!

This has been properly fixed [here](https://github.com/frappe/erpnext/pull/27742/commits/4e785b9491110b68b4128ceabacf6552387c9db9).

PR #27700, opened 2021-09-29
fix: remove stale doctypes and add msg for ecommerce refactor

ankush:
Closes https://github.com/frappe/erpnext/issues/27550 & ISS-21-22-06573

- delete stale doctypes and leave a message for all users when logging in. (only on sites where E commerce is enabled)

Screenshot 2021-09-29 at 11 13 34 PM

PR #27703, opened 2021-09-29
ci: check for stale doctypes in CI

ankush:
To avoid errors like: https://github.com/frappe/erpnext/pull/27313 https://github.com/frappe/erpnext/pull/27700


ankush:
this wont catch anything that didn't exist in v10 🤔

PR #27712, opened 2021-09-30
fix(India): Internal transfer check fix

deepeshgarg007:
If neither the Customer/Supplier nor the Company Address had a GSTIN added, then the transaction between these parties was wrongly considered an internal transaction.

Due to this when a PO was converted to Purchase Invoice, the taxes added in PO were removed when it got converted to Purchase Invoice

Steps to test:
1. Add a Company Address with no GSTIN
2. Add an address for a Supplier without GSTIN
3. Make a Purchase Order, with that company (company should have country as India) add some taxes in it
4. Using the menu option convert the Order to Invoice and make sure the same taxes from PO are copied in the invoice


deepeshgarg007:
@Mergifyio backport develop

deepeshgarg007:

PR #27717, opened 2021-09-30
fix(Org Chart): use attribute selectors instead of ID selector for node IDs with special characters

ruchamahabal:
## Problem:

- In Org Chart, individual nodes have an id set as Employee ID. 
- ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").
- If you have employee IDs created with special characters other than these like EMP/01674 having a forward slash, the node selection fails and org chart rendering breaks

image

## Fix:

- Use Attribute selectors like ```$(`[id="${node.id}"]`)``` instead of node IDs ```$(`#${node.id}`)``` to overcome this since the employee IDs can be any form.

image

### DIFF
--- a/cypress/integration/test_organizational_chart_desktop.js
+++ b/cypress/integration/test_organizational_chart_desktop.js
@@ -6,7 +6,7 @@ context('Organizational Chart', () => {
 
 	it('navigates to org chart', () => {
 		cy.visit('/app');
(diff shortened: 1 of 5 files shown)


ruchamahabal:

PR #27720, opened 2021-09-30
fix: Website Items with same Item name unhandled, thumbnails missing

marination:
## Website Item Naming
**Issue:**
- If two items have different item codes and **same item name**, while creating website items via patch, it throws a Duplicate Error
- This is because the `website_item_name` (copied from item name) is used to name the Website Item

**Fix:**
-  Use naming series for Website Item and re-run patch to make sure any pending items to be published are published.

## Thumbnails missing
- `make_thumbnail` was skipped in patches due to which Website Items have missing thumbnails
- This impacts the cart, search, recommended items, etc.
**Fix:**
- In the website item creation patch (that will be re-run), for existing Website Items check if thumbnail is missing. If yes, trigger `make_thumbnail`
- For new Website Items thumbnails will now be created after removing the patch condition in  `make_thumbnail`

## Miscellanous
- Website Item Page (full page view) breaks if **Enable Shopping Cart** is disabled in **E Commerce Settings**
- ‘Request for Quote’ button & cart summary not visible if **Enable Checkout** is disabled in **E Commerce Settings**


marination:

PR #27725, opened 2021-10-01
fix: option to limit reposting in certain timeslot

ankush:
If repost queue is large, then processing it during working hours can cause timeout errors while doing regular operations. 

This PR adds a configuration for tweaking the time when reposting should happen. Options:
- Toggle limiting to a timeslot.
- Select timeslot/day (minimum 10 hours)
- Select a weekday where limits don't apply.

Screenshot 2021-10-01 at 12 42 52 PM


ankush:
@mergifyio backport version-13-hotfix version-13-pre-release

PR #27735, opened 2021-10-01
fix: Sales Order cancellation message

GangaManoj:
_Problem:_

When cancelling a Sales Order which has a Sales Invoice in the draft state linked to it:
- The user is directed to 'cancel' the linked invoice, even though only submitted invoices can be cancelled.
- If the invoice has more than one Item, the name of the invoice will be repeated multiple times(as many times as there are Items in it).

Screenshot 2021-10-02 at 2 21 33 AM

_Fix:_

- Edit the message to instruct the user to delete the linked invoice instead.
- Only display the name of the invoice once, irrespective of the number of Items in it.

_Additional changes:_

- Remove redundant code
- Add tests


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -241,14 +241,24 @@ def check_nextdoc_docstatus(self):
...
-			where t1.name = t2.parent and t2.sales_order = %s and t1.docstatus < 2""",
+			where t1.name = t2.parent and t2.sales_order = %s and t1.docstatus = 1""",
 			self.name)
 
 		if submit_rv:
 			submit_rv = [get_link_to_form("Sales Invoice", si) for si in submit_rv]
 			frappe.throw(_("Sales Invoice {0} must be cancelled before cancelling this Sales Order")
 				.format(", ".join(submit_rv)))
 
+		draft_rv = frappe.db.sql_list("""select distinct t1.name
```
deepeshgarg007: Instead of running two queries, just get both of them in a single query and also select `docstatus` along with `name`. Post query you can segregate the invoices and show the validation message accordingly

GangaManoj: Should I keep the draft_rv thing and get rid of everything else in the function? Because all the other conditions check for submitted docs, which means the errors won't ever be thrown, because the Cancel All dialog box would show up.

deepeshgarg007 (approved):

PR #24603, opened 2021-02-10
feat: E-commerce Refactor

marination:
- [x]  Create **Website Item** to handle all Website Related Data
- [x] Publish Item from Item Master (create Website Item)
- [x] Consolidated **E-commerce Settings** (replace Products and Shopping Cart Settings)
- [x] Drive Portal via **Website Item** and **E Commerce Settings** only (cleanup item and make stuff obsolete)
- [x] Patch for:  Items ➡️   Website Items  (needs load testing)
- [x] Patch: Shopping Cart & Products Settings ➡️    E Commerce Settings
- [x] Patch: Homepage Products :- Items ➡️    Website Items 
- [x] Fix Tests

------
- [x] All Categories page (configurable)
- [x] Wishlist/Add to Cart (Card Actions) 
- [x] Wishlist Full Page
- [x] Item Full Page (better arrangement of info, add to wishlist) 
- [x] Ratings and Reviews
- [x] Capturing discounts
- [x] Shopping Cart Page (remove items in a click and bug fixes)
- [x] Grid and List view Toggling (Listing Page)
- [x] Search
- [x] Server side Test cases
- [x] Review changes
-----------------------------------

Redisearch Docs: https://github.com/frappe/erpnext_documentation/pull/375
[Official Documentation](https://docs.erpnext.com/docs/v13/user/manual/en/e_commerce/set_up_e_commerce)

[Unofficial E-commerce guide](https://docs.google.com/document/d/17s6VxWeihTImSG2toR5Ma1jNq9nNAaYueP-dimNdQ4I/edit?usp=sharing)

PR #27742, opened 2021-10-02
 feat: add `total_billing_hours` to Sales Invoice (fp #26783)

sagarvora:
- Front-port of https://github.com/frappe/erpnext/pull/26783
- Fixes [invalid reference to older function](https://github.com/frappe/erpnext/pull/27694)
- Also fixes same field appearing twice in **Sales Invoice Timesheet** (`column_break_5`)

To be backported again for consistency between branches and applying fixes 😅


sagarvora:
@nabinhait Let's not accept PRs directly into `version-13-hotfix`. These are hard to front-port ([another example](https://github.com/frappe/erpnext/pull/24603)).

PR #27749, opened 2021-10-02
fix: Display appropriate message for Payment Term discrepancies in Payment Entry

GangaManoj:
_Problem:_

When a Payment Entry is created for an Invoice, but the Payment Term in the Payment Entry does not correspond with the ones used in the Invoice, the following message is displayed:

Screenshot 2021-10-02 at 8 50 37 PM

This doesn't give a clear indication of the problem, i.e. the mismatch in Payment Terms.

_Fix:_

Display the following message instead.

Screenshot 2021-10-02 at 9 07 46 PM


deepeshgarg007 (approved):

deepeshgarg007:

PR #27753, opened 2021-10-02
fix: Remove pointless buttons from Payment Order

GangaManoj:
_Proposed Change:_

- The Payment Order References table is read-only, but has an 'Add Row' button, which is clearly pointless.

image


nextchamp-saqib:

Issue #26746, opened 2021-07-30
Changes to Company abbreviation don't propagate properly to Cost Centers, Tax templates, Warehouses etc.


blewsky:
## Description of the issue
Changing the abbreviation for a Company, changes the abbreviations appended in the Chart of Accounts but not to Cost Centers. 

## Steps to reproduce the issue
1. Rename company abbreviation using `Abbr.` Field in Company Doctype. This is the pop-up:
![image](https://user-images.githubusercontent.com/87971956/127698327-5521149c-8289-4a88-82b2-5beebe57ab10.png)
2. Save changes, wait until changes affect the entries in the database and reload company page.
2. (Check Chart of Accounts. The abbreviations added to each Account name are changed.)
3. Go-to Cost Centers. The Cost Centers still have the old abbreviations.

### Observed result
- Abbreviations in the name of `Cost Center` are not changed according to the change in `Company` doctype.
- To make things worse: It doesn't work to rename the Cost Centers manually to match the new abbreviations. 

### Expected result
Changes to the abbreviation in `Company` doctype should directly propagate correctly and affect the abbreviation used in the name of all `Cost Center` entries.

## Additional information
ERPNext: v13.7.0 (HEAD)
Frappe Framework: v13.7.0 (HEAD)

blewsky:
It seems this problem is bigger than previously thought. The company abbreviations are also not correctly propagated to the Tax Templates and Warehouses. The weird thing is, that some Warehouses etc. are renamed correctly and others not. It seems this "feature" (= renaming a company abbreviation) should be overhauled and checked for bugs. A company can potentially have many Accounts, Warehouses, Tax templates etc. that can be affected by the renaming of the company abbreviation.  Not having proper names after a change in the abbreviation, can create confusion or other bugs. Therefore, I think this is a serious issue that should be fixed.

Personally, I think it is not a good practice to add the company abbreviation automatically to all these dependent fields for two reasons:
1. Propagating changes is very error-prone (as we can see here).
2. People might want a different naming procedure
(e.g. I dislike the space characters between the doctype name and the company abbreviation others might want to hide the company abbreviation because they don't need it.)

blewsky:
Related discussion in the forum:
https://discuss.erpnext.com/t/renaming-company-abbreviation-is-broken-how-to-fix-cost-center-names/78628

blewsky:
[...]

PR #27766, opened 2021-10-04
refactor!: remove abbreviation renaming

nextchamp-saqib:
Problem 1: Renaming abbreviation is a heavy task if a user has a lot of transactions (links to abbreviated documents).
For eg. renaming one single account which has links to 20k+ documents, raises a "Too many commits in one transaction" error.

Problem 2: Some of the doctype names are abbreviated, but those doctypes are not included while renaming, for eg. Item Tax Template. Hence after renaming, item tax templates have old abbreviations.

Abbreviation only serves one purpose, that is to differentiate between 2 companies' bootstrapped documents such as accounts, cost center, etc. Ideally one should not need to rename abbreviations. Users can still set the abbreviation at the time of creating companies.


ankush (approved):

blewsky:
Probably closes https://github.com/frappe/erpnext/issues/26746 ?

nextchamp-saqib:

PR #27767, opened 2021-10-04
fix(Lead): Email ID need not be unique

netchampfaris:
A lead can come back after say 2 years. This validation doesn't make sense to me. Am I missing something?

### DIFF
--- a/erpnext/crm/doctype/lead/lead.py
+++ b/erpnext/crm/doctype/lead/lead.py
@@ -36,7 +36,6 @@ class Lead(SellingController):
 		self.set_lead_name()
 		self.set_title()
 		self.set_status()
-		self.check_email_id_is_unique()
 		self.validate_email_id()
 		self.validate_contact_date()
 		self._prev = frappe._dict({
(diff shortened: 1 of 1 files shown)


rtdany10:
> A lead can come back after say 2 years. This validation doesn't make sense to me. Am I missing something?

We can re-use the old lead document in that case?

netchampfaris:
> We can re-use the old lead document in that case?

Maybe. But is it worth the effort? If a user wants context, it's just a filter away.

rtdany10:
> > We can re-use the old lead document in that case?
> 
> Maybe. But is it worth the effort? If a user wants context, it's just a filter away.

We have a call center. Only creators/lead owners can view their leads. So, filters might fail in that scenario. We even had to make the mobile number field unique to avoid lead duplication.

Agent A doesn't have access to lead created by Agent B. Hence, this uniqueness is a necessary for us. Else Agent A and B might create multiple leads with same email IDs. And no leads would like to be contacted by different agents without context of previous conversation.

This is just a scenario I've pointed out. This uniqueness helps avoid lead duplication. If the code is removed, we can still customize the form and make the field unique, so that shouldn't be an issue. But a small enterprises running ERPNext without much technical knowledge might have to get professional help to do that.

rtdany10:
@netchampfaris I believe this PR is a step in the right direction. Keeping this validation hardcoded isn't right.

ruchamahabal:
> @netchampfaris I believe this PR is a step in the right direction. Keeping this validation hardcoded isn't right.

**Allow Lead Duplication based on Emails** configuration is being added via https://github.com/frappe/erpnext/pull/27788/

PR #27768, opened 2021-10-04
feat: employee initial work history updated when transfer is performed

mohammedyusufshaikh:
Before this feature when an employee was transferred their initially work history was not updated in the internal work history table only the new transfer details were updated in the table so it was an issue to track their previous details

So after this feature the initial history as well as the new transfer details all this will be updated in the employee internal work history table.

Lets take and example of a sample employee detail
Screenshot 2021-10-04 at 1 32 47 PM

Before this feature was added and a transfer would be performed only the recent transfer details would be shown
Screenshot 2021-10-04 at 1 36 56 PM

After this feature is added both the previous and and recent history will be show
Screenshot 2021-10-04 at 1 35 07 PM

`no-docs`


erpnext/hr/doctype/employee_transfer/test_employee_transfer.py
```diff
@@ -57,3 +61,65 @@ def test_new_employee_creation(self):
 		self.assertTrue(transfer.new_employee_id)
 		self.assertEqual(frappe.get_value("Employee", transfer.new_employee_id, "status"), "Active")
 		self.assertEqual(frappe.get_value("Employee", transfer.employee, "status"), "Left")
+
+	def test_employee_history(self):
+		name = frappe.get_value("Employee", {"first_name": "John", "company": "Test Company"}, "name")
+		doc = frappe.get_doc("Employee",name)
+		record_count  = len(doc.internal_work_history)
```
ruchamahabal:
```suggestion
		record_count = len(doc.internal_work_history)
```

erpnext/hr/utils.py
```diff
@@ -44,9 +52,30 @@ def update_employee(employee, details, date=None, cancel=False):
...
 			internal_work_history[item.fieldname] = item.new
+
 	if internal_work_history and not cancel:
 		internal_work_history["from_date"] = date
 		employee.append("internal_work_history", internal_work_history)
+
+	if cancel:
+		filters = {}
+		for d in details:
+			for h in employee.internal_work_history:
```
ruchamahabal:
```suggestion
			for history in employee.internal_work_history:
```
Better naming?

erpnext/hr/utils.py
```diff
@@ -44,9 +52,30 @@ def update_employee(employee, details, date=None, cancel=False):
...
+					designation = d.new
+					filters["designation"] = designation
+				if d.property == "Branch" and h.branch == d.new:
+					branch = d.new
+					filters["branch"] = branch
+				if date and date == h.from_date:
+					filters["from_date"] = date
+		if filters:
+			frappe.db.delete("Employee Internal Work History", filters)
+			frappe.db.commit()
```
ruchamahabal:
```suggestion
```

We can remove the commit here

erpnext/hr/utils.py
```diff
@@ -30,6 +30,14 @@ def set_employee_name(doc):
 		doc.employee_name = frappe.db.get_value("Employee", doc.employee, "employee_name")
 
 def update_employee(employee, details, date=None, cancel=False):
+	if not employee.internal_work_history and not cancel:
+		employee.append("internal_work_history",{
```
ruchamahabal:
```suggestion
		employee.append("internal_work_history", {
```

erpnext/hr/utils.py
```diff
@@ -44,9 +52,30 @@ def update_employee(employee, details, date=None, cancel=False):
 		setattr(employee, item.fieldname, new_data)
 		if item.fieldname in ["department", "designation", "branch"]:
 			internal_work_history[item.fieldname] = item.new
+
 	if internal_work_history and not cancel:
 		internal_work_history["from_date"] = date
 		employee.append("internal_work_history", internal_work_history)
+
+	if cancel:
```
ruchamahabal, line 60: We can separate the code from this line onwards into a new function?

`delete_employee_work_history`

erpnext/hr/utils.py
```diff
@@ -30,6 +30,14 @@ def set_employee_name(doc):
 		doc.employee_name = frappe.db.get_value("Employee", doc.employee, "employee_name")
 
 def update_employee(employee, details, date=None, cancel=False):
```
ruchamahabal: Should we rename this to `update_employee_work_history` since it is only doing that?
Update the name wherever this is getting called.

erpnext/hr/doctype/employee_transfer/test_employee_transfer.py
```diff
@@ -13,9 +14,12 @@
 
 class TestEmployeeTransfer(unittest.TestCase):
 	def setUp(self):
+		frappe.db.sql("""delete from `tabEmployee Transfer`""")
```
ruchamahabal: Add `tearDown` too with `frappe.db.rollback`. You won't have to use ```frappe.db.sql("""delete from `tabEmployee Transfer`""")``` then.

erpnext/hr/doctype/employee_transfer/test_employee_transfer.py
```diff
@@ -57,3 +63,65 @@ def test_new_employee_creation(self):
...
+		name = frappe.get_value("Employee", {"first_name": "John", "company": "Test Company"}, "name")
+		doc = frappe.get_doc("Employee",name)
+		record_count = len(doc.internal_work_history)
+		self.assertEqual(record_count, 2)
+		data = frappe.db.get_list("Employee Transfer", filters={"employee":name}, fields=["*"])
+		doc = frappe.get_doc("Employee Transfer", data[0]["name"])
+		doc.cancel()
+		employee_doc = frappe.get_doc("Employee",name)
+		record_count = len(employee_doc.internal_work_history)
+		self.assertEqual(record_count, 1)
```
ruchamahabal: Check the data set in line items in this test too: after transfer submission and cancellation. Not just the length of records.

ruchamahabal (approved):

ruchamahabal:

Issue #27521, opened 2021-09-15
New subscriptions not generated


barredterra:
### Reproduce

I created a new **Subscription** in July. ERPNext correctly generated the invoice for July. Then it stopped generating new invoices. Now I have a subscription like this:

![Bildschirmfoto 2021-09-15 um 15 35 30](https://user-images.githubusercontent.com/14891507/133443641-31286936-1d5e-4ccc-84e4-b4ecb85c6d53.png)
![Bildschirmfoto 2021-09-15 um 15 36 55](https://user-images.githubusercontent.com/14891507/133443806-92e7369e-9127-4c4e-b8b3-3fdd06ad788e.png)

- The last **Sales Invoice** was generated for July.
- "Current Invoice Start Date" and "Current Invoice End Date" in the **Subscription** are set to September.
- Payment Entries are not entered into ERPNext, so Sales Invoices will remain "Overdue" and Subscription will remain "Unpaid"

### Problem

When I click the "Fetch Subscription Updates" button, nothing happens. I would expect that it generates the missing **Sales Invoice** for August. Or the **Sales Invoice** for August should have been generated automatically in the first place.

#### Observations

- "Current Invoice Start Date" and "Current Invoice End Date" are set regardless of the _actual_ invoices generated. I would expect that this is set to the last invoice generated or to the period after that.
    https://github.com/frappe/erpnext/blob/73666982c7d9cbec4e9706aeb8ab41dc8ad7040c/erpnext/accounts/doctype/subscription/subscription.py#L414-L415

- `Subscription.is_postpaid_to_invoice()` and `Subscription.is_prepaid_to_invoice()` both return `False` for my data, therefore no invoice will be generated.
    https://github.com/frappe/erpnext/blob/73666982c7d9cbec4e9706aeb8ab41dc8ad7040c/erpnext/accounts/doctype/subscription/subscription.py#L331-L343

- When I manually reset "Current Invoice Start Date" and "Current Invoice End Date" to July, ERPNext will correctly create the invoice for August.

### Data to reproduce

<details>
<summary>Subscription</summary>
<p>

```json
{
    "name": "ACC-SUB-2021-00028",
    "owner": "Administrator",
    "creation": "2021-07-14 14:35:14.002988",
    "modified": "2021-09-15 15:17:56.845582",
    "modified_by": "Administrator",
    "idx": 0,
    "docstatus": 0,
    "customer": "CUST-20110135",
    "status": "Unpaid",
[...]

PR #27782, opened 2021-10-04
fix(Subscription): reorder updation of end date

hrwX:
- closes [issue/#27521](https://github.com/frappe/erpnext/issues/27521)
- This issue persists since the current end date is updated before `is_postpaid_to_invoice` and `is_prepaid_to_invoice` is run, and both end up returning `False` as `nowdate` will always be smaller than `current_invoice_end`.


barredterra (approved):
Looks good to me. Please port to `version-13-hotfix` and `develop`.

sagarvora:
Closing this, will merge the auto-backport from #27783

PR #27789, opened 2021-10-05
fix: Only calculate first_respone_time if SLA is set

GangaManoj:
_Issue:_

First responses aren't being sent on Issues if the SLA field is not set. The following error pops up:

Screenshot 2021-10-05 at 10 57 18 AM

_Fix:_

Only calculate `first_respone_time` if SLA is set. `first_response_time` calculation requires SLAs, to fetch the working hours, thereby causing this error when it's not set.


AfshanKhan:

AfshanKhan (approved):

PR #27794, opened 2021-10-05
fix: batch_no not mapped from PR to Stock Entry

noahjacob:
### Issue
- Batch number isn't getting mapped when trying to create a stock entry from Purchase Receipt.

batch_map

### After Fix

batch_map_fix


ankush:

ankush (approved):

PR #27798, opened 2021-10-05
fix: using DN for transfer w/o internal customer

ankush:
DN allows "material transfer" by specifying target warehouse.  https://docs.erpnext.com/docs/v13/user/manual/en/stock/articles/material-transfer-from-delivery-note#enable-target-warehouse

This is mostly meant for internal customer/supplier use case, however since it was allowed before many users are using this for making DN to regular customers also. The use-case is sort of valid hence restoring the functionality unless a better alternative is made available. 

Issue introduced in https://github.com/frappe/erpnext/pull/27086

### DIFF
--- a/erpnext/controllers/selling_controller.py
+++ b/erpnext/controllers/selling_controller.py
@@ -424,7 +424,7 @@ class SellingController(StockController):
 					or (cint(self.is_return) and self.docstatus==2)):
 						sl_entries.append(self.get_sle_for_source_warehouse(d))
 
-				if d.target_warehouse and self.get("is_internal_customer"):
+				if d.target_warehouse:
 					sl_entries.append(self.get_sle_for_target_warehouse(d))
 
 				if d.warehouse and ((not cint(self.is_return) and self.docstatus==2)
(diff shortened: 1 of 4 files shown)


erpnext/controllers/selling_controller.py
```diff
@@ -559,6 +559,12 @@ def validate_target_warehouse(self):
 				frappe.throw(_("Row {0}: Delivery Warehouse ({1}) and Customer Warehouse ({2}) can not be same")
 					.format(d.idx, warehouse, warehouse))
 
+		if not self.get("is_internal_customer") and any(d.get("target_warehouse") for d in items):
+			msg = _("Target Warehouse set for some items but the customer is not an internal customer.")
+			msg += " " + _("This {} will be treated as material transfer.").format(_(self.doctype))
```
marination, line 564: Can this be a toast ? It will be disrupting for those who have such transfers often from UX POV.
```suggestion
			msg = _("Target Warehouse is set for some items but the customer is not an internal customer.")
			msg += " " + _("This {} will be treated as material transfer.").format(_(self.doctype))
```

ankush: fixed via https://github.com/frappe/erpnext/commit/2cfafede4408c943d003a04985cadb869f811c21

PR #27819, opened 2021-10-05
fix: check leave type and make sure there is no duplicatd

zaqoutabed:
when `allocate_earned_leaves` in `scheduler_events` is start
if there is more than on leave type the annual allocation will be zero!

Screenshot from 2021-10-05 15-50-33


erpnext/hr/doctype/leave_policy/leave_policy.py
```diff
@@ -11,8 +11,18 @@
 
 class LeavePolicy(Document):
 	def validate(self):
+		self.validate_leave_type()
+		
 		if self.leave_policy_details:
 			for lp_detail in self.leave_policy_details:
+				if lp_detail.leave_type in leaves:
+					frappe.throw(_("Leave Policy {} is selected in row {}".format(lp_detail.leave_type)))
```
ankush:
```suggestion
					frappe.throw(_("Leave Policy {} is selected in row {}").format(lp_detail.leave_type))
```

Translations require that string formatting is applied after translation function call.

erpnext/hr/doctype/leave_policy/leave_policy.py
```diff
@@ -11,8 +11,18 @@
 
 class LeavePolicy(Document):
 	def validate(self):
+		self.validate_leave_type()
+
```
ankush:
```suggestion

```

erpnext/hr/doctype/leave_policy/leave_policy.py
```diff
@@ -11,8 +11,18 @@
...
+		self.validate_leave_type()
+		
 		if self.leave_policy_details:
 			for lp_detail in self.leave_policy_details:
+				if lp_detail.leave_type in leaves:
+					frappe.throw(_("Leave Policy {} is selected in row {}").format(lp_detail.leave_type))
 				max_leaves_allowed = frappe.db.get_value("Leave Type", lp_detail.leave_type, "max_leaves_allowed")
 				if max_leaves_allowed > 0 and lp_detail.annual_allocation > max_leaves_allowed:
 					frappe.throw(_("Maximum leave allowed in the leave type {0} is {1}").format(lp_detail.leave_type, max_leaves_allowed))
+
```
ankush:
```suggestion

```

PR #27420, opened 2021-09-09
fix: gp incorrect valuation rate si_list missing dn_detail

dj12djdjs:
The gross profit report only reports correct valuation rate if the Delivery Note is created first and Sales Invoice is created second.
This is because on creation of Sales Invoice from Delivery Note: delivery_note, and dn_detail are mapped to the Sales Invoice Item.

If Sales Invoice is created first, or Sales Order then Sales Invoice and Delivery Note are only linked by Sales Order valuation rate uses some last known valuation rate. Resulting in Gross Profit being incorrect.

This pr maps missing dn_detail and delivery_note to si_list where possible.

closes #27400 
closes #27297

Issue #27400, opened 2021-09-08
Gross Profit: Incorrect Valuation Rate Depending on Sales Flow


dj12djdjs:
## Description of the issue
ERPNext selling documentation notes there are 5 ways to process sales.
```
                   / ----- > Delivery Note
1. Sales Order -> |
                   \  ----- > Sales Invoice

2. Sales Order -> Delivery Note -> Sales Invoice
3. Sales Order -> Sales Invoice -> Delivery Note
4. Sales Invoice -> Delivery Note
5. Delivery Note -> Sales Invoice
```

![image](https://user-images.githubusercontent.com/29856401/132581192-b0551393-b186-45ff-838a-88e649575889.png)

If you process sales using methods 2 - 5 gross profit valuation rate is correct. However, using method 1 results in some sort of "last known" valuation rate as shown in issue: #27297 

My findings per sales flow are explained below:

1. Material Receipt: Item B  1000Nos x $0.98
2. ACC-SINV-2021-00007: Valuation rate is correct. (This is the first sale so method doesn't matter.)
3. Material Receipt: Item B  100Nos X $2.50.
4. ACC-SINV-2021-00008: Valuation rate is correct using method 2.
5. Material Receipt: Item B  100Nos X $3.00.
6. ACC-SINV-2021-00009: Valuation rate is incorrect using method 1.
7. ACC-SINV-2021-00010: Valuation rate is correct using method 3.

The problem happens when Sales Invoice and Delivery Note are not directly linked to each other.

![image](https://user-images.githubusercontent.com/29856401/132582553-99c025e8-f295-44e5-9503-a5b1855a2788.png)

## Context information (for bug reports)

**Output of `bench version`**
```
develop
version-13
version-12
```

## Proposed Resolution:

1. Perhaps when sales invoice or delivery note is submitted it will link dn_detail by looking it up through sales order.
2. Modify gross profit report to look for delivery note through sales order during processing.

aakvatech:
Seen this as well on v13, but not sure if it was due to old data of v13 beta or whether it still is a problem.

dj12djdjs:
@aakvatech it's definitely still a problem. I tried it with fresh data on a fresh install.

dj12djdjs:
It seems I may have initially been wrong method 3 and 4 as they don't seem to be working now. I think I've narrowed it down if Sales Invoice Item does not contain delivery_note, or dn_detail field Gross Profit does not calculate properly.

[...]

PR #27827, opened 2021-10-05
test(accounts): Add basic testing for gross profit report.

dj12djdjs:
These tests start out as failing. You can see in this [discussion](https://discuss.erpnext.com/search?q=gross%20profit%20after%3A2020-12-31) query there are lots of people posting about inaccurate gross profit reporting. I have encountered it as well.

These tests are check gross profit after 3 different sales flows illustrated in this issue: #27400
I've tested this PR #27420 resolves these tests.

The test is checked by comparing the total row against the amounts from default accounts: Income, and Cost of Sold Goods.

On a side note, this is the first test case I've ever written so all criticism is very much welcome :)


dj12djdjs:
T

PR #27834, opened 2021-10-06
fix: use ceil in case of whole uoms for reorder qty

18alantom:
### Description
- Reorder qty is calculated from stock projected qty.
- The UOM used here may be different say _Nos_ as compared to Purchase UOM say _Pack of 10_ (conversion factor will be 10).
- So if shortage is of 7 (_Nos_) this will cause a reorder of 0.7 (_Pack of 10_).
- This pr will cause `ceil(0.7)` i.e. `1` if the Purchase UOM should be a whole number.

(Internal ref: ISS-21-22-07039)

### DIFF
--- a/erpnext/stock/reorder_item.py
+++ b/erpnext/stock/reorder_item.py
@@ -150,7 +150,7 @@ def create_material_request(material_requests):
 							conversion_factor = frappe.db.get_value("UOM Conversion Detail",
 								{'parent': item.name, 'uom': uom}, 'conversion_factor') or 1.0
 
-					must_be_whole_number = frappe.get_value("UOM", uom, "must_be_whole_number")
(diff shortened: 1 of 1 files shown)


18alantom:

erpnext/stock/reorder_item.py
```diff
@@ -149,11 +150,16 @@ def _log_exception():
 							conversion_factor = frappe.db.get_value("UOM Conversion Detail",
 								{'parent': item.name, 'uom': uom}, 'conversion_factor') or 1.0
 
+					must_be_whole_number = frappe.get_value("UOM", uom, "must_be_whole_number")
```
ankush:
```suggestion
					must_be_whole_number = frappe.db.get_value("UOM", uom, "must_be_whole_number", cache=True)
```

This will probably get called 100s of time in same DB transaction. Better to cache the results.

ankush (approved):

PR #27840, opened 2021-10-06
fix: add cost center in gl entry for advance payment entry

AfshanKhan:
Problem:

When allocating advance payment in sales invoice and submitting it creates GL without cost center mentioned in sales invoice.
Because of this when checking Accounts Receivable Reports with a filter of cost center gives the paid sales invoice showing outstanding balance.

Steps to replicate:
1. Make a payment entry for a customer.
2. Now make a Sales Invoice with a cost center for that customer and allocate advance as the payment that was created.
3. After submitting Sales Invoice check Accounts Receivable Reports and add that cost center in filters.
4. The Sales invoice even thought it is paid is shown as one containing outstanding balance.


AfshanKhan:

nextchamp-saqib (approved):

PR #27884, opened 2021-10-10
fix: patch fails if accounts are frozen

nextchamp-saqib:


deepeshgarg007:

Issue #27842, opened 2021-10-06
search_link not working on finished item input in new BOM


reubenromandy:
## Description of the issue
Typing part numbers into the finished item input on the new BOM page doesn't deliver filtered results. Just keeps sending the top 20 without regard to input text. Seems to have happened after updating to Frappe 13.12
## Context information (for bug reports)

**Output of `bench version`**
```
erpnext 13.12.0
frappe 13.12.0
```

## Steps to reproduce the issue

1. open New BOM
2. start typing an existing part number into the item input at the top of the screen

### Observed result
no filtering of list occurs

### Expected result
see filtered items based on input
### Stacktrace / full error message

```
There is no error from error log or from browser console.
```

## Additional information
This input is the only one I've seen affected by this issue.

OS version / distribution, `ERPNext` install method, etc.

Description:    Debian GNU/Linux 10 (buster)
erpnext installed through install script via wget

szufisher:
Mixed both input txt and is stock item, has_variants into or_condtion is the root cause, the relevant code need to be changed as below( for reference only)

```
@frappe.whitelist()
@frappe.validate_and_sanitize_search_inputs
def item_query(doctype, txt, searchfield, start, page_len, filters):
	meta = frappe.get_meta("Item", cached=True)
	searchfields = meta.get_search_fields()

	order_by = "idx desc, name, item_name"

	fields = ["name", "item_group", "item_name", "description"]
	fields.extend([field for field in searchfields
		if not field in ["name", "item_group", "description"]])

	searchfields = searchfields + [field for field in [searchfield or "name", "item_code", "item_group", "item_name"]
		if not field in searchfields]

	query_filters = {
		"disabled": 0,
		"ifnull(end_of_life, '5050-50-50')": (">", today())
	}

	or_cond_filters = {}
	if txt:
		for s_field in searchfields:
			or_cond_filters[s_field] = ("like", "%{0}%".format(txt))

		barcodes = frappe.get_all("Item Barcode",
			fields=["distinct parent as item_code"],
			filters = {"barcode": ("like", "%{0}%".format(txt))})

		barcodes = [d.item_code for d in barcodes]
		if barcodes:
			or_cond_filters["name"] = ("in", barcodes)

		if filters and filters.get("is_stock_item"):
			fields.extend(['is_stock_item','has_variants'])
	else:
		or_cond_filters["is_stock_item"] = 1
		or_cond_filters["has_variants"] = 1

	if filters and filters.get("item_code"):
[...]

PR #27890, opened 2021-10-11
fix: bom item query

ankush:
- fix bom item query https://github.com/frappe/erpnext/issues/27842 
- basic test for item query

### DIFF
--- a/erpnext/manufacturing/doctype/bom/test_bom.py
+++ b/erpnext/manufacturing/doctype/bom/test_bom.py
@@ -382,8 +382,8 @@ class TestBOM(unittest.TestCase):
 		test_items = query(txt="_Test")
 		filtered = query(txt="_Test Item 2")
 
-		self.assertNotEqual(len(test_items), len(filtered), msg="Item filtering showing excessive reulsts")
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/bom/test_bom.py
```diff
@@ -375,6 +376,16 @@ def test_bom_with_process_loss_item(self):
...
 		self.assertRaises(frappe.ValidationError, bom_doc.submit)
 
+	def test_bom_item_query(self):
+		query = partial(item_query, doctype="Item", txt="", searchfield="name", start=0, page_len=20, filters={"is_stock_item": 1})
+
+		test_items = query(txt="_Test")
+		filtered = query(txt="_Test Item 2")
+
+		self.assertNotEqual(len(test_items), len(filtered), msg="Item filtering showing excessive reulsts")
+		self.assertTrue(0 < len(filtered) <= 3, msg="Item filtering showing excessive reulsts")
```
ankush:
```suggestion
		self.assertNotEqual(len(test_items), len(filtered), msg="Item filtering showing excessive results")
		self.assertTrue(0 < len(filtered) <= 3, msg="Item filtering showing excessive results")
```

PR #27904, opened 2021-10-11
fix: map missing fields in opportunity

pateljannat:
**Issue:**

1. When we create an Opportunity from Lead manually, the Opportunity picks the value of similar fields from Lead doctype.
2. The same does not happen when Opportunity is created from Lead using data import.

**Fix:**
1. Added logic on the server-side, which checks if a field in opportunity does not have a value and this field is present in Lead/Customer doctype then fetch that value and add it to the field in Opportunity.


erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -43,6 +44,15 @@ def validate(self):
 		else:
 			self.calculate_totals()
 
+	def map_fields(self):
+		for field in self.meta.fields:
+			if not self.get(field.fieldname):
+				try:
+					value = frappe.db.get_value(self.opportunity_from, self.party_name, field.fieldname)
+					frappe.db.set(self, field.fieldname, value)
```
ruchamahabal, line 52: Since this is called on validate, it will save the doc. Do we need `db.set` here?

pateljannat: So after getting the value from lead/customer it should be set in opportunity.

The fieldname is in field.fieldname. Doing `self[field.fieldname] = value` is not working. So I had to go with db.set

ruchamahabal (approved):

pateljannat:

pateljannat:

PR #27914, opened 2021-10-12
fix: Avoid automatic customer creation on website user login

marination:
**Issue:**
> TLDR: When a website user logs in, if they are not a customer and if cart is enabled, a customer was automatically created
- This user could have been created for blogs, or for other purposes, and not e-commerce.
- **This happens because:** on login, hooks triggers an event that sets the cart count. For this it tries to fetch the current users shopping cart (cart quotation). If it cannot find a quotation it creates a new one against this user. While doing this, it creates a customer if non-existent.
- If a quotation is found then its items count becomes the cart count, else an empty quotation(cart) is initialised.
- If a logged in user has ever added anything to cart, they are made a customer in the backend. So if this user is not a customer, we don't need to refresh the cart at all, it will be empty or non-existent.

**Fix:**
- Setting of cart count (refreshing the cart count) must happen only if the user logging in is a valid customer in the system. Otherwise the user has no business with the cart.
- Changed `check_customer_or_supplier` to `is_customer`. If supplier we were exiting the function anyway so why return it.

**To test:**
- Create a new website user (do not use this user for e-commerce)
- login as this user
- Check if customer was auto created after login

### DIFF
--- a/erpnext/shopping_cart/utils.py
+++ b/erpnext/shopping_cart/utils.py
@@ -1,8 +1,5 @@
 # Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 # License: GNU General Public License v3. See license.txt
-
-from __future__ import unicode_literals
-
 import frappe
 
 from erpnext.shopping_cart.doctype.shopping_cart_settings.shopping_cart_settings import (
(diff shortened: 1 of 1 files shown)


marination:
Not sure what test to write here, it just refreshes cart count. test case is too specific. 

Failing UI tests are unrelated

marination:

rasos:
We need to test the B2C scenario: a new user registers as s/he wants to buy an item with the cart. So that user is immediately a customer.

PR #27916, opened 2021-10-12
perf: improve sales invoice submission

nextchamp-saqib:
- skip `get_pricing_rules` if no pricing rule exists
  - `get_pricing_rules` is called for each item in the invoice, if the pricing rules doesn't exist then no point in calling the method
- fetch mode of payments data in a single query
- get total company stock only for purchase order
  - `company_total_stock` is only used in the purchase order item
- skip insertion of stock ledger entry

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -1997,8 +1997,7 @@ def get_mode_of_payments_info(mode_of_payments, company):
 			mp.name in (%s)
 		group by
 			mp.name
-		""",
-	(company, mode_of_payments), as_dict=1)
+		""", (company, mode_of_payments), as_dict=1)
 
 	return {row.get('mop'): row for row in data}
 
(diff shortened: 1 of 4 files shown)


erpnext/stock/stock_ledger.py
```diff
@@ -240,8 +239,7 @@ def __init__(self, args, allow_zero_rate=False, allow_negative_stock=None, via_l
 		self.verbose = verbose
 		self.allow_zero_rate = allow_zero_rate
 		self.via_landed_cost_voucher = via_landed_cost_voucher
-		self.allow_negative_stock = allow_negative_stock \
-			or cint(frappe.db.get_single_value("Stock Settings", "allow_negative_stock"))
+		self.allow_negative_stock = True
```
```suggestion
		self.allow_negative_stock = allow_negative_stock \
			or cint(frappe.db.get_single_value("Stock Settings", "allow_negative_stock"))
```
Intentional?

erpnext/accounts/doctype/sales_invoice/sales_invoice.py
```diff
@@ -1982,6 +1983,25 @@ def get_all_mode_of_payments(doc):
...
+			`tabMode of Payment Account` mpa,`tabMode of Payment` mp
+		where
+			mpa.parent = mp.name and
+			mpa.company = %s and
+			mp.enabled = 1 and
+			mp.name in (%s)
+		group by
+			mp.name
+		""",
+	(company, mode_of_payments), as_dict=1)
```
```suggestion
		""", (company, mode_of_payments), as_dict=1)
```

nextchamp-saqib:

PR #27921, opened 2021-10-12
fix: Multiple fixes for Partly Paid status in Sales Invoice and Purchase Invoice

pruthvi145:
## Issues
- When `base_grand_total` and `grand_total` are not same, status is being wrongly set.

## Changes made
 - Compare `outstanding_amount` with `base_grand_total` instead of `grand_total` since it will always be in base currency.
 - Improve `is_overdue` and `update_invoice_status`.
 - Consider `rounded_total` when it is not disabled.


sagarvora:
Reopened here: https://github.com/frappe/erpnext/pull/28019

PR #27924, opened 2021-10-12
fix: Item Variant selection empty popup on website

marination:
**Issue:**
- Screenshot 2021-10-12 at 7 59 26 PM
- This was tested earlier with items that happened to have the same name as the website item
- Website Item naming is now changed to naming series based. So this was missed.

**Fix:**
- Store `item_code` instead of website item name in `Select Variant` buttons data attribute
- It is used in `get_attributes_and_values` which was taking the website item name earlier and giving back a blank result.
Screenshot 2021-10-12 at 7 58 44 PM

### DIFF
--- a/erpnext/templates/generators/item/item_configure.html
+++ b/erpnext/templates/generators/item/item_configure.html
@@ -4,8 +4,8 @@
 <div class="mt-5 mb-6">
 	{% if cart_settings.enable_variants | int %}
 	<button class="btn btn-primary-light btn-configure font-md mr-2"
-		data-item-code="{{ doc.name }}"
-		data-item-name="{{ doc.item_name }}"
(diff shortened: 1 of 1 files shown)


nostahl:
ran into this issue as well after updating v13. after this hot fix applied the modal shows the variants now but clicking on the fields is empty.

nostahl:
@marination is there an additional component to this hot fix that populates the variant selector field?

marination:
@nostahl The variant items need to be published Website Items

nostahl:
I notice in e commerce settings If I select "hide variants" and save it unchecked that box after save... i was able to get the template product to work now with the variants, I don't want all the variant products showing as individual products on the site etc.

nostahl:
ahh all seems to be working with 13.13.0

PR #27934, opened 2021-10-13
fix: TDS round off not working from second transaction

deepeshgarg007:
Tax Withholding Round Off amount check was done only in the code block where the TDS for the very first transaction was calculated.

So for the subsequent transactions let's say second and so on the TDS amount was not getting rounded to the nearest integer

Moved the check to the code block where the final tax amount is returned so that the tax amount is rounded off for all types (TDS  and TCS both) and all transactions


deepeshgarg007:

deepeshgarg007:

PR #27937, opened 2021-10-13
fix: not authorized to update entries after freezing accounts

noahjacob:
### Issue
- Freezing date is set in the **"Accounts Settings"**.

image

- When trying to make any entry which needs to update the GL, it throws the following error.
- For Example, while trying to submit a Delivery Note after the freezing date, the error is still thrown.

image

- Reposts fail even if posting date is greater than the freezing date.

image

### Fix
- The condition that throws the error passes when the user is "Administrator" which is always the case when reposting or while updating entries. 
- Fixed the condition.

### DIFF
--- a/erpnext/patches/v13_0/requeue_failed_reposts.py
+++ b/erpnext/patches/v13_0/requeue_failed_reposts.py
@@ -1,0 +1,14 @@
+import frappe
+from frappe.utils import cstr
+
+def execute():
+
+	reposts = frappe.get_all("Repost Item Valuation",
+		{"status": "Failed", "modified": [">", "2021-10-6"] },
+		["name", "modified", "error_log"])
+
+	for repost in reposts:
+		if "check_freezing_date" in cstr(repost.error_log):
+			frappe.db.set_value("Repost Item Valuation", repost.name, "status", "Queued")
+
+	frappe.db.commit()
--- a/erpnext/patches/v13_0/requeue_failed_reposts.py
+++ b/erpnext/patches/v13_0/requeue_failed_reposts.py
@@ -1,14 +1,13 @@
 import frappe
 from frappe.utils import cstr
 
+
 def execute():
 
 	reposts = frappe.get_all("Repost Item Valuation",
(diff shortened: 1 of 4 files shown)


noahjacob:
Need to write a patch to re-queue failed reposts.

erpnext/patches/v13_0/requeue_failed_reposts.py
```diff
@@ -0,0 +1,14 @@
...
+
+	reposts = frappe.get_all("Repost Item Valuation",
+		{"status": "Failed", "modified": [">", "2021-10-6"] },
+		["name", "modified", "error_log"])
+
+	for repost in reposts:
+		if "check_freezing_date" in cstr(repost.error_log):
+			frappe.db.set_value("Repost Item Valuation", repost.name, "status", "Queued")
+
+	frappe.db.commit()
```
ankush:
```suggestion
```

Unnecessary: https://frappeframework.com/docs/user/en/api/database#patches

erpnext/patches/v13_0/requeue_failed_reposts.py
```diff
@@ -0,0 +1,14 @@
+import frappe
+from frappe.utils import cstr
+
```
ankush, line 3:
```suggestion


```

import convention - 2 blank lines. Automate this: https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

erpnext/patches/v13_0/requeue_failed_reposts.py
```diff
@@ -0,0 +1,14 @@
...
+def execute():
+
+	reposts = frappe.get_all("Repost Item Valuation",
+		{"status": "Failed", "modified": [">", "2021-10-6"] },
```
ankush:
```suggestion
		{"status": "Failed", "modified": [">", "2021-10-05"] },
```

dates should be zero-padded ideally so string comparison can also work on them.

ankush (changes requested):

ankush (approved):

PR #27939, opened 2021-10-13
test: remove commits from tests

ankush:
Removing transaction commits from all tests. Continuation of https://github.com/frappe/erpnext/pull/27496 

- [x] stock
- [x] selling
- [x] Manufacturing
- [ ] accounts
- [x] Buying

PS: prioritizing modules that affect stock first, but eventually will cover the full test suite.


ankush:
160 failing 🥲

ankush:
failing test fixed separately.

PR #27945, opened 2021-10-13
fix: use batch, serial in future sle check for neg qty

18alantom:
### Description
1. Uses batch and serial numbers filters to check future Stock Ledger Entries for negative values.
2. Checks future SLE for negative values for matching warehouse and batch.

_**Note**: FP of this [PR](https://github.com/frappe/erpnext/pull/25675) which was closed because changing base was more trouble than just cherry-picking onto develop._


18alantom:
Linter failing cause:
```bash
Trim Trailing Whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1

  File "/home/runner/.cache/pre-commit/repoqm7iopci/py_env-python3.8/bin/trailing-whitespace-fixer", line 8, in <module>
    sys.exit(main())
  File "/home/runner/.cache/pre-commit/repoqm7iopci/py_env-python3.8/lib/python3.8/site-packages/pre_commit_hooks/trailing_whitespace_fixer.py", line 95, in main
    if _fix_file(filename, md, chars):
  File "/home/runner/.cache/pre-commit/repoqm7iopci/py_env-python3.8/lib/python3.8/site-packages/pre_commit_hooks/trailing_whitespace_fixer.py", line 16, in _fix_file
    with open(filename, mode='wb') as file_processed:
PermissionError: [Errno 13] Permission denied: 'erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py'
```

erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -870,6 +870,50 @@ def test_additional_cost_distribution_non_manufacture(self):
...
+			dict(item_code=item_code,
+				qty=2,
+				from_warehouse=warehouse_names[0],
+				to_warehouse=warehouse_names[1],
+				batch_no=batch_no,
+				posting_date='2021-07-02',          # Illegal SE
+				purpose='Material Transfer')
+		]
+		illegal_sequence_executor = create_stock_entry_sequence_executor(illegal_sequence)
+		self.assertRaises(frappe.ValidationError, illegal_sequence_executor)
```
ankush:
```suggestion
		self.assertRaises(frappe.ValidationError, create_stock_entries, sequence_of_entries)
```

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises you can pass args/kwargs directly. Easier to read this way.

erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -940,3 +984,30 @@ def get_multiple_items():
...
+		item_code=item_code,
+		warehouse=warehouse_names[0],
+		valuation_rate=100,
+		qty=opening_qty,
+		batch_no=batch_no,
+	)
+	return warehouse_names
+
+def create_stock_entry_sequence_executor(sequence_of_entries):
+	return lambda: create_stock_entries(sequence_of_entries)
```
ankush:
```suggestion
```

erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -172,6 +172,40 @@ def validate_with_last_transaction_posting_time(self):
 					msg += "<br>" + "<br>".join(authorized_users)
 					frappe.throw(msg, BackDatedStockTransaction, title=_("Backdated Stock Entry"))
 
+	def validate_future_sle_by_batch_for_outgoing_entries(self):
+		allow_neg = frappe.db.get_single_value('Stock Settings', 'allow_negative_stock')
+		if not self.batch_no or allow_neg or self.actual_qty is None or self.actual_qty >= 0:
+			return
```
ankush:
```suggestion
		if not self.batch_no or self.actual_qty is None or self.actual_qty >= 0:
			return
		if frappe.db.get_single_value('Stock Settings', 'allow_negative_stock'):
			return
```

Reducing unnecessary query.

ankush (dismissed):

erpnext/stock/stock_ledger.py
```diff
@@ -1084,6 +1085,8 @@ def get_future_sle_with_negative_qty(args):
 			item_code = %(item_code)s
 			and warehouse = %(warehouse)s
 			and voucher_no != %(voucher_no)s
+			and ifnull(batch_no,'') = %(batch_no)s
+			and ifnull(serial_no,'') = %(serial_no)s
```
ankush: This has to be `like` comparison?

ankush:
I am undoing serial no check, which needs to be improved. For now, let's just merge the batch-wise check since it helps in case of backdated transactions that are incorrectly ordered backdated transactions?

ankush:
I will do some manual testing, possible cases:

- [x] 1. Remove more qty than exist in batch (works before PR too)
- [x] 2. Insert a transaction that makes intermediate qty go negative for a batch but overall stays positive. 
- [ ] 3. insert a transaction that makes intermediate qty for a batch ONLY go negative, another batch exists that makes overall quantity positive at the same time. (**FAILED**)



Alse note:
1. `actual_qty_after_transaction` alone isn't enough for mulit-batch items probably.

erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.py
```diff
@@ -170,6 +170,42 @@ def validate_with_last_transaction_posting_time(self):
...
+			voucher_condition = f"and not (voucher_no = '{voucher_no}' and voucher_type = '{voucher_type}')"
+
+		future_negative_entries = frappe.db.sql("""select
+					voucher_type, voucher_no, item_code, warehouse, posting_date
+				from `tabStock Ledger Entry`
+				where
+					warehouse = %(warehouse)s
+					and item_code = %(item_code)s
+					and batch_no = %(batch_no)s
+					and (qty_after_transaction + %(actual_qty)s) < 0
```
ankush, line 193: `qty_after_transaction` represents total quantity, so this isn't correct for multi-batch case.


while `actual_amt_check` check on stock ledger entry ensures current balance of `actual_qty` is not negative, it actually needs to ensure that the **cumulative sum** of `actual_qty` never goes negative.

ankush: IMO let's just get rid of this whole function and expand batch-wise check to consider running totals instead of final sum.

ankush:
continued in https://github.com/frappe/erpnext/pull/28782

PR #27953, opened 2021-10-13

latoile:
image

Add scripts and reference fields:
image

image


erpnext/regional/switzerland/setup.py
```diff
@@ -0,0 +1,31 @@
...
+            'Sales Invoice': [
+                dict(fieldname='reference_number', label='Reference Number',
+                    fieldtype='Data', insert_after='naming_series', length='25', unique='1', translatable='0'),
+                dict(fieldname='reference_number_full', label='Reference Number Full',
+                    fieldtype='Data', insert_after='reference_number', length='25', unique='1', translatable='0', hidden='1')
+            ]
+    }
+    create_custom_fields(custom_fields)
+
+def make_client_script():
```
deepeshgarg007, line 24: @latoile you do not need to make a custom client script like this. Any regional client-side code can be written in the `../doctype/sales_invoce/regional/switzerland.js`

You can refer this file as an example https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/doctype/sales_invoice/regional/india.js

deepeshgarg007 (changes requested):

deepeshgarg007:
Also add documentation for the changes made

PR #27954, opened 2021-10-14
fix: value_after_depreciation calculation

### LINKED ISSUE #27431: Creating existing, partially depreciated assets with correct depreciation is impossible

nextchamp-saqib:
Fixes #27431


erpnext/assets/doctype/asset/test_asset.py
```diff
@@ -682,6 +682,27 @@ def test_discounted_wdv_depreciation_rate_for_indian_region(self):
 		# reset indian company
 		frappe.flags.company = company_flag
 
+	def test_asset_1(self):
```
nextchamp-saqib:
```suggestion
	def test_expected_value_change(self):
```

nextchamp-saqib:

PR #27957, opened 2021-10-14
fix: patch to enable scheduled job for reposting

rohitwaghchaure:


erpnext/patches/v13_0/enable_scheduler_job_for_item_reposting.py
```diff
@@ -0,0 +1,7 @@
+import frappe
+
```
ankush, line 2:
```suggestion


```

ankush (dismissed):

PR #27967, opened 2021-10-15
fix: Account number and name incorrectly imported using COA importer

deepeshgarg007:
Account name and number were not correctly updated in the Account master when accounts were imported using the chart of accounts importer tool

For Eg: If an `account name`  is `Test Account` and `account number` is `1234` the account numbers is imported as blank and the `account name` is imported as `1234 - Test Account`

This PR fixes that issue

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
@@ -12,7 +12,7 @@ from six import iteritems
 from unidecode import unidecode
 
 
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

deepeshgarg007:

Ponyooooo:
have installed a new erpnext with the v13.13.0, and the chart of accounts imported did not recognize or maybe made all accounts as "is group" is 1. So all accounts imported are in group.

Olawale1:
Hi @deepeshgarg007 

The issue reported by @Ponyooooo is a serious one and we just encountered same now. All our ledgers are now imported as Group which means we can't post anything into them! Please help fix this urgently!

Thanks

PR #27990, opened 2021-10-18
fix: changes in schedules gets overwritten on save

noahjacob:
### Issue
- Any changes made to the schedule table would get overwritten since generate schedule is called on save.

main_bug

### Fix
- While validating, only call generate schedule if schedule table is empty or of no of visits are changed.


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -199,12 +199,24 @@ def validate_sales_order(self):
...
+	def validate_no_of_visits(self):
+		total_visits = 0
+		visits_changed = False
+		for v in self.items:
+			total_visits += v.no_of_visits
+
+		if total_visits != len(self.schedules):
+			visits_changed = True
+
+		return visits_changed
```
```suggestion
		return len(self.schedules) != sum(d.no_of_visits for d in self.items)
```

noahjacob:

noahjacob:

PR #27625, opened 2021-09-21
feat: add `Partly Paid` status in Invoices

sagarvora:
- Resolves https://github.com/frappe/erpnext/issues/26969
- Also uses `getdate()` instead of `CURDATE()` to get accurate dates based on timezones set in System Settings.

## Behaviour
If a part of the invoice amount has been received, `Partly Paid` state will be shown (instead of `Unpaid`). If any scheduled payment (as specified in the Payment Schedule table) is pending beyond it's due date, the `Overdue` state will be shown.

## Documentation
https://docs.erpnext.com/docs/v13/user/manual/en/accounts/purchase-invoice/edit?wiki_page_patch=a8b85aba51
https://docs.erpnext.com/docs/v13/user/manual/en/accounts/sales-invoice/edit?wiki_page_patch=709cb5df15

Reopened from #27310

PR #27992, opened 2021-10-18
fix: incorrect sales invoice status set if rounding adjustment is set

nextchamp-saqib:
Introduced with #27625

Case:
`due_date = any_future_date`

`grand_total = 499.5`
`rounding_adjustment = 0.5`
`rounded_total = 500`
`oustanding_amount = 500`

with the above values, the status of the invoice is set as `Overdue` because,
`is_overdue` returns `True` since `grand_total - outstanding_amount` is negative


nextchamp-saqib:
Already being worked on with https://github.com/frappe/erpnext/pull/27921

PR #27997, opened 2021-10-19
fix: Employee filter in YTD and MTD in salary slip

govindsmenokee:
The filter for YTD, MTD etc are based on employee name. This seems like an amateur mistake. It should be based on employee id.


pateljannat (approved):

pateljannat:
@Mergifyio backport develop

PR #28005, opened 2021-10-19
fix: Fetch thumbnail from Item master instead of regenerating

marination:
**Issue:**
- Due to thumbnail generation in patch, patch slowed down a lot. File is created , etc. which slows things down

**Fix:**
- Separate patch to update thumbnail field of website item from item master
- In `create_website_items`, only make web items, nothing else. If web item exists, skip the item.
- Added a progress bar to `create_website_items`


marination:

PR #28006, opened 2021-10-19
fix: POS consolidated invoice rounded total issue

nemesis189:
Ref issue: ISS-21-22-06296

**Problem:**
Consider the following 2 POS Invoices:
 <details>
<summary><b>invoice 1:</b></summary>

</details>

<details>
<summary><b>Invoice 2:</b></summary>

</details>

<details>
<summary><b>Consolidated Invoice of the above two:</b></summary>

</details>

- The total gets rounded in both the invoices and _0.50_ is deducted from the total of both invoices
- So the effect totals and the amount paid in both the invoices now are:  ***250 and 50***
- On consolidating these 2 invoices, the total becomes:
   ***250.5 + 50.5 = 301***
- The effective amount actually paid is ***250+50 = 300***, since the totals of individual invoices were rounded off
- In the consolidated invoice, the paid amount is calculated by adding paid amounts from individual pos invoices
- The  amount to be paid in the consolidated sales invoice gets recalculated as **301** after merging the two POS invoices
- Since only 300 was paid and the sales invoice has 301 as the amount to be paid, the consolidated invoice would remain in unpaid state and there would be an outstanding amount of ***Re. 1***

**Solution:**
<details>
<summary><b>Fixed Consolidated Invoice of the above two:</b></summary>

</details>

- Adding the `Rounding Adjustment` and `Rounded Total` of the individual POS invoices while merging to calculate the same for Consolidated Invoice
- In `calculate_taxes` method, rounding adjustment was being overriden so bypassed that for consolidated invoices
- In `calculate_totals` method, total_taxes_and_charges was being calculated even when there were no taxes applied, so added a condition for the same

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.py
+++ b/erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.py
@@ -114,6 +114,8 @@ class POSInvoiceMergeLog(Document):
 	def merge_pos_invoice_into(self, invoice, data):
 		items, payments, taxes = [], [], []
 		loyalty_amount_sum, loyalty_points_sum = 0, 0
+		rounding_adjustment, base_rounding_adjustment = 0, 0
+		rounded_total, base_rounded_total = 0, 0
 		for doc in data:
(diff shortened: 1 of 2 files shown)


nextchamp-saqib:

PR #28011, opened 2021-10-19
feat(e-commerce): publish bulk items as website items

dj12djdjs:
item-bulk-create-website-item


erpnext/stock/doctype/item/item_list.js
```diff
@@ -14,6 +14,31 @@ frappe.listview_settings['Item'] = {
 			return [__("Variant"), "green", "variant_of,=," + doc.variant_of];
 		}
 	},
+	onload: function(me) {
+		me.page.add_action_item(__("Create Website Item(s)"), function() {
```
marination: Would this be better? Because there's a button inside Item too named **Publish In Website**. It will be familiar and consistent imo
```suggestion
		me.page.add_action_item(__("Publish in Website"), function() {
```

erpnext/e_commerce/doctype/website_item/website_item.py
```diff
@@ -533,4 +533,16 @@ def make_website_item(doc, save=True):
...
+@frappe.whitelist()
+def make_bulk_website_items(items):
+	if isinstance(items, str):
+		items = json.loads(items)
+
+	website_items = []
+	for item in items:
+		doc = frappe.get_cached_doc("Item", item)
+		website_item_details = make_website_item(doc)
+		website_items.append(website_item_details)
```
marination: Could we add a progress bar here too ? In most cases people will publish a larger amount of items. It'd be nice to get feedback on the progress.

erpnext/stock/doctype/item/item_list.js
```diff
@@ -14,6 +14,31 @@ frappe.listview_settings['Item'] = {
...
+							message: __("Website Item {0} has been created.",
+								[repl('<a href="/app/website-item/%(item_encoded)s" class="strong">%(item)s</a>', {
+									item_encoded: encodeURIComponent(result[0]),
+									item: result[1]
+								})]
+							),
+							title: __("Published"),
+							indicator: "green"
+						});
+					});
```
marination: Again if there are around 50 items to publish, the msgprint does not serve the best UX (too many entries).

Here's a suggestion:
- If a progress bar is added that is enough feedback on the progress (as mentioned in the previous review).
- **In this callback**, the user can now just get a toast (`show_alert`) saying "50 Items were successfully published".
- If there are errors those can come as a collective msgprint in the end.

marination:
Also @dj12djdjs thanks for pushing this 😄

erpnext/stock/doctype/item/item_list.js
dj12djdjs: What are your thoughts to queue more then 50 items into a background job? If the user has thousands of items they are publishing for the first time. 50 items per batch is going to take a while.

marination: @dj12djdjs I'd say if its more than 50 items we can push it to the background. until 50 the progress bar should help. More than that background will be good.

The background job implementation can be similar item variants creation. You can check where and how `enqueue_multiple_variant_creation` is used

PR #28019, opened 2021-10-20
fix: incorrect status being set in Invoices

sagarvora:
Fixes issues described in #27921, #27992, #28023

Failing test due to newly introduced QB check. Focus currently is on getting this fix out, so skipping QB for now.

PR #28023, opened 2021-10-20
fix: setting proper status in Sales and Purchase Invoices

AfshanKhan:
### Problem:
If the Sales or Purchase Invoice have `Rounding Adjustment`, wrong value of status is getting set to the invoice.

### Steps to replicate:
**Case 1:**
1. Create a Sales or Purchase Invoice with posting date as today and make sure Rounding Adjustment have some value and it is in positive.
2. Add a payment term which sets due date to be after 30 days of posting date.
3. Submitting the invoice should set status as `Unpaid` but the status is being set as `Overdue`.

**Case 2:**
1. Create a Sales or Purchase Invoice with posting date as today and make sure Rounding Adjustment have some value and it is in negative.
2. Add a payment term which sets due date to be after 30 days of posting date.
3. Submitting the invoice should set status as `Unpaid` but the status is being set as `Partially Paid`.

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
@@ -1192,7 +1192,7 @@ class PurchaseInvoice(BuyingController):
 					self.status = 'Internal Transfer'
 				elif is_overdue(self):
 					self.status = "Overdue"
-				elif 0 < outstanding_amount < flt(self.grand_total, self.precision("grand_total")):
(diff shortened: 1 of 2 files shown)


sagarvora:
🤦🏼 
https://github.com/frappe/erpnext/pull/28019
https://github.com/frappe/erpnext/pull/27921
https://github.com/frappe/erpnext/pull/27992

sagarvora:
Closing. Please review mine: https://github.com/frappe/erpnext/pull/28019

PR #28025, opened 2021-10-20
fix: avoid resetting employee on amending timesheets

ruchamahabal:
**Before Fix**:

If a timesheet with employee field set is amended, it resets employee details and since it is a non-mandatory field user can miss this, save and submit the timesheet without setting the employee again. This is because the timesheet tries to set the employee from session user on load.

timesheet-employee-reset

**After Fix:**

Don't reset employee details if already set

employee-reset-fix


ruchamahabal:

PR #28042, opened 2021-10-20
fix: POS Closing Entry without linked invoices

nemesis189:
**Problem:**
Creating a POS Closing Entry without any invoices was not allowed earlier

**Solution:**
Now it is allowed.

**Steps:**
1. Create an opening entry through POS or manually
2. Create a closing entry against the same
3. Submit and see if it allows you to do so


nextchamp-saqib:

PR #28043, opened 2021-10-20
refactor: move process loss report to manufacturing

18alantom:
### Description
Moved _Process Loss Report_ to _Manufacturing_

> "Why was it in stock in the first place?"
> ¯\\\_(ツ)\_/¯


ankush (approved):

ankush:
Ignoring raw query. This is an existing query that's just moved.

PR #28050, opened 2021-10-21
feat: Competitor Tagging in Opportunity and Quotation

anupamvs:
test
docs - https://docs.erpnext.com/docs/v13/user/manual/en/CRM/opportunity#26-capture-the-reasons-and-competitors-for-lost-opportunities


erpnext/crm/doctype/opportunity/opportunity.js
```diff
@@ -60,6 +59,53 @@ frappe.ui.form.on("Opportunity", {
...
+					"fieldname": "lost_reasons",
+					"options": "Opportunity Lost Reason Detail",
+					"reqd": 1
+				},
+				{
+					"fieldtype": "Table MultiSelect",
+					"label": __("Competitors"),
+					"fieldname": "competitors",
+					"options": "Competitor Detail",
+					"reqd": 1
```
ruchamahabal: This need not be mandatory?

erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -79,7 +79,7 @@ erpnext.selling.QuotationController = class QuotationController extends erpnext.
 
 			if(doc.status!=="Ordered") {
 				this.frm.add_custom_button(__('Set as Lost'), () => {
-						this.frm.trigger('set_as_lost_dialog');
```
ruchamahabal: We can let it be in `sales_common.js` and add fields in the dialog based on the doctype instead of writing the function twice. Any reason to move?

ruchamahabal (changes requested):

erpnext/crm/doctype/opportunity/opportunity.json
```diff
@@ -486,15 +489,33 @@
    "label": "Grand Total",
    "options": "currency",
    "read_only": 1
+  },
+  {
+   "fieldname": "lost_detail_section",
+   "fieldtype": "Section Break",
+   "label": "Lost Detail"
```
ruchamahabal:
```suggestion
   "label": "Lost Reasons"
```

ruchamahabal (approved):

PR #28055, opened 2021-10-21
fix(UX): Add mandatory depends on condition for reference no and date fields

deepeshgarg007:
In payment entry if any of the accounts either Paid From or to is a bank account then user used to get a server side validation for missing reference no and date

In this case neither mandatory indicator shown on these fields nor the user was scrolled to the missing fields.
Added `mandatory_depends_on_condition` so that both of these things happen for better UX


deepeshgarg007:

PR #28071, opened 2021-10-23
fix : Get LTDS based on tax withholding category

deepeshgarg007:
Replace section code in Lower Deduction certificate with Tax Withholding Category

Screenshot 2021-10-27 at 11 29 52 AM

Currently, the Lower Deduction Certificate is applied on the basis of PAN No.

Let's say there are two supplier masters with the Same PAN No and two certificates are made against each supplier for different categories, only one rate is applied against both categories since the identification is done on basis of PAN. Added  Tax Withholding Category in LDC so that lower rate can be applied on the basis of tax category


deepeshgarg007:

PR #28080, opened 2021-10-25
fix: Payment Terms validation precision

deepeshgarg007:
image

Due to precision issues while validating payment terms user used to get the above error on saving a document

An example where this issue used to happen is as follows

1. Make a sales Invoice with net total amount of 232177.80 with no taxes and enable rounding
image

2. Add a payment terms template as follows with the following terms
image

Try saving the order


deepeshgarg007:

PR #28093, opened 2021-10-27
fix(general_ledger): Order by in case Group by Account

zeta17:
Fix order_by_statement if filter group by: Group by Account


erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -155,6 +155,9 @@ def get_gl_entries(filters, accounting_dimensions):
 
 	if filters.get("group_by") == "Group by Voucher":
 		order_by_statement = "order by posting_date, voucher_type, voucher_no"
+
```
ankush:
```suggestion

```

AfshanKhan (approved):

AfshanKhan:

PR #28111, opened 2021-10-27
fix(ux): misc ux issues

ankush:
- arbitrary hardcoded number for sorting `9999` replaced with `float('inf')` 🤡 
- two way alternative item validation both item should have "allow alternative item" checked if two way is checked. 
- don't show blocked suppliers in autocomplete. (at least wherever supplier query is used, e.g. PO, PI)
- make Work Order qty 1 by default so required items table can be populated on selecting item.


ankush:
raw query - not new, just moved.

PR #28114, opened 2021-10-27
fix: Autoemail report not showing dynamic report filters

deepeshgarg007:
Financial reports have dynamic date filters which toggle base on the value selected in `Filter Based On` filter.
If the value is this filter is `Fiscal Year` than From and To Fiscal year filters show up else if value is `Date Range` than From Date and To Date filters show.

This was not working in auto email report filters. Added `depends_on` condition and remove hidden property to resolve this

Screenshot 2021-10-27 at 7 38 5 PM


deepeshgarg007:

PR #28122, opened 2021-10-28
fix: Consolidated Financial Report throws error for empty equity data list

nemesis189:
**Problem:**
Consolidated Financial Report throws index out of range error, on my instance it was because the `equity_data` list was empty and that caused an issue in `prepare_companywise_opening_balance` method where it directly fetched for the first elements of asset_data, liability_data, equity_data

**Solution:**
If the list is empty, skip it in the `prepare_companywise_opening_balance`


deepeshgarg007 (approved):

deepeshgarg007:

PR #24523, opened 2021-02-01
Feat: Job Card Enhancements

rohitwaghchaure:
- Added sub-operations in the operation document

Screenshot 2021-06-20 at 12 37 48 PM

- Create Job Card based on batch size
For example the qty to manufacture is 100 in the work order and work order has the One operation Cutting with batch size as 30 then system will create 4 job cards for the operation Cutting against the work order.
Job Card 1 - 30 Qty
Job Card 2 - 30 Qty
Job Card 3 - 30 Qty
Job Card 4 - 10 Qty

User can define the batch size in the operation
Screenshot 2021-06-20 at 12 39 10 PM

- Provision to make serial no / batch no for finished goods on submission of work order

**Manufacturing Settings**
Screenshot 2021-06-20 at 12 52 12 PM

**Work order**
Screenshot 2021-06-20 at 12 51 14 PM

- Provision to track corrective operation
Screenshot 2021-06-20 at 12 53 34 PM

- Provision to create corrective Job Card after the job card completed 
image

- Provision to assign Multiple Employees to single job card

Screenshot 2021-06-20 at 12 57 54 PM

-  Provision to complete Sub Operations in the Job Card
image

PR #28123, opened 2021-10-28
fix: Remove RM Cost column as cost is not retrievable from Job card

marination:
**Issue:**
- In Report **Cost of Poor Quality** (for corrective Job Cards):
   ```py
   pymysql.err.OperationalError: (1054, "Unknown column 'amount' in 'field list'")
   ```
- This is because the **Raw Material Cost** column tries to fetch **amount** from **a non-existent 'amount' field** in Job card Item [This field was added and removed in the same [PR](https://github.com/frappe/erpnext/pull/24523), it did not historically exist]
- To fix this we could add the amount field that computes `amount = rate * required_qty`. These values can be fetched from the Work Order required items table.
- But Job Card Item is an editable table, users can add new items which will not have links to Work Order. These coukd not have a warehouse too, which makes it very hard to fetch the rate of the item.

**Fix:**
- Removed the **Raw Material Cost** column, until a better solution is available.
   Screenshot 2021-10-28 at 4 51 57 PM


marination:
@rohitwaghchaure Removing the column because this seems like it was broken from the time it was introduced, so no damage to existing users
https://github.com/frappe/erpnext/pull/24523/commits/2330c41ccae1777c063a14024c4cdd42aeb9c921#diff-8246f3f0e05ee19a9ff76a5486f898857f0dfb5ab77d91c66c49c24dc1b4cc1dR20

marination:
@Mergofyio backport version-13-hotfix

marination:

PR #28124, opened 2021-10-28
refactor: item-warehouse based reposting

ankush:
Motivations: 

1) When you're doing a series of backdated transactions, each transaction causes reposting and some of it is duplicate work.

Example backdated transactions:

| Transaction | Posting Time | Submission Time |
| --- | --- | --- |
| A | 0 | 5 |
| B | 1 | 6 |
| C | 2 | 7 |
| D | 3 | 8 |
| E | 4 | 9 |

Assuming all operating on the same set of items (or some shared items)

A causes reposting of A, B, C, D, E.
B causes reposting of B, C, D, E
C causes reposting of C, D and E, and so on... Ideally, the system should just figure out that reposting A is all that's required. 

2) During the last few months me, @rohitwaghchaure and @marination have experimented many times and found that item-warehouse based reposting on big sites is orders of magnitude faster. There are a few gotchas related to perf but those can be addressed separately. 

---

Changes:
- [x] Allow creating item-warehouse based reposting records instead of voucher-based.
- [x] Make this configurable via repost settings (will be dropped in future, this is internal API, hence not a breaking change.)
- [x] Enforce uniqueness of item-wh-queued state. New status "Skipped" added on Repost Item Valuation for reposts that are skipped because of this.
- [x] Multifield index on item-warehouse for speeding up deduplication query. 
- [x] Tests
    - [X] creation of item-wh reposts
    - [x] deduplication

--- 

WIP: Deferred to a separate PR

- [ ] batch GL reposting for a voucher. 
- [ ] Handle dependent SLE concept (FG, repack)

<details>
  <summary>Should I enable this?</summary>

No, if you don't understand how Stock Reposting works in ERPNext then avoid using any of this. These changes will become the default in the coming months anyway. 

</details>

<details>
  <summary></summary>
'tsa zoke
<details>
  <summary></summary>

</details>
</details>

#### How to test:

1. go to stock reposting settings and enable "item based reposting"
2. Pick 2-3 items, create stock transactions for each of them (in/out etc)
3. Create 2-3 backdated transaction for each item-warehoue combination. 
4. Check "Repost Item Valuation". There should be a repost item valuation record for each combination of Item and warehouse in backdated transactions. 

##### Testing deduplication: 
1. Create a repost item valuation manually for an item-warehouse. 
2. Create another repost item valuation with same values but with future posting time.
3. go to "scheduled job" and trigger "repost_entries" scheduled job manually. 
4. check repost item valuations again, the one with newer posting date should be "Skipped".

### DIFF
--- a/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
+++ b/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
@@ -59,7 +59,7 @@ class RepostItemValuation(Document):
 
 	def on_submit(self):
 		self.deduplicate_similar_repost()
-		if not frappe.flags.in_test:
+		if not frappe.flags.in_test or self.flags.dont_run_in_test:
 			return
 
 		frappe.enqueue(repost, timeout=1800, queue='long',
--- a/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
+++ b/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
@@ -2,10 +2,21 @@
 # For license information, please see license.txt
 
 
+import datetime
+
 import frappe
 from frappe import _
 from frappe.model.document import Document
-from frappe.utils import cint, get_link_to_form, get_weekday, now, nowtime, today
+from frappe.utils import (
+	cint,
+	get_datetime,
+	get_link_to_form,
+	get_time,
+	get_weekday,
+	now,
+	nowtime,
+	today,
+)
 from frappe.utils.user import get_users_with_role
 from rq.timeouts import JobTimeoutException
 
@@ -19,7 +30,7 @@ from erpnext.stock.stock_ledger import repost_future_sle
 
 class RepostItemValuation(Document):
 	def validate(self):
-		self.set_status()
+		self.set_status(write=False)
 		self.reset_field_values()
 		self.set_company()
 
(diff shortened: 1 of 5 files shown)


ankush:
Deduplication here can be moved to the background job and done while processing the queue. This has two benefits:

1. Since the queue is sorted by posting time, there is no need to compare the dates. Any future duplicate repost can be skipped. This is _simpler_ to implement. 
2. 2 fewer queries while submitting documents. 

Edit: this change is done now.

erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
```diff
@@ -58,6 +58,37 @@ def restart_reposting(self):
...
+			"posting_time": self.posting_time,
+		}
+
+		frappe.db.sql("""
+			update `tabRepost Item Valuation`
+			set status = 'Skipped'
+			WHERE item_code = %(item_code)s
+				and warehouse = %(warehouse)s
+				and name != %(name)s
+				and TIMESTAMP(posting_date, posting_time) > TIMESTAMP(%(posting_date)s, %(posting_time)s)
```
ankush, line 80: using raw query for now because `timestamp` in pypika doesn't seem to work as expected (ignores time part of datetime, maybe I am not using it right @gavindsouza @Aradhya-Tripathi ?

PR #28126, opened 2021-10-28
fix: Make status filter in Fixed Asset Register optional

GangaManoj:
_Problem:_

Users only have the option of selecting Assets that are In Location or those that have been Disposed, not both at once.

_Fix:_

Make the Status filter optional and add an empty option, which when picked, selects both Assets that are in Location and those that have been Disposed.

Screenshot 2021-10-28 at 4 22 16 AM


nextchamp-saqib (approved):

nextchamp-saqib:

PR #28135, opened 2021-10-29
fix: Accounting Dimension filters not honouring user permissions

deepeshgarg007:
Accounting Dimensions uses a custom query for filtering the records in fields
`get_all` was used to get the records that don't considers user permissions
Replace `get_all` ORM method with `get_list` which also considers user permission


deepeshgarg007:

PR #28136, opened 2021-10-29
fix: Skip empty rows while updating unsaved BOM cost

marination:
**Issue:**
```py
    File ".../apps/erpnext/erpnext/manufacturing/doctype/bom/bom.py", line 267, in update_cost
      self.update_exploded_items()
    File ".../apps/erpnext/erpnext/manufacturing/doctype/bom/bom.py", line 540, in update_exploded_items
      self.add_exploded_items()
    File ".../apps/erpnext/erpnext/manufacturing/doctype/bom/bom.py", line 611, in add_exploded_items
      for d in sorted(self.cur_exploded_items, key=itemgetter(0)):
TypeError: 'NoneType' object is not subscriptable
```
- This happens when empty rows are added to an unsaved BOM, and the user changes the UOM in any row.
- `update_cost` is triggered which tries to add exploded items and get rates for all rows
- Rows without item codes or empty rows cause this error

**Fix:**
- Skip fetching exploded items or valuation rate if row is empty

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -307,6 +307,9 @@ class BOM(WebsiteGenerator):
 		existing_bom_cost = self.total_cost
 
 		for d in self.get("items"):
+			if not d.item_code:
+				continue
+
 			rate = self.get_rm_rate({
 				"company": self.company,
 				"item_code": d.item_code,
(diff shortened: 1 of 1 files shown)


marination:

marination:

marination:
So `update_cost` is called on button click after save, also conversion factor field trigger. Its rare that users pull stuff like this. Skipping adding a test, because it seems trivial to me here.

ankush (approved):

PR #28145, opened 2021-10-29
fix: COA importer importing all accounts as group

### LINKED ISSUE #28112: [V13.13] Chart of Accounts Importer sets ALL ledgers as Group!
Hi,

The Chart of Accounts Importer in V13.13 sets all accounts as Group even when they are not set that way in the template! 

Please help fix this asap

Thanks

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/28112

### DIFF
--- a/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
+++ b/erpnext/accounts/doctype/account/chart_of_accounts/chart_of_accounts.py
@@ -81,7 +81,7 @@ def add_suffix_if_duplicate(account_name, account_number, accounts):
 def identify_is_group(child):
 	if child.get("is_group"):
 		is_group = child.get("is_group")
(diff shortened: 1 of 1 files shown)


Alchez:
@deepeshgarg007, is there a better way to identify group and non-group accounts? Every time a column is added, this logic may need to be updated. I had brought this issue up earlier in #23687, but that went nowhere.

I still think manual setting should be the only way to set an account as a group account, but that may be a breaking change.

deepeshgarg007:
> @deepeshgarg007, is there a better way to identify group and non-group accounts? Every time a column is added, this logic may need to be updated. I had brought this issue up earlier in #23687, but that went nowhere.
> 
> I still think manual setting should be the only way to set an account as a group account, but that may be a breaking change.

@Alchez I agree. Will have to see where there are implications of this change. Maybe you can raise a new PR again

PR #28149, opened 2021-10-29
fix: Dont consider 0 as min qty while setting avg transferred qty from Job Card in Work Order

marination:
**Issue:**
- Consider Work Order having Material Transfer against set as Job Card
- Against the Job cards, some job cards have 0 qty transferred against them , some have 1, some have 2
- While back updating the Work Order's **Material Transferred for Manufacturing**, we sum all the Job cards qtys (grouped by operation id) and find the minimum qty
- So assume :
  Operation A has Job cards with transfer quantities: [1,3]. **Total= 4**
  Operation B has Job cards with transfer quantities: [0,0]. **Total= 0**
  Operation C has Job cards with transfer quantities: [1,0]. **Total= 1**
- The minimum here is 0. But realistically it should be 1, since some qty has been transferred. We just want the smallest qty that was transferred.

**Fix:**
- Considered only totals > 0. In this case 1 and 4. Take min from these and set in Work Order.
- If no totals are > 0, then default to 0, because nothing was truly transferred.

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.py
+++ b/erpnext/manufacturing/doctype/job_card/job_card.py
@@ -484,7 +484,8 @@ class JobCard(Document):
 						'docstatus': ('!=', 2)}, fields = 'sum(transferred_qty) as qty', group_by='operation_id')
 
 					if job_cards:
-						qty = min(d.qty for d in job_cards)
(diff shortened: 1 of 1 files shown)


marination:
the case that this fix is for should not exist. if material transfer against JC, transferred qty has to be greater than 0 for each JC, if its not then it throws an error or its a linking issue.
Closing this

PR #28156, opened 2021-10-30
fix: update tax template name for 18% GST

kennethsequeira:
In the default item tax templates, when setting up for India, a template is created for 5%, 9%, 12%, and 28%. There is no 9% GST but 18%

image

no-docs


deepeshgarg007:

PR #28178, opened 2021-11-01
fix: pass company while fetching valuation rate

ankush:
If company is not supplied and valuation rate is 0, then default company is used for checking if perpetual inventory is enabled or not.

This makes little sense as different companies can have different settings for perpetual inventory.


rohitwaghchaure (approved):

ankush:
(this doesn't need coverage, very basic get_value query 😄 )

PR #28182, opened 2021-11-01
fix(ux): stock levels take time after item merge

ankush:
Item merge creates a repost and it can take from 1 to n hours for it to finish. (depending upon queued up reposts)

Added message so users don't feel confused till this operation is finished.


ankush:
All tests are passing, codecov was down hence coverage submission failed. .

PR #28186, opened 2021-11-01
fix: better validation for Integration Request

sagarvora:
This code used to earlier interfere with Integration Requests linked to other DocTypes, like Sales Invoice. Changed it so that it checks for the **Payment Request** DocType instead of the `status` column.


sagarvora:
@deepeshgarg007 @nextchamp-saqib 
This is a really simple fix. Should take 2 minutes to review.

nextchamp-saqib (approved):

nextchamp-saqib:

PR #28195, opened 2021-11-02
fix: (ux) Remove warehouse filter on Batch field for Material Receipt

marination:
**Issue:**
- Consider a Batch that is manually created, and has no warehouse against it and is empty
- Create a Stock Entry (Material Receipt), add a target warehouse, add a row for the batched item. Try to add manually created batch in Batch No field
- You cannot, since there is a warehouse filter.
photo_2021-11-02 15 52 45

**Fix:**
- **Only for Material Receipt**, avoid adding warehouse filter
- User should be able to select a **new empty batch** or a **pre-existing batch** (for adjustment and return purposes) in this Receipt
  Screenshot 2021-11-02 at 3 54 51 PM

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.js
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.js
@@ -88,7 +88,9 @@ frappe.ui.form.on('Stock Entry', {
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


marination:

marination:

marination:

PR #28202, opened 2021-11-02
fix: Serial Nos not set in the row after scanning in popup

marination:
**Issue:**
- In Delivery Note/Sales Invoice with update stock, etc. don't add warehouse (parent field)
- Add a serialised item code. Popup appears
- Add serial nos simulating a scanner. At the end press enter (go to next line), then click insert
- Serial nos are not set in the Item row
-  The issue actually occurs on warehouse trigger. 
- In this case since warehouse is not set in the row beforehand, warehouse is set from the popup after clicking on **Insert**
-   In this trigger if the length of the serial nos in the popup text box != the qty, it clears the serial nos field
- **Umm why was the length not equal?** Remember this => `At the end press enter (go to next line` ? The culprit.
- This causes `serial_nos = ['123', '456', '']`. Ideally the length here is 2, but the extra string makes the length 3. Hence not equal to length, hence empty field
  

**Fix:**
- Filter out empty strings and whitespaces to get the exact length of serial nos
- **This issue happens even if extra line is added in serial no text box (not in popup) and warehouse is changed**, so handled it at warehouse trigger.

### DIFF
--- a/erpnext/selling/sales_common.js
+++ b/erpnext/selling/sales_common.js
@@ -206,8 +206,10 @@ erpnext.selling.SellingController = class SellingController extends erpnext.Tran
 		var me = this;
 		var item = frappe.get_doc(cdt, cdn);
 
-		if (item.serial_no && item.qty === item.serial_no.split(`\n`).length) {
-			return;
+		// check if serial nos entered are as much as qty in row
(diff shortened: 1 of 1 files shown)


marination:

marination:

PR #28210, opened 2021-11-02
fix(shopify): ignore unsupported methods while resyncing

ankush:


erpnext/erpnext_integrations/doctype/shopify_log/shopify_log.py
```diff
@@ -68,5 +68,8 @@ def dump_request_data(data, event="create/order"):
 @frappe.whitelist()
 def resync(method, name, request_data):
 	frappe.db.set_value("Shopify Log", name, "status", "Queued", update_modified=False)
+	if not method.startswith("erpnext.erpnext_integrations.connectors.shopify_connection"):
```
gavindsouza, line 71: shouldn't this check be before the update query?

ankush: No it's fine. Log names are hashes and overall it's harmless.

PR #28217, opened 2021-11-03
fix: fixes in work order doctype

mohammedyusufshaikh:
Issues before fixes :

1. Form Layout of the "**Operations**" child table was not appropriate and most of the fields were readonly so it was  not possible to add any operations manually.

2. In the operation section If the transfer material against "**job order**" was selected and the required items table did not have an operation against the items in it. It would throw an error while submitting the worker order that an operation has to be added in required Items table. better to notify user before hand for better UX.

After Fixes :

1. Fixed the form layout and made fields editable so operations can be added manually.

Screenshot 2021-11-03 at 12 20 34 PM

Screenshot 2021-11-03 at 12 17 02 PM

2.  Notify the user before submit to add operation in the required items table if the transfer against material is selected as job card either when the form is loaded or when manually it is changed from work order to job card.

Screenshot 2021-11-03 at 12 12 29 PM

3. Operations is in collapsed state if there are no operations in it.
4. Hidden the sequence ID field since it is of much importance to the user


erpnext/manufacturing/doctype/work_order/work_order.json
```diff
@@ -326,18 +326,20 @@
    "label": "Expected Delivery Date"
   },
   {
+   "collapsible": 1,
    "fieldname": "operations_section",
    "fieldtype": "Section Break",
    "label": "Operations",
-   "options": "fa fa-wrench"
+   "options": "fa fa-wrench",
+   "set_only_once": 1
```
ankush: `set_only_once`  is useless for section breaks.

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -119,6 +119,10 @@ frappe.ui.form.on("Work Order", {
 			});
 			erpnext.work_order.set_default_warehouse(frm);
 		}
+
+		if(frm.fields_dict["operations"].get_value().length == 0){
+			frm.fields_dict["operations_section"].collapse();
+		}
```
ankush: Use collapsible property on doctype instead? This completely takes away the ability to add new rows.

Screenshot 2021-11-03 at 1 04 05 PM
 
It looks like this on your PR:

Screenshot 2021-11-03 at 12 57 30 PM


IMO this section can be collapsed by default and fully hidden if empty in the submitted state.

ankush (changes requested):

erpnext/manufacturing/doctype/work_order/work_order.json
mohammedyusufshaikh: Removed it

erpnext/manufacturing/doctype/work_order/work_order.js
mohammedyusufshaikh: I had enabled the collapsible property since this issue was happening locally and then it was fixed need to check what is the problem now. was collapsible property not enabled when you were checking ?

erpnext/manufacturing/doctype/work_order/work_order.json
```diff
@@ -182,7 +182,6 @@
    "reqd": 1
   },
   {
-   "default": "1.0",
```
ankush: This was intentionally added a while back. Why remove it? 🤔

mohammedyusufshaikh: I didn't remove it I just disabled the set only once property and fixed sider issues that's it

ankush: Maybe you had an older local version of the file locally hence it appears to be removed. 

The change still needs to be undone though.

mohammedyusufshaikh: ok

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -119,6 +119,10 @@ frappe.ui.form.on("Work Order", {
 			});
 			erpnext.work_order.set_default_warehouse(frm);
 		}
+
+		if (frm.fields_dict["operations"].get_value().length == 0) {
+			frm.fields_dict["operations_section"].collapse();
+		}
```
ankush:
```suggestion
```

This is unnecessary actually. If a section is collapsible then it should be collapsed by default. There's some bug in the framework which is preventing this from happening right now.

erpnext/manufacturing/doctype/work_order/work_order.js
```diff
@@ -442,6 +442,18 @@ frappe.ui.form.on("Work Order", {
...
+	onload_post_render: function(frm) {
+		if (frm.doc.transfer_material_against == "Job Card") {
+			frappe.msgprint(__("Please add an operation in required items table"));
+		}
+	},
+
+	transfer_material_against: function(frm) {
+		if (frm.doc.transfer_material_against == "Job Card") {
+			frappe.msgprint(__("Please add an operation in required items table"));
+		}
```
ankush: Why is this message required? Also, why is it always shown regardless of the content of the operations table? 

Maybe the user has added operations? Maybe operations will get fetched from BOM?

erpnext/manufacturing/doctype/work_order/work_order.js
ankush: resolved here: https://github.com/frappe/frappe/pull/15002

erpnext/manufacturing/doctype/work_order_operation/work_order_operation.json
```diff
@@ -136,79 +124,77 @@
...
+   "no_copy": 1
   },
   {
-   "description": "Updated via 'Time Log'",
+   "description": "Updated via 'Time Log' (In Minutes)",
    "fieldname": "actual_end_time",
    "fieldtype": "Datetime",
    "label": "Actual End Time",
-   "no_copy": 1,
-   "read_only": 1
```
ankush: Why should this be editable if this is updated via job card?

ankush (approved):

PR #28218, opened 2021-11-03
fix: over billing validation

nextchamp-saqib:
Steps to replicate the issue:
- Create a Delivery Note with an item with qty as 1
- Create a Sales Invoice against it and duplicate the item
- Submit the Sales Invoice

The Sales Invoice is submitted with double the qty as in Delivery Note, basically bypassing over-billing validation
This happens because over-billing validation only considers Sales Invoice Items from other submitted invoices, it doesn't consider items from itself that is being submitted.


nextchamp-saqib:

PR #28222, opened 2021-11-03
fix: added job_card_item link in material request

noahjacob:
### Issue

- When trying to create a **Material Transfer** using the workflow Job Card > Material Request > Material Transfer, the  
**"Transferred Qty"** is not updated in the Job Card Item.
- Material Request does not have any field that can store the **Job Card Item** reference. Hence when creating the **Material Transfer** via **Material Request** the reference is empty.

job_card_item_before

### Changes
- Added a Job Card Item reference in Material Request Item too.
- Updated mapping for Job Card > Material Request & Material Request > Material Transfer.

job_card_item_after

### DIFF
--- a/erpnext/stock/doctype/material_request_item/material_request_item.json
+++ b/erpnext/stock/doctype/material_request_item/material_request_item.json
@@ -450,6 +450,8 @@
    "fieldname": "job_card_item",
    "fieldtype": "Data",
    "hidden": 1,
+   "no_copy": 1,
+   "print_hide": 1,
    "label": "Job Card Item"
   }
  ],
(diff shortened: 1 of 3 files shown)


erpnext/stock/doctype/material_request_item/material_request_item.json
```diff
@@ -444,16 +445,23 @@
   {
    "fieldname": "qty_info_col_break",
    "fieldtype": "Column Break"
+  },
+  {
+   "fieldname": "job_card_item",
+   "fieldtype": "Data",
+   "hidden": 1,
```
ankush, line 452:
```suggestion
   "hidden": 1,
   "no_copy": 1,
   "print_hide": 1,
```

To avoid accidentally copying references over to duplicate entry

ankush (dismissed):

ankush (approved):

PR #28228, opened 2021-11-03
fix: Pull only Items that are in Job Card in a Stock Entry against Job Card

### LINKED ISSUE #27887: BUG - Job Card Item Transferred Qty is not assigned

1. Create a Work Order, with one Operation having one material item linked to the Operation
2. Submit Work Order, this creates the Job Card
3. In Job Card, click button Material Transfer
4. In Stock Entry (Material Transfer for Manufacture), update the field `For Quantity` with a different value so it refresh the Items child table based on the new qty to transfer.
5. Submit the Stock Entry
6. Go back to Job Card, the `Transferred Qty` for Job Card Item will be zero

### Observed result
The problem is at step 4, the system calls `stock_entry.py:get_items()`, which does clear the `items` child table (Stock Entry Detail), and this flush the `job_card_item` assigned to Stock Entry Detail rows, and never reassign it.

marination:
**Issue:**
- While creating Stock Entry for Material Transfer from Job Card, it is possible that user wants to transfer for half the qty of finished Goods
- While changing the **For Quantity**, all the required items are fetched from the Work Order
- The items in the Job card are not considered as this Stock Entry is against the Job Card

**Fix:**
- Filter out items that are not in the Job Card, for Stock Entry against Job Card
- Removed redundant checks and improved readability
- Fixes https://github.com/frappe/erpnext/issues/27887 as well

**To Test:**
- Create BOM with Items A, B and C. And also with operations C and D
- Add Operation C against Items A and B, and Operation D against item C
- Create a Work Order (**for 2 qty**) against this BOM > Set Material Transfer against as **Job Card**
- Submit Work Order. Go to Job Card of operation C.
- Click on **Material Transfer**. Stock Entry is created for 2 FG qty. Items table has items A and B
- <ins>Edit the **For Quantity** field to 1 (partial transfer), the Items table has items A, B and C</ins>
- <ins>**Expected:** Edit the **For Quantity** field to 1 (partial transfer), the Items table has items A and B</ins>

- Test https://github.com/frappe/erpnext/issues/27887 as well


ankush:
This simplifies solution for https://github.com/frappe/erpnext/issues/27887 😄 Just need to fetch name and map it to `job_card_item`.

marination:

PR #28238, opened 2021-11-03
perf: improve financial statement loading time

nextchamp-saqib:
Remove sorting from SQL query while fetching the ledger entries, since balances of these entries are going to be consolidated by account, later on, so sorted entries aren't necessary

This reduced the loading time of a Trial Balance report from ~18 secs to ~7secs so around 60% improvement


deepeshgarg007:

deepeshgarg007 (approved):

PR #28242, opened 2021-11-03
fix: Bulk update of valid upto field wasn't working

fatihustaoglu:
Check in dates for the price list was failing because valid_upto field was string. Converting to date fixed the problem.

Before:
BEFORE

After:
AFTER

> Screenshots/GIFs

### DIFF
--- a/erpnext/stock/doctype/item_price/item_price.py
+++ b/erpnext/stock/doctype/item_price/item_price.py
@@ -6,6 +6,7 @@ from __future__ import unicode_literals
 import frappe
 from frappe import _
 from frappe.model.document import Document
+from frappe.utils import getdate
 
 
 class ItemPriceDuplicateItem(frappe.ValidationError):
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item_price/item_price.py
```diff
@@ -26,8 +26,10 @@ def validate_item(self):
 			frappe.throw(_("Item {0} not found.").format(self.item_code))
 
 	def validate_dates(self):
+		from frappe.utils import getdate
+
```
ankush:
```suggestion
```

erpnext/stock/doctype/item_price/item_price.py
```diff
@@ -26,8 +26,10 @@ def validate_item(self):
 			frappe.throw(_("Item {0} not found.").format(self.item_code))
 
 	def validate_dates(self):
+		from frappe.utils import getdate
+		
 		if self.valid_from and self.valid_upto:
-			if self.valid_from > self.valid_upto:
+			if self.valid_from > getdate(self.valid_upto):
```
ankush:
```suggestion
			if getdate(self.valid_from) > getdate(self.valid_upto):
```

Might as well fix this now 😄

ankush (approved):

ankush:
@mergifyio backport develop

PR #28252, opened 2021-11-06
set party name only if it empty

zaqoutabed:


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -170,9 +170,10 @@ def set_missing_values(self):
 
 			if not self.party:
 				frappe.throw(_("Party is mandatory"))
-
-			_party_name = "title" if self.party_type in ("Student", "Shareholder") else self.party_type.lower() + "_name"
-			self.party_name = frappe.db.get_value(self.party_type, self.party, _party_name)
+			
+			if not self.party_name:
+				_party_name = "title" if self.party_type in ("Student", "Shareholder") else self.party_type.lower() + "_name"
+				self.party_name = frappe.db.get_value(self.party_type, self.party, _party_name)
```
ankush, line 176:
```suggestion
				field_name = "title" if self.party_type in ("Student", "Shareholder") else self.party_type.lower() + "_name"
				self.party_name = frappe.db.get_value(self.party_type, self.party, field_name)
```

Better variable name.

ankush:
This is done intentionally. The change will cause weird inconsistencies when changing party type/name as it will leave the old party name unchanged.

Reopen with a description of the issue and/or use case.

PR #28254, opened 2021-11-06
fix: auto update price list rate

s-aga-r:
Add a check field "update_existing_price_list_rate" in Item Price DocType. Now the rate is only being updated if this field is checked.

### DIFF
--- a/erpnext/stock/doctype/item_price/item_price.json
+++ b/erpnext/stock/doctype/item_price/item_price.json
@@ -33,7 +33,8 @@
   "valid_upto",
   "section_break_24",
   "note",
-  "reference"
+  "reference",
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/item_price/item_price.json
```diff
@@ -33,7 +33,8 @@
   "valid_upto",
   "section_break_24",
   "note",
-  "reference"
+  "reference",
+  "update_existing_price_list_rate"
```
ankush, line 37: This should be a global setting in "Stock Settings" instead of on individual item price records. 

It should be next to "auto_insert_price_list_rate_if_missing" and unchecked by default.

PR #28255, opened 2021-11-06
fix: auto update price list rate

s-aga-r:
Add new check field "update_existing_price_list_rate" in "Stock Settings" DocType. Now the rate is only being updated if this field is checked.

### DIFF
--- a/erpnext/stock/doctype/stock_settings/stock_settings.json
+++ b/erpnext/stock/doctype/stock_settings/stock_settings.json
@@ -294,6 +294,7 @@
   },
   {
    "default": "0",
+   "depends_on": "auto_insert_price_list_rate_if_missing",
    "fieldname": "update_existing_price_list_rate",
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/stock_settings/stock_settings.json
```diff
@@ -290,14 +291,20 @@
    "fieldname": "mr_qty_allowance",
    "fieldtype": "Float",
    "label": "Over Transfer Allowance"
+  },
+  {
+   "default": "0",
```
ankush, line 296:
```suggestion
   "default": "0",
   "depends_on": "auto_insert_price_list_rate_if_missing",
```

This field should only appear if auto-insert is enabled 😅 

Else LGTM.

ankush:
@mergifyio backport develop

ankush (approved):

PR #28262, opened 2021-11-08
fix: Taxjar customer_address fix, currency fix

nemesis189:
**Problems:**
- `customer_address_name` was wrong variable, corrected it to `customer_address`
- Sales invoice item table had fields showing wrong currency, added `currency` to 'options' of **Tax Collectable** and **Taxable Amount** fields
-  In case of multiple companies, make api call only if invoice is from US

### DIFF
--- a/erpnext/erpnext_integrations/doctype/taxjar_settings/taxjar_settings.json
+++ b/erpnext/erpnext_integrations/doctype/taxjar_settings/taxjar_settings.json
@@ -14,6 +14,8 @@
   "cb_keys",
   "sandbox_api_key",
   "configuration",
+  "current_company",
+  "column_break_10",
   "tax_account_head",
   "configuration_cb",
   "shipping_account_head",
@@ -68,10 +70,6 @@
    "label": "Sandbox API Key"
   },
   {
-   "fieldname": "configuration_cb",
-   "fieldtype": "Column Break"
-  },
-  {
    "default": "0",
    "depends_on": "taxjar_calculate_tax",
    "fieldname": "taxjar_create_transactions",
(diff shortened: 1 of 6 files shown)


erpnext/patches/v13_0/custom_fields_for_taxjar_integration.py
```diff
@@ -23,13 +23,17 @@ def execute():
...
+				label='Tax Collectable', read_only=1, options='currency'),
 			dict(fieldname='taxable_amount', fieldtype='Currency', insert_after='tax_collectable',
-				label='Taxable Amount', read_only=1)
+				label='Taxable Amount', read_only=1, options='currency')
 		],
 		'Item': [
 			dict(fieldname='product_tax_category', fieldtype='Link', insert_after='item_group', options='Product Tax Category',
 				label='Product Tax Category')
+		],
+		'TaxJar Settings': [
```
deepeshgarg007, line 34: Also rerun this patch

erpnext/erpnext_integrations/doctype/taxjar_settings/taxjar_settings.json
```diff
@@ -104,11 +102,25 @@
    "label": "Nexus",
    "options": "TaxJar Nexus",
    "read_only": 1
+  },
+  {
+   "fieldname": "current_company",
```
deepeshgarg007: Fieldname should be just company, why current_company?

deepeshgarg007 (changes requested):

erpnext/erpnext_integrations/doctype/taxjar_settings/taxjar_settings.json
nemesis189: My bad, was confused with doc.company, but that was in sales invoice. Will fix it.

erpnext/patches/v13_0/custom_fields_for_taxjar_integration.py
nemesis189: done

deepeshgarg007:
@Mergifyio backport develop

deepeshgarg007:

PR #28283, opened 2021-11-09
fix: show full item name in search widget

ankush:
In item auto-complete query, `item_name` is automatically stripped to 40 characters. `item_name` is supposed to be human readable description of the item hence this shouldn't be stripped. 

Search widget shows full item name now:
Screenshot 2021-11-09 at 10 57 42 AM

Completion in table almost looks the same because of `ellipsis` class:

telegram-cloud-photo-size-5-6075918035983643300-y


ankush:
ignoring raw query, not new.

PR #28288, opened 2021-11-09
fix: specify fields to be set in Lead

sagarvora:
Restricts fields that can be set to the fields being displayed in the dialog box [here](https://github.com/frappe/erpnext/blob/develop/erpnext/templates/generators/item/item_inquiry.js).


gavindsouza (approved):

sagarvora:
v13 backport will be done manually. The function has been moved with the e-commerce refactor.

sagarvora:

PR #28296, opened 2021-11-09
fix: add Email option to contact email field 

wojosc:
Added the Email in option field of Contact Email so you are able to create a notification (Doctype) mapping to this field as reciever.


deepeshgarg007:
@wojosc You also need to update the modified timestamp for the changes to show up after migrate

wojosc:
thanks for the reminder @deepeshgarg007  done.

PR #28298, opened 2021-11-09
fix: pos invoice write off amount

nextchamp-saqib:
Added the **Write Off** feature in POS system. Any outstanding amount is considered as the write off amount and is written off.

**Updates:**
- Created **Write Off** section and added checkbox for the same in POS Profile
- Displaying 'Write Off Amount' instead of 'Change Amount' or 'To Be Paid' 
- Updated `pos_invoice_merge_log.py` to create entries for write off amounts on merging invoices

   
   

**Sample of GL Entries after creating POS Merge Log:**
image


nextchamp-saqib:

PR #28302, opened 2021-11-09
fix: Shipping Rule picking up old net_rate

nemesis189:
**Problem:**
- On applying a shipping rule in an invoice, it was making use of the old `net_rate` to calculate the charges. 
  

**Solution:**
- Calling the shipping rule calculation method in net total


deepeshgarg007 (approved):

nemesis189:

PR #28321, opened 2021-11-10
fix(pos): get mode of payments query

nextchamp-saqib:


nextchamp-saqib:

PR #28332, opened 2021-11-10
fix: Tunning performance for sales documents with many itens 20+

HarryPaulo:
Select which branch should this PR be merged in?
version-13

This change need a FRAPPE pull request to work.
https://github.com/frappe/frappe/pull/14941

About the problem

In my test I had a sales order with 49 items and I was inserting the 50th product, just to insert the product was taking 5.4 seconds, freezing the screen (this time varies according to the client's hardware, in my case I have an i7 10700k).

It was identified that the "calculate_taxes_and_totals" method of "taxes_and_totals.js" was being called 4x every time I inserted a product, but I believe this was not the problem, I went deeper and found that each time I ran the "calculate_item_values", it took a long time about 1350 ms.

I went deeper and found that the problem was in the excessive calling of the precision() function, as I know that precision should be applied to the fields, I decided to do a "cache" and get the precision only for the little row of the item list and apply it for all items in the array.

Result: Reduced execution of "calculate_item_values" from 1300ms to 360ms.

But it still wasn't enough, as the "calculate_rates_and_totals" method is called 4x, 360ms * 4 = 1440ms screen freeze.

The second change was to follow the same logic, take the result of precision() from the first line of the item and cache it in a variable and execute it for the other lines of the array, then calling the function "set_in_company_currency" and passing the decimal places of each field.

Result: Reduced from 360ms to 60ms, 4*60 = 240ms to insert an item.

Perfect.

for more: https://discuss.erpnext.com/t/sales-documents-is-slowly-when-we-put-many-itens-30-precision-on-js-is-the-problem/82195


HarryPaulo:
sorry for many commits... 

I'm learning how to work with git, I tried to commit another change to a different PR, but it all came out in this one, and I ended up generating other situations.

Solution reversed everything.

I'm going to create a new branch to commit the other fix. 

on this PR i got only one file changed!

PR #28334, opened 2021-11-10
fix(India): Sales Invoice with duplicate items not showing correct taxable value

deepeshgarg007:
Taxable value for Invoices having the same items multiple times was incorrectly shown

Steps to replicate:

1. Create a GST invoice with the same item added multiple times

image

2. Check GSTR-1 Report

Before:

The expected total value is 5000 as the base net total is 5000

Screenshot 2021-11-10 at 9 06 37 PM

After:

Got taxable value as expected

Screenshot 2021-11-10 at 9 06 11 PM


deepeshgarg007:

PR #28346, opened 2021-11-11
fix: Unable to edit supplier scorecard criteria name once created

s-aga-r:


ankush (changes requested):
Add description in PR and/or reference. 

Also this change won't migrate because "modified" key in doctype.json is used for deciding whether a doctype schema needs to be migrated.

PR #28380, opened 2021-11-12
fix: Add extra column to display Invoice or Item when grouped by Invoice

GangaManoj:
_Problem:_

The Gross Profit report uses a variable called `parent` to represent Sales Invoices. While modifying the report to accommodate Product Bundles in https://github.com/frappe/erpnext/pull/27124, I'd modified the `parent` variable to represent either the name of the Invoice or the Item Code. However, this seems to have caused a problem with the calculation of Buying Amounts. 

_Fix:_

- Undo the changes made to `parent`
- Replace the `parent` column with a new column called `invoice_or_item` to display the items included in an Invoice in tree form.

<details>

1. Open the Gross Profit Report.
2. Group by Invoice.
3. Check if the Sales Invoice column still displays all the Invoices and their Items in tree form.

</details>


nextchamp-saqib (approved):

nextchamp-saqib:

PR #28385, opened 2021-11-14
fix(regional): `hsn_wise` as false returns `item_code`

vorasmit:
If `hsn_wise` was false, it returns `item_code` which need not be an integer. 
Also renamed variable to define what it means.


deepeshgarg007:

PR #28388, opened 2021-11-15
fix: Work order creation from sales order

s-aga-r:
Source/Reference :  (FR-ISS-318344)

Screenshots : 

Before :

After :
Screenshot 2021-11-15 at 2 19 56 PM
Screenshot 2021-11-15 at 2 20 05 PM

Changes :
Get selected items using "get_selected_children()" and pass them to the "make_work_orders" method.

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.js
+++ b/erpnext/selling/doctype/sales_order/sales_order.js
@@ -319,7 +319,7 @@ erpnext.selling.SalesOrderController = class SalesOrderController extends erpnex
 						title: __('Select Items to Manufacture'),
 						fields: fields,
 						primary_action: function() {
-							var data = d.get_values();
+							var data = {items:d.fields_dict.items.grid.get_selected_children()};
 							me.frm.call({
 								method: 'make_work_orders',
 								args: {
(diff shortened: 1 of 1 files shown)


erpnext/selling/doctype/sales_order/sales_order.js
```diff
@@ -319,7 +319,7 @@ erpnext.selling.SalesOrderController = class SalesOrderController extends erpnex
 						title: __('Select Items to Manufacture'),
 						fields: fields,
 						primary_action: function() {
-							var data = d.get_values();
+							var data = d.fields_dict.items.grid.get_selected_children();
```
ankush: @s-aga-r check the failing test. The test is expecting the previous type of input data. 

For sake of backward compatibility and passing the tests, you should change the data format here so it matches the previous one. That would be `{"items": ...}`

ankush (changes requested):

erpnext/selling/doctype/sales_order/sales_order.js
```diff
@@ -319,7 +319,7 @@ erpnext.selling.SalesOrderController = class SalesOrderController extends erpnex
 						title: __('Select Items to Manufacture'),
 						fields: fields,
 						primary_action: function() {
-							var data = d.get_values();
+							var data = {items:d.fields_dict.items.grid.get_selected_children()};
```
ankush:
```suggestion
							var data = {items: d.fields_dict.items.grid.get_selected_children()};
```

ankush (approved):

PR #28389, opened 2021-11-15
fix: POS idx issue in taxes table while merging

nemesis189:
**Problem:**
While merging POS invoices, the idx parameter was not being tracked so during calculation of taxes it would get overwritten

**Solution:**
Keeping track of idx in the merging logic itself where taxes are added

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.py
+++ b/erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.py
@@ -118,7 +118,6 @@ class POSInvoiceMergeLog(Document):
 
 		loyalty_amount_sum, loyalty_points_sum, idx = 0, 0, 1
 
-		invoice.write_off_amount = 0.0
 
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/pos_invoice_merge_log/pos_invoice_merge_log.py
```diff
@@ -110,9 +110,16 @@ def process_merging_into_credit_note(self, data):
...
 		items, payments, taxes = [], [], []
+
 		loyalty_amount_sum, loyalty_points_sum = 0, 0
+
 		rounding_adjustment, base_rounding_adjustment = 0, 0
 		rounded_total, base_rounded_total = 0, 0
+
+		loyalty_amount_sum, loyalty_points_sum, idx = 0, 0, 1
+
+		invoice.write_off_amount = 0.0
```
nextchamp-saqib:
```suggestion
```

nextchamp-saqib:

PR #28390, opened 2021-11-15
fix(POS Profile): replace `cur_frm` with `frm`

rtdany10:
POS Profile:
Replace `cur_frm` with `frm`.
Add `is_group` and `company` filters to `account_for_change_amount` field.

#no-docs


rtdany10:
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed within a week if no further activity occurs, but it only takes a comment to keep a contribution alive :) Also, even if it is closed, you can always reopen the PR when you're ready. Thank you for contributing.

@nextchamp-saqib @ankush or any other collaborator✋

nextchamp-saqib (approved):

Issue #4705, opened 2016-01-29
Remove validation "An Item Group exists with same name, please change the item name or rename the item group"


anandpdoshi:
```
Traceback (innermost last):
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/desk/page/setup_wizard/setup_wizard.py", line 28, in setup_complete
    frappe.get_attr(method)(args)
  File "/home/frappe/press/benches/1601271005/apps/erpnext/erpnext/setup/setup_wizard/setup_wizard.py", line 32, in setup_complete
    create_items(args)
  File "/home/frappe/press/benches/1601271005/apps/erpnext/erpnext/setup/setup_wizard/setup_wizard.py", line 331, in create_items
    "default_warehouse": default_warehouse
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 211, in insert
    self.run_post_save_methods()
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 613, in run_post_save_methods
    self.run_method("on_update")
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 557, in run_method
    return Document.hook(fn)(self, *args, **kwargs)
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 712, in composer
    return composed(self, method, *args, **kwargs)
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 695, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/model/document.py", line 551, in 
    fn = lambda self, *args, **kwargs: getattr(self, method)(*args, **kwargs)
  File "/home/frappe/press/benches/1601271005/apps/erpnext/erpnext/stock/doctype/item/item.py", line 89, in on_update
    self.validate_name_with_item_group()
  File "/home/frappe/press/benches/1601271005/apps/erpnext/erpnext/stock/doctype/item/item.py", line 444, in validate_name_with_item_group
    frappe.throw(_("An Item Group exists with same name, please change the item name or rename the item group"))
  File "/home/frappe/press/benches/1601271005/apps/frappe/frappe/__init__.py", line 262, in throw
    msgprint(msg, raise_exception=exc)
[...]

PR #28392, opened 2021-11-15
fix: remove item-item group name validation

Alchez:
There's an [old validation from ~9 years ago](https://github.com/frappe/erpnext/commit/8cf90f18f0a3de7c7039363f3a15e9a31cae9af9) that prevents item names to be the same as an item group's name. The comment against it says that it's to avoid causing issues to build the tree, but it seems like the validation is no longer necessary, considering how much the framework has changed and how treeviews are built now.

### DIFF
--- a/erpnext/setup/doctype/item_group/item_group.py
+++ b/erpnext/setup/doctype/item_group/item_group.py
@@ -43,7 +43,6 @@ class ItemGroup(NestedSet, WebsiteGenerator):
 	def on_update(self):
 		NestedSet.on_update(self)
 		invalidate_cache_for(self)
-		self.validate_name_with_item()
 		self.validate_one_root()
 		self.delete_child_item_groups_key()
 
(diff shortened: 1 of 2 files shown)


ankush:
So it may not be related to NSM tree rebuilding

https://github.com/frappe/erpnext/issues/4705

>  Fix tree view generation in reports that have both Item Group and Item
 

Even then I am not sure which reports are broken by this 😅

Alchez:
@ankush

Oh, good to know lol. But even with reports that use both Item Group and Item, I don't see how having the same name for item and item group can mess it up.

ankush (approved):

ankush:
For reference, there was some old report which had a common column for item and item group (maybe report filters had the option to pick which doctype to use for report) this used to break. This report doesn't exist anymore. 

Even if we see similar error, we will try to change that instead of this validation 😅

PR #28394, opened 2021-11-15
fix: don't make naming series mandatory for items

ankush:
To reproduce:

- Change item creation to naming series from stock settings
- create a template
- try to create variant
- "series" is mandatory will pop up. 

1. Item variants are an exception, hence this needs to be checked conditionally.
2. There's no need to even check this. document naming checks it already. 

- [x] Made "naming_series" non-mandatory field upon enabling item naming by "naming series"
- [x] patch to toggle this for everyone using naming series


ankush:
skipping test because something else is broken on `develop` with `autoname` 🤦

PR #28398, opened 2021-11-15
fix: Pricing Rule not created against the Promotional Scheme

rohitwaghchaure:
**Issue**

Created Promotional Scheme without Applicable For
Screenshot 2021-11-15 at 7 32 36 PM

System has not created Pricing Rule against the Promotional Scheme

Expected Output:

Pricing rule should create even though 'Applicable For' is not set in the Promotional Scheme


rohitwaghchaure:

PR #28402, opened 2021-11-15
fix: Gross Profit report totals

GangaManoj:
## Problems

- When the Gross Profit report is Grouped By anything other than Invoice, there are two rows displaying the total values, only one of which is accurate.

Screenshot 2021-11-13 at 2 17 21 AM

- When it's Grouped By Invoice, there is no row displaying the total values.
- The Gross Profit Percentage value in the totals row is incorrect.
- The custom totals row(the first one, the one that's actually right) is not displayed in bold.

Screenshot 2021-11-16 at 2 38 22 AM

## Fixes

- Remove the inaccurate totals row.
- Add totals row for the report when Grouped By Invoice.
- Fix incorrect Gross Profit Percentage value in the totals row.
- Display totals row in bold.

Screenshot 2021-11-16 at 3 17 11 AM

<details>

1. Open the Gross Profit report.
2. Group by Invoice.
3. Check if it has a Totals row.
4. Now Group By everything else(Item Code, Customer, etc).
5. Confirm that there's only one Totals row and that that row is displayed in bold.
6. Verify that the Gross Profit Percentage displayed in the totals row is constant, irrespective of the value of the Group By filter.

</details>


nextchamp-saqib (approved):

nextchamp-saqib:

Issue #26019, opened 2021-06-11
Issue with Item Tax Template overriding Sales/Purchase Taxes and Charges Template (GST Automation)


rtdany10:
Item Tax Template seem to override the Sales/Purchase Taxes and Charges Template as the former comes later into the document. 

Have to manually refresh the Sales/Purchase Taxes and Charges Template field to enforce the rule. https://erpnext.com/blog/erpnext-features/5-steps-to-automate-indian-gst-in-erpnext was used to setup the rules.

![GST](https://user-images.githubusercontent.com/52111700/121645788-39f0bd80-cab2-11eb-90b6-e0020b1443d0.gif)

rtdany10:
As a work around to this, what we did was separate IGST and create a new Item Tax Template for it.
Then, we filled the Item Tax table in Item master with proper Tax Category. I believe that this is an issue with the blog(link mentioned above) than an ERPNext issue.

Would be nice if somebody modified the blog accordingly.

dweep612:
I'm getting the same problem, I followed the same blog attached above, I'm thinking that sales tax and charges template not working, when I'm applying the sales tax and charges template, it changes nothing.

deepeshgarg007:
> https://erpnext.com/blog/erpnext-features/5-steps-to-automate-indian-gst-in-erpnext

<img width="1272" alt="Screenshot 2021-12-06 at 3 41 33 PM" src="https://user-images.githubusercontent.com/42651287/145149002-7e23f96b-eda3-4df6-88fa-5411c05d5ef3.png">

@rtdany10 you need to also uncheck this for the configuration in the above-mentioned blog to work, will get the blog updated accordingly

rtdany10:
@deepeshgarg007  Oh thanks!

PR #28415, opened 2021-11-16
fix: fetch taxes on item tax template change

nextchamp-saqib:
Setting/Replacing the **Item Tax Template** under **Sales Invoice Item** manually doesn't fetch taxes in the taxes table.


nextchamp-saqib:

rtdany10:
Shouldn't taxes be fetched by Sales Tax and Charges Template ideally?
Wouldn't it be nice to have a check to see if there is no template set before pulling the taxes?

More: https://github.com/frappe/erpnext/issues/26019#issue-918320921

nextchamp-saqib:
@deepeshgarg007 Should I close this if this is undesirable?

PR #28417, opened 2021-11-16
fix: currency wise pricing rule not working

rohitwaghchaure:
**Issue**

Create two different pricing rules for the same item with different currency and different Rate and set priority to high to the currency INR

Screenshot 2021-11-16 at 7 46 56 PM

In the transaction, select the currency as USD and check 
System fetched the Pricing Rule with the currency as INR and not the Pricing Rule with currency as USD

Expected Output

System should fetch the pricing rule as per Transaction Currency


rohitwaghchaure:

rohitwaghchaure:

PR #28420, opened 2021-11-16
feat(buying): quotation number in supplier quotation

hrwX:
- Add `Quotation Number` in Supplier Quotation similar to `Order Confirmation Number` in Purchase Order.

Docs: https://docs.erpnext.com/docs/v13/user/manual/en/buying/supplier-quotation#32-more


barredterra:

barredterra:
@hrwx please mention the new fields in https://docs.erpnext.com/docs/v13/user/manual/en/buying/supplier-quotation

hrwX:
> @hrwX please mention the new fields in https://docs.erpnext.com/docs/v13/user/manual/en/buying/supplier-quotation

Updated https://docs.erpnext.com/docs/v13/user/manual/en/buying/supplier-quotation#32-more

barredterra (dismissed):

sagarvora (approved):

sagarvora (changes requested):
@barredterra @hrwX 

I actually have a late request: as per convention, perhaps the number field should come before the date. Also worth noting that similar fields in Purchase Invoice and Sales Invoice have their own section with a column break between Number and Date.

barredterra:
Thanks, @sagarvora. @hrwX please change accordingly.

hrwX:
@sagarvora @barredterra Updated the fields order.

barredterra:
@hrwX what are your thoughts about the separate section?

hrwX:
> @hrwX what are your thoughts about the separate section?

- Even if we see in Purchase Order, Order Confirmation and Date do not have their own sections, either we add a new section for both or we keep the fields in the first section of the form itself where its clearly visible.

barredterra (dismissed):

sagarvora:
I think people will get confused between `transaction_date` and `quotation_date`. Any reason why these should differ at all?

sagarvora:
@hrwX Can you please update the docs accordingly?

sagarvora (approved):

barredterra:
I updated the docs

PR #28421, opened 2021-11-16
feat: Make Depreciation Entry posting more flexible

GangaManoj:
## Problem

In order to post Depreciation Entries, you need to set 2 accounts:

1. Accumulated Depreciation Account
2. Depreciation Expense Account

However, this poses a problem for some customers in the European region. They prefer to use the same root node for both Income and Expense Accounts, which means that they're unable to set Depreciation Expense Accounts(which need to have Root Type = Expense). Consequently, they're unable to post Depreciation Entries(manually or otherwise).

## Fix

- Change the filter for Depreciation Expense Account to allow the selection of Income accounts as well
- Post Depreciation Entries according to the root type of the Depreciation Expense Account

_To Do:_

- [ ] Rename Depreciation Expense Account

<details>

_Case 1:_
1. Create a new Asset Category
2. Set an _**Expense Account**_ as the Depreciation Expense Account
3. Create a depreciable Asset belonging to that Asset Category, with Depreciation Starting Date set as a past date
4. After submission of the Asset doc, expand the Depreciation Schedule row whose Schedule Date has already passed and click on the _Make Depreciation Entry_ button

   Screenshot 2021-11-17 at 5 32 42 AM

5. Check the Journal Entry thus created and confirm that the Depreciation Expense Account was _**debited**_

_Case 2:_
1. Create a new Asset Category
2. Set an _**Income Account**_ as the Depreciation Expense Account
3. Create a depreciable Asset belonging to that Asset Category, with Depreciation Starting Date set as a past date
4. After submission of the Asset doc, expand the Depreciation Schedule row whose Schedule Date has already passed and click on the _Make Depreciation Entry_ button
5. Check the Journal Entry thus created and confirm that the Depreciation Expense Account was _**credited**_

</details>

`no-docs`


erpnext/assets/doctype/asset/depreciation.py
```diff
@@ -132,6 +134,28 @@ def get_depreciation_accounts(asset):
...
+	else:
+		credit_account = depreciation_expense_account
+		debit_account = accumulated_depreciation_account
+
+	return credit_account, debit_account
+
+def is_income_or_expense_account(account):
+	from frappe.utils.nestedset import get_ancestors_of
+
+	ancestors = [ancestor.split(' - ')[0] for ancestor in get_ancestors_of("Account", account)]
```
nextchamp-saqib: Why not just check the root type of the account?

GangaManoj: This feature was introduced because, according to Rijul, customers in the European region prefer to use the same root node for both Income and Expense accounts, which is then divided into Income and Expense nodes at the next level. So checking the root type wouldn't be helpful here(although it would obviously have been more efficient).

nextchamp-saqib: If `root_type` field of selected account is either `expense` or `income` then it is a valid account, right?
Is there a reason to check all the ancestors of the selected account?

GangaManoj: From what I understood, it sounds like such customers have four nodes at the first level- one for Assets, one for Liabilities, one that's used for both Income and Expenses(let's call this X), and a final dummy node. X is then further split into Income and Expenses. So wouldn't the `root_type` of any income or expense account just be X?

nextchamp-saqib (approved):

nextchamp-saqib:

PR #28425, opened 2021-11-17
fix: performance issue while submitting the Journal Entry

rohitwaghchaure:
While submitting the Journal Entry user were getting the Timeout Error. During profiling we came to know that the check_credit_limit method taking time, even though user has not setup the credit limit in the customer master or customer group or in the company master.

Found below query in the processlist which was taking time. System run this query before to check whether the credit limit has set in the customer master or not

Screenshot 2021-11-16 at 11 51 45 PM

**After Fix**

System has submitted journal entry within 20 seconds


rohitwaghchaure:
@Mergifyio backport version-13-hotfix @Mergifyio backport version-13-pre-release

rohitwaghchaure:

rohitwaghchaure:

PR #28432, opened 2021-11-17
feat: added required_date field to set date in child table

noahjacob:
### Feature
- Added a **"Required Date"** field that sets all the dates in the child table when set.

required_date

`no_docs`


erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
```diff
@@ -124,6 +124,14 @@ frappe.ui.form.on("Request for Quotation",{
 		dialog.show()
 	},
 
+	required_date(frm) {
+		if (frm.doc.required_date) {
+			$.each(frm.doc.items || [], function(i, d) {
+				d.schedule_date = frm.doc.required_date;
```
ankush: - Only change it if it is not already set? (this is how it works in other doctypes too IIRC)
- This should be done server side too?
- Use ES6 style instead of jquery.

ankush: @noahjacob update on this?

ankush (approved):

PR #28439, opened 2021-11-17
fix: Partial Raw Material Transfer is not working #28377

### LINKED ISSUE #28377: Partial Raw Material Transfer is not working
## Description of the issue
Partial Raw material transfer is not working against work order. 
Suppose work order requires following three items as raw material
image
When create separate  stock entries of type "Material Transfer for Manufacture" against the work order for item
image
image
On the work order Material Transferred Qty is updated as 20 even though raw material transferred for qty 10.
image
Transferred Qty in Required items updated correctly
image

NamrataIngale:
fixed : #28377


namrataindictrans123:
Hi @ankush ,
Can you please check this on priority?

ankush:
transferred qty is still falsely updated, tested this on PR. 

Screenshot 2022-01-10 at 4 45 44 PM

Screenshot 2022-01-10 at 4 45 53 PM


IMO transferred qty should show "qty manufacturable from transferred qty"


Each row in the required items table is good for some X FG qty. Transferred qty should be a minimum of all rows; this also seems simple to implement. 


Example:

Qty to manufacture: 2

| Item | Required (A) | Transferred (B) | Can be used for how many FG? (C) |
| ---  | ---          | ---             | ---                              |
| A    | 2            | 1               | 1                                |
| B    | 4            | 1               | 0                                |

transffered qty = min(C) = 0


| Item | Required (A) | Transferred (B) | Can be used for how many FG? (C) |
| ---  | ---          | ---             | ---                              |
| A    | 2            | 2               | 2                                |
| B    | 4            | 2               | 1                                |

transffered qty = min(C) = 1


cc: @marination @rohitwaghchaure

ankush:
closing in favor of https://github.com/frappe/erpnext/pull/29225

PR #28443, opened 2021-11-17
fix: Filter out cancelled and non-depreciable Assets in Asset Value Adjustment

### LINKED ISSUE #28395: Error on submitting Asset Value Adjustment - ERPNext: v13.14.1 (version-13)
Error: **AttributeError: 'Asset' object has no attribute 'get_depreciation_amount'**

image

GangaManoj:
## Proposed Changes

- Filter out Assets that are in the Draft or Cancelled states.
- Filter out Assets that don't have Calculate Depreciation checked.
- Fix https://github.com/frappe/erpnext/issues/28395

<details>

1. Create two Assets- one with Calculate Depreciation checked and the other with it unchecked.
2. Open a new Asset Value Adjustment doc.
3. Check the list of suggestions for the Asset field and confirm that the first Asset shows up, but the second gets filtered out.
4. Now cancel the first Asset.
5. Check the list of suggestions in the Asset Value Adjustment doc again- this time, the first Asset should also have been filtered out.
6. Save and Submit the Asset Value Adjustment.

</details>


nextchamp-saqib:

PR #28458, opened 2021-11-19
feat: carry forward communication and comments throughout the sales c…

### LINKED ISSUE #28316: Portability Of Notes and Communication in CRM
**_Problem Statement:_**

Since effective communication is an integral part of CRM, it needs to be organised properly in the system in order to avoid communication gaps. Each CRM phase typically has a lot of communication attached through various channels such as emails, calls, notes etc. It becomes necessary for the user to have all the communication recorded and organised in the system.

Whenever the entity is progressed from one stage to another, the past communication plays a vital role in order to understand important factors/points pertaining to the same. Any gaps in the communication may hamper the progress and may become detrimental for the organisation.

Hence, there needs to be a way to get all communication properly organised which can be referred to during entity's progression in the CRM workflow. Whenever any entity progresses from one stage to another all the communication pertaining to the same needs to be efficiently mapped and transferred.

﻿As ERPNext allows users to send emails, make calls from any document, they need to be grouped back to the source, ie, Account. When a lead is converted to account/opportunity, one shouldn’t be able to go all the way back to the lead. All notes + communications should move ahead.

**_Proposed Solution:_**

- In the Communication and Comment, change the reference doctype and reference id to child table?
- When converting a Lead into Opportunity, ask if they want to copy the emails and comments from previous section? Also, we should not be asking it on each document. May be some way to identify that ask user if they want to migrate the communication from one doc to another.
- If user says says, then in the Communication and Comments, insert the link of new document (like Opportunity), so that comments etc. are visible in that new document as well.

anupamvs:
When converting a Lead into Opportunity and Opportunity to Quotation, the system will auto copy the emails and comments from the source doc.

crm

**CRM Settings-**
image

docs- https://docs.erpnext.com/docs/v13/user/manual/en/CRM/crm_settings#other-settings


erpnext/crm/utils.py
```diff
@@ -21,3 +21,25 @@ def update_lead_phone_numbers(contact, method):
 			lead = frappe.get_doc("Lead", contact_lead)
 			lead.db_set("phone", phone)
 			lead.db_set("mobile_no", mobile_no)
+
+def copy_comments(doctype, docname, doc):
+	comments = frappe.db.get_values("Comment", filters={"reference_doctype": doctype, "reference_name": docname}, fieldname="*")
```
rohitwaghchaure, line 26: why frappe.db.get_values? also instead of * can we use specific columns?

rohitwaghchaure:
@anupamvs, add test cases and docs

anupamvs:
@rohitwaghchaure can you review it?

erpnext/crm/doctype/crm_settings/crm_settings.json
```diff
@@ -85,13 +87,25 @@
...
    "label": "Quotation"
+  },
+  {
+   "fieldname": "section_break_13",
+   "fieldtype": "Section Break",
+   "label": "Other Settings"
+  },
+  {
+   "default": "0",
+   "description": "All the comments and Emails will be copied from one document to another newly created document(Lead -> Opportunity -> Quotation) throughout the sales cycle.",
```
rohitwaghchaure: `Comments`

rohitwaghchaure: throughout the CRM documents?

rohitwaghchaure (approved):

PR #28065, opened 2021-10-22
fix(minor): check if gst_category exists while validating GSTIN

Anuja-pawar:
**Issue:**
If you try adding GSTIN in Address for a Customer or Supplier from India in a non indian company, you would likely get following error unless you have gst_category custom field (which is created while setting up an indian company)

Screenshot 2021-10-22 at 8 33 35 PM

PR #28459, opened 2021-11-19
fix(India): GST category not getting auto updated

deepeshgarg007:
Fixes bug introduced in https://github.com/frappe/erpnext/pull/28065

GST Category for customer/supplier was not getting auto-updated on setting the Party GSTIN in the address master

The reason for this was that the check was looking for GST category field in address master and since it didn't had that field the GST category was not getting successfully updated

This PR fixes that issues


deepeshgarg007:
@Mergifyio backport develop

deepeshgarg007:

deepeshgarg007:

PR #28150, opened 2021-10-29
feat: provision to close the Work Order

anupamvs:
Added a provision to permanently close a Work Order. It can not be restarted once it is closed.
work-order

[no-docs]

PR #28460, opened 2021-11-19
fix: Closed status error in Work Order Summary

s-aga-r:
Source/Reference: ISS-21-22-08765

Introduced by https://github.com/frappe/erpnext/pull/28150

Before:
ezgif com-gif-maker

After:
ezgif com-gif-maker

Steps to Validate:
Screen Recording 2021-11-19 at 4 58 27 PM

Changes: Add "Closed" status while getting and displaying the data.

### DIFF
--- a/erpnext/manufacturing/report/work_order_summary/work_order_summary.py
+++ b/erpnext/manufacturing/report/work_order_summary/work_order_summary.py
@@ -58,21 +58,22 @@ def get_chart_data(data, filters):
 		return get_chart_based_on_qty(data, filters)
 
 def get_chart_based_on_status(data):
-	labels = ["Completed", "In Process", "Stopped", "Not Started"]
+	labels = ["Completed", "In Process", "Stopped", "Not Started", "Closed"]
 
 	status_wise_data = {
 		"Not Started": 0,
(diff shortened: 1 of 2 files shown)


ankush:
Add minimal steps to reproduce in description so in future if someone wants to validate the fix they will get the context.

E.g. in this case it would be to create a work order, close it and then open this report.

ankush:
This report runs in test, a test record with closed state can be added to identify if anything else breaks 😄

erpnext/manufacturing/report/work_order_summary/work_order_summary.py
```diff
@@ -58,21 +58,22 @@ def get_chart_data(data, filters):
 		return get_chart_based_on_qty(data, filters)
 
 def get_chart_based_on_status(data):
-	labels = ["Completed", "In Process", "Stopped", "Not Started"]
+	labels = ["Completed", "In Process", "Stopped", "Not Started", "Closed"]
```
ankush: Don't hardcode this? 

Use `get_meta("Work Order").get_options("status")` to get status options.

ankush (approved):

PR #28468, opened 2021-11-19
fix: coa balance rendering bug

nextchamp-saqib:
A company with a high volume of GL Entries takes time to display balance alongside the Company in the Chart of Accounts Tree.

CleanShot 2021-11-19 at 21 03 40

 There was a bug, (cannot replicate locally since a low number of GL Entries)
- User opens Chart of Accounts Tree
- Before balances has been rendered, the user expands an account
- Once the balances are rendered, due to expansion, the balance is rendered twice for unknown reasons

Fix:
Remove balances if exists before rendering on any node.


nextchamp-saqib:

PR #28469, opened 2021-11-19
fix: FY query returning None for new company

nemesis189:
**Problem:**
---
When creating a new company, Tax Withholding Categories are set up which queries for FY, which returns `None` if the FY has even a single company in its `Fiscal Year Companies` child table and the new company creation fails.  

**Solution:**
---
Removing the `company` parameter from the query fixes the problem.


deepeshgarg007:

deepeshgarg007:

PR #28471, opened 2021-11-20
fix: bug with qrcode generation due to default print format name

nextchamp-saqib:
If Default Print Format is in the Urdu language, the QR code generation fails with the below error

`UnicodeEncodeError: 'latin-1' codec can't encode characters in position 57-62: ordinal not in range(256)`

Fix:
Encode the URL parameters before creating a QRCode


deepeshgarg007:

deepeshgarg007:

PR #28497, opened 2021-11-22
fix(UX): Allocated Amount in Advances not updated on updating expense amount in Expense Claim

ruchamahabal:
**Before:**

1. Create a new Expense Claim for an employee with some Advance paid
2. On selecting the employee, the advance table fetches all the Employee Advances
3. Add items to the Expense Detail table. On entering the amount, it should update the **Grand Total** and even the **Allocated Amount** in the Advance table. Currently, the user has to do this manually.

Kapture 2021-11-22 at 16 36 14

**After:**

Trigger `calculate_grand_total` which will also call `update_employee_advance_claimed_amount` and update allocated amount in advances.

Kapture 2021-11-22 at 16 24 24

**Also fixed form labels (not capitalized):**

labels


ruchamahabal:

PR #28500, opened 2021-11-22
feat: report to see the consumed materials against the work order

rohitwaghchaure:
This report shows the information about the raw materials which has been transfer to work in progress warhouse as well consumed in manufacturing process.

work-order-consumed-materials

docs https://docs.erpnext.com/docs/v13/user/manual/en/manufacturing/reports/work-order-consumed-materials


rohitwaghchaure:

sharmanakul:
I think the batch info could also be included here.

PR #28501, opened 2021-11-22
fix: POS Item cart only taxes with amount displayed

nemesis189:
**Problem:**
---
POS Item cart would display the taxes in **Sales Charges and Taxes Template** as 0.0 even when it is iverriden by **Item Tax Template** applied to the items, it creates a clutter in the screen and is not required.
Screenshot 2021-11-22 at 6 58 43 PM

**Solution:**
---
The tax name, rate, amount would only be rendered if it has tax amount != 0
Screenshot 2021-11-22 at 6 56 52 PM


nextchamp-saqib:

nextchamp-saqib (approved):

PR #28504, opened 2021-11-22
fix: Employee link formatter showing incorrect value for Employee Name

ruchamahabal:
**Before:**

If a DocType has 2 _Employee_ Link fields and corresponding _Employee Name_ fields, the link formatter appends the _Employee Name_ value incorrectly.

fetch-from

incorrect

**After:**

Check if the value matches the employee attribute of the doc. Only then append the employee name

image


ruchamahabal:

PR #28513, opened 2021-11-23
fix: Display 'Total' before the totals row in the Gross Profit report

GangaManoj:
_Problem:_

In https://github.com/frappe/erpnext/pull/28380, I'd replaced `parent` with `invoice_or_row` as the first column when the Gross Profit report is grouped by Invoice. Around the same time, I'd added a new totals row to the report when grouped by Invoice(https://github.com/frappe/erpnext/pull/28402), which used `parent` to represent the first column. Consequently, after merging https://github.com/frappe/erpnext/pull/28380, the 'Total' in the Totals row is missing.

Screenshot 2021-11-23 at 6 55 17 AM

_Fix:_

Replace `parent` with `invoice_or_item` in the totals row.

Screenshot 2021-11-23 at 6 56 58 AM


nextchamp-saqib:

PR #28522, opened 2021-11-23
refactor: move Bin queries to qb

ankush:
- Move Bin.py queries to query builder.


ankush:
wontfix sider issues, sider's suggestions are bad for QB type chaining code and makes it less readable.

PR #24775, opened 2021-03-01
fix: reposting patch fixes

nabinhait:

PR #28523, opened 2021-11-23
fix: correct module for reloading doc

ankush:
fixes https://github.com/frappe/erpnext/pull/24775#issuecomment-976344581

### DIFF
--- a/erpnext/patches/v13_0/item_reposting_for_incorrect_sl_and_gl.py
+++ b/erpnext/patches/v13_0/item_reposting_for_incorrect_sl_and_gl.py
@@ -6,10 +6,19 @@ from erpnext.stock.stock_ledger import update_entries_after
 
 
 def execute():
-	for doctype in ('repost_item_valuation', 'stock_entry_detail', 'purchase_receipt_item',
-			'purchase_invoice_item', 'delivery_note_item', 'sales_invoice_item', 'packed_item'):
-		frappe.reload_doc('stock', 'doctype', doctype)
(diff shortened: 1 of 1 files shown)


ankush:

erpnext/patches/v13_0/item_reposting_for_incorrect_sl_and_gl.py
```diff
@@ -6,10 +6,19 @@
 
 
 def execute():
-	for doctype in ('repost_item_valuation', 'stock_entry_detail', 'purchase_receipt_item',
-			'purchase_invoice_item', 'delivery_note_item', 'sales_invoice_item', 'packed_item'):
-		frappe.reload_doc('stock', 'doctype', doctype)
-	frappe.reload_doc('buying', 'doctype', 'purchase_receipt_item_supplied')
+	doctypes_to_reload = [
```
sagarvora, line 9: Nitpicking:
This is unlikely to change, could've been a tuple. Like its constituents. 😄

ankush: Heh, should've used dictionary ideally as there are more doctypes than modules... thought about it but ultimately makes very little different as this patch likely won't change for a long time. 

Anyway I feel we need a better patch system to avoid this reloading stuff altogether.

sagarvora: I don't think there is a good alternative to reloading. Remember: we still want the old data when patching..

PR #28534, opened 2021-11-24
refactor(KSA VAT): QR Code as per ZATKA specification

ahmadpak:
Verified using the following app:
Play store
https://play.google.com/store/apps/details?id=com.posbankbh.einvoiceqrreader

App store
https://apps.apple.com/sa/app/e-invoice-qr-reader-ksa/id1580793042

image

### DIFF
--- a/erpnext/regional/saudi_arabia/utils.py
+++ b/erpnext/regional/saudi_arabia/utils.py
@@ -41,18 +41,16 @@ def create_qr_code(doc, method):
 			tlv_array = []
 			# Sellers Name
 
-			'''TODO: Fix arabic conversion'''
-			# seller_name = frappe.db.get_value(
-			# 	'Company',
-			# 	doc.company,
-			# 	'company_name_in_arabic')
+			seller_name = frappe.db.get_value(
+				'Company',
+				doc.company,
+				'company_name_in_arabic')
 
-			# if not seller_name:
-			# 	frappe.throw(_('Arabic name missing for {} in the company document'.format(doc.company)))
+			if not seller_name:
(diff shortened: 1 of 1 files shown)


erpnext/regional/saudi_arabia/utils.py
```diff
@@ -28,24 +31,76 @@ def create_qr_code(doc, method):
...
+			# 	doc.company,
+			# 	'company_name_in_arabic')
+
+			# if not seller_name:
+			# 	frappe.throw(_('Arabic name missing for {} in the company document'.format(doc.company)))
+
+			seller_name = doc.company
+			tag = bytes([1]).hex()
+			length = bytes([len(seller_name)]).hex()
+			value = seller_name.encode('utf-8').hex()
```
ahmadpak, line 54: @deepeshgarg007 This is done as per the mentioned specification in the docs but somehow does not convert Arabic name to proper hex code.

nesscale-com: @ahmadpak I think need to encode string before length function. 
			length = bytes([len(seller_name.encode('utf-8'))]).hex()

bhavesh95863: @ahmadpak I think needs to encode string before length.
			length = bytes([len(seller_name.encode('utf-8'))]).hex()

deepeshgarg007:
Merged via https://github.com/frappe/erpnext/pull/28605. Closing this PR

shiraski:
ho w to update this qr code my existing system

PR #28535, opened 2021-11-24
fix: cannot load company form

nextchamp-saqib:
Problem:
The default currency field is set as read-only if any transaction exists against the company.
If the query to `check_if_transactions_exist` takes more than 300 secs (timeout limit) the company form cannot be loaded. 

Fix:
Check for transactions later once the form has been loaded.


nextchamp-saqib:

PR #28538, opened 2021-11-24
fix: Use `get_all` instead of `get_list` for child doctype

ruchamahabal:
Student Attendance marking from the tool not working:

image

Use `get_all` instead of `get_list` for child doctype queries. This change is required to avoid the permission error due to the change made in https://github.com/frappe/frappe/pull/14526

Also fixed incorrect "fetch from" value in Course Schedule doctype for Instructor Name


ruchamahabal:
Not fixing sider issues for query builder. Not useful.

ruchamahabal:
No other instances found in:

- HR
- Payroll
- Projects
- CRM
- Education
- Non Profit

ruchamahabal:

PR #28539, opened 2021-11-24
fix: actual tax conversion in case of multicurrency invoices

nextchamp-saqib:
Steps to replicate:
- Create a multi-currency Purchase Invoice and modify the exchange rate
- Add an Item and add an Actual Tax of 10$
- Submit the Invoice
- Create Purchase Receipt using Create -> Purchase Receipt
- Check the Actual Tax Amount, it should be incorrect.

This happens because Actual Taxes are re-calculated based on the current exchange rate. And the expression for the same is incorrect.


nextchamp-saqib:

nextchamp-saqib:

PR #28550, opened 2021-11-24
refactor(SLA): Application of SLA and its fields

nextchamp-saqib:
### Problems Fixed:
- Hourly Job `set_service_level_agreement_variance` doesn't work since first parameter `doctype` is not passed
   - fix: Disabled Hourly Job
- Default Issue Priority is set as "Medium", which creates problems in finding an SLA
- Entity and Condition fields in SLA Form are confusing
    - fix: Added Dynamic Descriptions
- If "All Customer Group" is set SLA customer group, then if an Issue Customer which belongs to one of the decedent groups, SLA is not applied
- Response & Resolution Metrics doesn't consider Hold Time, if status changes from Replied to Closed
- SLA is always set as Failed in case if an Issue that was closed within SLA, is now opened after a long time.

### Refactors:
- Remove Variance Metrics
- Split Ongoing Status into First Response Due & Resolution Due to better convey the SLA Status


nextchamp-saqib:
Closed #28752

PR #28553, opened 2021-11-24
fix: Customer, Supplier heatmap data not rendering

shariquerik:
Caused by https://github.com/frappe/erpnext/pull/27302

**Issue:**
The data of heatmap for customer and supplier is not rendered.

**Reason:**
To get the timeline_data to display on heatmap `get_timeline_data` method should be in the customer.py and supplier.py
which was removed since it was considered an unused import. 

**Before:**
image

**After:**
image

To test this, test with any Customer or Supplier with some transaction history.

### DIFF
--- a/erpnext/buying/doctype/supplier/supplier.py
+++ b/erpnext/buying/doctype/supplier/supplier.py
@@ -13,8 +13,8 @@ from frappe.model.naming import set_name_by_naming_series, set_name_from_naming_
 
 from erpnext.accounts.party import (  # noqa
 	get_dashboard_info,
-	validate_party_accounts,
 	get_timeline_data,
+	validate_party_accounts,
 )
 from erpnext.utilities.transaction_base import TransactionBase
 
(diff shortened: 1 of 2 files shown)


shariquerik:
pre-commit `isort` will not allow this, suggest how to bypass it? or any better solution

```
def get_timeline_data(self):
	return get_timeline_data
```
can this be considered as a better solution?

ChillarAnand:
To ignore any line, you can add `# noqa` and it should ignore it.

```
from erpnext.accounts.party import get_dashboard_info, validate_party_accounts, get_timeline_data  # noqa
```

shariquerik:
> To ignore any line, you can add `# noqa` and it should ignore it.
> 
> ```
> from erpnext.accounts.party import get_dashboard_info, validate_party_accounts, get_timeline_data  # noqa
> ```

Still failing

ChillarAnand:
```
from erpnext.accounts.party import (  # noqa
	get_dashboard_info,
	get_timeline_data,
	validate_party_accounts,
)
```

This seems to be working. Can you try this?

shariquerik:
@ChillarAnand I found this
image
it is more descriptive

ChillarAnand (dismissed):

shariquerik:
@ChillarAnand  still failing
will try this
```
from erpnext.accounts.party import (  # noqa
	get_dashboard_info,
	get_timeline_data,
	validate_party_accounts,
)
```

ChillarAnand (approved):

PR #28555, opened 2021-11-25
fix: Validate Finished Goods for Independent Manufacture entries as well

marination:
**Issue:**
- For independent Manufacture Stock Entries (with no **From BOM** enabled and no Work Order), there was no validation if atleast one FG exists or not (atleast one must exist)
- No check for multiple FGs as well (should be blocked)
- Auto sets rows with target warehouse to scrap item because it cannot differentiate between scrap and FG because of no WO (it should not, user must manually set)

This PR fixes these.

**To Test:**
- Create a manual Stock Entry of type Manufacture, with From BOM disabled
- One item incoming (basic rate 100), one item outgoing (basic rate 200). Save. It should ask you to set atleast one FG
- Try setting both as FG it should raise error again
- Then just set one as FG and see if basic rate is recalculated based on RM.
- Also make a repack entry and see if this change is impacting it (is finished good checkbox in row should be autoset on save and cost of fg should be sum of the cost of rm)


marination:
Unrelated `test_deferred_expense_via_journal_entry` failing, fixed via https://github.com/frappe/erpnext/pull/29057

marination:

PR #28562, opened 2021-11-25
fix: patch failure due to new doctype

ankush:
v10 to latest migration is broken because prepared report doctype doesn't exist and in one of patches ERPNext tries to delete a report. 

This patch doesn't even need to be run pre-migration, long term solution: https://github.com/frappe/frappe/issues/15089


ankush:

PR #28565, opened 2021-11-25
fix: using DN for transfer w/o internal customer

fproldan:

The field `target_warehouse` depends on itself to be shown

Fix for changes introduced in this PR: https://github.com/frappe/erpnext/pull/27798

### DIFF
--- a/erpnext/stock/doctype/delivery_note_item/delivery_note_item.json
+++ b/erpnext/stock/doctype/delivery_note_item/delivery_note_item.json
@@ -468,7 +468,7 @@
    "width": "100px"
   },
   {
-   "depends_on": "eval:parent.is_internal_customer || doc.target_warehouse",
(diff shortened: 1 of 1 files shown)


ankush:
> Fix for changes introduced in this PR: #27798


That change was intentional 😅 It depends on internal customer **OR** itself so the field will be shown in case it's already set somehow. This is to avoid confusing behaviour where hidden target warehouse field is set. 


Can you describe what exactly broke after #27798?

fproldan:
I had not understood the "it's already set somehow" part. I thought the purpose of the PR was to make that field editable regardless of whether the customer or supplier is internal.

How the hidden target warehouse field can be set besides if the customer is internal?

ankush:
> How the hidden target warehouse field can be set besides if the customer is internal?


Server scripts / crazy customizations 😐

fproldan:
Okay, I understand the purpose of pr now. Sorry for the misunderstanding.

PR #28579, opened 2021-11-26
fix: KeyError for closed status

noahjacob:
### Pre-requisite
- Close a **Work Order**.

image

### Issue
image

- The error is thrown when trying to open the **Manufacturing Dashboard** if there are "Closed" Work Orders present.
-  There is no key for the "Closed" status.

### Fix
- Added the "Closed" key.


ankush:
fixed already: https://github.com/frappe/erpnext/pull/28460

PR #28580, opened 2021-11-26
feat: Bulk Transaction Processing

### LINKED ISSUE #28229: Bulk Transaction Processing
Problem Statement:
Allow to create bulk transactions to process multiple invoices/DN from Sales Order. Likewise for other common flows (PR/PI from PO).

Expected Solution:
Use bulk actions in list view to give options to create transactions in one go. This will save time in processing multiple transactions as opposed to one by one by UI, or Excel import.

mohammedyusufshaikh:
**Before this Feature**
Lets take an example to get better idea
1. If the user had to create multiple sales invoice from multiple sales order it had to be done one by one by going into the   particular sales order and then create sales invoice against it which consumed a lot of time.Similarly this was the case for many doctype's for which creating multiple docs from multiple docs was time consuming.
2. Problem was identified from this issue #28229 

**Tasks**
- [x] Notification to the user when job is completed and some minor tasks.
- [x] Add Logger For Transactions
- [x] Tests
- [x] Increase Code Coverage for codecov/project

Following operations can be performed in bulk:

Screenshot 2021-11-26 at 1 49 25 PM

**After this Feature**
1. To tackle this problem options are added into the actions menu. so when the user selects multiple invoices/orders the actions menu button appears from which user can select the bulk operation to be performed. thus saving a lot of time.
2. Since creation of multiple invoices/orders can take a lot of time it should not block user from doing other operations. so if there are more than 10 invoices/orders it will create a background job for the same.
3.  There might be situation where in while bulk creating of invoices/orders some exceptions might occur which might halt the creation of invoices/orders. There is a mechanism to bypass failing invoices/orders and continue forward creating other invoices/orders in the queue.

bulk_transaction_processing


erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+	from erpnext.stock.doctype.delivery_note import delivery_note
+	from erpnext.stock.doctype.purchase_receipt import purchase_receipt
+
+	i = 0
+	for d in deserialized_data:
+		try:
+			i+=1
+
+			# From Sales Order
+			if to_create == "Sales Invoice From Sales Order":
```
ankush: Too much code duplication here?

Python allows treating functions as first class objects, just create a mapping like this instead: 


`Tuple[from_doctype, to_doctype] -> Function`


If a particular function needs non-standard or extra default arguments you can create a partial function: https://docs.python.org/3/library/functools.html#functools.partial

ankush:
300 sider issues 😱 Just pass new files through something like https://prettier.io/ to auto format them.

ankush:
reopen after resolving open issues + some basic tests. 


JS code also seems to be duplicated across many different doctype.js files. Maybe create a JS util class / function somewhere?

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,138 @@
...
+				"Sales Order": {"Sales Invoice": sales_order.make_sales_invoice, "Delivery Note": sales_order.make_delivery_note, "Advance Payment": payment_entry.get_payment_entry},
+				"Sales Invoice": {"Delivery Note": sales_invoice.make_delivery_note, "Payment": payment_entry.get_payment_entry},
+				"Delivery Note": {"Sales Invoice": delivery_note.make_sales_invoice, "Packing Slip": delivery_note.make_packing_slip},
+				"Quotation": {"Sales Order": quotation.make_sales_order, "Sales Invoice": quotation.make_sales_invoice},
+				"Supplier Quotation": {"Purchase Order": supplier_quotation.make_purchase_order, "Purchase Invoice": supplier_quotation.make_purchase_invoice, "Advance Payment": payment_entry.get_payment_entry},
+				"Purchase Order": {"Purchase Invoice": purchase_order.make_purchase_invoice, "Purchase Receipt": purchase_order.make_purchase_receipt},
+				"Purhcase Invoice": {"Purchase Receipt": purchase_invoice.make_purchase_receipt, "Payment": payment_entry.get_payment_entry},
+				"Purchase Receipt": {"Purchase Invoice": purchase_receipt.make_purchase_invoice}
+			}
+	if to_doctype == "Advance Payment" or to_doctype == "Payment":
```
rohitwaghchaure: if to_doctype in ['Advance Payment', 'Payment']

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,138 @@
...
+	i = 0
+	for d in deserialized_data:
+		failed = []
+
+		try:
+			i+=1
+			doc_name = d.get('name')
+			task(doc_name, from_doctype, to_doctype)
+
+		except Exception as e:
```
ankush, line 44: This has potential commit failed data. 

E.g. if exception is thrown in `on_submit` method, a lot of data is already written, docstatus already changed etc etc that has to be rolledback in case of exception. 

Either use commit + rollback or new feature of savepoints: https://github.com/frappe/frappe/pull/15379

ankush (changes requested):

erpnext/utilities/bulk_transaction.py
mohammedyusufshaikh: Did the changes

erpnext/utilities/bulk_transaction.py
mohammedyusufshaikh: Did the changes

olamide2:
Eye on this

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.js
```diff
@@ -0,0 +1,21 @@
...
+				frappe.call({
+					method: "erpnext.bulk_transaction.doctype.bulk_transaction_log.bulk_transaction_log.retry_failing_transaction",
+					args: {}
+				}).then(() => {
+					frappe.show_alert(__("Retrying Failed Transactions"), 5);
+				});
+			},
+			()=>{
+
+			});
```
ankush:
```suggestion
```

dead code ?

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
```diff
@@ -0,0 +1,51 @@
...
+@frappe.whitelist()
+def retry_failing_transaction():
+    btp = frappe.qb.DocType("Bulk Transaction Log Detail")
+    data = (
+        frappe.qb.from_(btp)
+        .select(btp.transaction_name, btp.from_doctype, btp.to_doctype)
+        .distinct()
+        .where(btp.retried != 1)
+        .where(btp.transaction_status == "Failed")
+        .where(btp.date == fn.CurDate())
```
ankush: You can only retry what failed today? Any reason for this restriction? 🤔

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,194 @@
...
+			"Payment": payment_entry.get_payment_entry,
+		},
+		"Purchase Receipt": {"Purchase Invoice": purchase_receipt.make_purchase_invoice},
+	}
+	if to_doctype in ['Advance Payment', 'Payment']:
+		obj = mapper[from_doctype][to_doctype](from_doctype, doc_name)
+	else:
+		obj = mapper[from_doctype][to_doctype](doc_name)
+
+	obj.flags.ignore_validate = True
```
ankush, line 104: Ignoring validation is VERY dangerous. It also ignores running the `validate` method that does a lot of preprocessing e.g. setting value from rate and qty.  Check `validate` method of any doctype bulk transaction is being applied on, is it safe to not run all those things?

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,194 @@
...
+	if not failed_history:
+		frappe.msgprint(
+			"Creation of {0} Successfull".format(to_doctype),
+			title="Successfull",
+			indicator="green",
+		)
+
+	if len(failed_history) != 0 and len(failed_history) < len(deserialized_data):
+		frappe.msgprint(
+			"""Creation of {0} partially Successfull.
```
ankush: Missing translations.

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,194 @@
...
+			"""Creation of {0} Failed.
+				Check <b><a href="/app/bulk-transaction-log">Bulk Transaction Log</a></b>""".format(
+				to_doctype
+			),
+			title="Failed",
+			indicator="red",
+		)
+
+
+def record_exists(log_doc, doc_name, status):
```
ankush, line 180: This feels like a method on the object where first argument can be `self`. Any reason why this is a separate function?

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,194 @@
...
+	record = 0
+	for d in log_doc.get("logger_data"):
+		if d.transaction_name == doc_name and d.transaction_status == "Failed":
+			d.retried = 1
+			record = record + 1
+
+	log_doc.save()
```
ankush: Contrary to what function name `record_exists` suggests, this seems to do a lot of write operations. Maybe split these two things?

ankush (changes requested):

erpnext/bulk_transaction/doctype/bulk_transaction_log/test_bulk_transaction_log.py
```diff
@@ -0,0 +1,106 @@
...
+		transaction_processing(data, "Sales Order", "Sales Invoice")
+		retry_failing_transaction()
+		doc = frappe.get_doc("Bulk Transaction Log", str(date.today()))
+		for d in doc.get("logger_data"):
+			if d.get('transaction_name') == data[0]["name"]:
+				self.assertEqual(d.get('transaction_status'), "Failed")
+				self.assertEqual(d.retried, 1)
+
+def create_company():
+	if not frappe.db.exists('Company', 'Test Bulk'):
```
ankush: reuse some existing company, there are 20 in default test records. Creating a company with many fixtures takes a lot of time in tests; I don't think it's required in this case.

erpnext/public/js/bulk_transaction_processing.js
```diff
@@ -0,0 +1,30 @@
...
+					method: "erpnext.utilities.bulk_transaction.transaction_processing",
+					args: {data: checked_items, from_doctype: from_doctype, to_doctype: to_doctype}
+				}).then(()=> {
+
+				});
+				if (count_of_rows > 10) {
+					frappe.show_alert("Starting a background job to create {0} {1}", [count_of_rows, to_doctype]);
+				}
+			} else {
+				frappe.msgprint(__("Selected document must be in submitted state !"));
```
ankush:
```suggestion
				frappe.msgprint(__("Selected document must be submitted"));
```

We almost never need to use exclamation mark in user facing messages 😄

erpnext/utilities/bulk_transaction.py
ankush:
```suggestion
			"""Creation of {0} partially successful.
```

Typo; in other places too.

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
ankush: This is for retrying all failing transaction log from list view; IMO
1. It should signify it will only attempt today's failed transaction. 
2. Single log should also have retry. (feature request)

erpnext/utilities/bulk_transaction.py
ankush: Another reason for not skipping this is that `validate` function also populates dependent fields and it can lead to weird behavior where `validate` populates some field and if it's invalid it will give an error that's not visible in the document since it gets rolled back again.


In addition to that if something causes a document to fail _first save_ maybe that root cause needs to be fixed for better UX. Any examples?

ankush: Also frequently we end up inspecting how document is created ("was it created by list view" or "was it submitted by data import") etc cause they do things a little bit differently which can introduce a bug as existing code doesn't think it can happen. 

So IMO it's better to keep document creation similar to how it would've been if user manually created them using `"create > doctype"`


@rohitwaghchaure

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
mohammedyusufshaikh: In the list view it shows all the log docs with their name set as current date. It doesn't show all the transactions at once. so the user can retry that specific date's failing transaction
Screenshot 2022-01-27 at 12 34 36 PM

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
```diff
@@ -0,0 +1,51 @@
...
+    else:
+        job(data)
+
+
+def job(data):
+    for d in data:
+        failed = []
+        try:
+            task(d.transaction_name, d.from_doctype, d.to_doctype)
+        except Exception as e:
```
ankush: this will commit partial data, use savepoints/rollback like the other function which is also named job. Also possibly duplicate code?

erpnext/bulk_transaction/doctype/bulk_transaction_log/test_bulk_transaction_log.py
```diff
@@ -0,0 +1,106 @@
...
+	so.transaction_date = date.today()
+	so.append("items", {
+		"item_code": "MK",
+		"delivery_date": date.today(),
+		"qty": 10,
+		"rate": 80,
+	})
+
+	so.flags.ignore_validate = True
+	so.insert(ignore_mandatory=True)
```
ankush: not sure why in tests we need to ignore these things.

cypress/integration/test_bulk_transaction_processing.js
```diff
@@ -0,0 +1,44 @@
...
+		cy.get(".actions-btn-group > .btn-primary").click({ force: true });
+		cy.wait(3000);
+		cy.get(".dropdown-menu-right > .user-action > .dropdown-item")
+			.contains("Sales Invoice")
+			.click({ force: true });
+		cy.wait(3000);
+		cy.get(".modal-content > .modal-footer > .standard-actions")
+			.contains("Yes")
+			.click({ force: true });
+		cy.contains("Creation of Sales Invoice Successfull");
```
ankush:
```suggestion
		cy.contains("Creation of Sales Invoice successful");
```

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.js
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2021, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Bulk Transaction Log', {
+
+	refresh: function(frm) {
+		frm.add_custom_button(__('Retry Transactions'), ()=>{
```
ankush: - This shouldn't be shown if there's nothing failed. 
- The button is present on document indicating the whole document will be retried, though it's hardcoded to only retry today's failures. Very bad UX IMO. 

Screenshot 2022-02-02 at 5 51 50 PM

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,202 @@
...
+			indicator="green",
+		)
+
+	if len(failed_history) != 0 and len(failed_history) < len(deserialized_data):
+		frappe.msgprint(
+			_("""Creation of {0} partially successfull.
+				Check <b><a href="/app/bulk-transaction-log">Bulk Transaction Log</a></b>""").format(
+				to_doctype
+			),
+			title="Partially Successfull",
```
ankush:
```suggestion
			title="Partially successful",
```

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,202 @@
...
+	if not failed_history:
+		frappe.msgprint(
+			_("Creation of {0} successfull").format(to_doctype),
+			title="Successfull",
+			indicator="green",
+		)
+
+	if len(failed_history) != 0 and len(failed_history) < len(deserialized_data):
+		frappe.msgprint(
+			_("""Creation of {0} partially successfull.
```
ankush:
```suggestion
			_("""Creation of {0} partially successful.
```

erpnext/utilities/bulk_transaction.py
```diff
@@ -0,0 +1,202 @@
...
+			append_data_to_logger(
+				log_doc, doc_name, e, from_doctype, to_doctype, status, restarted
+			)
+			log_doc.save()
+
+
+def show_job_status(failed_history, deserialized_data, to_doctype):
+	if not failed_history:
+		frappe.msgprint(
+			_("Creation of {0} successfull").format(to_doctype),
```
ankush:
```suggestion
			_("Creation of {0} successful").format(to_doctype),
```

ankush (changes requested):

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
ankush: > It doesn't show all the transactions at once. so the user can retry that specific date's failing transaction

Can they though? The query only appears to retry today's failed transactions.

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
```diff
@@ -0,0 +1,58 @@
...
+
+class BulkTransactionLog(Document):
+    pass
+
+
+@frappe.whitelist()
+def retry_failing_transaction(log_date):
+    btp = frappe.qb.DocType("Bulk Transaction Log Detail")
+    data = (
+        frappe.qb.from_(btp)
```
ankush: This file appears to be using spaces. We use tabs everywhere. Mixing them can cause failures.

ankush (changes requested):

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
mohammedyusufshaikh: yes I will change it

mohammedyusufshaikh: Done

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
mohammedyusufshaikh: Added save points and rollback

erpnext/bulk_transaction/doctype/bulk_transaction_log/bulk_transaction_log.py
mohammedyusufshaikh: No it did not but now I have made the changes to retry that specific date's transaction

PR #28588, opened 2021-11-27
fix: incorrect bin qty on backdated reconciliation

ankush:
Bin quantities are updated by adding actual_qty from current SLE to
Bin's qty. In stock reconciliation qty is "reset" hence
qty_after_transaction is used to set qty in bin.

However, in the case of backdated stock reconciliation, there are future
transactions that will affect the final sum of quantity available in bin.

Note: This gets auto-corrected after reposting finishes for backdated
transactions, but Bin shows wrong quantity till then. Bin is used in item
dashboard hence it's confusing. Also, reposting is only supposed to be
for valuation, qty should be correct in real-time.

Solution: Instead of overriding Bin's quantity, replace it with last sle qty in case of backdated transactions.

To test:

1. Pick one item - warehouse combination.
2. Create some stock transactions.
3. Open Bin and verify that current quantity is correct.
4. Create a backdated stock reconciliation.
5. Confirm that Bin's quantity matches with stock ledger and stock
   balance. (earlier this used to get reset to whatever stock reco
   quantity was)
6. Cancel the stock reconciliation and verify stock ledger/balance
   again.

Example transactions ordered by creation; more test cases on  https://github.com/frappe/erpnext/pull/26158  

| Transaction | Posting date | Qty | qty in Bin |
| --- | ---  | --- | --- |
| Create A | 0    | +1 | 1 |
| Create B | -1 | 2 (reset via stock reco) | 3  (this used to be 2 before)| 
| Cancel B | -1 | 0 | 1 (this used to be 0 before) |

### DIFF
--- a/erpnext/stock/doctype/bin/bin.py
+++ b/erpnext/stock/doctype/bin/bin.py
@@ -130,8 +130,8 @@ def update_stock(bin_name, args, allow_negative_stock=False, via_landed_cost_vou
 	"""WARNING: This function is deprecated. Inline this function instead of using it."""
 	from erpnext.stock.stock_ledger import repost_current_voucher
 
-	update_qty(bin_name, args)
 	repost_current_voucher(args, allow_negative_stock, via_landed_cost_voucher)
+	update_qty(bin_name, args)
 
 def get_bin_details(bin_name):
 	return frappe.db.get_value('Bin', bin_name, ['actual_qty', 'ordered_qty',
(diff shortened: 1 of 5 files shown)


ankush:
this (and all other cases of bin update) are broken in the case of intermediate stock reconciliation.  refer: https://github.com/frappe/erpnext/pull/26158

erpnext/stock/doctype/bin/bin.py
```diff
@@ -130,19 +130,30 @@ def update_stock(bin_name, args, allow_negative_stock=False, via_landed_cost_vou
...
 	'reserved_qty_for_sub_contract'], as_dict=1)
 
 def update_qty(bin_name, args):
+	from erpnext.controllers.stock_controller import future_sle_exists
+
 	bin_details = get_bin_details(bin_name)
 
-	# update the stock values (for current quantities)
-	if args.get("voucher_type")=="Stock Reconciliation":
+	if future_sle_exists(args):
```
ankush, line 149: This is internally cached. So won't have any hit on perf for normal usecases (i.e. only backdated entries will cause extra query)

ankush:
Tests failing because actual qty is now getting picked from last (correct) sle 😐

ankush:
uncovered lines intentionally don't run in the tests. coverage is 100% actually 😄

PR #28589, opened 2021-11-27
fix(Timesheets): calculate to time

hrwX:
- When time logs are being entered in timesheets, when you enter the `from time` and `hours`, the `to time` will be calculated.
- But this functionality is missing if you create timesheets using an API.
- This PR fixes that by checking if `to time` is empty and calculates it based off `from time` and `hours`.


erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -7,7 +7,7 @@
 import frappe
 from frappe import _
 from frappe.model.document import Document
-from frappe.utils import flt, getdate, time_diff_in_hours
+from frappe.utils import flt, getdate, time_diff_in_hours, add_to_date
```
ankush:
```suggestion
from frappe.utils import add_to_date, flt, getdate, time_diff_in_hours
```

Use isort/precommit 😄

erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -136,10 +136,15 @@ def validate_dates(self):
 
 	def validate_time_logs(self):
 		for data in self.get('time_logs'):
+			self.set_to_time(data)
 			self.validate_overlap(data)
 			self.set_project(data)
 			self.validate_project(data)
 
+	def set_to_time(self, data):
+		if data.from_time and data.hours and not data.to_time == add_to_date(data.from_time, hours=data.hours):
```
ankush: This condition will almost never be true? Because `to_time` will be string and `add_to_date` returns datetime object. Maybe use `get_datetime`?

barredterra (commented):
@hrwX kindly add a test to cover this change.

erpnext/projects/doctype/timesheet/timesheet.py
```diff
@@ -136,10 +136,19 @@ def validate_dates(self):
...
 			self.set_project(data)
 			self.validate_project(data)
 
+	def set_to_time(self, data):
+		if not (data.from_time and data.hours):
+			return
+
+		_to_date = add_to_date(data.from_time, hours=data.hours, as_datetime=True)
+		if not data.to_time == _to_date:
+			data.to_time = _to_date
```
barredterra:
```suggestion
		_to_time = add_to_date(data.from_time, hours=data.hours, as_datetime=True)
		if data.to_time != _to_time:
			data.to_time = _to_time
```

barredterra:

barredterra (approved):

barredterra:
@ankush can we get this merged?

ruchamahabal (approved):

PR #28592, opened 2021-11-27
fix: do not add gst fields if no indian company

hrwX:
- Fixes the bug introduced with #26129
The patch for PR doesn't check if there is any company with country India that exists in the system while running the patch; due to this, the Payment Entry has a GST section irrespective of whether a company with country India exists in the system.


hrwX:

deepeshgarg007 (approved):

PR #28607, opened 2021-11-29
fix: updated packed_items getting fetched on Sales Return / Credit Note

noahjacob:
### Issue (Delivery Note / Sales Invoice)
- When creating a Sales Return from a **Delivery Note**, if the **Delivery Note** contains a **Product Bundle** item, on saving the return document, the updated packed_items are fetched instead of the packed_items from the **Delivery Note**.

<details>
<summary>Product Bundles</summary>

### Product Bundle Before Creating Delivery Note
image

### Updated Product Bundle

image

</details>

return_map

### Fix (Delivery Note and Sales Invoice)
- If document is a Return, mapped packed_items are kept as it is instead of fetching packed_items from **Product Bundle**.
- Packed_items qtys are recalculated if the item qty is changed.

return_map_fix

### Steps (Similar Steps for Credit Note)
- Create a **Product Bundle**.
- Create a **Delivery Note** with the **Product Bundle**.
- Update the **Product Bundle**.
- Create a "Sales Return" via the **Delivery Note**.
- Save the "Sales Return" document.

> edit by @ankush 

> More Test cases:
> - Multiple product bundles
> - partial returns (e.g. only 1 bundle returned from 2 delivered)


ankush:
Product bundle should be a submittable doctype with versioning like BOM! 😐

ankush:
This would apply to Sales Invoice with `update_stock` too right? Can you check and fix both cases in one PR then?

Perhaps this is best handled in that util function instead of a single doctype 😅

noahjacob:
> This would apply to Sales Invoice with `update_stock` too right? Can you check and fix both cases in one PR then?
> 
> Perhaps this is best handled in that util function instead of a single doctype 😅

Moved it to stock.utils and did the same for Credit Note.

erpnext/stock/utils.py
```diff
@@ -416,3 +416,12 @@ def is_reposting_item_valuation_in_progress():
 		{'docstatus': 1, 'status': ['in', ['Queued','In Progress']]})
 	if reposting_in_progress:
 		frappe.msgprint(_("Item valuation reposting in progress. Report might show incorrect item valuation."), alert=1)
+
+def calculate_mapped_packed_items_return(self):
```
ankush: function with `self` as argument? 😄

erpnext/stock/utils.py
```diff
@@ -419,6 +419,17 @@ def is_reposting_item_valuation_in_progress():
 	if reposting_in_progress:
 		frappe.msgprint(_("Item valuation reposting in progress. Report might show incorrect item valuation."), alert=1)
 
+
+def calculate_mapped_packed_items_return(return_doc):
+	parent_items = set([item.parent_item for item in return_doc.packed_items])
+	against_doc = frappe.get_doc(return_doc.doctype, return_doc.return_against)
+	for d_item, item in zip(against_doc.items, return_doc.items):
+		if d_item.item_code in parent_items:
+			for p_item, d_p_item in zip(return_doc.packed_items, against_doc.packed_items):
```
ankush, line 428: If you have two bundle this for loop adds last bundle to all references. It's not being checked which item they actually belong to here.

ankush: If you have two bundle this for loop adds last bundle to all references. It's not being checked which item they actually belong to here.

original

Screenshot 2022-01-21 at 11 08 23 AM



returned:
Screenshot 2022-01-21 at 11 08 44 AM

ankush (changes requested):

ankush (approved):

PR #28610, opened 2021-11-29
fix: Duplicate row gets added when "Get Items From" material request in Purchase Order

s-aga-r:
**Source/Reference: FR-ISS-314060**

**Issue(s): Purchase Order**

1. Duplicate row gets added when "Get Items From" same Material Request.

https://user-images.githubusercontent.com/63660334/143840476-3ef8603b-98b6-4d04-a7ad-3b41a1c21753.mp4

**Fix:**

1. Passed new parameters "row_id_fieldname" and "remove_common_rows" from purchase_order.py to "map_current_doc" in "utils.js".
2. Remove common row(s) by getting it's unique row-id from "opts.args.filtered_children".

**Result:**

https://user-images.githubusercontent.com/63660334/143844465-4aecd7d1-e48a-4c50-83b9-38ba40e31ec0.mov


erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -428,7 +428,8 @@ erpnext.buying.PurchaseOrderController = class PurchaseOrderController extends e
 					},
 					allow_child_item_selection: true,
 					child_fielname: "items",
-					child_columns: ["item_code", "qty"]
+					child_columns: ["item_code", "qty"],
```
ankush, line 431: `item_code` can appear twice because we allow repeating same item. Here row's `name` needs to be checked.

ankush (changes requested):

erpnext/buying/doctype/purchase_order/purchase_order.js
s-aga-r: We can add the same item more than once but cannot add an item twice from the same Material Request.
Screenshot 2021-11-29 at 5 24 48 PM

ankush: That's not what I meant. A material request can have multiple rows for same item code, so we shouldn't be checking for duplicate using `item_code`. Use `material_request_item` which is hash of row. 

Screenshot 2021-11-30 at 6 33 55 PM

ankush (changes requested):

erpnext/public/js/utils.js
```diff
@@ -629,41 +629,57 @@ erpnext.utils.map_current_doc = function(opts) {
...
-						already_set = true;
-						if (item_qty_map[d.item_code])
-							item_qty_map[d.item_code] += flt(d.qty);
-						else
-							item_qty_map[d.item_code] = flt(d.qty);
-					}
+			if (opts.remove_common_rows) {
+				$.each(cur_frm.doc.items, (parent_idx, parent_val) => {
+					$.each(opts.args.filtered_children, (child_idx, child_val) => {
+						if (parent_val[link_fieldname + "_item"] == child_val) {
```
s-aga-r: @ankush Here i am using "material_request_item" or "{link_fieldname}_item" to get the row hash.

ankush: We don't maintain this naming everywhere, hence we need to pass the actual field name 😅 

Check other doctypes like sales invoice for example. The function you're modifying is used by 10+ other doctypes.

s-aga-r: I understood, Btw It will work fine for the Sale Invoice 😅

erpnext/public/js/utils.js
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 // License: GNU General Public License v3. See license.txt
-frappe.provide("erpnext");
+frappe.provide("erpnext ");
```
ankush:
```suggestion
frappe.provide("erpnext");
```
why is this changed?

erpnext/buying/doctype/purchase_order/purchase_order.js
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2015, Frappe Technologies Pvt. Ltd. and Contributors
 // License: GNU General Public License v3. See license.txt
 
-frappe.provide("erpnext.buying");
+frappe.provide("erpnext.buying ");
```
ankush:
```suggestion
frappe.provide("erpnext.buying");
```

ankush (changes requested):

PR #28611, opened 2021-11-29
fix: changed fieldtype from int to float for the field Batch Size in the Work Order Operation table

rohitwaghchaure:
Screenshot 2021-11-29 at 4 41 47 PM


rohitwaghchaure:

PR #28613, opened 2021-11-29
fix: allow creating Shift Assignment for same day

ruchamahabal:
There is a validation that does not allow Shift Assignment creation with the same **Start Date** and **End Date**:

image

Removed it so that users can create Shift Assignments for the same day.


ruchamahabal:

PR #28627, opened 2021-11-29
fix: Paid invoices showing in AR/AP report

deepeshgarg007:
Paid Multicurrency Invoices still showing up in AR/AP report

Steps to test:
1. Make a multi-currency sales invoice from INR to USD at an exchange rate of 70
Screenshot 2021-12-02 at 6 40 30 PM

2. Make a standalone payment against the same customer at an exchange rate of 65
Screenshot 2021-12-02 at 6 44 04 PM

3. Reconcile this payment against the Invoice using the payment reconciliation tool
Screenshot 2021-12-02 at 6 44 42 PM

The status Of the invoice will be changed to paid but it is still visible in the Accounts Payable Report if you'll check using the inline data table filters

Screenshot 2021-12-02 at 7 03 26 PM

Solution: Check for outstanding amount in both company and party currency as well


deepeshgarg007:
@Mergifyio backport develop

aakvatech:
This problem will be come back from Journals as well that are affecting party. It is because of a need to have adjustment journal to adjust the exchange gain and loss balance that can occur due to party paying in different currency than the invoiced currency and there is an exchange rate difference resulting into a decimal value.

Not sure of the solution but will create separate issue so it can be handled in next sprint.

PR #28636, opened 2021-11-30
feat: Show Zero Values filter in consolidated financial statement

deepeshgarg007:
Added a filter to filter zero value rows in the Consolidated financial statement

Screenshot 2021-11-30 at 1 32 33 PM

From now zero value rows won't be seen by default, only after enabling this filter, zero value rows will be visible

On Enablig the filter:
Screenshot 2021-11-30 at 1 37 24 PM

On Disabling the filter:
Screenshot 2021-11-30 at 1 37 57 PM

`no-docs`


deepeshgarg007:

PR #28638, opened 2021-11-30
fix(Employee): enable no copy for lft rgt fields

ruchamahabal:
**Problem**

1. Create an employee - emp1
2. Duplicate the Document to create another employee - emp2
3. Set "Reports To" field in emp1 as emp2
4. NestedSet throws an error while validating the loop

image

**Fix**:

When you duplicate the employee the Left Index and Right Index fields also get copied so when you try to rebuild the hierarchy by setting Reports To, it will throw this error as lft and rgt are same for both the docs.

Enabled no copy for these fields

### DIFF
--- a/erpnext/hr/doctype/employee/employee.json
+++ b/erpnext/hr/doctype/employee/employee.json
@@ -726,6 +726,7 @@
    "fieldtype": "Int",
    "hidden": 1,
    "label": "lft",
+   "no_copy": 1,
    "read_only": 1
   },
   {
@@ -733,6 +734,7 @@
    "fieldtype": "Int",
    "hidden": 1,
    "label": "rgt",
+   "no_copy": 1,
    "read_only": 1
   },
   {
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/employee/employee.json
```diff
@@ -726,13 +726,15 @@
    "fieldtype": "Int",
    "hidden": 1,
    "label": "lft",
+   "no_copy": 1,
```
ankush, line 729: This should go to framework right? It never makes sense to copy lft/rgt.

ruchamahabal: The framework already handles this while adding nested set fields with no copy enabled when you enable **Is Tree** for doctypes

https://github.com/frappe/frappe/blob/76cc6bf10b6c981dd4c0e0d7e86a28937ce4bb06/frappe/core/doctype/doctype/doctype.py#L629-L651

Earlier making tree view doctypes wasn't configurable and these fields were manually added in most of the doctypes. Same with Employee.

Will enable no copy for other tree doctypes in ERPNext too.

ruchamahabal: Fixed in copy doc: https://github.com/frappe/frappe/pull/15128

But merging this anyway.

ruchamahabal:
Fixed in copy doc: frappe/frappe#15128

PR #28648, opened 2021-11-30
fix: Unable to search project by project name and other search fields in Sales Invoice

### LINKED ISSUE #27345: Include search fields in Project Link to other Transaction
## Description of the issue
1. Search field of a Project Link is not visible to other transactions except when linking only to a Sales Order.

image

But search field is not visible on Delivery Note Sales Invoice and the Buying Modules.

image

image

image

**Output of `bench version`**
```
ERPNext: v13.9.2 (HEAD)
Frappe Framework: v13.9.1 (HEAD)

this is also a bug on V12
```

deepeshgarg007:
In accounting doctypes like sales invoice for standard accounting dimensions like project and cost center a custom query is used, this query didn't use to consider searchfields which made it difficult for users to select links like project where most of the projects have naming series and is difficult for the user to know which project is what

Before:
Screenshot 2021-11-30 at 6 29 44 PM

After:
image

closes https://github.com/frappe/erpnext/issues/27345

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.js
@@ -517,15 +517,6 @@ cur_frm.fields_dict.write_off_cost_center.get_query = function(doc) {
 	}
 }
 
-// project name
-//--------------------------
-cur_frm.fields_dict['project'].get_query = function(doc, cdt, cdn) {
-	return{
-		query: "erpnext.controllers.queries.get_project_name",
-		filters: {'customer': doc.customer}
-	}
-}
-
 // Income Account in Details Table
 // --------------------------------
 cur_frm.set_query("income_account", "items", function(doc) {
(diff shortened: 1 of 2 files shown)


ankush:
@mergifyio backport develop

erpnext/accounts/doctype/sales_invoice/sales_invoice.js
```diff
@@ -517,15 +517,6 @@ cur_frm.fields_dict.write_off_cost_center.get_query = function(doc) {
 	}
 }
 
-// project name
-//--------------------------
-cur_frm.fields_dict['project'].get_query = function(doc, cdt, cdn) {
-	return{
-		query: "erpnext.controllers.queries.get_project_name",
```
ankush, line 524: Why remove this? maybe I don't have context here... is there any other place where query function is getting set?

deepeshgarg007: For accounting doctypes like sales invoice, `get_filtered_dimensions` is used. Since project and cost center are standard accounting dimensions, even for these two this is used in the sales invoice. Have kept the query because its still being called from Sales Order

ankush (approved):

deepeshgarg007:

PR #28656, opened 2021-11-30
fix: Taxjar Nexus list visible only if child table is visible

nemesis189:
**Issue:**
---
Taxjar nexus list would display blank section when empty. It is visible only when the child table is populated which happens only when the Update Nexus list button gets pressed

**Fix:**
---
- The section is only visible if the child table is visible


nextchamp-saqib (approved):

deepeshgarg007:

PR #28657, opened 2021-11-30
fix: update `modified` timestamp for delivery note

sagarvora:
For consistency with https://github.com/frappe/erpnext/pull/28645/commits/8f5b1ea3d2b809cb260bc5ce3b6899ab988fbf2d
This was inadvertently not updated in #27467, but I guess this is no longer a requirement with https://github.com/frappe/frappe/pull/14246

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.json
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.json
@@ -1315,7 +1315,7 @@
  "idx": 146,
  "is_submittable": 1,
  "links": [],
- "modified": "2021-10-08 14:29:13.428984",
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/delivery_note/delivery_note.json
```diff
@@ -1315,7 +1315,7 @@
  "idx": 146,
  "is_submittable": 1,
  "links": [],
- "modified": "2021-10-08 14:29:13.428984",
+ "modified": "2021-10-09 14:29:13.428984",
```
ankush, line 1318:
```suggestion
 "modified": "2021-12-01 14:29:13.428984",
```

PR #28658, opened 2021-11-30
fix: get_items method in stock_entry shouldnt raise error

mro-admin:
Issue: Some front end dialog boxes that creates a draft Stock Entry errors out because the stock_entry.get_items() method raises an error if an item has no valuation rate. An example of this is the "Start" button in Work Order which if used and a required item has no valuation rate will raise an error which will then lead to the stock entry draft not being created

Fix: set raise_error_if_no_rate=False . This will not affect the validation when a stock entry is submitted

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -1072,7 +1072,7 @@ class StockEntry(StockController):
 		self.set_actual_qty()
 		self.update_items_for_process_loss()
 		self.validate_customer_provided_item()
-		self.calculate_rate_and_amount()
+		self.calculate_rate_and_amount(raise_error_if_no_rate=False)
 
 	def set_scrap_items(self):
 		if self.purpose != "Send to Subcontractor" and self.purpose in ["Manufacture", "Repack"]:


erpnext/stock/doctype/stock_entry/stock_entry.py
```diff
@@ -1072,7 +1072,7 @@ def get_items(self):
 		self.set_actual_qty()
 		self.update_items_for_process_loss()
 		self.validate_customer_provided_item()
-		self.calculate_rate_and_amount()
+		self.calculate_rate_and_amount(raise_error_if_no_rate=False)
```
ankush, line 1075: I don't think this is a good solution to the problem you described. This will just delay the error till the last moment of submission, even when doing normal stock entries.  

When is the valuation rate not available? Mostly it's when the item is newly added but doesn't have any stock. This makes sense as you shouldn't be manufacturing using RM which isn't in stock. If it was in stock most it will have a valuation rate. Unable to save draft stock entry isn't a big deal here IMO. 


Any other use case where this is breaking?

ankush (changes requested):

ankush:
Closing, as I don't think delaying this validation till the last step is a good idea from UX perspective. Let me know if there are other use valid use cases and maybe we can think of alternative solutions for them.

PR #28667, opened 2021-12-01
test: dynamic Fiscal Year creation in tests

ankush:
Few tests are failing because FY 2021-2022 doesn't exist and reports are trying to fetch FY of 1 month from today. 

Root cause: FY up to 2021 were hardcoded in test records 🥲 so they were scheduled to fail today. 
Fix: Dynamic FY for test records, from 2011 to (today + 4 years)


gavindsouza:
Update `.gitignore` too? 🤔

ankush:
> Update .gitignore too? 🤔


 update what exactly? 🤔 😕

gavindsouza:
Oops. My bad

PR #28673, opened 2021-12-01
fix: misleading "Set Default X" fields after saving

### LINKED ISSUE #28644: All "set default value" fields are sort of confusing
There are many fields like "set default warehouse", "set delivery date" etc etc which have NO functional use other than setting a default on the form (client-side) and maybe server-side too. These defaults are propagated to child items but they can be edited and can cause very confusing behavior. 

One example:
Screenshot 2021-11-30 at 5 25 56 PM

s-aga-r:
### Issue(s): Purchase Receipt

_Problem:_
- When populating the "Accepted Warehouse" field it adds that warehouse as the default warehouse on all the item rows. This is helpful because often the same warehouse for all the items received. But if the user changes the warehouse for an item. In that case, this field becomes confusing as it is holding another warehouse.

_Proposed Solution:_
- If all warehouses in the child table(Items) are not matching the "default warehouse" field then set its value to None.

<details>
<summary>Images/GIF</summary>

_Before:_

https://user-images.githubusercontent.com/63660334/144282361-d6e8ef46-ae48-42ed-a86c-8136cf993b45.gif

_After:_

https://user-images.githubusercontent.com/25903035/130674996-a6d891e4-4220-4574-87fc-5b32bca6992a.png
</details>

Closes: https://github.com/frappe/erpnext/issues/28644


erpnext/stock/utils.py
```diff
@@ -416,3 +416,15 @@ def is_reposting_item_valuation_in_progress():
 		{'docstatus': 1, 'status': ['in', ['Queued','In Progress']]})
 	if reposting_in_progress:
 		frappe.msgprint(_("Item valuation reposting in progress. Report might show incorrect item valuation."), alert=1)
+
+def reset_default_field(doc, default_field, child_table_name, child_table_field):
```
ankush: This should have a docstring to describe what it does. 

Also a util that's only used once isn't a util 😜 we should extend this to all similar fields in other "set default X" type fields.

ankush (changes requested):

erpnext/stock/utils.py
ankush: Also add a 2 small test cases for util function. Mixed values and same values.


A minor optimization: if default_field is empty in document then this function doesn't even need to run.

s-aga-r:
@ankush Where to put this "test_utils.py" file

ankush:
@s-aga-r linter failing. Setup pre-commit locally to avoid these issues: https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

PR #28700, opened 2021-12-03
fix(Non Profit): fetch memberships for 80G certificate by from date only

ruchamahabal:
It's possible for a membership to start in the last month of this fiscal year and end in the first month of the next fiscal year. In such a case, this membership will never be fetched in any 80G certificate because it checks for both Start and End dates for memberships.

Ideally, only the **From Date** should be checked since 80G is handed out for payments done and payments are done at the start of the membership period so **To Date** should not come into the picture while fetching memberships.


ruchamahabal:

PR #28712, opened 2021-12-03
fix!: dont allow renaming warehouse primary key

ankush:
If a mistake is made while naming the warehouse it should be corrected using stock transfer + disabling the original warehouse. 

The display name can still be changed; only `name` can't be changed. 

We are working on a better tool for handling things like items, warehouse renames; until then this has to go because it has the potential to cause inconsistent data. 

Other change:
- remove autocommit from item rename


Ponyooooo:
is the renaming tool for the warehouse available already? why not remove the function on renaming the serial numbers and batch numbers also?

PR #28732, opened 2021-12-05
fix: fixed a method which always returns 404

### LINKED ISSUE #26136: erpnext.stock.get_item_details.get_item_tax_info  this method always returns 404 error Version 13.5.1
#25961 
erpnext.stock.get_item_details.get_item_tax_info  this method always returns 404 error
See the image attached below
image

Python Code seems be culprit here.
Line no :: 447 Inside for-loop of method get_item_tax_info **item_code[1] ===> to replaced with item_code**

vishalpandeynits:
closes  #26136

As described in the issue, the method get_item_tax_info is returning an error 404, and I have made the change by replacing item_code[1] with item_code as instructed in the issue itself.

I would also like to understand from core developers that why this PR fixes the issue of 404, to me it seems a fix of list/tuple index out of range or 'collection' object is not subscriptable. (PS: I am unaware of the structure of different variables used)

cc @ankush

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -469,18 +469,18 @@ def get_item_tax_info(company, tax_category, item_codes, item_rates=None, item_t
 		item_tax_templates = json.loads(item_tax_templates)
 
 	for item_code in item_codes:
-		if not item_code or item_code[1] in out or not item_tax_templates.get(item_code[1]):
+		if not item_code or item_code in out or not item_tax_templates.get(item_code):
 			continue
 
-		out[item_code[1]] = {}
-		item = frappe.get_cached_doc("Item", item_code[0])
-		args = {"company": company, "tax_category": tax_category, "net_rate": item_rates.get(item_code[1])}
+		out[item_code] = {}
+		item = frappe.get_cached_doc("Item", item_code)
+		args = {"company": company, "tax_category": tax_category, "net_rate": item_rates.get(item_code)}
 
 		if item_tax_templates:
-			args.update({"item_tax_template": item_tax_templates.get(item_code[1])})
+			args.update({"item_tax_template": item_tax_templates.get(item_code)})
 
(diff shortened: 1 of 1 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -469,18 +469,18 @@ def get_item_tax_info(company, tax_category, item_codes, item_rates=None, item_t
 		item_tax_templates = json.loads(item_tax_templates)
 
 	for item_code in item_codes:
-		if not item_code or item_code[1] in out or not item_tax_templates.get(item_code[1]):
+		if not item_code or item_code in out or not item_tax_templates.get(item_code):
 			continue
 
-		out[item_code[1]] = {}
-		item = frappe.get_cached_doc("Item", item_code[0])
```
rtdany10, line 476: This line tries to fetch something that doesn't exist and hence the 404, I believe. Feel free to correct if I'm wrong.

vishalpandeynits: I think in that case either of these two exceptions would be raised.
1. 'type(item)' is not subscriptable
2. index is out of range

rtdany10: No, `item_code` is a string. Therefore, `item_code[0]` returns the first character of that string.
As visible in #26136, the item_code is FG-00022 in the table, but on the popup, it says Item F.

rtdany10: Line 476 becomes `item = frappe.get_cached_doc("Item", "F")` then. And there is no item called "F" in the system and hence 404.

vishalpandeynits: Oh, I understood. Thanks a lot.

rtdany10: Happy to help, thank you for contributing :100:

rtdany10: Also try to follow these conventions for your commit messages - https://github.com/pvdlg/conventional-commit-types

PR #28745, opened 2021-12-06
refactor: map serial from schedule if only one

noahjacob:
### Before
- Serial No was not mapped while creating a **Visit** from a **Maintenance Schedule**.

serial_before

### Changes
- Serial No is now mapped while creating a **Visit** from a **Maintenance Schedule** if there is only one serial number.

serial_before

### DIFF
--- a/erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
+++ b/erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
@@ -323,10 +323,14 @@ def make_maintenance_visit(source_name, target_doc=None, item_name=None, s_id=No
 		target.maintenance_schedule = source.name
 		target.maintenance_schedule_detail = s_id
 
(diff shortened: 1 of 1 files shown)


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -323,9 +323,13 @@ def update_status_and_detail(source, target, parent):
...
 
-	def update_sales(source, target, parent):
+	def update_sales_and_serial(source, target, parent):
 		sales_person = frappe.db.get_value('Maintenance Schedule Detail', s_id, 'sales_person')
 		target.service_person = sales_person
+		serial_nos = get_serial_nos(target.serial_no)
+		if len(serial_nos) == 1:
+			target.serial_no = serial_nos[0]
+			return
 		target.serial_no = ''
```
```suggestion
		else:
			target.serial_no = ''
```

This clarifies the intent better IMO.

ankush (approved):

PR #28748, opened 2021-12-06
fix: Ageing in AR/AP report for advances

deepeshgarg007:
When ageing is viewed by due date no ageing is shown for advances received via Journal Entry or payment as they don't have any Due Date.

This leads to a mismatch in the outstanding amount and the amount when the ageing amount in all the ageing columns is added.

This PR considers the Jounral Entry's  and Payment Entry's posting date as the due date to show the ageing based on due date for such entries


deepeshgarg007:

PR #28763, opened 2021-12-06
Discount Amount is not calculating proper for different UOMs

lovinmaxwell:
If multiple UOM is available, the discount amount is applying based on item line UOM


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -387,7 +387,7 @@ def apply_price_discount_rule(pricing_rule, item_details, args):
 			if field not in item_details:
 				item_details.setdefault(field, 0)
 
-			item_details[field] += (pricing_rule.get(field, 0)
+			item_details[field] += (pricing_rule.get(field, 0) * args.get("conversion_factor", 1)
```
ankush, line 390: @lovinmaxwell 

Should this apply to both fields? the discount percentage is not affected by the UOM. 

Also check test failure and update the description to include a minimal reproducible example of the problem you're trying to fix.

ankush (changes requested):

PR #28776, opened 2021-12-07
feat(Regional): KSA E-Invoing optimizations and POS support

rtdany10:
Objectives:
- [x] Rename QR Code field to match localization
- [x] Add QR Code to POS Invoice
- [x] Print Format for POS Invoice with QR Code
- [x] In case of missing QR Code field while submitting invoice, create the field and then QR Code
- [x] Disable KSA print formats for other countries

#no-docs


rtdany10:

rtdany10:
Linters failing with error: `The PR contains a SQL query that may be re-written with frappe.qb`
Since we need to backport, I believe that can be ignored for the time being and be fixed in a future pull request.

erpnext/patches/v13_0/disable_ksa_print_format_for_others.py
```diff
@@ -0,0 +1,17 @@
...
+
+def execute():
+	company = frappe.get_all('Company', filters = {'country': 'Saudi Arabia'})
+	if company:
+		return
+
+	if frappe.db.exists('DocType', 'Print Format'):
+		frappe.reload_doc("regional", "print_format", "ksa_vat_invoice", force=True)
+		frappe.reload_doc("regional", "print_format", "ksa_pos_invoice", force=True)
+		frappe.db.sql("""UPDATE`tabPrint Format` SET disabled = 1
```
deepeshgarg007: @rtdany10 Even version-13 supports query builder, so this query can be rewritten using query builder

erpnext/regional/saudi_arabia/setup.py
```diff
@@ -13,6 +13,18 @@ def setup(company=None, patch=True):
...
 	make_custom_fields()
 
+def add_print_formats():
+	frappe.reload_doc("regional", "print_format", "detailed_tax_invoice", force=True)
+	frappe.reload_doc("regional", "print_format", "simplified_tax_invoice", force=True)
+	frappe.reload_doc("regional", "print_format", "tax_invoice", force=True)
+	frappe.reload_doc("regional", "print_format", "ksa_vat_invoice", force=True)
+	frappe.reload_doc("regional", "print_format", "ksa_pos_invoice", force=True)
+
+	frappe.db.sql("""UPDATE`tabPrint Format` SET disabled = 0 WHERE
```
deepeshgarg007: This can also re written using ORM

erpnext/regional/saudi_arabia/utils.py
```diff
@@ -143,5 +145,6 @@ def delete_vat_settings_for_company(doc, method):
 	if doc.country != 'Saudi Arabia':
 		return
 
-	settings_doc = frappe.get_doc('KSA VAT Setting', {'company': doc.name})
-	settings_doc.delete()
\ No newline at end of file
+	settings_doc = frappe.get_all('KSA VAT Setting', filters={'company': doc.name}, pluck='name')
```
deepeshgarg007: There is only one settings doc per company and this hook is called on deletion of the company, why do we need to use get_all?

deepeshgarg007 (changes requested):

erpnext/regional/saudi_arabia/utils.py
rtdany10: I believe the settings doc of the company can be manually deleted by the user. Therefore, we will have to check if the document exists by name and then delete it. How about restricting the delete permission?

erpnext/regional/saudi_arabia/setup.py
rtdany10: Done

erpnext/patches/v13_0/disable_ksa_print_format_for_others.py
rtdany10: Couldn't get `frappe.qb.update()` to work. Used `frappe.db.set_value()` instead.

rtdany10:
Unrelated test failure?

rtdany10:
@deepeshgarg007 tests all passing. Can we merge this PR?

PR #28782, opened 2021-12-08
fix: batch negative qty validation

ankush:
continuation from https://github.com/frappe/erpnext/pull/27945 

Change: 
- Validate that any backdated transaction can never cause an individual Batch's ledger to go negative at any time in the future. 

The simplest way to reproduce:
1. Create a batched item. Create one batch at T0, a second batch at T0+2 of equal quantities. 
2. Consume the second batch at T0+1. 

Observed: The Batch didn't exist at this time. The transaction gets submitted without any warning/error. 

Expected: Transaction can't be submitted. 

More use cases:
1. Try to consume more from a batch that produced qty in chronological order (no backdated entries) This was caught before this PR too. 
2. Add any transaction that makes the intermediate quantity of batch go negative but overall stays positive. 
3. Use multiple batches like in the example for reproducing above. 
4. **ANY** weird combination you can think of, since this validation constructs a batch's ledger in memory it should catch any and all possible instances of negative stock for a batch. 

TODO:
- [x] Need test case for exact example that broke the previous PR. https://github.com/frappe/erpnext/pull/27945

### DIFF
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -1089,17 +1089,36 @@ def validate_negative_qty_in_future_sle(args, allow_negative_stock=False):
 	allow_negative_stock = cint(allow_negative_stock) \
 		or cint(frappe.db.get_single_value("Stock Settings", "allow_negative_stock"))
 
-	if (args.actual_qty < 0 or args.voucher_type == "Stock Reconciliation") and not allow_negative_stock:
-		sle = get_future_sle_with_negative_qty(args)
-		if sle:
-			message = _("{0} units of {1} needed in {2} on {3} {4} for {5} to complete this transaction.").format(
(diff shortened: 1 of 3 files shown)


erpnext/stock/stock_ledger.py
```diff
@@ -1118,6 +1137,29 @@ def get_future_sle_with_negative_qty(args):
 		limit 1
 	""", args, as_dict=1)
 
+
+def get_future_sle_with_negative_batch_qty(args):
+	return frappe.db.sql("""
+		with batch_ledger as (
```
ankush, line 1143: I've thought a lot about this, there's no simple way to ensure performance + correctness both other than CTE + window function. LMK if anyone has a better suggestion. 

@rohitwaghchaure ?

ankush:

PR #28785, opened 2021-12-08
feat: Populate "Scrap Items" table based on "Scrap %" in BOM

### LINKED ISSUE #28740: Scrap percentage in the BOM child table
Problem
In BOM, In the raw material table rows, there is a field, Scrap %.
Currently, there is no provision to link this field with a scrap material table.

Expected
Based on the value provided in the field, calculate the scrap in the scrap material table.

s-aga-r:
### Feature(s): BOM

_Current Behaviour:_
- In the raw material table rows, there is a field, Scrap %. There is no provision to link this field with a scrap material table.

_Proposed Feature:_
- Based on the Scrap % of RM, the Scarp Items table will be populated.
- For UOMs which can only be whole numbers. If we calculate percentage in that case and that turns out to be a decimal number then trunc the stock_qty value.

<details>

<summary>Images/GIF</summary>

_Before:_

https://user-images.githubusercontent.com/63660334/145161858-11e80eb6-c5e0-4c6f-94d0-1ecf821bdef7.mov

_After:_

https://user-images.githubusercontent.com/63660334/146731922-fd00aafe-bfc4-439f-b1d3-e201a7842084.mov

</details>

**ToDo:**

- [x] Move JS functionality to Py
- [x] Tests
- [ ] Docs


marination:
@s-aga-r Please update docs [here](https://docs.erpnext.com/docs/v13/user/manual/en/manufacturing/articles/scrap-management) and add a test for this

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -140,6 +141,14 @@ def validate(self):
 		if not self.company:
 			frappe.throw(_("Please select a Company first."), title=_("Mandatory"))
 
+		items_in_msg = []
+		for idx, scrap_item in enumerate(self.scrap_items):
+			if isinstance(scrap_item.stock_qty, float) and frappe.db.get_value("UOM", scrap_item.stock_uom, "must_be_whole_number"):
+				self.scrap_items[idx].stock_qty = trunc(scrap_item.stock_qty)
+				items_in_msg.append(f"<b>{scrap_item.item_name}</b>")
+		if items_in_msg:
+			frappe.msgprint("Scrap cannot be in fraction for Item(s): " + ", ".join(map(str, items_in_msg)))
```
marination:
```suggestion
			frappe.msgprint("Scrap rounded down since it cannot be a fraction for Item(s): " + ", ".join(map(str, items_in_msg)))
```
Also make this a toast by passing flags: 
```py 
alert=True, indicator='orange'
```

marination: Push this snippet into another function called `validate_scrap_item_qty` maybe. Leaving stray snippets like this in `validate` looks messy.

erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -734,3 +740,44 @@ function trigger_process_loss_qty_prompt(frm, cdt, cdn, item_code) {
...
+			}
+		}
+
+		if (row) {
+			row.stock_qty = scrap_item_qty;
+			row.rate = item.rate;
+			row.amount = scrap_item_qty * item.rate;
+			row.stock_uom = item.stock_uom;
+			row.base_rate = item.base_rate;
+			row.base_amount = item.base_rate * scrap_item_qty;
```
marination: Do we need to update these ? This function is triggered only on changing scrap %, so only qty and related stuff like amount needs a change right ? why would the rate change?

s-aga-r: Yes, In some scenarios

Scenario: 
1. Add an item in the Items.
2. Set "Scrap %" (will populate a new row in Scrap)
3. Again add the same item with a different qty or rate and set "Scrap %".

Now, the existing row in Scrap will be updated with new values i.e qty and rate.

s-aga-r: No, This function is also triggered when we change item.qty and item.rate

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -140,6 +141,15 @@ def validate(self):
 		if not self.company:
 			frappe.throw(_("Please select a Company first."), title=_("Mandatory"))
 
+		# items_in_msg = []
+		# for idx, scrap_item in enumerate(self.scrap_items):
+		# 	if isinstance(scrap_item.stock_qty, float) and frappe.db.get_value("UOM", scrap_item.stock_uom, "must_be_whole_number"):
+		# 		self.scrap_items[idx].stock_qty = trunc(scrap_item.stock_qty)
+		# 		items_in_msg.append(f"<b>{scrap_item.item_name}</b>")
+		# if items_in_msg:
+		# 	frappe.msgprint("Scrap cannot be in fraction for Item(s): " + ", ".join(map(str, items_in_msg)))
```
marination: why is this left as a comment here?

s-aga-r: Forgot to remove

erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -734,3 +740,44 @@ function trigger_process_loss_qty_prompt(frm, cdt, cdn, item_code) {
...
+				item_name: item.item_name,
+				stock_qty: scrap_item_qty,
+				rate: item.rate,
+				amount: scrap_item_qty * item.rate,
+				stock_uom: item.stock_uom,
+				base_rate: item.base_rate,
+				base_amount: item.base_rate * scrap_item_qty,
+				is_process_loss: 0
+			});
+		}
```
marination: This can use lesser redundancy. All the common items can be one object, the same can update the row object . [Spread Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) is useful here.
Excuse the indentation, but you get the idea
```suggestion
        let row_data = {
              "stock_qty" = scrap_item_qty;
	      "rate" = item.rate;
	      "amount" = scrap_item_qty * item.rate;
	      "stock_uom" = item.stock_uom;
	      "base_rate" = item.base_rate;
	      "base_amount" = item.base_rate * scrap_item_qty;
        };
        
	if (not row) {
	      row = frm.add_child('scrap_items', {
	              item_code: item.item_code,
		      item_name: item.item_name,
		      is_process_loss: 0
	      });
	}
	row = { ...row_data }; // update the existing row object with new data
```

marination:
Rest looks fine, after docs and tests should be able to merge

marination:
@s-aga-r Updated the description with some ToDos needed to merge this

s-aga-r:
@marination 
1. Scrap Items table is not populating after save.
2. Did we need to assign an index manually for each scrap item row?

https://user-images.githubusercontent.com/63660334/151516975-02ee72b8-9129-411b-a9ed-5862b5f492d7.mov

ankush:
```python
doc.append("child_table_field", {})
```

this should be enough to add child items. Dictionary contains k:v pairs of child doctype properties. 


updated docs: https://frappeframework.com/docs/v13/user/en/api/document#docappend

s-aga-r:
@ankush Thanks :)


> ```python
> doc.append("child_table_field", {})
> ```
> 
> this should be enough to add child items. Dictionary contains k:v pairs of child doctype properties.
> 
> updated docs: https://frappeframework.com/docs/v13/user/en/api/document#docappend


@ankush Thanks :)

PR #28797, opened 2021-12-08
fix: validate if asset account is set against company

nextchamp-saqib:
Problem:
- Asset Category "Office Equipments" has a Fixed Asset Account set for Company 1
- Printer belongs to Asset Category Office Equipments
- Creating Purchase Invoice for Company 2 with Printer as item throws an error "Account is Required", if expense account is not explicitly set

Fix:
- Throw a proper error message if expense account is not found for the asset item

   CleanShot 2021-12-08 at 17 35 09


nextchamp-saqib:

PR #28816, opened 2021-12-10
chore: update stale rules and apply on issues

ankush:
Some changes in how the public issue tracker will work from now. These changes are necessary for us to use GitHub issue tracker as our primary issue tracker. 

1. Issue reporting format will be changed to capture module-info using github's new yaml based forms. 
2. Someone from the maintainer team (~50 people right now) needs to validate the issue be it a feature request or bug report by adding `valid` label on it. 
3. If the issue isn't validated and stays inactive for **60 days**, the stale bot will give a warning and close it in 7 days. 
4. Issues/PRs added to milestones or projects are exempted. 
5. Issues with `valid` and `to-validate` labels are ignored.
6. Current issues will be labeled `to-validate` and triaged over time. 

PS: PR rules remain the same. 

Will merge after finishing WIP tasks.


ankush:
The rationale behind this, from stalebot's readme:

> ## Is closing stale issues really a good idea?
> In an ideal world with infinite resources, there would be no need for this app.

> But in any successful software project, there's always more work to do than people to do it. As more and more work piles up, it becomes paralyzing. Just making decisions about what work should and shouldn't get done can exhaust all available resources. In the experience of the maintainers of this app—and the hundreds of other projects and organizations that use it—focusing on issues that are actively affecting humans is an effective method for prioritizing work.

> To some, a robot trying to close stale issues may seem inhospitable or offensive to contributors. But the alternative is to disrespect them by setting false expectations and implicitly ignoring their work. This app makes it explicit: if work is not progressing, then it's stale. A comment is all it takes to keep the conversation alive.

PR #28822, opened 2021-12-10
feat: Deferred Revenue and Expense report with actual and upcoming postings

### LINKED ISSUE #28225: Deferred Accounting Report for Income and Expenses
### Problem Statement:

1.  Since income and expenses are deferred, you don't come to know:
                   1.  The actual income booked in a current month
                   2.  The upcoming income which will be posted in coming months, as per start and end date
4.  The upcoming expense ------------------ // -----------------
5.  The expected against actual post
6.  Actual posting of expense as per the start and end date for each fiscal year / quarter / month
7.  Actual posting of income as per the start and end date for each fiscal year / quarter / month
 
### Expected Solution:

1.  Make new reports to provide this information

### Who is the end user? Define persona(s).

1.  Business owner, who wants to:
2.  see income and expense, with and without deferred posting
3.  Upcoming Income / Expense as per future deferred posting
4.  Account Manager / Auditor, who wants to review if the posting of deferred entries has happened correctly as per the Start and End Period set in the invoices.

ruthra-kumar:
### Deferred Revenue and Expense Report
This report aim to provide a simplified view of all deferred income/expense in a certain period. It can help in identifying the actual and expected posting.

### Deferred Revenue
def_revenue

### Deferred Expense
def_expense

### Documentation
https://docs.erpnext.com/docs/v13/user/manual/en/accounts/deferred_revenue/expense_report


erpnext/accounts/report/deferred_revenue_and_expense/deferred_revenue_and_expense.py
```diff
@@ -0,0 +1,436 @@
...
+		elif items[0].deferred_expense_account:
+			self.type = "Purchase"
+
+		self.items = []
+		# for each uniq items
+		self.uniq_items = set([x.item for x in items])
+		for item in self.uniq_items:
+			self.items.append(Deferred_Item(item, self, [x for x in items if x.item == item]))
+
+	def get_postings(self):
```
deepeshgarg007: Seems like a dead function, not called from anywhere

erpnext/accounts/report/deferred_revenue_and_expense/deferred_revenue_and_expense.js
```diff
@@ -0,0 +1,114 @@
...
+			"options": [
+				{ "value": "Revenue", "label": __("Revenue") },
+				{ "value": "Expense", "label": __("Expense") }
+			],
+			"default": "Revenue",
+			"reqd": 1
+		},
+		{
+			"fieldname" : "with_upcoming_postings",
+			"label": __("Show with upcoming revenue/expense"),
```
deepeshgarg007, line 88: This also shows unbooked values for the previous months, may better to call this label as "Show expected revenue/expense" and this should be unchecked by default

deepeshgarg007 (changes requested):

deepeshgarg007:
Fix the linting issues

erpnext/accounts/report/deferred_revenue_and_expense/deferred_revenue_and_expense.js
ruthra-kumar: Agreed. 
Along with this, I think it's better if it only shows the actual posting(bar graph) when unchecked. Thoughts?

deepeshgarg007: If someone wants to compare the expected and actual posting, how do they do that then?

erpnext/accounts/report/deferred_revenue_and_expense/deferred_revenue_and_expense.py
ruthra-kumar: removed

erpnext/accounts/report/deferred_revenue_and_expense/deferred_revenue_and_expense.js
ruthra-kumar: renamed and 'unchecked' by default.

deepeshgarg007:

PR #28831, opened 2021-12-11
Fix: Error when submit in Period Closing Voucher doctype

aymenit2008:
delete account_currency from get_pnl_gl_entry function because closing account head has same company currency and acc.account_currency may has different currency


aymenit2008:
image

PR #28832, opened 2021-12-12
fix: KSA QR field rename patch

deepeshgarg007:
Fixes an issue introduced in https://github.com/frappe/erpnext/pull/28776

The data from old field was not migrated to the new field as it didn't exist in the first place

### DIFF
--- a/erpnext/patches/v13_0/rename_ksa_qr_field.py
+++ b/erpnext/patches/v13_0/rename_ksa_qr_field.py
@@ -2,6 +2,7 @@
 # License: GNU General Public License v3. See license.txt
 
 import frappe
+from frappe.custom.doctype.custom_field.custom_field import create_custom_fields
 from frappe.model.utils.rename_field import rename_field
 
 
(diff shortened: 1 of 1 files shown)


erpnext/patches/v13_0/rename_ksa_qr_field.py
```diff
@@ -12,5 +13,20 @@ def execute():
 
 	if frappe.db.exists('DocType', 'Sales Invoice'):
 		frappe.reload_doc('accounts', 'doctype', 'sales_invoice', force=True)
+
+		# rename_field method assumes that the field already exists or the doc is synced
+		if not frappe.db.has_column('Sales Invoice', 'ksa_einv_qr'):
```
rtdany10, line 18: Shouldn't https://github.com/frappe/erpnext/blob/develop/erpnext/regional/saudi_arabia/utils.py#L19-L30 be enough? :thinking: 
It checks if the field is present, if not, creates, before sales invoice is submitted.

deepeshgarg007: Then there is no point of the patch then, the patch will get executed before the field is created and no data will me moved from the old field in the invoices

deepeshgarg007:

deepeshgarg007:

PR #28839, opened 2021-12-13
fix: remove migration hashes from doctypes

ankush:
there was an issue with migration hash: https://github.com/frappe/frappe/pull/14564


ankush:

PR #22055, opened 2020-05-30
feat: Selling Desk, Dashboard and Onboarding

marination:
<details><summary>Onboarding</summary>

- Onboarding on Desk
  Screenshot 2020-06-01 at 2 01 08 PM
 Screenshot 2020-06-01 at 2 06 33 PM

- _Form Tour_ of Selling Settings
  selling-settings

</details>

<details><summary>Selling Desk</summary>
 
  screencapture-ugh-in-8001-desk-2020-05-30-17_53_38
</details>

<details><summary>Reports</summary>

- **Sales Order Analysis Report** (replaced **Ordered Items to be Billed** and **Ordered Items to be Delivered and Billed**)
- Patch to delete above Reports
- Added **Sales Order Analysis Report** to Accounts and Stock Desk
- Added Chart, Filters and Color coding to Item-wise Sales History Report
- Added Chart to Quotation Trends Report
- Added Chart to Sales Order Trends Report
- Sorted Delivery Note Trends Report's Chart
</details>

<details><summary>Selling Dashboard</summary>

  screencapture-ugh-in-8001-desk-2020-05-30-17_52_36
</details>

**ToDo:**
- [x]  Fixtures for Chart, Numbers Cards and Dashboard
- [x]  Selling Onboarding

PR #28842, opened 2021-12-13
feat: new column 'Time taken to Deliver' in sales order analysis

### LINKED ISSUE #28753: Include column for 'Days taken to deliver' in Sales Order Analysis Report
Currently, the Sales Order Analysis reflects the Delay in Days took them to deliver the Ordered Items, they want one more column to reflect how many days they take to completely deliver a Sales Order in order to increase the efficiency and get more productive accordingly.

ruthra-kumar:
### Current Report
_'Sales Order Analysis'_ helps in quickly identifying billed amount, pending amount, delay in days and more information. Along with this a new column is added

### New Column - 'Time taken to Deliver'
This denotes the no of days taken to completely deliver an item. This helps in identifying bottlenecks and get more efficient.

### Image
time_taken_to_deliver

**no-docs**


erpnext/selling/report/sales_order_analysis/sales_order_analysis.py
```diff
@@ -259,6 +264,12 @@ def get_columns(filters):
 			"fieldname": "delay",
 			"fieldtype": "Data",
 			"width": 100
+		},
+		{
+			"label": _("Time Taken to Deliver"),
+			"fieldname": "time_taken_to_deliver",
+			"fieldtype": "Data",
```
nextchamp-saqib: Can we make this a Duration field? And store the seconds difference between `posting_date` & `transaction_date`?

ruthra-kumar: Changed to Duration

nextchamp-saqib (approved):

marination:
@nextchamp-saqib @ruthra-kumar This report was introduced via https://github.com/frappe/erpnext/pull/22055 to replace "Ordered Items to be Billed" and "Ordered Items to be Delivered and Billed"
The right join to DN makes it only fetch those transaction that have DNs against them, it does not fetch SOs that are pending to Deliver and Bill
It could use a test too, given that this could have been identified by CI

Also this column is a bit confusing with the Delay column, this needs documentation too https://docs.erpnext.com/docs/v13/user/manual/en/selling/sales-analytics

ruthra-kumar:
@marination thanks for reporting this. Will create a fix.

PR #28843, opened 2021-12-13
refactor: update_serial_no function for old Maintenance Visits

noahjacob:
### Issue

- Old documents which were saved before the refactor would not have the "Maintenance Schedule Detail" field set.

image

### Fix
- Using `maintenance_schedule` field in **"Maintenance Visit"** to fetch the serial nos instead of the **Maintenance Schedule Detail** row id, which isn't possible to identify on old docs.
- Patch to update this field on old docs that have `prevdoc_docname`.

**To Test:**
- Make an independent maintenance visit (not from maintenance schedule) and see if it works fine. Use serialised item in it and add serial nos > Add multiple serialised items and see if you can select appropriate serial nos (check the filters)
- Make a maintenance schedule with multiple items, 1 non serialised, 2 serialised. Make visits, see if it works and updates the schedule correctly


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -306,13 +307,27 @@ def get_pending_data(self, data_type, s_date=None, item_name=None):
...
+	if isinstance(item, str):
+		item = json.loads(item)
+		item = frappe._dict(item)
+
+	serial_nos = []
+	if s_id:
+		serial_nos = frappe.db.get_value('Maintenance Schedule Detail', s_id, 'serial_no')
+
+	# for old docs which were already saved.
+	elif item.get('prevdoc_docname'):
```
ankush: This can't be patched to handle changes in code? So we don't have to add code like this.

noahjacob: I had removed the mapping to these fields. So the patch could be copying this prevdoc_docname(if present) to the form level field(both are storing the maintenance schedule name)? I could just use the maintenance schedule name while fetching the serial nos but then I wouldn't be covering cases where the user could add duplicate items with a single serial no

serial

Currently it uses the linked row id from the schedules table in Maintenance Schedule, can't get in old docs since this was never linked. Should I just ignore this case and use the name?

noahjacob: Or could just not filter the serials for old docs(old behaviour).

noahjacob: Fetching serials with maintenance schedule name now, wrote a patch to update the maintenance schedule field.

noahjacob: What type of tests should I write for this?

erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -306,26 +305,28 @@ def get_pending_data(self, data_type, s_date=None, item_name=None):
 					return schedule.name
 
 @frappe.whitelist()
-def update_serial_nos(s_id):
-	serial_nos = frappe.db.get_value('Maintenance Schedule Detail', s_id, 'serial_no')
+def update_serial_nos(item_code, schedule=None):
```
marination: why is this called `update_serial_nos` when it just returns serial nos of an item from the Maintenance Schedule ?
`get_serial_nos_from_schedule` would be better ? or something like that

erpnext/maintenance/doctype/maintenance_visit/maintenance_visit.js
```diff
@@ -2,52 +2,45 @@
...
 		frm.set_query('customer_address', erpnext.queries.address_query);
 		frm.set_query('customer', erpnext.queries.customer);
 	},
-	onload: function (frm, cdt, cdn) {
-		let item = locals[cdt][cdn];
+	onload: function (frm) {
+		// filters for serial no based on item code
 		if (frm.doc.maintenance_type === "Scheduled") {
-			const schedule_id = item.purposes[0].prevdoc_detail_docname || frm.doc.maintenance_schedule_detail;
+			let item_code = frm.doc.purposes[0].item_code;
```
marination, line 14: this seems incorrect in cases where an extra item is added to the purposes table.
Consider the use case: An adhoc visit occurs, this wasn't scheduled so there is no schedule. A customer wanted immediate maintenance. So in an independent maintenance visit case i could add many rows.

The Serial No filter only considers row 1. It should be based on the item code of the current row.

The set_query below's logic is fine, but it is getting the incorrect item code. That should be more dynamic

marination (approved):

marination:

PR #28845, opened 2021-12-13
fix: incorrect amount based on payment days in timesheet salary slip

ruchamahabal:
## Problem

1. Create a Salary Structure with "**Salary Slip Based on Timesheet**" enabled.
2. Add other components to the structure with "**Depends on Payment Days**" checked.
3. Mark the Employee absent for a few days.
3. Create a Salary Slip. If the Payment Days are less than the Total Days, the system won't calculate amounts based on Payment Days for other components.

## Fix

It skips Payment Days calculation for all components when Salary Slip is based on Timesheet. It should only skip payment days calculation for the Timesheet Component and correctly calculate for others.


ruchamahabal:
Unrelated test case failing. Will fix separately

ruchamahabal:
@Mergifyio backport version-13-hotfix version-12-hotfix

PR #28877, opened 2021-12-14
fix: qty filter not working if apply_multiple_pricing_rules is enabled

nextchamp-saqib:
`filter_pricing_rules` doesn't work in case of `apply_multiple_pricing_rules` checked
`sorted_by_priority` filters the pricing rules, if no pricing rules are found after filtering, all the pricing rules are returned back

https://github.com/frappe/erpnext/blob/2718ddd784e8dc13b5fae1bd7c201c6b76e1391d/erpnext/accounts/doctype/pricing_rule/utils.py#L59-L76

<details>

1. Create a Pricing Rule for Buy 3 Get 10% Discount
   CleanShot 2021-12-14 at 15 00 25
2. Create a Pricing Rule for Buy 3 Get 1 Free
   CleanShot 2021-12-14 at 15 01 09
3. Create a Sales Invoice and the item on which the free item rule is set.
4. Even if the Qty is < 3, the free item is added in the Sales Invoice
   CleanShot 2021-12-14 at 15 06 23

</details>


erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -67,13 +67,12 @@ def sorted_by_priority(pricing_rules, args, doc=None):
 			if not pricing_rule.get('priority'):
 				pricing_rule['priority'] = 1
 
-			if pricing_rule.get('apply_multiple_pricing_rules'):
```
deepeshgarg007: What was the significance of this parameter earlier, completely removing this won't cause any issues in other scenarios? Isn't there a checkbox in the form corresponding to this parameter, will it be still useful?

nextchamp-saqib: `sorted_by_priority` should only sort the `pricing_rules` by priority. 
Whether multiple pricing rules are to be applied or not, should be decided at some other place. 

---

This should not break anything since `sorted_by_priority` is only called if (line 45) `apply_multiple_pricing_rules(pricing_rules)` is `True`

---

> Isn't there a checkbox in the form corresponding to this parameter, will it be still useful?

Yes, that check is already being used in `apply_multiple_pricing_rules()`
Also, this check can only be enabled from promotional schemes, I guess.

nextchamp-saqib: After discussing with @rohitwaghchaure, found out that this change is not needed. Hence removing it.

nextchamp-saqib:
@deepeshgarg007 review once again

PR #28887, opened 2021-12-14
fix: reset_expected_response_and_resolution on hold

nextchamp-saqib:


nextchamp-saqib:
Unrelated patch failing, will be fixed in frappe/frappe

PR #28724, opened 2021-12-03
fix: gross profit report showing zero valuation

mro-admin:
Issue: [(https://discuss.erpnext.com/t/gross-profit-100/7646/16)] Gross profit report will show zero valuation on some items

Fix: Ensure that the correct name for the parrenttype is used when being compared to the stock ledger entries

PR #28909, opened 2021-12-15
fix: gross profit report not properly calculating valuation rate

mro-admin:
original pull request https://github.com/frappe/erpnext/pull/28724#issuecomment-995035342

### DIFF
--- a/erpnext/accounts/report/gross_profit/gross_profit.py
+++ b/erpnext/accounts/report/gross_profit/gross_profit.py
@@ -335,7 +335,7 @@ class GrossProfitGenerator(object):
 		else:
 			my_sle = self.sle.get((item_code, row.warehouse))
 			if (row.update_stock or row.dn_detail) and my_sle:
-				parenttype, parent = row.parenttype, row.parent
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -335,7 +335,7 @@ def get_buying_amount(self, row, item_code):
 		else:
 			my_sle = self.sle.get((item_code, row.warehouse))
 			if (row.update_stock or row.dn_detail) and my_sle:
-				parenttype, parent = row.parenttype, row.parent
+				parenttype, parent = row.parenttype, (row.parent_invoice if row.parenttype == "Sales Invoice" and row.parent_invoice else row.parent)
```
deepeshgarg007, line 338: @mro-admin the issues caused by reusing the `parent` key seems to be fixed in this PR https://github.com/frappe/erpnext/pull/28483. Not sure if we still need this change. If you are still getting 0 valuation rates there might be some other issue. Is this only seen for product bundle items, then maybe its the parent being set as `None` while making product bundle rows

deepeshgarg007:
@mro-admin any updates on this?

mro-admin:
> @mro-admin any updates on this?

Can confirm its been fixed with that PR. Thanks!

PR #28916, opened 2021-12-16
fix: Multicurrency invoices using subscription

deepeshgarg007:
The user gets the following error on trying to create an invoice using subscription against a party having billing currency other than the company currency

image

Sets to test the pull request
1. Create a customer with billing currency other than company currency, Eg USD if company currency is INR
Screenshot 2021-12-17 at 12 18 05 PM

2. Create a subscription plan in party currency
image

3. Create a backdated subscription as per the following details
image

4. Click on the fetch subscription updates button on the top right and see if the invoice gets generated successfully
image

### DIFF
--- a/erpnext/accounts/doctype/subscription/subscription.py
+++ b/erpnext/accounts/doctype/subscription/subscription.py
@@ -23,6 +23,7 @@ from erpnext.accounts.doctype.accounting_dimension.accounting_dimension import (
 	get_accounting_dimensions,
 )
 from erpnext.accounts.doctype.subscription_plan.subscription_plan import get_plan_rate
+from erpnext.accounts.party import get_party_account_currency
 
 
 class Subscription(Document):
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

vanesaotaduy:
@deepeshgarg007 this pr doesn't solve the issue [Subscription currency #29036](https://github.com/frappe/erpnext/pull/29036).
When you set up the subscription plan with company currency and choose a price list in another currency, for example, USD, the invoice doesn't convert the amount. It takes the price in USD as it was in the company currency. Let me show you an example on V13:

I created this subscription plan by setting up ARS currency for the plan and choosing a Price List in USD.

image


The price for this product is 300 USD for the list price selected on the plan.

When issuing the invoice, takes the price as 300 ARS, which is wrong,  it should be converted at the exchange rate of 1175 ARS.

image

The correct amount should be 300 x 1175 = 352.500 ARS

PR #28917, opened 2021-12-16
fix: mapping to maintenance visit gets erased

noahjacob:
### Issue
- When trying to create a **Maintenance Visit** via **Maintenance Schedule**, no items get mapped.

mapping_broken

### Reason
- When mapping to visit, the form is still in the `__islocal` state.

image

### Fix
- Since the items are mapped to Visit only via Schedule, added the condition `frm.doc.maintenance_type == 'Unscheduled'` as it would only need to be cleared if the maintenance_type is of type "Unscheduled". (When creating visit via schedule, the type is set to "Scheduled")

### Notes
- The `frm.clear_table()` was introduced to removed the empty row on creating a new Unscheduled Visit.

empty_row


ankush:
@noahjacob  merging for now. Please fix the root cause (extra row) instead of using `clear_table`.

PR #28919, opened 2021-12-16
feat: Add currency in import download statement

mohsinalimat:
Issue:
When parent child company with different currency.
When upload statement at that time issue comes.

Solve:
After this changes done, I upload statement without any errors.


mohsinalimat:
@deepeshgarg007 Anything I miss?

PR #28929, opened 2021-12-17
fix: Refresh Stock Items table in Asset Repair after inserting Serial Numbers

GangaManoj:
_Problems Addressed:_

- https://github.com/frappe/erpnext/pull/28349 introduced the option to add Serial Numbers to Stock Items consumed during Asset Repairs. However, after inserting Serial Numbers using the Add Serial No button, the Serial No field in the Stock Items table isn't updated instantly, it only gets updated on expanding the row again. This PR fixes this by refreshing the Stock Items table after inserting the Serial Numbers.

<details>

1. Create a Serialized Item
2. Submit Stock Entry of type Material Receipt for it(or just create Serial No docs)
3. Create an Asset
4. Create an Asset Repair for it
5. Check the _Stock Consumed During Repair_ box
6. Enter the Serialized Item in the Stock Items table
7. Expand the row for the Serialized Item
8. Click on the _Add Serial No_ button and select the Serial No of your choice
9. Click the _Insert_ button

Screenshot 2021-12-17 at 8 37 21 AM

_Expected Result:_ The entered Serial Numbers should instantly be added to the _Serial No_ field.

Screenshot 2021-12-17 at 8 37 52 AM

</details>

- [minor] The Stock Items table uses fieldnames `valuation_rate` for `rate` and `consumed_qty` for `qty`, which is inconsistent with the rest of ERPNext and causes minor problems here and there. This PR renames them to just `rate` and `qty`.

### DIFF
--- a/erpnext/assets/doctype/asset_repair_consumed_item/asset_repair_consumed_item.json
+++ b/erpnext/assets/doctype/asset_repair_consumed_item/asset_repair_consumed_item.json
@@ -6,26 +6,12 @@
  "engine": "InnoDB",
  "field_order": [
   "item_code",
-  "valuation_rate",
-  "consumed_quantity",
+  "rate",
+  "qty",
   "total_value",
   "serial_no"
  ],
  "fields": [
-  {
-   "fetch_from": "item.valuation_rate",
-   "fieldname": "valuation_rate",
-   "fieldtype": "Currency",
-   "in_list_view": 1,
-   "label": "Valuation Rate",
-   "read_only": 1
-  },
-  {
-   "fieldname": "consumed_quantity",
-   "fieldtype": "Data",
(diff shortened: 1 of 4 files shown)


erpnext/assets/doctype/asset_repair_consumed_item/asset_repair_consumed_item.json
```diff
@@ -44,12 +30,26 @@
    "in_list_view": 1,
    "label": "Item",
    "options": "Item"
+  },
+  {
+   "fetch_from": "item.valuation_rate",
+   "fieldname": "rate",
```
ankush, line 36: These new fields won't get auto patched. You'll have to write patch to migrate existing data.

PR #28935, opened 2021-12-17
fix: Is Reverse Charge check in Tax Category

deepeshgarg007:
Added "Is Reverse Charge" custom field for tax categories for Indian Companies

Screenshot 2021-12-21 at 12 45 41 PM

If a user has done a GST setup where no source state is mentioned in the tax categories and a common template is used for all the states then there was no differentiation between regular and reverse charge templates and Reverse Charge template was auto fetched even for regular transactions

This PR fixes that issue by identifying RCM tax categories separately


deepeshgarg007:
@Mergifyio backport develop

PR #28945, opened 2021-12-18
test: test failure due to rounding

rtdany10:
```
FAIL  test_average_ratings_on_feedback_submission_and_cancellation
(erpnext.hr.doctype.interview_feedback.test_interview_feedback.TestInterviewFeedback)
  File "/home/runner/frappe-bench/apps/erpnext/erpnext/hr/doctype/interview_feedback/
        test_interview_feedback.py", line 62, in test_average_ratings_on_feedback_submission_and_cancellation
    self.assertEqual(avg_on_interview_detail, round(feedback_1.average_rating))
AssertionError: 3.5 != 4
```

Rounding produces different result in case of averages with decimal and hence the assertEqual fails.

### DIFF
--- a/erpnext/hr/doctype/interview_feedback/test_interview_feedback.py
+++ b/erpnext/hr/doctype/interview_feedback/test_interview_feedback.py
@@ -59,7 +59,7 @@ class TestInterviewFeedback(unittest.TestCase):
 		}, 'average_rating')
 
 		# 1. average should be reflected in Interview Detail.
-		self.assertEqual(avg_on_interview_detail, round(feedback_1.average_rating))
+		self.assertEqual(avg_on_interview_detail, feedback_1.average_rating)
 
 		'''For Second Interviewer Feedback'''
 		interviewer = interview.interview_details[1].interviewer


rtdany10:
Issue arises only if the average has decimal. Since random function generates ratings, the average might or might not have decimal. I assume, so far, the test passed cause the average didn't have a decimal and hence the round function really didn't have any effect? :confused: 

https://github.com/frappe/erpnext/blob/develop/erpnext/hr/doctype/interview_feedback/test_interview_feedback.py#L95-L101

P.S Tests all passing :smile:

rtdany10:

ankush:
this is failing because of https://github.com/frappe/frappe/pull/15203 

Change in code is required, instead of fixing test 😄

rtdany10:
> Change in code is required, instead of fixing test

Skeptical of that since the code assigns the value as it should, without rounding (?) :confused:  https://github.com/frappe/erpnext/blob/e85da7a03a45787ec785f1d0b5d0efc56f145149/erpnext/hr/doctype/interview_feedback/interview_feedback.py#L70

rtdany10:
My bad. Ratings are divided by number of starts and stored in db now.
So, yes, the code requires change. Shall I try fix it on this PR itself?

rtdany10:
My bad. Ratings are divided by number of stars and stored in db now.
So, yes, the code requires change. Shall I try fix it on this PR itself?

@ankush

ankush:
Let's leave this for now, we will soon start allowing half-star ratings so then rounding won't make much sense. 

ref: https://github.com/frappe/frappe/pull/15203#issuecomment-996642229

PR #28359, opened 2021-11-12
feat: ERPNext support for Postgres

c-p-b:
Companion Frappe PR: https://github.com/frappe/frappe/pull/15354

At a high level, mostly multisql was used when possible favoring changing MariaDB queries themselves. When the query had to be rewritten entirely, db independent syntax and frappe.db.sql was favored. The goal of this PR is to target as low-risk change as possible for existing installations.  It adds a bit of technical debt in that in the ideal world, ORM would be used where possible, there are perhaps several simpler queries that I touched that could have converted to using the ORM, but at the risk of introducing new bugs/behavior in the MariaDB queries I opted not to do that and instead favored using multisql and frappe.db.sql.

Caveats:

1. Little was known about underlying data behavior or what things are used for - As a result I tried to change the existing behavior of mdb queries as little as possible. Because of this, ORM was only minimally used, and generally the strategy of modifying existing query syntax to support both engines where possible and using multisql where not possible was followed.  If conversion to multisql became too ugly/unweildy, extra time was taken to port to the ORM
2. This was only tested against Postgres 13.2 and MariaDB 10.3
3. During pg date/time conversions and comparisons some explicit type casting was required for the engine to execute functions - some of this may  throw different errors in runtime than what you might see from mariadb and this also may introduce some performance implications
4. Using the collation for MariaDB that is by default provided by ERPNext means that queries are case insensitive in MariaDB. This PR accounts for that and changes reflect that queries are case sensitive for Postgres for the purposes of better data integrity. Therefore Postgres is more stringent than MariaDB in this regard
5. Currently this PR has changes in frappe that require auto_commit on certain calls to db_insert() and insert() functions. These auto_commits only run during installation and testing and only in specific parts of test data bootstrap. During normal runtime normal behavior is respected
6. The data structure of the output of several queries needed to be modified. This primarily affects conversions from frappe.db.get_* to frappe.qb and frappe.db.sql/multisql. Typically the old style queries would return nested tuples e.g. ((, )), and the newer style returns a tuple wrapped in a list e.g. [(, )]. The output of the queries is pretty well covered in tests as they are all consumed downstream so this should be a relatively low risk change.

In this PR:

1. Fixed all instances of the sql directly checking against the timestamp ‘0000-00…’ as this is not a valid date in Postgres (but it is for MariaDB). It is instead removed entirely as the Postgres equivalent is simply NULL which the MariaDB queries already check for
2. Converted all instances calling timestamp() to its equivalent addition operator (date + time) in pgsql
[...]

PR #28952, opened 2021-12-20
ci: postgres unittests configurations

c-p-b:
This PR simply enables running of postgres tests without doing anything else. It requires a couple of minor commits in Frappe repository: https://github.com/cpdeethree/frappe/tree/enable-postgres-tests

These commits are also included on the larger frappe and erpnext pr's with changes for postgres support for ERPNext: 

https://github.com/frappe/erpnext/pull/28359

Thus (minus the DEBUG branch pointer commit), all changes here are strict subsets of the above PR's

Depends on: https://github.com/frappe/frappe/pull/15354


ankush:
This is technically ready for review but can't be merged till https://github.com/frappe/frappe/pull/15354 is merged, hence marking it draft for now. 

Also I need to make this CI workflow conditional for postgres PRs.

ankush:
note: Reverted change to HSN code creation since it appears that something else is rolling back full transaction as apparent from CI run on this commit: https://github.com/frappe/erpnext/pull/28952/commits/d8649727ce451040f78cd9aeeb12e97dbb0850ec 

https://github.com/frappe/erpnext/runs/4684613490?check_suite_focus=true

ankush:
note: Reverted change to HSN code creation since it appears that something else is rolling back full transaction as apparent from CI run on this commit: https://github.com/frappe/erpnext/pull/28952/commits/d8649727ce451040f78cd9aeeb12e97dbb0850ec 

https://github.com/frappe/erpnext/runs/4684613490?check_suite_focus=true 


This is the root cause behind HSN codes not getting created properly in tests. I can fix this here but feels like this needs to be handled better by the framework. Improper error handling is causing rollback. Well, actually both changes are required, this kinda error handling in regional setup will cause weird state in DB. 

https://github.com/frappe/erpnext/blob/d8649727ce451040f78cd9aeeb12e97dbb0850ec/erpnext/regional/india/setup.py#L715-L723

cc: @gavindsouza @deepeshgarg007 


for postgres DB we are just rolling back everything on SQL exception 😐 

https://github.com/frappe/frappe/blob/96e1b7dcf9d39651aa72ea7af151aa15208e765b/frappe/database/database.py#L165-L166

ankush:
Merging just the CI change for now, rest will require a fix at framework level.

PR #28953, opened 2021-12-20
fix(UX): Optimize rate updation on changing price list

deepeshgarg007:
Consider a scenario where you selected more than 100 items in an Invoice and then later realize that the price list selected is incorrect

On changing the price list rate and other related fields are updated and triggers like `calculate_taxes_and_charges` are run for each row which freezes the form for minutes in some cases and the user has to wait for to make other updates

**Solution**: Do not call any triggers and just cal the required functions manually either at the item or parent level


deepeshgarg007:
@Mergifyio backport develop

PR #28966, opened 2021-12-20
fix: Stock Ageing Report - Negative Opening Stock

marination:
**Issue:**
Consider ledger:
> 1. -30, 1st Dec
> 2. 100, 2nd Dec
> 3. -10, 3rd Dec

- Stock Ageing Report ignored the first entry (no impact) as there is no existing qty in the FIFO Queue to consume from
- Hence, the total qty of the range columns here would show **30 qty more** than the actual value [**100 -10 (90)** instead of **-30+100-10 (60)**]

**Fix:**
- Added a case to handle empty FIFO Queue. It could be empty because it was all consumed or because there was never any incoming stock
- Either way, add the negative qty to the queue
- When stock is added (+ve) add it to the negative balance to neutralise it until the balance is positive.
- Code cleanup: new class for `FIFOSlots` to generate chronological FIFO queue
   Usage 
   ```py 
     FIFOSlots(filters).generate()
    ```
  Screenshot 2021-12-27 at 9 08 15 PM

**Dependencies/To Test:**
- Stock Ageing Report end-to-end with various ledger scenarios (can pick up some from the .md file in this PR)
- Report warehouse wise item balance age and value
- Report stock balance with `show ageing data` checked


marination:
Sider failing on .md file, remark-lint was disabled for this repo, idk why its still failing

marination:

PR #28968, opened 2021-12-20
fix: fetch appointment letter content in the same order as template

ruchamahabal:
**Problem**:

1. Create an Appointment Letter Template with terms

    image

2. Create an Appointment Letter with that template. The terms don't get fetched in the same order as the template. The system should follow the same sequence which is in Appointment Letter Template.
  
    appointment-letter

**Fix**:

While fetching terms, order by `idx` so that the same order is followed in all the letters generated using a template.


ruchamahabal:

PR #28970, opened 2021-12-20
feat(Regional): KSA VAT report optimization and tests

### LINKED ISSUE #28915: KSA VAT Report Calculation Error when using tax templetes
### Information about bug

@ahmadpak 
When using tax templetes vat is not being reflected in KSA VAT Reports
Screenshot 2021-12-16 at 3 05 15 PM

The Vat comes in Charts of accounts but not in KSA VAT reports, 
It comes only if we select item tax. 

Video
https://www.awesomescreenshot.com/video/6401231?key=3665f063c3ad99fec54a5bf63309e23b

### Version

v13

### Module

regional

### Version

ERPNext: v13.16.1 (HEAD)

Frappe Framework: v13.16.0 (HEAD)

### Installation method

FrappeCloud

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rtdany10:
KSA VAT Report showed correct data only if Item Tax Template was provided. This PR aims to produce accurate data even if taxes are added via tax table, thereby removing dependency on item tax template.

- [x] Optimize report loading time
- [x] Add company filter
- [x] Remove item tax template dependency
- [x] Add tests
- [x] Patch to remove zero and exempt rows from KSA VAT Setting
- [x] Remove Exempt and Zero VAT account creation
- [x] Change raw SQL to frappe.qb

Closes #28915
#no-docs


erpnext/regional/report/ksa_vat/ksa_vat.py
```diff
@@ -52,124 +51,74 @@ def get_data(filters):
...
 
-		for item in invoice_items:
-			# Summing up total taxable amount
-			if invoice.is_return == 0:
-				total_taxable_amount += item.net_amount
-
-			if invoice.is_return == 1:
-				total_taxable_adjustment_amount += item.net_amount
+	for inv in invoices:
+		invoice = frappe.get_doc(doctype, inv)
```
deepeshgarg007: get_doc in loops is not recommended especially in reports as there could be a lot of records and get doc will make the report very slow. You should get the data in a single query and then use it as needed similar to the previous implementation

erpnext/regional/report/ksa_vat/ksa_vat.py
```diff
@@ -1,13 +1,12 @@
-# Copyright (c) 2013, Havenir Solutions and contributors
+# Copyright (c) 2021, Havenir Solutions, Wahni Green Technologies and contributors
```
deepeshgarg007, line 1: There is nothing wrong with this but for consistency would be better you added the standard "Frappe Technologies Pvt. Ltd. and Contributors" to this as this is for the official repo

erpnext/regional/report/ksa_vat/ksa_vat.py
```diff
@@ -52,124 +51,74 @@ def get_data(filters):
...
-	tax_amount = 0
-	if item_wise_tax_detail and len(item_wise_tax_detail) > 0:
-		item_wise_tax_detail = json.loads(item_wise_tax_detail)
-		for key, value in item_wise_tax_detail.items():
-			if key == item_code:
-				tax_amount = value[1]
-				break
-
-	return tax_amount
+			item_itemised_tax = itemised_tax.get(item)
```
deepeshgarg007: The main reason for adding the dependency of item tax template was distinguishing between zero-rated and exempted items as those have to be reported separately. That identification was done based on item tax template. How that is done in this implementation?

deepeshgarg007 (changes requested):

erpnext/regional/report/ksa_vat/ksa_vat.py
rtdany10: Agreed, will stick to that from now onward :)

erpnext/regional/report/ksa_vat/ksa_vat.py
rtdany10: Regional setup already adds `is_zero_rated` and `is_exempt` check boxes to Item doctype. Utilizing that would be better than relying on Item Tax Template. Will make changes in the code accordingly.

erpnext/regional/report/ksa_vat/ksa_vat.py
rtdany10: Done :+1:

rtdany10:
Depends on [#29175](https://github.com/frappe/erpnext/pull/29175)

erpnext/regional/report/ksa_vat/ksa_vat.py
```diff
@@ -49,140 +49,183 @@ def get_columns():
...
+	for inv in invoices:
+		acc = inv["account_head"]
+		if tax_details.get(acc):
+			tax_details[acc]["total_tax"] += inv["base_tax_amount"]
+			# 'item_wise_tax_detail': '{"Item Code":[Tax Rate, Tax Amount]}'
+			item_wise = json.loads(inv["item_wise_tax_detail"])
+			if not inv["is_return"]:
+				for x in item_wise:
+					if item_wise[x][1] > 0.0:
+						tax_details[acc]["taxable_amount"] += item_wise[x][1]*100/item_wise[x][0]
```
deepeshgarg007: Can't we just use the net total as taxable value? This back calculating may go wrong in many cases. In cases when users may add taxes as actual amount rather than based on net total

Also, my general concern is there might be users who will be using this item tax template config and might not have the nil-rated and exempted checks updated in the Item master. How do they come to know of this change? Their reports will go wrong.

The item tax template in KSA Vat Settings kind of becomes redundant now and it's not making it independent of the item tax template

rtdany10: > Can't we just use the net total as taxable value?

In an invoice with 2 items where one is taxed at 5% and the other is zero/exempt, the net total is not the taxable amount. Hence, it won't be possible to consider the net total as total taxable amount.

> This back calculating may go wrong in many cases. In cases when users may add taxes as actual amount rather than based on net total

Yes, agreed. Maybe a better way would be to use the tax rate(set in the account master) to calculate the taxable amount. That way, even if the user add tax as actual amount, we still we be able to calculate the taxable amount. 

But the tax rate field in accounts doctype isn't mandatory. How about replacing Item Tax Template with a tax rate field in the KSA VAT Settings? 

> Also, my general concern is there might be users who will be using this item tax template config and might not have the nil-rated and exempted checks updated in the Item master. How do they come to know of this change? Their reports will go wrong.

We can run a patch to tick `Is Zero Rated/Is Exempt` if that item has the corresponding Item Tax Template in the taxes table in Item master. We can also run a patch to tick those fields in transactions that are already completed.

We could also place a hook on Sales and Purchase Invoices to tick `Is Zero Rated/Is Exempt` (in the transaction as well as in the Item master) if Item Tax Template set is Zero or Exempt.

We could keep this hook for a period of time, update the docs, inform the users and then slowly remove the hook once we're confident enough.

Depending upon Item Tax Template isn't convenient as:
- It isn't a mandatory field in the transaction and people might forget to fill it.
- In some cases, people will add the correct Item Tax Template but will tamper with the taxes table. Item Tax Template remains unaffected by these tampering. In such cases, tax account ledger will have one data and VAT report another, causing audit issues.
- With issues like #29122, people prefer to stick on to Sales/Purchase Taxes and Charges Template

Therefore, I believe, it is always better and convenient to use the data from the taxes table to generate the report.
@deepeshgarg007

rtdany10:
.

erpnext/regional/report/ksa_vat/ksa_vat.py
rtdany10: @deepeshgarg007

PR #28983, opened 2021-12-21
feat: allow user to change the parent company

anupamvs:
**Issues-** 
- Currently, users are not able to change the parent company of a company.
- To view the Consolidated Financial Statements, the User would like to make these companies child companies under one Parent company.

**Fix-**
- Made Parent company field editable.
- rebuild tree, if the parent company is changed.
parent_company

docs- https://docs.erpnext.com/docs/v13/user/manual/en/setting-up/company-setup#210-change-parent-company


erpnext/setup/doctype/company/company.py
```diff
@@ -396,6 +401,13 @@ def on_trash(self):
 		if not frappe.db.get_value('GL Entry', {'company': self.name}):
 			frappe.db.sql("delete from `tabProcess Deferred Accounting` where company=%s", self.name)
 
+	def check_parent_changed(self):
+		frappe.flags.parent_company_changed = False
+
+		if not self.is_new() and \
+			self.parent_company != frappe.get_cached_value("Company",  self.name,  "parent_company"):
```
nextchamp-saqib:
```suggestion
			self.parent_company != frappe.db.get_value("Company",  self.name,  "parent_company"):
```

`frappe.get_cached_value` calls `get_doc` internally if cached doc is not found.

anupamvs:

nextchamp-saqib (approved):

PR #28988, opened 2021-12-21
style: better labels for SLE fields

ankush:
These things are very confusing to new devs, specifically, the "actual" part is quite pointless. This just renames few labels, not touching the field names for now. (too much patching and breaking change in code)

Actual Qty -> Qty change
Actual Qty After Transaction -> Qty After Transaction
Stock Value -> Balance Stock Value
Stock Value Difference -> Change in Stock Value
Stock Queue (FIFO) -> FIFO Queue (qty, rate)

### DIFF
--- a/erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.json
+++ b/erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.json
@@ -150,7 +150,7 @@
    "fieldtype": "Float",
    "in_filter": 1,
    "in_list_view": 1,
-   "label": "Actual Quantity",
+   "label": "Qty Change",
    "oldfieldname": "actual_qty",
    "oldfieldtype": "Currency",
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/stock_ledger_entry/stock_ledger_entry.json
```diff
@@ -219,14 +219,14 @@
...
-   "label": "Stock Value Difference",
+   "label": "Change in Stock Value",
    "options": "Company:company:default_currency",
    "read_only": 1
   },
   {
    "fieldname": "stock_queue",
    "fieldtype": "Text",
-   "label": "Stock Queue (FIFO)",
+   "label": "FIFO Queue (qty, rate)",
```
ankush:
```suggestion
   "label": "FIFO Stock Queue (qty, rate)",
```

PR #28996, opened 2021-12-22
fix: Issues in Bank Reconciliation tool

deepeshgarg007:
Read about bank reco in ERPNext https://docs.erpnext.com/docs/v13/user/manual/en/accounts/bank-reconciliation

The mapping of the bank accounts is maintained at the bank level. But however, one bank could have both types of bank account like Asset (Savings and Current Account) and Liability like (Credit Card Accounts). Now there was an issue while reconciling the bank transactions of the bank account of type liability. The amount to be reconciled was shown as 0 for such transactions and it became difficult to reconcile such a transaction

Steps to test the issue:
1. Create a bank and add mapping for bank statement import
2.  Make a bank account and map a company account of type liability
3. Import bank statements for the above-created bank account.
4. Add a JV for expenses using such a bank account and check


deepeshgarg007:
@Mergifyio backport develop

PR #28998, opened 2021-12-22
feat: Allowing non stock items in POS

nemesis189:
**Issue:**
- Non stock items not allowed in POS, making changes to allow the same.
- Removed item qty indicator for non stock items

`no-docs`


erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -41,7 +41,7 @@ def validate(self):
 		self.validate_serialised_or_batched_item()
 		self.validate_stock_availablility()
 		self.validate_return_items_qty()
-		self.validate_non_stock_items()
+		# self.validate_non_stock_items()
```
nextchamp-saqib:
```suggestion
```

nextchamp-saqib: Also delete the method

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -475,9 +477,12 @@ def get_stock_availability(item_code, warehouse):
...
 		return bin_qty - pos_sales_qty
-	else:
-		if frappe.db.exists('Product Bundle', item_code):
-			return get_bundle_availability(item_code, warehouse)
+	elif frappe.db.exists('Product Bundle', item_code):
+		return get_bundle_availability(item_code, warehouse)
+	#To continue the flow considering a service item
+	elif frappe.db.get_value('Item', item_code, 'is_stock_item') == 0:
+		return 0
+
```
nextchamp-saqib:
```suggestion

	else:
		if frappe.db.exists('Product Bundle', item_code):
			return get_bundle_availability(item_code, warehouse)
		else:
			# Is a service item
			return 0
```

nextchamp-saqib: Also, we need to show service items as available without quantity in the Item Details section

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -99,7 +99,7 @@ def get_items(start, page_length, price_list, item_group, pos_profile, search_te
 		), {'warehouse': warehouse}, as_dict=1)
 
 	if items_data:
-		items_data = filter_service_items(items_data)
+		# items_data = filter_service_items(items_data)
```
nextchamp-saqib:
```suggestion
```

nextchamp-saqib: Remove the `filter_service_items` method too

nextchamp-saqib (changes requested):

erpnext/selling/page/point_of_sale/pos_controller.js
```diff
@@ -637,11 +637,16 @@ erpnext.PointOfSale.Controller = class {
 		const bold_warehouse = warehouse.bold();
 		const bold_available_qty = available_qty.toString().bold()
 		if (!(available_qty > 0)) {
-			frappe.model.clear_doc(item_row.doctype, item_row.name);
-			frappe.throw({
-				title: __("Not Available"),
-				message: __('Item Code: {0} is not available under warehouse {1}.', [bold_item_code, bold_warehouse])
-			})
+			frappe.db.get_value('Item', item_row.item_code, 'is_stock_item').then(({message}) => {
```
nextchamp-saqib: This `get_value` can be eliminated if you return `is_stock_item` along with `available_qty` from whitelisted method `get_stock_availability`

nextchamp-saqib:
For service items, remove the quantity indicator from Item Selector

nextchamp-saqib:
Continued with https://github.com/frappe/erpnext/pull/29556

PR #29005, opened 2021-12-23
fix(test): Leave Allocation validation against Leave Application after submit

ruchamahabal:
**Problem**:

image

Test failing as the test data considers days in Feb (edge case): 23/2/2022 - 2/3/2022 (8 days). But the test assumes 11 days (as in a normal month), hence validation error is not raised.

**Fix:**

Use dynamic leave allocation instead of hardcoded values.


ruchamahabal:

PR #29009, opened 2021-12-23
fix: Start date validation for deferred invoices

deepeshgarg007:
For more info on deferred revenue refer: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/deferred-revenue

Service Start Date for deferred revenue booking should not be before the invoice posting date. Many times users creating invoices in the current month and enter service start dates of the previous months which cause issues while booking deferred revenue if the books for previous months are frozen and deferred revenue booking fails


deepeshgarg007:

PR #29014, opened 2021-12-23
fix!: disable renaming on serial nos

ankush:
New serial nos can simply be deleted and recreated if a mistake is made.

modifying delivered/manufactured/bought sr no should not be allowed without amending doc.

> Edit: For future reference, a lot of old data that was renamed could happen to not be appropriately renamed in the SLE since its a text field. We want to avoid hacky stuff to incorporate all this - @marination


marination (approved):

marination:
@Mergifyio version-13-hotfix

marination:

PR #29024, opened 2021-12-24
feat: Provision to assign multiple payroll cost centers against a single employee

nabinhait:
**Existing:**
Earlier we had a provision to assign a payroll cost centre against an Employee via Department/Employee master.

**New Feature:**
Now you can assign multiple cost centres against an employee. This is a very common use-case, where one employee contributes to multiple departments or cost centers, hence the salary expenses associated with that employee should be distributed against multiple cost centres.

- A new child table "Payroll Cost Centers" has been added inside "Salary Structure Assignment"
    -  Here, you can add multiple cost centres with percentages
    - If the table is left blank, it will fetch payroll cost centre from Employee/Department
    - Total percentages should be 100

Screenshot 2021-12-24 at 4 42 10 PM

Docs: https://docs.erpnext.com/compare?wiki_page=docs/v13/user/manual/en/human-resources/payroll-entry&&compare=1a037a4c97 (the docs changes are already live by mistake)


ruchamahabal:
Not fixing sider issues for frappe.qb. Not helpful

ruchamahabal:
Raised a PR for fixing unrelated failing tests: https://github.com/frappe/erpnext/pull/29051

ruchamahabal (approved):

Issue #29029, opened 2021-12-25
invoice not getting currency when  created from Subscription base on Subscription Plan Currency


MohammedAlNazili:
### Information about bug

After creating a subscription plan based on different company default currency and adding a subscription for customer and price base one price list.
when creating invoices it passes the default currency and the balance show not correct

**Subscription Plan**
![1](https://user-images.githubusercontent.com/56793163/147381493-c23d7d89-1bf2-4d73-91f6-604b5c8db801.png)
**Item Price**
![2](https://user-images.githubusercontent.com/56793163/147381495-d1bb6d49-eeb8-4f4c-b07c-b1cf0582a439.png)
**Sales Invoices**
![3](https://user-images.githubusercontent.com/56793163/147381498-46cde96b-a7ee-4811-81d9-028c6ec85282.png)

### Module

accounts

### Version

ERPNext: v13.17.0 (version-13)

Frappe Framework: v13.17.1 (version-13)

### Installation method

manual install

deepeshgarg007:
Fixed via https://github.com/frappe/erpnext/pull/28916

PR #29036, opened 2021-12-26
Subscription currency

MohammedAlNazili:
https://github.com/frappe/erpnext/issues/29029#issue-1088571522

Base on issue we have created ...

After creating a subscription plan based on different company default currency and adding a subscription for customer and price base one price list.
when creating invoices it passes the default currency and the balance show not correct

**Subscription Plan**

**Item Price**

**Sales Invoices**

### Module

accounts

### Version

ERPNext: v13.17.0 (version-13)

Frappe Framework: v13.17.1 (version-13)

### Installation method

manual install

_No response_

### DIFF
--- a/erpnext/accounts/doctype/subscription/subscription.py
+++ b/erpnext/accounts/doctype/subscription/subscription.py
@@ -275,6 +275,15 @@ class Subscription(Document):
 		self.validate_end_date()
 		self.validate_to_follow_calendar_months()
 		self.cost_center = erpnext.get_default_cost_center(self.get('company'))
+		self.validate_plan_currency()
+
+	def validate_plan_currency(self):
+		for p in self.plans:
+			plan_currency = frappe.db.get_value("Subscription Plan",p.plan, ["currency"])
+			for pp in self.plans:
+				cur_currency = frappe.db.get_value("Subscription Plan",pp.plan, ["currency"])
+				if cur_currency!=plan_currency:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -371,10 +380,12 @@ def create_invoice(self, prorate):
 
 		# Subscription is better suited for service items. I won't update `update_stock`
 		# for that reason
-		items_list = self.get_items_from_plans(self.plans, prorate)
+		items_list, temp_currency = self.get_items_from_plans(self.plans, prorate)
 		for item in items_list:
 			item['cost_center'] = self.cost_center
 			invoice.append('items', item)
+			# currency
+			invoice.currency = temp_currency
```
deepeshgarg007, line 388: @MohammedAlNazili this doesn't look like a correct fix to me. One subscription could have multiple subscription plans and by this logic, the invoice will be made in the currency of the last subscription plan in the child table.
The invoice current should be the party account currency and independent of the subscription plan

I have tried to resolve this in the following PR. Please update your site and check if this helps
https://github.com/frappe/erpnext/pull/28916

For now closing this PR

MohammedAlNazili: I tried to make subscriptions accept subscription plans in one currency in entry and by accepting one currency I can take first or last.
but with Customer currency, I think it is better.
Thanks

PR #29042, opened 2021-12-27
fix: ignore links while setting default notification templates in Settings

ruchamahabal:
Backport of: https://github.com/frappe/erpnext/pull/29019


ruchamahabal:
test fix will be backported separately

PR #29046, opened 2021-12-27
fix: filter out Claimed employee advances in Expense Claim

ruchamahabal:
## Problem:

1. Create an Employee Advance. Create a Payment Entry to pay the advance.
2. Create an Expense Claim against the advance, now the entire advance is claimed.
    
    claimed-advanc
     
3. Create a new Expense Claim, the advance still shows up for selection. 

    advance-shown

## Fix:

Filter out already claimed advances for selection. There was already a filter existing in the current codebase for:

`['paid_amount', '>', 'claimed_amount']`

However, search uses `get_list`, and column-based comparison is not supported in `get_list` or `get_all`. It was recently added with the query builder only for v14.

Easier way to filter out such advances is directly by the status field:

`['status', '!=', 'Claimed']`

not-available-adv


ruchamahabal:

PR #29055, opened 2021-12-28
fix: rename non existent doctype field to the right one

developmentforpeople:
Background job for settings membership expired status: `set_expired_status` considers `status` (non existent field) instead of `membership_status`


ruchamahabal (approved):

ruchamahabal:

PR #29062, opened 2021-12-28
fix: erase UOMs when changing stock UOM

anupamvs:
**Issue-** When item's default stock UOM is changed, the old conversion factors become unusable, erase UOM conversion list upon updating stock UOM. 

Screenshot 2022-01-12 at 7 25 45 PM

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -492,18 +492,20 @@ class Item(WebsiteGenerator):
 		context.shopping_cart = get_product_info_for_website(self.name, skip_quotation_creation=True)
 
 	def add_default_uom_in_conversion_factor_table(self):
-		uom_conv_list = [d.uom for d in self.get("uoms")]
-		if self.stock_uom not in uom_conv_list:
-			ch = self.append('uoms', {})
-			ch.uom = self.stock_uom
-			ch.conversion_factor = 1
+		if not self.is_new() and self.has_value_changed("stock_uom"):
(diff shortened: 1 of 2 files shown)


anupamvs:

erpnext/stock/doctype/item/item.py
```diff
@@ -492,18 +492,26 @@ def set_shopping_cart_data(self, context):
...
+		uoms_list = []
+		if not self.is_new():
+			if self.stock_uom != frappe.get_value("Item", self.name, "stock_uom"):
+				idx = 1
+				for d in self.get("uoms"):
+					if d.conversion_factor == 1 and d.uom != self.stock_uom:
+						self.remove(d)
+					else:
+						d.idx = idx
+						idx += 1
```
ankush: 1. mutating list while iterating over it will lead to runtime errors.
2. it's simpler to just reset the index instead of this complicated logic. Once all modifications are done...

```python
for idx, row in enumerate(doc.child_table, start=1):
    row.idx = idx
```

ankush (changes requested):

ankush (changes requested):
just realized something very basic. 

When stock UOM change, the old conversions don't make sense at all. They all need to be removed 😄 

e.g. 
- Nos is default stock UOM
- 20 Nos = 1 Box
- You change default stock UOM to Kg and now conversion is 20 Nos = 1KG now.

IMO lets just erase the whole table and show toast that UOM conversions are cleared.

PR #29067, opened 2021-12-29
fix: broken patches

saurabh6790:

> Screenshots/GIFs


erpnext/patches/v13_0/make_homepage_products_website_items.py
```diff
@@ -13,4 +13,7 @@ def execute():
 		row.item_code = web_item
 
 	homepage.flags.ignore_mandatory = True
-	homepage.save()
\ No newline at end of file
+	try:
+		homepage.save()
+	except frappe.exceptions.LinkValidationError:
```
ankush: ignore links instead? ignoring exception will commit a potentially partial change. 

`homepage.flags.ingore_links = True`

erpnext/patches/v13_0/update_old_loans.py
```diff
@@ -24,6 +24,15 @@ def execute():
 	frappe.reload_doc('loan_management', 'doctype', 'loan_repayment')
 	frappe.reload_doc('loan_management', 'doctype', 'loan_repayment_detail')
 	frappe.reload_doc('loan_management', 'doctype', 'loan_interest_accrual')
+
+	#handle type casting for is_cancelled field
+	frappe.db.sql('''UPDATE `tabGL Entry` SET is_cancelled =
```
ankush: this is a change in GL entry doctype so doesn't belong in `update_old_loans.py`; either separate patch or reordering of patches is required.

ankush (changes requested):

saurabh6790:
@ankush fixed suggested changes !

deepeshgarg007:
@saurabh6790 can you fix the linting issues. You'll have to set up pre-commit locally as well. For this you can follow
https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

erpnext/patches/v12_0/update_is_cancelled_field.py
```diff
@@ -3,14 +3,21 @@
...
-	try:
-		frappe.db.sql("UPDATE `tabStock Ledger Entry` SET is_cancelled = 0 where is_cancelled in ('', NULL, 'No')")
-		frappe.db.sql("UPDATE `tabSerial No` SET is_cancelled = 0 where is_cancelled in ('', NULL, 'No')")
+	#handle type casting for is_cancelled field
 
-		frappe.db.sql("UPDATE `tabStock Ledger Entry` SET is_cancelled = 1 where is_cancelled = 'Yes'")
-		frappe.db.sql("UPDATE `tabSerial No` SET is_cancelled = 1 where is_cancelled = 'Yes'")
+	for doc_mapper in (('stock','Stock Ledger Entry'),
+		('stock','Serial No'),
+		('accounts','GL Entry')):
```
ankush: @deepeshgarg007 @saurabh6790 

GL entry has never had this select field for `is_cancelled` in any production version prior to v13; we need to check if this column even exists before running the query or patch will fail. 

Is this to accommodate any beta branch which had yes/no for is_cancelled?

ankush (changes requested):

erpnext/patches/v12_0/update_is_cancelled_field.py
saurabh6790: Not remembering  the site but yes added this as it was breaking in v11 to v13 migrations.

ankush:

erpnext/patches/v12_0/update_is_cancelled_field.py
```diff
@@ -3,14 +3,19 @@
...
+	)
 
-		frappe.db.sql("UPDATE `tabStock Ledger Entry` SET is_cancelled = 1 where is_cancelled = 'Yes'")
-		frappe.db.sql("UPDATE `tabSerial No` SET is_cancelled = 1 where is_cancelled = 'Yes'")
+	for module, doctype in module_doctypes:
+		if frappe.db.has_column(doctype, "is_cancelled"):
+			frappe.db.sql('''UPDATE `tab{doctype}` SET is_cancelled = CASE
+					WHEN is_cancelled = 'No' THEN 0
+					WHEN is_cancelled = 'Yes' THEN 1
+					ELSE 0
```
ankush: If this patch is retried everything will become 0 😄

ankush:
@mergifyio copy develop

saurabh6790:
@ankush are you working on this?  It's been 27 days now!

+1 for adding the check of field type.

But I still believe, there is no need of two queries. The field type check and sql case statement will do job for you in one go.

ankush:
> there is no need of two queries.

Added it before adding type check cause re-running could cause the entire column to become 0; best to leave it... it doesn't take much time and it runs one time only.

PR #29072, opened 2021-12-29
fix: Must not be able to start Job Card if it is related to Work Order that is not started yet

### LINKED ISSUE #29071: Cannot start Job Card if related to Work Order not started yet
Create a WO (from BOM with operation and Material transfer for WO)
Submit WO => Start WO => Stock transfer to WIP warehouse

Unfortunately you don't have enough stock to transfer to WIP, so you have WO in Submit status, Stock Entry in Draft, and Job card in draft.

On job card, you shouldn't be able to start counting time if work order is not started.

FHenry:
related to #29071


FHenry:
Hi @marination Do you need more explanation from me about this PR ?

johnlongland:
It will be great if this can be included into the core. Any idea when this might happen ?

erpnext/manufacturing/doctype/job_card/job_card.js
```diff
@@ -73,7 +73,24 @@ frappe.ui.form.on('Job Card', {
 		if (frm.doc.docstatus == 0 && !frm.is_new() &&
 			(frm.doc.for_quantity > frm.doc.total_completed_qty || !frm.doc.for_quantity)
 			&& (frm.doc.items || !frm.doc.items.length || frm.doc.for_quantity == frm.doc.transferred_qty)) {
-			frm.trigger("prepare_timer_buttons");
+
+			//if Job Card is link to Work Order, the job card must not be able to start if Work Order not "Started"
+			// and if stock mvt for WIP is required
+			if (frm.doc.work_order) {
+				frappe.db.get_value('Work Order', frm.doc.work_order, 'skip_transfer').then((r) => {
```
marination: Lets get whatever we need in one db call. `result` will have to be accessed as per whatever is returned (`r.message` wont be required)
```suggestion
				frappe.db.get_value('Work Order', frm.doc.work_order, ['skip_transfer', 'status']).then((result) => {
```

erpnext/manufacturing/doctype/job_card/job_card.js
```diff
@@ -73,7 +73,24 @@ frappe.ui.form.on('Job Card', {
...
+					if (r.message.skip_transfer !== 1 ) {
```
marination: The last else can be eliminated by combining both if conditions.
Also does "Completed" work order need a timer button?
```suggestion
					if (result.skip_transfer !== 1  && result.status == 'In Process') {
					     frm.trigger("prepare_timer_buttons");
					}
```

FHenry:
@marination I've follow your instruction, and it's more readable like that. But result is not enought, we need result.message, like is it done line 95. It seems that frappe.db.get_value return result as r.message

marination (dismissed):

marination (approved):

FHenry:

PR #29085, opened 2021-12-30
fix(Selling,E-Commerce): Shopping cart quotation without website item.

dj12djdjs:
**Problem / Motivation**
Our customer is browsing our e-store and cannot find what they are looking for. They call a customer service rep to see if the item is available. The rep adds the item into the quotation and saves it. Without realizing, that item hasn't been added as a Website Item yet. 

The customer goes to visit their shopping cart to realize the web page is broken. Caused by: https://github.com/frappe/erpnext/blob/498d933e9ce7c18d6c595d2e0f575645dc020d81/erpnext/e_commerce/shopping_cart/cart.py#L277-L286
<details>
<summary>Error Snapshot</summary>

image
</details>

**Proposed Resolution**
If the quotation is of type Shopping Cart validate that all item's have a Website Item associated with it.
Screenshot 2022-04-22 at 1 21 03 PM


dj12djdjs:
Failing test is unrelated to this PR

dj12djdjs:
@marination I think this PR will need the e-commerce-fix label since it validates with Website Item doctype not available in develop yet.

dj12djdjs:
@deepeshgarg007 @nextchamp-saqib any thoughts about this?

marination:
@dj12djdjs sorry haven't been able to actively review PRs. This seems valid.

PS: You can open future e-commerce related stuff on develop. Its merged there too

erpnext/selling/doctype/quotation/quotation.py
```diff
@@ -38,6 +39,14 @@ def validate_valid_till(self):
 		if self.valid_till and getdate(self.valid_till) < getdate(self.transaction_date):
 			frappe.throw(_("Valid till date cannot be before transaction date"))
 
+	def validate_shopping_cart_items(self):
+		if self.order_type != "Shopping Cart":
+			return
+
+		for item in self.items:
+			if not frappe.db.exists("Website Item", {"item_code": item.item_code}):
+				frappe.throw(_("Item {0} must be a website item for Shopping Cart quotations").format(item.item_code))
```
marination:
```suggestion
            has_web_item = frappe.db.exists("Website Item", {"item_code": item.item_code})
            template = frappe.db.get_value("Item", item.item_code, "variant_of")
            if not has_web_item and template:
                has_web_item = frappe.db.exists("Website Item", {"item_code": has_template})
                
			if not has_web_item:
				frappe.throw(_("Row #{0}: Item {1} must be a Website Item for Shopping Cart quotations").format(item.idx, item.item_code))
```

**The condition here needs to _not_ throw the validation if it is a variant item whose template is published.**

Item variants without website items are allowed to be added to cart. I've encountered that some users do not create individual web pages for variants at all, and here Website Item seems wasteful.
Reference: https://github.com/frappe/erpnext/pull/29383

Fixing this should fix the failing server side test on this PR

dj12djdjs: This is good to know. I never use variants so I didn't even think of this.

marination (approved):

PR #29086, opened 2021-12-30
fix: leave_allocation variable not being defined

sudo-vaibhav:
ISSUE

Recently while hosting erpnext for my company, I got an error while doing leave allocations, as mentioned in this issue.
https://discuss.erpnext.com/t/leave-policy-bulk-allocation-error/84246

This is a fix, which moves the assignment inside the if check, which makes more sense to me, and should prevent this error from occuring.


ruchamahabal (approved):

ruchamahabal:

PR #29114, opened 2022-01-03
fix: from time and to time not updated in drag and drop action

### LINKED ISSUE #28727: Drag and drop function doesn't work in Course Schedule Calendar
## Description of the issue

Try to drag an drop some Course Schedule in Calendar on monthe, week or day.
It is seems like works, but after reloading of page falls into back.

**Output of `bench version`**
```
ERPNext: v13.13.0 (version-13)

Frappe Framework: v13.13.0 (version-13)
```

1. Go to Course Schedule Calendar /app/course-schedule/view/calendar/default
2. Choose Calendar View.
3. Drag some course schedule and drop it to another time or day.
4. Reload page.

### Observed result

After reloading of page course schedule falls into back.

### Expected result

New time or date should be save in Course Schedule.

```
(paste here)
```

## Additional information

https://user-images.githubusercontent.com/3992823/144697982-a83d759c-dae4-4969-96a3-b1d66d3fecdd.mp4

mohammedyusufshaikh:
Before Fix: 

The Course Schedule Calendar did not update **from time** and **to time** when a drag and drop action is performed in time slots of a day 
drag_and_drop_in_time_slots

After Fix:

**from time** and **to time** updated correctly so new changes are saved

drag_and_drop_in_time_slots_fix

This issue was identified from here [GitHub Issue](https://github.com/frappe/erpnext/issues/28727)

There is also a bug identified from this issue which doesn't allow user to drag and drop events to a different date which required a fix in the framework. A PR has been raised for the same [#15509](https://github.com/frappe/frappe/pull/15509)

### DIFF
--- a/erpnext/education/doctype/course_schedule/course_schedule_calendar.js
+++ b/erpnext/education/doctype/course_schedule/course_schedule_calendar.js
@@ -1,12 +1,10 @@
 frappe.views.calendar["Course Schedule"] = {
 	field_map: {
-		// from_datetime and to_datetime don't exist as docfields but are used in onload
 		"start": "from_time",
 		"end": "to_time",
 		"id": "name",
 		"title": "course",
 		"allDay": "allDay",
-		"schedule_date": "schedule_date"
 	},
 	gantt: false,
 	order_by: "schedule_date",
--- a/erpnext/education/doctype/course_schedule/course_schedule_calendar.js
+++ b/erpnext/education/doctype/course_schedule/course_schedule_calendar.js
@@ -1,11 +1,12 @@
 frappe.views.calendar["Course Schedule"] = {
 	field_map: {
 		// from_datetime and to_datetime don't exist as docfields but are used in onload
-		"start": "from_datetime",
-		"end": "to_datetime",
+		"start": "from_time",
+		"end": "to_time",
 		"id": "name",
 		"title": "course",
-		"allDay": "allDay"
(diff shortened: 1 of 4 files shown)


erpnext/education/doctype/course_schedule/course_schedule_calendar.js
```diff
@@ -1,11 +1,12 @@
 frappe.views.calendar["Course Schedule"] = {
 	field_map: {
 		// from_datetime and to_datetime don't exist as docfields but are used in onload
```
ankush:
```suggestion
```

misleading old comment.

erpnext/education/doctype/course_schedule/course_schedule_calendar.js
```diff
@@ -1,11 +1,12 @@
...
-		"start": "from_datetime",
-		"end": "to_datetime",
+		"start": "from_time",
+		"end": "to_time",
 		"id": "name",
 		"title": "course",
-		"allDay": "allDay"
+		"allDay": "allDay",
+		"schedule_date": "schedule_date"
```
ankush:
```suggestion
		"allDay": "allDay"
```

this bit can be handled in doctype controller like we discussed. 

Basically if the start field is DateTime and schedule date != start date then it can be updated in validating function. (add comment there as it's handling very specific case - calendar updates)

PR #29115, opened 2022-01-03
fix: Deferred revenue booking for multi currency invoices via JV

deepeshgarg007:
Deferred Revenue for multi-currency Invoices was not booked for the correct amount if booked via JV. Is was still considering the amount in party currency rather than the base amount


deepeshgarg007:

PR #29117, opened 2022-01-03
chore: remove framework patch for custom fields

shadrak98:
While migrating from older versions to the latest versions few patches were breaking.
This patch is moved to framework


erpnext/patches/v13_0/item_reposting_for_incorrect_sl_and_gl.py
```diff
@@ -12,6 +12,7 @@ def execute():
 			("stock", "purchase_receipt_item"),
 			("stock", "delivery_note_item"),
 			("stock", "packed_item"),
+			("stock", "stock_ledger_entry"),
```
ankush: This is already done in a separate patch. It's not wise to reload it here.

Schema for SLE actually changed between v12 and v13 and hence the patch that migrates schema should only touch it.

ankush (changes requested):

erpnext/patches/v13_0/item_reposting_for_incorrect_sl_and_gl.py
shadrak98: The patches were still breaking during migration.. do we need to reorder the patch then?

ankush: the patch already runs before this one; it's called "update_is_cancelled_field"

ankush: https://github.com/frappe/erpnext/blob/964b911b32d71e3069ad0f942b447cd25cb4da1c/erpnext/patches/v12_0/update_is_cancelled_field.py#L12

shadrak98: Ahh, this is strange. Nvm, I will remove it. Thanks!

erpnext/patches/v13_0/validate_options_for_data_field.py
```diff
@@ -9,18 +9,18 @@
...
-                    options=NULL
-                WHERE
-                    name=%s
-            """, (field))
+        if field.options not in data_field_options:
+            custom_field = frappe.qb.DocType('Custom Field')
+            frappe.qb.update(
+                custom_field
+            ).set(
+                custom_field.options, "NULL"
```
ankush: this is the string `"null"` and not SQL `NULL`. 😄 

It should be `None` IIRC

ankush: also this whole patch doesn't belong in ERPNext. Framework added this validation, so patch/fix should be in framework.

shadrak98: Yeah thanks for pointing this out!
Will close this one and raise a PR in FF

ankush: instead of closing this, reuse it to remove that framework patch? 😅

PR #29126, opened 2022-01-04
fix: POS items added to cart despite low qty

nemesis189:
**Issue:**
- Go to point of sale.
- Add items into the cart.
- Notice if the available quantity of the item is 2 it will still keep on adding the item in the cart. (It gives an error but increases the number of item in the cart).

**Expected Result:**
Although we are getting the error, the item gets added in the cart.If the item is not available it should not get added in the cart as well.

**Fix:**
Item is not adde to cart and error is thrown instead of alert


nextchamp-saqib (approved):

nextchamp-saqib:

PR #29137, opened 2022-01-04
feat: Payment Terms Status report

### LINKED ISSUE #28758: Generate Invoice based on Payment Terms
Here are the points to be considered or challenges when creating a payment against a Payment Term:
**Challenge:** 
Since the Amount in the Sales Invoice is determined based on Rate and Qty, in order to do 50% billing for an invoice (for eg.) we will have to set either qty to 50% of rate to 50%. Resetting Qty may not be ideal, as it can lead to an issue if “Update Stock” is also checked. We cannot deliver 0.50 TV. Even now, we change the item’s Qty to get the pending invoice amount in the invoice.

**Payment Terms:**
If an invoice is made against a Payment Term, then there should be no other Payment Term fetched from Sales Order. Currently, we fetch all the payment terms of the sales order into the invoice.

**Benefits**
This design will help us extra a reports like:
**Sales Order pending to be billed for a specific term**
**Sales Order outstanding as per the payment terms**

ruthra-kumar:
## Payment Terms Status Report
This report aims to provide current status of payment terms based on the associated invoices against the SO.
All invoices associated with Sales Order are fetched and assigned to terms in FIFO method. Statuses shown in the report are calculated at runtime and does not affect the state in the database.
payment terms status

When Individual SO's are selected, a chart is used to show the payment and paid amount for each terms.
charts for Payment terms

Documentation can be found [**here**]( https://docs.erpnext.com/docs/v13/user/manual/en/accounts/payment_terms_status_report )


ruthra-kumar:
commenting to remove inactive flag

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.js
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2016, Frappe Technologies Pvt. Ltd. and contributors
```
deepeshgarg007:
```suggestion
// Copyright (c) 2022, Frappe Technologies Pvt. Ltd. and contributors
```

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
```diff
@@ -0,0 +1,211 @@
...
+	columns = [
+		{
+			"label": _("Sales Order"),
+			"fieldname": "name",
+			"fieldtype": "Link",
+			"options": "Sales Order",
+			"read_only": 1,
+		},
+		{
+			"label": _("Submitted"),
```
deepeshgarg007:
```suggestion
			"label": _("Posting Date"),
```

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
```diff
@@ -0,0 +1,211 @@
...
+			"fieldname": "submitted",
+			"fieldtype": "Date",
+			"read_only": 1
+		},
+		{
+			"label": _("Payment Term"),
```
deepeshgarg007, line 23:
```suggestion
			"label": _("Payment Terms Template"),
```

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
```diff
@@ -0,0 +1,211 @@
...
+		{
+			"label": _("Payment Amount"),
+			"fieldname": "payment_amount",
+			"fieldtype": "Currency",
+			"read_only": 1,
+		},
+		{
+			"label": _("Paid Amount"),
+			"fieldname": "paid_amount",
+			"fieldtype": "Currency",
```
deepeshgarg007, line 51: Different companies could have different default currencies. Currently for all companies current from Global Defaults will be picked. You'll have to add a hidden column for current and use that as Options for all the Currency columns to avoid this

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def get_columns():
+	columns = [
+		{
+			"label": _("Sales Order"),
+			"fieldname": "name",
+			"fieldtype": "Link",
+			"options": "Sales Order",
+			"read_only": 1,
```
deepeshgarg007: Is read-only applicable here? I mean here columns are by default read-only or some new feature has been introduced recently

deepeshgarg007 (changes requested):

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
ruthra-kumar: This field denotes the individual term in the template. I think, labelling it 'Payment Terms Template' would cause confusion.

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
ruthra-kumar: removed.

erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
ruthra-kumar: added new hidden column as suggested

deepeshgarg007:

letsfkccp:
How is this feature going? Thank you.
-Generate Invoice based on Payment Terms

PR #29146, opened 2022-01-04
fix(UX): validate setup on clicking Mark Attendance button in Shift Type

ruchamahabal:
## Problem:

The **Mark Attendance** button on shift type calls the `process_auto_attendance` class method

image

This method only processes attendance if the setup for fields is done in Shift Type. However, if the setup is missing user is not prompted and a message for successful attendance marking is shown which is perceived as a bug since no attendance records are created.

https://github.com/frappe/erpnext/blob/32197355add09ff4626dc704bc02a42f003d8b85/erpnext/hr/doctype/shift_type/shift_type.py#L18-L20

## Fix:

Explicitly check for missing configurations (Enable Auto Attendance, Process Attendance After, Last Sync of Checkin) before calling the method. 

image

### DIFF
--- a/erpnext/hr/doctype/shift_type/shift_type.js
+++ b/erpnext/hr/doctype/shift_type/shift_type.js
@@ -4,15 +4,32 @@
 frappe.ui.form.on('Shift Type', {
 	refresh: function(frm) {
 		frm.add_custom_button(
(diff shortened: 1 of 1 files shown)


ruchamahabal:

PR #29163, opened 2022-01-06
fix: Earned Leave allocation from Leave Policy Assignment

ruchamahabal:
## Problem:

Earned Leave allocation via Leave Policy Assignment is incorrect.

**Steps to replicate:**

1. Create a Leave Type for Earned Leave with monthly frequency
    image
2. Create a Leave Policy with Earned Leave
3. Create Leave Policy Assignment based on Leave Period. The Leave Allocation created from this assignment should allocate 0 leaves first since it is an Earned Leave and the scheduler will allocate leaves based on the set frequency (eg: if the policy is for 12 earned leaves for Monthly frequency, then the employee should get 0 leaves at first and 1 leave allocated at the end of every month). However, currently, it allocates all 12 leaves on assignment submission. 

   earned-leave-bug

## Fix:

The `get_leaves_for_passed_months` does not reset leaves allocated if months passed are 0 i.e if the allocation is done in the first month of the leave period

earned-leave-assignment-corrected


ruchamahabal:

PR #29175, opened 2022-01-06
fix: seperate custom field creation for KSA

### LINKED ISSUE #29110: Emirate Field Wrongly Created for Country = KSA
### Information about bug

Emirate requirement in address or anywhere else is mainly for UAE based users. If a company is created where country = Kingdom of Saudi Arabia during the setup the address doctype adds a field for Emirates:

The system lazily copies the setup from UAE one:

https://github.com/frappe/erpnext/blob/f07ff92a35c876e024703856cc2c0d31437b6dd5/erpnext/regional/united_arab_emirates/setup.py#L59

https://github.com/frappe/erpnext/blob/f07ff92a35c876e024703856cc2c0d31437b6dd5/erpnext/regional/united_arab_emirates/setup.py#L105

image
image

These are both for an account created for a company setup in KSA

Expected result: Emirates field should only show up when country is selected as UAE/United Arab Emirates

### Module

accounts

### Version

13

### Installation method

_No response_

_No response_

rtdany10:
If a company is created where country is Kingdom of Saudi Arabia, the system lazily copies the custom field creation from UAE setup. This PR aims to add only KSA specific fields upon setup.
Closes #29110

### DIFF
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -312,7 +312,7 @@ erpnext.patches.v14_0.delete_healthcare_doctypes
 erpnext.patches.v13_0.update_category_in_ltds_certificate
 erpnext.patches.v13_0.create_pan_field_for_india #2
 erpnext.patches.v14_0.delete_hub_doctypes
-erpnext.patches.v13_0.create_ksa_vat_custom_fields
+erpnext.patches.v13_0.create_ksa_vat_custom_fields # 07-01-2022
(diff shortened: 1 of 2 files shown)


rtdany10:
We will need to run custom field creation patch once more for existing users. (https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v13_0/create_ksa_vat_custom_fields.py)

Will shifting the order in patches.txt suffice?
@deepeshgarg007

deepeshgarg007 (approved):

PR #29176, opened 2022-01-06
fix: Inconsistency in calculating outstanding amount

deepeshgarg007:
Screenshot 2022-01-07 at 11 55 21 AM

image

There was inconsistency in calculating the grand total and outstanding amount which led to multiple errors like an unwanted round off entry being passed even though rounding was disabled

The status of the invoice gets changed to partly paid even if there was no payment against the invoice.

The issue is replicable only in some specific scenarios one of which is as follows. To replicate follow these steps

Step 1: Make a new purchase invoice.
Step 2: Change the currency to USD (should be a multicurrency invoice) and keep the exchange rate as 3.75
Step 3: Add two items with a rate of 437.5, and don't apply any taxes.
Step 4: Disable rounded total and save the invoice.


deepeshgarg007:

PR #29179, opened 2022-01-06
fix: incorrect scrap item qty

rohitwaghchaure:
## **Issue**

The scrap qty in the Manufacture stock entry calculated incorrectly

Steps to replicate the issue

1. Create work order with qty 1000 and 2 operations
2. Enable "Skip Material Transfer to WIP Warehouse" in the work order and submit it.
2. On submission of the work order, system has created 2 job card with 1000 qty each against 2 operations
3. Modified "Qty To Manufacture" from 1000 to 800 in the first job and added two scrap items as per below image and then submitted

Screenshot 2022-01-07 at 12 39 51 AM

5. Open the second job card and Modified "Qty To Manufacture" from 1000 to 800 in the job card and added two scrap items as per below image and then submitted

Screenshot 2022-01-07 at 12 41 46 AM

6. After that Open the Work Order and clicked on finish button, system has open the popup to set the Qty for Manufacture
7. Set the Qty for Manufacture as 800 and clicked on Create button
Screenshot 2022-01-07 at 12 44 30 AM

8. System has opened the stock entry where the scrap items are showing the incorrect quantity 

Screenshot 2022-01-06 at 2 11 35 PM

First item should show qty as 40 and second as 25

## **After Fix**

System showing the correct qty for the scrap items
Screenshot 2022-01-06 at 2 38 04 PM

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/test_work_order.py
+++ b/erpnext/manufacturing/doctype/work_order/test_work_order.py
@@ -2,7 +2,7 @@
 # License: GNU General Public License v3. See license.txt
 
 import frappe
-from frappe.utils import add_months, cint, flt, now, today
+from frappe.utils import add_days, add_months, cint, flt, now, today
 
 from erpnext.manufacturing.doctype.job_card.job_card import JobCardCancelError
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:

PR #29182, opened 2022-01-07
fix: get project from PO into payment entry

### LINKED ISSUE #16662: Payment entry from Purchase Invoice does NOT catch Project from PO line
I want to check profitability on Project base and encounter one problem with payment entries created from a Purchase Invoice.

The item in my PO is linked to a certain project

### expected behavior 
when I use "Make Payment" from that PO, the Project which is connected to the item lines in the PO should be fetched in the payment entry automatically

### actual behavior
the connected Project is **not** fetched automatically in the PE and has to be selected manually

ruthra-kumar:
Fix for [Issue 16662]( https://github.com/frappe/erpnext/issues/16662)

Now, creating payments for Purchase orders will pull the project from the line items.

output

`no-docs`


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1522,6 +1523,8 @@ def get_payment_entry(dt, dn, party_amount=None, bank_account=None, bank_amount=
 	pe.paid_amount = paid_amount
 	pe.received_amount = received_amount
 	pe.letter_head = doc.get("letter_head")
+	if dt == 'Purchase Order':
```
deepeshgarg007: This will also be applicable for other transactions like "Sales Order", "Purchase Receipt", "Delivery Note", "Sales Invoice" and "Purchase Invoice". Can we make this more generic to handle for all doctypes?

ruthra-kumar: We can avoid 'Purchase Receipt' and 'Delivery Note', as there are no process available to create Payment Entry from them.

deepeshgarg007:

PR #29185, opened 2022-01-07
feat(UX): Option to exclude holidays while marking monthly attendance

ruchamahabal:
## Problem:

While marking monthly Attendance in bulk for an employee from the attendance list view, the Attendance checklist includes all the days of the month for which attendance is not marked. The user has to manually uncheck holidays while doing so which is cumbersome.

holidays

## Fix:

Add an option to **Exclude Holidays** while marking attendance in bulk.

exclude-holidays

Docs: https://docs.erpnext.com/docs/v13/user/manual/en/human-resources/attendance/edit?wiki_page_patch=8cd07a867b


ruchamahabal:

PR #29189, opened 2022-01-07
fix(asset): check if fixed asset account is set against company

nextchamp-saqib:
If you create an asset against a company and there are no accounts set in the asset category for that company, an error "Account not found" is thrown while making ledger entries on submission of asset.


erpnext/assets/doctype/asset/asset.py
```diff
@@ -608,7 +608,16 @@ def get_purchase_document(self):
...
+		fixed_asset_account = get_asset_category_account('fixed_asset_account', None, self.name, None, self.asset_category, self.company)
+		if not fixed_asset_account:
+			frappe.throw(
+				_("Set {0} in asset category {1} for company {2}").format(
+					frappe.bold("Fixed Asset Account"),
+					frappe.bold(self.asset_category),
+					frappe.bold(self.company),
+				),
+				title=_("Account not Found"),
+			)
```
nextchamp-saqib, line 620:
```suggestion
			)
		return fixed_asset_account
```

PR #29208, opened 2022-01-10
fix(India): NIL Rated, Exempted and non gst invoices in GSTR-1 report

deepeshgarg007:
Added Nil Rated/Exempted and Non-GST section in GSTR-1 report

image

Steps to check:
1. Create Exempted Item (Check "Is Nil Rated and Exempted" check in item master)
2. Make a regular GST Invoice and select the Exempted item
3. Base on the type of customer (Register or Unregistered) and place of supply (Instate or out state) the taxable value should be shown accordingly in the report as shown above


deepeshgarg007:

PR #29211, opened 2022-01-10
fix: Exclude unpublished items while fetching items from other item groups

marination:
**Issue:**
- Assume Item A belonging to Item Group **Products** and Item B belonging to Item Group**Accessories**
- Add **Accessories** to Website Item A's **Website Item Groups table**.
- In Item Group **Accessories** page, we expect to see Item B and also Item A (which actually belongs to **Products** but should be visible in **Accessories** as well)
- Now, **unpublish** Item A (untick the published button in the website item)
- Item A is still visible in the **Accessories** page (it should not be)

**Fix:**
- Add `published` as a filter while fetching Items from Website Item Groups table too


marination:
Exchange Rate tests are failing because exchangerate API service is down
Also feels trivial to write a test for this ( pls dont kill me this wont break again i swear)

PR #29216, opened 2022-01-10
fix: use `get_all` instead of `get_list` while fetching SLA doctypes

ruchamahabal:
`get_sla_doctypes` is called on the `app_ready` event to show the SLA dashboard on whichever doctypes it is configured for. Due to this, users who have nothing to do with SLA get insufficient permission pop-up

image

Use `get_all` instead of instead of `get_list` to ignore permissions


ruchamahabal:

PR #29219, opened 2022-01-10
fix: Behaviour of Item Variants Cache generation

marination:
**Issue:**
- Currently for certain attributes to be pulled in the **Item Configuration** popup, the Variant Item needs to be published or needs to be a Website Item.
- E.g. if colour **Red** must be there in the popup, at least one variant with colour 'Red' must be a Website Item.
- In some cases users don't want full page views/Website Items for variants. They only want it to be selectable in the Variant Selector.
- **Website Item and Variant popup should ideally have no direct dependence.**
- This resulted in users making Website Items for each variant (which is a bit useless since no full page views of variants are required)and increased db size

**Old Behaviour:**
-  Before the e-commerce refactor, **all variant attributes were pulled given that the variant was not disabled.**
- That means all variants will be able to be added to cart unless they are disabled

**Fix:**
- Revert to old behaviour. Showing select variants will be taken up later as a better designed feature
- Deleted unused functions that contributed to nothing and did unnecessary computation. All variant/attributes computation is done by `ItemVariantsCache`. So its not need in `get_context`

### **To Test:**
Item Variant popup in template item web page
**Case 1:**
- Make template e.g. tshirt (and publish) and variants large-red, large-green, medium-red, medium-green and small-red
- Disable variant small-red
- go to popup on template web page, size dropdown should only show **large** and **medium** , since item with small attribute is disabled and theres no other item with this attribute.

**Case 2:**
- Make template e.g. tshirt (and publish) and variants large-red, large-green, medium-red, medium-green and small-red
- select size "small". 
- Now open the color dropdown, only **red** should be a valid attribute value since only red is available in small size (small-red item)

ToDo:
- [x] Tests

### DIFF
--- a/erpnext/e_commerce/doctype/website_item/website_item.py
+++ b/erpnext/e_commerce/doctype/website_item/website_item.py
@@ -203,16 +203,15 @@ class WebsiteItem(WebsiteGenerator):
 		context.body_class = "product-page"
 
 		context.parents = get_parent_item_groups(self.item_group, from_item=True) # breadcumbs
-		self.attributes = frappe.get_all("Item Variant Attribute",
+		self.attributes = frappe.get_all(
+			"Item Variant Attribute",
 			fields=["attribute", "attribute_value"],
-			filters={"parent": self.item_code})
+			filters={"parent": self.item_code}
+		)
 
 		if self.slideshow:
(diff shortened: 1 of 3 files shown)


marination:
Unrelated test_job_card_scrap_item failing

PR #29222, opened 2022-01-10
fix(pos): cannot ignore pricing rule for one particular invoice

nextchamp-saqib:
1. Create a POS Invoice using Form View
2. Click on Ignore Pricing Rule
3. Add an Item that has some Pricing Rule
4. Save the document

You should see pricing rule has been applied and Ignore Pricing Rule is unchecked which shouldn't be the case. 

Expected Behaviour:
Ignore Pricing Rule check shouldn't be overwritten from POS Profile. It should be only fetched initially from POS Profile


nextchamp-saqib:

PR #29225, opened 2022-01-10
fix: partial transfers against work order

### LINKED ISSUE #28377: Partial Raw Material Transfer is not working

ankush:
closes https://github.com/frappe/erpnext/issues/28377 

Problem:
- Currently multiple transfer entries of increments total transferred qty by "for qty" of stock entry. 
- This is all good till "for qty" and qty inside stock entry tables match up but if you partially transfer qty then this computation is nonsensical. (refer to original issue above for example)
- Because of this incorrect computation additional qty can not be transferred. 

Change:

Summary: Transferred qty == "Minimum finished goods quantity that can be produced from transferred material" (yeah that's a mouthful, refer example below to understand better)

Example:

Qty to manufacture: 2

| Item | Required (A) | Transferred (B) | Can be used for how many FG? (C) |
| ---  | ---          | ---             | ---                              |
| A    | 2            | 1               | 1                                |
| B    | 4            | 1               | 0                                |

transferred qty = min(C) = 0

| Item | Required (A) | Transferred (B) | Can be used for how many FG? (C) |
| ---  | ---          | ---             | ---                              |
| A    | 2            | 2               | 2                                |
| B    | 4            | 2               | 1                                |

transferred qty = min(C) = 1

| Item | Required (A) | Transferred (B) | Can be used for how many FG? (C) |
| ---  | ---          | ---             | ---                              |
| A    | 2            | 2               | 2                                |
| B    | 4            | 4               | 2                                |

transferred qty = min(C) = 2

TODO:
- [x] unit test(s) for manufacturable qty logic
- [x] Validation for row-level transferred qty against allowance (should we?)
- [ ] Fix broken test / affected consumption logic.

PS: use split diff or browse commit by commit. Unified diff is unreadable for this change.


ankush:
This is becoming a bit too complicated with "for qty" logic used for consumption, so it's best fixed after a thorough refactoring of consumption logic. 


For now workaround for this issue is :
1. DO additional partial transfers with 0 as "for qty"
2. If some order is stuck, transfer additional material directly instead of against a work order.

PR #29228, opened 2022-01-10
fix(Payroll): Cannot submit salary slips from amended payroll entry.

dj12djdjs:
**Problem**
When cancelling a payroll entry, the amended payroll entry retains the transaction state of the previously cancelled payroll entry for the linked salary slips.
Example:
payroll-entry-ammend

**After PR**
payroll-entry-ammend-fix

### DIFF
--- a/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
+++ b/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
@@ -60,8 +60,8 @@ class PayrollEntry(Document):
 	def on_cancel(self):
 		frappe.delete_doc("Salary Slip", frappe.db.sql_list("""select name from `tabSalary Slip`
 			where payroll_entry=%s """, (self.name)))
-		self.salary_slips_created = False
-		self.salary_slips_submitted = False
+		self.db_set("salary_slips_created", 0)
+		self.db_set("salary_slips_submitted", 0)
 
 	def get_emp_list(self):
 		"""
--- a/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
+++ b/erpnext/payroll/doctype/payroll_entry/payroll_entry.py
@@ -60,6 +60,8 @@ class PayrollEntry(Document):
 	def on_cancel(self):
 		frappe.delete_doc("Salary Slip", frappe.db.sql_list("""select name from `tabSalary Slip`
 			where payroll_entry=%s """, (self.name)))
+		self.salary_slips_created = False
+		self.salary_slips_submitted = False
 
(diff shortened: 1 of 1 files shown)


dj12djdjs:
Manually checked, the linter error is committed to database.

erpnext/payroll/doctype/payroll_entry/payroll_entry.py
```diff
@@ -60,6 +60,8 @@ def validate_employee_details(self):
 	def on_cancel(self):
 		frappe.delete_doc("Salary Slip", frappe.db.sql_list("""select name from `tabSalary Slip`
 			where payroll_entry=%s """, (self.name)))
+		self.salary_slips_created = False
+		self.salary_slips_submitted = False
```
ruchamahabal:
```suggestion
		self.db_set("salary_slips_created", 0)
		self.db_set("salary_slips_submitted", 0)
```

https://github.com/frappe/erpnext/blob/8f0d10a1b791fb6c9471fe8aec785e5bd6ac5321/erpnext/payroll/doctype/payroll_entry/payroll_entry.py#L699

Use `db_set` while doing operations in `on_cancel` or `on_submit` since it's not explicitly committed to db.

ruchamahabal:
This is fairly understandable from the title of the PR but can you update the description with a screenshot and some description too?

ankush:
> Manually checked, the linter error is committed to database.


Refresh the page 😄

erpnext/payroll/doctype/payroll_entry/payroll_entry.py
dj12djdjs: Good to know, thanks!

dj12djdjs:
> This is fairly understandable from the title of the PR but can you update the description with a screenshot and some description too?

Will do

ruchamahabal:
We should also enable `no_copy` for these fields as this issue will occur in case someone duplicates the payroll entry too.

dj12djdjs:
@ruchamahabal I can't figure out why Payroll Entry doesn't have a duplicate option. I think this has been disabled somewhere.
image

Should I still add to no_copy flags?

ruchamahabal:
Payroll Entry has **Hide Copy** enabled. We don't need `no_copy` then.

image

ruchamahabal (approved):

ruchamahabal:
@mergifyio backport develop

ruchamahabal:
@dj12djdjs Thanks for the fix. Next time you raise a PR, it's better to raise it on develop then we can backport it to older versions.

PR #29233, opened 2022-01-11
fix: Added validation for serial nos selected at POS invoice level

nemesis189:
**Issue:**
After auto fetching the serial nos in the POS screen if the serial is modified incorrectly no error will be thrown and the invoice gets submitted. It is thrown only during consolidation.

**Fix:**
Added validation for checking if the selected serial nos are valid or not.

<details>
<summary><b>Steps:</b> </summary>
<ol>
<li>
 Create a serialized item and create a Serial No for that item.
   

   
</li>
<li> In POS Screen add the serialized item, after setting the quantity auto fetch the serial nos</li>
<li> Edit one of the serial nos</li>
   </li>
<li> After checkout it will throw a validation now:
    
 
   **Note: Before the fix it would simply go ahead and submit the invoice and show the bill**
</li>
</details>

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -214,6 +214,7 @@ class POSInvoice(SalesInvoice):
 			frappe.throw(error_msg, title=_("Invalid Item"), as_list=True)
 
 	def validate_serial_nos(self):
+		from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
 		error_msg = []
 		for item in self.get("items"):
 			serialized = item.get("has_serial_no")
@@ -221,7 +222,7 @@ class POSInvoice(SalesInvoice):
 			if serialized:
 				valid_serial_nos = frappe.get_all('Serial No', filters={'item_code':item.get("item_code")}, pluck="name")
 				invalid_serials = ""
-				for serial_no in item.get('serial_no').split("\n"):
+				for serial_no in get_serial_nos(item.get('serial_no')):
 					if serial_no not in valid_serial_nos:
 						invalid_serials = ", " if invalid_serials else "" + invalid_serials + serial_no
 						msg = (_("Row #{}: Following Serial numbers for item {} are <b>invalid</b>: {}").format(item.idx, frappe.bold(item.get("item_code")), frappe.bold(invalid_serials)))
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -40,6 +40,7 @@ class POSInvoice(SalesInvoice):
 		self.validate_item_cost_centers()
 		self.validate_warehouse()
 		self.validate_serialised_or_batched_item()
+		self.validate_serial_nos()
 		self.validate_stock_availablility()
 		self.validate_return_items_qty()
 		self.validate_non_stock_items()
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -212,6 +213,26 @@ def validate_serialised_or_batched_item(self):
 		if error_msg:
 			frappe.throw(error_msg, title=_("Invalid Item"), as_list=True)
 
+	def validate_serial_nos(self):
+		error_msg = []
+		for item in self.get("items"):
+			serialized = item.get("has_serial_no")
+			valid_serial_nos = frappe.get_all('Serial No', filters={'item_code':item.get("item_code")}, pluck="name")
```
nextchamp-saqib: Move this below `if` statement

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -212,6 +213,25 @@ def validate_serialised_or_batched_item(self):
...
 
+	def validate_serial_nos(self):
+		error_msg = []
+		for item in self.get("items"):
+			serialized = item.get("has_serial_no")
+			msg = ""
+			if serialized:
+				valid_serial_nos = frappe.get_all('Serial No', filters={'item_code':item.get("item_code")}, pluck="name")
+				invalid_serials = ""
+				for serial_no in item.get('serial_no').split("\n"):
```
ankush: Import this function instead of manually splitting it. 

https://github.com/frappe/erpnext/blob/d7b0a47d7cbfcc03322ade1a31d3520caddb6483/erpnext/stock/doctype/serial_no/serial_no.py#L472-L477

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -212,6 +213,25 @@ def validate_serialised_or_batched_item(self):
...
+					if serial_no not in valid_serial_nos:
```
ankush, line 232: You can do basic set difference here instead. 

`invalid_serial_nos = serial_no_in_doc - available_serial_no`

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -212,6 +213,25 @@ def validate_serialised_or_batched_item(self):
 		if error_msg:
 			frappe.throw(error_msg, title=_("Invalid Item"), as_list=True)
 
+	def validate_serial_nos(self):
+		error_msg = []
+		for item in self.get("items"):
+			serialized = item.get("has_serial_no")
+			msg = ""
+			if serialized:
+				valid_serial_nos = frappe.get_all('Serial No', filters={'item_code':item.get("item_code")}, pluck="name")
```
ankush: maybe only active ones should be fetched?

ankush: This still means same serial no can be sold twice if not reserved like for batches 😅

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
nemesis189: I felt the loop could be simpler and helps with better error message

ankush: You can get the same error message. Anyway, doesn't matter much.

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
nextchamp-saqib: Yes, we will have to check for reserved serial nos too. 
That can be queried from POS Invoice Item

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -212,6 +213,26 @@ def validate_serialised_or_batched_item(self):
...
 			frappe.throw(error_msg, title=_("Invalid Item"), as_list=True)
 
+	def validate_serial_nos(self):
+		from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
+		error_msg = []
+		for item in self.get("items"):
+			serialized = item.get("has_serial_no")
+			msg = ""
+			if serialized:
+				valid_serial_nos = frappe.get_all('Serial No', filters={'item_code':item.get("item_code"), "status":'Active'}, pluck="name")
```
nextchamp-saqib: This should be fetched using `auto_fetch_serial_number` since it takes into account the `pos_reserved_serial_nos`

nextchamp-saqib (changes requested):

PR #28798, opened 2021-12-08
fix: misleading "Set Default X" fields after saving

s-aga-r:
### Issue(s): Purchase Receipt

_Problem:_
- When populating the "Accepted Warehouse" field it adds that warehouse as the default warehouse on all the item rows. This is helpful because often the same warehouse for all the items received. But if the user changes the warehouse for an item. In that case, this field becomes confusing as it is holding another warehouse.

_Proposed Solution:_
- If all warehouses in the child table(Items) are not matching the "default warehouse" field then set its value to None.

<details>
<summary>Images/GIF</summary>

_Before:_

https://user-images.githubusercontent.com/63660334/144282361-d6e8ef46-ae48-42ed-a86c-8136cf993b45.gif

_After:_

https://user-images.githubusercontent.com/25903035/130674996-a6d891e4-4220-4574-87fc-5b32bca6992a.png
</details>

Closes: https://github.com/frappe/erpnext/issues/28644 

Affected doctypes:
- [x] Purchase Invoice
- [x] Sales Invoice
- [x] Sales Order 
- [x] Purchase Order
- [x] Stock Entry
- [x] Purchase Receipt
- [x] Delivery Note
- [x] Material request

PR #29257, opened 2022-01-12
fix: Avoid resetting Default wh fields for Manufacture Entry

marination:
Introducted via https://github.com/frappe/erpnext/pull/28798

**Issue:**
- Source/target is blank on some rows in a Manufacture Stock Entry
- Users still want an overall reference of the warehouses involved in this mfg entry
- Scrap item may have different warehouse (just one row), it should be overlooked

**Fix:**
- Avoid resetting Default source/target warehouse in manufacture entry alone
- For Repack entry users most like will not use these fields since each row has to have a different warehouse. And repack entries aren't auto generated from anywhere, like Manufacture entries are generated from Work Orders
- If there's 1 entry in `child_table_values`, dont do anything. **There's nothing to reset**. Reset and set to None only if there's more than one value for a field in the table rows.

### DIFF
--- a/erpnext/controllers/tests/test_transaction_base.py
+++ b/erpnext/controllers/tests/test_transaction_base.py
@@ -4,73 +4,72 @@ import frappe
 
 
 class TestUtils(unittest.TestCase):
-    def test_reset_default_field_value(self):
-        doc = frappe.get_doc({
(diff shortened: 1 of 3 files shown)


marination:

PR #29271, opened 2022-01-13
fix: Don't validate FG in repack entry

marination:
https://github.com/frappe/erpnext/pull/28555 missed the case of multiple FGs being possible in repack entries.

- An item could repacked into 3/4 smaller items
- In this case 3/4 rows will have `is_finished_item` auto checked
- This should be allowed in case of repack

**Fix:**
- The only check for repack should be if theres no row with `is_finished_item` checked, throw an error
- multiple fg should be allowed

**To test:**
- case 1: test repack entry with one item consumed and one item created (one has source wh and second row has target)
- case 2: test repack entry with one item consumed and 3 other items created (one item of say 90 qty is repacked into 3 items of 30 qty each)
- Check impact on cost
- Check if `is_finished_item` checked is checked on the right rows

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -738,9 +738,9 @@ class StockEntry(StockController):
 
 	def validate_finished_goods(self):
 		"""
-			1. Check if FG exists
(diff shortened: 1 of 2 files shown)


marination:

PR #29290, opened 2022-01-14
fix: billed amount in delivery note items

s-aga-r:
### Issue(s): Delivery Note

_Problem:_
- Billed amount becomes "0" if we create a Delivery Note from a Sales Order after a Sales Return.

_Proposed Solution:_
- Consider returned quantity at the time of submitting the delivery note and update the billed amount for every item.

_Steps to Reproduce:_
- Create Sales Order
- Create Sales Invoice 
- Create Delivery Note(from Sales Order)
- Create Sales Return(from Delivery Note)
- Again Create a Delivery Note(from Sales Order)

<details>
<summary>Images/GIF</summary>

_Before:_

https://user-images.githubusercontent.com/63660334/151803718-dcd0d9f6-8010-4bd0-bad6-9bd0edc28cb9.mov

_After:_

https://user-images.githubusercontent.com/63660334/151803606-ddd075d3-200d-45d6-a4b9-85d4f21033c0.mov
</details>


erpnext/patches/v13_0/set_billed_amount_in_returned_dn.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2022, Frappe and Contributors
+# License: GNU General Public License v3. See license.txt
+
+import frappe
```
deepeshgarg007: This patch will try to update Delivery Notes right from the very start. The books must be frozen for most of the users for the previous fiscal years so not a good idea to change those now and this patch will also take a lot of time for some users. Hence I'll suggest removing this patch and handle on a case to case basis if needed on the site

Also, please add an example with some actual values like quantities and amount so that the fix can be easily tested or replicated

deepeshgarg007 (approved):

deepeshgarg007:
@Mergifyio backport develop

PR #29292, opened 2022-01-14
fix: exclude existing serial numbers while auto creating new

rtdany10:
In cases where Serial No is auto generated, consider the serial number template is XYZ.### and that the counter is at 0.

Assume I create a purchase receipt with 1 qty and manually input the serial number as XYZ005.
And if I create another purchase receipt with 10 qty and no serial number given, the system will try to use serial numbers XYZ001 to XYZ010. In that case, I will get a validation error regarding XYZ005.

Further: https://discuss.erpnext.com/t/serial-number-generation-problem/84831

This PR aims to solve the issue by checking if the serial number already exists or not.
#no-docs

### DIFF
--- a/erpnext/stock/doctype/serial_no/serial_no.py
+++ b/erpnext/stock/doctype/serial_no/serial_no.py
@@ -402,9 +402,15 @@ def update_serial_nos(sle, item_det):
 def get_auto_serial_nos(serial_no_series, qty):
 	serial_nos = []
 	for i in range(cint(qty)):
-		serial_nos.append(make_autoname(serial_no_series, "Serial No"))
+		serial_nos.append(create_sno(serial_no_series))
 
 	return "\n".join(serial_nos)
+
+def create_sno(serial_no_series):
+	sno = make_autoname(serial_no_series, "Serial No")
+	if frappe.db.exists("Serial No", sno):
+		sno = create_sno(serial_no_series)
+	return sno
 
 def auto_make_serial_nos(args):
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/serial_no/serial_no.py
```diff
@@ -402,10 +402,16 @@ def update_serial_nos(sle, item_det):
...
-		serial_nos.append(make_autoname(serial_no_series, "Serial No"))
+		serial_nos.append(create_sno(serial_no_series))
 
 	return "\n".join(serial_nos)
 
+def create_sno(serial_no_series):
+	sno = make_autoname(serial_no_series, "Serial No")
+	if frappe.db.exists("Serial No", sno):
+		sno = create_sno(serial_no_series)
+	return sno
```
ankush: overall fix is okay; some suggestions on naming:

function should be probably be called "get_new_serial_number", along with argument "series" it makes it pretty clear as to what it does. 

Also `sno` is bad variable name; `sr_no` maybe?

---

up to you: Add a test case for this behaviour. (just encode the steps you described in a unit test 😄)

ankush (changes requested):

erpnext/stock/doctype/serial_no/serial_no.py
rtdany10: Have renamed. Will also add a test case. :)

rtdany10:

ankush (approved):

PR #29296, opened 2022-01-15
feat: LIFO valuation

ankush:
Similar to FIFO valuation except when consuming last received material is consumed first. 

https://www.investopedia.com/articles/02/060502.asp

- LIFO can be enabled globally or on an individual item. 
- Negative inventory handling is equivalent to current FIFO handling. 
- PS: LIFO is banned by some accounting bodies, make sure your regional bodies allow using it: https://www.investopedia.com/articles/investing/042115/why-last-first-out-banned-under-ifrs.asp

<details>
  <summary></summary>

my first feature in ERPNext

</details>

todo:
- [x] rebase
- [x] some more manual testing
- [x] docs https://docs.erpnext.com/docs/v13/user/manual/en/stock/articles/calculation-of-valuation-rate-in-fifo-and-moving-average

To test:
- automated tests added. To check behavior create new item with LIFO valuation and run sales/purchase cycles (A lot of different ways it can be tested)
- check that FIFO/Moving average still work as expected.

### DIFF
--- a/erpnext/stock/doctype/item/item.json
+++ b/erpnext/stock/doctype/item/item.json
@@ -346,7 +346,7 @@
    "fieldname": "valuation_method",
    "fieldtype": "Select",
    "label": "Valuation Method",
-   "options": "\nFIFO\nMoving Average"
+   "options": "\nFIFO\nMoving Average\nLIFO"
   },
   {
    "depends_on": "is_stock_item",
--- a/erpnext/stock/doctype/stock_settings/stock_settings.json
+++ b/erpnext/stock/doctype/stock_settings/stock_settings.json
@@ -92,7 +92,7 @@
    "fieldname": "valuation_method",
    "fieldtype": "Select",
    "label": "Default Valuation Method",
-   "options": "FIFO\nMoving Average"
+   "options": "FIFO\nMoving Average\nLIFO"
   },
   {
    "description": "The percentage you are allowed to receive or deliver more against the quantity ordered. For example, if you have ordered 100 units, and your Allowance is 10%, then you are allowed to receive 110 units.",
@@ -305,7 +305,7 @@
  "index_web_pages_for_search": 1,
  "issingle": 1,
  "links": [],
- "modified": "2021-11-06 19:40:02.183592",
+ "modified": "2022-01-15 17:42:53.174865",
  "modified_by": "Administrator",
  "module": "Stock",
  "name": "Stock Settings",
@@ -324,5 +324,6 @@
  "quick_entry": 1,
  "sort_field": "modified",
  "sort_order": "ASC",
+ "states": [],
  "track_changes": 1
 }
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -16,7 +16,7 @@ from erpnext.stock.utils import (
 	get_or_make_bin,
 	get_valuation_method,
 )
-from erpnext.stock.valuation import FIFOValuation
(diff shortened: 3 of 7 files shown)


erpnext/stock/utils.py
```diff
@@ -259,29 +260,25 @@ def get_valuation_method(item_code):
...
+	ValuationKlass = LIFOValuation if method == "LIFO" else FIFOValuation
+
+	stock_queue = ValuationKlass(previous_stock_queue)
 	if flt(qty) >= 0:
-		total = sum(f[0] for f in previous_stock_queue)
-		return sum(flt(f[0]) * flt(f[1]) for f in previous_stock_queue) / flt(total) if total else 0.0
+		total_qty, total_value = stock_queue.get_total_stock_and_value()
+		return total_value / total_qty if total_qty else 0.0
 	else:
-		available_qty_for_outgoing, outgoing_cost = 0, 0
```
ankush, line 268: A smallish "breaking" behavior here. This will keep on using the last rate for computing the average incoming rate in case of negative inventory. Incoming rate is anyway nonsensical for inventory that doesn't exist. 

The benefit is, now it's more in line with actual consumption logic.

erpnext/stock/doctype/item/item.json
```diff
@@ -4,7 +4,7 @@
  "allow_import": 1,
  "allow_rename": 1,
  "autoname": "field:item_code",
- "creation": "2013-05-03 10:45:46",
+ "creation": "2022-01-11 15:03:47.588455",
```
gavindsouza:
```suggestion
 "creation": "2013-05-03 10:45:46",
```
oof

--

this is fixed via https://github.com/frappe/frappe/pull/15763

erpnext/stock/doctype/stock_settings/stock_settings.json
```diff
@@ -1,6 +1,6 @@
 {
  "actions": [],
- "creation": "2013-06-24 16:37:54",
+ "creation": "2022-01-11 15:03:48.230814",
```
gavindsouza:
```suggestion
 "creation": "2013-06-24 16:37:54",
```

erpnext/stock/stock_ledger.py
```diff
@@ -701,14 +701,18 @@ def get_moving_average_values(self, sle):
 						sle.voucher_type, sle.voucher_no, self.allow_zero_rate,
 						currency=erpnext.get_company_currency(sle.company), company=sle.company)
 
-	def update_fifo_values(self, sle):
+	def update_fifo_lifo_values(self, sle):
```
gavindsouza:
```suggestion
	def update_queue_values(self, sle):
```

Felt the named method didn't make much sense. Also, how about adding type hints? I'm not sure what `sle` is supposed to be without reading the code above or below.

erpnext/stock/stock_ledger.py
```diff
@@ -461,7 +461,7 @@ def process_sle(self, sle):
 					self.wh_data.qty_after_transaction += flt(sle.actual_qty)
 					self.wh_data.stock_value = flt(self.wh_data.qty_after_transaction) * flt(self.wh_data.valuation_rate)
 				else:
-					self.update_fifo_values(sle)
+					self.update_fifo_lifo_values(sle)
```
```suggestion
					self.update_queue_values(sle)
```

erpnext/stock/stock_ledger.py
ankush: +1 on typing, right now sles are either _dict or actual sle document, often interchangeably 🥲

PR #29304, opened 2022-01-16
fix: dont update sle values from get_gl_entries

ankush:
Summary: `get_gl_entries` is adding average valuation rate when SLE has no stock value difference.

Steps to reproduce:
1. Make sure FIFO is the valuation method. Make ~3 stock receipts; 1 with 0 valuation. 
2. Consume material; when 0 valued material is consumed instead the average rate is used to compute value difference. 

Fix: Just assume what the stock ledger has done is right and don't try to modify it when making GL entries. 

This was added in https://github.com/frappe/erpnext/pull/7583; now there's no reason to assume that 0 rate is from a bug (and if there is, root cause should be fixed instead of applying average rate)


rohitwaghchaure (approved):

rohitwaghchaure:

PR #29317, opened 2022-01-17
feat: additional doctypes access for ESS user

ruchamahabal:
Additional doctypes access for ESS User

`no-docs`

Depends on: https://github.com/frappe/frappe/pull/15707
Not adding tests, it's just a default configuration change

To test:

Check if a user with **Employee Self Service** role is able to access these doctypes with relevant permissions:

https://github.com/frappe/erpnext/blob/6c1c85be0c7318692dc3544b4be644b45c31143a/erpnext/setup/install.py#L207-L233


ruchamahabal:

PR #29323, opened 2022-01-17
refactor: compute batch ledger in python

ankush:
Some users don't have latest mariadb version which causes failure in
this validation.


ankush:

ankush:
keeping this on v13 only for now.

PR #29359, opened 2022-01-19
fix: bypass selling price validation for free item

KrithiRamani:
User has enabled  "Validate Selling Price for Item Against Purchase Rate or Valuation Rate" checkbox in selling settings. When a free item is added and rate is set to 0, the validation gets triggered and forces the user to enter a rate. 

This fix wouldn't throw the validation message.


rohitwaghchaure:

PR #29193, opened 2022-01-07
fix: deferred report division by zero exception

ruthra-kumar:
Fixing issue where 'Deferred Revenue and Expense Report' throws exception if deferred item has no period (0 months).

Screenshot 2022-01-07 at 6 05 18 PM

PR #29360, opened 2022-01-19
test: unit test for deferred item with zero months

ruthra-kumar:
Refactored [_test_deferred_revenue_and_expense.py_](https://github.com/frappe/erpnext/blob/5ae44f3a02eea5c3d4d1d688deec7a932842937e/erpnext/accounts/report/deferred_revenue_and_expense/test_deferred_revenue_and_expense.py)

- Clearing old entries has a more generalised approach.
- New Test Case has been added for 'ZeroDivisionError' bug in report, fixed by https://github.com/frappe/erpnext/pull/29193

### DIFF
--- a/erpnext/accounts/report/deferred_revenue_and_expense/test_deferred_revenue_and_expense.py
+++ b/erpnext/accounts/report/deferred_revenue_and_expense/test_deferred_revenue_and_expense.py
@@ -232,6 +232,91 @@ class TestDeferredRevenueAndExpense(unittest.TestCase):
 		]
 		self.assertEqual(report.period_total, expected)
 
+	def test_zero_months(self):
+		self.clear_old_entries()
+		# created deferred expense accounts, if not found
+		deferred_revenue_account = create_account(
+			account_name="Deferred Revenue",
+			parent_account="Current Liabilities - _CD",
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/deferred_revenue_and_expense/test_deferred_revenue_and_expense.py
```diff
@@ -17,10 +17,42 @@
...
+		deferred_invoices = (
+			qb.from_(pinv)
+			.join(pinv_item)
+			.on(pinv.name == pinv_item.parent)
+			.select(pinv.name)
+			.where(pinv_item.enable_deferred_expense == 1)
+			.run()
+		)
+		if deferred_invoices:
+			qb.from_(pinv).delete().where(pinv.name.isin(deferred_invoices)).run()
```
ankush, line 51: can you confirm if this is actually working? 

AFAIK `isin` expects a shallow iterable like `[name1, name2]` whereas SQL results are typically nested one level like this: `[(name1,), (name2,)]`

Screenshot 2022-01-21 at 11 40 46 AM

ankush (changes requested):

deepeshgarg007:

Issue #29363, opened 2022-01-19
Sales Invoice with return checked and include payment issues


mro-admin:
### Information about bug

When a sales invoice is created with the include payment is checked and a return is created from it. The payment method changes to the default payment method according to the pos profile which is not the payment method from the invoice the return is created from.

This seems like it is being caused by line 620 in erpnext.controllers.taxes_and_totals_py where a call to self.update_paid_amount_for_return is made with this code

```
		default_mode_of_payment = frappe.db.get_value('POS Payment Method',
			{'parent': self.doc.pos_profile, 'default': 1}, ['mode_of_payment'], as_dict=1)

		if default_mode_of_payment:
			self.doc.payments = []
			self.doc.append('payments', {
				'mode_of_payment': default_mode_of_payment.mode_of_payment,
				'amount': total_amount_to_pay,
				'default': 1
			})

		self.calculate_paid_amount()
```

Seems like an unnecessary call. Will create a PR if im correct but i want input from someone else

### Module

accounts, other

### Version

Frappe Framework: v13.18.0 (version-13)
ERPNext: v13.18.0 (version-13)

deepeshgarg007:
https://github.com/frappe/erpnext/pull/23599

Already a PR was merged a while back to resolve this. In other cases its intentional

PR #29364, opened 2022-01-20
fix: Only add default payment if there are no payments configured

mro-admin:
Issue:[Sales Invoice with return checked and include payment issues]( https://github.com/frappe/erpnext/issues/29363)


mro-admin:
checked the failng test. failed test 1 and 3 doesn't seem to be related to the change

but failing test 2 is because the test is assuming that when an invoice has 2 payments and a return is created the return value must be summed and the default pos payment must be used at all times. This is not always the case. We might want to return from different accounts and I would argue that most of the time returns will be returned from the account they are paid from.

Would like some opinion on this because this issue is causing us not to be able to return pos invoices by just using the "Create">"return" button most of the time

PR #29366, opened 2022-01-20
fix: Cleanup empty rows on bank statement import

deepeshgarg007:
Bank Statement Import fails if there are empty rows in the template as follows and gives the below error

```
  File "apps/frappe/frappe/utils/background_jobs.py", line 104, in execute_job
    method(**kwargs)
  File "apps/erpnext/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py", line 111, in start_import
    add_bank_account(data, bank_account)
  File "apps/erpnext/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py", line 147, in add_bank_account
    row[bank_account_loc] = bank_account
IndexError: list assignment index out of range
```

image

Added a function to clean up empty rows before importing

### DIFF
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py
@@ -17,6 +17,7 @@ from openpyxl.styles import Font
 from openpyxl.utils import get_column_letter
 from six import string_types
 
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@Mergifyio backport develop version-13-pre-release

PR #29374, opened 2022-01-20
fix: ignore pricing rule in transactions

nextchamp-saqib:
Fixed an issue that can be replicated as:
- Add an Item to the Sales Invoice that has some Pricing Rule defined. For eg. 20% off on Apple Earbuds
- Save the Sales Invoice - Make sure Pricing Rule should have been applied
- Check Ignore Pricing Rule and save the Sales Invoice
- The rate of the item is still the discounted rate

Changes made:
- Set `rate` as `price_list_rate` if pricing rule is removed
- `set_missing_item_details()` will set pricing rule fields if pricing rule is removed


nextchamp-saqib:
Merging as unrelated tests are failing

PR #29375, opened 2022-01-20
fix: (enhance) BOM Operations Report

marination:
**Issue:**
- Report **BOM Operations Time** had no filters and a dump of all the BOM data. This made it hard to filter data BOM wise
- Also the Time in Mins field was inaccurate showing Integer data (like 10.56 showed as 10)
- Due to adding rows as they came and not maintaining data BOM-wise while computing the report, random operations belonging to BOM A, where shown under BOM B. (**Row 11 [BOM-BDSH-001-002] shows 4 operations, while it only has 2 in the BOM**)
Screenshot 2022-01-20 at 1 43 26 PM

**Fix:**
- Added filters in the Report for BOM ID, Item Code and Workstation
- BOM ID is a Multiselect filter, so that users can view certain BOMs of an Item
- Converted Raw SQL to frappe.qb and added method to get filtered data
- Changed fieldtype of 'Time in mins' from Int to Float
- Get BOM wise grouped data to  keep order and accurate grouping in report
Screenshot 2022-01-20 at 1 42 52 PM

**To Test:**
- All Filters
- Include a simple BOM with operations
- A BOM with operations having a sub-assembly with operations. Check if both sub-assembly and BOM show up on report. Subassembly BOM Count should be 1 in sub-assemble BOM row
- Make multiple BOMs with operations (different or same items doesn't matter). Make sure operations are shown under correct BOM.


marination:
Report does not do any extra computation, just formats data that's it. The basic test in `manufacturing/reporyt/test_reports.py` should be enough

marination:

PR #28570, opened 2021-11-26
fix: purchase to Stock UOM conversion on Production Plan

MaxMorais:
The new Production Planning Tool, didnt take into consideration the UOM conversion for items that have a different purchase UOM.

With that, it will generate wrong material requests to purchase!

As a simple case

- Item XPTO (10 units available in stock)
   - Stock UOM: Unit <-- must be a whole number
   - Purchase UOM: Box of 100 <-- must be a whole number

BOM ABC:
     - XPTO - 4 units required

Production Plan for ABC
     - Produce 50 units

Actual Result
     - Stock Transfer of 10 units
     - Purchase Transfer of 40 Box of 100 (leading to 4000 units in Stock)

Expected Result
     - Stock Transfer of 10 units
     - Purchase Transfer of 1 Box of 100 (leading to 100 units in Stock)

PR #29376, opened 2022-01-20
fix: show stock UOM for material transfers

ankush:
To reproduce: follow https://github.com/frappe/erpnext/pull/28570

In the production plan, the material request table shown was showing purchase UOM for all rows; it was handled correctly internally just UOM displayed was incorrect. 

Before:
image

After: 
Screenshot 2022-01-20 at 2 23 45 PM


ankush:
Tests passing; codecov is down. Ignoring.

PR #29380, opened 2022-01-20
fix: "sr no created" misleading message

ankush:
To reproduce:

1. create 2 serialized items, 1 with series and 2nd with no series (i.e. you manually need to input sr no)
2. create stock receipt of both items, 1st with auto generated serial nos and 2nd manual.
3. Save and submit. 

Screenshot 2022-01-20 at 4 19 44 PM

Root cause: msgprints just go brrrr but the entire transaction has actually been rolled back so nothing got committed to database i.e. no new serial nos. So it's misleading. 

Solution: on rollback clear all messages except last one. It's ugly; MAYBE framework should provide a default solution for this. 

PS: doesn't work :lol: rollback observers are called after processing message. 
 https://github.com/frappe/frappe/blob/1e70c02e43b3a7212e4c35598721dcf1d183c249/frappe/app.py#L86-L94


ankush:
This was a bad hack; best handled at framework level.

PR #29382, opened 2022-01-20
fix(patch): delete hospitality custom fields

ankush:
Missed out the deletion of custom fields as pointed out here: https://github.com/frappe/erpnext/pull/29167#discussion_r788444630

to test: 
- just see if the sales invoice page opens or not.
- check if restaurant custom field got deleted from custom fields list.


ankush:
IMO this should just be part of `delete_doc(doctype)` for a simple reason: not deleting linked custom fields makes the whole form unusable (it just logs you out totally without any hints! LMAO)


https://github.com/frappe/frappe/blob/1e70c02e43b3a7212e4c35598721dcf1d183c249/frappe/model/delete_doc.py#L68

There could be one more delete here

```python
frappe.db.delete("Custom Field", {"options": name})
```

@gavindsouza

ruchamahabal (approved):

PR #29383, opened 2022-01-20
fix: Cart & Popup Logic of Item variant without Website Item

marination:
Fix over https://github.com/frappe/erpnext/pull/29219

## **Issue:**
- Lets assume Item **Tshirt** has many variants which don't need individual Web Pages. We just want the variants to be selected based on Color/Size and added to cart
- [This fix](https://github.com/frappe/erpnext/pull/29219) made sure that we don't need a Website Item of variants to select them from the template's web page
- While getting the exact matched item, it used to try fetch all product details, in which it to fetch Website Item details (in this case variants don't have website items):
image
- Another issue was display in shopping cart would also break because it relied on Website Item.

## **Fix:**
**To consider:** There's only two actions with variants not having a Website Item - add to cart & fetch in popup
- While getting the exact match in template web item popup, fetch only price details, not entire product details since it's not even used. Only price is used for display
- For Variant items without Website Items, get web data like image from the template. But get name from Item master itself
- Below you will see the name is from the Variant Item master, and the image is from the Template Item
   Screenshot 2022-01-20 at 7 46 05 PM

**todo:**
- [x] Provide right selling price list to `get_price`
- [x] Test full flow from popup to cart

**To Test:**
- Against Template Item **Tshirt**, create a variant **Tshirt-Small-Red** (keep a human readable item name)
- Publish the Template Item only
- Create Item Price for variant **Tshirt-Small-Red**
- Go to template item web page > Click on **Select Variant** > select attributes for variant **Tshirt-Small-Red**
- This item should get an exact match and its price should be visible in popup too
- Add variant to cart
- Go to cart and check it's name, image and price (if no image in variant item master, it should fetch from template website item). Name should be same as Variant Item master

Try the same flow with another template item whose variant does have a published website item

### DIFF
--- a/erpnext/e_commerce/shopping_cart/cart.py
+++ b/erpnext/e_commerce/shopping_cart/cart.py
@@ -42,7 +42,7 @@ def get_cart_quotation(doc=None):
 
 	if not doc.customer_address and addresses:
 		update_cart_address("billing", addresses[0].name)
-	print("doc>>", doc, type(doc))
+
 	return {
 		"doc": decorate_quotation_doc(doc),
 		"shipping_addresses": get_shipping_addresses(party),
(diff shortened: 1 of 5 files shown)


marination:

PR #29389, opened 2022-01-21
fix: update status on visit cancel

### LINKED ISSUE #29372: Completion statuses not updated appropriately in Maintenance Schedules upon cancellation of maintenance visits
### Information about bug

Issue : Completion statuses are not updated appropriately in Maintenance Schedules upon cancellation of maintenance visits 
 
Edge Case for https://github.com/frappe/erpnext/pull/28843

Steps to reproduce : 

1. Create a maintenance schedule with multiple items, 1 non serialised, 2 serialised. 
2. Create visit for both serialised items together, save and submit. Check if the modal for selection now excludes this item 
3. Create visit for non serialised item, modal for selection of item should show only the item for which visit hasn't been created yet. Save and submit the visit created. 
4. Check for schedules generated in Schedules Table 
      a. For serialised items it should show proper mapping for serial numbers
      b. Completion Status should be updated appropriately based on the status of maintenance visit
      c. The counter for maintenance visit in connections section should be updated properly as per number of visits created
5.Now cancel the created visits, reload and check the Completion Status under Schedules table, it isn't updated. 

Screen Recording : 

Edge Case

Work around - Manually update the status in Schedules table 

### Module

support

### Version

V 13.19.0

### Installation method

_No response_

_No response_

noahjacob:
### Issue
- On cancelling the **Visit**, the status would not update in the linked **Schedule**.

cancel_bug

### Fix
cancel_fix

### DIFF
--- a/erpnext/maintenance/doctype/maintenance_schedule/test_maintenance_schedule.py
+++ b/erpnext/maintenance/doctype/maintenance_schedule/test_maintenance_schedule.py
@@ -4,6 +4,7 @@
 import unittest
 
 import frappe
+from frappe.utils import format_date
 from frappe.utils.data import add_days, formatdate, today
 
 from erpnext.maintenance.doctype.maintenance_schedule.maintenance_schedule import (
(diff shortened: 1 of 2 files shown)


noahjacob:
#29372

PR #29398, opened 2022-01-21
fix: Wrong variable name breaks Cart Items rendering

marination:
**issue:**
After adding a variant to cart from the popup , having no image (nothing in template web item too), cart breaks:
```py
jinja2.exceptions.UndefinedError: 'erpnext.selling.doctype.quotation_item.quotation_item.QuotationItem object' has no attribute 'web_item_name'
```

**fix:**
- Correct variable name
- Test cart items rendering

### DIFF
--- a/erpnext/e_commerce/shopping_cart/test_shopping_cart.py
+++ b/erpnext/e_commerce/shopping_cart/test_shopping_cart.py
@@ -168,6 +168,9 @@ class TestShoppingCart(unittest.TestCase):
 
 		self.assertEqual(doc.get("items")[0].item_name, "Test-Tshirt-Temp-S-R")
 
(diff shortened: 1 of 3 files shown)


erpnext/e_commerce/shopping_cart/test_shopping_cart.py
```diff
@@ -168,6 +168,9 @@ def test_add_item_variant_without_web_item_to_cart(self):
 
 		self.assertEqual(doc.get("items")[0].item_name, "Test-Tshirt-Temp-S-R")
 
+		# test if items are rendered without error
+		frappe.render_template("templates/includes/cart/cart_items.html", cart)
```
ankush, line 172: brilliant 😆

PR #29400, opened 2022-01-21
fix: Wrong variable name breaks Cart Items rendering

marination:
Port of https://github.com/frappe/erpnext/pull/29398
**issue:**
After adding a variant to cart from the popup , having no image (nothing in template web item too), cart breaks:
```py
jinja2.exceptions.UndefinedError: 'erpnext.selling.doctype.quotation_item.quotation_item.QuotationItem object' has no attribute 'web_item_name'
```

**fix:**
- Correct variable name
- Test cart items rendering


marination:
I try debugging and everything passes... Cant replicate the failing test either. Plus time crunch for release :(
Screenshot 2022-01-21 at 6 07 33 PM

PR #29404, opened 2022-01-21
fix: Previous document currency getting overridden

deepeshgarg007:
**Issue**: Document Currency from Purchase Receipt gets overridden in Purchase Invoice

### Steps to replicate

1. Make sure the supplier you are using doesn't have a default currency set
2. Make a purchase receipt in currency other than Company currency
3. Make a Purchase Invoice from the receipt using the "Create Invoice" button, make sure the currency in Purchase Receipt is mapped in Purchase Invoice


deepeshgarg007:
@Mergifyio backport develop

PR #29405, opened 2022-01-21
fix(healthcare): populate practitioner selected in form to check availability popup

akashkrishna619:
- Practitioner field was getting cleared intermittently while loading the check availability popup because of department field on change
Before
fix: handle department on_change to conditionally clear practitioner

- Display practitioner name in the check availability dialog
PR

- Lab Test, Vital Signs and Sample Collection had two way fetch from for inpatient record and patient, removed fetch_from for patient link field
https://github.com/frappe/healthcare/pull/51#discussion_r782778183
fix: remove fetch_from for patient link field


erpnext/healthcare/doctype/patient_appointment/patient_appointment.js
```diff
@@ -332,14 +332,17 @@ let check_and_set_availability = function(frm) {
 
 		d.set_values({
 			'department': frm.doc.department,
-			'practitioner': frm.doc.practitioner,
```
ChillarAnand: This is needed for initial form loading?

akashkrishna619: Thanks, added it back

ChillarAnand (approved):

PR #29419, opened 2022-01-23
fix: incorrect raw materials quantity in manufacture stock entry

rohitwaghchaure:
**Issue**

1. Created BOM with one raw material and set the qty as 1
2. Set backflushed based on Material Transfer for Manufacture
2. Created work order for 100 qty
3. Created Materials Transfer and changed the qty from 100 to 110 for the raw material
4. Created Manufacture entry with qty as 90 and system as set the raw material's qty as 100 which is correct.
5. Created one more manufacture entry with qty as 10, but system has set the raw material's qty as 11 which is incorrect
6. The correct qty should be 10 and not 11.


rohitwaghchaure:

PR #29425, opened 2022-01-24
fix: honour 'include holidays' setting while marking attendance for leave application

ruchamahabal:
## Steps to replicate:

- Create a Leave Type with **Include holidays within leaves as leaves** disabled

    holidays

- Create a Leave Application that has holidays in between.

     leaves
 
- Since the above setting is disabled, the holidays in between should not be considered as leaves but holidays. 
- Even though the **Total Leave Days** are calculated correctly (after excluding holidays), attendance is marked for 6 days (holidays included) with the status "On Leave"

## Fix:

Honour **Include holidays within leaves as leaves** while marking attendance too, and not just while calculating Total Leave Days.

leave-das

P.S: These had no side effects like affecting leave balance because total leave days are/were calculated fine. Just attendance marking was corrected.


ruchamahabal:

PR #29426, opened 2022-01-24
fix: Use get for conditionally available fields while setting missing values

marination:
**To Replicate Issue:**
- Add custom field `supplier` in Material Request
- Create a new MR > Save > watch the world burn

**Issue:**
- Due to custom field "supplier" and missing field "supplier_address", dot operator breaks in `buying_controller > set_missing_values > get_party_details`
   ```py
    File "apps/erpnext/erpnext/controllers/buying_controller.py", line 74, in set_missing_values
      doctype=self.doctype, company=self.company, party_address=self.supplier_address, 
      shipping_address=self.get('shipping_address'),
    AttributeError: 'MaterialRequest' object has no attribute 'supplier_address'
    ```
- This scenario is brought to you by **customisation**

**Fix:**
- Make sure to use `get` instead of just dot operator if field is in some doctypes and not all

**Ambiguity:**
- Adding such custom fields with core field names, triggers JS functions also that technically aren't supposed to be triggered. The user won't have a clue. Not sure how to go about that

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -70,9 +70,18 @@ class BuyingController(StockController, Subcontracting):
 
 		# set contact and address details for supplier, if they are not mentioned
 		if getattr(self, "supplier", None):
-			self.update_if_missing(get_party_details(self.supplier, party_type="Supplier", ignore_permissions=self.flags.ignore_permissions,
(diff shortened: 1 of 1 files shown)


marination:
Skipping tests here, its a defensive fix

marination:
Gave it some thought. One way to go about controlling what is triggered where is to have a clear map of fields and documents.
But that is ultra tedious. We use `where field is present` in a lot of places.
Not worth the effort, ending it at this fix

marination:

PR #29428, opened 2022-01-24
fix: Add shipping charges to taxes only if applicable

deepeshgarg007:
Recently server-side implementation for shipping rule was added. For old invoices generated before this change following error was seen when making a subsequent document (Sales Invoice, Delivery Note, etc) from it.

```
 File "apps/erpnext/erpnext/controllers/taxes_and_totals.py", line 368, in get_current_tax_amount
  self.set_item_wise_tax(item, tax, tax_rate, current_tax_amount)
 File "apps/erpnext/erpnext/controllers/taxes_and_totals.py", line 376, in set_item_wise_tax
  if tax.item_wise_tax_detail.get(key):
AttributeError: 'NoneType' object has no attribute 'get'
```


deepeshgarg007:
@Mergifyio backport develop

PR #29433, opened 2022-01-24
fix: extend serial_no string clean up to packed items and WO

ankush:
Extend https://github.com/frappe/erpnext/pull/26596 to:
1. Packed items table serial no field (in DN/SI)
2. Work Order serial no field (for use in reservation against SO)

To test:
- Leading/trailing space in these field should get stripped
- nothing much else.


erpnext/stock/doctype/serial_no/serial_no.py
```diff
@@ -465,6 +465,13 @@ def get_serial_nos(serial_no):
 	return [s.strip() for s in cstr(serial_no).strip().upper().replace(',', '\n').split('\n')
 		if s.strip()]
 
+def clean_serial_no_string(serial_no: str) -> str:
+	if not serial_no:
+		return ""
+
+	serial_no_list = get_serial_nos(serial_no)
```
ankush, line 472: reusing `get_serial_nos` since it does what previous `clean_serial_nos` was doing (+ little extra by uppercasing) and `get_serial_nos` is used in used in other business logic too so using it here also makes sense IMO.

ankush:
All test failures are related to framework changes and fixed now; won't rerun CI (to avoid waste)

ref:

https://github.com/frappe/frappe/pull/15739
https://github.com/frappe/frappe/pull/15729

PR #29438, opened 2022-01-24
refactor!: remove amazon mws integration

s-aga-r:
Move Amazon MWS Integration from core to ecommerce_integrations


ankush:
@s-aga-r we will need to add a patch that removes doctypes from db. Example: https://github.com/frappe/erpnext/pull/27313

general checklist here: https://frappe.io/handbook/splitting-modules 



Also remove this patch: https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v12_0/rename_mws_settings_fields.py

ankush (approved):

ankush:
@s-aga-r add warning patch on v13.

PR #29449, opened 2022-01-25
feat: updated so/po_required dialogs, added Sales/Purchase Order create buttons in buying/selling workflows

### LINKED ISSUE #29608: UX: Inducing ideal practice on the sales/purchase side for linkages among the documents.
Consider that a sales invoice is created against a sales order. Now in the event that the user splits any of the items in the sales invoice item table, the newer row of the item does not get linked with the sales order.

Now consider another use case in which the user directly creates a sales invoice without creating the sales order. Even though there may exist some sales orders which are corresponding to the same customer and items, the sales invoice does not get linked with the sales orders.

In both the cases above, the sales orders are just left hanging and the status of the sales orders do not get updated.

**Possible Solution:**

When a user directly create sales invoice/delivery, inform him something on these lines:
“This invoice is not linked to any Sales Order. For better tracking of deliveries and payments, we recommend making an invoice from a Sales Order or linking one. Click here to create a new Sales Order”
Also if he chooses not to proceed with the sales order, then inform the user of the related transaction settings.
image

noahjacob:
### Changes
- When trying to create a non-linked **Delivery Note** or **Sales Invoice**, the following dialog appears.

image

- When a Customer is selected, if there are open sales orders against that customer, the following description is updated.

image

- If there are no open sales orders against that customer, a dialog would appear prompting you to create one.

image

si_dialog

- A **Sales Order** can also be created via the **Create** button if the **Setting** is set to **"No"**.

image

- If the user clicks on **Sales Order**,  i.e he wants to create a Sales Order, it would redirect to the **Sales Order** form with the current document mapped onto the  **Sales Order**.
- On submitting the **Sales Order**, a link is created between the **Sales Order** and the **Delivery Note/Sales Invoice.**

so_link

Similarly this process is also working for Buying Module.(Purchase Order)

### Tasks

- [x] Selling Module(Sales Order)
- [x] Buying Module(Purchase Order)


TurkerTunali:
There should be some settings to disable this feature. It will not be handy at all situations.

PR #29450, opened 2022-01-25
feat: provision to exclude exploded items in the BOM

### LINKED ISSUE #17738: [Feature Request] Allow different Levels in BOM Explosion
**Requirement:**
Possibility to use different, defined levels in BOM explosion. 

**Simplified example BOM:**

image

**Scenario:**
In this example the BOM of the product contains the sub-assemblies _Item 1_ and _Item 2_.
To manufacture the product _Item 1_ (1st level) shall be consumed. Item 2 shall be transparent so that _Item 2.1_, _Item 2.2_ and _Item 2.3_ (2nd level) are consumed directly. 

**Problem:**
Currently the system only allows to use the first level or the last level (called multi-level). But the last level is not always what you need e.g. if you have multiple stages of subcontracting. The production planning should allow a higher granuality in defining the level resolution of boms. 

Right now the production planning only offers two ways of resolution:

1. Use 1st level: _Item 1_ and _Item 2_
2. Use multi-level: _Item 1.1.1_, _Item 1.1.2_, _Item 1.2_, _Item 1.3_, _Item 2.1_, _Item 2.2_, _Item 2.3_

**Possible Solution:**
Define a check-box in the BOM that indicates whether the BOM can be exploded or not. If the BOM does not allow further explosion the item should be treated as a leaf. The resolution should stop in that case. 

In the scenario mentioned above, _Item 1_ would not allow further explosion.
The explosion stops at _Item 1_ while _Item 2_ is still transparent.

image

rohitwaghchaure:
**Issue**

In the BOM, when we select the sub-assembly item then system auto set the BOM in the BOM No field. Even if user don't want bom and remove the BOM from the BOM No field. On save it again set the BOM to the bom no field. So currently no option to exclude the exploded items for the particular bom

**Solution**

To fix the issue, we have added "Do Not Explode" checkbox in the BOM Item table. If the "Do Not Explode" checkbox is on then system will not show the BOM No field and won't added the exploded items.

image

**Use Case**

Lets say a company manufacture the Laptop which required two sub-assembly items as Motherboard and Keyboard. The company do the manufacturing once they received the order from the customer. The Manufacturing Time required for the Motherboard is more than the Keyboard, therefore the company do the manufacturing of Motherboard individually irrespective of the sales orders and kept in the stock. As the item Motherboard already in the stock it help to reduce the overall Manufacturing Time of the item Laptop. 

Now While preparing the BOM for the Laptop in the ERPNext, they don't want to explode the BOM of the item Motherboard but they want to explode the BOM of the item Keyboard. Therefore we have added checkbox "Do Not Explode" for the BOM Item. With this user will enable the checkbox "Do Not Explode" for the item Motherboard and not for the item Keyboard

#docs  https://docs.erpnext.com/docs/v13/user/manual/en/manufacturing/bill-of-materials/edit?wiki_page_patch=1787f5fedc

Fixed https://github.com/frappe/erpnext/issues/17738

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -331,7 +331,7 @@ frappe.ui.form.on("BOM", {
 			});
 		});
 
-		if (has_template_rm) {
+		if (has_template_rm && has_template_rm.length) {
 			dialog.fields_dict.items.grid.refresh();
 		}
 	},
@@ -467,7 +467,8 @@ var get_bom_material_detail = function(doc, cdt, cdn, scrap_items) {
 				"uom": d.uom,
 				"stock_uom": d.stock_uom,
 				"conversion_factor": d.conversion_factor,
-				"sourced_by_supplier": d.sourced_by_supplier
+				"sourced_by_supplier": d.sourced_by_supplier,
+				"do_not_explode": d.do_not_explode
 			},
 			callback: function(r) {
 				d = locals[cdt][cdn];
@@ -639,6 +640,13 @@ frappe.ui.form.on("BOM Operation", "workstation", function(frm, cdt, cdn) {
 		}
 	});
 });
+
+frappe.ui.form.on("BOM Item", {
+	do_not_explode: function(frm, cdt, cdn) {
+		get_bom_material_detail(frm.doc, cdt, cdn, false);
+	}
+})
+
 
 frappe.ui.form.on("BOM Item", "qty", function(frm, cdt, cdn) {
 	var d = locals[cdt][cdn];
--- a/erpnext/manufacturing/doctype/bom/bom.py
+++ b/erpnext/manufacturing/doctype/bom/bom.py
@@ -203,6 +203,10 @@ class BOM(WebsiteGenerator):
 		for item in self.get("items"):
 			self.validate_bom_currency(item)
 
+			item.bom_no = ''
+			if not item.do_not_explode:
+				item.bom_no = item.bom_no
+
 			ret = self.get_bom_material_detail({
 				"company": self.company,
 				"item_code": item.item_code,
@@ -214,8 +218,10 @@ class BOM(WebsiteGenerator):
 				"uom": item.uom,
 				"stock_uom": item.stock_uom,
 				"conversion_factor": item.conversion_factor,
-				"sourced_by_supplier": item.sourced_by_supplier
+				"sourced_by_supplier": item.sourced_by_supplier,
+				"do_not_explode": item.do_not_explode
 			})
+
 			for r in ret:
 				if not item.get(r):
 					item.set(r, ret[r])
@@ -266,6 +272,9 @@ class BOM(WebsiteGenerator):
 			 'include_item_in_manufacturing': cint(args.get('transfer_for_manufacture')),
 			 'sourced_by_supplier'		: args.get('sourced_by_supplier', 0)
 		}
+
+		if args.get('do_not_explode'):
+			ret_item['bom_no'] = ''
 
 		return ret_item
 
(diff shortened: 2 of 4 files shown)


ankush:
If an item is excluded from exploded items why is it added in BOM's tree? 🤔

rohitwaghchaure:
> If an item is excluded from exploded items why is it added in BOM's tree? 🤔

Customer wants to explode the items for specific BOMs and not for other specific BOMs

djpalshikar:
Explaining the use case here :
Product (High Level BOM) -
 * Sub-Assembly 1  : Kept in stock, as manufactured separately by a different department
 * Sub-Assembly 2 : Sub assembly will be exploded in manufacturing

Currently the exploded BOM would have both of these exploded sub-assemblies. While in my case some some assemblies will always be in stock while other will not.

This feature does not hamper or affect any other operation.

djpalshikar:
> If an item is excluded from exploded items why is it added in BOM's tree? thinking

Just to clarify the, if the check box is set, this item will be **included** in the exploded BOM items table as it is. It will not be exploded, even if there is a default BOM

ankush:
> Explaining the use case here :
> Product (High Level BOM) -

> Sub-Assembly 1 : Kept in stock, as manufactured separately by a different department
> Sub-Assembly 2 : Sub assembly will be exploded in manufacturing
> Currently the exploded BOM would have both of these exploded sub-assemblies. While in my case > some some assemblies will always be in stock while other will not.


Hi @djpalshikar 

IMHO This use case is better implemented using Production Plan -> Work Order workflow. 

Production Plan prepares a list of subassembly required so it's easier to use existing stock instead of manufacturing it OR assign that work order for subasssembly to another department.

Basically, when you use "multi-level" BOM you lose the ability to have parallel traceability or configurability of individual assembly/subassembly in BOM's tree. 

Have you considered implementing that workflow instead of a single work-order for multi-level BOM? If not can you share what's blocking you from implementing that workflow?


---

Regardless of it, this feature still helps in other cases (Make/Buy decisions when sometimes you buy subassembly and sometimes you manufacture on case to case basis)

erpnext/manufacturing/doctype/bom/bom.js
```diff
@@ -331,7 +331,7 @@ frappe.ui.form.on("BOM", {
 			});
 		});
 
-		if (has_template_rm) {
+		if (has_template_rm && has_template_rm.length) {
```
ankush, line 334: PS: this was an issue in framework; still makes sense to fix it here also: https://github.com/frappe/frappe/pull/15734

ankush (approved):

ankush:
failing test is unrelated, fixed here: https://github.com/frappe/frappe/pull/15739

djpalshikar:
@ankush yes it works for the make/buy case too. Just wanted to justify the  requirement for future references:
In my case the decision to explode or not to explode is a 'Engineering Design' decision, which is taken once. This is to make the  decision process for production faster.

Our Mechanical Production Manager, need not spend time on deciding whether he should explode Electronics Department assembly. 
For electronics department, the rise in reserved_qty_for_production is sufficient to trigger a Material Request for Manufacture.  

Yes, we loose the parallel traceability via BOMs., however the same is achieved by keeping batch nos for 'stock' sub-assemblies.

Finally regarding Workflows:
- One of our departments manufactures some Sub-assemblies only in large batches (electronics)
- The other department(Mechanical) manufactures Work Order even for 1 nos.

sahil48:
For our use case, we make wiring harnesses to stock and not against order using auto-reorder. For this reason, we plan the wiring harnesses (auto-reorder) separate from the instrument (against production plan), so it matches with what Rohit suggested as the case. Can the item have default value set by user on Item Master for when it is placed in the BOM's item list (On Item Master, say if item by default should have Do not Explode Checked or Unchecked when added to the BOM list)?

djpalshikar:
@rohitwaghchaure has this been ported and merged with V13 as well, I cannot see this on my account.

ankush:

ankush:
@djpalshikar should be part of the next release c. Wednesday.

PR #29451, opened 2022-01-25
feat: Provisional accounting for expenses

deepeshgarg007:
### Step 1: Enable Provisional Accounting in Company Master and set default provision account
Screenshot 2022-01-28 at 1 10 42 PM

### Step 2: Make a Purchase Receipt For a non-stock item
image

Expected GL Entries on posting of Purchase Receipt

image

### Step 3: Make a Purchase Invoice against the Purchase Receipt
image

Expected GL Entries on posting Purchase Invoice 
 
Against Purchase Invoice:
image

Aginst Purchase Receipt (GL entries against purchase receipt should be reversed), remember to remove group by filter in General Ledger report to view this:

image

Now you may be wondering why we are doing all this, lets try and understand this with an example. In many scenarios lets say a services is received and the expense is incurred but the Invoice is done later, so this expense doesn't flow into your month end MIS if the purchase invoice is made in the later month. This feature aims to solve that issue. On posting a purchase receipt for that Non Stock Item provisional accounting entries will be passed and the same will be reversed on posting the actual purchase invoice on the purchase invoice posting date.  
 
docs: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/purchase-invoice/edit?wiki_page_patch=4c596d26e3


nabinhait (dismissed):

nabinhait:
@deepeshgarg007 I think we should add a field for "Provisional Account" in the Purchase Receipt so that the user can change it in the transaction. It can be set by default using the default provisional account defined in the company.

deepeshgarg007:

erpnext/stock/doctype/purchase_receipt/purchase_receipt.json
```diff
@@ -1144,16 +1146,30 @@
...
+   "collapsible": 1,
+   "fieldname": "accounting_details_section",
+   "fieldtype": "Section Break",
+   "label": "Accounting Details"
+  },
+  {
+   "fieldname": "provisional_expense_account",
+   "fieldtype": "Link",
+   "hidden": 1,
+   "label": "Provisional Expense Account",
```
nabinhait, line 1160: Fetch from company if blank

nabinhait (approved):

PR #29454, opened 2022-01-25
fix(sales order): incorrect no. of items fetched while creating dn

nextchamp-saqib:
### Issue:

If a Sales Order has more than 50 items (has multiple pages in item table), and there are some undelivered items. Some undelivered items have a different delivery date.
If some of the "undelivered items" belong to the next page of the items table for eg. an "undelivered item" has an index of 55 (like 51+)

Then while creating Delivery Note using Create button, those "next page" items are not fetched. 

### Why?

The `frm.fields_dict['items'].grid.grid_rows` contains only first 50 rows hence filtering only works on first 50 item rows

https://github.com/frappe/erpnext/blob/c6a17d1966174feb448a732a1014bf4fac4bc32b/erpnext/selling/doctype/sales_order/sales_order.js#L503-L509


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -565,6 +565,12 @@ def update_item(source, target, source_parent):
 	}
 
 	if not skip_item_mapping:
+		def condition(doc):
+			if frappe.flags.args and frappe.flags.args.delivery_dates:
```
nextchamp-saqib, line 570:
```suggestion
			# make_mapped_doc sets js `args` into `frappe.flags.args`
			if frappe.flags.args and frappe.flags.args.delivery_dates:
```

PR #29456, opened 2022-01-25
fix: Incorrect packing list for recurring items & code cleanup

marination:
Introduced via https://github.com/frappe/erpnext/pull/27124

### 1) Recurring Items in Sales transactions
- Add two **same** bundle items in a delivery note on different rows, with different quantities
- The packed items table only shows packed items for the first row and ignores the second recurring item row

### 2) Code comprehensibility
- Long functions
- looping over the packed items table over and over again
- Hard to understand/overly hacky  logic due to confusing function names that do a lot more than stated

## **Fix:**
- Consider recurring bundle items
- `make_packing_list` in **packed_item.py** should give a clear overview of the sequence of events
- smaller functions and lesser nested calls
- reset table if any deletion takes place. No. of loops to regenerate/update the table is the same. Deleting rows separately and looping again adds to the complexity.
- Get rate of packed item from price list. Fallback on Item master's `valuation_rate` if no price is set as per price list

**ToDo:**
- [x] Rewrite price updation logic and add price fetching for packed items
- [x] Tests

**To test:**
- Test product bundle addition/updation/deletion via normal form and update items
- Can test in SO and DN, behaviour is same throughout
- Test recurring items test case
- Test packed items that have batches/serial nos too
- Test enabling "calculate bundle price from packed items" in Selling Settings. Check how packed items price is fetched firstly. Earlier it was blindly pulled from Item master. Now it will fetch based on price list rate/buying price and fallback on  Item master's `valuation_rate`


marination:

PR #29460, opened 2022-01-25
feat: Refund entry against loans

deepeshgarg007:
- Separated out payment account in Loan Type and Loan Doctype into two separate accounts"Disbursement Account" and "Repayment Account" to handle scenarios where Loan Disbursement and  Repayment can happen into two separate bank accounts.  Also added a patch for this to update the account values accordingly

- Add a whitelisted method to create an excess amount refund against loans via Journal Entry

`no-docs`


deepeshgarg007:

PR #29461, opened 2022-01-25
fix: item dashboard move/add functionality behaviour

### LINKED ISSUE #20827: Add button on item dashboard bypasses workflow
image

The above button allow stock entries that bypass the workflow to be created.

ankush:
closes https://github.com/frappe/erpnext/issues/20827

Problem:

1. Currently item dashboard allows creating inventory out of thin air with 2 clicks 🪄 (against stock adjustment account) which is often not desired by many (with exception of devs who just want to quickly test things by adding arbitrary qty and rate)
2. This also bypasses workflows set up by organizations for approval of such stock transactions. 

Proposed change: The default option should be "Open stock entry" instead of auto-submitting it. 

Additional smallish fix: only show leaf/child warehouses in target warehouse list. 

https://user-images.githubusercontent.com/9079960/151048025-68e6c462-0a3a-4e34-99b2-cd77a6309231.mov

To test:

- check the move/add functionality of the item dashboard. Nothing else should be affected by this change.


ankush:
dont have time for this right now; will reopen later.

PR #29471, opened 2022-01-27
fix: Further sort sales_order_analysis to get consistent response

surajshetty3416:
- Since the data was sorted by `transaction_date`, the report sorting order used to be inconsistent every time if there were lots of records with the same transaction date. This is used to create problems while exporting reports as "Excel" with filters.

For more details
https://frappe.io/app/issue/ISS-21-22-06182

more info:

> while exporting data as Excel with column filter, we pass “visible_idx” from client-side… but due to inconsistent response from the server (DB) the… server picks wrong column as visible row.


surajshetty3416:

PR #29474, opened 2022-01-27
Patch 1

smehata:
stock leger undefined issue[](url
Screenshot from 2022-01-27 12-23-20
)


ankush:
@smehata I think wrong base branch was selected; if you reopen I can move it to correct one

PR #28418, opened 2021-11-16
fix(e-invoicing): totals validation of e-invoices

nextchamp-saqib:
To prevent repeated errors and create logs to help in debugging

PR #29480, opened 2022-01-27
revert: "refactor!: drop e-invoicing integration from erpnext"

nextchamp-saqib:
Bringing back e-invoicing integration into v14 as the beta is being released. 

The development of the Indian regional app developed by @sagarvora  would include e-invoicing integration. So until the app is ready, e-invoicing has to stay in v14

This PR will restore the doctypes & custom fields deleted in the previous PR (#26940)

---

Also, port new fixes & features of e-invoicing from v13 branch
#27375, #28084, #28418, #28547, #28235


erpnext/patches/v12_0/add_einvoice_summary_report_permissions.py
```diff
@@ -0,0 +1,20 @@
+from __future__ import unicode_literals
```
ankush, line 1: remove `future` and `six` imports. It's right now working only because some package has those dependency, we don't. It will start breaking randomly in future.

nextchamp-saqib: Done

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -0,0 +1,1171 @@
...
+)
+from frappe.utils.scheduler import is_scheduler_inactive
+from pyqrcode import create as qrcreate
+
+from erpnext.regional.india.utils import get_gst_accounts, get_place_of_supply
+
+
+@frappe.whitelist()
+def validate_eligibility(doc):
+	if isinstance(doc, frappe.string_types):
```
ankush:
```suggestion
	if isinstance(doc, str):
```

This will fail too 😄 

It HAS failed before: https://github.com/frappe/erpnext/pull/26947

nextchamp-saqib:

PR #29498, opened 2022-01-29
fix(ui): Only allow csv files to be uploaded with Rename Tool

dj12djdjs:
closes https://github.com/frappe/frappe/issues/15613

### DIFF
--- a/erpnext/utilities/doctype/rename_tool/rename_tool.js
+++ b/erpnext/utilities/doctype/rename_tool/rename_tool.js
@@ -13,6 +13,12 @@ frappe.ui.form.on("Rename Tool", {
 	},
 	refresh: function(frm) {
 		frm.disable_save();
+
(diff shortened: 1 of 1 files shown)


dj12djdjs:
Who is in charge of this PR? 👀

erpnext/utilities/doctype/rename_tool/rename_tool.js
```diff
@@ -11,6 +11,17 @@ frappe.ui.form.on("Rename Tool", {
 			}
 		});
 	},
+	onload_post_render: function(frm) {
+		$(frm.get_field('file_to_rename').$input).click(() => {
+			frm.get_field('file_to_rename')
```
ankush: better handled by setting df.option 😅 

will push the change.

dj12djdjs: Noted, thank you

PR #29504, opened 2022-01-29
fix: Employee Onboarding and Separation UX

### LINKED ISSUE #29470: Employee Onboarding & Separation Improvements
### Information about bug

Reference : [26738](https://github.com/frappe/erpnext/pull/26738)

Issues - 

EMPLOYEE ONBOARDING : 

1. Employee Onboarding link missing from Employee’s connections form. 
2. When an employee is created from Job Offer, the offer date should be mapped as is from the Job Offer record itself. Right now there’s no such mapping 

Test case 3

3. If we create employee onboarding for an employee then the details like designation, department, grade and holiday list should be fetched for that employee as per the same mentioned in the Employee record. Also employee name linkage is missing in employee onboarding record on save. 

Test case 4

4. On cancellation of employee onboarding, all the related projects and tasks get deleted altogether. The user has no way to find out if these projects or tasks got deleted.  We can add a confirmation message, like something that says "If you cancel this employee onboarding record, related projects and tasks will be deleted permanently. Do you want to cancel ?" An example : 

message example

5.  On cancellation of employee onboarding, if I amend the same and try to create afresh it still says that an employee onboarding record exists. So workaround is to delete the cancelled document and then create again. However let’s say for Sales Order user can still create record for same customer without having to create new record altogether. 

cancel onboarding handling

Expected behaviour : While validating it should check for document status as cancelled and then let the user proceed accordingly instead of preventing the user

6. In the project created against employee onboarding, under Start and End Dates section, there's a field called "Actual Time (in hrs)". This time is calculated based on timesheets. A small message indicating on what basis this is calculated would be helpful 

image

EMPLOYEE SEPARATION  : 

1. Can we completely remove the Required for Onboarding checkbox from activities child table in Employee Separation Template? This doesn't make sense to be present in the separation doctypes 

image

2. On creation of employee separation, projects and tasks are created. When viewing tasks in Gantt, all tasks are shown appropriately but in calendar view one task is missing 

Incorrect task count

Test instance : https://v14-pre-release.frappe.cloud/

### Module

HR

### Version

V14

### Installation method

_No response_

_No response_

ruchamahabal:
Closes https://github.com/frappe/erpnext/issues/29470

### DIFF
--- a/erpnext/controllers/employee_boarding_controller.py
+++ b/erpnext/controllers/employee_boarding_controller.py
@@ -132,12 +132,16 @@ class EmployeeBoardingController(Document):
 
 	def on_cancel(self):
 		# delete task project
-		for task in frappe.get_all('Task', filters={'project': self.project}):
+		project = self.project
+		for task in frappe.get_all('Task', filters={'project': project}):
 			frappe.delete_doc('Task', task.name, force=1)
-		frappe.delete_doc('Project', self.project, force=1)
+		frappe.delete_doc('Project', project, force=1)
 		self.db_set('project', '')
 		for activity in self.activities:
 			activity.db_set('task', '')
+
+		frappe.msgprint(_('Linked Project {} and Tasks deleted.').format(
+			project), alert=True, indicator='blue')
 
 
 @frappe.whitelist()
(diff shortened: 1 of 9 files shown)


ruchamahabal:

Ponyooooo:
will this fix be included on the v13 so that we can utilize and test the revisions? Right now the HR feature for the employee onboarding and separation are both not used because of the issues raised.

PR #29519, opened 2022-01-30
fix: bin uniqueness

ankush:
- In very rare cases concurrent updates can cause the creation of duplicate Bins.
- In some cases, item/warehouse is not specified for a new bin, such bins are of no use.

Fix:
- Add DB level unique constraint
- Handle constraint failure but only from known util methods.
- Add mandatory constraint on item_code and warehouse
- Add pre-model-sync patch for accommodating new constraints and fixing broken bins if any.

### DIFF
--- a/erpnext/patches/v13_0/add_bin_unique_constraint.py
+++ b/erpnext/patches/v13_0/add_bin_unique_constraint.py
@@ -11,35 +11,47 @@ from erpnext.stock.stock_balance import (
 
 
 def execute():
+	delete_broken_bins()
+	delete_and_patch_duplicate_bins()
 
-	duplicate_rows = frappe.db.sql("""
+def delete_broken_bins():
+	# delete useless bins
+	frappe.db.sql("delete from `tabBin` where item_code is null or warehouse is null")
+
+def delete_and_patch_duplicate_bins():
+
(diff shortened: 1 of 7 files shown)


s-aga-r (dismissed):
@ankush everything looks good

erpnext/patches/v13_0/add_bin_unique_constraint.py
```diff
@@ -0,0 +1,57 @@
...
+
+		for broken_bin in existing_bins:
+			frappe.delete_doc("Bin", broken_bin.name)
+
+		qty_dict = {
+			"reserved_qty": get_reserved_qty(duplicate_bin.item_code, duplicate_bin.warehouse),
+			"indented_qty": get_indented_qty(duplicate_bin.item_code, duplicate_bin.warehouse),
+			"ordered_qty": get_ordered_qty(duplicate_bin.item_code, duplicate_bin.warehouse),
+			"planned_qty": get_planned_qty(duplicate_bin.item_code, duplicate_bin.warehouse),
+			"actual_qty": get_balance_qty_from_sle(duplicate_bin.item_code, duplicate_bin.warehouse)
```
ankush: production/subcontract reserved qty need to be updated

PR #29522, opened 2022-01-30
refactor!: dynamically compute bom_level

ankush:
https://github.com/frappe/erpnext/pull/26240 introduced a field called "BOM level" to properly sort subassemblies in the production plan. This however relied on value present in BOM. 

Since a BOM's level is not a static value but depends on context, this has to be computed dynamically. Thankfully `indent` is exactly that, starting with the 0th level and recursively adding the level as it goes. (another alternative was to use `BOMTree` class which generates this on the fly too)

Breaking Change:
- The BOM level shown now is opposite of what it used to be and hence sorted in a reverse manner. The point is to make work order for the lowest level of subassemblies first and work upwards in the tree so you always have subcomponents available. 

https://github.com/frappe/erpnext/blob/157b4b33fe633ed96c7e36fdcc81c9f2245241ec/erpnext/manufacturing/doctype/production_plan/production_plan.py#L563

Secondary fix:

- Sort subassembly items globally instead of doing it for a single production item. (refer test for more info)

Example:

Suppose we are manufacturing 2 items with nested tree like this:
```
Assembly
┣━ SubAssembly1
┃  ┣━ ChildPart1
┃  ┣━ ChildPart2
┣━ SubAssembly2
┃  ┣━ ChildPart3
┣━ SubAssembly3
┃  ┣━ SubSubAssy1
┃  ┃  ┣━ ChildPart4
┣━ ChildPart5
┣━ ChildPart6
┣━ SubAssembly4
┃  ┣━ SubSubAssy2
┃  ┃  ┣━ ChildPart7


MegaDeepAssy
┣━ SecretSubassy
┃  ┣━ SecretPart
┃  ┃  ┣━ VerySecret
┃  ┃  ┃  ┣━ SuperSecret
┃  ┃  ┃  ┃  ┣━ Classified
```

Then generated list of subassembly should be like this:

```
SuperSecret
VerySecret
SubSubAssy1
SubSubAssy2
SecretPart
SubAssembly1
SubAssembly2
SubAssembly3
SubAssembly4
SecretSubassy
```

Basically, items that are at the lowest level in BOM tree should appear first. Similar to Work Order fix: https://github.com/frappe/erpnext/pull/25886


rohitwaghchaure (approved):

rohitwaghchaure:

djpalshikar:
@ankush how will the his change affect the BOM update tool ?

ankush:
> @ankush how will this change affect the BOM update tool?


It doesn't affect the BOM update tool. 

We were using this field only for sorting production plan subassemblies and nothing else.

PR #29531, opened 2022-01-31
fix: not able to declare opportunity lost without competitors

### LINKED ISSUE #29252: Able to save Lost opportunity without providing reason
### Information about bug

1. Create an opportunity, Save
2. Mark it as Lost, a modal appears wherein you provide the Lost reason (mandatory) and Competitors.
3. Just close the modal or make it disappear by clicking in the dark area
4. you are still able to Save the Opportunity. 

THe user is not able to provide the reason as both the modal and the sections do not appear

Possible solution
when the user selects the status 'Lost' also make the section appear, if not through modal the user can select the reason from the section.
image

### Module

CRM

### Version

v14

### Installation method

_No response_

_No response_

anupamvs:
issue- 
in opportunity doc Select the lost reason and click on Declare Lost. System throws this error
telegram-cloud-photo-size-5-6323301890170925858-y

Closes #29252

### DIFF
--- a/erpnext/selling/sales_common.js
+++ b/erpnext/selling/sales_common.js
@@ -499,7 +499,7 @@ frappe.ui.form.on(cur_frm.doctype, {
 					method: 'declare_enquiry_lost',
 					args: {
 						'lost_reasons_list': values.lost_reason,
-						'competitors': values.competitors,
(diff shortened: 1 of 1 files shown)


anupamvs:

PR #29535, opened 2022-01-31
fix: do not hide Loan Repayment Entry field in salary slip

ruchamahabal:
When loan repayment is done via Salary Slip, a Loan Repayment entry is made on salary slip submission and linked in the Salary Slip Loans child table. However, this link is hidden, making it difficult for the user to discover the repayment entry made. Disable `hidden` for this field.


ruchamahabal:
Unrelated test case failures will be fixed in: https://github.com/frappe/erpnext/pull/29539

ruchamahabal:

PR #29536, opened 2022-01-31
feat: Cost Center Allocation

nabinhait:
## Cost Center Allocation

Cost Center Allocation is a feature using which the general ledger entry against a cost center can be split against multiple cost centers. In the Cost Center Allocation document, you can define allocation percentages of the child cost centers.

In a growing business, it becomes a necessity to analyse the income/expenses against each business unit of the organisation. And to do that, we need to treat each businees unit as a cost center and book income/expenses against the cost center. But if we need to split it every time at the transaction level manually, it becomes very difficult. That's when this Cost Center Allocation feature comes to the rescue.

In ERPNext, we just need to define the allocation between multiple cost centers for a specific master/main cost center. Then whenever we book an invoice or expense transaction against the main cost center, the system automatically split it based on allocation and posts gl entries against each child cost center.

cost-center-allocation

### GL Entries against Transaction
While booking any transaction against the main cost center, the system automatically split the GL Entry against it and posts multiple gl entries based on the applicable cost center allocation record.

general-ledger-based-on-cost-center-allocation

(GL Entries against a Sales Invoice has been splitted based on Cost Center Allocation)

### What changed from the earlier Distributed Cost Center?
In version 13, we have a feature called [Distributed Cost Center](https://docs.erpnext.com/docs/v13/user/manual/en/accounts/distributed-cost-center) using which you can define the allocation percentages between multiple cost centers. Earlier, we used to do it inside Cost Center master. Based on the allocation, we used to get the multiple financial reports based on the distribution, but it had no effect in the actual GL Entries. 
It has certain drawbacks, like it does not take care of changes in the distribution overtime. In the real world, cost center allocation keeps changing over time and based on updated allocation, system should book income/expenses against it. It cannot be calculated run-time.

### Patch
A patch has been written to migrate the data from old system to the new.

### Docs
https://docs.erpnext.com/docs/v13/user/manual/en/accounts/new?wiki_page_patch=fee1a96a05

### Test Cases:

- Validations:
    - Total allocation percentage is 100
    - From  Date must be after last gl entry against main cost center
    - Main cost center cannot be entered as a child
    - Main cost center cannot be used in other allocations as child
    - Child cost center is not used in other allocation as the main

- Logic:
    - Multiple GL Entries are posted based on allocation

- Reports:
    - Code related to the old distributed cost center has been deleted from the following reports, hence needs manual testing.
        - Financial Statements
        - General Ledger
        - Profitability Analysis
        - Budget Variance Report


sammish:
@nabinhait 
congratulations  Nice change "Valid From "
image

jigneshpshah:
How can this https://docs.erpnext.com/docs/v14/user/manual/en/accounts/distributed-cost-center be deleted ?

As distributed cost center is no more available in v14.
As such it was a good feature to view the Financial Statements based on the likely distribution % ! :-)

sammish:
@jigneshpshah Its available in v14 which called cost center allocation

jigneshpshah:
@sammish please read my comments again 😃

PR #25735, opened 2021-05-17
feat: Employee reminders

NagariaHussain:
- [x] New Reminders Section in HR Settings
- [x] Work Anniversary Reminders
- [x] Holiday Reminders

Screenshot 2021-08-24 at 2 55 29 PM

Show a warning in case a change in frequency of holiday reminders will lead to missing reminders for a date range:

https://user-images.githubusercontent.com/34810212/118953441-70875c80-b97a-11eb-8a12-489e5cb66318.mp4

Refactoring:

1. New `employee_reminders` and `test_employee_reminders` modules
2. is_holiday: "Explicit is better than implicit"
3. get_holidays_for_employee 
4. HR Settings Reminders Section
5. Holiday Reminders on the day of the event was removed in: https://github.com/frappe/erpnext/pull/25735/commits/836dfffc3093d5af16be5e85426e965a22e7ef21

Tests:

1. is_holiday: Previously there were no explicit tests for this function
2. Reminder emails
 
`no-docs` (temp)

PR #29548, opened 2022-02-01
fix: employee reminders fixes

ruchamahabal:
Introduced via https://github.com/frappe/erpnext/pull/25735 as mentioned in https://github.com/frappe/erpnext/pull/25735#issuecomment-1026481465

- Send Reminders to Active employees only
- Don't send reminders if there are no holidays: 
  
   image

- Use optional plural in work anniversary reminder message: years -> year(s)

  1-years

### DIFF
--- a/erpnext/hr/doctype/employee/employee_reminders.py
+++ b/erpnext/hr/doctype/employee/employee_reminders.py
@@ -51,7 +51,7 @@ def send_advance_holiday_reminders(frequency):
 			raise_exception=False
 		)
 
-		if not (holidays is None):
+		if holidays:
 			send_holidays_reminder_in_advance(employee, holidays)
 
 def send_holidays_reminder_in_advance(employee, holidays):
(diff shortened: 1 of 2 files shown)


ruchamahabal:

erpnext/hr/doctype/employee/employee_reminders.py
```diff
@@ -225,7 +235,7 @@ def get_work_anniversary_reminder_text_and_message(anniversary_persons):
 			names.append(person_text)
 			# Number of years completed at the company
 			completed_years = getdate().year - person['date_of_joining'].year
-			person_text += f" completed {completed_years} years"
+			person_text += f" completed {completed_years} year(s)"
```
ruchamahabal, line 238: There is no sense in testing this. - -
													   -

ruchamahabal, line 238: Don't need to cover this change in tests. Merging without green tick

PR #29578, opened 2022-02-01
fix: Future recurring period calculation for additional salary

nabinhait:
In recurring additional salary, when you set the start date from the current payroll period and the end date from the next payroll period, it is used to calculate the wrong number of recurring months. For tax calculation, the recurring period should be calculated only based on the current payroll period.

Example:
        Payroll Period: 01-01-2021 to 31-03-2022
	From Date: 01-10-2021
	To Date: 31-05-2022

Before fix: 
Recurring months = 8 months

After fix:
Recurring months = 6 months

Based on recurring months, the system calculates recurring additional salaries for pending months of the year which contributes to calculating net taxable salary for the employee.


nabinhait:

nabinhait:

PR #29601, opened 2022-02-02
fix: Incorrect provisional profit and loss in balance sheet

deepeshgarg007:
If there are more than one balance sheet root accounts of one root type for eg "Provisions" of root type "Liability" in this case the Balance sheet shows some balance in Provisional Profit and Loss even though the previous fiscal years has been closed as it was considering only one root account. This PR solves that issue

Screenshot 2022-02-02 at 10 45 39 PM


deepeshgarg007:

PR #29607, opened 2022-02-03
feat: show stock value difference on Stock Ledger report

ankush:
Show stock value difference and sum of stock value difference so it's easy to infer the accounting impact of a particular voucher. 

Screenshot 2022-02-03 at 10 52 35 AM

TODO:
~~don't show totals for all the rows. (requires FW change)~~ deferred, not a small change as it seemed first - https://github.com/frappe/frappe/issues/15866 

`no-docs`


erpnext/stock/report/stock_ledger/stock_ledger.json
```diff
@@ -1,24 +1,27 @@
...
- "is_standard": "Yes", 
- "modified": "2017-02-24 20:14:07.626391", 
- "modified_by": "Administrator", 
- "module": "Stock", 
- "name": "Stock Ledger", 
- "owner": "Administrator", 
- "ref_doctype": "Stock Ledger Entry", 
- "report_name": "Stock Ledger", 
- "report_type": "Script Report", 
+ "add_total_row": 1,
```
ankush, line 2: this is the only change, rest is autoformat from FW.

PR #29609, opened 2022-02-03
fix: Zero rated exports in GSTR-3B report

deepeshgarg007:
Export Invoices without payment of taxes were not showing up in GSTR-3B report

Screenshot 2022-02-06 at 7 04 33 PM

Steps to test:
1. Make an overseas customer with GST Category as "Overseas" and "Export Type" as "Without Payment Of Tax"
2. Create a GST Sales Invoice against this customer, taxes won't be applied since the export type is without payment of tax
3. Generate GSTR-3B report for this month and check if the taxable value show up as seen above


deepeshgarg007:

PR #29624, opened 2022-02-03
chore: update de.csv

wojosc:
minor translation error


ankush:
Hi, while we have our translation system back up the PRs are fine but can you please add correct title as per semantic commit guideline?
`chore: update german translations` is fine.

wojosc:
> Hi, while we have our translation system back up the PRs are fine but can you please add correct title as per semantic commit guideline? `chore: update german translations` is fine.

Thanks for the remark. Will do in future!

PR #27923, opened 2021-10-12
refactor: E-commerce (port to develop)

marination:
Depends on https://github.com/frappe/frappe/pull/15831

----------

front port of https://github.com/frappe/erpnext/pull/24603

ToDo:

- [x] Finish Porting all commits (fixes and from https://github.com/frappe/erpnext/pull/24603)
- [x] Test functionality manually (need to fix build issue on cart and item full page)
- [x] fix server side tests (cherry pick fixes)

PR #29625, opened 2022-02-03
fix: Load control & dialog assets with `cart.js`

surajshetty3416:
Previously, control & dialog assets were getting loaded only on visiting "/cart", but since cart.js sets up various events on the homepage, dialog-related assets are required up-front.

fixes https://github.com/frappe/frappe/issues/15869
image

---

The issue was introduced via https://github.com/frappe/erpnext/pull/27923

----------------------------------
## Final Fix:
**Issue:**
- `cart.js` was added to the erpnext-web bundle, so it loaded on every single page
- It requires controls bundle from frappe to make dialogs in it's file, which it relies on `cart.html` for
- Since only js was loaded on login, the lack of controls bundle broke the  js asset > and thus the login page wouldn't work (static page)
- This file shouldn't have been in the bundle since it is specific to `cart.html`, not a generic utility file unlike the others.

**Fix:**
- Move `cart.js` to /templates/pages with `cart.html`. They are tightly bound and exclusive to eachother
- Remove cart.js import from bundle


marination:
@surajshetty3416 I think cart.js should not be in the bundle, I must have absent mindedly put that in. Its js for `cart.html`. Not generic enough to be in the bundle. 
Relocating the file and removing the import in the bundle would solve this issue.
It should not impact the login page 🥲 
I'll relocate it and test if it breaks anything

marination:
Login works:


Cart works too:

 
Can confirm for other pages too (Source: trust me)

marination (approved):

marination:

PR #29626, opened 2022-02-03
fix: subscription recovery

barredterra:


deepeshgarg007:
@barredterra is there anything pending here or you just wanted to add the test case?

barredterra:
@deepeshgarg007 I wanted to show that **Subscription** breaks in this case — the test fails. Maybe @hrwX will be able to fix **Subscription** so that the test will pass.

PR #29635, opened 2022-02-04
fix: Incorrect tax template in Sales Invoice via data import

deepeshgarg007:
Correct GST template based on party details was not getting fetched when Importing Sales Invoice via data import tool
This PR fixes that issue


deepeshgarg007:

deepeshgarg007:
@Mergifyio backport develop

PR #29639, opened 2022-02-04
perf: Weed out disabled variants via sql query instead of pythonic looping separately

marination:
## Issue:
- If the number of variants are large (almost 2lakhs), the query to get variants and attribute data takes time
- If the no.of disabled attributes is large as well, the list comprehension weeding out disabled variants takes forever
- We dont need to loop over the variants data so many times
- Avoid any `if a in list(b)` is best when the iterables have tremendous data

**Fix:**
- Use Query to weed out disabled items by doing a simple `JOIN`. Its much faster than `[2 lakh items] not in [61k items]`

**Stats:**
- Earlier Job would time out even in long queue (over 45 mins)
  Screenshot 2022-02-08 at 11 46 23 AM

- Now:
  Screenshot 2022-02-08 at 11 29 40 AM
> - **Max: 1min 38 secs** for approx 1lakh 70k Items` 
> - **Min: 17 secs for approx 2 lakh items**

> (including disabled items weeded out in query)

**ToDo:**
- [x] Load testing and perf stats (currently tested the queries and estimated the impact on prod)

**To test:**
- If cache building is enqueued and is executed correctly (Disable a variant and check if popup in website now excludes this variant)
- Post that, if selector works as expected (try creating a new template-variant and testing that on the portal too)


marination:
Unrelated `test_back_dated_entry_not_allowed` failing

marination:

PR #29641, opened 2022-02-04
fix: Ignore linked invoices on Journal Entry cancel

deepeshgarg007:
The user used to get below warning on canceling a Journal Entry that has been allocated to invoice as an Advance payment and the Sales Invoice also used to get canceled on canceling the Journal Entry

image

Only journal entry should be canceled and the invoice should be just delinked and not canceled.

Steps to test:

1. Create an advance Journal Entry against a customer
2. Create Sales Invoice against the same customer and allocate advance payment entry.
3. Cancel the payment entry and check, Sales Invoice should not get canceled


deepeshgarg007:

PR #29642, opened 2022-02-04
fix: Regenerate packed items on newly mapped doc

### LINKED ISSUE #29634: Incorrect packing list calculation while updating quantity from SO --> DN
### Information about bug

Issue :  When adding single product bundle with qty as 1 on one row and same product bundle with qty as 3 on another row and check if packing list is generated correctly in SO. Now submit SO and create DN, delete first row's qty to 2. Packing list for qty 1 still appears and incorrect packing list with incorrect qty is generated 

Reference : https://github.com/frappe/erpnext/pull/29456

Steps used to reproduce : 

1. In SO add single product bundle with qty as 1 on one row and same product bundle with qty as 3 on another row
2. Now save and submit SO.
3. Create DN and update the first row's qty to 2 and let second row remain as is.

Screen Recording : 

Product Bundle case issue

Observation : 

Issue no. 1 - Packing list for qty 1 still appears. It should have been cleared
Issue no. 2 - Packing list for qty 3 gets duplicated again

### Module

selling

### Version

V 13.20.0

### Installation method

_No response_

_No response_

marination:
Fixes https://github.com/frappe/erpnext/issues/29634

**Fix:**
- Cannot determine action on newly mapped DN (from SO) that hasn't been inserted
- Rows could have been deleted, updated, added, etc. before first save
- In this case , reset packing table
- Should work as expected after fix

https://user-images.githubusercontent.com/25857446/152569034-2575df96-6f33-449e-b445-9808a8c583d0.mov

### DIFF
--- a/erpnext/stock/doctype/packed_item/packed_item.py
+++ b/erpnext/stock/doctype/packed_item/packed_item.py
@@ -27,8 +27,7 @@ def make_packing_list(doc):
 
 	stale_packed_items_table = get_indexed_packed_items_table(doc)
 
-	if not doc.is_new():
-		reset = reset_packing_list_if_deleted_items_exist(doc)
+	reset = reset_packing_list(doc)
 
 	for item_row in doc.get("items"):
 		if frappe.db.exists("Product Bundle", {"new_item_code": item_row.item_code}):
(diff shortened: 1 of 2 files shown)


marination:

PR #29650, opened 2022-02-05
fix: earned leaves not allocated if assignment is created on month-end based on Leave Period

ruchamahabal:
## Background

1. On the creation of a **Leave Policy Assignment** for **Earned Leaves**, allocation of leaves happens in 2 stages:
    - If this assignment is done in the latter months of the leave period, it allocates leaves for past months. 
    - then, a daily background job runs for allocating "earned" leaves at the end of every month (in case of monthly frequency)
2. For eg: the leave period starts from Jan 2022 and you create an assignment in Feb 2022 then it will allocate leave for Jan and then allocate leaves for Feb onwards during month-end as per your frequency set in the leave type

## Problem
1. If you create an assignment on month-end, say 31st Jan 2022, and the policy period starts from 1st Jan 2022, then 1 leave should be allocated since it is the last day of the month.
2. But this does not happen because as per the current logic, the month has not "passed".
3. Also, the daily background job would run at 12:00 am on 31st Jan 2022, but if you create an assignment after it, then the job is already over and it will next run on 1st Feb 2022, but this is not the last day of the month, so leave is never allocated for Jan.

<details> 

<summary> Steps to Replicate </summary>

- Leave Type configuration

  image

- Leave Policy:
   
  image

- Leave Policy Assignment created for Jan 2022-Dec 2022 on 31st Jan 2022
  
  image

- 0 _Total Leaves Allocated_ in Leave Allocation although Jan has passed (it is 31st)
   
   image

</details>

## Fix

1. If the assignment is created on the last day of the month within the leave period, consider it in "past months" because the employee should already "earn" this leave.
2. Add a check in the background job allocation to determine if the leave is not already allocated, to avoid duplication.
3. Add tests for both cases.

1 Leave allocated for Jan on policy assignment creation. Check _Total Leaves Allocated_

image

- [x] Handle carry-forwarded leaves while checking if leaves are already allocated

### DIFF
--- a/erpnext/hr/doctype/leave_application/test_leave_application.py
+++ b/erpnext/hr/doctype/leave_application/test_leave_application.py
@@ -546,7 +546,7 @@ class TestLeaveApplication(unittest.TestCase):
 		from erpnext.hr.utils import allocate_earned_leaves
 		i = 0
 		while(i<14):
-			allocate_earned_leaves()
+			allocate_earned_leaves(ignore_duplicates=True)
 			i += 1
 		self.assertEqual(get_leave_balance_on(employee.name, leave_type, nowdate()), 6)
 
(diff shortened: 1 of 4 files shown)


ruchamahabal:

PR #29654, opened 2022-02-05
revert: BU Schlüssel (a21f76f)

barredterra:
Turns out this produces more problems than it solves:

- Added "40" to every row to disable automatic VAT deduction
- Problem: the target software complains if there's not VAT to disable

We just need to configure the right accounts (without automatic tax) to solve this problem properly.


barredterra:

deepeshgarg007:

PR #29659, opened 2022-02-06
fix: Billing status for zero amount ref doc

deepeshgarg007:
Billing Status was not getting updated for 0 value Sales Order on creation of invoice against them

Steps to test:

1. Create a Sales Order with rates for all items as 0, make sure the grand total is 0
2. Create a Sales Invoice against the order, on submitting the Invoice the status of Invoice should change from "Not Billed" to either "Fully Billed" or "Partially Billed"


deepeshgarg007:

PR #29660, opened 2022-02-06
fix: Add missing key in Loan

abhinavxd:
* Added missing `disbursed_amount` field for Loan.
* Without which  `get_pending_principal_amount:510` which is in same file throws error.


deepeshgarg007:

PR #29674, opened 2022-02-07
 fix(ux): make stock entry type the title field

ankush:
- remove standard title field
- make "stock entry type" show up as title field instead. 

Screenshot 2022-02-07 at 1 07 57 PM

If users want something else they can customize it easily by adding a new field and making it title field.

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.json
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.json
@@ -8,7 +8,6 @@
  "engine": "InnoDB",
  "field_order": [
   "items_section",
-  "title",
   "naming_series",
   "stock_entry_type",
   "outgoing_stock_entry",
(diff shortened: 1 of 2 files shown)


swistakm:
@ankush This is a bit disappointing (and disturbing) to discover this type change after the minor update of  ERPNext.

> If users want something else they can customize it easily by adding a new field and making it title field.

Users of my instance were genuinely using that field to label their Stock Entries. After the update it is gone. Of course, I could recover it using custom field but adding it back this way made me feel a bit anxious.

ankush:
@swistakm 

> Users of my instance were genuinely using that field to label their Stock Entries. 

It's standard, read only field that's forcefully set to another field in code. We can't support badly done customisation.


If you need it back just add custom field called "title" and everything will like it used to be before.

PR #29679, opened 2022-02-07
fix: ignore cancelled SLEs

ankush:
TODO:
- [x] check impact on LCV (which deliberately cancels and resubmits entries) (unaffected cause PR/PI have different GL entry computation function)
- [x] impact on other transactions using this method. 

PS: This does not affect you if you only create docs using Desk or API. Defensive fix in a way.


ankush:
that BOM report failure is unrelated, fixed separately. 
https://github.com/frappe/erpnext/pull/29684

PR #29681, opened 2022-02-07
fix(ux): make "allow zero valuation rate" readonly if "s_warehouse" is set

s-aga-r:
Whenever source warehouse is used you can't say "allow zero valuation". Source warehouse means the item already exists and getting consumed, so whatever rate exists has to be utilized. So making the checkbox "allow zero valuation rate" read-only in cases where it doesn't make sense. 

https://user-images.githubusercontent.com/63660334/152768964-02aff30b-d235-4fa5-9a9a-dd4947350232.mov


erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
```diff
@@ -346,7 +346,8 @@
    "fieldtype": "Check",
    "label": "Allow Zero Valuation Rate",
    "no_copy": 1,
-   "print_hide": 1
+   "print_hide": 1,
+   "read_only_depends_on": "eval: doc.s_warehouse != \"\""
```
ankush: `!!doc.s_warehouse` would be better?

By default the field is `undefined` when user empties it it becomes `""`

erpnext/stock/doctype/stock_entry_detail/stock_entry_detail.json
```diff
@@ -346,7 +346,8 @@
    "fieldtype": "Check",
    "label": "Allow Zero Valuation Rate",
    "no_copy": 1,
-   "print_hide": 1
+   "print_hide": 1,
+   "read_only_depends_on": "eval: doc.s_warehouse"
```
ankush: Tested locally and found one more UX problem.

You can toggle the checkbox and then select s_warehouse, then it will have an incorrect value AND won't let you change it. 

Should be fixable by setting the checkbox to zero whenever the source warehouse is set.

PR #29690, opened 2022-02-07
fix: ignore rate validation for work order

bhavesh95863:
Not able to start work order because one of my row materials does not have a rate and gives the below error message so we should ignore the rate validation for making stock entry from the work order.
image


ankush:
@mergifyio copy develop

ankush (approved):

PR #29691, opened 2022-02-07
added Modul translation

wojosc:
added translations for Moduls for the US local

https://en.wikipedia.org/wiki/Inventory


deepeshgarg007:
@wojosc you can add a custom translation for this
This cannot be accepted in core

PR #29696, opened 2022-02-07
fix: do not reset asset_category

hrwX:
- Asset category is being reset when `set_asset_naming_series` is called since it resets the docs which lead to `asset_category` being reset to `null`, since `frappe.model.sync` resets the doc.
- For now, have fixed this for assets.
- Steps to reproduce the issue
    - Create a new **Item**
    - Enable _Is Asset_
    - Focus the field _Asset Category_, scroll down and click "Create New Asset Category"
    - Fill out and save the **Asset Category**
- Before Fix

    https://user-images.githubusercontent.com/14891507/151782340-5abe13b6-b5d5-4774-a415-3b088515bdf0.mov

- After Fix

    Peek 2022-02-07 18-15

- Partially fixes [Frappe/Issue/15810](https://github.com/frappe/frappe/issues/15810).


barredterra (commented):
Works as expected

erpnext/stock/doctype/item/item.js
```diff
@@ -165,11 +165,11 @@ frappe.ui.form.on("Item", {
...
 
-		frm.call({
-			method: "set_asset_naming_series",
-			doc: frm.doc,
-			callback: function() {
+		frappe.call({
+			method: "erpnext.stock.doctype.item.item.get_asset_naming_series",
+			callback: function(r) {
 				frm.set_value("is_stock_item", frm.doc.is_fixed_asset ? 0 : 1);
+				frm.asset_naming_series = r.message;
```
barredterra: Maybe there's a better way to pass this value to `set_asset_naming_series()`?

barredterra:
Failing unit tests are unrelated

erpnext/stock/doctype/item/item.js
hrwX: Lemme check this one

nextchamp-saqib (approved):

PR #29705, opened 2022-02-08
fix: Set Pending Qty in Prod Plan after updating Work Order

marination:
**Issue:**
- **Pending Qty** in Production Plan Items table stays **unchanged** even after Work Order is completed against it
- For **independent rows** (not pulled from SO or MR), **Pending Qty** is never set even in Draft. It's always 0 from start to end.

**Fix:**
- Set Pending Qty in Prod Plan after updating Work Order
- Initialise Pending Qty on save only for independent rows. After submit this field will be updated by WO/Stock Entry
- Rearrange functions

----

**To Test:**
#### Case 1:  Production Plan containing Sales Order/MR
- Create a Sales Order for an Item
- Pull previous Sales Order into a new Production Plan. Follow usual flow, Save Production Plan
- The Pending Qty in the Items table will be equal to the (Qty - Ordered Qty) from the SO
- In our case since its a fresh SO, `pending_qty = planned_qty`
- Submit > Create Work Order against Prod Plan > Finish Work Order (complete it)
- Check **Pending Qty** in Production Plan's Items Table
- Cancel the Manufacture Stock Entry and see impact on **Pending Qty**

#### Case 2:  Production Plan containing independent rows
- Create a Production Plan
- Add an Item manually in the table. Save Production Plan
- The Pending Qty in the Items table will be equal to the Planned Qty (since no SO or MR)
- Save. Change Planned Qty and Save again. Check Pending Qty.
- Submit > Create Work Order against Prod Plan > Finish Work Order (complete it)
- Check **Pending Qty** in Production Plan's Items Table
- Cancel the Manufacture Stock Entry and see impact on **Pending Qty**

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -285,7 +285,7 @@ class ProductionPlan(Document):
 
 		self.db_set("total_produced_qty", self.total_produced_qty, update_modified=False)
 
-	def update_produced_qty(self, produced_qty, production_plan_item):
+	def update_produced_pending_qty(self, produced_qty, production_plan_item):
 		for data in self.po_items:
 			if data.name == production_plan_item:
 				data.produced_qty = produced_qty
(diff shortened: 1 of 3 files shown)


marination:
Patch test breaking due to framework
Server test failing due to unrelated `test_attendance_update_for_exclude_holidays`

marination:

PR #29706, opened 2022-02-08
fix: french translation

FHenry:
As process from [translate.frappe](https://translate.erpnext.com/) seem to not be used for french translation I try to put it here.

If there is other more relevant place, please let me know.


ankush:
The new translation contribution system is WIP.

PR #23112, opened 2020-08-21
feat:  Leave policy assignment

Anurag810:
### Leave Policy Assignment
Screenshot 2020-08-20 at 6 50 19 PM
On granting leave

image

### Grant Leaves in bulk
Screenshot 2020-08-20 at 6 50 01 PM

### Bulk Leave policy Assignment
Screenshot 2020-08-20 at 4 45 36 PM

### leave allocation Linking and validation for Leave policy assignment 
Screenshot 2020-08-20 at 6 50 39 PM

Docs: https://github.com/frappe/erpnext_documentation/pull/136

PR #29711, opened 2022-02-08
fix: Earned Leave allocation based on joining date fixes

ruchamahabal:
Earned Leave allocation based on joining date not working due to incorrect field name. This is wrong right from the time the feature was introduced in https://github.com/frappe/erpnext/pull/23112

Field name is `based_on_date_of_joining`.
image

Code was trying to access `based_on_date_of_joining_date`

- [x] Part 2 of the fix https://github.com/frappe/erpnext/pull/29650: Allocate leaves for past months for Joining Date type too, currently, it only considers for leave policy
- [x] Add Tests for this second type

#### Based on Joining Date with `based_on_date_of_joining` _disabled_ in Leave Type

- Should set _Effective From = DOJ_
- Should allocate leaves at the end of every month
- If the assignment is happening on the last day of the month, should consider that month in total leaves allocated too

#### Based on Leave Period with `based_on_date_of_joining` _disabled_ in Leave Type

- Should set _Effective From = Start date of Leave Period_
- Should allocate leaves at the end of every month
- If the assignment is happening on the last day of the month, should consider that month in total leaves allocated too

#### Based on Joining Date with `based_on_date_of_joining` enabled in Leave Type 

- Should set _Effective From = DOJ_
- Should grant leaves on the day of joining every month (eg: DOJ=12th, so allocation should happen on 12th of every month)
- If the assignment is happening on the day of joining of the month, should consider that month in total leaves allocated too

#### Based on Leave Period with `based_on_date_of_joining` enabled in Leave Type 

- Should set _Effective From = Start date of Leave Period_
- Should grant leaves on the day of joining every month (eg: DOJ=12th, so allocation should happen on 12th of every month)
- If the assignment is happening on the day of joining of the month, should consider that month in total leaves allocated too

### DIFF
--- a/erpnext/hr/doctype/leave_policy_assignment/leave_policy_assignment.py
+++ b/erpnext/hr/doctype/leave_policy_assignment/leave_policy_assignment.py
@@ -8,7 +8,7 @@ from math import ceil
 import frappe
 from frappe import _, bold
 from frappe.model.document import Document
-from frappe.utils import date_diff, flt, formatdate, get_datetime, get_last_day, getdate
+from frappe.utils import date_diff, flt, formatdate, get_last_day, getdate
 
 
 class LeavePolicyAssignment(Document):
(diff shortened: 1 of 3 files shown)


ruchamahabal:

PR #29714, opened 2022-02-08
fix: Fetch valuation rate for stock items consumed during asset repair

GangaManoj:
## Problem:

The following table is used to enter the details of Stock Items consumed during Asset Repairs. 

Screenshot 2022-02-08 at 5 44 13 PM

The Valuation Rate was being fetched from the Valuation Rate field of the Item, however, this field isn't always filled by the user. In such cases, this field will remain empty (since it's a read-only field), preventing the user from submitting the repair.

## Fix:

The Valuation Rate will now be fetched from the last Stock Ledger Entry created for that Item in that Warehouse. If there aren't any, the following alternatives will be considered:

1. Valuation Rate from the Item form
2. Standard Rate from the Item form
3. Price List Rate

<details>

1. Create a Stock Item, SI.
2. Submit a Purchase Receipt for SI (this will create a Stock Ledger Entry) with Rate = R and Warehouse = W.
3. Create an Asset Repair.
4. Check 'Stock Consumed During Repair'.
5. Enter Warehouse W in the Stock Consumption Details section.
6. Enter item SI in the Stock Items table.
7. Confirm that valuation rate R was fetched.
8. Create a second Stock Item with Valuation Rate = VR.
9. Add this item to the Asset Repair as well.
10. Confirm that VR was fetched.
11. Create a third Stock Item with Valuation Rate = 0 and Standard Selling Rate = SSR.
12. Add this item to the Asset Repair as well.
13. Confirm that SSR was fetched.

</details>


ankush:
Just use existing function that's present in stock/utils.py?

PR #29715, opened 2022-02-08
fix: Reserved for Production calculation considered closed work orders

aaronmenezes:
### Issue:
- In a Warehouse for Bins, Item quantities Reserved for Production value were incorrect. 
- For an item, there was a mismatch between the sum of item quantities used in Work orders with states 'In-Progress', 'Not Started', and  Reserved qty for Production in Bin.
- This issue arises because the calculation for quantities Reserved for Production considered work orders that were closed.

### Fix:
- The updated calculation for quantities Reserved for Production to exclude work orders that are "Stopped", "Completed", "Closed".

caused by https://github.com/frappe/erpnext/pull/28150


ankush:

ankush (dismissed):

PR #29716, opened 2022-02-08
fix: Loan repayment via Salary Slip

deepeshgarg007:
The user gets the following error message while submitting salary slip with Loan paid via salary if the payroll payable account has account type as "Payable" due to missing party types in GL Entry

image

Steps to test:
1. Create a Loan Type
2. Create a loan against an employee and mark it as repay from Salary
3. Create Loan Disbursement
4. Process loan interest using Process Loan Interest Accrual Doctype
5. Make a salary slip against that same employee, make sure the payroll payable account type is updated as "Payable"
6. Try and submit the Salary Slip


deepeshgarg007:
@Mergifyio backport develop

PR #29717, opened 2022-02-08
fix: avoid list mutation within loop

govindsmenokee:
Prevent list mutation within loop leading to incorrect data

> Screenshots/GIFs


ankush:
@mergifyio backport develop

PR #29724, opened 2022-02-09
feat: API method to created consolidated purchase invoice from receipts

deepeshgarg007:
`no-docs`

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -2,6 +2,8 @@
 # License: GNU General Public License v3. See license.txt
 
 
+import json
+
 import frappe
 from frappe import _, throw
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -926,3 +926,17 @@ def get_item_account_wise_additional_cost(purchase_document):
 						account.base_amount * item.get(based_on_field) / total_item_cost
 
 	return item_account_wise_cost
+
+# API method to created consolidated Purchase Invoice against multiple receipts
+@frappe.whitelist()
+def create_consolidated_purchase_invoice(purchase_receipts, save=0, submit=0):
+	target_doc = None
+	for purchase_receipt in purchase_receipts:
```
nextchamp-saqib, line 940: `purchase_receipts` could be a stringified JSON, you should probably `json.loads` if it is a string

PR #29729, opened 2022-02-09
fix: allow `regional_overrides` hook to be set in subsequent apps

sagarvora:
This change is necessary in preparation for separating the regional module into multiple apps.


sagarvora:
Current functionality still works as expected:
https://github.com/frappe/erpnext/runs/5124351497?check_suite_focus=true#step:12:396

nextchamp-saqib (approved):

PR #29736, opened 2022-02-10
fix: time out error while making work orders from production plan

rohitwaghchaure:
**Issue**
While making 1135 work orders (both fininshed goods and sub-assembly items) system was throwing the timeout error.

Profiling before changes
Screenshot 2022-02-10 at 10 57 40 AM

**After Changes**

Screenshot 2022-02-10 at 12 29 39 PM

Reduced time from 165 seconds to 106 seconds

Support ticket ISS-21-22-11869


rohitwaghchaure:

PR #29774, opened 2022-02-13
fix: Patch fixes

deepeshgarg007:


erpnext/patches/v14_0/update_opportunity_currency_fields.py
```diff
@@ -6,8 +6,8 @@
 
 
 def execute():
-	frappe.reload_doc('crm', 'doctype', 'opportunity')
-	frappe.reload_doc('crm', 'doctype', 'opportunity_item')
+	frappe.reload_doc('crm', 'doctype', 'opportunity', force=True)
+	frappe.reload_doc('crm', 'doctype', 'opportunity_item', force=True)
```
ankush: shouldn't be required!

We use hash-based migration now, so if the file is changed it should reload.

PR #29776, opened 2022-02-13
fix: show user id in emp group table

rtdany10:
User ID is a read only field that is fetched from the Employee document.
Users, while adding new rows, fail to acknowledge if the field is correctly fetched as it isn't visible in the table columns. 

Before:
image

After:
image


ruchamahabal (approved):

ruchamahabal:
Unrelated test case failure. Merging

ruchamahabal:

PR #29783, opened 2022-02-14
style: force auto-formatted code

### LINKED ISSUE #29762: proposal: bulk-format python code with `black` and enforce in pre-commit
Pros:
1. Consistent, no-nonsense formatting. No need to ever think about it EVER again. 
2. No need to wait for sider to check and re-push minor formatting fixes ever. (just like how flake8 now forces checks locally before pushing)

Cons:
1. Github GUI still doesn't support .git-blame-ignore but it will soon.
2. Short term pains (current open PRs will get buncha conflicts)

(leave 👍 or 👎 to vote and if you've serious objection please explain)

ankush:
closes https://github.com/frappe/erpnext/issues/29762

Summary: Add pre-commit hook for auto-formatting all python using our black fork. 

If you have a pre-commit setup locally you don't need to do anything. If not, follow contribution guideline here: https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting 

TODO:

- [x] agree internally on new style configs (99 chars max length and continue using tabs)
- [x] bump black version
- [x] add final commits in git blame ignore list (will commit directly post merge) Github now support ignoring revs from UI too: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view 

Fix known blockers:
- [x] Translate strings split to new lines by black:  https://github.com/frappe/frappe/pull/16409 

PS: won't fix sider issues, instead of disabling style issue reported by sider since they are irrelevant now. :)


ankush:
All semgrep issues are unrelated/pre-existing, they are triggered due to change of lines:

Screenshot 2022-03-26 at 6 01 55 PM



I'll wait for https://github.com/frappe/frappe/pull/16409; else this is ready for merge 😬

ankush:
After inspecting the diff, I feel 70% of the changes improve readability and 30% of them reduce it. The 30% however is because of badly laid out code that probably needs to be refactored ANYWAY. This change will at least ensure you don't get to push long unreadable over-indendent things without it looking ugly. 

Example: https://github.com/frappe/erpnext/blob/21bf187988ef6f091a7730a7c95c1a2a98a4b5b9/erpnext/stock/doctype/serial_no/serial_no.py#L277

barredterra:
@ankush can you elaborate on the decision to fork black, cutting us off from updates:

- Why didn't we switch to spaces if anything gets bulk-formatted anyways?
- If we're keeping tabs, why fork black? Instead we could, for example, add a small hook that replaces leading spaces with tabs.

ankush:
> Why didn't we switch to spaces if anything gets bulk-formatted anyways?


Explained here: https://github.com/frappe/frappe/pull/16453#issuecomment-1082971526 


>  decision to fork black, cutting us off from updates

We did it a long time ago for other projects. There's hardly a 2 line diff between fork and original. Black isn't something that needs to be updated often so eh, who cares.

barredterra:
Thanks, makes sense to keep tabs. 👍

> We did it a long time ago for other projects.

They didn't receive updates either, if I'm not mistaken. 😄  The black fork will most probably just sit there, untouched for the next 5 years.

> Black isn't something that needs to be updated often

Meanwhile, black had 5 releases in the last 6 months.

ankush:
> Black isn't something that needs to be updated often

Emphasis on this. We aren't missing _anything_.

PR #29788, opened 2022-02-14
fix: Generate Warehouse wise FIFO Queue always and later aggregate if required

marination:
## Issue:
Consider the following ledger with **different warehouses:**
<table>
 <th>No.</th>
  <th>Document</th>
  <th>Date</th>
  <th>Qty</th>
  <th>Warehouse</th>
  <tr>
    <td>1</td>
    <td>Stock Reco</td>
    <td>1st Feb</td>
    <td>1200</td>
    <td>WH 1</td>
  </tr>
  <tr>
    <td>2</td>
    <td>Stock Reco</td>
    <td>2nd Feb</td>
    <td>500</td>
    <td>WH 2</td>
  </tr>
  <tr>
    <td>3</td>
    <td>Delivery Note</td>
    <td>3rd Feb</td>
    <td>-10</td>
    <td>WH 2</td>
  </tr>
</table>

> TLDR: Total Balance (across warehouses) should be **1200 + 490 (500-10) = 1690**  but instead is **500 - 10 = 490**
- Our current calculation, **while `show_warehouse_wise_stock` is disabled**, ignores warehouse completely and maintains only item wise data
- In the case above, when reco is done for **WH 2**, `FIFOSlots` does not know/care about which warehouse this reco is for.
- **So it resets the entire previous balance**, even if part of the balance belonged to **WH 1**
Screenshot 2022-02-14 at 8 36 58 PM

## Fix:
- Maintain Queue and balances warehouse wise and later aggregate if `show_warehouse_wise_stock` is disabled
- Due to warehouse wise data, sequential stock recos can rightfully reset the entire balance qty of one warehouse only
Screenshot 2022-02-14 at 8 35 37 PM

**To test:**
- Test above scenario with and without `Show Warehouse Wise Stock` enabled in Report **Stock Ageing**
- Test Report **Stock Balance** for the same with `Show Ageing Data` enabled
- Test Report **Warehouse Wise Item Balance Age and Value**


marination:
Patch test breaking due to framework

marination:

PR #29791, opened 2022-02-14
fix: Fixes in TDS payable monthly report

deepeshgarg007:
- Cancelled entries were being considered in TDS Payable monthly report, which led to wrong amount being visible in the reports
- Supplier wise filtering fix and considers debit entries (TDS Refund)
- Ignore TDS payment entries


deepeshgarg007:

PR #29803, opened 2022-02-15
fix: Update SO via Work Order made from MR (attached to SO)

marination:
**Issue:**
> **SO -> MR -> WO**
- Create a Sales Order for an FG Item (with BOM and Raw Materials)
- Create a Material Request of type **Manufacture** from this Sales Order. Submit.
- Create a Work Order from this Material Request. Submit and Finish.
- On completion of WO, **Material Request is back updated, but Sales Order is not** (check SO items table, Work Order Qty needs to be updated)

**Fix:**
- Add SO Item reference in WO from MR (that was made from SO)
  Screenshot 2022-02-15 at 2 23 37 PM

**To test**:
- Above workflow
- Also check impact on cancellation


marination:

marination:

PR #29804, opened 2022-02-15
feat!: batch-wise valuation rates

### LINKED ISSUE #29468: Value $ difference between Stock Entry and Stock Ledger when issuing multiple batches
### Information about bug

Here are the steps to reproduce.

Create new Item-1, `Has Batch No` = Yes, `Valuation Method` = FIFO
Create a Purchase Receipt, Item-1, qty:10, batch-50, 50$
Create a Purchase Receipt, Item-1, qty:10, batch-30, 30$

Here is the Stock Ledger Entry for this item.
Stock_queue is showing 10 @ 50.00$ and 10 @ 30.00$
All good.
image

Next, create a Stock Entry - Material Issue
Add line 1 - Item-1, qty:10, batch-50
Add line 2 - Item-1, qty:10, batch-30

Now look at the rates one each item lines, both lines are showing 50$, although second line should be at 30$.
Total Outgoing value shows 1000$ even though I have only 800$ of value in stock. 
image

When we submit the Stock Entry, the Stock Ledger is showing the correct amounts.
image

Please fix the Stock Entry screen to pick the right amount when issuing multiple batches.
Thanks

### Module

stock

### Version

v13.16.0

### Installation method

manual install

_No response_

ankush:
continued from: https://github.com/frappe/erpnext/pull/26914 

NOTE: This feature is only applicable to items with FIFO/LIFO valuation. 

Problem:
- Currently batch numbers are NOT considered while consuming material i.e. outgoing rate is computed on item globally regardless of individual batch selected.

To understand the difference, consider the following incoming transactions, assume the same item *I*:
|# |Type   |Qty|Rate|Warehouse|Batch No|
|--|-------|---|----|---------|--------|
|1 |Inward |1  |100 |A        |X       |
|2 |Inward |1  |150 |A        |Y       |

If an outward entry is to be made:
|# |Type   |Qty|Warehouse|Batch No|
|--|-------|---|---------|--------|
|3 |Outward| -1|A        |Y       |

what will the item valuation be if the item uses *FIFO Valuation*, *Moving
Average Valuation*?

**Before :: FIFO: 100, MA: 125**
**After :: 150 (for FIFO, Moving average will be unchanged)**

Changes:
- Use moving average rate for each batch internally.
- Any incoming batch uses incoming rate * qty for computed value (just as before)
- While consuming a batched item: current batch value and qty is computed to derive the average rate of a particular batch and then this rate is used for consuming.

**Breaking changes:**
- All new batches created after the patch will consider batch-wise valuation rates.
- Stock Ledger entries of batched items won't have a stock queue. (no point in maintaining them, need to confirm side effect of this change on reports)

TODO:
 - [x] moving avg rates inside a batch
 - [x] get batchwise valuation / incoming rate
 - [x] negative stock handling
 - [x] tests moving average rates (single batch)
 - [x] test moving average rates (transfers)
 - [x] port tests from previous PR related to stock reco and entris
 - [x] port test related to reposting/stock entry/multi-batch
 - [x] check all get_incoming_rate calls
 - [x] check all get_valuation_rate calls
 - [x] mix data test (old/new batches) (FIFO/LIFO)

`no-docs` (users don't have to do anything for this)

#### Tests
lots of automated tests added for asserting known scenarios. The most important bits to test are as follows:

- Use Multi-batch, multi-item stock entries. Consume, transfer and manufacture items. Ensure the correct rate of consumption in ALL cases. 
- From UI make sure that correct valuation rates are displayed when batches are selected. These usually have no impact on the business logic in most cases but are important for UX. Affected documents: Stock entry (transfer/consume/issue), Delivery Note, Sales Invoice with updated stock. 
- Backdated / out of order entries should ensure correct valuation (after reposting finishes)
- Check the "stock value difference" of the individual stock transactions to assert the correct consumption rate. 
- check behaviour with negative stock. 
- Assert old behavior on old existing batches. 
- Create a new batch for items with old batches and check if valuation works as expected in both cases. 
- Assert moving average rates inside a single batch by adding additional quantity in the same batch with different rates and then consuming at an average rate.
- Go wild with any stock transaction that affects batches: purchase invoice, purchase receipt, sales invoice, delivery note, stock entry, Stock Reconciliation, Landed cost voucher, packed items (as part of bundle), POS invoice. 

Co-authored-by: Alan Tom <2.alan.tom@gmail.com>


ankush:
Note: on hold for discussing whether this should be enabled on moving average.

ankush:
(failures unrelated)

erpnext/stock/doctype/batch/batch.py
```diff
@@ -110,11 +110,15 @@ def after_delete(self):
 
 	def validate(self):
 		self.item_has_batch_enabled()
+		self.set_batchwise_valuation()
```
marination, line 113: so if anyone disables/enables an old batch or adds description to it, it will be set to 1 ?

ankush: it's "set only once" so won't change it but throw an error. 🥲 

Nice catch, will fix it separately.

ankush: Screenshot 2022-02-21 at 6 12 09 PM


can confirm locally

ankush: https://github.com/frappe/erpnext/pull/29916

TurkerTunali:
Should also Manufacture stock entry affected from batch valuation rate? But test section in this PR states that "Affected documents: Stock entry (transfer/consume/issue), ..."

ankush:
@TurkerTunali yeah, no special handling for manufacturing type... it's also affected with same logic. 

Manufacturing is just the consumption of RM and production of FG as far as stock ledger is concerned.

PR #29815, opened 2022-02-15
fix: if an item code is too long, truncate before setting BOM name

Alchez:
<details>
<summary>Traceback (while trying to save BOM)</summary>

```python
  File "apps/frappe/frappe/app.py", line 68, in application
  File "apps/frappe/frappe/api.py", line 55, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/__init__.py", line 1214, in call
    return fn(*args, **newargs)
  File "apps/frappe/frappe/desk/form/save.py", line 21, in savedocs
    doc.save()
  File "apps/frappe/frappe/model/document.py", line 287, in save
    return self._save(*args, **kwargs)
  File "apps/frappe/frappe/model/document.py", line 309, in _save
    return self.insert()
  File "apps/frappe/frappe/model/document.py", line 252, in insert
    self.db_insert()
  File "apps/frappe/frappe/model/base_document.py", line 368, in db_insert
    ), list(d.values()))
  File "apps/frappe/frappe/database/database.py", line 146, in sql
    self._cursor.execute(query, values)
  File "env/lib/python3.6/site-packages/pymysql/cursors.py", line 148, in execute
    result = self._query(query)
  File "env/lib/python3.6/site-packages/pymysql/cursors.py", line 310, in _query
    conn.query(q)
  File "env/lib/python3.6/site-packages/pymysql/connections.py", line 548, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "env/lib/python3.6/site-packages/pymysql/connections.py", line 775, in _read_query_result
    result.read()
  File "env/lib/python3.6/site-packages/pymysql/connections.py", line 1156, in read
    first_packet = self.connection._read_packet()
  File "env/lib/python3.6/site-packages/pymysql/connections.py", line 725, in _read_packet
    packet.raise_for_error()
  File "env/lib/python3.6/site-packages/pymysql/protocol.py", line 221, in raise_for_error
    err.raise_mysql_exception(self._data)
  File "env/lib/python3.6/site-packages/pymysql/err.py", line 143, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.DataError: (1406, "Data too long for column 'name' at row 1")
```

</details>


ankush:
This will allow inserting but mess up with versioning right? because for versioning item code is used to find existing BOMs. 

https://github.com/frappe/erpnext/blob/fa57a38cc1bd7462bfe3e2a6e0bc92fcf8c17a84/erpnext/manufacturing/doctype/bom/bom.py#L112-L113

Screenshot 2022-02-16 at 12 32 48 PM



(and even if it works it looks like this 🙈 )

Screenshot 2022-02-16 at 12 31 48 PM


Not sure who _really_ needs >100 character item codes TBH. A better solution could be to increase length in DB table in such odd cases.

Alchez:
@ankush 

> A better solution could be to increase length in DB table in such odd cases.

I agree with this, but I'm not sure if there's an easy way to increase the maxlength of the `name` field for BOM. And would it persist after an update?

Alchez:
@ankush I've made some changes so that versioning no longer relies on the item, and simply the current version, which is what the old code was looking for too.

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -103,25 +104,27 @@ class BOM(WebsiteGenerator):
...
 
-			# split by item
-			names = [name.split(self.item, 1) for name in names]
-			names = [d[-1][1:] for d in filter(lambda x: len(x) > 1 and x[-1], names)]
+		prefix = self.doctype
+		suffix = "%.3i" % index  # convert index to string (1 -> "001")
+		bom_name = f"{prefix}-{self.item}-{suffix}"
 
-			# split by (-) if cancelled
-			if names:
```
ankush, line 116: @Alchez new logic for versioning doesn't consider cancelled items (check last test case I added which fails)

ankush (changes requested):

erpnext/manufacturing/doctype/bom/bom.py
Alchez: @ankush, oh thanks for the test case! I've added a fix for checking cancelled BOMs. It's fairly implicit, but I'm not sure about a better way to do it.

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -134,6 +137,30 @@ def autoname(self):
...
+		# - BOM/ITEM/001
+		# - BOM/ITEM/001-1
+		# - BOM-ITEM-001
+		# - BOM-ITEM-001-1
+		valid_bom_parts = list(filter(lambda x: len(x) > 1 and x[-1], bom_parts))
+
+		# extract the current index from the BOM parts
+		if not valid_bom_parts:
+			# handle cancelled and submitted documents
+			indexes = [cint(part[2] for part in valid_bom_parts)]
```
Alchez, line 157: @ankush

Do you think there's a better way to do this? Should we even bother with this anymore and simply let the system fail with an error message?

If we do, and an item has a `-` or '/' character in its name, then this logic will fail.

I was thinking I could remove all amended documents entirely by adding it as a filter in the `get_all` above:

```python
existing_boms = frappe.get_all(
	  "BOM",
	  filters={"item": self.item, "amended_from": ["is", "not set"]},
	  pluck="name"
)
```

That should let me change this back to:

```suggestion
			indexes = [cint(part[-1] for part in valid_bom_parts)]
```

which, although still implicit, is better than the current state, I think.

ankush (approved):
LGTM.

PR #29816, opened 2022-02-15
fix: Stock Ageing Transfer Bucket logic for Repack Entry with split batch rows

marination:
## **Issue:**
Consider Repack Entry with **Same Item-Warehouse**:
> Item 1  >  -50   > Batch 1
Item 1  >  -50   > Batch 2
Item 1  >  +100 > Batch 3
> **TLDR: Above ledger caused a balance difference of 50 in stock ageing**

- Repack entries with split consumption rows (mostly due to batch) would generate inaccurate time bucket balances in FIFO queue for Stock Ageing
- For such an entry technically there's no material **coming in**. Just existing material repurposed.
- This is why we maintain `transferred_item_details` in `FIFOSlots`.
- We maintain qty that was consumed from the FIFO Queue via repack in `transferred_item_details`. Which is then re-added to the FIFO Queue and adjusted in `transferred_item_details`
- After consuming 50, 50 qty from the FIFO Queue, `transferred_item_details` looks like : `[[50, date], [50, date]]`
- Now for the **100** that is incoming we need to put it back in the FIFO Queue
- The older logic simply popped the first slot from `transferred_item_details`, thus making it : `[[50, date]]` and only adding 50 to the FIFO Queue
- The remaining 50, was left in `transferred_item_details`, which caused a difference of 50 in the generated FIFO Queue.
 
## Fix:
- Consume `transferred_item_details` just like outgoing stock is consumed in the FIFO Queue (ref: `__compute_outgoing_stock()`)
- So each bucket in `transferred_item_details` : `[[50, date], [50, date]]` is consumed given there's equivalent incoming stock
- If incoming stock is more than transfer bucket qty, excess incoming stock will be added as new slot in the FIFO Queue

## To Test:
Consider Ledger below (note voucher nos):
Screenshot 2022-02-15 at 7 44 27 PM
- Before:
  Screenshot 2022-02-15 at 7 48 39 PM

- After:
  Screenshot 2022-02-15 at 7 47 02 PM

## More Cases:
- Repack where Item 1 is consumed and **Item 2** is produced. Check Stock Ageing for **Item 2**.
- Repack (negative stock) while having 0 stock, so outgoing SLE causes negative stock, and incoming makes stock 0 again. For same Item-Warehouse
- Repack: Overconsume. 26 consumed, 1 produced (more consumed, less produced)
- Repack: Overconsume with split rows (multiple rows), and produce less in one row
- Repack: Overproduce. Consumed 28, Produced 30
- Repack: Overproduce with split rows (multiple rows), and produce more in one row

You can check the test file and simulate similar ledgers if you want to


marination:
Screenshot 2022-02-15 at 7 54 25 PM

marination:
Unrelated `test_item_search` failing

marination:

marination:

erpnext/stock/report/stock_ageing/stock_ageing.py
```diff
@@ -12,6 +12,7 @@
 from erpnext.stock.doctype.serial_no.serial_no import get_serial_nos
 
 Filters = frappe._dict
+precision = cint(frappe.db.get_single_value("System Settings", "float_precision"))
```
gavindsouza, line 15: Breaks multi-tenancy!

gavindsouza (commented):
Great job on the tests + docs btw :trollface:

PR #29819, opened 2022-02-16
test: set correct DocType

sagarvora:
- Fixes incorrect DocTypes causing failing tests [here](https://github.com/frappe/frappe/pull/15979#issuecomment-1040323674).
- Minor optimisation in `test_salary_slip`


erpnext/payroll/doctype/salary_slip/test_salary_slip.py
```diff
@@ -687,28 +687,35 @@ def make_employee_salary_slip(user, payroll_frequency, salary_structure=None):
...
+		get_salary_component_account(doc, company_list)
 
 def get_salary_component_account(sal_comp, company_list=None):
 	company = erpnext.get_default_company()
 
 	if company_list and company not in company_list:
 		company_list.append(company)
 
-	sal_comp = frappe.get_doc("Salary Component", sal_comp)
+	if isinstance(sal_comp, str):
```
sagarvora: This could alternatively check for `Document` to allow filters instead of `str`.

ankush (approved):

PR #29835, opened 2022-02-16
chore: Move patch that updates SO from WO to v13

marination:
Moving a patch between folders, nothing to see here.
Introduced via https://github.com/frappe/erpnext/pull/29803


marination:
Unrelated `test_item_search` failing

PR #29837, opened 2022-02-16
fix: made lead status field read-only

anupamvs:
**Issue-** User is not able to change the lead status
lead-status

**Fix-** Made Lead status field read-only since status is managed by StatusUpdater.

### DIFF
--- a/erpnext/crm/doctype/lead/lead.json
+++ b/erpnext/crm/doctype/lead/lead.json
@@ -471,11 +471,12 @@
  "idx": 5,
  "image_field": "image",
  "links": [],
- "modified": "2021-08-04 00:24:57.208590",
+ "modified": "2022-02-16 18:49:09.569678",
  "modified_by": "Administrator",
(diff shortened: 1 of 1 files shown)


erpnext/crm/doctype/lead/lead.json
```diff
@@ -140,6 +140,7 @@
    "oldfieldname": "status",
    "oldfieldtype": "Select",
    "options": "Lead\nOpen\nReplied\nOpportunity\nQuotation\nLost Quotation\nInterested\nConverted\nDo Not Contact",
+   "read_only": 1,
```
ruchamahabal, line 143: Status updater only takes care of these:

https://github.com/frappe/erpnext/blob/ecfa6344c02b18455637455742202c7fadf1b35d/erpnext/controllers/status_updater.py#L18-L23

What about some other status like Replied, Interested, Do no contact, etc? These need to be updated manually.

PR #29846, opened 2022-02-17
fix(Timesheet): convert time logs to datetime while checking for overlap

ruchamahabal:
## Steps to replicate

1. Create a timesheet with an employee linked.
2. Set From Time for one log as 10:00. Set hours to 1
3. Set From Time for another log as 11:00. Set hours to 1
4. These time logs are actually non-overlapping: 10-11 and 11-12. So overlap validation shouldn't be thrown but it will be thrown.

overlap

## Fix

After https://github.com/frappe/erpnext/pull/28589 `to_time` values in time logs are set on the server-side. The `set_to_time` function also converts value set on client-side:

image

While checking for overlap, the function compares strings. So this happens:

image

Convert values to datetime before comparing for overlap.

## Additional Changes

The following fields from the Timesheet Detail table are not being used anywhere and they are only visible when the parent is Work Order. But Work Order doctype doesn't have any child table for timesheet detail. Hence removing these fields and the code related to it (dead code)

`completed_qty`
`workstation`
`operation`
`operation_id`

### DIFF
--- a/erpnext/projects/doctype/timesheet/test_timesheet.py
+++ b/erpnext/projects/doctype/timesheet/test_timesheet.py
@@ -151,6 +151,35 @@ class TestTimesheet(unittest.TestCase):
 		settings.ignore_employee_time_overlap = initial_setting
 		settings.save()
 
+	def test_timesheet_not_overlapping_with_continuous_timelogs(self):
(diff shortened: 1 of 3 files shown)


ruchamahabal:

PR #29869, opened 2022-02-17
fix: GSTIN filter for GSTR-1 report

deepeshgarg007:
Added company GSTIN filters in GSTR-1 Report. Currently, only an address filter is available in the GSTR-1 report.
Users who have multiple addresses in the same state cannot view the report for a single GSTIN as the GSTIN is issued state-wise. This filter will help users to do that.

image


deepeshgarg007:

PR #29870, opened 2022-02-17
refactor: subscriptions

hrwX:
- Reduced complexity associated with invoice generation function, got rid of `is_postpaid_to_invoice` and `is_prepaid_to_invoice`.
```
def can_generate_new_invoice(self) -> bool:
    if self.cancelation_date:
	    return False
    elif self.generate_invoice_at_period_start and (
	    getdate(frappe.flags.current_date) == getdate(self.current_invoice_start)
	    or self.is_new_subscription()
    ):
	    return True
    elif getdate(frappe.flags.current_date) == getdate(self.current_invoice_end):
	    if self.has_outstanding_invoice() and not self.generate_new_invoices_past_due_date:
		    return False
    
	    return True
    else:
	    return False
```
- The invoices won't be generated randomly, they will only be generated either at the beginning of the period or the end of the period depending on the configuration, eliminating the randomness of earlier implementation.
- Removed redundant child table for saving invoices. Added a link field in Sales and Purchase Invoice


erpnext/accounts/doctype/subscription/subscription.json
```diff
@@ -263,11 +250,23 @@
...
+   "group": "Purchase",
+   "link_doctype": "Purchase Invoice",
+   "link_fieldname": "subscription"
+  },
+  {
+   "group": "Sales",
+   "link_doctype": "Sales Invoice",
+   "link_fieldname": "subscription"
+  }
+ ],
```
barredterra:
```suggestion
 "links": [
  {
   "group": "Buying",
   "link_doctype": "Purchase Invoice",
   "link_fieldname": "subscription"
  },
  {
   "group": "Selling",
   "link_doctype": "Sales Invoice",
   "link_fieldname": "subscription"
  }
 ],
```

erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -2,36 +2,48 @@
...
+	pass
+
+
 class Subscription(Document):
 	def before_insert(self):
 		# update start just before the subscription doc is created
 		self.update_subscription_period(self.start_date)
 
-	def update_subscription_period(self, date=None, return_date=False):
+	def update_subscription_period(
```
barredterra: Instead of the `return_date` flag, we could extract a method `get_subscription_period()` that can be called independently.

erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -101,286 +124,276 @@ def get_current_invoice_end(self, date=None):
...
+				_("Billing Interval in Subscription Plan must be Month to follow calendar months")
+			)
+
+	def after_insert(self) -> None:
 		# todo: deal with users who collect prepayments. Maybe a new Subscription Invoice doctype?
 		self.set_subscription_status()
 
-	def generate_invoice(self, prorate=0):
+	def generate_invoice(
+		self, from_date: Optional[bool] = None, to_date: Optional[bool] = None
```
barredterra:
```suggestion
		self, from_date: Optional[date] = None, to_date: Optional[date] = None
```

erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -101,286 +124,276 @@ def get_current_invoice_end(self, date=None):
...
-		self.save()
-
-		return invoice
+		Backwards compatibility
+		"""
+		return self.create_invoice(from_date=from_date, to_date=to_date)
 
-	def create_invoice(self, prorate):
+	def create_invoice(
+		self, from_date: Optional[bool] = None, to_date: Optional[bool] = None
```
barredterra:
```suggestion
		self, from_date: Optional[date] = None, to_date: Optional[date] = None
```

sam-anam1:
Hi all.

please in need help regarding this matter [post it ](https://discuss.erpnext.com/t/cancelling-subscriptions/87170)



Would like to use subscriptions with web shared clients
but the problem is when we cancelling subscriptions it will generate invoice
as the documentation

> If you are cancelling an active subscription, an invoice will immediately be generated. 
The generated invoice will be on pro-rata basis by default. If you want ERPNext always create an invoice for the full amount, uncheck the ‘Prorate’ field in Subscription Settings.
[Cancelling Subscriptions](https://docs.erpnext.com/docs/v12/user/manual/en/accounts/articles/how-to-manage-subscriptions-with-erpnext#cancelling-subscriptions)

it would be great has an option for that "no creating invoices when Cancelling & Restarting".
It dose make sense to me that not even make invoice when “uncheck Generate Invoice At Beginning Of Period” maybe.

However, the reason here as mention as example shared hosting plans
when customer cancel subscription due ICANN or even breaking rules such copyright or requesting for cancellation to changing domain extension etc… there is no way for refund or pro-rata basis… or full amount… so we wouldn’t wanna create an invoice or such.

Is there any side script or anyway to

Disable generating SI on Manually Cancelling and Restarting Subscription?
Disable Subscription Processing In The Background
or even other solution if it could change the manually status
We just wouldn’t go with custom code or doctype to avoid ERPNext updates

Many thanks!!

deepeshgarg007:
@hrwX is this complete? Can you update the description and resolve the conflicts?

federicocalvo:
Hi, I don't know if it should be taken out as I don't see the detail of what the refactor looks like.

Something that conceptually should be possible to do, is to choose the payment gateway in the subscription, and not in the plan.

It is good that the option of the gateway is in the plan, but if I have 5 gateways, it forces me to make 5 plans for a product, since it is the user who chooses the payment method.

I just wanted to know if you raised it or should I create an issue outside.

hrwX:
@sam-anam1 @federicocalvo This PR is a refactor for the existing functionality, I won't be adding any new functionality in this PR.
It would be great if you create a Github issue and tag it as improvement.

hrwX:
> improvement

I am in the process of doing it, will ping you once the PR is ready

federicocalvo:
@hrwX done! #30571

nameduser0:
@hrwX does this PR fix the issue someone told me about (I haven't checked myself) that if you create a subscription on 2022-03-15 the next invoice will be generated on 2022-05-01 and misses out 2022-04-01?

Is `Generate Invoice At Beginning Of Period` the difference between billing in advance and billing in arrears?
It's also not clear if this is the same thing as `Follow Calendar Months`?

Also the documentation mentions `Generate Invoice Days Early` which doesn't appear to exist.

Also, is this going into v13 too?

sam-anam1:
> @sam-anam1 @federicocalvo This PR is a refactor for the existing functionality, I won't be adding any new functionality in this PR.
> It would be great if you create a Github issue and tag it as improvement.

@hrwX here it is [30640](https://github.com/frappe/erpnext/issues/30640)

deepeshgarg007:
@hrwX is this ready for review? Please reopen once you start working on this again

hrwX:
I effed up this branch, reopening a new one here #30963

PR #29872, opened 2022-02-18
fix: JobCard TimeLog to_date

s-aga-r:
### Issue(s): Job Card Item

<details>
<summary>Steps to Reproduce:</summary>

1. Check your Time Zone in **System Settings**.

image

2. Change the Time Zone for the User.

image

_Note: User Time Zone must be ahead of System Settings Time Zone. In the current case, Asia/Tokyo is 3 hours and 30 minutes ahead of Asia/Kolkata._

3. Now create a **Work Order** with **Operations**.

image

4. Create a **Job Card** and try to start the Job.

image

</details>

_Problem:_

- We are getting from_time from [here](https://github.com/frappe/erpnext/blob/c36bd7e1a6fe48c5fff4765e843571a0d6560dd1/erpnext/manufacturing/doctype/job_card/job_card.js#L252). now_datetime() will return current DateTime (User Time Zone).
- to_date will be None since we are starting the job for the first time.
- [Here](https://github.com/frappe/erpnext/blob/c36bd7e1a6fe48c5fff4765e843571a0d6560dd1/erpnext/manufacturing/doctype/job_card/job_card.py#L65) we are comparing from_time with to_time. get_datetime() will return current DateTime (System Settings Time Zone) if datetime_str is None (As in our case).
- In the case where the User's TimeZone is ahead of the System Settings TimeZone will create this type of problem.

_Proposed Solution:_

- Skip comparison if to_time is None.


ankush (approved):

ankush:
(some more deliberations required in TZ-related problems but that's unrelated to this PR hence merged)

PR #29874, opened 2022-02-18
refactor: trigger generate schedule when any change made in items table

### LINKED ISSUE #29564: Serial Number not auto - updated on save in Maintenance Schedule
### Information about bug

Issue : Whenever a serial number is updated in Items' child table , the same doesn't get updated in the schedule child table.

Steps to reproduce : 

1. Create a Maintenance Schedule and select serialised item and enter serial number for that item say SE0021
2. Now the Schedule would be generated automatically. Hit save and check serial number in schedule child table, it is correct i.e SE0021
3. Now change the serial number (say SE0022)  in the items child table and click save. It still shows previous serial number i.e SE0021.

WORKAROUND : Click on Generate Schedule to update the serial number in the schedule child table 

Screen Recording : 

Serial Number In Maintenance

### Module

support

### Version

V13.20.0

### Installation method

_No response_

_No response_

noahjacob:
### Issue
- When changing any field in the items table, the user would need to click on generate schedule to get the new schedule with the appropriate changes.

maint_gen_bug

### Changes
- Generate schedule now automatically triggers when making changes in fields like serial_no, sales_person, etc.

maint_gen_fix

Closes - [#29564](https://github.com/frappe/erpnext/issues/29564)


erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.js
```diff
@@ -152,12 +152,27 @@ erpnext.maintenance.MaintenanceSchedule = class MaintenanceSchedule extends frap
...
+		}
+	}
+
+	sales_person(doc) {
+		let me = this;
+		if (doc.docstatus == 0) {
+			me.frm.call('generate_schedule');
+		}
+	}
+
```
marination: instead of calling `generate_schedule`  conditionally from JS and Py, call it purely from server side ?
Create one function on the py side that checks if generate schedule needs to be called and calls it on `validate`. The JS file needs a clean up, lets not add more to it maybe.

Probably first map out under what circumstances you want to regenerate/generate schedule in `validate` method

noahjacob: Basically I need to call it whenever there's a change in the items table, how do I do that from server side. I can handle everything except the date change.

marination: @noahjacob it should broadly be regenerated on save if the items table has changed or different serial_no/sales_person, etc.(you'd know what other factors are there)
- check items table in `doc_before_save` and compare with current doc object, if there are changes (in these certain fields in the items table), regenerate the schedule.
- This will happen on every **Save**
- Point is this should be API friendly too, just making changes via JS doesn't help.

erpnext/maintenance/doctype/maintenance_schedule/maintenance_schedule.py
```diff
@@ -197,6 +197,23 @@ def validate_sales_order(self):
...
+			schedule_generated = False
+			for field in fields:
+				b_doc = prev_item.as_dict()
+				doc = item.as_dict()
+				if cstr(b_doc[field]) != cstr(doc[field]):
+					schedule_generated = True
+					self.generate_schedule()
+					break
+			if schedule_generated:
+				break
```
ankush: any reason to track this flag? Why not just return from the whole function call? 

Also `validate_items_table_change` shouldn't be generating schedule, the name is misleading. Either it should return bool indicating change in items, or name should be changed 😅

ankush (changes requested):

PR #29879, opened 2022-02-18
fix: Validate party account with company

deepeshgarg007:
Users were able to add incorrect party account in Customer/Supplier masters and there was no validation to prevent this and this caused issues in transactions later. Added a validation to prevent this. The account added should belong to the company added in the row and not to some other company 

Screenshot 2022-02-23 at 11 19 50 AM


deepeshgarg007:

PR #29884, opened 2022-02-19
refactor!: remove India specific code

sagarvora:
### Currently kept as-is

- HR Related Code (to be moved to HRMS app)
  - Salary Slip Overrides
  - Custom Fields
  - Salary Components
  - Default Gratuity Rule
  - Following Additional Patches:
    - `setup_gratuity_rule_for_india_and_uae`
    - `move_tax_slabs_from_payroll_period_to_income_tax_slab`
    - `check_is_income_tax_component`
  - Following Reports:
    - Income Tax Deductions
    - Professional Tax Deductions
    - Provident Fund Deductions
 
- Lower Deduction Certificate and related code (to be ported later)

### Notes for reviewer
- Certain patches have been kept back with code that applied to countries outside India
  -  `removed_roles_from_gst_report_non_indian_account`
  - `create_gst_payment_entry_fields`
  
- Changes in `asset_value_adjustment.py` were not ported because they are redundant. The function `get_depreciation_amount` has been already patched using regional overrides and ported to India Compliance.
- A test in purchase invoice was changed to create fixtures that have been moved to India Compliance.


ankush:
Reopen when it's time?

erpnext/setup/setup_wizard/data/country_wise_tax.json
```diff
@@ -1163,481 +1163,6 @@
 		}
 	},
 
-	"India": {
```
deepeshgarg007, line 1166: How is this created in the new app? Some users might not have the India compliance app before setup, if they choose India in the setup wizard what COA will the first company have?

sagarvora: We added a event using hooks in the Company doctype's `on_update` method. It checks if `frappe.flags.country_change` is set (from ERPNext) and installs these defaults based on that.

sagarvora: In case of installing the app after creating the company, user can click a button in their Company doctype to trigger this manually.

sagarvora: However, I can also build a way to detect if the defaults are set for existing companies when India Compliance is installed and then if they are not set, I can create them on install.

vorasmit:
- [x] Patch to remove masters with warning

erpnext/setup/setup_wizard/data/country_wise_tax.json
sagarvora: solved in https://github.com/resilient-tech/india-compliance/pull/152

PR #29887, opened 2022-02-19
fix: Ledger entries on LIA for term loans

deepeshgarg007:
Ledger entries for payable principal amount should not be passed while booking loan Interest Accruals for term loans.

This leads to an incorrect income being booked in the Interest Income Account and an Incorrect amount in Loan As well which is not set off during the repayment


deepeshgarg007:

deepeshgarg007:

PR #29895, opened 2022-02-21
fix: cashflow mapper form view not loading

ankush:
`Cash Flow Mapping Template Details` was not marked as child doctype so it doesn't have parent fields and hence form load fails. 

PS: only affects new installs of v14. 

fix:
- convert the doctype to child table
- add a test to "load" all doctypes to avoid this in future.


erpnext/accounts/doctype/cash_flow_mapping_template_details/cash_flow_mapping_template_details.json
```diff
@@ -1,94 +1,33 @@
...
- "beta": 0, 
- "creation": "2018-02-08 10:18:48.513608", 
- "custom": 0, 
- "docstatus": 0, 
- "doctype": "DocType", 
- "document_type": "", 
- "editable_grid": 1, 
- "engine": "InnoDB", 
+ "actions": [],
+ "creation": "2022-02-11 11:25:05.336846",
```
gavindsouza:
```suggestion
 "creation": "2018-02-08 10:18:48.513608",
```

We should get this Frappe bug fixed 🥲

erpnext/tests/test_zform_loads.py
```diff
@@ -0,0 +1,29 @@
...
+
+import frappe
+from frappe.desk.form.load import getdoc
+
+
+class TestFormLoads(unittest.TestCase):
+
+	def test_load(self):
+		doctypes = frappe.get_all("DocType", {"istable": 0, "issingle": 0, "is_virtual": 0}, pluck="name")
+
```
gavindsouza:
```suggestion
		erpnext_modules = frappe.get_all("Module Def", filters={"app_name": "erpnext"}, pluck="name")
		doctypes = frappe.get_all("DocType", {"istable": 0, "issingle": 0, "is_virtual": 0, "module": ("in", erpnext_modules)}, pluck="name")

```

gavindsouza (changes requested):

PR #29913, opened 2022-02-21
fix: Block merging items if both have product bundles

marination:
**Issue:**
- Create two Items (non stock) and create Product Bundles for each
- Now merge one of these into the other
- The items are merged, and the Product Bundle of the older item stays as is, and its **Parent Item** changes to the new item
- Now when this new item code (second item) is used in transactions, it pulls double rows(from both bundles) in the packing items table
- Both the bundles should not exist

**Fix:**
- Make user delete Product Bundle of first Item (that is being absorbed by the merge), so that only one bundle exists
   Screenshot 2022-02-21 at 4 13 24 PM


erpnext/stock/doctype/item/item.py
```diff
@@ -462,6 +463,18 @@ def validate_properties_before_merge(self, new_name):
 			msg += ": \n" + ", ".join([self.meta.get_label(fld) for fld in field_list])
 			frappe.throw(msg, title=_("Cannot Merge"), exc=DataValidationError)
 
+	def validate_duplicate_product_bundles_before_merge(self, old_name, new_name):
+		"Block merge if both old and new items have product bundles."
+		bundle = frappe.get_value("Product Bundle",filters={"new_item_code": old_name})
+		if bundle:
```
marination: check if new item has a bundle too

PR #29921, opened 2022-02-21
fix: boarding task dates not set when activity begins on is set to 0

ruchamahabal:
## Steps to Replicate:

1. Create an Employee Onboarding with some task starting on day 0 (first day).
2. Submit. The task created for the project on submission will not have start and end dates set.

day-9

start

## Fix:

Check if `begins_on` and `duration` is not None since 0 evaluates to False and dates are not set

dates


ruchamahabal:
Unrelated test `test_execute_all_stock_reports` failing.

ruchamahabal:
This feature does not exist on v13. Backporting not needed.

PR #29922, opened 2022-02-21
fix(LMS): program enrollment does not give any feedback

ruchamahabal:
Before:

1. Create a Program
2. Enable Learning Management System in Education Settings
3. Sign up as a new user on [site_name]/lms
4. On the program page, click on enroll. The dialog does not show any message even though program enrollment gets created in the background
image

After:

image


ruchamahabal:

PR #29928, opened 2022-02-22
fix: Account filter in PSOA

deepeshgarg007:
https://docs.erpnext.com/docs/v13/user/manual/en/accounts/process-statement-of-accounts

Recently in General Ledger, the Accounts filter was converted into a multi-select filter so that the users can select and filter multiple accounts altogether. This change led to an error in the Process Statement of Account. Records having an account filter set were not able to generate or download the statement

Screenshot 2022-02-25 at 4 52 56 PM


deepeshgarg007:

PR #29940, opened 2022-02-22
fix: Item discounts for quotation and other docs

deepeshgarg007:
### Why?
For instance, consider an instance where the user duplicates a Quotation on which itemized discount is applied. The rate in the duplicated quotation is not as per the discount in the new quotation. It won't be since many triggers are not called when a doc is duplicated

### Solution
The discount still won't be applied just on duplicate, but it should be applied on saving the quotation based on price list rate
Discount on the server-side is not applied if the item rate is set even if the discount percentage or discount amount is set.
The discount should be applied on price list rate if the discount rate or amount is mentioned irrespective of the rate


deepeshgarg007:

PR #29592, opened 2022-02-02
fix: Program Enrollment tests

ruchamahabal:

PR #29946, opened 2022-02-22
[29759] Here is the fix for the issue . An SQL , which returns a tupl…

veenapromantia:
…e was assigned to a date variable.

> Screenshots/GIFs

### DIFF
--- a/erpnext/education/doctype/program_enrollment/program_enrollment.py
+++ b/erpnext/education/doctype/program_enrollment/program_enrollment.py
@@ -61,15 +61,9 @@ class ProgramEnrollment(Document):
 			frappe.throw(_("Student is already enrolled."))
 
 	def update_student_joining_date(self):
-		table = frappe.qb.DocType('Program Enrollment')
-		date = (
-			frappe.qb.from_(table)
-				.select(Min(table.enrollment_date).as_('enrollment_date'))
-				.where(table.student == self.student)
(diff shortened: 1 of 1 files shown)


erpnext/education/doctype/program_enrollment/program_enrollment.py
```diff
@@ -61,15 +61,9 @@ def validate_duplication(self):
...
-		date = (
-			frappe.qb.from_(table)
-				.select(Min(table.enrollment_date).as_('enrollment_date'))
-				.where(table.student == self.student)
-		).run(as_dict=True)
-
-		if date:
-			frappe.db.set_value("Student", self.student, "joining_date", date[0].enrollment_date)
+		date = frappe.db.sql("select min(enrollment_date) from `tabProgram Enrollment` where student= %s", self.student)
+		((date,),) = frappe.db.sql("select min(enrollment_date) from `tabProgram Enrollment` where student= %s", self.student)
```
ankush, line 65: - l64 and l65 queries are same but repeated? 
- using raw SQL where it was already using QB is not okay. We are trying to make ERPNext db agnostic. 


Also, the code seems to do exact same thing. Instead of the dictionary result, it's destructuring the tuple.

ankush:
this is already fixed: https://github.com/frappe/erpnext/pull/29592

PR #29955, opened 2022-02-23
feat: Scheduling Multiple shifts and Auto Attendance

### LINKED ISSUE #29210: Error on creating a non - overlapping shift assignment for same day for same employee
### Information about bug

Issue : When creating a non - overlapping shift assignment for same day for same employee the error still appears :

Steps to reproduce :

1. Create a shift assignment (say morning shift from 8 - 4) for an employee for some date ( keep start date and end date same)
2. Create another shift assignment ( say evening shift from 5- 9 ) for the same employee for the same date

Screenshot : 

image

### Module

HR

### Version

13.18.0

Reference : 

https://frappe.io/app/pre-release-test/PRT-14-12-21-328290

### Installation method

_No response_

_No response_

ruchamahabal:
## Problem

- Earlier, only 1 Shift Type could be assigned to an employee for one day. 
- However, in some organizations, employees could be working in multiple shifts (Morning Shift, Evening Shift) on the same day.

### Schedule multiple shifts

- Users can make multiple non-overlapping Shift Assignments for the same employee on the same day

### Shift Request

- Users can make multiple non-overlapping Shift Requests for the same employee on the same day

### Fetching shift in Employee Check In

- When employee check-ins are logged via biometric or manually, the exact shift based on the timestamp is fetched. Before, shifts used to be fetched based on date. Now they are timestamp specific.
- Handles shifts spanning over 2 different days (eg: 23:00 - 2:00), midnight shifts.
- Shift fetching used to fail on timing boundaries, this is fixed.
- Handles 2 consecutive shifts if they are overlapping within grace periods.

### Attendance

- Earlier, only 1 attendance record could be marked for 1 day.
- Now multiple attendance records can be marked for each shift. Eg: employee can be absent for one shift, present for another on the same day
- Auto Attendance now, takes care of marking shift specific attendance and not for the whole day
- Earlier, employees with no attendance records on a specific date used to be marked absent via Auto Attendance. Now, this also checks for shift-wise attendance records along with the date. So, only if attendance is not marked for a particular shift/for the whole day, the employee will be marked absent.

### Attendance Reports

- Added shift-wise attendance against employees
- Added legend for status abbr with full form and color.
- Formatter for data to quickly identify cells with lesser cognitive load.
- Refactored Monthly Attendance Sheet code. Can still be better but will leave it for the next iteration.

**Before**:
image

**After**:
image

### Impact on Payroll

If there are multiple attendance records marked for a single day with separate shifts, each record will be considered as 1 Payment Day in payroll. So if an employee works 2 shifts in a single day this contributes to 2 payment days in Payroll. No code level changes are done for this

Documentation: Added docs for Monthly Attendance Sheet only since others are mostly backend changes: https://docs.erpnext.com/docs/v14/user/manual/en/human-resources/human-resources-reports/edit?wiki_page_patch=ea798a28e2
Also closes https://github.com/frappe/erpnext/issues/29210

### ToDo

- [x] Allow scheduling multiple shifts
- [x] Shift fetching in employee check-ins
- [x] Multiple attendance records for the same day
- [x] Auto Attendance for multiple shifts
- [x] Handle impact on attendance reports
- [x] Refactor Shift Request Overlapping validation
- [x] Test cases
- [x] Documentation


erpnext/hr/doctype/shift_type/shift_type.py
```diff
@@ -66,70 +72,66 @@ def mark_absent_for_dates_with_no_attendance(self, employee):
...
 		doc = frappe.get_doc('Shift Type', shift[0])
 		doc.process_auto_attendance()
 
-def get_filtered_date_list(employee, start_date, end_date, filter_attendance=True, holiday_list=None):
-	"""Returns a list of dates after removing the dates with attendance and holidays
-	"""
-	base_dates_query = """select adddate(%(start_date)s, t2.i*100 + t1.i*10 + t0.i) selected_date from
-		(select 0 i union select 1 union select 2 union select 3 union select 4 union select 5 union select 6 union select 7 union select 8 union select 9) t0,
-		(select 0 i union select 1 union select 2 union select 3 union select 4 union select 5 union select 6 union select 7 union select 8 union select 9) t1,
-		(select 0 i union select 1 union select 2 union select 3 union select 4 union select 5 union select 6 union select 7 union select 8 union select 9) t2"""
```
ankush, line 174: image

naquiuddin:
> Most companies use payment days as a measure in salary components to calculate payroll amounts. Current payment days calculation is very straightforward. If an attendance record is found, include that day in payment days. Now that there are multiple attendance records created for a single day.
> 
> * This either needs conversion to working hours (will break backward compatibility? Users might already have `payment_days` in the component formulae)
> * or Normalization (to evaluate the payment day between 0 and 1 based on all the shifts) - very complex - can go wrong.
> * or something else?

Rather than converting calculation based on payment_days or normalisation, can we provide a Select Dropdown in salary component called *Depends On* and show these options -  _No dependency_, _Shift Payment Days_, Total_Shift_Payment_Days, _Total Working Hours_, _Total Working Minutes_, Shift_Working_Hours, Shift_Working_Minutes ?

Using this setting the calculation method/function can be selected in Salary slip code as well.

**Total Payment Days**
- In earning component this would be Total salary for all working days regardless of how many shifts a user works in a day
- In deduction component this would be Total Salary that has to be deducted for all LWPs.
- This component would be useful if employee has fixed salary amount per month, regardless of how many shifts they work

**Shift Payment Days** 
- In earning component is similar to Payment Days, when employee works for only one shift. for eg: 9 to 5. However when more than one shift is there the salary would be sum of multiple shifts.
- In Deduction component this will be salary deducted for LWPs, in multiple shifts, Sum of Salary of LWPs per shift.
- This will be useful when different shifts have different salary amount.

**Shift Working Hours or Minutes** 
- In earning component it would be salary of working hours or minutes per shift.
- In deduction component it would be salary of missed hours or minutes per shift
- This will be useful when different shifts have different Hourly Rate or Minute Rate. Hourly/Minute Rate is not fixed on employee but rather depends on Shift.

**Total Shift Working Hours or minutes**
- In earning component it would be Hourly Rate * (sum of all shift working hours or minutes)
- In deduction component it would be Hourly Rate * (sum of all shift working hours or minutes)
- This will be useful when all shifts have same Hourly rate or Hourly Rate is fixed on Employee.


P.S. **Working Minutes** is something we use in our organisation. Similar is used in middle east areas. Because based on my knowledge in Islamic Shariah employee salary has to be calculated based on minutes and not hours.

naquiuddin:
Additionally we have to edit Leave Application Form. The From Date and To Date fields has to change to the type DateTime. so that we can calculate which shift hours user was on leave.
For example if user takes leave from 9 to 10 but his shift lies between 9 to 12 then in that case 9 to 10 has to be accounted as leave or break time during the shift.

ruchamahabal:
Unrelated test failures

nabinhait (approved):

ruchamahabal:
Reverted https://github.com/frappe/erpnext/pull/30587 and re-merging PR https://github.com/frappe/erpnext/pull/30588 without squashing since this has a high blast radius and needs a good commit history. 🙈

PR #29962, opened 2022-02-23
fix: Exotel Call Log fixes

nemesis189:
- When creating a call the status was always set as ringing even for missed calls, fixed to show the correct status:
   **No Answer** if no agent picks up the call
   **Canceled** if the customer hangs up before any agent picks up the call 

- Added field to display name of employee to whom the call is directed towards


erpnext/erpnext_integrations/exotel_integration.py
```diff
@@ -33,10 +33,25 @@ def handle_end_call(**kwargs):
 
 @frappe.whitelist(allow_guest=True)
 def handle_missed_call(**kwargs):
-	update_call_log(kwargs, 'Missed')
+	status = ""
+	CallType = kwargs.get("CallType")
```
surajshetty3416:
```suggestion
	call_type = kwargs.get("CallType")
```
according to our naming convention, we use CapitalCase for class names or for query builder instances and snake case for variables

erpnext/public/js/call_popup/call_popup.js
```diff
@@ -140,6 +140,15 @@ class CallPopup {
 			}, {
 				'fieldtype': 'Section Break',
 				'hide_border': 1,
+			}, {
+				'fieldname': 'type_of_call',
+				'label': 'Type Of Call',
+				'fieldtype': 'Select',
+				'options': '\nFrappe Cloud Queries/Plan\nEnterprise Plans\nPartnership\nSupport\nBilling/Renewal\nOpen source / Junk',
```
surajshetty3416: 👀

"Frappe Cloud Queries" or other options here are not really relevant for all telephony users... better to have a link field instead.

erpnext/public/js/call_popup/call_popup.js
```diff
@@ -140,6 +140,15 @@ class CallPopup {
 			}, {
 				'fieldtype': 'Section Break',
 				'hide_border': 1,
+			}, {
+				'fieldname': 'type_of_call',
+				'label': 'Type Of Call',
```
surajshetty3416:
```suggestion
				'label': 'Call Type',
```
simpler?

erpnext/telephony/doctype/call_log/call_log.json
```diff
@@ -134,15 +138,34 @@
    "fieldname": "call_details_section",
    "fieldtype": "Section Break",
    "label": "Call Details"
+  },
+  {
+   "depends_on": "to",
+   "fieldname": "employee_call_directed_to",
+   "fieldtype": "Data",
+   "label": "Employee Call Directed To",
```
surajshetty3416:
```suggestion
   "label": "Call Received By",
```
?

erpnext/telephony/doctype/call_log/call_log.py
```diff
@@ -33,6 +33,16 @@ def before_insert(self):
 		if lead:
 			self.add_link(link_type='Lead', link_name=lead)
 
+		# Add Employee Name
+		if self.is_incoming_call():
+			# Taking the last 10 digits of the number
+			emp_number_reversed = (self.get("to"))[-1:-11:-1]
+			emp_number = emp_number_reversed[-1::-1]
+
+			employee = frappe.get_all("Employee", filters={"cell_number":["like","%"+emp_number+"%"]}, fields=["first_name", "middle_name", "last_name", "user_id"])
```
surajshetty3416:
```suggestion
			employee = frappe.get_all("Employee", filters={
				"cell_number": ["like", "%"+emp_number+"%"]
			}, fields=["first_name", "middle_name", "last_name", "user_id"])
```

surajshetty3416 (changes requested):

erpnext/telephony/doctype/call_log/call_log.json
nemesis189: call may not always be 'received'(picked up/ attended) by the agent so i felt 'directed to' would make more sense, anyway I'll change

erpnext/telephony/doctype/call_log/call_log.py
```diff
@@ -33,6 +33,18 @@ def before_insert(self):
 		if lead:
 			self.add_link(link_type='Lead', link_name=lead)
 
+		# Add Employee Name
+		if self.is_incoming_call():
+			# Taking the last 10 digits of the number
+			emp_number_reversed = (self.get("to"))[-1:-11:-1]
+			emp_number = emp_number_reversed[-1::-1]
```
surajshetty3416: use `strip_number` instead? Also, we avoid short-forms for variable names... let's keep it `employee_number`.

erpnext/telephony/doctype/call_log/call_log.py
```diff
@@ -33,6 +33,18 @@ def before_insert(self):
...
+			employee = frappe.get_all("Employee", filters={
+				"cell_number": ["like", "%"+emp_number+"%"]
+			}, fields=["first_name", "middle_name", "last_name", "user_id"])
```
surajshetty3416: you can extend `get_employees_with_number` method to make this work.

erpnext/telephony/doctype/call_log/call_log.py
```diff
@@ -33,6 +33,18 @@ def before_insert(self):
...
+
+			self.call_received_by = get_employee_name(employee[0])
```
surajshetty3416: Let's keep `call_rececived_by` a link field. We can also, enable the [title for link](https://frappeframework.com/docs/v13/user/en/basics/doctypes/form_&_view_settings#title-field) for Employee doctype.

surajshetty3416 (changes requested):

surajshetty3416:
@nemesis189 let's also introduce basic tests for this integration.

erpnext/telephony/doctype/call_log/call_log.py
nemesis189: Done

erpnext/telephony/doctype/call_log/call_log.py
nemesis189: tweaked and used the method as suggested

surajshetty3416 (dismissed):

surajshetty3416:

PR #29966, opened 2022-02-23
fix(lead): reload contact before updating links

barredterra:
Contact might have changed since it was created. In this case the user saw an error message ("contact has been modified") and was unable to save the lead.

The error only shows when you have hooks on Contact that modify Contact.


ankush:
> The error only shows when you have hooks on Contact that modify Contact.


Can you describe this in a bit more detail? or share what your hook is doing?  

Hooks shouldn't be triggering `.save` which is the only way to cause "has been modified" error. 😅 


Use `db_set` instead if calling from after-save hooks.

barredterra:
@ankush **Lead** inserts a **Contact** at one point in time (and stores a reference to the object). This causes other hooks on **Contact** to fire, which might modify the contact on the database level.
Later, **Lead** modifies the same **Contact** object from before (without reloading) and tries to save it again. If meanwhile the database state has changed, this produces an error.

We can avoid the error by reloading the contact before updating it.

### Produces Error
```python
class Lead(Document):
    def before_insert(self):
        self.contact_doc = frappe.get_doc({
            "first_name": "Original",
            # ...
        }).insert()

    def after_insert(self):
        self.contact_doc.first_name = "I'm Second"
        self.save()

class Contact(Document):
    def on_update(self):
        frappe.db.set_value(self.doctype, self.name, "fist_name", "I'm First")
```

### Works 
```python
class Lead(Document):
    def before_insert(self):
        self.contact_doc = frappe.get_doc({
            "first_name": "Original",
            # ...
        }).insert()

    def after_insert(self):
        self.contact_doc.reload()
        self.contact_doc.first_name = "I'm Second"
        self.save()

class Contact(Document):
    def on_update(self):
        frappe.db.set_value(self.doctype, self.name, "fist_name", "I'm First")
```

ankush:
@barredterra thanks for the explanation. This is very conflicting indeed. 

`.reload()` will ignore any other changes done in `lead.py` (while there aren't any right now, there could be in the future. 😐 Not sure what's the best/safe solution here. 


The simplest way to avoid this would be not to set `update_modified=False` in Contact hooks. That way "this document is modified" validation won't trigger. 

cc: @ruchamahabal / @anupamvs

ankush:
@barredterra perhaps it would be better to move to reload right next to insert? 

```python
class Lead(Document):
    def before_insert(self):
        self.contact_doc = frappe.get_doc({
            "first_name": "Original",
            # ...
        }).insert()
        self.contact_doc.reload()

    def after_insert(self):
        self.contact_doc.first_name = "I'm Second"
        self.save()

class Contact(Document):
    def on_update(self):
        frappe.db.set_value(self.doctype, self.name, "fist_name", "I'm First")

```

ankush (approved):

PR #29978, opened 2022-02-24
fix: Commission not applied while making Sales Order from Quotation

deepeshgarg007:
A feature to grant commission on certain items was introduced via PR https://github.com/frappe/erpnext/pull/27467

This works fine when a Sales Invoice/Sales Order was created directly and commission rate was applied but not when a Sales Order/ Invoice is created from quotation and the commission rate is mapped from the quotation.

Steps to test:
1. Make a quotation
2. Convert quotation to Sales Order.
3. Check if amount eligible for commission and commission amount are correct


deepeshgarg007:

PR #29979, opened 2022-02-24
fix: Multi-currency bank reconciliation fixes

deepeshgarg007:
image

The user gets the above error when trying to make reconcile a Bank Transaction with a multi-currency payment entry

The issue is seen when the Paid From Account and Paid To account currency is the same but that currency is not the same as the company currently.

In this case, `base_paid_amount` amount was considered for reconciliation but that is not the same as the amount in Bank Transaction which led to the error.


deepeshgarg007:

PR #29987, opened 2022-02-25
fix: Get MRs that are yet to be received but fully ordered in Report `Requested Items to Order and Receive`

marination:
**Issue:**
- MRs that were fully ordered but not received/partially received were not visible in the Report **Requested Items to Order and Receive**
  Screenshot 2022-02-25 at 12 13 38 PM

**Fix:**
- Remove incorrect query clause that only check if ordered qty < 100
- MR should be visible in report until fully received (cycle complete)
   Screenshot 2022-02-25 at 12 12 58 PM

**To test:**
- Create an MR of type **Purchase** for 10 Qty
- Create a PO against this MR for 10 Qty. Check Report
- Create a PR against this PO for 5 Qty (partial). Check Report
- Report must have this MR in it, as it not fully received/purchase cycle is not yet complete.
- Make another PR for remaining 5 Qty, MR must disappear from report


marination:

Issue #29975, opened 2022-02-23
payment method in POS do not show non default mode of payment


andrewhartono:
### Information about bug

after checkout on POS only default payment method is available while others payment method do not available or do not show

### Module

selling

### Version

ERPNext: v13.21.0 (HEAD)

Frappe Framework: v13.21.0 (HEAD)

### Installation method

FrappeCloud

AlaaBadry:
i made the last update but still have the same problem

AlaaBadry:
@nextchamp-saqib is this final solution or still under test

nextchamp-saqib:
@AlaaBadry The fix has been merged to `develop` and `version-13-hotfix`. I have tested the fix, it works. 

The issue still exists in the v13.21.0 version. It'll be fixed with v13.22.0.

andrewhartono:
@nextchamp-saqib thank you, the new release has been published and available on frappecloud, but after update to v13.22.0 there is new error on POS: server error: AttributeError: 'NoneType' object has no attribute 'options'
i've created new issue: https://github.com/frappe/erpnext/issues/30128
now the POS can not even be used, it can not add other item, can not checkout, can not pay.

andrewhartono:
@nextchamp-saqib The error actually triggers on every form that need to choose items and item quantity such as:
purchase order, purchase receipt, purchase invoice, sales order, delivery note, sales invoice, etc.
we can not save or submit after error, how do we work? please solve and release the fix soon.

andrewhartono:
@nextchamp-saqib Thank you, the problem has been resolved, the payment method is available now

Issue #29959, opened 2022-02-23
Pricing rule is applied even if 'ignore pricing rule' is checked in POS profile


sharmanakul:
### Information about bug

This is related to https://github.com/frappe/erpnext/issues/29354

Setup
1. ignore pricing rule is checked
2. create item and set a pricing rule on it

Steps
1. open POS
2. select the item and qty such that the pricing rule becomes applicable
3. CLick on Checkout
4. Here, discount is not displayed (and rightly so)
5. Click on 'Complete order', now in the invoice the discount is displayed.

### Module

accounts

### Version

v13, 14

sharmanakul:
if you create a POS invoice from the form view, then the 'ignore pricing rule' is displayed as unchecked despite having it checked in the POS profile

PR #29991, opened 2022-02-25
fix(pos): minor fixes

### LINKED ISSUE #29981: MOPs set in POS Profile disappear on checkout before complete
### Information about bug

As mentioned in #29597, the **POS Invoice** form is now saved on checkout,
https://github.com/frappe/erpnext/blob/4d8d4acec4ff486505f1e99c53e3c7ebdd9944a7/erpnext/selling/page/point_of_sale/pos_controller.js#L717-L720

However, the **POS Invoice** controller runs the below code during _validate_, that filters out all payment rows that has zero or no `amount` set. 
https://github.com/frappe/erpnext/blob/f87ba6ac0a07bd3f0f7195afa9ba6071cfcc21d1/erpnext/accounts/doctype/pos_invoice/pos_invoice.py#L441-L445

As no `amount` has been entered in the payment rows apart from the default one, all other mops apart from the default is removed. So when the pos page renders the payment methods (again), it does so only with the default mop set in the **POS Profile**

### Module

accounts, selling

### Version

ERPNext: v13.21.0 (version-13)
Frappe Framework: v13.21.0 (version-13)

### Installation method

docker

_No response_

nextchamp-saqib:
Fixes https://github.com/frappe/erpnext/issues/29981, https://github.com/frappe/erpnext/issues/29975, https://github.com/frappe/erpnext/issues/29959

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -439,7 +439,6 @@ class POSInvoice(SalesInvoice):
 			self.paid_amount = 0
 
 	def set_account_for_mode_of_payment(self):
-		self.payments = [d for d in self.payments if d.amount or d.base_amount or d.default]
 		for pay in self.payments:
 			if not pay.account:
 				pay.account = get_bank_cash_account(pay.mode_of_payment, self.company).get("account")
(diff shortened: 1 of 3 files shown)


nextchamp-saqib:
Unrelated tests failing.

Local tests are passing.

CleanShot 2022-02-25 at 15 33 19

CleanShot 2022-02-25 at 15 34 01

CleanShot 2022-02-25 at 15 34 34

nextchamp-saqib:

PR #29997, opened 2022-02-25
fix: org chart connectors not rendered when Employee Naming is set to Full Name

ruchamahabal:
## Steps to Replicate

- In HR Settings, set **Employee Naming By** to **Full Name**
- Create 2 employees (one reporting to another) having spaces in the name
- The connectors for nodes will not be rendered for such nodes

connec

Also, sometimes after clicking on **Expand All** button following error is thrown:

```python
  File "apps/frappe/frappe/app.py", line 68, in application
  File "apps/frappe/frappe/api.py", line 55, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/__init__.py", line 1214, in call
    return fn(*args, **newargs)
TypeError: get_all_nodes() missing 1 required positional argument: 'company'
```

## Fix

- Use attribute selectors while finding parent and child nodes to connect to handle spaces.

   test-1

- Fix object scope in `get_all_nodes`

### DIFF
--- a/erpnext/public/js/hierarchy_chart/hierarchy_chart_desktop.js
+++ b/erpnext/public/js/hierarchy_chart/hierarchy_chart_desktop.js
@@ -304,12 +304,13 @@ erpnext.HierarchyChart = class {
 	}
 
 	get_child_nodes(node_id) {
+		let me = this;
 		return new Promise(resolve => {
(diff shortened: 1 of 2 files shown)


ruchamahabal:

PR #30012, opened 2022-02-28
feat: added provision to apply discount on RFQ from the portal

anupamvs:
- Added provision on the portal to apply discount while creating supplier quotation from the RFQ. 
rfq

- Fixed List View UI


anupamvs:

PR #30015, opened 2022-02-28
refactor: removed validation to check zero qty

rohitwaghchaure:
1. Set backflush bashed on as "Material Transferred for Subcontract" in the buying settings
2. Create the subcontracted Purchase Order
3. Transfer extra materials in batches
4. While making purchase receipt, update the Consumed Qty of one raw material to zero 
5. Try to save the purchase receipt
6. You will get the error that "the consumed qty cannot be zero for the item XYZ"

**Why user wants to keep the qty as Zero?**
User has transferred extra materials which supplier has not consumed and user wants to return non consumed materials back to the their store warehouse

**Issue**
Screenshot 2022-02-28 at 2 22 43 PM

### DIFF
--- a/erpnext/controllers/subcontracting.py
+++ b/erpnext/controllers/subcontracting.py
@@ -363,20 +363,12 @@ class Subcontracting():
 			return
 
 		for row in self.get(self.raw_material_table):
-			self.__validate_consumed_qty(row)
-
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:

PR #30025, opened 2022-02-28
fix: Exchange rate not getting set in payment entry

deepeshgarg007:
image

The user gets the above error while making a payment entry for some roles (not replicable for all roles) if there is a multi-company setup and no default company is set in global defaults.

This happens because exchange rates are fetched on loading the form but this fails since there is no default company


deepeshgarg007:
@Mergifyio backport develop

PR #30039, opened 2022-03-01
fix: Nil and Exempted values in GSTR-3B Report

deepeshgarg007:
The following section is GSTR-3B report was showing incorrect data because of error in logic

image

This PR fixes that issue


deepeshgarg007:

PR #30041, opened 2022-03-01
fix: incorrect balance serial no in stock ledger report

noahjacob:
### Issue
- When certain filters are set, the balance serial no are incorrect.
- The balance serials are updated solely on the basis of the previous rows within the filters. 

For example, if I create a delivery note and set the voucher no as the filter, the report cannot update the balance since the previous serials aren't in the filter range. The balance only updates correctly when you reset the filters to also fit in the previous entries.
stock_report_balance_Serial

### Fix
- Fetch the available serial nos at first instead of setting it to an empty list as default.

Screenshot 2022-03-01 at 6 28 47 PM

### Steps To Reproduce
- Have some serial numbers in the warehouse for that item.
- Create a delivery note with a qty lower than the available.
- View the stock ledger report with the delivery note voucher no as the filter.


erpnext/stock/report/stock_ledger/stock_ledger.py
```diff
@@ -70,7 +71,10 @@ def update_available_serial_nos(available_serial_nos, sle):
 	serial_nos = get_serial_nos(sle.serial_no)
 	key = (sle.item_code, sle.warehouse)
 	if key not in available_serial_nos:
-		available_serial_nos.setdefault(key, [])
+		stock_balance = get_stock_balance_for(sle.item_code, sle.warehouse, sle.date.split(' ')[0],
+			sle.date.split(' ')[1], batch_no=sle.batch_no)
```
ankush:
```suggestion
			sle.date.split(' ')[1])
```

Not sure why batch no filter is required here 🤔 

otherwise lgtm.

ankush (approved):

PR #30050, opened 2022-03-02
fix: check for debit credit difference even after round-off adjustment

nextchamp-saqib:
While making general ledger entries, any fractional debit credit difference is adjusted into the round-off account. 
However in some cases, after adjusting with the round-off account, debit credit difference still exists and the ledger entries are posted with it.

This PR just validates for debit credit difference even after round-off adjustment has been made and stops the transaction if the difference is greater than allowance (eg. 0.5)


nextchamp-saqib:
Keep Alive

erpnext/accounts/general_ledger.py
```diff
@@ -302,12 +302,29 @@ def validate_cwip_accounts(gl_map):
...
+	debit_credit_diff = get_debit_credit_difference(gl_map, precision)
+	if abs(debit_credit_diff) > allowance:
+		raise_debit_credit_not_equal_error(debit_credit_diff, voucher_type, voucher_no)
+
+	elif abs(debit_credit_diff) >= (1.0 / (10**precision)):
+		make_round_off_gle(gl_map, debit_credit_diff, precision)
+
+	debit_credit_diff = get_debit_credit_difference(gl_map, precision)
+	if abs(debit_credit_diff) > allowance:
+		raise_debit_credit_not_equal_error(debit_credit_diff, voucher_type, gl_map[0].voucher_no)
```
nextchamp-saqib:
```suggestion
		raise_debit_credit_not_equal_error(debit_credit_diff, voucher_type, voucher_no)
```

PR #30057, opened 2022-03-02
fix(Timesheet): fetch exchange rate only if currency is set

ruchamahabal:
In Timesheets if the currency field is not set fetching exchange rate fails:

```python
  File "apps/frappe/frappe/app.py", line 66, in application
  File "apps/frappe/frappe/api.py", line 54, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/__init__.py", line 1232, in call
    return fn(*args, **newargs)
TypeError: get_exchange_rate() missing 1 required positional argument: 'from_currency'
```

Only call `get_exchange_rate` once the currency is set.


ruchamahabal:

PR #30081, opened 2022-03-06
fix: website_image field not copied when creating Website Items

nameduser0:
The `create_website_items` v13 patch assumes that the website image for an item is in the `image` field.
This is clearly not correct since there is a `website_image` field in both the original `Item` and `Website Item` doctypes.

## This means upgrades from v12 to v13 lose images on the homepage.

Introduces a new patch to copy the field across.

This needs someone with an idea of the ERPNext roadmap to figure out if it was intended to remove the `website_image` field on the `Website Item` or not. `image` and `website_image` seem redundant but why doesn't the original patch copy the image from `Item.website_image` to `Website Item.image` ??

PS why isn't the website_image_alt field used in the home.html template instead of item_name ??

### DIFF
--- a/erpnext/patches/v13_0/create_website_items_imagefix.py
+++ b/erpnext/patches/v13_0/create_website_items_imagefix.py
@@ -1,0 +1,9 @@
+import frappe
+
+def execute():
+	"Copy website_image field from Item if image is not already populated"
+	for website_item_name in frappe.db.get_list('Website Item', {'image': ("=", "")}, pluck='name'):
+		witem = frappe.get_doc('Website Item', website_item_name)
+		image = frappe.db.get_value('Item', witem.item_code, 'website_image')
+		witem.image = image
+		witem.save()
--- a/erpnext/patches.txt
+++ b/erpnext/patches.txt
@@ -351,3 +351,4 @@ erpnext.patches.v13_0.update_disbursement_account
 erpnext.patches.v13_0.update_reserved_qty_closed_wo
 erpnext.patches.v13_0.amazon_mws_deprecation_warning
 erpnext.patches.v13_0.set_work_order_qty_in_so_from_mr
+erpnext.patches.v13_0.create_website_items_imagefix


marination:
@casesolved-co-uk  I think this was genuinely missed out. In most cases I've observed the image and website image have matched and both have co-existed.
This makes sense `website_image` should have been copied over not image

`website_image_alt` is used as alt text for images, which is present AFAIK

erpnext/patches/v13_0/create_website_items_imagefix.py
```diff
@@ -0,0 +1,9 @@
+import frappe
+
+def execute():
+	"Copy website_image field from Item if image is not already populated"
+	for website_item_name in frappe.db.get_list('Website Item', {'image': ("=", "")}, pluck='name'):
+		witem = frappe.get_doc('Website Item', website_item_name)
+		image = frappe.db.get_value('Item', witem.item_code, 'website_image')
+		witem.image = image
+		witem.save()
```
marination, line 9: We can do this for all Website Items where the `website_image` in Item does not match that of Website Item (in case of differing values or empty values).
The update could include a simple db update (query) to map `website_image` and thumbnail (for safety). The save will make the patch very slow.
I'll push this change

nameduser0: Do you mean you're going to rewrite it? I agree it could be written more efficiently with a db.set_value.
But like I said, worth checking `home.html` to see which image field it uses (image) and was intended to use and figuring out whether the `website_image` field is required at all? The DocType `Image Field` setting is set to `image`.
While you're looking at the template you'll notice `website_image_alt` is not used.

marination:
ToDo:

- [x] Remove `image` field from Website Item , only use `website_image`
- [x] No need for patch as it attempted to populate `image` with `website_image`. We can just use `website_image` for e-commerce related stuff.
- [x] Handle migration (create_website_items patch)
- [x] Handle Item -> Website Item
- [x] Check all portal pages that use `image` field

marination:
In depth fix done [here](https://github.com/frappe/erpnext/pull/31469). The same will be ported to v13

PR #30091, opened 2022-03-07
feat: Include child item group products in Item Group Page & cleanup

marination:
Documentation: https://docs.erpnext.com/docs/v13/user/manual/en/e_commerce/item_group_wise_product_listing

**Case:**
- Item Group **Seating** as child item groups -> **Dining Chairs** and **Lounge Chairs**
- When the **Seating** Item Group page is visited on the Portal, only Items with **Item Group = Seating** or having Seating in **Website Item Groups**, are pulled.
- The user wants **Dining Chairs** and **Lounge Chairs** and Lounge Chairs to be pulled on the **Seating** page too.
- This can be driven via **Website Item Groups** but it gets too tedious to set Website Item Groups for so many sub groups.
Screenshot 2022-03-07 at 1 14 15 PM

**Feature:**
- Added a checkbox `Include Descendants` in the Item Group. Now users can check this for the Item Groups they want.
   Screenshot 2022-03-07 at 1 19 35 PM
- This will pull in Items belonging to child Item groups as well, with filters applicable on the same
- This PR also cleans up the Website section in Item Group.
- Item Group filters will now be build separately in `query.py` instead of `api.pi` earlier. `api.py` must only pass data not construct data based on business logic.
Screenshot 2022-03-07 at 1 14 32 PM


marination:
Unrelated failures

PR #30092, opened 2022-03-07
feat(sales order): added required_by_date in Update Items dialog, SO form & SO items table

kkulloters:
**Details of the feature request:**
To add Reqd By Date in the Sales order form, sales order items table, and on update items dialog

**Solution:**
1. Added Reqd By Date in Sales Order and Sales Order Item Doctype
2. Added Reqd By Date column in Update Items Dialog

Screenshot of the feature request:
**Before**
OLD

**After**
New

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -1906,6 +1906,7 @@ def set_order_defaults(parent_doctype, parent_doctype_name, child_doctype, child
 	child_item.warehouse = get_item_warehouse(item, p_doc, overwrite_warehouse=True)
 	conversion_factor = flt(get_conversion_factor(item.item_code, child_item.uom).get("conversion_factor"))
 	child_item.conversion_factor = flt(trans_item.get('conversion_factor')) or conversion_factor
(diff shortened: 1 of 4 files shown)


ankush:
"delivery date" is "expected delivery date" ~ "required date".. or at least that's how I have interpreted it always 😅 

Do we need to duplicate this info?

kkulloters:
Hello @ankush 

Functionally, our company uses the following dates:

> Required By Date = The date that the customer wants to receive the assemblies.
> Delivery Date = Actual date that the company sends the goods to the customer.

The reason for the two dates is for transparency purposes. Ideally, the goods are sent to the customer on the exact day that they require it. However, in manufacturing companies, the actual delivery date can be earlier or later than the "Required By Date".

Having separate columns in the Sales Order allows the customers to plan for the early or delayed arrival of goods properly.
And this can also be the basis for a Supplier rating from the customer's point of view.

Note: Especially in today's situation, wherein there is a worldwide allocation of electronic components, and a disruption in the supply chain management across the globe, the planning for logistics is of paramount importance for any organization to stay competitive.

ankush:
Since this is rather specific to your use case (which can be achieved via custom fields easily) adding another date field will make this more confusing IMO.

PR #30130, opened 2022-03-09

latoile:
image

image

image

`no-docs`

Following the request: https://github.com/frappe/erpnext/pull/29028#issuecomment-1062654108


deepeshgarg007:
@latoile we have recently decided to move the localization to separate region-wise respective applications. We will ask you reopen and port this once the app for this region is ready. Closing this PR for now.

PR #30136, opened 2022-03-09
fix: Remove tax invoice no field

deepeshgarg007:
Introduced via https://github.com/frappe/erpnext/pull/30133

Remove wrongly added field


deepeshgarg007:

PR #30146, opened 2022-03-09
refactor: Add exception handling in background job within BOM Update Tool

marination:
> List View throws an error due to https://github.com/frappe/frappe/issues/16302 (its still usable)

**Issue:**
- BOM Update Tool jobs run in the background, there's no way to know about failures, if any (due to timeout or bug)
- This PR adds error logging and enhances the UX

**Working:**
- The **Replace** button is inactive until both BOMs are mentioned. On clicking on **Replace**, a **BOM Update Log** is created. This log maintains state and lets user know of failure.
   
- This log has two types: **Replace BOM** and **Update Cost**. Each is self explanatory as to what job is being executed.
- List View:
   Screenshot 2022-03-31 at 12 53 55 PM

- Failed Log:
  Screenshot 2022-03-17 at 1 05 44 PM
- Completed Log
  Screenshot 2022-03-17 at 1 07 03 PM
- Failed Log with simulated timeout error:
  

**Todo:**
- [x] Add BOM Update Log with status
- [x] Tests for Log


erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+	new_bom = boms.get("new_bom")
+
+	unit_cost = get_new_bom_unit_cost(new_bom)
+	update_new_bom(unit_cost, current_bom, new_bom)
+
+	frappe.cache().delete_key("bom_children")
+	parent_boms = get_parent_boms(new_bom)
+
+	with click.progressbar(parent_boms) as parent_boms:
+		pass
```
ankush: Is this required? if yes, following should be part of contextmanager scope?

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+	for bom in parent_boms:
+		bom_obj = frappe.get_cached_doc("BOM", bom)
```
ankush: Since cache eviction isn't perfect, should we rely on the cached doc? 😅 

Any changes done via raw SQL / unsafe ORM methods might be lost.

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+			bom_obj.save_version()
+
+
+def update_new_bom(unit_cost: float, current_bom: str, new_bom: str) -> None:
+	bom_item = frappe.qb.DocType("BOM Item")
+	frappe.qb.update(bom_item).set(bom_item.bom_no, new_bom).set(bom_item.rate, unit_cost).set(
+		bom_item.amount, (bom_item.stock_qty * unit_cost)
+	).where(
+		(bom_item.bom_no == current_bom) & (bom_item.docstatus < 2) & (bom_item.parenttype == "BOM")
+	).run()
```
ankush, line 101: This looks ugly (it is my fault though 🥲)


All select queries look better but updates are problematic since they aren't wrapped


Just wrapping it in `( )` makes it pretty:


 
Screenshot 2022-03-30 at 3 24 43 PM

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+
+	with click.progressbar(parent_boms) as parent_boms:
+		pass
+	for bom in parent_boms:
+		bom_obj = frappe.get_cached_doc("BOM", bom)
+		# this is only used for versioning and we do not want
+		# to make separate db calls by using load_doc_before_save
+		# which proves to be expensive while doing bulk replace
+		bom_obj._doc_before_save = bom_obj
+		bom_obj.update_new_bom(unit_cost, current_bom, new_bom)
```
ankush: BOM methodd has a different signature: https://github.com/frappe/erpnext/blob/ebf00946c91bf03105533d46c85e9b405cc7d62a/erpnext/manufacturing/doctype/bom/bom.py#L700 

perheps the new function added in this file is what you meant to call?

If yes, then this needs to be revisited, cause direct DB updates won't be present on BOM object from cache.

ankush: also add a test for this, since currently build is passing _somehow_

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+		boms = frappe._dict(boms or {})
+
+		if update_type == "Replace BOM":
+			replace_bom(boms)
+		else:
+			update_cost()
+
+		doc.db_set("status", "Completed")
+
+	except (Exception, JobTimeoutException):
```
ankush:
```suggestion
	except Exception:
```

Second one inherits from `Exception` so it's same.

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -0,0 +1,163 @@
...
+	bom = frappe.qb.DocType("BOM")
+	new_bom_unitcost = (
+		frappe.qb.from_(bom).select(bom.total_cost / bom.quantity).where(bom.name == new_bom).run()
+	)
+
+	return flt(new_bom_unitcost[0][0])
+
+
+def run_bom_job(
+	doc: "BOMUpdateLog", boms: Optional[Dict] = None, update_type: Optional[str] = "Replace BOM"
```
ankush:
```suggestion
	doc: "BOMUpdateLog", boms: Optional[Dict[str, str]] = None, update_type: str = "Replace BOM"
```

- `boms` maps old name to new name, so encoding that info
- `update_type` isn't really optional, it just has a default value. For instance it can't be `None`? 

---

I see, the "Update Cost" isn't explicitly added when doing update cost. 


Type should be `Literal["Replace BOM", "Update Cost"]`

ankush (changes requested):

erpnext/manufacturing/doctype/bom_update_tool/bom_update_tool.py
```diff
@@ -1,136 +1,67 @@
...
+def update_cost() -> None:
+	"""Updates Cost for all BOMs from bottom to top."""
 	bom_list = get_boms_in_bottom_up_order()
 	for bom in bom_list:
 		frappe.get_doc("BOM", bom).update_cost(update_parent=False, from_child_bom=True)
 
-	frappe.db.auto_commit_on_many_writes = 0
+
+def create_bom_update_log(
+	boms: Optional[Dict] = None, update_type: str = "Replace BOM"
```
ankush: why accept `None` as boms here? 🤔

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: removing it,log is good enough for this cut. Job in normal cases shouldn't take that long also.

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: yeah it auto formatted this lol

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: I'll rename the new function.
meant to call the bom object function, must've gotten mixed up.
It passes because theres no checks and the snippet below always continues 😿 
Screenshot 2022-03-30 at 4 28 31 PM

erpnext/manufacturing/doctype/bom_update_tool/bom_update_tool.py
marination: In case of "Update Cost" type log, boms can be `None`, they dont need to be passed

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: @ankush Removed `update_new_bom()` in https://github.com/frappe/erpnext/pull/30146/commits/620575a9012a9759c6285558ac25c6709c4e92cc

Did the same thing `update_new_bom_in_bom_items()` does. Double check ?

ankush: Yep, both seem to be doing same thing.

Actually no, update cost wasn't working since it's updating child items but not really saving to db 🤦 (db_update only updates main document not child items)

marination:
@ankush as discussed will fix cost updation inconsistencies in a different PR. Good to go for now ?

ankush:
unrelated test failing.

ankush (approved):

marination:
Fixed the failing test in https://github.com/frappe/erpnext/pull/30509

PR #30147, opened 2022-03-10
feat(accounts_controller): Allow update item in Update Item button in PO and SO

BevanTony1:
Feature Details:
--
Allow Sales Manager users to change the item_code in Purchase Order and Sales Order using the "Update Item" button.

Users should be able to update the items in the Sales Order/Purchase Order. By updating the item_code directly, the current process for updating submitted SO/PO is deleting the item and using the add button.

Purchase Order:
--
Purchase Order (submitted)

Sales Order:
--
Sales Order (Submitted)

Validations:
--
Only users with the Sales Manager role are only ones allowed to perform the update.
Non-salesmanager

if item line is partially billed or delivered. Prevent item code from being updated.

**Purchase Order**
Purchase Order (billed)
**Sales Order**
Sales Order (billed)

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -2044,7 +2044,7 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
 		else:
 			check_doc_permissions(parent, 'write')
 			child_item = frappe.get_doc(parent_doctype + ' Item', d.get("docname"))
-
+			prev_item_code, new_item_code = child_item.item_code,d.get('item_code')
 			prev_rate, new_rate = flt(child_item.get("rate")), flt(d.get("rate"))
 			prev_qty, new_qty = flt(child_item.get("qty")), flt(d.get("qty"))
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@BevanTony1 Please add a test case for this and move it to the develop branch

ankush:
Ah, I forgot to comment on this. 

This is also problematic in case the item row is already linked with some deliveries or other documents.

Deleting and adding a new line seems like a perfectly fine workaround. This whole feature is a workaround for amending documents in case of "minor" changes like qty and rate. Change of item code is not a small change!

BevanTony1:
> Deleting and adding a new line seems like a perfectly fine workaround. This whole feature is a workaround for amending documents in case of "minor" changes like qty and rate. Change of item code is not a small change!
Hello @ankush,

Thank you for your feedback.

It "frappe.throw" if the item line is partially received/delivered so users shouldn't be able to update item_codes with item lines with linked orders/deliveries.

I also think this is a better workaround for deleting then adding item lines

PR #30152, opened 2022-03-10
fix: Sales and Purchase return optimization

deepeshgarg007:
Trying to create a return invoice (Credit Note) against a purchase invoice that has a lot of items (around 250+) and when the database size is comparatively large (have a lot of other purchase invoices) the credit note creation request gets timed out because the `get_returned_qty_map_for_row` has a query which takes a lot of time. 

Solution:
Modified the query to use an already existing index due to which the query time is reduced. Also added the same index for some missing doctypes

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -926,3 +926,6 @@ def get_item_account_wise_additional_cost(purchase_document):
 						account.base_amount * item.get(based_on_field) / total_item_cost
 
 	return item_account_wise_cost
+
+def on_doctype_update():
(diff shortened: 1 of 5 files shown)


erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
```diff
@@ -926,3 +926,6 @@ def get_item_account_wise_additional_cost(purchase_document):
 						account.base_amount * item.get(based_on_field) / total_item_cost
 
 	return item_account_wise_cost
+
+def on_doctype_update():
+	frappe.db.add_index("Purchase Receipt", ["supplier", "is_return", "return_against"])
```
nextchamp-saqib, line 931: You will also have to modify the doctype to update the indexes

deepeshgarg007:

PR #30160, opened 2022-03-10
fix: 'save_quotations_as_draft' checkbox not honoured

marination:
**Issue:**
- Quotations were submitted even though `save_quotations_as_draft` was enabled in E commerce Settings

**Fix:**
- Make sure `request_for_quotation` considers `save_quotations_as_draft`
- Added test for checkout disabled quote


marination:
Unrelated failures

PR #30174, opened 2022-03-10
fix: Allow on Submit for Material Request Item Required Date

bhattdevarsh:
### ISSUE
User not able to change/update the Required Date after Submitting, but the main(parent) document has "Allow on Submit" for Required date.

### SOLUTION
Allow user to Update the Required Date Item wise after Submitting the document.

### DIFF
--- a/erpnext/stock/doctype/material_request/material_request.py
+++ b/erpnext/stock/doctype/material_request/material_request.py
@@ -85,7 +85,7 @@ class MaterialRequest(BuyingController):
 
 	def before_update_after_submit(self):
 		self.validate_schedule_date()
-		
+
 	def validate_material_request_type(self):
 		""" Validate fields in accordance with selected type """
 
--- a/erpnext/stock/doctype/material_request/material_request.py
+++ b/erpnext/stock/doctype/material_request/material_request.py
@@ -83,6 +83,9 @@ class MaterialRequest(BuyingController):
 		self.reset_default_field_value("set_warehouse", "items", "warehouse")
 		self.reset_default_field_value("set_from_warehouse", "items", "from_warehouse")
 
+	def before_update_after_submit(self):
+		self.validate_schedule_date()
+		
 	def validate_material_request_type(self):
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/material_request_item/material_request_item.json
```diff
@@ -177,6 +177,7 @@
    "fieldtype": "Column Break"
   },
   {
+   "allow_on_submit": 1,
```
ankush, line 180: If we want to allow this then we also need to validate by adding same validation in `before_update_after_submit` methods on material request.


ref:
https://github.com/frappe/erpnext/blob/79a51b24812725a541c3accf68fdde004582da5d/erpnext/stock/doctype/material_request/material_request.py#L59-L62

https://github.com/frappe/frappe/blob/a886d793f29fb3c291620958286b52b3665c1ebc/frappe/model/document.py#L955-L961

ankush (changes requested):

erpnext/stock/doctype/material_request_item/material_request_item.json
bhattdevarsh: I think Validate does not run on Update, can you please explain the reason of updating Validate code? @ankush

ankush: It **should** run on update, right now it doesn't :) that's what I am asking for adding.

ankush (approved):

erpnext/stock/doctype/material_request/material_request.py
```diff
@@ -83,6 +83,9 @@ def validate(self):
 		self.reset_default_field_value("set_warehouse", "items", "warehouse")
 		self.reset_default_field_value("set_from_warehouse", "items", "from_warehouse")
 
+	def before_update_after_submit(self):
+		self.validate_schedule_date()
+
```
ankush:
```suggestion

```

whitepsace

PR #30179, opened 2022-03-10
fix: Filter fields in E Commerce Settings (use Website Item)

MostafaFekry:
https://github.com/frappe/erpnext/pull/30733 for details

Update get meta from Item to Website Item
To allow Website Item Fields to display meta fields

**ToDo:**

- [x] test patch with table multiselect
- [x] tests


ankush:
@MostafaFekry please be a bit more descriptive. So it's easier for others to review and merge. 

What problem does this PR fix? 


same for https://github.com/frappe/erpnext/pull/30180

ankush:
The fix doesn't seem to work 🤔 
Screenshot 2022-03-28 at 6 09 47 PM


This isn't working cause `fieldname` is reserved property. So this field has to be changed... perhaps to autocomplete.

ref: https://github.com/frappe/erpnext/pull/30416 

cc @marination

marination:
> The fix doesn't seem to work 🤔 Screenshot 2022-03-28 at 6 09 47 PM
> 
> This isn't working cause `fieldname` is reserved property. So this field has to be changed... perhaps to autocomplete.
> 
> ref: #30416
> 
> cc @marination

@ankush  It works for me 🤔 


Also, autocomplete is a feature specific to v14. Since we don't want to port features cant do this for v13. It'll just be the same bad UX with correctness.
I'll push v13 suitable commits from https://github.com/frappe/erpnext/pull/30733 here

ankush:
@marination tried https://github.com/frappe/erpnext/pull/30733 works locally. Not sure why it didn't work for me before 👀

marination:
Unrelated tests breaking due to merge conflicts in the codebase via another PR

marination (approved):

PR #30180, opened 2022-03-10
fix: e_commerce_settings.py

MostafaFekry:
Update get meta from Item to Website Item doctype to check validate_field_filters


ankush:
merged with https://github.com/frappe/erpnext/pull/30179

PR #30181, opened 2022-03-10
fix: Shipping rule application fixes

deepeshgarg007:
image

The shipping rule was not getting applied on changing the net total. When the net total was not falling into any of the slabs mentioned in the shipping rule, the tax amount for shipping rule was not getting set to 0


deepeshgarg007:

PR #30183, opened 2022-03-10
Update order.html

sam-anam1:
It dose make sense to be {{ doc.title }} instead of {{ party_name }} as well as its customer portal
Those would be helpful which every column has own class  when styling by using "Website Settings --> HTML Header, Robots and Redirects"

> Screenshots/GIFs


sam-anam1:
I may clear it those requests are None of these classes exist in CSS
it just would be helpful and useful using Website Settings --> HTML Header, Robots and Redirects at customer portal end

sam-anam1:
@ankush  and this please gotta closed,

PR #30184, opened 2022-03-10
Update fields class projects.html

sam-anam1:
Those it would be really helpful that fields has own class when styling for mobile or desktop for Customer portal end; by using "Website Settings --> HTML Header, Robots and Redirects"


sam-anam1:
I may clear it those requests are None of these classes exist in CSS
it just would be helpful and useful using Website Settings --> HTML Header, Robots and Redirects at customer portal end

sam-anam1:
@ankush  and this please gotta closed, as mentioned.. it just that the goal most fields has own css at portal end.

PR #30185, opened 2022-03-10
chore: class selectors for customizability

sam-anam1:
Really helpful for customer portal
 those fields has own class when styling for mobile or desktop by using Website Settings --> HTML Header, Robots and Redirects

### DIFF
--- a/erpnext/templates/includes/projects/project_row.html
+++ b/erpnext/templates/includes/projects/project_row.html
@@ -1,11 +1,11 @@
 {% if doc.status == "Open" %}
   <div class="web-list-item transaction-list-item">
     <div class="row">
-      <div class="col-xs-2 ProjectLink">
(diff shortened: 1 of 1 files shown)


erpnext/templates/includes/projects/project_row.html
```diff
@@ -1,11 +1,11 @@
 {% if doc.status == "Open" %}
   <div class="web-list-item transaction-list-item">
     <div class="row">
-      <div class="col-xs-2">
+      <div class="col-xs-2 ProjectLink">
```
ankush: None of these classes exist in CSS 👀

ankush: are you proposing this for customizability?

sam-anam1:
please as I said this just apply by Website Settings --> HTML Header, Robots
I would be useful for customer portal

PR #30191, opened 2022-03-11
fix: incorrect payable amount for loan closure

abhinavxd:
- For Loan closure, the penalty amount should also be added to the payable amount


deepeshgarg007:

PR #30206, opened 2022-03-13
fix: Do not update ignore prcing rule check implicitly

deepeshgarg007:
`Ignore Pricing Rule` was implicitly enabled when converting an order to an Invoice. This should not be done as this confuses the users


deepeshgarg007:

PR #30207, opened 2022-03-13
fix: Do not consider cancelled entries

deepeshgarg007:
Do not consider cancelled entries in Trial Balance for Party Report


deepeshgarg007:

PR #30216, opened 2022-03-14
fix: max_qty validation condition in WO

s-aga-r:
### Issue(s): Work Order

_Problem:_
- Since we can have decimal value for Qty. If the Max Qty is less than "1" the error will be thrown.

_Proposed Solution:_
- Error will throw if the planned qty is not greater than "0".


ankush:
@s-aga-r can you add a simple test for this? Just need to create WO with <1 qty to make sure they work.

ankush:
@mergifyio copy develop

ankush (approved):

PR #30223, opened 2022-03-14
fix: disable "get items" buttons from submitted production plan

ankush:
telegram-cloud-photo-size-5-6159229316215451305-y

In bizarre cases when the doc is submitted and these buttons are clicked again it adds new rows to the production items table. 

If work orders are created from this client-side doc, they will be linked to a non-existing plan item.

Fix: remove buttons and patch data.


ankush:
client-side behavior. Not possible to unit test.

PR #30225, opened 2022-03-14
fix: Search query of payroll entry reference in Journal Entry

nabinhait:
Issue:
In the Journal Entry child table, "Reference name" does not populate if "Reference Type" is selected as "Payroll Entry".

Screenshot 2022-03-14 at 2 18 20 PM

Fix:
After this fix, it will show all the Payroll Entry records against which bank entry has not been created.

root cause: double escaping.


nabinhait:
@Mergifyio backport develop

ankush (approved):

PR #30228, opened 2022-03-14
fix: While creating Payment Request from other forms, open a new Payment Request form without saving

nabinhait:
Issue:

Currently, while creating a Payment Request from other forms, the system saves a Payment Request and then redirect it to the new record. It does not allow to user to choose a different series for the Payment Request.

Fix:

Redirect to the new Payment Request after mapping proper data but without saving the record. It will allow the user to choose any other series.

### DIFF
--- a/erpnext/accounts/doctype/payment_request/payment_request.py
+++ b/erpnext/accounts/doctype/payment_request/payment_request.py
@@ -376,8 +376,8 @@ def make_payment_request(**args):
 		if args.order_type == "Shopping Cart" or args.mute_email:
 			pr.flags.mute_email = True
 
-		pr.insert(ignore_permissions=True)
 		if args.submit_doc:
+			pr.insert(ignore_permissions=True)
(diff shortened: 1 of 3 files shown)


nabinhait:
@Mergifyio backport develop

nabinhait:

nabinhait:

erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -376,8 +376,8 @@ def make_payment_request(**args):
 		if args.order_type == "Shopping Cart" or args.mute_email:
 			pr.flags.mute_email = True
 
-		pr.insert(ignore_permissions=True)
 		if args.submit_doc:
+			pr.insert(ignore_permissions=True)
```
ankush, line 380: Why? This is a huge breaking change. Earlier it was inserting a document and now it just returns an in-memory document.

PR #30230, opened 2022-03-14
fix: KSA E-Invoice QR Code showing wrong VAT amount

deepeshgarg007:
VAT Amount incorrect in KSA VAT Amount

Screenshot 2022-03-14 at 4 21 34 PM

Before (Incorrect):
Before

After (Correct):
After

### DIFF
--- a/erpnext/regional/saudi_arabia/utils.py
+++ b/erpnext/regional/saudi_arabia/utils.py
@@ -130,7 +130,7 @@ def create_qr_code(doc, method=None):
 def get_vat_amount(doc):
 	vat_settings = frappe.db.get_value('KSA VAT Setting', {'company': doc.company})
 	vat_accounts = []
-	vat_amount  = 0
+	vat_amount = 0
 
 	if vat_settings:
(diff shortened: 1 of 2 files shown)


erpnext/regional/saudi_arabia/utils.py
```diff
@@ -127,6 +127,22 @@ def create_qr_code(doc, method=None):
 		doc.db_set('ksa_einv_qr', _file.file_url)
 		doc.notify_update()
 
+def get_vat_amount(doc):
+	vat_settings = frappe.db.get_value('KSA VAT Setting', {'company': doc.company})
+	vat_accounts = []
+	vat_amount = 0
+
+	if vat_settings:
+		vat_settings_doc = frappe.get_doc('KSA VAT Setting', vat_settings)
```
nextchamp-saqib:
```suggestion
		vat_settings_doc = frappe.get_cached_doc('KSA VAT Setting', vat_settings)
```

deepeshgarg007:

PR #29065, opened 2021-12-29
feat: feature to create 1 PL for several SO irrespective of customers and t…

KrithiRamani:
…o create several DNs grouped by each customer

Documentation Updates:

https://docs.erpnext.com/docs/v13/user/manual/en/stock/delivery-note/edit?wiki_page_patch=eb709e059f
https://docs.erpnext.com/docs/v13/user/manual/en/stock/pick-list/edit?wiki_page_patch=24f2742f11

**Test Cases:**

Customer is not mandatory in PL

Allow selection of multiple SO from several customers

Do not edit Picked Qty in PL. Submit the PL
	- Picked Qty should = Stock Qty

Let qty = 10. Enter 5 as picked qty and submit
	- Picked Qty should remain as 5 and should not be set to stock qty

On submit of PL
	-  “% Picked” (new field) should be updated in SO.
	- “Picked Qty”(new field) should be updated in corresponding SO item row

Create PL for an SO where 
	* “% Picked” is 100%. - On submission, should throw validation message
		* “% Picked” is < 100 % - Should allow submission until 100%

Create PL from multiple SO by clicking on ‘Get Items’
	- submit and create DN from PL 
	- should create separate DNs for each customer.

Create PL by adding items manually(not from SO). Then select an SO also. PL now has both items from SO and those not from SO
	- submit and create DN from PL
	- should create 1 DN for items added manually
	- separate DNs for items from SO, grouped by customer

PR #30238, opened 2022-03-14
feat: Create single PL/DN from several SO. 

KrithiRamani:
New PR from latest develop to avoid rebase (old PR https://github.com/frappe/erpnext/pull/29065 )

Documentation Updates:

https://docs.erpnext.com/docs/v13/user/manual/en/stock/delivery-note/edit?wiki_page_patch=eb709e059f
https://docs.erpnext.com/docs/v13/user/manual/en/stock/pick-list/edit?wiki_page_patch=24f2742f11

Test Cases:

Customer is not mandatory in PL

Allow selection of multiple SO from several customers

Do not edit Picked Qty in PL. Submit the PL
- Picked Qty should = Stock Qty

Let qty = 10. Enter 5 as picked qty and submit
- Picked Qty should remain as 5 and should not be set to stock qty

On submit of PL
- “% Picked” (new field) should be updated in SO.
- “Picked Qty”(new field) should be updated in corresponding SO item row

Create PL for an SO where
* “% Picked” is 100%. - On submission, should throw validation message
* “% Picked” is < 100 % - Should allow submission until 100%

Create PL from multiple SO by clicking on ‘Get Items’
- submit and create DN from PL
- should create separate DNs for each customer.

Create PL by adding items manually(not from SO). Then select an SO also. PL now has both items from SO and those not from SO
- submit and create DN from PL
- should create 1 DN for items added manually
- separate DNs for items from SO, grouped by customer

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.json
+++ b/erpnext/selling/doctype/sales_order/sales_order.json
@@ -130,6 +130,7 @@
   "per_delivered",
   "column_break_81",
   "per_billed",
+  "per_picked",
   "billing_status",
   "sales_team_section_break",
   "sales_partner",
@@ -1514,13 +1515,19 @@
    "fieldtype": "Currency",
    "label": "Amount Eligible for Commission",
    "read_only": 1
+  },
+  {
+   "fieldname": "per_picked",
+   "fieldtype": "Percent",
+   "label": "% Picked",
+   "read_only": 1
   }
  ],
  "icon": "fa fa-file-text",
  "idx": 105,
  "is_submittable": 1,
  "links": [],
- "modified": "2021-10-05 12:16:40.775704",
+ "modified": "2022-03-15 21:38:31.437586",
  "modified_by": "Administrator",
  "module": "Selling",
  "name": "Sales Order",
@@ -1594,6 +1601,7 @@
  "show_name_in_global_search": 1,
  "sort_field": "modified",
  "sort_order": "DESC",
+ "states": [],
  "timeline_field": "customer",
  "title_field": "customer_name",
  "track_changes": 1,
--- a/erpnext/selling/doctype/sales_order_item/sales_order_item.json
+++ b/erpnext/selling/doctype/sales_order_item/sales_order_item.json
@@ -23,6 +23,7 @@
   "quantity_and_rate",
   "qty",
   "stock_uom",
+  "picked_qty",
   "col_break2",
   "uom",
   "conversion_factor",
@@ -798,12 +799,17 @@
    "fieldtype": "Check",
    "label": "Grant Commission",
    "read_only": 1
+  },
+  {
+   "fieldname": "picked_qty",
+   "fieldtype": "Float",
+   "label": "Picked Qty"
   }
  ],
  "idx": 1,
  "istable": 1,
  "links": [],
- "modified": "2022-02-24 14:41:57.325799",
+ "modified": "2022-03-15 20:17:33.984799",
  "modified_by": "Administrator",
  "module": "Selling",
  "name": "Sales Order Item",
--- a/erpnext/stock/doctype/pick_list/pick_list.py
+++ b/erpnext/stock/doctype/pick_list/pick_list.py
@@ -73,7 +73,7 @@ class PickList(Document):
 		for item in so_doc.get('items'):
 			total_picked_qty += flt(item.picked_qty)
 			total_so_qty += flt(item.stock_qty)
-		total_picked_qty = total_picked_qty	+ picked_qty
+		total_picked_qty=total_picked_qty + picked_qty
 		per_picked = total_picked_qty/total_so_qty * 100
 
 		so_doc.db_set("per_picked", flt(per_picked) ,update_modified=False)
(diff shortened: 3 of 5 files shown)


rohitwaghchaure:

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -24,8 +26,21 @@ def validate(self):
 	def before_save(self):
 		self.set_item_locations()
 
+		# set percentage picked in SO
+		for location in self.get('locations'):
+			if location.sales_order and frappe.db.get_value("Sales Order",location.sales_order,"per_picked") == 100:
+				frappe.throw("Row " + str(location.idx) + " has been picked already!")
```
ankush, line 32: Missing translations. This message all other similar messages should be old-style format string wrapped in `_()`


https://frappeframework.com/docs/v13/user/en/translations

erpnext/selling/doctype/sales_order/sales_order.json
```diff
@@ -1514,13 +1515,19 @@
    "fieldtype": "Currency",
    "label": "Amount Eligible for Commission",
    "read_only": 1
+  },
+  {
+   "fieldname": "per_picked",
+   "fieldtype": "Percent",
+   "label": "% Picked",
+   "read_only": 1
```
ankush, line 1523: This should have `no_copy`

erpnext/selling/doctype/sales_order_item/sales_order_item.json
```diff
@@ -798,12 +799,17 @@
    "fieldtype": "Check",
    "label": "Grant Commission",
    "read_only": 1
+  },
+  {
+   "fieldname": "picked_qty",
+   "fieldtype": "Float",
+   "label": "Picked Qty"
```
ankush, line 806: This should have `no_copy` and read only.

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -37,6 +52,32 @@ def before_submit(self):
 			frappe.throw(_('For item {0} at row {1}, count of serial numbers does not match with the picked quantity')
 				.format(frappe.bold(item.item_code), frappe.bold(item.idx)), title=_("Quantity Mismatch"))
 
+	def before_cancel(self):
+		#update picked_qty in SO Item on cancel of PL
+		for location in self.get('locations'):
+			if location.sales_order_item:
+				self.update_so(location.sales_order_item,0,location.item_code)
+
+	def update_so(self,so_item,picked_qty,item_code):
```
ankush, line 61: missing whitespace? Multiple instances of this. 😅 

for future reference: https://peps.python.org/pep-0008/

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -37,6 +52,32 @@ def before_submit(self):
...
+
+		total_picked_qty = 0
+		total_so_qty = 0
+		for item in so_doc.get('items'):
+			total_picked_qty += flt(item.picked_qty)
+			total_so_qty += flt(item.stock_qty)
+		total_picked_qty=total_picked_qty + picked_qty
+		per_picked = total_picked_qty/total_so_qty * 100
+
+		so_doc.db_set("per_picked", flt(per_picked) ,update_modified=False)
```
ankush, line 79: This can be done once per SO instead of once per item.

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -37,6 +52,32 @@ def before_submit(self):
 			frappe.throw(_('For item {0} at row {1}, count of serial numbers does not match with the picked quantity')
 				.format(frappe.bold(item.item_code), frappe.bold(item.idx)), title=_("Quantity Mismatch"))
 
+	def before_cancel(self):
+		#update picked_qty in SO Item on cancel of PL
+		for location in self.get('locations'):
+			if location.sales_order_item:
+				self.update_so(location.sales_order_item,0,location.item_code)
```
ankush, line 59: This does _nothing_ on cancellation 😄 Picked qty remains as is since it's already there in already_picked. Here we need to reverse impact by adding negative qty to already picked qty.

ankush: please add a test for this.

erpnext/stock/doctype/pick_list/test_pick_list.py
```diff
@@ -329,9 +328,9 @@ def test_pick_list_for_items_with_multiple_UOM(self):
 			'purpose': 'Delivery',
 			'locations': [{
 				'item_code': '_Test Item',
-				'qty': 1,
-				'stock_qty': 5,
-				'conversion_factor': 5,
+				'qty': 2,
+				'stock_qty': 1,
+				'conversion_factor': 0.5,
```
ankush, line 333: Not sure why an unrelated test is being modified here.

erpnext/stock/doctype/pick_list/pick_list.py
ankush: also use of `!` in user facing messages should be avoided.

erpnext/stock/doctype/pick_list/pick_list.py
ankush: reference.

Created SO, created PL, cancelled PL. Picked qty doesn't go back to zero. 

Screenshot 2022-03-17 at 6 48 04 PM

Screenshot 2022-03-17 at 6 48 12 PM

PR #30254, opened 2022-03-15
fix: Payment Request Amount calculation in case of multi-currency

nabinhait:
**Issue:**
Payment Request Amount is calculated incorrectly if the PO currency is different from the party account currency and there is any advance against the PO.

Purchase Order:

PO Currency = USD
Company Currency = INR
Party Account Currency = INR
Conversion Rate = 70

Grand Total = $100
Advance Paid = INR 3500

**Calculation Before fix:**
The system was calculating the outstanding amount as $(100 - 3500) = -$3400, hence giving an error saying "Payment Request is already created"

**Calculation After fix:**
Outstanding Amount = $(100 - 3500/70) = $50


nabinhait:

nabinhait:
@mergifyio backport develop

PR #30255, opened 2022-03-15
Version 13 hotfix

mcterzioglu:
Hi,

   I tHnik the payment status is wrong cause When a foreign currency invoice is issued, it should show the difference as a foreign currency difference. But now it looks like it's overdue. The invoice status should be "exchange rate differences".

In this situation we create exchange rate differences invoice in Turkey. Whoever owns the foreign exchange difference has to issue an invoice for the exchange difference.

An example of foreign exchange calculation is as follows:
<html>
<body>

Date | Transaction | Amount (currency) | Currency | Setup | Amount (TL)
-- | -- | -- | -- | -- | --
03.01.2022 | Invoice to customer | 100,00 | USD | 13,60 | 1.360 TL
28.01.2022 | customer payment | -100,00 | USD | 13,63 | -1.363 TL
  | Balance |   |   |   | -3 TL

</body>
</html>

In the above example; Company A's customer B was invoiced for 100 USD, including VAT, and a payment of 100 USD was received from firm B in return, and the account was closed commercially. However, this account was not closed in the legal books, and in the legal books of company A (this amount should also be in the legal records of customer B), there is a balance of 3 TL in customer B account.

In order to close the TL balance in the legal books, company A has to issue a 3 TL "Exchange rate difference" invoice to customer B, including VAT.

"The party with the money in favor of the TL account issues the foreign exchange invoice. In the example above, since firm A is in an advantageous position in its TL account (it has collected more than the invoice amount it has issued).

Therefore, customer B has to issue a foreign exchange invoice. If the exchange rate decreased after the invoice was issued and the TL amount paid was less than the TL amount of the invoice, this time customer B would issue a foreign exchange invoice to company A.

Right at this point, "How is the exchange difference invoice issued?" Let's answer the question. In general, the foreign exchange difference invoice is issued just like other types of invoices. But there are a few important points to note:

It should be arranged in Turkish lira.
VAT calculation should be made according to the exchange rate difference and should be processed in this way on the invoice.
It should be stated in the explanation section of the invoice that there is a "currency difference invoice".
If the other party is an e-invoice payer, the invoice should be issued in e-invoice format."

image

image

> Screenshots/GIFs


ghost:
Sider gave up on analyzing this pull request because of too many changed files.

ankush:
Your PR doesn't have any new commits wrt base version-13-hotfix

ankush:
reopen after fixing branch.

PR #30264, opened 2022-03-15
fix: report sales payments summary

HarryPaulo:
PS: on report filters Check **payment_detail** must be checked

I found two problems in the report:

1 - When searching for the change amount, the system was searching based on the mode_of_payment field, but as this is a field of the payment method description, it is not correct.

Changed to base type = 'Cash'

2 - In the select that searches for payments made through Payment Entry, it was bringing a wrong value for the following situation:

Ex: Let's say I have 5 sales invoices for the same customer
$50 / each invoice
$250 total

I create only 1 Payment Entry, paying $250 by linking 5 invoices of $50.

In the report this was showing 5x250= $1250...

Changed base_paid_amount to allocated_amount.

3 - it's not a problem, just for better view and wrong duplicate payment methods...

I did a select with group by for the 3 select inside.

### DIFF
--- a/erpnext/accounts/report/sales_payment_summary/sales_payment_summary.py
+++ b/erpnext/accounts/report/sales_payment_summary/sales_payment_summary.py
@@ -199,31 +199,39 @@ def get_mode_of_payment_details(filters):
 	invoice_list = get_invoices(filters)
 	invoice_list_names = ",".join('"' + invoice['name'] + '"' for invoice in invoice_list)
 	if invoice_list:
-		inv_mop_detail = frappe.db.sql("""select a.owner, a.posting_date,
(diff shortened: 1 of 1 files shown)


ankush:
@deepeshgarg007 

merged with lint errors?

HarryPaulo:
@ankush what meens lint erros? i can analyse and fix it, if had a error

ankush:
https://github.com/frappe/erpnext/runs/5555230851?check_suite_focus=true 

whitespace error as per "linter check"

This is a big problem but pollutes diff in future when someone else has to modify this file. 


You can check these locally by setting up pre-commit: https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

HarryPaulo:
i craeted a new pull request for "linter erros" 
https://github.com/frappe/erpnext/pull/30345

PR #30266, opened 2022-03-16
fix: Validate income/expense account in sales and purchase invoice

deepeshgarg007:
Validation for accounts added in the transaction is done when GL Entry is passed against them w.r.t a company.

In case of a deferred income/expense transaction entry against income/expense account is not passed during submission so if a user selects an incorrect account it won't be validated and later while booking deferred revenue/expense it fails because of the validation

Added a validation to validate income/expense account during save instead of on GL Entry creation


deepeshgarg007:

PR #30269, opened 2022-03-16
fix: Do not update ignore pricing rule check implicitly

deepeshgarg007:
Continuation of PR: https://github.com/frappe/erpnext/pull/30206

Removed missing updates


deepeshgarg007:

PR #30284, opened 2022-03-17
fix: Clean and fixes in Dimension-wise Accounts Balance Report

deepeshgarg007:
Dimension-wise account balance report was not showing any data because of the incorrect query.
Fixed the query and did an overall cleanup of the report


deepeshgarg007:
@Mergifyio backport develop version-13-pre-release

PR #30287, opened 2022-03-17
fix: custom cash flow mapper doesn't show any data

nextchamp-saqib:
- Convert `frappe.db.sql` to `frappe.qb`
- use `row.account` instead of `row.name`
   -  `row.name` was used as `account_names` to query `gl_entries`

### DIFF
--- a/erpnext/accounts/report/cash_flow/custom_cash_flow.py
+++ b/erpnext/accounts/report/cash_flow/custom_cash_flow.py
@@ -30,12 +30,12 @@ def get_mappers_from_db():
 def get_accounts_in_mappers(mapping_names):
 	return frappe.db.sql('''
 		select cfma.name, cfm.label, cfm.is_working_capital, cfm.is_income_tax_liability,
-		cfm.is_income_tax_expense, cfm.is_finance_cost, cfm.is_finance_cost_adjustment
+		cfm.is_income_tax_expense, cfm.is_finance_cost, cfm.is_finance_cost_adjustment, cfma.account
 		from `tabCash Flow Mapping Accounts` cfma
 		join `tabCash Flow Mapping` cfm on cfma.parent=cfm.name
 		where cfma.parent in (%s)
 		order by cfm.is_working_capital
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/cash_flow/custom_cash_flow.py
```diff
@@ -30,12 +30,12 @@ def get_mappers_from_db():
...
 	return frappe.db.sql('''
 		select cfma.name, cfm.label, cfm.is_working_capital, cfm.is_income_tax_liability,
-		cfm.is_income_tax_expense, cfm.is_finance_cost, cfm.is_finance_cost_adjustment
+		cfm.is_income_tax_expense, cfm.is_finance_cost, cfm.is_finance_cost_adjustment, cfma.account
 		from `tabCash Flow Mapping Accounts` cfma
 		join `tabCash Flow Mapping` cfm on cfma.parent=cfm.name
 		where cfma.parent in (%s)
 		order by cfm.is_working_capital
-	''', (', '.join('"%s"' % d for d in mapping_names)))
+	''', (', '.join('%s' % d for d in mapping_names)))
```
nextchamp-saqib, line 38: This prevents queries like `parent in (\"Temporary Opening\", )`

erpnext/accounts/report/cash_flow/custom_cash_flow.py
```diff
@@ -124,27 +124,27 @@ def setup_mappers(mappers):
 		)
 
 		for label in account_types_labels:
-			names = [d['name'] for d in account_types if d['label'] == label[0]]
+			names = [d['account_name'] for d in account_types if d['label'] == label[0]]
```
nextchamp-saqib, line 127: `names` are used to query the `gl entries`, it was originally `child_row.name` which now is changed to `child_row.account`

jigneshpshah:
@nextchamp-saqib  any idea when this will be merged to version-13 main branch  ?

nextchamp-saqib:
It should be updated on or before 30th March

PR #30304, opened 2022-03-18
fix: Add permission for KSA VAT documents

deepeshgarg007:
KSA Vat documents like KSA Vat Settings were not visible for users who already had existing companies for Saudi Arabia as in the earlier patch only the custom fields were created but the required permission to the docs was not given.

Added a new patch to give permission to those docs.


deepeshgarg007:

PR #30317, opened 2022-03-21
fix: P&L account validation on cancellation

deepeshgarg007:
Validation for Profit and Loss account in GL Entry was added around a year back.

Canceling opening entries posted before this validation with the P&L account raises a validation while posting reverse ledger entries and the user is unable to cancel the document. The modification in the condition will allow users to cancel the old documents


deepeshgarg007:

PR #30331, opened 2022-03-21
fix(ux): warning for disabled carry forwarding in Policy Assignment

ruchamahabal:
## Problem:
- While creating Leave Policy Assignment there is an option to add unused leaves from the previous allocation (carry forwarding)

   unused

- On submission, this creates allocations but does not carry forward leaves for leave types with "Is Carry Forward" unchecked. This happens silently so it confuses the user about why leaves weren't carried forward.

## Fix:

Warn the user about Leave Types which don't have carry forwarding enabled so that they can enable this (if needed) before submission:

image


ruchamahabal:
Unrelated failures

PR #30336, opened 2022-03-21
fix: Product Filters Lookup

marination:
## **Issue:**
- Filter search field had bad visibility and was not functional at all
   

- Field Filter Lookup
  

- Attribute Filter Lookup


marination:
all failures unrelated

PR #30118, opened 2022-03-08
fix: leave allocation records query

ruchamahabal:
**Problem**:

Syntax error in query due to missing field sanitization:

```Python
  SELECT
  	SUM(CASE WHEN is_carry_forward = 1 THEN leaves ELSE 0 END) as cf_leaves,
  	SUM(CASE WHEN is_carry_forward = 0 THEN leaves ELSE 0 END) as new_leaves,
  	MIN(from_date) as from_date,
  	MAX(to_date) as to_date,
  	leave_type
  FROM `tabLeave Ledger Entry`
  WHERE
  	from_date <= %(date)s
  	AND to_date >= %(date)s
  	AND docstatus=1
  	AND transaction_type="Leave Allocation"
  	AND employee=%(employee)s
  	AND is_expired=0
  	AND is_lwp=0
  	and leave_type='Special Leave - Husband's Death'
  GROUP BY employee, leave_type
	
Traceback (most recent call last):
  File "apps/frappe/frappe/app.py", line 68, in application
    response = frappe.api.handle()
  File "apps/frappe/frappe/api.py", line 55, in handle
    return frappe.handler.handle()
  File "apps/frappe/frappe/handler.py", line 31, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/handler.py", line 68, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "apps/frappe/frappe/__init__.py", line 1214, in call
    return fn(*args, **newargs)
  File "apps/frappe/frappe/__init__.py", line 630, in wrapper_fn
    retval = fn(*args, **get_newargs(fn, kwargs))
  File "apps/frappe/frappe/desk/query_report.py", line 244, in run
    result = generate_report_result(report, filters, user, custom_columns)
  File "apps/frappe/frappe/__init__.py", line 630, in wrapper_fn
    retval = fn(*args, **get_newargs(fn, kwargs))
  File "apps/frappe/frappe/desk/query_report.py", line 86, in generate_report_result
    res = get_report_result(report, filters) or []
  File "apps/frappe/frappe/desk/query_report.py", line 70, in get_report_result
    res = report.execute_script_report(filters)
  File "apps/frappe/frappe/core/doctype/report/report.py", line 133, in execute_script_report
    res = self.execute_module(filters)
  File "apps/frappe/frappe/core/doctype/report/report.py", line 150, in execute_module
    return frappe.get_attr(method_name)(frappe._dict(filters))
[...]

PR #30342, opened 2022-03-21
test: get leave allocation records query

ruchamahabal:
Added tests for the query changed in https://github.com/frappe/erpnext/pull/30118 as per this discussion: https://github.com/frappe/erpnext/pull/30118#issuecomment-1073801909


ruchamahabal:
CI failing due to the changes made in what `frappe.db.exists` returns: 

image

Probably this PR: https://github.com/frappe/frappe/pull/16200 (credits for the guess: @ankush)

**Before**:

image

**After**:

image

We can rectify the usage here, but many other apps might be dependent on this behavior. Thoughts? @surajshetty3416 @barredterra

barredterra:
@ruchamahabal good question, feel free to revert my PR for the time being.

I would, however, argue that somebody misused `frappe.db.exists` when they actually needed `frappe.get_all`. Here's the code in question:

https://github.com/frappe/erpnext/blob/01fd3adedf289aa18947310f06a1b6b38ba7ecd6/erpnext/crm/doctype/appointment/test_appointment.py#L10-L20

This code would actually improve with the new behavior because we could get rid of the `[0][0]` indexes.

ankush:
I agree that this is bad usage, even this API of `exists` should have never returned anything other than Boolean, but alas we can't break this because of 

https://www.hyrumslaw.com/ 

Let's limit this to v14 maybe?

barredterra:
@ankush I agree. Backporting was not intended anyways. I'll send a fix for mentioned issue.

ankush:
I'd propose keeping old behaviour that returned get_all style results but limit to 1 entry since that's what `exists` signifies, unlikely that anyone is using exists to do exactly what get_all does. (Fetching all rows is really stupid too 😆)



@gavindsouza @surajshetty3416

barredterra:
@ankush, the behavior was not consistent before, a mix of `get_value` and `get_all`. The goal of my PR was to keep possible return types simple and limited: `str or None`.

ankush:
keeping the breaking change and added note in https://github.com/frappe/frappe/wiki/Migrating-to-Version-14 (this needs to be updated with all breaking changes 🙈 )

Issue #29122, opened 2022-01-04
Taxes from Item Tax Templates does not get calculated and added to the document.


harshit-30:
### Information about bug

Steps to Replicate:
Item should not have any item tax template assigned.
Create a Sales Invoice/ Purchase Invoice/ Purchase Order/ Sales Order.
Select the item and edit row and add any Item Tax Template for that particular Item.
Even after Saving the Taxes are not added.

Expected Result:
Taxes should be calculated and added in the document.

Refer the screen recording attached.
![image](https://user-images.githubusercontent.com/46772424/148027987-88a4cadd-68c7-4714-b794-3e77706e1df4.png)

![image](https://user-images.githubusercontent.com/46772424/148028012-0c9ead9a-1187-4537-91ff-bb1da541ea52.png)

### Module

accounts

### Version

13

nitniyo:
Hey Team any update when will this be released ?

PR #30343, opened 2022-03-21
fix: Taxes not getting fetched from item tax template

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/29122


deepeshgarg007:
@Mergifyio backport develop

PR #30346, opened 2022-03-21
fix: usage of frappe.db.exists

barredterra:
Resolves https://github.com/frappe/erpnext/pull/30342#issuecomment-1074101391, intoduced by slight API change in https://github.com/frappe/frappe/pull/16200 (v14+).

Do not backport.

---
Sorry for the bulk formatting of `test_appointments.py`. It used spaces instead of tabs and my IDE is configured to not allow this.

### DIFF
--- a/erpnext/crm/doctype/appointment/appointment.py
+++ b/erpnext/crm/doctype/appointment/appointment.py
@@ -229,5 +229,5 @@ def _get_employee_from_user(user):
 		{'doctype': 'Employee', 'user_id': user})
 	if employee_docname:
 		# frappe.db.exists returns a tuple of a tuple
-		return frappe.get_doc('Employee', employee_docname[0][0])
+		return frappe.get_doc('Employee', employee_docname)
 	return None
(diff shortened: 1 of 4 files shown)


barredterra:
I didn't touch the code that the linter complains about and it would need a big refactor.

barredterra:
Failure of `TestSupplierScorecardVariable.test_path_exists` will be resolved with https://github.com/frappe/frappe/pull/16361

erpnext/crm/doctype/appointment/appointment.py
```diff
@@ -229,5 +229,5 @@ def _get_employee_from_user(user):
 		{'doctype': 'Employee', 'user_id': user})
 	if employee_docname:
 		# frappe.db.exists returns a tuple of a tuple
-		return frappe.get_doc('Employee', employee_docname[0][0])
```
ankush, line 232: we should replace `exists` call altogether, if we want to fetch value it makes more sense to use `get_value`.

ankush:
I can't push the changes. Can you enable maintainer edits? 

or here's the patch: https://gist.github.com/ankush/2e3697c7d675c5c57a93b002a3f4bf0a

ankush:
moved to https://github.com/frappe/erpnext/pull/30350

PR #30348, opened 2022-03-22
feat: Subscription option to apply advance payment received in sales invoice generated

man4ever84:
Subscription doctype currently does not have a mechanism to automatically account for any prepayment when a new invoice is created.

Proposed change : Add a "Apply Advance Payment Automatically" checkbox in the Subscription doctype, and add logic to call set_advance() in Subscription.py when invoice is being created.  The invoice generated will then apply any advance payment made accordingly and show the correct outstanding amount.
 
Before
image

After
image


deepeshgarg007:
@man4ever84 Can you add a test case for this scenario

man4ever84:
> @man4ever84 Can you add a test case for this scenario

Yes I'll try to figure out how to create one this weekend.  This is my 1st time contributing so there is a lot to learn.

deepeshgarg007:
@man4ever84 I think your branch is not rebased properly. Please rebase and then open a new pull request

PR #30349, opened 2022-03-22
refactor: minor setup cleanup

ankush:
- remove dead and duplicate code related to sample docs.
- use current year for FY generation in `before_test` (this avoid boilerplate fixture generation in UI tests)


ankush:
tested new site setup locally.

The code doesn't appear in coverage because it runs before tests start.

PR #30356, opened 2022-03-22
fix: ignore already fetched serial no

ankush:
exclude_sr_nos is sent as JSON string of list, so load it before operating on it.

telegram-cloud-photo-size-5-6183807283746681146-y

### DIFF
--- a/erpnext/stock/doctype/serial_no/serial_no.py
+++ b/erpnext/stock/doctype/serial_no/serial_no.py
@@ -3,6 +3,7 @@
 
 
 import json
+from typing import List, Optional, Union
 
 import frappe
 from frappe import ValidationError, _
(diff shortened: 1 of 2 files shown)


marination (dismissed):

marination:
tested, works fine

ankush:
telegram-cloud-photo-size-5-6183807283746681370-y


This is bit flaky for some reason, will investigate.

query when this fails. Batch filter isn't being applied at all. 
```sql

SELECT `tabSerial No`.`name`,`tabSerial No`.`batch_no`,`tabBatch`.`name` FROM `tabSerial No` LEFT JOIN `tabBatch` ON `tabSerial No`.`batch_no`=`tabBatch`.`name` WHERE `tabSerial No`.`item_code`='844275405428b582' AND `tabSerial No`.`warehouse`='_Test Warehouse - _TC' AND COALESCE(`tabSerial No`.`sales_invoice`,'')='' AND COALESCE(`tabSerial No`.`delivery_document_no`,'')='' AND COALESCE(`tabBatch`.`expiry_date`,'4000-12-31')>='2021-01-01' ORDER BY `tabSerial No`.`creation` LIMIT 5
```


This happens when batch nos are integers and `safe_json_loads()` makes it int. 😆

PR #30358, opened 2022-03-22
fix: flaky salary slip email test

ankush:
`via_payroll_entry` a global flag is set while doing payroll entry bulk submission to avoid sending email, this is however never unset after it's done. 

Long term solution: need to change `FrappeTestCase` so flags are reset across different test runs.


ruchamahabal (approved):

ruchamahabal:
Unrelated stock test failures. Merging

PR #30363, opened 2022-03-22
Revert "fix: Do not update ignore pricing rule check implicitly"

marination:
Reverts frappe/erpnext#30269

> Part 2 of https://github.com/frappe/erpnext/pull/30206

Mapped doc rates are unset due to this change
**case:**
- Item A has buying Item Price 100
- User changes price to 150 from 100 in PO. Submit PO.
- Create Pr from PO. Rate is set as 100 (item price). PO rate is fully ignored
- This is due to `apply_price_list` that resets rates as per price list in `transaction.js`. This portion is run as `ignore_pricing_rule` is set to 0 by the linked PR
  Screenshot 2022-03-22 at 8 15 29 PM

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -1667,6 +1667,7 @@ def make_maintenance_schedule(source_name, target_doc=None):
 @frappe.whitelist()
 def make_delivery_note(source_name, target_doc=None):
 	def set_missing_values(source, target):
+		target.ignore_pricing_rule = 1
 		target.run_method("set_missing_values")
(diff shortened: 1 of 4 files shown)


melBoh:
Hi ! 
I see the PR was closed ? does it mean it'll never be reverted ? 
If so, can I pull it one way or another nonetheless ? It's critical for me
Thank you

marination:
@melBoh this was closed because https://github.com/frappe/erpnext/pull/30365 was added as a deeper fix

melBoh:
Hi ! I've tried it again, and sadly it's not working for me
I opened an issue https://github.com/frappe/erpnext/issues/30986
If you can take a look, it'd be very much appreciated.
Thank you

PR #30365, opened 2022-03-22
fix: Changing item prices on converting orders/receipts to invoices

deepeshgarg007:
Change add in PR frappe/erpnext#30269

> Part 2 of https://github.com/frappe/erpnext/pull/30206

Mapped doc rates are unset due to this change
**case:**
- Item A has buying Item Price 100
- User changes price to 150 from 100 in PO. Submit PO.
- Create Pr from PO. Rate is set as 100 (item price). PO rate is fully ignored
- This is due to `apply_price_list` that resets rates as per price list in `transaction.js`. This portion is run as `ignore_pricing_rule` is set to 0 by the linked PR
  Screenshot 2022-03-22 at 8 15 29 PM

Added an onload check to ignore price list trigger on converting order/receipt to invoice


deepeshgarg007:

Issue #30392, opened 2022-03-24
Item Rate changes after saving


kanedai:
### Information about bug

When I make a quote, modify the rate, save it, and the Rate will change to another value (this problem in the early V13 in 2021)
<img width="320" alt="image" src="https://user-images.githubusercontent.com/4647923/159829582-c7233b35-344a-4a5e-a273-200e69b6f986.png">
<img width="309" alt="image" src="https://user-images.githubusercontent.com/4647923/159829631-1534ddad-762d-48b6-a261-3c5b6cac0d28.png">

### Module

accounts

### Version

Frappe - 13.23.0
Erpnext - 13.23.1 (latest version)

backendoadmin:
Duplicated:
https://github.com/frappe/erpnext/issues/30375

aakvatech:
![image](https://user-images.githubusercontent.com/35020381/159976815-78b26d3a-d4d4-4895-b472-fa81c4fdb76f.png)

discount_percentage is calculated by (discount_amount / price_list_rate) and applying the precision (usually 2 or 3 decimals) = rounded discount_percentage
Then item rate is calculated by calculating an incorrect discount_amount from the (price_list_rate * rounded discount_percentage) and then deducting the incorrect discount_amount from price_list_rate

backendoadmin:
I think its fixed with https://github.com/frappe/erpnext/pull/30406

PR #30393, opened 2022-03-24
fix: 30392 Item Rate changes after saving #30392

szufisher:

> Screenshots/GIFs

reproducible steps

1. item price for selling is 1000
2. create quotation for the item, change rate to 990 , save
3. change quotation item rate to other value, let us say 980, save, system will change rate back to 990, it is caused by the wrong discount amount.

this issue impacted user's daily operation, hot fix to be released as soon as possible.


deepeshgarg007:
@szufisher please explain the entire scenario when this issue happens

szufisher:
> @szufisher please explain the entire scenario when this issue happens

updated description with reproducible steps.

PR #30395, opened 2022-03-24
fix: Write off amount wrongly calculated in POS Invoice

deepeshgarg007:
Currently, the change amount is dependent on the write-off amount and the write-off amount is dependent on the change amount. The write-off is auto-calculated if there is a change amount. All the behaviors leads to incorrect calculation of write-off amount in some cases as both the amounts are interdependent on each other.

Write off amount should not be dependent on any field since its a user input field and the only way to auto calculate the write off amount is when it has to be equal to the outstanding amount

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/test_sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/test_sales_invoice.py
@@ -812,12 +812,37 @@ class TestSalesInvoice(unittest.TestCase):
 		pos.append("payments", {'mode_of_payment': 'Bank Draft', 'account': '_Test Bank - TCP1', 'amount': 50})
 		pos.append("payments", {'mode_of_payment': 'Cash', 'account': 'Cash - TCP1', 'amount': 60})
 
-		pos.change_amount = 5.0
+		pos.write_off_outstanding_amount_automatically = 1
 		pos.insert()
 		pos.submit()
 
 		self.assertEqual(pos.grand_total, 100.0)
-		self.assertEqual(pos.write_off_amount, -5)
+		self.assertEqual(pos.write_off_amount, 0)
+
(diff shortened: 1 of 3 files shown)


deepeshgarg007:
@Mergifyio backport develop

deepeshgarg007:
@Mergifyio backport develop

HarryPaulo:
@deepeshgarg007 hello!! with this change, write off is not working properly.

look at my examples:
I made a sales with grand total $ 9,5

Case 1 - If customer pay to me $10, automatically field **Change Amount** receive $0,5.
If customer says I don't need to give change, the i set ZERO to field **Change Amount** and **write off** must be setted $-0,5


In the example I just presented, it doesn't matter if the **write_off_outstanding_amount_automatically** check is checked, it must assign as write off the amount removed from the change.

Case 2 - With **write_off_outstanding_amount_automatically** checked:

Using the same sale, the Customer would have to pay me $9.5 (outstanding amount), but he says he only has $9 and asks if I would sell the product for that amount, if I agree:
I inform him that I received $9, mark the check **write_off_outstanding_amount_automatically** and automatically the diference amount I should receive will be entered as a write off  = $0,5.


With the change made, the write off is not behaving as per the two cases I pointed out above.

HarryPaulo:
@deepeshgarg007 can u explain why this changes was made? 

**Actually**:

Example:
Sales: $9,90
Change Amount: $0,1

1. if i remove the $0,1 from change amount, on save system set change amount again.
2. if i set "write off" without remove "Change Amount", if i try to remove change amount system cannot remove, keeping Write Off and Change Amount

**How I imagine it should be**
Change amount and Write Off work together, How it need to work:

1. If I enter Write Off, the amount should be removed from the "Change Amount"
2. If I reduce the change amount, the reduction amount should be set in the write off.

PR #30399, opened 2022-03-24
fix: broken production item links on production plan

ankush:
Production Plan tables for po_items and sub_assembly_items are prepared client-side so both don't exist at the time of first save or modifying.

Any "links" created are invalid (`production_plan_item`). This change retains the temporary name so it can be relinked server-side after naming is performed.

Steps to reproduce:

1. Start production plan.
2. Add PO items
3. fetch subassembly items
4. save
5. subassembly item rows will have `production_plan_item` that's temp name like `new-production-plan...`


marination:
@ankush Tested this works as expected. LGTM

marination (approved):

Issue #30394, opened 2022-03-24
Item rate changes on save


rtdany10:
### Information about bug

Item rate gets changed upon saving.
Occurring due to rate being recalculated from price list using discount percentage.

https://user-images.githubusercontent.com/52111700/159854484-90af7669-d1e8-4b0a-a253-e4a41d70a8ca.mp4

### Module

accounts

### Version

ERPNext: v13.23.0 (HEAD)
Frappe Framework: v13.23.0 (HEAD)

### Installation method

FrappeCloud

szufisher:
same as this one https://github.com/frappe/erpnext/issues/30392, pull request under review

rtdany10:
> same as this one #30392, pull request under review

Can you link the PR here, please?

@ankush @deepeshgarg007 Please look into this. Invoicing has become an issue.

szufisher:
https://github.com/frappe/erpnext/pull/30393

rtdany10:
> #30393

I doubt this. 

My issue arises due to rate being recalculated from price list using discount percentage. Discount percentage has limited precision, therefore, when the rate is recalculated by subtracting the discount percentage from price list rate, it goes wrong.

szufisher:
@rtdany10  you are right, the PR is part of the solution, we also encountered the discount percentage precision issue, the workaround solution is to change the discount percentage precision to 9 via customize form. hopefully the core team developer can  find a permanent fix to address this critical issue.

deepeshgarg007:
@rtdany10 Can you let me know the exact steps to follow to replicate the issue. I am not able to replicate the issue

rtdany10:
> @rtdany10 Can you let me know the exact steps to follow to replicate the issue. I am not able to replicate the issue

@deepeshgarg007 Create an item, set price list rate, sell that item at a lower rate. Discount percentage will be calculated based on that, and upon saving, rate is recalculated from price list rate.

deepeshgarg007:
> > @rtdany10 Can you let me know the exact steps to follow to replicate the issue. I am not able to replicate the issue
> 
> @deepeshgarg007 Create an item, set price list rate, sell that item at a lower rate. Discount percentage will be calculated based on that, and upon saving, rate is recalculated from price list rate.

Yes, I tried the same no issues. Maybe you have some other precision, some other setting
![Price List](https://user-images.githubusercontent.com/42651287/159945647-5cfe0337-6a20-4f17-bf25-a75f1873f3ac.gif)

[...]

PR #30406, opened 2022-03-25
fix: Rate change issue on save and mapping from other doc

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/30394


deepeshgarg007:
@Mergifyio backport develop version-13-pre-release version-13

PR #30419, opened 2022-03-25
fix: move item tax to item tax template patch

nextchamp-saqib:
When Item Tax is converted into Item Tax Template while migrating to v12, the new Item Tax Templates are created. If another item has the same Item Taxes, the same newly created template is used, however, the template name was incorrectly stored to check for duplicates. 

If a template has already been created is checked with the following lines and returns the `template name`
https://github.com/frappe/erpnext/blob/33863cd6abd9c42ae5989fed9c54a3234e076cb6/erpnext/patches/v12_0/move_item_tax_to_item_tax_template.py#L86-L88

But later the `dict` is updated with the `title` of the `template`
https://github.com/frappe/erpnext/blob/33863cd6abd9c42ae5989fed9c54a3234e076cb6/erpnext/patches/v12_0/move_item_tax_to_item_tax_template.py#L136-L137

---
The problem only exists in v13 as the `autoname` of `Item Tax Template` has been modified in v13


nextchamp-saqib:
CleanShot 2022-03-28 at 17 33 51@2x

It shouldn't break

PR #30420, opened 2022-03-25
fix: unsupported operand type(s) for +=: 'int' and 'NoneType'

nextchamp-saqib:


erpnext/accounts/report/cash_flow/custom_cash_flow.py
```diff
@@ -435,15 +435,15 @@ def _get_account_type_based_data(filters, account_names, period_list, accumulate
 				(GLEntry.voucher_type != 'Period Closing Voucher') &
 				(GLEntry.account.isin(account_subquery))
 			)
-		).run()
+		).run(pluck=True)
```
ankush: this wont be portable on v13, pluck only exists on v14 😐

PR #30438, opened 2022-03-27
fix: Incorrect default amount to pay for POS invoices

deepeshgarg007:
Bug introduced in https://github.com/frappe/erpnext/pull/30181

The `calculate_shipping_charge` function is an async function. Once the shipping charges are applied taxes are calculated again and this async behavior was causing incorrect calculation of the total amount to pay and in turn left the invoice as overdue


deepeshgarg007:
@Mergifyio backport develop

PR #30445, opened 2022-03-27
refactor: barcode scanning in stock transactions

ankush:
### Refactors for barcode scanner

- Decouple all barcode scanner code from `transaction.js` to `erpnext.utils.BarcodeScanner` No idea why this was in `transaction.js` 🤷 
- Don't **hardcode ANYTHING**, you can now specify the field name of the child table, and child table fields if required when using the barcode scanner on transactions that don't follow "conventional" naming. 
- Reduce toast duration from 7 sec to 5 sec (should we do 3 instead?) When scanning fast these toasts become a nuisance. 
- If a serial number already exists in a row then don't re-add it (duh)
- Trigger `table_add` event after adding a new child item (to auto-populate defaults like a warehouse) This wasn't working before because of the missing trigger when adding row via scanner. Check if default warehouses are propagating in new rows. (careful review required for this bit)
- Split function to set item values into separate functions. Rationale: A loop that sets all key:value pairs but checks if key is `serial_no` and does something else is not a sign of good maintainable code 😅 
- Reverted this fix as it didn't seem to work and added a specific attribute on the item row instead: https://github.com/frappe/erpnext/pull/16806 (need check for side effects if any)


szufisher:
is it possible also support barcode for different UOM, e.g item has default UOM as PC, there is barcode for UOM Box, when scan the box barcode, in items table, both the quantity and UOM will be handled correctly.

ankush:
@szufisher good idea, adding another field for UOM in barcode table should be enough I think. Conversions are already handled by stock transactions.


However this won't work with batch numbers.

ankush:
Split the PR to avoid scope creep, this is already way past its original scope 😅 

Stock reco scanning: https://github.com/frappe/erpnext/pull/30446

PR #30446, opened 2022-03-28
feat: Barcode scanning in Stock Reconciliation

### LINKED ISSUE #9557: Barcode scanning in stock reconciliation
Is it me or we can't use the barcode scanner in the stock reconciliation form? Isn't it a feature that would be useful? Would be a lot quicker than having to type each item codes.

Thanks!

G.

ankush:
### Barcode scanning in stock reconciliation 

During some warehouse ops you might want to reconcile stock just by scanning directly into ERP instead of noting it down separately. This flow would go like this: 

1. Create a new stock reconciliation. 
2. Select warehouse and type of reconciliation. 
3. Select default warehouse (new field/feature)
4. Start scanning items one by one, it will keep counting items in each rows and if new items are encountered, a new row is added (existing functionality from barcode scanner)

The main problem in stock reco was auto fetching of quantity and serial nos. This is addressed by introducing "scan mode". 

Enable "scan mode" before using a barcode scanner for stock reconciliation. Once enabled, ERPNext won't fetch existing quantity or serial nos and you'll need to keep incrementing/adding them using a scanner. (UX review required for this part)

In action: 

https://user-images.githubusercontent.com/9079960/160292580-22093089-3146-4879-be93-5e73b4ad1bd2.mov

depends on: https://github.com/frappe/erpnext/pull/30445 (hard dependency)
depends on: https://github.com/frappe/erpnext/pull/30516
closes https://github.com/frappe/erpnext/issues/9557 

TODO:

- [x] Docs  https://docs.erpnext.com/docs/v14/user/manual/en/stock/stock-reconciliation


ankush:
Removed auto enabling of "scan mode" on the first scan after discussing with @marination it's kinda implicit and those who want this always enabled can anyway do it from Customize Form.

ankush:
removed auto enabling of "scan mode" on first scan after discussing with @marination

dj12djdjs:
I see in the spec it is setup to pull the default warehouse. Any thoughts about handling the following scenarios:

1. If the item has multiple warehouses for the specific company.
       - Maybe a second stage after scanning the item to select the warehouse?
 2. Scanning an Item that is thought to be out of stock, having the behaviour automatically pull default warehouse could result in counter recording the stock into the wrong location. Counter should specify the location where item is found.

I think most warehouses will have a barcode on their "Warehouse"/"Bin" that would make input of warehouse much more simple if counter scanned the item, then scanned the bin. 

What do you think?

ankush:
> If the item has multiple warehouses for the specific company.
> - Maybe a second stage after scanning the item to select the warehouse?


You can still do this, don't select a default warehouse, and after the scanning process add it manually. Though it would be much better to do it warehouse-wise i.e. you reconcile one warehouse in one stock reconciliation, save and submit, move on to the next warehouse. 


> Scanning an Item that is thought to be out of stock, having the behaviour automatically pull default warehouse could result in counter recording the stock into the wrong location. Counter should specify the location where item is found

I don't follow the reasoning here, the "default warehouse" field is set by user, NOT pulled from item's defaults. 

> I think most warehouses will have a barcode on their "Warehouse"/"Bin" that would make input of warehouse much more simple if counter scanned the item, then scanned the bin.

Hmm, this would require shifting input manually or programming the focus to move from "scan barcode" field to warehouse field 😅

dj12djdjs:
@ankush You're right, when I initially looked at the PR last night I didn't even notice the "Default Warehouse" field. I thought it was going to be pulled from the Item master default warehouse.

So in this case, it could easily be done by user customizations to allow user to scan the bin they are at, then it would fill default warehouse field, then processed to scan item and enter qty.

marination:
@ankush everything else looks aight except Items with Batch and Serial No's UX.

The qty comes as 0 after scanning (i think due to missing batch?), when batch is selected qty appears. 

In the older cut it fetched total qty irrespective of batch, and then when user selects a batch, the batch qty is overwritten.

The batch selection bit is not very intuitive is it ?

ankush:
changes wrt original PR:
1. Fixed all interactions with serial selector AND auto-populating of item, sr, batch details. 
2. Fixed issue where some other trigger was overriding data from scan. 

Pending issue:
1. In non-scan mode fetched serials don't respect selected batch no. (this isn't a shallow fix, will spin it off in separate PR, temp workaround added for now.)

johnlongland:
Good day Ankush. Is there an expected date of release for this ? Very excited to see/try/check this !

ankush:
>  Good day Ankush. Is there an expected date of release for this ? Very excited to see/try/check this !



Hi, @johnlongland most new features will now only land in v14 directly.

PR #30449, opened 2022-03-28
fix: misc setup fixes

ankush:
- Disable currency exchange validation: This reduces ~4-5 seconds of setup time and chances of setup failure. It's implicitly ignored anyway, not sure why we attempt to do it in first place. 
- Explicitly ignore mandatory fields during single updates (This should really be handled by the framework)


ankush:
Screenshot 2022-03-28 at 2 16 38 PM


🤔 

maybe we can add tests for everything that's expected to be there after setup

PR #30461, opened 2022-03-29
fix(asset): do not validate warehouse on asset purchase

nextchamp-saqib:
If the default warehouse is not set for the company, and the user tries to create a Purchase Invoice with Update Stock for a purchase of an Asset, then the Warehouse Missing error is raised which doesn't make sense since Asset is a non-stock item.


erpnext/accounts/doctype/purchase_invoice/purchase_invoice.py
```diff
@@ -250,7 +250,7 @@ def validate_with_previous_doc(self):
 	def validate_warehouse(self, for_validate=True):
 		if self.update_stock and for_validate:
 			for d in self.get("items"):
-				if not d.warehouse:
+				if not d.warehouse and not d.is_fixed_asset:
```
ankush: Expand this to the general case? Check if item `is_stock_item` ?

PR #30466, opened 2022-03-29
feat(purchase order/sales order): allow update item in Update Item bu…

BevanTony1:
Feature Details:
--
Allow Sales Manager users to change the item_code in Purchase Order and Sales Order using the "Update Item" button.

Users should be able to update the items in the Sales Order/Purchase Order. By updating the item_code directly, the current process for updating submitted SO/PO is deleting the item and using the add button.

Purchase Order:
--
Purchase Order (submitted)

Sales Order:
--
Sales Order (Submitted)

Validations:
--
Only users with the Sales Manager role are only ones allowed to perform the update.
Non-salesmanager

if item line is partially billed or delivered. Prevent item code from being updated.

**Purchase Order**
Purchase Order (billed)
**Sales Order**
Sales Order (billed)

**Add test Case**
https://github.com/frappe/erpnext/pull/30147

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -2044,6 +2044,7 @@ def update_child_qty_rate(parent_doctype, trans_items, parent_doctype_name, chil
 			check_doc_permissions(parent, 'write')
 			child_item = frappe.get_doc(parent_doctype + ' Item', d.get("docname"))
 
+			prev_item_code, new_item_code = child_item.item_code,d.get('item_code')
 			prev_rate, new_rate = flt(child_item.get("rate")), flt(d.get("rate"))
 			prev_qty, new_qty = flt(child_item.get("qty")), flt(d.get("qty"))
 			prev_con_fac, new_con_fac = flt(child_item.get("conversion_factor")), flt(d.get("conversion_factor"))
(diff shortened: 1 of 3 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -2069,6 +2071,23 @@ def validate_quantity(child_item, d):
...
+						frappe.throw(_("Cannot update item: partial bill or partial delivery exists."))
+					else:
+						item_data = frappe.get_doc("Item", d.get("item_code"))
+						child_item.item_code = d.get("item_code")
+						child_item.item_name = item_data.item_name
+						child_item.description = item_data.description
+						child_item.stock_uom = item_data.stock_uom
+						child_item.uom = item_data.stock_uom
+			else:
+				frappe.throw(_("User must be Sales Manager to update Item Code."))
```
ankush, line 2089: - We do not hardcode roles like this in code. Only Submit permission should be checked IMO since this is updating a submitted document. 
- Also same code runs for the purchase cycle where the Sales manager role doesn't make sense.

ankush (changes requested):
Still some concerns that were mentioned in the previous PR: https://github.com/frappe/erpnext/pull/30147#issuecomment-1073891481 

Need to look at a lot of code that will be affected by SO/PO item codes changing without with same `child_row.name`.

PR #30476, opened 2022-03-29
fix: validate 0 transfer qty in stock entry

ankush:
If qty and conversion factors are close to zero then stock qty becomes 0. This is usually incorrect data but currently, there's no good error message and it fails during SLE submission. 

Screenshot 2022-03-29 at 4 33 40 PM


ankush:

PR #30486, opened 2022-03-30
fix: Account currency validation

### LINKED ISSUE #24477: Clearing account currency sets it to company currency.
## Reproduce

1. Create an **Account** and set the "Currency" field to USD.
2. Make a **GL Entry** against that **Account**.
3. Open the **Account**, clear its "Currency" field and save.

## Expected behaviour

The "Currency" field cannot change if there is an existing **GL Entry**.

## Problem

The "Currency" field gets set to company currency. It is not possible to change it afterwards.

## Possible solution

Swap the if-statements here:

https://github.com/frappe/erpnext/blob/907e2990d064957b336717a77d2013ac5718c564/erpnext/accounts/doctype/account/account.py#L157-L163

---

See [discuss](https://discuss.erpnext.com/t/multi-currency-accounting-changing-currency/70588?u=rmeyer) for the original report.

deepeshgarg007:
fixes: https://github.com/frappe/erpnext/issues/24477

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -204,7 +204,9 @@ class Account(NestedSet):
 		if not self.account_currency:
 			self.account_currency = frappe.get_cached_value("Company", self.company, "default_currency")
 
-		elif self.account_currency != frappe.db.get_value("Account", self.name, "account_currency"):
+		gl_currency = frappe.db.get_value("GL Entry", {"account": self.name}, "account_currency")
+
+		if gl_currency and self.account_currency != gl_currency:
 			if frappe.db.get_value("GL Entry", {"account": self.name}):
 				frappe.throw(_("Currency can not be changed after making entries using some other currency"))
 
--- a/erpnext/accounts/doctype/account/test_account.py
+++ b/erpnext/accounts/doctype/account/test_account.py
@@ -264,6 +264,72 @@ class TestAccount(unittest.TestCase):
 		self.assertRaises(frappe.ValidationError, acc.save)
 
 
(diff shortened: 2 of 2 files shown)


deepeshgarg007:

erpnext/accounts/doctype/account/account.py
```diff
@@ -204,7 +204,9 @@ def validate_account_currency(self):
 		if not self.account_currency:
 			self.account_currency = frappe.get_cached_value("Company", self.company, "default_currency")
 
-		elif self.account_currency != frappe.db.get_value("Account", self.name, "account_currency"):
+		gl_currency = frappe.db.get_value("GL Entry", {"account": self.name}, "account_currency")
+
+		if gl_currency and self.account_currency != gl_currency:
 			if frappe.db.get_value("GL Entry", {"account": self.name}):
```
barredterra, line 210: This line became redundant (If `gl_currency` has a value, the GL Entry likely exists).

erpnext/accounts/doctype/account/test_account.py
```diff
@@ -241,6 +241,28 @@ def test_child_company_account_rename_sync(self):
 		for doc in to_delete:
 			frappe.delete_doc("Account", doc)
 
+	def test_validate_account_currency(self):
```
barredterra, line 244: Correct me if I'm wrong, but it seems like this test does not cover the problem described in the issue.

barredterra (commented):
Thanks for fixing! Found two minor issues, maybe you'd like to address them :)

(As always, I'm not a big fan of merging without reviews.)

PR #30488, opened 2022-03-30
fix: web form filters for project

ahmadpak:
image

Filters on web form list view are not working from customer's login.

On selection of Status→ Completed the web form is not getting reflected & same with other fields also


deepeshgarg007:
@ahmadpak can you add a gif showing the current state and a basic test case for this if possible

deepeshgarg007:
@ahmadpak can you add a gif showing the current state and a basic test case for this if possible

ahmadpak:
Hey @deepeshgarg007 I am afraid I might not be able to do that as I am not working on this issue at the moment. If that is a deal breaker for the PR let me know I will try to share the gif

PR #30494, opened 2022-03-30
fix: update `get_cached_value` usage based on changes in definition

sagarvora:
depends on https://github.com/frappe/frappe/pull/16454

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -375,12 +375,12 @@ def get_pricing_rule_for_item(args, price_list_rate=0, doc=None, for_validate=Fa
 
 def update_args_for_pricing_rule(args):
 	if not (args.item_group and args.brand):
-		try:
-			args.item_group, args.brand = frappe.get_cached_value(
-				"Item", args.item_code, ["item_group", "brand"]
-			)
-		except frappe.DoesNotExistError:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -375,12 +375,12 @@ def get_pricing_rule_for_item(args, price_list_rate=0, doc=None, for_validate=Fa
 
 def update_args_for_pricing_rule(args):
 	if not (args.item_group and args.brand):
-		try:
-			args.item_group, args.brand = frappe.get_cached_value(
-				"Item", args.item_code, ["item_group", "brand"]
-			)
-		except frappe.DoesNotExistError:
+		item = frappe.get_cached_value("Item", args.item_code, ("item_group", "brand"))
```
ankush, line 378: This is another thing, that's been bothering me. The shape of return could be `[None, None]` if not found? 🤔 

just a thought. Not in scope of this PR or parent one.

ankush (approved):

erpnext/accounts/doctype/pricing_rule/pricing_rule.py
ankush: Eh, probably bad idea since the return value will become truthy then.

sagarvora: I made same comment in parent PR 😄

PR #30497, opened 2022-03-30
fix: remove naming series from bin, repost queue

ankush:
Removing naming series from:
1. Bin
2. Repost queue

These doctypes are not user facing and dont really need naming series.
Current implementation of naming makes stock transaction sequential if
these documents are to be created during submission.

Note: If users want it they can get it back by adding naming series from "customize form".

Don't backport.


ankush:
failures unrelated. AFAIK we don't use names of these doctypes anywhere. 

also happening in other PRs: https://github.com/frappe/erpnext/runs/5755472751?check_suite_focus=true 

will investigate later 🥲

PR #30499, opened 2022-03-30
feat: 'customer' column and more filter to Payment terms status report

ruthra-kumar:
## Change
New field 'Customer' has been added to 'Payment Terms Status for Sales Order' report output. Additional filters Customer, Customer Group, Item and Item group have been added as well.

Screenshot 2022-03-30 at 6 51 52 PM

Screenshot 2022-03-31 at 9 41 32 AM

**no-docs**

### DIFF
--- a/erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
+++ b/erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
@@ -73,6 +73,55 @@ def get_columns():
 	return columns
 
 
+def get_descendants_of(doctype, group_name):
+	group_doc = qb.DocType(doctype)
+	# get lft and rgt of group node
+	lft, rgt = (
+		qb.from_(group_doc).select(group_doc.lft, group_doc.rgt).where(group_doc.name == group_name)
+	).run()[0]
+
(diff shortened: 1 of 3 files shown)


erpnext/selling/report/payment_terms_status_for_sales_order/payment_terms_status_for_sales_order.py
```diff
@@ -92,16 +116,29 @@ def get_so_with_invoices(filters):
 
 	so = qb.DocType("Sales Order")
 	ps = qb.DocType("Payment Schedule")
+	cust = qb.DocType("Customer")
```
deepeshgarg007: Why do we need to do a join with the `Customer` and `Item` table, can't this information be retrieved from the `Sales Order` and `Sales Order Item Table` respectively?

ruthra-kumar: oh, I wasn't aware that these are available in `Sales Order` and `Sales Order Item`. I'll remove it.

deepeshgarg007:

OMNERP:
why isn't this report showing in the reports section of selling module? It is accessible via search bar and report list only.

PR #30500, opened 2022-03-30
fix(India): Tax fetching based on tax category

deepeshgarg007:
Taxes based on templates linked via tax category was not getting fetched properly. This was replicable only for Indian companies as this was conflicting with regional tax fetching logic for India.

Scenario when the issue is replicated
1. Make a customer with GST address and GSTIN
2. Add tax category to customer master and sales taxes and charges template
3. Try creating an invoice against that customer


deepeshgarg007:

Issue #29595, opened 2022-02-02
Unallocated amount is inaccurate on multi-currency Payment Entry


Archais:
### Information about bug

The issue is that when you are processing a payment entry in a currency other than base, the unallocated_amount can end up being inaccurate, especially with higher exchange rates, e.g. 210 : 1.

In my example of the issue we are running a base currency of USD and we are receipting payment in ZWL. The customer has paid us ZWL$1,160,250.00 at an exchange rate of 210:1 (represented in ERPNext as 0.004761905) which evaluates to USD$5,525.00. This part is accurate, however because ERPNext does the allocations in base currency then converts back to the selected currency to give you your unallocated_amount, we get a value of ZWL$1,160,249.94. So we are losing ZWL$0.06 on the transaction but there is no USD difference because when converted it is less than USD$0.01. Thus there is a loss in value of the customer's money without the value going anywhere.

### Module

accounts

### Version

Frappe version - 13.19.0
ERPNext version - 13.19.0

### Installation method

manual install

fproldan:
👍🏻

Issue #23936, opened 2020-11-17
Unallocated amount is incorrectly calculated in Payment Entry when paying an Invoice in Non Base currency


palashjhabak:
## Description of the issue
When a PI / SI is raised in a non Base currency (lets say USD) and we try to pay it through an account in the same none base currency (USD), the unallocated amount is wrongly calculated. Let me try to explain it in detail with an example. Lets say I have a supplier who deals in USD. The Supplier has raised me an invoice of 5000 USD and at the time book booking exchange rate w.r.t my currency was lets say 16. And at a later date I make a manual payment entry with an amount of 10000 USD ( paying from an account with currency USD ) and the exchange rate on that day was 15 which was entered in the PE for. Now unallocated amount in this scenario should be 5000 USD but the current implementation calculates it as 4333 USD. Now it calculates it incorrectly because it adds up all the allocated amount first in the base currency with exchange rate present in the PI and then calculates the unallocated amount in base currency and converts it back to USD with the exchange rate of PE but in reality, it shouldnt matter at all as the unallocated amount should just be 5000 USD as we are paying in USD to a PI raised in USD

## Context information (for bug reports)

**Output of `bench version`**
```
(paste here)
```

## Steps to reproduce the issue

1.
2.
3.

### Stacktrace / full error message

```
(paste here)
```

## Additional information

OS version / distribution, `ERPNext` install method, etc.

fproldan:
@palashjhabak Hi, were you able to fix it?

palashjhabak:
@fproldan its been a long while since I raised the issue so don't exactly remember but I will check if I did something in my custom app to fix it and let you know. 

PS: apologies for such a late reply

dj12djdjs:
https://github.com/frappe/erpnext/issues/34680

dj12djdjs:
Another example:

1. Setup company with base currency CAD
2. Create a Purchase Invoice at $100.00 value of USD currency using 1.3417 exchange rate.
3. Create Payment Entry of $1000 USD paid $1360.00 CAD at 1.360 exchange rate.
4. Allocated $100 of the $1000 Payment Entry against the $100.00 Purchase Invoice.

**Expected Unallocated Amount:** $900 USD
**Observed Unallocated Amount:** $901.35 USD

ruthra-kumar:
This issue should be fixed post the multi-currency refactor in v14 and above, https://github.com/frappe/erpnext/issues/34680.

PR #30504, opened 2022-03-30
fix(accounts): payment entry posts incorrect amt to party ledger

dj12djdjs:
Closes:
- https://github.com/frappe/erpnext/issues/29595
- https://github.com/frappe/erpnext/issues/23936

### Problem Statement
Before this PR, `unallocated_amount` for a `Payment Entry` of type `Pay` would be calculated by the following formula:
(`base_paid_amount` + `base_total_taxes_and_charges` - (`total_deductions` + `base_total_allocated_amount`)) / `target_exchange_rate`

Given the following values the above formula is incorrect, resulting in an incorrect amount posting to **GL**.
_`unallocated_amount = 2,061.64 = (3,883.88 + 0 - (0 + 935.73)) / 1.430`_
_`total_allocated_amount = 679.00`
_`received_amount = 2,716.00`_

The above values will be posted to the party **GL** at the following amount.
_`total_allocated_amount + unallocated_amount`_
_`debit_account_currency = 2,740.64 = 679.00 + 2,061.64`_

Therefore, a difference of 24.64 is posted incorrectly to the party ledger.
_`total_received - debit_account_crrency = -24.64`_

### Proposed Solution
Change the unallocated_amount formula to the following:
((`base_paid_amount` + `base_total_taxes_and_charges`) / `target_exchange_rate`) - `total_allocated_amount`

_`unallocated_amount = 2,037.00 = ((3,883.88 + 0) / 1.43) - 679.00`_
_`total_allocated_amount = 679.00`
_`received_amount = 2,716.00`_

This will result in the correct amount being posted to ledger
_`2,716.00 = 2,037.00 + 679.00`_


dj12djdjs:
Test for this PR is done in `test_payment_entry_against_si_usd_to_inr`

dj12djdjs:
Lol thought this new formula was working for some reason 😆 It's not

dj12djdjs:
@deepeshgarg007  @nextchamp-saqib could this PR get a review please? It's very critical as incorrect amount is getting posted to party ledger.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -878,18 +878,19 @@ frappe.ui.form.on('Payment Entry', {
...
-					unallocated_amount = (frm.doc.base_received_amount + total_deductions + frm.doc.base_total_taxes_and_charges
-						- frm.doc.base_total_allocated_amount) / frm.doc.source_exchange_rate;
-			} else if (frm.doc.payment_type == "Pay"
-				&& frm.doc.base_total_allocated_amount < frm.doc.base_paid_amount - total_deductions
-				&& frm.doc.total_allocated_amount < frm.doc.received_amount + (total_deductions / frm.doc.target_exchange_rate)) {
-					unallocated_amount = (frm.doc.base_paid_amount + frm.doc.base_total_taxes_and_charges - (total_deductions
-						+ frm.doc.base_total_allocated_amount)) / frm.doc.target_exchange_rate;
+			if(frm.doc.payment_type === "Receive"
+				&& frm.doc.base_total_allocated_amount < frm.doc.base_received_amount + total_deductions) {
+				unallocated_amount = (
```
deepeshgarg007: @dj12djdjs What happens when there is a deduction or tax applied in the Payment Entry. I see that is not considered in the new formula. Can you please elaborate with an example considering deductions as well

deepeshgarg007:
@dj12djdjs any update on the deduction amount query?

dj12djdjs:
> @dj12djdjs any update on the deduction amount query?

@deepeshgarg007 I'll push today.

dj12djdjs:
@deepeshgarg007 I need to investigate the failing expense claim test. 

Regarding taxes and charges, to be honest I don't know what it's primary use case is. I wasn't able to find any information in the Payment Entry docs. I re-wrote the formula based on my understanding of the previous one.

I've setup a test site with this patch installed and the erpnext test data. Could you test the taxes and charges are working as expected?
https://fix-validate-pe-ex-rate.frappe.cloud/
User: Administrator
Pass: 2c8db28

dj12djdjs:
@deepeshgarg007 Any feedback on this? I've been using this patch for 1 week. I haven't received any complaints of error.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -870,36 +870,45 @@ frappe.ui.form.on('Payment Entry', {
...
+			target_exchange_rate,
+			source_exchange_rate,
+		} = frm.doc;
+
+		const exchange_rate = payment_type === "Pay" ? target_exchange_rate : source_exchange_rate;
+		const base_amount = payment_type === "Pay" ? base_paid_amount : base_received_amount;
+
+		if (party &&
+			total_allocated_amount < (base_amount - total_deductions) / exchange_rate) {
+			unallocated_amount = (
```
deepeshgarg007, line 889: @dj12djdjs I am not sure if the deduction still has any impact on the calculation of `unallocated_amount`. If you'll check the formula the net total deduction is still 0 as the same value gets added and subtracted. Will be helpful if you add an example/test case here with a deduction in the case of Receive and Pay as well.

Also this unit test `test_multiple_payment_entries_against_expense ` seems to be related to this change. This will also have to be fixed

deepeshgarg007 (changes requested):

dj12djdjs:
This PR has become over complicated. Will re-open in a new PR.

PR #30510, opened 2022-03-31
fix: Taxes getting overriden from mapped to target doc

deepeshgarg007:
While mapping a source doc (for eg, Purchase Receipt) to a target doc (Purchase Invoice) the taxes should not change and the taxes added manually by the user in the source doc should persist in the target doc

On mapping the doc the `get_party_details` trigger used to reset the taxes from the tax templated added in the doc and the taxes manually added by the user gets removed. The change will prevent this from happening


deepeshgarg007:

deepeshgarg007:

PR #30516, opened 2022-03-31
fix(barcode_scanner)!: improve interactions with batch/serial selector

ankush:
Changes:
1. When serial no is scanned, also add the batch no of it in the child table. 
2. (breaking change) don't trigger scan on barcode field (this was circular trigger and old way to scan items). Scanning now only works using the "Scan Barcode" field. 
3. When scanned information is "sufficient" for populating details, avoid prompting with serial/batch selector. 

Scanning multiple serial no with different batches will now be on separate rows

Screenshot 2022-03-31 at 2 40 24 PM
 

To test:
check behavior on: 1. stock entry (issue), DN, SI (with update stock)

Expected behaviour:

| What is scanned | Normal item               | batched item                            | serialized item                       | batch+serial item                       |
| :-:             | :-:                       | :-:                                     | :-:                                   | :-:                                     |
| Barcode         | fetch item code, incr qty ✅ | fetch item code, show batch selector ✅   | fetch item code, show serial selector ✅ | fetch item show batch + serial selector ✅ |
| Batch number    | N/A                       | fetch item and batch, dont show dialog ✅ | n/a                                   | fetch item, batch, show serial selector ✅ |
| Serial number   | N/A                       | n/a                                     | fetch item and sr, dont show dialog ✅  | fetch item, sr, batch dont show dialog ✅ |

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -404,13 +404,11 @@ erpnext.TransactionController = class TransactionController extends erpnext.taxe
 	}
 
 	barcode(doc, cdt, cdn) {
-		const d = locals[cdt][cdn];
-		if (!d.barcode) {
+		var d = locals[cdt][cdn];
+		if(d.barcode=="" || d.barcode==null) {
 			// barcode cleared, remove item
 			d.item_code = "";
 		}
-		// flag required for circular triggers
-		d._triggerd_from_barcode = true;
 		this.item_code(doc, cdt, cdn);
 	}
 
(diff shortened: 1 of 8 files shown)


ankush:
Stock entry: works fine
Delivery note and sales invoice: triggers serial no selector even though both batch and serial no are added.  🤦

ankush:
What works:
1. Skipping of selector when info is present. 

What doesn't work:
1. setting correct values. Due to the convoluted callback hell of item_code, batch_no and serial_no triggers, things are getting overridden in an unpredictable manner.

PR #30528, opened 2022-03-31
refactor!: change "is_subcontracted" field type from "Select" to "Check"

s-aga-r:
Change the "is_subcontracted" field type from "Select" to "Check" in Purchase Order, Purchase Receipt, Purchase Invoice, Supplier Quotation.


erpnext/controllers/buying_controller.py
```diff
@@ -340,9 +340,9 @@ def get_supplied_items_cost(self, item_row_id, reset_outgoing_rate=True):
 
 	def validate_for_subcontracting(self):
 		if not self.is_subcontracted and self.sub_contracted_items:
-			frappe.throw(_("Please enter 'Is Subcontracted' as Yes or No"))
+			frappe.throw(_("Please check or uncheck 'Is Subcontracted'"))
```
ankush: This validation should just be removed. Checkbox will always be 0 or 1.

erpnext/patches.txt
```diff
@@ -334,6 +334,7 @@ erpnext.patches.v13_0.delete_bank_reconciliation_detail
 erpnext.patches.v13_0.enable_provisional_accounting
 erpnext.patches.v13_0.non_profit_deprecation_warning
 erpnext.patches.v13_0.enable_ksa_vat_docs #1
+erpnext.patches.v14_0.change_is_subcontracted_fieldtype
```
ankush: Verify if there are any intermediate patches that depend on the "is_subcontracted" field, if they do then this patch has to be moved up.

erpnext/patches/v14_0/change_is_subcontracted_fieldtype.py
```diff
@@ -0,0 +1,12 @@
...
+
+import frappe
+
+
+def execute():
+	for doctype in ["Purchase Order", "Purchase Receipt", "Purchase Invoice", "Supplier Quotation"]:
+		docs = frappe.get_all(doctype, fields=["name", "is_subcontracted"])
+		for doc in docs:
+			value = "1" if doc.is_subcontracted == "Yes" else "0"
+			frappe.db.set_value(doctype, doc.name, "is_subcontracted", value, update_modified=False)
```
ankush: This will take a very long time on large databases. Write raw SQL / QB. 

Also patches should ideally be idempotent i.e. rerunning them shouldn't cause side effects. Rerunning this patch right now will just make entire column `0`



Example: 

https://github.com/frappe/erpnext/blob/f754e76067c64336a008e800e788e722efb63452/erpnext/patches/v12_0/update_is_cancelled_field.py#L19-L34

erpnext/stock/doctype/bin/bin.py
```diff
@@ -79,7 +79,7 @@ def update_reserved_qty_for_sub_contracting(self):
 				& (se.name == se_item.parent)
 				& (po.name == se.purchase_order)
 				& (po.docstatus == 1)
-				& (po.is_subcontracted == "Yes")
+				& (po.is_subcontracted)
```
ankush: Does this query work? I'd suggest being explicit with `is_subcontracted == 1`

ankush (changes requested):

ankush (approved):

casesolved-co-uk (commented):
You didn't mean to change this file. This will have introduced a bug.
erpnext/accounts/report/tax_detail/test_tax_detail.json

s-aga-r:
> You didn't mean to change this file. This will have introduced a bug. erpnext/accounts/report/tax_detail/test_tax_detail.json

@casesolved-co-uk Can you create a GitHub issue for the same with a description of the bug?

PR #30542, opened 2022-04-01
fix: Ignore disabled tax categories

deepeshgarg007:
Query should not consider disabled tax categories


deepeshgarg007:

PR #30553, opened 2022-04-04
fix(india): minor e-invoicing fixes

nextchamp-saqib:
Fixes:
- If the mode of payment > 18 characters, the e-invoice portal throws an error
  - Since the mode of payment info is not mandatory, it can be omitted.
- Fix JSON parse error while previewing GST E-Invoice before generating IRN
   CleanShot 2022-04-04 at 14 38 29@2x


nextchamp-saqib:
Unrelated tests failing

PR #30555, opened 2022-04-04
fix: Ignore user perm for party account company

deepeshgarg007:
User will user permission set for Company DocType is not able to view customers having multiple party accounts belonging to different companies

image

Screenshot 2022-04-04 at 12 55 38 PM

Screenshot 2022-04-04 at 12 55 55 PM

To solve this ignored user permissions for company field in party account DocType


deepeshgarg007:

PR #30558, opened 2022-04-04
refactor: rename method in subcontracting.py

s-aga-r:
Remove prefix __ from the method name in subcontracting.py, as we don't use this type of naming convention.

### DIFF
--- a/erpnext/controllers/subcontracting.py
+++ b/erpnext/controllers/subcontracting.py
@@ -14,30 +14,30 @@ class Subcontracting:
 			return
 
 		self.raw_material_table = raw_material_table
-		self.__identify_change_in_item_table()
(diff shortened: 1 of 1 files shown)


ankush:
this was done intentionally as subcontracting class is inherited using "Multiple-inheritance"  @s-aga-r 

Also all the listed methods are private methods. So if not `__` they should be at least prefixed with `_`

s-aga-r:
> this was done intentionally as subcontracting class is inherited using "Multiple-inheritance" @s-aga-r
> 
> Also all the listed methods are private methods. So if not `__` they should be at least prefixed with `_`

@nabinhait Should we proceed with `_` or `__`?

PR #30564, opened 2022-04-04
 fix: if accepted warehouse not selected during rejection then stock ledger not created

rohitwaghchaure:
- Create purchase receipt with rejected qty and rejected warehouse
- Don't select the accepted warehouse
- Submit the purchase receipt, you'll see the SLE entry has not created for the rejected qty in rejected warehouse


ankush:
@rohitwaghchaure related test failing

Screenshot 2022-04-04 at 5 33 24 PM

PR #30575, opened 2022-04-04
fix: Do not apply shipping rule for POS transactions

deepeshgarg007:
POS transactions doesn't involve any shipping so shipping rule should not be applied on POS transactions


deepeshgarg007:
@Mergifyio backport develop

rakshitsukanta:
@deepeshgarg007 ,
Please give suggestion, 
Suppose in a scenario customer walks in orders an item (bulky like Furniture etc.) pays by POS and it requires a delivery charge to ship the product to customer. How to charge shipping for this?

PR #30578, opened 2022-04-05
chore: Accessibility for E-commerce Doctypes

marination:
- Added accessibility in Item master for **Website Item**
  Screenshot 2022-04-05 at 12 10 00 PM

- Added Prices, Stock and E-com Settings access from Website Item
  Screenshot 2022-04-05 at 12 31 25 PM
  
---------------
> Not making a workspace right now, because it is turning out very painful to make a standard workspace


marination:
Item.js needs a cleanup, will do in another PR

marination:
changes have nothing to do with server tests

PR #30584, opened 2022-04-05
refactor!: remove DATEV integration

barredterra:
The DATEV integration follows the trend and moves into a separate app.

- the DATEV Integration will be maintained at https://github.com/alyf-de/erpnext_datev (version-14 branch)
- the German Compliance and Localization will be maintained at https://github.com/alyf-de/erpnext_germany

PR #30598, opened 2022-04-06
chore(deps)!: Drop `pandas` as dependency

### LINKED ISSUE #27047: refactor: remove pandas dependency
Currently, only 2 functionalities in ERPNext use pandas:
- Sales funnel (which does JUST 1 transformation using pandas, this usage is seriously a crutch)
- German tax reports (DATEV)

I am proposing to refactor and remove this dependency because:
- It's a huge library (35+ MB) and comes with a lot of transitive dependencies. 
- It has caused _many_ installation failures in past and most newbies can't figure out why. This has improved a lot in recent years as NumPy and Pandas packaging has improved, but still, this will always cause some mess IMO. 

Cons:
- It's actually quite expressive and useful in reports. It's just that not many seem to be using it. 

TODO:

- [x] remove pandas usage from core erpnext https://github.com/frappe/erpnext/pull/30597
- [x] move datev stuff to separate app.  https://github.com/frappe/erpnext/pull/30584 
- [x] drop dependency https://github.com/frappe/erpnext/pull/30598

ankush:
depends on https://github.com/frappe/erpnext/pull/30597
depends on https://github.com/frappe/erpnext/pull/30584 

Closes https://github.com/frappe/erpnext/issues/27047 (read for more details)


sagarvora:
Pandas took 500M * number of Gunicorn workers (I've seen 17 on sites) in RAM, so great job removing it! Have been thinking about it for a while now. Another one on my list is `openpyxl` (this takes ~400M RAM per worker). I'm planning to replace it with `pylightxl` and `xlsxwriter` (both of these have no dependencies - they use the standard library)

gavindsouza (approved):

PR #30602, opened 2022-04-06
fix: Deferred Revenue/Expense Account validation

deepeshgarg007:
Users were able to submit the Deferred Invoices without specifying the deferred income/expense account in the invoices and later the deferred booking used to fail because of the missing account.

Added the missing validation for this


deepeshgarg007:

PR #30606, opened 2022-04-06
fix: Exchange gain and loss button in Payment Entry

deepeshgarg007:
Currently, the button to add exchange gain and loss as deductions show up in the UI only when the paid from and paid to account has different currencies and the currency is different currencies.

There could be scenarios where both the accounts have the same currency but still there is an exchange gain and loss due due to different exchange rates at the time of invoicing and make payment

Updated the condition to show up the button in both the scenarios


deepeshgarg007:

PR #30613, opened 2022-04-06
fix(UX): enable Track Changes, show allocation history for earned leaves allocated via scheduler

ruchamahabal:
## Background:

If Leave Policy Assignment is created for Earned Leave, initially 0 leaves are allocated and then allocation happens every month based on the configuration set in Leave Type. 
eg Policy: **Annual Allocation**: 24, **Frequency**: Monthly
In this case, 2 leaves will be allocated every month for the employee via scheduler. 

## Problem:

- Users are directly shown Total Leaves Allocated in Leave Allocation.
   image
- There is no trace of when the scheduler allocates these leaves because this happens via `db_set`
- "New Leaves Allocated" field is editable after submission even for Earned Leave type with policy assignment which doesn't make sense because this allocation happens via scheduler. Users can reset leaves to 0 directly, without any trace and ledgers do not handle this.
- There is one more setting that decides when this leave will be allocated via the scheduler in Leave Type master. If Based on Date of Joining is enabled, it allocates leave every month on the Joining Date of the employee, else on month-end.

  image
- Basically, users have to see 3 different things and do the mental math to derive the number of leaves shown in the allocation: the annual allocation, the number of months that have passed, and the based on DOJ setting.

## Fix: 

- Enable Track Changes for Leave Allocation (useful for allocations not happening via a scheduler)
- Make New Leaves Allocated input read-only if the earned leave Allocation is linked to Leave Policy Assignment (indicating that this will happen via scheduler)
- Show the history of earned leaves allocated along with the date and the info of whether they were allocated based on DOJ or not

Based on DOJ:

log

Usual:

image

### DIFF
--- a/erpnext/hr/doctype/leave_allocation/leave_allocation.js
+++ b/erpnext/hr/doctype/leave_allocation/leave_allocation.js
@@ -34,6 +34,15 @@ frappe.ui.form.on("Leave Allocation", {
 				});
 			}
 		}
+
+		// make new leaves allocated field read only if allocation is created via leave policy assignment
+		// and leave type is earned leave, since these leaves would be allocated via the scheduler
+		if (frm.doc.leave_policy_assignment) {
+			frappe.db.get_value("Leave Type", frm.doc.leave_type, "is_earned_leave", (r) => {
+				if (r && cint(r.is_earned_leave))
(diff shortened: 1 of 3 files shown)


ruchamahabal:
Not adding tests since it's just a comment being added to the doc timeline. Earned leave calculation already has coverage. Unrelated test failures

PR #30626, opened 2022-04-07
feat(india): e-invoicing for intra-state union territory transactions

nextchamp-saqib:
Added a new column in GST Settings - Accounts table

CleanShot 2022-04-07 at 13 10 55@2x

Users can set UTGST accounts and the same will be considered as an SGST account for e-invoicing.
Later the same will be used in other applicable places as well like GST Reports etc.


mohsinalimat:
Could you please support v13 as well.

PR #30634, opened 2022-04-07
fix: dont reassign mutable (list) to a different field

sagarvora:
Search regex for those interested: `[^=]*=[^=]+=.*\[\]`

Identified here: https://github.com/frappe/erpnext/runs/5866861172?check_suite_focus=true#step:12:843

### DIFF
--- a/erpnext/accounts/report/purchase_register/purchase_register.py
+++ b/erpnext/accounts/report/purchase_register/purchase_register.py
@@ -124,11 +124,11 @@ def get_columns(invoice_list, additional_table_columns):
 		_("Purchase Receipt") + ":Link/Purchase Receipt:100",
 		{"fieldname": "currency", "label": _("Currency"), "fieldtype": "Data", "width": 80},
 	]
-	expense_accounts = (
-		tax_accounts
-	) = (
-		expense_columns
-	) = tax_columns = unrealized_profit_loss_accounts = unrealized_profit_loss_account_columns = []
+
+	expense_accounts = []
+	tax_accounts = []
+	tax_columns = []
+	unrealized_profit_loss_accounts = []
 
 	if invoice_list:
 		expense_accounts = frappe.db.sql_list(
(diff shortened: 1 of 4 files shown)


erpnext/accounts/report/purchase_register/purchase_register.py
```diff
@@ -124,11 +124,11 @@ def get_columns(invoice_list, additional_table_columns):
...
-	expense_accounts = (
-		tax_accounts
-	) = (
-		expense_columns
-	) = tax_columns = unrealized_profit_loss_accounts = unrealized_profit_loss_account_columns = []
+
+	expense_accounts = []
+	tax_accounts = []
+	tax_columns = []
+	unrealized_profit_loss_accounts = []
```
sagarvora, line 131: removed two variables since they are being re-assigned below

erpnext/accounts/report/purchase_register/purchase_register.py
```diff
@@ -124,11 +124,10 @@ def get_columns(invoice_list, additional_table_columns):
...
 	]
-	expense_accounts = (
-		tax_accounts
-	) = (
-		expense_columns
-	) = tax_columns = unrealized_profit_loss_accounts = unrealized_profit_loss_account_columns = []
+
+	expense_accounts = []
+	tax_accounts = []
+	unrealized_profit_loss_accounts = []
```
sagarvora, line 130: three variables already being init below, skipped here.

erpnext/selling/doctype/customer/customer.py
```diff
@@ -100,7 +100,8 @@ def validate(self):
 	@frappe.whitelist()
 	def get_customer_group_details(self):
 		doc = frappe.get_doc("Customer Group", self.customer_group)
-		self.accounts = self.credit_limits = []
+		self.accounts = []
+		self.credit_limits = []
```
sagarvora, line 104: this only worked because `doc.append` always re-initialised the table if the value was falsy 🤣

erpnext/selling/doctype/customer/customer.py
```diff
@@ -100,7 +100,8 @@ def validate(self):
 	@frappe.whitelist()
 	def get_customer_group_details(self):
 		doc = frappe.get_doc("Customer Group", self.customer_group)
-		self.accounts = self.credit_limits = []
```
sagarvora, line 103: this only worked because `doc.append` always re-initialised the table if the value was falsy 🤣

ankush (approved):

PR #30638, opened 2022-04-07
fix: use empty row when batch is scanned

ankush:
Currently, if you scan barcode/serial no, first empty row is used. However, this is not done for batch number scans. This PR makes the behavior consistent for all the cases. 

continuation of https://github.com/frappe/erpnext/pull/30445 & https://github.com/frappe/erpnext/pull/30516


ankush:
This has UI tests now :smile:

Issue #30465, opened 2022-03-29
GSTR-1 Download JSON error


mohsinalimat:
### Information about bug

![Screenshot 2022-03-29 at 12 06 25 PM](https://user-images.githubusercontent.com/973676/160550306-832b3c2a-0f21-474b-94a2-eab34a3d0b6b.png)
![Screenshot 2022-03-29 at 12 11 36 PM](https://user-images.githubusercontent.com/973676/160550311-bbef60d3-50ab-4217-b273-52582b736689.png)
![Screenshot 2022-03-29 at 12 12 52 PM](https://user-images.githubusercontent.com/973676/160550314-48b9fe05-aa88-4f1d-8779-5aac1c182961.png)

### Module

accounts

### Version

Frappe-version = v13.23.0
ERPNext-version = v13.23.3

### Installation method

easy-install

### Relevant log output / Stack trace / Full Error Message.

```shell
Traceback (most recent call last):
  File "apps/frappe/frappe/app.py", line 68, in application
    response = frappe.api.handle()
  File "apps/frappe/frappe/api.py", line 55, in handle
    return frappe.handler.handle()
  File "apps/frappe/frappe/handler.py", line 31, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/handler.py", line 68, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "apps/frappe/frappe/__init__.py", line 1213, in call
    return fn(*args, **newargs)
  File "apps/erpnext/erpnext/regional/report/gstr_1/gstr_1.py", line 812, in get_json
    gstin = get_company_gstin_number(filters.get("company"), filters.get("company_address"))
  File "apps/erpnext/erpnext/regional/report/gstr_1/gstr_1.py", line 1177, in get_company_gstin_number
    frappe.throw(_("Please set valid GSTIN No. in Company Address {} for company {}").format(
  File "apps/frappe/frappe/__init__.py", line 444, in throw
    msgprint(msg, raise_exception=exc, title=title, indicator='red', is_minimizable=is_minimizable, wide=wide, as_list=as_list)
  File "apps/frappe/frappe/__init__.py", line 423, in msgprint
    _raise_exception()
  File "apps/frappe/frappe/__init__.py", line 378, in _raise_exception
    raise raise_exception(msg)
frappe.exceptions.ValidationError: Please set valid GSTIN No. in Company Address  for company <b>Company Name</b>
```

[...]

PR #30651, opened 2022-04-08
fix: Download JSON for GSTR-1 report

deepeshgarg007:
Stale filters that we set on load were getting passed while downloading the JSON and leading to errors


mohsinalimat:
https://github.com/frappe/erpnext/issues/30465

deepeshgarg007:

PR #30654, opened 2022-04-08
fix: update translation

FHenry:
Following #30474


FHenry:

PR #30662, opened 2022-04-09
fix: Implicit ignore pricing rule check on returns

deepeshgarg007:
When "Ignore Pricing Rule" is enabled system tries to remove the pricing rules that are applied and reset the rates to the pricelist rate

https://github.com/frappe/erpnext/blob/71f72458bfb3371b30ca240efa10e3b4602c6f62/erpnext/accounts/doctype/pricing_rule/pricing_rule.py#L298

In case of a making a return for a document on which the pricing rule is applied ignore pricing rule was automatically getting checked and system was trying to remove the applied pricing rules and reset the rate on submissing which led to the below 
error even though there was no change in rate

image


deepeshgarg007:
@Mergifyio backport develop

PR #30675, opened 2022-04-11
feat: Item-wise provisional accounting for service items

deepeshgarg007:
Added default provisional account in item defaults in the Item master and move the provisional account from parent to the item table so that users can do item-wise provisional accounting and track item-wise expenses.

Screenshot 2022-04-14 at 12 07 06 PM

Screenshot 2022-04-14 at 12 24 01 PM

`no-docs`


deepeshgarg007:

PR #30677, opened 2022-04-11
refactor: stock balance report

ankush:
- [x] set up tests
- [x] query conversions, remove all raw SQL. 
- [x] cover most used scenarios in tests
- [x] minor code cleanup here and there. 

don't backport

### DIFF
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -7,11 +7,13 @@ from typing import Optional, TypedDict
 
 import frappe
 from frappe import _
+from frappe.query_builder.functions import CombineDatetime
 from frappe.utils import cint, date_diff, flt, getdate
+from frappe.utils.nestedset import get_descendants_of
+from pypika.terms import ExistsCriterion
 
 import erpnext
 from erpnext.stock.report.stock_ageing.stock_ageing import FIFOSlots, get_average_age
-from erpnext.stock.report.stock_ledger.stock_ledger import get_item_group_condition
 from erpnext.stock.utils import add_additional_uom_columns, is_reposting_item_valuation_in_progress
 
 
@@ -32,8 +34,6 @@ def execute(filters: Optional[StockBalanceFilter] = None):
 	is_reposting_item_valuation_in_progress()
 	if not filters:
 		filters = {}
-
-	to_date = filters.get("to_date")
 
 	if filters.get("company"):
 		company_currency = erpnext.get_company_currency(filters.get("company"))
@@ -62,6 +62,7 @@ def execute(filters: Optional[StockBalanceFilter] = None):
 
 	_func = itemgetter(1)
 
+	to_date = filters.get("to_date")
 	for (company, item, warehouse) in sorted(iwb_map):
 		if item_map.get(item):
 			qty_dict = iwb_map[(company, item, warehouse)]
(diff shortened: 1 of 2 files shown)


erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -215,66 +231,67 @@ def get_columns(filters):
...
 
-	if filters.get("warehouse_type") and not filters.get("warehouse"):
-		conditions += (
-			" and exists (select name from `tabWarehouse` wh \
-			where wh.warehouse_type = '%s' and sle.warehouse = wh.name)"
-			% (filters.get("warehouse_type"))
+	if warehouse := filters.get("warehouse"):
+		children = get_descendants_of("Warehouse", warehouse, ignore_permissions=True)
+		# TODO: check perf impact
+		query = query.where(sle.warehouse.isin(children + [warehouse]))
```
ankush: At first glance, this implementation using `isin` appears to be much faster because of removal of correlated subquery. 

TODO: check on very large database. 

Patch to move to previous `exists` query if required


```patch
diff --git a/erpnext/stock/report/stock_balance/stock_balance.py b/erpnext/stock/report/stock_balance/stock_balance.py
index d06a3992e0..3e136848f5 100644
--- a/erpnext/stock/report/stock_balance/stock_balance.py
+++ b/erpnext/stock/report/stock_balance/stock_balance.py
@@ -9,7 +9,7 @@ import frappe
 from frappe import _
 from frappe.query_builder.functions import CombineDatetime
 from frappe.utils import cint, date_diff, flt, getdate
-from frappe.utils.nestedset import get_descendants_of
+from pypika.terms import ExistsCriterion
 
 import erpnext
 from erpnext.stock.report.stock_ageing.stock_ageing import FIFOSlots, get_average_age
@@ -233,6 +233,7 @@ def get_columns(filters: StockBalanceFilter):
 
 def apply_conditions(query, filters):
 	sle = frappe.qb.DocType("Stock Ledger Entry")
+	warehouse_table = frappe.qb.DocType("Warehouse")
 
 	if to_date := filters.get("to_date"):
 		query = query.where(sle.posting_date <= to_date)
@@ -243,7 +244,6 @@ def apply_conditions(query, filters):
 		query = query.where(sle.company == company)
 
 	if warehouse_type := filters.get("warehouse_type") and not filters.get("warehouse"):
-		warehouse_table = frappe.qb.DocType("Warehouse")
 		query = (
 			query.join(warehouse_table)
 			.on(warehouse_table.name == sle.warehouse)
@@ -251,9 +251,17 @@ def apply_conditions(query, filters):
 		)
 
 	if warehouse := filters.get("warehouse"):
-		children = get_descendants_of("Warehouse", warehouse, ignore_permissions=True)
-		# TODO: check perf impact
-		query = query.where(sle.warehouse.isin(children + [warehouse]))
+		lft, rgt = frappe.db.get_value("Warehouse", warehouse, ["lft", "rgt"])
+		chilren_subquery = (
+			frappe.qb.from_(warehouse_table)
+			.select(warehouse_table.name)
+			.where(
+				(warehouse_table.lft >= lft)
+				& (warehouse_table.rgt <= rgt)
+				& (warehouse_table.name == sle.warehouse)
+			)
+		)
+		query = query.where(ExistsCriterion(chilren_subquery))
 
 	return query
 

```

ankush:
Each warehouse having roughly 1000 unique item and 6000-10000 SLEs
Single warehouse:
Before: 50ns
After: 32ns


Group of 5 warehouse:
Before: 57ns
After: 35ns


A very marginal difference which is eliminated once # of child warehouses ~15-20 and will be somewhat slower on numbers higher than that. Not sure if this is worth it. The main difference will likely come from rewriting the whole qty and value logic in SQL.

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -215,66 +233,75 @@ def get_columns(filters):
 	return columns
 
 
-def get_conditions(filters):
-	conditions = ""
-	if not filters.get("from_date"):
-		frappe.throw(_("'From Date' is required"))
```
marination, line 221: From Date is not validated ? Its required for opening stock or sumn right ?

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -215,66 +233,75 @@ def get_columns(filters):
...
+def apply_conditions(query, filters):
+	sle = frappe.qb.DocType("Stock Ledger Entry")
+	warehouse_table = frappe.qb.DocType("Warehouse")
 
-	if filters.get("to_date"):
-		conditions += " and sle.posting_date <= %s" % frappe.db.escape(filters.get("to_date"))
+	if to_date := filters.get("to_date"):
```
marination, line 243: oo walrus noice

erpnext/stock/report/stock_balance/stock_balance.py
ankush: Yes, it's used. But it doesn't _have to be mandatory_, felt this was unnecessary validation. Removing it means opening stock computation won't be done, which is fine?

ankush: Oh wait, clientside JS prevents it from loading, if unset. 🤦

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -215,66 +233,75 @@ def get_columns(filters):
...
-				where wh.lft >= %s and wh.rgt <= %s and sle.warehouse = wh.name)"
-				% (warehouse_details.lft, warehouse_details.rgt)
-			)
 
-	if filters.get("warehouse_type") and not filters.get("warehouse"):
-		conditions += (
-			" and exists (select name from `tabWarehouse` wh \
-			where wh.warehouse_type = '%s' and sle.warehouse = wh.name)"
-			% (filters.get("warehouse_type"))
+	if warehouse := filters.get("warehouse"):
```
marination: Instead of checking twice if warehouse filter exists
```py
if warehouse := filters.get("warehouse"):
    # snippet
elif warehouse_type := filters.get("warehouse_type"):
```

erpnext/stock/report/stock_balance/stock_balance.py
marination: > Yes, it's used. But it doesn't have to be mandatory, felt this was unnecessary validation. Removing it means opening stock computation won't be done, which is fine?

Apart from JS, its also unexpected behaviour for users using this report via API for years (we have some)

erpnext/stock/report/stock_balance/stock_balance.py
ankush: fixed

erpnext/stock/report/stock_balance/stock_balance.py
ankush: Ah got it, better to put this back 😅

marination (approved):

PR #30686, opened 2022-04-12
feat: Ignore permlevel for specific fields

nabinhait:
**Current Behaviour:**
Currently, if a field has a higher permlevel and a user does not have permission to that permlevel, on saving the document system resets the value of the field if there are any changes.

**Solution:**
There are some use-cases where the value of the field is changed by the code, not by the user. To allow those changes, introduced a document level flag to mention those specific fields for which permlevel will be ignored.

**Actual Use Case:**
In the ERPNext sales cycle, users want to set a higher permlevel for the price_list field because they want to control and it is set based on configuration.
On loading a transaction, the default value for price_list is set based on default_price_list set on Selling Settings. Then on the selection of Customers, it gets overwritten by the customer's default price list. But on saving the transaction, the system resets the price_list field's value to the original (system settings). This change will fix it.

Associated PR on the Frappe: https://github.com/frappe/frappe/pull/16590

docs: no-docs

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -22,6 +22,9 @@ class QtyMismatchError(ValidationError):
 
 
 class BuyingController(StockController, Subcontracting):
+	def __setup__(self):
+		self.flags.ignore_permlevel_for_fields = ["buying_price_list", "price_list_currency"]
+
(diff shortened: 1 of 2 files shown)


nabinhait:

deepeshgarg007 (approved):

PR #30688, opened 2022-04-12
fix: Map Production Plan company in subassembly WO created from it

### LINKED ISSUE #30505: bug: creating Work Order from Production Plan must be made in Default Company else 'Warehouse Finished Goods - ABC does not belong to company None'
### Information about bug

There is a check when creating Work Orders from a Production Plan that the `For Warehouse` belongs to the Global Default - Default Company. It is quite possible that multiple companies could be using the same instance with separate warehouses and doing manufacturing at the same time.

The Work Order should be created for the Company set in the Production Plan.

Screen Shot 2022-03-30 at 22 15 56

### Module

manufacturing

### Version

v13.19.0

### Installation method

_No response_

_No response_

marination:
Closes https://github.com/frappe/erpnext/issues/30505

**Issue:**
- Prod Plan made for company **TextileMart**, but while making WO, subassembly WO takes global default company **RetailMart**

- https://user-images.githubusercontent.com/25857446/163110562-5484c72e-6694-42e5-92f1-f6fdeb4eccd6.mov

**Fix:**
- Map company in Sub assembly WO. The same happens automatically due to queried company while making FG WO

- https://user-images.githubusercontent.com/25857446/163111689-9ad292df-5051-443a-b8c7-013d1e499ef4.mov

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/production_plan.py
@@ -500,6 +500,7 @@ class ProductionPlan(Document):
 
 		for supplier, po_list in subcontracted_po.items():
 			po = frappe.new_doc("Purchase Order")
+			po.company = self.company
 			po.supplier = supplier
 			po.schedule_date = getdate(po_list[0].schedule_date) if po_list[0].schedule_date else nowdate()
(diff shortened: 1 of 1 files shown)


marination:
too smol for tests

PR #30689, opened 2022-04-12
fix: Do not show disabled dimensions in reports

deepeshgarg007:
Disabled accounting dimensions are also fetched and shown in the financial reports

Added a default `{disabled: 0}` in the `get_accounting_dimensions` function


deepeshgarg007:

PR #30700, opened 2022-04-13
fix: Payment reco query with max invoice and payment amount limit

deepeshgarg007:
Specifying maximum and minimum amount limits in Payment Reconciliation leads to the below errors. This PR fixes the query for it

Screenshot 2022-04-13 at 1 10 58 PM


deepeshgarg007:
@Mergifyio backport develop

PR #30707, opened 2022-04-13
fix: Remove "Values Out of Sync" validation

marination:
## **Issue:**
-  Consider the following two consecutive transactions (Receipts) and their impacts:
   Screenshot 2022-04-14 at 5 18 34 PM
- During V-001, **Warehouse A** did not have an account against it. So company inventory account "Stock In Hand" was used to book incoming stock value
- In V-002, **Warehouse A** has its account set to **1234 - Inventory**

Now to check if account balance of account **1234 - Inventory** matches with stock value booked against it:
- We get all warehouses with account = **1234 - Inventory**
- Get their current stock value, sum it and check if it is equal to account balance of **1234 - Inventory**
- Which is wrong here since some of **Warehouse A**'s stock value is booked in account **Stock In Hand**
- So you get stock value = **160** and account balance = **80**, **_mismatch_**

In such a case we cannot check the stock balance against a particular account as the account could have contributed partially. And getting its exact contribution is tedious.

So the values are wrong anyway.
> Tried drafting some possible approach [here](https://docs.google.com/spreadsheets/d/18xjX0V85W9BPIhT8tJFzMv0AbE_PmdVdWFMV9jUKMuE/edit#gid=0), but this will be computationally expensive.

## **Conclusion:**
- The validation computes incorrect values for an account that contributes partially to warehouse's value.
- Even if there is a difference there's not much a user can do
- Checking it at the JV level is senseless. Blindly making users post another JV will lead to more error.
- Users making JV (not prompted by system) usually know what they are doing
- Removed from reposting because it just sets job as failed. Repost goes through anyway. Theres no impact and the user can't do anything about it.

**ToDo:**
- [x] Remove validation from repost as well (useless, stock and account value comparison report can be looked at) and just comment out the function (for future)

> To whoever re-adds this next: Please consider the first point in this description. Also document how this is calculated and impact on different cases


ankush (approved):

ankush:
unrelated flaky tests.

PR #30709, opened 2022-04-13
test: flaky PR close test

ankush:


ankush:
The original issue got fixed in: https://github.com/frappe/erpnext/pull/30709

PR #30710, opened 2022-04-13
fix: Exchange gain and loss on advance Journal Entry allocation

deepeshgarg007:
Due to no reference exchange being pulled in the advances posted via Journal Entry entire advance amount was being booked as exchange gain and loss adjustment. Pulled reference exchange rate to fix this and added test

image

Also ignored cascade cancellation of advance entries on cancellation of invoice. The advance entries should be just unlinked and not get cancelled


deepeshgarg007:

PR #30711, opened 2022-04-14
fix(call log): Convert `Data` to `Link` for type_of_call

surajshetty3416:
Convert `Data` to `Link` for `type_of_call`, `employee_user_id` and `call_received_by` fields to establish proper linking.

New fields added [in this PR](https://github.com/frappe/erpnext/pull/29962) were all data fields instead of Link fields 😑


surajshetty3416:
Minor fix. Review not required. Merging.

PR #30716, opened 2022-04-14
fix: update translation

FHenry:
fix: update translation


erpnext/translations/fr.csv
```diff
@@ -271,7 +271,7 @@ Assessment Report,Rapport d'Évaluation,
 Assessment Reports,Rapports d'évaluation,
 Assessment Result,Résultat de l'Évaluation,
 Assessment Result record {0} already exists.,Le Résultat d'Évaluation {0} existe déjà.,
-Asset,Atout,
+Asset,Actif - Immo.,
```
ankush, line 274: This doesn't quite feel correct. Can you share some sources? 
Screenshot 2022-04-14 at 5 24 36 PM

marination: @ankush https://github.com/frappe/erpnext/pull/30593#discussion_r844291392
Similar thing in the older PR. Seems like **Actif Immobilisation** is used as "Fixed Assets"

FHenry:

PR #30721, opened 2022-04-16
fix: SO's without delivery note will also be fetched

ruthra-kumar:
### Fix bug on Sales order analysis report
Report will also fetch SO's that doesn't have any Delivery Note created for them.

Screenshot 2022-04-16 at 11 17 58 AM
Screenshot 2022-04-16 at 10 38 00 AM

`docs:` https://docs.erpnext.com/docs/v13/user/manual/en/selling/sales-analytics#4-sales-order-analysis-report


deepeshgarg007:

deepeshgarg007:

PR #30724, opened 2022-04-16
fix: Consistent customer and supplier forms

deepeshgarg007:
Made placement of fields in Customer and Supplier doctypes more consistent

#### Customer Form
image

#### Supplier Form
image


ChillarAnand:
@deepeshgarg007 Supplier form screenshot is attached in place of customer form?

With before/after screenshots, it would be easy to visualise the changes.

ChillarAnand (approved):

PR #30725, opened 2022-04-16
fix: update translation (Russian)

Altizo:
This is a series of corrections after a cursory check of the translation. This PR fixes a mistranslation that messes up the UX a lot.
_(These are just some translation fixes. There will be many changes in the future)_


marination:
@Altizo is this PR ready for review or are there more changes expected ?

Altizo:
Thanks @marination , yes, there will be many more changes, but I think it would be better to prepare a separate PR for them.

Due to the peculiarities of the language, the translation is terrible in some places, and a lot of work needs to be done to correct it.

marination (approved):

PR #30726, opened 2022-04-17
fix: dependent gle reposting

ankush:
Minimal reproducible example:
1. Pick an item and two warehouses A, B. All transaction are with 1 qty.
2. Receive stock in A @ X rate
3. Transfer stock to B (@ X rate, implicitly)
4. Consume from B (@X rate implicitly) GL entry will be posted with X amount
5. Receive backdated stock in A with Y rate. This will repost and correct entries in B warehouse but won't repost GL entires for transaction mentioned step # 4 

Fix: keep track of dependent transactions that are processed even if they don't directly depend on item-wh; and send them to GLE reposting function.

TODO:
- [x] check resuming with multi-item-wh (manual test)
- [x] sorting concern
- [ ] refactor `future_sle_exists` (in separate PR)

### DIFF
--- a/erpnext/accounts/test/test_utils.py
+++ b/erpnext/accounts/test/test_utils.py
@@ -1,10 +1,17 @@
 import unittest
 
+import frappe
 from frappe.test_runner import make_test_objects
 
 from erpnext.accounts.party import get_party_shipping_address
-from erpnext.accounts.utils import get_future_stock_vouchers, get_voucherwise_gl_entries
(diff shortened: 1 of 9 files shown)


ankush:
Screenshot 2022-04-19 at 6 42 55 PM


Tested locally by forcing job failure. The last 4 appeared one at a time after each retry.

marination (approved):

PR #30728, opened 2022-04-18
fix: Price changing on creating Sales return from Delivery Note

deepeshgarg007:
Rate for items having pricing rule gets changed when making a sales return through a delivery note and the user gets the below invalid validation

image


deepeshgarg007:

PR #30729, opened 2022-04-18
chore: Add semantic releases

deepeshgarg007:
- Automatically updated version (version bump)
- Auto-generate release notes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,7 +2,7 @@ name: Generate Semantic Release
 on:
   push:
     branches:
-      - test-release
+      - version-13
 jobs:
   release:
     name: Release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,25 @@
+name: Generate Semantic Release
+on:
+  push:
+    branches:
+      - test-release
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Checkout Entire Repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup Node.js v14
+        uses: actions/setup-node@v2
+        with:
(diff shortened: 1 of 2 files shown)


.releaserc
```diff
@@ -0,0 +1,25 @@
+{
+	"branches": ["version-13"],
+	"plugins": [
+		"@semantic-release/commit-analyzer",
+		"@semantic-release/release-notes-generator",
+		[
+			"@semantic-release/exec", {
+				"prepareCmd": 'sed -ir "s/[0-9]*\.[0-9]*\.[0-9]*/${nextRelease.version}/" erpnext/__init__.py'
```
ankush, line 13: IIRC this one gladly bumps it to the next major version 😅 

We should block it in case of a major version bump.

deepeshgarg007: This it will do only in case any breaking changes are merged which ideally should not go in the `version-13` branch anymore. Will anyways see if I can add a safety check

.releaserc
```diff
@@ -0,0 +1,30 @@
...
+		[
+			"@semantic-release/git", {
+				"assets": ["erpnext/__init__.py"],
+				"message": "chore(release): Bumped to Version ${nextRelease.version}\n\n${nextRelease.notes}"
+			}
+		],
+		[
+			"@semantic-release/github", {
+				"assets": [
+					{"path": "dist/*"},
```
ankush: I don't think we need assets here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,25 @@
+name: Generate Semantic Release
+on:
+  push:
+    branches:
+      - version-13
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-18.04
```
ankush:
```suggestion
    runs-on: ubuntu-latest
```

ankush (approved):

PR #30745, opened 2022-04-19
fix: Correct precision for asset value after full schedule

GangaManoj:
### Issue:

Ideally, `asset_value_after_useful_life` should be the same as `expected_value_after_useful_life`, however, the precision applied on both are different, leading to minor inconsistencies which could possibly throw errors.

### Fix:

Apply precision for `expected_value_after_useful_life` on `asset_value_after_useful_life`, instead of that of `gross_purchase_amount`.

<details>

ISS-21-22-14006

</details>


nextchamp-saqib (approved):

aakvatech:
https://github.com/frappe/erpnext/issues/31600

PR #30748, opened 2022-04-19
fix: monthly attendance sheet

barredterra:
### Problem
In the report "Monthly Attendance Sheet" we have the option to show a "Summarized View". This includes the sum of leaves taken by **Leave Type**. Currently this shows zeros everywhere, regardless of the actual data.

before

### Cause
This is because we tried to match column definitions (`"LEAVE TYPE::Float:120"`, stored in `leave_list`) with Leave Types (`"LEAVE TYPE"`, stored in `leaves`). They are never equal, so we output 0.

https://github.com/frappe/erpnext/blob/4389100fa1ec134eb121de275d88d0f1c9d1ec59/erpnext/hr/report/monthly_attendance_sheet/monthly_attendance_sheet.py#L225-L229

### Solution

Use leave types instead of column definitions for the comparison.

after

This PR targets `version-13-hotfix` directly because the report has been refactored in `develop` and the problem most likely doesn't exist there.

### DIFF
--- a/erpnext/hr/report/monthly_attendance_sheet/monthly_attendance_sheet.py
+++ b/erpnext/hr/report/monthly_attendance_sheet/monthly_attendance_sheet.py
@@ -57,11 +57,10 @@ def execute(filters=None):
 
 	data = []
 
-	leave_list = None
+	leave_types = None
 	if filters.summarized_view:
-		leave_types = frappe.db.sql("""select name from `tabLeave Type`""", as_list=True)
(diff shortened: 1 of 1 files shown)


ruchamahabal (approved):

ruchamahabal:
Unrelated testcase failure

tajff:
Is it possible to show the total number of hours of late entry, and number of hours of early exit?

PR #30754, opened 2022-04-20
fix: First preference to parent cost center rather than round off cost center

deepeshgarg007:
There are scenarios where the user maintains branch-wise cost centers. In such scenarios, they also want the cost center in the rounding adjustment entry to be passed branch-wise rather than a single cost center. Sales and Purchase Invoice already have a cost center field that users can update, this cost center is also updated branch wise, hence this cost center should be given first preference for rounding adjustment entry cost center before the round off cost center

#### Todo
- [x] Consider custom accounting dimensions for rounding adjustment entry as well


deepeshgarg007:

PR #30759, opened 2022-04-20
fix: Depreciate Asset before generating GL Entries on sale

GangaManoj:
## Issue

If an Asset worth Rs 75,000 that has been in use since Jan 1, 2021 and depreciates annually gets sold on Feb 20, 2021, it would create a final Depreciation Entry on the date of sale, to account for its depreciation from Jan 1, 2022 to Feb 20, 2022.

Screenshot 2022-04-20 at 4 49 04 PM

However, the GL Entries to account for Asset disposal are created _before_ the final depreciation is posted, not after. This causes the Accumulated Depreciation Entries to be posted incorrectly - in this case, the Accumulated Depreciation Amount will be taken as 15,000 and not 19,479.45 (i.e. the final depreciation up to the date of sale will be ignored).

<details>
<summary>How it should be</summary>

Screenshot 2022-04-20 at 4 58 51 PM

</details>

## Fix

- During Asset sale: Depreciate Asset _before_  the GL Entries are generated, not after.
- During Asset return: Reverse depreciation _before_ the GL Entries are generated, not after.

<details>

ISS-21-22-14250

</details>


GangaManoj:
@nextchamp-saqib hey, this is that PR I was talking about. Squash merge it please.

GangaManoj:
@mergifyio backport version-13

PR #30762, opened 2022-04-20
feat: support product bundles in picklist

ankush:
Currently if you import sales order which has product bundle it does nothing and gives rather useless message "bundle item not in stock". This is because product bundles aren't supported by picklist. 

- [x] Add check to avoid importing product bundle items in picklist. 
- [x] Import packed item list for the bundle in picklist instead. 
- [x] On submit/cancel update picking status back to the original bundle (based on max possible bundle qty)
- [x] While creating DN add the original bundle item back in DN items.
- [x] Update stock info (like warehouse) on DN packed items.
- [x] Tests

Unrelated changes:
- Faster picked qty updates
- Fix pick list and SO form fields for tracking qty
- hide picklist button when it's not possible. 
- Fix pick list validation (against stock qty instead of SO qty)
- Consider already picked qty while preparing new picklist. 

`no-docs`

note to reviewers: commits are mostly atomic, so easier to read the diff commit-by-commit.

### DIFF
--- a/erpnext/stock/doctype/pick_list/pick_list.py
+++ b/erpnext/stock/doctype/pick_list/pick_list.py
@@ -618,7 +618,6 @@ def add_product_bundles_to_delivery_note(
 	for so_row, item_code in product_bundles.items():
 		sales_order_item = frappe.get_doc("Sales Order Item", so_row)
 		dn_bundle_item = map_child_doc(sales_order_item, delivery_note, item_mapper)
-		# TODO: post process packed items and update stock details
 		dn_bundle_item.qty = pick_list._compute_picked_qty_for_bundle(
 			so_row, product_bundle_qty_map[item_code]
 		)
--- a/erpnext/stock/doctype/pick_list/pick_list.py
+++ b/erpnext/stock/doctype/pick_list/pick_list.py
@@ -261,6 +261,8 @@ class PickList(Document):
 
 	def _compute_picked_qty_for_bundle(self, bundle_row, bundle_items) -> float:
 		"""Compute how many full bundles can be created from picked items."""
+		precision = frappe.get_precision("Stock Ledger Entry", "qty_after_transaction")
+
 		possible_bundles = []
 		for item in self.locations:
 			if item.product_bundle_item != bundle_row:
@@ -270,7 +272,7 @@ class PickList(Document):
 				possible_bundles.append(item.picked_qty / qty_in_bundle)
 			else:
 				possible_bundles.append(0)
-		return min(possible_bundles)
+		return flt(min(possible_bundles), precision or 6)
 
 
 def validate_item_locations(pick_list):
--- a/erpnext/stock/doctype/pick_list/pick_list.py
+++ b/erpnext/stock/doctype/pick_list/pick_list.py
@@ -517,13 +517,7 @@ def create_delivery_note(source_name, target_doc=None):
 	if sales_dict:
 		delivery_note = create_dn_with_so(sales_dict, pick_list)
 
-	is_item_wo_so = 0
-	for location in pick_list.locations:
-		if not location.sales_order:
-			is_item_wo_so = 1
-			break
-	if is_item_wo_so == 1:
-		# Create a DN for items without sales orders as well
+	if not all(item.sales_order for item in pick_list.locations):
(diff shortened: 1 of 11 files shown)


ankush:
Future scope of improvement while picking: instead of mapping fully qty open a multi-select dialog box with a field to specify how much qty to pick.

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -86,20 +97,16 @@ def update_so(self, so_item, picked_qty, item_code):
 				frappe.throw(
 					_(
 						"You are picking more than required quantity for {}. Check if there is any other pick list created for {}"
-					).format(item_code, so_doc.name)
+					).format(item_code, item.sales_order)
```
marination, line 103: Bug: in multi-uom cases, `picked_qty` is in stock UOM and `actual_qty` is in UOM specified in SO.
It always throws the error even though I am picking half qty
Eg: Item Double Dozen Eggs > Stock UOM: Nos > 1 Box = 24 Nos

Place SO for 10 Boxes (240 Nos) > Pick 5 Boxes (120 Nos) > It says I am overpicking

marination, line 100: Bug: in multi-uom cases, `picked_qty` is in stock UOM and `actual_qty` is in UOM specified in SO.
It always throws the error even though I am picking half qty
Eg: Item Double Dozen Eggs > Stock UOM: Nos > 1 Box = 24 Nos

Place SO for 10 Boxes (240 Nos) > Pick 5 Boxes (120 Nos) > It says I am overpicking

ankush: Nice catch. Need to fetch conversion factor specified in SO row too ig 🤔

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -217,6 +225,57 @@ def group_similar_items(self):
...
+		possible_bundles = []
+		for item in self.locations:
+			if item.product_bundle_item != bundle_row:
+				continue
+
+			if qty_in_bundle := bundle_items.get(item.item_code):
+				possible_bundles.append(item.picked_qty / qty_in_bundle)
+			else:
+				possible_bundles.append(0)
+		return flt(min(possible_bundles), precision or 6)
```
marination: Does not respect "Must be Whole UOM", and shouldn't go to FP anyway

Bundle: Packed Item = 1:3
So technically I can pick 12 items  = 4 Bundles
But it allows me to pick 13 items = 4.333 Bundles

Let's not allow it to go FP, it doesn't make sense to have 4.3 bundles. Yet to come across a legit use case for this

marination (changes requested):

erpnext/stock/doctype/pick_list/pick_list.py
ankush: Changed field to stock_qty for comparison and added tests: https://github.com/frappe/erpnext/pull/30762/commits/8207697e43a8baa3353c0c984e8e99b14fcb5b55 

I am bit conflicted about UOM of `picked_qty` field. So far it's been in stock_qty. Total picking percent is also on basis of stock_qty. 

It's next to qty in stock UOM field so maybe it's fine

Screenshot 2022-04-27 at 12 39 05 PM

ankush: will update label of picked qty field to indicate it's in stock UOM.

marination: We usually back update in stock qty only, saves headache of converting. All % is also calculated w.r.t to stock qty only.

erpnext/stock/doctype/pick_list/pick_list.py
ankush: I will round it down for now. I also highly doubt if there's legit use case of this.

ankush:
TODO:
- [x] consider pending qty = qty - max(delivered, picked)
- [x] dont map picked qty from packing list to PL

marination:
@ankush Dont show "make Pick List" if 100% picked
Rest LGTM

ankush:
@marination done. Skipping CI (only small JS change)

marination (approved):

PR #30774, opened 2022-04-21
fix(postgres): minimum required changes for postgres builds

ankush:
- Ignore duplicate fixtures in India setup (it's ignored by catching exception _anyway_)
- Fix several queries which are blocking creation of test fixtures. 

TODO:
- [ ] Fix stock_ledger.py queries (separate PR) All stock ledger-related queries completely block tests from starting.


ankush:
merging this. Postgres tests are gonna fail for a foreseeable future 😄

PR #30778, opened 2022-04-22
fix: Do not validate while creating accounting dimension

deepeshgarg007:
Accounting dimension creation fails if there is any error in any of the custom fields as all the fields in a doctype are validated while creating any custom field.

To prevent this failure, ignore validation on accounting dimension creation.


deepeshgarg007:

PR #30779, opened 2022-04-22
fix: Shopping cart quotation without website item.

marination:
Front port of https://github.com/frappe/erpnext/pull/29085

**Problem / Motivation**
Our customer is browsing our e-store and cannot find what they are looking for. They call a customer service rep to see if the item is available. The rep adds the item into the quotation and saves it. Without realizing, that item hasn't been added as a Website Item yet. 

The customer goes to visit their shopping cart to realize the web page is broken. Caused by: https://github.com/frappe/erpnext/blob/498d933e9ce7c18d6c595d2e0f575645dc020d81/erpnext/e_commerce/shopping_cart/cart.py#L277-L286
<details>
<summary>Error Snapshot</summary>

image
</details>

**Proposed Resolution**
If the quotation is of type Shopping Cart validate that all item's have a Website Item associated with it.
Screenshot 2022-04-22 at 1 21 03 PM

### DIFF
--- a/erpnext/selling/doctype/quotation/quotation.py
+++ b/erpnext/selling/doctype/quotation/quotation.py
@@ -55,8 +55,20 @@ class Quotation(SellingController):
 			return
 
 		for item in self.items:
-			if not frappe.db.exists("Website Item", {"item_code": item.item_code}):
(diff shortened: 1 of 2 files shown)


marination:
unrelated failure

PR #30781, opened 2022-04-22
fix: translation (german)

wojosc:
fixed the short word for March in german language


marination:
Hi @wojosc I'd encourage giving attention to semantic commits going ahead (commit messages should have the right pre-fix/structure instead of just being one continuous line). 
Here's a great resource https://www.conventionalcommits.org/en/v1.0.0/

marination (approved):

PR #30802, opened 2022-04-25
fix: Allow multi-currency opening invoices

deepeshgarg007:
image

The user gets the above error when while making an opening invoice against a customer having a default current other than the company currency


deepeshgarg007:
@Mgergifyio backport version-13-hotfix

deepeshgarg007:

PR #30806, opened 2022-04-26
fix: Consistent accounting dimensions across Sales and Purchase docs

deepeshgarg007:
image

Add parent level accounting dimensions in Sales Order, Purchase Order, and Purchase Receipts to make accounting dimensions consistent in all Sales and Purchase documents


deepeshgarg007:

jigneshpshah:
From the design perspective, would like to know do @deepeshgarg007  foresee addition of accounting dimensions in child item tables of Sales Order, Purchase Order, Purchase Receipt ?

PR #30807, opened 2022-04-26
test: flaky employee leave test

ankush:
Future tests fail with "unable to find `_T-Leave Allocation-00001`".


ankush:
fixed here: https://github.com/frappe/erpnext/pull/30810

PR #30811, opened 2022-04-26
chore: Warn users about multiple warehouse accounts

marination:
**Issue:**
- During audits or even otherwise tallying an account's value with it's warehouses' value is hard because it's difficult to map exactly how much value is booked in which account.
- [This case](https://github.com/frappe/erpnext/pull/30707) (warehouse value across multiple accounts) is what contributes to it.
- Users don't sometimes understand the unnecessary complexity it creates or the importance of planning account heads before hand

**Solution:**
- Get distinct accounts that warehouse value has been booked against
- If same account as the one being set, ignore
- If not same account or multiple accounts: warn user that it makes it harder to track mismatches
- This is shown on changing or setting warehouse account only (which should ideally not happen often at all!)
   Screenshot 2022-04-26 at 4 49 10 PM

### DIFF
--- a/erpnext/stock/doctype/warehouse/warehouse.py
+++ b/erpnext/stock/doctype/warehouse/warehouse.py
@@ -89,7 +89,7 @@ class Warehouse(NestedSet):
 				.on(ac.name == gle.account)
 				.select(gle.account)
 				.distinct()
-				.where((sle.warehouse == name) & (ac.root_type == "Asset"))
+				.where((sle.warehouse == name) & (ac.account_type == "Stock"))
 				.orderby(sle.creation)
 				.run(as_dict=True)
 			)
--- a/erpnext/stock/doctype/warehouse/warehouse.py
+++ b/erpnext/stock/doctype/warehouse/warehouse.py
@@ -36,6 +36,9 @@ class Warehouse(NestedSet):
 				self.set_onload("account", account)
 		load_address_and_contact(self)
 
+	def validate(self):
+		self.warn_about_multiple_warehouse_account()
+
 	def on_update(self):
 		self.update_nsm_model()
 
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -70,6 +73,53 @@ def on_trash(self):
...
+		# WH account is being changed or set get all accounts against which wh value is booked
+		if self.account != old_wh_account:
+			accounts = get_accounts_where_value_is_booked(self.name)
+			accounts = [d.account for d in accounts]
+
+			if not accounts or (len(accounts) == 1 and self.account in accounts):
+				# if same singular account has stock value booked ignore
+				return
+
+			warning = _("Warehouse's Stock Value has already been booked in the foll.g accounts:")
```
ankush:
```suggestion
			warning = _("Warehouse's Stock Value has already been booked in the following accounts:")
```

erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -70,6 +73,53 @@ def on_trash(self):
...
+
+			return (
+				frappe.qb.from_(sle)
+				.join(gle)
+				.on(sle.voucher_no == gle.voucher_no)
+				.join(ac)
+				.on(ac.name == gle.account)
+				.select(gle.account)
+				.distinct()
+				.where((sle.warehouse == name) & (ac.root_type == "Asset"))
```
marination: the `root_type` condition i'm not sure about. AC Receivables also comes under assets. I thought of using Account Type but Ive noticed a lot of people don't set it
```suggestion
				.where((sle.warehouse == name) & (ac.root_type == "Asset"))
```

cc: @ankush

ankush: It should be similar to this? 

`account_type == "Stock"`

https://github.com/frappe/erpnext/blob/98b28abde0cdca40c48600b4f579494924f7f9cc/erpnext/accounts/utils.py#L1272-L1298

marination:
Skipping tests because its a msgprint

ankush (approved):

PR #30818, opened 2022-04-27
feat: request_for_quotation - show supplier name

ahmadpak:
While trying to create supplier quotation from request for quotation. The dialog box only shows supplier IDs which makes it difficult to identify if supplier ID is different from supplier name. 
So if the supplier series are activated in the buying settings or if supplier ID is not equal to supplier name. It would be very helpful to view both the information while trying to create supplier quotation

In the example below. One has different supplier ID and supplier name and vice versa

image

`no-docs`


erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
```diff
@@ -87,14 +87,20 @@ frappe.ui.form.on("Request for Quotation",{
...
-	make_suppplier_quotation: function(frm) {
+	make_supplier_quotation: function(frm) {
 		var doc = frm.doc;
 		var dialog = new frappe.ui.Dialog({
 			title: __("Create Supplier Quotation"),
 			fields: [
 				{	"fieldtype": "Select", "label": __("Supplier"),
 					"fieldname": "supplier",
-					"options": doc.suppliers.map(d => d.supplier),
+					"options": doc.suppliers.map(d => {
```
deepeshgarg007: @ahmadpak just make this field a link field. Users will be able to configure what they want to see as search fields and supplier names will also be by default visible

deepeshgarg007 (changes requested):

erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
ahmadpak: @deepeshgarg007 Thank you for the feedback. will do so

ahmadpak:
@deepeshgarg007 Can you please check now

erpnext/buying/doctype/request_for_quotation/request_for_quotation.js
```diff
@@ -87,16 +87,24 @@ frappe.ui.form.on("Request for Quotation",{
...
 		var doc = frm.doc;
 		var dialog = new frappe.ui.Dialog({
 			title: __("Create Supplier Quotation"),
 			fields: [
-				{	"fieldtype": "Select", "label": __("Supplier"),
+				{	"fieldtype": "Link",
+					"label": __("Supplier"),
 					"fieldname": "supplier",
-					"options": doc.suppliers.map(d => d.supplier),
+					"options": 'Supplier',
```
marination, line 98:
```suggestion
					"options": "Supplier",
```

marination (approved):

PR #30819, opened 2022-04-27
feat: leave application - exclude non weekly-off/public holidays

ahmadpak:
In leave type we have the option to include holidays within leaves as leaves. By default ERPNext will include all kind of holidays as leaves. While this is very helpful but there are two cases were a company would want to exclude special holidays. 

1. Non weekly-off by a company. It could a company anniversary or celebration of a special day to the company
2. Public holidays.

image

Now we have two new options to exclude from the leave days while applying for leave application. These options would show up only if you check the box include holidays within leaves as leaves. As they are not applicable in the other case
1. Non weekly off
2. Public holidays

image

And in the holiday list we can set appropriate settings with the new checkbox
image

docs: https://docs.erpnext.com/docs/v14/user/manual/en/human-resources/leave-type/edit?wiki_page_patch=f32fa6b2eb


erpnext/hr/doctype/holiday/holiday.json
```diff
@@ -44,17 +45,24 @@
   {
    "fieldname": "section_break_4",
    "fieldtype": "Section Break"
+  },
+  {
+   "default": "0",
+   "fieldname": "public",
+   "fieldtype": "Check",
+   "label": "Public"
```
ruchamahabal, line 53: Holidays that aren't weekly offs are essentially public holidays or those granted by the company. I don't see any incentive in adding an extra checkbox.

Also, in general, this design is not good because I can check both weekly and public at the same time! Always use select fields in such cases where only one should be selected at a time.

erpnext/hr/doctype/leave_type/leave_type.json
```diff
@@ -211,15 +213,30 @@
...
+   "fieldtype": "Check",
+   "label": "Exclude non weekly off holidays within leaves"
+  },
+  {
+   "default": "0",
+   "depends_on": "eval: doc.include_holiday",
+   "fieldname": "exclude_public_holidays_within_leaves",
+   "fieldtype": "Check",
+   "label": "Exclude public holidays within leaves"
   }
```
ruchamahabal, line 230: Again, the design isn't really good. What if I check both these options, then "Include holidays within leaves as leaves" is useless. Would suggest keeping a single checkbox "Exclude weekly offs".

erpnext/hr/doctype/leave_application/leave_application.py
```diff
@@ -126,6 +126,17 @@ def validate_applicable_after(self):
 					holidays = 0
 					if not frappe.db.get_value("Leave Type", self.leave_type, "include_holiday"):
 						holidays = get_holidays(self.employee, date_of_joining, self.from_date)
+					else:
+						if frappe.db.get_value(
```
ruchamahabal, line 130: This has a higher blast radius than just a Leave Application. The "include holidays" setting also affects payment days calculation in Salary Slip.

ruchamahabal (commented):
@ahmadpak Thanks a lot for taking out time to contribute but not sure how important this feature is. IMO, this just adds more fluff that isn't needed. Don't see a lot of people asking for it.

The main incentive for having this feature was for "Sandwich Leave Policy".

If it's for a very narrow use case, can it be handled via server scripts?

erpnext/hr/doctype/holiday/holiday.json
ahmadpak: 1. I have put it that way to differentiate company holidays from public holidays. 
2. Even if we check both checkbox together it still makes sense because there could be a public holiday on a weekend

That is the reason I have not suggested a select box. The extra check box provides better reporting as well

erpnext/hr/doctype/leave_type/leave_type.json
ahmadpak: It will not make Include holidays within leaves as leaves useless. I have explain with an example

Weekly Offs are checked as weekly off in the holidays table. Lets take an example of a week where:

1. Sat and Sun are weekly off
2. Tuesday to Thursday are public holidays
3. Company decided to give Monday as a holiday as well

Include holidays within leaves as leaves checkbox is important to implement sandwich policy
Exclude non weekly off holidays within leaves is important in the case above where company would want to not deduct as un paid leaves and similarly for the public leaves.

I can give you example of KSA. It is law there that you can not deduct public holidays even it is a sandwich leave policy. So we need to exclude those holidays

erpnext/hr/doctype/leave_application/leave_application.py
ahmadpak: Thank you for pointing that out. I can re-check for any negative effects on the Salary Slip. I purposely did that as obviously payment days would change if any holidays are excluded from the leave days

dj12djdjs:
I like the idea of this, having multiple types of holidays. This is good for reporting. I just don't think checkboxes is the way to go. I'm sure there is more than just `Public`, and `Weekly` holidays.

What if we introduced a new doctype called `Holiday Type` which could contain all the options as well as user specified holiday types. Ex. `Public`, `Weekly`, `No pay, no penalty`, `Retreat`, etc... 

Possibly, this could provide additional flexibility in Salary Component calculations.

When `Include holidays within leaves as leaves` the user could select as many holiday types to exclude through a Table Multiselect field.

I don't have any experience using the leave system, just my thoughts.

ruchamahabal:
@ahmadpak Thanks a lot for your time and efforts in making this contribution. Although currently, we don't think we really need this feature as we haven't got a lot of requests around it. The design approach for the feature can also be a lot better to provide more flexibility over rigidity. 

Will have to close this for now. But you can surely open a GitHub issue thread for this where we get enough use cases and impact of this feature so that we find a direction to design this with better UX.

PR #30822, opened 2022-04-27
fix: filters not working in Shift Assignment Calendar view

ruchamahabal:
Manual port of https://github.com/frappe/erpnext/pull/30752 due to multiple changes in v14


ruchamahabal:
Unrelated test failures

PR #30831, opened 2022-04-27
fix: update default_item_manufacturer on rename

govindsmenokee:
update default_item_manufacturer on rename of Manufacturer Doctype

> Screenshots/GIFs


ankush:
IMO this is better fixed by converting the field to Link type. It's data field right now which is why it's not auto updated.

ankush:
closing in favour of https://github.com/frappe/erpnext/pull/30835

PR #30832, opened 2022-04-28
feat(stock): Pick List Scan

### LINKED ISSUE #30821: Scanning in picklist
Add support for updating picked qty using a barcode scanner.

dj12djdjs:
closes: https://github.com/frappe/erpnext/issues/30821

### Proposed Solution
Please note for the examples bellow there are 4 items Item A - Item D with barcode 'a' corresponding to Item A, etc...

- [x] Restrict `BarcodeScanner` from creating new rows.
 By default `BarcodeScanner` will create a row if batch_no or item_code doesn't exist.
 In the case of pick list only the rows generated from sales order or other supported documents
 should be considered when scanning. Ex. We are trying to add Item B through barcode 'b' but it doesn't add row
 because Item B isn't in the list of items to pick.
feat-pick-list-scan_prevent-new-row

- [x] Restrict `BarcodeScanner` from exceeding maximum pickable qty provided by pick list.
  By default `BarcodeScanner` will continue to increment the quantity of the scanned item.
  In the case of pick list a maximum quantity will be reached. I've modified the `BarcodeScanner` opts to support this.
feat-pick-list-scan_max-qty

- [x] Support multiple rows with the same item code.
feat-pick-list-scan_multiple-item_codes

- [x] Prevent Picked Qty from being fulfilled automatically when pick list is submitted. There should be some validation Ex: Picker didn't pick all the items.
dont allow delivery note with full qty picked

- [x] Might be good to have an option prompt for qty after a scan. This way if a picking order calls for 50pcs It won't be necessary to scan all 50pcs. Just one, then input qty to increment. 
prompt qty

**Please post your suggestions / feedback.**


erpnext/stock/doctype/pick_list/pick_list.json
```diff
@@ -126,11 +130,32 @@
...
    "label": "Group Same Items",
    "print_hide": 1
+  },
+  {
+   "fieldname": "column_break_13",
+   "fieldtype": "Column Break"
+  },
+  {
+   "default": "0",
+   "fieldname": "scan_mode",
```
ankush, line 153: I don't think this is required. We can always show the scan_barcode field (which has global setting (in stock settings) to hide it if required 😄 


This was required in stock reconciliation because there the default behaviour and scan behaviour were quite conflicting in nature. Here there are no triggers on picked qty.

ankush:
>  Restrict BarcodeScanner from creating new rows.

Makes sense for picklist. I was gonna add this too. 


>  Restrict BarcodeScanner from exceeding the maximum pickable qty provided by picklist.

This also makes sense. However, IMO this check should be implemented in Pick List's JS and not in the barcode scanner? Do you know anywhere else where max qty validation will be required with barcode scanning?


>  Support multiple rows with the same item code.

Behavior shown in GIF looks good 👍 


>  Prevent Picked Qty from being fulfilled automatically when pick list is submitted. There should be some validation Ex: Picker didn't pick all the items.


Yeah, this will be a good addition. We can just check before_submit if all stock_qty == picked_qty and if not we can warn the user. 


>  Support Simultaneous pickers scanning items for pick list

Not sure if I understand what this means 😅 

Two pickers working on same doc on different computers? I highly doubt that's doable because of the way the framework works right now.

erpnext/stock/doctype/pick_list_item/pick_list_item.json
```diff
@@ -187,11 +188,16 @@
    "hidden": 1,
    "label": "Product Bundle Item",
    "read_only": 1
+  },
+  {
+   "fieldname": "barcode",
```
ankush: This isn't required IMO. 

This was the legacy way to scan barcodes. (Open row > click on barcode field and scan it)

dj12djdjs:
> > Support Simultaneous pickers scanning items for pick list
> 
> Not sure if I understand what this means 😅
> 
> Two pickers working on same doc on different computers? I highly doubt that's doable because of the way the framework works right now.

I kind of figured this. It is one of the requirements I have for my client. I will have to develop some middle man to handle this.

szufisher:
please also consider catch and handle scan same serial number multi times, currently duplicate serial number will be added to serial no field and quantity increase by 1. program need to prevent this and feedback as warning.

ankush:
> I kind of figured this. It is one of the requirements I have for my client. I will have to develop some middle man to handle this.

You can use real-time updates using socket.io for this. Real-time synchronization of normal data structures like JSON (document object) would quite likely lead to conflicts and unexpected behavior 😄 You'll have to implement some kind of locking/synchronization primitives on top of the document object to prevent races. 


> please also consider catch and handle scan same serial number multi times, currently duplicate serial number will be added to serial no field and quantity increase by 1. program need to prevent this and feedback as warning.

This is fixed in develop version. Since it was part of bigger changes it didn't get backported, maybe I should port fixes backto v13 😅

erpnext/stock/doctype/pick_list_item/pick_list_item.json
dj12djdjs: Okay, I will remove this. I only added it because I noticed it was in all other doctypes using barcode.

dj12djdjs:
> This also makes sense. However, IMO this check should be implemented in Pick List's JS and not in the barcode scanner? Do you know anywhere else where max qty validation will be required with barcode scanning?

@ankush I agree with what you're saying. I was originally implementing with that concept in mind, until I came to the realization I needed to be able to scan multiple rows of the same item code. The barcode scan API now relies specifying a maximum quantity so it knows to move to the next item.
https://github.com/dj12djdjs/erpnext/blob/8053f2dbcd835602d86dca520b8629b0ac866656/erpnext/public/js/utils/barcode_scanner.js#L199-L202

dj12djdjs:
> > please also consider catch and handle scan same serial number multi times, currently duplicate serial number will be added to serial no field and quantity increase by 1. program need to prevent this and feedback as warning.
> 
> This is fixed in develop version. Since it was part of bigger changes it didn't get backported, maybe I should port fixes backto v13

@szufisher It seems this might be fixed in develop branch. I'm assuming serial_no has something to do with batch qty. I have no experience with this workflow.

I've setup test site, could you check if the behavior is working as expected for serial no.?
https://feat-picklist-scan.frappe.cloud/
demo@example.com
password

ankush:
> @ankush I agree with what you're saying. I was originally implementing with that concept in mind, until I came to the realization I needed to be able to scan multiple rows of the same item code. The barcode scan API now relies specifying a maximum quantity so it knows to move to the next item.


makes sense. Let's go with max_qty for now, if I can think of a better solution I'll comment / implement later.

dj12djdjs:
> > @ankush I agree with what you're saying. I was originally implementing with that concept in mind, until I came to the realization I needed to be able to scan multiple rows of the same item code. The barcode scan API now relies specifying a maximum quantity so it knows to move to the next item.
> 
> makes sense. Let's go with max_qty for now, if I can think of a better solution I'll comment / implement later.

@ankush I'm going to be implementing barcode scan for Packing Slip as well. Maybe max_qty will be useful over there.

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -10,6 +10,11 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 		this.serial_no_field = opts.serial_no_field || "serial_no";
 		this.batch_no_field = opts.batch_no_field || "batch_no";
 		this.qty_field = opts.qty_field || "qty";
+		this.max_qty_field = opts.max_qty_field || null;
+		this.allow_new_row = opts.allow_new_row;
+		if (this.allow_new_row === undefined || this.allow_new_row === null) {
+			this.allow_new_row = true;
+		}
```
ankush: Using negation of this is simpler. 

`this.dont_allow_new_row` which by default is correct `undefined`.

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -72,6 +77,15 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 		}
 
 		if (!row) {
+			if (!this.allow_new_row) {
+				frappe.show_alert({
+					message: __("Maximum quantity scanned for this barcode."),
```
ankush:
```suggestion
					message: __("Maximum quantity scanned for item {0}.", [item_code]),
```

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -42,8 +42,13 @@ def before_submit(self):
...
 			# if the user has not entered any picked qty, set it to stock_qty, before submit
-			if item.picked_qty == 0:
-				item.picked_qty = item.stock_qty
+			if item.picked_qty < item.stock_qty:
+				frappe.throw(
+					_("Row {0} is short by {1} {2}").format(
+						item.idx, item.stock_qty - item.picked_qty, item.stock_uom
+					),
+					title=_("Pick List Incomplete"),
+				)
```
ankush, line 50: This breaks current behavior for those who don't scan things. They just want to pick 100% qty and submit the pick list.

erpnext/stock/doctype/pick_list/test_pick_list.py
```diff
@@ -411,6 +411,8 @@ def test_pick_list_for_items_with_multiple_UOM(self):
 			}
 		)
 		pick_list.set_item_locations()
+		for location in pick_list.locations:
+			location.picked_qty = location.stock_qty
```
ankush: Modifying tests like this is a red flag 😬  it is indeed breaking an existing workflow.

erpnext/stock/doctype/pick_list/pick_list.py
ankush: we might have to add "scan_mode" afterall to differentiate between users adding qty via scanning vs auto-picking (current behavior)

ankush (changes requested):

erpnext/public/js/utils/barcode_scanner.js
ankush: alternatively: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

erpnext/stock/doctype/pick_list/test_pick_list.py
dj12djdjs: Resolved by https://github.com/frappe/erpnext/pull/30832/commits/4a19c1c19d23bd833d845631a0ddc8a04141a32d

erpnext/public/js/utils/barcode_scanner.js
dj12djdjs: Resolved by : https://github.com/frappe/erpnext/pull/30832/commits/7ae89dedd59815cc0dea63a3ecf67b2355ea4425

erpnext/stock/doctype/pick_list/pick_list.py
dj12djdjs: Resolved by https://github.com/frappe/erpnext/pull/30832/commits/d48ab81622de00c1f088caa435468476cfd9c727

dj12djdjs:
@ankush I've added another feature for prompting qty. It will be useful so user doesn't have to scan an item so many times for high qty. Let me know your thoughts.

ankush:
> @ankush I've added another feature for prompting qty. It will be useful so user doesn't have to scan an item so many times for high qty. Let me know your thoughts.


We have also had a similar request, though UX for this needs to be reviewed. Some users would want to scan everything, and some would want a prompt. 

There are even cases where users want prompt for buying cycle (where large qty are bought) but individual scans for the selling cycle where qty is usually small. 😬 



cc: @marination

dj12djdjs:
> > @ankush I've added another feature for prompting qty. It will be useful so user doesn't have to scan an item so many times for high qty. Let me know your thoughts.
> 
> We have also had a similar request, though UX for this needs to be reviewed. Some users would want to scan everything, and some would want a prompt.
> 
> There are even cases where users want prompt for buying cycle (where large qty are bought) but individual scans for the selling cycle where qty is usually small. 😬
> 
> cc: @marination

These are also very useful, without these features barcode scan feature is weak. Although, I think I should publish these in a separate PR as they are more for the whole of the barcode scan API.

erpnext/stock/doctype/pick_list/pick_list.json
```diff
@@ -126,11 +131,38 @@
    "fieldtype": "Check",
    "label": "Group Same Items",
    "print_hide": 1
+  },
+  {
+   "fieldname": "section_break_15",
+   "fieldtype": "Section Break"
```
ankush, line 138: This section can be hidden when doc is non-draft.

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -182,7 +209,11 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 
 	get_row_to_modify_on_scan(item_code) {
 		// get an existing item row to increment or blank row to modify
-		const existing_item_row = this.items_table.find((d) => d.item_code === item_code);
+		const existing_item_row = this.items_table.filter((d) => {
+			const [qty, max_qty] = [d[this.qty_field], d[this.max_qty_field] || null];
+			return d.item_code === item_code && ((max_qty === null) || (qty < max_qty));
+		}).splice(0, 1).pop();
```
ankush: This max_qty magic is only happening for non-batched items. There's a separate function for fetching row that match item_code and batch_no both, right above this function.

erpnext/stock/doctype/pick_list/pick_list.json
ankush: The alignment of these fields is bit ugly but idk if we can do anything about it, so lets ignore for now. 


Screenshot 2022-05-07 at 5 20 16 PM

erpnext/stock/doctype/pick_list/pick_list.py
```diff
@@ -41,8 +41,15 @@ def before_save(self):
 	def before_submit(self):
 		update_sales_orders = set()
 		for item in self.locations:
-			# if the user has not entered any picked qty, set it to stock_qty, before submit
-			if item.picked_qty == 0:
+			if self.scan_mode and item.picked_qty < item.stock_qty:
+				frappe.throw(
+					_("Row {0} is short by {1} {2}").format(
```
ankush:
```suggestion
					_("Row {0} is short by {1} {2}").format(
```

Better wording than "short by" 😅 

Perheps: `Row {} picked qty is less than required qty, additional {pending_qty} {uom} required.`

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -106,9 +119,23 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
...
-		frappe.model.set_value(row.doctype, row.name, item_data);
+		return new Promise(resolve => {
+			const increment = (value = 1) => {
+				const item_data = {item_code: item_code};
+				item_data[this.qty_field] = Number((row[this.qty_field] || 0)) + Number(value);
+				frappe.model.set_value(row.doctype, row.name, item_data);
+			};
+
+			if (this.prompt_qty) {
+				frappe.prompt(__("Please enter quantity for item {0}", [item_code]), ({value}) => {
```
ankush, line 117: Side-effect of this, idk if intentional or not: fractional qty scanning is now also possible 😬

erpnext/stock/doctype/pick_list/pick_list.js
```diff
@@ -158,6 +158,18 @@ frappe.ui.form.on('Pick List', {
...
+		const opts = {
+			frm,
+			items_table_name: 'locations',
+			qty_field: 'picked_qty',
+			max_qty_field: 'qty',
+			dont_allow_new_row: true,
+			prompt_qty: frm.doc.prompt_qty
+		};
+		const barcode_scanner = new erpnext.utils.BarcodeScanner(opts);
+		barcode_scanner.process_scan();
```
ankush, line 173: Good to see separating out barcode scanner code actually being useful here 😬

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -10,6 +10,9 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 		this.serial_no_field = opts.serial_no_field || "serial_no";
 		this.batch_no_field = opts.batch_no_field || "batch_no";
 		this.qty_field = opts.qty_field || "qty";
+		this.max_qty_field = opts.max_qty_field || null;
+		this.dont_allow_new_row = opts.dont_allow_new_row;
+		this.prompt_qty = opts.prompt_qty;
```
ankush, line 18: It would be great if we can add basic info about what these fields do. Some are obvious, rest not so much.

ankush (changes requested):
last round of minor changes. Tested locally, mostly LGTM. 

Parking batch/serial UX for later, it's bit awkward in picklist right now.

erpnext/stock/doctype/pick_list/pick_list.json
ankush: We can't do column breaks like this right now due to the way layout works:

`|--|-|-|`

ankush: test comment

ankush: test

ankush: test

ankush (dismissed):

ankush:
@dj12djdjs please update the docs page. There are some differences here for "prompt" and "scan mode"

dj12djdjs:
@ankush  Will do, apologies for the delay

dj12djdjs:
@ankush Should I target the doc for V14 only?

ankush:
@dj12djdjs yes. This depends on changes to barcode scanner that aren't on v13.

dj12djdjs:
docs update https://docs.erpnext.com/compare?wiki_page=8c53d48a87&&compare=0fc7ad55c8

PR #30834, opened 2022-04-28
fix: precision loss when transferring 

ankush:
- Stock Entry transfer relies on "incoming_rate" to determine added value in target warehouse during transfers. 
- Incoming rate was getting rounded off while setting. This means if rate was more _precise_ than currency precision then remaining value wont be added to target warehouse. 
- This results in very small rounding adjustment which in turn results in GLE for stock adjustment account. 

This change disables rounding of incoming rate to preserve incoming value = outgoing value invariant. 

Example (same as test case):
1. Material bought at 156,526 for 20 tonne where stock UOM is KG. So per kg rate is 156.526 (but currency precision is 2 so this would get rounded off to 156.53 when transferring)
2. This rounding off results in small rounding adjustment. 1000 * 156.526 - 1000 * 156.53 = 4.0. 

This only happens when purchase UOM is higher than stock UOM and rate doesn't fit in currency precision.

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.py
@@ -672,7 +672,8 @@ class StockEntry(StockController):
 					batch_no=d.batch_no,
 				)
 
-			d.basic_rate = flt(d.basic_rate, d.precision("basic_rate"))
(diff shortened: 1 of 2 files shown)


marination (approved):

marination:
Tested locally works as expected

PR #30835, opened 2022-04-28
fix: convert default_item_manufacturer to link field

ankush:
better fix for https://github.com/frappe/erpnext/pull/30831 

`default_item_manufacturer` on Item doctype is not a link field so it gets out-of-sync when renamed. 

Fix: convert to link field and discard invalid data.


ankush:
Test not required. Link field renaming is a framework feature.

PR #30837, opened 2022-04-28
fix: Respect system precision for user facing balance qty values

marination:
**Issue:**
> **System Precision**: **3**

Consider the following ledger:
<table>
<th>
Qty Diff
</th>
<th>
Bal Qty (for users)
</th>
<th>
Bal Qty (in db)
</th>

<tr>
<td>0</td>
<td>559.833</td>
<td>559.8327</td>
</tr>

<tr>
<td>-470.84</td>
<td>88.993</td>
<td>88.9927</td>
</tr>
</table>

- Now try subtracting 100 from balance (post a Sales Invoice), it says qty is insufficient and it needs more (100-88.9927)  **11.0073** Qty
- The deficiency shown to users does not respect system precision. It could have 7 places after the decimal too. Really bad UX

**Solution**:
- Round off qty deficiency in system precision so that the negative stock validation considers rounded off precision as thr right value and users can adjust stock as per this value
- E.g. Here users will be prompted to add more **11.007** qty instead of **11.0073**
- The Ledger will now look like:
<table>
<th>
Qty Diff
</th>
<th>
Bal Qty (for users)
</th>
<th>
Bal Qty (in db)
</th>

<tr>
<td>0</td>
<td>559.833</td>
<td>559.8327</td>
</tr>

<tr>
<td>-470.84</td>
<td>88.993</td>
<td>88.9927</td>
</tr>

<tr>
<td>11.007</td>
<td>100.000</td>
<td>99.9997</td>
</tr>

<tr>
<td>-100.000</td>
<td>0.000</td>
<td>-0.0003</td>
</tr>
</table>

--------------
**To the reviewer:**
> The diff of **-0.0002999999999957481** in the last SLE will be rounded up to **0.000** and so no error is raised since diff is 0. This will however **technically** causes some negative stock (that again if rounded becomes 0)

**Todo:**

- [x] Adjust precision for future negative qty validation
- [x] Tests


erpnext/stock/stock_ledger.py
```diff
@@ -381,11 +380,9 @@ def __init__(
...
 		self.build()
 
-	def get_precision(self):
-		company_base_currency = frappe.get_cached_value("Company", self.company, "default_currency")
-		self.precision = get_field_precision(
-			frappe.get_meta("Stock Ledger Entry").get_field("stock_value"), currency=company_base_currency
-		)
+	def set_precision(self):
+		self.flt_precision = cint(frappe.db.get_default("float_precision")) or 2
+		self.currency_precision = cint(frappe.db.get_default("currency_precision")) or 2
```
ankush: This is often not set and derived from number format 😐 

https://github.com/frappe/frappe/blob/a59e5d7aacf2ddcea8885591c191e9c93f7984d0/frappe/model/meta.py#L730-L733

ankush: IMO we should just change this in the framework and always set currency precision. If unset by the user it can be derived from the format, but this weird fallback is just not easily understood by users)

Also this way it's simpler to query too!

marination: @ankush but the fallback is in the case that currency precision is not set and somehow it doesnt get anything from number format also. Most cases its not likely to use the fallback `or 2`

ankush: No, I meant on most sites currency precision is not defined. It's derived from the number format selected by the user. 

Fallback of 2 is fine.

ankush: https://github.com/frappe/frappe/blob/develop/frappe/core/doctype/system_settings/system_settings.json#L162

`currency_precision` has no defaults and is not mandatory also. So on most sites this field wont have any value.

marination:
The test is getting a bit hard. Values set are not honouring system precision, will take a look at EOW

erpnext/stock/stock_ledger.py
ankush: @marination IMO we should not change this rounding behaviour of currency now, it will cause different rounding when new code is applied.

marination: will leave `stock_value` as is

marination: @ankush `stock_value` now uses `get_field_precision` (consistent behaviour as before).
Rest, use precision only for error message data (consistent with what users can actually input to restore data as per precision)

ankush (approved):

yhx5773489:
.qmbox p.p1 {margin: 0.0px 0.0px 0.0px 0.0px; font: 12.0px Helvetica; -webkit-text-stroke: #000000}
.qmbox span.s1 {font-kerning: none}



________________________________
大连新程软件有限公司　　　
郭峰
ADDR :大连市沙河口区会展路115号
（百年汇A座）22楼
TEL: 0411-8475-4616
FAX: 0411-8475-4617
***@***.***
URL:  http://www.newlandsystem.com

PR #30844, opened 2022-04-29
fix(UX): record reason for skipping attendance or marking absent for auto attendance

ruchamahabal:
## Problem

In the Auto Attendance scheduler job there are multiple reasons why an employee is marked Absent or attendance is skipped for check-in records. The user has to compare the calculated working hours with the shift timings and investigate Employee Check-ins to find out why the employee was marked Absent or why attendance was skipped for the employee.

## Fix

**Add comments to the Attendance record when the scheduler runs:**

For not meeting the working hours threshold:

image

Due to missing Employee Checkins:

image

**Add relevant comments to Employee Checkin when the scheduler runs:**

The Skip Auto Attendance checkbox is editable so this comment is required to distinguish whether it was skipped during the scheduler job or skipping was enabled by the user during log creation:

image

_Adding comments of type "Comment" instead of "Info" because comments are visible in the report view and can be exported._

### DIFF
--- a/erpnext/hr/doctype/employee_checkin/employee_checkin.py
+++ b/erpnext/hr/doctype/employee_checkin/employee_checkin.py
@@ -5,7 +5,7 @@
 import frappe
 from frappe import _
 from frappe.model.document import Document
-from frappe.utils import cint, get_datetime
+from frappe.utils import cint, get_datetime, get_link_to_form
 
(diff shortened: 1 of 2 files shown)


ruchamahabal:
`skip_attendance_in_checkins` is already covered in tests. Not adding tests for comments.

PR #30850, opened 2022-04-30
fix(pos): freeze screen while processing pos invoices

HarryPaulo:
In situations where the cashier has many POS Invoice 100+, the screen is very slow because the rules were created in javascript, so it happens that the user clicks more than once on the save button because he thinks that nothing is being done.

When the user clicks the button several times, the expected value of the closing is totally wrong...


HarryPaulo:
waiting for a return

PR #30854, opened 2022-05-01
fix(minor): discount accounting patch

rmehta:

> Screenshots/GIFs

### DIFF
--- a/erpnext/patches/v14_0/discount_accounting_separation.py
+++ b/erpnext/patches/v14_0/discount_accounting_separation.py
@@ -2,8 +2,10 @@ import frappe
 
 
 def execute():
-	doc = frappe.get_doc("Accounts Settings")
-	discount_account = doc.enable_discount_accounting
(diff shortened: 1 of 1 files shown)


erpnext/patches/v14_0/discount_accounting_separation.py
```diff
@@ -2,8 +2,10 @@
 
 
 def execute():
-	doc = frappe.get_doc("Accounts Settings")
-	discount_account = doc.enable_discount_accounting
+	data = frappe.db.sql(
+		'select value from tabSingles where doctype="Accounts Settings" and field="enable_discount_accounting"'
+	)
+	discount_account = data and data[0][0] or 0
 	if discount_account:
```
ankush, line 9: This should be casted to integer, else it's always true. @deepeshgarg007 @rahib-hassan

rmehta:
Nice catch Ankush! https://github.com/frappe/erpnext/pull/30858

PR #30869, opened 2022-05-02
fix: Unlink and delete batch created from stock reco on cancel 

marination:
Port of https://github.com/frappe/erpnext/pull/23192/

### DIFF
--- a/erpnext/stock/doctype/stock_entry/test_stock_entry.py
+++ b/erpnext/stock/doctype/stock_entry/test_stock_entry.py
@@ -656,9 +656,9 @@ class TestStockEntry(FrappeTestCase):
 	def test_serial_batch_item_stock_entry(self):
 		"""
 		Behaviour: 1) Submit Stock Entry (Receipt) with Serial & Batched Item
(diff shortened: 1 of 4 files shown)


erpnext/stock/doctype/stock_entry/test_stock_entry.py
```diff
@@ -653,6 +653,109 @@ def test_serial_cancel(self):
...
+			target="_Test Warehouse - _TC",
+			qty=1,
+			basic_rate=100,
+			batch_no=batch_no,
+		)
+		serial_no2 = get_serial_nos(se2.items[0].serial_no)[0]
+
+		batch_qty = get_batch_qty(batch_no, "_Test Warehouse - _TC", item.item_code)
+		self.assertEqual(batch_qty, 2)
+		frappe.db.commit()
```
ankush: Commit/rollbacks are probably not required. savepoints can be added if reqd 😅

marination: Yup fixing these locally

ankush (approved):

PR #30870, opened 2022-05-02
fix(UX): misleading stock entry lables

### LINKED ISSUE #30867: Field "Value Difference" in Stock Entry doctype calculation wrong way round.
### Information about bug

I think the Value Difference field is supposed to calculate the value added to an item. The text says "Total Value Difference(Out - In) but the code calculates (In - Out)

From stock_entry.py
`def set_total_incoming_outgoing_value(self):`
		`self.total_incoming_value = self.total_outgoing_value = 0.0`
		`for d in self.get("items"):`
			`if d.t_warehouse:`
				`self.total_incoming_value += flt(d.amount)`
			`if d.s_warehouse:`
				`self.total_outgoing_value += flt(d.amount)`

		`self.value_difference = self.total_incoming_value - self.total_outgoing_value`

### Module

stock

### Version

v13.21.0
v13.21.0

### Installation method

manual install

_No response_

ankush:
1. flip field location (consumption on left, receipt on right like warehouses)
2. Change incorrect label on value difference field

closes https://github.com/frappe/erpnext/issues/30867

before
Screenshot 2022-05-02 at 2 43 50 PM

after

image

### DIFF
--- a/erpnext/stock/doctype/stock_entry/stock_entry.json
+++ b/erpnext/stock/doctype/stock_entry/stock_entry.json
@@ -46,9 +46,9 @@
   "items",
   "get_stock_and_rate",
   "section_break_19",
-  "total_incoming_value",
-  "column_break_22",
   "total_outgoing_value",
+  "column_break_22",
+  "total_incoming_value",
   "value_difference",
   "additional_costs_section",
   "additional_costs",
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/stock_entry/stock_entry.json
```diff
@@ -386,15 +386,15 @@
...
+   "label": "Total Outgoing Value (Consumption)",
    "options": "Company:company:default_currency",
    "print_hide": 1,
    "read_only": 1
   },
   {
    "fieldname": "value_difference",
    "fieldtype": "Currency",
-   "label": "Total Value Difference (Out - In)",
+   "label": "Total Value Difference (In - Out)",
```
marination:
```suggestion
   "label": "Total Value Difference (Incoming - Outgoing)",
```
A bit hard to visually cross check right now because instinct was to do LHS - RHS

PR #30879, opened 2022-05-02
fix(india): keyerror while generating e-way bill from an e-invoice

### LINKED ISSUE #30878: Unable to Generate E-Way Bill ( Part A ) due to KeyError(None).
### Information about bug

PROBLEM :
When an IRN is created and then the user tries to generate Eway Bill (Part A) using Generate Eway Bill (E-Invoicing) KeyError(None) is raised and it prevents the user from creating Eway Bill ( Part A ).

STEPS TO REPRODUCE.
Create a sales invoice
E Invoicing > Generate IRN
E Invoicing > Generate E-Way Bill
Leave Mode Of Transport blank
Add transporter (which is all that is required to create E-way Bill Part A ).
Submit 

### Module

regional

### Version

develop

### Installation method

manual install

_No response_

maharshivpatel:
When an IRN is created and the user tries to Generate E-way Bill (Part A) using Generate Eway Bill (E-Invoicing). KeyError(None) is raised due to invoice.mode_of_transport being None from Generate E-Way Bill Dialog and it prevents the user from creating Eway Bill ( Part A ) which can be generated by only providing transporter.

I have made a small change that prevents KeyError and allow the user to create E-Way bill ( Part A ).

This Closes #30878


maharshivpatel:
I have many bug fixes and UI/UX improvements for E-invoice & Eway Bill but as Rushabh suggested on the forum that we should start with small bug fixes so I am creating small issues and fixing them one at a time.

However, this is inefficient so I would like to know your opinion on what should I do? continue with small fixes for now or create one big pull request with many fixes and improvements?

nextchamp-saqib:
Hey @maharshivpatel  Thanks for all the contribution!

I would say it's your choice, you can raise one PR for all the fixes/improvements that describes each major change OR small ones with just like these
Just that it'll take more time merging it, as there might possibly be multiple changes requested

maharshivpatel:
@nextchamp-saqib Thank you for your response then i will continue with small fixes.

PR #30881, opened 2022-05-02
refactor: Assets module

GangaManoj:


erpnext/assets/doctype/asset/asset.py
```diff
@@ -1,100 +1,74 @@
...
 
-		self.status = self.get_status()
+	def after_insert(self):
+		if not self.is_serialized_asset and self.calculate_depreciation:
+			# if this is moved to validate(), an error will be raised while
+			# linking depr schedules to assets during the first save
+			create_depreciation_schedules(self)
+
+	def before_submit(self):
+		super().before_submit()
```
nextchamp-saqib, line 31: You don't need to do this

nextchamp-saqib, line 31: You don't need to do this

PR #30884, opened 2022-05-03
fix: Ignore loan repayments made from salary slip

deepeshgarg007:
Loan repayments made via salary slip should not be considered in any Bank Reconciliation tool as the amount does not actually hit the bank account and the amount is deducted from Salary.


deepeshgarg007:
@Mergifyio backport develop

PR #30897, opened 2022-05-04
fix: Consider party and party type as well in group by consolidated view

deepeshgarg007:
General Ledger was not showing entries party wise and just an incorrect party was shown when group by voucher, add party in consolidated view as well

Journal Entry:
image

Before:
Screenshot 2022-05-04 at 4 03 22 PM

After:
Screenshot 2022-05-04 at 4 02 12 PM


deepeshgarg007:

PR #30908, opened 2022-05-05
feat(india): E Invoice Eway Bill Distance is calculated automatically.

### LINKED ISSUE #30907: feat(india): E Invoice Eway Bill Distance can be calculated automatically.
If the distance is passed as ‘0’ (Zero) and the distance between the PIN codes provided is available in E Way Bill database, the system will replace the distance as per the database. Passing ‘0’ is an indication that the user wants the system to take the distance as per the E Way Bill database

https://docs.ewaybillgst.gov.in/apidocs/version1.03/generate-eway-bill.html ( Validation no 3 )

* When using Adaequare E Invoice API
In ERPNext If the User doesn't provide a distance. By default, it is 0 and the distance is calculated automatically using pin codes but users have no way to have this information. In the case of auto calculation of distance that distance should also be updated in the sales invoice's distance field.

maharshivpatel:
Passing Distance as ‘0’ is an indication that the user wants the eway bill system to take the distance as per the E Way Bill database.

I have added the description to notify the user about this and added functionality to save that distance in Sales Invoice Distance Field & show the auto-calculated distance.

I have also added an alert to let the user know whether the request was successful or not.

* I used msgprint to show the user that distance was auto-calculated. this is a temporary solution to notify the user that distance was auto-calculated until a better non-destructive method is used.

* I am currently working on that and for some reason, sanitize_error_message or raise_error is not working.

This Closes #30907 

`no-docs`

### DIFF
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -99,8 +99,8 @@ erpnext.setup_einvoice_actions = (doctype) => {
 									...data
 								},
 								freeze: true,
-								callback: () => frm.reload_doc() || d.hide(),
-								error: () => d.hide()
+								callback: () => frappe.show_alert({ message:__('E-Way Bill Generated successfully'), indicator:'green'}, 7) && frm.reload_doc() && d.hide(),
+								error: () => frappe.show_alert({ message:__('E-Way Bill was not Generated'), indicator:'red'}, 7)
 							});
 						},
 						primary_action_label: __('Submit')
@@ -202,7 +202,8 @@ const get_ewaybill_fields = (frm) => {
 			'fieldname': 'distance',
 			'label': 'Distance (in km)',
 			'fieldtype': 'Float',
-			'default': frm.doc.distance
+			'default': frm.doc.distance,
+			'description': 'By Default, It will be calculated automatically using Pincodes.<br> (You can also enter it Manually)',
 		},
 		{
 			'fieldname': 'transporter_col_break',
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -99,7 +99,14 @@ erpnext.setup_einvoice_actions = (doctype) => {
 									...data
 								},
 								freeze: true,
-								callback: () => frappe.show_alert({ message:__('E-Way Bill Generated successfully'), indicator:'green'}, 7) && frm.reload_doc() && d.hide(),
+								callback: () => {
+									frappe.show_alert({
+										message: __('E-Way Bill Generated successfully'),
+										indicator: 'green'
+									}, 7);
+									frm.reload_doc();
+									d.hide();
+								},
 								error: () => frappe.show_alert({ message:__('E-Way Bill was not Generated'), indicator:'red'}, 7)
 							});
 						},
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -203,7 +203,7 @@ const get_ewaybill_fields = (frm) => {
 			'label': 'Distance (in km)',
 			'fieldtype': 'Float',
 			'default': frm.doc.distance,
-			'description': 'By Default, It will be calculated automatically using Pincodes.<br> (You can also enter it Manually)',
+			'description': 'Set as zero to auto calculate distance using pin codes',
 		},
 		{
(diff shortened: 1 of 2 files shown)


deepeshgarg007:
@maharshivpatel Many such fixes already exist in this [app](https://github.com/resilient-tech/india-compliance). All the India-related localizations are planned to be moved to this app from version-14 onwards. The fixes you are raising can be in the version-13 branch or this app if you intend to do them in version-14

maharshivpatel:
@deepeshgarg007 link to the app is not working however i am aware of this [app](https://github.com/frappe/erpnext_gst_compliance) but it didn't receive any update for quite some time and some of the recent fixes & activity was happening here so i assumed that when time will come you will separate the existing code into the separate app.

erpnext/regional/india/e_invoice/einvoice.js
```diff
@@ -202,7 +202,8 @@ const get_ewaybill_fields = (frm) => {
 			'fieldname': 'distance',
 			'label': 'Distance (in km)',
 			'fieldtype': 'Float',
-			'default': frm.doc.distance
+			'default': frm.doc.distance,
+			'description': 'By Default, It will be calculated automatically using Pincodes.<br> (You can also enter it Manually)',
```
nextchamp-saqib:
```suggestion
			'description': 'Set as zero to auto calculate distance using pin codes',
```

erpnext/regional/india/e_invoice/einvoice.js
```diff
@@ -99,8 +99,8 @@ erpnext.setup_einvoice_actions = (doctype) => {
 									...data
 								},
 								freeze: true,
-								callback: () => frm.reload_doc() || d.hide(),
-								error: () => d.hide()
+								callback: () => frappe.show_alert({ message:__('E-Way Bill Generated successfully'), indicator:'green'}, 7) && frm.reload_doc() && d.hide(),
```
nextchamp-saqib:
```suggestion
								callback: () => {
									frappe.show_alert({
										message: __('E-Way Bill Generated successfully'),
										indicator: 'green'
									}, 7);
									frm.reload_doc();
									d.hide();
								},
```

erpnext/regional/india/e_invoice/einvoice.js
```diff
@@ -99,8 +99,8 @@ erpnext.setup_einvoice_actions = (doctype) => {
...
+								error: () => frappe.show_alert({ message:__('E-Way Bill was not Generated'), indicator:'red'}, 7)
```
nextchamp-saqib:
```suggestion
								error: () => {
									frappe.show_alert({
										message: __('E-Way Bill was not Generated'),
										indicator: 'red'
									}, 7);
									d.hide();
								}
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -1113,6 +1113,18 @@ def generate_eway_bill(self, **kwargs):
...
+				if res.get("info"):
+					info = res.get("info")
+					# when we have more features (responses) in eway bill, we can add them using below forloop.
+					for msg in info:
+						if msg.get("InfCd") == "EWBPPD":
+							pin_to_pin_distance = int(re.search(r"\d+", msg.get("Desc")).group())
+							frappe.msgprint(
+								_("Auto Calculated Distance is {} KM.").format(str(pin_to_pin_distance)),
+								title="Notification",
+								indicator="green",
```
nextchamp-saqib, line 1125: `msgprint` with `show_alert=1` could be used here?

nextchamp-saqib (changes requested):

erpnext/regional/india/e_invoice/einvoice.js
maharshivpatel: 'E-Way Bill was not Generated ' has trailing whitespace.

erpnext/regional/india/e_invoice/utils.py
maharshivpatel: I have added alert=True.

erpnext/regional/india/e_invoice/einvoice.js
nextchamp-saqib: Can you split this into multiple lines manually?

maharshivpatel: Ok I will do it just now

maharshivpatel: done

nextchamp-saqib (approved):

PR #30911, opened 2022-05-05
fix: German translations for Employee doctype

barredterra:


barredterra:

marination (approved):

PR #30913, opened 2022-05-06
fix: double future qty updates

ankush:
image

This is an edge case, where if you have transactions like these then canceling the middle transaction results in double processing of future quantity. 

The incorrect data resolves itself in some time once the backdated entry is processed. 

This is because:

1. Future qty for those SLE with same posting time are updated in `update_entries_after`

Here all SLE that match posting date time (after removing milliseconds) are processed.
https://github.com/frappe/erpnext/blob/c4d20ce15969e14fdecf952b97fd4a8f20fbc71b/erpnext/stock/stock_ledger.py#L440-L461

2. Same are again updated when doing update future qty.

https://github.com/frappe/erpnext/blob/c4d20ce15969e14fdecf952b97fd4a8f20fbc71b/erpnext/stock/stock_ledger.py#L1301-L1314

Since step 1 updates all current posting time SLEs, only strictly newer SLEs should be updated while updating qty. 

Both these function calls in this order (i.e. this exact code execution path) ONLY occurs when you are making backdated entry. 

TODO:
- [x] check impact on similar query part for stock reco
https://github.com/frappe/erpnext/blob/c4d20ce15969e14fdecf952b97fd4a8f20fbc71b/erpnext/stock/stock_ledger.py#L1377-L1385 

Unrelated fix:

Sort SLEs while picking the next stock reco entry: 
 https://github.com/frappe/erpnext/pull/30913/commits/0fb6062dbcacd6213a64e53b78a3957d4b598838#diff-7a6f48ad1157b2728d67c44c2bdb26439c7eec44a4241b915070f63fef695f95R1382

### DIFF
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -1317,7 +1317,7 @@ def update_qty_in_future_sle(args, allow_negative_stock=False):
 		datetime_limit_condition = get_datetime_limit_condition(detail)
 
 	frappe.db.sql(
-		"""
+		f"""
 		update `tabStock Ledger Entry`
 		set qty_after_transaction = qty_after_transaction + {qty_shift}
 		where
@@ -1328,9 +1328,7 @@ def update_qty_in_future_sle(args, allow_negative_stock=False):
 			and timestamp(posting_date, time_format(posting_time, %(time_format)s))
 				> timestamp(%(posting_date)s, time_format(%(posting_time)s, %(time_format)s))
 		{datetime_limit_condition}
-		""".format(
-			qty_shift=qty_shift, datetime_limit_condition=datetime_limit_condition
-		),
+		""",
 		args,
 	)
 
@@ -1381,6 +1379,7 @@ def get_next_stock_reco(args):
 					and creation > %(creation)s
 				)
 			)
+		order by timestamp(posting_date, posting_time) asc, creation asc
 		limit 1
 	""",
 		args,
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -1303,6 +1303,8 @@ def update_qty_in_future_sle(args, allow_negative_stock=False):
 	datetime_limit_condition = ""
 	qty_shift = args.actual_qty
 
+	args["time_format"] = "%H:%i:%s"
+
 	# find difference/shift in qty caused by stock reconciliation
 	if args.voucher_type == "Stock Reconciliation":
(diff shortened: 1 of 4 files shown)


erpnext/stock/stock_ledger.py
```diff
@@ -1323,12 +1325,8 @@ def update_qty_in_future_sle(args, allow_negative_stock=False):
...
 			and voucher_no != %(voucher_no)s
 			and is_cancelled = 0
-			and (timestamp(posting_date, posting_time) > timestamp(%(posting_date)s, %(posting_time)s)
-				or (
-					timestamp(posting_date, posting_time) = timestamp(%(posting_date)s, %(posting_time)s)
-					and creation > %(creation)s
-				)
-			)
+			and timestamp(posting_date, time_format(posting_time, %(time_format)s))
+				> timestamp(%(posting_date)s, time_format(%(posting_time)s, %(time_format)s))
```
ankush, line 1329: Not sure 100% about this yet, rationale is:
1. update_entries_after updates ALL SLEs that match current timestamp in stripped timestamp format. 
2. So future qty updates should ONLY update entries that are **strictly** greater than posting time.

marination: @ankush In this case with the example in the test case, since both transactions have exactly same time stamp, how do we determine the chronology ?

ankush: Chronology is still determined by these three columns: posting date, posting time, creation


It's highly unlikely and near impossible for two docs to have exact creation time down to milliseconds 😅 (unless it's manually fiddled with)


--

This edge case is happening because user added a new transaction with previous posting time, but frontend doesn't allow exact posting time, so it's before in chronological order even though it's created later. (the same exact test case)



Screenshot 2022-05-06 at 1 33 25 PM

ankush: IMO we should strip down posting time to what's visible even when it's not edited manually? This is very confusing.

ankush: If we were stripping it down to nearest seconds then this confusion and subsequently the bug wouldn't exist 😅

ankush: removed datetime formatting from this report: 

Screenshot 2022-05-06 at 1 40 57 PM

ankush:
unrelated `test_tax_rule` failing (on like 80-90% PRs now, its time has come 🥲 https://github.com/frappe/erpnext/pull/30934 )

PR #28775, opened 2021-12-07
feat: Reconciliation in POS using Credit Note

nemesis189:
<h3>Updates:</h3>

- Adding Sales Invoice credit note with outstanding as an advance payment
- Added 'Credit Note' as a Mode of Payment
- Added a checkbox in POS Invoice called `Ignore Payments for Return` to skip payment validation while creating return credit notes via Point Of Sale
---

<details open>
<summary><b>Pre requisite Feats added:</b></summary>
<ol>
<li>Adding Sales Invoice credit note with outstanding as an advance payment </li>
<ul>
<li>Earlier only Payment entires and Journal Entries could be added as an advance for payment in the Advances table.</li>
<li>Now, any sales invoice credit note with outstanding < 0 can be used for advance payment</li>
</ul>

<li>Added 'Credit Note' as a Mode of Payment:</li>
<ul>
<li>A new Mode Of Payment 'Credit Note' has been added which works only for POS currently, it fetches POS Invoice credit notes with negative outstanding</li>
<ul>
</details>

---

<h3>Feature Description:</h3>

- In Point of Sale now we can make use of a POS invoice credit note with negative outstanding for payments. 
- We can select the 'Credit Note' as the mode of payment where all the POS invoice credit notes(returns) with negative outstanding would be listed down and the total of the selected invoice would be added to the `paid_amount` and the outstanding amount of the selected credit note becomes 0.
- While closing a POS, if CN is used as a mode of payment for any invoice, a third consolidated invoice - **Consolidated Advances Invoice** is created along with the usual two other (consolidated sales invoice and consolidated credit return) consolidated invoices.
---
<details open>
<summary><b>Consider the following scenario (also the steps):</b></summary>
<ol>
<li>The first day multiple transactions(invoice) are made by the customer - Make a new opening entry for POS
<li>Create a few pos invoices(3-4) and close the POS
<li>Now the next day one of the invoices needs to be returned so for that - make a new opening entry again(for a new day) and make return for 1-2 invoices through the UI created in the above steps, **while doing so make sure the checkbox `Ignore Payments For Return` is checked(Add it to pos screen via POS Settings). ** This step will make POS return credit notes with negative outstanding which would be used for reconciliation or as payments
<li>Make a new pos sale and in payments select 'Credit Note' as mode of payment and select one of the credit notes created in the above steps.
<li>Complete the transaction, make another transaction with mode of payment as Credit Note.
<li>Close this POS session by creating a closing entry.
[...]

PR #30927, opened 2022-05-08
fix(accounts): payment entry unallocated amount

### LINKED ISSUE #23936: Unallocated amount is incorrectly calculated in Payment Entry when paying an Invoice in Non Base currency

### LINKED ISSUE #29595: Unallocated amount is inaccurate on multi-currency Payment Entry

dj12djdjs:
Continuation of https://github.com/frappe/erpnext/pull/30504

closes: https://github.com/frappe/erpnext/issues/29595
closes: https://github.com/frappe/erpnext/issues/23936

** Allocating References
This shows the unallocated amount being changes as rows become allocated.

allocate references

** Write Offs
Using the "deductions" table, the unallocated_amount can be manipulated. It's primary use is for rounding, and writeoffs.

maniplute unallocated_amount

** Exchange gain/loss
Using the "losses table, the difference_amount can be manipulated. It's primary use if for declaring the difference between currencies. 
manipluate difference amount


erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -871,16 +871,25 @@ frappe.ui.form.on('Payment Entry', {
...
 
 		if(frm.doc.party) {
-			if(frm.doc.payment_type == "Receive"
+			if(frm.doc.payment_type === "Receive"
 				&& frm.doc.base_total_allocated_amount < frm.doc.base_received_amount + total_deductions
 				&& frm.doc.total_allocated_amount < frm.doc.paid_amount + (total_deductions / frm.doc.source_exchange_rate)) {
-					unallocated_amount = (frm.doc.base_received_amount + total_deductions + frm.doc.base_total_taxes_and_charges
-						- frm.doc.base_total_allocated_amount) / frm.doc.source_exchange_rate;
-			} else if (frm.doc.payment_type == "Pay"
+					unallocated_amount = (
```
dj12djdjs, line 875: Need to look into this again. Just found out Allocated credit notes is not working as expected

dj12djdjs, line 875: Need to look into this. Allocating credit not working as expected

dj12djdjs, line 875: Need to look into this. Allocating credit not working as expected

dj12djdjs:
Disregard comment about allocation for credit note.

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -871,16 +871,25 @@ frappe.ui.form.on('Payment Entry', {
...
+						((frm.doc.base_received_amount + frm.doc.base_total_taxes_and_charges
+							 + (total_deductions > 0 ? total_deductions : 0))
+						/ frm.doc.source_exchange_rate)
+						- frm.doc.total_allocated_amount
+					);
+
+			} else if (frm.doc.payment_type === "Pay"
 				&& frm.doc.base_total_allocated_amount < frm.doc.base_paid_amount - total_deductions
```
dj12djdjs: This is still some instance where this condition results in unallocated_amount = 0 even though there is still a balance remaining to be allocated.

dj12djdjs:
I'm starting to think the only way to solve this is to separate deductions/writeoffs and exchange rate deductions into separate  child tables.

Having a condition that forces unallocated_amount to be zero is not working. It should be produced by addition and subtraction.

The only way I can think to solve this is to subtract exchange deductions from total_deductions. Then calculate the unallocated_amount

dj12djdjs:
Gonna post some screen recordings in a few hours.

dj12djdjs:
I don't know why the linter keeps failing. I've run `pre-commit run --all-files` multiple times and is passing on my machine.

dj12djdjs:
@deepeshgarg007 This is working. Can you please review.

dj12djdjs:
@deepeshgarg007 Just checking up, did you have a chance to review this ?

deepeshgarg007:
@dj12djdjs I am reviewing this and will also take a second opinion internally if we need a separate table for losses

deepeshgarg007:
@dj12djdjs I am reviewing this and will also take a second opinion internally if we need a separate table for losses

dj12djdjs:
> @dj12djdjs I am reviewing this and will also take a second opinion internally if we need a separate table for losses

If you guys can find a way to make it work in 1 table that would be more convenient.
 
I don't have much experience in accounting. This is the only way I could make it work.

dj12djdjs:
@deepeshgarg007 any update from the team on this matter?

dj12djdjs:
@deepeshgarg007 Is there a time frame for this?

OneiricIT:
@dj12djdjs does this have any relation to being able to pull credit note amounts sitting on account as payment method for other invoices?
Reference:
https://github.com/frappe/erpnext/pull/28775

dj12djdjs:
> @dj12djdjs does this have any relation to being able to pull credit note amounts sitting on account as payment method for other invoices? Reference: #28775

@OneiricIT no, this PR is just about unallocated calculation method

deepeshgarg007:
Closing this since a new [PR](https://github.com/frappe/erpnext/pull/31272) is raised for this.

PR #30934, opened 2022-05-09
refactor: tax rule validity query

ankush:
This query is reason behind flaky test `test_tax_rule`. 

posting_date isn't being specified when fetching in doc, but it is when doing it in tests, so two different queries result in two different outcome. 

Screenshot 2022-05-09 at 2 16 14 PM

TODO:
- [x] `posting_date` field doesn't exist on quotation, tax rule is receiving `None`


ankush:
dont port.

PR #30938, opened 2022-05-09
fix: allow to use formatting for the field to_discuss in opportunity

rohitwaghchaure:
**Issue**
The users are receiving emails regarding their upcoming follow-ups on Opportunities.
But the email does not mention the exact Opportunity ID for the follow-up making it a very tedious task for the user to track and identify the exact opportunity.

**Solution**

Formatting on the field to discuss

image

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.py
+++ b/erpnext/crm/doctype/opportunity/opportunity.py
@@ -9,7 +9,7 @@ from frappe import _
 from frappe.email.inbox import link_communication_to_document
 from frappe.model.mapper import get_mapped_doc
 from frappe.query_builder import DocType
-from frappe.utils import cint, cstr, flt, get_fullname
+from frappe.utils import cint, flt, get_fullname
 
 from erpnext.crm.utils import add_link_in_communication, copy_comments
 from erpnext.setup.utils import get_exchange_rate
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:

erpnext/crm/doctype/opportunity/opportunity.py
```diff
@@ -215,20 +215,20 @@ def add_calendar_event(self, opts=None, force=False):
...
-					opts.description = "Contact lead " + cstr(self.party_name)
+					opts.description = f"Contact lead {self.party_name}"
 
 			opts.subject = opts.description
-			opts.description += ". By : " + cstr(self.contact_by)
+			opts.description += f". By : {self.contact_by}"
 
 			if self.to_discuss:
-				opts.description += " To Discuss : " + cstr(self.to_discuss)
+				opts.description += f" To Discuss : {frappe.render_template(self.to_discuss, {"doc": self})}"
```
ankush:
```suggestion
				opts.description += f" To Discuss : {frappe.render_template(self.to_discuss, {'doc': self})}"
```

Syntax error here... does this work locally? 🤔

rohitwaghchaure: Not tested post rebase, fixed

PR #30943, opened 2022-05-09
fix: dont keep milliseconds in posting_time

ankush:
In stock ledger, ties are always broken using creation, hidden milliseconds part is
confusing and can cause unexpected ordering like https://github.com/frappe/erpnext/pull/30913 

This was added here: https://github.com/frappe/erpnext/pull/10975/commits/c50b85fae2c8df147cbda782d040216adcbc43b2#diff-e48d53ef3463d9915ffdf9cc9d34fafb96ff861579bdbcb07243d3a6741dd01eR28 
(looks like just for tests? )


ankush:
will reopen someday later

PR #30950, opened 2022-05-10
fix: Set actual qty and basic rate in SE on warehouse triggers (`get_warehouse_details`)

marination:
**Issue:**
- On creating a Stock Entry from a Material Transfer Material Request, `actual_qty` in the child table is not populated while mapping
- On tinkering with the warehouse as well it is not populated, which is ideally the main trigger for it. It randomly get's populated on tinkering with other fields like qty, rate, etc.
- On create a PR (unsaved) from PO, adjusting the accepted qty does not trigger total/grand total change due to https://github.com/frappe/erpnext/pull/30728. (`apply_price_list` called from `qty()` recalculates totals which is skipped due to PR being a mapped doc)

**Fix:**

- [x] On source/target warehouse trigger in Stock Entry, populate `actual_qty` and `basic_rate` immediately
- [x] On Stock Entry mapping `actual_qty` and `basic_rate` should be populated given source warehouse
- [x] Recalculate totals on qty change in mapped Purchase Receipt


marination:
Unrelated `test_tax_rule` failing

PR #30956, opened 2022-05-10
refactor: use utility methods for docstatus

barredterra:
Use utility methods for docstatus that have been introduced in https://github.com/frappe/frappe/pull/15515


ankush:
@barredterra test failures are related. (loyalty points)

PR #30963, opened 2022-05-10
feat: subscription refactor

hrwX:
- Reopened the PR #29870 here as the branch was messed up.
- Reduced complexity associated with invoice generation function, got rid of `is_postpaid_to_invoice` and `is_prepaid_to_invoice`.
```
def can_generate_new_invoice(self) -> bool:
    if self.cancelation_date:
	    return False
    elif self.generate_invoice_at_period_start and (
	    getdate(frappe.flags.current_date) == getdate(self.current_invoice_start)
	    or self.is_new_subscription()
    ):
	    return True
    elif getdate(frappe.flags.current_date) == getdate(self.current_invoice_end):
	    if self.has_outstanding_invoice() and not self.generate_new_invoices_past_due_date:
		    return False
    
	    return True
    else:
	    return False
```
- The invoices won't be generated randomly, they will only be generated either at the beginning of the period or the end of the period depending on the configuration, eliminating the randomness of earlier implementation.
- Removed redundant child table for saving invoices. Added a link field in Sales and Purchase Invoice
- New Subscription Layout
    image

> no-docs


federicocalvo:
Hi, this is the most important doctype/function for any service company. And it is becoming more and more common, so not only does a refactor make sense, but it already marks a horizon with improvements.

PR #30976, opened 2022-05-11
fix: payments duplicate on pos closing entry

HarryPaulo:
**Problem**:  This error does not always happen, in 98% of the cases the closing always works, only in some situations it happens to duplicate the payment methods.

Sometimes the user will close the cashier and the payments are duplicated, the payment methods referring to the opening are inserted in the JS in an asynchronous way and then the function is called that inserts from the POS Invoice what was sold and each form of payment 

image

We can see in the image selected in blue that the payment methods for sales were first inserted, we can see selected in red the payment methods for opening the POS was inserted later (without the sales values grouped), which should not happen.

**Solution**
Call "get_pos_invoices" after load the opening payments methods.

### DIFF
--- a/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
+++ b/erpnext/accounts/doctype/pos_closing_entry/pos_closing_entry.js
@@ -65,7 +65,6 @@ frappe.ui.form.on('POS Closing Entry', {
 		if (frm.doc.pos_opening_entry && frm.doc.period_start_date && frm.doc.period_end_date && frm.doc.user) {
 			reset_values(frm);
 			frm.trigger("set_opening_amounts");
-			frm.trigger("get_pos_invoices");
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


nextchamp-saqib:
@HarryPaulo 

Can you check if the following changes also works?

```
set_opening_amounts(frm) {
	return frappe.db.get_doc("POS Opening Entry", frm.doc.pos_opening_entry)
	...
}
```


```
frappe.run_serially([
	frm.trigger("set_opening_amounts"),
	frm.trigger("get_pos_invoices"),
])
```

nextchamp-saqib (approved):

HarryPaulo:
@nextchamp-saqib [nextchamp-saqib](https://github.com/nextchamp-saqib)

I thought about using frappe.run_serially, but I know it won't work.

Even though the name is intuitive saying "run_serially", if you make a request with the server (JS is asynchronous, that means that js will call the request and without having the result it will exit and go to the next one).

run_serially waits for the javascript function to run and goes to the next one, if I use this way, I'll have to force and wait for the result of the request, I don't think it's the best alternative.

Better alternative is to call inside the return of the set_opening_amounts function



in this case of my other pull request the run_serially was needed.
https://github.com/frappe/erpnext/pull/30850

HarryPaulo:
@nextchamp-saqib 
Just the return will wait the request

PR #30978, opened 2022-05-12
fix: Multiple fixes in GST reporting

deepeshgarg007:
### 1. SEZ and Export Invoices without payment of tax in GSTR-1 report

Currently, SEZ invoices without payment of tax did not use to show up in the GSTR-1 report since no GST was applied in it, this PR fixes that, SEZ and Export invoices (without payment of tax) also showed tax rates as 0 as no GST was applied. This was incorrect, invoices even without payment of taxes should be reported based on their HSN tax rate

For this have introduced a new table GST Tax rate in GST HSN Code doctype which will be used to fetch GST rates for items in such invoices
 
image

### 2. UOM fix in HSN wise summary of outward supplies

If two items have different UOM but the same HSN code if sold in a single month then in HSN-Wise summary of outward supply on one UOM is shown and the qty is wrong

### 3. Sync tax rates in HSN Codes

Screenshot 2022-05-14 at 12 30 25 PM

Tax rates for some HSN codes have been bootstrapped. The sync rates button can be used to sync the system HSN codes

### 4. Other misc fixes:
- Validation for WOPAY invoices
- Auto set eligibility for ITC in RCM invoices

### DIFF
--- a/erpnext/__init__.py
+++ b/erpnext/__init__.py
@@ -4,7 +4,7 @@ import frappe
 
 from erpnext.hooks import regional_overrides
 
-__version__ = "13.29.0"
+__version__ = "13.27.0"
 
 
 def get_default_company(user=None):
--- a/erpnext/__init__.py
+++ b/erpnext/__init__.py
@@ -4,7 +4,7 @@ import frappe
 
 from erpnext.hooks import regional_overrides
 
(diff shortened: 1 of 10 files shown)


deepeshgarg007:
@Mergifyio backport develop

PR #30987, opened 2022-05-12
fix: prevent bypassing forced valuation rate

ankush:
Internal transfers are equivalent to stock entry of "transfer" type. The rate isn't supposed to be editable. However, by changing the margin the rate change limitation can be bypassed right now. 

if you edit "margin_rate_or_amount" after saving DN then based on the selected margin the rate gets updated which isn't the valuation rate.

Screenshot 2022-05-12 at 6 44 39 PM

### DIFF
--- a/erpnext/controllers/buying_controller.py
+++ b/erpnext/controllers/buying_controller.py
@@ -307,14 +307,15 @@ class BuyingController(StockController, Subcontracting):
 				if self.is_internal_transfer():
 					if rate != d.rate:
 						d.rate = rate
-						d.discount_percentage = 0
-						d.discount_amount = 0
 						frappe.msgprint(
 							_(
 								"Row {0}: Item rate has been updated as per valuation rate since its an internal stock transfer"
 							).format(d.idx),
 							alert=1,
 						)
+					d.discount_percentage = 0.0
+					d.discount_amount = 0.0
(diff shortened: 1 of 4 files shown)


erpnext/controllers/selling_controller.py
```diff
@@ -447,15 +447,16 @@ def set_incoming_rate(self):
 						rate = flt(d.incoming_rate * d.conversion_factor, d.precision("rate"))
 						if d.rate != rate:
 							d.rate = rate
+							frappe.msgprint(
```
ankush, line 450: unrelated: only show alert when rate is changed.

erpnext/controllers/buying_controller.py
```diff
@@ -307,14 +307,15 @@ def set_incoming_rate(self):
...
-						d.discount_amount = 0
 						frappe.msgprint(
 							_(
 								"Row {0}: Item rate has been updated as per valuation rate since its an internal stock transfer"
 							).format(d.idx),
 							alert=1,
 						)
+					d.discount_percentage = 0.0
+					d.discount_amount = 0.0
+					d.margin_rate_or_amount = 0.0
```
ankush, line 318: unrelated: always reset these fields regardless of change in rate.

deepeshgarg007 (approved):

PR #30989, opened 2022-05-12
fix: Fix pro rata calculation for monthly depreciation

GangaManoj:
## Issue

For an Asset that depreciates on a monthly basis, if the Available-for-use Date is the first of a month, but the Depreciation Start Date is on the 15th of that month, then the first and last depreciation amounts only needs to be a fraction  of the standard depreciation amount. In the case of monthly depreciation, the system checks this by checking the value of the variable `has_pro_rata` and the iteration number. However, in the case of existing assets that had already booked their first depreciation, this is not enough, since the first iteration does not correspond to the first depreciation, only the first depreciation in this system.

## Fix

Check the number of depreciations booked, in addition to the iteration number and `has_pro_rata`.

<details>

ISS-21-22-14391

</details>

### DIFF
--- a/erpnext/assets/doctype/asset/asset.py
+++ b/erpnext/assets/doctype/asset/asset.py
@@ -353,10 +353,9 @@ class Asset(AccountsController):
 					if self.allow_monthly_depreciation:
 						# month range is 1 to 12
 						# In pro rata case, for first and last depreciation, month range would be different
-						if (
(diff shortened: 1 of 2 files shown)


nextchamp-saqib:
@GangaManoj 

The new test is passing on `version-13-hotfix`  too

CleanShot 2022-05-13 at 12 26 57@2x

Did you miss enabling `allow_monthly_schedule`?

PR #30990, opened 2022-05-12
fix: Item rate reset on changing posting date

### LINKED ISSUE #30341: Item price resets on document date change
### Information about bug

After creating a sales invoice and adding an item to the items table, the item's (pricelist price) appears in the "Rate" column by default
But if you change the "Rate" price manually, and then change the "Date" of the document, the price resets to the default "pricelist price".

What's expected is that the modified "Rate" of the item stays the same even after the date of the document is changed.

### Module

selling

### Version

ERPNext: v13.20.1 (version-13)

Frappe Framework: v13.20.1 (version-13)

### Installation method

_No response_

_No response_

deepeshgarg007:
fixes: https://github.com/frappe/erpnext/issues/30341

Changing date triggers currency and then a series of cascading triggers start. It triggers `conversation_rate` which the again applies price list and the rate gets reset to price list currency again

Solution: Instead of simply triggering `conversion_rate` set coversion_rate to 1 which will trigger `conversion_rate` only if there is a change

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -1578,7 +1578,6 @@ erpnext.TransactionController = class TransactionController extends erpnext.taxe
 	}
 
 	apply_price_list(item, reset_plc_conversion) {
-		debugger
 		// We need to reset plc_conversion_rate sometimes because the call to
 		// `erpnext.stock.get_item_details.apply_price_list` is sensitive to its value
 		if (!reset_plc_conversion) {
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

ankush (approved):

Archais:
Hi. I've discovered a bug that is caused by the changes that were made in this commit. When creating a Sales Order from a Quotation (with a currency other than base), the variables cause the if statement to trigger the else block. Previously, this was not an issue however, with the new changes the exchange rate is set 1 instead of using the previously specified exchange rate or even at least the one specified in the Currency Exchange table.

PR #31004, opened 2022-05-13
fix: duplicate entries in payment terms report output

ruthra-kumar:
### Fix
Duplicate line items appear in Payment Terms status report for SO's with more than one Items. This change will fix that.


deepeshgarg007:

PR #31017, opened 2022-05-14
fix(stock): don't reserve qty on sales return.

dj12djdjs:
### Before PR
When a sales return is issued, the qty returned will be marked as "Reserved Qty".
beforepr dont reserve qty

### After PR
When a sales return is issued, the qty returned won't reflect reserved qty.
afterpr dont reserver qty full

### References
https://discuss.erpnext.com/t/reserved-item-on-submit-of-sales-order-not-coming-back-to-stock-on-sales-return/51436
https://discuss.erpnext.com/t/after-return-the-product-appears-with-reserved-quantity-in-the-stock-summary/34829


dj12djdjs:
Hoping I don't have to re-write that query using the new query builder 😅

dj12djdjs:
This might need a patch, maybe calling get_reserved_qty and updating each row in tabBin ?

ankush:
>  Hoping I don't have to re-write that query using the new query builder 😅

Not required for fixes. Required for new queries.

ankush:
@dj12djdjs can you add a short test to avoid regression? 

Useful utils:
- `make_item()` to make random item, 
- `make_sales_order`
- `make_delivery_note`
- `make_return_doc`

dj12djdjs:
@ankush I cannot get the test to pass. It works when I do it over the browser, but in the test environment the status of the documents is not even changing. I'm not sure how to proceed.

dj12djdjs:
> @ankush I cannot get the test to pass. It works when I do it over the browser, but in the test environment the status of the documents is not even changing. I'm not sure how to proceed.

Looks like this https://github.com/frappe/erpnext/pull/31017/commits/4d0fa85a755eb86d5011b0f31dcb6d01c0a74f4d fixed it

dj12djdjs:
I need to look into why the test is still failing. I was thinking make_item is generating a unique item based on a hash so there shouldn't be any other stock based on that item. I guess I was wrong.

ankush:
@dj12djdjs will look into it today 👍

dj12djdjs:
@ankush Looks good now

ankush (dismissed):

ankush:
@dj12djdjs this is kind of a breaking behaviour. 


E.g. There are users for whom returned qty means they might have to send more to fulfill the order and hence reservation makes sense for them. 


Others can close the sales order to remove reservations if further deliveries are not to be made.

dj12djdjs:
> @dj12djdjs this is kind of a breaking behaviour.
> 
> E.g. There are users for whom returned qty means they might have to send more to fulfill the order and hence reservation makes sense for them.
> 
> Others can close the sales order to remove reservations if further deliveries are not to be made.

Good point, maybe we can add something in the stock settings. Something like, 
- [ ] Reserve Quantity on Sales Return (Default checked, so users can opt out of this)

LOL, I guess I'm gonna have to convert that query using the Query Builder if that's the case. Should be easier to maintain this condition if the query is dynamically built.

dj12djdjs:
bump

dj12djdjs:
I have no idea how to open this again 🤣

ankush:
Wait, you're unable to open closed PR? 👀 

looks like you pushed a commit after PR got closed https://github.com/frappe/erpnext/pull/31271/commits/c9dce9cb962e7bbda5911cba0ac8d2f4c76f61c2 

still weird, only forced pushed branches don't reopen else it should be fine.

dj12djdjs:
@ankush That's correct. I was thinking if I pushed something to the branch it might re-open. Sadly, it didn't.

Issue #31003, opened 2022-05-13
Fail to save IN time In Attendance marked through Employee Checkin


iamejaaz:
### Information about bug

- Fail to save In time when `Shift Type` settings are:
![image](https://user-images.githubusercontent.com/67804911/168221732-c4fe50ea-b087-4dd5-9f6b-9809080c1e68.png)
- Attendance snapshot ⬇️ 
![image](https://user-images.githubusercontent.com/67804911/168221649-6528fa0b-963d-46c9-90a2-53d51c24c3ff.png)

### Module

HR

### Version

ERPNext: v14.0.0-beta.4 (version-14-beta)
Frappe Framework: v14.0.0-beta.3 (version-14-beta)

iamejaaz:
IMO, Adding `in_time = in_log.time` here will save IN time in Attendance 
![image](https://user-images.githubusercontent.com/67804911/168241750-2cbb5959-fde1-4a3c-845e-444013c93add.png)

PR #31022, opened 2022-05-16
fix: Fail to save IN time In Attendance marked through Employee Checkin

SherinKR:

Fail to save IN time In Attendance marked through Employee Checkin

> Screenshots/GIFs

Issue Reference : #31003


ruchamahabal:
@SherinKR There are extra commits in this PR that are unrelated to the fix that seem to be merged already. Can you update the branch with just the required commit?

ruchamahabal (commented):
@SherinKR Tested this. Thanks a lot for your efforts but it's an incorrect fix just as captured in the failing test.

Consider the following logs:

7:19 OUT
**7:49 IN**
8:49 OUT
8:59 IN
9:29 OUT
9:49 IN
11:19 IN
**12:49 OUT**

In this case, the ones marked in bold should be IN and OUT but in this fix, `in_time` is set after the for loop, meaning all logs have been processed and thus we get the last IN here.

Instead, it should update `in_time` as soon as it finds the first IN i.e. 7:49. 

Raised a fix: https://github.com/frappe/erpnext/pull/31029

PR #31026, opened 2022-05-16
fix: remove item attribute limit from variant selector

ankush:
This limit was raised from 20 to 500, well someone has 500 attribute values too :D 

Soln: just remove this limit totally. 

ref: https://github.com/frappe/erpnext/pull/13155


marination:
Only concern i see is `get_list` taking a while due to unusually large values, that could probably use `freeze: true` but thats also unlikely so will let that be

marination (approved):

PR #31034, opened 2022-05-16
fix: disable pricing rules for internal transfers

ankush:
Extends https://github.com/frappe/erpnext/pull/30987 

Problem: Pricing rules can also bypass the forced valuation rate required by the internal transfers. 
Fix: disable pricing rules on internal transfers completely. 

Check test case and linked PR above for steps to reproduce.

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -148,6 +148,7 @@ class AccountsController(TransactionBase):
 
 		self.validate_inter_company_reference()
 
+		self.disable_pricing_rule_on_internal_transfer()
 		self.set_incoming_rate()
 
 		if self.meta.get_field("currency"):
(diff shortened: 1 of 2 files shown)


erpnext/controllers/accounts_controller.py
```diff
@@ -382,6 +383,14 @@ def validate_inter_company_reference(self):
 				msg += _("Please create purchase from internal sale or delivery document itself")
 				frappe.throw(msg, title=_("Internal Sales Reference Missing"))
 
+	def disable_pricing_rule_on_internal_transfer(self):
+		if not self.get("ignore_pricing_rule"):
```
marination:
```suggestion
		if not self.get("ignore_pricing_rule") and self.is_internal_transfer():
```

ankush: oof. Moved this from some other place.

ankush:
unrelated precision error (in test only) if it happens again, will look into.

PR #31039, opened 2022-05-16
fix: Chart data for monthly periodicity in Cash Flow report

AlexandreLumertz:
fix: Only 'total' value was reported in the chart.
In this case, only the first period was graphed.

Problem: https://prnt.sc/jYG3plkV-XnN


deepeshgarg007:

Issue #29806, opened 2022-02-15
erpnext.utilities.product.get_price get wrong price


gn306029:
### Information about bug

My Item have multiple item price in same price list, and same price list the every record's valid_from and valid_upto is not overlap
like this:
<img width="690" alt="截圖 2022-02-15 下午4 20 39" src="https://user-images.githubusercontent.com/20434666/154022748-7ebe006a-0c45-4169-a564-b03775f770a0.png">

but when i try
<img width="655" alt="截圖 2022-02-15 下午4 35 05" src="https://user-images.githubusercontent.com/20434666/154023678-bbd55bea-873d-400d-aadf-71c454bb8a6b.png">

price_list_rate should be 4.6, because today(2022/02/15) is between 2022-01-01 and 2910-12-31, but result's price_list_rate is 7.5

### Module

other

### Version

Frappe Version: 13
Erpnext Version: 13

szufisher:
checking the extracted code below, it seems current date as filtering condition is missing.

price = frappe.get_all("Item Price", fields=["price_list_rate", "currency"],
			filters={"price_list": price_list, "item_code": item_code})

```
def get_price(item_code, price_list, customer_group, company, qty=1):
	from erpnext.e_commerce.shopping_cart.cart import get_party

	template_item_code = frappe.db.get_value("Item", item_code, "variant_of")

	if price_list:
		price = frappe.get_all("Item Price", fields=["price_list_rate", "currency"],
			filters={"price_list": price_list, "item_code": item_code})

		if template_item_code and not price:
			price = frappe.get_all("Item Price", fields=["price_list_rate", "currency"],
				filters={"price_list": price_list, "item_code": template_item_code})

		if price:
			party = get_party()
			pricing_rule_dict = frappe._dict({
				"item_code": item_code,
				"qty": qty,
				"stock_qty": qty,
				"transaction_type": "selling",
				"price_list": price_list,
				"customer_group": customer_group,
				"company": company,
				"conversion_rate": 1,
				"for_shopping_cart": True,
				"currency": frappe.db.get_value("Price List", price_list, "currency")
			})

			if party and party.doctype == "Customer":
				pricing_rule_dict.update({"customer": party.name})

			pricing_rule = get_pricing_rule_for_item(pricing_rule_dict)
			price_obj = price[0]
```

gn306029:
Thanks your response.

When I modify get_price source code
```python
[...]

PR #31040, opened 2022-05-16
fix: filters - added date filter on get_price()

SherinKR:
    
    Added date filter on get_price() with current date

   Currently in erpnext.utilities.product.get_price() , It doesn't check with any kind of date validation or filters. Added Current Date filter and applied with Item Price Validity.

> Screenshots/GIFs

 Reference Issue : #29806

### DIFF
--- a/erpnext/utilities/product.py
+++ b/erpnext/utilities/product.py
@@ -87,14 +87,24 @@ def get_price(item_code, price_list, customer_group, company, qty=1):
 		price = frappe.get_all(
 			"Item Price",
 			fields=["price_list_rate", "currency"],
-			filters={"price_list": price_list, "item_code": item_code},
+			filters=[
+				["price_list", "=", price_list],
+				["item_code", "=", item_code],
+				["valid_from", "<=", frappe.utils.nowdate()],
+				["valid_upto", ">=", frappe.utils.nowdate()]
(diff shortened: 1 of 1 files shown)


erpnext/utilities/product.py
```diff
@@ -87,14 +87,24 @@ def get_price(item_code, price_list, customer_group, company, qty=1):
 		price = frappe.get_all(
 			"Item Price",
 			fields=["price_list_rate", "currency"],
-			filters={"price_list": price_list, "item_code": item_code},
+			filters=[
+				["price_list", "=", price_list],
+				["item_code", "=", item_code],
+				["valid_from", "<=", frappe.utils.nowdate()],
+				["valid_upto", ">=", frappe.utils.nowdate()],
```
ankush, line 94: This is probably incorrect. 

valid_from and valid_upto both can be empty on a price list, in which case it means "no limit"

ankush (changes requested):

erpnext/utilities/product.py
ankush: @SherinKR why was this comment marked resolved?

deepeshgarg007:
Closing this PR because of no response, please reopen again when issues are resolved

PR #31042, opened 2022-05-16
fix: TypeError in add_indicator_for_multicompany

dpk404:
If a customer has done transactions with multiple companies, the add (+)  buttons in connection dashboard will not appear. This is caused by TypeError during stats area creation for customer. Error msg in console looks like 
**"TypeError: e.dashboard.stats_area.removeClass is not a function"**
which is created by **"frm.dashboard.stats_area.removeClass('hidden');"**
So we can use existing show() function to do the similar job 
`frm.dashboard.stats_area.show();`

### DIFF
--- a/erpnext/public/js/utils.js
+++ b/erpnext/public/js/utils.js
@@ -125,7 +125,7 @@ $.extend(erpnext.utils, {
 	},
 
 	add_indicator_for_multicompany: function(frm, info) {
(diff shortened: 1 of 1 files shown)


ankush (approved):

ankush:
before
Screenshot 2022-05-17 at 11 53 26 AM


after
Screenshot 2022-05-17 at 11 53 09 AM

PR #31043, opened 2022-05-17
fix: reset barcode scan field

dj12djdjs:
The barcode scan field was not doing a reset after each scan, most likely caused by the promise introduced in https://github.com/frappe/erpnext/pull/31018

Not sure how I missed this when I made the initial PR. Anyways, resetting directly with frappe.model is working.

cc: @ankush


ankush:
I can't reproduce the issue, for me it's resetting just fine. 

Can you share steps to reproduce this?

dj12djdjs:
Ha! I think I was at the screen for too long when I made this fix. For whatever reason it's working now. 🤷

PR #31052, opened 2022-05-17
chore: additional of hidden & hide total without tax on taxes.html & total.html respectively.

Douglas-Nkubitu:
"taxes.html", one that displays the taxes on sales/purchase documents, added a hidden custom field, then the additional check on the taxes.html. This enables hiding and unhiding some tax lines on a need basis. For instance, If you have service charge, catering levy, and vat, you can hide one from the print-out.
Sales Taxes   Charges

Then there's "total.html", added hide_total_without_tax custom field on the sales order/quotation/invoice, the idea is if someone wants to hide total without tax when doing the print-out it will be possible.
Hide Total (Without Tax)


erpnext/hooks.py
```diff
@@ -625,3 +625,21 @@
 additional_timeline_content = {
 	"*": ["erpnext.telephony.doctype.call_log.call_log.get_linked_call_logs"]
 }
+
+fixtures = [
```
ankush, line 629: Fixtures are for others to customize. They don't make much sense when making changes in ERPNext core directly.

ankush (commented):
Does this change warrant the addition of a new field? This can also be achieved using a print format right?

erpnext/hooks.py
Douglas-Nkubitu: Yes, hidden and hide_total_without_tax on sales taxes and charges and sales order/quotation/invoice doctype, then also added the two fields on the taxes.html and also total.html.

deepeshgarg007:
@Douglas-Nkubitu I don't think this is needed in core ERPNext at the moment, will close this PR

PR #31054, opened 2022-05-17
fix: Job Card excess transfer behaviour

marination:
**Issue:**
> TLDR: You can transfer as much excess against a Job Card (no way to restrict), which is also inconsistent with the front end's **Material Transfer** button

-  Make sure **Allow Excess Material Transfer** in Mfg Settings is disabled
- Make a Work Order with **Transfer Against: Job Card**  (operations and items) > Submit
- Go to one of the Job Cards (must have items) > Create two Material Transfer entries in **Draft** state
- Submit them one by one. You can submit both and the `transferred_qty` in the Job Card is double the required qty
- This is wrong since there's no way to control accidental over transfer
- Also inconsistent with the front end. We hide **Material Transfer** button if no pending qty to transfer and **Allow Excess Material Transfer** in Mfg Settings is disabled. Server side should also respect that

**Fix:**
- Block excess transfer of items if not allowed in settings.
- Behaviour made consistent with js behaviour (button disappears if not pending and not allowed in settings)
- Test for same case
   Screenshot 2022-05-17 at 5 35 24 PM

> Check is done item wise since total transferred qty in Job Card is not reliable

### DIFF
--- a/erpnext/manufacturing/doctype/job_card/job_card.py
+++ b/erpnext/manufacturing/doctype/job_card/job_card.py
@@ -531,11 +531,9 @@ class JobCard(Document):
 
 		def _validate_over_transfer(row, transferred_qty):
 			"Block over transfer of items if not allowed in settings."
-			allow_excess = frappe.db.get_single_value("Manufacturing Settings", "job_card_excess_transfer")
 			required_qty = frappe.db.get_value("Job Card Item", row.job_card_item, "required_qty")
 			is_excess = flt(transferred_qty) > flt(required_qty)
-
-			if is_excess and not allow_excess:
+			if is_excess:
 				frappe.throw(
 					_(
 						"Row #{0}: Cannot transfer more than Required Qty {1} for Item {2} against Job Card {3}"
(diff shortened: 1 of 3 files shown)


erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -522,23 +526,50 @@ def get_current_operation_data(self):
 			},
 		)
 
-	def set_transferred_qty_in_job_card(self, ste_doc):
+	def set_transferred_qty_in_job_card_item(self, ste_doc):
+		from frappe.query_builder.functions import Sum
+
+		def _validate_over_transfer(row, transferred_qty):
+			"Block over transfer of items if not allowed in settings."
+			allow_excess = frappe.db.get_single_value("Manufacturing Settings", "job_card_excess_transfer")
```
ankush: Move this out to the outer function? So we don't need to do this computation at all if the excess transfer is allowed?

Rest LGTM.

erpnext/manufacturing/doctype/job_card/job_card.py
```diff
@@ -522,23 +526,50 @@ def get_current_operation_data(self):
 			},
 		)
 
-	def set_transferred_qty_in_job_card(self, ste_doc):
+	def set_transferred_qty_in_job_card_item(self, ste_doc):
```
ankush, line 529: also, if possible lets avoid renaming public function 😅

marination: had to be done at some point. The function names almost clash and are so confusing, cant understand the purpose.
There is `set_transferred_qty` and `set_transferred_qty_in_job_card`. Both set transferred qty in job card, now guess how they are different 🥲

PR #31058, opened 2022-05-17
fix: barcode scan resolve after model is updated

dj12djdjs:
It was impossible to validate the scanned row because the row is returned before the model is finished updating. This will only resolve the scanned row after all the fields in the model have been updated.

Here is what the "validation" is:
image

Example before PR:
The row has been updated to 15 qty. However, the resolved scanned row still shows the value before the model was updated.
pickedqty before

Example after PR:
The resolved row is up to date with the model.
pickedqty after

Related: https://github.com/frappe/erpnext/pull/31018

CC: @ankush


erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -92,12 +88,11 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
...
-			this.show_scan_message(row.idx, row.item_code, qty);
-		});
-		this.set_serial_no(row, serial_no);
-		this.set_batch_no(row, batch_no);
-		this.set_barcode(row, barcode);
+		const qty = await this.set_item(row, item_code);
+		this.show_scan_message(row.idx, row.item_code, qty);
+		await this.set_serial_no(row, serial_no);
+		await this.set_batch_no(row, batch_no);
+		await this.set_barcode(row, barcode);
```
ankush: This is practically all sequential code now, so better to use `frappe.run_serially` ?

https://github.com/frappe/frappe/wiki/Developer-Cheatsheet#running-tasks-serially

ankush: One more thing, all dependent triggers are not ran serially, so if any fields have triggers that call DB and update it later then those changes will likely not be present on row you receive 😅

dj12djdjs: Resolved https://github.com/frappe/erpnext/pull/31058/commits/4ee6ab1b62b8c966492d86a4a2a21d5aed97e0b9

ankush (approved):

PR #31062, opened 2022-05-18
fix: Leave Encashment calculations

### LINKED ISSUE #23448: V12 HR, Leave Encashment BUGS.

### LINKED ISSUE #24706: Incorrect Encashable leave count
Description- 
 
1. We have allocated privilege leaves to employee for financial year april 2020 to march 2021

Screenshot (943)

2. Employee has used 2.5 leaves out of that

Screenshot (950)

3. now when We are trying to encash the privilege leaves, it is showing available balance = 4, ideally it should be 1.5 & employee should get encashment for 1.5 leaves
Screenshot (949)

4. If we submit the record employee is getting encashment for 4 leaves &  leave count in employee dashboard is -2.5

1. Allocate leaves-(category leave should be encashable)
2. Take some leaves so leave balance will not be equal to allocated leaves
3. now try to create leave encashment for balanced leaves

### Observed result
System is providing encashment for all the allocated leaves, even if employee has used some leaves & does not have that much leave balance
### Expected result
Leave encashmenyt should be done as per balance leaves, not for allocated leaves`

## Additional information
Leave ledger entry:

Type                             To date     entry | From date | To date | Entry
-- | -- | -- | --
Allocated leaves | 01-01-21 | 31-01-21 | 10
Leave application | 10-01-21 | 10-01-21 | -1
Leave application | 26-01-21 | 27-01-21 | -2
 
Existing: | From date | To date | Entry
  | 01-01-21 | 31-01-21 | 7
  | 05-01-21 | 20-01-21 | -1
 
MR | From date | To date | Entry
  | 01-01-21 | 31-01-21 | 10
  | 05-01-21 | 20-01-21 | 0
 
Expected | From date | To date | Entry
  | 01-01-21 | 31-01-21 | 7
  | 05-01-21 | 20-01-21 | 9

### LINKED ISSUE #28020: Leave encashment from date issue in query in version ERPNext version 12
image

In file expnext/erpnext/hr/doctype/leave_encashment/leave_encashment.py on line number 82 leave_allocation query is not correct and not get the from date. While on line 70 from date is using and due that from date is not working.

ruchamahabal:
## Problem:

The Leave Balance calculation in Leave Encashment is incorrect:

`get_leave_allocation` doesn't fetch `allocation.from_date`:

https://github.com/frappe/erpnext/blob/9fb7b49b435887d7e48505830389ede0d740ccff/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L128-L136

but `from_date` is being accessed in `get_unused_leaves` for getting encashable leave balance which is None:

https://github.com/frappe/erpnext/blob/9fb7b49b435887d7e48505830389ede0d740ccff/erpnext/hr/doctype/leave_encashment/leave_encashment.py#L107-L111

Also, the above formula is not correct for calculating leave balance. It's subtracting unused leaves from allocated leaves. Instead, it should remove used leaves from leaves allocated to get the encashable balance. So if 10 leaves are allocated 6 leaves are used, balance should be 4 but this does 10 - 4 = 6. 🤷🏻‍♀️

This feature seems to be broken since the time it was introduced. It also has a test but the test was incorrect causing no failures. 

## Fix:

Fetch `from_date` in the query

Change the leave balance calculation formula to:

```python
self.leave_balance = (
	allocation.total_leaves_allocated
	- allocation.carry_forwarded_leaves_count
	# adding this because the function returns a -ve number
	+ get_leaves_for_period(
		self.employee, self.leave_type, allocation.from_date, self.encashment_date
	)
)
```

Subtract the leaves taken from leaves allocated to get the balance leaves that can be encashed.

Closes https://github.com/frappe/erpnext/issues/28020, Closes https://github.com/frappe/erpnext/issues/23448, Closes https://github.com/frappe/erpnext/issues/24706

### DIFF
--- a/erpnext/hr/doctype/leave_encashment/leave_encashment.py
+++ b/erpnext/hr/doctype/leave_encashment/leave_encashment.py
@@ -7,7 +7,7 @@ from frappe import _
 from frappe.model.document import Document
 from frappe.utils import getdate, nowdate
 
-from erpnext.hr.doctype.leave_allocation.leave_allocation import get_unused_leaves
+from erpnext.hr.doctype.leave_application.leave_application import get_leaves_for_period
 from erpnext.hr.doctype.leave_ledger_entry.leave_ledger_entry import create_leave_ledger_entry
 from erpnext.hr.utils import set_employee_name, validate_active_employee
 from erpnext.payroll.doctype.salary_structure_assignment.salary_structure_assignment import (
@@ -107,7 +107,10 @@ class LeaveEncashment(Document):
 		self.leave_balance = (
 			allocation.total_leaves_allocated
 			- allocation.carry_forwarded_leaves_count
-			- get_unused_leaves(self.employee, self.leave_type, allocation.from_date, self.encashment_date)
+			# adding this because the function returns a -ve number
+			+ get_leaves_for_period(
+				self.employee, self.leave_type, allocation.from_date, self.encashment_date
+			)
 		)
 
 		encashable_days = self.leave_balance - frappe.db.get_value(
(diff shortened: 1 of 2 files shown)


ruchamahabal:
Also, I don't know if this is how it should be. The **Encashment threshold days** setting seems to be adding some functionality that should work exactly opposite to what it's doing right now maybe?

**As per docs and functionality:**

image

Which organization has a threshold above which you can encash? Like if you have 15 leaves, you cannot encash 5 but you can encash 10? 

image

**Expected (Maybe)**:

Shouldn't it be like; you can only encash up to 5 leaves because 5 is the threshold. Currently, it's very confusing for the user.
Maybe we should have minimum and maximum thresholds as separate fields. I am not sure about how HRs use this. 

cc: @nabinhait

ruchamahabal:
> Expected (Maybe):
> 
> Shouldn't it be like; you can only encash up to 5 leaves because 5 is the threshold. Currently, it's very confusing for the user.
> Maybe we should have minimum and maximum thresholds as separate fields. I am not sure about how HRs use this.

Will look into this in a separate PR.

PR #31064, opened 2022-05-18
feat: added a filter for item group on "get items" stock reconciliation

HarryPaulo:
A filter was added for the Item Group, it is very common for inventory reconciliation to be performed by item group session within the company... so this change will make it much easier.

https://prnt.sc/Mo2tr7ULfGQW


HarryPaulo:
@ankush

PR #31065, opened 2022-05-18
feat: added a filter for item group on "get items" stock reconciliation

HarryPaulo:
A filter was added for the Item Group, it is very common for inventory reconciliation to be performed by item group session within the company... so this change will make it much easier.

https://prnt.sc/Mo2tr7ULfGQW

### DIFF
--- a/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.js
+++ b/erpnext/stock/doctype/stock_reconciliation/stock_reconciliation.js
@@ -104,6 +104,12 @@ frappe.ui.form.on("Stock Reconciliation", {
 				}
 			},
 			{
+				label: "Item Group",
+				fieldname: "item_group",
+				fieldtype: "Link",
+				options: "Item Group"
+			},
+			{
 				label: __("Ignore Empty Stock"),
 				fieldname: "ignore_empty_stock",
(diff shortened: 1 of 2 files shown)


HarryPaulo:
@ankush can you guide me if there is any plugin/extension for VSCode to help with code indentation, to validate linters in development?

dj12djdjs:
@HarryPaulo Install the pre-commit package ERPNext is using as a linter. 

`pip3 install pre-commit`

Then in the ERPNext repo directory before you commit run `pre-commit run --all-files` It will automatically format the code.
I think there is a way to automate it so when you commit it will run. But I've never set that up.

ankush:
`pre-commit install` sets up a git hook to automate it. 

https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

HarryPaulo:
waiting reviewers

ankush:
@HarryPaulo this way we will keep adding one filter at a time, "select items" dialogs already support generic filters which will let user pick any filter they want (check out pick list or sales order for example)



Please send a PR if you want to add that generic filtering.

HarryPaulo:
@ankush okk i will analyze and create it next week, thx.

PR #31069, opened 2022-05-19
fix: remove "scrap %" field

### LINKED ISSUE #28740: Scrap percentage in the BOM child table

ankush:
This field does nothing, there's a scrap items table below that's actually used for specifying scrap.

Screenshot 2022-05-19 at 12 04 41 PM


muchai:
There was a recent pr to fix this: https://github.com/frappe/erpnext/pull/28785

Sometimes, you only know the approximate % of scrap to expect. This feature would have made it easier to capture that.

PR #31072, opened 2022-05-19
perf: BOM Update Tool

### LINKED ISSUE #30960: Performance of BOM Update Tool and Correctness
**Case**: BOM count in the system is in Lakhs. (470K)
- BOM Update Tool times out
- Child rows are only updated for operating cost

Basic refactor was done in https://github.com/frappe/erpnext/pull/30146

ToDo:

- [x]  See how timeout can be optimised and notified

- [ ] Add version log for updates made by Tool (check with perf)

- [x] Fix update_cost in BOM, only op cost db updates child, rest does not get updated. Also rm cost is db_updated twice. Correctness needs to be fixed and tests are needed

marination:
<details><summary> First Draft </summary>

##### `get_boms_in_bottom_up_order`
- Create child-parent map once and fetch value from child key to get parents
- Get parents recursively for a leaf node (get all ancestors)
- Approx. 44 secs for 4lakh 70k BOMS (used to go on for over 5-10 mins earlier, maybe more idk)

##### `update_cost`
- Remove double RM cost calculation
- Update child items in db as well (optionally)
- Don't regenerate exploded items for each BOM while updating cost (only in update cost job/bom function)🤦🏻‍♀️
- Approx 5 minutes for 20k BOMs. [WIP]
</details>

> There were correctness issues that could all be solved with a newer AlGoRiThM, that took preference. So first correctness >> then perf

Consider the following BOMs and their dependencies:

BOM-structure drawio
- While updating all BOMs, we update the **Raw Material BOMs** or leaf nodes first and work our way up
- An important thing here: **We must only update a BOM after all its _Dependencies_ are resolved/updated**
- That means: α (alpha) cannot be updated until A,B,1,2,3,4 are updated
- For this we update BOMs **one level at a time**, and **within a level we process BOMs parallelly as they are independent.**
- BOMs **within each level** are split into batches of 20k (takes 5 mins) and queued for updation
- Once a level is done a Cron Job will prep and start processing the next level

#### Working:
- Given the above example, on submitting the BOM Update Log, **Raw Material BOMs** will be processed first (batch size = 20k). The status will be **In Progress**

  
- These Jobs will keep updating their row in the table in the Log and keep checking if all other jobs of this level are done (via `bom_batches` table)
- The **Current Level** is also maintained in the Progress Section, which will increment as we progress up the levels
- **Once a level is complete** and the cron (every 5 mins) determines a level is complete (all rows of current level have status **Pending** -> **Completed**), it will prepare next level parent boms (dependants) and kickstart the next level

   Screenshot 2022-05-24 at 2 17 49 PM
- **How do we determine "Parent BOMs" ?** : By checking immediate parents with no **_Dependencies_**. Here **1** has immediate parents **A** and **α (alpha)**. **A** can be considered since it has no unresolved/non-updated **_Dependencies_**, but **α (alpha)** has  A and B as unresolved dependencies. **So we proceed with A as parent of 1 in the next level**.
- A Cron Job will run every 5 minutes and check for **In Progress** jobs and start processing the next level **if all jobs of current level are completed**
- The Log will be set as **Complete** when there are no more **_Dependants_/ Parent BOMs** to resolve/update

-----------------------------
 > #### Important change/fix to test
> - Rate updation as per **Raw Material Cost as per** used to not work, which now does
> - Rates are reset rightfully on save and fetched as per Valuation/Last Purchase/Price List
----------------------------
**To Test:**
- Create BOM with 3 levels (RM BOM, Subassembly BOM that uses RM BOM, FG BOM that uses Subassemble BOM)
- Set RM cost as per **Valuation Rate** > Change valuation rate of Item in RM BOM > Update all costs via tool > check
- Set RM cost as per **Valuation Rate** with different BOM currency (USD)> Change valuation rate of Item in RM BOM > Update all costs via tool > check
- Set RM cost as per **Price List** > Change Price List rate of Item in RM BOM > Update all costs via tool > check
- Set RM cost as per **Price List** in different Price List currency > Change Price List rate of Item in RM BOM > Update all costs via tool > check

**Todo:**

- [x] Clearing BOM Update Logs periodically (or atleast the table in it)


marination:
Update cost tests are breaking. Will fix while writing tests

marination:
Was an issue with concurrent jobs and incorrect locking sequence causing lost connections to db server.
Fixed and works as expected.
4 jobs parallelly take ~ 12-14 minutes (4*20k = 80,000 BOMs)

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -42,123 +51,192 @@ def validate_bom_items(self):
...
+	try:
+		doc.db_set("status", "In Progress")
 
+		if not frappe.flags.in_test:
+			frappe.db.commit()
 
-def get_parent_boms(new_bom: str, bom_list: Optional[List] = None) -> List:
-	bom_list = bom_list or []
-	bom_item = frappe.qb.DocType("BOM Item")
+		frappe.db.auto_commit_on_many_writes = 1
```
ankush, line 93: This can be removed now (there are commits at N processed BOMs)

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -42,123 +51,192 @@ def validate_bom_items(self):
...
 
+	set_values_in_log(update_doc.name, values, commit=True)
 
-def get_new_bom_unit_cost(new_bom: str) -> float:
-	bom = frappe.qb.DocType("BOM")
-	new_bom_unitcost = (
-		frappe.qb.from_(bom).select(bom.total_cost / bom.quantity).where(bom.name == new_bom).run()
-	)
+	if frappe.flags.in_test:
+		return current_boms, current_level
```
ankush: Keeping prod and test code similar will help. 

We can use `enqueue(..., now=frappe.flags.in_test)` to run things in tests immediately. Doesn't simulate IRL behavior but "close enough"

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -42,123 +51,192 @@ def validate_bom_items(self):
...
-			update_cost()
+		batch_row = update_doc.append(
+			"bom_batches", {"level": current_level, "batch_no": batch_no, "status": "Pending"}
+		)
+		batch_row.db_insert()
 
-		doc.db_set("status", "Completed")
+		if frappe.flags.in_test:
+			# skip background jobs in test
+			return boms_to_process, batch_row.name
```
ankush: Same comment as before, most flags.in_test conditions (except commits) seem to/should be replacable.

erpnext/manufacturing/doctype/bom_update_log/test_bom_update_log.py
```diff
@@ -55,42 +63,59 @@ def test_bom_update_log_validate(self):
...
 
-		# teardown (undo replace impact) due to commit
-		boms = frappe._dict(
-			current_bom=self.boms.new_bom,
-			new_bom=self.boms.current_bom,
-		)
-		log2 = enqueue_replace_bom(
-			boms=self.boms,
+
+def update_cost_in_all_boms_in_test():
```
ankush, line 61: This *might* be replaceable with actual prod code using `now`. That will be more maintainable in case design/logic changes in future.

erpnext/manufacturing/doctype/bom_update_log/bom_updation_utils.py
```diff
@@ -0,0 +1,224 @@
...
+		# this is only used for versioning and we do not want
+		# to make separate db calls by using load_doc_before_save
+		# which proves to be expensive while doing bulk replace
+		bom_obj._doc_before_save = bom_obj
+		bom_obj.update_exploded_items()
+		bom_obj.calculate_cost()
+		bom_obj.update_parent_cost()
+		bom_obj.db_update()
+		if bom_obj.meta.get("track_changes") and not bom_obj.flags.ignore_version:
+			bom_obj.save_version()
```
ankush:
```suggestion
		bom_obj.save_version()
```

both condition already seem to be part of `Document.save_version`


Also check if version is getting created, since `_doc_before_save` is required which isn't triggered unless post save methods are run... maybe it can be added manually. 


PS: tested locally, no version docs found for bom.

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
```diff
@@ -42,123 +51,192 @@ def validate_bom_items(self):
...
+
+	current_boms = {}
+	values = {}
+
+	if update_doc.status == "Queued":
+		# First level yet to process. On Submit.
+		current_level = 0
+		current_boms = get_leaf_boms()
+		values = {
+			"processed_boms": json.dumps({}),
```
ankush, line 120: This can grow quite large in DB for numbers we are dealing with. Maybe add a weekly/monthly job to discard old successful BOM update logs. 


reminds me of this: https://github.com/frappe/frappe/issues/17014

marination: I can clear this field once the log is complete. Its kinda useless after completion. Its used to mainly check if lower levels were processed while going to next level.

> Maybe add a weekly/monthly job to discard old successful BOM update logs.

How old is old 🙈 ? @ankush

ankush: Most logs are deleted at 30 days intervals in log settings... lemme see if we can speed up doctype agnostic log settings 👀

ankush: Just clearing that field might be cool too 👍

gavindsouza: Just emptying the child table (`bom_batches` ) upon update success should do. Let's not add yet another cleanup job for failed tasks? That bit can be manual - or handled by the likes of "Log Settings".

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: I think you meant to remove it in `update_cost_in_level` 😄

ankush: Yep, this one can stay 😬

erpnext/manufacturing/doctype/bom_update_log/bom_updation_utils.py
ankush: On second thought generating versions for all BOMs, every day seems a bit too much. 😅

ankush:
Tested with deeply nested BOM with different costs for each RM and some ops cost. 

#### RM cost
- Started with everything valued @ 0
- Update all bottom-most item (RM) to some valuation rate and did BOM update (scheduled job trigger manually)

Exploded item cost == individual subassy costs. 
Screenshot 2022-06-09 at 2 05 45 PM

---


#### Ops cost

- Used same BOM, revised and replaced a revision with ops. 
- Ran update tool
Screenshot 2022-06-09 at 2 20 14 PM


Increased levels  (by moving this full subassy on another FG)

Screenshot 2022-06-09 at 2 29 02 PM



Partitioning / update logic LGTM 👍 


---

unrelated feature request (I added these columns in BOM explorer (v. hacky implementation) but this might help in actually exploring costs 😬 )

erpnext/manufacturing/doctype/bom_update_log/bom_updation_utils.py
marination: fixed locally, version was not created because the faux `_doc_before_save` pointed to the bom_obj so there was never any difference between the two. Just had to make a copy and assign to `_doc_before_save`

erpnext/manufacturing/doctype/bom_update_log/bom_update_log.py
marination: @ankush will work on auto clearing logs in a separate PR. Need some clarity on what data should be maintained from user POV. For now `processed_boms` is cleared

ankush (approved):

PR #31075, opened 2022-05-19
feat: make Customer/Supplier Group and Territory optional

barredterra:
_Customer/Supplier Group_ and _Territory_ is not useful for SMBs who operate locally. They always have to fill dummy values or set dummy defaults.

For me, a mandatory field implies that the value is strictly necessary for some piece of code to work properly. Is this the case here? If not, these should be optional and can be made mandatory by the user, if needed.

> no-docs

### DIFF
--- a/erpnext/buying/doctype/supplier/supplier.json
+++ b/erpnext/buying/doctype/supplier/supplier.json
@@ -157,6 +157,7 @@
    "width": "50%"
   },
   {
+   "allow_in_quick_entry": 1,
    "fieldname": "supplier_group",
    "fieldtype": "Link",
    "in_list_view": 1,
@@ -164,8 +165,7 @@
    "label": "Supplier Group",
    "oldfieldname": "supplier_type",
    "oldfieldtype": "Link",
-   "options": "Supplier Group",
-   "reqd": 1
+   "options": "Supplier Group"
   },
   {
    "default": "Company",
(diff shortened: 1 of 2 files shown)


ankush:
@barredterra there's a default value for this in selling settings, they are auto-added in quick entry form. 

Screenshot 2022-05-20 at 10 25 10 AM


> a mandatory field implies that the value is strictly necessary for some piece of code to work properly. Is this the case here?

They have been mandatory for years, so lots of code **assumes** that these fields will exist. 

This won't fail in unit tests easily but in real use you might see some reports breaking.

hrwX:
> @barredterra there's a default value for this in selling settings, they are auto-added in quick entry form.
> 
> Screenshot 2022-05-20 at 10 25 10 AM
> 
> > a mandatory field implies that the value is strictly necessary for some piece of code to work properly. Is this the case here?
> 
> They have been mandatory for years, so lots of code **assumes** that these fields will exist.
> 
> This won't fail in unit tests easily but in real use you might see some reports breaking.

- Easier option to handle this is to set Customer Group and Territory in Selling Settings and Supplier Group in Buying Settings.

barredterra:
@ankush should we try to make them optional in `develop`? I'll check the occurrences. And, being on develop only, we'll have some time to notice if something breaks and fix it.

PR #31077, opened 2022-05-20
fix: Loan repayment entries for payroll payable account

deepeshgarg007:
The user was not getting a proper loan balance against an employee for a loan because no employee was added against the payroll payable account while processing Salary Slip. This is done so that no other employee for eg, an Accountant can directly link the final payable amount to a particular employee.

image

So removed the party type from loan repayment entries and added validation for account type in Payroll Entry.

image


deepeshgarg007:

PR #31085, opened 2022-05-21
fix: change project's actual_start_date fieldtype from Data to Date

sersaber:
I have noticed that project's “actual_start_date” is not shown with the same format that other start/end dates in some reports. See image below.

project_report

This is because the field actual_start_date is defined as "Data" instead of "Date" in the doctype. I have looked through the code and the fields actual_start_date and actual_end_date are managed in the same way, but actual_end_date is defined as "Date" while actual_start_date is defined as "Data". I think this could be a typo.

### DIFF
--- a/erpnext/projects/doctype/project/project.json
+++ b/erpnext/projects/doctype/project/project.json
@@ -458,7 +458,7 @@
  "index_web_pages_for_search": 1,
  "links": [],
  "max_attachments": 4,
- "modified": "2022-01-29 13:58:27.712714",
+ "modified": "2022-05-25 22:45:06.108499",
  "modified_by": "Administrator",
  "module": "Projects",
  "name": "Project",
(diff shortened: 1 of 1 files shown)


ruchamahabal (changes requested):
@sersaber You also need to update the `modified` timestamp to sync the changes on migrate.

sersaber:
I think I have made the change suggested by @[ruchamahabal](https://github.com/ruchamahabal), but I'm not sure if it's correct.

ruchamahabal (approved):

ruchamahabal:
@sersaber A note for future contributions: Commit messages must follow the conventional commit specifications:
https://www.conventionalcommits.org/en/v1.0.0-beta.2/

ankush:
This will likely require a patch in case some bad data is present in this column that can't be converted.

ruchamahabal:
> This will likely require a patch in case some bad data is present in this column that can't be converted.

This field is read-only and is set from timesheets, so only dates would be present in it. No open-ended data

PR #31087, opened 2022-05-22
chore: update ru translate

Altizo:


ankush:
@Altizo please follow semantic commits 

https://www.conventionalcommits.org/en/v1.0.0/

PR #31088, opened 2022-05-22
feat(india): Check E-way Bill Status before Cancellation.

maharshivpatel:
NIC doesn't provide e-way bill cancellation API in production so instead I have made changes to check e-way bill status using **Get E-Waybill Details by IRN** API and update the status accordingly.

I have also created functions to fetch and update e-way bill details in the sales invoice using IRN that can be useful in cases where users have created/canceled e-way bills using the portal.

Note: I will send separate PR  for the client-side implementation of the " fetch and update e-way bill details" feature.
`no-docs`


maharshivpatel:
@nextchamp-saqib 
Hey, it's been over a week did you get time to look at my PR?
Are there any problems with this PR that I can solve?
Is there anything I can do to make this PR merge faster?

PR #24327, opened 2021-01-07
feat: Leave Policy Assignment Refactor

Anurag810:
**Documentation PR:** https://github.com/frappe/erpnext_documentation/pull/321

- [x] Added Employee name in Bul assignment.
- [x] removed automatic assignment
- [x] removed unnecessary state
- [x] Create allocation on submitting
- [x] Added dashboard
- [x] Same small fixes

fix

PR #31097, opened 2022-05-23
fix: remove leave policy assignment creation patch

ruchamahabal:
- This patch was written in https://github.com/frappe/erpnext/pull/23112 where leaves were granted explicitly after creating policy assignments.
- This was refactored in https://github.com/frappe/erpnext/pull/24327 where this redundant "Grant Leaves" button was removed and leaves were granted on submission of policy assignment.
- This patch doesn't seem to be required since policy assignments never used to get created automatically.
- Moreover, creating assignments without Leave Period information, breaks this patch a lot during migrations and there is no other way to patch this.
- If successful, it just allocates leaves arbitrarily on migration based on defaults set in the employee master which is also not correct.


ruchamahabal:
Will look into the failing test `test_unmarked_days` in a separate PR

PR #31100, opened 2022-05-23
fix(india): duplicate qrcode and hide button

maharshivpatel:
Problem: allows qrcode generation even if it exists and makes API call even though data will most likely be available in signed_qr_code field.

Solution: I have added logic on the server & client side to prevent the duplicate generation of qrcode and use signed_qr_code if available this will allow generating qrcode even after 2 days. if all fails it will get qrcode using api.

Note: I have added __unsaved check on the client side this is for future proofing.

> currently, there is no way to change anything in the invoice once you generate IRN however this is not the correct approach as there are many fields that have nothing to do with Invoice details that we send to the IRP ( government ) and we should be able to change fields that don't affect data we sent to the IRP. I am working on it and will send separate PR or this.

### DIFF
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -11,7 +11,7 @@ erpnext.setup_einvoice_actions = (doctype) => {
 
 			if (!invoice_eligible) return;
 
-			const { doctype, irn, irn_cancelled, ewaybill, eway_bill_cancelled, name, __unsaved } = frm.doc;
+			const { doctype, irn, irn_cancelled, ewaybill, eway_bill_cancelled, name, qrcode_image, __unsaved } = frm.doc;
 
 			const add_custom_button = (label, action) => {
 				if (!frm.custom_buttons[label]) {
@@ -175,7 +175,23 @@ erpnext.setup_einvoice_actions = (doctype) => {
 			}
 
 			if (irn && !irn_cancelled) {
+				let is_qrcode_attached = false;
+				if (qrcode_image && frm.attachments) {
+					let attachments = frm.attachments.get_attachments()
+					if (attachments.length != 0) {
+						for (let i = 0; i < attachments.length; i++) {
+							if (attachments[i].file_url == qrcode_image) {
+								is_qrcode_attached = true;
+								break;
+							}
+						}
+					}
+				}
+				if (!is_qrcode_attached || !qrcode_image) {
 				const action = () => {
+					if (frm.doc.__unsaved) {
+						frappe.throw(__('Please save the document to generate QRCode.'));
+					}
 					const dialog = frappe.msgprint({
 						title: __("Generate QRCode"),
 						message: __("Generate and attach QR Code using IRN?"),
@@ -195,6 +211,7 @@ erpnext.setup_einvoice_actions = (doctype) => {
 					dialog.show();
 				};
 				add_custom_button(__("Generate QRCode"), action);
+			}
 			}
 		}
 	});
--- a/erpnext/regional/india/e_invoice/einvoice.js
+++ b/erpnext/regional/india/e_invoice/einvoice.js
@@ -177,7 +177,7 @@ erpnext.setup_einvoice_actions = (doctype) => {
 			if (irn && !irn_cancelled) {
 				let is_qrcode_attached = false;
 				if (qrcode_image && frm.attachments) {
-					let attachments = frm.attachments.get_attachments()
+					let attachments = frm.attachments.get_attachments();
 					if (attachments.length != 0) {
 						for (let i = 0; i < attachments.length; i++) {
 							if (attachments[i].file_url == qrcode_image) {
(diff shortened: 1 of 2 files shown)


erpnext/regional/india/e_invoice/utils.py
```diff
@@ -1010,13 +1010,29 @@ def bulk_generate_irn(invoices):
...
 
 	def fetch_and_attach_qrcode_from_irn(self):
-		qrcode = self.get_qrcode_from_irn(self.invoice.irn)
-		if qrcode:
-			qrcode_file = self.create_qr_code_file(qrcode)
-			frappe.db.set_value("Sales Invoice", self.invoice.name, "qrcode_image", qrcode_file.file_url)
-			frappe.msgprint(_("QR Code attached to the invoice"), alert=True)
-		else:
-			frappe.msgprint(_("QR Code not found for the IRN"), alert=True)
+		is_qrcode_file_attached = frappe.db.exists(
```
```suggestion
		is_qrcode_file_attached = self.invoice.qrcode_image and frappe.db.exists(
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -1010,13 +1010,29 @@ def bulk_generate_irn(invoices):
...
+				frappe.msgprint(_("QR Code attached to the invoice."), alert=True)
+			else:
+				qrcode = self.get_qrcode_from_irn(self.invoice.irn)
+				if qrcode:
+					qrcode_file = self.create_qr_code_file(qrcode)
+					frappe.db.set_value("Sales Invoice", self.invoice.name, "qrcode_image", qrcode_file.file_url)
+					frappe.msgprint(_("QR Code attached to the invoice."), alert=True)
+				else:
+					frappe.msgprint(_("QR Code not found for the IRN"), alert=True)
+		elif is_qrcode_file_attached and self.invoice.qrcode_image:
```
```suggestion
		else:
```

erpnext/regional/india/e_invoice/einvoice.js
```diff
@@ -175,27 +175,44 @@ erpnext.setup_einvoice_actions = (doctype) => {
...
+						for (let i = 0; i < attachments.length; i++) {
+							if (attachments[i].file_url == qrcode_image) {
+								is_qrcode_attached = true;
+								break;
 							}
-						},
+						}
+					}
+				}
+				if (!is_qrcode_attached || !qrcode_image) {
```
nextchamp-saqib:
```suggestion
				if (!is_qrcode_attached) {
```

If `qrcode_image` is not set then `is_qrcode_attached` will always be `false`

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -1010,13 +1010,29 @@ def bulk_generate_irn(invoices):
...
+			{
+				"attached_to_doctype": "Sales Invoice",
+				"attached_to_name": self.invoice.name,
+				"file_url": self.invoice.qrcode_image,
+				"attached_to_field": "qrcode_image",
+			},
+		)
+		if not is_qrcode_file_attached:
+			if self.invoice.signed_qr_code:
+				self.attach_qrcode_image(update_url=True)
```
```suggestion
				self.attach_qrcode_image()
				frappe.db.set_value("Sales Invoice", self.invoice.name, "qrcode_image", self.invoice.qrcode_image)
```

We can avoid the `kwarg` if it is only used once

nextchamp-saqib (changes requested):

nextchamp-saqib:
> currently, there is no way to change anything in the invoice once you generate IRN however this is not the correct approach as there are many fields that have nothing to do with Invoice details that we send to the IRP ( government ) and we should be able to change fields that don't affect data we sent to the IRP. I am working on it and will send a separate PR or this.

I had implemented a [validation](https://github.com/frappe/erpnext_gst_compliance/blob/bd78ba4556f524fa4f8a6a87a2f21e6b202aa37c/erpnext_gst_compliance/erpnext_gst_compliance/doctype/e_invoice/e_invoice.py#L667-L691) for the same, it compares the e-invoice JSON before and after a field was modified, if the before & after JSON was the same, then the modification was allowed. 

However, this isn't needed anymore right? Once the e-invoice is generated on invoice submission, the invoice need not be edited anymore

Or do you foresee any other scenario where this validation would be needed?

maharshivpatel:
> Or do you foresee any other scenario where this validation would be needed?

Current `if irn` will make the **"allow on submit"** fields not editable so this [validation](https://github.com/frappe/erpnext_gst_compliance/blob/bd78ba4556f524fa4f8a6a87a2f21e6b202aa37c/erpnext_gst_compliance/erpnext_gst_compliance/doctype/e_invoice/e_invoice.py#L667-L691) will be needed.


For example, The issue that I am working on right now is that the user doesn't want to create the e-way bill as the invoice value is less than 50000.

if the user adds transport details then an e-way bill will be generated so the temporary fix would be the user would add it after IRN generation but this is not allowed due to the `if irn` statement.

>  I am working on a more user-friendly solution for the above issue. 


There is [e_inovice](https://github.com/frappe/erpnext/tree/develop/erpnext/regional/india/e_invoice) inside regional that is supposed to be removed in v14 then there is the **[erpnext_gst_compliance](https://github.com/frappe/erpnext_gst_compliance)** app and maybe there is a private repo that @deepeshgarg007  mentioned [here](https://github.com/frappe/erpnext/pull/30908#issuecomment-1119214422) so which one of these will be available in v14 as in erpnext_gst_compliance there aren't any commits after Jan 6, 2022, so it lacks some of the important fixes and is not tested yet in production. we (the public) don't have access to the repo that deepeshgarg007 mentioned

nextchamp-saqib:
> Current `if irn` will make the **"allow on submit"** fields not editable so this [validation](https://github.com/frappe/erpnext_gst_compliance/blob/bd78ba4556f524fa4f8a6a87a2f21e6b202aa37c/erpnext_gst_compliance/erpnext_gst_compliance/doctype/e_invoice/e_invoice.py#L667-L691) will be needed.

We can remove the validation altogether. If IRN will be generated on submission, then modifying an invoice value after submission is not possible. So is there a need to keep the validation that some fields of the invoice cannot be modified?

> so which one of these will be available in v14

Currently, the changes in the current `develop` will be available in v14. Both of those apps are not yet in production.

nextchamp-saqib (approved):

PR #31101, opened 2022-05-23
fix: Improve button labels in Warehouse

barredterra:
File was formatted like 💩 
-> Format. I know this is discouraged, but at some point somebody has to take the blame...

Labels "Group to Non-Group" / "Non-Group to Group" were not descriptive.
-> Changed to "Convert to Ledger" / "Convert to Group".

Context was missing for translations.
-> Added "Warehouse" context.

Queries were set during `onload` and `refresh`.
-> Set queries during `setup`.

Also fixed german translations ("Convert to Group" already existed, therefore I just removed the old label).


barredterra:
@backport version-13-hotfix

ankush:
> File was formatted like 💩


we should introduce autoformat for JS too at some point 🥲

I like prettier but it makes some files totally shit in frappe codebase. Need to tweak the config so it's tolerable by _most_ devs.

erpnext/stock/doctype/warehouse/warehouse.js
```diff
@@ -1,88 +1,96 @@
...
-			}
+		frm.add_custom_button(
+			frm.doc.is_group
+				? __("Convert to Ledger", null, "Warehouse")
+				: __("Convert to Group", null, "Warehouse"),
+			function () {
+				convert_to_group_or_ledger(frm);
+			},
+			"fa fa-retweet",
+			"btn-default"
```
ankush:
```suggestion
```

Not sure why is this added? Icons aren't supported


Screenshot 2022-05-24 at 10 47 56 AM

ankush (approved):

PR #31107, opened 2022-05-24
refactor: Accounts Receivable/Payable Report will use Payment Ledger

ruthra-kumar:
## Accounts Receivable/Payable Report
Change in this PR
- Accounts Receivable/Payable report has been refactored to use Payment Ledger as data source
- unit test for AR/AP report refactored as well.


centaur78:
Hi @ruthra-kumar

I getting the below reports... where both the Accounts Receivable and Payable have entries dating back from 2020 which are not show here. This was showing when GL entry was used as source

image

and  

image


Error 

image

ruthra-kumar:
@centaur78 Check if the old GL postings have been migrated to Payment Ledger. If not run patch 'erpnext.patches.v14_0.migrate_gl_to_payment_ledger' on your site. If you are still facing this issue, create an issue ticket

centaur78:
Hi @ruthra-kumar 

How do you run a specific patch like 'erpnext.patches.v14_0.migrate_gl_to_payment_ledger' ??

I did bench update --patch but it still shows the same result

ankush:
Check patch log doctype, and see if there are any with `migrate_gl_to_payment_ledger` already. If so it has already ran on your site.

centaur78:
> Check patch log doctype, and see if there are any with `migrate_gl_to_payment_ledger` already. If so it has already ran on your site.

Hi @ankush 
yes just checked... there is no erpnext.patches.v14_0.migrate_gl_to_payment_ledger in the logs ..

So how would I run this patch specifically ?

ruthra-kumar:
@centaur78 

use
`bench --site <sitename> run-patch --force erpnext.patches.v14_0.migrate_gl_to_payment_ledger`

You have to clear 'Payment Ledger Entry' table beforehand using DB console though. Otherwise, you will get duplicate entries.

PR #31113, opened 2022-05-24
fix: don't fetch item_code if already exists.

dj12djdjs:
The problem is if you're processing a return where an old barcode no longer exists in the item master. You'll get an error Barcode for item X not found. 

The solution, if the item_code already exists, no need to fetch it.


ankush:
Ah, I came across this just recently!

ankush (approved):

PR #31115, opened 2022-05-25
feat: packing slip barcode scanning

dj12djdjs:
## Changes to Existing Features
1. Refactored old `cscript` to `frappe.ui.form.on`
2. Moved "Get Items" button from above "items" table into the forms custom button container.

**Before**
image

**After**
image

## New Features
1. Barcode Scanning
The barcode scan field will only be displayed if the delivery note is selected, and packing no. is set.
This is because `delivery_note`, `from_case_no`, and `to_case_no` are all required to validated scanned row.

2. Barcode Scan Validation
Scan Barcode - Packing Slip

**Note**: These validations did already exist, but only validated when packing slip is saved. This is not practical when using barcode scanner. If there is a error in the scanned item the operator needs to be notified immediately, and not when the pallet has been fully packed and operator attempts to save and submit the packing slip. 

## Showcase
1. Scanning an item that doesn't belong to the selected delivery note.
invalid item code

2. Over scan qty of a item while packing. In this example `Item B` 100 units are being delivered.
over qty

Will update docs if PR gets approved. Looking forward to hearing your thoughts.

### DIFF
--- a/erpnext/stock/doctype/packing_slip/packing_slip.js
+++ b/erpnext/stock/doctype/packing_slip/packing_slip.js
@@ -11,7 +11,7 @@ frappe.ui.form.on("Packing Slip", {
 	refresh(frm) {
 		frm.toggle_display("misc_details", frm.doc.amended_from);
 
-		if (frm.doc.delivery_note && frm.doc.status === 0) {
+		if (frm.doc.delivery_note && frm.doc.docstatus === 0) {
 			frm.add_custom_button(__("Get Items"), () => get_items(frm));
 		}
 	},
--- a/erpnext/stock/doctype/packing_slip/packing_slip.js
+++ b/erpnext/stock/doctype/packing_slip/packing_slip.js
@@ -51,12 +51,12 @@ frappe.ui.form.on("Packing Slip", {
 		// this as a fallback to reference.
 		const backup = frm.doc.items.map(item => ({...item}));
 
-		function after_scan(scanned_row, event) {
-			if (event === "remove") {
+		function after_scan(scanned_row, exc_type) {
+			if (exc_type === "InvalidPackingSlipItem") {
 				frm.get_field("items").grid.grid_rows[scanned_row.idx - 1].remove();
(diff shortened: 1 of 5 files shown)


erpnext/stock/doctype/packing_slip/packing_slip.js
```diff
@@ -1,51 +1,109 @@
...
+				frappe.msgprint(__("Item {0} packed quantity has been reverted to {1}",
+					[row.item_code, row.qty]));
+			}
+		}
+
+		barcode_scanner.process_scan().then(scanned_row => {
+			frm.call({
+				doc: frm.doc,
+				method: "validate_scanned_item",
+				args: {scanned_row}
```
dj12djdjs: `args: {scanned_row}` isn't required. It will be removed.

erpnext/stock/doctype/packing_slip/test_packing_slip.py
```diff
@@ -1,11 +1,8 @@
...
 # See license.txt
 
-import unittest
-
-# test_records = frappe.get_test_records('Packing Slip')
 from frappe.tests.utils import FrappeTestCase
 
 
-class TestPackingSlip(unittest.TestCase):
+class TestPackingSlip(FrappeTestCase):
```
dj12djdjs, line 7: I'm not sure if testing will be required for packing slip... Maybe UI testing for barcode scan validation.

dj12djdjs: I suppose it could be good to add some tests to ensure qty is correct with multiple packing slips against a delivery note.

ankush:
@dj12djdjs can you rebase the branch to remove merged commits? Also check sider issues.

dj12djdjs:
@ankush rebase done

dj12djdjs:
@ankush  I've thought of some flaws with this design. I'm going to convert to draft for now. I'll update tasks for this PR.

dj12djdjs:
This was a bit pre-mature, there is a lot of working to do on packing slip. We're going to demo in our organization for some time. Once we get the correct flow, I'll re-open the PR .

PR #31126, opened 2022-05-25
chore: update translation fr for BOM

FHenry:
fix: update translation fr for BOM

Note to reviewer, yes google says (in the past) 
Bill of material = Liste des matériaux 
but now it's are more accurate  
image

I can assure you, the is no ERP in french that use "Liste des matériaux" to describe the BOM feature.
"Liste des matériaux" can exists but it will be more "picking list" (that will be another PR ;-)


ankush:
Thanks, these do make more sense 😄

FHenry:

PR #31129, opened 2022-05-25
fix: Loan Doc query in Bank Reconciliation Statement

deepeshgarg007:
`Repay From Salary` condition was not getting applied in the query, rewrote the query to fix this


deepeshgarg007:

PR #31136, opened 2022-05-26
refactor: Payment Reconciliation tool will use Payment Ledger

ruthra-kumar:
## Payment Reconciliation tool

Refactoring the reconciliation tool to use Payment ledger for pulling invoices and return invoices with outstanding balances.
Changes in this PR
- outstanding invoices and outstanding return invoices are fetched from payment ledger
- 'get_outstanding_invoices' function in utils.py has been refactored
- 'get_outstanding_reference_documents' function in payment_entry.py has been refactored
- helper class to query Payment Ledger
- unit tests for PR tool
- update outstanding amount using payment ledger

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -6,7 +6,7 @@ import json
 from functools import reduce
 
 import frappe
-from frappe import ValidationError, _, scrub, throw
+from frappe import ValidationError, _, qb, scrub, throw
 from frappe.utils import cint, comma_or, flt, getdate, nowdate
 
 import erpnext
@@ -1195,6 +1195,9 @@ def get_outstanding_reference_documents(args):
 	if args.get("party_type") == "Member":
 		return
 
+	ple = qb.DocType("Payment Ledger Entry")
+	common_filter = []
+
 	# confirm that Supplier is not blocked
 	if args.get("party_type") == "Supplier":
 		supplier_status = get_supplier_block_status(args["party"])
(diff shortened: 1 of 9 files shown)


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1231,16 +1240,19 @@ def get_outstanding_reference_documents(args):
 			condition += " and {0} between '{1}' and '{2}'".format(
 				fieldname, args.get(date_fields[0]), args.get(date_fields[1])
 			)
+			common_filter.append(ple.posting_date[args.get(date_fields[0]) : args.get(date_fields[1])])
```
deepeshgarg007:
```suggestion
			common_filter.append(ple.posting_date[args.get(date_fields[0]) == args.get(date_fields[1])])
```
```suggestion




			common_filter.append(ple.posting_date[args.get(date_fields[0]) : args.get(date_fields[1])])
```

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/payment_ledger_entry/payment_ledger_entry.py
```diff
@@ -18,5 +31,119 @@ def validate_account(self):
...
 
+	def validate_account_details(self):
+		"""Account must be ledger, active and not freezed"""
+
+		ret = frappe.db.sql(
+			"""select is_group, docstatus, company
+			from tabAccount where name=%s""",
+			self.account,
+			as_dict=1,
+		)[0]
```
ankush, line 42: - why raw sql? 
- `ret` is bad variable name 😅, `account` or `account_data` adds more info.

PR #31143, opened 2022-05-26
feat(india): Generate E-way Bill with IRN Settings

maharshivpatel:
Problem: If Transport details are available while generating IRN ewaybill will be generated without consent.
user also doesn't have an indicator if eway bill will be generated.
Cases:
1. The sales Invoice was duplicated so transport details were copied and eway  is generated.
2. The invoice value is less than 50000 so the user doesn't want to create eway bill but it is still generated.
3. currently, transport fields are not shown inside the transporter section as intended so the user can't change or add all the required fields to generate eway bill.

### Current vs Proposed Flow 
Eway Flow

I have added e-way bill settings and renamed the label of the advanced setting to GSP Setting as it was more appropriate

E Invoice Settings

**Note:** I have considered "" value as "Always Ask" and "Always" for eway_auto_generate and eway_when_generate respectively ~~as there is some bug and because of that the default value of the select field is not honored.~~

### Generate IRN Preview Changes
If ewaybill will be generated.

generate eway preview

If ewaybill will not be generated.

don't generate eway preview

### GST E-Invoice Print Format Changes

If IRN is not generated:

print format no irn

If Eway Bill Details are available:
print format eway details

`no-docs`


maharshivpatel:
@nextchamp-saqib hope you are doing well, are there any issues/improvements that I can solve to make this PR merge faster?

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -624,8 +624,21 @@ def make_einvoice(invoice):
...
+			or (
+				(invoice.mode_of_transport and invoice.mode_of_transport != "Road")
+				or (invoice.mode_of_transport == "Road" and invoice.vehicle_no)
+			)
+		)
+		and not invoice.is_return
+		and generate_ewb == "true"
+	):
+		if invoice.mode_of_transport == "Road" and not invoice.vehicle_no:
+			frappe.throw(_("Eway Bill : Vehicle No is mandatory if Mode of Transport is Road"))
```
```suggestion
			frappe.throw(_("Vehicle No is mandatory to generate Eway Bill if Mode of Transport is Road"))
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -624,8 +624,21 @@ def make_einvoice(invoice):
...
+		(
+			invoice.transporter
+			or (
+				(invoice.mode_of_transport and invoice.mode_of_transport != "Road")
+				or (invoice.mode_of_transport == "Road" and invoice.vehicle_no)
+			)
+		)
+		and not invoice.is_return
+		and generate_ewb == "true"
+	):
```
```suggestion
	if (
		(
			invoice.transporter
			or (invoice.mode_of_transport and invoice.mode_of_transport != "Road")
			or (invoice.mode_of_transport == "Road" and invoice.vehicle_no)
		)
		and not invoice.is_return
		and generate_ewb == "true"
	):
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -979,6 +992,30 @@ def generate_irn(self):
...
+				"""
+				# right now eway bill errors are shown, we can add more when available using below forloop.
+				for msg in info:
+					if msg.get("InfCd") == "EWBERR":
+						for desc in msg.get("Desc"):
+							frappe.msgprint(
+								_(desc.get("ErrorMessage")),
+								title="E-Way Bill not generated",
+								indicator="orange",
+							)
```
nextchamp-saqib: This could be extracted into a separate method?

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -979,6 +992,30 @@ def generate_irn(self):
 						Contact ERPNext support to resolve the issue."
 					)
 
+			if res.get("info"):
+				info = res.get("info")
```
```suggestion
				info = res.get("info", [])
```

erpnext/regional/india/e_invoice/utils.py
```diff
@@ -979,6 +992,30 @@ def generate_irn(self):
...
+    				                "ErrorMessage": "Vehicle number  should be passed in case of transportation mode is Road."
+    				            }
+    				        ]
+    				    }
+    				]
+				"""
+				# right now eway bill errors are shown, we can add more when available using below forloop.
+				for msg in info:
+					if msg.get("InfCd") == "EWBERR":
+						for desc in msg.get("Desc"):
```
```suggestion
						for desc in msg.get("Desc", []):
```

nextchamp-saqib (changes requested):
Minor changes

erpnext/regional/india/e_invoice/utils.py
maharshivpatel: I can extract for code readability but other than that there is really no other use for this as of now only case where IRN status is successful and there is an error inside info is when we have an e-way bill error.

nextchamp-saqib: Yes, I meant extracting it for code readability only
Even in the `einvoice.js`, the `Generate IRN` action has become huge. It could also be split into smaller functions.

maharshivpatel: I will commit separate method soon

maharshivpatel: I have made changes as discussed.

PR #31155, opened 2022-05-27
fix: Exchange rate reste to 1 on making mapped doc

### LINKED ISSUE #31063: conversion factor issue
Hi. I've discovered a bug that is caused by the changes that were made in this commit. When creating a Sales Order from a Quotation (with a currency other than base), the variables cause the if statement to trigger the else block. Previously, this was not an issue however, with the new changes the exchange rate is set 1 instead of using the previously specified exchange rate or even at least the one specified in the Currency Exchange table.

_Originally posted by @Archais in https://github.com/frappe/erpnext/issues/30990#issuecomment-1129834011_

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/31063


deepeshgarg007:
@Mergifyio backport version-13-hotfix version-13

PR #31166, opened 2022-05-27
fix(accounts): validate payment entry references with latest data.

### LINKED ISSUE #31138: Payment overallocation against invoice
### Information about bug

Invoices can get over allocated and outstanding amount goes to negative. Attaching video for reference.

https://user-images.githubusercontent.com/52111700/170423673-cf6bf64f-5fd6-4cc7-8b80-4eddc3ff1bff.mp4

### Module

accounts

### Version

ERPNext: v13.30.0 (HEAD)
Frappe Framework: v13.29.1 (HEAD)

### Installation method

FrappeCloud

_No response_

dj12djdjs:
Closes: https://github.com/frappe/erpnext/issues/31138

Problem:

The system allows multiple draft payment entries to be made for a single invoice, which is fine, but when the first payment entry is submitted, the outstanding amount of the invoice in the second payment entry isn't accurate anymore. But the system still allows the user to submit the entry, resulting in negative outstanding amounts.

Solution:

During validation of allocated amount in the payment entry, update each row in references with latest information before comparing amounts. If the user tries to submit another payment entry but the corresponding invoice has already been fully paid, the system would throw an error to let the user know. If the invoice has been already partly paid, the user would be told the same and would be asked to use the 'Get Outstanding Invoice' button to get the latest outstanding amount.

Unrelated: I discovered fetching documents for advance payment via payment entry form was also broken. Fixed in https://github.com/frappe/erpnext/pull/31166/commits/1ca20b97d066d3758fe85dd0bc6b9cc65d449bb3

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -174,10 +174,10 @@ class PaymentEntry(AccountsController):
 		for d in self.get("references").copy():
 			latest = latest_lookup.get((d.reference_doctype, d.reference_name))
 
-			# The reference has already been allocated.
-			if not latest:
+			# The reference has already been allocated, or partially allocated.
+			if not latest or d.outstanding_amount != latest.outstanding_amount:
 				frappe.throw(
-					_("{0} {1}, has already been allocated.").format(d.reference_doctype, d.reference_name)
+					_("{0} {1}, has already been allocated after the creation of Payment Entry {2}.").format(d.reference_doctype, d.reference_name, self.name)
 				)
 
 			d.outstanding_amount = latest.outstanding_amount
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -181,11 +181,7 @@ class PaymentEntry(AccountsController):
 				self.remove(d)
 				continue
 
-			d.due_date = latest.due_date
-			d.total_amount = latest.invoice_amount
 			d.outstanding_amount = latest.outstanding_amount
-			d.bill_no = latest.bill_no
-			d.payment_term = latest.payment_term
 
 			if (flt(d.allocated_amount)) > 0:
 				if flt(d.allocated_amount) > flt(d.outstanding_amount):
--- a/erpnext/accounts/doctype/payment_entry/payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/payment_entry.py
@@ -177,7 +177,9 @@ class PaymentEntry(AccountsController):
 			# The reference has already been allocated, or partially allocated.
 			if not latest or d.outstanding_amount != latest.outstanding_amount:
 				frappe.throw(
-					_("{0} {1}, has already been allocated after the creation of Payment Entry {2}.").format(d.reference_doctype, d.reference_name, self.name)
+					_("{0} {1}, has already been allocated after the creation of Payment Entry {2}.").format(
+						d.reference_doctype, d.reference_name, self.name
+					)
 				)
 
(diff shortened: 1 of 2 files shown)


dj12djdjs:
I'm not sure if the tests are failing from this PR. I'll have a further look today.

dj12djdjs:
I really don't know why those two tests keep failing. They don't fail on my local machine.

erpnext/accounts/doctype/payment_entry/test_payment_entry.py
```diff
@@ -758,6 +758,21 @@ def test_payment_entry_against_onhold_purchase_invoice(self):
...
+	def test_payment_entry_over_allocate_amount(self):
+		si = create_sales_invoice()
+
+		pe_draft = get_payment_entry("Sales Invoice", si.name)
+		pe_draft.insert()
+
+		pe = get_payment_entry("Sales Invoice", si.name)
+		pe.submit()
+
+		pe_draft.validate()
```
deepeshgarg007: In this case payment entry is getting submitted, since the reference row gets removed and it might be getting allocated as an advance which the users might not want, shouldn't this throw a validation error instead?

Also, you should not call validate method explicitly, it real scenarios the old document would already be validated, it should directly be called from submitting, you should remove the validate call its not needed

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/payment_entry/test_payment_entry.py
dj12djdjs: > In this case payment entry is getting submitted, since the reference row gets removed and it might be getting allocated as an advance which the users might not want, shouldn't this throw a validation error instead?

This is a good point. I will modify.

> Also, you should not call validate method explicitly, it real scenarios the old document would already be validated, it should directly be called from submitting, you should remove the validate call its not needed

Noted, I will  modify.

dj12djdjs:
@deepeshgarg007 I've added another check incase the a reference get's partially allocated between two payment entries.
I'm using the `outstanding_amount` to check this. 
https://github.com/dj12djdjs/erpnext/blob/1ef44c815743d6c97dd7efe270730a4782f170af/erpnext/accounts/doctype/payment_entry/payment_entry.py#L174-L175

I'm a confused why `get_payment_entry` makes the reference rows in this way. For example I create:
image

Output:
image

Shouldn't the outstanding be the same as the sales invoice outstanding (100) and the allocated be 50 ?

rtdany10:
@dj12djdjs @deepeshgarg007

rtdany10:
@dj12djdjs @deepeshgarg007 any update on this?

dj12djdjs:
@rtdany10 I think my solution to this isn't getting accepted. I'll think about it and try submitting a different solution.

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1449,7 +1487,7 @@ def get_orders_to_be_billed(
 	if voucher_type:
 		doc = frappe.get_doc({"doctype": voucher_type})
 		condition = ""
-		if doc and hasattr(doc, "cost_center"):
+		if doc and hasattr(doc, "cost_center") and doc.cost_center is not None:
```
deepeshgarg007:
```suggestion
		if doc and hasattr(doc, "cost_center") and doc.cost_center:
```

deepeshgarg007 (dismissed):

anandbaburajan:
Thanks @dj12djdjs!

ruchamahabal:
Payment Entry against Employee Advance (HR) stopped functioning after this PR:
Tests missing for Payment Entry against Employee Advance. Only Employee Advance -> Journal Entry flow gets tested in HR. Will add tests

image

image

Fixed in: https://github.com/frappe/erpnext/pull/35741

PR #31168, opened 2022-05-28
fix: Auto Insert Item Price If Missing when discount & blank UOM

### LINKED ISSUE #31167: Auto Insert Item Price If Missing doesn't work as intended.
### Information about bug

[ Item Price List ] 

**1. Auto Insert Item Price If Missing doesn't save the actual price_list_rate when the discount is used.**

**how to reproduce:**
 1. set Auto Insert Item Price If Missing as checked (✓) in Stock Settings.
 2. set Allow User to Edit Price List Rate in Transactions as checked (✓) in Selling Settings.
 3. create or choose an item that doesn't have a price_list_rate already set.
 4. add an item to the Sales Invoice with a 100₹ (price_list_rate) - 20% (discount) = 80₹ (Item Price).
 
The expected behavior is to add a price of 100₹ but instead, 80₹ is added to the Item Price List.

**2. UOM is set to blank which should be set as stock_uom of the item because the price is converted to stock_uom.**

**For example:**
Item have a conversion of [  1 box = 20 Nos ] and stock_uom = Nos and sales_uom = Box.
add an item to the Sales Invoice with a 1000₹ (price_list_rate) - 20% (discount) = 800₹ (Item Price).
item price will be 80₹ (due to the above-mentioned bug) and UOM will be set as blank as a result, when an item is added to the next invoice it will consider  80₹ for 1 box which is again incorrect behavior.

https://user-images.githubusercontent.com/39730881/170816976-b05ce47d-6cdd-4176-b42f-b366dad18eec.mp4

### Module

accounts, stock

### Version

ERPNext: v14.x.x-develop

### Installation method

manual install

_No response_

maharshivpatel:
Problem: Auto Insert Item Price If Missing will save the "price after discount" instead of "price before discount".

Solution: I have added price_list_rate = item_price + discount instead of price_list_rate = item_price and the reason for doing it this way instead of directly getting price_list_rate from args is that in cases where price_list_rate is not available this will still work and is less prone to bugs IMHO.

Problem: UOM is set to blank which should be set as stock_uom of the item because the price is converted to stock_uom.

Solution: I have added uom=stock_uom instead of blank and this will fix the item price being wrong due to non-conversion from stock_uom to sale_uom.
### Before 

https://user-images.githubusercontent.com/39730881/170821375-d895a883-c932-4605-9c61-60213e9ef3ad.mp4

### After

https://user-images.githubusercontent.com/39730881/170821391-3e3c2df5-d90a-4bd5-9600-300b9d9811f4.mp4

This closes #31167


erpnext/stock/get_item_details.py
```diff
@@ -849,6 +852,7 @@ def insert_item_price(args):
 						"item_code": args.item_code,
 						"currency": args.currency,
 						"price_list_rate": price_list_rate,
+						"uom": args.stock_uom,
```
ankush, line 855: This doesn't seem correct to me 🤔 Use could've picked a UOM that's different from stock UOM. E.g. A box = 100 Nos and now the price will be added for Nos instead of Box right? 


Similar problem exists for "update item price" functionality right above this line.

ankush:
@maharshivpatel can you add 1 unit test for this behaviour (where discount is present)

erpnext/stock/get_item_details.py
maharshivpatel: @ankush  I will add a test but If I understand you correctly you are saying that if the user adds 1 box that has 100 Nos.
Price will be added for Nos instead of Box but this is the intended behavior as you can see [here](https://github.com/frappe/erpnext/blob/3abd947982e45816109d0024a208436317a0ab94/erpnext/stock/get_item_details.py#L826) price is also converted as per the stock_uom by `(args.rate + args.discount_amount) / args.get("conversion_factor")` and it will be converted to sales UOM at the time of next SI Creation [ box or any other] as intended. 

Example:
Price of 1 Nos = 1 Box [1000₹] / Nos in Box [100] 
1 Nos = 10 ₹

Price_List_Rate is added as below
(
  item: any_product,
  UOM: "Nos",
  Price: 10 ₹
)
If I misunderstood your point can you please elaborate?

ankush: @maharshivpatel looks good. I didn't see the change in the context of how the price list rate is being derived right now. 


Perhaps this can be changed later someday to support inserting rates based on UOM only.

maharshivpatel:
I have ran all of the failed tests of  [Python Unit Tests (1)](https://github.com/frappe/erpnext/runs/6658838839?check_suite_focus=true#logs) on my local setup and they were successful but they failed on CI.

can someone please take a look.

what is the suggested approach for semgrep error [ Avoid using `args` as only argument unless absolutely necessary. ]

PR #27157, opened 2021-08-26
fix: don't allow BOM's item code at any level of child items

ankush:
Steps to reproduce:
1. Create BOM for Item A.
2. In child item, again add Item A.
3. Save. 

Observed:
BOM gets saved.

Expected:
validation error. 

Solution:
Never allow BOM's item at any level of exploded child items. 

closes https://github.com/frappe/erpnext/issues/26747

PR #31175, opened 2022-05-30
fix: allow recursive BOMs as long as they don't recursively explode

### LINKED ISSUE #31159: allow single level recursive BOMs when child item itself doesn't have BOM
This case isn't "recursive" in technical manner since child item isn't exploded further it stops there. 
example:

`Curd (1L) = Milk 1L + curd (10g)`

ankush:
This PR: https://github.com/frappe/erpnext/pull/27157 blocked the creation of BOMs that where the item is present in any level of BOM. However, some use cases are _simplified_ by using recursive BOM, hence adding this functionality back but only when item doesn't have BOM itself. This can be done by using "Do not explode": https://github.com/frappe/erpnext/pull/29450

New logic: 
- Parent BOM No. can never appear in BOM's tree
- Parent item code can appear as long as it's a child item without a BOM. 

Unrelated changes:
- Show "New version" button (duplicating is not obvious to users)
- Fix issue with BOM no getting erased and reset with default. 

closes #31159


ankush:
Some unrelated test failing, probably due to translations caching... will fix separately 😅

rohitwaghchaure (approved):

PR #31184, opened 2022-05-31
fix(accounts): Ignore Cancelled GL Entries

Mitchy25:
Profitability Analysis includes 'is_cancelled' GL Entries which means that the profit numbers are incorrect. This change will ensure that the profit figures ignore cancelled GL Entries.


deepeshgarg007:

PR #31194, opened 2022-05-31
fix: incorrect billed_qty in sales order analysis report when multiple delivery notes for item

ruthra-kumar:
## Issue
Incorrect billed_qty is calculated when Sales Order items have multiple Delivery notes against them.

Changes in this PR:
- separate query for fetching elapsed seconds 
- unit test for asserting report output on various fields

## Before
Screenshot 2022-05-31 at 5 26 45 PM

## After
Screenshot 2022-05-31 at 5 25 19 PM


deepeshgarg007:

PR #31210, opened 2022-06-01
fix: Pluralize year text instead of optional bracket

NagariaHussain:
## The Problem

The email sent for **Work anniversary reminders** uses optional brackets (`"(s)"`) to suffix number of years. It looks odd and even worse if multiple people have anniversaries on the same day (see below for an instance).

## The Fix

Conditionally add `"year"` or `"years"` based on number of years completed at the Company.

**Before**:

>  "Today John Doe completed 2 year(s) and Jenny completed 1 year(s) at our Company!

**After**:

> "Today John Doe completed 2 years and Jenny completed 1 year at our Company!

### DIFF
--- a/erpnext/hr/doctype/employee/employee_reminders.py
+++ b/erpnext/hr/doctype/employee/employee_reminders.py
@@ -257,8 +257,7 @@ def get_work_anniversary_reminder_text_and_message(anniversary_persons):
 def get_pluralized_years(years):
 	if years == 1:
 		return "1 year"
-	else:
-		return f"{years} years"
+	return f"{years} years"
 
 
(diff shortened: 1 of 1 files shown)


erpnext/hr/doctype/employee/employee_reminders.py
```diff
@@ -254,6 +254,13 @@ def get_work_anniversary_reminder_text_and_message(anniversary_persons):
 	return reminder_text, message
 
 
+def get_pluralized_years(years):
+	if years == 1:
+		return "1 year"
+	else:
+		return f"{years} years"
```
ruchamahabal:
```suggestion
	return f"{years} years"
```

ruchamahabal:
Unrelated test failure `test_sales_invoice_from_timesheet`

PR #31216, opened 2022-06-01
fix: Trial Balance failing to ignore Finance Book

### LINKED ISSUE #31214: Trial Balance not considering Finance Book
### Information about bug

_Opening(Dr)_ and _Opening(Cr)_ columns aggregates all entries with or without _Finance Book_ when this filter is left blank. Otherwise when, this filter is set, the resulting values are as expected.

The discrepancy can be observed when this report is compared against **Report: Balance Sheet**.

### Module

accounts

### Version

ERPNext: v13.32.0
Frappe Framework: v13.31.0

### Installation method

docker

_No response_

sunhoww:
Fixes #31214.

This issue was occurring because the removed code block only considers cases if the `finance_book` filter was set. When `finance_book` is not set, the resulting query would execute over all **GL Entry** irrespective of the value of the `GL Entry.finance_book`. However, the expected query would have been to return all `GL Entry.finance_book` which are not `company_fb` or empty.

Therefore, this block has been replaced with one that's unabashedly pulled from `financial_statements.py`.
https://github.com/frappe/erpnext/blob/3974fbbb6e0f4f41b292b870b80d6eab300f5e32/erpnext/accounts/report/financial_statements.py#L494-L499

For cases where the `finance_book` filter is not set, the condition could evaluate to 
```sql
... AND (finance_book in (NULL, NULL, '') OR finance_book IS NULL) ...
```
which should work although probably not ideal.

### DIFF
--- a/erpnext/accounts/report/trial_balance/trial_balance.py
+++ b/erpnext/accounts/report/trial_balance/trial_balance.py
@@ -160,14 +160,10 @@ def get_rootwise_opening_balances(filters, report_type):
 	if filters.project:
 		additional_conditions += " and project = %(project)s"
 
-	if filters.finance_book:
-		fb_conditions = " AND finance_book = %(finance_book)s"
-		if filters.include_default_book_entries:
-			fb_conditions = (
-				" AND (finance_book in (%(finance_book)s, %(company_fb)s, '') OR finance_book IS NULL)"
-			)
-
-		additional_conditions += fb_conditions
+	if filters.get("include_default_book_entries"):
+		additional_conditions += "AND (finance_book in (%(finance_book)s, %(company_fb)s, '') OR finance_book IS NULL)"
+	else:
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

dawoodjee:
I'm on v14.20 and still having this issue

tb vs gl

dawoodjee:
I'd to remove this part entirely, using the **Include Default Book Entries** condition

**frappe/apps/erpnext/erpnext/accounts/report/trial_balance/trial_balance.py**
```
if filters.get("include_default_book_entries"):
    pass
else:
```

**/home/frappe/frappe/apps/erpnext/erpnext/accounts/report/financial_statements.py**

```
if filters.get("include_default_book_entries"):
    pass
```

I think best to not add the entire condition if no finance books are specified

PR #31219, opened 2022-06-02
fix: Parent dimension filters in orders

deepeshgarg007:
Extension to: https://github.com/frappe/erpnext/pull/30806

The sales Order cost center field was showing group cost centers and later while converting these orders to invoices there was a validation error.

Moved accounting dimension filters to Sales and Purchase common to apply filters on all Sales and Purchase docs


deepeshgarg007:

PR #31223, opened 2022-06-02
fix: display currencies in validation message.

dj12djdjs:
It will be helpful to display the currency of the party and the document to help diagnose the issue.

Especially during failing test. It will take much greater time to find what the currency of these test documents are. Example: 
https://github.com/frappe/erpnext/runs/6712083194?check_suite_focus=true


deepeshgarg007:

Issue #31079, opened 2022-05-20
Print/PDF for financial statement reports displays either wrong date range or wrong fiscal year


dgerhardt:
### Information about bug

Reports for financial statements including "Consolidated Financial Statement", "Profit and Loss Statement", "Balance Sheet" and "Cash Flow" can be filtered either by fiscal year or by date range (`filter_based_on`). When one of them is changed, the other one isn't updated automatically and is just ignored for the purpose of filtering. But on print/PDF report, both filters are displayed. The date range and fiscal year will display inconsistent values.

Example:
The fiscal year 2022 is currently the default. I now open the Consolidated Financial Statement which uses 2022 and 2022-01-01 to 2022-12-31. I now change the fiscal year to 2021 and open the print view. It displays:
```
Filter Based On: Fiscal Year
Start Date: 2022-01-01
End Date: 2022-12-31
Start Year: 2021
End Year: 2021
```

I would expect that either the inactive filter is not displayed or that the values are updated so fiscal year and date range are consistent.

### Module

accounts

### Version

Frappe version 13.29.1
ERPNext version 13.30.0

PR #31225, opened 2022-06-03
fix: Print/PDF for financial statement reports displays either wrong date range or wrong fiscal year

vishdha:
fixed issue: https://github.com/frappe/erpnext/issues/31079

While changing fiscal year it's not updating period_start_date and period_end_date. so added this change to show valid info in print format.

image

### DIFF
--- a/erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.js
+++ b/erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.js
@@ -50,7 +50,15 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 				"fieldtype": "Link",
 				"options": "Fiscal Year",
 				"default": frappe.defaults.get_user_default("fiscal_year"),
-				"reqd": 1
+				"reqd": 1,
+				on_change: () => {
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.js
```diff
@@ -50,15 +50,31 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 				"fieldtype": "Link",
 				"options": "Fiscal Year",
 				"default": frappe.defaults.get_user_default("fiscal_year"),
-				"reqd": 1
+				"reqd": 1,
+				on_change: () => {
+					frappe.model.with_doc("Fiscal Year", frappe.query_report.get_filter_value('from_fiscal_year'), function(r) {
+						let start_fy = frappe.model.get_doc("Fiscal Year", frappe.query_report.get_filter_value('from_fiscal_year'));
```
deepeshgarg007: Use get_value instead of get_doc?

deepeshgarg007 (changes requested):

erpnext/accounts/report/consolidated_financial_statement/consolidated_financial_statement.js
vishdha: updated

PR #31229, opened 2022-06-04
chore: Release for version 14 beta

deepeshgarg007:


maharshivpatel:
> Why is canceling ewaybill disabled?

NIC have depreciated e-way bill cancel api from e-invoice api

user will have to cancel e-way bill using portal and then update the status in erpnext

9D9E6AD7-417A-4ACE-BCEC-BB81F4CD5B06

PR #31230, opened 2022-06-05
fix: Remove redundant query

deepeshgarg007:
Getting deducted taxes separately was an unnecessary query. Handled the same in the tax query and reduced one extra query


deepeshgarg007:

PR #31231, opened 2022-06-05
fix(India): Supplies from composite dealer not showing up

deepeshgarg007:
image

Supplies from composite dealer was not showing up in GSTR-3B report list


deepeshgarg007:

PR #31232, opened 2022-06-05
fix: update fr translation

FHenry:
fix: update fr translation


erpnext/translations/fr.csv
```diff
@@ -2811,7 +2811,7 @@ Stock Value,Valeur du Stock,
 Stock balance in Batch {0} will become negative {1} for Item {2} at Warehouse {3},Solde du stock dans le Lot {0} deviendra négatif {1} pour l'Article {2} à l'Entrepôt {3},
 Stock cannot be updated against Delivery Note {0},Stock ne peut pas être mis à jour pour le Bon de Livraison {0},
 Stock cannot be updated against Purchase Receipt {0},Stock ne peut pas être mis à jour pour le Reçu d'Achat {0},
-Stock cannot exist for Item {0} since has variants,Stock ne peut pas exister pour l'Article {0} puisqu'il a des variantes,
+Stock cannot exist for Item {0} since has variants,Stock ne peut pas exister pour l&#39;Article {0} puisqu'il a des variantes,
```
marination: Is `&#39;` required? I tried making this change locally and:
Screenshot 2022-06-09 at 7 33 42 PM

l'Article works just fine right ?

FHenry: Some translation can work with ' but it will always works everywhere with `&#39; `
Not related to french language but frappe js framework that newly (develop branch) deal strangely with quote in translation

For example (actual translation)
`Stock Ledger,Livre d'Inventaire,`
cause a js errors (as you can see there is missing buton because js crash)

`Uncaught (in promise) Error: Syntax error, unrecognized expression: a[data-label='Livre%20d'Inventaire']`

image

That's why I suggest 
`Stock Ledger,Livre d&#39;Inventaire,`
image



I hesitate to review all french translation to change simple quote by `&#39;` because it always works, but I prefers to go one by one

Where your right is that I let a simple quote into the same translation (line 2814), I've just update it

FHenry: You'r right it seems that my quote problem only affect doctype name (that have to be displayed into View or Action card button), so I let the normal simple quote on this reviewed line sentence, but the problem for field name or doctype name is still there, I've just have to check where the translations are used

FHenry: @marination I've send a fix for the simple quote button name problem into frappe, So I re-change `&#39;` to simple quote. Better that way

marination: @FHenry Awesome !! 😄

erpnext/translations/fr.csv
```diff
@@ -1352,7 +1352,7 @@ Item Description,Description de l'Article,
 Item Group,Groupe d'Article,
 Item Group Tree,Arborescence de Groupe d'Article,
 Item Group not mentioned in item master for item {0},Le Groupe d'Articles n'est pas mentionné dans la fiche de l'article pour l'article {0},
-Item Name,Nom de l&#39;article,
+Item Name,Nom de l'article,
 Item Price added for {0} in Price List {1},Prix de l'Article ajouté pour {0} dans la Liste de Prix {1},
 "Item Price appears multiple times based on Price List, Supplier/Customer, Currency, Item, UOM, Qty and Dates.","Le prix de l&#39;article apparaît plusieurs fois en fonction de la liste de prix, du fournisseur / client, de la devise, de l&#39;article, de l&#39;unité de mesure, de la quantité et des dates.",
```
marination:
```suggestion
"Item Price appears multiple times based on Price List, Supplier/Customer, Currency, Item, UOM, Qty and Dates.","Le prix de l'article apparaît plusieurs fois en fonction de la liste de prix, du fournisseur / client, de la devise, de l'article, de l'unité de mesure, de la quantité et des dates.",
```

erpnext/translations/fr.csv
```diff
@@ -9871,8 +9871,42 @@ Show Barcode Field in Stock Transactions,Afficher le champ Code Barre dans les t
...
+{} Pending,{} En attente.s
+{} To Bill,{} à facturer
+{}  To Receive,{} A recevoir
+{} Active,{} Actif.ve(s)
+{} Open,{} Ouvert.e(s)
+Incorrect Data Report,Rapport de données incohérentes
+Incorrect Serial No Valuation,Valorisation inccorecte par Num. Série / Lots
+Incorrect Balance Qty After Transaction,Equilibre des quantités aprés une transaction
+Interview Type,Type d&#39;entretien
+Interview Round,Cycle d&#39;entretien
```
marination:
```suggestion
Interview Type,Type d'entretien
Interview Round,Cycle d'entretien
```

erpnext/translations/fr.csv
```diff
@@ -1352,9 +1352,9 @@ Item Description,Description de l'Article,
...
 Item Group not mentioned in item master for item {0},Le Groupe d'Articles n'est pas mentionné dans la fiche de l'article pour l'article {0},
-Item Name,Nom de l&#39;article,
+Item Name,Nom de l'article,
 Item Price added for {0} in Price List {1},Prix de l'Article ajouté pour {0} dans la Liste de Prix {1},
-"Item Price appears multiple times based on Price List, Supplier/Customer, Currency, Item, UOM, Qty and Dates.","Le prix de l&#39;article apparaît plusieurs fois en fonction de la liste de prix, du fournisseur / client, de la devise, de l&#39;article, de l&#39;unité de mesure, de la quantité et des dates.",
+"Item Price appears multiple times based on Price List, Supplier/Customer, Currency, Item, UOM, Qty and Dates.","Le prix de l'article apparaît plusieurs fois en fonction de la liste de prix, du fournisseur / client, de la devise, de l'article, de l'unité de mesure, de la quantité et des dates.",
 Item Price updated for {0} in Price List {1},Prix de l'Article mis à jour pour {0} dans la Liste des Prix {1},
 Item Row {0}: {1} {2} does not exist in above '{1}' table,Ligne d&#39;objet {0}: {1} {2} n&#39;existe pas dans la table &#39;{1}&#39; ci-dessus,
```
marination:
```suggestion
Item Row {0}: {1} {2} does not exist in above '{1}' table,Ligne d'objet {0}: {1} {2} n'existe pas dans la table '{1}' ci-dessus,
```

marination (approved):

PR #31233, opened 2022-06-05
fix: Reverse provisional entries on Purchase Invoice cancel

deepeshgarg007:
For context read https://docs.erpnext.com/docs/v13/user/manual/en/accounts/purchase-invoice#45-provisional-accounting-for-expenses-only-for-non-stock-items

On cancellation of Purchase Invoice against a service Purchase Receipt, the provisional entries posted against the Purchase Receipt were not getting canceled and on amending the Purchase Invoice against reverse provisional entries were passed due to which the balances went off

To avoid this simply cancel reversal entries on Purchase Invoice cancel


deepeshgarg007:

PR #31253, opened 2022-06-06
fix: leave balance for earned leaves in backdated Leave Application dashboard

ruchamahabal:
## Problem

1. Create Leave Policy Assignment for earned leaves with 8 leaves allocated at first.
	
	allocation

2. Earned Leaves are allocated in April and May end.
3. Try creating a backdated leave application for April. It shows balance as 12 and 4 leaves expired. Although 12 leaves weren't available on April 1st, only 8 were available and there were no expired leaves

image

This problem only occurs in the case of Earned Leaves because they are added based on policy frequency, unlike other leaves which are allocated all at once on allocation creation.

The leave balance and leave balance summary reports, and balance in the leave application all work fine. The issue is only with the leave application dashboard.

## Fix

Don't consider the entire allocation, only consider the ledger entries that fall within that period.

image


ruchamahabal:
Unrelated test failure

PR #31258, opened 2022-06-07
fix: Close unsecured terms loans

deepeshgarg007:
Currently, there was no way to close an unsecured term loan, they would reach up to Loan Closure requested state and there was no further operation to be carried out to close the loan

Screenshot 2022-06-07 at 1 19 10 PM

Add an option to close such loans


deepeshgarg007:

PR #31264, opened 2022-06-07
fix: internal transfer GLE validation

ankush:
While doing internal transfers if both accounts are the same a GLE is made with equal credit/debit, this fails validation while submitting the doc. 

Soln: toggle credit to negative debit (like done in other transactions) 

ps: check added test case for example


deepeshgarg007 (approved):

deepeshgarg007:

PR #31271, opened 2022-06-07
fix(stock): don't reserve qty on sales return.

dj12djdjs:
bringing https://github.com/frappe/erpnext/pull/31017 back to life


dj12djdjs:
@ankush Just wanted to checkup, have you thought of any way to implement this without disturbing existing workflows ?

ankush:
@dj12djdjs we can add an option in selling settings. 

The default behavior should be consistent with current behavior.

dj12djdjs:
@ankush Ready for review

dj12djdjs:
@ankush Sorry about that, I just realized there was more conflicts 😅

dj12djdjs:
@ankush Is there any issues with this ?

s-aga-r (changes requested):
@dj12djdjs please resolve conflicts.

barredterra (changes requested):
Changes to `erpnext/selling/doctype/sales_order/sales_order.json` seem to be unrelated; please revert these.

erpnext/stock/stock_balance.py
```diff
@@ -94,51 +98,85 @@ def get_balance_qty_from_sle(item_code, warehouse):
 
 
 def get_reserved_qty(item_code, warehouse):
```
rohitwaghchaure, line 96: This function looks very complex and difficult to read, could you split the code into multiple methods?

dj12djdjs: @rohitwaghchaure Honestly, I don't even fully understand the query. I just reverted it back to the previous query and used f string to inject the condition.

rohitwaghchaure (approved):

Issue #23068, opened 2020-08-18
Write off is lost upon payment entry. Wrong outstanding shown.


rtdany10:
## Description of the issue
Write off in sales invoice is lost upon payment entry and the outstanding amount shown is wrong.

## Steps to reproduce the issue
![errorgif](https://user-images.githubusercontent.com/52111700/90506483-fd925080-e171-11ea-93fd-a3a944d0957c.gif)

dj12djdjs:
Ignore previously linked PRs. Unrelated.

PR #31272, opened 2022-06-07
fix: payment entry unallocated_amount

### LINKED ISSUE #23936: Unallocated amount is incorrectly calculated in Payment Entry when paying an Invoice in Non Base currency

### LINKED ISSUE #29595: Unallocated amount is inaccurate on multi-currency Payment Entry

dj12djdjs:
alternative solution for https://github.com/frappe/erpnext/pull/30927

3rd time the charm 😉

In my previous PR I really couldn't think about how to manage the difference between write-offs and gain/loss
Finally, I found we can solve this by zeroing unallocated_amount and difference_amount if they equal each other.

This is much simpler than maintaining two different tables.

closes: https://github.com/frappe/erpnext/issues/29595
closes: https://github.com/frappe/erpnext/issues/23936


deepeshgarg007:
@dj12djdjs I couldn't replicate https://github.com/frappe/erpnext/issues/23068 on develop branch and also could not understand how this PR fixes this issue.

I am still not sure about "I found we can solve this by zeroing unallocated_amount and difference_amount if they equal each other" this logic

If you specify the values that you tried locally while testing and elaborate on this more I'll be able to review this better.

dj12djdjs:
Hello @deepeshgarg007

I've attached a database of the figures I've used if you'd like to test it out on both branches to compare the results locally.
[db.sql.gz](https://github.com/frappe/erpnext/files/9167911/db.sql.gz)

Regarding the issue you referred to, you're correct. That issue is not related to this PR. I've removed it.


**Before PR**

develop

**After PR**

fix-pe


You can see the results for yourself here:
user: demo@demo.com
pass: thispasswordsucks123

Before PR: https://develop-pe.frappe.cloud
With PR: https://fix-pe.frappe.cloud

fproldan:
👍🏻

PR #31278, opened 2022-06-08
[DEV] accounts: added filter with columns in account receivable report.

Solufyin:
**`BASE FEATURE`**

- Account Receivable has a filter like finance book, accounts, cost center, customer, aging base on date,  Payment terms, custom group, sales person, territory, and other filters. currency is appearing in a column in a report but it shows only the default company currency. Currency is shown and but party account-wise currency does not show in a report. 

**`DEVELOPED FEATURE`**

- Added account currency filter and columns like account wise currency, currency-wise invoice amount in Accounts Receivable report.

- Here is attached video for developed feature in Accounts Receivable report.

https://user-images.githubusercontent.com/34390782/172552724-5a071098-bb75-4600-bcb6-6b3f07fbed19.mp4

Thanks.

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.js
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.js
@@ -148,6 +148,12 @@ frappe.query_reports["Accounts Receivable"] = {
 			"options": "Territory"
 		},
 		{
+			"fieldname": "account_currency",
+			"label": __("Account Currency"),
+			"fieldtype": "Link",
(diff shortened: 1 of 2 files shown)


deepeshgarg007:
@Solufyin I don't think viewing the report in multiple currencies makes any sense. If you filter by a specific receivable/payable account or customer/supplier you get the report in the account or party-specific currency. Not sure what's the use case here. Please update the use case and reopen the PR. Right now I don't think this feature is needed.

PR #31280, opened 2022-06-08
fix: locale Currency and Float setting in update_employee

RJPvT:
In fieldtypes locale settings (example NL) . and , changes whereby the field is inproperly filled


marination:
@RJPvT please send a PR to `develop` branch as well

marination:
`update_employee` is only actively used in v12. No references in v13 or develop except for declaration.

marination (approved):

PR #31284, opened 2022-06-08
fix(ux): Add tabs in Item

rmehta:
image


chechani:
Thanks to include the Tab item doctype that deserves TabBreak field most ..

We found 3 issues in Item Doctype after updating
1. There is a link field for Doctype Billing Type, which is not there in the System.
2. There are 2 details tab
3. No data can be seen in any of the details Tab & Dashboard Tab refer screenshot
Screenshot 2022-06-08 at 20 31 00
Screenshot 2022-06-08 at 20 31 56
Screenshot 2022-06-08 at 20 32 09
Screenshot 2022-06-08 at 20 32 14

rmehta:
@chechani did you add new tabs via customize form?

chechani:
No, It's a fresh install for testing without any customisation. In Item doctype,  the Hubsync field (why does it exist?) is already there through the customisation hence Details tab is visible twice probably. Details Tab is visible sometimes and sometimes not that is an issue if we create a new doctype too and we created an issue already about that https://github.com/frappe/frappe/issues/16645  . 

Our observation is that if you refresh the page Details page will be visible definitely and hence we have a workaround through a client script to refresh the page for a doctype that has a tabbreak field.

ankush:
Faced this locally as well, seems like a problem with tabs, not just this PR. (no errors or anyhing, some tabs just dont load)

Screenshot 2022-06-09 at 12 56 36 PM

rmehta:
https://github.com/frappe/frappe/pull/17108/files#diff-1dc5b7a41767010f76111c10a9c3680d518f6bc395aa0a35c8b530c4bb8db3dfL6 should fix it

chechani:
@rmehta , it worked well! Thanks a lot and salute to you!

yadavyk:
Use case : While creating template item, the "maintain stock" checkbox has to be unchecked so that item variants can be created. However upon checking "maintain stock" , the inventory tab hides. GIF for reference : 

inventory_tab

@rmehta , is this intentional ?

rmehta:
@yadavyk will push a fix

PR #31285, opened 2022-06-08
fix: make Quickbooks Migrator picklable

sagarvora:
Depends on https://github.com/frappe/frappe/pull/17115

Uses features available only in Python 3.


ankush:
@sagarvora we can leave this as is? Not sure why/where we'd ever need to cache this doctype 😄 Since get_doc doesn't implicitly push everything to cache IMO this isn't a problem anymore. 

Reopen if reqd.

PR #31286, opened 2022-06-08
fix: Added total of future amount, Remaining Balance in Payable and Receivable reports

Solufyin:
**`Version`**
ERPNext: v14.0.0-develop

**`Accounts Receivable`**

**Before**:
- In Accounts Receivable, when group by customer and show future payments is checked then future payment amount and remaining balance subtotal and total lines does not show.

https://user-images.githubusercontent.com/34390782/172615148-df2a925b-5966-437d-bcdc-d1bc3d69716d.mp4

**After Development**:
- Added code for future payment amount and remaining balance of subtotal and total.

https://user-images.githubusercontent.com/34390782/172615947-f2c30984-956a-41bf-bcfa-a0dffd476dea.mp4

**`Accounts Payable`**

**Before**:

- In Accounts Payable, there is no option to show future payments.
image

**After Development**:
- Added filter like Show Future Payments, and update query in get_future_payments_from_journal_entry() method.

https://user-images.githubusercontent.com/34390782/172619166-6c773054-2ba1-40ab-b53d-12452b4302e4.mp4

Thanks.


erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -563,7 +565,7 @@ def get_future_payments_from_journal_entry(self):
 				jea.party,
 				jea.party_type,
 				je.posting_date as future_date,
-				sum({0}) as future_amount,
+				sum('{0}') as future_amount,
```
deepeshgarg007, line 549: Why this change? I think the query is working as it is right now

Solufyin: Hi @deepeshgarg007,

Earlier in the account payable report the future payment filter with column was not available.
So, in order to fulfill the change for both account receivable report and account payable report I did this change and its working.

Thanks.

deepeshgarg007: @Solufyin I understood the reason for the PR, my concern is regarding the quotes in the query `sum('{0}')`, they were not there before and the report was working fine, why add them now?

Solufyin: Hi @deepeshgarg007,

Because of the use of ```sum({0})``` then only the account receivable report worked but the account payable report did not work with the added new feature. So added ```sum('{0}')``` then after checked so both reports are worked properly.

Thanks.

Solufyin: Hi @deepeshgarg007 , @nextchamp-saqib,

Is there any update?

PR #31291, opened 2022-06-08
fix: Depreciate asset on the last day of the month if its depr start date falls on the last day of its month

GangaManoj:
## Issue

If an Asset that depreciates monthly starts depreciating on, say, Feb 28, then it would depreciate next on March 28, then April 28 and so on. However, customers usually intend to have the asset depreciate on the last day of each month in this case- i.e. Feb 28 should be followed by March 31 and April 30 instead.

## Fix

This PR checks if the `depreciation_start_date` falls on the last day of its month, and if so, modify the schedule dates so that they'd each fall on the last day of their respective months.

Screenshot 2022-06-09 at 2 38 31 AM

### DIFF
--- a/erpnext/assets/doctype/asset/asset.py
+++ b/erpnext/assets/doctype/asset/asset.py
@@ -257,6 +257,7 @@ class Asset(AccountsController):
 				number_of_pending_depreciations += 1
 
 			skip_row = False
+			should_get_last_day = self.is_last_day_of_the_month(finance_book.depreciation_start_date)
 
 			for n in range(start[finance_book.idx - 1], number_of_pending_depreciations):
 				# If depreciation is already completed (for double declining balance)
@@ -269,6 +270,9 @@ class Asset(AccountsController):
 					schedule_date = add_months(
 						finance_book.depreciation_start_date, n * cint(finance_book.frequency_of_depreciation)
 					)
+
+					if should_get_last_day:
+						schedule_date = get_last_day(schedule_date)
 
 					# schedule date will be a year later from start date
 					# so monthly schedule date is calculated by removing 11 months from it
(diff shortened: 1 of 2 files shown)


erpnext/assets/doctype/asset/asset.py
```diff
@@ -405,6 +409,14 @@ def make_depreciation_schedule(self, date_of_sale):
...
 						)
 
+	def is_last_day_of_the_month(self, depreciation_start_date):
+		last_day_of_the_month = get_last_day(depreciation_start_date)
+
+		if getdate(last_day_of_the_month) == getdate(depreciation_start_date):
+			return True
+		else:
+			return False
+
```
marination:
```suggestion
		return getdate(last_day_of_the_month) == getdate(depreciation_start_date)
```

erpnext/assets/doctype/asset/test_asset.py
```diff
@@ -1261,6 +1261,32 @@ def test_asset_cost_center(self):
...
+			"2020-02-28",
+			"2020-03-31",
+			"2020-04-30",
+			"2020-05-31",
+			"2020-06-30",
+			"2020-07-15",
+		]
+
+		for i, schedule in enumerate(asset.schedules):
+			self.assertEqual(expected_dates[i], schedule.schedule_date)
```
marination: Failing CI
```suggestion
			self.assertEqual(getdate(expected_dates[i]), schedule.schedule_date)
```

GangaManoj:
Fix included in https://github.com/frappe/erpnext/pull/31429, thereby closing this.

GangaManoj:
Fix included in https://github.com/frappe/erpnext/pull/31429, therefore closing this.

PR #31292, opened 2022-06-09
fix: update disburesement account from old loan

zeta17:

> Screenshots/GIFs


ankush:
@zeta17 please update the description, also fixing a patch usually requires rerunning it fully or rerunning part of it.

deepeshgarg007:
@ankush  This patch doesn't require rerunning, This is only for users migrating from v12 to v13

deepeshgarg007:
@zeta17 There is already a patch taking care of this when the disbursement account was introduced, don't think this change is needed

https://github.com/frappe/erpnext/blob/develop/erpnext/patches/v13_0/update_disbursement_account.py

Closing this for now, please reopen if this is otherwise

PR #31294, opened 2022-06-09
fix(India): Incorrect taxable in GSTR-3B report

deepeshgarg007:
Taxable value for invoice having the same item repeated multiple times was not getting reported correctly


deepeshgarg007:

PR #31302, opened 2022-06-09
fix: Reset represents company on disabling internal customer and supplier

deepeshgarg007:
The `Represents Company` field is dependent on the `is_internal_customer` and `is_internal_suppliers` fields in Customer and Supplier doctypes respectively and it is a unique field that means there can be only one customer/supplier representing a particular internal company at a time

Many times users uncheck `is_internal_(customer/supplier)` from a record but don't clear the represented company field which leads to issues in creating new ones.

 Added a check to auto clear this field


deepeshgarg007:

ankush (approved):

PR #31322, opened 2022-06-10
fix: Company address filter in quotation

deepeshgarg007:
1. Address filter in the quotation in the company address field (should show only addresses linked to the company)
2. For India, sales taxes and charges template fetching based on company
3. User was able to make N no. of sales order against a single quotation using the Create Sales Order button in the quotation, added a Partially Ordered status to control the button viewing logic


deepeshgarg007:

PR #31325, opened 2022-06-10
feat: item_reorder - Add Request from

ahmadpak:
While configuring item reorder levels. If the type of request is transfer. Its good to have the option to set the from warehouse. The new field request from will only display if the request type is transfer

In this PR. If the user sets the request from. The system will set the from warehouse while auto creating material request of type transfer

image

image

docs: `no-docs`


marination:
@ahmadpak 
1) Can you please allow edits by maintainers here? Makes it easier for us to update the branch or push fixes to speed up the PR merger
   Screenshot 2022-06-13 at 10 34 10 AM

2) Please add a basic test for this: create item > add reorder info to item and save > create stock reco such that item's projected qty goes below re-order level > call `reorder_item` > check the MRs created (filter out MR with new item) and see if warehouse is set and qty is correct
You can refer to `_test_auto_material_request` in `test_stock_entry.py`

3) Update https://docs.erpnext.com/docs/v13/user/manual/en/stock/item#34-automatic-reordering section and add the link to the page in the PR description

ahmadpak:
@marination I am little lost here. I do not see this checkbox.

For point 2 and 3, I will try doing that too.

image

ankush:
> @marination I am little lost here. I do not see this checkbox.



Organization's forks aren't allowed to do this AFAIK. There's no way for you to enable maintainer edits, that only works with individual forks.

marination:
@ankush @ankush my bad, you can ignore the checkbox bit

PR #31327, opened 2022-06-11
fix: transaction date gets unset in material request

sagarvora:
## Issue
If `transaction_date` is set in args (as in `transaction.js` before this PR):
`out` gets the same `transaction_date` as `args` in `get_basic_details` and the date remains unchanged in `doc`

If `transaction_date` is not set in args (as in `material_request.js`):
`out` doesn't get the `transaction_date` in `get_basic_details`, and the date consequently gets unset

## Solution
- don't pass `transaction_date` at all if we're overwriting `args` anyway
- move `get_basic_details` to below where args are overwritten
- set `transaction_date or posting_date` as `transaction_date` in `args` (as per current behavior in `transaction.js`). The reason why this is harmless is because docs have either `transaction_date` or `posting_date`, but not both.
- remove all references of `posting_date` to simplify code

A side effect of this is exchange rate being correct on Invoices. Since `get_item_details.validate_conversion_rate` only considers `args.transaction_rate`, this rate was incorrect for documents with `posting_date` instead of `transaction_date`.

### DIFF
--- a/erpnext/public/js/controllers/transaction.js
+++ b/erpnext/public/js/controllers/transaction.js
@@ -453,7 +453,6 @@ erpnext.TransactionController = class TransactionController extends erpnext.taxe
 							is_pos: cint(me.frm.doc.is_pos),
 							is_return: cint(me.frm.doc.is_return),
 							is_subcontracted: me.frm.doc.is_subcontracted,
-							transaction_date: me.frm.doc.transaction_date || me.frm.doc.posting_date,
 							ignore_pricing_rule: me.frm.doc.ignore_pricing_rule,
 							doctype: me.frm.doc.doctype,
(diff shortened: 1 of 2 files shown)


marination:
tested conversion rate and item price fetch locally on Purchase Order (doc with `transaction_date`) and Purchase Receipt (doc with `posting_date`). Works as expected.
Logic seems sound

@sagarvora Can you update the branch once ? Once the newer workflow runs pass we're good to go

marination (dismissed):

sagarvora:
Linter failing due to https://github.com/returntocorp/semgrep/issues/5562
Thanks @ankush for helping me find above issue.

Merging this as of now.

sagarvora:

PR #31345, opened 2022-06-14
refactor: use db independent offset syntax

c-p-b:
This syntax is supported by both MariaDB and Postgres


erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -107,7 +107,7 @@ def get_items(start, page_length, price_list, item_group, pos_profile, search_te
 		ORDER BY
 			item.name asc
 		LIMIT
-			{start}, {page_length}""".format(
+			{page_len} offset {start}""".format(
```
ankush:
```suggestion
			{page_length} offset {start}""".format(
```

Typo

ankush (approved):

PR #31350, opened 2022-06-14
fix: Conversion rate validation for multi-currency invoices

deepeshgarg007:
There are some cases when the exchange rate API doesn't work and the exchange rate in multi-currency invoices is not fetched. In such cases, a multi-currency invoice is submitted with an exchange rate of 1.

There was already validation for this on Purchase Invoice, commodified it and added it on Sales Invoice as well.


erpnext/controllers/accounts_controller.py
```diff
@@ -1838,6 +1838,17 @@ def create_advance_and_reconcile(self, party_link):
...
 		jv.submit()
 
+	def check_conversion_rate(self):
+		default_currency = erpnext.get_company_currency(self.company)
+		if not default_currency:
+			throw(_("Please enter default currency in Company Master"))
+		if (
+			(self.currency == default_currency and flt(self.conversion_rate) != 1.00)
+			or not self.conversion_rate
+			or (self.currency != default_currency and flt(self.conversion_rate) == 1.00)
```
ankush, line 1858: Conversion rate between few currencies can actually be 1.0 (e.g. pegged currencies)

Lets avoid validating that (or maybe just give warning instead)

deepeshgarg007:

PR #31354, opened 2022-06-14
fix(minor): move variants to separate tab

rmehta:

> Screenshots/GIFs


erpnext/manufacturing/workspace/manufacturing/manufacturing.json
```diff
@@ -415,52 +415,38 @@
...
+   "link_to": "BOM",
    "type": "DocType"
   },
   {
-   "color": "Green",
-   "format": "{} Active",
-   "label": "BOM",
-   "link_to": "BOM",
-   "restrict_to_domain": "Manufacturing",
-   "stats_filter": "{\n    \"is_active\": 1\n}",
```
ankush, line 432: Why these filters are removed? 👀 They are used for showing that count on workspace. 
Screenshot 2022-06-15 at 12 21 40 PM


Rest LGTM, tested locally

ankush (approved):

PR #31356, opened 2022-06-14
refactor: use CURRENT_DATE instead of CURDATE()

c-p-b:
This syntax is supported by both postgres and mariadb


ankush (approved):

ankush:
unrelated patch failure 👀

PR #31358, opened 2022-06-14
refactor: DB independent quoting and truthy/falsy values

c-p-b:
Minor, straightforward changes to db independent syntax for table quoting, spacing, and truthy/falsy coercion


erpnext/accounts/doctype/loyalty_program/test_loyalty_program.py
```diff
@@ -247,7 +247,7 @@ def create_sales_invoice_record(qty=1):
 			"due_date": today(),
 			"posting_date": today(),
 			"currency": "INR",
-			"taxes_and_charges": "",
+			"taxes_and_charges": '',
```
ankush: This shouldn't be affecting it, it's python string :p

ankush:
@cpdeethree can you once confirm the reason behind reformatting python strings to use single-quotes? 

We now autoformat code using black, so it will be turned back to double quotes by black unnless it's considered unsafe like `' "double quotes inside" '`. I don't think using single quote is a problem outside of SQL strings. 


reference: https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

c-p-b:
As noted in our discussion, it looks like some of these changes happened when I originally went through and changed multiple double quoted values at once in the same file - even if it wasn't necessary. I will verify that there are no implications for postgres compatibility with the reformat you added (reverting back to the original)

c-p-b:
Verified that PG has no problem with the changes you updated with the black commit 👍

erpnext/accounts/doctype/subscription/subscription.py
```diff
@@ -148,8 +148,7 @@ def get_billing_cycle_and_interval(self):
 		billing_info = frappe.db.sql(
 			"select distinct `billing_interval`, `billing_interval_count` "
 			"from `tabSubscription Plan` "
-			"where name in %s",
-			(plan_names,),
+			"where name in {plan_names}".format(plan_names="('" + "','".join(plan_names) + "')"),
```
ankush: Is this a problem with postgres `IN` queries? if so we should handle it in framework and not here

ankush: created frappe issue: https://github.com/frappe/frappe/issues/17231

erpnext/assets/doctype/asset_maintenance/asset_maintenance.py
```diff
@@ -54,8 +54,10 @@ def assign_tasks(asset_maintenance_name, assign_to_member, maintenance_task, nex
 	}
 	if not frappe.db.sql(
 		"""select owner from `tabToDo`
-		where reference_type=%(doctype)s and reference_name=%(name)s and status="Open"
-		and owner=%(assign_to)s""",
+		where reference_type=%(doctype)s and reference_name=%(name)s and status='Open'
+		and owner={assign_to}""".format(
+			assign_to="('" + team_member + "')"
+		),
```
ankush: Not sure why this change is required?

ankush: NVM this looks like it's buggy code 😮‍💨

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -1305,7 +1305,7 @@ def item_query(doctype, txt, searchfield, start, page_len, filters):
 		if not field in searchfields
 	]
 
-	query_filters = {"disabled": 0, "ifnull(end_of_life, '5050-50-50')": (">", today())}
+	query_filters = {"disabled": 0, "end_of_life": (">", today())}
```
ankush, line 1308: This isn't correct. if end of life is not specified than it's considered to have infinite end of life.

erpnext/stock/report/incorrect_balance_qty_after_transaction/incorrect_balance_qty_after_transaction.py
```diff
@@ -73,7 +73,7 @@ def get_stock_ledger_entries(report_filters):
 		"Stock Ledger Entry",
 		fields=fields,
 		filters=filters,
-		order_by="timestamp(posting_date, posting_time) asc, creation asc",
+		order_by="posting_date asc, posting_time asc, creation asc",
```
ankush, line 76: @cpdeethree I think this is fine 

however in the past, people have suggested relying on the timestamp function instead of sorting columns individually, do you know of any real differences between these two?

erpnext/stock/doctype/warehouse/warehouse.py
```diff
@@ -161,8 +161,8 @@ def get_children(doctype, parent=None, company=None, is_root=False):
 
 	fields = ["name as value", "is_group as expandable"]
 	filters = [
-		["docstatus", "<", "2"],
-		['ifnull(`parent_warehouse`, "")', "=", parent],
+		["ifnull(`docstatus`, 0)", "<", 2],
```
ankush: IMO this condition doesn't make sense. The warehouse has no canceled status 😄

ankush (changes requested):

erpnext/accounts/doctype/subscription/subscription.py
ankush: Fixed by rewriting query

erpnext/stock/doctype/warehouse/warehouse.py
ankush: removed this

ankush (dismissed):

PR #31361, opened 2022-06-14
refactor: add SELECT columns in statements to GROUP BY

c-p-b:
If we have the following

```sql
SELECT column1, column2
...
GROUP BY column1
```

MariaDB implicitly inserts column2 into the GROUP BY clause. Postgres does not, it errors. 

This change explicitly takes those situations and specifies them in the GROUP BY clause.


ankush:
reference for mariadb: https://mariadb.com/kb/en/group-by/

> You can use any of the grouping functions in your select expression. Their values will be calculated based on all the rows that have been grouped together for each result row. **If you select a non-grouped column or a value computed from a non-grouped column, it is undefined which row the returned value is taken from.** This is not permitted if the ONLY_FULL_GROUP_BY [SQL_MODE](https://mariadb.com/kb/en/sql-mode/) is used.




If I understood this right, it's not same as implicitly adding in group by clause and adding them to group-bys will result in potentially more # of rows. 


Example:

Table:
| ColA | ColB |
| ---  | --- |
| "X"  | 1 |
| "X" |  2 |
| "Y" | 1 |

Query: 
```sql
select * from table group by colA
```

Result:

| ColA | ColB | |
| ---  | --- | :-- |
| "X"  | 1 OR 2 |  <- UB |
| "Y" | 1 | |


Query if implicit group by: 
```sql
select * from table group by colA, colB
```

| ColA | ColB |
| ---  | --- |
| "X"  | 1 |
| "X" |  2 |
| "Y" | 1 |



This will require reviewing each query case-by-case, some might even be buggy if UB affects them.  😢

ankush:
ref: https://github.com/frappe/erpnext/pull/28359#issuecomment-1277105369

PR #31372, opened 2022-06-15
chore: Fix a potential variable misuse bug

LostBenjamin:
Hi,

This pull request is a fix to a potential variable misuse bug at `erpnext/accounts/doctype/pricing_rule/pricing_rule.py`. Please check the changes.

It seems to me that the variable in the if condition should be `d` to check if `d` is `None`. If the variable should indeed be `field`, it would be better to put the condition in a separate if statement.

Best,
Jingxuan

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.py
@@ -37,7 +37,7 @@ class PricingRule(Document):
 	def validate_duplicate_apply_on(self):
 		if self.apply_on != "Transaction":
 			field = apply_on_dict.get(self.apply_on)
-			values = [d.get(frappe.scrub(self.apply_on)) for d in self.get(field) if field]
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -37,7 +37,7 @@ def validate(self):
 	def validate_duplicate_apply_on(self):
 		if self.apply_on != "Transaction":
 			field = apply_on_dict.get(self.apply_on)
-			values = [d.get(frappe.scrub(self.apply_on)) for d in self.get(field) if field]
+			values = [d.get(frappe.scrub(self.apply_on)) for d in self.get(field) if d]
```
marination: Let's check for an empty table before hand, so both checks stay
```suggestion
			if not field:  # table is empty
			    return
			values = [d.get(frappe.scrub(self.apply_on)) for d in self.get(field) if d]
```

marination (dismissed):

marination (approved):

PR #31405, opened 2022-06-20
feat: Cash and Non trade discounts in Sales Invoice

deepeshgarg007:
In many cases, the discount given on Invoices on the grand total amount is not an itemized discount. Its a cash or a non-trade discount, means that this type of discount doesn't affect the tax value (GST) applied on the items and is reported on the invoice discount while E-Invoicing

image

If the discount applied on grand total is a Cash or Non-Trade discount user can check the "Is Cash Or Non Trade" tick and it won't be affecting any item values or have any effect on taxes else it will be considered as an itemized discount

`no-docs`


deepeshgarg007:
Unrelated test case failing, merging this

deepeshgarg007:

PR #31421, opened 2022-06-22
fix: add UOM validation for planned-qty

### LINKED ISSUE #30234: UOM not validated in production plan or work order
### Information about bug

non-fractional UOM gets validated at last stage (stock entry) in manufacturing cycle. 

e.g. test written in this PR shouldn't pass since selected UOM is non-fractional: https://github.com/frappe/erpnext/pull/30216/files 

### Module

manufacturing

### Version

v13 / v14

### Installation method

_No response_

_No response_

s-aga-r:
#30234

### DIFF
--- a/erpnext/manufacturing/doctype/production_plan/test_production_plan.py
+++ b/erpnext/manufacturing/doctype/production_plan/test_production_plan.py
@@ -679,15 +679,23 @@ class TestProductionPlan(FrappeTestCase):
 		self.assertFalse(pp.all_items_completed())
 
 	def test_production_plan_planned_qty(self):
-		pln = create_production_plan(item_code="_Test FG Item", planned_qty=0.55)
-		pln.make_work_order()
-		work_order = frappe.db.get_value("Work Order", {"production_plan": pln.name}, "name")
-		wo_doc = frappe.get_doc("Work Order", work_order)
-		wo_doc.update(
-			{"wip_warehouse": "Work In Progress - _TC", "fg_warehouse": "Finished Goods - _TC"}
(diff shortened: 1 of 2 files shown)


erpnext/manufacturing/doctype/production_plan/test_production_plan.py
```diff
@@ -679,7 +679,14 @@ def test_qty_based_status(self):
...
 	def test_production_plan_planned_qty(self):
-		pln = create_production_plan(item_code="_Test FG Item", planned_qty=0.55)
+		from erpnext.stock.doctype.item.test_item import make_item
+
+		fg_item = make_item(properties={"is_stock_item": 1, "stock_uom": "_Test UOM 1"}).name
+		bom_item = make_item().name
+
+		make_bom(item=fg_item, raw_materials=[bom_item], source_warehouse="_Test Warehouse - _TC")
+
+		pln = create_production_plan(item_code=fg_item, planned_qty=0.55)
```
ankush: THe issue is not about making this test pass or using correct values here. The issue actually applies to end user too. 

If you use Nos and set qty as 0.5 production plan currently allows doing that... which doesn't make sense. 

Screenshot 2022-06-22 at 12 25 28 PM

s-aga-r:
Failing tests are not related. #[00807ab](https://github.com/frappe/erpnext/pull/31421/commits/00807abe314344fec1ec636ac5f281473589f7f1)

marination (approved):

marination:
Can confirm unrelated failing tests

PR #31425, opened 2022-06-22
fix(Salary Slip): Components not updated when amount evaluates to 0 due to payment days

ruchamahabal:
## Steps to Replicate:

- Salary component **Professional Tax** is dependent on Gross Pay and Bonus and these components are dependent on payment days:
   formula
- If the Payment Days are altered after saving by updating "Leave Without Pay", the gross changes to 12000 from 15000 and the professional tax should be 0 as per the formula. But this amount isn't updated because it fails the check made before calling `update_component`

  before

## Fix:

Explicitly check whether the `amount is not None` so that 0 valued components get updated after payment days are altered:

after


ruchamahabal:
This will add 0 valued components to the table though 🤔 (which doesn't happen right now). Need to handle that.

PR #31429, opened 2022-06-22
fix: Non-annual depreciation by WDV method

GangaManoj:
## Issues

1. Depreciation by WDV method only works if the Asset depreciates annually, as the Rate of Depreciation is calculated by using asset life in years.
2. If an Asset that depreciates monthly starts depreciating on, say, Feb 28, then it would depreciate next on March 28, then April 28 and so on. However, customers usually intend to have the asset depreciate on the last day of each month in this case- i.e. Feb 28 should be followed by March 31 and April 30 instead.

## Fixes

1. Take `number_of_depreciations` instead of asset life in years to calculate the Rate of Depreciation.

eg: For an Asset purchased for Rs 10,000 that started being used on 15/02/2022 and has a salvage value of Rs 5000, the Rate of Depreciation would be 12.9% if it depreciates monthly, yielding the following schedule:

Screenshot 2022-06-22 at 11 12 35 PM

2. Check if the `depreciation_start_date` falls on the last day of its month, and if so, modify the schedule dates so that they'd each fall on the last day of their respective months.

Screenshot 2022-06-09 at 2 38 31 AM

<details>

ISS-22-23-00424

</details>


GangaManoj:

nabinhait (approved):

PR #31438, opened 2022-06-23
feat!: Only Project Manager and administror able to view Costing and Margin section in Project

vishdha:
Costing and Billing section and Margin section info visible to user with project Manager role only

### DIFF
--- a/erpnext/patches/v14_0/update_project_permission.py
+++ b/erpnext/patches/v14_0/update_project_permission.py
@@ -1,0 +1,10 @@
+import frappe
+from frappe.permissions import add_permission, update_permission_property
+from frappe.core.page.permission_manager.permission_manager import remove
+
+def execute():
+	frappe.reload_doc("projects", "doctype", "project")
+
+	add_permission("Project", "Projects Manager", 1)
+	update_permission_property("Project", "Projects Manager", 1, "write", 1)
+	remove("Project", "All", 1)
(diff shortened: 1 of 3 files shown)


barredterra:
@vishdha did you test this? It doesn't seem to work for me. I tried the following:

1. Create a new site with erpnext
2. Create a new **Project** as system manager
3. Create a new **User** with role “Project User” (nothing else)
4. Log in as project user

    -> Can still see section costing & billing and margin.

(I checked the Project DocType, it was synced correctly.)

ankush:
> Can still see section costing & billing and margin.



Should we be adding role-based perm levels in the core? 

This is specific to individual organizations and should ideally be configured during the implementation phase 😅

barredterra:
@ankush I agree, but in this case we thought that the requirement is common enough. Frappe is probably one of the few organizations where everyone is allowed to see the financial details of a project. In my experience, most companies want to hide them from project users.

IMO it makes sense that financial details get a different permlevel than other project fields.

erpnext/patches/v14_0/update_project_permission.py
```diff
@@ -0,0 +1,10 @@
+import frappe
+from frappe.permissions import add_permission, update_permission_property
+from frappe.core.page.permission_manager.permission_manager import remove
+
```
barredterra, line 4:
```suggestion


```

vishdha:
Closing this as customized PR added

barredterra:
Let's get this merged anywas, if Frappe agrees

PR #31439, opened 2022-06-23
fix: General Ledger and TB opening entries mismatch issues

deepeshgarg007:
**1. General Ledger Issue when the "Show Opening Entries" check is enabled**

An "Opening Entry" against an account should be considered as an opening entry irrespective of the date when they were posted. Currently opening entries having posting greater than the GL report `to_date` were getting ignored when the "Show Opening Entries" check was enabled.

**2. Removed account fetching for Opening Entries in Journal Entry** 

On selecting `voucher_type` as Opening Entry in Journal Entry all balance sheet accounts except stock accounts are fetched. Have rarely seen anyone using this. It causes more annoyance to most users who don't need this than helping them.

**3. Make the finance book field in Journal Entry read-only**

The Finance book field in Journal Entry was only added for the purpose of single Asste depreciations via multiple methods as a compliance requirement for some companies. Many users end up using that without having proper knowledge or context about it and later end up having issues while reporting

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -149,22 +149,6 @@ frappe.ui.form.on("Journal Entry", {
 					}
 				});
 			}
-			else if(frm.doc.voucher_type=="Opening Entry") {
-				return frappe.call({
-					type:"GET",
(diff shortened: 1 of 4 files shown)


deepeshgarg007:

nabinhait (approved):

PR #31446, opened 2022-06-24
fix: Quotation and Sales Order item sync

deepeshgarg007:
Had recently introduced partial billing status in Quotation https://github.com/frappe/erpnext/pull/31322

Even after creating a Sales Order against all items of the quotation user was able to make a new Sales Order against it
Synced items between Sales Order and Quotation so that only remaining items are ordered


deepeshgarg007:

PR #31460, opened 2022-06-27
fix: 'attribute error' on AR/AP report with delivery note filter

ruthra-kumar:
Fixing attribute error on AR/AP report with 'Show Delivery Note' filter selected.
Screenshot 2022-06-27 at 5 32 22 PM

### DIFF
--- a/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
+++ b/erpnext/accounts/report/accounts_receivable/accounts_receivable.py
@@ -128,6 +128,7 @@ class ReceivablePayableReport(object):
 					credit_note_in_account_currency=0.0,
 					outstanding_in_account_currency=0.0,
 				)
+			self.get_invoices(ple)
 
 			if self.filters.get("group_by_party"):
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/accounts_receivable/accounts_receivable.py
```diff
@@ -128,6 +128,7 @@ def init_voucher_balance(self):
 					credit_note_in_account_currency=0.0,
 					outstanding_in_account_currency=0.0,
 				)
+			self.get_invoices(ple)
```
ankush, line 131: Just a note on naming, a method called `get_SOMETHING` should not be mutating state. `get` implies "read something and return it to me"

ruthra-kumar: Yeah, yeah. I'm off the same opinion. But, even before the refactor, `get` in function name is loosely used from a class instance point of view. Since the refactor is to integrate payment ledger, only the underlying function logic is changed and the names are left as is. 😄

PR #31462, opened 2022-06-27
fix: open lead and opportunities based on today's event

nabinhait:
- Restored city, state and country fields
- open lead and opportunities based on today's event

### DIFF
--- a/erpnext/crm/doctype/opportunity/opportunity.json
+++ b/erpnext/crm/doctype/opportunity/opportunity.json
@@ -32,9 +32,12 @@
   "column_break_23",
   "industry",
   "market_segment",
-  "column_break_31",
-  "territory",
   "website",
+  "column_break_31",
+  "city",
+  "state",
+  "country",
+  "territory",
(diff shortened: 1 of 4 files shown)


erpnext/crm/doctype/opportunity/opportunity.json
```diff
@@ -463,9 +466,8 @@
   },
   {
    "fieldname": "no_of_employees",
-   "fieldtype": "Select",
-   "label": "No of Employees",
-   "options": "1-10\n11-20\n21-30\n31-100\n11-50\n51-200\n201-500\n101-500\n500-1000\n501-1000\n>1000\n1000+"
+   "fieldtype": "Int",
```
ankush, line 469: @nabinhait won't this require separate patch to migrate text data to integers?

ankush: looks like it was added in last PR only, so doesn't affect deployments yet...

PR #31469, opened 2022-06-28
chore: Make `image` field obsolete in Website Item (redundant)

marination:
via https://github.com/frappe/erpnext/pull/30081
**Issue:**
- `website_image` and `image` both coexist in Website Item, with `website_image` being used the most
- `website_image` may always have a value. If `image` is set, `website_image` used to be auto set in older versions. Else `website_image` could be set explicitly.
- `website_image` has been used for thumbnail generation as well
- Need single source of truth for images in Website Item and remove redundant fields

**Fix:**
- **Delete `image` field from Website Item** and set `website_image` as form's image field for uploads
- Remove instances of `image` field access via Website Item
- Item -> Web Item via **Form UI**: Map Item's `image` to Web Item's `website_image`
- Item -> Web Item via **patch** (for v12 items): `website_image` will be **mapped** and  `thumbnail`
- Remove magic that auto-sets `website_image` from `image` in Website Item
- No patch needed: If `image` was set, `website_image` was 100% set. If `website_image` is still unset, there was no image
Screenshot 2022-07-07 at 2 35 59 PM

> For v12: **image** = **website_image** unless **image** is not set

**To Test:**
- Item -> Website Item: via Desk (Publish Item) [make sure Item has image]
- Check products page, cart, wishlist, recommended items, homepage


marination:
Unrelated linter failure

PR #31473, opened 2022-06-28
fix: validate item dashboard material transfer entry

dj12djdjs:
porting the validation from `version-13`

https://github.com/frappe/erpnext/blob/43cb4fcbb1ffcb0f9ffd1662f6c57d665dc47b21/erpnext/stock/dashboard/item_dashboard.js#L260-L266


ankush:
is this really required? I removed it because now we just open Stock Entry page, there user will be asked to update missing info 👀 


I guess it makes UX somewhat better.

dj12djdjs:
> is this really required? I removed it because now we just open Stock Entry page, there user will be asked to update missing info 👀
> 
> I guess it makes UX somewhat better.

The reason we need this is because we sell into negative stock to customers, then receive the product later. This is a guard for us so operators can't make typo inputing qty when transferring stock between bins.

PR #31493, opened 2022-06-30
fix: Internal PI link in Sales Invoice

deepeshgarg007:
Screenshot 2022-07-02 at 6 27 14 PM

Added dashboard link for Internal Transfer Purchase Invoices in Sales Invoice doctype.
Also added missing logic for inwarding only unbilled items rather than simply copying items from Sales Invoice to Purchase Invoice


deepeshgarg007:

PR #31497, opened 2022-06-30
fix: translated message for credit limit blocking

HarryPaulo:
Translation for the blocking message when the customer has no credit to submit an invoice.

doctype to reproduce: Sales Invoice


HarryPaulo:
waiting reviewers

PR #31506, opened 2022-07-02
fix: link error on supplier deletion

safvanhuzain:
When deleting suppliers with primary address, system throws link error.
supplier


s-aga-r (changes requested):
@safvanhuzain add test for this.

PR #31513, opened 2022-07-04
fix: incorrect rate in BOM exploded items

s-aga-r:
Exploded Item rate showing incorrect value in the saved state.

https://user-images.githubusercontent.com/63660334/177125719-2cee7bc5-b6e3-4eca-801b-ceaaa0f17b48.mov


s-aga-r:
Failing tests are not related #[88e2b04](https://github.com/frappe/erpnext/pull/31513/commits/88e2b047f1fdd2c3a4ddc4e8cfeb584661c394d1)

erpnext/manufacturing/doctype/bom/bom.py
```diff
@@ -189,8 +189,8 @@ def validate(self):
 		self.validate_transfer_against()
 		self.set_routing_operations()
 		self.validate_operations()
-		self.update_exploded_items(save=False)
 		self.calculate_cost()
+		self.update_exploded_items(save=False)
```
marination, line 193: `calculate_cost` requires updated exploded items, which is why the old sequence was as such.
Can you explain the approach here ?

marination: Okay I assume its getting stale Items table row rate right ? then makes sense

s-aga-r: Yeah, Can we make this rate field read-only? bcoz rate is set by the system based on the "Rate Of Materials Based On" field and It also solves this wrong rate problem.

marination: @s-aga-r  So earlier any custom rate entered by the user was respected and rates were not reset as per the **Rate Of Materials Based On** field (valuation/last purchase rate, etc)
A recent fix changed that so it does respect the **Rate Of Materials Based On** field, but should users be allowed to tweak the rate?
My opinion aligns with yours, make the rate field read only. @rohitwaghchaure will that work or we missing a use case?

s-aga-r: @rohitwaghchaure @marination 

image

marination: @s-aga-r they want line item level price decision i.e. from item to item it will change where the rate should be picked from. Right now its on the BOM document level.
What they are asking for is a new feature. That can be excluded from this PR and dealt with separately.
They can add "valuation_rate" on Item master if they dont want it as 0 (workarounds)

marination:
@s-aga-r You can make the rate field read only and add a test to change qty in items table and check reflected changes in the exploded items table on save.
2 things to get this PR merged

PR #31516, opened 2022-07-04
fix: Incorrect provisional expense booking while reposting

deepeshgarg007:
In cases where a single Purchase Receipt contains both stock and non-stock items is reposted provisional GL Entries with no account is posted which leads to incorrect balances in reports


deepeshgarg007:

PR #31521, opened 2022-07-05
fix: components in the same table don't get updated value of prev payment-days based component

ruchamahabal:
## Steps to Replicate

- Salary Component Basic has "Depends on Payment Days" enabled.
- Salary Component Bonus (which is in the same table: earnings) is dependent on this component by formula.

	salary-struc
- Salary Slip with 6 LWP. 
	As per the formula `1249.5 if (B>=MBS) else (B * .0833)` the Bonus amount should be 
	`(B * 0.0833) = 12096.77 * 0.0833 = 1008.0  `
	since `B(12096.77) >= MBS(15000)` is false.
- But the amount is 1250 because the Bonus formula is not getting the updated value of Basic (12096.77). It is using the old basic amount = 15000. 
	
	bonus

- This works fine if a component in another table is referring to a component in the current table. i.e if deductions component is dependent on earnings component by formula. But doesn't work if components from the same table are dependent. This is because, after processing earnings, the `get_data_for_eval` is called again for fetching all salary component's updated values. 

## Fix

- All formulae calculation uses `data` which has values for all the salary components to evaluate the amount. Once the amount is changed as per payment days, update the amount in `data` too, so that other components get the updated value.

bonus-recalculated

- Added validation for the scenario mentioned in: https://github.com/frappe/erpnext/pull/31521#issuecomment-1174598298

	image

### DIFF
--- a/erpnext/payroll/doctype/salary_slip/salary_slip.py
+++ b/erpnext/payroll/doctype/salary_slip/salary_slip.py
@@ -626,7 +626,7 @@ class SalarySlip(TransactionBase):
 		for struct_row in self._salary_structure_doc.get(component_type):
 			amount = self.eval_condition_and_formula(struct_row, data)
 			if amount is not None and struct_row.statistical_component == 0:
-				self.update_component_row(struct_row, amount, component_type)
+				self.update_component_row(struct_row, amount, component_type, data=data)
 
 	def get_data_for_eval(self):
 		"""Returns data for evaluating formula"""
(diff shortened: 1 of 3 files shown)


ruchamahabal:
One problem:

After this update: If a component like Bonus (dependent on another payment-days-based component) has `Depends on Payment Days` checked, it will change the amount twice as per payment days. 

In that case, we can show some message like this:

image

But it can be confusing and wordy(?)

ruchamahabal:
Unrelated asset test failing

nabinhait (approved):

ruchamahabal:
Unrelated test failure: `test_sales_invoice_against_supplier`

PR #31526, opened 2022-07-05
fix: update fr translations

FHenry:
Update french translations

Note to reviewer :
the litteral translation of Work Order is Ordre de travail
Work (en) = Travail (fr)
Order (en) = Ordre (fr)

But in ERP system in french we commonly use "Ordre de fabrication" (basically we use the acronym OF). 
The closest way english translation would be Manufacture Order, this is not the standard DocType name but it ERPNext implement the feature of Manufacture Order with Work Order. It's most common to use ERPNext Work Order to manufacture something than just do operation without something to produce

in french 
https://www.odoo.com/fr_FR/app/manufacturing-features
in english
https://www.odoo.com/app/manufacturing-features

In fact ERPNext Work Order are Manufacturing Order
Work Order in Odoo are ERPNext Job Card


erpnext/translations/fr.csv
```diff
@@ -4400,7 +4400,7 @@ Motor Assessment Scale,Échelle d&#39;évaluation motrice,
 The following {0} were created: {1},Les {0} suivants ont été créés: {1},
 Work Orders,Bons de travail,
 The {0} {1} created sucessfully,Le {0} {1} a été créé avec succès,
-Work Order cannot be created for following reason: <br> {0},L&#39;ordre de travail ne peut pas être créé pour la raison suivante:<br> {0},
+Work Order cannot be created for following reason: <br> {0},L&#39;ordre de fabrication ne peut pas être créé pour la raison suivante:<br> {0},
```
marination:
```suggestion
Work Order cannot be created for following reason: <br> {0},L'ordre de fabrication ne peut pas être créé pour la raison suivante:<br> {0},
```

marination (approved):

PR #31537, opened 2022-07-06
refactor: create index for payment ledger

ruthra-kumar:
Create DB index for Payment Ledger Entry.


erpnext/accounts/doctype/payment_ledger_entry/payment_ledger_entry.py
```diff
@@ -147,3 +147,8 @@ def on_update(self):
 			update_voucher_outstanding(
 				self.against_voucher_type, self.against_voucher_no, self.account, self.party_type, self.party
 			)
+
+
+def on_doctype_update():
+	frappe.db.add_index("Payment Ledger Entry", ["against_voucher_type", "against_voucher_no"])
+	frappe.db.add_index("Payment Ledger Entry", ["voucher_type", "voucher_no"])
```
ankush: Do you need both multi-column AND single-column indexes for these? 


I am guessing most queries will have where clause on both `voucher_type` and `voucher_no` (But don't go by my assumption, look at most used queries)

would recommend reading this: https://use-the-index-luke.com/sql/where-clause/the-equals-operator/concatenated-keys

ruthra-kumar: Makes sense. Thanks for the recommendation. 👍🏾

deepeshgarg007 (approved):

PR #31549, opened 2022-07-07
chore: remove missed code for India localisation

sagarvora:
- Address Template
- DocType **HSN Tax Rate**
- HSN Code Data
- Unlink GST HSN Code in Item (will get updated if India Compliance is subsequently installed)
- Add check to ensure that patches don't run for users who have already installed India Compliance.


sagarvora:
> * Add check to ensure that patches don't run for users who have already installed India Compliance.

Deprecated docs are removed separately when India Compliance is installed in [this post-install patch](https://github.com/resilient-tech/india-compliance/blob/develop/india_compliance/patches/post_install/remove_deprecated_docs.py).

PR #31552, opened 2022-07-07
feat: Allow many doctypes to be used as Sales Order Item previous doc

dj12djdjs:
This PR makes it easier to add multiple mappings to Sales Order from more than just Quotation

As long as these conditions are met, you can map any custom doctype to Sales Order:
1. The doctype inherits StatusUpdater
2. The doctype contains a matching set of Order Types in relation to Sales Order.

Requires: https://github.com/frappe/frappe/pull/17445

`no-docs`


dj12djdjs:
I'm currently stuck on this, as the dashboard is setup to always link `prevdoc_docname` with Quotation.

https://github.com/frappe/erpnext/blob/832a863d1e1af0a0def8a2305a4376abcc003962/erpnext/selling/doctype/sales_order/sales_order_dashboard.py#L15

I need to find a way to make this only match if `prevdoc_doctype === Quotation`

deepeshgarg007:
@dj12djdjs Can you please specify the use case as well. Other than Quotation what other doctype you wish to link?

dj12djdjs:
@deepeshgarg007 I can't think of anything native to ERPNext. This is specifically for better integration with custom apps.

If you want to use a sales method other than quotation, this will make integrating to Sales Order a breeze. No need to add a custom field to link it. Just specify the `prevdoc_doctype`, and `prevdoc_docname` and override `override_doctype_dashboards` to add your doctype. All done.

dj12djdjs:
@deepeshgarg007 Is this functionality something frappe team is interested in keeping for sales order?

PR #31556, opened 2022-07-08
fix: Route condition set with proper filter

nihantra:
**`Version`**
ERPNext: v13.35.0 (HEAD)
Frappe Framework: v13.35.0 (HEAD)

**Before:**
- In Member doctype, when clicking Accounting Ledger then redirect to General Ledger report but the filter does not work with a customer. and when clicking Accounts Receivable then redirects to the Accounts Receivable report but one issue is like the member filter has not in the report.

https://user-images.githubusercontent.com/99652762/176614354-67ae2059-21d5-4ba9-ab7c-af11a4678423.mp4

**After:**
- After developing, if the customer has in member form and clicks Accounting Ledger then will redirect to General Ledger report with customer filter.
```
if (frm.doc.customer) {
	frappe.set_route('query-report', 'General Ledger', {party_type: 'Customer', party: frm.doc.customer});
} else {
	frappe.set_route('query-report', 'General Ledger', {party_type: 'Member', party: frm.doc.name});
}
```

- And Also set conditions for the Accounts Receivable report
```
if (frm.doc.customer) {
	frm.add_custom_button(__('Accounts Receivable'), function() {
		frappe.set_route('query-report', 'Accounts Receivable', {customer: frm.doc.customer});
	});
}
```
https://user-images.githubusercontent.com/99652762/176615788-5eae6fbc-4743-47d8-85c3-40103bddfad3.mp4

Thank You!


erpnext/non_profit/doctype/member/member.js
```diff
@@ -21,12 +21,15 @@ frappe.ui.form.on('Member', {
...
+					frappe.set_route('query-report', 'General Ledger', {party_type: 'Customer', party: frm.doc.customer});
+				} else {
+					frappe.set_route('query-report', 'Accounts Receivable', {party_type: 'Member', party: frm.doc.name});
+				}
 			});
 
 			frm.add_custom_button(__('Accounts Receivable'), function() {
-				frappe.set_route('query-report', 'Accounts Receivable', {member:frm.doc.name});
+				frappe.set_route('query-report', 'Accounts Receivable', {customer: frm.doc.customer});
 			});
```
marination: @nihantra Can we hide this button if there's no customer set? Going to AR report without customer is pointless. Lets not give users the option unless valid

nihantra: Hi @marination,
As per your according, added a condition like if have a customer in member doctype then AR will appear.


https://user-images.githubusercontent.com/99652762/180206359-4b1ce1cf-97ee-42eb-a20d-86af9cdee385.mp4

marination (approved):

marination:
@ChillarAnand @ruchamahabal  will have to port this to the non profit app too right ?

ruchamahabal:
> @ChillarAnand @ruchamahabal will have to port this to the non profit app too right ?

@marination I will do it. Thanks for reviewing

PR #31560, opened 2022-07-08
fix: Maintain Stock Value

nihantra:
**`Version`**

**Before:**

- In Item doctype, when Maintain Stock checked to unchecked then an error show like
```
pymysql.err.OperationalError: (1054, "Unknown column 'parent' in 'field list'")
```

https://user-images.githubusercontent.com/99652762/178058194-6fd0546b-f7b5-471e-a993-5af6613f9472.mp4

**After:**
- If condition and after elif condition is not set properly then set it via tab.

https://user-images.githubusercontent.com/99652762/178058259-100b7e92-7455-4fe6-8807-bba8831139ae.mp4

Thank You!

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -958,13 +958,13 @@ class Item(Document):
 				):
 					return linked_doc
 
-			elif linked_doc := frappe.db.get_value(
-				doctype,
-				filters,
-				["parent as docname", "parenttype as doctype"],
-				as_dict=True,
-			):
-				return linked_doc
+				elif linked_doc := frappe.db.get_value(
+					doctype,
+					filters,
+					["parent as docname", "parenttype as doctype"],
+					as_dict=True,
+				):
+					return linked_doc
 
 	def validate_auto_reorder_enabled_in_stock_settings(self):
 		if self.reorder_levels:
--- a/erpnext/stock/doctype/item/test_item.py
+++ b/erpnext/stock/doctype/item/test_item.py
@@ -778,6 +778,22 @@ class TestItem(FrappeTestCase):
 			item.has_batch_no = 1
 			item.save()
 
+	def test_update_is_stock_item(self):
+		# Step - 1: Create an Item
+		item = make_item(properties={"is_stock_item": 1})
+
+		# Step - 2: Set is_stock_item = 0
+		item.is_stock_item = 0
+		item.save()
+		item.reload()
+		self.assertEqual(item.is_stock_item, 0)
+
+		# Step - 3: Again set is_stock_item = 1
+		item.is_stock_item = 1
+		item.save()
+		item.reload()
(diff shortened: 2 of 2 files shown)


s-aga-r (changes requested):
Add a simple test case:
* Create an Item. 
* Update Maintain Stock value (1 --> 0, 0 --> 1).

nihantra:
Hi @s-aga-r,

As per your scenario, create a video of a demo sample for a new item.

https://user-images.githubusercontent.com/99652762/178113162-37632e7e-3733-45a7-89ed-80eb614217ef.mp4

Thank You!

s-aga-r:
> Hi @s-aga-r,
> 
> As per your scenario, create a video of a demo sample for a new item.
> 
>  new_item.mp4 
> Thank You!

You need to add a test case [here](https://github.com/frappe/erpnext/blob/075d2c3fe1871d521a737ca5c47774965ceeef5e/erpnext/stock/doctype/item/test_item.py), see other methods for reference.

nihantra:
Hi @s-aga-r,

Please check the test case

Thank You!

s-aga-r:
> Hi @s-aga-r,
> 
> Please check the test case
> 
> Thank You!

@nihantra Check the last commit #01f2a43 for reference.

nihantra:
> > Hi @s-aga-r,
> > Please check the test case
> > Thank You!
> 
> @nihantra Check the last commit #01f2a43 for reference.

Thank You @s-aga-r for explaining the full scenario of the test case and how to test and commit the test case.

s-aga-r (approved):

marination:
@s-aga-r This fix seems wrong to me. The elif was intentionally outside. If the doctype is not PI Item or SI Item then we check if this doctype has a record with this item code.
All the doctypes are child doctypes, except Product Bundle. The breakage is happening due to Product Bundle from what i can see. Thats the only thing that needs to be tweaked

erpnext/stock/doctype/item/item.py
```diff
@@ -958,13 +958,13 @@ def _get_linked_submitted_documents(self, changed_fields: List[str]) -> Optional
...
-				as_dict=True,
-			):
-				return linked_doc
+				elif linked_doc := frappe.db.get_value(
+					doctype,
+					filters,
+					["parent as docname", "parenttype as doctype"],
+					as_dict=True,
+				):
+					return linked_doc
```
marination, line 967: This can be reverted. The `elif` is outside so that it handles:
``` py
if doctype not in ("Purchase Invoice Item","Sales Invoice Item",)
```

From the `linked_doctypes` list only Product Bundle will not have a parent or parenttype.

On line `946` after `if doctype == "Product Bundle":` we can fetch the Product Bundle details and return that back. That should cover not allowing to change important fields if Product Bundle exists against item

@s-aga-r seems okay? Or am i missing something?

PR #31566, opened 2022-07-11
fix: Allow multi currency invoice against single party account

deepeshgarg007:
A validation was introduced in this https://github.com/frappe/erpnext/pull/26916, some users don't view reports in Party's billing currency and are okay with the exchange rate differences in the reports. Such users who are aware of the effects of using a single account for multicurrency and are okay with it can enable this and use one single account for multicurrency

By default, this has been kept disabled so that unaware users don't end up using the same account for multicurrency invoices.

Screenshot 2022-07-11 at 1 38 05 PM


deepeshgarg007:

barredterra:
Thanks for the fix, @deepeshgarg007!

PR #31576, opened 2022-07-12
fix: credite note for returned delivery note updates SO's billed percentage

ruthra-kumar:
### Issue
1. Create SO -> DN -> SI
2. Create Return Delivery against DN -> Credite Note

Credit note should update Percentage billed in Sales Order.

### Fix
Loosen display criteria for 'Update billed amount in Sales Order' check box. This is so user can make use of this option in above case.


deepeshgarg007:

PR #31579, opened 2022-07-12
fix: display customer name on picking list

dj12djdjs:
For those who use naming series for customer, it is difficult to know what customer is assigned to the picking list at first glance.


erpnext/stock/doctype/pick_list/pick_list.json
```diff
@@ -159,11 +160,18 @@
    "fieldname": "prompt_qty",
    "fieldtype": "Check",
    "label": "Prompt Qty"
+  },
+  {
+   "depends_on": "eval:doc.purpose==='Delivery' && doc.customer",
+   "fetch_from": "customer.customer_name",
+   "fieldname": "customer_name",
+   "fieldtype": "Data",
+   "label": "Customer Name"
```
marination: Make it read only ?

marination (approved):

PR #31583, opened 2022-07-12
fix: (india) (e-invoice) CN / DN with 0 qty and discount issues

maharshivpatel:
Qty 0 is allowed when creating Credit or Debit Note this caused ZeroDivisionError during unit_rate calculation.
fixed the issue by adding required conditionals.

There were multiple issues due to discounts fixed them, added an option for user to choose to report discount or net rate in e-invoice and added tests for discount cases.


deepeshgarg007:

PR #31584, opened 2022-07-12
fix: picking list picked_qty cannot be greater than stock_qty

dj12djdjs:
This PR adds a validation that will ensure user doesn't pick more than the required qty.


dj12djdjs:
linter is passing locally

image

PR #31587, opened 2022-07-13
fix: Make Payment via Journal / Payment Entry

Solufyin:
**`Version`**

**Before:**
- Make Payment via Journal Entry does not show in Accounts Setting because it's is hide from Accounts Setting.
image

**After:**
- Remove hide option in Make Payment via Journal Entry from Accounts Setting.
image

Thank You!

### DIFF
--- a/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
+++ b/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
@@ -100,7 +100,7 @@
    "default": "0",
    "fieldname": "make_payment_via_journal_entry",
    "fieldtype": "Check",
-   "hidden": 1,
+   "hidden": 0,
    "label": "Make Payment via Journal Entry"
   },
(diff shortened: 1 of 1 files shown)


Solufyin:
Hi @deepeshgarg007, @nextchamp-saqib and @ruthra-kumar,

Is there any problem in PR?

deepeshgarg007:
@Solufyin We don't want users to use Journal Entry for payments and user Payment Entry only as in Journal Entry many cases are not handled. The check was hidden for backward compatibility so the users who have already configured can keep on using it. Any reason you need JVs to make payments instead of payment entries?

Solufyin:
Hi @deepeshgarg007,

I know both sides' payment entries do not manage and I haven't any problem.
But some client payment entries create via Journal Entry and some are used Payment Entry.

If convenient for the changes then go ahead with the process.

Thank You!

Issue #31536, opened 2022-07-05
translation mistake on new-employee-checkin


billy995:
### Information about bug

Can't find file to change german translation  IM to EIN

![grafik](https://user-images.githubusercontent.com/48571007/177422779-a0a738a0-0386-4f49-9d68-07639acfd1e5.png)

### Module

HR

### Version

Frappe Framework: v13.29.1 (version-13)
ERPNext: v13.30.0 (version-13)

### Installation method

manual install

dj12djdjs:
Hello @billy995,

Here is the file location for German translations.
https://github.com/frappe/erpnext/blob/0b822dade01c16a0f8319744113885f6b53bdcf4/erpnext/translations/de.csv#L6006

billy995:
Thanks dj12djdjs, is it possible that you will fix this bug in the next version? Otherwise there could be translation mistakes for  German-speaking users

PR #31596, opened 2022-07-14
Update de.csv

billy995:
translation mistake on new-employee-checkin
IN,IM to IN,EIN


erpnext/translations/de.csv
```diff
@@ -6003,7 +6003,7 @@ CN,CN,
 DE,DE,
 ES,ES,
 FR,FR,
-IN,IM,
+IN,Ein,
```
marination, line 6007: @billy995  hey can you give some instances where **Ein** is used as IN (context IN/OUT) ? From everything I can find online **IN** translates to IN (english) and **IM** translated to IN THE (english)

EIN seems to translate to "a/an" (article)

billy995:
https://github.com/frappe/erpnext/issues/31536#issue-1294820239

Marica ***@***.***> schrieb am Di., 19. Juli 2022, 11:03:

erpnext/translations/de.csv
billy995: @marination  on new-employee-checkin [example : stamp in = einstempeln   stamp out = ausstempeln] in = ein, out = aus
grafik

marination: Had to check if IN was used elsewhere in other context. Nothing so far so this is good to go

marination (approved):

PR #31599, opened 2022-07-14
fix: Supplier details in TDS monthly report

deepeshgarg007:
Supplier details for TDS deducted via Journal Entry were not visible in TDS payable monthly report if the party row was not the first row in the accounts table, added a fallback in the report to fix this.


deepeshgarg007:

PR #31601, opened 2022-07-14
perf: reduce db calls for warehouse wise stock value chart

dj12djdjs:
Count of Warehouses: 13579
Count of Bins: 17702

**Before**
Warehouse Wise Before

**After**
Warehouse Wise Afterpng


marination:
@dj12djdjs How many Warehouses and Bin records are these benchmarks based on? Can you mention those in the description?

dj12djdjs:
@marination The description has been updated.

marination (approved):

PR #31602, opened 2022-07-15
fix(india): e-way bill json for unregistered gst category

rtdany10:
Currently, for unregistered customers, E-Way Bill JSON can only be generated via Delivery Note.
Have added provision to generate it via Sales Invoice as well.


rtdany10:
@deepeshgarg007 Please check

PR #31609, opened 2022-07-16
fix(India): Inward supplies from Composition Supplier in GST 3B report

deepeshgarg007:
Invoices based on GST category were not getting segregated properly because gst category was not selected in the query, which led to incorrect reporting of inward supplies from a composite supplier.


sagarvora:
@deepeshgarg007 Please port to IC if applicable.

PR #31621, opened 2022-07-18
fix(accounts): `query_for_outstanding` apply common_filter to whole of query.

### LINKED ISSUE #31607: Payment Ledger Entry `get_voucher_outstandings` doesn't consider reconciliations before filter posting date.
### Information about bug

Given the filters posting date from 2022-01-01 to 2022-07-15 "Get Outstanding" will produce only 1 invoice as shown below:

example 2022-01-01 to 2022-07-15

If we reduce the filter to only 1 month instead of the full year, for example 2022-06-15 to 2022-07-15 magically two invoices appear as shown below:
 
example 2022-06-15 to 2022-07-15

In the second example there is 1 invoice with $3.66 outstanding.

Let's take a look at the Payment Entry Ledger.

You will see, **highlighted in teal** the credit entries before 2022-06-15 have been reconciled against this invoice.
Because these entries are limited out of the query by the posting date filter. A balance of $3.66 still shows remaining on the already closed invoice.

image

### Module

accounts

### Version

frappe develop
erpnext develop

### Installation method

manual install

_No response_

dj12djdjs:
closes https://github.com/frappe/erpnext/issues/31607

Changes to `QueryPaymentLedger`:
1. Remove common_filter from `query_voucher_amount`, and `query_voucher_outstanding` queries.
2. Join `Payment Ledger Entry` to `cte_query_voucher_amount_and_outstanding` and applied common_filter to it.

TestCase:
Added a test case to ensure invoices allocated before the from_invoice_date filter won't show as unallocated.


dj12djdjs:
Closing as fixed in: https://github.com/frappe/erpnext/pull/31737

PR #31622, opened 2022-07-18
feat(stock): port quick stock entry from version-13

dj12djdjs:
`no-docs`

I'm not sure why it was decided to open the stock entry in form view from the item dashboard. It takes much longer to move items between locations.

## Proposal
Retain version-13 behavior when Create Stock Entry is clicked, and new develop behavior of opening the stock entry in form view when edit is clicked.  

image

image


ankush:
> Currently item dashboard allows creating inventory out of thin air with 2 clicks 🪄 (against stock adjustment account) which is often not desired by many (with exception of devs who just want to quickly test things by adding arbitrary qty and rate)

> This also bypasses workflows set up by organizations for approval of such stock transactions.


Two reasons why it was removed. Maybe we can keep previous behavior for stock movement, but adding stock randomly is only done by devs or for testing. It's a bad practice and shouldn't be done without even letting user know what is being done (a stock entry)

dj12djdjs:
@ankush This makes sense. I'll work on making this compliant.

dj12djdjs:
The more I look into making this workflow compliant, the more I dread this. Since I don't use workflows in any of my implementations I just don't understand all the mechanics. I'm just going to override the item dashboard in a custom app.

If enough people are also interested in bringing back the quick stock entry feature then it might be worth it. However, right now I haven't seen any complaints about it other than 1 organization. Therefor, I'm going to close this PR.

PR #31631, opened 2022-07-19
fix: slow stock reposting

rohitwaghchaure:
1) Slow Stock Reposting

Recently we have made [changes](https://github.com/frappe/erpnext/pull/31519) which would enqueue the same item with different warehouse so that system won't throw the timeout error. This changes causing the slow reposting issue.

To fix this issue, before enqueue check whether the same item and warehouse exists in the queue or not and if it's exists then check the posting date. If the posting date is less than the current posting date then update the posting date.

```
def update_distinct_item_warehouses(self, dependant_sle):
		key = (dependant_sle.item_code, dependant_sle.warehouse)
		val = frappe._dict({"sle": dependant_sle})
		if key not in self.distinct_item_warehouses:
			self.distinct_item_warehouses[key] = val
			self.new_items_found = True
		else:
			existing_sle_posting_date = (
				self.distinct_item_warehouses[key].get("sle", {}).get("posting_date")
			)
			if getdate(dependant_sle.posting_date) < getdate(existing_sle_posting_date):
				val.sle_changed = True
				self.distinct_item_warehouses[key] = val
				self.new_items_found = True
```

2) Item and Warehouse reposting

In case of item and warehouse based reposting, system partially repost with status as **Completed**. The cause for this issue is if because of timeout the system has reposted Repost Item Valuation partially and when the reposting runs again system failed to pick the correct items which has to be repost.

Screenshot 2022-07-19 at 12 59 14 PM

### DIFF
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -209,8 +209,12 @@ def repost_future_sle(
 	via_landed_cost_voucher=False,
 	doc=None,
 ):
-	if not args and voucher_type and voucher_no:
-		args = get_items_to_be_repost(voucher_type, voucher_no, doc)
+
+	items_to_be_repost = get_items_to_be_repost(
+		voucher_type=voucher_type, voucher_no=voucher_no, doc=doc
+	)
+	if items_to_be_repost:
+		args = items_to_be_repost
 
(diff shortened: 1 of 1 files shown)


rohitwaghchaure:
@Mergify backport version-13-hotfix

PR #31637, opened 2022-07-20
fix: Removed 'Allow Monthly Depreciation' checkbox

nabinhait:
### Fix 1
Remove **Allow Monthly Depreciation** checkbox from Asset for the following reasons:
- User can set depreciation frequency as 1 for monthly depreciation
- It sets the wrong schedule in case it is checked but the frequency is not set as 1

** Added patch to correct number and frequency of depreciations

### Fix 2
In the case of the WDV method, earlier the rate of depreciation was always set as an integer. Now it is set as a decimal number with precision 2.


nabinhait:

aakvatech:
This will resolve errors in the past caused by monthly depreciation.

What would the users use if they want calculation of WDV depreciation on yearly basis, but divide that number by 12 and pass monthly depreciations?

PR #31639, opened 2022-07-20
fix: assign duplicate_items_msg outside conditional

maharshivpatel:
duplicate_items_msg was defined inside if the statement  of stock_item so when duplicate was found in non_stock_item it raised referenced before assignment error.


ankush:
@maharshivpatel send all PRs to develop branch always (unless it's not relevant on develop) and then backport. 

https://frappe.io/handbook/release_process#branching-convention
https://frappe.io/handbook/frappe-git-tips#branching-guideline-at-frappe

ankush (approved):

PR #31652, opened 2022-07-20
Fix Material Request from Sales Order

MaxMorais:
When we create an Material Request from a Sales Order, ERPNext didn't take into consideration if the item has stock or not

https://discuss.erpnext.com/t/material-request-from-sales-order-urgent/12404

> Screenshots/GIFs


rohitwaghchaure:
@MaxMorais can you make it configurable?

MaxMorais:
@rohitwaghchaure Sure, I was thinking about this!

I think we should be able to select the type of Material Request we wanna make, and have different behaviors!

Cuz, for Stock Transfer or Buy, make sense be the lower items
But for Stock Issue, Manufacturing etc, need to be the whole SO

PR #31659, opened 2022-07-21
fix: ensure defaults removed in bad frappe patch get set again

sagarvora:


ankush (approved):

ankush:
Reference: https://github.com/frappe/frappe/pull/17575

PR #31676, opened 2022-07-22
fix: enable tax withholding checkbox in PI with supplier_tds

ruthra-kumar:
## Issue
1. Create PO without applying **tax_withholding_amount**, whose supplier has default tax withholding category.
2. Create PI from [1].
'Apply tax withholding amount' option is disabled in PI from [2].

## Fix
Add supplier_tds on the onload attribute of PI if the supplier has one.


deepeshgarg007:

deepeshgarg007:

PR #31686, opened 2022-07-22
fix: refer precision if given outstanding amount field

kunhimohamed:

Refer the precision for outstanding_amount field in "Sales Invoice", "Purchase Invoice" or "Fees" Doctypes. when update from gl_entry.


kunhimohamed:
it's already mentioned in set_status function.

PR #31690, opened 2022-07-25
fix: misc payments fixes

phot0n:
Since we're splitting payments app from frappe, updating
ERPNext's stuff dependant on that as well

Added payments as a required app for erpnext - as erpnext has some doctypes which are dependant on it (like Payment Request, Payment Gateway Account, etc) and other apps dependant on erpnext might also use it.

### DIFF
--- a/.github/workflows/patch.yml
+++ b/.github/workflows/patch.yml
@@ -105,7 +105,6 @@ jobs:
               echo "Updating to v$version"
               branch_name="version-$version-hotfix"
 
-
               git -C "apps/frappe" fetch --depth 1 upstream $branch_name:$branch_name
               git -C "apps/erpnext" fetch --depth 1 upstream $branch_name:$branch_name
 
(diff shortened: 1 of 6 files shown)


surajshetty3416:
Unrelated file linter is failing. Ignoring.

surajshetty3416 (approved):

.github/workflows/patch.yml
```diff
@@ -127,6 +127,8 @@ jobs:
           pyenv global $(pyenv versions | grep '3.10')
           rm -rf ~/frappe-bench/env
           bench -v setup env
+          bench pip install -e ./apps/payments
           bench pip install -e ./apps/erpnext
 
           bench --site test_site migrate
+          bench --site test_site install-app payments
```
phot0n, line 134: wondering if we even need this here 😅 

side thinking: we could once try to move this line to L122 - could probably work & might be more accurate 🤔

phot0n: > side thinking: we could once try to move this line to L122 - could probably work & might be more accurate 🤔

Nope doesn't work ..the file import payments app is using doesn't exist in v13 (since it had undergone a major refactor in v14)

PR #31699, opened 2022-07-26
fix: hero image not loading in portal homepage

ruchamahabal:
## Before:

Hero Image setup in the Homepage is not loading: 

image

image

## After:

bg

### DIFF
--- a/erpnext/templates/pages/home.css
+++ b/erpnext/templates/pages/home.css
@@ -1,7 +1,6 @@
 /* csslint ignore:start */
 {% if homepage.hero_image %}
 .hero-image {
-	background-image: url("{{ homepage.hero_image }}");
 	background-size: cover;
 	padding: 10rem 0;
 }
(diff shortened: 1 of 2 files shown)


paulstary:
I am experiencing this exact same problem with v14.0.0-beta.6 running on frappe.cloud (hosted). It appears the fix has been incorporated into v13 so assume this would apply to later versions as well? 

Perhaps this still does not work on a hosted instance? Any suggestions why this is still not working for me?

ruchamahabal:
@paulstary while uploading the hero image check if it's private. Framework by default now marks every attachment as private. Set it to public and try

paulstary:
Thank you for your quick response. I'm sure the image is public. I have tried both ways just to make sure. I have tried uploaded images and hosted images but none ever show on the Home page.

The .hero-image style shows the placeholder for the image but does not load the image.

image

Here is my Homepage form. I am using the Hero Section Based on: **Default**. Is this correct?

image

I have saved and reload the page many time. You can see the image is good and opens at <https://rayva.dev/img/Strings_Theme.jpg>.

image

Here is the Website with the Inspector open.

image

... and after manually inserting the image link into the url( ) of the .hero-image style.

image

It appears that the {{ homepage.hero_image }} placeholder is never set to the image URL by the form.

PR #31741, opened 2022-07-30
fix: check item_code in all rows of po_items

s-aga-r:
Source / Reference: ISS-22-23-01552

There can be rows which do not have item-code selected in po-items. Check all rows for item-code in po_items before fetching the sub-assembly items, throw an error if not available.

Screenshot 2022-07-30 at 10 18 23 AM


erpnext/manufacturing/doctype/production_plan/production_plan.json
```diff
@@ -358,7 +358,7 @@
    "options": "Production Plan Sub Assembly Item"
   },
   {
-   "depends_on": "eval:doc.po_items && doc.po_items.length && doc.docstatus == 0",
+   "depends_on": "eval:doc.po_items && doc.po_items[0].item_code && doc.docstatus == 0",
```
rohitwaghchaure: why item code not added by the User in po_items table?

rohitwaghchaure: Add condition for sub_assembly_items

s-aga-r: > why item code not added by the User in po_items table?

By default, we are provided with an empty row (item_code not selected) in **po_items** while creating a new Production Plan. The current condition applied on **Get Sub Assembly Items** button is **po_items && po_items.length**, which is always true when we create a new Production Plan, which means the **Get Sub Assembly Items** is visible. If there is no **po_item(s)** and the user tries to get sub-assembly items by clicking on **Get Sub Assembly Items**, an error will be thrown. Ideally, the **Get Sub Assembly Items** should be hidden if there are no **po_items**.

We have different ways to do the same:
1. Check the tem_code in the 0th index. (This PR)
2. Checking each row in JS and hiding the button accordingly.

s-aga-r: > Add condition for sub_assembly_items

Want to hide the **Get Sub Assembly Items** button, based on **po_items**.

s-aga-r: Description Image: In this case, if we click on "Get Sub Assembly Items" an error will be thrown saying that **'NoneType' object is not iterable"**

erpnext/manufacturing/doctype/production_plan/production_plan.py
```diff
@@ -639,6 +639,9 @@ def get_sub_assembly_items(self, manufacturing_type=None):
 		sub_assembly_items_store = []  # temporary store to process all subassembly items
 
 		for row in self.po_items:
+			if not row.item_code:
+				frappe.throw(_(f"Row #{row.idx}: Please select Item Code in Assembly Items"))
```
```suggestion
				frappe.throw(_("Row #{0}: Please select Item Code in Assembly Items").format(row.idx))
```
F-strings don't work with translations.

ankush (changes requested):

ankush (approved):

PR #31750, opened 2022-08-01
fix: minor URL link

rohitwaghchaure:
Incorrect RFQ URL link sent to supplier due to which supplier wouldn't able to open the RFQ page on web portal


rohitwaghchaure:

PR #31761, opened 2022-08-02
fix: specify allowed doctype in queries

sagarvora:


ankush:
Most queries have tests afaik.

ankush (approved):

PR #31779, opened 2022-08-04
Bug add accouting dimension in asset repair

ruthra-kumar:
## Issue
When asset repair cost is capitalized, GL entries are posted. In such case, validation error is thrown for accounting dimension that are mandatory for P/L accounts. There is no provision to enter the new dimensions in the **Accounting Dimensions** section.

## Fix
Include **Asset Repair** in dimension creation hook.


deepeshgarg007:

PR #31780, opened 2022-08-04
Fix: Loan pending principal amount 

abhinavxd:
- Use the common function to get the pending principal amount in `Loan Write Off`.
- Changes in Loan Repayment & Loan Disbursement doctypes to allow change of repayment & disbursement accounts while creating a new doc.


deepeshgarg007:

PR #31799, opened 2022-08-08
fix: process loan interest accrual

abhinavxd:
- Terms loans interest would be accrued twice when process loan interest accrual is created without selecting a loan & loan type and only posting date is selected.


deepeshgarg007:

PR #31804, opened 2022-08-09
fix: incorrect incoming rate set for inter transfer purchase receipt

rohitwaghchaure:
**Issue**

1. Make Intra Company transfer transaction for Item A with valuation rate as 100
2. Make Delivery Note and Purchase Receipt 
3. Create backdated purchase entry for Item A with incoming rate as 200 in a way that the previous delivery note's valuation rate should change to new rate as 200
4. System update the valuation rate to 200 for the previous delivery note
5. But System not update the valuation rate in the Purchase Receipt which causing incorrect stock and account value


rohitwaghchaure:

PR #31806, opened 2022-08-09
ci: ensure unique group ID to prevent workflows from getting cancelled

sagarvora:
- Use `event_name` to differentiate between `push`, `workflow_dispatch` and `pull_request`
- Fallback to `run_id` (only for `workflow_dispatch` event)

References:
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
- https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

### DIFF
--- a/.github/workflows/patch.yml
+++ b/.github/workflows/patch.yml
@@ -11,7 +11,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: patch-develop-${{ github.event.number }}
+  group: patch-develop-${{ github.event_name }}-${{ github.event.number || github.event_name == 'workflow_dispatch' && github.run_id || '' }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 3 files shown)


ankush (dismissed):

ankush:
`run_id` fallback is problematic sorta? We don't wanna keep running on branches all the time. If new changes are pushed last one should be killed. 😅 

For PRs it won't matter.

sagarvora:
> * Fallback to `run_id`

With this, each `push` event will be treated as a separate group. Is this the intended behavior or should one push event cancel another?

sagarvora:
> `run_id` fallback is problematic sorta? We don't wanna keep running on branches all the time. If new changes are pushed last one should be killed. sweat_smile

was just confirming this on TG 😅

ankush:
> Is this the intended behavior or should one push event cancel another?


New pushes should cancel previous builds. It's because of practical constraints, we only have 20 concurrent builds for all repos on this org.

PR #31816, opened 2022-08-09
fix: limited options for no-of-employees in the crm documents

nabinhait:
Limited options for no-of-employees in the Lead, Opportunity and Prospect documents.

Old options:
```
1-10
11-20
21-30
31-100
11-50
51-200
201-500
101-500
500-1000
501-1000
>1000
1000+
```

New Options:
```
1-10
11-50
51-200
201-500
501-1000
1000+
```


nabinhait:

PR #31817, opened 2022-08-09
fix: limit pos recent order page result

ruthra-kumar:
## Issue
POS Recent Order page crashes if the no of invoice grows larger in size.

Screenshot 2022-08-09 at 8 35 12 PM

Screenshot 2022-08-09 at 8 36 00 PM

## Fix
Query result should be limited by default.


ruthra-kumar:

PR #31822, opened 2022-08-10
fix(pos): error while consolidating pos invoices

nextchamp-saqib:
In some rare cases, while merging multiple POS Invoices into a Sales Invoice, the **sum** of the Grand Total of all the POS Invoices is not equal to the **sum** of the Paid Amount
The difference is usually minimal, like 1.2 or 0.5
This difference is then considered as outstanding which then leads to an unpaid invoice, or in the case of returns stops the consolidation

This PR adds a new field to POS Profile to set a Write-Off Limit, to automatically write off such fractional amounts while consolidating


deepeshgarg007:

PR #31830, opened 2022-08-11
fix: Unable to make payment entry against Fees using education app

deepeshgarg007:
Users are unable to create a payment entry against Fees doctype using the Education app


deepeshgarg007:

PR #31860, opened 2022-08-17
fix: delete custom fields on deletion of inventory dimension

rohitwaghchaure:
- Delete custom fields on deletion of the inventory dimension
- Removed rename functionality for the inventory dimension
- Set value from the parent form (Using fieldname or doctype name)
image


rohitwaghchaure:
@mergifyio backport version-14-hotfix

PR #31863, opened 2022-08-17
fix: not able to issue expired batches

rohitwaghchaure:
While issuing expired batches system throwing an error
Screenshot 2022-08-17 at 12 20 27 PM


rohitwaghchaure:
@mergifyio backport version-13-hotfix  version-14-hotfix

PR #31894, opened 2022-08-19
fix: TDS calculation for advance payment

maharshivpatel:
```
"against_voucher": ["is", "not set"]
```
was used in query due to which if TDS was added on **advance** payment vouchers and then reconciled against purchase invoice. it will not find those vouchers and consider it as first-time threshold ( which is incorrect ) due to which it will calculate Tax for all transactions.


deepeshgarg007:

PR #31898, opened 2022-08-19
Fix upload invoice suppliers

balabofficial:
fix: bug uploading supplier invoice in regional IT, when upload invoice quantity fields are wrong.
fixed moving initial vars in def prepare_items_for_invoice() inside loop for reinit every time.


deepeshgarg007:
@balabofficial Please rebase your PR and then reopen

PR #31907, opened 2022-08-21
feat(hr): multiple shift v13

alaa-alsalehi:
We need to add multiple shifts to v13 we copy these modifications from [this PR](https://github.com/frappe/erpnext/pull/29955) we fix different issues in dependencies, tests and python old version and be sure the test cases work fine
this is the same description of the original PR
### What Changes?
**Schedule multiple shifts**

- Users can make multiple non-overlapping Shift Assignments for the same employee on the same day

### Shift Request

- Users can make multiple non-overlapping Shift Requests for the same employee on the same day

### Fetching shift in Employee Check In

- When employee check-ins are logged via biometric or manually, the exact shift based on the timestamp is fetched. Before, shifts used to be fetched based on date. Now they are timestamp specific.
- Handles shifts spanning over 2 different days (eg: 23:00 - 2:00), midnight shifts.
- Shift fetching used to fail on timing boundaries, this is fixed.
- Handles 2 consecutive shifts if they are overlapping within grace periods.

### Attendance

- Earlier, only 1 attendance record could be marked for 1 day.
- Now multiple attendance records can be marked for each shift. Eg: an employee can be absent for one shift, present for another on the same day
- Auto Attendance now, takes care of marking shift-specific attendance and not for the whole day
- Earlier, employees with no attendance records on a specific date used to be marked absent via Auto Attendance. Now, this also checks for shift-wise attendance records along with the date. So, only if attendance is not marked for a particular shift/for the whole day, the employee will be marked absent.

### Attendance Reports

- Added shift-wise attendance against employees
- Added legend for status abbr with full form and color.
- Formatter for data to quickly identify cells with lesser cognitive load.
- Refactored Monthly Attendance Sheet code. Can still be better but will leave it for the next iteration.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -8,15 +8,24 @@
 #
 # $ git config blame.ignoreRevsFile .git-blame-ignore-revs
 
+# Replace use of Class.extend with native JS class
+1fe891b287a1b3f225d29ee3d07e7b1824aba9e7
+
 # This commit just changes spaces to tabs for indentation in some files
 5f473611bd6ed57703716244a054d3fb5ba9cd23
 
-# Whitespace trimming throughout codebase
-9bb69e711a5da43aaf8c8ecb5601aeffd89dbe5a
-f0bcb753fb7ebbb64bb0d6906d431d002f0f7d8f
+# Whitespace fix throughout codebase
+4551d7d6029b6f587f6c99d4f8df5519241c6a86
+b147b85e6ac19a9220cd1e2958a6ebd99373283a
 
-# imports cleanup
-4b2be2999f2203493b49bf74c5b440d49e38b5e3
+# sort and cleanup imports
+915b34391c2066dfc83e60a5813c5a877cebe7ac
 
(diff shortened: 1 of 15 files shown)


alaa-alsalehi:
@ruchamahabal if we rebase it to hotfix there is a chance to accept it if yes we will do it

ruchamahabal:
@ServeMe Since v14 has been released as a major version, we won't be adding or backporting any features to v13. Only fixes will go in for v13. This feature was specifically developed for v14 only.

alaa-alsalehi:
@ruchamahabal I am aware of this rule but we have a need to add it for this release. Any suggestion on how we can do that without breaking the cycle?
we currently have a fork of ERPNext and we merge on it any other suggestions?

ruchamahabal:
@ServeMe Yes, you can either add this feature to your fork or simply upgrade to v14 (better in terms of future maintenance) after assessing the breaking changes

https://github.com/frappe/erpnext/wiki/Migration-Guide-to-ERPNext-version-14

PR #31912, opened 2022-08-21
fix: departments tree creation

zaqoutabed:
departments tree creation when create new company

> When Create new company and the user language not English, will accord error when create departments tree and the tree will not be created.

Screenshot from 2022-08-21 16-28-28

### DIFF
--- a/erpnext/setup/doctype/company/company.py
+++ b/erpnext/setup/doctype/company/company.py
@@ -223,102 +223,65 @@ class Company(NestedSet):
 		)
 
 	def create_default_departments(self):
-		records = [
-			# Department
+		departments_list = [
 			{
-				"doctype": "Department",
 				"department_name": _("All Departments"),
 				"is_group": 1,
-				"parent_department": "",
-				"__condition": lambda: not frappe.db.exists("Department", _("All Departments")),
 			},
 			{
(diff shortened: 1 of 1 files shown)


erpnext/setup/doctype/company/company.py
```diff
@@ -223,102 +223,65 @@ def create_default_accounts(self):
...
-		rebuild_tree("Department", "parent_department")
+		for dep_record in departments_list:
+			if not frappe.db.exists("Department", "{0} - {1}".format(dep_record["department_name"], self.abbr)):
+				department = frappe.get_doc(
+					{
+						"doctype": "Department",
+						"department_name": dep_record["department_name"],
+						"is_group": dep_record.get("is_group", 0),
+						"company": self.name,
+						"parent_department": "{0} - {1}".format(_("All Departments"), self.abbr) if not dep_record.get("is_group", 0) else "",
```
deepeshgarg007, line 279: All Departments won't have abbreviation at the end, all test cases failing because of it. Please fix and reopen

zaqoutabed: @deepeshgarg007 
I think abbreviations are required !
because each company has its own departments

Issue #29183, opened 2022-01-07
perf: fast af stock balance report


ankush:
stock balance report serializes all SLEs and then applies stock ledger logic on it to transform the whole history of item-warehouse into a single line result for the final balance.

Proposal: move all this computation to raw SQL(qb). It's easily 100-1000x faster on large databases. 

PS: Some things like ageing data might be tricky but the straightforward computation should be moved to raw SQL.

ankush:
another instance of problem: https://github.com/frappe/erpnext/issues/31528 

> With almost 1.1 million SLEs the background job `run_background` prematurely terminates Seems like its a memory issue.
> 
> SLEs could be queried and looped over in batches. Build the resultant data incrementally <img alt="Screenshot 2022-07-05 at 7 10 58 PM" width="1397" src="https://user-images.githubusercontent.com/25857446/177343954-2037300f-1b6e-4d52-a1c5-d299ef30f0e9.png">
> 
> Or bump up the memory

PR #31925, opened 2022-08-22
fix: consider from date in SLE query

rtdany10:
In cases where from date isn't provided, the system throws error, but it doesn't apply the from date in the query if provided.


rtdany10:
@s-aga-r please check

s-aga-r (approved):

s-aga-r:
> @s-aga-r please check

@rtdany10 Add a test case for the same.

s-aga-r (changes requested):
Add a test case for the same.

rtdany10:
@s-aga-r I'm not sure why the existing test case is failing. Any pointers?

s-aga-r:
> @s-aga-r I'm not sure why the existing test case is failing. Any pointers?

@rtdany10 not sure, but it can be because tests have from_filter and now it's applying in the query?

rtdany10:
> > @s-aga-r I'm not sure why the existing test case is failing. Any pointers?
> 
> @rtdany10 not sure, but it can be because tests have from_filter and now it's applying in the query?

@s-aga-r  I'm not sure, cause there are no hard-coded values, comparison occurs between SLE values and report values. I'll try digging deeper.

s-aga-r:
> > > @s-aga-r I'm not sure why the existing test case is failing. Any pointers?
> > 
> > 
> > @rtdany10 not sure, but it can be because tests have from_filter and now it's applying in the query?
> 
> @s-aga-r I'm not sure, cause there are no hard-coded values, comparison occurs between SLE values and report values. I'll try digging deeper.

Hi, @rtdany10 any update on this? Got a Support Ticket related to this.

erpnext/stock/report/stock_balance/test_stock_balance.py
```diff
@@ -53,9 +52,11 @@ def assertInvariants(self, rows):
 						ORDER BY timestamp(posting_date, posting_time) desc, creation desc)
 						AS rn
 					FROM `tabStock Ledger Entry`
-					where is_cancelled=0
+					where posting_date >= %s and posting_date <= %s and is_cancelled=0
```
ankush, line 55: Never modify test code to them like to pass other tests! 

It is encoding a known supported business use case. 😅

erpnext/stock/report/stock_balance/stock_balance.py
```diff
@@ -259,7 +259,9 @@ def apply_conditions(query, filters):
 	sle = frappe.qb.DocType("Stock Ledger Entry")
 	warehouse_table = frappe.qb.DocType("Warehouse")
 
-	if not filters.get("from_date"):
+	if from_date := filters.get("from_date"):
+		query = query.where(sle.posting_date >= from_date)
```
ankush, line 263: This is incorrect. Stock balance report has to fetch all SLEs, the from date is only used for computing opening qty. 

This is somewhat inefficient but that's how it's implemented. It's _correct_ at least.

ankush (changes requested):

erpnext/stock/report/stock_balance/stock_balance.py
ankush: @s-aga-r

ankush: also refer https://github.com/frappe/erpnext/issues/29183

PR #31942, opened 2022-08-24
fix: include payment against PO in AR/AP report

ruthra-kumar:
Payments against Purchase Order/Sales Order should be included in Accounts Receivable/Payable report under the same Voucher no as outstanding amount.

Before:
before_ar_rpt

After:
after_ar_rpt


ruthra-kumar:

PR #31949, opened 2022-08-24
fix: Scrap Asset Accounting Dimensions

maharshivpatel:
While scrapping assets, accounting dimensions are not passed to the journal entry, so I added them to fix the issue.


deepeshgarg007:

muchai:
Somewhat related, please add dimensions on asset capitalization too.

PR #31950, opened 2022-08-24
fix: display amount in account currency if party is supplied

ruthra-kumar:
Accounts Receivable/Payable report should display amount in account currency, if party is supplied.
Company Currency:
Screenshot 2022-08-25 at 10 09 32 AM
Account curreny:
Screenshot 2022-08-25 at 10 09 22 AM


deepeshgarg007:

PR #31951, opened 2022-08-24
fix: Purposes not set in Maintenance Visit

rohitwaghchaure:
While making Maintenance Visit from Maintenance Schedule the purposes table has not set any value

1) Go to Maintenance Visit DocType
2) Select any series and then Select customer as "ABC"
3) Click on the "Get Items From" and select "Maintenance Schedule".
4) Select the Maintenance schedule Document
5) Purposes table showing blank


rohitwaghchaure:

PR #31954, opened 2022-08-24
feat: better Item Price list view

FHenry:
Users don't need technical ID of Item Price in list, but they need,by default, the Price List of the Item Price

Before
image

After

image

`no-doc`


deepeshgarg007:
@FHenry Can you please add a screenshot of how does it look post the change

FHenry:
> @FHenry Can you please add a screenshot of how does it look post the change

@deepeshgarg007 Hi, sure, it's done. You'r right it's better to show difference in screenshots, I'll do that in next PR about changing default list settings. I'm not sure that ID list filter is needed, what do you think ?

FHenry:
@deepeshgarg007 I added the reference column by default as it's supplier or customer of item price, it can be very useful as well. 
I also check that ID standard filter is not removable (cannot find how..., sorry) , we have to let it here, even if it's not usefull for users

deepeshgarg007 (approved):

FHenry:

FHenry:

PR #31966, opened 2022-08-25
fix: default supplier not set in the PP

rohitwaghchaure:
In the production plan -> Under sub assembly table -> Default Supplier has not set from Item master in case of Subcontracting item


rohitwaghchaure:

PR #31967, opened 2022-08-25
fix: material request connection on work order

rohitwaghchaure:
Added material request connection on work order

Screenshot 2022-08-25 at 12 10 56 PM


rohitwaghchaure:
https://github.com/Mergifyio backport version-14-hotfix

PR #31978, opened 2022-08-25
emergency hotfix: change in Python Error handling

RJPvT:
suddenly non existence of ["catagory"] creates an NoneType error in pyhton (change in handling python 3.5 ?). since tags are not strictly needed all execptions should pass.

### DIFF
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
@@ -226,7 +226,7 @@ def new_bank_transaction(transaction):
 	try:
 		tags += transaction["category"]
 		tags += ["Plaid Cat. {}".format(transaction["category_id"])]
-	except KeyError:
+	except Exception:
 		pass
 
 	if not frappe.db.exists("Bank Transaction", dict(transaction_id=transaction["transaction_id"])):
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
@@ -226,7 +226,7 @@ def new_bank_transaction(transaction):
 	try:
 		tags += transaction["category"]
 		tags += ["Plaid Cat. {}".format(transaction["category_id"])]
-	except (KeyError, NoneType):
+	except (KeyError, TypeError):
 		pass
 
 	if not frappe.db.exists("Bank Transaction", dict(transaction_id=transaction["transaction_id"])):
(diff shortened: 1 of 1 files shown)


RJPvT:
@surajshetty3416 could you check ? stopped working for me since last update, so far as i can see no changes in ERPNEXT happened in the codebase fot this function

RJPvT:
addition after more research.

as of feb this year Plaid had been making changes to their catagory system : https://plaid.com/blog/transactions-categorization-taxonomy/ which means that some catagories can return None as they are changed. This produces the error. I still think as of this that all exceptions should pass. catagory (tags) is optional and not important enough to warrant holding error.

erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
```diff
@@ -226,7 +226,7 @@ def new_bank_transaction(transaction):
 	try:
 		tags += transaction["category"]
 		tags += ["Plaid Cat. {}".format(transaction["category_id"])]
-	except KeyError:
+	except Exception:
```
deepeshgarg007: Instead of simply ignoring all the exceptions can you just add the one that you got and can be ignored

deepeshgarg007 (changes requested):

erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
RJPvT: I have made the change. still, because the tags are not mandatory i question the importance to make this a failing error. i would think the creation of the document is more important than the labels in this case. 
perhaps make the other exceptions an assignment to the admin but not failing creation ?

erpnext/erpnext_integrations/doctype/plaid_settings/plaid_settings.py
```diff
@@ -226,7 +226,7 @@ def new_bank_transaction(transaction):
 	try:
 		tags += transaction["category"]
 		tags += ["Plaid Cat. {}".format(transaction["category_id"])]
-	except KeyError:
+	except (KeyError, TypeError):
 		pass
```
Alchez, line 230: I don't think we need a `try-except` here

```suggestion
	if transaction.get("category"):
		tags.extend(transaction.get("category"))
	if transaction.get("category_id"):
		tags.append("Plaid Cat. {}".format(transaction.get("category_id")))
```

RJPvT: Hi Alchez. 

Fine by me. 

@deepeshgarg007 do youagree?

RJPvT:
@deepeshgarg007 could you check the review and final code ?

it is a breaking error for me so would like it pushed a.s.a.p.

PR #31980, opened 2022-08-25
refactor: use redis-py instead of redisearch

### LINKED ISSUE #31300: Use Redis-py client over deprecated redisearch
### Information about bug

Frappe `develop` uses Redis 3.5.3. I tried upgrading it to use 4.3.3 (latest) but couldn't. Essentially we can't upgrade the Redis client because the redisearch library was deprecated and merged into redis-py but the APIs are not added as is - equivalent APIs exist which means we'll have to put slightly more time to do the upgrade.

### Module

ecommerce

### Version

Frappe develop
ERPNext develop

### Installation method

_No response_

_No response_

ankush:
closes https://github.com/frappe/erpnext/issues/31300 

depends on https://github.com/frappe/frappe/pull/17921


akhilnarang:
@mergifyio backport version-14-hotfix

PR #31988, opened 2022-08-26
chore: remove precision on discount_percentage of Sales Invoice Item

ruthra-kumar:
## Issue
Discount percentages of 99.999% gets rounded up to 100% effectively making the item rate 0. Caused by precision '2'.

## Fix
Remove precision on discount_percentage field of Sales Invoice Item and use system default '3'.


deepeshgarg007:

PR #31996, opened 2022-08-26
feat: don't copy item title to description

barredterra:
We used to copy the item name into the description, if the description was empty.

Why? It doesn't add information. What if the item name says enough and doesn't need a description?

Todo:

- [ ] make description optional in transaction items

> no-docs

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -18,7 +18,6 @@ from frappe.utils import (
 	now_datetime,
 	nowtime,
 	strip,
-	strip_html,
 )
 from frappe.utils.html_utils import clean_html
 
@@ -82,9 +81,6 @@ class Item(Document):
 	def validate(self):
 		if not self.item_name:
 			self.item_name = self.item_code
-
-		if not strip_html(cstr(self.description)).strip():
-			self.description = self.item_name
 
 		self.validate_uom()
 		self.validate_description()


sagarvora:
> Why?

I think a lot of user code relies on description - print formats and what not. It's better to have _something_ there in case the description hasn't been manually updated.

Changing the Item DocType alone could be fine I think, as long as the Item Name still gets copied as description in transactions as a fallback.

barredterra:
@sagarvora why would the item description be mandatory in a transaction – is this a legal requirement somewhere? Print formats can handle missing values pretty well, they work just as fine for optional fields.

sagarvora:
> is this a legal requirement somewhere?

It's not, AFAIK.



> Print formats can handle missing values pretty well, they work just as fine for optional fields.

You're missing the point. People expect some value to show up in their invoice for the Item. Most users currently have description showing up in their print formats - they will now see... nothing (if the description isn't set). So they would have to do something like `{{ item.description or item.item_name }}` in all their print formats. Except, even that won't work. If the description has `ql-editor` HTML inside it but no content, it will still get chosen over `item_name`.

Changes like the one you're proposing are good ideally, but are generally not preferred to avoid compatibility issues.

barredterra:
> You're missing the point.

Correct, and I still don't get it 😄

> People expect some value to show up in their invoice for the Item.

Yes. That value will be Item Code and Item Name, which are rightfully mandatory.

Now assume you have some items for which a description doesn't exist. Why force users to make up a description or print the item name a second time? This seems pretty arbitrary to me.

No need to do `{{ item.description or item.item_name }}`, you can just do `{{ item.item_name }} {{ item.description }}`. If the description is empty, it's empty, the name will always be there.

sagarvora:
> No need to do `{{ item.description or item.item_name }}`, you can just do `{{ item.item_name }} {{ item.description }}`. If the description is empty, it's empty, the name will always be there.

Ok, lets assume this is the new implementation:

An existing user X has about 3000 items in her ERP. For a majority of these items, her item description is the same as item name (it's been the default behavior for a long time!). Now when she creates a new item in her ERP, if she implements the solution you're proposing, she will have to either:
- Change all existing items so that they doesn't have a description, or
- Be happy with item name showing up twice in all her transactions. (I doubt she will be.)

So basically, you're forcing users to update all their existing items. One of my sites has 4,50,000+ items!

---

With that being said, I understand your use-case. You might want to sync this data with an e-commerce site where it isn't the norm for the item name to be a part of the description. We can make this feature configurable and let users decide what they want.

barredterra:
Okay, I guess we're assuming different default behaviors: My assumed default behavior is that the user prints item name and description, with the description being different from the name. Your assumed default behavior is that the user prints only the description, which already includes the item name?

> Be happy with item name showing up twice in all her transactions. (I doubt she will be.)

That's exactly the problem I'm trying to solve here 😄. If we print the fields Item Name and Description, the item name will show up twice, if it doesn't have a different description.

However, I understand your concern regarding the impact on existing systems.

barredterra:
@sagarvora how about we keep the copying but make the description optional? This way users can at least remove the description, if they don't need it.

sagarvora:
> how about we keep the copying but make the description optional

- We can keep the copying configurable, IMO. Users who don't want it can toggle it in **Stock Settings** or some other single DocType.
- Ideally the copying should happen in `before_insert` instead of `validate`. If a user removes the description manually, it can be annoying to find it back on save.
- Keeping the description optional in transactions _should be okay_ - we'd need to check usage in ERPNext and dependent apps to be sure.

ankush:
I have thought about removing this long time ago too, we didn't because of print formats 😄 

@rohitwaghchaure can also confirm.

barredterra:
For now we'll be using a workaround in our print formats:

- Enter "-" as description, if it should be empty
- `{% if (doc.description | striptags) != "-" %}{{ doc.description }}{% endif %}`

PR #32005, opened 2022-08-29
fix: lost quotation not to expired

niyazrazak:
photo_2022-08-29_14-48-45

Analyse the image you can see the user is already marked this quotation as lost. After the expire date of quotation the quotation is changed into expired and user need ato again mark the quitation as lost. 

If it is not okay. Please test from your side
@ankush


deepeshgarg007:

PR #32026, opened 2022-08-30
fix: force delete old report docs

ankush:
telegram-cloud-photo-size-5-6084706599928180794-y


faisal-zahoor:
hy @ankush ,
you deleted reports but these are still used in the make_demo function, can you guide me to any alternate reports of these old reports?

PR #32030, opened 2022-08-30
fix: Loan Interest accruals for 0 rated loans

deepeshgarg007:
Removed zero interest check introduced in RP https://github.com/frappe/erpnext/pull/31799

Front ported fixes from https://github.com/frappe/erpnext/pull/32011


deepeshgarg007:

Issue #32040, opened 2022-08-31
Journal Entry Template


Rgottsch1:
### Information about bug

If I create the first Journal Entry Template - it goes in fine - (see first screen shot)

![Journal Entry Template created fine](https://user-images.githubusercontent.com/29579531/187734840-67d21ea6-ef19-4b13-8573-a7e4ccb5b943.png)

If I then create another one - it doesn't populate the Series and one cannot choose a series from the drop down.  Only way around it is to reload and than I can access with a series listed, and need to click on reload again if need to make another eachtime.

![Journal Entry Template created fine](https://user-images.githubusercontent.com/29579531/187735645-23d14d44-403a-4408-a231-ab4e80290372.png)

![JOURNAL ENTRY ERROR](https://user-images.githubusercontent.com/29579531/187735327-c3d365c5-cfec-4af4-94c6-671aa7407f84.png)

### Module

accounts

### Version

Frappe Framework
Installed Apps
ERPNext: v13.36.3 (version-13)

Frappe Framework: v13.37.1 (version-13)

### Installation method

manual install

PR #32045, opened 2022-09-01
fix: Naming series in Journal Entry Template

Solufyin:
**Version:**

ERPNext: v14.0.3 (version-14)
Frappe Framework: v14.5.0 (version-14)
Payments: v0.0.1 (develop)

___

**Before:**

- When creating a new journal entry template the first time naming series appears but when duplicating the journal entry template or back to the journal entry template and again creating a new journal entry template the naming series does not appear. But when reloading the system and creating the journal entry template then will be naming series appear.

https://user-images.githubusercontent.com/34390782/187846470-0c94d74d-0338-4ca2-8c5e-be5b1d822181.mp4

**After:**
- After the form event set, when When creating a new journal entry, duplicating the entry, and back to the journal entry template and again creating a new journal entry template the naming series will appear.
- For cross-checking, also multi-currency wise and without multi-currency account show in Accounting Entries (Table).

https://user-images.githubusercontent.com/34390782/187847897-9ba79020-881b-42ca-9744-129cd35a05f7.mp4

___

Issue: #32040

It problem is also in version 13.

___

Thank You!

### DIFF
--- a/erpnext/accounts/doctype/journal_entry_template/journal_entry_template.js
+++ b/erpnext/accounts/doctype/journal_entry_template/journal_entry_template.js
@@ -2,7 +2,7 @@
 // For license information, please see license.txt
 
 frappe.ui.form.on("Journal Entry Template", {
-	setup: function(frm) {
+	refresh: function(frm) {
 		frappe.model.set_default_values(frm.doc);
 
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

PR #32049, opened 2022-09-01
Allow Item Templates in Work Order Items

djpalshikar:
Item Variants can be a part of the items tables, however, when BOM items are fetched to be a part of the Work Order items, item variants were being filtered out.
The filtering out does not serve a purpose. Having Item variants in BOMs allows for template like behaviour.

> Screenshots/GIFs


djpalshikar:
@rohitwaghchaure please review this. It will go a long way, trying to handle a very wide variety of BOM.

PR #32053, opened 2022-09-02
fix: KSA VAT report multi currency amount issue

hamzaali15:
In KSA VAT report amount is not showing correctly for multi currencies because net_amount field is fetched instead of base_net_amount

image
image

As you can see in picture base amount is **751.90 SAR** and base tax is **37.59 SAR** but when we show in report **200** in USD shows but VAT amount is **37.59 SAR** which is correct. I have attach the reference picture
image

I have amend the code and changed net_amount to base_net_amount. After changing the code it looks like
image


deepeshgarg007:

PR #32081, opened 2022-09-04
fix: allow statical component to be assigned with a company

ahmadRagheb:
fix for https://github.com/frappe/hrms/issues/134 
make sure we can assign statical components to the company. so we can use it in the system ( salary structure )


ahmadRagheb:
idk why ?? it's a bug please review it

ahmadRagheb:
@ruchamahabal

PR #32086, opened 2022-09-05
fix: QR Code multi currency issue

hamzaali15:
When try to scan qr code on app it is showing correct values for multi currencies because it is not getting base amount
image
image
As above images showing:
1. Grand Total is **105** in USD and **394.60** in SAR
2. Tax Amount is **5** in USD and **18.79** in SAR

whereas when try to scan QR code of invoice it is not fetching the base amount but using base currency symbol
image
Before changing in the code this QR code is showing Invoice Total: **105 SAR** & VAT Total: **5 SAR** while it should show **105 USD** or **394.60 SAR**
image
After amend the code it is showing **394.60 SAR** insted of **105 USD** which is correct

### DIFF
--- a/erpnext/regional/saudi_arabia/utils.py
+++ b/erpnext/regional/saudi_arabia/utils.py
@@ -87,7 +87,7 @@ def create_qr_code(doc, method=None):
 		tlv_array.append("".join([tag, length, value]))
 
 		# Invoice Amount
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@Mergifyio backport version-14-hotfix develop

Issue #31688, opened 2022-07-22
Bank Statement Import - Unable to import, throws an error 


nidhipurohit11:
### Information about bug

Bank Statement Import - When downloaded the template, edited it, saved it in the correct format, and tried to import then it's throwing an error - "Value for Import Log cannot be a list"

![Screenshot 2022-07-22 at 8 24 59 PM](https://user-images.githubusercontent.com/82337328/180468242-5c2d136d-f54b-4533-a410-218d099af768.png)

### Module

accounts

### Version

ERPNext: v14.0.0-dev (HEAD)
Frappe Framework: v14.0.0-dev (HEAD)

### Installation method

FrappeCloud

### Relevant log output / Stack trace / Full Error Message.

```shell
request.js:438 Traceback (most recent call last):
  File "apps/frappe/frappe/app.py", line 69, in application
    response = frappe.api.handle()
  File "apps/frappe/frappe/api.py", line 54, in handle
    return frappe.handler.handle()
  File "apps/frappe/frappe/handler.py", line 45, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/handler.py", line 83, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "apps/frappe/frappe/__init__.py", line 1579, in call
    return fn(*args, **newargs)
  File "apps/frappe/frappe/desk/form/save.py", line 22, in savedocs
    doc.save()
  File "apps/frappe/frappe/model/document.py", line 298, in save
    return self._save(*args, **kwargs)
  File "apps/frappe/frappe/model/document.py", line 335, in _save
    self._validate()
  File "apps/frappe/frappe/model/document.py", line 525, in _validate
    self._validate_length()
  File "apps/frappe/frappe/model/base_document.py", line 884, in _validate_length
    for fieldname, value in self.get_valid_dict(ignore_virtual=True).items():
  File "apps/frappe/frappe/model/base_document.py", line 328, in get_valid_dict
    frappe.throw(_("Value for {0} cannot be a list").format(_(df.label)))
  File "apps/frappe/frappe/__init__.py", line 521, in throw
    msgprint(
  File "apps/frappe/frappe/__init__.py", line 489, in msgprint
    _raise_exception()
  File "apps/frappe/frappe/__init__.py", line 441, in _raise_exception
    raise raise_exception(msg)
frappe.exceptions.ValidationError: Value for Import Log cannot be a list
```

ruthra-kumar:
[...]

PR #32110, opened 2022-09-07
fix: error on bank statement import

### LINKED ISSUE #31288: Bank Statement Import always fails due to type error (v14-beta)
### Information about bug

The Bank Statement Import function seems to be completely unworking at the moment. I was unable to get it to work correctly, regardless of import method or data format.

The issue appears to be mismatched type assumptions for the variable `import_log`. In the doctype controller, `import_log` is defined as a list, whereas in the javascript frontend it is assumed to be a JSON string.

The problematic line is this one:
```javascript
let import_log = JSON.parse(frm.doc.import_log || "[]");
```
...which appears in the file three different times:
https://github.com/frappe/erpnext/blob/b29edb302ca91e5f1e81761b604b88e733c39c3f/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js#L143-L144
https://github.com/frappe/erpnext/blob/b29edb302ca91e5f1e81761b604b88e733c39c3f/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js#L311-L312
https://github.com/frappe/erpnext/blob/b29edb302ca91e5f1e81761b604b88e733c39c3f/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js#L441-L442

JSON.parse fails because `frm.doc.import_log` is being passed from the controller not as a JSON string but rather as an empty list. This also creates a problem with saving the Bank Statement Import list, as the database expects a serialized string as well.

### Module

other

### Version

Frappe v14-develop
ERPNext v14-develop

### Installation method

_No response_

```shell
  File "apps/frappe/frappe/desk/form/save.py", line 23, in savedocs
    doc.save()
  File "apps/frappe/frappe/model/document.py", line 288, in save
    return self._save(*args, **kwargs)
  File "apps/frappe/frappe/model/document.py", line 325, in _save
    self._validate()
  File "apps/frappe/frappe/model/document.py", line 512, in _validate
    self._validate_length()
  File "apps/frappe/frappe/model/base_document.py", line 870, in _validate_length
    for fieldname, value in self.get_valid_dict(ignore_virtual=True).items():
  File "apps/frappe/frappe/model/base_document.py", line 316, in get_valid_dict
    frappe.throw(_("Value for {0} cannot be a list").format(_(df.label)))
  File "apps/frappe/frappe/__init__.py", line 503, in throw
    msgprint(
  File "apps/frappe/frappe/__init__.py", line 478, in msgprint
    _raise_exception()
  File "apps/frappe/frappe/__init__.py", line 430, in _raise_exception
    raise raise_exception(msg)
frappe.exceptions.ValidationError: Value for Import Log cannot be a list
```

ruthra-kumar:
## Issue
Bank Statement Import throws 'Value for Import log cannot be List'. 
This is caused by name collision between Framework's File importer's internal attribute and Bank Statement import log - `import_log` field.

https://github.com/frappe/frappe/blob/b47205ff01d3276d87c590c9ad0d9595e6c5eb45/frappe/core/doctype/data_import/importer.py#L50-L61

`import_log` overridden on call to `get_preview_from_template`

https://github.com/frappe/erpnext/blob/008542b7155791d3f5cbbd4701fde2d953be3802/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.py#L86-L88
## Fix
Renamed `import_log` to `statement_import_log`.

Fixes: #31288 , #31688

### DIFF
--- a/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
+++ b/erpnext/accounts/doctype/bank_statement_import/bank_statement_import.js
@@ -141,7 +141,7 @@ frappe.ui.form.on("Bank Statement Import", {
 	},
 
 	show_import_status(frm) {
-		let import_log = JSON.parse(frm.doc.import_log || "[]");
+		let import_log = JSON.parse(frm.doc.statement_import_log || "[]");
 		let successful_records = import_log.filter((log) => log.success);
 		let failed_records = import_log.filter((log) => !log.success);
 		if (successful_records.length === 0) return;
@@ -309,7 +309,7 @@ frappe.ui.form.on("Bank Statement Import", {
 	// method: 'frappe.core.doctype.data_import.data_import.get_preview_from_template',
 
 	show_import_preview(frm, preview_data) {
-		let import_log = JSON.parse(frm.doc.import_log || "[]");
+		let import_log = JSON.parse(frm.doc.statement_import_log || "[]");
 
 		if (
 			frm.import_preview &&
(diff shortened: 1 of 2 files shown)


ruthra-kumar:

PR #32113, opened 2022-09-07
fix: internal transfer flow

rohitwaghchaure:
image

1. Added purchase order reference in sales document
2. Added Set From Warehouse in the Purchase Order 
3. Fixed warehouses in the sales order and purchase receipt for internal transfer flow


deepeshgarg007:

PR #32117, opened 2022-09-07
fix: Reduce font size for Process Statement of accounts print/pdf

deepeshgarg007:
The Print statement of accounts was inheriting the font size from the base template which is specifically designed to show pages in "Portrait Orientation"

The PSOA sends AR/AP reports which are usually in landscape mode and due to the big font size the report was taking more pages unlike the one made from the general ledger, so reduced the font size to accommodate more rows.


deepeshgarg007:

deepeshgarg007:

PR #32135, opened 2022-09-08
fix: reposting not working for internal transferred purchase receipt

rohitwaghchaure:
**Issue**

If user has created backdated transactions which has impact on the future Internal Transfer entries like Inter Delivery Note and Inter Purchase Receipt. System do reposting for Inter Delivery Note as per new rate bit not for Inter Purchase Receipt. Due to which the outward and inward cost is not same.

image


rohitwaghchaure:

rohitwaghchaure:
https://github.com/Mergifyio backport version-13-hotfix

PR #32137, opened 2022-09-08
refactor(minor): delete linked payment ledger entries on source doc deletion

ruthra-kumar:
With 'Delete Accounting and Stock Ledger Entries on deletion of Transaction' toggled in Account settings and upon deletion of source document, corresponding Payment Ledger Entries should be deleted.


ruthra-kumar:

PR #32139, opened 2022-09-08
fix: remove EmployeeBoardingController

barredterra:
`EmployeeBoardingController` is part of hrms now. This caused errors when updating Project and hrms was not installed.

https://github.com/frappe/hrms/pull/55


ruchamahabal:
Unrelated accounting dimension tests failing

ruchamahabal (approved):

PR #32144, opened 2022-09-09
fix: Rate for internal PI have non stock UOM items

deepeshgarg007:
For Internal transfers within the same company, the net rate was fetched incorrectly for Purchase Invoices.

For Eg, Lets an internal Sales Invoice is made for Item A with UOM "Kg" for 1 Kg having the stock UOM as grams
while making an internal Purchase Invoice against that SI the rate for the Item is fetched from the corresponding SI items
`internal_rate` field. The `internal_rate` is maintained in stock UOM and making a PI in non-stock UOM fetched an incorrect rate as the conversion factor was not considered.

This PR fixes that issue


deepeshgarg007:

PR #32150, opened 2022-09-09
refactor: BOM Stock Calculated report, fix required-qty

s-aga-r:
Source / Reference: ISS-22-23-02270

- BOM Items

  

- BOM Item(s) Qty and Stock Qty

  

- Case 1: When BOM Qty is 1

  

  - BOM Stock Calculated Report
 
    - Before
    
      

    - After

      

- Case 2: When BOM Qty is more than 1

  

    - BOM Stock Calculated Report

      - Before

        

      - After

        

#32735


rohitwaghchaure (dismissed):

s-aga-r:
#32735

PR #32156, opened 2022-09-11
fix: remove duplicate call to ple creation

ruthra-kumar:
Remove duplicate call for PLE creation on reverse gl creation event.


ruthra-kumar:

PR #32184, opened 2022-09-12
fix(Salary Slip): set default amount to 0 if None

ruchamahabal:
**Problem**:

Components dependent on additional salary components by formulae error out during calculation if the additional salary default amount is None.
Add 0 as a fallback while setting default amount in structures

image


tajff:
The Saudi system, if the Saudi employee, must monthly deduct part of his salary, even if the employee took an unpaid leave for a month, a message appears that cannot be approved because the balance is too small.
image

PR #32190, opened 2022-09-12
fix: item wise sales register taxes and charges

maharshivpatel:
I have added a separate column for other charges. Instead of adding all values to tax_total, it checks if account_type is tax, and then only it adds to total_tax otherwise it adds to the total_other_charges.

### Sales Invoice taxes table:

Invoice taxes table

### Item-wise Sales Register Before:

Before

### Item-wise Sales Register After:

After


deepeshgarg007:

deepeshgarg007:

PR #32221, opened 2022-09-15
fix: consider posting time for internal transfer PO

rohitwaghchaure:
Posting time field is not there in the purchase order document therefore consider current time as posting time while calculating incoming rate


nabinhait (approved):

nabinhait:
@rohitwaghchaure needs to be backported?

PR #32222, opened 2022-09-15
fix: abbreviation issue on renaming cost center

### LINKED ISSUE #31562: Renaming cost center automatically adds old company abbreviation
### Information about bug

ERPnext automatically adds the company abbreviation after each cost center. 
After a company abbreviation was changed, renaming a cost center does not work properly. ERPnext still adds the old company abbreviation. It is impossible to remove this part of the name, since ERPnext handles it internally (only the front part of the cost center is editable - not the abbreviation itself). 

A similar issue was previously raised here:
https://github.com/frappe/erpnext/issues/26746
This PR addresses a similar issue but doesn't solve the problem.
https://github.com/frappe/erpnext/pull/27766

**Suggested solution:** 
Allow the modification of the abbreviated part of the name which is currently impossible.

### Module

accounts

### Version

ERPNext: v13.35.0 (HEAD)
Frappe Framework: v13.35.0 (HEAD)

### Installation method

FrappeCloud

_No response_

nabinhait:
Closes #31562

### DIFF
--- a/erpnext/accounts/doctype/account/account.py
+++ b/erpnext/accounts/doctype/account/account.py
@@ -37,7 +37,7 @@ class Account(NestedSet):
 	def autoname(self):
 		from erpnext.accounts.utils import get_autoname_with_number
 
-		self.name = get_autoname_with_number(self.account_number, self.account_name, None, self.company)
+		self.name = get_autoname_with_number(self.account_number, self.account_name, self.company)
 
 	def validate(self):
 		from erpnext.accounts.utils import validate_field_number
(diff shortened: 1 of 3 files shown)


nabinhait:

blewsky:
Hi @nabinhait ,
Waited for the release today and only realized now that this was only pushed to the v14-hotfix. Will this be also released to v13? 
(Sorry I couldn't find how any guidance online how it is decided which PRs are backported and which not. Our operations depend on this hotfix and due to other dependencies we cannot upgrade to v14, yet).
Thank you.

PR #32226, opened 2022-09-15
fix: calculate depreciation properly on asset sale entry and scrap entry

anandbaburajan:
#### Issue 1

1. Create asset with the date 01-07-2022 for 12 monthly depreciation
2. Make depreciation entries for July and August
3. Scrap the asset on 27th of September

You will see that the system isn't calculating depreciation for the 27 days of September and hence the Acc. Dep. and Profit & Loss account figure aren't correct in Journal Entry.

#### Issue 2

1. Create asset with the date 01-07-2022 for 12 monthly depreciation
2. Make depreciation entries for July and August
3. Sell the asset on 27th of September

You will see that the system is calculating depreciation for the 27 days of September but same is not getting added in the Acc. Dep. and Profit & Loss account in Journal Entry.

--

Now the system correctly calculates asset's depreciation for the days till it is scraped or sold.


nabinhait:
@AnandBaburajan Can you please add a proper description in the PR? It's not good to add an internal support issue link for future reference because community devs don't have access to it.

nabinhait (dismissed):

anandbaburajan:
> @AnandBaburajan Can you please add a proper description in the PR? It's not good to add an internal support issue link for future reference because community devs don't have access to it.

Ah, done.

nabinhait (approved):

nabinhait:

PR #32233, opened 2022-09-16
refactor: rewrite Production Plan queries in QB

s-aga-r:
#32735


shreyanshj (commented):
In our environment, with this patch (QB update), the Sales Order Items being fetched are not excluding those Sales Order Items which don't have a default BoM. This leads to a warning about a particular Sales Order Item not having a BoM and the fetching fails. When this patch is removed, the behavior falls back to normal. And being only a restructuring, the functional impact isn't much.

I don't have sufficient method to recreate (as I have a customer owned data) - so, just a note if someone observes something similar.

(I will try to isolate to exact line and update when possible)

PR #32245, opened 2022-09-16
feat(stock): Audible indication of barcode scan status.

dj12djdjs:
In some cases, just displaying a short alert regarding the scan status isn't enough. If the user doesn't look back at their screen fast enough after scanning they should visually scan the whole `items`? table just to confirm if the scan was successful or not.

Proposed solution is to add an audible noise to indicate the scan status.

- Adds `play_success_sound` to barcode scan API.
- ADds `play_fail_sound` to barcode scan API.

`no-docs`


dj12djdjs:
cc: @rohitwaghchaure @s-aga-r

Issue #32249, opened 2022-09-17
Dropdown menu is not working in employee leave balance summary report


ArunaDevraj:
### Information about bug

[README.md](https://github.com/frappe/erpnext/files/9591447/README.md)
![1663410262687](https://user-images.githubusercontent.com/100012141/190852190-4101fc72-2f2d-47ac-b17d-6ca5c698037c.jpg)
![1663410262678](https://user-images.githubusercontent.com/100012141/190852200-e44de0cc-57c1-4ccf-8df0-52556e255edf.jpg)

### Module

HR

### Version

ERPNext: v13.35.3 (version-13)

Frappe Framework: v13.35.2 (version-13)

PR #32252, opened 2022-09-17
fix: Change Department fieldtype to Link in Leave Balance Report

Solufyin:
**Version:** 

ERPNext: v13.39.0 (version-13)
Frappe Framework: v13.41.1 (version-13)
___
Issue: #32249
___

**Before:**
- When add column from the report then does not show link fields like employee and department.

https://user-images.githubusercontent.com/34390782/190858295-33044a5a-c6a4-40c5-bd39-87d7372ba0d6.mp4

**After:**
- After set the link field option, when add columns from the report then link fields will appear.

https://user-images.githubusercontent.com/34390782/190858479-00984ba0-672d-48f6-91ef-da627afb0198.mp4

___
Version 14 PR: https://github.com/frappe/hrms/pull/59

Thank You!

### DIFF
--- a/erpnext/hr/report/employee_leave_balance_summary/employee_leave_balance_summary.py
+++ b/erpnext/hr/report/employee_leave_balance_summary/employee_leave_balance_summary.py
@@ -22,9 +22,9 @@ def execute(filters=None):
 
 def get_columns(leave_types):
 	columns = [
-		_("Employee") + ":Link.Employee:150",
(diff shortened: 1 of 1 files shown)


ruchamahabal (approved):

ArunaDevraj:
After adding the links and refresh the site, it shows like : IndexError: list index out of range

PR #32267, opened 2022-09-19
fix: create contact after lead doctype insert

niyazrazak:
when validate lead multiple contact has been created when its is placed on before_insert so i changed its to after insert.


deepeshgarg007:
@niyazrazak Even `before_insert` is called once when the record is created for the first time, don't think this is the issue for multiple contact creation. Can you please add some GIFs and a test case to verify? Closing this PR until then

PR #32272, opened 2022-09-19
fix: Incoming rate precision fixes for intra company transfer

deepeshgarg007:
While doing inter-company stock transfers via Sales/Purchase Invoices for items having multiple valuation rates there might be cases that a small difference in stock change might be seen between source and target warehouse SLEs

This might lead to incorrect values and stock and account value mismatch issues as well
So in order to avoid this added a GL Entry for the diff and booked it as COGS


nabinhait (dismissed):

deepeshgarg007:

erpnext/controllers/stock_controller.py
```diff
@@ -191,6 +193,47 @@ def get_gl_entries(
 					elif sle.warehouse not in warehouse_with_no_account:
 						warehouse_with_no_account.append(sle.warehouse)
 
+			if abs(sle_rounding_diff) > 0.1 and (
```
nabinhait: Instead of a hardcoded value (0.1), you can use a value based on the precision of the debit/credit field of GLE.
Ref: https://github.com/frappe/erpnext/blob/develop/erpnext/accounts/general_ledger.py#L402

erpnext/accounts/doctype/sales_invoice_item/sales_invoice_item.json
```diff
@@ -820,6 +820,7 @@
    "label": "Incoming Rate (Costing)",
    "no_copy": 1,
    "options": "Company:company:default_currency",
+   "precision": "6",
```
nabinhait, line 823: The same should be done for DN and also in PR/PI for the valuation_rate field.

erpnext/controllers/stock_controller.py
```diff
@@ -191,6 +193,47 @@ def get_gl_entries(
...
+				self.get("is_internal_customer") or self.get("is_internal_supplier")
+			):
+				asset_account = ""
+				if self.get("is_internal_customer"):
+					asset_account = warehouse_account[item_row.get("target_warehouse")]["account"]
+				elif self.get("is_internal_supplier"):
+					asset_account = warehouse_account[item_row.get("from_warehouse")]["account"]
+
+				expense_account = item_row.get("expense_account")
+				gl_list.append(
```
nabinhait, line 205: We need a discussion on GLE

nabinhait (changes requested):

PR #32284, opened 2022-09-20
fix: remove no_copy for ignore_pricing_rule

maharshivpatel:
### Problem: 
no_copy checked on ignore_pricing_rule breaks sales cycle.

### Steps to Reproduce:
1. create pricing rule for any item.
2. create sales order where condition of pricing rule is matched
3. check ignore pricing rule and manually enter price
4. save and submit
5. create sales invoice against it
6. pricing rule is applied on sales invoice as ignore_pricing_rule was not copied and all of the item prices are overridden as per pricing rule.

### Before :
 

https://user-images.githubusercontent.com/39730881/191172867-98f9bb16-de54-4a30-98e9-79c0caf2db6d.mov

### After :

https://user-images.githubusercontent.com/39730881/191172167-0431ae13-5205-4da5-a733-1c6269c470c5.mov


deepeshgarg007:

PR #32296, opened 2022-09-20
fix: get amount in words for debit note

ruthra-kumar:
Get amount in words for Debit Note.
Screenshot 2022-09-20 at 5 35 10 PM


ruthra-kumar:
@Mergifyio backport version-12-hotfix version-13-hotfix version-14-hotfix

PR #32303, opened 2022-09-21
fix: difference amount calculation and popup on payment reconciliation

### LINKED ISSUE #29081: Not getting the Option to account the Exchange Gain/Loss in case of Multi-Currency Transaction in the Payment reconciliation Window which was available in the earlier versions 12
### Information about bug

One Sample Use case is taken to understand the Issue/Bug with respect to the above issue.

1. Created one Supplier with the USD.

image
image
image

2. Record one Purchases Invoice Entry with that Supplier- Say 1 Qty @ 1000 USD Exchange Rate- 1 USD= 80 Rs
image

3.Recording the Payment Entry with respect to that Supplier for USD1000  Exchange Rate- 1 USD= 75 Rs
image

4. Now, When we go to the Payment Reconciliation window to reconcile the same - that USD 1000 - E.R- 80 PI entry ith the USD-1000- E.R-75, then the system shows the below error
image

image

image

image

In earlier version 12 it shows the Window to account this Rs 5000 Exchange Gain/Loss Amount but in Version 13 it is not showing
Please check the below screenshot in version 12.7
image

image

image

You can see the above image where it gives an option to accounts that difference Rs 5000 to the Exchange Gain/Loss Entry 

This window is not coming in Versions 13 onwards.

Kindly consider this bug as a lot of Foreign Currencies entries that need to be reconciled through Payment reconciliation is not happening.

-

### Module

accounts

### Version

image

### Installation method

FrappeCloud

```shell
Debit and Credit not equal for Payment Entry #ACC-PAY-2021-00001. The difference is 5000.0.
```

ruthra-kumar:
Difference amount calculation was broken on Payment Reconciliation tool. PR fixes this.
Screenshot 2022-09-21 at 9 25 31 PM
Screenshot 2022-09-21 at 9 25 50 PM

### DIFF
--- a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
+++ b/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
@@ -14,6 +14,7 @@ from erpnext.accounts.utils import (
 	QueryPaymentLedger,
 	get_outstanding_invoices,
 	reconcile_against_document,
+	update_reference_in_payment_entry,
 )
 from erpnext.controllers.accounts_controller import get_advance_payment_entries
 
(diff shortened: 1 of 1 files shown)


ruthra-kumar:

PR #32020, opened 2022-08-30
fix:total value in all key

nishibakabeer:
Gross and net profit report showing wrong values in monthly quarterly and half yearly filters which is the total value
@nextchamp-saqib @deepeshgarg007 @ankush 

> Please provide enough information so that others can review your pull request:

> Explain the **details** for making this change. What existing problem does the pull request solve?

> Screenshots/GIFs

PR #32310, opened 2022-09-21
fix: total value in all keys

nishibakabeer:
Gross and net profit report showing wrong values in monthly quarterly and half yearly filters which is the total value @ruthra-kumar added in develop branch as suggested ( https://github.com/frappe/erpnext/pull/32020)


ruthra-kumar:

PR #32341, opened 2022-09-24
Update gstr_2.py

Barathpalanisamy:
GSTR2 report


deepeshgarg007:
@Barathpalanisamy Please add a proper description for the fix and then reopen again

PR #32368, opened 2022-09-27
chore: remove discount accounting logic from PI and PI Item

ruthra-kumar:
1. Clean up Purchase Invoice and Purchase Invoice Item doctypes. Removed discount account related fields and logic.
3. Clean up buying_settings.json


deepeshgarg007:

PR #32370, opened 2022-09-27
fix: Not allowing to return expired batches using purchase return

rohitwaghchaure:
**Issue**
System was not allowing to make purchase return entry against expired batch by throwing below validation

Screenshot 2022-09-27 at 11 52 32 AM

**After Fix**

User will be able to create purchase invoice return or purchase receipt return against the expired batch.


mujeerhashmi:
Backport needed.

rohitwaghchaure:
> Backport needed.

Added labels already, these labels will create backport PRs

PR #32378, opened 2022-09-27
fix: Add return against indexes for POS Invoice

deepeshgarg007:
`set_status` is called twice while creating a POS Invoice, once on `validate` and once on `submit`. The `set_status` method in `pos_invoice.py` (similar to sales invoice) has an if block which check for return invoices to set the status as "Credit Note Issued"

The very same query in POS Invoice took around ~5 sec (depending on the no of existing POS Invoices) for a specific site. Post adding the index the POS invoice submission took 3 seconds instead of 13 seconds

telegram-cloud-photo-size-5-6165440435925923276-y

Also showed up in slow query logs
telegram-cloud-photo-size-5-6165440435925923281-y

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -743,4 +743,4 @@ def add_return_modes(doc, pos_profile):
 
 
 def on_doctype_update():
-	frappe.db.add_index("POS Invoice", ["customer", "is_return", "return_against"])
+	frappe.db.add_index("POS Invoice", ["return_against"])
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -740,3 +740,7 @@ def append_payment(payment_mode):
 		]:
 			payment_mode = get_mode_of_payment_info(mode_of_payment, doc.company)
 			append_payment(payment_mode[0])
+
+
+def on_doctype_update():
+	frappe.db.add_index("POS Invoice", ["customer", "is_return", "return_against"])
```
ankush: Just a simple return against index should be enough. It has high enough cardinality ~= unique return. 

That can be managed from doctype directly 👀

ankush (changes requested):

deepeshgarg007:

PR #32379, opened 2022-09-27
fix: POS only validate QTY if is_stock_item

maharshivpatel:
### **POS issues:**

1. POS invoice raised " Item not available " validation error even though the item is non_stock.
2. In the case of Product Bundle that have non_stock item `Available Qty at Warehouse` was calculated incorrectly ( always 0 ).

### PRODUCT BUNDLE 
product bundle with non stock item

### Before POS

https://user-images.githubusercontent.com/39730881/192609045-cc8eff15-1a15-4163-8356-b3acb35db13c.mov

### After POS

https://user-images.githubusercontent.com/39730881/192609052-f481a365-fdb4-469f-a08c-ee55ed350bce.mov


nextchamp-saqib (approved):

deepeshgarg007:

PR #32389, opened 2022-09-27
fix: don't validate qty if not document is not being submitted.

dj12djdjs:
During the receipt process scanning an items barcode will increment the "Accepted Qty" field in the items table.

The Accepted qty should start from zero if the user is scanning barcodes. 

The user is unable to save the document into a draft status because of this validation.

**Proposed Solution** 
Only validate the qty if the document is being submitted.


deepeshgarg007:
@dj12djdjs can you please add a GIF as well for this, I am unable to exactly understand the use case

dj12djdjs:
@deepeshgarg007 Please see attached user flow.

1. User creates Purchase Receipt
2. Set All Accept Qty to Zero
3. Save document as Draft as not to lose data.
4. Scan items to increment Accepted qty

*Before PR*
123-before

*After PR*
123-after

deepeshgarg007:
@dj12djdjs Why would a user follow this flow is what I am still not able to understand, why wouldn't they simply scan and add the items, there are users who want this validation

dj12djdjs:
> @dj12djdjs Why would a user follow this flow is what I am still not able to understand, why wouldn't they simply scan and add the items, there are users who want this validation

@deepeshgarg007 The validation still occurs, just on submit. I don't understand the reason to validate a "Draft" document. If the user hasn't completed their job and wants to save for later.

If you just scan the items without loading them from Purchase Invoice or Purchase Order first, then the reference will not be used and the respective Purchase Order will not be closed. 

Let me try to explain the scenarios in detail.

Option 1: Create a new Purchase Receipt against a given supplier, then scan the physical items which will load them into the Purchase Receipt Item table.
Pro: This will ensure you receive exactly what you scan.
Con: There won't be any reference to a Purchase Invoice or Purchase Order.

Option 2: Create a new Purchase Receipt pre loaded with item's expected from the delivery via the `Get From` -> (`Purchase Order`, `Purchase Invoice`) and set all the accepted qty to zero.
The user can than scan each of the items incrementing the accepted qty.

Let's say the delivery is 60 pallets, with thousands of items. If worker doesn't have time to finish scanning the whole delivery, they can save it as draft and continue scanning the next day.

A few days later, worker finishes scanning all items received. The worker submits the purchase receipt. If everything expected in the delivery is scanned the purchase receipt will post to stock successfully. If there are any item's not scanned at all. The worker will be notified via the validation that X item has zero accepted qty and won't be able to proceed until ether the row is removed manually, or the worker found this item they missed and scan the qty.

Does this use-case make sense?
Please let me know if you have any comments.

rohitwaghchaure:
@dj12djdjs 

Consider in purchase receipt there are 200 line items then as per your logic user will change the qty to zero for each line item and then they will manually scan each item again to enter quantity? this might be overwhelming for the user

In my opinion system should set the QTY based on Barcode Scan only when there is no scanned item exists in the line items table

We need to fix the functionality of the scan barcode field
In my opinion the scan barcode field should work as follow

1. If line item exists with quantity and if user scanning serial nos as barcode then No of Serial Nos and Qty should be same
2. If line item exists with quantity and if user scanning batch no as Barcode then system won't update Qty but set batch no in the batch no field
4. If line item exists with quantity and if user scanning Barcode then system should show scanned qty and not update QTY field.

I am working on this issue and will open the PR ASAP

Closing this PR Now and will update you once PR is ready

rohitwaghchaure:
@dj12djdjs 

I have opened the [PR](https://github.com/frappe/erpnext/pull/32799
) to fix this issue, kindly test and share your feedback

PR #32394, opened 2022-09-28
fix: fetch swift number in payment request from bank doctype

maharshivpatel:
There isn't direct link between payment_request and bank as it was Read Only field so swift_number wasn't fetched using Fetch form. I fixed it by making bank Link field.

### Bank with SWIFT Number
Bank with SWIFT Number

### Before Payment Request:
https://user-images.githubusercontent.com/39730881/192717277-71d092be-ad8d-4fe2-8c36-872e97fa567b.mov

### After Payment Request:
https://user-images.githubusercontent.com/39730881/192717339-96b6d322-2548-4992-a3f4-5363882973d9.mov


deepeshgarg007:

PR #32399, opened 2022-09-28
fix: don't count half day in absent days in Monthly Attendance Sheet summarized view

ruchamahabal:
port of: https://github.com/frappe/hrms/pull/68


ruchamahabal:
Unrelated test failure: test_get_voucher_wise_gl_entry

PR #32404, opened 2022-09-28
fix(ux): show `Make Purchase Invoice` button based on permission

s-aga-r:
ref: ISS-22-23-02516

192753472-91b45af7-7ed8-41d2-ae34-8e84ffd00446

Problem: The "Make Purchase Invoice" button is visible even if the logged-in user doesn't have the create permission.


s-aga-r:
@rohitwaghchaure

PR #32408, opened 2022-09-28
feat: Tab Breaks in all Sales and Purchase cycle transactions

nabinhait:
Tabs introduced in the following documents:
- Quotation
- Sales Order
- Delivery Note
- Sales Invoice
- Material Request
- Supplier Quotation
- Purchase Order
- Purchase Receipt
- Purchase Invoice

**Tab Pattern**
sales_invoice_with_tabs

`no-docs`


rtdany10:
@nabinhait the more info tab seems to be cluttered in the GIF

deepeshgarg007 (changes requested):
1. Got this error when I tried to create a new Quotation, some `depends_on` expression is invalid

image

2. The middle column remains empty which looks a bit weird, this is seen in all sales and purchase transactions, maybe move the net weight field there or use just two column layout for this section

Screenshot 2022-10-05 at 4 22 32 PM

3. This same is also seen in the items section in some transactions, the latter one in Purchase Order looks better

Screenshot 2022-10-05 at 4 31 25 PM

Screenshot 2022-10-05 at 5 03 53 PM

4. Line break missing for totals section in PO

Screenshot 2022-10-05 at 5 11 09 PM

5. "More Information" and "Additional Information" instead of "More Info" and "Additional Info"
6. One general concern is that this messes up some of the regional custom field ordering that is created via custom apps which we can control and fix to some extent, the user-created custom fields will just have to be fixed manually

deepeshgarg007:

PR #32412, opened 2022-09-28
fix: Incorrect TCS amount deducted in Sales Invoice

deepeshgarg007:
Advances were getting considered incorrectly while calculating TCS.

Payments against invoices that are already considered in the amount for check threshold was also getting considered which let to incorrect calculation of TCS amount


deepeshgarg007:

PR #32424, opened 2022-09-30
feat: Repayment schedule types for term loans

deepeshgarg007:
Previously the loans of type "Is Term Loan" supported only one type of loan schedule, In certain cases, the loan schedule has to follow calendar months and the interest calculation also has to be pro-rated for starting partial months.

In order to handle such scenarios introduced a Repayment Schedule Type in Loan Type
Screenshot 2022-10-22 at 10 47 22 PM

There are two types available now
1. Monthly as per repayment start date (as per current repayment schedule generation logic 
2. Pro-rated as per calendar months

For the "Pro-rated as per calendar months" the Interest Calculation Start Date (`repayment_start_date`) is considered as the date from which the interest calculation starts and the first date of the next month is considered as the first payment date and from there on as per calendar months the interest is calculated based on the balance amount for 30 days each.

There is also one more option available when the "Pro-rated as per calendar months" is chosen as to when the repayment date should be, should it be "First day of the upcoming month" or the "End day of the current month"

The repayment schedule will look as follows

Screenshot 2022-09-30 at 3 26 15 PM

Screenshot 2022-09-30 at 3 26 49 PM

`no-docs`


deepeshgarg007:

PR #32436, opened 2022-09-30
fix: add validation for non-stock item in SCR

s-aga-r:
ref: ISS-22-23-02628


s-aga-r:
The failing test is not related.

PR #32441, opened 2022-09-30
fix: fixes query builder order by issue in postgres

shridarpatil:
fixed query builder order by issue in Postgres


ankush:
@shridarpatil can you share error you're facing? This is maybe better fixed at framework level so we don't have to repeat ourselves in every query.

ankush (approved):

PR #32443, opened 2022-09-30
fix: wrong status when delivery note has returns

HarryPaulo:
Some delivery notes are getting "To Bill" status, even with payment of the amount.

This happens when goods are exchanged.
Ex: I sold the product, the customer paid, the product had a problem, I make a return, and create a new Delivery Note to deliver.

How to Play:
1 - Create Sales Order
2 - Create sales invoice
3 - Create Delivery Note
At this moment the first delivery is made, the order will be as "Completed"

4 - Create a Delivery Note Return.
The Sales Order will be real, for an EXCHANGE to be carried out, of which I will be. the were to return

5 - Create a new Delivery Note for Sales Orders, to deliver as they were returned.
"Delivery Note" created, has the status "To Bill", but in this process the amounts were already PAID at the beginning, in step 2

With my change, the system will consider the returns for signing the Status


HarryPaulo:
waiting for reviewers...

deepeshgarg007:
@HarryPaulo Can you please add a unit test for this scenario as well so that in case any other change breaks this again we can easily identify

HarryPaulo:
i will create the test for this scenario, but i dont have time at now... soon...

HarryPaulo:
@deepeshgarg007 idk how i can do a unit test for it

PR #32456, opened 2022-10-01
fix: update with new Frappe color. fix #32455

### LINKED ISSUE #32455: Change the active/selected timeslot's color on appointment booking screen
It still uses the old ERPNext design, that was #5e64ff
image

Please change to the new color: #2490ef
image

Thank you

backendoadmin:
Update with new Frappe color. fix #32455

### DIFF
--- a/erpnext/www/book_appointment/index.css
+++ b/erpnext/www/book_appointment/index.css
@@ -45,7 +45,7 @@
 
 .time-slot.selected {
     color: white;
-    background: #5e64ff;
+    background: #2490ef;
 }
 
 .time-slot.selected .text-muted {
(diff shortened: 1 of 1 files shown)


erpnext/www/book_appointment/index.css
```diff
@@ -45,7 +45,7 @@
 
 .time-slot.selected {
     color: white;
-    background: #5e64ff;
+    background: #2490ef;
```
ankush:
```suggestion
    background: var(--primary-color);
```
This should work too?

ankush (changes requested):

erpnext/www/book_appointment/index.css
backendoadmin: Idk, I'm just using Frappe Cloud on public bench yet, so I cannot test your version, but it seems correct to me.

ankush (approved):

PR #32476, opened 2022-10-03
Item Wise TDS Calculation

niralisatapara:
**### For calculating Item wise TDS in Purchase Invoice.**

Step 1:  **To calculate TDS item wise, added apply_tds checkbox in Purchase Invoice Item.**
image

Step 2: **Calculate Tax Withholding Net Total based where apply_tds is checked.**
image

Step 3: **Computing TDS based on Tax Withholding Category**
image

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
@@ -1,1505 +1,1515 @@
 {
- "actions": [],
- "allow_import": 1,
- "autoname": "naming_series:",
- "creation": "2013-05-21 16:16:39",
- "doctype": "DocType",
- "document_type": "Document",
- "engine": "InnoDB",
- "field_order": [
-  "title",
-  "naming_series",
-  "supplier",
-  "supplier_name",
-  "tax_id",
-  "due_date",
-  "tax_withholding_category",
-  "column_break1",
-  "company",
-  "posting_date",
-  "posting_time",
-  "set_posting_time",
-  "is_paid",
-  "is_return",
-  "apply_tds",
-  "amended_from",
-  "accounting_dimensions_section",
-  "cost_center",
-  "dimension_col_break",
-  "project",
-  "supplier_invoice_details",
-  "bill_no",
-  "column_break_15",
-  "bill_date",
-  "returns",
-  "return_against",
-  "section_addresses",
-  "supplier_address",
-  "address_display",
-  "contact_person",
-  "contact_display",
-  "contact_mobile",
-  "contact_email",
-  "col_break_address",
-  "shipping_address",
-  "shipping_address_display",
-  "billing_address",
-  "billing_address_display",
-  "currency_and_price_list",
-  "currency",
-  "conversion_rate",
-  "column_break2",
-  "buying_price_list",
-  "price_list_currency",
-  "plc_conversion_rate",
-  "ignore_pricing_rule",
-  "sec_warehouse",
-  "set_warehouse",
-  "rejected_warehouse",
-  "col_break_warehouse",
-  "set_from_warehouse",
-  "supplier_warehouse",
-  "is_subcontracted",
-  "items_section",
-  "update_stock",
-  "scan_barcode",
-  "items",
-  "pricing_rule_details",
-  "pricing_rules",
-  "raw_materials_supplied",
-  "supplied_items",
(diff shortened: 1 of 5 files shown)


erpnext/accounts/doctype/purchase_invoice/test_purchase_invoice.py
```diff
@@ -1574,6 +1574,35 @@ def test_batch_expiry_for_purchase_invoice(self):
 
 		self.assertTrue(return_pi.docstatus == 1)
 
+	def test_without_tds(self):
```
deepeshgarg007, line 1577: Please move these tests to `test_tax_withholding_category.py`. Other tax withholding tests are there, so good to have all the tests in the first place

erpnext/accounts/doctype/purchase_invoice/test_purchase_invoice.py
```diff
@@ -1682,6 +1711,86 @@ def make_purchase_invoice(**args):
 			pi.submit()
 	return pi
 
+def make_purchase_invoice_tds(**args):
```
deepeshgarg007, line 1714: This seems like duplicate code, `test_tax_withholding_category.py` already creates tax withholding purchase invoices, please use the same approach

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -865,6 +866,12 @@
    "label": "Product Bundle",
    "options": "Product Bundle",
    "read_only": 1
+  },
+  {
+    "default": "1",
```
deepeshgarg007, line 871: A patch to update this check in old items is required

erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json
```diff
@@ -1,1505 +1,1515 @@
...
+    "default": "0",
+    "fieldname": "is_old_subcontracting_flow",
+    "fieldtype": "Check",
+    "hidden": 1,
+    "label": "Is Old Subcontracting Flow",
+    "read_only": 1
+   },
+   {
+     "default": "0",
+     "fieldname": "tax_withholding_net_total",
```
deepeshgarg007, line 1427: A base currency field is also needed for multi-currency transactions.
Also, a patch to update `net_total` in old invoices to this field is needed.

https://github.com/frappe/erpnext/blob/07c4a7483869a06c1e00a2a71c1909fcf5cd9ed6/erpnext/accounts/doctype/tax_withholding_category/tax_withholding_category.py#L373

Check this method mentioned above, many users withheld taxes using "Journal Entry" and "Payment Entry" as well so we check the credit value directly from `GL Entries` to get get the already withheld tax amount. You'll have to modify this method to get the amount from Purchase Invoices using `tax_withholding_net_total ` and using GL entries for other doctypes

erpnext/controllers/taxes_and_totals.py
```diff
@@ -58,12 +58,23 @@ def _calculate(self):
 		self.initialize_taxes()
 		self.determine_exclusive_rate()
 		self.calculate_net_total()
+		self.calculate_tax_withholding_net_total()
```
deepeshgarg007, line 61: Since you have entered this field just for purchase invoices, this method can be called inside validate event in `purchase_invoice.py`

erpnext/controllers/taxes_and_totals.py
```diff
@@ -58,12 +58,23 @@ def _calculate(self):
...
 		self.calculate_taxes()
 		self.manipulate_grand_total_for_inclusive_tax()
 		self.calculate_totals()
 		self._cleanup()
 		self.calculate_total_net_weight()
 
+	def calculate_tax_withholding_net_total(self):
```
deepeshgarg007, line 68: This can also be moved to `purchase_invoice.py`

deepeshgarg007 (changes requested):
Congratulations @niralisatapara for your first PR, a tough topic that you have picked up as your first contribution, requesting some changes so that we can merge this to the core, have left comments on specific lines wherever possible, adding some general comments below.

- This is also used for applying TCS on the Sales side, which seems to be completely ignored as of now
- Please go through the pull request checklist and setup linter to avoid failing styling checks https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

PR #32477, opened 2022-10-03
Austria chart of accounts update

rasos:
improved mapping of VAT accounts - please consider to move to verified folder, as this chart is being used in production since many years


deepeshgarg007:
@rasos You can move to verified folder and raise a PR. For your next PR refer https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist

PR #32478, opened 2022-10-03
feat(JE): trigger account field when fetched from template

### LINKED ISSUE #32409: Incorrect account currency shown in new journal entry
### Information about bug

I am using accounts in USD, EUR and CHF (the latter being the company currency). 

When creating a new journal entry involving accounts in a currency different from the company currency, the correct currency is shown in the _Account Currency_ column of the _Accounting Entries_ table.

However, when either _duplicating_, or creating from a _template_, those same accounts will show, and use, the company currency, not their own currency, in that column, as well as in the detailed entry form. 

The only way to correct this is to delete the account entry from the table and to re-enter it.

image

### Module

accounts

### Version

Frappe v14.5.0
ERPNext v14.1.2

### Installation method

manual install

_No response_

rtdany10:
Closes #32409

#no-docs

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -173,8 +173,8 @@ frappe.ui.form.on("Journal Entry", {
 var update_jv_details = function(doc, r) {
 	$.each(r, function(i, d) {
 		var row = frappe.model.add_child(doc, "Journal Entry Account", "accounts");
-		row.account = d.account;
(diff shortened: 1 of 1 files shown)


rtdany10:
Also, need to uncheck **No Copy** from currency field.
https://github.com/frappe/erpnext/issues/32409#issuecomment-1264238195

deepeshgarg007:

deepeshgarg007:
@rtdany10 Not sure about the no copy thing yet, merging this change for now

PR #32497, opened 2022-10-04
chore: set `Quality Inspection` status based on readings status

s-aga-r:
ref: ISS-22-23-02678

- Set the Quality Inspection status based on the status of the reading:
  - **Case - 1**: When there are one or more readings with `rejected` status and parent `manual inspection` is unchecked, then parent status should be set to `rejected`.
  - **Case - 2**: When all readings have `accepted` status and parent `manual inspection` is unchecked, then parent status should be set to `accepted`.
  - **Case - 3**: When parent `manual inspection` is checked, then parent status should not be changed.


s-aga-r:
@rohitwaghchaure 

Added the requested changes:
1. Make readings status mandatory.
2. Set parent status to rejected if one or more readings with rejected status.

rtdany10:
~@s-aga-r I would say, move this validation to `on_submit`.~
~Let the users input values and save without throwing the error and validate only on submit.~

~Scenario: QC team might consist of multiple people testing for multiple parameters. So, one person might not update all parameters.~

rtdany10:
Just realized that the code doesn't throw any errors but only silently updates the status. Maybe we should inform the user with a message? @s-aga-r

PR #32500, opened 2022-10-05
feat: refactoring activation level logic 

shadrak98:
### Current Implementation:
Previously, the activation level was calculated on the basis of minimum records created against each doctype.

But once the activation level is calculated, it doesn't differ much from the next day even though the user has not interacted with the system at all. 

### Purposed Change: 
Instead of counting the doctypes, it is better to only calculate records created in the last 24 hrs to check the daily activation level.

### DIFF
--- a/erpnext/utilities/activation.py
+++ b/erpnext/utilities/activation.py
@@ -60,7 +60,7 @@ def get_level():
 	if frappe.db.get_single_value("System Settings", "setup_complete"):
 		activation_level += 1
 
-	communication_number = frappe.db.count("Communication", dict(communication_medium="Email"))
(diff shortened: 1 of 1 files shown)


erpnext/utilities/activation.py
```diff
@@ -59,11 +68,15 @@ def get_level():
...
 		activation_level += 1
 
-	level = {"activation_level": activation_level, "sales_data": sales_data}
+	level = {"activation_level": activation_level, "sales_data": sales_data, "records": records}
 
 	return level
 
 
+def get_new_records(doctype):
+	return frappe.db.count(doctype, dict(creation=("between", (last_24hrs, now()))))
```
ankush, line 77: Since creation doesn't have index so this will get quite slow on large sites. 
We can build time series info from total records too (total vs timestamp) 

So no need to capture this data separately.

PR #32519, opened 2022-10-06
fix: Explicitly update modified

ankush:
required after https://github.com/frappe/frappe/pull/18301

People use Bin doctype's modified timestamp to write integrations, hence updating it explicitly. 

Found out cause my old-timey test started failing 😄 https://github.com/frappe/erpnext/pull/29746


s-aga-r (approved):
LGTM

PR #32529, opened 2022-10-08
feat(report):added account wise redirection

finbyz:
>Problem Statement:

When someone is checking balance-sheet in ERPNext, he will get total Payable in Creditors Account, Total Receivable in Debtors Account.

Now the obvious question will be who are the those creditors or debtors, which is available in our Account Payable and Receivable Summary report. By default in balance sheet report the click on account leads to redirection in general ledger report from all the accounts. Looking general ledger of creditors/debtors without party doesn’t make sense to anyone.
Same issue is with Stock in hand account and Fixed asset accounts.

>Solution:

We have tried to improve this user experience by changing redirect to reports based on Account Type as below:

1. Payable -> Accounts Payable Summary
2. Receivable -> Accounts Receivable Summary
3. Stock -> Stock Balance
4. Fixed Asset, Accumulated Depreciation ->  Asset Depreciations and Balances
5. All Other account type will continue to point to General Ledger Report as current.

> Video

https://user-images.githubusercontent.com/18363620/193403987-e6e1e30d-1479-4322-9d86-2681372eb05f.mp4

`no-docs`


deepeshgarg007:
@finbyz Did some code cleanup, and removed report redirections for Stock and Asset account types as they were not making much sense, for stock accounts there can be warehouse-specific accounts, and just simply redirecting to the stock balance report for all accounts is not making much sense.

Same for assets, different asset categories can have  different account, and just redirecting without proper asset category just doesn't help

deepeshgarg007:

PR #32536, opened 2022-10-08
fix: PO cancel post advance payment cancel against PO

### LINKED ISSUE #32287: Cannot delete or cancel because Purchase Order is linked with GL Entry
### Information about bug

Hi ,

GL Entries with 'Is Cancelled' checked are preventing associated documents (like Purchase Order) from being Cancelled. This shouldn't be so. Please see associated thread below:

https://discuss.erpnext.com/t/cannot-delete-or-cancel-because-purchase-order-is-linked-with-gl-entry/78615

Kindly help fix asap

Thanks

### Module

accounts

### Version

ERPNext: v13.36.5 (version-13)
Frappe Framework: v13.39.1 (version-13)

### Installation method

_No response_

_No response_

deepeshgarg007:
fixes: https://github.com/frappe/erpnext/issues/32287

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.py
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.py
@@ -349,7 +349,7 @@ class PurchaseOrder(BuyingController):
 		update_linked_doc(self.doctype, self.name, self.inter_company_order_reference)
 
 	def on_cancel(self):
-		self.ignore_linked_doctypes = "Payment Ledger Entry"
+		self.ignore_linked_doctypes = ("GL Entry", "Payment Ledger Entry")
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

dawoodjee:
Unfortunately this doesn't solve for the GL because the GL is not submittable, therefore cannot be cancelled

dawoodjee:
> Cannot delete or cancel because Payroll Entry HR-PRUN-2022-00015 is linked
with GL Entry ACC-GLE-2022-38178

I also had this problem when deleting Payroll Entries. Adding line 93 in the payroll_entry.py resolved it.

image

PR #32549, opened 2022-10-10
fix: type-cast while saving an item

Alchez:
**Problem:**

When a non-batch item is saved, the system adds unnecessary logs for version changes.

This happens because the item fields in this PR were being automatically modified and hard-set to `None` during runtime, due to which the version-checker would assume that there's a meaningful change in those fields.

image


Alchez:
@rohitwaghchaure @s-aga-r

If this looks good, can we also backport it to v14?

ankush (approved):

PR #32557, opened 2022-10-11
fix: Value error on validation of POS invoices with Serial Nos

ruthra-kumar:
## Issue
Value error on POS Invoice validation. 
1. Create POS Invoice with 1 item and Serial no
2. Make a return Invoice for [1]
3. Set docstatus = 2 for [1]. (Cancelling [1] cancels [2] automatically, if done through desk. But, if such scenario arises where only [1] is cancelled or deleted, next step throws error)
4. Make a POS Invoice with same serial no as [1] and Save

Screenshot 2022-10-11 at 8 46 36 AM

## Fix
use Set() instead of List


nextchamp-saqib (approved):

ruthra-kumar:

ruthra-kumar:

PR #32563, opened 2022-10-11
fix: consider sales rate as incoming rate for transit warehouse in purchase flow

rohitwaghchaure:
**Issue**

In Inter Transfer case when you make purchase receipt against the deliver note, system calculate the valuation rate in the stock ledger entry as per FIFO and not consider the valuation rate as per the delivery note.

### DIFF
--- a/erpnext/stock/stock_ledger.py
+++ b/erpnext/stock/stock_ledger.py
@@ -589,6 +589,15 @@ class update_entries_after(object):
 		sle.stock_queue = json.dumps(self.wh_data.stock_queue)
 		sle.stock_value_difference = stock_value_difference
 		sle.doctype = "Stock Ledger Entry"
+
+		if (
+			sle.voucher_type in ["Purchase Receipt", "Purchase Invoice"]
+			and sle.voucher_detail_no
+			and sle.actual_qty < 0
+			and frappe.get_cached_value(sle.voucher_type, sle.voucher_no, "is_internal_supplier")
+		):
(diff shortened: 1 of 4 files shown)


erpnext/stock/stock_ledger.py
```diff
@@ -589,6 +589,15 @@ def process_sle(self, sle):
...
 		sle.stock_value_difference = stock_value_difference
 		sle.doctype = "Stock Ledger Entry"
+
+		if (
+			sle.voucher_type in ["Purchase Receipt", "Purchase Invoice"]
+			and sle.voucher_detail_no
+			and sle.actual_qty < 0
+			and frappe.get_cached_value(sle.voucher_type, sle.voucher_no, "is_internal_supplier")
+		):
+			sle.outgoing_rate = get_incoming_rate_for_inter_company_transfer(sle)
```
nabinhait, line 599: The `outgoing_rate` should be set at the beginning of the `process_sle` function, because otherwise the system will not consider this outgoing_rate for calculating `valuation_rate`

nabinhait (approved):

erpnext/stock/stock_ledger.py
nabinhait: The `outgoing_rate` should be set at the beginning of the `process_sle` function, because otherwise the system will not consider this outgoing_rate for calculating `valuation_rate`

nabinhait (changes requested):

nabinhait (dismissed):

PR #32566, opened 2022-10-11
fix: pricing rule item code UOM apply & conversions

maharshivpatel:
There were multiple issues with pricing rule's apply_on_field ==  "item_code" ( UOM : blank, other_than_stock_uom etc ). I have added the required checks and proper conversions.

### Before:
https://user-images.githubusercontent.com/39730881/195068408-1a3a2d4a-7aca-4b3e-8c54-397dd471ba68.mov

### After:
https://user-images.githubusercontent.com/39730881/195422233-1283adb7-447b-441b-8fbd-2a3ab8d74caa.mov


erpnext/accounts/doctype/pricing_rule/pricing_rule.py
```diff
@@ -440,14 +442,19 @@ def apply_price_discount_rule(pricing_rule, item_details, args):
 		if pricing_rule.currency == args.currency:
 			pricing_rule_rate = pricing_rule.rate
 
+		# TODO https://github.com/frappe/erpnext/pull/23636 solve this in some other way.
 		if pricing_rule_rate:
 			# Override already set price list rate (from item price)
 			# if pricing_rule_rate > 0
-			item_details.update(
-				{
-					"price_list_rate": pricing_rule_rate * args.get("conversion_factor", 1),
```
deepeshgarg007, line 448: So the conversion factor coming from the args should be the `rule_conversion_factor` itself. Why this is not coming investigating and fixing that would be a proper fix

erpnext/accounts/doctype/pricing_rule/utils.py
```diff
@@ -111,6 +111,12 @@ def _get_pricing_rules(apply_on, args, values):
 		)
 
 		if apply_on_field == "item_code":
+			if args.get("uom") != args.get("stock_uom"):
+				item_conditions += (
+					" and ({child_doc}.uom='{item_uom}' or IFNULL({child_doc}.uom, '')='')".format(
+						child_doc=child_doc, item_uom=args.get("uom")
+					)
+				)
```
deepeshgarg007: What this exactly does I am not able to understand, please explain this as well

deepeshgarg007 (changes requested):
Also need a test case for this, please add that too

deepeshgarg007:

sjain17:
Thank you for fixing this bug; however, it is still not considering the max qty limitations.

PR #32594, opened 2022-10-13
fix: Party account for multi-order invoices

deepeshgarg007:
While pulling items from multiple orders in the Sales/Purchase Invoice default party account from the company master is fetched rather than the party-specific account mentioned in the customer and supplier master

Party Account:
Screenshot 2022-10-13 at 3 11 43 PM

Before:
Before

After:
After


deepeshgarg007:

PR #32598, opened 2022-10-13
refactor: split ple creation function into two

ruthra-kumar:
Refactor `create_payment_ledger_entry` function. PLE map generation is separated out into a new function.


ruthra-kumar:

PR #32601, opened 2022-10-13
Fix error when import invoice

balabofficial:
When quantity it’s 1 the cycle import with quantity of the last item import because qty = 1 it’s out of for cycle. With this mod now work fine.

> Screenshots/GIFs


deepeshgarg007:
@balabofficial The description is not explained well and it is hard to understand the exact issue here, please update the description and reopen the PR

PR #32607, opened 2022-10-15
chore: Remove HRMS related code

deepeshgarg007:
Removed while fixing https://github.com/frappe/hrms/pull/85

Also added no copy for journal entry reference fields


ruchamahabal (approved):

ruchamahabal:
Unrelated test case failure (opportunity: `test_load`)

PR #32617, opened 2022-10-17
fix: default enabled 'Automatically Fetch Payment Terms from Order' in accounts settings

rohitwaghchaure:
**Issue**

When you make purchase invoice from the purchase order then system won't fetch the Payment Terms from the Purchase Order to Purchase Invoice if the "Automatically Fetch Payment Terms from Order" checkbox has disabled in the Accounts Settings.

**Solution**

Make default behaviour enabled "Automatically Fetch Payment Terms from Order" in the Accounts Setting

### DIFF
--- a/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
+++ b/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
@@ -180,7 +180,7 @@
    "label": "Enable Custom Cash Flow Format"
   },
   {
-   "default": "0",
+   "default": "1",
    "fieldname": "automatically_fetch_payment_terms",
    "fieldtype": "Check",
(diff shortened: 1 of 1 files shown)


bosue:
Enabling this setting by default makes a lot of sense. One might even ask why this switch exists at all.

It is really hard to see a reason someone would create a Sales Order with Payment Terms, but then wouldn‘t want these Payment Terms be preserved when creating the Sales Invoice.

See for example this: #36108. Even devops don‘t expect this to be just an optional feature. It‘s so much expected that people would try and come up with a PR to implement a feature deemed missing.

There used to be a few rough edges, bust most of them have been fixed, such as #35285, and if any problems remain, these will have to be fixed anyway.

For all these reasons, I really think this PR should be reopened!

Edit: one remaining bug I found, is #38030.

PR #32620, opened 2022-10-17
feat: page break in SoA pdf

rtdany10:
- [X] Fetch customer even if no email is found.
- [X] Ability to add page break inside PDF after each SoA when printed consolidated.

https://user-images.githubusercontent.com/52111700/196178937-807e7db1-f277-4636-8c4e-425823f71e54.mp4

#no-docs

### DIFF
--- a/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.py
+++ b/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.py
@@ -128,7 +128,8 @@ def get_report_pdf(doc, consolidated=True):
 	if not bool(statement_dict):
 		return False
 	elif consolidated:
-		result = "".join(list(statement_dict.values()))
+		delimiter = '<div style="page-break-before: always;"></div>' if doc.include_break else ""
+		result = delimiter.join(list(statement_dict.values()))
 		return get_pdf(result, {"orientation": doc.orientation})
 	else:
 		for customer, statement_html in statement_dict.items():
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.py
```diff
@@ -240,8 +242,8 @@ def fetch_customers(customer_collection, collection_name, primary_mandatory):
 		if int(primary_mandatory):
 			if primary_email == "":
 				continue
-		elif (billing_email == "") and (primary_email == ""):
-			continue
+		# elif (billing_email == "") and (primary_email == ""):
```
deepeshgarg007: Why is this commented out?

rtdany10: This line of code would continue the iteration if no email was found. Email is necessary only when I send it via email, not when I download the consolidated PDF. Hence, removed this code and added validation on email sending function

deepeshgarg007: We usually don't leave any commented out code, remove if its not needed

rtdany10: Updated :)

rtdany10:
@mergifyo backport-version-13

rtdany10:

rtdany10:
@deepeshgarg007 could you back port this as well, please?

PR #32625, opened 2022-10-17
feat: Editable Sales Invoice

deepeshgarg007:
Post introduction of E-Invoicing in India and other countries some invoice-wise details have to be reported on the government portals in real-time, one of these details is also the cancellation of invoices in case there are any.

There are certain internal accounting details like account head or cost center that don't have to be reported and are only used for internal purposes. But updating details like these will require the user to cancel and amend the invoice. Since canceling and amending the invoice changes the invoice number, E-invoice also has to be canceled and regenerated which is kind of an unnecessary hassle.

So to make the user's life a bit easier we have allowed some fields mentioned above to be editable even after submission, a GIF of how this works can be seen below

Editable Invoice

This is more or less like canceling and resubmitting the invoice but without actually canceling it or changing the invoice number.

An exhaustive list of fields that has an accounting impact and are allowed to be edited post submit is as follows

1. Additional Discount Account (Sales Invoice), Unrealised P/L Account (for internal transfers via invoice)

3. Cash/Bank Account, Account for Change Amount, Write Off Account, Loyalty Redemption Account (Sales Invoice, POS enabled)
 
5. Accounting Dimensions at parent level (only cost center has by default “allow on submit” enabled, for custom dimensions it will have to be enabled in specific custom fields)
 
7. Income, COGS, and Discount Account (Sales Invoice Item)
 
9. Accounting Dimensions (Sales Invoice Item)

11. Account Head and Dimensions (Sales Taxes and Charges Table)

13. Is Opening Entry “Yes” and “No” select field

`no-docs`


ruthra-kumar:
@deepeshgarg007 Awesome work on developing this feature.

There are 2 minor and 1 critical bug while handling `docstatus` change. Once fixed, feature should be ready.

Minor:
1. `Repost Accounting Entries` is still visible for cancelled documents. It doesn't post anything. But, i think its better to hide this button. 
2. Amending [1] still has the `Repost Accouting Entries` button visible. Trigger does a repost which is not required.

Critical:
3. Triggering repost on the Amended document from [2] posts ledger entries while Invoice is still in `draft` status. If you proceed and Submit, now there are duplicate ledger entries.

[1] - 

https://user-images.githubusercontent.com/3272205/199954323-08c73f61-6e41-491b-ab01-8df7e8b9fe57.mov

[2] - 

https://user-images.githubusercontent.com/3272205/199954564-249d39a7-da62-43f7-b006-57fb8724e7b4.mov


[3] - 

https://user-images.githubusercontent.com/3272205/199954612-cedab69c-ba37-43a1-a370-06d26b2b545f.mov



Resetting `repost_required` on cancellation should fix the above issues.

ruthra-kumar:
Duplicating an Invoice that is yet do a repost, copies the `repost_required` flag as well. Triggering the repost on the copied doc has the same effect as [3] from the earlier comment. Resetting`repost_required` on `db_insert` should probably fix this.

https://user-images.githubusercontent.com/3272205/200231246-ab7e3061-bc17-4389-ab73-4ee7ea7b7e19.mov



Reposting cancels Deferred Income - Critical:
1. Make Invoice with deferred Revenue
2. Process deferred revenue for few months.
3. Update invoice and do a Repost.
This cancels the revenue realized through deferred accounting.

[1] - 

https://user-images.githubusercontent.com/3272205/200231335-eb296167-1a37-4692-ab92-0de4ecd64264.mov

[2], [3] - 

https://user-images.githubusercontent.com/3272205/200231370-2604a08e-467e-44a0-ad68-f7f15ff6a9ec.mov

deepeshgarg007:
Unrelated linter check failing

ruthra-kumar:
Works fine on Basic `docstatus` changes and Deferred Accounting.

ruthra-kumar (approved):

czed666:
how to get these changes?

PR #32629, opened 2022-10-17
perf: cache barcode scan result

dj12djdjs:
- Added cached barcode result up to 2 minuets to reduce latency between continues (same item) scans.

Locally this saves ~2-5ms per scan.
Remote server: Untested.

The time saving really isn't that much, but it's something.

### DIFF
--- a/erpnext/stock/utils.py
+++ b/erpnext/stock/utils.py
@@ -553,6 +553,15 @@ def check_pending_reposting(posting_date: str, throw_error: bool = True) -> bool
 
 @frappe.whitelist()
 def scan_barcode(search_value: str) -> Dict[str, Optional[str]]:
+	def set_cache(data: dict):
+		frappe.cache().set(f"erpnext:barcode_scan:{search_value}", json.dumps(data), 120)
+
+	def get_cache() -> Optional[Dict[str, Optional[str]]]:
+		if data := frappe.cache().get(f"erpnext:barcode_scan:{search_value}"):
+			return json.loads(data)
+
+	if scan_data := get_cache():
+		return scan_data
 
 	# search barcode no
 	barcode_data = frappe.db.get_value(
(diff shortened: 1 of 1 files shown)


ankush:
> The time saving really isn't that much, but it's something.

Is it anything? The columns are indexed so it's not like the DB query is slow. 😬 


image




A typical scan request is completed within 3 milliseconds!

erpnext/stock/utils.py
```diff
@@ -552,7 +554,16 @@ def check_pending_reposting(posting_date: str, throw_error: bool = True) -> bool
 
 
 @frappe.whitelist()
-def scan_barcode(search_value: str) -> Dict[str, Optional[str]]:
+def scan_barcode(search_value: str) -> BarcodeScanResult:
+	def set_cache(data: BarcodeScanResult):
+		frappe.cache().set(f"erpnext:barcode_scan:{search_value}", json.dumps(data), 120)
+
+	def get_cache() -> Optional[BarcodeScanResult]:
+		if data := frappe.cache().get(f"erpnext:barcode_scan:{search_value}"):
```
ankush, line 562: `redis.set` / `redis.get` shouldn't be used (they don't respect multitenancy) use set_value / get_value instead.

ankush (changes requested):

ankush:
Benchmarked `scan_barcode` using get_value/set_value and redis only (no local caching):

- Baseline: 1.2 ms
- With this change: 0.03 ms 🥴 

So _it's something_! Though I recon biggest time consumer here would be the latency itself, so the locality of the server matters more than anything. Can't beat speed of light 🚀 

If it's for repeat scan of same items only then even better option would be limited client side caching.

erpnext/stock/utils.py
```diff
@@ -552,7 +554,16 @@ def check_pending_reposting(posting_date: str, throw_error: bool = True) -> bool
...
+			return json.loads(data)
```
ankush:
```suggestion
		frappe.cache().set_value(f"erpnext:barcode_scan:{search_value}", data, expires_in_sec=120)

	def get_cache() -> Optional[BarcodeScanResult]:
		if data := frappe.cache().get_value(f"erpnext:barcode_scan:{search_value}"):
			return data
```

get/set don't respect multitenancy.

You also dont need to handle serializing/deserializing with set_value/get_value.

ankush (changes requested):

PR #32635, opened 2022-10-18
fix: Integrity Error on PLE while submitting sales invoice

ruthra-kumar:
Avoid updating outstanding amount using PLE on Receivable/Payable Account heads that doesn't have `party_type` and  `party` field value supplied.
Screenshot 2022-10-18 at 5 35 09 PM


ruthra-kumar:

PR #32639, opened 2022-10-19
fix: Pricing rule item group consider UOM

maharshivpatel:
Same as #32566.

Pricing rule's apply_on_field == "item_group" didn't check for  UOM. I have added the required checks.


deepeshgarg007:

PR #32641, opened 2022-10-19
fix: allow to create Sales Order from expired Quotation

barredterra:
The company should always be able to accept an order, even if the **Quotation** is expired.

Let's assume, the customer orders a couple days "too late". What are we supposed to tell them? Sorry, we'd love to take your order, but the quotation is already expired so we technically cannot serve you? 😅

Legally speaking, the validity date gives us the option, _but not the obligation_, to refuse the order.


ankush (approved):
Makes sense. This isn't compliance, it's just about processes that indidual org has.

PR #32692, opened 2022-10-24
fix: Clear invoice table post importing invoices

### LINKED ISSUE #32326: A few Bugs on Opening Invoice Creation Tool Tool
### Information about bug

One, When upload button is used, the system does not capture party and amount, despite them being on the import template. The go around is to use the Data Import Tool, which imports the data correctly.

Two, and this is the major issue, after invoices are created from the tool, they are not deleted, and the Create Invoices button is still active, means a user can click on the button, and end up with duplicate opening invoices.

### Module

accounts

### Version

ERPNext: v14.1.0 (HEAD)
Frappe Framework: v14.6.0 (HEAD)

### Installation method

FrappeCloud

_No response_

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/32326


deepeshgarg007:

PR #32711, opened 2022-10-25
Optional Hungarian COA for microenterprises #32688

backendoadmin:
#32688


backendoadmin:
Some account's name still too long.

Issue #31415, opened 2022-06-09
Discount and Margin section open-close issue


nidhipurohit11:
## Description of the issue - 
"Discount and Margin" section once opened, gets closed automatically again. Also when entering any percentage/amount then it gets closed immediately, so to see the change have to open the section again.

## Steps to reproduce the issue

1. Ensure that item price exists for an item (if not then add from Item Price list doctype)
2. Open any form like quotation/sales order and select this item, view its details from the child table 
3. Below the price list field, a section should appear "Discount and Margin"
4. Click this section.

### Observed result
- Clicking this section closes it back immediately, also once opened then entering %/amount closes it back.

### Expected result
- Only a single click to open it should keep this section open, and making changes in the % or amount field should reflect without closing the section.

https://user-images.githubusercontent.com/82337328/172848117-7670670f-d198-4b1c-afd9-3f18f40d5012.mov

shariquerik:
@nidhipurohit11 @Komal-Saraf0609 
> Clicking this section closes it back immediately

I am not able to reproduce this

> once opened then entering %/amount closes it back

This I am able to reproduce.

Komal-Saraf0609:
@shariquerik the first issue is replicable on my end as well. We can discuss if you can connect.

PR #32717, opened 2022-10-26
fix: Add condition for discount section collapse

### LINKED ISSUE #24847: annoying fold of discount and margin section
when setting discounts and margin in transaction (quotation, sales order, invoice) inside the item list (detail) there is an annoying fold when entering values:
margin_fold

Section should not fold after input.

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/24847 https://github.com/frappe/erpnext/issues/31415


deepeshgarg007:

PR #32567, opened 2022-10-11
fix: deduplicate Payment Ledger on Item value repost

ruthra-kumar:
Upon Item value repost, duplicate entries are created in Payment Ledger which affect the Invoice outstanding. Fixing in this change.

1. Invoice: Screenshot 2022-10-11 at 5 03 20 PM
2. Backdated stock entry: Screenshot 2022-10-11 at 2 58 11 PM
3. Repost triggered from [2]: Screenshot 2022-10-11 at 2 58 38 PM
4. Duplicate PLE entries: Screenshot 2022-10-11 at 2 58 01 PM
5.  Incorrect invoice outstanding: Screenshot 2022-10-11 at 4 55 40 PM

PR #32742, opened 2022-10-28
feat: Tool to repost PLE manually

ruthra-kumar:
## Repost PLE
Tool is primarily intended for manual correction of [Duplicate Payment Ledger entries on Item Value Repost](https://github.com/frappe/erpnext/pull/32567)

## Usage:
1. Use Filters to Bulk select or Add Vouchers manually and Submit 
Screenshot 2022-11-08 at 10 53 32 AM

2. Once submitted, a background job will be triggered. 
Screenshot 2022-11-08 at 10 53 41 AM

3. After successful run,`repost_status` will be updated 
Screenshot 2022-11-08 at 10 53 47 AM

`no-docs`

### DIFF
--- a/erpnext/accounts/doctype/repost_payment_ledger/repost_payment_ledger.js
+++ b/erpnext/accounts/doctype/repost_payment_ledger/repost_payment_ledger.js
@@ -1,0 +1,53 @@
+// Copyright (c) 2022, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on('Repost Payment Ledger', {
+	setup: function(frm) {
+		frm.set_query("voucher_type", () => {
+			return {
+				filters: {
+					name: ['in', ['Purchase Invoice', 'Sales Invoice', 'Payment Entry', 'Journal Entry']]
+				}
+			};
+		});
+
+		frm.fields_dict['repost_vouchers'].grid.get_field('voucher_type').get_query = function(doc) {
+			return {
+				filters: {
+					name: ['in', ['Purchase Invoice', 'Sales Invoice', 'Payment Entry', 'Journal Entry']]
+				}
+			}
+		}
+
+		frm.fields_dict['repost_vouchers'].grid.get_field('voucher_no').get_query = function(doc) {
+			if (doc.company) {
+				return {
+					filters: {
+						company: doc.company,
+						docstatus: 1
+					}
+				}
+			}
+		}
+
+	},
+	refresh: function(frm) {
+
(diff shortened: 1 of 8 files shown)


erpnext/accounts/doctype/repost_payment_ledger/repost_payment_ledger.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class RepostPaymentLedger(Document):
+	def __init__(self, *args, **kwargs):
+		super(RepostPaymentLedger, self).__init__(*args, **kwargs)
+		self.vouchers = []
+
+	def validate(self):
+		if not self.company:
+			frappe.throw(_("Company is Mandatory"))
```
deepeshgarg007: Needs checking again?
Both of these seem to have `reqd` as 1

erpnext/accounts/doctype/repost_payment_ledger/repost_payment_ledger.js
```diff
@@ -0,0 +1,53 @@
...
+						docstatus: 1
+					}
+				}
+			}
+		}
+
+	},
+	refresh: function(frm) {
+
+		if (frm.doc.docstatus==1 && ['Queued', 'Failed'].find(x => x == frm.doc.repost_status)) {
```
deepeshgarg007, line 36: This seems like an unnecessary click/step for the user I guess. The job should be by default Queued on submit and the status should be updated as "Queued". By default, the "Queued" status might be misleading to the user. Showing a progress bar will also be a good to have to show the user the realtime status of the job

ruthra-kumar: I've changed it to trigger job on submit.

As repost happens in single transaction, Progress bar is not possible.

erpnext/accounts/doctype/repost_payment_ledger/repost_payment_ledger.py
ruthra-kumar: Removed

deepeshgarg007:

dawoodjee:
Which ERPNext versions has this been merged to?

I'm on v14.11 and its not working

First the function `is_job_queued` was absent in the latest stable Frappe 14.20. 

`if not frappe.utils.background_jobs.is_job_queued(job_name):`

After adding it, it just says its running in the background but not doing anything. I'll be debugging as I wait for a response

ruthra-kumar:
> 

@dawoodjee This issue is fixed in v14.13.0

PR #32768, opened 2022-10-30
fix: Budget validation for main cost center

### LINKED ISSUE #32273: Cost center allocation not working
### Information about bug

Previously in version 13 it is called "Distributed Cost Center" which is a feature in cost center doctype, now in version 14 it has been separate out and called "Cost center allocation" which is a diff doctype.

So, it works in version 13 when it is "Distributed Cost Center", but it not working in version 14 "Cost center allocation".

Step to reproduce:-
- 3 cost center which "is group" does not check. 
- Create "Cost center allocation", put 1 of the 3 created cost center as main, other 2 as sub cost center, set as 80% and 20%.
- Create Budget for the Main Cost center.
- Go to "budget variance report" to check for result

### Module

accounts

### Version

ERPNext: v14.0.3 (version-14)
Frappe Framework: v14.5.0 (version-14)

### Installation method

manual install

_No response_

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/32273

### DIFF
--- a/erpnext/accounts/doctype/budget/budget.py
+++ b/erpnext/accounts/doctype/budget/budget.py
@@ -107,7 +107,7 @@ class Budget(Document):
 		self.naming_series = f"{{{frappe.scrub(self.budget_against)}}}./.{self.fiscal_year}/.###"
 
 
-def validate_expense_against_budget(args):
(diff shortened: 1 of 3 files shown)


deepeshgarg007:

PR #32776, opened 2022-10-31
perf: use `get_cached_value` instead of `db.get_value` in controllers

### LINKED ISSUE #32720: perf: Use get_cached_value instead of db.get_value for validate_enabled_taxes_and_charges
When you have a doctype created that uses taxes and charges, this function fires for each line item. e.g. Delivery Note. The validate_enabled_taxes_and_charges function does a DB call on a master data multiple times instead of leveraging the get_cached_value function. This causes unnecessary system utilization.

Using get_cached_value will reduce the number of SQL calls.

Expected solution:
Change from:
```if frappe.db.get_value(taxes_and_charges_doctype, self.taxes_and_charges, "disabled"):```
to
```if frappe.get_cached_value(taxes_and_charges_doctype, self.taxes_and_charges, "disabled"):```

### PR DESCRIPTION
**Pull request by DaizyModi**
Replaced `db.get_value` with `get_cached_value` for masters doctype in **Controllers** file.

### Before
```
In [2]: %timeit doc.validate_enabled_taxes_and_charges()
528 µs ± 3.3 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### After (without local cache)
```
In [3]: %timeit doc.validate_enabled_taxes_and_charges()
5.85 µs ± 64.2 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
Closes #32720

### DIFF
--- a/erpnext/controllers/stock_controller.py
+++ b/erpnext/controllers/stock_controller.py
@@ -57,7 +57,7 @@ class StockController(AccountsController):
 			make_reverse_gl_entries(voucher_type=self.doctype, voucher_no=self.name)
 
 		provisional_accounting_for_non_stock_items = cint(
-			frappe.db.get_value(
+			frappe.get_cached_value(
 				"Company", self.company, "enable_provisional_accounting_for_non_stock_items"
 			)
 		)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: erpnext/controllers/stock_controller.py [hunk 1]
```diff
@@ -449,14 +451,15 @@ def update_inventory_dimensions(self, row, sl_dict) -> None:
 
 				# Get value based on doctype name
 				if not sl_dict.get(dimension.target_fieldname):
-					fieldname = frappe.get_cached_value(
-						"DocField", {"parent": self.doctype, "options": dimension.fetch_from_parent}, "fieldname"
-					)
```
sagarvora: no newline after block begins.

ankush:
@DaizyModi / @sagarvora make sure the numbers are without using local.cache{} between loops that timit executes (probably by writing simple test wrapper)

Otherwise it will be grossly misleading. I've seen simple queries + db.value_cache often being faster than get_cached_value. 

Also db.value_cache is more reliable from cache eviction pov.

deepeshgarg007:

PR #32777, opened 2022-10-31
fix: Reset advance paid amount on Order cancel and amend

deepeshgarg007:
Advance paid amount doesn't get reset when an Order against which advance has been paid is cancelled without cancelling the payment entry


deepeshgarg007:

PR #32779, opened 2022-10-31
fix: Mode of payment for returns in POS Sales Invoice

deepeshgarg007:
While making a return for a Sales Invoice (with Is POS check enabled) the mode of payment always used to reset to default one on saving overriding user entered mode of payment.

Modified the logic to do that only for the pending amount


deepeshgarg007:

PR #32790, opened 2022-11-01
fix: pro_rata_amount calculation in assets tests

anandbaburajan:
This is to fix two of the [failing](https://github.com/frappe/erpnext/actions/runs/3363147889/jobs/5584235627) tests in the assets module due to incorrect pro_rata_amount calculation.
Thanks to @rohitwaghchaure for letting me know about the failing tests.

### DIFF
--- a/erpnext/assets/doctype/asset/test_asset.py
+++ b/erpnext/assets/doctype/asset/test_asset.py
@@ -221,7 +221,7 @@ class TestAsset(AssetSetup):
 			asset.precision("gross_purchase_amount"),
 		)
 		pro_rata_amount, _, _ = asset.get_pro_rata_amt(
-			asset.finance_books[0], 9000, add_months(get_last_day(purchase_date), 1), date
+			asset.finance_books[0], 9000, get_last_day(add_months(purchase_date, 1)), date
 		)
 		pro_rata_amount = flt(pro_rata_amount, asset.precision("gross_purchase_amount"))
 		self.assertEquals(accumulated_depr_amount, 18000.00 + pro_rata_amount)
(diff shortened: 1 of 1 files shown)


erpnext/assets/doctype/asset/test_asset.py
```diff
@@ -221,7 +221,7 @@ def test_scrap_asset(self):
 			asset.precision("gross_purchase_amount"),
 		)
 		pro_rata_amount, _, _ = asset.get_pro_rata_amt(
-			asset.finance_books[0], 9000, add_months(get_last_day(purchase_date), 1), date
+			asset.finance_books[0], 9000, get_last_day(add_months(purchase_date, 1)), date
```
AnandBaburajan, line 224: The previous way worked only for months with 30 days and February, not for months with 31 days.

erpnext/assets/doctype/asset/test_asset.py
```diff
@@ -221,7 +221,7 @@ def test_scrap_asset(self):
 			asset.precision("gross_purchase_amount"),
 		)
 		pro_rata_amount, _, _ = asset.get_pro_rata_amt(
-			asset.finance_books[0], 9000, add_months(get_last_day(purchase_date), 1), date
```
AnandBaburajan, line 224: I wanted to get the last day of the month, but the previous method doesn't work for months with 31 days.

anandbaburajan, line 224: This essentially tried to get the last day of the previous month but doesn't work if the current month has 30 days or is February.

PR #32794, opened 2022-11-01
fix: Disable tax included prices for internal transfers

deepeshgarg007:
Internal transfers via the Sales and Purchase Transactions system by default force valuation rate for items.

This is not an issue while making Sales, but while receiving the item via a receipt or an invoice the rate is fetched from the corresponding outward document (Delivery Note or Sales Invoice). And including tax in that same rate causes a drop in the valuation

Hence to avoid this, disabling tax included rates for internal transfers


deepeshgarg007:

PR #32799, opened 2022-11-01
fix: Scan Barcode UX

rohitwaghchaure:
## **Issue**

When user scan the Barcode in the Scan Barcode field, if item already exists with the quantity then system increase the quantity. 

For example

In Purchase Order we have set the quantity as 100, so when user creates the Purchase Receipt against the purchase order system fetch the items from the purchase order along with the Quantity. Here if User scan barcode system increase the QTY  field value. Ideally system should provide provision to capture the Scanned Qty but we don't have this field.

#### **GIF to show the issue in Purchase Receipt**
issue-barcode-scan

## **Solution**

Once user scan the barcode system will open the Dialog and we'll capture the Scanned Qty. The scanned qty will be update in the child table. If the "Quantity to Scan" is more than the "Scanned Qty" then system will add the remaining qty in another row (Useful in the Batch case).

after-fix-barcode-scan

### DIFF
--- a/erpnext/public/js/utils/barcode_scanner.js
+++ b/erpnext/public/js/utils/barcode_scanner.js
@@ -103,6 +103,7 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 				row = frappe.model.add_child(this.frm.doc, cur_grid.doctype, this.items_table_name);
 				// trigger any row add triggers defined on child table.
 				this.frm.script_manager.trigger(`${this.items_table_name}_add`, row.doctype, row.name);
+				this.frm.has_items = false;
 			}
 
 			if (this.is_duplicate_serial_no(row, serial_no)) {
@@ -157,8 +158,10 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 				frappe.prompt(__("Please enter quantity for item {0}", [item_code]), ({value}) => {
 					increment(value).then((value) => resolve(value));
 				});
-			} else {
+			} else if (this.frm.has_items) {
 				this.prepare_item_for_scan(row, item_code, barcode, batch_no, serial_no);
+			} else {
+				increment().then((value) => resolve(value));
 			}
 		});
 	}
@@ -207,7 +210,6 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 						return;
 					}
 
-					debugger
 					if (e.target.value) {
 						this.scan_api_call(e.target.value, (r) => {
 							if (r.message) {
@@ -280,7 +282,7 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 	}
 
 	update_dialog_values(scanned_item, r) {
-		const {item_code, barcode, batch_no, serial_no, uom} = r.message;
+		const {item_code, barcode, batch_no, serial_no} = r.message;
 
 		this.dialog.set_value("barcode_scanner", "");
 		if (item_code === scanned_item &&
@@ -357,7 +359,6 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 	}
 
 	async set_batch_no(row, batch_no) {
-		debugger
 		if (batch_no && frappe.meta.has_field(row.doctype, this.batch_no_field)) {
 			await frappe.model.set_value(row.doctype, row.name, this.batch_no_field, batch_no);
 		}
--- a/erpnext/public/js/utils/barcode_scanner.js
+++ b/erpnext/public/js/utils/barcode_scanner.js
@@ -47,32 +47,38 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
 				return;
 			}
 
-			frappe
-				.call({
-					method: this.scan_api,
-					args: {
-						search_value: input,
-					},
-				})
-				.then((r) => {
-					const data = r && r.message;
-					if (!data || Object.keys(data).length === 0) {
-						this.show_alert(__("Cannot find Item with this Barcode"), "red");
-						this.clean_up();
-						this.play_fail_sound();
-						reject();
-						return;
-					}
+			this.scan_api_call(input, (r) => {
+				const data = r && r.message;
+				if (!data || Object.keys(data).length === 0) {
+					this.show_alert(__("Cannot find Item with this Barcode"), "red");
+					this.clean_up();
+					this.play_fail_sound();
+					reject();
+					return;
+				}
 
-					me.update_table(data).then(row => {
-						this.play_success_sound();
-						resolve(row);
-					}).catch(() => {
-						this.play_fail_sound();
-						reject();
-					});
+				me.update_table(data).then(row => {
+					this.play_success_sound();
+					resolve(row);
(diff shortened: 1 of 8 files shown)


erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -150,11 +158,184 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
...
+				fieldtype: "Data",
+				fieldname: "barcode_scanner",
+				options: "Barcode",
+				label: __("Scan Barcode"),
+				onchange: (e) => {
+					if (!e) {
+						return;
+					}
+
+					debugger
```
dj12djdjs: remove break point
```suggestion
```

erpnext/public/js/utils/barcode_scanner.js
```diff
@@ -150,11 +158,184 @@ erpnext.utils.BarcodeScanner = class BarcodeScanner {
...
+				read_only: 1,
+				hidden: 1
+			});
+		}
+
+		return fields;
+	}
+
+	update_dialog_values(scanned_item, r) {
+		const {item_code, barcode, batch_no, serial_no, uom} = r.message;
```
dj12djdjs: `uom` isn't being used in this scope?

dj12djdjs (changes requested):

dj12djdjs:
I like your approach for this. There is one thing I think needs to be validated. If the user is scanning into the purchase receipt, if there are any items missing from the scan. The user should be notified instead of just posting to stock. Example:


pr-scan-missing-qty


```patch
diff --git a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
index f85c478..1da4c2f 100644
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
@@ -5,6 +5,7 @@
 import frappe
 from frappe import _, throw
 from frappe.desk.notifications import clear_doctype_notifications
+from frappe.model.docstatus import DocStatus
 from frappe.model.mapper import get_mapped_doc
 from frappe.utils import cint, flt, getdate, nowdate
 
@@ -117,6 +118,7 @@ class PurchaseReceipt(BuyingController):
 		self.validate_uom_is_integer("stock_uom", "stock_qty")
 		self.validate_cwip_accounts()
 		self.validate_provisional_expense_account()
+		self.validate_scanned_qty()
 
 		self.check_on_hold_or_closed_status()
 
@@ -128,6 +130,21 @@ class PurchaseReceipt(BuyingController):
 		self.reset_default_field_value("rejected_warehouse", "items", "rejected_warehouse")
 		self.reset_default_field_value("set_from_warehouse", "items", "from_warehouse")
 
+	def validate_scanned_qty(self):
+		if self.docstatus == DocStatus.submitted():
+			raise_error = False
+			if any([d.has_item_scanned for d in self.items]):
+				for item in self.items:
+					if not item.has_item_scanned:
+						raise_error = True
+						frappe.msgprint(
+							_("Row {0}: {1} hasn't been scanned.").format(item.idx, item.item_code),
+							_("Item(s) not Scanned"),
+							indicator="red",
+						)
+			if raise_error:
+				raise frappe.ValidationError
+
 	def validate_cwip_accounts(self):
 		for item in self.get("items"):
 			if item.is_fixed_asset and is_cwip_accounting_enabled(item.asset_category):
```

erpnext/public/js/utils/barcode_scanner.js
rohitwaghchaure: Fixed

erpnext/public/js/utils/barcode_scanner.js
rohitwaghchaure: Fixed

rohitwaghchaure:
> I like your approach for this. There is one thing I think needs to be validated. If the user is scanning into the purchase receipt, if there are any items missing from the scan. The user should be notified instead of just posting to stock. Example:
> 
> pr-scan-missing-qty [ pr-scan-missing-qty ](https://user-images.githubusercontent.com/29856401/199537298-349f3364-be6a-4e57-9bfd-bd652510f7aa.gif) [ ](https://user-images.githubusercontent.com/29856401/199537298-349f3364-be6a-4e57-9bfd-bd652510f7aa.gif)
> 
> ```diff
> diff --git a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
> index f85c478..1da4c2f 100644
> --- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
> +++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.py
> @@ -5,6 +5,7 @@
>  import frappe
>  from frappe import _, throw
>  from frappe.desk.notifications import clear_doctype_notifications
> +from frappe.model.docstatus import DocStatus
>  from frappe.model.mapper import get_mapped_doc
>  from frappe.utils import cint, flt, getdate, nowdate
>  
> @@ -117,6 +118,7 @@ class PurchaseReceipt(BuyingController):
>  		self.validate_uom_is_integer("stock_uom", "stock_qty")
>  		self.validate_cwip_accounts()
>  		self.validate_provisional_expense_account()
> +		self.validate_scanned_qty()
>  
>  		self.check_on_hold_or_closed_status()
>  
> @@ -128,6 +130,21 @@ class PurchaseReceipt(BuyingController):
>  		self.reset_default_field_value("rejected_warehouse", "items", "rejected_warehouse")
>  		self.reset_default_field_value("set_from_warehouse", "items", "from_warehouse")
>  
> +	def validate_scanned_qty(self):
> +		if self.docstatus == DocStatus.submitted():
> +			raise_error = False
> +			if any([d.has_item_scanned for d in self.items]):
> +				for item in self.items:
> +					if not item.has_item_scanned:
> +						raise_error = True
> +						frappe.msgprint(
> +							_("Row {0}: {1} hasn't been scanned.").format(item.idx, item.item_code),
> +							_("Item(s) not Scanned"),
> +							indicator="red",
> +						)
> +			if raise_error:
> +				raise frappe.ValidationError
> +
>  	def validate_cwip_accounts(self):
>  		for item in self.get("items"):
>  			if item.is_fixed_asset and is_cwip_accounting_enabled(item.asset_category):
> ```


This no need to put in the core.. can be achieve using Server Script / Client Script

dj12djdjs:
> This no need to put in the core.. can be achieve using Server Script / Client Script

@rohitwaghchaure Sounds good. I tested this PR. This is a very nice improvement to barcode scanning flow!

rohitwaghchaure:
> > This no need to put in the core.. can be achieve using Server Script / Client Script
> 
> @rohitwaghchaure Sounds good. I tested this PR. This is a very nice improvement to barcode scanning flow!

Thanks a lot for giving your valuable feedback.

TurkerTunali:
I think that, this feature should have an "Enabled" check in the Stock Settings.

PR #32816, opened 2022-11-02
fix: update advance paid in SO/PO from Payment Ledger

ruthra-kumar:
`advance_paid` in Sales and Purchase Order will be calculated and updated from Payment Ledger


ruthra-kumar:

PR #32844, opened 2022-11-04
fix: Auto advance allocation against partial invoices

deepeshgarg007:
Incorrect amount gets allocated in invoices while partially billing a Sales Order

Typical scenario when this is seen is
- User makes a Sales Order and makes an Advance Payment against it
- Make a standalone advance payment against the same customer
- Partially bill the created Sales Order

The advances allocated are incorrect

Before:
image

After:
image


deepeshgarg007:

Issue #27297, opened 2021-09-01
Gross profit report uses last known valuation rate instead of valuation rate at time of invoice.


dj12djdjs:
## Description of the issue
Left side is Gross Profit Report
Right side is Stock Ledger Report

The records are lined up exactly with each other. 

Where the rows are highlighed yellow you can see we received stock at a lower price then previously. Resulting in the valuation rate declining. The problem is gross profit report is using the latest known valuation rate for that warehouse instead of the valuation rate known at time of sale.

![image](https://user-images.githubusercontent.com/29856401/131684299-b23684a0-061b-4c01-bc52-7c2d0c11e2f5.png)

```
Frappe Framework: v13.9.0
ERPNext: v13.9.2
```

PR #32866, opened 2022-11-06
fix: incorrect buying amount on Gross Profit

### LINKED ISSUE #31786: [Gross Profit Report] Wrong calculation of buying amount
### Information about bug

Purchase Invoice:
image

Sales Invoice:
image

Delivery Note:
image

So far,
We know that buying amount is 900, the selling amount is 100,000. So, the gross profit is 100,000 - 900 = 99,100.
But as per ERPNext's Gross Profit report, it is only 99,000(attached video for ref).

https://user-images.githubusercontent.com/52111700/183245282-30fd91c9-38aa-4c76-9e6d-630fa32550dd.mp4

### Module

accounts

### Version

ERPNext: v13.36.2 (HEAD)
Frappe Framework: v13.36.3 (HEAD)

### Installation method

FrappeCloud

_No response_

ruthra-kumar:
Incorrect buying amount for Invoices(`update_stock` unchecked and no `warehouse`) but has Delivery Notes.

Steps:
1. Create `Material Receipt` with same item for 2 different rates 
Screenshot 2022-11-08 at 7 51 48 AM

2. Make Sales Invoice for one Item without updating the stock 
Screenshot 2022-11-08 at 7 50 30 AM

3. Make Delivery Note for [2] 
Screenshot 2022-11-08 at 7 50 40 AM

Before:
Buying amount: 150(taken as average of last transaction)
Screenshot 2022-11-08 at 7 54 15 AM

After:
Buying amount: 100
Screenshot 2022-11-08 at 7 54 43 AM

fixes: #31786, #27297, #27400

### DIFF
--- a/erpnext/accounts/report/gross_profit/gross_profit.py
+++ b/erpnext/accounts/report/gross_profit/gross_profit.py
@@ -3,7 +3,8 @@
 
 
 import frappe
-from frappe import _, scrub
+from frappe import _, qb, scrub
+from frappe.query_builder import Order
 from frappe.utils import cint, flt, formatdate
 
 from erpnext.controllers.queries import get_match_cond
@@ -398,6 +399,7 @@ class GrossProfitGenerator(object):
 		self.average_buying_rate = {}
 		self.filters = frappe._dict(filters)
 		self.load_invoice_items()
+		self.get_delivery_notes()
 
 		if filters.group_by == "Invoice":
 			self.group_items_by_invoice()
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/gross_profit/gross_profit.py
```diff
@@ -477,6 +479,17 @@ def process(self):
 		if self.grouped:
 			self.get_average_rate_based_on_group_by()
 
+	def has_delivery_notes(self, row):
+		for x in self.delivery_notes:
```
deepeshgarg007: It will go through all the delivery notes and items for each and every Sales Invoice/Row. Could be a heavy operation for users who have a lot of Delivery Notes. Can we make some sort of map to check this?

ruthra-kumar: Created a Map.

deepeshgarg007:

rtdany10:
@deepeshgarg007 @ruthra-kumar will this be backported to v13?

Issue #32818, opened 2022-11-02
Warehouse wise item balance and Ageing


maveezmh:
### Information about bug

Some of the reports in ERP including warehouse wise item balance and age. the column width of the report is not as per the title. its getting adjusted as per the data below. this is creating lots of concerns. everytime the report is opened we have to increase the column width. its a huge issue when you have more than 10 warehouses and everytime you need to increase width of each column to see warehouse name.

Refer to image below 
![image](https://user-images.githubusercontent.com/112148467/199423197-fb233a9d-1298-4129-b71b-9c8bff60b3d6.png)

### Module

stock

### Version

ERPNext: v13.40.3 (HEAD)
Frappe Framework: v13.42.0 (HEAD)

Same issue in V14

### Installation method

FrappeCloud

PR #32874, opened 2022-11-07
fix: Increase columns width in Warehouse wise Item Balance Age and Value

Solufyin:
**Version**

ERPNext: v14.5.1 (version-14)
Frappe Framework: v14.14.2 (version-14)
___
Issue: #32818
___
**Before:** 
- The column's width does not correctly set as the column's title name. If you increase the column for some time after then reload the system then happens to reset the column's width size and set it again manually.

image

**After:**
- When entering the report then will at least show starting title of the column of the warehouse. 
- Also set the increase in the columns width of Value, Age, and Total Qty.

image

Thank You!


deepeshgarg007:

PR #32883, opened 2022-11-07
fix: Project filter in timesheet

### LINKED ISSUE #23306: Project List doesn't get updated in Timesheet
Not sure why this is happening but in pull-down list of "Project" in a Timesheet entry the list if projects is not being updated. It's stuck at some older point in time. If you use the pull-down to "Create a New Project" it will show up for that entry only. Also, closing a project does not stop it from showing up in this area.

I can't share screenshots due to projects being proprietary, and it looks like the demo or beta sites are not up so I don't have any data. Should be easy to replicate though.

Versions:
ERPNext: v12.11.2 (version-12)

ERPNext Support: v0.0.3 (master)

Frappe Framework: v12.9.1 (version-12)

Journeys: v0.0.1 (version-12)

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/23306

### DIFF
--- a/erpnext/manufacturing/doctype/workstation/workstation.py
+++ b/erpnext/manufacturing/doctype/workstation/workstation.py
@@ -100,9 +100,7 @@ def get_default_holiday_list():
 
 def check_if_within_operating_hours(workstation, operation, from_datetime, to_datetime):
 	if from_datetime and to_datetime:
-		if not cint(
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

DrZoidberg09:
@deepeshgarg007 Since this PR I cannot select projects anymore in my timesheet (without notable error message). There is just no project to select. Could you please check?

PR #32899, opened 2022-11-08
fix: conditions to show "Update Item"

barredterra:


melanieroedel:
The described functionality in the documentation is fairly complex. 

The documentation suggests that it is possible to 

1) update the quantities in the Sales Order as long as no Purchase Receipt is created against it, and 
2) update the prices in the Sales Order as long as no Purchase Invoice is created against it.

This is the behavior we would like to achieve. However, in `version-13` the "Update Items" button disappears if either one is created. 

See:
https://docs.erpnext.com/docs/v13/user/manual/en/buying/articles/amending-purchase-order-after-submit

Edit by @barredterra

barredterra:
Closing for now

PR #32900, opened 2022-11-08
fix: Exchange gain and loss booking on multi-currency invoice reconciliation

deepeshgarg007:
#### Issue 1

Exchange gain and loss were not getting booked (difference amount was not getting calculated) while allocating payments made using journal entry against invoices using payment reconciliation

#### Issue 2

While allocating payments made using payment entry against an invoice, the difference amount gets calculated only when the full payment amount is allocated against the invoice. In case the payment was a clubbed payment for multiple invoices and only a partial amount is allocated then the difference amount is not calculated.

Replace the current logic with the one used while allocating advances in Invoices. This logic works for most scenarios.

Some minor currency symbol fixes, the allocated account currency should be in party currency. Before it was shown in the company curreny as seen in the below screenshot

Screenshot 2022-11-08 at 12 12 27 PM

If there is a difference a dialog box will appear asking the user to select the difference account and also an intimation that a journal entry will be created for booking the exchange gain and loss account


ruthra-kumar (approved):

deepeshgarg007:

deepeshgarg007:

ruthra-kumar:

Issue #32813, opened 2022-11-02
Temporary Work Stations


SvbZ3r0:
**Is your feature request related to a problem? Please describe.**
Currently, workstations describe machines owned and operated by the company. However, this need not always be the case in reality. In my industry, we have contract workers who work for day wages. These are people who specialize in one particular operation. They have no permanent link to any one company, but find work in the industry everyday. 
In ERPNext terms, these people are workstations. They perform a single operation at a certain efficiency for a certain cost for the whole day. The problem is that a company might have need for 100 workers today, but just 20 tomorrow. Or, it might have need for 80 people, but only 50 show up.

**Describe the solution you'd like**
What I propose is to have a template workstation describing a worker, the operation they perform, their wage and working hours. Everyday, the manager in charge of the operation tells the system how many workers are present that day. Let's say 50, for the sake of this example. ERPNext creates 50 (virtual?) workstations, and adjusts the production plan accordingly. At the end of the day, these temporary workstations are deleted, and their output efficiency is attributed to the template workstation. Rinse and repeat the following day.

**Describe alternatives you've considered**
The current alternative is to edit the worstation everyday. The Production Capacity field would denote the number of workers. The cost fields are all set to actual cost x no. of workers. This process is tedious and error prone, but it works.

**Additional context**

SvbZ3r0:
The alternative solution I've described here doesn't actually work since we'll have to edit the time required for every operation whenever we change the workstation capacity.

PR #32911, opened 2022-11-09
feat: Workstation Type for BOM

rohitwaghchaure:
### Workstation Type 

Screenshot 2022-11-09 at 4 18 25 PM

### Workstation master
Screenshot 2022-11-09 at 4 59 23 PM

 

### Workstation Type selection in the BOM

Screenshot 2022-11-09 at 4 18 41 PM

### Auto set available workstation on job card based on Workstation Type 
image

#docs https://docs.erpnext.com/docs/v14/user/manual/en/manufacturing/workstation_type


s-aga-r:
https://github.com/frappe/erpnext/pull/32937 This will fix the linter for `job_card.py`

SvbZ3r0:
Hi @rohitwaghchaure, would it make sense to incorporate a feature like https://github.com/frappe/erpnext/issues/32813 within this PR?

PR #32913, opened 2022-11-09
fix: set stock UOM in args to ensure item price is fetched

sagarvora:
## Issue

If the following the conditions are satisfied, item price was not fetched earlier.

- the default Stock UOM is the items table is not the same as the Default Stock UOM for the selected item
- Item price is saved in stock UOM and converted to sales UOM

### Before

before item price fix

### After

after item price fix


s-aga-r (approved):
LGTM

PR #32921, opened 2022-11-10
feat: Add net value filter on PSOA

### LINKED ISSUE #30917: Process statement of account doesn't have " Show Net Values in Party Account " option
### Information about bug

If we are reconcile Sales Invoice with Credit Notes, then system generates a Journal Entry if the amount is not matching exactly.
It generated a Journal Entry with same party account is being debited and credited which is absolutely fine in terms of accounting. However, when we generate a General Ledger we have an option to avoid these kind of entries via clicking on  **"Show Net Values in Party Account"** and the statement of account to party can be generated from here and that is working fine.

At the same time if we use Process Statement of Account this  **Show Net Values in Party Account** is not available and again those Same Account Credit / Debit entries are showing up in the report which is misleading.

Can we have the checkbox available in Process Statement of Accounts as well?

### Module

accounts

### Version

Currently, I am using the below git repo -
Frappe version-13
ERPNext version-13

### Installation method

_No response_

```shell
There is no error but the value of General ledger for Party with Show Net Values in Party Account =1 and Process Statement of Account Report are not matching to each other.
```

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/30917

`no-docs`

### DIFF
--- a/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.json
+++ b/erpnext/accounts/doctype/process_statement_of_accounts/process_statement_of_accounts.json
@@ -23,6 +23,7 @@
   "fetch_customers",
   "column_break_6",
   "primary_mandatory",
+  "show_net_values_in_party_account",
   "column_break_17",
   "customers",
   "preferences",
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

PR #32930, opened 2022-11-11
fix: ambiguous 'cost_center' while using payment reconciliation

ruthra-kumar:
Steps to replicate:
1. Customize `Journal Entry` and create a new field `cost_center`
2. Open Payment Reconciliation tool
3. Apply Filter on `Cost Center` and trigger `Get Unreconciled Entries`
Screenshot 2022-11-11 at 10 01 17 AM

Maybe a Validation must be added in the Framework to restrict creation of custom fields with same name as standard. Should apply recusively on child tables as well.


phot0n:
> Maybe a Validation must be added in the Framework to restrict creation of custom fields with same name as standard

this happens in framework (atleast in develop/v14) - not completely sure about v13

https://user-images.githubusercontent.com/32034600/201376143-8e5bb265-38a8-4e5d-8a11-2879261d9477.mp4

PR #32933, opened 2022-11-11
feat: separating depreciation schedule from assets into a new doc

anandbaburajan:
`Depreciation Schedule` was a child table in `Asset`, but this approach had a major problem -- every time an `Asset` would undergo a change that would modify its `Depreciation Schedule`, the table would just be overwritten and no record would be maintained of its previous state.

To overcome this, I’ve created a new doctype -- the `Asset Depreciation Schedule`, which has `Depreciation Schedule` as a child table. Now, when a depreciable `Asset` is created, an `Asset Depreciation Schedule` would be created in the `Draft` state. On submitting the `Asset`, the `Asset Depreciation Schedule` gets submitted as well and its status changes to `Active`.

When an `Asset` undergoes any change that should modify its schedule (if it undergoes a repair that extends its life, for instance), the current active `Asset Depreciation Schedule` gets cancelled with a note explaining why, and a new active `Asset Depreciation Schedule` is created with the new schedule.

A separate `Asset Depreciation Schedule` is created for each `Finance Book` in an `Asset`.

This PR also includes 5-6 bug fixes (mostly finance books related) which were in the same parts of code I was working on.


erpnext/assets/doctype/asset/asset.py
```diff
@@ -1132,34 +864,49 @@ def update_existing_asset(asset, remaining_qty):
...
+			"This schedule was created when Asset {0} was updated after being split into new Asset {1}."
+		).format(
+			get_link_to_form(asset.doctype, asset.name), get_link_to_form(asset.doctype, new_asset_name)
 		)
+		new_asset_depr_schedule_doc.notes = notes
+
+		current_asset_depr_schedule_doc.flags.should_not_cancel_depreciation_entries = True
+		current_asset_depr_schedule_doc.cancel()
+
+		new_asset_depr_schedule_doc.submit()
```
deepeshgarg007, line 909: Could see similar code in the same file below as well, can this be moved to a function to reduce duplication?

erpnext/patches/v15_0/create_asset_depreciation_schedules_from_assets.py
```diff
@@ -0,0 +1,80 @@
...
+
+			asset_depr_schedule_doc.insert()
+
+			if asset.docstatus == 1:
+				asset_depr_schedule_doc.submit()
+
+			update_depreciation_schedules(asset.name, asset_depr_schedule_doc.name, fb_row.idx)
+
+
+def get_details_of_draft_or_submitted_depreciable_assets():
```
deepeshgarg007, line 29: For users having a few assets records this might not be an issue, but for users having a lot of assets this patch might take some time, will it be possible to get all info in one single query and make a sort of map and use that to get the details?

deepeshgarg007 (commented):
Just a couple of suggestions, the rest all looks good to me
Please also start updating the docs and raise a draft request for the visible user changes and link it in the description

erpnext/patches/v15_0/create_asset_depreciation_schedules_from_assets.py
anandbaburajan: Ah, yep. Do you have any existing patches in mind doing something like that which I can refer to?

deepeshgarg007 (approved):

erpnext/patches/v15_0/create_asset_depreciation_schedules_from_assets.py
anandbaburajan: @deepeshgarg007 done: https://github.com/frappe/erpnext/pull/35867.

PR #32953, opened 2022-11-14
chore(payment_entry): Remove dead validations

gavindsouza:
Remove or enable `PaymentEntry.validate_bank_accounts`? Not sure what enabling this now will cause. Or if it is even required. At this point, all this check does is add multiple DB calls and has nothing to show for.

Commented out via #22559 (2+ years ago)

ref: https://github.com/frappe/erpnext/pull/22559/files#r1018934765


deepeshgarg007 (approved):

deepeshgarg007:

PR #32958, opened 2022-11-14
fix: link `Item Price` brand to `Brand` doctype.

dj12djdjs:
If there are item price records assigned a brand, and that brand gets renamed. The brand shown in the item price will remain outdated.

Changing the brand field from Read Only to Link will force the item price brand to be updated to the renamed brand.

### DIFF
--- a/erpnext/stock/doctype/item_price/item_price.json
+++ b/erpnext/stock/doctype/item_price/item_price.json
@@ -1,6 +1,7 @@
 {
  "actions": [],
  "allow_import": 1,
+ "autoname": "hash",
  "creation": "2013-05-02 16:29:48",
  "description": "Multiple Item prices.",
  "doctype": "DocType",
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/item_price/item_price.json
```diff
@@ -218,11 +219,10 @@
  "idx": 1,
  "index_web_pages_for_search": 1,
  "links": [],
- "modified": "2022-09-02 16:33:55.612992",
+ "modified": "2022-11-14 17:26:49.052007",
  "modified_by": "Administrator",
  "module": "Stock",
  "name": "Item Price",
- "name_case": "Title Case",
```
deepeshgarg007, line 225: @dj12djdjs Was this auto changed or done intentionally?

dj12djdjs: I think this was an auto change. I'll fix and push it

dj12djdjs: I'm not sure what the "name_case" key was related to. I'm thinking it's to do with naming. I know item price originally is a random hash for PK. So I specified that in the doctype definition.

deepeshgarg007:

PR #32962, opened 2022-11-15
fix: Write Off section visibility for non POS Invoices

### LINKED ISSUE #23068: Write off is lost upon payment entry. Wrong outstanding shown.

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/23068

Write off section is intended only for POS invoices as there is no separate payment entry made and the write-off has to be handled within the sales invoice itself.

Currently, it is also visible for nonpos invoices which should not be the case as write off for them should be handled in payment entries and not the invoices.

Adding write off to non-pos invoices leads to incorrect outstanding amount updates on posting GL Entries

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.json
@@ -2098,7 +2098,7 @@
   {
    "collapsible": 1,
    "collapsible_depends_on": "write_off_amount",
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

PR #32970, opened 2022-11-15
Revert "fix: get `consumed_qty` based on `received_qty` in SCR"

s-aga-r:
Reverts frappe/erpnext#32867


s-aga-r:
created: https://github.com/frappe/erpnext/pull/32971

PR #32983, opened 2022-11-16
fix: Opening journal entry templates

deepeshgarg007:
image


deepeshgarg007:

PR #32993, opened 2022-11-16
fix: validate doctype argument in get children method

saurabh6790:


ankush:
This change is not safely portable to v14 as it will add hard dependency on related frappe 14.x version. 


lets not do this.

PR #33003, opened 2022-11-17
fix: precision in asset tests

anandbaburajan:
This is to fix the failing `test_gle_made_by_asset_sale` test due to incorrect precision [here](https://github.com/frappe/erpnext/blob/65e855bfff6b0e2ca935c3f81a907ad9b6eb7e2d/erpnext/assets/doctype/asset/test_asset.py#L304).

```
-  ('_Test Gain/Loss on Asset Disposal - _TC', 52064.52, 0.0),
+  ('_Test Gain/Loss on Asset Disposal - _TC', 52064.520000000004, 0.0),
```


anandbaburajan:
Thanks @s-aga-r for letting me know about the failing test.

PR #33005, opened 2022-11-17
feat(pos): multiple item prices

### LINKED ISSUE #32922: Barcode wise UOM not working in Point of Sale
### Information about bug

I have created 2 barcodes of one item in Item Barcode with 2 different UOM i.e., Nos, and Box. 

The Barcode UOM is appearing fine if we scan the item barcode in a sales invoice, sales order, or stock entry is created; however, while scanning the barcode in Point of Sale it is not showing UOM as per scanned barcode. It always shows stock_uom irrespective of other UOM mentioned in the Item Barcode table.

### Module

selling

### Version

ERPNext: v14.5.1 (HEAD)
Frappe Framework: v14.14.2 (HEAD)

### Installation method

FrappeCloud

```shell
There is no error but Item Barcode shows incorrect UOM.
```

ssiyad:
- [x] show uom with product price
- [x] multiple item (variant) depending on uom


erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -17,45 +17,79 @@
...
-				"stock_uom",
-				"image as item_image",
-				"is_stock_item",
-			],
-			as_dict=1,
-		)
+	item_code = result.get("item_code", search_term)
+	serial_no = result.get("serial_no", "")
+	batch_no = result.get("batch_no", "")
+	barcode = result.get("barcode", "")
```
barredterra, line 23: Are you aware that 
- `{}.get("barcode", "")` is equal to `{}.get("barcode") or ""`, but
- `{"barcode": None}.get("barcode") or ""` returns `""` while `{"barcode": None}.get("barcode", "")` returns `None`?

So, your changes could produce `None` values, while the former code didn't. Is that what was intended here?

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -17,45 +17,79 @@
...
+				}
+			)
 
-		item_stock_qty, is_stock_item = get_stock_availability(item_code, warehouse)
-		price_list_rate, currency = frappe.db.get_value(
-			"Item Price",
-			{"price_list": price_list, "item_code": item_code},
-			["price_list_rate", "currency"],
-		) or [None, None]
+	item_stock_qty, _ = get_stock_availability(item_code, warehouse)
```
barredterra: `_()` is the translation function in most parts of ERPNext. Redefining it here may lead to confusion.

erpnext/selling/page/point_of_sale/point_of_sale.py
```diff
@@ -17,45 +17,79 @@
...
+	item_stock_qty = item_stock_qty // item.get("conversion_factor")
+	item.update({"actual_qty": item_stock_qty})
+
+	price = frappe.get_list(
+		doctype="Item Price",
+		filters={
+			"price_list": price_list,
+			"item_code": item_code,
+		},
+		fields="*",
```
barredterra:
```suggestion
		fields=["uom", "stock_uom", "currency", "price_list_rate"],
```

erpnext/selling/page/point_of_sale/point_of_sale.py
ssiyad: Valid point, but `search_for_serial_or_batch_or_barcode_number()` (which use `scan_barcode()`) does not return explicit `None` values.

- https://github.com/frappe/erpnext/blob/develop/erpnext/stock/utils.py#L557
- https://stackoverflow.com/a/6130800/11143333

deepeshgarg007:
keep-alive

PR #33013, opened 2022-11-17
fix: validate repost item valuation against accounts freeze date

rtdany10:
Peek 2022-11-17 15-33

I have taken into consideration the role allowed to override. Maybe we shouldn't do that?
@deepeshgarg007

#no-docs

### DIFF
--- a/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
+++ b/erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
@@ -37,8 +37,9 @@ class RepostItemValuation(Document):
 		if not acc_settings.acc_frozen_upto:
 			return
 		if getdate(self.posting_date) <= getdate(acc_settings.acc_frozen_upto):
-			if acc_settings.frozen_accounts_modifier and frappe.session.user in get_users_with_role(
+			if (
 				acc_settings.frozen_accounts_modifier
+				and frappe.session.user in get_users_with_role(acc_settings.frozen_accounts_modifier)
 			):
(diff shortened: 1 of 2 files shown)


erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
```diff
@@ -25,6 +25,25 @@ def validate(self):
...
+	def validate_accounts_freeze(self):
+		acc_settings = frappe.db.get_value(
+			"Accounts Settings",
+			"Accounts Settings",
+			["acc_frozen_upto", "frozen_accounts_modifier"],
+			as_dict=1,
+		)
+		if not acc_settings.acc_frozen_upto:
+			return
+		if acc_settings.frozen_accounts_modifier and self.owner in get_users_with_role(
```
deepeshgarg007: The doc owner can be forced set to some other owner as well I guess, doesn't seem reliable. Can we check this with the current session user as we do at other places?

deepeshgarg007:
Also, this seems like a `fix` and not a `feat`

deepeshgarg007:
@rtdany10 Any update on this PR?

rtdany10:
> @rtdany10 Any update on this PR?

Sorry for the delay, have changed it to session user.

Also, this:
> I have taken into consideration the role allowed to override. Maybe we shouldn't do that?

erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
rtdany10: Updated

PR #33022, opened 2022-11-17
feat: Incoterms in buying and selling

barredterra:
https://en.wikipedia.org/wiki/Incoterms

- Add doctype **Incoterm** with code, title and description
    - naming by code
    - title field and translated links enabled (users will see the translated title in the link fields)
- Add links to **Incoterm** in Selling transactions (**Quotation**, **Sales Order**, **Delivery Note**, **Sales Invoice**)
- Add links to **Incoterm** in Buying transactions (**Request for Quotation**, **Supplier Quotation**, **Purchase Receipt**, **Purchase Invoice**)
- Replace select field in **Shipment** with link and migrate existing shipments
- Create **Incoterm** records via patch
- Create **Incoterm** records after install
- Add german translations for incoterm titles

### Doubts

- Should users just use the existing "Terms" doctype for this?
    - Pro: incoterms are a subset terms.
    - Contra: terms might contain additional company-specific clauses, while incosterms are standardized. A combination of both might make more sense.


erpnext/patches/v14_0/create_incoterms_and_migrate_shipment.py
```diff
@@ -0,0 +1,32 @@
...
+	create_incoterms()
+	migrate_shipments()
+
+
+def migrate_shipments():
+	if not frappe.db.count("Shipment"):
+		return
+
+	OLD_VALUES = [
+		# old_value: (code, title),
```
barredterra:
```suggestion
```

barredterra:
Will add docs once reviews and related changes are through.

erpnext/setup/doctype/incoterm/incoterm.py
```diff
@@ -0,0 +1,24 @@
...
+class Incoterm(Document):
+	pass
+
+
+def create_incoterms():
+	"""Create Incoterm records from incoterms.csv."""
+	import os
+	from csv import DictReader
+
+	with open(os.path.join(os.path.dirname(__file__), "incoterms.csv"), "r") as f:
```
deepeshgarg007, line 17: Any specific reason for reading them from a separate CSV file rather than simply hardcoding a dictionary/object in this file itself?

barredterra: Separating data from code, non-technical people can modify/extend it without having to know about python

PR #33032, opened 2022-11-18
fix: Accounting Dimension filtering for Sales and Purchase Report

deepeshgarg007:
Sales and Purchase Invoices should be filtered based on the parent-level accounting dimensions instead of the item level. Parent and child tables can have different dimensions and viewing the parent-level reports after applying dimension filters might give different results when inline filters in the data table are applied.

With Inline filters
image

With Report filters
image


deepeshgarg007:

PR #33051, opened 2022-11-20
fix: Timesheet timer button

### LINKED ISSUE #27878: Timesheet Timer goes to an inconsistent state on clicking "Complete"
## Description of the issue

In Timesheet Timer, clicking on "Complete" results in:

-  a "No changes in document" popup
- the "Save" button not being shown (it remains "Submit") and 
- the "Start Timer" button not being shown (it stays "Resume Timer").

However, editing the row shows the "Completed" checkbox checked (so the buttons I mentioned should say "Save" instead of "Submit" and "Start Timer" instead of "Resume Timer")

This can be worked around by making any minor change to the row, which will then enable the "Save" button and "Start Timer" button.

It's still pretty dangerous because if you don't notice you can easily leave the document unsaved, with the result that you probably bill the wrong amount of time.

**Output of `bench version`**
```
frappe@39103186755b:~/bench$ bench version
erpnext 13.11.1
frappe 13.11.0
```

1. Start a Timesheet timer
2. Click on "Complete" to stop it

### Observed result

The UI is not updated, and the fact that the timer has stopped is not saved

### Expected result

The UI is updated, showing "Save" instead of "Submit" and "Start Timer" instead of "Resume Timer".

Ideally the document would be also saved automatically, but that's just my preference.

```
N/A
```

## Additional information

ERPNext running in docker container from image `pipech/erpnext-docker-debian:v13-py3-latest`

```$ docker -v
Docker version 20.10.5+dfsg1, build 55c4c88
```
Host OS: Debian Bullseye

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/27878

### DIFF
--- a/erpnext/public/js/projects/timer.js
+++ b/erpnext/public/js/projects/timer.js
@@ -48,21 +48,23 @@ erpnext.timesheet.control_timer = function(frm, dialog, row, timestamp=0) {
 	var $btn_complete = dialog.$wrapper.find(".playpause .btn-complete");
 	var interval = null;
 	var currentIncrement = timestamp;
-	var initialised = row ? true : false;
+	var initialized = row ? true : false;
 	var clicked = false;
 	var flag = true; // Alert only once
 	// If row with not completed status, initialize timer with the time elapsed on click of 'Start Timer'.
 	if (row) {
-		initialised = true;
+		initialized = true;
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

PR #33062, opened 2022-11-21
fix: create rounding gl entry for PCV during gle post processing

nabinhait:
**Issue:**
While submitting Period Closing Voucher, the error "Debit != Credit" appeared with a difference of 33.

**Solution:**
After investigation came to know that on postprocessing of `make_gl_entries`, there was a difference of 0.02, that's why the system was creating an extra round-off-gle. And in this process, the system tries to adjust the values if there is already any gle for round-off-account in the current voucher's gl entries. Generally, we expect only a single existing gle for round-off-account for any voucher.
But in the case of Period Closing Voucher, there can be multiple gl entries against round-off-account based on different accounting dimensions. And that's why we should not try to adjust the rounding difference against the existing round-off-gle for PCV. The system should create a new gle for the rounding difference.


deepeshgarg007:

PR #32646, opened 2022-10-19
feat: Item Wise TDS Calculation

niralisatapara:
###  For calculating Item wise TDS in the Purchase Invoice

Step 1:  **To calculate TDS item-wise, added apply_tds checkbox in Purchase Invoice Item.**
image

Step 2: **Calculate Tax Withholding Net Total based where apply_tds is checked.**
image

Step 3: **Computing TDS based on Tax Withholding Category**
image

`no-docs`

PR #33063, opened 2022-11-21
feat: item wise tds in purchase order

niralisatapara:
**For calculating Item wise TDS in the Purchase Order.**
Reference of pull request  #32646 of purchase invoice.

`no-docs`


deepeshgarg007:
@niralisatapara Can you please also replicate a test case for the Purchase Order same as the Purchase Invoice so that such changes in the Purchase Invoice affecting PO are not missed again.

deepeshgarg007:

PR #33068, opened 2022-11-21
fix: Don't show payment button for invoices on hold

### LINKED ISSUE #20183: On version:-  v12.0.5 (version-12) , Payment entry gets created for holding invoice and without having reference of PI on PE

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/20183

### DIFF
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
@@ -81,7 +81,7 @@ erpnext.accounts.PurchaseInvoice = class PurchaseInvoice extends erpnext.buying.
 		}
 
 		if(doc.docstatus == 1 && doc.outstanding_amount != 0
-			&& !(doc.is_return && doc.return_against)) {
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

PR #33069, opened 2022-11-21
fix: Remove unnecessary filters from Journal Entry

### LINKED ISSUE #20456: JV Type as Credit Note should not include "Is Return (Credit Note) = 1" in filter
On v12 Cloud version.

When creating a credit note via Journal Voucher, user select Type = Credit Note.
image

User enters details and expand the accounts table to apply a credit on the Debtor (Customer)'s account. User wants to specifically credit the customer's account down to a specific sales invoice. This is possible by setting `Reference Type` to `Sales Invoice`.

However, when user proceeds to fill out `Reference Name`, the script/filter applies to only show 
* doc status = 1 
* Outstanding amount != 0
* Customer = customer defined in the JV
... and a bunch of other filters that are important to help speed up the search.

Screenshot 2020-01-28 at 7 01 19 PM

However, one condition is `Is Return (Credit Note) = 1` for the Sales Invoice. This is very limiting and should be removed from the conditions, as there is use case where user is actually using the said JV document being created as the *Credit Note itself*. Hence, user is applying the amount indicated in the JV to deduct against the receivables of that specified sales invoice (or the payable related to that specific SI from the point of view of the customer).

Ask: remove this `Is Return (Credit Note) = 1` condition in the reference name search filter.

deepeshgarg007:
fixes: https://github.com/frappe/erpnext/issues/20456

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -253,9 +253,6 @@ erpnext.accounts.JournalEntry = class JournalEntry extends frappe.ui.form.Contro
 				var party_account_field = jvd.reference_type==="Sales Invoice" ? "debit_to": "credit_to";
 				out.filters.push([jvd.reference_type, party_account_field, "=", jvd.account]);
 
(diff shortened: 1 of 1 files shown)


deepeshgarg007:

PR #33071, opened 2022-11-22
fix: advance paid updated in account currency for SO/PO

ruthra-kumar:
Update `advance_paid` of Sales/Purchase Order in account currency using Payment Ledger.

regression: https://github.com/frappe/erpnext/pull/32816


ruthra-kumar:

Issue #32556, opened 2022-10-11
Retail POS - Returns


tjomtek:
### Information about bug

When processing a return in the pos retail module of erpnext:
If the amount paid by the customer is exactly the same amount owed by the customer, on-return the system produces the "payment method" option as expected.
However , if the customer pays more than the total owed , then the 'payment method" option does not appear. i.e the total amount for an item bought is 155 and the customer gives 160 , their change is 5, the system handles this nicely but does not known how to handle it on return.

### Module

accounts, selling, other

### Version

ERPNext: v13.36.1 (version-13)

Frappe Framework: v13.36.3 (version-13)

Screenshot with "payment method" on return with exact amounts no change issued:

![Screenshot from 2022-10-11 08-13-09](https://user-images.githubusercontent.com/77873407/195016464-1dea13ec-718d-46a2-94f7-f5fc8e20e6a0.png)

Screenshot with no "payment method" option when the system issues change to the customer.

![Screenshot from 2022-10-11 08-53-49](https://user-images.githubusercontent.com/77873407/195017227-f8119e0c-7cfb-4deb-86d0-b923b8bcfef6.png)

If you click on checkout at this point the result is as screen shot below 
![Screenshot from 2022-10-11 08-58-31](https://user-images.githubusercontent.com/77873407/195017944-da5373b9-ffc6-4866-a242-bc192da786e8.png)

ssiyad:
Probably a duplicate of #31896

Issue #31896, opened 2022-08-19
POS Return not Working


ravi854:
### Information about bug

POS return not working,
When doing pos return, it shown a message "Total payment amount cant be greater than (invoice imount)"
![pos (2)](https://user-images.githubusercontent.com/85276497/185581425-8d438f1d-7200-4611-a476-f5d8613ba671.gif)

### Module

accounts

### Version

Frappe Framework: v14.0.2 (version-14)
ERPNext: v14.0.0 (version-14)

PR #33074, opened 2022-11-22
fix(pos): set return amount as net amount

ssiyad:
ref:
  - https://github.com/frappe/erpnext/issues/32556
  - https://github.com/frappe/erpnext/issues/31896


ssiyad:
https://github.com/frappe/erpnext/pull/33091

PR #33093, opened 2022-11-23
fix: cannot update auth token until token expiry if credentials change

### LINKED ISSUE #30673: India E Invoice - GSP Client Credential when reset from GSP's website, ERPNext still send old Auth Token
### Information about bug

As per the below code in `erpnext/regional/india/e_invoice/utils.py ` ERPNext will fetch a new auth_token only when the token expires. But if the client secret has been reset from GSP portal and updated in the site_config still it used the old token till it expires.

```
def get_auth_token(self):
		if time_diff_in_seconds(self.e_invoice_settings.token_expiry, now_datetime()) < 150.0:
			self.fetch_auth_token()

		return self.e_invoice_settings.auth_token
```

### Module

accounts, regional

### Version

Frappe Version - v13.25.0 (version-13)
ERPNext Version - v13.25.2 (version-13)

### Installation method

easy-install

_No response_

nextchamp-saqib:
Fixes #30673

### DIFF
--- a/erpnext/regional/doctype/e_invoice_settings/e_invoice_settings.py
+++ b/erpnext/regional/doctype/e_invoice_settings/e_invoice_settings.py
@@ -10,3 +10,8 @@ class EInvoiceSettings(Document):
 	def validate(self):
 		if self.enable and not self.credentials:
 			frappe.throw(_("You must add atleast one credentials to be able to use E Invoicing."))
+
+		prev_doc = self.get_doc_before_save()
+		if prev_doc.client_secret != self.client_secret or prev_doc.client_id != prev_doc.client_id:
+			self.auth_token = None
+			self.token_expiry = None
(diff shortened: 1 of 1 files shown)


erpnext/regional/doctype/e_invoice_settings/e_invoice_settings.py
```diff
@@ -10,3 +10,8 @@ class EInvoiceSettings(Document):
 	def validate(self):
 		if self.enable and not self.credentials:
 			frappe.throw(_("You must add atleast one credentials to be able to use E Invoicing."))
+
+		prev_doc = self.get_doc_before_save()
+		if prev_doc.client_secret != self.client_secret or prev_doc.client_id != prev_doc.client_id:
```
nextchamp-saqib:
```suggestion
		if prev_doc.client_secret != self.client_secret or prev_doc.client_id != self.client_id:
```

PR #33138, opened 2022-11-28
fix(ux): Action buttons in Bank Reconciliation

### LINKED ISSUE #22618: Bank Reconciliation does not consider Opening Entries causing total_amount to be distored
## Description of the issue
The opening entries are not considered during bank reconciliation, making the total amount show incorrect value, vis-a-vis the actual bank balance as per the General Ledger

https://github.com/frappe/erpnext/blob/260bf54c994ab43d770f874dea33fc52d04e00ff/erpnext/accounts/doctype/bank_clearance/bank_clearance.py#L37
This line ignores the opening balances and therefore the totals are incorrect. It will be better to included it so that the totals match as per the bank statement

**Output of `bench version`**
```
erpnext 12.10.0
frappe 12.7.3
```

1. Create opening balance for a bank account. e.g. INR1,000.00
2. Create payment entries for that bank account e.g. 50, 40, 30.
3. Open Bank Reconciliation (Bank Clearance)
4. Get unreconciled entries

### Observed result
The total will show for the unreconciled entries of INR50 + 40 + 30 = INR 120.00 only. The actual balance is INR880.00

### Expected result
The actual balance is INR880.00 as there was an opening balance of INR1,000.00

There is no error message.

## Additional information
Ubuntu 16.04.5 LTS
ERPNext easy install used

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/22618

Removed "Total Amount" field from the form, that field was never meant to show the account balance and was just causing confusion

Removed doc buttons and added primary buttons for consistency

### DIFF
--- a/erpnext/accounts/doctype/bank_clearance/bank_clearance.js
+++ b/erpnext/accounts/doctype/bank_clearance/bank_clearance.js
@@ -37,6 +37,14 @@ frappe.ui.form.on("Bank Clearance", {
 
 	refresh: function(frm) {
 		frm.disable_save();
+
+		if (frm.doc.account && frm.doc.from_date && frm.doc.to_date) {
+			frm.add_custom_button(__('Get Payment Entries'), () =>
+				frm.trigger("get_payment_entries")
+			);
+
+			frm.change_custom_button_type('Get Payment Entries', null, 'primary');
+		}
 	},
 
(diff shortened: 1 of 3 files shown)


deepeshgarg007:

deepeshgarg007:

PR #33146, opened 2022-11-28
fix: reload Currency Exchange Settings in patch

barredterra:
Got an error while migrating a backup that didn't have Currency Exchange Settings


deepeshgarg007:

PR #33148, opened 2022-11-28
Update tax_rule.json

wojosc:
added required option to field that it is visible on UI. If not set you get an error when trying to save a Tax Rule.
image

In the field we link a Template for Sales/Purchase Taxes & Charges

### DIFF
--- a/erpnext/accounts/doctype/tax_rule/tax_rule.json
+++ b/erpnext/accounts/doctype/tax_rule/tax_rule.json
@@ -129,6 +129,7 @@
    "fieldname": "tax_category",
    "fieldtype": "Link",
    "label": "Tax Category",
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@wojosc Tax Category is not a required field for generating a tax rule. Why it has been marked as mandatory? This seems unnecessary. Unable to understand the issue here

wojosc:
@deepeshgarg007 thanks for your feedback. Sorry, it was late, must of made a mistake. The field is sales_tax_template when selecting Sales in tax_type and purchase_tax_template when selecting Purchase. These are the fields that should have a mandatory condition.

...you might be quicker in changing this. Let me know if I can do something further. Thank you.

PR #33162, opened 2022-11-29
fix: Tax withholding net total for PI in reports

deepeshgarg007:
Item-wise TDS was introduced in PR https://github.com/frappe/erpnext/pull/32861

Accommodated this in the reports to show amounts just for the Item on which TDS is applicable
**Note**: This only works for Purchase Invoices for other doctypes will still show same amounts


deepeshgarg007:

PR #33170, opened 2022-11-29
fix(UX): Alert on change of item rate in Stock Entry

### LINKED ISSUE #28927: show toast when zero valuation rate is forced
When valuation rate is changed to 0 automatically, it can be quite confusing to the user.

show a toast message when allow zero valuation is causing rate to become 0 (If it's changed; if it was already 0 then no need).

ref: ISS-21-22-09500

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/28927


s-aga-r (approved):

deepeshgarg007:

PR #33182, opened 2022-11-30
fix: Error on making stock entry from material request

### LINKED ISSUE #22856: Incorrect / misleading error message when item value = 0 and creating STE
When trying to create a new STE (Stock Entry) either from an MR or if you pull the MR into the new STE, and the one of the items in the MR has a zero value, the following error comes up
image

It suggests to "If the item is transacting as a Zero Valuation Rate item in this entry, **please enable 'Allow Zero Valuation Rate' in the Stock Entry Item table**." (highlight mine)

However, since the items never get pulled into the STE or the STE doesn't get created,then user cannot do anything really. So either allow the transfer or fix the error message.

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/22856


deepeshgarg007:

PR #32836, opened 2022-11-03
perf: use `get_cached_value` instead of `db.get_value` in accounts module

DaizyModi:
Similar to #32776 

Applied on **Accounts** Module

### Before

```
In [1]: from erpnext.accounts.doctype.fiscal_year.fiscal_year import get_from_and_to_date
In [2]: %timeit get_from_and_to_date('2022-2023')
2.52 ms ± 63.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### After
```
In [1]: from erpnext.accounts.doctype.fiscal_year.fiscal_year import get_from_and_to_date
In [2]: %timeit get_from_and_to_date('2022-2023')
5.84 µs ± 203 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

PR #33191, opened 2022-12-01
fix: key error while filtering on date range and reporting on foreign currency

ruthra-kumar:
Financial Reports - P/L and Balance sheet, throw Key error while filtering on date range and reporting on a foreign currency.

Screenshot 2022-12-01 at 3 35 14 PM

1. Regression in `get_from_and_to_date` by https://github.com/frappe/erpnext/pull/32836 https://github.com/frappe/erpnext/blob/d23b5d8f2fc9bdf2e0f267fbaf38bdf11167fc87/erpnext/accounts/doctype/fiscal_year/fiscal_year.py#L162-L164

2. `get_currency` should handle both Fiscal Year and Manual date range in General Ledger, P/L Report and Balance sheet. https://github.com/frappe/erpnext/blob/d23b5d8f2fc9bdf2e0f267fbaf38bdf11167fc87/erpnext/accounts/report/utils.py#L31

### DIFF
--- a/erpnext/accounts/doctype/fiscal_year/fiscal_year.py
+++ b/erpnext/accounts/doctype/fiscal_year/fiscal_year.py
@@ -169,5 +169,6 @@ def auto_create_fiscal_year():
 
 
 def get_from_and_to_date(fiscal_year):
-	fields = ["year_start_date as from_date", "year_end_date as to_date"]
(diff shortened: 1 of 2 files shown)


deepeshgarg007:

PR #33214, opened 2022-12-04
fix(v13 patch): website_image field not copied when creating Website Items

casesolved-co-uk:
When upgrading, item images are not copied to the new Website Items


barredterra (changes requested):
Hi @casesolved-co-uk, thanks for your contribution. Kindly have a look at the [Pull Request Checklist](https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist), in particular the sections **Which Branch** and **Linting**.

Maybe it might be useful to have a look at the existing methods and patches that create a **Website Item** from an **Item** and fix these first, then create a patch that takes care of the remaining "failed" migrations.

marination:
@casesolved-co-uk Check this out https://github.com/frappe/erpnext/pull/31469. This makes sure `website_image` is used as that is the only image field fitting the ecommerce context.
This solves the issue for anyone trying to **create Website Items**. You could test it out if you want.

It was fixed via an issue you raised earlier https://github.com/frappe/erpnext/pull/30081

deepeshgarg007:
@marination Can we close this PR?

PR #33219, opened 2022-12-05
fix: Allow item rate updates for non-stock invoices

deepeshgarg007:
Currently, the valuation rate is forced for items in internal invoices even when the "update stock" check is not enabled.

In many scenarios, the user doesn't want to show the actual valuation rates to the receiving party of the invoices and use the price list rate only for such outgoing transactions.

This change allows users to make a Delivery Note/ Purchase Receipt for moving stock in and out and make invoices against them with the price list rate and apply taxes on them


deepeshgarg007:

PR #33250, opened 2022-12-07
fix: validate if group account is set in item defaults as expense account

MyuddinKhatri:
Problem - 
While submitting stock entry if the expense account set is a group account it gives error
image

Solution - 
Added validation in Item master if user sets Default expense account in Item defaults which is group enabled it throws error and in Company if user sets default Inventory account which is group  Account it throws error.
image

image


MyuddinKhatri:
@deepeshgarg007 opened this https://github.com/frappe/erpnext/pull/33353

PR #33265, opened 2022-12-08
refactor!: drop ecommerce in favor of webshop

ssiyad:
Most E-Commerce related logics/files has been moved into a separate repo (https://github.com/frappe/webshop), and can be safely removed from ERPNext.

There are still references remaining to shopping-cart, which will be removed in subsequent PRs


ssiyad:
keep-alive

dj12djdjs:
@ssiyad What happened to this?

rmehta:
I don't think we should do this anymore. Maybe deprecate it in the next version?

PR #33274, opened 2022-12-08
fix: update payment entry references

dj12djdjs:
### Description
Before this PR the following code was executed before adding rows to the deductions table.
```python
payment_entry.flags.ignore_validate_update_after_submit = True
payment_entry.setup_party_account_field()
payment_entry.set_missing_values()
payment_entry.set_amounts()
```

This PR modified the above statements to execute after adding rows to the deductions table.

### Before PR
After payment reconciliation is done, difference amount remains in payment entry document.
image

### After PR
The difference amount has been updated after payment reconciliation. 
image


erpnext/accounts/test/test_utils.py
```diff
@@ -73,6 +77,43 @@ def test_stock_voucher_sorting(self):
...
+		payment_entry = get_payment_entry(purchase_invoice.doctype, purchase_invoice.name)
+		payment_entry.target_exchange_rate = 62.9
+		payment_entry.paid_amount = 15725
+		payment_entry.deductions = []
+		payment_entry.insert()
+
+		self.assertEqual(payment_entry.difference_amount, -4855.00)
+		payment_entry.references = []
+
+		update_reference_in_payment_entry(
```
deepeshgarg007: @dj12djdjs In this test case you seem to be doing manually everything, an ideal test would be to make a standalone payment entry first and then reconcile using payment reconciliation and then check if the difference amount was updated or not

erpnext/accounts/test/test_utils.py
```diff
@@ -73,6 +76,47 @@ def test_stock_voucher_sorting(self):
...
+		payment_reconciliation.get_unreconciled_entries()
+		payment_reconciliation.allocate_entries(
+			{
+				"payments": [d.__dict__ for d in payment_reconciliation.payments],
+				"invoices": [d.__dict__ for d in payment_reconciliation.invoices],
+			}
+		)
+		for d in payment_reconciliation.invoices:
+			# Reset invoice outstanding_amount because allocate_entries will zero this value out.
+			d.outstanding_amount = d.amount
```
dj12djdjs, line 111: I have to reset the `outstanding_amount` here because it will get set to zero here during `allocate_entries()` https://github.com/frappe/erpnext/blob/cabaed9ed2526e2649d173c806f6987d3377b0c3/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py#L245 

This will then fail during the `validate_allocation()`
https://github.com/frappe/erpnext/blob/cabaed9ed2526e2649d173c806f6987d3377b0c3/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py#L367-L373

PR #33275, opened 2022-12-08
fix: use highest precision for exchange rate.

dj12djdjs:
Update payment entry exchange rate fields to use highest precision setting.

This is to follow the same convention as:
`Sales Invoice`: 

https://github.com/frappe/erpnext/blob/7d41f63c5dc409878c98304a361bf21589ac0552/erpnext/accounts/doctype/sales_invoice/sales_invoice.json#L615

`Purchase Invoice`: 

https://github.com/frappe/erpnext/blob/7d41f63c5dc409878c98304a361bf21589ac0552/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.json#L492


deepeshgarg007:

PR #33278, opened 2022-12-09
fix(accounts): allow Payment Request and Payment Entry for POS Invoice

casesolved-co-uk:
Fixes the Payment Request and Payment Entry for POS Invoices

Almost entirely just adds `mode_of_payment` in various places

Looks like a big change but most is formatting using black (with the tabs put back).
Thats' why it hasn't been squashed.

Also adds a hook after Point of Sale payment rendering

version-14-hotfix

## Note

version-13-hotfix

Required for v13 because I have a marketplace app that adds payments to Point of Sale for v13 & v14. Let me know if you want me to provide the v13 backport.


casesolved-co-uk:
No problem, I'll just monkey-patch my module

PR #32923, opened 2022-11-10
fix: Maintain same rate between Quotation and Sales Order

deepeshgarg007:
Fixes: https://github.com/frappe/erpnext/issues/29976

Issue #29976, opened 2022-02-24
[Selling Settings]Maintain Same Rate Throughout Sales Cycle doesn't work


rtdany10:
### Information about bug

Same rate is not enforced even after enabling the setting. One can easily change the rates.

https://user-images.githubusercontent.com/52111700/155473701-94575e87-f0b0-411f-81a7-b59ec7f1f1de.mp4

### Module

Selling / Accounts

### Version

ERPNext: v13.21.0 (HEAD)
Frappe Framework: v13.21.0 (HEAD)

### Installation method

FrappeCloud

ankush:
IIRC this setting applies to quotation yet.

PR #33281, opened 2022-12-09
fix: maintain same rate throughout sales cycle doesn't work with quotation

tundebabzy:
Issue #29976  was partly fixed by #32923 but the problem still persists. The reason is because an incorrect fieldname was passed to the `validate_rate_with_reference_doc` method.

This commit fixes it.

version-14-hotfix

version-13-hotfix

> Screenshots/GIFs


tundebabzy:
@rohitwaghchaure can this be backported to version 13 as well since it is a bugfix?

PR #33305, opened 2022-12-12
feat: incoterm named place

barredterra:
An Incoterm always goes along with a named place of delivery / pickup.

### Before

before

### After

after

- move _Tax Category_ to the first column, above _Taxes and Charges Template_, because the category may determine the template
- move _Shipping Rule_ to the middle column
- move _Incoterm_ to the third column
- add _Named Place_ below _Incoterm_ (hidden until an Incoterm is selected)

Docs: https://docs.erpnext.com/docs/v14/user/manual/en/selling/articles/payment-terms/new-wiki?wiki_page_patch=5dc8e9ee3e


deepeshgarg007:

deepeshgarg007:
image

@barredterra Looks like the title didn't save, can you please update it? Merging the PR for now

PR #33309, opened 2022-12-13
refactor: translatable strings and guard clause

barredterra:
- The error message was not translatable the way it was composed. -> Refactor to get rid of the variables.
- Convert leading `if`-statement to guard clause, thus reducing indentation (condition changed according to [this rule](https://docs.sourcery.ai/Reference/Built-in-Rules/refactorings/de-morgan/))
- Replace unneeded comprehension with generator in `sum()`

Note: "Hide whitespace" may be helpful to make sense of this diff.


deepeshgarg007:

PR #33340, opened 2022-12-14
Feature: Quotation Templates

bernhein:
Until now, users can only import items into a quotation from opportunities. But preconfigured quotations or items are sometimes also very useful.

Over the button "Get Items From" the User can get items from a "Quotation Template".

image

The user has then the ability to multi-select items from Templates

image

If the user clicks on "Get Items", the items from the Quotation Templates are loaded into the quotation and the prices etc. are automatically updated.

Over the button "Make Quotation Template" the user can create a new one.
image


bernhein (commented):
This file should not be in the pull request, sorry!
All others are fine and need to be imported

PR #33353, opened 2022-12-15
fix: validate if group account is set in item defaults as expense account

MyuddinKhatri:
Problem - 
While submitting stock entry if the expense account set is a group account it gives error
image

Solution - 
Added validation in Item master if user sets Default expense account in Item defaults which is group enabled it throws error and in Company if user sets default Inventory account which is group  Account it throws error.
image

image


deepeshgarg007:
@MyuddinKhatri Test case and linter check failing, please fix both of them, the fix looks okay

deepeshgarg007:
@MyuddinKhatri This won't be fixed just by updating the branch. You'll need to update the test cases.
For the linter check setup pre-commit on your local system, please refer to this for the setup https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

MyuddinKhatri:
> @MyuddinKhatri This won't be fixed just by updating the branch. You'll need to update the test cases. For the linter check setup pre-commit on your local system, please refer to this for the setup https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#linting

I'll update PR with tests and lint.

deepeshgarg007:
@MyuddinKhatri Please fix the translation issues

deepeshgarg007:
@MyuddinKhatri Please reopen the PR once the linter issues are resolved

PR #33358, opened 2022-12-15
fix: remove unnecessary permissions from Appointment and Appointment Booking Settings

DaizyModi:


erpnext/crm/doctype/appointment/appointment.py
```diff
@@ -240,3 +242,14 @@ def _get_employee_from_user(user):
 	if employee_docname:
 		return frappe.get_doc("Employee", employee_docname)
 	return None
+
+
+def add_agent_assignment(args):
```
barredterra:
```suggestion
def assign_agents(doctype: str, name: str, agents: list[str]) -> None:
```

Maybe this signature would improve readability a bit.

erpnext/www/book_appointment/index.py
```diff
@@ -90,23 +94,27 @@ def get_available_slots_between(query_start_time, query_end_time, settings):
 
 @frappe.whitelist(allow_guest=True)
 def create_appointment(date, time, tz, contact):
-	format_string = "%Y-%m-%d %H:%M:%S"
-	scheduled_time = datetime.datetime.strptime(date + " " + time, format_string)
+	contact = json.loads(contact)
+	datetime_obj = datetime.datetime.strptime(date + " " + time, "%Y-%m-%d %H:%M:%S")
```
barredterra:
```suggestion
	from frappe.utils.data import get_datetime

	contact = json.loads(contact)
	datetime_obj = get_datetime(date + " " + time)
```

We can use frappe's `get_datetime` util here

DaizyModi: Yes. We will refactor it in another PR.

deepeshgarg007:
Didn't see the squash label :sweat_smile:

PR #33362, opened 2022-12-15
feat: Ignore company related doctype for other apps via hooks

deepeshgarg007:
Ignore company linked doctypes from other apps via hooks on clearing transactions

`no-docs`


deepeshgarg007:

sagarvora:
Alternatively, ignoring children of single doctypes might've worked as well. We are already ignoring singles:
https://github.com/frappe/erpnext/blob/5f1b2263620a489ba3f68e494b4d7200c0fae898/erpnext/setup/doctype/transaction_deletion_record/transaction_deletion_record.py#L119

PR #33372, opened 2022-12-16
feat: add after_refresh hook to item dashboard

dj12djdjs:
When using the ItemDashboard and developer wants to make some modification after data is fetched. An after_refresh hook is required.

`no-docs`

### DIFF
--- a/erpnext/stock/dashboard/item_dashboard.js
+++ b/erpnext/stock/dashboard/item_dashboard.js
@@ -97,11 +97,14 @@ erpnext.stock.ItemDashboard = class ItemDashboard {
 		};
 
 		var me = this;
-		return frappe.call({
+		frappe.call({
 			method: this.method,
 			args: args,
 			callback: function (r) {
 				me.render(r.message);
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@dj12djdjs Will this hook be called via some custom app?

dj12djdjs:
> @dj12djdjs Will this hook be called via some custom app?

@deepeshgarg007 Yes, the idea is to add some event listeners to the newly created DOM elements. My example is:
```js
frm.item_dashboard = new erpnext.stock.ItemDashboard({
  parent: frm.fields_dict.item_code.wrapper,
  item_code: frm.doc.item_code,
  page_length: 20,
  method: 'pps.apps.whms.get_dashboard',
  template: 'material_transfer_dashboard_list',
  after_refresh() {
    $(frm.item_dashboard.result).find("[data-type='warehouse']").click(function () {
       const warehouse = $(this).attr("data-name");
       frm.set_value("scan_barcode", warehouse);
    });
  }
});
```

PR #33387, opened 2022-12-19
fix: daily scheduler to identify and fix stock transfer entries having incorrect valuation

rohitwaghchaure:
- Daily scheduler will execute to identify and fix Stock Transfer entries created one day before and has incorrect valuation.
- The Stock Entry type Material Transfer, Send to Subcontractor, Material Transfer for Manufacture


rmehta:
What is the root cause of this entry? Seems like a very implicit way of fixing!

@nabinhait

PR #33397, opened 2022-12-20
feat: adding warehouse filter for gross profit report

Gokulnath17:
In gross profit report, there is no warehouse filter, so i will add warehouse filter in gross profit report report.

**Uses:** If some company use warehouse for their branch, so warehouse filter is used for that types companies and this feature also used for all types of companies.

image


erpnext/accounts/report/gross_profit/gross_profit.js
```diff
@@ -50,6 +50,12 @@ frappe.query_reports["Gross Profit"] = {
 			"fieldtype": "Link",
 			"options": "Sales Person"
 		},
+		{
+			"fieldname": "warehouse",
+			"label": __("Warehouse"),
+			"fieldtype": "Link",
+			"options": "Warehouse"
+		},
```
ruthra-kumar, line 66: 1. Use `get_query` to filter Warehouses on Company. Currently it shows warehouses from all the companies on a multi company setup. `get_query` ex: https://github.com/frappe/erpnext/blob/a7b011aed4c879e077a3a3e81c6834dc2555eac1/erpnext/accounts/report/accounts_payable_summary/accounts_payable_summary.js#L65-L73
3. When a root node is selected, all of its child nodes should be included in the filter. You can use the utlity function for that. https://github.com/frappe/erpnext/blob/a7b011aed4c879e077a3a3e81c6834dc2555eac1/erpnext/stock/report/stock_ledger/stock_ledger.py#L430

deepeshgarg007:
@Gokulnath17 Do you plan to fix this? Else we will close the PR

erpnext/accounts/report/gross_profit/gross_profit.js
Gokulnath17: Now fixed kindly check it...

> @Gokulnath17 Do you plan to fix this? Else we will close the PR
Updated

ruthra-kumar: > When a root node is selected, all of its child nodes should be included in the filter. You can use the utlity function for that.

@Gokulnath17  This is not fixed

PR #33418, opened 2022-12-22
fix: payment terms and sales partner filter issue in Accounts Receivable report

ruthra-kumar:
TypeError while using Payment Terms Template and Sales Partner Filters in Accounts Receivable Report
Screenshot 2022-12-22 at 10 56 44 AM


ruthra-kumar:

PR #33439, opened 2022-12-26
fix: Default dimensions on fetching items from BOM

deepeshgarg007:
Default dimension gets cleared in Stock Entry when items are fetched from BOM


deepeshgarg007:

PR #33444, opened 2022-12-26
fix: consider child nodes while getting bin details

### LINKED ISSUE #28766: get_item_details() args.warehouse results in 0 actual_qty when group node is supplied
https://github.com/frappe/erpnext/blob/ef53eb7fcda8a58c62a11aef3c8a7dc16f5e3f0b/erpnext/stock/get_item_details.py#L29

If args.warehouse is a group node then https://github.com/frappe/erpnext/blob/ef53eb7fcda8a58c62a11aef3c8a7dc16f5e3f0b/erpnext/stock/get_item_details.py#L988 will return actual_qty 0

# Expected Result
return the sum of projected_qty, actual_qty, and reserved_qty  for all nodes under the specified group.

s-aga-r:
closes: https://github.com/frappe/erpnext/issues/28766

Before:

image

After:

image

image

### DIFF
--- a/erpnext/stock/get_item_details.py
+++ b/erpnext/stock/get_item_details.py
@@ -1179,21 +1179,22 @@ def get_bin_details(item_code, warehouse, company=None, include_child_warehouses
 	bin_details = {"projected_qty": 0, "actual_qty": 0, "reserved_qty": 0}
 
 	if warehouse:
+		from frappe.query_builder.functions import Coalesce, Sum
+
 		from erpnext.stock.doctype.warehouse.warehouse import get_child_warehouses
 
 		warehouses = get_child_warehouses(warehouse) if include_child_warehouses else [warehouse]
-		bin_details = frappe.db.get_value(
-			"Bin",
-			filters={"item_code": item_code, "warehouse": ["in", warehouses]},
-			fieldname=[
-				"sum(projected_qty) as projected_qty",
-				"sum(actual_qty) as actual_qty",
-				"sum(reserved_qty) as reserved_qty",
-			],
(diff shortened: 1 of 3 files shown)


erpnext/stock/get_item_details.py
```diff
@@ -1171,14 +1175,28 @@ def get_projected_qty(item_code, warehouse):
...
+		bin_details = (
+			frappe.qb.from_(bin)
+			.select(
+				Sum(bin.projected_qty).as_("projected_qty"),
+				Sum(bin.actual_qty).as_("actual_qty"),
+				Sum(bin.reserved_qty).as_("reserved_qty"),
+			)
+			.where((bin.item_code == item_code) & (bin.warehouse.isin(warehouses)))
+		).run(as_dict=True)[0]
+		bin_details = {k: 0 if not v else v for k, v in bin_details.items()}
```
rohitwaghchaure:
```
bin_details = {k: 0 if not v else v for k, v in bin_details.items()}
```
What is the use of this? bin_details is already a dictionary

s-aga-r: This is to replace the `None` values with `0`. This happens when there is no Bin for the key (item_code, warehouse). 

Example: 

image

image

s-aga-r: Let's use `frappe.db.get_value()` instead.

rohitwaghchaure: Use COALESCE(SUM(projected_qty),0)

rohitwaghchaure (dismissed):

PR #33449, opened 2022-12-27
fix: Random behaviour while picking items using picklist

deepeshgarg007:
"Get Item Locations" button returned values in random order if the batches didn't have any expiry dates. This in turn led to randomly picking on warehouses, splitting of rows, and overriding of user-entered picked qty incorrectly


deepeshgarg007:
@Mergifyio backport version-14-hotfix version-14

PR #33456, opened 2022-12-27
fix: Multiple rows for same warehouse and batches in pick list

deepeshgarg007:
Continuation of https://github.com/frappe/erpnext/pull/33449

There were still multiple rows getting added for the key `(item_code, warehouse, uom, batch, reference)` which is undesired. Grouped the location by key again post picking the items


deepeshgarg007:

PR #33468, opened 2022-12-28
fix: remove unnecessary permissions from Appointment and Appointment Booking Settings

DaizyModi:
Manual Backport of #33358

### DIFF
--- a/erpnext/crm/doctype/appointment/appointment.py
+++ b/erpnext/crm/doctype/appointment/appointment.py
@@ -5,10 +5,10 @@
 from collections import Counter
 
 import frappe
+import frappe.share
 from frappe import _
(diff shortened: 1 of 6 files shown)


erpnext/crm/doctype/appointment/appointment.py
```diff
@@ -6,7 +6,9 @@
 
 import frappe
 from frappe import _
+from frappe.desk.form.assign_to import add as add_assignment
 from frappe.model.document import Document
+from frappe.share import add
```
sagarvora: `add` isn't intuitive. either import `frappe.share` and use `frappe.share.add` or import `add` as `add_share`.

PR #33509, opened 2023-01-03
fix: from Purchase Order, while creating sales order, is internal customer check was missing in SO

vishdha:
Step1. Create Internal Sales order from PO
Step2. Internal Customer is fetch but is internal customer check was not set
this change will set is_internal_customer = 1

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/sales_invoice.py
@@ -2136,6 +2136,8 @@ def make_inter_company_transaction(doctype, source_name, target_doc=None):
 			currency = frappe.db.get_value("Customer", details.get("party"), "default_currency")
 			target_doc.company = details.get("company")
 			target_doc.customer = details.get("party")
+			target_doc.is_internal_customer = 1
+			target_doc.ignore_pricing_rule = 1
 			target_doc.selling_price_list = source_doc.buying_price_list
 
 			update_address(


vishdha:
Backport to version-14

vishdha:
Can someone review and merge this

deepeshgarg007:
@vishdha In Sales Order `is_internal_customer` check is fetched from the customer master. If that was not fetched means the customer was not an internal customer, don't think this change is required. Was this happening even after save?

vishdha:
> @vishdha In Sales Order `is_internal_customer` check is fetched from the customer master. If that was not fetched means the customer was not an internal customer, don't think this change is required. Was this happening even after save?

Hello @deepeshgarg007 the thing you suggested is correct. but just try to replicate this. 
1. Create internal purchase Order
2. Then create Internal Sales Order from Purchase Order
3. once Sales Order form open it contain Internal Customer but is_internal_customer check was not set.
4. our problem is that we use different series for SO. for Normal So - so.###, for Internal So - ISO.##### based on Is_internal_customer checkbox
5. because this is_internal_customer not set when save document series get set to normal SO.### even if our customer is internal customer.
6. Once Series get set we cannot change series after save
7. same you look for line number 2109 this changes applied for SO to PO.

vishdha:
https://user-images.githubusercontent.com/6947417/212834587-f7e535d6-1bd5-48c9-9e7f-092293f06df0.mp4

deepeshgarg007:
image

@vishdha This is set for the `is_internal_customer` field, this also works on the server side there is no way this doesn't work
The use case of the naming series that you have mentioned can be handled via customizations. Closing this PR

PR #33524, opened 2023-01-03
refactor: revamp process loss feature & added tab breaks

### LINKED ISSUE #33094: Rate of Finished Goods is not calculating correctly in case there is a process Loss Involved
### Information about bug

We are in a Paint Manufacturing Industry.

1. We have a use case where out of Total Qty- 97, out of Which 1 Qty is consider as a Process Loss (Scrap), 
2. When we record the Manufacture Entry of Qty-97, 1 Qty taken as a Scrap, Hence 96 is the actual Qty which is Actual Yield Qty
3. But the Valuation of the above Qty of FG Item is showing wrong- It shows wrong Valuation- It is showing wrong valuation =13521.26/97=139.3944, wheras it should show- 13521.26/96=140.8465, since the actual yield qty is 96 whereas it is considering even the Process Loss Qty also which should not be as this qty is totally loss for me.

image

image

image

image

image

### Module

manufacturing

### Version

image

### Installation method

FrappeCloud

_No response_

rohitwaghchaure:
### Current Design

- User has to set the FG item as Process Loss item in the BOM under Scrap Item table
- If BOM has process loss item then In the Manufacture stock entry, system adds FG item as Is Process Loss item 
- This Process Loss item then add in the Stock with Zero Valuation Rate

Screenshot 2023-01-03 at 7 15 44 PM

### Current Design Drawbacks

- It force to add Process Loss item in the Stock
- The Process Loss shouldn't be the part of Stock
- The cost of the process loss should be added in the Produced Stock

### New Design 

With the new design, system won't be added the Process Loss item in the Stock.

To Set the Process Loss in the BOM

Screenshot 2023-01-03 at 7 02 24 PM

Process Loss impact in the Manufacture Stock Entry

Screenshot 2023-01-03 at 7 01 30 PM

In the above image the "For Quantity" was 100, so FG item should also 100 but as the Process Loss Qty is 20 therefore the FG Item qty set as 80.

Closes https://github.com/frappe/erpnext/issues/33094

### DIFF
--- a/erpnext/manufacturing/doctype/bom/bom.js
+++ b/erpnext/manufacturing/doctype/bom/bom.js
@@ -4,7 +4,7 @@
 frappe.provide("erpnext.bom");
 
 frappe.ui.form.on("BOM", {
-	setup: function(frm) {
+	setup(frm) {
 		frm.custom_make_buttons = {
 			'Work Order': 'Work Order',
 			'Quality Inspection': 'Quality Inspection'
@@ -65,11 +65,11 @@ frappe.ui.form.on("BOM", {
 		});
 	},
 
-	onload_post_render: function(frm) {
+	onload_post_render(frm) {
 		frm.get_field("items").grid.set_multiple_add("item_code", "qty");
 	},
 
-	refresh: function(frm) {
+	refresh(frm) {
 		frm.toggle_enable("item", frm.doc.__islocal);
 
 		frm.set_indicator_formatter('item_code',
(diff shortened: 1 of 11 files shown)


goapaints:
Screenshot 2023-01-24 232207

Manufacturing setting is not created for cost merging of process loss in FG.

goapaints:
@rohitwaghchaure . Please address this bug as mentioned above.

muchai:
Documentation for this please.

PR #33528, opened 2023-01-03
feat: Copy project_name, from_time, to_time from timesheet details to sales invoice.

batonac:
This is a very small patch to the make_sales_invoice function in timesheet.py, allowing the function to copy the project_name, from_time, and to_time fields from the Timesheet Detail to the Sales Invoice Timesheet.

These fields are identical in name and purpose in both documents, but never populated in the current make_sales_invoice script. Interestingly, they do copy when using the "Fetch Timesheet" function from the Sales Invoice, which was what I had used until I figured out this better solution.


deepeshgarg007:
@batonac Please raise PR on `develop` branch, no contributions are directly accepted on the `version-14` branch

PR #33544, opened 2023-01-05
fix: disappear employee image

niyazrazak:
https://user-images.githubusercontent.com/76736615/210723589-87d62851-351f-4c82-ac05-87935a04b073.mp4

Step to replicate

1. Add a employee
2. add image to employee
3. create a user without profile image
4. select userid  in employee

### DIFF
--- a/erpnext/setup/doctype/employee/employee.json
+++ b/erpnext/setup/doctype/employee/employee.json
@@ -180,8 +180,6 @@
    "read_only": 1
   },
   {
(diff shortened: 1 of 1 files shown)


niyazrazak:
@ankush can you review this PR?

ruchamahabal (commented):
Removing "fetch from" isn't the ideal solution. Fetch if empty is enabled here but not respected. So this should fix it instead: https://github.com/frappe/frappe/pull/19586

PR #33547, opened 2023-01-05
fix: incorrect `rate` and `amount` in MR Item

### LINKED ISSUE #25747: v13 bug: foreign currency price copied from sales order->material request->supplier quotation as company currency

**Output of `bench version`**
```
(paste here)
```

1. create and submit new sales order in foreign currency with manual input price e.g 50 USD for new item 
2. from sales order create material request, price 50 and amount copied to material request, but currency changed to company code currency CNY. item price is auto created as 50 CNY
3. create supplier quotation against the material request, 50 CNY adopted to newly created supplier quotation

### Expected result
while copy the price from sales order to material request, for foreign currency case, the exchange rate need to be considered.

```
(paste here)
```

## Additional information

Version 13.2.2

s-aga-r:
closes: #25747

Item Price:

image

Sales Order:

image

Material Request:

image


erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -602,7 +602,12 @@ def update_item(source, target, source_parent):
 			},
 			"Sales Order Item": {
 				"doctype": "Material Request Item",
-				"field_map": {"name": "sales_order_item", "parent": "sales_order"},
+				"field_map": {
+					"name": "sales_order_item",
+					"parent": "sales_order",
+					"base_rate": "rate",
```
rohitwaghchaure: Sales Rate and Purchase Rate won't be the same, maybe you can not mapped the rate fields from sales order to MR. Let the rate on the MR will be fetched from the buying price list.

s-aga-r: Currently, the system is not fetching the rate from buying price list. Will add the fix for that.

PR #33553, opened 2023-01-05
chore: pin ubuntu version for tests

barredterra:
This will make installing dependencies easier, since some of them have builds for a specific ubuntu version (for example, wkhtmltopdf). Also, now it will not fail unexpectedly whenever GitHub decides to deploy the latest ubuntu release.

### DIFF
--- a/.github/workflows/patch.yml
+++ b/.github/workflows/patch.yml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     timeout-minutes: 60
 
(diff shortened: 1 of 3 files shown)


ankush:
I unpinned it so we are ready for next version whenever they are out. It was running 18.04 even though 20.04 was out for long time. 

CI failures are no big deal, at least we will come to know about that as soon as it fails. 


Let's not pin this back, ubuntu-latest exist for this reason only :) 


We can add wkhtml conditionally by checking ubuntu version instead..

barredterra:
Okay, we have conflicting goals here :D

Closing this PR.

PR #33579, opened 2023-01-09
fix(stock entry): wrong valuation rate in repack

rtdany10:
Before:

https://user-images.githubusercontent.com/52111700/211315916-d6a3e899-fa70-40be-bb39-6ab3f28d3bd5.mp4

After:

https://user-images.githubusercontent.com/52111700/211315944-b4f595fb-b9fe-499e-b6c2-c18e086cd63f.mp4


s-aga-r (changes requested):
@rtdany10 please add a test case for the same.

rtdany10:
> @rtdany10 please add a test case for the same.

Will send a PR with test case.

@rohitwaghchaure can we backport this to v13 as well?

PR #33581, opened 2023-01-09
refactor: convert heatmap queries to QB

ankush:
Uses new `UnixTimestamp` function, don't backport.

### DIFF
--- a/erpnext/projects/doctype/project/project.py
+++ b/erpnext/projects/doctype/project/project.py
@@ -7,6 +7,8 @@ from email_reply_parser import EmailReplyParser
 from frappe import _
 from frappe.desk.reportview import get_match_cond
 from frappe.model.document import Document
+from frappe.query_builder import Interval
+from frappe.query_builder.functions import Count, CurDate, Date, UnixTimestamp
 from frappe.utils import add_days, flt, get_datetime, get_time, get_url, nowtime, today
 
 from erpnext import get_default_company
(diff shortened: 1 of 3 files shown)


ankush:
Tested manually, none of these are tested in CI to best of my knowledge (overkill too)

barredterra:
In cases where the database server has a different timezone from the site's **System Settings**, this seems to produce incorrect results. Demonstrated by this CI failure:

```
======================================================================
 FAIL  test_heatmap_data (erpnext.stock.doctype.item.test_item.TestItem.test_heatmap_data)
  File "/home/runner/frappe-bench/apps/erpnext/erpnext/stock/doctype/item/test_item.py", line 682, in test_heatmap_data
    self.assertTrue(one_year_ago <= timestamp <= now)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    count = 1
    data = {1775692800: 1}
    now = 1775679080.6278868
    one_year_ago = 1744056680.6278868
    self = <erpnext.stock.doctype.item.test_item.TestItem testMethod=test_heatmap_data>
    time = <module 'time' (built-in)>
    timestamp = 1775692800
AssertionError: False is not true
```

https://github.com/frappe/erpnext/actions/runs/24154573095/job/70493792157?pr=54137

> The server interprets date as a value in the current [time zone](https://mariadb.com/docs/server/reference/data-types/string-data-types/character-sets/internationalization-and-localization/time-zones) and converts it to an internal value in [UTC](https://mariadb.com/docs/server/reference/data-types/string-data-types/character-sets/internationalization-and-localization/coordinated-universal-time). – https://mariadb.com/docs/server/reference/sql-functions/date-time-functions/unix_timestamp

That means the failure can happen with no recent change at all, just depending on when the code runs:

- site date has already rolled over to Apr 9 in `Asia/Kolkata`
- the heatmap bucket gets converted as midnight UTC
- the test compares that against `time.time()` before UTC midnight
- result: a “future” timestamp for a few hours

If the shard containing `test_item.py` reaches this test before UTC midnight but after IST midnight, and
if nothing else aborts that shard first, then that run should fail.

Another possibility is that first system timezone is set to A, test records are created, system timezone is changed to B, tests are executed. Where the exact timing of test record creation and timezone changes varies, to produce unpredictable results.

PR #33347, opened 2022-12-15
feat: Add operating cost based on bom quanity without creating job card

vishdha:
Problem Statement: Across Manufacturing organisation with multiple operation and illiterate job worker its difficult to create job card. Hence it should be record operating cost per finished good Item. 
Step 1. Add Per BOM quantity operating cost in bom
Step 2. You will see Planned operating cost in BOM
Setp 3. In Manufactured Stock Entry it will record operating_cost_per_bom_quantity * FG Quantity 

Why this change:
1. Need operating cost but want to skip job card need sku based operating cost.

Step1. Add operating cost per BOM quanity. operating cost = 45*10 = 450 as shown in pic
BOM-Makhana_7-107-006

Final Manufactured stock entry without job card
Manufacture-MAT-STE-2022-00088

GIF:

https://user-images.githubusercontent.com/6947417/207784936-d9159dfb-c2de-43a0-a6b7-2d41503b8cac.mp4

`no-docs`

PR #33590, opened 2023-01-09
Revert "feat: Add operating cost based on bom quanity without creating job card"

rohitwaghchaure:
Reverts frappe/erpnext#33347


rohitwaghchaure:
Missing modified time in the JSON file

PR #33603, opened 2023-01-10
FIX: Remove usage of "six.string_types"

rahulporuri:
`six` is no longer a dependency so relying on `six.string_types` should be causing an error. This PR replaces `six.string_types` with just `str`.


ankush:
@frappe/erpnext-maintainers 

related: Setup pyupgrade and force new syntax: https://github.com/frappe/frappe/pull/17367

PR #33495, opened 2022-12-31
fix(stock): missing ordered_qty in get_bin_details

dj12djdjs:
closes: https://github.com/frappe/erpnext/issues/33473

PR #33617, opened 2023-01-11
Revert fix(stock): missing ordered_qty in get_bin_details #33495

s-aga-r:
revert: https://github.com/frappe/erpnext/pull/33495


ankush:
@s-aga-r this is a wrong fix. We dont need to revert this PR. The problem is with us storing dynamic data in DB through `onload`.

s-aga-r:
> @s-aga-r this is a wrong fix. We dont need to revert this PR. The problem is with us storing dynamic data in DB through `onload`.

@ankush closing this.

PR #33619, opened 2023-01-12
fix: rewrite logic for duplicate check in Item Attribute

### LINKED ISSUE #15467: Item Variant Abbreviations always defaulting to uppercase
I am not sure if this is by design, but for us it is quite annoying to have the abbreviations when creating item variants always convert to uppercase values.

It seems like those are only meant for T-Shirts where I can imagine this to be okay, but we would like to store different product variants in there like:

* Google-Apps
  * -Unlimited
  * -For-Business
  * -Enterprise

Going further it would also help to completely override (not append) for variant values as sometimes the Data we would have to store is not matching a pattern (Google-Apps-Enterprise would be 1010020020 in reality, which we got to translate in the backend already).

### LINKED ISSUE #22256: Why is it necessary to have upper case item attribute abbreviation?
https://github.com/frappe/erpnext/blob/f97c5dad82084664d0b2a04e79f8b79ca36346a7/erpnext/stock/doctype/item_attribute/item_attribute.py#L56

Shouldn't it be based on whatever is written? If the customer wants, then the attribute can be upper cased using Javascript

SvbZ3r0:
Previously, Item Attribute values were not checked for case-insensitive duplicates, and Item tttribute abbreviations were forced to be uppercase. This commit fixes both problems.

Fixes: #22256
Partially fixes: #15467

### DIFF
--- a/erpnext/stock/doctype/item_attribute/item_attribute.py
+++ b/erpnext/stock/doctype/item_attribute/item_attribute.py
@@ -75,11 +75,9 @@ class ItemAttribute(Document):
 		values, abbrs = [], []
 		for d in self.item_attribute_values:
 			if d.attribute_value.lower() in map(str.lower, values):
-				frappe.throw(
-					_("Attribute value: {0} must appear only once").format(d.attribute_value.title()))
+				frappe.throw(_("Attribute value: {0} must appear only once").format(d.attribute_value.title()))
 			values.append(d.attribute_value)
 
 			if d.abbr.lower() in map(str.lower, abbrs):
-				frappe.throw(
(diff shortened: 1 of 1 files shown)


s-aga-r:
@SvbZ3r0 please fix the failing test.

SvbZ3r0:
I've fixed the formatting, but I'm not sure what the other failing tests have to do with any changes I made here.

SvbZ3r0:
@s-aga-r This is passing all tests. Let me know if I need to change anything else.

s-aga-r (approved):

PR #33622, opened 2023-01-12
fix: Updating SO throws ordered_qty not allowed to change after submission

ruthra-kumar:
Fixes Not allowed to change 'Order Qty' after submission while updating Sales Order
Screenshot 2023-01-12 at 1 25 11 PM

regression: https://github.com/frappe/erpnext/pull/33495


ruthra-kumar:

PR #33624, opened 2023-01-12
fix: apply user perm in sales register

rtdany10:


rtdany10:
The reason I raised this PR was, in a multi company scenario:
If a user has access only to **Company B** and the default company as per global defaults is **Company A**, then, that user will be able to view details of the **Company A** in this(and other) reports. Company is not a mandatory filter. The report by default opens with empty company filter and outputs all data, irrespective of perms.

rtdany10:
We need a better solution for this, maybe something like setting the session/global default based on user perms.
Thoughts @deepeshgarg007 @ruthra-kumar @ankush 

Edit: Even if we set the default company, the user can still empty out the company filters in this report and view details of the other :/

deepeshgarg007:
@rtdany10 Lets make company mandatory for all these reports, will it solve your issue then?

rtdany10:
> @rtdany10 Lets make company mandatory for all these reports, will it solve your issue then?

Ideally, yes, that should fix the issue.

But there is another concern that popped up in my mind: If the management wants to view the full data of multiple companies, there won't be any way to do that :grimacing:

deepeshgarg007:
Add a hidden company column in the reports and then check if that works, this way only the company records for which the user has permission will be fetched.

rtdany10:
.

PR #33646, opened 2023-01-13
Revert "fix: Updating SO throws ordered_qty not allowed to change after submission"

ruthra-kumar:
Reverts frappe/erpnext#33622


ruthra-kumar:

PR #33662, opened 2023-01-15
revert: Reverting changes done on 33495

ruthra-kumar:
This is causing issues in Drop ship creation. https://github.com/frappe/erpnext/pull/33495#issuecomment-1381979649

### DIFF
--- a/erpnext/stock/doctype/item/test_item.py
+++ b/erpnext/stock/doctype/item/test_item.py
@@ -106,7 +106,6 @@ class TestItem(FrappeTestCase):
 			"conversion_factor": 1.0,
 			"reserved_qty": 1,
 			"actual_qty": 5,
-			"ordered_qty": 10,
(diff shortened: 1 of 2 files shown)


s-aga-r:
@ruthra-kumar https://github.com/frappe/erpnext/pull/33617#issuecomment-1378691488

ruthra-kumar:
> @ruthra-kumar [#33617 (comment)](https://github.com/frappe/erpnext/pull/33617#issuecomment-1378691488)

@s-aga-r 
A dynamic `ordered_qty` will break Purchase Order creation from Sales Order as mentioned in the description. A safer approach(without using onload()) to load dynamic quantities(projected, actual, reserved ) is needed.  But, even that approach shouldn't fetch `ordered_qty`.

ruthra-kumar:

PR #33675, opened 2023-01-16
chore: ERPNext setup wizard cleanup

deepeshgarg007:
Did a minor cleanup to resolve the above-mentioned issues

- Removed the image attachment fields, very less users upload them during setup
- Fields like Company Abbreviation and Bank Account were not quite necessary and understood by a user signing up for the first time so removed the fields completely, the Company Abbreviation field is actually hidden as it’s still needed for the setup and is by default taken as the initials of the name of the company
- Added a default company tagline for the website page and removed it from the setup wizard 
- Changed a few slide titles

Post the cleanup the ERPNext setup wizard is reduced from two slides to one

image


dj12djdjs:
@deepeshgarg007 I think we can add Company Abbreviation back? It isn't possible to change after initial setup. And every single company I've setup on ERPNext I've customized the abbreviation.

PR #33268, opened 2022-12-08
fix: creating internal sales order from PO not fetching Material Request

vishdha:
fixed: material Request not fetch into sales order for Internal sales order in Purchase order issue fixed

Issue #27874, opened 2021-10-08
Create Sales Invoices from Delivery Note : Missing dashboard link to Delivery note


FHenry:
Reopen  #27359 close by merge, but merge was reverted by #23570
So the bug still occurs until a fix is develop

sharmanakul:
![image](https://user-images.githubusercontent.com/88872881/141494308-6223edd2-4d85-46e0-bd39-e0c3253b8174.png)
numbers are not displayed next to the link.

FHenry:
I fix it with a hook on custom app, for one of my integration the use case is more relevant with it.

wojosc:
Hi,

I am not sure if we are all on the same issue here. I have the case, that I created a DN from SO
![grafik](https://user-images.githubusercontent.com/22279621/157233316-3fbf34cc-dee8-4dd5-9d61-ac1989ec2c94.png)

then created SI from DN. On DN I can see the links to SO and SI
![grafik](https://user-images.githubusercontent.com/22279621/157233477-ce009220-8c92-4f09-86fd-881c838cab62.png)

On the Sales invoice there is no backlink to DN though
![grafik](https://user-images.githubusercontent.com/22279621/157233575-124fe33f-7b8d-4691-ad23-bd29fce9c326.png)

Although the link exists on item level on the SI
![grafik](https://user-images.githubusercontent.com/22279621/157233664-1b510196-1caa-4470-8acf-e4025616fe6a.png)

thank you

FHenry:
Yes the link is into line details, but connexion dashboard is a more direct link. Users don't want to click on each lines details (but in version-13, it's more easy to add column into items table)

PR #33690, opened 2023-01-16
fix: Sales Order Connections Tabs do not show linked Material Request or "+" button  (intoduce by #33304)

FHenry:
ERPNext develop, versoin-14, version-13

How to reproduce :
Create a Sales order
Create a Material request from the sales order (use create blue button as "+" is missing) => Go to Connections tabs => Go back to originated Sales Order
Check Sales Order Connections Tab, the materials Request is not here.

Before PR:

https://user-images.githubusercontent.com/1050053/212759149-7daaed86-d408-4319-ba39-e6ea5d95a8c4.mp4

After PR:

https://user-images.githubusercontent.com/1050053/212758598-3cdba92d-f8c3-4a34-b139-ab9238816b97.mp4

Note To Reviewer:
 - I don't know why @s-aga-r have (or I have) different result, but I test from fresh install without any custom app or customization in version-13, version-14 and develop, and this PR broke exactly what it should to fix 
 - Linters error is not related with this change.

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order_dashboard.py
+++ b/erpnext/selling/doctype/sales_order/sales_order_dashboard.py
@@ -14,7 +14,6 @@ def get_data():
 		},
 		"internal_links": {
 			"Quotation": ["items", "prevdoc_docname"],
-			"Material Request": ["items", "material_request"],
 		},
 		"transactions": [
 			{


FHenry:
When Sales Order is displayed
There is the Frappe API Rest Call
/api/method/frappe.desk.notifications.get_open_count
 
doctype | Sales Order
name | SAL-ORD-2023-00001
items | ["Sales  Invoice","Pick List","Delivery Note","Maintenance Visit","Material  Request","Purchase Order","Project","Work Order","Quotation","Auto  Repeat","Payment Entry","Payment Request","Journal Entry"]

URI param "items" already include Material Request, The is no need to specify it into get_data() on sales_order_dashboard.py. 

Either Add "Material Request": ["items", "material_request"], in get_data() on sales_order_dashboard.py is strange, shouldn't it be "Material Request": ["items", "sales_order"], but there is no need as frappe already manage it with "transaction" section ? (manage by frappe/model/meta.py : add_doctype_links()

That can explain why result of API call didn't return at all Material Request count as the field material_request don't exists into Material Request Item

```
{
	"message": {
		"count": [
			{
				"name": "Sales Invoice",
				"count": 0
			},
			{
				"name": "Pick List",
				"count": 0
			},
			{
				"name": "Delivery Note",
				"count": 0
			},
			{
				"name": "Maintenance Visit",
				"count": 0
			},
			{
				"name": "Purchase Order",
				"count": 0
			},
			{
				"name": "Project",
				"count": 0
			},
			{
				"name": "Work Order",
				"count": 0
			},
			{
				"name": "Auto Repeat",
				"count": 0
			},
			{
				"name": "Payment Entry",
				"count": 0
			},
			{
				"name": "Payment Request",
				"count": 0
			},
			{
				"name": "Journal Entry",
				"count": 0
			}
		]
	}
}
```

s-aga-r:
@FHenry there are two flows, I have added the fix for the [Internal Transfer](https://github.com/frappe/erpnext/pull/33268#issuecomment-1349011101) but it breaks the SO --> MR [MR link in SO]. AFAIK the `material_request` and `material_request_item` fields not getting updated when a Material Request gets submitted. To fix this we need to update the same on submit of MR.

FHenry:
@s-aga-r Ok, Thank you very much for clarification.

So that a similar problem than this issue https://github.com/frappe/erpnext/issues/27874
We actually can't have more than 1 way to count related linked elements

s-aga-r:
> @s-aga-r Ok, Thank you very much for clarification.
> 
> So that a similar problem than this issue #27874 We actually can't have more than 1 way to count related linked elements

@FHenry will be able to see the MR ref(s) in SO if we set the `material_request` and  'material_request_item' in SO on submission of MR, but I am not sure about the `+` button will be visible or not. Will try it locally and raise a PR if it works.

PR #33710, opened 2023-01-18
Removed an unnecessary check in code which always evaluates to true

openrefactory:
In file: deferred_revenue_and_expense.py, method: generate_report_data, a logical expression uses the identity operator. A new object is created inside the identity check operation and then used for matching identity. Since this is a distinct, new object, it will not have identity an match with anything else. As a result, the identity check will have a logical short circuit. Moreover in line 378, we already create an entry inside a list. So, the check for emptiness is unnecessary. Currently, the test expression in line 381 is passing all the time.

We removed that unnecessary test expression.


ruthra-kumar (approved):

ruthra-kumar:

PR #33735, opened 2023-01-19
fix: use stock qty to calculate POS reserved stock

xeanth8:
Use `stock_qty` when calculating POS reserved stock. Currently it uses the `qty` field which doesn't represent UOM's conversion factor, making the available stock incorrect.


vishdha:
backport to version 14

PR #33739, opened 2023-01-19
fix: calculate correct amount for qty == 0

barredterra:
When entering `0` as line item quantity, the calculated amount was not correct. The code checked for `!item.qty`, which is `true` when `item.qty == 0`, and in this case used `1` (or `-1` for returns) as the qty instead, resulting in an incorrect amount.

Before:

```python
qty = 0
rate = 10
amount = (qty or 1) * rate  # == 10
```

After:

```python
qty = 0
rate = 10
amount = (1 if qty is None else qty) * rate  # == 0
```

Fixed in JS by checking for `item.qty === undefined`.

Also used the opportunity to replace the jQuery `$.each` with a `for .. of` loop.

Internal reference: LAN-714


s-aga-r:
@barredterra the `amount` and `net_total` is not calculating correctly:

image

PR #33770, opened 2023-01-21
refactor(region): Splitting of KSA(Saudi Arabia) Region from ERPNext.

sankarsubramaniankvs:
This Pull Request Will remove the code related to Saudi Arabia Region from ERPNext. It will come in a separate app that will be published in frappe marketplace and will be used only by companies that need KSA Compliance. The idea behind it is to remove code unnecessary for a user. ERPNext should only have generic code that can be used by everyone.


rtdany10:
@sankarsubramaniankvs are you guys working on phase 2 of KSA e-Invoicing?

sankarsubramaniankvs:
Yes. We are working on cleartax integration for KSA region. That covers phase 2

erpnext/patches/v13_0/disable_ksa_print_format_for_others.py
```diff
@@ -3,7 +3,7 @@
 
 import frappe
 
-from erpnext.regional.saudi_arabia.setup import add_print_formats
+from ksa.install import add_print_formats
```
deepeshgarg007, line 6: All the patches should also be moved to the custom app, there should be no dependency on any other app

erpnext/setup/setup_wizard/operations/taxes_setup.py
```diff
@@ -108,11 +108,16 @@ def from_detailed_data(company_name, data):
 
 def update_regional_tax_settings(country, company):
 	path = frappe.get_app_path("erpnext", "regional", frappe.scrub(country))
-	if os.path.exists(path.encode("utf-8")):
+	if os.path.exists(path.encode("utf-8")) or frappe.scrub(country) == 'saudi_arabia':
 		try:
-			module_name = "erpnext.regional.{0}.setup.update_regional_tax_settings".format(
-				frappe.scrub(country)
-			)
+			if frappe.scrub(country) == 'saudi_arabia':
```
deepeshgarg007, line 113: This validation is not needed

deepeshgarg007 (changes requested):
No git history is being transferred right now on the repo https://github.com/8848digital/KSA
Please follow https://github.com/frappe/erpnext/wiki/Guide-to-splitting-modules-from-ERPNext

PR #33784, opened 2023-01-22
refactor: TaxController, TaxesandChargesTemplate

barredterra:
- `TaxesandChargesTemplate`: common base class for Sales and Purchase Taxes and Charges Template
- `TaxController`: common base class for transaction taxes and charges (child table)


barredterra:
Let's say this was for practice and needs a second attempt.

PR #33788, opened 2023-01-23
fix: manual depr entry not updating asset value [v14]

anandbaburajan:
1. When manual depreciation entries (created using the `Create Depreciation Entry` option in the `Manage` dropdown of assets with `calculate_depreciation` not checked) are submitted, the value of the asset isn't updated, so fixed that.
2. We have quite a few functions which (sometimes incorrectly) performed the same task of fetching an asset's current value after depreciation, so refactored all of them and now there's just one function.
3. The fixed asset register and the chart on asset documents weren't reflecting manual depr entries, so fixed them.
4. Wrote a patch to update value of all assets which have manual depr entries created for them.
5. In the fixed asset report register, assets with finance books weren't filtered properly resulting in an `TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'` error, so fixed it.


erpnext/accounts/doctype/journal_entry/journal_entry.py
```diff
@@ -225,6 +226,32 @@ def apply_tax_withholding(self):
...
+
+		processed_assets = []
+
+		for d in self.get("accounts"):
+			if (
+				d.reference_type == "Asset" and d.reference_name and d.reference_name not in processed_assets
+			):
+				processed_assets.append(d.reference_name)
+
+				asset = frappe.get_doc("Asset", d.reference_name)
```
deepeshgarg007: Use `get_value` instead of `get_doc`

deepeshgarg007 (changes requested):

barredterra:
Is there a corresponding PR that implements these changes on the develop branch?

anandbaburajan:
@barredterra not yet, I've to think more about some things. I'll open a PR in a day or two.

barredterra:
@AnandBaburajan isn't the correct procedure to send PRs to develop, then (partially) backport them? This way we can avoid features going missing with an upgrade. For example, if somebody upgrades to develop right now, your features will not be there.

anandbaburajan:
> @AnandBaburajan isn't the correct procedure to send PRs to develop, then (partially) backport them?

Yep, but like I said, I've to think more about some of those features/fixes (there might be some design changes) which would go into develop and fixing bugs in v13 and v14 was the first priority so opened this PR.

barredterra:
Understood. IMHO it doesn't make sense to solve a bug in a lower version and "maybe sometime" (to exaggerate) in the `develop` branch. This has lead to problems in the past and will continue to do so, unless we follow the proper process.

PR #33789, opened 2023-01-23
chore: update translations for naming series error

anandbaburajan:
Updating a naming series error message at https://github.com/frappe/frappe/pull/19711, so changing the translations here.


ankush:
@AnandBaburajan this wont work cause change in quotes. Translations need to match exact string.

PR #33795, opened 2023-01-23
 refactor: Selling Settings, Quotations

mnatalia:
We need a possibility to create several Sales Orders from Quotation. So Selling Settings doctype and quotation.js were updated.

- `Selling Settings`: added new field `allow_against_multiple_quotations`
- `Quotation`: common base class for transaction taxes and charges (child table)

> Screenshots/GIFs


mnatalia:
sorry for  bad PR, will fix and update

PR #33808, opened 2023-01-24
fix: make title field readonly

### LINKED ISSUE #33806: Can't rename Lead
### Information about bug

Creating lead then save it. Then rename it, won't rename it.

### Module

CRM

### Version

ERPNext: v14.12.1 (version-14)
Frappe Framework: v14.22.0 (version-14)

### Installation method

manual install

```shell
No error, but appears a success message and the rename is not done.
```

dj12djdjs:
closes: https://github.com/frappe/erpnext/issues/33806#issuecomment-1402804212

Make the title field read_only so when the user clicks the title, the rename dialog won't appear.

### DIFF
--- a/erpnext/crm/doctype/lead/lead.json
+++ b/erpnext/crm/doctype/lead/lead.json
@@ -312,7 +312,8 @@
    "fieldtype": "Data",
    "hidden": 1,
    "label": "Title",
-   "print_hide": 1
+   "print_hide": 1,
+   "read_only": 1
   },
   {
    "fieldname": "language",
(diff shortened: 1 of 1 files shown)


dj12djdjs:
please backport version-14

erpnext/crm/doctype/lead/lead.json
```diff
@@ -514,11 +515,10 @@
  "idx": 5,
  "image_field": "image",
  "links": [],
- "modified": "2022-10-13 12:42:04.277879",
+ "modified": "2023-01-24 18:20:05.044791",
  "modified_by": "Administrator",
  "module": "CRM",
  "name": "Lead",
- "name_case": "Title Case",
```
deepeshgarg007, line 521: @dj12djdjs Why is this removed?

dj12djdjs: @deepeshgarg007 I'm not sure. I only checked Read Only on the title field. Clicked Save.

This actually happened last on my last PR changing a doctype definition. Seems anytime I save a doctype this line gets removed.

PR #33815, opened 2023-01-25
fix: disfuctional cost center filter on Journal Entries

ruthra-kumar:
Cost center filter didn't apply on Journal entries on Payment Reconciliation tool. Has been fixed now. Added a test case to catch regression.


ruthra-kumar:

PR #33822, opened 2023-01-26
fix: enable customs in Selling Workpace by default

FHenry:
Like in all other version (version-14 and develop), and like all other defualt Workspace, there is no reason the hide custom doctype and custom report by default in this workspace


deepeshgarg007:
@FHenry This won't have any effect unless the modified timestamp is updated. Please update from UI and then reopen the PR

FHenry:
@deepeshgarg007 done here #33853 
May be next time, could you just ping me and not close the PR ?

PR #33835, opened 2023-01-27
fix(patch): validation error on cost center allocation migration

ruthra-kumar:
If Distributed cost centers have GL postings on patch run date, patch fails with validation error.
Skip `from_date` validation during patch run.

validation_error


ruthra-kumar:

PR #33850, opened 2023-01-28
fix: Currency symbol for tax withholding net total field

deepeshgarg007:
After Fix:
Screenshot 2023-01-28 at 7 08 20 PM


bhavyajicomputers:
thanks

PR #33852, opened 2023-01-28
fix: Ignore linked JE on JE cancellation

deepeshgarg007:
image

Ignore and just unlike the reconciled Journal Entry on canceling


anandbaburajan:
I was going to do the same because of:

image

It's a depreciation entry for an asset. A user shouldn't need to cancel all those linked documents just to cancel that entry. I'll add all those doctypes (mentioned in the screenshot) to `ignore_doctypes_on_cancel_all` in another PR.

PR #33863, opened 2023-01-29
feat(stock): Support more barcodes in an item by validate the barcode with the barcodenumber module

developsessions:
I implemented a feature add the support for more barcode types than just an EAN or UPC-A. I used the already available barcodenumber module for this.
Some people had already an request to have an feature like this.

`no-docs`

### DIFF
--- a/erpnext/stock/doctype/item/item.py
+++ b/erpnext/stock/doctype/item/item.py
@@ -358,7 +358,7 @@ class Item(Document):
 					check_list.append(d.item_tax_template)
 
 	def validate_barcode(self):
-		from stdnum import ean
+		import barcodenumber
 
 		if len(self.barcodes) > 0:
 			for item_barcode in self.barcodes:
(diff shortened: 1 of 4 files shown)


erpnext/stock/doctype/item_barcode/item_barcode.json
```diff
@@ -25,7 +25,7 @@
    "fieldtype": "Select",
    "in_list_view": 1,
    "label": "Barcode Type",
-   "options": "\nEAN\nUPC-A"
```
ankush, line 28: This will invalidate existing UPC-A barcodes. Write a thin compatibility layer to translate ERPNext types and barcode library types. Can be a simple constant dictionary in file.

ankush (changes requested):
Add tests for newly supported barcode types. 

Example: https://github.com/frappe/erpnext/blob/1b967f37afe61f74f64307215848d7a811b759ba/erpnext/stock/doctype/item/test_item.py#L568-L615

developsessions:
Can anyone start the github workflows again?

developsessions:
Restart the workflow please

developsessions:
@ankush is this request ready to merge?

deepeshgarg007 (approved):

erpnext/stock/doctype/item/item.py
```diff
@@ -985,6 +982,22 @@ def validate_auto_reorder_enabled_in_stock_settings(self):
 				)
 
 
+def convert_erpnext_to_barcodenumber(erpnext_number):
+	convert = {
+		"UPC-A": "UPCA",
+		"CODE-39": "CODE39",
+		"EAN": "EAN13",
+		"EAN-12": "EAN",
+		"EAN-8": "EAN8",
```
ankush, line 991: This is somewhat problematic, earlier we allowed all different EAN types to be considered as just "EAN"


reference  https://github.com/frappe/erpnext/issues/34233

ankush: Let's relax it. The point of this validation is to avoid errors, they shouldn't become a nuisance :)

developsessions: @ankush: Should we implement the behaviour as before? (Allow all different EAN types to be considered as just "EAN")
If yes, I can do that. I was not clear about the previous behavior.

ankush: @developsessions yeah that would be better for existing users. Otherwise resaving item will start failing.

developsessions: @ankush you're right. I will prepare the change and create an pull request #34250 for this fix.

developsessions:
When gets this feature released to version-14?
Because I would need it there.

PR #33874, opened 2023-01-30
feat: Support for Alternative Items in Quotation

### LINKED ISSUE #33609: Alternative line items in Quotation
Summary: I want to offer multiple variations of an item or service at different prices in the same **Quotation**. Only the main item or service shall be included in the total. The customer can decide to order any one of the proposed variations.

Let's say I'm selling flights. I want to tell my customer the prices for Economy, Business and First class. The Quotation should look like this:

Sr | Item Name | Is Alternative | Qty | Rate | Amount
---|------------|:--------------:|----:|------:|---------:
1 | Economy | | 2 | 150 | 300
2 | Business | ✓ | 2 | 200 | 400
3 | First Class | ✓ | 2 | 400 |  800
4 | **Total** | | | | **300**

The first line item is the main product offered, in this case. All following line items that have _Is Alternative_ checked, are alternatives to the first line item. Note that an alternative line item may have the same item code as the main item (but a different description).

The amount of alternative line items should be excluded from the total. It does not make sense to include them in the total, because that would make the value much higher than anything the customer will ever pay (1.500 instead of the maximum possible value of 800 for two first class tickets).

When the **Quotation** is turned into an order, exactly one of these positions must be chosen. The alternative line item only applies to **Quotation**. In **Sales Order** etc. all line items are final. For example, the **Sales Order** might look like this:

Sr | Item Name | Qty | Rate | Amount
---|------------|----:|------:|---------:
1 | Business | 2 | 200 | 400
2 | **Total** | | | **400**

This counts as a full order of all quoted items.

Internal reference: DEP-431

marination:
> **Note:** Not to be confused with DocType "Item Alternative"

Documentation Draft: https://docs.erpnext.com/docs/v14/user/manual/en/selling/quotation/edit-wiki?wiki_page_patch=b1a88d79a9

Please check https://github.com/frappe/erpnext/issues/33609 to understand the issue

<details>
<summary>Impact on Quotation</summary>

> **Note:** It is important to maintain the right order i.e. Alternative Item rows must follow a Non-alternative item row. Grouping will be done on this basis
> Screenshot 2023-02-06 at 4 38 13 PM

- Users can add an alternative item in the Quotation to give the customer a chance to choose among items in the quotation
- To add an alternative item: add a row and check **Is Alternative** 
    Screenshot 2023-02-06 at 4 31 18 PM

- **Totals**: Only consider original (non-alternative Item)
  Screenshot 2023-01-31 at 5 02 44 PM
- **Taxes**: Applied accordingly
  Screenshot 2023-01-31 at 5 03 45 PM
</details>

<details>
<summary>Mapping to Sales Order</summary>
Consider:
Screenshot 2023-01-31 at 5 05 31 PM

- User will be asked to select an item from among the original-alternatives set, to map in the Sales Order
   

- One of the alternatives is selected, the same is mapped and the rest in the set are skipped. Apart from that the simple items are also mapped:
   
</details>

<details>
<summary>Use Cases</summary>

- Item Code has alternatives which are different Items
   Screenshot 2023-02-06 at 4 51 28 PM
- Service Item has alternatives and non-alternatives sharing the same Item code. The case here is that the service costing is very dynamic and subjective and so the **same code** with different Item name/Description is used in the Quotation.
 Screenshot 2023-02-06 at 4 53 46 PM

</details>

### Todo

- [x]  Partially ordered status
- [x] Tests


erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -220,17 +234,111 @@ erpnext.selling.QuotationController = class QuotationController extends erpnext.
 			}
 		})
 	}
+
+	show_alternative_items_dialog() {
+		var me = this;
```
deepeshgarg007:
```suggestion
		let me = this;
```

deepeshgarg007 (commented):
The dialog won't work if someone uses the "Get Items From" option directly from the sales order, right?

marination:
@deepeshgarg007 Yes it won't. I'll fix that

marination:
@deepeshgarg007 in SO/SI item selection is not possible/doesn't make sense, because i don't think any user wants to select alternatives from a mountain of rows from multiple Quotations.

So i added a disclaimer in the SO dialog saying that if alternatives are involved we would recommend mapping from the Quotation. And then if the user proceeds anyway, only non alternative rows are mapped in the Sales Order (the rows that sum up to the total in the quotation)
Screenshot 2023-02-20 at 9 21 19 PM

maveezmh:
@marination how about having a provision where user can deliver an alternative because given item is not aviailable in stock

PR #33886, opened 2023-01-30
fix: increase the width of Posting Date Column on General Ledger Report

iramluism:
Despite the posting date on General Ledger Report dont render collapsed, the column name no render properly.  In this change incremented the width of column to 120. And it avoid to extend the width manually to see the column name. 

Before 
image

Then
image


deepeshgarg007:
@iramluism Already did via https://github.com/frappe/erpnext/pull/33858
Closing this PR

PR #33902, opened 2023-01-31
fix: Amount for debit and credit notes with 0 qty line items

deepeshgarg007:
Partially reverts: https://github.com/frappe/erpnext/pull/33739

There are scenarios where the user creates Debit and Credit notes with 0 qty just to adjust the amount against a previously issued invoice. No goods are returned or shipped in these scenarios just the amount against that item is adjusted.


barredterra:
@deepeshgarg007 one of our customers ran into this issue again. It freaks people out to see `0 * x = x` 😄 . Should we maybe make this configurable?

PR #33904, opened 2023-01-31
fix: incorrect actual_qty in bin

rohitwaghchaure:


s-aga-r (commented):
Why rewriting `qb` in `sql`? `qb` has `for_update()`
https://github.com/frappe/erpnext/blob/c5834c1db41db1379d428026ce1c8985919ff8b0/erpnext/stock/doctype/pick_list/pick_list.py#L444

PR #33912, opened 2023-02-01
fix: don't allow to cancel non completed repost item valuation

rohitwaghchaure:
If user trying to cancel the non completed repost item valuation then system will throw the below error 

Screenshot 2023-02-01 at 10 16 58 AM


erpnext/stock/doctype/repost_item_valuation/repost_item_valuation.py
```diff
@@ -83,6 +83,17 @@ def set_status(self, status=None, write=True):
 		if write:
 			self.db_set("status", self.status)
 
+	def on_cancel(self):
+		self.validate_repost_cancelled()
```
ankush, line 87: There's similar `before_cancel` hook already. Maybe modify if to add if any corner cases are being missed.

PR #33944, opened 2023-02-03
fix: loss Opportunity Report Issue

Komal-Saraf0609:
On creating a lost opportunity it doesn't get shown up in the "Lost Opportunity" report.  Refer below SS:

Before:

Opportunity:
Screenshot 2023-02-03 at 8 28 18 PM

Report:
Screenshot 2023-02-03 at 8 29 32 PM

After:
Screenshot 2023-02-03 at 8 31 18 PM

Same issue existed for "Lead Details" report as well. Fixed the same in the above PR.

### DIFF
--- a/erpnext/crm/report/lead_details/lead_details.py
+++ b/erpnext/crm/report/lead_details/lead_details.py
@@ -4,6 +4,7 @@
 
 import frappe
 from frappe import _
+from frappe.utils import get_datetime
 
 
 def execute(filters=None):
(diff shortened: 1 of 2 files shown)


erpnext/crm/report/lead_details/lead_details.py
```diff
@@ -69,6 +70,10 @@ def get_columns():
 
 
 def get_data(filters):
+	if filters.get("from_date") and filters.get("to_date"):
```
deepeshgarg007, line 73: @Komal-Saraf0609 This doesn't seem like a correct fix, can you please elaborate on how did you reach this fix?

PR #33947, opened 2023-02-03
fix: exchange rate revaluation errors

dj12djdjs:
1. f4d580ac6025639f87fb106282d9e13be7acf9a3 
Previous Code:
https://github.com/frappe/erpnext/blob/2afed7de0393a09e00773aa300a35ab56f56e663/erpnext/accounts/doctype/exchange_rate_revaluation/exchange_rate_revaluation.py#L215-L232

In the event d.balance is < 0 and d.balance_in_account_currency == 0 then
gain_loss = current_exchange_rate * 0
This will always result in 0. Therefor, keeping the company currency balance open.

2. 08fd991ccae9abcffe9f7c0e07f6165a07b40fd6
Journal Entry will fail to submit if both debit, and credit are zero.

3. e801be203c09bb492d22e69cf649a6e81a4aed30
If there is any remaining balance at the end of revaluation of exchange rates, add that difference to exchange/loss account.


dj12djdjs:
Failing tests are from Purchase Order, failing on all PRs.

If this logic get's approved I'll write some tests for the 3 above scenarios.

deepeshgarg007 (approved):

PR #33990, opened 2023-02-08
Fix for not getting employee expense approver & Adding Contact Channels Doctype

nipunanr:
1. Fix for not loading employee's default expense approver when selecting employee
2. Adding Doctype Contact Channels for creation of Contact Channels (Eg: Phone, Email, IM, Portal, etc)


deepeshgarg007:
@nipunanr Since you are adding new doctypes this looks more like a feature rather than a fix and this should go in the latest separated-out app

HRMS has been moved out as a separate app from version-14 onwards
The repo link is https://github.com/frappe/hrms

PR #34002, opened 2023-02-10
fix: should never get cutomer price on purchase document

FHenry:
In some case in Purchase Order the field Customer is set with the customer coming from 
`Material Request `
or 
`erpnext/erpnext/stock/doctype/material_request/material_request.make_purchase_order_based_on_supplier`

and in this case the get_item_detail and _in fine_ get_item_price 

```
	if args.get("customer"):
			conditions += " and customer=%(customer)s"
        elif args.get("supplier"):
	        conditions += " and supplier=%(supplier)s"
```

Then customer win... on Purchase document this is a non sense


erpnext/stock/get_item_details.py
```diff
@@ -88,8 +88,20 @@ def get_item_details(args, doc=None, for_validate=False, overwrite_warehouse=Tru
...
 	update_party_blanket_order(args, out)
 
+	current_customer = args.customer
+	# Never try to find a supplier price if customer is set
+	# these Doctype
+	if (
+		args.get("doctype") == "Purchase Order"
+		or args.get("doctype") == "Purchase Receipt"
+		or args.get("doctype") == "Purchase Invoice"
+	):
```
s-aga-r: Use `in` keyword to check the doctype:

`if (args.get("doctype") in ["Purchase Order", "Purchase Receipt", "Purchase Invoice"]):`

s-aga-r (changes requested):

deepeshgarg007:
@FHenry Can you please add a test case for this scenario as well so that such an issue is not repeated again

FHenry:
> @FHenry Can you please add a test case for this scenario as well so that such an issue is not repeated again

@deepeshgarg007 Ok, I'm on it

FHenry:
> @FHenry Can you please add a test case for this scenario as well so that such an issue is not repeated again

@deepeshgarg007 ok test is done

FHenry:

FHenry:

FHenry:
@deepeshgarg007, @s-aga-r 
 Is it possible to add backport label it to  version-13-hotfix and version-14-hotfix or the process is make other PRs ?

EDIT: Sorry I didn't see the you've already put version-14-hotfix, the question remain for version-13-hotfix

PR #34021, opened 2023-02-12
fix: rare instances of IntegrityError while cancelling journals against cr note

ruthra-kumar:
On rare instances, outstanding amount calculation returns None, which causes IntegrityError.
image


ruthra-kumar:

PR #34040, opened 2023-02-13
Update fields after item row removal

sameer55chauhan:
Case: when one enters multiple items and then if one item is removed from the child table then all the doctype fields like Quantity and Amount should be updated automatically, instead of the validation at the before save. 

1. Items selected
Screenshot from 2023-02-13 17-26-05

2. One item row being deleted
Screenshot from 2023-02-13 17-26-15

4. Fields auto updation after row removal
Screenshot from 2023-02-13 17-26-20


deepeshgarg007:
@sameer55chauhan This already happens, don't think this is needed. Might be some other issue at your end. Please check the browser console for any errors. Closing this PR

PR #34046, opened 2023-02-13
fix: BOM import failed as use same label field for Raw Materials Item table and Scrap Item table

FHenry:
fix: #33994 

Note to reviewer : all is in the Issue

PR #34053, opened 2023-02-14
fix: rename duplicate field name with same type into a DocType to avoid import Error

FHenry:
Following #34046 

The fields rename extracted from 

> SELECT df.parent,df.label,df.fieldname,df.fieldtype,count(df.name) FROM tabDocField as df
> INNER JOIN tabDocType as dt on dt.name = df.parent
> WHERE df.parenttype='DocType' AND df.fieldtype NOT IN ('Section Break','Column Break')
> AND dt.allow_import = 1
> GROUP BY df.parent,df.label,df.fieldtype
> HAVING count(df.name)>1

### DIFF
--- a/erpnext/accounts/doctype/pricing_rule/pricing_rule.json
+++ b/erpnext/accounts/doctype/pricing_rule/pricing_rule.json
@@ -472,7 +472,7 @@
    "description": "If rate is zero them item will be treated as \"Free Item\"",
    "fieldname": "free_item_rate",
    "fieldtype": "Currency",
-   "label": "Rate"
+   "label": "Rate of Free Item"
   },
   {
    "collapsible": 1,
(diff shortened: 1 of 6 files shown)


erpnext/accounts/doctype/pricing_rule/pricing_rule.json
```diff
@@ -472,7 +472,7 @@
    "description": "If rate is zero them item will be treated as \"Free Item\"",
    "fieldname": "free_item_rate",
    "fieldtype": "Currency",
-   "label": "Rate"
+   "label": "Rate of Free Item"
```
s-aga-r:
```suggestion
   "label": "Free Item Rate"
```

erpnext/stock/doctype/item/item.json
```diff
@@ -706,7 +706,7 @@
    "depends_on": "enable_deferred_expense",
    "fieldname": "no_of_months_exp",
    "fieldtype": "Int",
-   "label": "No of Months"
+   "label": "No of Months (expense)"
```
s-aga-r:
```suggestion
   "label": "No of Months (Expense)"
```

s-aga-r (changes requested):

s-aga-r (approved):

PR #34058, opened 2023-02-14
fix: manual depr entries in asset_depreciations_and_balances report and some misc bugs [v14]

anandbaburajan:
- Manual depreciation entries weren't reflecting in the `asset_depreciations_and_balances` report, so fixed that.
- The `opening_accumulated_depreciation` was shown incorrectly in the asset chart, so fixed that.
- Asset values were sometimes shown with weird precision in the chart and `fixed_asset_register` report, so fixed that.

Note: backport to v13


anandbaburajan:
@mergify backport version-14

PR #34060, opened 2023-02-14
fix: multiple Point of Sale conversion issue resolved

vishdha:
1. stock projected qty report shows qty as Reserved for POS it should be stock qty
2. POS not  allowed to deduct 1 unt item having KGS as STOCK UOM - Resolved by UOM Conversion

Default UOM 1 = KG
UNT 1 = .09Kg or 90g

.81 kg = 810g = 9 UNT

before Change: System not allowed to sales 8 UNT = 720g = .72kg, Ideally it should allowed sales .72kg when we have .81kg

after Change: System allowed to sales 8 UNT = 720g = .72kg , Remaining qty after sales 0.09kg = 90g = 1 unt

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -161,7 +161,7 @@ class POSInvoice(SalesInvoice):
 
 		bold_item_name = frappe.bold(item.item_name)
 		bold_extra_batch_qty_needed = frappe.bold(
-			abs(available_batch_qty - reserved_batch_qty - item.qty)
+			abs(available_batch_qty - reserved_batch_qty - item.qty * item.conversion_factor)
 		)
 		bold_invalid_batch_no = frappe.bold(item.batch_no)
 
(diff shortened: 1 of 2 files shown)


vishdha:
Backport to version 14

vishdha:
@s-aga-r @rohitwaghchaure @deepeshgarg007 Can someone look into it

s-aga-r (approved):

s-aga-r (changes requested):
@vishdha please add a test-case for the same.

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -161,7 +161,7 @@ def validate_pos_reserved_batch_qty(self, item):
 
 		bold_item_name = frappe.bold(item.item_name)
 		bold_extra_batch_qty_needed = frappe.bold(
-			abs(available_batch_qty - reserved_batch_qty - item.qty)
+			abs(available_batch_qty - reserved_batch_qty - item.qty * item.conversion_factor)
```
deepeshgarg007: Why not just use item.stock_qty instead?

vishdha: yeah that is also correct

erpnext/selling/page/point_of_sale/pos_controller.js
```diff
@@ -522,7 +522,7 @@ erpnext.PointOfSale.Controller = class {
 
 			const from_selector = field === 'qty' && value === "+1";
 			if (from_selector)
-				value = flt(item_row.qty) + flt(value);
+				value = flt(item_row.qty * item_row.conversion_factor) + flt(value);
```
s-aga-r:
```suggestion
				value = flt(item_row.stock_qty) + flt(value);
```

s-aga-r (commented):
@vishdha why using `conversion_factor` here?

s-aga-r (approved):

PR #34167, opened 2023-02-21
chore: add german translations

PatrickDEissler:


erpnext/translations/de.csv
```diff
@@ -9916,3 +9916,5 @@ Cost and Freight,Kosten und Fracht,
 Delivered at Place,Geliefert benannter Ort,
 Delivered at Place Unloaded,Geliefert benannter Ort entladen,
 Delivered Duty Paid,Geliefert verzollt,
+Discount Validity,Gültigkeit des Nachlasses,
+Discount Validity Based On,Gültigkeit des Nachlasses basierend auf,
```
barredterra:
```suggestion
Discount Validity,Frist für den Rabatt,
Discount Validity Based On,Frist für den Rabatt berechnet sich nach,
```

In the other fields, „Discount“ already gets translated as „Rabatt“. Let's stick to this for consistency.

PatrickDEissler (commented):
Thank you Raffael for the two corrections. They are accepted and included in this patch.

barredterra (approved):

PR #34170, opened 2023-02-22
fix: fields mapping from Material Request to Transfer Material

mohsinalimat:
fields mapping from Material Request to Transfer Material


s-aga-r:
@mohsinalimat please update the description and use semantic commits.

s-aga-r:
@mohsinalimat I will raise a fix for this, the current fix also sets the `Target Warehouse` for Material Transfer (In Transit)

PR #34176, opened 2023-02-23
fix: ui freeze upon item selection in sales invoice

ruthra-kumar:
**Issue:** Desk UI Freezes when selecting an item in Sales Invoice. 
Possible Fix: Call to `get_batch_no` doesn't return any valid data. Yet, the else block sets 'batch_no' to undefined.


ruthra-kumar:

PR #34189, opened 2023-02-23
fix: user shouldn't able to make item price for item template

rohitwaghchaure:
In Item Price, the Item Templates are also selectable. As Templates are NOT transaction items, they should not be selectable in the Item Price doctype.

### DIFF
--- a/erpnext/stock/doctype/item_price/item_price.js
+++ b/erpnext/stock/doctype/item_price/item_price.js
@@ -2,7 +2,18 @@
 // License: GNU General Public License v3. See license.txt
 
 frappe.ui.form.on("Item Price", {
-	onload: function (frm) {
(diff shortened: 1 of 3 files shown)


aashaybinaykia:
This feature however eliminates the need to add Item Prices for all variants if the item price is the same for all of them.  @rohitwaghchaure

TurkerTunali:
Can we do this by adding filters in the transaction? For instance we also can't select disabled items and non-stock items. Also we can't select sales items in the buying transactions. So, filters and controllers makes this possible.

Because we don't have a tool to set the item variant prices! Customers can't update 100s of variants each month.

The price logic checks for variant price first. If it doesn't exist it gets the price from the Template. So users can set and update item price for all variants easily via Template Prices.

PR #34192, opened 2023-02-23
fix: remove non-maintained Tax Detail report and missing workspace links

casesolved-co-uk:
Removing contributed Tax Detail report because:

- It is out of date and has bugs, for example it does not support transaction-based taxes and charges templates, only Item Tax Templates
- I will maintain it as part of our free UK module on frappe cloud
- Undocumented feature because the docs never made it in: https://github.com/frappe/erpnext_documentation/pull/322

Also, whilst removing from the workspace, some Saudi Arabia (KSA) components have been removed

version-14-hotfix


barredterra (changes requested):
Please also remove the report from here:

https://github.com/frappe/erpnext/blob/c3d455653e717b539a6fc567fcd8ddbdbf5c717f/erpnext/accounts/test/test_reports.py#L33-L36

Not sure about backporting to v14... might there be people using this report in production?

casesolved-co-uk:
> Please also remove the report from here:

Done

> Not sure about backporting to v14... might there be people using this report in production?

If they found it and managed to figure out how to use it. I doubt it. My module has a copy called `Tax Ledger` that my customers use

barredterra (approved):

PR #34207, opened 2023-02-24
fix: permission error while calling get_work_order_items

ruthra-kumar:
'Insufficient Permission for Sales Order' while trying to create Work Order from Sales Order's drop down option with only `Read` access for Sales Orders.

scr1

## Cause:
`get_work_order_items` is a whitelisted document method.

https://github.com/frappe/erpnext/blob/e26c6dc76b16707c31e9a0277c2710a4c5cb43f0/erpnext/selling/doctype/sales_order/sales_order.py#L417-L460

With recent change in FW's permission handling https://github.com/frappe/frappe/pull/18666, calling document method is considered as 'Update after Submit' action and user must have 'Submit' access to proceed. Else, will get above error.

## Fix:
Convert `get_work_order_items` to Standalone whitelisted method.

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.js
+++ b/erpnext/selling/doctype/sales_order/sales_order.js
@@ -309,9 +309,12 @@ erpnext.selling.SalesOrderController = class SalesOrderController extends erpnex
 
 	make_work_order() {
 		var me = this;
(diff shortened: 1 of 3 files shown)


ruthra-kumar:

PR #34212, opened 2023-02-25
fix: Remove missing DocField in fetch_from

brian-pond:
Removing value "prevdoc_detail_docname.sales_person" from the `fetch_from` metadata.  This DocField no longer exists in Maintenance Visit Purpose.


s-aga-r:
The field `prevdoc_detail_docname` was removed in https://github.com/frappe/erpnext/pull/25358

s-aga-r (approved):

PR #34225, opened 2023-02-27
fix: set `from_warehouse` and `to_warehouse` while mapping SE

s-aga-r:
Steps to replicate:

1. Create Material Transfer (Purpose = Material Transfer).
2. Create > Material Transfer OR Material Transfer (In Transit).

Problem: The fields `from_warehouse` and `to_warehouse` didn't get copied from Material Request to Stock Entry.

ref: https://github.com/frappe/erpnext/pull/34170


mohsinalimat:
LGTM

mohsinalimat:
LGTM

PR #34235, opened 2023-02-28
feat: set landed cost based on purchase invoice rate

rohitwaghchaure:
If the purchase receipt and purchase invoice rate is not same and user wants to adjust the valuation rate of the item as per purchase invoice cost then currently they have to make Landed cost voucher but with this feature system will adjust the valuation rate as per purchase invoice rate. To enable this feature you need to  enabled checkbox "Set Landed Cost Based on Purchase Invoice Rate" (see below image) in the buying settings. 
Screenshot 2023-02-28 at 3 45 20 PM

**Use Case**

- Created Purchase Receipt for item A with rate as 100 
- System has booked Stock In Hand with 100 rate
- After 2 days, user has created purchase invoice against the above purchase receipt
- After 2 days because of change in exchange rate the rate in the invoice changed to 150
- Now purchase receipt has rate 100 where as Purchase invoice has rate 150
- If you want to adjust the Stock In Hand with Purchase invoice Rate (150) then this feature will be helpful

Documentation: https://docs.erpnext.com/buying-settings#25-set-landed-cost-based-on-purchase-invoice-rate

### DIFF
--- a/erpnext/buying/doctype/buying_settings/buying_settings.json
+++ b/erpnext/buying/doctype/buying_settings/buying_settings.json
@@ -18,7 +18,7 @@
   "pr_required",
   "column_break_12",
   "maintain_same_rate",
-  "adjust_incoming_rate_based_on_purchase_invoice_rate",
+  "set_landed_cost_based_on_purchase_invoice_rate",
   "allow_multiple_items",
   "bill_for_rejected_quantity_in_purchase_invoice",
   "disable_last_purchase_rate",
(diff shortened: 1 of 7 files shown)


erpnext/stock/doctype/purchase_receipt_item/purchase_receipt_item.json
```diff
@@ -1007,12 +1008,20 @@
    "fieldtype": "Check",
    "label": "Has Item Scanned",
    "read_only": 1
+  },
+  {
+   "fieldname": "adjust_incoming_rate",
+   "fieldtype": "Currency",
+   "label": "Adjust Incoming Rate (Purchase Invoice)",
```
nabinhait: Rate Difference with Purchase Invoice

erpnext/buying/doctype/buying_settings/buying_settings.json
```diff
@@ -147,14 +148,22 @@
...
    "fieldtype": "Check",
    "label": "Show Pay Button in Purchase Order Portal"
+  },
+  {
+   "default": "0",
+   "depends_on": "eval: !doc.maintain_same_rate",
+   "description": "Users can enable the checkbox If they want to adjust the incoming rate (set using purchase receipt) based on the purchase invoice rate.",
+   "fieldname": "adjust_incoming_rate_based_on_purchase_invoice_rate",
+   "fieldtype": "Check",
+   "label": "Adjust Incoming Rate Based on Purchase Invoice Rate"
```
nabinhait: Update Landed Cost Based on Purchase Invoice Rate

nabinhait (changes requested):

PR #34241, opened 2023-02-28
fix: pos return throwing amount greater than grand total

### LINKED ISSUE #34134: POS return is not possible, totals get wrong calculated, message: Paid amount + Write Off Amount can not be greater than Grand Total
### Information about bug

In the Point of sale app "app/point-of-sale", click on Toggle Recent Orders, filter the paid orders and make a return.
The article qty is negative which is ok, but the prices and total is positive which should also be negative because its a return.
If we click on the blue Checkout button, we get the dialog window with the message "Paid amount + Write Off Amount can not be greater than Grand Total".

image

The validation occurs in the file sales_invoice.py:
image

I do not know what is the cause of the problem, hopefully someone can help or give a hint.

ErpNextReturnsNotPossible

### Module

selling

### Version

Frappe Framework: v15.x.x-develop () (develop)
Payments: v0.0.1 (develop)

### Installation method

docker

_No response_

ruthra-kumar:
Credit Notes against POS Invoices throwing 'Paid Amount + Write off amount cannot be greater than Grand Total'.

If `item.qty` is `0`, just do rate*-1 as the else block will only process Credit/Debit notes
regression: https://github.com/frappe/erpnext/pull/33902

### DIFF
--- a/erpnext/public/js/controllers/taxes_and_totals.js
+++ b/erpnext/public/js/controllers/taxes_and_totals.js
@@ -131,8 +131,8 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 					item.net_amount = item.amount = flt(item.rate * item.qty, precision("amount", item));
 				}
 				else {
-					let qty = item.qty || 1;
-					qty = me.frm.doc.is_return ? -1 * qty : qty;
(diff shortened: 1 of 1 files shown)


ruthra-kumar:

PR #34257, opened 2023-03-01
feat: Closing balance for period closing and reporting

deepeshgarg007:
Introducing a new doctype "Account Closing Balance" for calculating opening balances in Financial Statements

Adding this brings in some minor restrictions on the way users post accounting entries in ERPNext.
Since the calculation of the closing balances is tied to the period closing voucher, users won't be able to change any accounting entry or post opening-entries before the period closing voucher's posting date. Only once the period closing voucher is canceled previous entries can be updated.

This also has support for custom accounting dimensions so dimension-based filtering is also possible. If there are PCVs posted and no account closing balances have been created then by default GL entries are used for computation as per the current behavior

### Why is this required?

Right now for calculating the opening balances or even account balances the reports and functions go through all the previous GLs (could be of just 1 fiscal year or 10 years for some) even when practically there is not much change in the entries

These reports and functions usually take a lot of time or fail for users dealing with more than a million GL entries, so pre-computing balances for such time frames help with such performance improvements.

On a typical site with 1.5 million GL entries running the trial balance report without "Account Closing Balance" the report takes around 21s whereas with "Account Closing Balance" it comes down to 8s

Before:
Screenshot 2023-03-08 at 2 27 25 PM

After:
Screenshot 2023-03-08 at 2 27 40 PM

ToDo List
- [x] Update Trial Balance report
- [x] Patch to create closing balance records
- [x] Update financial statements (Balance Sheet)

`no-docs`


erpnext/accounts/general_ledger.py
```diff
@@ -566,6 +572,28 @@ def check_freezing_date(posting_date, adv_adj=False):
...
+	last_pcv_date = frappe.db.get_value(
+		"Period Closing Voucher", {"docstatus": 1, "company": company}, "max(posting_date)"
+	)
+
+	if last_pcv_date and getdate(posting_date) <= getdate(last_pcv_date):
+		message = _("Books have been closed till the period ending on {0}").format(
+			formatdate(last_pcv_date)
+		)
+		message += "</br >"
+		message += _("You cannot create any new accounting entries till this date.")
```
nabinhait:
```suggestion
		message += _("You cannot create/amend any accounting entries till this date.")
```

nabinhait (changes requested):

nabinhait (approved):

JitendraSAW:
@deepeshgarg007 , Can you help us as we can check this Use Case, Can you provide any List of Transactions or Flow we can check the same.

PR #34258, opened 2023-03-01
chore: Make finance book read only

deepeshgarg007:
The change must have gotten lost while backporting https://github.com/frappe/erpnext/pull/31439

Hence manually back porting


trysyss:
In my latest install, this is now read only and I cannot unset the flag. It causes the Finance book to be hidden in Journal Entry. Prior to ERPNext: v14.15.1 (version-14), the field showing is Ok.

PR #34259, opened 2023-03-01
fix!: Parse float as per number format in quality inspection

ankush:
fixes https://github.com/frappe/frappe/pull/20147 

Quality inspection stores reading as `Data` hence all the computation can get messed up if number format uses different symbols for comma and decimal.


ankush:
This should not be backported as many users are just accustomed to using different format for adding readings knowing this bug.

PR #34262, opened 2023-03-01
fix(General Ledger): Wrap unexpectedly long word 

surajshetty3416:
**Before:**
Screenshot 2023-03-01 at 1 15 33 PM

**After:**
Screenshot 2023-03-01 at 4 12 13 PM

A more generic fix has been done with [this](https://github.com/frappe/frappe/pull/20152) but since General Ledger is using custom format it needed a separate fix.


surajshetty3416:

PR #34267, opened 2023-03-01
fix: Do not calculate commission post submit

deepeshgarg007:
image

### DIFF
--- a/erpnext/selling/sales_common.js
+++ b/erpnext/selling/sales_common.js
@@ -253,7 +253,7 @@ erpnext.selling.SellingController = class SellingController extends erpnext.Tran
 	}
 
 	calculate_commission() {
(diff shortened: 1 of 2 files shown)


erpnext/selling/sales_common.js
```diff
@@ -253,7 +253,7 @@ erpnext.selling.SellingController = class SellingController extends erpnext.Tran
 	}
 
 	calculate_commission() {
-		if(!this.frm.fields_dict.commission_rate) return;
+		if(!this.frm.fields_dict.commission_rate || this.frm.doc.docstatus != 0) return;
```
barredterra:
```suggestion
		if(!this.frm.fields_dict.commission_rate || this.frm.doc.docstatus === 1) return;
```

Shouldn't this match the server side condition? Alternatively, they could both be `> 0`.

PR #34273, opened 2023-03-02
fix: `rejected_serial_no` not getting copied from PR to PR(Return)

s-aga-r:
Source / Ref: ISS-22-23-05650

Issue(s):

1. The `Rejected Serial No` is not getting copied from the Purchase Receipt Rejected Serial No field.
2. The `Serial No` field is required even if the `Accepted Quantity` is 0 (Return Purchase Receipt).


s-aga-r:
Unrelated failing test.

PR #34284, opened 2023-03-02
fix: Default sales team not getting set

deepeshgarg007:
Default sales team not getting set when converting quotation to sales order


Mutantpenguin:
Oh well, with this change we can't add a SalesTeam to our SalesOrder anymore programatically because this code here always overwrites it.

Everything via code:
1. Create a SalesOrder
2. Save it -> this new code runs
3. Create a new SalesTeam and append it to the SalesOrder
4. Save the Sales Order -> this new code runs because the field `sales_team` is nowhere to be found in `self.meta` and overwrites the SalesTeam we set ourselves.

I can provide example code if necessary.

PR #34287, opened 2023-03-03
fix: Allocate tax loss to tax account head on early payment discount

marination:
Docs update: 
https://docs.erpnext.com/docs/v14/user/manual/en/accounts/payment-terms/edit-wiki?wiki_page_patch=9f963e285a
https://docs.erpnext.com/docs/v14/user/manual/en/accounts/accounts-settings/edit-wiki?wiki_page_patch=7b3e920d00

## Early Payment Discount: Loss Allocation
Early Payment Discount Deductions in payment entry can be split into income loss and tax loss by enabling **Book Tax Loss on Early Payment Discount** in the Accounts Settings:
  Tax_Settings_Revised

**Consider the foll.g Sales Invoice:**
Screenshot 2023-03-03 at 1 39 58 PM

**Payment Entry Before:**
Screenshot 2023-03-03 at 1 34 54 PM

**Payment Entry After:**
deductions

## Other Fixes

1. **Deductions amount value must always be in terms of base currency/company currency**: Calculate and convert the deductions amount into base currency. They were being set in the Supplier/Customer currency which is wrong since these deductions are on the Company account heads
2. **Back update `discounted amount` in Invoice's Payment Schedule rows based on Discount Type**: Discounted Amount was always calculated in percentages on submitting a Payment Entry which gave the wrong value if it was an amount based discount.
3. **Set label and value formatting currency of `Amount` in deductions as base currency/company currency**: For multi-currency SI and PE, the deductions table's amount field label showed the company currency but the value was formatted in the Customer/Supplier currency
4. **If `Allocate Payment Based on Payment Terms` is checked, then handle payment schedule amount extraction and back updation in the right currency**: 
     - Payment Schedule Amounts are always in transaction currency. 
     - While fetching amount from Payment Schedule in the PE, if accounting is in base currency, convert the value into base currency.
     - Similarly while back **Paid Amount** into Payment Schedule, convert back into trans.n currency IF reference doc is multi-currency and PE is not (PE is in base currency)
5. **Provision to apply discount on recording payment late**: Party could have paid on time but payment is recorded late. Prompt for reference date so that discount is applied while mapping. **This is only prompted if early payment discount is applied on current document.**
    

<hr>

**TODO:**

- [x] Check with multi currency
- [x] Tests for discount %, discount amount and multi currency
- [x] Docs


erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1911,13 +1930,99 @@ def apply_early_payment_discount(paid_amount, received_amount, doc):
...
 		if total_discount:
 			money = frappe.utils.fmt_money(total_discount, currency=doc.get("currency"))
 			frappe.msgprint(_("Discount of {} applied as per Payment Term").format(money), alert=1)
 
-	return paid_amount, received_amount, total_discount
+	return paid_amount, received_amount, total_discount, valid_discounts
+
+
+def set_early_payment_discount_loss(pe, doc, valid_discounts, discount_amount):
+	"""Split early bird discount deductions into Income Loss & Tax Loss."""
```
barredterra:
```suggestion
	"""Split early payment discount into Income Loss & Tax Loss."""
```

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1911,13 +1930,99 @@ def apply_early_payment_discount(paid_amount, received_amount, doc):
...
+	for account, loss in tax_discount_loss.items():
+		total_tax_loss += loss
+		pe.append(
+			"deductions",
+			{
+				"account": account,
+				"cost_center": pe.cost_center or frappe.get_cached_value("Company", pe.company, "cost_center"),
+				"amount": flt(loss * doc.get("conversion_rate", 1), precision),
+			},
+		)
```
barredterra:
```suggestion
		total_tax_loss += loss
		amount = flt(loss * doc.get("conversion_rate", 1), precision)
		if amount == 0.0:
			continue

		pe.append(
			"deductions",
			{
				"account": account,
				"cost_center": pe.cost_center or frappe.get_cached_value("Company", pe.company, "cost_center"),
				"amount": amount,
			},
		)
```

Don't add deduction if the `amount == 0`. This can happen when the Sales Taxes and Charges table contains rows that don't apply to any item on that invoice.

barredterra (commented):
I tried to create PE from a SI in company currency

- ... with multiple tax rows and early payment discount
- ... with a single tax row and no early payment discount

This worked very well 👍🏼 

What didn't work correctly was to create a SI in another currency, while the GL entries use base amounts / company currency. In this case, the paid amount (in company currency) was wrong.
<details>
<summary>SI</summary>

Bildschirmfoto 2023-03-06 um 22 56 27

</details>
<details>
<summary>PE</summary>

_Paid Amount (EUR)_ is wrong, deductions are correct

Bildschirmfoto 2023-03-06 um 22 57 02

Bildschirmfoto 2023-03-06 um 22 57 20

</details>

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1889,20 +1931,26 @@ def set_paid_amount_and_received_amount(
...
+def apply_early_payment_discount(paid_amount, received_amount, doc, party_account_currency):
 	total_discount = 0
+	valid_discounts = []
 	eligible_for_payments = ["Sales Order", "Sales Invoice", "Purchase Order", "Purchase Invoice"]
 	has_payment_schedule = hasattr(doc, "payment_schedule") and doc.payment_schedule
+	is_multi_currency = party_account_currency != doc.company_currency
 
 	if doc.doctype in eligible_for_payments and has_payment_schedule:
 		for term in doc.payment_schedule:
 			if not term.discounted_amount and term.discount and getdate(nowdate()) <= term.discount_date:
```
barredterra:
```suggestion
			if not term.discounted_amount and term.discount and doc.reference_date <= term.discount_date:
```

The relevant date for determining if the discount applies is the date the transaction was received. If this would make things easier, we could argue that the posting date would work as well – I guess they can be identical for most practical purposes.

Imagine the customer paid yesterday, but I'm recording the payment today. The discount should still apply, if yesterday was the last day of validity.

Another possibility would be to ask the user whether the early payment discount should be applied. This would help in cases where the company accepts discounted payments even after the validity date, to avoid trouble.

barredterra (commented):
Issues from the last review have been resolved. 👍🏼 

Something that should go into the docs is that this feature (currently) only works when mapping a payment from an invoice, not when creating a payment and then fetching outstanding invoices.

One small annoyance I encountered was that with odd amounts I can get an additional deduction of `0.01`. For example, when two deductions have been rounded down but the difference adds up to > `0,005`. This deduction of `0.01` has no account set. We could either set the company's _Default Round Off Account_ here or change the calculation method to avoid rounding differences. This can be recreated with the example SI and PE from the last review.

erpnext/accounts/doctype/payment_entry/payment_entry.py
marination: I think reference date would be safest since it's purpose is to specify the actual cheque date or receipt date but it is not set until after the mapping occurs :(

marination:
@deepeshgarg007 Mind taking a look 👀?

deepeshgarg007 (commented):
The tax loss booking should be optional IMO. In most case this is [Non Trade (Cash Discount)](https://byjus.com/commerce/difference-between-trade-discount-and-cash-discount/) and don't affect tax, at least this how it is in India

Also the docs link says v13, should be v14

barredterra:
A supplier can reduce their GST liability when granting a cash discount, provided certain conditions are met. According to Section 15(3) of the CGST Act, 2017, the value of the supply shall not include any discount if:

1. The discount is given before or at the time of supply and has been duly recorded in the invoice issued in respect of such supply.
2. The discount is given after the supply has been effected, if:
    a) The discount is established in terms of an agreement entered into at or before the time of such supply and specifically linked to relevant invoices.
    b) The input tax credit attributable to the discount, on the basis of the document issued by the supplier, has been reversed by the recipient of the supply.

By meeting these conditions, the supplier can effectively reduce the taxable value of the supply, resulting in a lower GST liability. However, it is essential to ensure that all applicable conditions are met and necessary documentation is maintained to support the reduction in GST liability.

Summarized from https://carajput.com/learn/whether-liable-to-gst-on-discount-received-for-early-payments.html

---

Under the GST regime, Section 15 of the CGST Act, 2017 governs discounts. Discounts given before or at the time of supply can be reduced from the taxable value if mentioned on the invoice. Discounts given after the supply can also be reduced from the taxable value if the discount is agreed upon before the supply, linked to a specific supply invoice, and the Input Tax Credit (ITC) attributable to the discount is reversed by the buyer. GST liability of the supplier can be reduced if both the supplier and receiver are aware of the discount before supply. There is no differentiation between trade and cash discounts in GST; instead, they are segregated into two categories: those given before or at the time of supply, and those given after the time of supply.

Summarized from https://taxguru.in/goods-and-service-tax/tax-treatment-tradecash-discounts-gst-regime.html

deepeshgarg007:
@barredterra Didn't exactly get your last response, are you suggesting the tax loss booking should not be optional?

barredterra:
I couldn't access the information you pointed to (blocked in Europe), but the articles I found seemed to support our use case. ("GST liability of the supplier can be reduced if both the supplier and receiver are aware of the discount before supply. There is no differentiation between trade and cash discounts in GST")

Anyway, making it configurable shouldn't be an obstacle. First rule of ERPs: everything must be configurable. 😅

marination:
@deepeshgarg007 It is configurable now in the Accounts Settings (see description). Do let me know if any other major changes are needed

deepeshgarg007 (approved):

PR #34311, opened 2023-03-06
fix: cancel exchange gain/loss postings upon payment cancellation

ruthra-kumar:
Cancel Exchange Gain/Loss GL entries for an Invoices if the Advance Payment Entry with a difference exchange rate is cancelled.


ruthra-kumar:
Will add test cases for this.

ruthra-kumar:
V14 needs a refactor on how Exchange GL entries are posted. A separate Journal has to be posted to make it easy to track and cancel when needed.

Abandoning this approach.

PR #34313, opened 2023-03-06
perf: Stock Entry (Material Transfer)

s-aga-r:
**Source / Ref:** ISS-22-23-05690

**Changes:** Moved the queries from for loop to a separate function and get the data at once from DB instead of getting it in every iteration of for loop.

Reduces the Execution Time from **144 --> 64s**.


s-aga-r:
Unrelated failing test

PR #34301, opened 2023-03-04
refactor: use renamed timezone utils

barredterra:
Depends on https://github.com/frappe/frappe/pull/20253

PR #34340, opened 2023-03-07
refactor: use renamed timezone utils

barredterra:
Wait for release of https://github.com/frappe/erpnext/pull/34301 before merging


ankush:
@barredterra lets wait for a while before merging this. Unnecessary dependency on 14.x.x

barredterra:
Sure, but who will remember if there's no open PR?

Issue #34027, opened 2023-02-13
Quality Goal


Ponyooooo:
### Information about bug

error on the objective field of the Quality Goal?

![image](https://user-images.githubusercontent.com/60758795/218362043-1c03e73a-e69c-424a-a71c-012c2e00b940.png)

The objective field cannot be edit once the goal field has an entry and once it is saved

![image](https://user-images.githubusercontent.com/60758795/218362079-f3acb68c-26f8-4e3a-a939-d151a34c8db1.png)

### Module

quality

### Version

ERPNext: v13.45.0 (HEAD)
Frappe Framework: v13.49.0 (HEAD)

### Installation method

FrappeCloud

dj12djdjs:
### Steps to reproduce.

Can't type anything in Objective field.

mihir-kandoi:
This is no longer replicable. Closing.

PR #34344, opened 2023-03-08
fix: objective field set readonly on the Quality Goal form

iramluism:
When you define a goal on the **Quality Goal** form, the objective field become read only. 
It happens due the objective field handle the goal field as link type, and try to fetch the objective value from it.  

In this PR, the `fetch_from` property is set empty to avoid fetch this value from goal field.

### DIFF
--- a/erpnext/quality_management/doctype/quality_goal_objective/quality_goal_objective.json
+++ b/erpnext/quality_management/doctype/quality_goal_objective/quality_goal_objective.json
@@ -1,4 +1,5 @@
 {
+ "actions": [],
  "autoname": "format:{####}",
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@iramluism I didn't actually get the fix, if you don't want the value to be overridden then you can maybe enable the fetch_if_empty check and customize it as per your behavior. Closing this PR as this doesn't seem necessary

iramluism:
Ok, but this field is fetching from a data type field. It raise an error in this case.

QwQuan:
Quality Goal #34027
I reproduced this bug

https://github.com/frappe/erpnext/issues/34027#issue-1581572449

PR #34371, opened 2023-03-09
feat: Making Quotation Templates possible

intellimomentum:
In the Quotation form, the button "Get Items From", let you only import items from opportunities.
This feature makes it possible to import pre-defined items from so called "Quotation Templates":

image

image

image


intellimomentum:
Any questions about the feature?
It works very well in our company and increases the quotation creation speed massively!

dj12djdjs:
I think this is pretty useful.

Just wondering, could the same thing be dome with the existing Product Bundle functionality?

intellimomentum:
Hello,

In many aspects, the functionality is similar.
However, there are a few significant differences:
- Product Bundles behave like a fixed configured product, it is not
possible to change the number of a product in the product bundle.
- You don't always want to group the individual products. Offers for PV
installations are a good example of this. Basically all customers need the
same products for which a few simple quote templates can be created: Solar
panels, cables, inverters, electrical installation ... However, there are
always differences in the number of panels (to give just one example).
- Product Bundles are a single product from the viewpoint of a customer.
(One price, just included items but not individually listed with a price
and position in a quotation)

Hope this helps

Yours sincerely
Bernhard Hein


Am Sa., 11. März 2023 um 06:35 Uhr schrieb Devin Slauenwhite <
***@***.***>:

erpnext/selling/doctype/quotation/quotation.js
```diff
@@ -349,3 +349,81 @@ frappe.ui.form.on("Quotation Item", "stock_balance", function(frm, cdt, cdn) {
...
+					  var row=frm.add_child("items"); // add row
+					  frm.refresh_fields("items"); // Refresh table
+
+					  // Copy-Paste Operation
+					  let idx = items_idx+1;
+					  let fields = ['item_code', 'qty'];
+					  for(var m = 0; m < fields.length; m++){
+						  frm.get_field("items").grid.grid_rows[idx].doc[fields[m]] = item[fields[m]];
+						  frm.get_field("items").grid.grid_rows[idx].refresh_field(fields[m]);
+					  }
```
barredterra:
```suggestion
					for (field of ['item_code', 'qty']) {
						let row = frm.get_field("items").grid.grid_rows[items_idx + 1];
						row.doc[field] = item[field];
						row.refresh_field(field);
					}
```

barredterra (changes requested):
Please format your code with prettier (use tabs) and remove useless comments like "Again, the Doctype we want to pick up". This will likely increase the chances of someone reviewing this. For details, please check the [contribution guidelines](https://github.com/frappe/erpnext/wiki/Contribution-Guidelines).

And maybe we can get a comment from @deepeshgarg007, if he's open for a feature like this.

PR #34387, opened 2023-03-10
refactor(patch): remove inner join to improve SQL performance

ruthra-kumar:
Changes in PR:
1. Improving patch performance by removing inner join
2. Remarks field is also moved
3. Remove 'migrate_remarks_from_gl_to_payment_ledger' from patches.txt as the field is moved in this 'migrate_gl_to_payment_ledger'


ruthra-kumar:

PR #34408, opened 2023-03-12
chore: delete remarks migration patch

ruthra-kumar:
Versions higher than V14.18.2, 'remarks' will be moved in 'migrate_gl_to_payment_ledger'


ruthra-kumar:

PR #34409, opened 2023-03-12
fix: cache session filters on Bank Reconciliation tool

ruthra-kumar:
Allow concurrent use of Bank Reconciliation Tool. Filters are cached using session id (sid).


barredterra (commented):
Works as expected 👍🏼

Could we use `frappe.session.user` as the key, so that filters will survive a logout? Having to reconfigure filters after each logout seems unnecessary.

ruthra-kumar:
@barredterra This was actually a hacky way to solve the problem. Will not be using this solution.

The main problem here is the document `save` event was used as as trigger to pull bank transactions and refresh the tool.
https://github.com/frappe/erpnext/blob/ca10e2bb9ff61bc1c9ffcd1e7a2a14d6ec4f4d51/erpnext/accounts/doctype/bank_reconciliation_tool/bank_reconciliation_tool.js#L77C2-L79

Since Bank Reconciliation Tool is a single doctype, if 2 users are using the tool concurrently, the filters applied(save trigger) by one user will reflect for the other user. Ideally, the `save` trigger should not have been used.

A better approach would be a separate button to pull transactions, similar to how Payment Reconciliation works. Payment Reconciliation already allows for concurrent usage without caching.

Better approach: https://github.com/frappe/erpnext/pull/34501

PR #34419, opened 2023-03-13
test: Update values in Sales Invoice tests

deepeshgarg007:
Sales Invoice test fixes after addition of https://github.com/frappe/frappe/pull/20298


ankush:
Ay thanks for fixing 🙏

PR #34428, opened 2023-03-13
Fix: Ensure revaluation calculation consistency

dgehri:
PR ensure that currency revaluation calculations are performed consistently between the exchange rate revaluation and the generated journal entries.


deepeshgarg007:
@dgehri Can you please explain more in detail what was the issue what does this PR fixes? Unable to get the fix. Please reopen after updating the description

PR #34432, opened 2023-03-13
Adding Subscription Link to Invoices

mayanknahata007:
Before
Screenshot 2023-03-13 at 9 45 06 PM
In the subscription section of the invoices, there is no link to the subscription it was created from.

After:
Screenshot 2023-03-13 at 9 44 35 PM

Added the field which automatically gets populated along with the other data.

Now the invoices could also be filtered via subscription value.
Screenshot 2023-03-13 at 9 44 16 PM


barredterra (commented):
One problem with this is that we'll get circular linking: **Subscription** links to **Sales Invoice** and **Sales Invoice** links back to **Subscription**. Then you can delete neither one.

I think the link from **Sales Invoice** to **Subscription** is the correct way, in theory, but we'll also need to remove the link from **Subscription** to **Sales Invoice** and refactor the code to handle this change.

barredterra:
> I think the link from Sales Invoice to Subscription is the correct way, in theory, but we'll also need to remove the link from Subscription to Sales Invoice and refactor the code to handle this change.

Are you planning to implement this? We've tried to "fix" subscriptions a couple of times before and failed, so from my experience it's not an easy task.

mayanknahata007:
> > I think the link from the Sales Invoice to the Subscription is the correct way, in theory, but we'll also need to remove the link from the Subscription to Sales Invoice and refactor the code to handle this change.
> 
> Are you planning to implement this? We've tried to "fix" subscriptions a couple of times before and failed, so from my experience, it's not an easy task.

Yes, I plan to. 
I wanted the subscription to be linked to the invoices to have consolidation at the invoice level as well, and maybe add additional fetching to the invoices via subscription doctype.
Since it is coded in the core, not sure if a custom app would work here.

The easiest approach would be not to make it "read-only" and maybe remove the link from draft invoices for deletion. I'm sure it wont be the right approach though :P

But yes, would be interested and working on improving it. Also maybe create reminders before the invoices are generated.

Let's collab XD

Mood93:
Hello, new to ERPNext/Frappe ecosystem. Would it be possible and preferred to create an intermediary DocType to handle a many-to-many relationship between subscription and invoice?

Is there a use case for multiple subscriptions to be on a single invoice? E.g. Customer is subscribed to multiple products but only want to send one invoice?

What about subscriptions having multiple invoices? E.g. With insurance in the US, you have a term length of 6mo or 12mo but you can pay monthly, quarterly, semiannually, and annually; each with different rates/discounts.

deepeshgarg007:
@mayanknahata007 Closing this PR for now, is not the right fix

federicocalvo:
@barredterra @mayanknahata007 subscriptions is one of the great forgotten ones, and I find it hard to understand why.

In our case we had to make a lot of corrections, which is difficult for us to provide because it is somewhat modified. It is a section where it is very difficult to test changes and simulate cases.

But it is the axis of invoicing for any service company, contemplated for example that Frappe is one of them, I think that by handling the subscriptions from another side with the theme of Press.
Otherwise, it is hard to understand why they would not improve a section that would benefit themselves.

The last attempt I remember is this one, which did not prosper:
https://github.com/frappe/erpnext/pull/30963

barredterra:
I ended up reimplementing subscriptions for one of our customers:
 https://github.com/alyf-de/simple_subscription

The existing subscription feature would need a complete reimplementation as well, IMO.

PR #34454, opened 2023-03-14
fix(accounts): allow Payment Request and Payment Entry for POS Invoice

casesolved-co-uk:
Adds Mode of Payment support to payment entry and payment request for POS Invoice

Use case:
The POS Invoice allows multiple payments by different payment methods both in the UI and database. This PR adds functionality to be able to take payments for each payment method separately using payment requests and payment entries, which in turn connect with the various payment gateways available.

version-14-hotfix


vishdha (commented):
Hi @casesolved-co-uk what is use case here. 
Generally POS not allowed zero invoice. Stilll We book unpaid Invoice with 1 minimum amount.
1. We need to consolidate POS Invoice in Closing POS Entry. 
2. Unpaid Sales Invoice get generated automatically then we can do payment entry.
3. If we do Payment entry from POS and by the time consolidated state. Did this change make Sales invoice paid?
4. What is effect on Accounting leadger

casesolved-co-uk:
@vishdha I have updated the PR description above. Hopefully this answers your questions which aren't very clear.

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -1622,6 +1627,24 @@ def get_reference_details(reference_doctype, reference_name, party_account_curre
 
 		if reference_doctype in ("Sales Invoice", "Purchase Invoice"):
 			outstanding_amount = ref_doc.get("outstanding_amount")
+		elif reference_doctype == "POS Invoice":
+			if not mode_of_payment:
```
deepeshgarg007, line 1665: `mode_of_payment` is an optional parameter, why throw an error if mode of payment not found?

erpnext/accounts/doctype/payment_request/payment_request.py
```diff
@@ -49,16 +49,17 @@ def validate_payment_request_amount(self):
 			get_existing_payment_request_amount(self.reference_doctype, self.reference_name)
 		)
 
-		ref_doc = frappe.get_doc(self.reference_doctype, self.reference_name)
-		if not hasattr(ref_doc, "order_type") or getattr(ref_doc, "order_type") != "Shopping Cart":
-			ref_amount = get_amount(ref_doc, self.payment_account)
+		if existing_payment_request_amount:
```
deepeshgarg007, line 52: So in the very first payment request if the Payment Amount is more that the order/invoice amount this check will be bypassed since we are checking for `existing_payment_request_amount`

erpnext/accounts/doctype/payment_entry/payment_entry.py
casesolved-co-uk: Because `mode_of_payment` is mandatory for POS Invoice

erpnext/accounts/doctype/payment_request/payment_request.py
casesolved-co-uk: Suspect this is a conflict merge issue because these changes were done a long time ago and the same code has been changed by other people. It's going to be a while until I can look at it again

PR #34456, opened 2023-03-15
fix: Gross Profit reports Invoices with -ve qty for Invoices with Cr Notes

ruthra-kumar:
## Issue
1. Create Invoice with 2 rows of same item.
2. Make a credit note for [1]
3. GP report for [1] will have -ve qty and incorrect selling amount due to the -ve quantity

Cr note is overallocated to the Parent Invoice while calculating Average rate. This only happens if multiple rows have the same item.

## Changes in this PR:
1. Overallocation issue mentioned above has been fixed.
2. Cr notes with 'return_against'(parent invoice) will be ignored by the report. Item quantities of those Cr notes are applied against the Original Invoice, there is no need to report them separately.
3. Standalone CR Notes will be reported as usual.

Invoice SINV-23-00056 with multiple rows of same item.
Screenshot 2023-03-15 at 11 18 26 AM

Cr Note, SINV-23-00058, against SINV-23-00056
Screenshot 2023-03-15 at 11 18 18 AM

Standalone Cr Note SINV-23-00057
Screenshot 2023-03-15 at 11 18 31 AM

## Before Fix:
1. Item quantities of SINV-23-00056 goes to -ve due to overallocation mentioned above. 
2. It's Cr Note SINV-23-00058 is also reported as a separate Invoice
without_fix

## After Fix:
1. SINV-23-00056 has correct qty(0) as the Cr note should negate the Original invoice.
2. Cr Note SINV-23-00058 is not reported. 

with_fix

### DIFF
--- a/erpnext/accounts/report/gross_profit/gross_profit.py
+++ b/erpnext/accounts/report/gross_profit/gross_profit.py
@@ -741,6 +741,8 @@ class GrossProfitGenerator(object):
 		if self.filters.to_date:
 			conditions += " and posting_date <= %(to_date)s"
 
+		conditions += " and (is_return = 0 or (is_return=1 and return_against is null))"
+
 		if self.filters.item_group:
 			conditions += " and {0}".format(get_item_group_condition(self.filters.item_group))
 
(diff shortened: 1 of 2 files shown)


ruthra-kumar:

ruthra-kumar:

PR #34466, opened 2023-03-16
fix: unwanted difference amt while reconciling vouchers from base currency account

ruthra-kumar:
## Issue
With `Allow multi-currency invoices against single party account` enabled, Sales/Purchase Invoices can be submitted in foreign currency to a company currency account. While reconciling such vouchers difference amount was mistakenly calculated.

Added validation based on Account currency to fix this.


ruthra-kumar:

PR #34468, opened 2023-03-16
ci: use version specific payments repo

phot0n:


ankush:
Change is fine :+1: , make sure there are patches and FC branches are updated accordingly so users can transition smoothly. 


You can look at what we did for HR for example.

ankush (approved):

PR #34508, opened 2023-03-20
feat: add verified chart of accounts for colombia in two variants

blaggacao:
Each PR can be cherry-picked, indicivually.

- `needs-tests`: it would probably dry up the scraped source if there was automated tests involving a scraping procedure.
- The content (including the description) is part of the government decret and thus is in the public domain
- The implementation of `account_info{_html}` seems a bit hacky: I'd rather have `account_info` simply put read-only for non-Account Manager, but unfortunately, a read-only _HTML Editor_ field is reneders as text, instead of HTML. That may be bug. If it can be / is fixed, I'm happy to rebase.

Bonus:
I'd have liked to make a tooltip on hover so that the description is made readily available when hovering over an account. Is that possible? Is that desired? I'd appreciate any guidance.


blaggacao:
This is ready for review.

blaggacao:
This is ready for review.

deepeshgarg007 (commented):
@blaggacao The COA added looks fine but a lot of things added is not explained well enough. Why two fields `account_info` and `account_info_html`? The COA only has `account_info` field.

Why is this erpnext/accounts/doctype/account/chart_of_accounts/verified/gen_co_chart_of_accounts.py
added is also not clear to me

Can you please clarify these doubts? Also add screenshots of how the COA tree will be rendered with the account info field, I think should not be added in the tree view.

blaggacao:
@deepeshgarg007 I decided to remove the orthogonal display-improvements from this PR and I documented, as requested, the scraping script in the file header.

This change makes this PR simply provide an up-to-date Colombian CoA in two variants.

Should someone desire to recreate the data, or should an update require to do so, the CoA can be re-scraped making use of the script.

Once this is merged, I can reopen a separate PR for the display improvements with screenshots, etc.

erpnext/accounts/doctype/account/chart_of_accounts/verified/scrape_co_chart_of_accounts.py
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env python
+"""
+This script scrapes the colombian chart of accounts in two variants from https://puc.com.co
```
deepeshgarg007: What is the purpose of this, does this even work?

blaggacao: Yes, this has been the script that created the CoA in the first place. From time to time, the official PUC may change and re-scraping would then be the most reproducible update strategy.

blaggacao: If that update use case doesn't justify the additional code (with it's potentially long maintenance tail), I can just remove this file or post it in some gist or so.

Let me know what the preferences are.

blaggacao: @deepeshgarg007 should we keep this file in here or rather push it into a gist or something?

deepeshgarg007: Separating it out would be preferred

blaggacao: @deepeshgarg007 : we're good I removed that file and placed a link to the gist into the commit message instead.

blaggacao:
@deepeshgarg007 I rebased on `develop` but I believe it is unrelated that the Documentation GH Action check is failing.

PR #34511, opened 2023-03-20
fix: Gross Profit reports Invoices with -ve qty for Invoices with Cr Notes (manual backport to version 13)

ruthra-kumar:
Manual backport of https://github.com/frappe/erpnext/pull/34456


Craftint:
Hi

For return invoices, credit note invoice is the one showing in gross profit with qty reduced but amount is showing whole amount with gross profit 100% which is incorrect.

image

PR #34512, opened 2023-03-20
Fix: Cost Per Unit Property read-only removed

nomi-gee:
Field Cost per unit had a property read_only checked which was causing the issue and the Field was not visible even after depends on return True

> Screenshots/GIFs


deepeshgarg007:
@nomi-gee I think the cost per unit value gets set by the system only. How does removing the read-only check help here. If the value is set, it should be visible, this doesn't look like the issue

nomi-gee:
> 
@deepeshgarg007 
Can you show how the system is setting its value when we have check enabled "Set Operating Cost Based On BOM Quantity" because the system is not setting the value from anywhere. You need to check this first.

PR #34521, opened 2023-03-21
fix: POS not picking up pos profile company address instead fetch any random company address

vishdha:
NCR-POP-UP-Retail-Store

Farmley-VJ-Business-Tower-201303-Billing (1)

Before:
ACC-PSINV-2023-00690

after:
ACC-PSINV-2023-00355


vishdha:
backport to version-14

PR #34541, opened 2023-03-21
Update serial_no_batch_selector.js

hereabdulla:
As there is a hard limit of 20 nos in frappe.db.get_list, auto fetched serial numbers more than 20 is not added in the new rows,this change avoid this error

### DIFF
--- a/erpnext/public/js/utils/serial_no_batch_selector.js
+++ b/erpnext/public/js/utils/serial_no_batch_selector.js
@@ -261,7 +261,8 @@ erpnext.SerialNoBatchSelector = class SerialNoBatchSelector {
 
 			return frappe.db.get_list("Serial No", {
 				filters: { 'name': ["in", selected_serial_nos]},
-				fields: ["batch_no", "name"]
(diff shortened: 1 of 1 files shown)


erpnext/public/js/utils/serial_no_batch_selector.js
```diff
@@ -261,7 +261,8 @@ erpnext.SerialNoBatchSelector = class SerialNoBatchSelector {
 
 			return frappe.db.get_list("Serial No", {
 				filters: { 'name': ["in", selected_serial_nos]},
-				fields: ["batch_no", "name"]
+				fields: ["batch_no", "name"],
+				limit: 999
```
deepeshgarg007, line 265: @hereabdulla 999 looks like a large limit for a UI component, is this actually needed? Can you share a gif of how more than 100 serial no looks like in the selector?

hereabdulla: but it is taking only 20 nos while fetching but we might have needed more than that

PR #34551, opened 2023-03-22
fix: Tax Inclusive pricing feature added to pos

vishdha:
Tax Inclusive which calculate tax within given rate

NCR-POP-UP-Retail-Store (1)

With Tax Inclusive 

https://user-images.githubusercontent.com/6947417/226828194-983824e4-ced3-4c49-9b63-19d4a1cf6877.mp4

Without Tax Inclusive 

https://user-images.githubusercontent.com/6947417/226828233-2e43bca5-2803-4715-8a38-188cfd4187fd.mp4

### DIFF
--- a/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
+++ b/erpnext/accounts/doctype/pos_invoice/pos_invoice.py
@@ -520,6 +520,11 @@ class POSInvoice(SalesInvoice):
 			if self.taxes_and_charges and not len(self.get("taxes")):
 				self.set_taxes()
 
+			# fetch tax inclusive charges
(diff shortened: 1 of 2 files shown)


deepeshgarg007:
If this is checked in the Sales Taxes and Charges template this will be applied in the invoices as well, why is the check needed?

deepeshgarg007:
Don't think this is needed, please reply and then reopen the issue if required

vishdha:
User is using same Sales Taxes and Charges template for pos or normal sales invoice for that reason that was needed but it ok we can create duplicate and assign it to pos profile. 
But UX wise this feature was required but your suggested work flow user can not guess easily

PR #34554, opened 2023-03-22

Jeevansyriac:
when we click on landed cost voucher button from a purchase invoice connection.
it doesn't automatically fill in certain fields in the child table of the Landed Cost voucher


deepeshgarg007:
@Jeevansyriac There is syntax error in this, please fix them and then reopen the PR

image

PR #34555, opened 2023-03-22
fix: valuation rate issue while making stock entry from PO

rohitwaghchaure:
1. Create subcontracted Purchase Order
2. Clicked on Transfer Materials to Supplier
3. System has not opened the stock entry form and throw an error that "Valuation Rate Missing"
4. Ideally this validation should throw on stock entry form and not on the purchase order form


rohitwaghchaure:
@mergifyio backport version-13

PR #34559, opened 2023-03-23
feat(asset_maintenance.json):Add 3 yearly in periodicity dropdown

BevanTony1:
Request: We want to have an additional option (3 Yearly) in the Asset Maintenance Task child table

Sample PR:
image


jeremythegreat01:
For visibility - @dottenbr

deepeshgarg007:
@BevanTony1 Just adding this in the select field options won't work. There are server-side implications to this, please handle those as well and then reopen the PR

PR #34564, opened 2023-03-23
feat: Serial No Normalization and Serial Batch Bundle

rohitwaghchaure:
New Doctype "Serial and Batch Bundle"

image

Inward Serial Nos using Serial and Batch Selector
inward_serial_nos

Inward Batches using Serial and Batch Selector
inward_batches

Outward Serial Nos using Serial and Batch Selector

outward_serial_nos

Outward Batches using Serial and Batch Selector
outward_batches

Configuration to Auto Pickup Serial / Batch Based on FIFO / LIFO / Expiry
image

What's new with Serial and Batch Bundle 

- Serial No field will be link field and not a small text
- Easy to rename, maintain valuation rate for the serial no
- Users no need to add multiple rows in case of Batch item in transactions like Purchase Receipt, Delivery Note, Stock Entry etc
- Users can bundle multiple batches in single record and linked to child table in the transactions.

- [x] Serial Nos with Small Text field removed from all transactions
- [x] Add Separate doctype Serial and Batch Bundle
- [x] New design of Serial and Batch Selector
- [x] Add Serial and Batch Bundle for Purchase Receipt
- [x] Add Serial and Batch Bundle for Delivery Note
- [x] Add Serial and Batch Bundle for Rejected Qty
- [x] Add Serial and Batch Bundle for Material Transfer
- [x] Add Serial and Batch Bundle for Stock Entry
- [x] Add Serial and Batch Bundle for Manufacture, Repack
- [x] Add Serial and Batch Bundle for Subcontracting
- [x] Add Serial and Batch Bundle for Stock Reco (add Auto for Current Serial and Batch)
- [x] Add serial and batch bundle for Packed Item
- [x] Update Warehouse and Status in the Serial Nos
- [x] Add Serial and Batch Bundle for Inter Company Transfer
- [x] Validate Negative Stock for Serial and Batch Bundle on Serial Batch Bundle
- [x] Don’t allow Future Transactions for Serial Nos (Add validation in Serial and Batch Bundle)
- [x] Auto create serial no and batch from Work Order
- [x] GIT Stock Entry
- [x] Warranty / AMC Details on serial no
- [x] Check all return scenarios 
- [x] Report to track history and status of serial nos
- [x] POS
- [x] Pick List
- [x] Repost quantity of Stock Reco in case of Batch
- [x] Fix Travis
- [x] Add new test cases

Docs https://docs.erpnext.com/docs/v14/user/manual/en/stock/serial-and-batch-bundle


muchai:
- Does this mean a user will be inputting qty needed, then select Fetch Based On FIFO, and the system auto allocates batches?
- Either way, this a great rewrite. I saw a customer decide not to use Batch Inventory this week, when they saw the experience of Users on Batch based inventory on POS.
- Also curious, # Will be deprecated in v16 comments

rohitwaghchaure:
> * Does this mean a user will be inputting qty needed, then select Fetch Based On FIFO, and the system auto allocates batches?

When the user clicked on Pick Serial / Batch, the system will set the default quantity in the POP up based on the quantity set in the transactions. Users can change the quantity as well in the modal which will be updated in the transactions after the selection of serial no/batches

> * Either way, this a great rewrite. I saw a customer decide not to use Batch Inventory this week, when they saw the experience of Users on Batch based inventory on POS.

Thank you! if you have any suggestion please share

> * Also curious, # Will be deprecated in v16 comments

Denormalization structure to the normalization structure we have made a lot of changes regarding serial / batch in the database. We have introduced a new doctype "Serial and Batch Bundle" which is now compulsory to use to pick / Add the serial nos/batches in the transaction. But for the old data, we are still using the Denormalization structure and for the new record, the normalization structure will be used. We have decided this because writing a patch to convert old data to a normalization structure was expensive and the system was taking a lot of time to complete the patch, especially in the case of large databases. So we have kept the old structure for old data and this structure will be removed in version 16.

As this feature has breaking changes we have decided not to backport this feature on version 14 and we'll release this feature in version 15.

muchai:
- When the user clicked on Pick Serial / Batch, the system will set the default quantity in the POP up based on the quantity set in the transactions. Users can change the quantity as well in the modal which will be updated in the transactions after the selection of serial no/batches

So if a user types qty 10 on the transactions, and this qty 10, can only be fulfilled by three batches, and user clicks on Pick Serial / Batch, the system fills three lines with three different batch nos/serial nos on the pop? If so, can the **Pick Serial / Batch** be avoided/automated?

- Thank you! if you have any suggestion please share

I will. Let me get to test it first.

- Denormalization structure to the normalization structure we have made a lot of changes regarding serial / batch in the database. We have introduced a new doctype "Serial and Batch Bundle" which is now compulsory to use to pick / Add the serial nos/batches in the transaction. But for the old data, we are still using the Denormalization structure and for the new record, the normalization structure will be used. We have decided this because writing a patch to convert old data to a normalization structure was expensive and the system was taking a lot of time to complete the patch, especially in the case of large databases. So we have kept the old structure for old data and this structure will be removed in version 16.

I see.

- As this feature has breaking changes we have decided not to backport this feature on version 14 and we'll release this feature in version 15.

Alright.

rohitwaghchaure:
> So if a user types qty 10 on the transactions, and this qty 10, can only be fulfilled by three batches, and user clicks on Pick Serial / Batch, the system fills three lines with three different batch nos/serial nos on the pop? If so, can the **Pick Serial / Batch** be avoided/automated?

Yes will add configuration to pick auto serial nos / batches (in Stock Settings). The popup is useful when user wants to pick serial nos / batches manually

PR #34567, opened 2023-03-23
fix: recalculate WDV rate after asset repair [v13]

anandbaburajan:
In https://github.com/frappe/erpnext/pull/34520, I recalculated the `expected_value_after_useful_life` after an asset repair in case the depreciation method was WDV, but it was incorrect. The `rate_of_depreciation` is what's supposed to be recalculated.


anandbaburajan:
@mergify backport version-13

PR #34596, opened 2023-03-27
feat: Reconcile Payments in background

ruthra-kumar:
For extremely large volumes of Invoices and Payments, Payment Reconciliation tool may timeout. This feature allows for running the same reconciliation tool in the background. 

Documentation can be found [here](https://docs.erpnext.com/docs/v14/user/manual/en/accounts/semi-auto-payment-reconciliation)

Changes:
1. 2 new doctypes: `Process Payment Reconciliation` and `Process Payment Reconciliation Log` have been introduced. These reuse the existing reconciliation logic. `Process Payment Reconciliation` will have similar filters as Payment Reconciliation. Once submitted, 'Reconcile in Background' button will be availabe. Once triggered, `Process Payment Reconciliation Log` will be created. Payments are allocated in FIFO order. 
Screenshot 2023-04-04 at 5 34 07 PM
Screenshot 2023-04-04 at 5 34 28 PM
Screenshot 2023-04-04 at 5 34 33 PM
Screenshot 2023-04-04 at 5 34 39 PM
Screenshot 2023-04-04 at 5 35 01 PM

2. Added a checkbox in `Accounts Settings` to enable/disable this feature 
3. Added few validations in Payment Reconciliation tool to check for running jobs with same filter. 
Screenshot 2023-04-04 at 5 37 58 PM
Screenshot 2023-04-04 at 5 38 36 PM


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -360,14 +357,29 @@
    "fieldname": "show_balance_in_coa",
    "fieldtype": "Check",
    "label": "Show Balances in Chart Of Accounts"
+  },
+  {
+   "default": "0",
+   "fieldname": "enable_auto_reconciliation",
+   "fieldtype": "Check",
+   "label": "Enable Auto Reconciliation"
```
deepeshgarg007: Explicitly specify payment reconciliation?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2023, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on("Auto Reconcile", {
```
deepeshgarg007: "Process Payment Reconciliation" keeping the currency naming convention like "Process Deferred Accounting" and "Process Loan Interest Accrual"?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
...
+			return {
+				"filters": {
+					"name": ["in", Object.keys(frappe.boot.party_account_types)],
+				}
+			}
+		});
+		frm.set_query('receivable_payable_account',  function(doc, cdt, cdn) {
+			return {
+				filters: {
+					"company": locals[cdt][cdn].company,
```
deepeshgarg007:
```suggestion
					"company": doc.company,
```

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
...
+					"is_group": 0,
+					"account_type": frappe.boot.party_account_types[locals[cdt][cdn].party_type]
+				}
+			};
+		});
+		frm.set_query('cost_center', function(doc, cdt, cdn) {
+			return {
+				filters: {
+					"company": locals[cdt][cdn].company,
```
deepeshgarg007:
```suggestion
					"company": doc.company,
```

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
...
+	onload: function(frm) {
+		// set queries
+		frm.set_query("party_type", function() {
+			return {
+				"filters": {
+					"name": ["in", Object.keys(frappe.boot.party_account_types)],
+				}
+			}
+		});
+		frm.set_query('receivable_payable_account',  function(doc, cdt, cdn) {
```
deepeshgarg007: Such queries should be under `refresh`. They should refresh if the company is changed

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
...
+				filters: {
+					"company": locals[cdt][cdn].company,
+					"is_group": 0,
+				}
+			};
+		});
+		frm.set_query('bank_cash_account', function(doc, cdt, cdn) {
+			return {
+				filters:[
+					['Account', 'company', '=', locals[cdt][cdn].company],
```
deepeshgarg007:
```suggestion
					['Account', 'company', '=', doc.company],
```

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
```diff
@@ -0,0 +1,64 @@
...
+					['Account', 'is_group', '=', 0],
+					['Account', 'account_type', 'in', ['Bank', 'Cash']]
+				]
+			};
+		});
+	},
+	refresh: function(frm) {
+		if (frm.doc.docstatus==1 && ['Queued', 'Running'].find(x => x == frm.doc.status)) {
+			var execute_btn = __("Reconcile in Background")
```
deepeshgarg007:
```suggestion
			let execute_btn = __("Reconcile in Background")
```

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.json
```diff
@@ -0,0 +1,172 @@
...
+   "report": 1,
+   "role": "System Manager",
+   "share": 1,
+   "write": 1
+  }
+ ],
+ "sort_field": "modified",
+ "sort_order": "DESC",
+ "states": [],
+ "title_field": "company"
```
deepeshgarg007: Should have a naming series, right? Why title with company?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+class AutoReconcile(Document):
+	def validate(self):
+		self.validate_receivable_payable_account()
+		self.validate_bank_cash_account()
+
+	def validate_receivable_payable_account(self):
+		if self.receivable_payable_account:
+			if not frappe.db.get_list(
+				"Account", filters={"name": self.receivable_payable_account, "company": self.company}
+			):
```
deepeshgarg007:
```suggestion
			if self.company != frappe.db.get_value("Account", receivable_payable_account, "company"):
```

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+				frappe.throw(
+					_("Receivable/Payable Account: {0} doesn't below to company {1}").format(
+						frappe.bold(self.receivable_payable_account), frappe.bold(self.company)
+					)
+				)
+
+	def validate_bank_cash_account(self):
+		if self.bank_cash_account:
+			if not frappe.db.get_list(
```
deepeshgarg007: Update same as suggested above?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+			):
+				frappe.throw(
+					_("Bank/Cash Account {0} doesn't below to company {1}").format(
+						frappe.bold(self.bank_cash_account), frappe.bold(self.company)
+					)
+				)
+
+	def before_save(self):
+		self.status = "Draft"
+		self.error_log = ""
```
deepeshgarg007: By default status for saved doc is "Draft", do we need this?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+
+	def before_submit(self):
```
deepeshgarg007: Shouldn't this be `on_submit`?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+		self.error_log = ""
+
+
+@frappe.whitelist()
+def get_progress(docname: str | None = None) -> float:
+	progress = 0.0
+	if docname:
+		reconcile_log = frappe.db.get_list(
+			"Auto Reconcile Log", filters={"auto_reconcile": docname}, as_list=1
+		)
```
deepeshgarg007: Use `get_value`?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+		if reconcile_log:
+			processed = frappe.db.get_value("Auto Reconcile Log", reconcile_log[0][0], "reconciled_entries")
+			total = frappe.db.get_value("Auto Reconcile Log", reconcile_log[0][0], "total_allocations")
```
deepeshgarg007: Merge in a single query

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+			elif total == 0:
+				progress = 100
+
+			return progress
+
+	return progress
+
+
+def get_pr_instance(doc: str):
+	auto_reconcile = frappe.get_doc("Auto Reconcile", doc)
```
deepeshgarg007: Use get_value instead of get_doc, not using any child table values so better to use get_value instead of get_doc

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+	for field in fields:
+		d[field] = auto_reconcile.get(field)
+	pr.update(d)
+	return pr
+
+
+def is_job_running(job_name: str) -> bool:
+	jobs = frappe.db.get_list("RQ Job", filters={"status": ["in", ["started", "queued"]]})
+	for x in jobs:
+		if x.job_name == job_name:
```
deepeshgarg007: `get_list` by default will get name field I guess, or will it return `job_name`.
Also can be done by a single `get_value` query

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
```diff
@@ -0,0 +1,407 @@
...
+			)
+		)
+
+		return
+
+	if not is_scheduler_inactive():
+		res = is_any_doc_running()
+		if not res:
+			# Get Queued item and start the process
+			queued = frappe.db.get_list(
```
deepeshgarg007: `get_all` is preferred on server side, `get_list` does permission checks and can cause unwanted errors in some cases

erpnext/accounts/doctype/auto_reconcile_log/auto_reconcile_log.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2023, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.ui.form.on("Auto Reconcile Log", {
+	refresh(frm) {
+		var progress = 0;
```
deepeshgarg007:
```suggestion
		let progress = 0;
```

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
ruthra-kumar: Yes. I mean, by default not every one needs this feature. Moreover, as this is still in development, better to let the user enable this explicitly if they want it.

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.js
ruthra-kumar: Huh.. It does have a common theme to it. I'll rename both of them.

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
deepeshgarg007: No, I meant the label, "Enable Auto Reconciliation" is ambiguous, it could be bank or payment

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.json
ruthra-kumar: Naming Series added

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
ruthra-kumar: Label has been changed

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: Removed it.

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: Using `on_submit` controller now

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: Done

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: merge?

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: This doctype doesn't have any child table. It only has the filters.

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: Parameters like "fields" are not yet implemented for `RQ Job` virtual doctype. Normal `get_all` output includes all the fields.
`get_value` is unsupported as well.
Screenshot 2023-04-01 at 2 55 14 PM

erpnext/accounts/doctype/auto_reconcile_log/auto_reconcile_log.js
ruthra-kumar: addressed

erpnext/accounts/doctype/auto_reconcile/auto_reconcile.py
ruthra-kumar: replaced with `get_all` as suggested

ruthra-kumar: addressed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+class ProcessPaymentReconciliation(Document):
+	def validate(self):
+		self.validate_receivable_payable_account()
+		self.validate_bank_cash_account()
+
+	def validate_receivable_payable_account(self):
+		if self.receivable_payable_account:
+			if self.company != frappe.db.get_value("Account", self.receivable_payable_account, "company"):
+				frappe.throw(
+					_("Receivable/Payable Account: {0} doesn't below to company {1}").format(
```
deepeshgarg007:
```suggestion
					_("Receivable/Payable Account: {0} doesn't belong to company {1}").format(
```

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+						frappe.bold(self.receivable_payable_account), frappe.bold(self.company)
+					)
+				)
+
+	def validate_bank_cash_account(self):
+		if self.bank_cash_account:
+			if self.company != frappe.db.get_value("Account", self.bank_cash_account, "company"):
+				frappe.throw(
+					_("Bank/Cash Account {0} doesn't below to company {1}").format(
```
deepeshgarg007:
```suggestion
					_("Bank/Cash Account {0} doesn't belong to company {1}").format(
```

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+	def on_submit(self):
+		self.db_set("status", "Queued")
+		self.db_set("error_log", None)
+
+	def on_cancel(self):
+		self.db_set("status", "Cancelled")
+
+
+@frappe.whitelist()
+def get_progress(docname: str | None = None) -> float:
```
deepeshgarg007: We have `frappe.publish_realtime` to update progress, tried that instead of this approach?

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation_list.js
```diff
@@ -0,0 +1,14 @@
+frappe.listview_settings['Process Payment Reconciliation'] = {
+	add_fields: ["status"],
+	get_indicator: function(doc) {
+		var colors = {
```
deepeshgarg007:
```suggestion
		let colors = {
```

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+				frappe.db.set_value("Process Payment Reconciliation", doc, "status", "Completed")
+
+
+@frappe.whitelist()
+def is_any_doc_running(for_filter: str | dict | None = None) -> tuple:
+	running_doc = None
+	if for_filter:
+		if type(for_filter) == str:
+			for_filter = frappe.json.loads(for_filter)
+		running_doc = frappe.db.get_all(
```
deepeshgarg007: Use `frappe.db.exists` or `frape.db.get_value` instead? `get_all` seems unnecessary

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+				"status": "Running",
+				"company": for_filter.get("company"),
+				"party_type": for_filter.get("party_type"),
+				"party": for_filter.get("party"),
+				"receivable_payable_account": for_filter.get("receivable_payable_account"),
+			},
+			as_list=1,
+		)
+	else:
+		running_doc = frappe.db.get_all(
```
deepeshgarg007: Use `frappe.db.exists` or `frape.db.get_value` instead? `get_all` seems unnecessary

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.js
```diff
@@ -0,0 +1,71 @@
...
+					['Account', 'company', '=', doc.company],
+					['Account', 'is_group', '=', 0],
+					['Account', 'account_type', 'in', ['Bank', 'Cash']]
+				]
+			};
+		});
+	},
+	refresh: function(frm) {
+		if (frm.doc.docstatus==1 && ['Queued', 'Running'].find(x => x == frm.doc.status)) {
+			var execute_btn = __("Reconcile in Background")
```
deepeshgarg007:
```suggestion
			let execute_btn = __("Reconcile in Background")
```

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+					)
+				)
+
+	def before_save(self):
+		self.status = ""
+		self.error_log = ""
+
+	def on_submit(self):
+		self.db_set("status", "Queued")
+		self.db_set("error_log", None)
```
deepeshgarg007, line 40: There is also a log document, right? we can log the errors there itself

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+		reconcile_log = frappe.db.get_value(
+			"Process Payment Reconciliation Log", filters={"process_pr": docname}, fieldname="name"
+		)
+		if reconcile_log:
+			processed = frappe.db.get_value(
+				"Process Payment Reconciliation Log", reconcile_log, "reconciled_entries"
+			)
+			total = frappe.db.get_value(
+				"Process Payment Reconciliation Log", reconcile_log, "total_allocations"
+			)
```
deepeshgarg007: Can be fetched in a single query?

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+					job_name=job_name,
+					doc=docname,
+				)
+				frappe.msgprint(_("Background Job {0} triggered").format(job_name))
+	else:
+		frappe.msgprint(_("Scheduler is Inactive. Can't trigger job now."))
+
+
+@frappe.whitelist()
+def trigger_reconciliation_for_queued_docs():
```
deepeshgarg007: Only trigger for queued docs, needs whitelisting?

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+	Fetch queued docs and start reconciliation process for each one
+	"""
+	if not frappe.db.get_single_value(
+		"Accounts Settings", "enable_payment_reconciliation_in_background"
+	):
+		frappe.throw(
+			_(
+				"Payment Reconciliation through backgound Job has been disabled. Enable it through {}"
+			).format(get_link_to_form("Accounts Settings", "Accounts Settings"))
+		)
```
deepeshgarg007: Jobs will result in an error if settings is disabled, should simply return

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
```diff
@@ -0,0 +1,500 @@
...
+					method="erpnext.accounts.doctype.process_payment_reconciliation.process_payment_reconciliation.fetch_and_allocate",
+					queue="long",
+					timeout="3600",
+					is_async=True,
+					job_name=job_name,
+					enqueue_after_commit=True,
+					doc=doc,
+				)
+		else:
+			reconcile_log = frappe.get_doc("Process Payment Reconciliation Log", log[0][0])
```
deepeshgarg007: You can simply pluck name, won't have to do log[0][0

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: Removed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: Addressed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: Addressed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: If it just returns, the  Cron Job will show up as Completed. User will have no way to know that the feature was disabled. The `Process Payment Reconciliation` doc, which will be in queued status, will also not have any indication that Job didn't run due to disabled feature.

I think it's better to throw an exception here.

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.js
ruthra-kumar: changed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation_list.js
ruthra-kumar: changed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: typo fixed

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: replaced with get_value

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: Yes. But, if the log document large(large no of allocation) just opening the log itself will be time consuming. So, i think its better to update in Parent doc as well. So, if the job fails, the error msg can be easily identified here

erpnext/accounts/doctype/process_payment_reconciliation/process_payment_reconciliation.py
ruthra-kumar: `frappe.publish_realtime` would be unsuitable in this case. We need a listener and the publishing event should happen in realtime. Say if the job paritally failed, then there won't be any progress published. Or even if the Job ran successfully, then there won't be job to publish the progress.

The log `Process Payment Reconciliation Log` has `reconciled_entries` variable which holds count of successfully reconciled entries. Just querying that is enough to know the progress.

PR #34600, opened 2023-03-27
fix: Multiple issues in purchase invoice submission

deepeshgarg007:
- There is a debit not equal to credit error if there is a multicurrency invoice with a lot of items, the base_net_total + taxes does not match the base grand total in such cases as it was calculated from party currency grand total using the conversion rate. Added a different logic for multicurrency invoices

- Did some minor perf optimization by removing queries from for loop


ruthra-kumar:
Minor difference (0.01) between base grand total and base round off total which doesn't happen on develop branch with the same exchange rate

**Exchange rate and taxes used for testing:**
Screenshot 2023-03-30 at 8 05 08 PM

Screenshot 2023-03-30 at 8 05 20 PM

**Develop:**
Screenshot 2023-03-30 at 8 08 34 PM

**PR:**
Screenshot 2023-03-30 at 8 02 18 PM

PR #34646, opened 2023-03-29
fix: asset monthly WDV and DD schedule and refactor [develop]

anandbaburajan:
Before:

Monthly depreciation calculations (i.e., with `frequency_of_depreciation` = 1) for WDV and DD methods were done incorrectly -- they were calculated just like how yearly depreciation calculations were done.

image

After:

Now, for monthly depreciation calculations for WDV and DD methods, the amounts are calculated on a yearly basis and then distributed monthly.

image

--

I've also refactored some parts of asset.py and asset_depreciation_schedule.py.


sagarvora:
@AnandBaburajan this may fail?
https://github.com/frappe/erpnext/blob/91a26608ee6a8cb09547e2d5059a36ae4daaa0d9/erpnext/assets/doctype/asset_value_adjustment/asset_value_adjustment.py#L165

anandbaburajan:
@sagarvora oops, thanks! Will fix.

PR #34675, opened 2023-03-30
feat: Auto set Party in Bank Transaction

### LINKED ISSUE #17: Automatically link Party Type and Party
_Party Type_ and _Party_ should be set automatically in **Bank Transaction**.

Match against **Supplier**, **Customer** and **Employee** records.

Prerequisite: add _Party IBAN_ field to **Bank Transaction**.

Flow:

1. check if earlier transactions with the same IBAN have a party linked and reuse that
2. try to match any **Bank Accounts** linked to the party
3. fuzzy-match full name against **Supplier**, **Customer** and **Employee** records

### LINKED ISSUE #34477: Improve Bank Reconciliation

Currently it's challenging to reconcile **Bank Transactions** with other transactions in ERPNext.

- **Bank Transaction** holds too little information to make sense of it. _Party Name_ and _Party IBAN_ as per bank statement are missing.
- It takes too many clicks to "pay" for a voucher with a **Bank Transaction**: create a **Payment Entry**, fill some data manually, go back to the **Bank Reconciliation Tool**, search the **Payment Entry**, reconcile with it.
- Matching **Bank Transactions** with parties and vouchers is still a very manual process. We can do some more "magic" to automate this.

- <s>Semi-automatic, self-improving matching of party names as per bank statement with parties in ERPNext. Fuzzy matching, correction by user, learn from past matches.</s>
- <s>Match parties by bank account number if their bank details are already in the system.</s>
- <s>Provide an easy way to store a parties bank details for future reference and matching.</s>
- Match **Bank Transactions** with vouchers. Likely matches:
    - <s>The voucher number is contained in the description of the Bank Transaction</s>
    - <s>There's an unpaid voucher for the same party with the same amount</s>
- <s>Provide the option to create and reconcile a **Payment Entry** with a (matched) voucher, in one click.</s>
- <s>Transparency: tell the user why a voucher was matched or why not</s>

**Possible checks**

a) Check for open vouchers with a similar amount as the **Bank Transaction** (5% tolerance) and in a certain time range around the bank transaction (two months backwards, 1 week forwards)
b) <s>Check for voucher numbers contained in the description of the **Bank Transaction**</s>
c) <s>Check for a similar party name (fuzzy matching with low tolerance)</s>
d) <s>Check for exact party names contained in the description of the **Bank Transaction**</s>
e) <s>Check if previous **Bank Transactions** with the same bank and account number have been matched with a party</s>
f) <s>Check if there's a party with the same bank and account number</s>
g) Transaction number in **Bank Entry** matches transaction number in **Bank Transaction**

- Automatically reconcile if `a AND (b OR e OR f OR g)`
- Propose party if `c OR d`
- Propose voucher if `a OR b OR g`
- Automatically set party `e OR f OR (a AND b)`

Competition:

- https://www.youtube.com/live/U1EYxuZ7Q2E?feature=share
- https://learn.microsoft.com/en-ie/dynamics365/business-central/bank-how-reconcile-bank-accounts-separately

Related:

- https://github.com/frappe/erpnext/issues/17685

marination:
Resolves part of https://github.com/frappe/erpnext/issues/34477
Ref: https://github.com/alyf-de/banking/issues/17
Docs: https://docs.erpnext.com/docs/v14/user/manual/en/accounts/bank-reconciliation

## Issue

- Bank Transaction holds almost no information about the Counter Party (the party we are paying/receiving from) as per the Bank Statement. 
- This counter party information is not likely to exactly match the Party in the system 
- Due to which the fields **Party Type** and **Party** are almost always blank. This makes identification hard in the **Bank Reconciliation Tool**
- Bank Transactions are usually bulk imported, hence setting these values manually are painful

## Enhancement

> **Note**: This matching is done fairly conservatively. It's better to not have a match than to have multiple wrong matches.

This PR adds an optional automatic Party matcher.

1. Enable it in **Accounts Settings**. Fuzzy matching can be optionally enabled
   Screenshot 2023-05-18 at 1 31 57 PM

2. Register primary/most used bank details like _Account Number_ & _IBAN_ against **Customer**, **Supplier** & **Employee** via Bank Account or Employee DocType (for Employees). This is optional and will make the match more accurate if set by users.

3. Added fields to hold Counter party information in **Bank Transaction**.(_The population of these values will depend on integrations that import transactions or the users that upload the bank statement_ )
   Screenshot 2023-04-11 at 12 27 31 AM

4.  If Account details are absent in the Bank Transaction, **fuzzy matching** is _optionally_ done using the **Party Name (Bank Statement)** or **Description**. These are matched against Customer/Supplier/Employee names.
5. **The matching is done and values are set on Submit if auto matching is enabled**
6. Users can correct the Party & Party Type, if incorrect, after submission. This will only impact the edited transaction.
    Screenshot 2023-04-11 at 1 18 19 AM

## Working

<details><summary>Expand to view the working</summary>

Automatch Flow

- **If more than one match was found against a description/party name, with the same score**, we skip matching and do not set a party in the bank transaction. We do not want to automate close calls, the user must decide these.

</details>

**TODO:**

- [x] Provision for user to correct mapping
- [x] Provision for users to create a mapping  (they can manually set the party in the transaction & also create a Bank Party Mapper)
- [x] Test
- [x] Docs
- [x]  Description
- [x] Use old bank fields to get party against bank data. Don't introduce new ways of storing bank data.
- [x] Optional fuzzy matching + If multiple similar matches, skip automatching

> **For the reviewer:** Hit me up for test data


pyproject.toml
```diff
@@ -12,6 +12,7 @@ dependencies = [
     "pycountry~=20.7.3",
     "Unidecode~=1.2.0",
     "barcodenumber~=0.5.0",
+    "rapidfuzz~=2.15.0",
```
ankush, line 15: _Just commenting on dependency_

This has compiled components but wheels seem to be available for most py versions on linux, also seems to be actively maintained so it should be fine.

marination:
@ Whoever reviews this: Hit me up for test data

barredterra (commented):
Tested the matching, LGTM 👍🏼 

Two doubts:

- [x] Why add bank details to **Customer** and **Supplier** instead of fetching linked **Bank Accounts**?
- [x] Why **Bank Party Mapper** instead of searching old **Bank Transactions**, which should hold the same information?
   -> "If the match was found on fuzzy matching 'description', no mapper record is created". We could no make this distinction when searching old **Bank Transactions**.

marination:
> Why add bank details to Customer and Supplier instead of fetching linked Bank Accounts?

@deepeshgarg007 Could use your inputs here.

There's no uniform way to register Bank Accounts with Parties 😢.
For Employee: Bank details are registered in separate fields in the doctype
For other parties: We have to select a Party in the Bank Account. But if there are 2 parties that share the same account then we have to duplicate the Bank Account and link the party

What would be ideal is multiple parties linked to a Bank Account, this way everything is in the same place. 
What I did here was add Bank Account No. and IBAN fields to three party types. But that's perhaps creating yet another way to link bank data to party. It's all ...messy 😵‍💫

Any suggestions ?

marination:
@deepeshgarg007 any thoughts ?

vorasmit:
How about working with matching rules (like we have tax rules or pricing rules)?

Users create rules based on the amount or description or reference no or any bank transaction field for that matter.
eg: If the description contains "ABC" and the Deposit amount is positive --> Party Type is Customer and Party is ABC Ltd.
or If the description contains "Interest" and the Deposit amount is positive --> Suggest the creation of Interest Entry (with Interest account)

Can be extended to cover any other use cases that may be useful.

marination:
@vorasmit  Thanks for your inputs 😺
> If the description contains "ABC" and the Deposit amount is positive --> Party Type is Customer and Party is ABC Ltd

This kind of already happens. Party matches are given priority depending on the if it is a withdrawal or deposit. If it is a deposit then we are receiving a payment, that means Customer should be matched first. Employee and Supplier are matched if nothing is found among Customers, because it is possible (but not often) that a Supplier gives you a refund which is a deposit in your account. 
`If the description contains "ABC" ` is checked via fuzzy matching, creating rules for this seems kind of redundant here and way too much work for a user since bank transaction descriptions are very volatile.
Users can manually create a Bank Party Mapper entry linking the account number/IBAN/account holder name to the Party which will correctly match in the future. Again description is left out due to volatility at the moment.

Also matching rules sort of suggest that there are too many possibilities/ways to match the party that they need to be configured. But I see very limited things for a quick match at the moment that I have tried my best to incorporate. 

**What could change perhaps is** giving users the option to select which field must be matched? Depending on what is used to fetch bank transactions, some fields could be useless. Like the Plaid integration only works with the description field, so the Party Name/ Account Holder becomes kinda useless at the moment. That can be determined in the future, it's a rather small feature/fix.

> If the description contains "Interest" and the Deposit amount is positive --> Suggest the creation of Interest Entry (with Interest account)

I believe this falls out of scope of this PR. We're just trying to match a party here to make it easier in the Bank Reconciliation Tool. But this does fall in the scope of https://github.com/frappe/erpnext/issues/34477 as a whole, feel free to leave a comment on the issue or collaborate, there's a lot more we have to do to close the issue :)

vorasmit:
I appreciate your efforts here.

Problems would not arise if there is an exact match with the account number and the party is set.

However, there are certain limitations of fuzzy matches.

Where there are similar party names, it would be either always correct or always incorrect. This in turn adds responsibility to check additionally if the party set is correct every time.

This could be beneficial in the long run, if there was additional logic built-in, to look for historical matches giving it a priority.

How are descriptions usually set by the bank and how rules could make things more firm?
With respect to India, if NEFT or RTGS is performed, the description would contain the name of the party as added to net banking. If a cheque is issued or received, it would contain the name as per the cheque.

Where a rule is created, it's almost certain of what the user is expecting. Hence the recommendation. The second example, I gave earlier was just to showcase how it could be extended to simplify reconciliation. This is how it is done in Zoho Books, and it seems to be useful.

barredterra:
> Where there are similar party names, it would be either always correct or always incorrect. This in turn adds responsibility to check additionally if the party set is correct every time.

@vorasmit the way Marica implemented this, you'll only have to fix wrong matches once. The system remembers and in the future this party/account will always be matched correctly.

vorasmit:
For the fuzzy match, I am not sure if a mapper is created. It's not possible to create the same against the description either.

https://github.com/frappe/erpnext/pull/34675/files#diff-9b46ef4b0625a3eeb63b99694b64928323b206ed346ac92d95d5788299549acbR25

Edit:
For a client, we did the recent implementation, they have over 20k customers. Many are similar and need suffixes as to from which area they belong. 

Final thoughts:  If you want many people to use this feature, it is important to ensure that it does not generate unexpected results.

marination:
You have a valid concern here. Many similar names could cause chaos.
I think your rules suggestion can be incorporated into bank party mapper. I'll keep the scope to party matching for now and work with the use case you provided. 
Knowing which key to look for is helpful, since we don't have that right now we resort to fuzzy matching. Thanks @vorasmit

marination:
@deepeshgarg007 Docs and description has been updated. The latest changes resolve the following ambiguities:

> Why add bank details to Customer and Supplier instead of fetching linked Bank Accounts?
   
Bank details are searched for in Bank Account now. In the case of **Employee**, the Bank Account is searched and then the Employee doctype as it has bank account fields in it historically.
No new ways to store bank details

> @vorasmit's concern: "For a client, we did the recent implementation, they have over 20k customers. Many are similar and need suffixes as to from which area they belong."

I did brainstorm this quite a bit and discussed this with @barredterra. Smit suggested rules but that would cause users to create way too many rules to match their plethora of parties. Could not really come up with anything else that wasn't bad UX.

So **fuzzy matching is optional**. In the docs, users are encouraged to test the feature out first. The matching might be a wee bit problematic where extremely similar names exist (e.g. Adithya Medical& Surgical vs Adithya Medical & Surgical vs Adithya Medical And Surgical) but will work great for data where there is not so much similarity.

I also increased the match score cutoff to 80.

Additionally fuzzy matching does not set any value if multiple matches with the same score were obtained for a query.
E.g. if the query in the Bank Transaction is "A.K. Medical" and two parties exist such as "A.K. Medical Hall" and "A.K. Medical Store", both will be accurate matches. In this case we do not make the system decide between close calls and leave it blank for the user to set it themselves or not set it at all.

barredterra (approved):
Tested matching by party name, description and IBAN. LGTM.

Todo: Update docs and PR description (to not confuse future readers).

marination:
@barredterra docs and PR description updated ✅

PR #34687, opened 2023-03-31

Jeevansyriac:
when we click on landed cost voucher button from a purchase invoice connection.
it doesn't automatically fill in certain fields in the child table of the Landed Cost voucher


deepeshgarg007:
@Jeevansyriac I don't think this still works. Please test this locally once before sending a PR

image

PR #34711, opened 2023-04-03
fix: use stock qty to calculate POS reserved stock

vishdha:
Use stock_qty when calculating POS reserved stock. Currently it uses the qty field which doesn't represent UOM's conversion factor, making the available stock incorrect.

https://github.com/frappe/erpnext/pull/33735


vishdha:
@deepeshgarg007  Please review and merge

PR #34726, opened 2023-04-03
fix: wrong qty of remaining work orders to be created when using "Create" > "Work Order"

danjeremynavarro:
This fixes the wrong calculation of remaining items for work order when using the "Create" > "Work Order" button on Sales Order


danjeremynavarro:
bump

marination:
@danjeremynavarro Please push this fix on the `develop` branch as well

PR #34727, opened 2023-04-04
feat: Auto allocate advance payments only against orders

deepeshgarg007:
Screenshot 2023-04-03 at 10 58 10 PM

In many cases, users only want to allocate advance payment if it's allocated against the orders mentioned in the Sales Invoice. Added a check that allows users to do this from the UI and via APIs

`no-docs`


precihole:
Hello, @deepeshgarg007 

Kindly do same for Advance Payment for Purchase Order also.

PR #34735, opened 2023-04-04
'Make Asset Movement' button translation fix in asset_list.js

ihesamyou:
'Make Asset Movement' button is not translatable in 'Asset List' in the app.
asset_list.js is edited to fix this.


anandbaburajan:
Thanks @ihosseinu.

PR #34738, opened 2023-04-04
fix: misc regional improvements

sagarvora:
- Simpler `get_region`
- Set `frappe.flags.company` where possible
- Use `functools.wraps` to preserve function signature in `allow_regional`


sagarvora:
This doesn't really need a review, merging.

PR #34739, opened 2023-04-04
chore!: remove override to set item tax based on HSN code

sagarvora:
We are now using the `doc_events` hook in India Compliance instead (https://github.com/resilient-tech/india-compliance/pull/556)


sagarvora:
This PR can be optionally backported a bit later to avoid inconsistencies with older versions of India Compliance.

s-aga-r (approved):

PR #34777, opened 2023-04-07
fix: time_in_minutes should return integer value

noec764:
fix: In job card doctype, the function time_in_minutes() should return integer value, so I change division to entire division by 60.

image


deepeshgarg007:
@noec764 Why the time in minutes should always be an integer? Can you please add some reasoning to this as well in the description

deepeshgarg007:
@noec764 Please explain the reason and then reopen the PR

PR #34778, opened 2023-04-07
fix: improper usage of get_depreciation_amount in asset_value_adjustment

anandbaburajan:
I missed to take care of `get_depreciation_amount` in `asset_value_adjustment.py` in #34646, so fixed that.


sagarvora (approved):
LGTM

PR #34801, opened 2023-04-10
fix: batch qty conversion factor issue fixed in pos transaction

vishdha:
Problem: Batch Qty not taking as stock qty.
In Batch .2 kgs  means 200 gram stock available.
If we try to sell 50g stock form it should allow to sell
which will be 200g - 50g = 150g it should allow but gives error

Before:

https://user-images.githubusercontent.com/6947417/230882688-2b705d9b-db64-49a0-a2d1-fc52581525f1.mp4

After:

https://user-images.githubusercontent.com/6947417/230882762-d9b805ae-6f5b-4b5c-9ea1-829b28d846ce.mp4


vishdha:
Backport to Version 14

vishdha:
@s-aga-r @rohitwaghchaure Please review and Merge this

erpnext/selling/page/point_of_sale/pos_controller.js
```diff
@@ -559,8 +559,10 @@ erpnext.PointOfSale.Controller = class {
 
 				item_row = this.frm.add_child('items', new_item);
 
-				if (field === 'qty' && value !== 0 && !this.allow_negative_stock)
-					await this.check_stock_availability(item_row, value, this.frm.doc.set_warehouse);
+				if (field === 'qty' && value !== 0 && !this.allow_negative_stock) {
+					const qty_needed = field === 'qty' ? value * item_row.conversion_factor : item_row.qty * value;
```
deepeshgarg007: It will enter inside if block only if the field is `qty`. Whats the point of adding a conditional operator here?

vishdha: If multi uom case it fetch conversion factor otherwise value always be 1 for example explain above if that remove it restrict user to use multiple uom setup. Default uom qty get validated so example 1 kg and product is 50g so imagine 800g = .8kg item available so it should allow to sell 160 unit of 50g. But it say item not available thats reason it added here

vishdha:
@deepeshgarg007 any update regarding this

erpnext/selling/page/point_of_sale/pos_controller.js
deepeshgarg007: But my point is that condition is already validated on if condition, why it is needed inside the lf block as well?

vishdha:
closing this since new PR raised for same https://github.com/frappe/erpnext/pull/34917

PR #34867, opened 2023-04-15
fix: batch_no field mapping from Purchase Receipt to Purchase Invoice

mohsinalimat:
batch_no field mapping from Purchase Receipt to Purchase Invoice


rohitwaghchaure:
@mohsinalimat Why you want to add batch_no field in the purchase invoice? If the update stock is disabled then there is no need of batch no in the purchase invoice

PR #34895, opened 2023-04-17
fix: use `CombineDatetime` instead of `Timestamp` in QB queries

s-aga-r:
Change `Timestamp` to `CombineDatetime` for queries written in `frappe.qb`.


erpnext/stock/doctype/batch/batch.py
```diff
@@ -192,7 +192,7 @@ def get_batch_qty(
 				posting_time = nowtime()
 
 			query = query.where(
-				Timestamp(sle.posting_date, sle.posting_time) <= Timestamp(posting_date, posting_time)
+				UnixTimestamp(sle.posting_date, sle.posting_time) <= UnixTimestamp(posting_date, posting_time)
```
ankush: `CombineDatetime` is what you're looking for not unix timestamp?

s-aga-r: @ankush Is there any difference between sql `timestamp` and frappe.qb `UnixTimestamp`? As we are using sql `timestamp` in most of the queries.

PR #34896, opened 2023-04-17
fix: labels and translations

barredterra:
- Fix some german translations of **Expense Claim**
- Consistent labels and translations in **Project**/**Task** (sections "Start and End Dates", "Costing and Billing")

   For example, we used to have "Total Costing Amount (via Timesheets)" in **Project** and "Total Costing Amount (via Time Sheet)" in **Task**. This is merged to become "Total Costing Amount (via Timesheet)" in both **Project**  and **Task**.
   
   The source strings are updated for all languages, while the translation stays the same.

- Fix some german translations of **Stock Entry**

> no-docs


barredterra:
CI is failing because of some translation files with bad structure, which was not caused by this PR. Will this fix in a separate PR: https://github.com/frappe/erpnext/pull/34897

deepeshgarg007:
@barredterra Can you resolve conflicts on this PR?

barredterra:
Done, @deepeshgarg007

PR #34897, opened 2023-04-17
fix: translation files

barredterra:
All translations should be of structure `source string,translated string,context`

- remove leading "DocType: XXX", if any
- remove leading path and line number, if any
- remove leading comma, if any
- add trailing comma, if missing (three columns total)

Related: https://github.com/frappe/frappe/pull/20750


barredterra:
@ankush do you see any issues with this? I think it will improve translation coverage and ensure that we have better data before migrating to pot-files.

Issue #34908, opened 2023-04-18
Item rate in Asset Repair is Not Retrieved in V14


ihesamyou:
### Information about bug

I have been using item table of asset repair in V13 and incoming rate of items fetch correctly. The problem is in V14, when the get_incoming_rate API returns 0 or throws "zero valuation rate" error. Please note that I created two identical new items in V13 and V14 for testing this API and the data is not the problem here (in stock ledger report they have rate in the selected warehouses). Also our customizations in this form does not affect this issue.

In V13 we get valuation rate of 10,000,000 for item 123456:

![V13 Asset Repair](https://user-images.githubusercontent.com/86075967/232755097-6898f237-61df-42ba-bb17-29b159a39e3f.JPG)

While in V14 we get valuation rate of 0 for item 123456:

![V14 Asset Repair](https://user-images.githubusercontent.com/86075967/232755536-3ab140c6-91ea-4d91-920e-143fe9aa3f08.JPG)

### Module

stock

### Version

ERPNext V14

### Installation method

None

ihesamyou:
Fixed!

PR #34909, opened 2023-04-18
Get incoming rate v14 fix

ihesamyou:
There are two problems with get_incoming_rate API in v14 stock module. You should call this method in Asset Repair Stock Items table to see this problem (As stated in [This Issue](https://github.com/frappe/erpnext/issues/34908)) :

1- In get_incoming_rate function the in_rate value is calculated based on certain conditions. In each condition a certain function is called. These functions should assign None to in_rate if there is no data for that item with the given args but they assign 0 to in_rate even if there is no data for the given args and item (meaning None). This results in the last condition (see code below) to not be executed. 

V14 condition

This problem is because of "or 0" and "else 0" in the code.
I suspect this is something that is missed because of the changes in the logic of the last condition in V14.
The first commit will fix this problem.

2- The second problem is that the voucher_no value in v14 is None while in v13 it was empty string. voucher_no is passed to get_valuation_rate function and is used in two queries from stock_ledger.py.
In v13 which voucher_no is passed as empty string, these queries work correctly while in v14 which the voucher_no is passed as None, these queries return no value and as the result the "Valuation Rate Missing" Error is thrown (based on  get_valuation_rate method in stock_ledger.py).

Here are the args and queries execueted in v13 (compare voucher_no with v14):

V13

And here are the args and queries execueted in v14:

V14

If you run these queries on database you see that we get no value if the voucher_no = Null.
I could not find out why voucher_no = "" in v13 and why voucher_no = None in v14 considering that the only thing changed about voucher_no in v14 is the scope that the voucher_no is defined.
Moving voucher_no assignment into the last condition (as it was before in v13) fixes this problem.


ihesamyou:
@rohitwaghchaure Is the issue being investigated now?

PR #34910, opened 2023-04-18
fix: allow submit delivery note when the sales order was billed...

HarryPaulo:
Thinking of a flow below, where the Delivery Note and the Sales Invoice are generated from the sales order.

image

When the customer is checking for bypass_credit_limit_check_at_sales_order = 1 the system is forcing the validation of credit in the delivery note.

This makes sense because if the delivery note did not come from the invoice and skipped the credit validation on the order, it must validate on the delivery note.

However, if the order is already 100% invoiced even if the customer does not have credit, it means that he paid in advance or paid the invoice, which makes it possible for the delivery note not to validate the credit limit as it already happened on the invoice of sale and must not appear on the delivery note.

### DIFF
--- a/erpnext/stock/doctype/delivery_note/delivery_note.py
+++ b/erpnext/stock/doctype/delivery_note/delivery_note.py
@@ -271,6 +271,9 @@ class DeliveryNote(SellingController):
 	def check_credit_limit(self):
 		from erpnext.selling.doctype.customer.customer import check_credit_limit
 
+		if self.per_billed == 100:
+			return
+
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@HarryPaulo What will happen in case of partial invoicing and delivery of that partial invoiced items?
The percentage billed will not be 100 in that scenario

HarryPaulo:
@deepeshgarg007 in this case, follow the normal flow and perform credit validation.

PR #34917, opened 2023-04-19
fix: batch qty conversion factor issue fixed in pos transaction

vishdha:
Problem: Batch Qty not taking as stock qty.
In Batch .2 kgs  means 200 gram stock available.
If we try to sell 50g stock form it should allow to sell
which will be 200g - 50g = 150g it should allow but gives error

Before:

https://user-images.githubusercontent.com/6947417/230882688-2b705d9b-db64-49a0-a2d1-fc52581525f1.mp4

After:

https://user-images.githubusercontent.com/6947417/230882762-d9b805ae-6f5b-4b5c-9ea1-829b28d846ce.mp4


vishdha:
@deepeshgarg007 check this pr

PR #34918, opened 2023-04-19
perf: Journal Entries

s-aga-r:
**Source / Ref:** ISS-22-23-06298

refactor `get_stock_value_on()` to get Stock Value for a list of warehouses.

### DIFF
--- a/erpnext/stock/utils.py
+++ b/erpnext/stock/utils.py
@@ -9,9 +9,9 @@ import frappe
 from frappe import _
 from frappe.query_builder.functions import CombineDatetime, IfNull, Sum
 from frappe.utils import cstr, flt, get_link_to_form, nowdate, nowtime
-from pypika.terms import ExistsCriterion
 
 import erpnext
+from erpnext.stock.doctype.warehouse.warehouse import get_child_warehouses
 from erpnext.stock.valuation import FIFOValuation, LIFOValuation
 
 BarcodeScanResult = Dict[str, Optional[str]]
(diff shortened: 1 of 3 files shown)


erpnext/stock/utils.py
```diff
@@ -57,39 +58,39 @@ def get_stock_value_on(warehouse=None, posting_date=None, item_code=None):
...
-		ORDER BY timestamp(posting_date, posting_time) DESC, creation DESC
-	""".format(
-			condition
-		),
-		values,
-		as_dict=1,
-	)
+		query = query.where(sle.item_code == item_code)
+
+	stock_ledger_entries = query.run(as_dict=True)
```
ankush: why are we fetching all SLE to sum them up in python, we can just sum them up in SQL also :grimacing:

s-aga-r: > why are we fetching all SLE to sum them up in python, we can just sum them up in SQL also grimacing

That's the objective of this PR.

s-aga-r: @ankush 

```
sle_map = {}
	for sle in stock_ledger_entries:
		if not (sle.item_code, sle.warehouse) in sle_map:
			sle_map[(sle.item_code, sle.warehouse)] = flt(sle.stock_value)
```

Here, we are only considering the first row(item_code, warehouse). How can we achieve the same in qb?

rohitwaghchaure: Sum(stock_value_difference) group by item_code, warehouse will solve this

rohitwaghchaure: Stock reco needs to handle seperately using case statement

PR #34953, opened 2023-04-20
fix: `PermissionError` in Work Order

s-aga-r:
**Source / Ref:** ISS-23-24-00209

Use `__onload` to get the value in JS instead of `get_doc`.

### DIFF
--- a/erpnext/manufacturing/doctype/work_order/work_order.js
+++ b/erpnext/manufacturing/doctype/work_order/work_order.js
@@ -625,20 +625,18 @@ erpnext.work_order = {
 								// all materials transferred for manufacturing, make this primary
 								finish_btn.addClass('btn-primary');
 							}
-						} else {
-							frappe.db.get_doc("Manufacturing Settings").then((doc) => {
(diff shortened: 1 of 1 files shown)


erpnext/manufacturing/doctype/work_order/work_order.py
```diff
@@ -73,6 +73,9 @@ def onload(self):
 		self.set_onload("material_consumption", ms.material_consumption)
 		self.set_onload("backflush_raw_materials_based_on", ms.backflush_raw_materials_based_on)
 		self.set_onload("overproduction_percentage", ms.overproduction_percentage_for_work_order)
+		self.set_onload(
+			"overproduction_percentage_for_work_order", ms.overproduction_percentage_for_work_order
+		)
```
rohitwaghchaure: Use overproduction_percentage field, why new field overproduction_percentage_for_work_order?

s-aga-r: just missed it

PR #34969, opened 2023-04-20
fix: per_billed condition for Payment Entry

barredterra:
Old behavior:

- v13: if SO has been billed `>= 0 %`, you cannot create a Payment Entry against the SO
- v14: if SO has been billed `>= 99.99 %`, you cannot create a Payment Entry against the SO

New behavior:

- v14: if SO has been billed `>= 100 % + over_billing_allowance`, you cannot create a Payment Entry against the SO

(same applies to PO)

Use case: in services, orders can get more expensive during fulfillment. Overbilling allows us to put more on the sales invoice. But it was not possible to create overbilled advance payments against the SO, due to above validations.


barredterra:
@deepeshgarg007 no more fixes for v13?

PR #35000, opened 2023-04-23
fix: v14, Bank Reconcile Tools not cover case JV debit bank

kittiu:
Simply put, currently, Bank Reconcile Tools, can only reconcle Bank Transaction with Journal Entry on case Payout (JV with Credit Bank)

Steps:
- Create Bank Transaction with Deposit amount = 1
- Create Journal Entry with
  - Debit Bank Account = 1
  - Credit Bank Fee = 1
- Do Bank Reconcile for this record, you can reconcile, but,
  -  Journal Entry's Clearance Date is set
  - But Bank Transaction is not reconciled.

This PR fix that.


kittiu:
@deepeshgarg007 @ruthra-kumar Anything else to move this forward? Thanks!

PR #35010, opened 2023-04-24
fix: value of depreciable assets not updating after manual depr entry [v14]

anandbaburajan:
- A user had to create a depreciation entry manually for an asset when he later realised that the amounts in the asset's depreciation schedule weren't matching with the auditor's book. But since our system didn't handle the case of manual depreciation entries for depreciable assets, the value of the asset wasn't updated. So I fixed that.
- The Asset Depreciations and Balances report didn't reflect the depreciation values for manual depreciation entries, so fixed that as well. 
- The system required users to cancel the asset and the asset movement when they tried to cancel a particular depreciation entry, so added those doctypes to journal entry doctype's ignore_doctypes_on_cancel_all.
- Added validation to make sure that depreciation entries have voucher type set to 'Depreciation Entry'.


saeedkola:
Hello @anandbaburajan 
Is this below condition really needed in journal_entry.py?
`and d.account_type == "Depreciation"`

I have a case where depreciation expense accounts set in the asset category have account_type=Expense and Workflows are set up on Journal Entry.
As a result after the journal entry is approved by the workflow the value_after_depreciation is not getting updated in the asset.
Now there is no other place where we are validating the depreciation expense account to have account type as Depreciation. Neither in Company defaults nor in Asset category. 
How will the user know that this is a must in order for the feature to work correctly?
How can this be dealt in the best possible way?

anandbaburajan:
Thanks @saeedkola. I've added a validation here https://github.com/frappe/erpnext/pull/36659 for the account in asset category.

PR #35024, opened 2023-04-25
Change default Customer naming to Naming Series

aynugek:
On a new installation, a customer is identified by their name. This becomes a problem if you get customers who share names as it raises a duplicate exception. I think the default naming scheme should be series, and then an administrator can change it to the customer's name if need be as opposed to it being the other way around.


deepeshgarg007:
@aynugek I don't think this change should be made, if there is a duplication the system by default handles this. Closing this PR

aynugek:
Alright. Some time back I used to get that error when trying to create new customers by using the REST APIs. How does it handle the situation currently? Does it append numbers, as is the case with ammended invoices?

PR #35032, opened 2023-04-25
chore: set correct currency symbol in salary register for multi-currency salary slip

saurabh6790:


saurabh6790:

PR #35057, opened 2023-04-25
fix:"Customer" and "Sales order" field add in Task

IgorA100:
A Task can be without a Project.
But a task can refer to a Customer !


IgorA100:
Is there something wrong ?

deepeshgarg007:
@IgorA100 I don't think this is a generic requirement. These links can also be added as customization by the users needed. Closing this PR

PR #35061, opened 2023-04-26
refactor: don't book exchange gain/loss for sales/purchase orders

ruthra-kumar:
### Don't book Exchange Gain/Loss while referencing Sales/Purchase Order

Sales/Purchase Orders don't have any accounting impact. So, Booking Gain/Loss for a Multi-Currency Order doesn't make sense.


deepeshgarg007 (commented):
See similar logic being used for the exchange rate in both places, maybe move this to a function and use that?

Also if we can add a simple unit test for this

PR #35085, opened 2023-04-27
fix: handle finance book properly in trial balance and general ledger [develop]

anandbaburajan:
### Users not using any finance books

They don't need to change/do anything. Trial balance would reflect gl entries with fb == none by default.

### Users using one finance book "fb" (accidental case)

Trial balance would reflect gl entries with fb == none by default. So they need to set default_finance_book as "fb" in their company settings. Then trial balance would reflect gl entries with fb == "fb" or fb == none by default. Otherwise they would need to set their finance book "fb" every time in the trial balance filters.

#### If they have set a default_finance_book "fb1"

Trial balance would reflect gl entries with fb == "fb1" or fb == none by default
- If they need trial balance to reflect gl entries with fb == none, they need to uncheck include_default_book_entries
- If they need trial balance to reflect gl entries with fb == "fb2" or fb == none, they need to uncheck include_default_book_entries and then choose "fb2"

#### If they have not set a default_finance_book

Trial balance would reflect gl entries with fb == none by default
- If they need trial balance to reflect gl entries with fb == "fb1" or fb == none, they need to choose "fb1"
- If they need trial balance to reflect gl entries with fb == "fb2" or fb == none, they need to choose "fb2"

### DIFF
--- a/erpnext/accounts/report/financial_statements.py
+++ b/erpnext/accounts/report/financial_statements.py
@@ -541,7 +541,9 @@ def apply_additional_conditions(doctype, query, from_date, ignore_closing_entrie
 			company_fb = frappe.get_cached_value("Company", filters.company, "default_finance_book")
 
 			if filters.finance_book and company_fb and cstr(filters.finance_book) != cstr(company_fb):
-				frappe.throw(_("To use a different finance book, please uncheck Include Default Book Entries"))
+				frappe.throw(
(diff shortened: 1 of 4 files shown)


deepeshgarg007:
@AnandBaburajan Can you update the same behaviour in General Ledger, so that there are no mismatch issues anymore?

anandbaburajan:
@deepeshgarg007 yep, done.

deepeshgarg007 (approved):

PR #35112, opened 2023-05-01
refactor(Gross Profit): simplify group by invoice logic

ruthra-kumar:
SImplifying `Group By`: 'Invoice' logic in Gross Profit report for better readability


ruthra-kumar:
bump

Issue #34920, opened 2023-04-19
Asset Depreciation Schedule is Wrong!


Olawale1:
### Information about bug

Hi ,

There seems to be a serious problem with the Asset Depreciation Schedule in the recent versions of ERPNext (V13 and V14). It becomes very obvious in existing assets that are long lasting... the schedule extends beyond the Gross Purchase Amount and an extra line is added with a negative value (apparently in some form of attempt to bring the value back to the Gross Purchase Amount) !

![Asset Depreciation Issue 01](https://user-images.githubusercontent.com/9269337/233007740-f18edec8-4efb-4494-aa22-cae0085af058.png)
![Asset Depreciation Issue 02](https://user-images.githubusercontent.com/9269337/233007811-05597ae3-2b21-45f6-a9f2-da8753d6fe99.png)

This needs urgent fixing

Kind regards,

### Module

assets

### Version

ERPNext: v13.49.13 (version-13)
Frappe Framework: v13.54.0 (version-13)

Same in V14

### Installation method

None

Olawale1:
This issue is affecting both Straight Line and Manual Depreciation methods which I believe are the most widely used… It needs to be fixed urgently

Thanks

anandbaburajan:
@Olawale1, 6 number of depreciations booked means that the asset has been depreciating for 6 months (since the frequency of depreciation is set to 1). But the asset was available for use in March and the current month is May, so it hasn't been 6 months. Can you please explain? I guess our docs for existing assets need some more improvement?

Olawale1:
Hi @AnandBaburajan 

Even with that adjustment, schedule is still wrong! Pls see screenshot below:

![Screenshot_20230507-180954](https://user-images.githubusercontent.com/9269337/236696966-cfe3e1fe-1807-4eab-ae59-b3aca2a2cc92.jpg)

The schedule should have 336 Depreciations, instead it has 306 !

Olawale1:
Previously, the simple logic for existing assets was that the system subtracts the Number of Depreciations already booked for Total Number of Depreciations to know how many Depreciations should be included in the schedule

Also, Sum of Gross Accumulated Depreciation and Expected Value after Useful Life was subtracted from Gross Purchase Amount to know what value should be evenly spread across the schedule with the last row absorbing whatever slight difference there may be due to rounding

I really don't  think the issue of whether or not the Number of Depreciations booked from the Available for use

Olawale1:
[...]

PR #35117, opened 2023-05-01
fix: handle expected_value_after_useful_life properly in asset value adjustment

anandbaburajan:
Asset Value Adjustment creates wrong depreciation schedule for assets with expected_value_after_useful_life, so fixed that.


Olawale1:
Hi @AnandBaburajan @deepeshgarg007 

Thanks a lot for this fix but the entire schedule on the Asset is wrong... not only after adjustment

Reference: https://github.com/frappe/erpnext/issues/34920

Pls take a look if possible... could be the same issue affecting both

Many thanks

anandbaburajan:
Hi @Olawale1, I missed to see the issue you opened. Will take a look, thanks!

PR #35140, opened 2023-05-03
fix: remove code that causes upscrolling

Didiman1998:
**Title: Fix scrolling issue in variant dialog window for single variants**

**Description:**
This pull request fixes an issue in the variant dialog window for single variants in ERPNext. When the dialog window is too long, it keeps scrolling back to the top, making it difficult to use.

To fix this issue, we have removed some of the extra lines of code that were causing the problem, while keeping the same functionality. With this fix, users will be able to easily scroll through the dialog window without any interruptions.

Thanks for considering this pull request.

**BEFORE**
problem_before_my_change

**AFTER**
problem_solved_after_my_change

### DIFF
--- a/erpnext/stock/doctype/item/item.js
+++ b/erpnext/stock/doctype/item/item.js
@@ -772,12 +772,6 @@ $.extend(erpnext.item, {
 					if (modal) {
 						$(modal).removeClass("modal-dialog-scrollable");
 					}
-				})
-				.on("awesomplete-close", () => {
-					let modal = field.$input.parents('.modal-dialog')[0];
-					if (modal) {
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@Didiman1998 Would be helpful if you can add a before and after gif to help understand the issue better

Didiman1998:
@deepeshgarg007 

Sure thing! 
I just captured my screen and added before-after in the description.

Thank you for testing.

Best Regards
Dietmar Fischer

Didiman1998:
@deepeshgarg007 Current stand?

WebsiteDeveloper:
@deepeshgarg007 would it be possible to have this backported for version 14 and/or 13?

PR #35145, opened 2023-05-03
fix: based on status_update.py update opportunity status to converted…

HarryPaulo:
Looking at status_updater.py, the logic for the opportunity to be converted is to create a **Sales Order**.

However, currently this status does not update, I create the sales order and the status of the opportunity remains as **Quotation**.

So to solve the problem, when updating the Quotation status, it already updates the Opportunity status.

On submit sales order set opportunity status to **Converted**, on Cancel set status to **Quotation** back.

### DIFF
--- a/erpnext/selling/doctype/sales_order/sales_order.py
+++ b/erpnext/selling/doctype/sales_order/sales_order.py
@@ -217,6 +217,7 @@ class SalesOrder(SellingController):
 					frappe.throw(_("Quotation {0} is cancelled").format(quotation))
 
 				doc.set_status(update=True)
+				doc.update_opportunity("Converted" if flag == "submit" else "Quotation")
 
 	def validate_drop_ship(self):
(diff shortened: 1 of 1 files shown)


HarryPaulo:
I wondered if this type of treatment should be carried out in a personalized way according to the client...

Why can there be cases where the customer considers it converted when the order is placed, in other cases when it is invoiced...

But what led to this change was looking at the status_updater.py for the Opportunity

deepeshgarg007:
@HarryPaulo There is a function to check if the opportunity has any ordered quotation named `has_ordered_quotation`. Can you also check why this didn't work and remove it if not getting called or needed?

HarryPaulo:
@deepeshgarg007, Currently in Quotation, the Opportunity status only changes on: on_cancel, on_submit...

As I am making a sales order, it does not change the status, so my change was made.

PR #35156, opened 2023-05-04
fix: Incorrect Earned Leaves Proration

JunKangChin:
If today date is after or equal to employee's joining date, current month should be considered in earned leaves proration.

## Test Cases
- [x] Today Date: 1, Joining Date: 5, Result: +0
- [x] Today Date: 5, Joining Date: 5, Result: +1
- [x] Today Date: 24, Joining Date: 5, Result: +1


ruchamahabal (approved):

ruchamahabal:
Already fixed in https://github.com/frappe/hrms

PR #35162, opened 2023-05-04
When creating land cost voucher from PI or PR, the system takes base rate instead of base net rate, the problem is if we apply discount to the PI/PR then at land cost voucher the discounted rate does not shows it shows base rate and base net rate.

kunhimohamed:
> When creating land cost voucher from PI or PR, the system takes base rate instead of base net rate, so if we apply discount to the invoice it will not shows. 

> Changed *get_items_from_purchase_receipts* method of *LandedCostVoucher* class.


deepeshgarg007:
@kunhimohamed Please add a description and then reopen the PR

kunhimohamed:
Changed the description but am not able to see the re-open button.

PR #35189, opened 2023-05-06
fix: handle empty FBs properly in TB and GL [v14]

anandbaburajan:
https://github.com/frappe/erpnext/pull/35136 didn't handle finance books set to "", so fixed that.


anandbaburajan:
@mergify backport version-14

PR #35216, opened 2023-05-08
fix: broken save on empty row existance

rtdany10:
https://user-images.githubusercontent.com/52111700/236823586-00374089-bcdc-411b-bbb8-34d3ece4edfc.mp4

Problem:
Unable to save the document if the first save action was initiated with an empty row. Why leave the first row empty? Cause of multi select.

Cause:
- [`remove_empty_rows`](https://github.com/frappe/frappe/blob/develop/frappe/public/js/frappe/form/save.js#L43C6-L83) uses [`get_all_docs`](https://github.com/frappe/frappe/blob/develop/frappe/public/js/frappe/model/model.js#L791C10-L802) to fetch the child table rows.
- `get_all_docs` loops the document and if any key(field) is an array, treats it as a table
- [`_calculate_taxes_and_totals`](https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/taxes_and_totals.js#L93-L106) adds a new key `_items` to the doc.
- `_items` is not an actual table, but a clone of the actual `items` field.
- `get_all_docs` treats `_items` and `items` as different table causing duplicated rows
- Trying to clear this duplicate empty row throws an error and takes away the user's ability to save.

### DIFF
--- a/erpnext/public/js/controllers/taxes_and_totals.js
+++ b/erpnext/public/js/controllers/taxes_and_totals.js
@@ -92,7 +92,7 @@ erpnext.taxes_and_totals = class TaxesAndTotals extends erpnext.payments {
 
 	_calculate_taxes_and_totals() {
 		const is_quotation = this.frm.doc.doctype == "Quotation";
-		this.frm.doc._items = is_quotation ? this.filtered_items() : this.frm.doc.items;
+		this.frm._items = is_quotation ? this.filtered_items() : this.frm.doc.items;
 
 		this.validate_conversion_rate();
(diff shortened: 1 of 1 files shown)


rtdany10:
Introduced via: https://github.com/frappe/erpnext/pull/33874

rtdany10:
@ruthra-kumar @deepeshgarg007 @marination

ruthra-kumar (approved):

ruthra-kumar:
LGTM

ruthra-kumar:
LGTM

PR #35248, opened 2023-05-10
refactor: get default contact or address

barredterra:
Change `get_party_shipping_address` and `get_default_contact` to use `frappe.get_all`.

Not a pure refactor:

- earlier, we always returned some ~~**Address** or~~ **Contact**, no matter if it had "is_primary", "is_shipping", etc. set. Now we return primary contacts or None.

    This is an improvement IMO: we don't set stupid defaults but let the user choose explicitly when we don't have a good default.
    
- ~~earlier we queried all **Address** or **Contact**, disregarding user permissions. Now we query only the ones a user can see.~~

    ~~If this is not desirable for some reason, we can switch to `get_all`~~


PatrickDEissler:
@ruthra-kumar 
Hi Ruthra,
do you have any chance to check on this PR? 
I think the refactor makes sense as it brings more clarity into contact selection.

deepeshgarg007:
@barredterra The same is also used by website users and using `get_list` for the address query is leading to permission errors, which can be seen in the failing CI tests, maybe for now use `get_all` instead of `get_list`?

barredterra:
Sorry Deepesh, I have missed that. I changed it to use `get_all` now.

ruthra-kumar:
@barredterra 

> * earlier, we always returned some **Address** or **Contact**, no matter if it had "is_primary", "is_shipping", etc

This, I'm unable to confirm. Locally, existing code does work for me -- only pulling shipping **Address** or primary **Contact**. Any specific case that i can replicate with?

> * earlier we queried all **Address** or **Contact**, disregarding user permissions. Now we query only the ones a user can see.

Without `get_all` website users will face permission issues as @deepeshgarg007  mentioend. With `get_all`, the above intent is negated.

We can merge this, at the least for the rewrite of `frappe.db.sql` to `frappe.get_all`

barredterra:
> This, I'm unable to confirm. Locally, existing code does work for me -- only pulling shipping Address or primary Contact. Any specific case that i can replicate with?

If you don't have a primary contact, any non-primary contact used to get pulled. This was only the case for Contact, Address already worked fine.

> We can merge this, at the least for the rewrite of frappe.db.sql to frappe.get_all

Yes, I have already switched to `get_all` after Deepesh's review. 👍🏼

erpnext/accounts/party.py
```diff
@@ -850,33 +852,23 @@ def get_dashboard_info(party_type, party, loyalty_program=None):
...
+			["Dynamic Link", "link_doctype", "=", doctype],
+			["Dynamic Link", "link_name", "=", name],
+			["disabled", "=", 0],
+		],
+		or_filters=[
+			["is_shipping_address", "=", 1],
+			["address_type", "=", "Shipping"],
+		],
+		pluck="name",
+		limit=1,
```
barredterra, line 878:
```suggestion
		limit=1,
		order_by="is_shipping_address DESC",
```

Make sure primary shipping address comes first

erpnext/accounts/party.py
```diff
@@ -910,31 +903,23 @@ def get_partywise_advanced_payment_amount(
...
+			["Dynamic Link", "link_doctype", "=", doctype],
+			["Dynamic Link", "link_name", "=", name],
+		],
+		or_filters=[
+			["is_primary_contact", "=", 1],
+			["is_billing_contact", "=", 1],
+		],
+		pluck="name",
+		limit=1,
+		order_by="is_billing_contact DESC",
```
barredterra:
```suggestion
		order_by="is_primary_contact DESC, is_billing_contact DESC",
```

Keep old order

ruthra-kumar:
> If you don't have a primary contact, any non-primary contact used to get pulled. This was only the case for Contact, Address already worked fine.

Yup. Contact does seem to have that issue.

ruthra-kumar:
@barredterra 

Sorry about the force push from my end. Looks like you were some making some changes. If possible can you push again?

Will merge once the CI completes.

erpnext/accounts/party.py
```diff
@@ -861,22 +863,23 @@ def get_party_shipping_address(doctype, name):
...
-		"from `tabDynamic Link` dl join `tabAddress` ta on dl.parent=ta.name "
-		"where "
-		"dl.link_doctype=%s "
-		"and dl.link_name=%s "
-		"and dl.parenttype='Address' "
-		"and ifnull(ta.disabled, 0) = 0 and"
-		"(ta.address_type='Shipping' or ta.is_shipping_address=1) "
-		"order by ta.is_shipping_address desc, ta.address_type desc limit 1",
-		(doctype, name),
+	shipping_addresses = frappe.get_list(
```
barredterra:
```suggestion
	shipping_addresses = frappe.get_all(
```

erpnext/accounts/party.py
```diff
@@ -910,31 +913,32 @@ def get_partywise_advanced_payment_amount(
...
-			FROM `tabDynamic Link` dl
-			INNER JOIN `tabContact` c ON c.name = dl.parent
-			WHERE
-				dl.link_doctype=%s AND
-				dl.link_name=%s AND
-				dl.parenttype = 'Contact'
-			ORDER BY is_primary_contact DESC, is_billing_contact DESC
-		""",
-		(doctype, name),
+	contacts = frappe.get_list(
```
barredterra:
```suggestion
	contacts = frappe.get_all(
```

barredterra:
@ruthra-kumar everything done

PR #35255, opened 2023-05-11
fix: depreciation schedule for existing assets [v14]

anandbaburajan:
The depreciation schedule generated for existing assets was incorrect, so fixed that. Fixes https://github.com/frappe/erpnext/issues/34920 (more details here).


Olawale1:
Thanks a lot for this @AnandBaburajan 

Hope it gets merged soon so we can test

Thanks again

PR #35256, opened 2023-05-11
fix: depreciation schedule for existing assets [dev]

### LINKED ISSUE #34920: Asset Depreciation Schedule is Wrong!

anandbaburajan:
The depreciation schedule generated for existing assets was incorrect, so fixed that. Fixes https://github.com/frappe/erpnext/issues/34920 (more details here).


deepeshgarg007:
@AnandBaburajan Testcase failing, please fix that as well. The fix looks okay

anandbaburajan:
@deepeshgarg007, seems like there are two cases for existing assets. Earlier we handled the first one (demonstrated in our test case) but we didn't handle the second one (demonstrated in https://github.com/frappe/erpnext/issues/34920). This PR right now handles the second one but not the first one. I'm trying to find the difference between the two cases and how to handle both.

PR #35278, opened 2023-05-12
fix: Update de.csv

wojosc:
added many fixes on the base of 'No' which is often wrongly translated to 'Kein'.

Also removed translation of Naming Seris like ACC-INV-.YYYY.- to ACC-INV-.YYYY.-

this can be applied for v13 and v14


deepeshgarg007:
@barredterra Can you check this once?

barredterra (approved):

PR #35311, opened 2023-05-15
fix: update workstation hour rate when workstation change in job card

Vishnu7025:
fix: update workstation hour rate when workstation change in job card


Vishnu7025:
backport not working for the develop branch.

@rohitwaghchaure

PR #35330, opened 2023-05-16
refactor: cr/dr note should be standalone even when created from another invoice

### LINKED ISSUE #35343: Credit/Debit Note should be available for Reconciilation by default
When a Credit/Debit note is created from a parent invoice, it is directly linked to it and cannot be unlinked. This means if there are Payment Entries or Journals agianst that same invoice, Cr/Dr note cannot be used for clearing future invoice from same customer.

Expectation: 
Cr/Dr notes should be available for reconciliation even if it has a parent. To achieve this Cr/Dr Notes will post Ledger entries against itself.

ruthra-kumar:
Credit/Debit Notes should post ledger against itself(standalone), even if they are created against a parent invoice. This way, they can be reconciled against a different invoice from same customer/supplier.

### DIFF
--- a/erpnext/accounts/doctype/payment_entry/test_payment_entry.py
+++ b/erpnext/accounts/doctype/payment_entry/test_payment_entry.py
@@ -702,7 +702,50 @@ class TestPaymentEntry(FrappeTestCase):
 		pe2.submit()
 
 		# create return entry against si1
-		create_sales_invoice(is_return=1, return_against=si1.name, qty=-1)
+		cr_note = create_sales_invoice(is_return=1, return_against=si1.name, qty=-1)
+		si1_outstanding = frappe.db.get_value("Sales Invoice", si1.name, "outstanding_amount")
+
+		# create JE(credit note) manually against si1 and cr_note
+		je = frappe.get_doc(
+			{
+				"doctype": "Journal Entry",
+				"company": si1.company,
+				"voucher_type": "Credit Note",
+				"posting_date": nowdate(),
+			}
+		)
+		je.append(
+			"accounts",
+			{
+				"account": si1.debit_to,
+				"party_type": "Customer",
+				"party": si1.customer,
+				"debit": 0,
+				"credit": 100,
+				"debit_in_account_currency": 0,
+				"credit_in_account_currency": 100,
+				"reference_type": si1.doctype,
+				"reference_name": si1.name,
+				"cost_center": si1.items[0].cost_center,
+			},
+		)
+		je.append(
+			"accounts",
+			{
+				"account": cr_note.debit_to,
+				"party_type": "Customer",
+				"party": cr_note.customer,
+				"debit": 100,
+				"credit": 0,
+				"debit_in_account_currency": 100,
+				"credit_in_account_currency": 0,
+				"reference_type": cr_note.doctype,
+				"reference_name": cr_note.name,
+				"cost_center": cr_note.items[0].cost_center,
+			},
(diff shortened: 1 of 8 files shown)


barredterra:
Hi Ruthra, I wanted to review this and here's what I did:

1. Create a Credit Note against an existing, overdue **Sales Invoice**
2. Create a new **Sales Invoice**

I observed two things:

- The old **Sales Invoice**, that was returned, didn't get marked as "Credit Note Issued". Is this on purpose?
- The Credit Note did not become available for reconciliation with the new invoice. Neither pre-submit via "Fetch advance payments" nor post-submit via **Payment Reconciliation**.

ruthra-kumar:
@barredterra Thanks for reviewing this.

Due to other priority tasks, this change has been in the backburner. Will revist it probably by this weekend.

> * The old **Sales Invoice**, that was returned, didn't get marked as "Credit Note Issued". Is this on purpose?

No. But, I think this can addressed. Most probably some refactor required in the `set_status` method.

> * The Credit Note did not become available for reconciliation with the new invoice. Neither pre-submit via "Fetch advance payments" nor post-submit via **Payment Reconciliation**.

I did notice this when I initially made the change. Payment Reconciliation tool filters Cr/Dr Notes that have its `return_against` field set. So, even though ledger entries are posted against itself, the tool ignore them. 
https://github.com/frappe/erpnext/blob/b4db5e9561abc6cfae1797bfb9e657ab3253ecd2/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py#L176-L178

barredterra:
Linking this discussion her, for reference: https://github.com/frappe/erpnext/issues/36170

dj12djdjs:
@ruthra-kumar I think https://github.com/frappe/erpnext/pull/36049 needs to ship with this PR. Without it, the negative outstanding won't be reconcilable via Payment Reconciliation

ruthra-kumar:
PR is in working state now.

**Changes to existing design:**

1. Credit/Debit Notes by default will be available in Payment Reconciliation tool. 
2. Since Credit/Debit notes are posting ledger entries against itself, they will have -ve outstanding and not the parent invoice.

ruthra-kumar:
> I think https://github.com/frappe/erpnext/pull/36049 needs to ship with this PR. Without it, the negative outstanding won't be reconcilable via Payment Reconciliation


@dj12djdjs Due to 
> Since Credit/Debit notes are posting ledger entries against itself, they will have -ve outstanding and not the parent invoice.

Invoice will not have -ve outstanding and will not be reported by reconciliation tool. Rather the Cr/Dr note will show up there.
So, no need for https://github.com/frappe/erpnext/pull/36049

ruthra-kumar:
@barredterra 

> The old Sales Invoice, that was returned, didn't get marked as "Credit Note Issued". Is this on purpose?

With this change, we might be breaking the requirement for this status. An Invoice requires,

1. `outstanding_amount` to be -ve
2. `return_against` of a Cr note to reference this
3. It itself should not be a Return.

to get `Credit Note Issued` status.

Due to,
> Since Credit/Debit notes are posting ledger entries against itself, they will have -ve outstanding and not the parent invoice.

[1] will not be met, so `Credit Note Issued` will never be set.

We can simply remove [1] from the requirement, then it breaks the users expection that if Credit note is issued, then outstanding will be -ve. But, i think this is unavoidable and should be expected going forward.

Thoughts?

barredterra:
Makes sense 👍

PR #35335, opened 2023-05-17
fix: tds incorrectly calculated for invoice that are below threshold

ruthra-kumar:
Two purchase invoices for the same supplier, using different tax withholding categories will have this issue.

Consider 2 tax withholding categories like below, 

| Category | single | cumulative |
|-|-|-|
| cat1     |    100 |        500 |
| cat2     |   1000 |       5000 |

1. PINV1 of net total: 105/- uses cat1. TDS is calculated as it breached single threshold
2. PINV2 of net total: 200/- uses cat2. TDS incorrectly calculated as PINV1 already has TDS calculated and `consider_party_ledger_amount` is enabled.


ruthra-kumar:

ruthra-kumar:

PR #35365, opened 2023-05-19
feat: provision to make reposting entries from Stock and Account Value Comparison Report

rohitwaghchaure:
### Provision to create reposting entries from Stock and Account Value 

Screenshot 2023-05-19 at 6 54 50 PM

###  On Click of button user will get below Message

Screenshot 2023-05-19 at 6 42 45 PM

### If period has closed, then system won't allow you to make Reposting Entries
Screenshot 2023-05-19 at 6 47 09 PM

### On Success 
Screenshot 2023-05-19 at 6 47 37 PM

#no-docs


s-aga-r:
Backporting this to v13 as well

Issue #35373, opened 2023-05-22
Party Name along with Party Number (Code) in Reports.


hpotabatti:
In most of the report currently, the Party Code is being displayed.
As most of the businesses and industries use Party Codes to keep and maintain records but would still prefer to see the Party Names to identify them in reports, it would be very beneficial to display the Party Names along with Party codes in the reports throughout the system
Eg: in below General Ledger report, it becomes very difficult to identify the exact customer/party based on the Code mentioned in the report.
![image](https://github.com/frappe/erpnext/assets/92014378/43ebce45-e8cf-409f-af0a-a1ccfc14aef6)

Would really help if a Name is also displayed along with the code.

PR #35386, opened 2023-05-23
feat(report):added party name along with party number (code)

precihole:
#35373

### DIFF
--- a/erpnext/accounts/report/general_ledger/general_ledger.py
+++ b/erpnext/accounts/report/general_ledger/general_ledger.py
@@ -203,6 +203,12 @@ def get_gl_entries(filters, accounting_dimensions):
 		as_dict=1,
 	)
 
+	# fetch party name using frappe.db.get_value
+	for gl_entry in gl_entries:
+    	party_type = gl_entry.party_type
+    	if party_type:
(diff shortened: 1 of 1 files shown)


erpnext/accounts/report/general_ledger/general_ledger.py
```diff
@@ -203,6 +203,12 @@ def get_gl_entries(filters, accounting_dimensions):
 		as_dict=1,
 	)
 
+	# fetch party name using frappe.db.get_value
+	for gl_entry in gl_entries:
+    	party_type = gl_entry.party_type
+    	if party_type:
+        	gl_entry.party_name = frappe.db.get_value(party_type, gl_entry.party, f"{party_type.lower()}_name")
```
deepeshgarg007, line 210: @precihole Can you please add a screenshot of how the report looks post the change? Also, a query is not recommended in for loop, you can get the details in a single query and make a map and use that in the report.

For customers not using naming series for customers and suppliers the same info will be shown twice, so this scenario will also have to be handled

deepeshgarg007:
@precihole closing this PR. Please reopen once the requested changes are added

PR #35409, opened 2023-05-24
refactor: Workspace cleanup

nabinhait:
- Changed the order of workspaces based on the importance
- Made frappe's default Settings and Integrations workspaces hidden and merged links into ERPNext Settings and ERPNext Integrations respectively
- Deleted Retail and Utilities workspaces

Depends on [#frappe/frappe/21100](https://github.com/frappe/frappe/pull/21100)


erpnext/patches/v14_0/cleanup_workspaces.py
```diff
@@ -0,0 +1,9 @@
+import frappe
+
+
+def execute():
+	for ws in ["Retail", "Utilities"]:
+		frappe.delete_doc_if_exists("Workspace", ws)
+
+	for ws in ["Integration", "Settings"]:
+		frappe.db.set_value("Workspace", ws, "public", 0)
```
ankush, line 9: Mark it hidden instead?

erpnext/patches/v14_0/cleanup_workspaces.py
```diff
@@ -0,0 +1,9 @@
+import frappe
+
+
+def execute():
+	for ws in ["Retail", "Utilities"]:
+		frappe.delete_doc_if_exists("Workspace", ws)
+
+	for ws in ["Integration", "Settings"]:
```
ankush:
```suggestion
	for ws in ["Integrations", "Settings"]:
```

PR #35415, opened 2023-05-25
fix: Wrong Company address in multi-company Sales Order/Invoice

### LINKED ISSUE #34837: Multi Company - Wrong company address in documents
### Information about bug

Steps to replicate:
1. Create multiple companies
2. Create a new Sales Order/Invoice
3. Select the party
4. Change the company

The company address remains that of the default company. If we change the order of steps and first change the company and then select the party, it changes the address. But if we change the company after we have already set the party/customer, it doesn't.

Issues cause by this? 
1. Wrong company address in invoices, obv
2. E-Invoice and E-Way Bill (India specific) are generated based on company address set in the doctype. This issue causes E-Invoices and E-Way Bills of Company B to be generated from portal of Company A. Messing up the whole taxation.

Temporary resolution:
Triggered(`frm.trigger`) customer event in Sales Order/Invoice on company change.
Added a server side validation for address check.

@deepeshgarg007

### Module

accounts, buying, selling

### Version

Latest

### Installation method

FrappeCloud

_No response_

GursheenK:
Added the client-side script for updating the company address correctly whenever the **Company** field changes in Sales Order/Invoices. 

The company address gets updated irrespective of the order of steps followed for changing - 

- Party/Customer
- Company

Resolves #34837

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.json
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.json
@@ -157,7 +157,7 @@
   "party_account_currency",
   "inter_company_order_reference",
   "is_old_subcontracting_flow",
-  "dashboard"
+  "connections_tab"
  ],
  "fields": [
   {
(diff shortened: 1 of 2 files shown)


rtdany10:
@GursheenK why the changes in purchase order?

```
company: function(frm) {
    frm.trigger("customer");
}
```
is all the code change required imo.

Issue #32735, opened 2022-10-28
Stock, Manufacturing and Buying Reports Refactor


s-aga-r:
- [x] `frappe.db.sql` --> `frappe.qb`
- [x] Fix broken reports and filters
- [ ] Add missing test cases.

PR #35437, opened 2023-05-27
fix: incorrect `POS Reserved Qty` in `Stock Projected Qty` Report

s-aga-r:
**Source / Ref:** ISS-23-24-01015

**Issue:** The `Reserved for POS Transactions` shows incorrect qty in the `Stock Projected Qty` report. The `Is Return` filter is missing in the `get_pos_reserved_qty` function.

**Steps to Replicate:**
- Create POS Invoice.
- Create Consolidated Sales Invoices.
- Create a Return Against POS Invoice.
- Check the `Stock Projected Qty` report for the respective Item and Warehouse.

https://github.com/frappe/erpnext/issues/32735


vishdha:
https://github.com/frappe/erpnext/pull/34711

@s-aga-r this change breaking multi UOM structure.
Currently it uses the qty field which doesn't represent UOM's conversion factor, making the available stock incorrect.

erpnext/accounts/doctype/pos_invoice/pos_invoice.py
```diff
@@ -674,18 +675,22 @@ def get_bin_qty(item_code, warehouse):
...
-		(item_code, warehouse),
-		as_dict=1,
-	)
+	p_inv = frappe.qb.DocType("POS Invoice")
+	p_item = frappe.qb.DocType("POS Invoice Item")
+
+	reserved_qty = (
+		frappe.qb.from_(p_inv)
+		.from_(p_item)
+		.select(Sum(p_item.qty).as_("qty"))
```
vishdha, line 684: it should be stock_qty otherwise.
in warehouse have .5 kgs material of UNT of 100g. 5 qty UNT = .5 kgs but above change will reserved 5kgs which is blocking transaction for available stocks

PR #35459, opened 2023-05-29
fix: retention stock entry: grab conversion factor from source

marc-dll:
When creating new retention stock entry, by default, the uom/stock_uom conversion factor is set to 1 whatever the UOMs


marc-dll:
https://github.com/Mergifyio backport version-13-hotfix

marc-dll:
V13 and V14 backports would be nice :) Thanks

s-aga-r (approved):

PR #35465, opened 2023-05-30
fix: rate not fetching properly for inter transfer purchase order

rohitwaghchaure:
If the supplier is internal supplier and user is creating the transfer entry within a same company of two different branches then only fetch valuation rate else get the rate from the item price.

[Refer code](https://github.com/frappe/erpnext/blob/develop/erpnext/controllers/accounts_controller.py#L1868)


ggraza:
Its suggestion to add check box on account doctype for Enable / Disable Intercompany valuation rate

Facing same issue,

ERPNext: v14.8.0 (version-14)
Frappe Framework: v14.16.0 (version-14)
Payments: v0.0.1 (develop)

PR #35503, opened 2023-05-31
fix(ux): throw if no row selected to create repost entries

s-aga-r:
**Source / Ref:** ISS-23-24-01158

[Screencast from 2023-05-31 15-22-51.webm](https://github.com/frappe/erpnext/assets/63660334/8a59a7bc-9e84-4285-a984-a6ef7234e519)


s-aga-r:
https://github.com/frappe/erpnext/pull/35365

PR #35522, opened 2023-06-01
refactor!: removing loan management module

anandbaburajan:
Loan Management module is going to be a new app: https://github.com/frappe/lending. It would still depend on ERPNext.


erpnext/hooks.py
```diff
@@ -521,11 +515,22 @@
 	"Account Closing Balance",
 ]
 
-# get matching queries for Bank Reconciliation
 get_matching_queries = (
 	"erpnext.accounts.doctype.bank_reconciliation_tool.bank_reconciliation_tool.get_matching_queries"
 )
 
+get_matching_vouchers_for_bank_reconciliation = "erpnext.accounts.doctype.bank_reconciliation_tool.bank_reconciliation_tool.get_matching_vouchers_for_bank_reconciliation"
```
marination, line 522: @anandbaburajan Why are there 2 ways to plug in custom queries from other apps ? i.e. `get_matching_queries` and `get_matching_vouchers_for_bank_reconciliation`.
Earlier `get_matching_queries`  was used to plug in constructed custom queries, it just returned queries to be run.

Now that exists with `get_matching_vouchers_for_bank_reconciliation` which returns results of custom queries and appends them to the result in erpnext.

So now we can use `get_matching_queries` OR `get_matching_vouchers_for_bank_reconciliation` depending on if we want to return a query or results I guess. But this feels kinda confusing and there isn't any uniformity and also I'm curious about the motivation behind this.

anandbaburajan: @marination I agree it's confusing. About the motivation, it was actually a short-term fix for this problem: the code around `get_matching_queries` handles only raw sql queries and not qb ones, and loan related queries are written in qb and rest of the queries are still in raw sql. So the long-term solution would be to convert the rest to qb.

marination: @anandbaburajan Ah okay. Well you can return qb queries as a string without `.run()` and they can be used in `frappe.db.sql` 😄. I'll do the `sql -> qb` conversion anyway in https://github.com/frappe/erpnext/pull/36893 and make changes here too!

PR #35523, opened 2023-06-01
refactor!: Remove custom cashflow report mapper

deepeshgarg007:
Deprecating Custom Cash Flow Mapper as it has bugs and isn't that customizable as per user's requirements.
Will be good if this is redone from scratch is there a need for this


muchai:
There's a use for it, albeit rare.
We have one use case, need. The bugs you mention prevented us making it work for them..

PR #35536, opened 2023-06-02
fix: Ignore permissions while submitting account closing balance record

nabinhait:
Ignore permissions while submitting the account closing balance record


ruthra-kumar:

Issue #34282, opened 2023-03-02
Record advance payment as a liability


barredterra:
It is possible to fetch advances received into a **Sales Invoice**, reducing the payable amount. For this to work, the incoming advance payment currently has to be recorded as a negative entry on the receivable account. However, the advance payment is a liability for the receiving company, as long as the goods/services are not delivered. Recording the liability as a negative receivable violates the offsetting prohibition:

> IFRS: Offsetting of assets and liabilities is only allowed under restrictive conditions.
> German GAAP: Offsetting assets and liabilities is only allowed where an entity has a legally enforceable right to offset.
– https://www.pwc.com/gx/en/ifrs-reporting/pdf/ifrs-vs-german-gaap-similarities-and-differences_final2.pdf (page 8)

### Proposed solution

Advance payments need to be recorded on a separate liability account ("Advances Received"). In the invoice, we can fetch advance payments from this account. On submission of the invoice, the advance payment should be cleared and only the remaining amount left on the "receivable" account.

#### Sales Order

Sales Order over 200 net total plus 19% VAT equals 238 grand total

#### Advance Payment Entry

Payment of 50% in advance, Payment Entry against Sales Order. VAT is due on receipt of the payment.

GL Entries:

Type | Amount | Account | Root Type
-----|----:|----|---
Debit | 119 | Bank | Asset
Credit | 100 | Advances Received | Liability
Credit | 19 | VAT | Liability

So at this point in time we should have a total liability of 119. (Currently, ERPNext would show only show that our _receivables_ are reduced by 119, which is not correct.)

#### Sales Invoice

Sales Invoice over 200 net total plus 19% VAT equals 238 grand total.

Deducting 119 advances received. And 119 remain outstanding.

GL Entries:

Type | Amount | Account | Root Type
-----|----:|----|---
Credit | 200 | Revenue | Income
Credit | 19 | VAT | Liability
Debit | 119 | Receivables | Asset
Debit | 100 | Advances Received | Liability

> Disclaimer: I'm not an accountant and in english it's even harder – the GL Entries might be total nonsense and are only for illustrative purposes.

---

Similar, but not 100% identical issue: https://blog.prolecto.com/2016/07/08/solved-netsuite-customer-deposits-and-advance-payment-accounting/

szufisher:
it is a long await feature, it should also apply to purchase. accounts payable and accounts receivable reports needed to be adapted accordingly.

getroommanagement:
[...]

PR #35537, opened 2023-06-02
feat: Record Advance Payments as Liability

GursheenK:
Ability to let users record advance payments as a liability instead of a negative asset.

The current flow allows the user to record **Advance Payments** by storing it as a **Negative Asset** in General Ledger. It uses the Default Receivable account from the Company master to achieve this. 

Added the option to allow users to **Book Advance Payments as Liability**. When selected, the user needs to provide the Default Advances Received Account and the Default Advances Paid Account in the Company master.

Any Advance Payment Entries will therefore use the liability account. Any new Sales/Purchase Invoices that in turn fetch the Advance Payment Entries show the account used for the entry. 

Similar behavior for any Payment Reconciliations.

The Default Advances Received / Default Advances Paid accounts can be set from the following Doctypes for multiple companies - 

- Customer / Customer Group
- Supplier / Supplier Group

If they are not set in any of the above, the default liability account from the Company master is used.

Issue #34282

### DIFF
--- a/erpnext/accounts/doctype/party_account/party_account.json
+++ b/erpnext/accounts/doctype/party_account/party_account.json
@@ -6,7 +6,9 @@
  "engine": "InnoDB",
  "field_order": [
   "company",
-  "account"
+  "account",
+  "advances_received_account",
+  "advances_paid_account"
  ],
  "fields": [
   {
@@ -22,14 +24,26 @@
    "fieldname": "account",
    "fieldtype": "Link",
    "in_list_view": 1,
-   "label": "Account",
+   "label": "Default Account",
+   "options": "Account"
+  },
+  {
+   "fieldname": "advances_received_account",
+   "fieldtype": "Link",
+   "label": "Advances Received Account",
+   "options": "Account"
+  },
+  {
+   "fieldname": "advances_paid_account",
+   "fieldtype": "Link",
+   "label": "Advances Paid Account",
    "options": "Account"
   }
  ],
  "index_web_pages_for_search": 1,
  "istable": 1,
  "links": [],
- "modified": "2022-04-04 12:31:02.994197",
+ "modified": "2023-06-02 13:00:06.885744",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Party Account",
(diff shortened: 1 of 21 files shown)


mujeerhashmi:
Can this be back ported to version-14 ?

erpnext/accounts/doctype/party_account/party_account.json
```diff
@@ -6,7 +6,9 @@
  "engine": "InnoDB",
  "field_order": [
   "company",
-  "account"
+  "account",
+  "advances_received_account",
```
deepeshgarg007, line 10: The way `account` column is used for both parties like customer and supplier, maybe the same column `advance_account` can be used for both customer and supplier, with no need for separate columns (fields). Just a filter for respective account type in the masters would be enough

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -318,6 +346,10 @@ frappe.ui.form.on('Payment Entry', {
 		}
 	},
 
+	company: function(frm){
+		frm.trigger('party');
+	},
+
```
deepeshgarg007, line 352: Looks like a change from other PR, remove this

erpnext/controllers/accounts_controller.py
```diff
@@ -2136,45 +2144,41 @@ def get_advance_journal_entries(
...
+			(journal_acc.exchange_rate)
 		)
+		.where(journal_acc.account.isin(party_account)
+	 	& (journal_acc.party_type == party_type)
+	 	& (journal_acc.party == party)
+	 	& (journal_acc.is_advance == 'Yes')
+	 	& (journal_entry.docstatus == 1)
+	 	)
+    )
+	if party_type == "Customer":
```
deepeshgarg007, line 2168: Else condition missing

deepeshgarg007 (changes requested):

erpnext/controllers/accounts_controller.py
```diff
@@ -871,24 +873,30 @@ def set_advances(self):
...
 
 	def get_advance_entries(self, include_unallocated=True):
 		if self.doctype == "Sales Invoice":
-			party_account = self.debit_to
 			party_type = "Customer"
 			party = self.customer
 			amount_field = "credit_in_account_currency"
 			order_field = "sales_order"
 			order_doctype = "Sales Order"
+			party_account = frappe.db.get_values("Company", {"company_name": self.company}, ["default_receivable_account", "default_advances_received_account"])
```
deepeshgarg007, line 890: This can also be a different one setup in the customer/supplier master. Right now we are just considering from the company

PR #35540, opened 2023-06-02
fix: column formatting in Bank Reconciliation Tool

barredterra:
- format "Date" column
- format "Party" column as (dynamic) link
- fix closing tag and quoting in "Actions" button

Bildschirmfoto 2023-06-02 um 16 15 38


marination:
Translate the 'Get Unreconciled Entries' text in `change_custom_button_type`. There is no primary action when the system is in german

marination (commented):
Translate the 'Get Unreconciled Entries' text in change_custom_button_type. There is no primary action when the system is in german

barredterra:
@marination that seems to be unrelated to this PR, but I fixed it here: https://github.com/frappe/erpnext/pull/35564

marination (approved):

PR #35549, opened 2023-06-03
fix: set Phone option on contact_{mobile,phone} for SMS notification

blaggacao:
Only fields with options 'Phone' are selectable as recepients_by_field
in the notification doctype.

This fix sets sane defaults across all doctypes that have
contact_{mobile,phone} fields

This is a side effect of [this frappe implementation](https://github.com/frappe/frappe/blob/26722b1a1cd63a456e23406c6abe3acfee12b406/frappe/email/doctype/notification/notification.js#L69-L73)


blaggacao:
image

blaggacao:
@deepeshgarg007 Would you mind whitelisting the workflow runs?

blaggacao:
Thanks for merging!

PR #35554, opened 2023-06-04
feat: added method for getting ordered quantity in the  supplier scorecard variable.

Vishnu7025:
feat: added method for getting ordered quantity in the supplier scorecard variable.
@ankush


s-aga-r (changes requested):
@Vishnu7025 Use [Frappe Query Builder](https://frappeframework.com/docs/v14/user/en/api/query-builder) instead of raw sql queries. May I know why you have added this function and from where it gets called?

PR #35559, opened 2023-06-05
feat: Purchase Receipt Planning

kittiu:
This module add new feature, Purchase Receipt Planning, which normally useful if for i.e., construction where we normally pay by planned installment.

Design
* On Purchase Order, add new checkbox "Has Receipt Plan"
* If checked, user must plan receipt portions which should sum to 100%
* Validate Receipt Plan on Submit and after Submit.
* Create Purchase Receipt will now has dialog for user to choose the Installment (which has not been used).

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.js
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.js
@@ -219,7 +219,13 @@ erpnext.buying.PurchaseOrderController = class PurchaseOrderController extends e
 			if(doc.status != "Closed") {
 				if (doc.status != "On Hold") {
 					if(flt(doc.per_received) < 100 && allow_receipt) {
-						cur_frm.add_custom_button(__('Purchase Receipt'), this.make_purchase_receipt, __('Create'));
+						if (doc.has_receipt_plan) {
+							cur_frm.add_custom_button(__('Purchase Receipt'), function() {
+								me.make_purchase_receipt_by_plan();
+							}, __('Create'));
+						} else {
+							cur_frm.add_custom_button(__('Purchase Receipt'), this.make_purchase_receipt, __('Create'));
+						}
 						if (doc.is_subcontracted) {
 							if (doc.is_old_subcontracting_flow) {
 								if (me.has_unsupplied_items()) {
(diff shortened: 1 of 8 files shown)


kittiu:
image

image

image

image

kittiu:
Note: I was trying to make this into develop branch, but I just can't install the develop branch in my local. Not sure it broken or not.

cc @umairsy

kittiu:
@rohitwaghchaure @s-aga-r any comment with this PR
or @NagariaHussain

rohitwaghchaure:
@kittiu 

Upon clicking 'Create Purchase Receipt,' the modal opens with options such as 'Installment 1,' 'Installment 2,' and 'Installment 3.' If a user selects 'Installment 3' before completing 'Installment 1' and 'Installment 2,' the system currently allows them to proceed. Should the system force the user to create the purchase receipts sequentially? Additionally, should the system allow the creation of purchase receipts every time the user selects the 'Installment 3' option?

rohitwaghchaure:
> Upon clicking 'Create Purchase Receipt,' the modal opens with options such as 'Installment 1,' 'Installment 2,' and 'Installment 3.' If a user selects 'Installment 3' before completing 'Installment 1' and 'Installment 2,' the system currently allows them to proceed. Should the system force the user to create the purchase receipts sequentially? Additionally, should the system allow the creation of purchase receipts every time the user selects the 'Installment 3' option?

@kittiu Can you please reply on this

kittiu:
@rohitwaghchaure sorry, I miss manage my email. 

> Should the system force the user to create the purchase receipts sequentially?

Yes this is my intention, because I used to have cases when we agree with supplier to change order. But you point is valid in case we prevent user error also. What about, we lock the order as you mentioned but allow changing the order of Receipt Plan instead? This may cover both control and flexibility.

>Additionally, should the system allow the creation of purchase receipts every time the user selects the 'Installment 3' option?

I don't quite get this, can you elaborate?

Thanks.

rohitwaghchaure:
@kittiu We should lock the receipt plan line row once the receipt has created and user shouldn't be able to create the another purchase receipt against the already completed receipt plan.

Also instead of Dropdown in the modal, you can show the child table of receipt plan in the modal while creating the purchase receipt .

kittiu:
> @kittiu We should lock the receipt plan line row once the receipt has created and user shouldn't be able to create the another purchase receipt against the already completed receipt plan.

Yes

> Also instead of Dropdown in the modal, you can show the child table of receipt plan in the modal while creating the purchase receipt .

I don't quite get this, why should we show the whole table instead of dropdown, even only pending plan. Imagine, I have 20 installment, although I already have don't 2 installment, should I show the 18 installment remaining (without slect which installment to use, as we will use the next one anyway).

May be I don't get the picture you want to say. Are there some sample in system now I can see what you mean.

rohitwaghchaure:
> May be I don't get the picture you want to say. Are there some sample in system now I can see what you mean.

243297033-8aa120d6-3f15-481b-b656-c3b023e2414a

In the image above, you are displaying installments in the dropdown. Instead of this, could you please include a table with installment information, similar to how it is shown in the picture below?

image

kittiu:
> > May be I don't get the picture you want to say. Are there some sample in system now I can see what you mean.
> 
> 243297033-8aa120d6-3f15-481b-b656-c3b023e2414a
> 
> In the image above, you are displaying installments in the dropdown. Instead of this, could you please include a table with installment information, similar to how it is shown in the picture below?
> 
> image

Seem like a good idea to have this as table, if user can select multiple purchase and create them at once, show warning if selection is not in order, i.e., choose 2 (without 1),  choose 1 and 3 (without 2)....

WDYT?

rohitwaghchaure:
> WDYT?

Yes correct, please do changes.

kittiu:
@rohitwaghchaure not sure why this is auto closed. Can you reopen it?

Now, I have made the dialog as you mentioned earlier. With some changes,
* Do not allow select the installment that was already Done.
* Do not allow to select multiple, as it will need server side creation which is different from existing (we might do it in the future, but not for now).
* Do not require sequential selection, I think it can be too rigid.

Here is some screen shot,

image

Thanks!

PR #35565, opened 2023-06-05
When creating land cost voucher from PI or PR, the system takes base rate instead of base net rate, the problem is if we apply discount to the PI/PR then at land cost voucher the discounted rate does not shows it shows base rate and base net rate

kunhimohamed:
…mount

> Screenshots/GIFs


barredterra:
The title is not the description 😁

PR #35567, opened 2023-06-06
fix(ux): serial and batch bundle status

s-aga-r:
Before:

image

image

After:
image

image

### DIFF
--- a/erpnext/stock/doctype/serial_and_batch_bundle/serial_and_batch_bundle_list.js
+++ b/erpnext/stock/doctype/serial_and_batch_bundle/serial_and_batch_bundle_list.js
@@ -1,0 +1,11 @@
+// Copyright (c) 2023, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.listview_settings["Serial and Batch Bundle"] = {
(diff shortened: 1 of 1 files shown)


erpnext/stock/doctype/serial_and_batch_bundle/serial_and_batch_bundle_list.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2023, Frappe Technologies Pvt. Ltd. and contributors
+// For license information, please see license.txt
+
+frappe.listview_settings["Serial and Batch Bundle"] = {
+	add_fields: ["is_cancelled"],
```
ankush, line 5: @rohitwaghchaure @s-aga-r 

In new doctypes we can just follow our current conventions. Where `docstatus == 2` is considered cancelled. 

No need to have this separate field like GLE/SLE.

rohitwaghchaure: @ankush Yes but the Serial and Batch Bundle should not be submitted / cancelled by the user. In this case on cancel of the document through code we have to cancel the respective Serial and Batch Bundle.

PR #35594, opened 2023-06-07
feat: added support for mandatory dimensions per account without applying restrictions on dimension values

nikkothari22:
In ERPNext, we currently have the ability to set certain accounting dimensions mandatory and allow/restrict certain values only for specific accounts using [Accounting Dimension Filter](https://docs.erpnext.com/docs/v12/user/manual/en/accounts/accounting-dimension-filter).

Once these restrictions are applied, ERPNext:
1) filters link fields for the dimension throughout the application accordingly. 
2) validates GL and Payment Ledger Entry submissions by either throwing an "MandatoryAccountDimensionError" or an "InvalidAccountDimensionError". 

However, in some cases, we might only want to make dimensions mandatory for specific accounts, without restricting the values. So, for example, we might want "Cost Center" to be mandatory when entries are made against the "Plants and Machineries" account, but we do not want to restrict the entry to only specific projects.

This pull request aims to solve that problem by doing the following:

1. Adding a checkbox "Apply restriction on dimension values" in "Accounting Dimension Filter". If this is not checked, we should only validate for the mandatory dimensions and not restrict the values for said dimension. If this is checked, the feature just works as earlier.

image

2. The "Applicable Dimensions" table is not mandatory by default, and only mandatory if the checkbox above is checked. 

image

3. If the checkbox is unchecked, we set "allow_or_restrict" to "Restrict" and clear out the applicable dimensions table. This check is also added on the backend (`before_save`). This would work since we are using the same feature as before - but with "Restrict" to nothing. 

4. The SQL query that generates the `dimension_filter_map` had to be changed to a LEFT JOIN so that we fetch rows without a dimension value as well.

Hence, after applying these changes, the system supports making accounting dimensions mandatory for specific accounts and applying restrictions if needed.

#### Case 1: Only make dimension mandatory for certain accounts.

Accounting Dimension Filter: 
image

Query filter shows all options when making Journal Entry:
image

Error thrown when trying to submit the Journal Entry without adding in required dimension:
image

#### Case 2: Make dimension mandatory and only allow certain values

Accounting Dimension Filter: 
image

Query filter only shows allowed options (in this case "Main - TCC" and not "Test - TCC") when making Journal Entry:
image

Attempting to enter in value that's not allowed throws an error:
image
image

#### Case 3: Make dimension mandatory and do not allow certain values

Accounting Dimension Filter:
image

Query filter does not show restricted values (in this case "Main - TCC") when making Journal Entry:
image

Attempting to enter in value that has been restricted throws an error on submit:
image

image

`no-docs`


nikkothari22:
Hey guys, 

Any update on this?

deepeshgarg007:
@nikkothari22 Merging this in the `develop` branch. Can be backported to v14 once the docs are updated

nikkothari22:
@deepeshgarg007 Thanks!

Where do I update the docs?

nikkothari22:
Hey @deepeshgarg007 - do I just update the docs in the ERPNext Wiki directly and ask for a review?

PR #35604, opened 2023-06-07
fix: reconcile invoice against credit note.

### LINKED ISSUE #35603: Cannot reconcile multi-currency Purchase Invoice against Purchase Invoice Return
### Information about bug

# Use Case
1. Purchase Goods from Supplier
2. Pay Supplier in Full
3. Return Partially Defective Goods
4. Purchase Goods from Supplier
5. Apply Defective Credit to new Purchase Invoice.

# Steps to Reproduce
1. Create a Purchase Invoice $1,000.00 @ 1.35
6. Create Purchase Invoice (Return) $100.00 @ 1.40  **Important: Not against the original invoice.**
7. Attempt to Reconcile the the Purchase Invoice (Return) against the Purchase Invoice.

image

# Cause
During the payment reconciliation process
https://github.com/frappe/erpnext/blob/0c12d4d3c5ebb7d5bd6cf70dfde026c810a7b66a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py#L330-L331
is always executed, creating a journal entry for the difference in exchange rate.

Then this is executed:
https://github.com/frappe/erpnext/blob/0c12d4d3c5ebb7d5bd6cf70dfde026c810a7b66a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py#L336-L337

The exchange gain and loss difference need's to be booked in the above transaction for the payment reconciliation to become successful.

Related: https://github.com/frappe/erpnext/issues/34680

### Module

accounts

### Version

ERPNext: v14.0.0-dev (HEAD)

Frappe Framework: v15.0.0-dev (HEAD)

### Installation method

FrappeCloud

_No response_

dj12djdjs:
closes: https://github.com/frappe/erpnext/issues/35603

### DIFF
--- a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
+++ b/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
@@ -316,6 +316,7 @@ class PaymentReconciliation(Document):
 
 		entry_list = []
 		dr_or_cr_notes = []
+		difference_entries = []
 		for row in self.get("allocation"):
 			reconciled_entry = []
 			if row.invoice_number and row.allocated_amount:
@@ -328,13 +329,15 @@ class PaymentReconciliation(Document):
 				reconciled_entry.append(payment_details)
 
 				if payment_details.difference_amount:
-					self.make_difference_entry(payment_details)
+					difference_entries.append(
+						self.make_difference_entry(payment_details, do_not_save_and_submit=bool(dr_or_cr_notes))
+					)
 
 		if entry_list:
 			reconcile_against_document(entry_list, skip_ref_details_update_for_pe)
 
 		if dr_or_cr_notes:
-			reconcile_dr_cr_note(dr_or_cr_notes, self.company)
+			reconcile_dr_cr_note(dr_or_cr_notes, difference_entries, self.company)
 
 	@frappe.whitelist()
 	def reconcile(self):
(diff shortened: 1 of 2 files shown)


ruthra-kumar:
@dj12djdjs 

This approach doesn't link the Invoice with the Exchange gain/loss amount.

The Invoice will still be missing that Exchange Gain/Loss amount as there is no reference to it. Until it is linked, the purchase invoice will have outstanding.

A separate JE of type `Exchange Gain/Loss` should be created as well.

dj12djdjs:
@ruthra-kumar I'm not sure I understand what you mean the exchange gain/loss is not linked? It is linked via the journal entry reconciling the invoice against the credit.

Example:
image


From there the invoice is now closed:
image


I believe the original code before this PR was attempting to separate the transaction across two journal entries. But I don't think it's possible as one of the journal entries will never be balanced this way.

ruthra-kumar:
> I believe the original code before this PR was attempting to separate the transaction across two journal entries. But I don't think it's possible as one of the journal entries will never be balanced this way.

Got it. Will test some cases locally and will update you.

ruthra-kumar:
LGTM

ruthra-kumar (approved):

dj12djdjs:
@ruthra-kumar Just curious about the draft. Did you find any issues with this PR?

ruthra-kumar:
@dj12djdjs Tested on few edge cases, like, allocating a payment and debit note at the same time. Looks to be working.
Thanks for the Fix.

PR #35609, opened 2023-06-08
feat: Record Advance Payment as a liability

GursheenK:
The current flow allows the user to record - 

- **Advance Payments Received** from Customers as a **Negative Asset** in the General Ledger
- **Advance Payments Paid** to Suppliers as a **Negative Liability** in the General Ledger

The **Book Advance Payments in Separate Party Account** feature introduces the option to use separate accounts for recording these Advance Payments. The option can be selected from the **Company Doctype** > **Advance Payments Section**

When selected, the user needs to provide the Default Advances Received Account and the Default Advances Paid Account in the Company master.
<br>
Screenshot 2023-11-13 at 4 10 57 PM
<br>
Default Advance Accounts can be set for a Customer/Customer Group/Supplier/ Supplier Group with respect to multiple companies from the **Accounting Tab** > **Default Accounts Table**
<br>
Screenshot 2023-11-13 at 4 14 32 PM
<br>

> Note : If they are not set in the Customer or Supplier Doctypes, then the Default Advance Account mentioned in the Company Master will be used for recording the Advance Payments of that particular party.

### **Example**

A Payment Entry for receiving `₹1000` from a customer is submitted. The General Ledger Entries created for the same will use the Liability Account instead of the Receivable Account since it is an Advance Payment. 
<br>
>Screenshot 2023-11-13 at 4 16 47 PM

<br>

Suppose a Sales Invoice with a total of `₹500` is created for the same customer and the Advance Payments are fetched and allocated using the _Get Advances Received_ button. In such a case, when the Sales Invoice is submitted, the General Ledger for the Payment Entry will reflect the following changes.

<br>

>Screenshot 2023-11-13 at 4 19 54 PM

<br>

Since the total allocated amount to the Sales Invoice is `₹500`, the same has been debited from the Liability Account. The Liability Account will then hold the remaining `₹500` as an Advance Payment.

<br>

`no-docs`

Issue #34282


s-aga-r:
@deepeshgarg007 @GursheenK Why two [PRs](https://github.com/frappe/erpnext/pull/35537)?

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -86,12 +86,36 @@ def validate(self):
 	def on_submit(self):
 		if self.difference_amount:
 			frappe.throw(_("Difference Amount must be zero"))
+		book_advance_payments_as_liability = frappe.get_value(
+			"Company", {"company_name": self.company}, "book_advance_payments_as_liability"
+		)
+		if book_advance_payments_as_liability:
```
deepeshgarg007: Simply call `get_liability_account` from here and check if `book_advance_payments_as_liability` is enabled or not inside the function itself

erpnext/accounts/doctype/payment_entry/payment_entry.py
```diff
@@ -86,12 +86,36 @@ def validate(self):
 	def on_submit(self):
 		if self.difference_amount:
 			frappe.throw(_("Difference Amount must be zero"))
+		book_advance_payments_as_liability = frappe.get_value(
```
deepeshgarg007: This should be on `validate`, not on submit

erpnext/accounts/doctype/payment_entry/payment_entry.js
```diff
@@ -708,7 +712,7 @@ frappe.ui.form.on('Payment Entry', {
 				if(r.message) {
 					var total_positive_outstanding = 0;
 					var total_negative_outstanding = 0;
-
+					console.log(r.message);
```
deepeshgarg007: Remove this

deepeshgarg007 (changes requested):

rtdany10:
~@deepeshgarg007 will this be backported?~
Just saw the label, but any possibility?

PR #35613, opened 2023-06-08
fix: calculate wdv depr schedule properly for existing assets [v14]

anandbaburajan:
https://github.com/frappe/erpnext/pull/35458 wasn't enough, hence this PR.


anandbaburajan:
@mergify backport version-14

PR #35626, opened 2023-06-09
fix: modify filters for account in journal entry

ljain112:

- In taxes account setup, account currency for tax account is not been set.
- Only Accounts with default currency are allowed while creating journal Entry without multi currency so tax accounts are not shown.
- User Not able to create opening entries for tax account because currency not set in tax account.
[Discuss Issue Link](https://discuss.frappe.io/t/how-to-add-gst-opening-balance-to-erpnext/105618)

Modifying query in journal entry doctype to include accounts with empty currency also.

> Screenshots/GIFs

Account Currency empty in Tax accounts.
Screenshot from 2023-06-09 18-00-55

Filter Query in Journal Entry Requiring Default currency.
Screenshot from 2023-06-09 18-11-47

### DIFF
--- a/erpnext/accounts/doctype/journal_entry/journal_entry.js
+++ b/erpnext/accounts/doctype/journal_entry/journal_entry.js
@@ -575,7 +575,7 @@ $.extend(erpnext.journal_entry, {
 		};
 		if(!frm.doc.multi_currency) {
 			$.extend(filters, {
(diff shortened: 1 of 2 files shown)


rtdany10:
Won't it be better to tweak the account filter in JE?
If selected currency(JE) is same as the default company currency, list all accounts with no currency set and also the ones with that particular currency.

PR #35629, opened 2023-06-09
fix: added process loss in job card

### LINKED ISSUE #34978: V14 Process Loss should be considered as part of qty to manufacture in Work Order
### Information about bug

New Process Loss feature has been added in Version 14 for Process Manufacturing Industry.
In this flow, if this feature is used, work order shows as below, after completing stock entry:
image

Here the quantity to manufacture as per WO was 120, but actual manufactured qty was 118 and 2 was process loss.

Now the Work Order status is still set as "Not Started" and it asks to create Stock entry for process loss qty of 2. This should not be the case as 118 is the yield of the process and 2 is the loss and as such the process is completed.

BUG LOGIC:
WO is complete if:
Qty to manufacture = Manufactured Qty
CORRECT LOGIC:
WO is complete if:
Qty to manufacture = (Manufactured Qty + Process Loss)

### Module

manufacturing

### Version

ERPNext: v14.22.0 

Frappe Framework: v14.33.0 

### Installation method

FrappeCloud

_No response_

rohitwaghchaure:
## **Issue**

If the quantity to manufacture is not same as completed qty in job card then system throw the error that "quantity to manufacture should be same as completed qty"

Screenshot 2023-06-09 at 10 38 07 PM

But what if there is process loss and qty has not produced as per plan? 

## **Solution**
To solve this problem, added provision to set process loss qty in the job card

Screenshot 2023-06-09 at 11 20 22 PM

If you're using the Routing and first operation has process loss then in the second operation you won't be able to produce more qty than first operation completed qty. If you tried produce more qty then system will throw below error 

image

System will auto capture the Process Loss (from job cards) in the Manufacture Stock Entry
Screenshot 2023-06-09 at 11 27 28 PM

After completion of 9 quantity system won't allow to make further stock entries against the work order for 1 quantity because 1 quantity was process loss. (Fixed https://github.com/frappe/erpnext/issues/34978)
Screenshot 2023-06-09 at 11 30 41 PM


precihole:
Hello, @rohitwaghchaure 

How to partially complete the job card. Like i have 10 qty in job card but i want to complete only for 5 but it set remaning qty as loss.
image

goapaints:
@rohitwaghchaure , in multi-level BOM, when the above process Item is used, the qty required should be a proportion of (quantity to manufacture - process loss), else estimation will be less. If you agree, i will raise a bug.

PR #35630, opened 2023-06-10
Fix: CSS not applied to product title

trustedcomputer:
Backport of PR #35582 in develop branch, which never got tagged for version-14-hotfix or version-13-hotfix.


trustedcomputer:
I created this PR so the changes made in my previous PR could be backported to version-14 and version-13. If this is not the proper procedure, please advise what the proper procedure is. I am new at this! =)

PR #35636, opened 2023-06-11
fix: miscellaneous

s-aga-r:
Serial and Batch Bundle:
- Add filter `disabled=0` for Batch No.
- Don't show all DocTypes in the `Voucher Type` list.
- DocType {DocType} Item not found.
  image
- AttributeError: 'NoneType' object has no attribute 'precision'
  image
- Add `is_cancelled=0` filter for SBB.
- Add filters for SBB in `Asset Capitalization, Asset Repair, Installation Note, Job Card, Pick List, Quotation, Stock Entry, Stock Reconciliation and Subcontracting Receipt`
  image


rohitwaghchaure:
@s-aga-r make fixes in transaction.js file https://github.com/frappe/erpnext/blob/develop/erpnext/public/js/controllers/transaction.js#L125-L136

PR #35638, opened 2023-06-12
idea: Reconcile Mechanism

kittiu:
**NOTE:** this is just to share idea, not a compete PR.

IMHO, Reconciliation should be generic for all Accounts (given, is_reconcile=True) and should be done at GL Entry level (nothing should related to Documents).

From user / developer perspective, it should be like,

1. Setup an Account with is_reconcile=True
2. When GL Entry created, set Residual = debit-credit account first.
    * gl1, Debit AR = 100, Residual = 100
3. When opposite GL Entry created, set Residual first
    * gl1, Credit AR = 90, Residual = -90
 4. Call new function **reconcile_gl([gl1, gl2])**, and residual amount gets updated,
    * gl1, Debit AR = 100, Residual = 10
    * gl2, Credit AR = 90, Residual = 0

The concept is, when every GL Entry is being reconciled, it will create **Partial Reconcile Entry**

image

(If document is cancelled, the Amount Reconciled will set to zero)

And then update the residual amount at the GL Entry itself

image

As you can see from the mechanism above, things can get simpler in many areas.
1. We can have Open Items report for all account codes (is_reconcile = True)
3. We can display the Residual amount from GL Entry to the front Document easily (i.e., Invoice, Expense, Journal Entry, etc.)
4. We can use this mechanism for reconciliation in many other area easily using **reconcile_gl()** function (but for GL with against_voucher field, it is already automatic)
5. I have my own case, for Undue VAT in invoice, to be reconciled into VAT on Payment.

### Partial Reconcile Entry (new doctype)
- Debit GL Entry (link to GL Entry)
- Credit GL Entry (link to GL Entry)
- Reconcile Amount (amount in company currency)
- Full Reconcile Entry (mark all related GL Entries's residual become zero, used for grouping)

### Account (existing doctype)
- Is Reconcile = True for account that can reconcile, i.e., creditor, debtor, undue tax, advance, etc. (practically on any BS accounts)

### GL Entry (existing doctype)
- Is Reconcile
- Residual
- Full Reconcile Entry (mark all related GL Entries's residual become zero, used for grouping)

### Full Reconcile Entry (new doctype)
- Name, i.e., F1, F2, ...

### Tested for Cases

* 1 Invoice -> 1st Payment (full)
* 1 Invoice -> 1st payment (partial) -> 2nd payment (now full)
* 2 Invoices -> 1st Payment (full)
* 2 Invoices -> Arbitrary amount on payments
* Payment Reconcile (match invoice with payment allocation)
* Expense -> Payment
* Cancel cases
* Invoice / Credit Note

### Issues
* Problem with GL Entry auto rename, and it do not update its link field in Partial Reconcile Entry.
* Why Full Reconcile Entry do not show Links to GL Entry (but to Partial Reconcile Entry)

### DIFF
--- a/erpnext/accounts/doctype/account/account.json
+++ b/erpnext/accounts/doctype/account/account.json
@@ -21,6 +21,7 @@
   "column_break1",
   "parent_account",
   "account_type",
+  "is_reconcile",
   "tax_rate",
   "freeze_account",
   "balance_must_be",
@@ -186,13 +187,19 @@
    "fieldname": "disabled",
    "fieldtype": "Check",
    "label": "Disable"
+  },
+  {
+   "default": "0",
+   "fieldname": "is_reconcile",
+   "fieldtype": "Check",
+   "label": "Is Reconcile"
   }
  ],
  "icon": "fa fa-money",
  "idx": 1,
  "is_tree": 1,
  "links": [],
- "modified": "2023-04-11 16:08:46.983677",
+ "modified": "2023-06-10 16:48:34.726345",
  "modified_by": "Administrator",
  "module": "Accounts",
  "name": "Account",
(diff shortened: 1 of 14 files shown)


kittiu:
I have done this a separated module.

https://github.com/kittiu/bs_reconcile

PR #35642, opened 2023-06-12
Payment against credit notes will be considered as payment against parent invoice in Accounts Receivable/Payable report

ruthra-kumar:
1. Create Sales Invoice
2. Make Payment Entry for [1]
3. Make Credit Note for [1]
4. Create another Sales Invoice
5. Manually link Cr Note from [3] to [4] using Journal Entry

Accounts Receivable/Payable will report the Invoice [1] and [3] as having outstanding.

Changes:
- [x] AR/AP report should handle such cases.
- [x] Payment Reconciliation tool should not report that Credit Note [3] as outstanding invoice

### DIFF
--- a/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
+++ b/erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
@@ -6,7 +6,6 @@ import frappe
 from frappe import _, msgprint, qb
 from frappe.model.document import Document
 from frappe.query_builder.custom import ConstantColumn
-from frappe.query_builder.functions import IfNull
 from frappe.utils import flt, get_link_to_form, getdate, nowdate, today
 
 import erpnext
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
```diff
@@ -204,6 +211,9 @@ def get_invoice_entries(self):
 			accounting_dimensions=self.accounting_dimension_filter_conditions,
 		)
 
+		cr_dr_notes = [x.voucher_no for x in self.return_invoices]
```
ruchamahabal, line 214: HR tests for Payment Reco with Expense Claim failing after this PR: 
https://github.com/frappe/hrms/actions/runs/5230489656/jobs/9468777474?pr=586 

`self.return_invoices` is only set for parties: customer & supplier in `get_dr_or_cr_notes` so references to `self.return_invoices` fail in HR (party is Employee)

image

ruthra-kumar: Checking

ruthra-kumar: addressed: https://github.com/frappe/erpnext/pull/35659

Issue #34680, opened 2023-03-31
Multi Currency Issues


ruthra-kumar:
## Issues pertaining to Multi-Currency Invoicing in ERPNext V14

1. Invoices that has Advance Payment with Exchange Gain/Loss has incorrect outstanding amount in V14
2. Cancelling Invoice/Payment from [1] leave behind Exchange Gain/Loss JE's with references removed. Which make it hard to track from where the JE was originally generated.
3. Accounts Receivable/Payable report should have provision to include Exc Gain/Loss incurred on each Invoice

Tasks:
- [x] Don't book Exchange Gain/Loss for Sales/Purchase Orders https://github.com/frappe/erpnext/pull/35061
- [x] Refactor `Advances` section in Sales and Purchase Invoices. It should submit Journal of type `Exchange Gain/Loss` https://github.com/frappe/erpnext/pull/35644
- [x] Refactor `Payment Entry`. It should create a Journal of type `Exchange Gain/Loss` for each reference with gain/loss https://github.com/frappe/erpnext/pull/35644

deepeshgarg007:
Related Issues 
- https://github.com/frappe/erpnext/issues/34676
- https://github.com/frappe/erpnext/issues/34605
- https://github.com/frappe/erpnext/issues/34584
- https://github.com/frappe/erpnext/issues/34407
- https://github.com/frappe/erpnext/issues/34345
- https://github.com/frappe/erpnext/issues/34754

ruthra-kumar:
1. Payment Reconciliation
2. Advances Section in Sales/Purchase Invoice
3. Deduction section in Payment Entry

All of these book Exchange Gain/Loss amount in different ways. [1] posts a JE of type `Exchange Gain/Loss`.
[2] & [3] post GL entries directly in the parent voucher name. For all these methods, when a parent is cancelled, the exchange gain/loss is left behind.

dj12djdjs:
> 



> 1. Payment Reconciliation
> 2. Advances Section in Sales/Purchase Invoice
> 3. Deduction section in Payment Entry
> 
> All of these book Exchange Gain/Loss amount in different ways. [1] posts a JE of type `Exchange Gain/Loss`. [2] & [3] post GL entries directly in the parent voucher name. For all these methods, when a parent is cancelled, the exchange gain/loss is left behind.

[...]

PR #35644, opened 2023-06-12
refactor: booking exchange gain/loss amount through journal

ruthra-kumar:
While Reconciling Multi-Currency Invoices, Exchange Gain/Loss amount will be booked through a Journal Entry of type `Exchange Gain or Loss`. This applies for Sales/Purchase Invoices, Payment Entries and Journal Entries.

Ex:
1. Create Sales Invoice with Exchange Rate 80 /-
2. Create Payment with Exchange Rate 75 /-
3. Reconcile [1] and [2]

Exchange difference of 5 /- will be booked through a Journal Entry of type `Exchange Gain or Loss`. When either of the parent document is cancelled, the Exchange Gain Or Loss is cancelled as well. This happens on the background without any input from user.

Following scenarios are covered:
1. Sales/Purchase Invoice with Advance payment/Journal
2. Payment/Journal Made against Sales/Purchase Invoice
3. Reconciling Invoices against Payment/Journal

part of: https://github.com/frappe/erpnext/issues/34680


erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
```diff
@@ -363,11 +363,11 @@ def reconcile_allocations(self, skip_ref_details_update_for_pe=False):
...
-				if payment_details.difference_amount and row.reference_type not in [
-					"Sales Invoice",
-					"Purchase Invoice",
-				]:
-					self.make_difference_entry(payment_details)
+				# if payment_details.difference_amount and row.reference_type not in [
+				# 	"Sales Invoice",
+				# 	"Purchase Invoice",
+				# ]:
+				# 	self.make_difference_entry(payment_details)
```
deepeshgarg007: Remove comments

erpnext/controllers/accounts_controller.py
```diff
@@ -965,67 +965,197 @@ def set_advance_gain_or_loss(self):
...
+								reverse_dr_or_cr + "_in_account_currency": 0,
+								reverse_dr_or_cr: abs(d.exchange_gain_loss),
+							}
 						)
-					)
+
+						journal_entry.append("accounts", journal_account)
+
+						journal_entry.save()
+						journal_entry.submit()
```
deepeshgarg007: Too big function, maybe break into smaller functions for easier readability?

erpnext/controllers/accounts_controller.py
```diff
@@ -1087,9 +1217,15 @@ def update_against_document_in_jv(self):
 			reconcile_against_document(lst)
 
 	def on_cancel(self):
-		from erpnext.accounts.utils import unlink_ref_doc_from_payment_entries
+		from erpnext.accounts.utils import (
+			cancel_exchange_gain_loss_journal,
+			unlink_ref_doc_from_payment_entries,
+		)
+
+		if self.doctype in ["Sales Invoice", "Purchase Invoice", "Payment Entry"]:
```
deepeshgarg007: How are payments via Journal Entry handled? Or is this only limited to payment entry for now?

erpnext/accounts/utils.py
```diff
@@ -638,28 +639,47 @@ def update_reference_in_payment_entry(
...
+		)
+		if journals:
+			exchange_journals = frappe.db.get_all(
+				"Journal Entry",
+				filters={
+					"name": ["in", [x[0] for x in journals]],
+					"voucher_type": "Exchange Gain Or Loss",
+					"docstatus": 1,
+				},
+				as_list=1,
```
deepeshgarg007, line 678: you can do `pluck="name"` to directly get a list of names of Journal Entry

erpnext/controllers/tests/test_accounts_controller.py
```diff
@@ -0,0 +1,501 @@
...
+		return customer_name
+
+
+class TestAccountsController(FrappeTestCase):
+	"""
+	Test Exchange Gain/Loss booking on various scenarios
+	"""
+
+	def setUp(self):
+		self.create_company()
```
deepeshgarg007, line 61: New company creation takes a lot of time, if that can be avoided and an existing company can be used would be better

erpnext/accounts/utils.py
```diff
@@ -638,28 +639,47 @@ def update_reference_in_payment_entry(
...
+	if parent_doc.doctype in ["Sales Invoice", "Purchase Invoice", "Payment Entry"]:
+		journals = frappe.db.get_all(
+			"Journal Entry Account",
+			filters={
+				"reference_type": parent_doc.doctype,
+				"reference_name": parent_doc.name,
+				"docstatus": 1,
+			},
+			fields=["parent"],
+			as_list=1,
```
deepeshgarg007, line 667: you can do pluck="name" to directly get a list of names of Journal Entry

erpnext/controllers/accounts_controller.py
```diff
@@ -965,67 +965,197 @@ def set_advance_gain_or_loss(self):
...
-			for d in self.get("advances"):
-				if d.exchange_gain_loss:
-					is_purchase_invoice = self.get("doctype") == "Purchase Invoice"
-					party = self.supplier if is_purchase_invoice else self.customer
-					party_account = self.credit_to if is_purchase_invoice else self.debit_to
-					party_type = "Supplier" if is_purchase_invoice else "Customer"
-
-					gain_loss_account = frappe.get_cached_value(
-						"Company", self.company, "exchange_gain_loss_account"
+	def make_exchange_gain_loss_journal(self, args=None) -> None:
```
deepeshgarg007: What happens when this is called from `sales_invoice.py` or `purchase_invoice.py`?  I can only see handling for Journal Entry and Payment Entry

erpnext/controllers/accounts_controller.py
deepeshgarg007: Show an alert to the user for Journal Entry creation. Also, add links in the Sales/Purchase and other documents dashboard (Connections tab) wherever applicable for the Journal Entry

deepeshgarg007 (changes requested):

erpnext/accounts/doctype/payment_reconciliation/payment_reconciliation.py
ruthra-kumar: removed

erpnext/controllers/accounts_controller.py
ruthra-kumar: function has been refactored

erpnext/controllers/accounts_controller.py
ruthra-kumar: This logic has been refactored. New logic handles Journal Entry as well.

erpnext/accounts/utils.py
ruthra-kumar: `pluck` will not be suitable here. Since we are filtering on `reference_name` in Journal Entry Accounts, `parent` is required to get the Journal name and its type. Which is required to make sure we can only cancelling `Exchange Gain Or Loss` entries.

erpnext/accounts/utils.py
ruthra-kumar: pluck will not be suitable here. Since we are filtering on reference_name in Journal Entry Accounts, parent is required to get the Journal name and its type. Which is required to make sure we are only cancelling Exchange Gain Or Loss entries.

erpnext/controllers/accounts_controller.py
ruthra-kumar: Added

erpnext/controllers/accounts_controller.py
ruthra-kumar: Existing design of linking payments or journals to other Doctypes requires modifying the said payments or journals.
References table for Payment Entry and Accounts table in Journals are modified to achieve this.

Exchange Gain/Loss booking logic relies on this existing design. Whether payment is made as an advance or normal payment, whether linked during invoice creation or reconciled later, the References table will always be updated. Same is true for Journals.

We are using this existing design to make Exchange Gain Or Loss only from Payment Entries or Journal Entries.

erpnext/controllers/tests/test_accounts_controller.py
ruthra-kumar: Addressed. Reusing '_Test Company'.

deepeshgarg007 (approved):

thida-bcn:
> While Reconciling Multi-Currency Invoices, Exchange Gain/Loss amount will be booked through a Journal Entry of type `Exchange Gain or Loss`. This applies for Sales/Purchase Invoices, Payment Entries and Journal Entries.
> 
> Ex:
> 
> 1. Create Sales Invoice with Exchange Rate 80 /-
> 2. Create Payment with Exchange Rate 75 /-
> 3. Reconcile [1] and [2]
> 
> Exchange difference of 5 /- will be booked through a Journal Entry of type `Exchange Gain or Loss`. When either of the parent document is cancelled, the Exchange Gain Or Loss is cancelled as well. This happens on the background without any input from user.
> 
> Following scenarios are covered:
> 
> 1. Sales/Purchase Invoice with Advance payment/Journal
> 2. Payment/Journal Made against Sales/Purchase Invoice
> 3. Reconciling Invoices against Payment/Journal
> 
> part of: #34680

is this possible to post only one JE for each payment
currently when we made one payment with multi invoice "Exchange Gain or Loss" JE post multi entry.
example
Payment with 3 Sales Invoice, system posted 3 JE of "Exchange Gain or Loss" JE 
Payment with 10 Sales Invoice, system posted 10 JE of "Exchange Gain or Loss" JE 
Another is Bug
when we made payment reconciliation
one payment and 5 invoices  >> system posted 25 JE of "Exchange Gain or Loss" JE

ruthra-kumar:
> one payment and 5 invoices >> system posted 25 JE of "Exchange Gain or Loss" JE

@thida-bcn 
The expected behaviour is, For example, if there are 2 invoices linked to a payment, there will be 2 JE's of 'Exchange Gain or Loss'. It should not post multiple JE's for each references. Does the payment have invoices linked before reconciliation?

Can you give simpler scenario on the reconciliation bug?

Fredoluka:
I want to sincerely thank the team that worked on this because currency revaluation issues have been a torn in the flesh of ERPNext for organizations that use multicurrency. Many headaches have been solved.
However, there are still some issues that need the attention of this team.
1. Unrealised Exchange  Gain or Loss is not an income account. It should be reported in the Equity. From the company doctype where the account is defined, currently, you cannot select any account type Equity. It is only the Income account type that can be selected. Please help to make a selection of Equity account type possible.
2. For suppliers' and customers' balances revaluation, we observed the revalued balances contain the element of exchange gain or loss which are tagged as unrealized exchange gain and loss, and when the invoices that made up the balances are paid, new exchange gain and loss are calculated and posted without reference to the unrealized exchange gain and loss that have been previously booked. This amount in most cases cannot be equal to the amount previously charged to the customer/supplier account during the previous revaluation thereby making the customer /supplier balance in the base currency not to be correct.
Let there be a workaround to solve the problem.

PR #35650, opened 2023-06-12
feat: Ledger Preview

deepeshgarg007:
Accounting and Stock Ledger previews for buying and selling cycle, will gradually add to other doctypes as well wherever applicable

image

image

`no-docs`


erpnext/public/js/utils/ledger_preview.js
```diff
@@ -0,0 +1,76 @@
...
+				})
+			}, __("Preview"));
+		}
+	},
+
+	show_stock_ledger_preview(frm) {
+		let me = this
+		if(!frm.is_new() && frm.doc.docstatus == 0) {
+			frm.add_custom_button(__('Stock Ledger'), function() {
+				frappe.call({
```
ankush, line 28: Use `get` request here. Safer IMO.

erpnext/controllers/stock_controller.py
```diff
@@ -845,6 +845,153 @@ def add_gl_entry(
...
 
 
+@frappe.whitelist()
+def show_accounting_ledger_preview(company, doctype, docname):
+	frappe.db.savepoint("show_accounting_ledger_preview")
+
+	filters = {"company": company, "include_dimensions": 1}
+	doc = frappe.get_doc(doctype, docname)
+
+	gl_columns, gl_data = get_accounting_ledger_preview(doc, filters)
```
ankush, line 853: Wrap these in try...except? 

Otherwise some write can be committed by mistake.

IMO no savepoint is required. Do a full rollback instead.

ankush (changes requested):

PR #35672, opened 2023-06-13
fix: make showing taxes as table in print configurable

### LINKED ISSUE #35662: Hardcoded overridden print templates
https://github.com/frappe/erpnext/blob/42f4f80e0cc4fc6a52f4bce99234b8f1b8ddc395/erpnext/controllers/print_settings.py#L30-L35

These should be made configurable, it's confusing as to why a table renders as 3-4 lines instead of... well, a table. 

It should be made configurable OR a separate component can be somehow provided to render fields with standard renderer.

anandbaburajan:
Closes #35662

Adding a new Accounts setting called `show_taxes_as_table_in_print` to let users choose whether they want to have taxes as a table in print formats or they want to go with the `templates/print_formats/includes/taxes.html` template.

### DIFF
--- a/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
+++ b/erpnext/accounts/doctype/accounts_settings/accounts_settings.json
@@ -34,6 +34,7 @@
   "book_tax_discount_loss",
   "print_settings",
   "show_inclusive_tax_in_print",
+  "show_taxes_as_table_in_print",
   "column_break_12",
   "show_payment_schedule_in_print",
   "currency_exchange_section",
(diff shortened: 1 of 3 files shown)


erpnext/accounts/doctype/accounts_settings/accounts_settings.json
```diff
@@ -376,14 +377,20 @@
    "fieldname": "auto_reconcile_payments",
    "fieldtype": "Check",
    "label": "Auto Reconcile Payments"
+  },
+  {
+   "default": "0",
```
ankush, line 382: 1. Defaults should usually be consistent with current behaviour to avoid breaking changes. 
2. Singles can not sync defaults on existing sites so you should either patch it or rewrite business logic so default of `0` makes sense. i.e. `dont show taxes as table in print` (reword it)

```suggestion
   "default": "1",
```

ankush (changes requested):
Good first step towards fixing long pending annoyance :rocket:

erpnext/accounts/doctype/accounts_settings/accounts_settings.json
ankush: Ignore. brainfart moment.

PR #35687, opened 2023-06-14
Fix:General Ledger Report

nomi-gee:

Upon selecting the accounting dimension in General Ledger Report it throws an error

Small syntax fix

> Screenshots/GIFs

image


deepeshgarg007:
@nomi-gee This doesn't look like a proper fix, tested local and it fails when multiple values are select. Will need more info to investigate what exactly your error was

PR #27528, opened 2021-09-15
feat: Payments based dunning

barredterra:
Related to #26969

Test this PR at https://gh-imp-2021-01268.castlecraft.co.in (User: mossdiana@hotmail.com, Password: 7b97830487)

Create dunning for all overdue payments of a customer. We use a table similar to the payment schedule in sales invoice to pull all overdue scheduled payments into the Dunning.

- Rate of Interest is the same for all overdue payments.
- Interest amount is calculated for each overdue payment.
- Dunning Fee is applied only once per dunning.

### Changes in accounting

- Dunning amount will no longer be booked to receivables. The reason is that the dunning amount is not paid in many cases / is only intended as a warning. If paid, the dunning amount will be recorded as additional income in Payment Entry. See [here](https://www.rechnungswesenforum.de/forum/buchfuehrung-buchhaltung/allgemein/forderung-mahngebuehren.423727/) and [here](https://www.haufe.de/finance/haufe-finance-office-premium/mahnung-und-mahnverfahren-73-buchung-mahngebuehren-und-verzugszinsen_idesk_PI20354_HI2015327.html) for further reading (in german).

Preview:

new_dunning

Docs: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/dunning/edit?wiki_page_patch=894eeb58e3
Discuss: https://discuss.erpnext.com/t/dunning-mahnung-warning-notice/49499/30?u=rmeyer

To do:

- [x] Refactor `resolve_dunning`
- [x] Add patch
- [x] Adjust docs
- [x] Update test cases to use new mandatory **Company** field in **Dunning Type**.

Issue #26969, opened 2021-08-16
Mark Sales Invoice as Overdue when first payment becomes overdue


barredterra:
**Is your feature request related to a problem? Please describe.**

In **Sales Invoice** we can specify a payment schedule. For example, 50 % after 14 days, 50 % after 30 days. Currently the latest date in the payment schedule is used as the invoice's due date. Only when the last payment becomes overdue, the invoice status changes to "Overdue".

However, we regard a **Sales Invoice** as overdue based on the scheduled payments. Once the first payment is overdue, we want to create a **Dunning**, reminding the customer that he is late on this payments.

**Describe the solution you'd like**

- The status of a **Sales Invoice** depends on the payment schedule instead of due date
- Set status to "Overdue" once a scheduled payment is overdue
- Introduce a new status "Partially Paid" that is displayed when payments have been received according to the payment schedule.
- ~Extend **Dunning** to consider only overdue payments and calculate overdue days correctly. Probably it would make sense to copy the **Sales Invoice**'s payment schedule to the **Dunning**.~ -> Create a new Issue for this

**Describe alternatives you've considered**
> A clear and concise description of any alternative solutions or features you've considered.

**Additional context**

The option "Allocate Payment Based On Payment Terms" is set in **Payment Terms Template**.

mymi14s:
👍

PR #35689, opened 2023-06-14
feat: Payments based dunning

marination:
An attempt to revive https://github.com/frappe/erpnext/pull/27528

--------------------------

Related to #26969

Create dunning for all overdue payments of a customer. We use a table similar to the payment schedule in sales invoice to pull all overdue scheduled payments into the Dunning.

- Rate of Interest is the same for all overdue payments.
- Interest amount is calculated for each overdue payment.
- Dunning Fee is applied only once per dunning.

### Changes in accounting

- Dunning amount will no longer be booked to receivables. The reason is that the dunning amount is not paid in many cases / is only intended as a warning. If paid, the dunning amount will be recorded as additional income in Payment Entry. See [here](https://www.rechnungswesenforum.de/forum/buchfuehrung-buchhaltung/allgemein/forderung-mahngebuehren.423727/) and [here](https://www.haufe.de/finance/haufe-finance-office-premium/mahnung-und-mahnverfahren-73-buchung-mahngebuehren-und-verzugszinsen_idesk_PI20354_HI2015327.html) for further reading (in german).

Preview:

new_dunning

Docs: https://docs.erpnext.com/docs/v13/user/manual/en/accounts/dunning?editWiki=1&wikiPagePatch=318ad1abe6
Discuss: https://discuss.erpnext.com/t/dunning-mahnung-warning-notice/49499/30?u=rmeyer

--------------------------

**ToDo:**
- [x] Resolve conflicts
- [x] Manual tests of functionality
- [x] Test of migration patch
- [x] Fix: Manually added Payment Schedule without template or terms keeps dunning unresolved. Partial Fix: #35798 (users must take care of this. A note is added to the docs)
- [x] Possibly add some unit tests 
- [x] Apply changes suggested by reviewers / community
- [x] Check and update the docs


barredterra (dismissed):
Tested, LGTM

marination:
@deepeshgarg007 This is tested, reviewed and ready to merge. You can give it a last check and let us know if it looks okay.

marination:
@deepeshgarg007 the patch reverses GL entries for unresolved dunnings. Will it be problematic if there are some old unresolved dunnings ? From before the **Accounts Frozen Till Date** ? Those should stay untouched right ?

marination:

muchai:
This could use a backport to version 14 please.

bosue:
@muchai: This is a massive change, so backporting would break lots of custom code in v14 IMO. Also, it‘s still experimental, so not fully settled in v15. We all loved the feature and wanted to see it in v15, but IMO there‘s still much to do until it may be considered stable and be backported. So instead, you might want to upgrade to v15?

But that‘s just my two cents. I personally don‘t care if it‘s backported. I just can‘t consider it finished, and while having quite many advantages, there‘s a couple of drawbacks, including lost features that should probably be reinstated before backporting.

PR #35694, opened 2023-06-14
fix: reference error while using exchange rate revaluation

ruthra-kumar:
Screenshot 2023-06-14 at 8 31 42 PM


ruthra-kumar:

PR #35697, opened 2023-06-14
chore: fr translation lead vs prospect

FHenry:
In version-14 the Prospect Doctype was introduce, so in french the fact Lead Doctype and Prospect Doctype have the same translation "Prospect" is really confusing.

Lead as in ERPNext concept could be translate as "first contact"=>"Premier Contact", but no one use this, the english word "Lead" is pass in common language  in french, all sales team users understand it as it is implement in ERPNext.


FHenry:

FHenry:
@deepeshgarg007 Could you please backport to version-14 please ?

PR #35777, opened 2023-06-19
feat: Demo setup

deepeshgarg007:
Ability to setup demo data on new installations

Demo Setup

##### Masters
- Item, Item Group
- Customer, Customer Group
- Supplier, Supplier Group

##### Transactions
- Sales Invoice
- Purchase Invoice
- Payment Entry

`no-docs` -  THIS IS INTERNAL TO ERPNext. This can be changed/removed any time.


ankush:
Just some notes to check before review

- [x] We should add a test for this. A test that doesn't do anything apart from triggering demo setup on new site and checking that it completes successfully. The last demo-data setup became unusable as it had no tests and schema/validations went out of sync over the years.
- [ ] Data should hydrate reports and graphs with enough variety to give users that "aha" moment of finding _"insights"_ from their data.
- [ ] Where will this option to add/remove demo data be present, few options:
    - [x] setup wizard - easily discoverable
    - [ ] first onboarding task in home onboarding widget - slightly less discoverable
- [x] Erasure of demo data: company, transactions, masters.
- [ ] "discoverability" tours using demo data setup.

git-avc:
Hi:

Maybe you can consider to insert business data using the current year, instead fixed dates. This way, this feature will no get "outdated" on reports, etc ...

erpnext/public/build.json
```diff
@@ -19,6 +19,7 @@
 		"public/js/queries.js",
 		"public/js/sms_manager.js",
 		"public/js/utils/party.js",
+		"public/js/utils/demo.js",
```
ankush: This file is dead actually.

ankush:
- [ ] hydrate buying/selling dashboards
- [ ] make default demo show "profit" :woozy_face: 
- [ ] have more than 1 warehouses by default for stock dashboard

ankush (approved):

PR #33165, opened 2022-11-29
refactor: Exchange rate revaluation to handle accounts with zero account balance

ruthra-kumar:
Exchange Rate Revaluation will handle Foreign accounts with zero balance on account currency but has balance on company currency.

Changes in this PR:
1. New Journal type `Exchange Gain/Loss`. This type skips validations: `validate_debit_and_credit_amount` and `set_amounts_in_company_currency` which helps to submit GL's with '0' debit/credit value in either currencies. 
2. https://github.com/frappe/erpnext/blob/66dbf94151ca15d1911b2133f5d05dc3c6593f72/erpnext/accounts/report/utils.py#L81 has been modified to handle GL's from [1]. 
3. Exchange Rate Revaluation will now create 2 Journals. 
Screenshot 2022-12-02 at 9 43 16 PM

Discussion thread:
https://discuss.erpnext.com/t/how-to-maintain-correct-account-balance-in-multiple-currencies-in-case-of-foreign-bank-account/96839

Solution:
https://discuss.erpnext.com/t/how-to-maintain-correct-account-balance-in-multiple-currencies-in-case-of-foreign-bank-account/96839/9?u=ruthra

Conder a scenario where a Foreign currency account has balance in company currency but not in account currency. 'HDFC USD - TC' in the below screenshot.
Screenshot 2022-12-02 at 9 41 15 PM

ERR can submit a journal to move only the company currency balance to Exchange Gain/Loss account. 
Screenshot 2022-12-02 at 9 43 16 PM

Screenshot 2022-12-02 at 9 43 31 PM

Screenshot 2022-12-02 at 9 43 43 PM

General Ledger:
Company/Base Currency:
Screenshot 2022-12-03 at 7 23 24 AM

Account Currency:
Screenshot 2022-12-03 at 7 23 29 AM

PR #35794, opened 2023-06-20
fix: Exchange Rate Revaluation should only post on the currency that has balance in a 'zero' balance account

ruthra-kumar:
On Accounts that have balance only in account currency, Journals created by Exchange Rate Revaluation cause debit - credit mismatch in Trial Balance report. ERR should only post to currencies that have balance.


dineshpanchal93:
Are we going to push these changes in version-13 ?

ruthra-kumar:
@dineshpanchal93 
`Exchange Rate Revaluation` doctype has gone through a major refactor back in January https://github.com/frappe/erpnext/pull/33165 which was done on v14. The changes in this PR is for the refactored v14 version. This can't be backported to v13.

Issue #35795, opened 2023-06-20
frappe.exceptions.DoesNotExistError: DocType KSA VAT Setting not found


ashish-greycube:
### Information about bug

When migrating from v12->v13 migration fails with following traceback
This seems like a bug in ERPNext. While migrating the doctype is not reloaded before adding permission.

```

Executing erpnext.patches.v13_0.enable_ksa_vat_docs #1 in kayan.frappe.cloud (_f8e4e799afa4f162)

Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 110, in <module>
    main()
  File "/home/frappe/frappe-bench/apps/frappe/frappe/utils/bench_helper.py", line 20, in main
    click.Group(commands=commands)(prog_name="bench")
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/frappe/frappe-bench/env/lib/python3.7/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/frappe/frappe-bench/apps/frappe/frappe/commands/__init__.py", line 31, in _func
    ret = f(frappe._dict(ctx.obj), *args, **kwargs)
[...]

PR #35797, opened 2023-06-20
fix: frappe.exceptions.DoesNotExistError: DocType KSA VAT Setting

ashish-greycube:
fix: frappe.exceptions.DoesNotExistError: DocType KSA VAT Setting not found

When migrating from v12->v13 migration fails
While migrating the doctype is not reloaded before adding permission. #35795

> Screenshots/GIFs


ashish-greycube:
Context : frappecloud.com/helpdesk/ticket : 422114

PR #35816, opened 2023-06-20
fix: onchange of UOM for "Update Items"

sameer55chauhan:
To reproduce the error: in any document that has an "Update Items" Button, if we add a new row using update items and then select the UOM then the previous code results in error.

Have fixed the same.
Please let me know if there is a way to detect in onchange which row was modified by the user in child table in a Dialog.


deepeshgarg007:
@sameer55chauhan The fix doesn't look correct based on the initial review that I am doing. Not advised to make a server-side call for each item. Please update what is the current error that you get and also try and explain the fix in more detail. Closing this PR for now

PR #35818, opened 2023-06-20
fix: group item reorder by (warehouse, material_request_type)

dj12djdjs:
This allows you to reorder via purchase and transfer at the same time.
image

Use Case:
You order 10pc from supplier. Lead time is 2 weeks.
You have a primary warehouse containing majority of stock, you will get two material requests. You can transfer from main warehouse or wait for purchase.


erpnext/stock/doctype/item/item.py
```diff
@@ -395,12 +395,12 @@ def validate_barcode(self):
...
+		warehouse_material_request_type: list[tuple[str, str]] = []
 		for d in self.get("reorder_levels"):
 			if not d.warehouse_group:
 				d.warehouse_group = d.warehouse
-			if d.get("warehouse") and d.get("warehouse") not in warehouse:
-				warehouse += [d.get("warehouse")]
+			if (d.get("warehouse"), d.get("material_request_type")) not in warehouse_material_request_type:
+				warehouse_material_request_type += [(d.get("warehouse"), d.get("material_request_type"))]
 			else:
 				frappe.throw(
```
s-aga-r, line 405: @dj12djdjs update the error msg

erpnext/stock/doctype/item/item.py
```diff
@@ -395,17 +395,17 @@ def validate_barcode(self):
...
-					_("Row {0}: An Reorder entry already exists for this warehouse {1}").format(
-						d.idx, d.warehouse
-					),
+					_(
+						"Row {0}: An Reorder entry already exists for this warehouse and reorder type ({1}, {2})"
```
s-aga-r:
```suggestion
						"Row #{0}: A reorder entry already exists for warehouse {1} with reorder type {2}."
```

s-aga-r (changes requested):

s-aga-r (approved):

PR #35821, opened 2023-06-21
Fixes issue of asset value_after_depreciation field getting updated twice if workflow is enabled in Journal Entry

saeedkola:

The asset value_after_depreciation field is getting updated 2 times if workflow is enabled on the Journal Entry Doctype
The make_depreciation_entry def checks if workflow exists for JE, and leaves the depreciation entry in draft state, but it still updates the value_after_depreciation field in the finance_book table of the asset.
And after the workflow is completed the update_asset_value() method updates it again, reducing it even further, for the same depreciation entry

This fix will not update the asset value if JE has workflow, it will only update when JE is submitted
> Screenshots/GIFs
Asset depreciation issue

### DIFF
--- a/erpnext/assets/doctype/asset/depreciation.py
+++ b/erpnext/assets/doctype/asset/depreciation.py
@@ -137,7 +137,6 @@ def make_depreciation_entry(asset_name, date=None):
 			je.flags.planned_depr_entry = True
 			je.save()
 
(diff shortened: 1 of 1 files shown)


anandbaburajan:
Thanks @saeedkola! LGTM.

PR #35825, opened 2023-06-21
fix: multiple fixes in reconciliation tools

ruthra-kumar:
1. Permission error while using Payment Reconciliation tool. Happens if user only has 'Accounts User' role. Screenshot 2023-06-21 at 2 20 54 PM
2.  While reconciling bank transaction, Journal Entry creation throws cost center validation error, as cost center from different company is used in Journal. This only happens on multi company setup. 
Screenshot 2023-06-21 at 9 55 16 PM


ruthra-kumar:

PR #35828, opened 2023-06-21
fix: use correct fieldname for purchase receipt column in item_wise_purchase_register report

phot0n:
Purchase Receipt not visible in item-wise purchase register

before:

after:


erpnext/accounts/report/item_wise_purchase_register/item_wise_purchase_register.py
```diff
@@ -87,7 +87,7 @@ def _execute(filters=None, additional_table_columns=None, additional_query_colum
 				"project": d.project,
 				"company": d.company,
 				"purchase_order": d.purchase_order,
-				"purchase_receipt": d.purchase_receipt,
+				"purchase_receipt": purchase_receipt,
```
phot0n, line 90: I'm not completely sure about this change.

@deepeshgarg007 could you please take a look?

PR #35830, opened 2023-06-21
fix: incorrect cost center error in bank reconciliation

ruthra-kumar:
While reconciling bank transaction, Journal Entry creation throws cost center validation error, as cost center from different company is used in Journal. This only happens on multi company setup.
Screenshot 2023-06-21 at 9 55 16 PM


ruthra-kumar:
This was included in https://github.com/frappe/erpnext/pull/35825

PR #35831, opened 2023-06-21
fix!: make reorder_level not required

dj12djdjs:
**Use Case:** You want to auto create a material transfer for 2pc when stock level reaches zero.


barredterra:
Breaking changes like these (`<` -> `<=`)  should be documented in the migration guide! Suddenly all users get unexpected **Material Requests** after upgrading to v15.

PR #35832, opened 2023-06-21
fix: asset capitalization

anandbaburajan:
- Now users won't need to create the target asset -- it would be done automatically.
- Now the issue of "incorrect number of General Ledger Entries found" when all the consumable assets are non-depreciable has been fixed -- the single unnecessary entry won't be posted.

https://docs.erpnext.com/docs/v14/user/manual/en/asset/asset-capitalization


muchai:
Please include documentation for Asset Capitalization on the asset manual. It's usually missing there.

anandbaburajan:
@muchai https://docs.erpnext.com/docs/v14/user/manual/en/asset/asset-capitalization.

muchai:
> @muchai https://docs.erpnext.com/docs/v14/user/manual/en/asset/asset-capitalization.

Thank you very much.

PR #35836, opened 2023-06-22
fix!: UX of supplier linking with supplier users on portal pages

### LINKED ISSUE #35772: fix UX of Supplier linking with supplier users on portal pages
Right now this is how it works:

 `Supplier <dynamic link> Contact <link> User account with supplier role`

If all of these are satisfied then the user can access their PO/RFQs etc. The "Contact" part in the middle is neither documented nor discoverable. 

I propose creating a separate table for adding portal users like this to simplify linking of users:

image

TODO:

- [ ] Go through current codebase for applying permissions
- [x] Create a new child table doctype for linking users.
- [x] Add this table to supplier and customer. 
- [x] Modify the original permission check hook to consider this table instead.
- [x] Consider indexing for performance
- [x] Write patch to auto-populate existing portal users based on old business logic
- [ ] Document the changes

GursheenK:
The current flow of adding Users for a particular Supplier that are allowed to access the Supplier Portal includes performing the following steps - 

1. Create a Supplier/Customer. 
2. Create a User with the Supplier/Customer Role.
3. Add a contact from the Supplier Doctype to add the User to the list of users who can access that particular Supplier's Portal.

This creates a dynamic link from the Contact to the Supplier.

**Changes**

After adding the Child Table to the Supplier and Customer Doctypes, the flow is easier to follow - 
1. Create a Supplier.
2. From the Portal Users Tab, directly adding/removing one or multiple users to the list.

Closes #35772


erpnext/buying/doctype/supplier/supplier.py
```diff
@@ -62,6 +62,13 @@ def validate(self):
 
 		validate_party_accounts(self)
 		self.validate_internal_supplier()
+		self.validate_portal_users()
+
+	def validate_portal_users(self):
+		for portal_user in self.portal_users:
+			user_roles = frappe.get_roles(portal_user.user)
+			if "Supplier" not in user_roles:
+				frappe.throw(_("Portal Users must have Supplier Role."))
```
ankush: Let's automatically add this role. This role doesn't grant any permission besides considering user as "Supplier". 

- Use ORM, so if user has write to add this role then only they'll be able to do it. 
- We should give alert with `frappe.msgprint(alert=True` that we have automatically added this role to user. 
- If user doesn't have right to grant this role then throw error.

erpnext/buying/doctype/supplier/supplier.py
```diff
@@ -153,3 +160,22 @@ def get_supplier_primary_contact(doctype, txt, searchfield, start, page_len, fil
...
+
+def get_users_query():
+	user = frappe.qb.DocType("User")
+	hasRole = frappe.qb.DocType("Has Role")
+	q = (
+		frappe.qb.from_(user)
+		.inner_join(hasRole)
+		.on(user.name == hasRole.parent)
+		.select(user.name, user.full_name)
+		.where(hasRole.role == "Supplier")
```
ankush: If possible lets use `frappe.get_list` instead of raw query. Raw query wont respect permission rules that site might have configured. 


Something like this will work:
`frappe.get_list("User", {"role": "Supplier"}, ["name", "full_name"])`

Also, no need for two separate functions here IMO. Original function can call get_list directly.

erpnext/buying/doctype/supplier/supplier.py
```diff
@@ -153,3 +160,22 @@ def get_supplier_primary_contact(doctype, txt, searchfield, start, page_len, fil
...
+	)
+	return q
+
+
+@frappe.whitelist()
+def get_users_with_supplier_role(doctype, txt, searchfield, start, page_len, filters):
```
ankush: - Use `txt` parameter for searching user by name.
- Use `start` and `page_len` parameter to limit results (usually 20 are returned)

erpnext/controllers/website_list_for_contact.py
```diff
@@ -230,31 +230,29 @@ def get_customers_suppliers(doctype, user):
...
 
 	return customers if has_customer_field else None, suppliers if has_supplier_field else None
 
 
+def get_parents_for_user():
+	portal_users = frappe.qb.DocType("Portal Users")
+	q = (
+		frappe.qb.from_(portal_users)
+		.select(portal_users.parent)
+		.where(portal_users.user == frappe.session.user)
```
ankush: Since here we are filtering by `user` column. Enabling index on `user` column will help if there are many users in system. 

This can be enabled by just checking "Search index" on the docfield.

erpnext/controllers/website_list_for_contact.py
```diff
@@ -230,31 +230,29 @@ def get_customers_suppliers(doctype, user):
...
+	)
+	res = q.run(as_dict=True)
+	res = [i["parent"] for i in res]
```
ankush:
```suggestion
	results = q.run(pluck="parent")
```

- Use full names, short names are usually bad for readability. 
- `pluck` can pick the specific field you want.

https://github.com/frappe/erpnext/wiki/Naming-Guidelines

erpnext/selling/doctype/customer/customer.js
```diff
@@ -63,6 +63,12 @@ frappe.ui.form.on("Customer", {
 				}
 			}
 		});
+
+		frm.set_query("user", "portal_users", function() {
+			return {
+				query: "erpnext.selling.doctype.customer.customer.get_users_with_customer_role",
+			};
```
ankush: Maybe custom query is not required at all. All you need is filters right? :grimacing: 


This filter will allow showing website users too (disabled by default) `filter: {"ignore_user_type": True}`

ankush (changes requested):

erpnext/selling/doctype/customer/customer.js
ankush: reference: https://github.com/frappe/frappe/blob/4f70200bd5a259095b34e255ed664d58dfdea924/frappe/core/doctype/user/user.py#L961-L972

ankush (approved):

ankush:
@GursheenK update migration guide: https://github.com/frappe/erpnext/wiki/Migration-Guide-to-ERPNext-nightly-version-(future-v15)-%5BWIP%5D 


just need to explain the change that you did and how existing portal users need to change this behaviour. 


Also see the behaviour of assigning default role, because customer are self-signup often they were automatically created, contact created and they could access customer portal by default. We can add similar behaviour back by populating portal user added to automatically created customer I think.





image

PR #35837, opened 2023-06-22
refactor: increase precision for current exc rate in Exchange Rate Revaluation

ruthra-kumar:
Increase currency precision for 'current exchange rate' field in Exchange Rate Revalution master. This is to avoid rounding mismatch when re-evaluating a forex account.


ruthra-kumar:

PR #35851, opened 2023-06-22
chore: asset scrap and restore fixes [v14]

anandbaburajan:
Added better error message when users tried to cancel JE for asset scrap. Also fixed the issue of `list index out of range` when users tried to restore a scrapped non-depreciable asset.


anandbaburajan:
@mergify backport version-14

PR #33991, opened 2023-02-08
feat: allow arbitrary purchase order items

casesolved-co-uk:
Businesses may make arbitrary purchases of lots of items that they do not wish to make item_codes for, for example a prototype manufacturer.

Makes item_code & description non-mandatory, and sets conversion factor default to 1

Also purchase receipt and RFQ

Matches the current Sales and Purchase Invoice behaviour

You might think that you can just customise the doctype, but mandatory cannot be removed from standard fields using customisation.
Another argument for this is that the modules in erpnext were presumably split because they're supposed to be independent and self-contained, but requirements like this creates dependencies between the buying, selling and stock modules which defeats the objective.

Let me know if you want the mandatory properties added as property setters to maintain historic consistency?

version-14-hotfix
version-13-hotfix
no-docs

PR #33430, opened 2022-12-25
fix: allow arbitary items in Quotation and Sales Order

SvbZ3r0:
Arbitary items are required while quoting for an engineered-to-order product. There can't be an item in your database for every  request your customer makes.
Allowing arbitary items in Sales Orders follows the same logic. Billing of arbitary items is already allowed.

closes https://github.com/frappe/erpnext/issues/33429

The Buying side of this issue can possibly be implemented in a similar fashion, but I'm not sure if it'll break anything.

PR #35856, opened 2023-06-22
feat: allow arbitrary purchase order items (continued)

dj12djdjs:
Continuation of: https://github.com/frappe/erpnext/pull/33991

- [ ] Add setting to enable this feature in buying setting.
	- [ ] test: property setter doesn't conflict with existing property setters. (When enabling, and disabling the feature)

- [ ] test: purchase flow with feature enabled/disabled.

- [ ] add documentation including real world use case.


casesolved-co-uk:
@dj12djdjs thanks for picking this up.

Note however my initial intention (and I have more code not shared) was to make this also applicable to sales transactions too.

So I would perhaps put the setting in accounts settings rather than buying?

I started with purchases so that acceptance might be easier.

I can perhaps merge and push my other changes to your PR?

dj12djdjs:
@casesolved-co-uk That would be much appreciated, I've granted you write permission to my fork.

SvbZ3r0:
Looking at how long the development for this feature has been going on, I'm worried my fix for this issue on the Sales and Quotations (#33430) side of the coin might have overlooked quite a few things. The only change I made was to mark the `item_code` field non-mandatory. 
Would you guys care to take a look?

casesolved-co-uk:
Active 🙄

PR #35858, opened 2023-06-23
fix: show non-depreciable assets in fixed asset register

anandbaburajan:
Non-depreciable assets weren't being shown in the fixed asset register after the recent refactor, so fixed that.


anandbaburajan:
@mergify backport version-14

PR #35875, opened 2023-06-24
feat: display description of accounts to normal users

blaggacao:
Accounts Manager can set the description to help auxiliary accountants
understand the usage policy of certain accounts.

Add official (as in the law) descriptions to colombian CoA.

**For the Account Manager**
image

**For everybody else**
image


ernestoruiz89:
I my opinion is better use "Text Editor" field instead HTML field

blaggacao:
I had thought about this for a while. But ultimately one more common use case that I can see in a real world scenario is something like:

| Credit | Debit |
| --- | --- |
| instructions | instructions |

So plain text would not be able to reflect the required formatting of that use case.

Between markdown and html, I'm not sure what'd be better.

But we could change it to markdown instead (which is a superset of html anyways).

ernestoruiz89:
> I had thought about this for a while. But ultimately one more common use case that I can see in a real world scenario is something like:
> 
> Credit	Debit
> instructions	instructions
> So plain text would not be able to reflect the required formatting of that use case.
> 
> Between markdown and html, I'm not sure what'd be better.
> 
> But we could change it to markdown instead (which is a superset of html anyways).

I dont see an accountant writing HTML code, instead Text Editor is a easy to use

image

blaggacao:
Ah! Got ya! "Text Editor" means "WYSIWYG". Didn't realize that. Thx! Yeah, I agree!

:heavy_check_mark: suggestion adopted in latest push.

blaggacao:
@deepeshgarg007 @ruthra-kumar IMO, this is ready to merge. I implemented @ernestoruiz89 ' sugestion to use the `Text Editor` field type.

earona:
Hello,
It would be interesting to have this functionality

earona:
Hello @blaggacao you can reformulate the PR by adding no-docs, so that it is merged.

Thank you

blaggacao:
@earona This is fully ready to review/merge. It's up to the maintainers to decide whether this is a priority and can be considered.

In the meanwhile you could [patch](https://patch-diff.githubusercontent.com/raw/frappe/erpnext/pull/35875.patch) your local checkout. I'm probably going to rebase this once in a while.

PR #35908, opened 2023-06-27
perf: improve item wise register reports

anandbaburajan:
Improved the `item_wise_sales_register` and `item_wise_purchase_register` by removing the unnecessary `get_item_details` function (and adding the same logic to the query in the `get_items` function) and memoized the values of scrubbed tax fields.


anandbaburajan:
@mergify backport version-14

PR #35910, opened 2023-06-27
fix: also check on_hold

RJPvT:
in reference to my notification in https://github.com/frappe/erpnext/pull/35788 payment entry can get hold, recieve a release date en get hold released but NOT a changed release date. in that case on_hold =0 but it still wont show in this list and making a payment entry is not possibile.

it should always check for on_hold because that is the main field.


anandbaburajan:
> in reference to my notification in https://github.com/frappe/erpnext/pull/35788

@RJPvT did you mean your comment in https://github.com/frappe/erpnext/pull/35674?

RJPvT:
Yes wrong number, sorry ;)

PR #35918, opened 2023-06-28
fix: asset movement

anandbaburajan:
Asset Movement's UX was confusing -- the unnecessary fields for each 'purpose' weren't made read-only properly hence users couldn't figure out how to ~transfer~ issue an Asset to an employee. The docs are insufficient too, so will update them as well.


anandbaburajan:
@mergify backport version-14

PR #35944, opened 2023-06-30
fix: Expense Account filter in Sales Invoice

deepeshgarg007:


gbm001:
@deepeshgarg007

Another user has commented that this breaks Purchase Order (the comment is on the commit). This change also breaks Purchase Invoice - you can no longer select an asset account expense head on a Purchase Invoice item.
Documentation here:
https://docs.erpnext.com/docs/user/manual/en/purchase-invoice#4-1-accounting-impact

You should be able to select an Asset account.

PR #35950, opened 2023-06-30
feat: add Lead to Address And Contact report

FHenry:
feat: add Lead to Address And Contact report

Note to reviewer :

If merge
 - please add back-port to version-14-hotfix and version-13-hotfix label
 - I'll  make a PR to add Prospect (but it will be only for version-14-hotfix, but now I would like Lead for version-13)


deepeshgarg007 (approved):

FHenry:
@deepeshgarg007 
Could you please add back port to version-13 and version-14

PR #35955, opened 2023-07-01
fix: incorrect reposting causing stock adjustment entry

rohitwaghchaure:
### **Issue**

1. Created Purchase Receipt for the item A with 100 Qty and Rate as 273.45 in the warehouse Stores
2. Created return entry against above purchase receipt for 2 quantity 
3. Created Stock Transfer Entry from "Stores" to "Warehouse A"  for 20 Qty
4. Created Stock Transfer Entry from "Stores" to "Warehouse B"  for 20 Qty
5. Created Stock Transfer Entry from "Stores" to "Warehouse C"  for 20 Qty
6. Created Stock Transfer Entry from "Warehouse A" to "Stores"  for 20 Qty (This entry has an issue)
7. After that created Backdated Landed Cost Voucher against the Purchase Receipt (created at point no 1)
8. In the landed cost voucher reduced the cost from 273.45 to 250
9. System has auto created Repost Item Valuation entry after the landed cost voucher
10. Then checked the Stock Ledger Entries and found that the system has changed the valuation rate from 273.45 to 250 for all future stock ledgers except for the Stock Entry which was created at point no 6 ("Warehouse A" to "Stores")
11. Because of this system has created Stock Adjustment gl entries against the "Material Transfer" entry

Stock Entry has no Stock Difference
Screenshot 2023-07-01 at 9 20 51 AM

But still has Stock Adjustment Entry
Screenshot 2023-07-01 at 9 20 44 AM

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/test_purchase_receipt.py
+++ b/erpnext/stock/doctype/purchase_receipt/test_purchase_receipt.py
@@ -1797,6 +1797,121 @@ class TestPurchaseReceipt(FrappeTestCase):
 
 		self.assertEqual(abs(data["stock_value_difference"]), 400.00)
 
+	def test_purchase_receipt_with_backdated_landed_cost_voucher(self):
+		from erpnext.controllers.sales_and_purchase_return import make_return_doc
+		from erpnext.stock.doctype.landed_cost_voucher.test_landed_cost_voucher import (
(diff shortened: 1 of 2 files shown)


rohitwaghchaure:
@Mergifyio backport develop

rohitwaghchaure:

PR #35975, opened 2023-07-03
feat: added proforma invoice doctype

RitvikSardana:


s-aga-r:
@RitvikSardana setup https://pre-commit.com/, use https://www.conventionalcommits.org/en/v1.0.0/

erpnext/selling/doctype/proforma_invoice/proforma_invoice.py
```diff
@@ -0,0 +1,143 @@
...
+
+	automatically_fetch_payment_terms = cint(
+		frappe.db.get_single_value("Accounts Settings", "automatically_fetch_payment_terms")
+	)
+	if automatically_fetch_payment_terms:
+		doclist.set_payment_schedule()
+
+	doclist.set_onload("ignore_price_list", True)
+
+	return doclist
```
deepeshgarg007: This function can be commodified using the one in `sales_order.py`. No need to make a new one

erpnext/selling/doctype/sales_order/sales_order.py
```diff
@@ -804,6 +804,166 @@ def postprocess(source, doc):
...
+		frappe.db.get_single_value("Accounts Settings", "automatically_fetch_payment_terms")
+	)
+	if automatically_fetch_payment_terms:
+		doclist.set_payment_schedule()
+
+	doclist.set_onload("ignore_price_list", True)
+
+	return doclist
+
+
```
deepeshgarg007, line 965: Duplicate code

deepeshgarg007 (changes requested):

PR #36001, opened 2023-07-04
fix: accepted warehouse and rejected warehouse can't be same

rohitwaghchaure:
Added validation 

Screenshot 2023-07-04 at 5 27 52 PM


erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
```diff
@@ -423,6 +423,7 @@
   {
    "fieldname": "rejected_warehouse",
    "fieldtype": "Link",
+   "ignore_user_permissions": 1,
```
barredterra, line 426: Why do we need to ignore user permissions here?

erpnext/controllers/buying_controller.py
```diff
@@ -437,17 +437,20 @@ def validate_purchase_return(self):
...
-						)
+				if not item.rejected_warehouse:
+					msg = (
+						f"Row #{item.idx}: Rejected Warehouse is mandatory against rejected Item {item.item_code}"
 					)
+					frappe.throw(_(msg), title=_("Mandatory Rejected Warehouse"))
+
+			if item.get("rejected_warehouse") and (item.get("rejected_warehouse") == item.get("warehouse")):
+				msg = f"Row #{item.idx}: Accepted Warehouse and Rejected Warehouse cannot be same"
+				frappe.throw(_(msg), title=_("Invalid Warehouse"))
```
barredterra: These error messages will be untranslatable. `_("{0} {1}").format(p1, p2)` was the correct syntax.

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
rohitwaghchaure: - Create the User Permission for the Warehouse and enable the IS Default
- Then try to create the Purchase Receipt
- System auto set the same value in the Accepted warehouse and rejected warehouse based on the User Permission
- On save, you will get the error that "Accepted warehouse and Rejected warehouse cannot be same", even though you have not set the rejected warehouse.

erpnext/controllers/buying_controller.py
rohitwaghchaure: Fixed

barredterra: I don't see any fix. 🤔  Basically, we cannot use f-strings for translatable messages. We need to first call the translation function, then call `.format()` on the result

erpnext/accounts/doctype/purchase_invoice_item/purchase_invoice_item.json
barredterra: Okay, so in oder to be able to select a _different_ rejected warehouse, we need to ignore any existing user permission. Makes sense.

erpnext/controllers/buying_controller.py
rohitwaghchaure: Fixed

PR #36008, opened 2023-07-05
refactor: Leave Application should not be in hook.py

kittiu:
I am not sure this is correct, but AFAIK, Leave Application should be in hrms app, is it not?

### DIFF
--- a/erpnext/hooks.py
+++ b/erpnext/hooks.py
@@ -83,7 +83,7 @@ update_website_context = [
 my_account_context = "erpnext.e_commerce.shopping_cart.utils.update_my_account_context"
 webform_list_context = "erpnext.controllers.website_list_for_contact.get_webform_list_context"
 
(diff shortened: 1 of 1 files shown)


barredterra:
@kittiu please also create a PR to https://github.com/frappe/hrms that adds this line there, so that we don't lose any functionality.

kittiu:
> @kittiu please also create a PR to https://github.com/frappe/hrms that adds this line there, so that we don't lose any functionality.

Thanks and done.

https://github.com/frappe/hrms/compare/develop...kittiu:hrms:patch-1

ruchamahabal:
> https://github.com/frappe/hrms/compare/develop...kittiu:hrms:patch-1


@kittiu Can you create the pull request on hrms?

kittiu:
https://github.com/frappe/hrms/pull/691
Sorry for previous mistake. Here you go.

PR #36024, opened 2023-07-07
feat: Unit Price Items

### LINKED ISSUE #36012: Unit Price Contract
In construction, it is common to make unit price contracts. For example, the **Quotation** and **Sales Order** indicate that one square meter of roofing costs $100, while the total number of square meters and the total amount remain unknown. Upon completion of the work, the result is measured and invoiced, for example 50 sqm of roofing resulting in a total of $5000.

This is currently not possible in ERPNext. In a **Quotation** or **Sales Order**, we cannot prevent the system from calculating a line amount and we cannot set the line quantity to 0. In a **Sales Invoice** we cannot freely increase the quantity due to the overbilling validation.

Add the possibility to mark a line item as "Is Unit Price". This should prevent the calculation of a line amount, thus also excluding this line item from the total. If "Is Unit Price" is selected, the line quantity can be fixed to 1. The overbilling validation should skip such lines.

**Additional context**

- https://uk.practicallaw.thomsonreuters.com/w-002-7055
- https://www.netsuite.com/portal/resource/articles/accounting/unit-price-contract.shtml
- https://de.wikipedia.org/wiki/Einheitspreisvertrag (German)

marination:
Closes https://github.com/frappe/erpnext/issues/36012 (Refer for more Info)

#### Concept
- **Unit Price Items** or **Unit Price Contracts** are a way to sell/agree upon a list of items whose quantity cannot be determined at the time of the sale but Rate/Unit Price can.
- The quantity is determined once the job is fully or partially done and the agreement is to bill as per the Unit Price

<details><summary> <b>Examples:</b> </summary>

-  **Sales Order for a Consultation firm:**
<table>
<th>Item</th>
<th>Rate</th>
<th>Qty</th>

<tr>
  <td>Consulting (per hour)</td>
  <td>$60</td>
  <td>?</td>
</tr>
</table>

-  **Sales Order for a Construction Company for Road Work:**
<table>
<th>Item</th>
<th>Rate</th>
<th>Qty</th>

<tr>
  <td>Labour (per hour)</td>
  <td>$60</td>
  <td>?</td>
</tr>
<tr>
  <td>Subcontracting/Partner Costs (per hour)</td>
  <td>$40</td>
  <td>?</td>
</tr>
</table>

- The Sales Order/Quotation in both cases is agreed upon based on the unit cost/rate. 
- The ? Quantities will be determined once the job is done **E.g.** The hours put in for road work will be updated once half or all of the work is done. Invoicing is done on the basis of actual quantities.

</details>

### Solution:

- Rows in a Quotation or Sales Order can be marked as **Is Unit Price Item**. These rows can have 0 qty as their qty is unknown
- Screenshot 2023-07-07 at 5 30 29 PM
- Screenshot 2023-07-07 at 5 32 34 PM

- **Why 0 qty ?** 1 Qty might be confusing as users might assume 1 qty is being sold in actuality. Also this makes handling billing, etc. easy with the least amount of code changes
- While mapping from SO > SI, the Unit Price row qty is mapped as-is. Since the final qty is indeterminate, no pending qty can be calculated
    

Users can decide to manage their Sales Orders in two ways

1. Keep updating the Qty in the Sales Order
    - Once a portion of work is done, users can update the qty in the SO > Create a SI for the updated qty
2. Keep the SO Qty 0
    - For those who might think constant updation SO is adding friction or are not particular about tracking via SO, can leave the quantities to be 0.
    - They can keep making Sales Invoices with actual quantities against the SO with 0 qty. The Billed % will always show as 0
    - **The Sales Order will have to be manually Closed**. Since there's no way to know when the Sales Order is truly Completed.

ToDo:

- [ ] Docs
- [ ] Tests

### DIFF
--- a/erpnext/controllers/accounts_controller.py
+++ b/erpnext/controllers/accounts_controller.py
@@ -797,10 +797,15 @@ class AccountsController(TransactionBase):
 		return gl_dict
 
 	def validate_qty_is_not_zero(self):
-		if self.doctype != "Purchase Receipt":
-			for item in self.items:
-				if not item.qty:
-					frappe.throw(_("Item quantity can not be zero"))
+		if self.doctype == "Purchase Receipt":
+			return
+
+		for item in self.items:
+			if not item.qty and not item.get("is_unit_price_item"):
+				frappe.throw(
+					_("Row #{0}: Item quantity can not be zero").format(item.idx),
+					title=_("Invalid Quantity"),
+				)
 
 	def validate_account_currency(self, account, account_currency=None):
 		valid_currency = [self.company_currency]
--- a/erpnext/controllers/status_updater.py
+++ b/erpnext/controllers/status_updater.py
@@ -543,7 +543,7 @@ class StatusUpdater(Document):
 				)[0][0]
 			)
 
-			per_billed = (min(ref_doc_qty, billed_qty) / ref_doc_qty) * 100
+			per_billed = (min(ref_doc_qty, billed_qty) / (ref_doc_qty or 1)) * 100
 
 			ref_doc = frappe.get_doc(ref_dt, ref_dn)
 
--- a/erpnext/selling/doctype/quotation_item/quotation_item.json
+++ b/erpnext/selling/doctype/quotation_item/quotation_item.json
@@ -48,6 +48,7 @@
   "base_net_amount",
   "pricing_rules",
   "stock_uom_rate",
+  "is_unit_price_item",
   "is_free_item",
   "is_alternative",
   "has_alternative_item",
@@ -661,12 +662,20 @@
    "label": "Has Alternative Item",
    "print_hide": 1,
    "read_only": 1
+  },
+  {
+   "default": "0",
+   "fieldname": "is_unit_price_item",
+   "fieldtype": "Check",
+   "in_list_view": 1,
+   "label": "Is Unit Price Item",
+   "print_hide": 1
   }
  ],
  "idx": 1,
  "istable": 1,
  "links": [],
- "modified": "2023-02-06 11:00:07.042364",
+ "modified": "2023-07-07 12:12:39.691139",
  "modified_by": "Administrator",
  "module": "Selling",
  "name": "Quotation Item",
--- a/erpnext/selling/doctype/sales_order/sales_order.py
+++ b/erpnext/selling/doctype/sales_order/sales_order.py
@@ -901,7 +901,9 @@ def make_sales_invoice(source_name, target_doc=None, ignore_permissions=False):
 		target.debit_to = get_party_account("Customer", source.customer, source.company)
 
 	def update_item(source, target, source_parent):
-		target.amount = flt(source.amount) - flt(source.billed_amt)
+		pending_amount = flt(source.amount) - flt(source.billed_amt)
+		target.amount = 0 if source.get("is_unit_price_item") else pending_amount
+
 		target.base_amount = target.amount * flt(source_parent.conversion_rate)
 		target.qty = (
 			target.amount / flt(source.rate)
@@ -939,7 +941,7 @@ def make_sales_invoice(source_name, target_doc=None, ignore_permissions=False):
 					"parent": "sales_order",
 				},
 				"postprocess": update_item,
-				"condition": lambda doc: doc.qty
+				"condition": lambda doc: (doc.qty or doc.get("is_unit_price_item"))
 				and (doc.base_amount == 0 or abs(doc.billed_amt) < abs(doc.amount)),
 			},
 			"Sales Taxes and Charges": {"doctype": "Sales Taxes and Charges", "add_if_empty": True},
--- a/erpnext/selling/doctype/sales_order/sales_order.py
+++ b/erpnext/selling/doctype/sales_order/sales_order.py
@@ -902,7 +902,8 @@ def make_sales_invoice(source_name, target_doc=None, ignore_permissions=False):
 
 	def update_item(source, target, source_parent):
 		pending_amount = flt(source.amount) - flt(source.billed_amt)
-		target.amount = 0 if source.get("is_unit_price_item") else pending_amount
+		# 0 Amount rows (as seen in Unit Price Items) should be mapped as it is
+		target.amount = pending_amount if flt(source.amount) else 0
 
 		target.base_amount = target.amount * flt(source_parent.conversion_rate)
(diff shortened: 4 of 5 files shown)


SvbZ3r0:
@marination world it be possible to extend this to Purchases too? Like we issue Unit Price PO and create invoices against it for every delivery.

bosue:
Nice, but do we really need that “Is unit price item“ checkbox?

Whenever qty. is zero, we can safely assume the price given is a unit price, can’t we?

marination:
@bosue 
> but do we really need that “Is unit price item“ checkbox?

Yes we do. For normal items there is a validation making sure that the qty is never 0. We need to know which rows to override this validation for. Removing the checkbox essentially allows any row to have 0 qty.

Historically users would sometimes forget to set the qty in a large list of items and this 0 values caused some issues during calculations (can't remember), so a validation was added to make sure that we aren't receiving 0 qty. We want to make sure that 0 was intentionally set and that the user understands that the row is a unit price item and didn't just make a mistake.

bosue:
I‘m so sorry for bothering you. Really love this feature but am still not convinced we need yet another checkbox adding even more clutter to the UI. Generally we need to aim for stripping down the UI as it is already too busy out-of-the-box.

> Historically users would sometimes forget to set the qty in a large list of items

Users will always forget one or the other thing. We may want to restrict Unit Price Items to particular items, customers or documents ([x] “Allow zero-quantity unit prices“) to disable the validation only for particular items, customers or documents. Most admins would want to allow it in Quotations only, others would allow it in Sales Orders as well.

We may even add a warning (to Sales Orders only): “For items x and y the quantity given is zero. Before submitting, please check the items should be listed by their unit price and not included in the total amount.”

> and this 0 values caused some issues during calculations

Well, then whatever issues come up would have to be fixed at their origin.

If others are okay with adding yet another checkbox to Quotation and Sales Order documents, please go ahead by any means! There’s more than enough we can do to streamline UX that doesn’t block this or any other new feature.

bosue:
@marination: I didn‘t want to discourage you or anybody at all. And still, just a little bit of pushback vs. the UI of it, and already it’s falling stale in absence of third opinions.

Again and more concretely, I will propose:

- adding a select “Allow zero-quantity Unit Price items“ with the options “No”, “Only in Quotations” and “In Quotations and Sales Orders” to the Transaction settings section in Selling Settings, default being “Only in Quotations.”
- leaving out the extraneous checkbox.
- adding a Status flag “Unit Price” — or “On-demand” — to any zero-qty item in any item list.
- ~adding a Status flag “0 qty” to any Sales Orders (not Quotations) containing a zero-qty item, as they will need to be manually closed.~

This will keep the UI slim, while still informing users of any possible mistake, and allowing admins control over the process. You raised the point that orders with zero-quantity items will obviously have to be manually closed. For some, this may add flexibility, for others it may add room for errors and manual checking routines. They can decide.

And replicating this to the buying side, per @SvbZ3r0, we would consider:
- adding a select “Allow zero-quantity Unit Price items“ with the options “No”, “Only in Material Requests” and “In Material Requests and Purchase Orders” to the Transaction settings tab in Buying Settings, default being “Only in Material Requests.”
- replicating the logic to the purchase side, leaving out the extraneous checkbox.
- adding a Status flag “Unit Price” — or “On-demand” — to any zero-qty item in any item list.
- ~adding a Status flag “0 qty” to any Purchase Orders (not Material Requests) containing a zero-qty item, as they will need to be manually closed.~

My two cents. If you still disagree, that‘s okay. Please go ahead and do it your way. It would be a shame if your PR got lost just because of arguing about the perfect implementation.

_edits:_
- @barredterra feel free to chime in and tell us how this would or wouldn’t fit your use case described in #36012? Appreciate it. 🙏
- incorporate the next comment’s recommendations.
- copyediting

bosue:
We might also want to differentiate between the various individual use case scenarios:

1. Sales order:
  - may be “on-demand” (on the customer’s demand) = Currently, we don’t know the exact quantity demanded, but once fulfilled, we will know it exactly, may enter the quantity and close the Sales order.
  - may be limited only by our own supply = We don’t know yet how much we can deliver, but the more, the better.
  - standing order = This happens regularly with varying quantities and we don’t want to go into details every time

2. Purchase order:
  - may be “on-demand” (on our own demand) = Currently, we don’t know the exact quantity we need, but once fulfilled, we will know it exactly, may enter the quantity and close the Purchase order.
  - may be limited only by the supplier’s supply = We don’t know yet how they can deliver, but the more, the better.
  - standing order = This happens regularly with varying quantities and we don’t want to go into details every time
  
To put back together all these slightly different cases, I think all use cases can still be sufficiently approximated by the term “On-demand.” On-demand is a well established and sufficiently broad concept that would still be associated with every single of these use cases, and easily understood.

Particularly the standing order use case doesn’t seem to me adequately covered either way, neither today nor with the PR. With the PR, at least this is a workaround.

The supply-limited use case seems less important. We don’t know how much we (or the supplier) can deliver, but probably we _do_ know how much they (or we) want?

So the “demand to be specified” use cases seem the dominant factor here.

Until we (may) have more specific implementations of all the individual use-cases, I think we should call this an “On-demand purchase order” resp. “On-demand sales order”.

The only thing I’m still slightly unsure about is the flag in the Sales Orders resp. Purchase Orders List. Do we really need a flag there? There will always be a fulfillment date given as the respective date fields are required. So if forgotten, at some point it should become tagged “Overdue”.

Or will it actually? In the case of an overdue Sales Order, yes. In the case of an overdue Purchase Order, no. See #37308. But it should.

Thus, while it may not be straightforward to specifically filter by “On-demand” (0 qty.), the overdue S / P Orders will have to be looked into anyway. Thus I think we should flag “on-demand” items within a S / P Order, but may go away without flagging “On-demand” S / P Orders within their list. For now.

bosue:
At least my very last comment was really hard to read, even for myself a few days later now. Many words about few lines of code, and while I can‘t look into anybody‘s head, I‘m wondering whether you‘re annoyed or frustrated or lost interest or something…

Basically it‘s all about the UI, while the actual feature is undisputed.

I don’t want to hijack your feature, but if you don‘t mind I‘ll create a new PR with my alternative approach, and then we (or the maintainers) may choose or demand further changes.

I came to learn that sketches or screen recordings or actual code may be better than many words.

bosue:
Work is continuing here: #37865. The individual steps may take a while, but eventually we‘ll get the feature implemented...

PR #36027, opened 2023-07-07
fix: don't allow negative rate

### LINKED ISSUE #31533: Creating Sales/Purchase Return with Negative Rate
### Overview
When creating a return with a negative rate, the transaction is treated as an invoice instead of a credit.
This can happen if operator gets confused when adding negative quantity and also adds a negative rate.

### Observed Behavior
The transaction is submitted as an invoice.

### Expected Behavior
A validation error should be thrown similar to when positive qty is entered for a return.

### Module

accounts, buying, selling

### Version

frappe: develop
erpnext: develop

### Installation method

manual install

_No response_

dj12djdjs:
closes: https://github.com/frappe/erpnext/issues/31533

### Context
If rate is **negative**, and is_return is `True` the sales invoice will be posted to the GL as a debit note, instead of a credit note.

### DIFF
--- a/erpnext/accounts/doctype/sales_invoice/test_sales_invoice.py
+++ b/erpnext/accounts/doctype/sales_invoice/test_sales_invoice.py
@@ -3371,6 +3371,13 @@ class TestSalesInvoice(unittest.TestCase):
 
 		set_advance_flag(company="_Test Company", flag=0, default_account="")
 
+	def test_sales_return_negative_rate(self):
+		si = create_sales_invoice(is_return=1, qty=-2, rate=-10, do_not_save=True)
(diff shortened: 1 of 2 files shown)


deepeshgarg007:
@dj12djdjs one test case failing, can you please fix that?

dj12djdjs:
@deepeshgarg007 CI is resolved.

ruthra-kumar:
LGTM

peterafahmy:
Hi, I have a case where i'm selling airline tickets. When there is a commission, I have it entered as a negative number in the purchase invoice to the airline. Now I can't do this anymore because of this fix. 
Can you help me bypass this? My erpnext is on a frappe server.

PR #36034, opened 2023-07-09
Revert "perf: improve item wise register reports (backport #35908)"

deepeshgarg007:
Reverts frappe/erpnext#35912


deepeshgarg007:

PR #36051, opened 2023-07-10
feat: Provision to auto create Exchange Rate Revaluation

ruthra-kumar:
Provision for auto creation of Exchange Rate Revaluation on a daily/weekly frequency.

Company master will have new section called `Exchange Rate Revaluation Settings`, through which this can be enabled.
Screenshot 2023-07-10 at 10 54 55 AM

Documentation can be found [here](https://docs.erpnext.com/docs/v14/user/manual/en/accounts/exchange-rate-revaluation#3-automate-exchange-rate-revaluation-creation).


umairsy:
Can we add some field description for ERR Journals?

PR #36058, opened 2023-07-10
fix: Ambiguous column error while submitting stock entry

mohsinalimat:
1052, "Column 'qty' in field list is ambiguous in latest version-14

error


s-aga-r (changes requested):
Rewrite the query in [query builder](https://frappeframework.com/docs/v14/user/en/api/query-builder)

mohsinalimat:
Can I do?

mohsinalimat:
I don't know how to do this.

Issue #34764, opened 2023-04-06
Need General Ledger Format with invoice details


smbcatalyst:
We require a general ledger format in details with invoice details. Currently the general ledger shows only Payable Amount in supplier case. When we do ledger checking with supplier details, our general ledger format is not showing invoice amount, GST amount, TDS, etc in supplier case.

e.g. If total is Rs. 100, GST is Rs. 18 and TDS deducted in that invoice is Rs. 10, then ERPNext shows Rs. 108 in general ledger. But the supplier details shows Rs. 118. For every transactions user needs to open each invoice and cross check the details. It will take hours to cross check the details of single supplier.

The purchase register report is the correct format for requirement but as it is purchase related so it is not considering payment entries.
We need a report where we can get the bifurcation as above with payment entries. 
[General_Ledger.xlsx](https://github.com/frappe/erpnext/files/11169974/General_Ledger.xlsx)

1: Current Format
![image](https://user-images.githubusercontent.com/113963271/230369359-bf584a0d-11a4-459d-82e4-ecb7fd49aaba.png)

2. Expected Format
![image](https://user-images.githubusercontent.com/113963271/230374943-fef5392b-591a-44bd-9416-394e498eb4d0.png)

3. If Possible Expected Format
![image](https://user-images.githubusercontent.com/113963271/230376209-f3c75534-7623-44cb-8305-8bcc551fed7e.png)

deepeshgarg007:
Fixed via https://github.com/frappe/erpnext/pull/36069

PR #36069, opened 2023-07-10
feat: include payments in purchase / sales register report

GursheenK:
Added the feature to enable fetching Payment Entries & Journal Entries along with Purchase/Sales Invoice details in Purchase/Sales Register Reports.

The `Include Payments` checkbox alters the flow in the following manner - 

1. When enabled, **all Invoice and Payment details** are fetched with their corresponding Tax Account details.
<br>
Screenshot 2023-07-14 at 11 23 30 AM
<br>
<br>

3. When disabled, **only Invoice details** are fetched with their corresponding Tax Account details.

Issue #34764

`no-docs`


deepeshgarg007 (changes requested):
This can be enhanced further to show a holistic ledger view for Supplier and Customer

- Maybe rename the check to "Show Ledger View" instead of "Include Payments"
- Show opening balances on the first row
- Few columns might have to be hidden and rearranged and added as well
        - A balance column should be added 
        - Show invoice and  payment amount in separate columns (Invoice and Payments)
        - Hide columns like Supplier/Customer Group, Purchase Order, Receipt, Currency, Rounded total, and Outstanding Amount which are common for a specific party
- Skip the total row for this view (there is a parameter that can be passed for this)

erpnext/accounts/report/utils.py
```diff
@@ -179,3 +187,193 @@ def get_values_for_columns(report_columns, report_row):
...
+				if frappe.get_cached_value("DocType", dimension.document_type, "is_tree"):
+					filters[dimension.fieldname] = get_dimension_with_children(
+						dimension.document_type, filters.get(dimension.fieldname)
+					)
+				fieldname = dimension.fieldname
+				query = query.where(parent_doc[fieldname] == filters.fieldname)
+	return query
+
+
+def get_opening_row(party_type, party, from_date, company):
```
deepeshgarg007, line 369: Also, check for the party in this (query), multiple parties use the same account so this will fetch balance for all parties. Also add a check for cancelled entries, should not pull any cancelled entries

erpnext/accounts/report/sales_register/sales_register.py
```diff
@@ -142,83 +187,150 @@ def get_columns(invoice_list, additional_table_columns):
...
+				"label": _("Warehouse"),
+				"fieldname": "warehouse",
+				"fieldtype": "Link",
+				"options": "Warehouse",
+				"width": 100,
+			},
+			{"fieldname": "currency", "label": _("Currency"), "fieldtype": "Data", "width": 80},
+		]
+	else:
+		columns += [
```
deepeshgarg007, line 266: Convert these columns to the new format as well?

erpnext/accounts/report/purchase_register/purchase_register.py
```diff
@@ -75,56 +103,104 @@ def _execute(filters=None, additional_table_columns=None):
...
-		_("Project") + ":Link/Project:80",
-		_("Bill No") + "::120",
-		_("Bill Date") + ":Date:80",
-		_("Remarks") + "::150",
-		_("Purchase Order") + ":Link/Purchase Order:100",
-		_("Purchase Receipt") + ":Link/Purchase Receipt:100",
-		{"fieldname": "currency", "label": _("Currency"), "fieldtype": "Data", "width": 80},
-	]
+	if not include_payments:
+		columns += [
```
deepeshgarg007, line 182: Convert to new column format?

erpnext/accounts/report/utils.py
```diff
@@ -179,3 +187,193 @@ def get_values_for_columns(report_columns, report_row):
...
+		)
+		.where((pe.party == filters.get(args.party)) & (pe.paid_to.isin(args.party_account)))
+		.orderby(pe.posting_date, pe.name, order=Order.desc)
+	)
+	query = get_conditions(filters, query, [pe], payments=True)
+	payment_entries = query.run(as_dict=True)
+	return payment_entries
+
+
+def get_conditions(filters, query, docs, payments=False):
```
deepeshgarg007: Instead of docs, we can make this more explicit, which would be easier to read and understand. Instead of `docs` parameter, we can pass the doctype which would be either Sales, Purchase Invoice, or Payment or Journal entry.
child_doc can be an optional parameter and docs[2] (Sales Invoice Payment) can be hard coded since its already inside the Sales Invoice if block

deepeshgarg007 (changes requested):

erpnext/accounts/report/utils.py
```diff
@@ -179,3 +187,198 @@ def get_values_for_columns(report_columns, report_row):
...
+			ConstantColumn("Opening").as_("account"),
+			Sum(gle.debit).as_("debit"),
+			Sum(gle.credit).as_("credit"),
+			(Sum(gle.debit) - Sum(gle.credit)).as_("balance"),
+		)
+		.where(
+			(gle.account.isin(party_account))
+			& (gle.party == party)
+			& (gle.posting_date < from_date)
+			& (gle.docstatus == 1)
```
deepeshgarg007: check for field `is_cancelled` = 0, docstatus won't help here

deepeshgarg007:
"Opening" label is not visible in the Sales Register report. While viewing the Purchase Register report saw a mismatch in the balance column 

image

GST Itemised Sales and Purchase Register also affected by this

deepeshgarg007:
image

No payment entries were fetched even though there were payments in the specified period

PR #36092, opened 2023-07-11
fix: paid_amount when the group is mode of payment

HarryPaulo:
OBS: **R$** is a currency

**Problem**: If i create a invoice with total amount of R$ 100, paying this with two mode of payments:
**Cash / Money:** R$ 80
**Credit Card:** R$ 20

Using the report **POS Register**, set the filter "group_by" based on **payment method**, column Paid Amount show R$ 100 for both mode of payments, this is wrong. 

image

**>> Now With the fix <<**

Now whent we group based on **payment method**, the amount will be based on mode of payment **Cash / Money:** R$ 80 and **Credit Card:** R$ 20

image

If we have change, this amount will be discounted on CASH type.

**Example 1:** one sale with amount R$ 90.
image
image

**Example 2:** 
This situation is not common, and is not allowed in many establishments, but it is the following, R$ 100 was sold, with payment of R$ 120 on the credit card, generating a change of R$ 20, the change will be made and deducted from the cash payment method.

image
image

As the amount came out of the drawer to give change, in this case it will be negative, deducting from the general balance of the payment in "Cash / Money"

### DIFF
--- a/erpnext/accounts/report/pos_register/pos_register.py
+++ b/erpnext/accounts/report/pos_register/pos_register.py
@@ -50,20 +50,20 @@ def get_pos_entries(filters, group_by_field):
 	order_by = "p.posting_date"
 	select_mop_field, from_sales_invoice_payment, group_by_mop_condition = "", "", ""
 	if group_by_field == "mode_of_payment":
-		select_mop_field = ", sip.mode_of_payment"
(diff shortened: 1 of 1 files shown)


HarryPaulo:
waiting for reviewers

deepeshgarg007 (approved):

PR #36098, opened 2023-07-12
ci: regen release notes with GH API

ankush:


.github/workflows/release_notes.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  regen-notes:
+    name: 'Regenerate release notes'
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Update notes
+      run: |
+        NEW_NOTES=$(gh api --method POST -H "Accept: application/vnd.github+json"  /repos/frappe/frappe/releases/generate-notes  -f tag_name=$RELEASE_TAG | jq -r '.body' | sed  -E '/^\* (chore|ci|test|docs|style)/d' )
```
ruchamahabal, line 32: Shouldn't it be /repos/frappe/**erpnext**/releases/generate-notes here instead of frappe repo?

Issue #30482, opened 2022-03-29
Add Accounting Dimension  to Profitability Analysis


thida-bcn:
At Profitability Analysis Report, we would like to get Based on 'Accounting Dimension'
Example,
If I had added Business Unit as Accounting Dimension, we would like to choose 'Business Unit' as Based on
Thanks
![image](https://user-images.githubusercontent.com/62122640/160653408-4956fa39-673a-4ec9-b66b-54584f28378f.png)

thida-bcn:
I supposed Accounting Dimension should be able to choose at Profitability Analysis

deepeshgarg007:
Fixed Via: https://github.com/frappe/erpnext/pull/36107

PR #36107, opened 2023-07-12
feat: filtering based on accounting dimensions in profitability analysis

GursheenK:
Option in **Profitability Analysis Report** that allows filtering results based on any **Accounting Dimensions** created.

Removed validation for `based_on` filter since default value is defined as **Cost Center**.

Issue #30482

`no-docs`

### DIFF
--- a/erpnext/accounts/report/profitability_analysis/profitability_analysis.js
+++ b/erpnext/accounts/report/profitability_analysis/profitability_analysis.js
@@ -18,7 +18,15 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
 				"fieldtype": "Select",
 				"options": ["Cost Center", "Project", "Accounting Dimension"],
 				"default": "Cost Center",
-				"reqd": 1
+				"reqd": 1,
+				"on_change": function(query_report){
+					let based_on = query_report.get_values().based_on;
+					if(based_on!='Accounting Dimension'){
+						frappe.query_report.set_filter_value({
+							accounting_dimension: ''
+						});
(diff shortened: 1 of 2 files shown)


erpnext/accounts/report/profitability_analysis/profitability_analysis.js
```diff
@@ -16,9 +16,30 @@ frappe.require("assets/erpnext/js/financial_statements.js", function() {
...
+				"on_change": function(query_report){
+					let based_on = query_report.get_values().based_on;
+					if(based_on!='Accounting Dimension'){
+						frappe.query_report.set_filter_value({
+							accounting_dimension: ''
+						});
+					}
+				}
+			},
+			{
```
deepeshgarg007, line 31: No need for a separate filter, this can be combined with the cost center and project filter

erpnext/accounts/report/profitability_analysis/profitability_analysis.py
```diff
@@ -60,7 +62,9 @@ def get_data(accounts, filters, based_on):
 		filters.get("company"),
 		filters.get("from_date"),
 		filters.get("to_date"),
-		based_on,
+		based_on
+		if based_on != "accounting_dimension"
+		else filters.accounting_dimension.replace(" ", "_").lower(),
```
deepeshgarg007: Use fieldname from the Accounting Dimension doctype instead of converting to lowercase

deepeshgarg007 (changes requested):

PR #36116, opened 2023-07-13
feat: add local holidays

barredterra:
Add controls for easy addition of local holidays. Dates are fetched from https://pypi.org/project/holidays/

https://github.com/frappe/erpnext/assets/14891507/dbdb25a4-a4d7-4789-808a-71e1b7c920b1

Docs: https://docs.erpnext.com/app/wiki-page-patch/efbd6143b1


erpnext/setup/doctype/holiday_list/holiday_list.py
```diff
@@ -21,25 +24,59 @@ def validate(self):
...
+	def get_supported_countries(self):
+		return list_supported_countries()
+
+	@frappe.whitelist()
+	def get_local_holidays(self):
+		if not self.country:
+			throw(_("Please select Country"))
+
+		existing_holidays = self.get_holidays()
+		system_language = frappe.db.get_single_value("System Settings", "language")
```
ruchamahabal:
```suggestion
		system_language = frappe.db.get_single_value("System Settings", "language", cache=True)
```

erpnext/setup/doctype/holiday_list/holiday_list.json
```diff
@@ -1,480 +1,166 @@
...
+  {
+   "fieldname": "country",
+   "fieldtype": "Select",
+   "label": "Country"
+  },
+  {
+   "depends_on": "country",
+   "fieldname": "subdivision",
+   "fieldtype": "Select",
+   "label": "Subdivision"
```
ruchamahabal: Can we use the Autocomplete field type here? It's just easier to search

erpnext/setup/doctype/holiday_list/holiday_list.py
```diff
@@ -21,25 +24,59 @@ def validate(self):
...
+			if d in existing_holidays:
+				continue
+
+			self.append("holidays", {"description": _(self.weekly_off), "holiday_date": d, "weekly_off": 1})
+
+		self.sort_holidays()
+
+	@frappe.whitelist()
+	def get_supported_countries(self):
+		return list_supported_countries()
```
ruchamahabal: Haven't gone through the API docs yet but is there any way we could populate country names as labels instead of codes?

erpnext/setup/doctype/holiday_list/holiday_list.js
```diff
@@ -6,13 +6,39 @@ frappe.ui.form.on("Holiday List", {
 		if (frm.doc.holidays) {
 			frm.set_value("total_holidays", frm.doc.holidays.length);
 		}
+
+		frm.call("get_supported_countries").then(r => {
+			frm.subdivisions_by_country = r.message;
+			frm.set_df_property("country", "options", Object.keys(r.message));
```
ruchamahabal: Keep the first option blank in case someone wants to reset it (both country & subdivision)? I think using Autocomplete will fix that.

erpnext/setup/doctype/holiday_list/holiday_list.py
barredterra: `cache` is `True` by default, but also not critical for a method that is usually used only once, IMO.

erpnext/setup/doctype/holiday_list/holiday_list.json
barredterra: Is there any docs on the Autocomplete field? Can't figure out how it works. Also, I don't like the UX of it very much... I'm always annoyed when I have to use these in reports. 😅

barredterra:
Country names are localized now: 

https://github.com/frappe/erpnext/assets/14891507/bb707abc-b70d-47f5-9838-3bc9760a4ece

erpnext/setup/doctype/holiday_list/holiday_list.json
ruchamahabal: I am not sure but I think you might be talking about this Multiselect List field. Yeah, the UX is off:

image


[Autocomplete](https://github.com/frappe/frappe/pull/15794) is different. Just an alternative to select field with label & value.

image

I think the docs are missing. Will add.

barredterra:
First option is now blank.

Select Country:

Bildschirmfoto 2023-07-14 um 12 35 21

Select Subdivision:

Bildschirmfoto 2023-07-14 um 12 35 32

erpnext/setup/doctype/holiday_list/holiday_list.json
ruchamahabal: Usage reference here: https://github.com/frappe/frappe/blob/5c4f6c64c62bcf1d8e7b8f7282e760c9e5e47830/frappe/core/doctype/document_naming_settings/document_naming_settings.js#L16-L29

You will have to change the fieldtype to Autocomplete and then change:

`frm.set_df_property("country", "options", countries);`

to

`frm.fields_dict.country.set_data(countries);`

barredterra:
Done @ruchamahabal

ruchamahabal (approved):

PR #36125, opened 2023-07-14
fix: apply discount on item after applying price list

anandbaburajan:
When a user changes the exchange rate for a currency in a SI, the rate of the items are automatically updated. But the items which have a discount applied get an incorrect rate because the discount isn't applied again after the rate update. This is due to the `apply_price_list` function (which is called from `conversion_rate` when the exchange rate changes) not applying discount once the rate changes, so fixed that.


thida-bcn:
> When a user changes the exchange rate for a currency in a SI, the rate of the items are automatically updated. But the items which have a discount applied get an incorrect rate because the discount isn't applied again after the rate update. This is due to the `apply_price_list` function (which is called from `conversion_rate` when the exchange rate changes) not applying discount once the rate changes, so fixed that.

there was problem with pricing rule, before update version discount apply well
after updating patch screen refresh frequently and discount amount change

PR #36132, opened 2023-07-14
fix: delivery note connection in sales invoice

anandbaburajan:
Sales Invoice connections doesn't reflect the connected Delivery Note, so fixed that.


anandbaburajan:
Closing because: https://github.com/frappe/erpnext/pull/23570

PR #36133, opened 2023-07-14
perf: index in `Item` and `Item Variant Attribute`

s-aga-r:
**Source / Ref**: ISS-23-24-02011

Add Index for columns used in [this](https://github.com/s-aga-r/erpnext/blob/9d27a25e5f2335386402f96f83a10729695b20c8/erpnext/e_commerce/variant_selector/item_variants_cache.py#L67-L77) query.


aakvatech:
Would you also share anonymized contents of the ISS-23-24-02011 as part of the description?

s-aga-r:
> Would you also share anonymized contents of the ISS-23-24-02011 as part of the description?

The ISS-23-24-02011 is for our reference. I don't think this PR needs any description (?) It's just a simple change to enable Index for some columns used in a query. If you're asking about performance improvements that will be negligible for most of the users as this user has ~1 Cr Item Variant Attributes.

PR #36142, opened 2023-07-14
fix: HSN code not showing up in GST itemised register reports

anandbaburajan:
This https://github.com/frappe/erpnext/pull/36072 revert didn't work for some reason, so doing again.


anandbaburajan:
@mergify backport version-13

PR #36162, opened 2023-07-17
fix: incorrect Reserved Qty for Production Plan in BIN for the multi-uom case

rohitwaghchaure:
There is an incorrect Reserved Quantity for the Production Plan in the BIN for the multi-uom case.
Screenshot 2023-07-17 at 3 34 18 PM

**After Fix**

image


rohitwaghchaure:

PR #36163, opened 2023-07-17
chore: use consistent quotes

sagarvora:
To avoid unnecessary version creation:

image


aakvatech:
Considerable reduction of version table size for Sales Invoice changes

sagarvora:
> Considerable reduction of version table size for Sales Invoice changes

Perhaps we should also write a patch for existing versions like these?

PR #36167, opened 2023-07-17
fix: rounding of percentage fields

barredterra:
Always round with precision of 2

- `per_delivered`
- `per_picked`
- `per_billed`
- `per_received`

These are of type "Percentage", which usually is a value between 0 and 100. Something that is delivered/picked/billed/received > 99.99% should probably be marked as done.

In the past we used a precision of 6 in some places. Orders would have to be very large for 0.004 % to be valuable enough for the order to to stay open.

### DIFF
--- a/erpnext/buying/doctype/purchase_order/purchase_order.js
+++ b/erpnext/buying/doctype/purchase_order/purchase_order.js
@@ -62,7 +62,7 @@ frappe.ui.form.on("Purchase Order", {
 	get_materials_from_supplier: function(frm) {
 		let po_details = [];
 
-		if (frm.doc.supplied_items && (frm.doc.per_received == 100 || frm.doc.status === 'Closed')) {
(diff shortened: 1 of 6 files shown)


rohitwaghchaure:
@barredterra Using the precision as 2 will also cause the issue, some clients want 6 and some 2. I guess we need a configuration to set this value.

barredterra:
They really wish for 0.4 parts in a million to keep their orders open? Should we just use the precision setting of the individual fields then?

ankush:
@barredterra can you share an example where it doesn't tip over to 100 when it's very near 100? 


Perhaps sparing use of `Decimal` type here can solve the problem.  https://docs.python.org/3/library/decimal.html

barredterra:
For example, after three payments/deliveries/receipts of 33.333% a transaction would still be unresolved because `flt(33.333 * 3) != 100` while `flt(33.333 * 3, 2) == 100`

PR #36196, opened 2023-07-19
fix: show tax withholding category details for customers

GursheenK:
**Problem**
- The TDS Payable Monthly Report fetches tax details of both Purchase Invoices and Sales Invoices. However, the party details and total amount(without tax) of only purchase invoices is displayed correctly.
- The results can only be filtered by Supplier.

Screenshot 2023-07-19 at 6 26 24 PM

<br>
<br>

**Solution**
- Added logic to fetch and show Customer details and the total amount(without tax) for Sales Invoices.
- Added a mandatory filter for Party Type which allows filtering results for both Suppliers and Customers.

Screenshot 2023-07-24 at 12 22 42 PM

Screenshot 2023-07-24 at 12 22 59 PM

<br>
<br>


erpnext/accounts/report/tax_withholding_details/tax_withholding_details.js
```diff
@@ -12,10 +12,27 @@ frappe.query_reports["TDS Payable Monthly"] = {
...
 		{
-			"fieldname":"supplier",
-			"label": __("Supplier"),
-			"fieldtype": "Link",
-			"options": "Supplier",
+			"fieldname":"party_type",
+			"label": __("Party Type"),
+			"fieldtype": "Select",
+			"options": ["Supplier", "Customer"],
+			"reqd": 1,
```
deepeshgarg007, line 19: Set the default value as "Supplier"

erpnext/patches.txt
```diff
@@ -262,6 +262,7 @@ erpnext.patches.v14_0.update_reference_due_date_in_journal_entry
 erpnext.patches.v15_0.saudi_depreciation_warning
 erpnext.patches.v15_0.delete_saudi_doctypes
 erpnext.patches.v14_0.show_loan_management_deprecation_warning
+execute:frappe.rename_doc("Report", "TDS Payable Monthly", "Tax Withholding Details", force=True)
```
barredterra, line 265: Can we move this to `[post_model_sync]`? Seems to break my migration otherwise (`rename_doc` triggers a query on `tabDocType.is_virtual`, which does not exists at that time).

PR #36205, opened 2023-07-20
Update utils.py

cvkmohan:
Purchase Order does not have a column called `posting_date` - therefore the name generation will fail. This is the lesser evil way of solving it - putting it in code.  Renaming the `transaction_date` column in the `Purchase Order` is a bigger evil.

> Screenshots/GIFs


ruthra-kumar:
This has been addressed in https://github.com/frappe/erpnext/pull/36229

PR #36208, opened 2023-07-20
fix: made item or warehouse filter mandatory

rohitwaghchaure:
If Stock Ledger Entries count is more than 10,000 then user should select the filter of Item Code or Warehouse to view the Batch Wise Balance History Report.

Screenshot 2023-07-20 at 12 59 38 PM

**Why Mandatory?**

Currently, users are facing performance issues and are unable to generate the report due to the lack of a filter on large data.

### DIFF
--- a/erpnext/stock/report/batch_wise_balance_history/batch_wise_balance_history.py
+++ b/erpnext/stock/report/batch_wise_balance_history/batch_wise_balance_history.py
@@ -10,10 +10,17 @@ from pypika import functions as fn
 
 from erpnext.stock.doctype.warehouse.warehouse import apply_warehouse_filter
 
+SLE_COUNT_LIMIT = 10_000
+
 
 def execute(filters=None):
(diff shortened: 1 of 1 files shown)


erpnext/stock/report/batch_wise_balance_history/batch_wise_balance_history.py
```diff
@@ -10,11 +10,18 @@
...
 from erpnext.stock.doctype.warehouse.warehouse import apply_warehouse_filter
 
+SLE_COUNT_LIMIT = 10_000
+
 
 def execute(filters=None):
 	if not filters:
 		filters = {}
 
+	sle_count = frappe.db.count("Stock Ledger Entry", {"is_cancelled": 0})
```
ankush, line 20: This query will likely do full table scan because `is_cancelled` is not indexed.

PR #36220, opened 2023-07-20
fix: Patch Dunnings after accounts were frozen

marination:
Fixes patch in https://github.com/frappe/erpnext/pull/35689

- Consider "Accounts Frozen Until" and Period Closing Voucher: Reverse Unresolved old Dunning entries after one of these dates (whichever is more recent)

### DIFF
--- a/erpnext/patches/v14_0/single_to_multi_dunning.py
+++ b/erpnext/patches/v14_0/single_to_multi_dunning.py
@@ -7,7 +7,15 @@ def execute():
 	frappe.reload_doc("accounts", "doctype", "overdue_payment")
 	frappe.reload_doc("accounts", "doctype", "dunning")
 
-	all_dunnings = frappe.get_all("Dunning", filters={"docstatus": ("!=", 2)}, pluck="name")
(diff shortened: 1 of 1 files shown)


erpnext/patches/v14_0/single_to_multi_dunning.py
```diff
@@ -7,7 +7,15 @@ def execute():
 	frappe.reload_doc("accounts", "doctype", "overdue_payment")
 	frappe.reload_doc("accounts", "doctype", "dunning")
 
-	all_dunnings = frappe.get_all("Dunning", filters={"docstatus": ("!=", 2)}, pluck="name")
+	filters = {"docstatus": ("!=", 2)}
```
marination, line 11: Shouldn't this be ?
```suggestion
	filters = {"docstatus": ("==", 1)}
```

barredterra: No, we also want to populate the overdue payments in draft Dunnings.

But line 52 could be: `if dunning.status != "Resolved" and dunning.docstatus == 1`

PR #36238, opened 2023-07-22
fix: POS closing with item name 

RitvikSardana:
**Problem :**
When u try to close a POS in POS Invoice Merge Log doctype for an item without the item code.
We get the following error :

telegram-cloud-photo-size-5-6185780954363180809-y

**Reason :** 
The doctype uses the function : 

which is present in Sales Invoice, this function is used when we update the stock.

**Solution :**  
Removed the above function, along with this in [sales_invoice.js]

Screenshot 2023-07-22 at 1 56 30 PM


deepeshgarg007:
@RitvikSardana Don't push comments. Also, remove the commented functions if they are not being used anywhere. Add a test case for this scenario in POS

RitvikSardana:
@deepeshgarg007, made the required changes

deepeshgarg007 (approved):

PR #36245, opened 2023-07-24
build(deps): update python-plaid 7.2.1 -> 14.1.0

blaggacao:

Plaid has meanwhile switch to an autogenerated openapi-based sdk.
Old versions of `plaid-python` are [no longer packages in `nixpkgs`](https://search.nixos.org/packages?channel=23.05&show=python311Packages.plaid-python&from=0&size=50&sort=relevance&type=packages&query=plaid-python) cc @teutat3s

See [migration guide in change log](https://github.com/plaid/plaid-python/blob/master/CHANGELOG.md#client-initialization)

### DIFF
--- a/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
+++ b/erpnext/erpnext_integrations/doctype/plaid_settings/plaid_connector.py
@@ -1,11 +1,13 @@
 # Copyright (c) 2018, Frappe Technologies Pvt. Ltd. and contributors
 # For license information, please see license.txt
 
+import json
+
 import frappe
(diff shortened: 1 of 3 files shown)


ankush:
@blaggacao quite hard to approve these changes without test setup. Did you test these out locally?

ref: https://plaid.com/docs/sandbox/test-credentials/ 


---

Unrelated: are you sure all these efforts to package for NixOS are worth it? Why not somehow package docker-based installs?

updating all these legacy things without proper tests is really hard and it will be even harder to maintain it constantly in future :sweat_smile:

PR #36247, opened 2023-07-24
fix(startup): coa drill down

blaggacao:

`erpnext/public/js/setup_wizard.js` does not provide positional argument `is_root` when calling `get_coa` from  `erpnext/accounts/utils.py`.

As a consequence, users can't drill down into accounts in the chart preview during the initial setup wizard.

To not change the api, setting `is_root=None`. However, it is not even used in the function. Please remove at will.

### DIFF
--- a/erpnext/accounts/utils.py
+++ b/erpnext/accounts/utils.py
@@ -1117,7 +1117,7 @@ def parse_naming_series_variable(doc, variable):
 
 
 @frappe.whitelist()
-def get_coa(doctype, parent, is_root, chart=None):
+def get_coa(doctype, parent, is_root=None, chart=None):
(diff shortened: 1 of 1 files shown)


deepeshgarg007:
@blaggacao seems to be working fine for me

image

Please explain and then reopen the PR

blaggacao:
@deepeshgarg007 Isn't it obvious? (I assumed it was :smile: )

https://github.com/frappe/erpnext/blob/c0642cf528e2054192f5d16dd9d5c1a8162069cf/erpnext/public/js/setup_wizard.js#L181-L185

does not pass `is_root` which is why this code path fails without a default. You just need to trigger the right code path :smile: (which is in the setup wizard, here).

PR #36254, opened 2023-07-24
fix: Remove auto repeat (subscription) button from Sales and Purchase documents

deepeshgarg007:
- These are mostly transactions and are mostly unique and have a lot of variance, doesn't make much sense to put them on auto-repeat
- The button label which said "Subscription" was also misleading as there is a Subscription doctype already and many users thought this will lead to the creation of a Subscription for the Sales Invoice

### DIFF
--- a/erpnext/stock/doctype/purchase_receipt/purchase_receipt.js
+++ b/erpnext/stock/doctype/purchase_receipt/purchase_receipt.js
@@ -187,12 +187,6 @@ erpnext.stock.PurchaseReceiptController = class PurchaseReceiptController extend
 				}
 				cur_frm.add_custom_button(__('Retention Stock Entry'), this.make_retention_stock_entry, __('Create'));
 
-				if(!this.frm.doc.auto_repeat) {
-					cur_frm.add_custom_button(__('Subscription'), function() {
-						erpnext.utils.make_subscription(me.frm.doc.doctype, me.frm.doc.name)
-					}, __('Create'))
-				}
-
 				cur_frm.page.set_inner_btn_group_as_primary(__('Create'));
 			}
 		}
--- a/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
+++ b/erpnext/accounts/doctype/purchase_invoice/purchase_invoice.js
@@ -97,12 +97,6 @@ erpnext.accounts.PurchaseInvoice = class PurchaseInvoice extends erpnext.buying.
 				cur_frm.add_custom_button(__('Return / Debit Note'),
 					this.make_debit_note, __('Create'));
 			}
-
-			if(!doc.auto_repeat) {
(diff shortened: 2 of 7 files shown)


bosue:
I see at this point, this is about to being merged.

However I would dare to disagree that there is no use in auto-repeating orders and sales.
In our current project I‘m going to implement continuing obligations such as rent and financial lease.

Certainly, in most cases we may peruse the actual subscription functionality, issue a single long-term rental bill once, not monthly bills, and accomplish the rest by monthly journal entries and direct debit transactions.
But there are cases where we do need a monthly invoice.

Same is conceivable for orders, though not needed here.

Instead of removing the functionality, IMO we should properly rename the button to „Auto-repeat“ rather than „Subscription“ and be (almost) set.

Almost, because the argument that some (at least minor) variation may be needed, is valid. It should be possible to use placeholders in the item name and/or description, such as:
- „Rent {MMMM} {YYYY}“ for „Rent January 2023“, „Rent February 2023“ etc.
- „Usage compensation {QQ} {YYYY}“ for „Usage compensation Q1 2023“ etc.

But this feature should be raised in a separate ticket.

deepeshgarg007:
@pancho-s There is a subscription feature to manage such scenarios. On orders, it still doesn't make sense to me.
Will be merging if no other objections are there

bosue:
For our usage of invoices without material shipping, the actual subscription feature actually is a viable starting base.

Whether others with more material fulfillment needs do complain – and it seems the auto-repeat functionality has been used [quite a bit](https://github.com/frappe/erpnext/issues?q=is%3Aissue+auto+repeat) – we’ll see in a few months. And if there are valid usecases that don‘t work anymore and aren‘t covered by subscriptions, it may even be brought back one or the other way.

Most importantly we should rename the button from „subscription“ to „auto-repeat“ to avoid confusion. This should be backported as well, IMO.

ankush:
@pancho-s you can still just go to auto-repeat doctype and use it as "subscription"

erpnext/stock/doctype/purchase_receipt/purchase_receipt.js
```diff
@@ -187,12 +187,6 @@ erpnext.stock.PurchaseReceiptController = class PurchaseReceiptController extend
 				}
 				cur_frm.add_custom_button(__('Retention Stock Entry'), this.make_retention_stock_entry, __('Create'));
 
-				if(!this.frm.doc.auto_repeat) {
-					cur_frm.add_custom_button(__('Subscription'), function() {
-						erpnext.utils.make_subscription(me.frm.doc.doctype, me.frm.doc.name)
```
ankush, line 192: Snce we are removing it the util code should also go away? 

https://github.com/frappe/erpnext/blob/caad4537c5107dc5b675e05dcf037142bcfd8c37/erpnext/public/js/utils.js#L264-L276

PR #36263, opened 2023-07-24
fix: allow both custodian and location while creating asset

anandbaburajan:
This https://github.com/frappe/erpnext/pull/36093 recent change in Asset Movement blocked users from setting both custodian and location while creating an asset, so fixed that.


anandbaburajan:
@Mergifyio copy develop

PR #36268, opened 2023-07-24
fix: bank reconciliation rounding preventing clearance

### LINKED ISSUE #35440: v14 Bank Reconciliation - does not clear all when reconciling multiple vouchers with a single Bank Transaction in a single step
### Information about bug

When there are multiple vouchers offered to match a single Bank Transaction, and the sum of the vouchers will clear the Bank Transaction, not all (and it appears only one) voucher is marked cleared, yet both are added to and reconciled against the Bank Transaction.

Screen Shot 2023-05-27 at 15 47 52

All vouchers need to be cleared.

Workaround: Reconcile each voucher against the Bank Transaction separately.

### Module

accounts

### Version

ERPNext >= v14.19

### Installation method

None

Introduced: #34476

casesolved-co-uk:
Fixes https://github.com/frappe/erpnext/issues/35440

version-14-hotfix

### DIFF
--- a/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
+++ b/erpnext/accounts/doctype/bank_transaction/bank_transaction.py
@@ -1,15 +1,30 @@
 # Copyright (c) 2019, Frappe Technologies Pvt. Ltd. and contributors
 # For license information, please see license.txt
 
+import math
+
 import frappe
-from frappe.utils import flt
+from frappe.utils import cint, flt
 
 from erpnext.controllers.status_updater import StatusUpdater
 
 
 class BankTransaction(StatusUpdater):
+	_currency_precision = None
+
+	def flt(self, value):
+		# frappe needs a money datatype
+		if self._currency_precision is None:
+			fraction_units = (
+				frappe.db.get_value("Currency", self.currency, "fraction_units", cache=True) or 100
+			)
+			self._currency_precision = cint(math.log(fraction_units, 10))
+		return flt(value, self._currency_precision)
+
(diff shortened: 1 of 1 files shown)


erpnext/accounts/doctype/bank_transaction/bank_transaction.py
```diff
@@ -1,15 +1,30 @@
...
+
 import frappe
-from frappe.utils import flt
+from frappe.utils import cint, flt
 
 from erpnext.controllers.status_updater import StatusUpdater
 
 
 class BankTransaction(StatusUpdater):
+	_currency_precision = None
```
ankush, line 13: This will be a class variable (?) and kinda behave weirdly in multi-tenant deploys. Should be instance variable instead.

ankush (changes requested):

erpnext/accounts/doctype/bank_transaction/bank_transaction.py
casesolved-co-uk: It will behave as a default and be overridden when initialised in an instance of the class

ankush: @casesolved-co-uk the override will continue to stay there when a request from another site is fired. 

It's basically a bug like this: https://github.com/frappe/frappe/pull/20052



```
worker starts: class attr is None
first request is fired: class attr is set to 2
second request is fired from some other site: class attr still stays 2 and site's preferences aren't respected.
```

casesolved-co-uk:
.

PR #36272, opened 2023-07-24
fix: set new purchase_receipt_amount on asset split

anandbaburajan:
If an asset with a purchase receipt was split, a `Gross Purchase Amount should be equal to purchase amount of one single Asset` error was thrown since the `purchase_receipt_amount` for the new asset wasn't updated. So fixed that.


anandbaburajan:
@mergify copy develop

PR #36303, opened 2023-07-25
fix: root type in account map for balance sheet

GursheenK:
**Problem**
The Balance Sheet creates an Account map for all accounts with a specific root type but the entries pulled from the GL Entry table are not filtered by the root type. No account information is fetched from the accounts map for entries having a different root type.

**Solution**
Added the filter for the root type while fetching the GL entries.


deepeshgarg007 (changes requested):
`set_gl_entries_by_account` is also called from the trial balance report where it needs to fetch all the accounts and not just for a specific root_type.

The `root_type` can be made an optional parameter and only called from the balance sheet
The same changes will also apply to consolidated financial statements

erpnext/accounts/report/financial_statements.py
```diff
@@ -417,13 +418,20 @@ def set_gl_entries_by_account(
...
 
 	accounts_list = frappe.db.get_all(
 		"Account",
-		filters={"company": company, "is_group": 0, "lft": (">=", root_lft), "rgt": ("<=", root_rgt)},
+		filters={
+			"company": company,
+			"is_group": 0,
+			"lft": (">=", root_lft),
+			"rgt": ("<=", root_rgt),
+			"root_type": root_type,
```
deepeshgarg007: Root type should be passed as a parameter to filters only if has a value, passing None will give undesired results

Issue #36291, opened 2023-07-25
Installation  Failed


aadith-sigzen:
### Information about bug

![Screenshot from 2023-07-25 13-29-44](https://github.com/frappe/erpnext/assets/130636662/2db53771-5a71-4b49-9226-efb3ae36c9e1)

### Module

projects

### Version

erpnext 14.31.3
frappe 14.41.0

### Installation method

manual install

### Relevant log output / Stack trace / Full Error Message.
Error: No module named 'holidays'

_No response_

NayanGajjar:
we are having the same error..

Version
erpnext 14.31.3
frappe 14.41.0

![MicrosoftTeams-image (16)](https://github.com/frappe/erpnext/assets/122445476/4570c07c-af5a-4c5c-af7d-9b842c3f304d)

ruthra-kumar:
try `bench setup requirements` on your bench directory. This should install missing packages.

PR #36307, opened 2023-07-25
fix:update import error

aadith-sigzen:
#36291 Installation Failed

issues : holidays  module not found 

Screenshot from 2023-07-25 13-29-44
fix: reference from https://pypi.org/project/holidays/


ruthra-kumar:
`holidays` is part of erpnext dependencies and will be installed by bench.

PR #36308, opened 2023-07-25
feat: "Transaction Type" field for Item Tax Template

leonardamsler:
### **Problem Addressed**:

Currently, there is no way to assign different tax accounts to an item based on whether the item is being sold (Sales Order, Sales Invoice) or purchased (Purchase Order, Purchase Invoice). 

The use of Sales/Purchase Tax Templates is not feasible, as it cannot handle multiple items with different tax rates within a single Sale/Purchase Order or Invoice.

"Tax Categories" also cannot be used to resolve this problem, as the default usage of this feature would not cater to the specific requirements in this context. While it may be possible to create two tax categories ("Sales" and "Purchase"), other categories such as "Inland" or "Outland" sales/purchases could not be superposed.

This feature would be highly beneficial for companies that deal with diverse tax accounts, engage in both purchasing and selling the same item, and handle multiple items in a single Purchase/Sales Order or Invoice.

### **Implementation**:

To address this issue, the following implementation was undertaken:

- A new field called "Transaction Type" was added to the child table of Item Tax Templates (ITT).
- The following three files were modified as part of this implementation:
    1. erpnext/accounts/doctype/item_tax_template_detail/item_tax_template_detail.json:
        - The "Transaction Type" field was added to the Doctype.
    2. erpnext/stock/get_item_details.py:
        - The get_item_tax_map() function was modified to include the "doctype" argument, enabling access to the type of transaction type (e.g., Sales Order, Purchase Order, etc.).
        - The function's logic was updated so that the tax rate is added only if the Transaction Type is appropriate ("All" or the same as the doctype type).
    3. erpnext/public/js/controllers/transaction.js:
        - The get_item_tax_map() function call was updated to pass the "args.doctype" argument.

This implementation has been carefully designed to avoid interfering with the current ERPNext system. By setting the default "Transaction Type" for an entry to "All," the tax rate is applied correctly even it no Transaction Type is selected.

It's worth noting that this issue had been previously raised and discussed in a Pull Request (PR) at https://github.com/frappe/erpnext/pull/24812. However, no changes were made, and the PR is now closed.

The documentation https://docs.erpnext.com/docs/v14/user/manual/en/accounts/item-tax-template could be updated as follows: 

- Replacing the images in the section 2.1 with the new appearance:
Capture d’écran 2023-07-27 à 10 28 13
Capture d’écran 2023-07-27 à 10 30 33

- Adding a new section 2.2 with name "Having multiple accounts". In that section write the following:
   1. Cut and past the second black quote of the 2.1 section
   
   2. Add a second paragraph: 
New paragraph 2 2

   3. Display an image of multiple accounts: 
Capture d’écran 2023-07-27 à 10 40 23

## Example of an Item Tax Template:
Capture d’écran 2023-07-25 à 16 06 07


deepeshgarg007:
@leonardamsler can you please add a basic test case for this

erpnext/public/js/controllers/transaction.js
```diff
@@ -1854,7 +1854,10 @@ erpnext.TransactionController = class TransactionController extends erpnext.taxe
 				args: {
 					company: me.frm.doc.company,
 					item_tax_template: item.item_tax_template,
-					as_json: true
+					as_json: true,
+					args: {
+						doctype: doc.doctype
+					}
```
ankush: args inside args? Why not just move it out.

leonardamsler: It has been changed now

leonardamsler:
Here is a demo about this feature and how it works.

https://github.com/frappe/erpnext/assets/93430036/617de728-190a-40c3-b975-1e324d468784

I just forgot to mention at 3' that items could be sold with different rates.
Also, having the default value "All" for all lines is equivalent to the current usage of Item Tax Templates.

deepeshgarg007:
@leonardamsler Can you add a basic unit test for this?

deepeshgarg007:
@leonardamsler I am closing this PR, please reopen once you add tests

cgallay:
@deepeshgarg007 I am taking over the work from @leonardamsler, I added some unit test. Could you please review it and let me know if that meets your criteria? I feel like my test is quite long since I had to create a company and all other DocTypes on which the test depends. But maybe there is an easier way to create those dependencies. Let me know.

leonardamsler:
@deepeshgarg007 It seems I do not have the rights to reopen the PR. Charles has pushed the tests on our git branch, so we want to reopen it. Let me know when I can do so.

deepeshgarg007:
@leonardamsler I have raised a new PR at https://github.com/frappe/erpnext/pull/37177. Maintainer edits were not allowed on this PR

ankush:
@leonardamsler if possible raise future PRs from a personal account instead of the company's fork. Github doesn't allow maintainer edits on org forks. 


updated https://github.com/frappe/erpnext/wiki/Pull-Request-Checklist#maintainer-edits

PR #36310, opened 2023-07-25
fix: group by in fixed asset register

anandbaburajan:
The group by feature in the Fixed Asset Register report wasn't implemented properly, so fixed that.


anandbaburajan:
@mergify copy develop

anandbaburajan:
@mergify copy version-13-hotfix

